PR #991, opened 2016-06-08
don't assume groups are non-empty

gustafsson:
grouping on an empty data frame should not crash. `combine` can't create additional columns but returning an empty data frame is better than crashing.

Similarly, showing an GroupedDataFrame in the REPL should not assume that it's non-empty.

### DIFF
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -36,4 +36,6 @@ module TestGrouping
     x = pool(collect(1:20))
     df = DataFrame(v1=x, v2=x)
     groupby(df, [:v1, :v2])
+
+    by(e->1, DataFrame(x=[]), :x)
 end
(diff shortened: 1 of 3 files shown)


test/grouping.jl
```diff
@@ -36,4 +36,6 @@ module TestGrouping
     x = pool(collect(1:20))
     df = DataFrame(v1=x, v2=x)
     groupby(df, [:v1, :v2])
+
+    by(e->1, DataFrame(x=[]), :x)
```
nalimilan: Could you check the return value too? Could be good to test a function which actually uses the data (when present), like `sum`.

gustafsson:
@nalimilan I implemented your comment, tests passed on `0.4` but failed on `latest` so I rebased. Appveyor still fails on RDA but I can't see how that's related to this PR.

nalimilan:
Thanks!

PR #993, opened 2016-06-10
Fix InexactError in join() due to too many levels

nalimilan:
PooledDataVecs() uses the smallest possible integer size, but
the DataFrames code stores higher references than that in the vector.
Always use UInt32, which will be enough for any reasonable number of
levels.

Maybe PooledDataVecs() should not compact its result by default
(which makes it type-unstable), but this should be considered as part
of a larger rework.

Fixes the second issue at https://github.com/JuliaStats/DataFrames.jl/issues/960.


nalimilan:
Failure on 0.5 is due to the missing `varzm`, and then it will be because of https://github.com/JuliaStats/DataArrays.jl/pull/198.

Issue #760, opened 2015-01-12
Don't show entire 1M-column DataFrame by default


garborg:

johnmyleswhite:
Is the number of rows a problem? Or a number of columns? I thought we only showed as many rows as would, in some sense, fit into the REPL window's height.

garborg:
```julia
d = Dict{Symbol,Vector}()
for _ = 1:2_000_000
    d[symbol(randstring(8))] = [1]
end
df = DataFrame(d)
```
prints to the REPL for minutes.

johnmyleswhite:
So it's the number of columns that matters, not the number of rows. We could follow dplyr's lead and only print the number of columns that fits in the screen horizontally. Alternatively, we could just cap things at a constant like 25.

garborg:
Oh yes, columns -- your request for clarification went right over my head. Either of those option, or something like two screen widths, seem fine to me.

johnmyleswhite:
Will try to look into this soon. We may need a heuristic for deciding whether we even calculate the screen width contribution of all columns since that could be slow on its own.

garborg:
Makes sense.

johnmyleswhite:
And thanks for closing the old issues. You must have the highest close rate of all DataFrames contributors by far now.

garborg:
Thanks. I'm a bit surprised there are still 49 watchers.

johnmyleswhite:
I suspect the watch count never decreases. (Except in response to this comment.)

bkamins:
I get this problem very often when working with real data.

I would recommend adding an additional argument to `show` and `showrows` that could be called for example `maxchunks` that would limit the number of printed chunks with the following rules:

* `maxchunks` less than `1`: print all chunks
* `maxchunks` equal to `1`: print one chunk (this should be default so that we get only one screen)
* `maxchunks` greater than `1`: print given number of chunks

If number of printed chunks would be limited by this argument an information that not all columns of data frame were printed should be given.

johnmyleswhite:
I suspect we should use the number of characters rather than a number of chunks since it's more immediately interpretable -- unless you already mean chunks to indicate number of characters.

bkamins:
I was thinking of a parameter that would allow to limit `nchunks` in `showrows`
https://github.com/JuliaStats/DataFrames.jl/blob/master/src/abstractdataframe/show.jl#L344
as this would be a simple modification of existing code and currently single chunk is already sized to fit the screen.

[...]

PR #995, opened 2016-06-12
Improved show for DataFrames

bkamins:
A proposal to solve #760.

Summary of changes:
- `show` and `showall` get a new argument `onechunk` which limits the number of printed chunks to 1 if `splitchunks` is `true` (with appropriate message in the summary)
- `showcols` gets two parameters `allcols` (if all columns of an `AbstractDataFrame` should be printed or only those fitting on the screen) and `values` (if sample of values of an `AbstractDataFrame` should be printed)
- `showcompact` is defined for `AbstractDataFrame` and `GroupedDataFrame`


nalimilan:
Thanks. These positional arguments are really getting out of hand. We should probably get rid of `splitchunks` (whose name is really confusing) and tell people to use `showcols` instead. Then `onechunk` could be renamed to `allcols`. What do you think?

I don't think `showcompact` is intended for this kind of use: as the docs say, it's mainly for scalar values to provide a short representation without type information to be used inside arrays.

src/abstractdataframe/show.jl
```diff
@@ -556,32 +574,77 @@ end
...
+#'        each column is printed
 #'
 #' @returns o::Void A `nothing` value.
 #'
 #' @examples
 #'
 #' df = DataFrame(A = 1:3, B = ["x", "y", "z"])
-#' showcols(df, true)
-function showcols(io::IO, df::AbstractDataFrame) # -> Void
+#' showcols(STDOUT, df)
```
nalimilan: The example was correct, `STDOUT` is implicit.

src/abstractdataframe/show.jl
```diff
@@ -556,32 +574,77 @@ end
...
+function showcols(io::IO, df::AbstractDataFrame, allcols::Bool = false, values::Bool = true) # -> Void
     println(io, summary(df))
     metadata = DataFrame(Name = _names(df),
                          Eltype = eltypes(df),
                          Missing = colmissing(df))
-    showall(io, metadata, true, @compat(Symbol("Col #")), false)
+    nrows, ncols = size(df)
+    if values && nrows > 0
+        if nrows == 1
+            metadata[:Values] = [Symbol(df[1, i]) for i in 1:ncols]
```
nalimilan: Why a symbol? I think the best strategy is to use `sprint(showcompact, df[1, i])` (same below). In the tests, make sure to add cases with `String` and `Float32` columns, which are corner cases (the first needs quotes to be included inside the string, the other needs the `f0` prefix to be removed due to compact printing).

src/abstractdataframe/show.jl
```diff
@@ -556,32 +574,77 @@ end
...
+#' @exported
 #' @description
 #'
 #' Render a summary of the column names, column types and column missingness
 #' count.
 #'
 #' @param io::IO The `io` to be rendered to.
 #' @param df::AbstractDataFrame An AbstractDataFrame.
+#' @param allcols::Bool If `false` (default) only a subset of columns
+#'        fitting on the screen is printed
```
nalimilan: MIssing ending dot here and in other docstrings. Could also use a comma after "(default)" here and elsewhere.

bkamins:
@nalimilan Thx for the comments. I will correct the PR.

For the record let me add that this kind of output also should be fixed:
```
df = DataFrame(x=["a", "\t", "\\", "\n", "\$", "z"], y=1:6)
6×2 DataFrames.DataFrame
│ Row │ x   │ y │
├─────┼─────┼───┤
│ 1   │ "a" │ 1 │
│ 2   │ "\t"  │ 2 │
│ 3   │ "\\" │ 3 │
│ 4   │ "\n"  │ 4 │
│ 5   │ "\$" │ 5 │
│ 6   │ "z" │ 6 │
```

src/abstractdataframe/show.jl
bkamins: I thould it should stay as the example with implicit `STDOUT` is given below for definition of `function showcols(df::AbstractDataFrame, allcols::Bool=false, values::Bool=true)`.

nalimilan: OK. Anyway until we turn these into real docstrings this is pretty abstract.

src/abstractdataframe/show.jl
bkamins: `sprint` would make `:Values` column a string and thus it would be rendered with additional enclosing double quotes when printed and symbol is printed "as is". Your suggestion is valuable though :) and I have modified the code.

nalimilan: OK. But still use it before converting to symbol, to be sure you get the compact form. There have been many issues with this with the recent changes in 0.5.

bkamins:
@nalimilan I hope I have covered all your comments correctly.

I have left `splitchunks` internally to differentiate the behavior of `show` and `showall` (similarly to arrays).

Additionally I have changed the formula calculating the width of the string so that `DataFrames` render correctly with escaped strings.

quinnj:
I think this would be a great fix; care to rebase?

bkamins:
Sure - I thought it was rejected.

quinnj:
Sorry, the package has struggled w/ maintenance over the last year or so as development moved over to DataTables.jl, but all that work has now been backported here and all development will now resume here.

nalimilan:
Sorry, I think I didn't finish the review because I wanted to get a complete understanding of the design space, and I didn't find the time for that at that point.

bkamins:
I have started to update the PR and have one issue to clarify.
There is a change how DataFrames performs `show` between latest release and master.

Consider running the following line in REPL:
```
df = DataFrame(A = 1:4,
               B = ["x\"", "y\n", "z\$", "ABC"],
               C = Float32[1.0, 2.0, 3.0, 4.0],
               D = Symbol[:ABC,Symbol("x\""),Symbol("y\n"),Symbol("z\$")])
```
on master it shows:
```
4×4 DataFrames.DataFrame
│ Row │ A │ B   │ C   │ D   │
├─────┼───┼─────┼─────┼─────┤
│ 1   │ 1 │ x"  │ 1.0 │ ABC │
│ 2   │ 2 │ y
   │ 2.0 │ x"  │
│ 3   │ 3 │ z$  │ 3.0 │ y
  │
│ 4   │ 4 │ ABC │ 4.0 │ z$  │
```
and latest release it shows:
```
4×4 DataFrames.DataFrame
│ Row │ A │ B     │ C   │ D   │
├─────┼───┼───────┼─────┼─────┤
│ 1   │ 1 │ "x\""  │ 1.0 │ ABC │
│ 2   │ 2 │ "y\n"   │ 2.0 │ x"  │
│ 3   │ 3 │ "z\$"  │ 3.0 │ y
  │
│ 4   │ 4 │ "ABC" │ 4.0 │ z$  │
```
and if we cast `df` to an array we get yet another output (it could be reproduced by `show` of DataFrame with column names and pipes added - here I want to concentrate on how the field values are printed):
```
julia> convert(Matrix, df)
4×4 Array{Any,2}:
 1  "x\""  1.0  :ABC
 2  "y\n"  2.0  Symbol("x\"")
 3  "z\$"  3.0  Symbol("y\n")
 4  "ABC"  4.0  Symbol("z$")
```
Which is the preferred target printing style? Or maybe yet some other option?
I personally would feel comfortable with the third one (show values like `show` for arrays) as it would be consistent.

quinnj:
I definitely think we should be consistent w/ array printing (last example). Show strings as quoted strings, as well as symbol that way. I think that's the only real solution if we want to be able to show unicode + control characters/whitespace and maintain the correct column widths.

nalimilan:
Agreed, the `Array` output looks like a good reference. Though printing quotes around strings is a bit verbose, and we could get rid of it if we printed the column eltype in a header (like [tibbles](https://github.com/tidyverse/tibble/blob/master/README.md) in R).

bkamins:
Thank you for the comments. Regarding refactoring of show I have some additional thoughts:
* if we omit `"` in strings how do we visually distinguish `"NA"` string from true `NA` (in R this is a problem with `"<NA>"`)
* how do we handle columns of custom types (normal and `Nullable` - whatever they will be eventually called) - in particular when their representation is very long (should some truncation be applied); in particular current `show` has problem for calculation of width of such structures, e.g.:
```
julia>     df = DataFrame(A=[[1:25;],"sdf"])
2×1 DataFrames.DataFrame
│ Row │ A                                                                                           │
├─────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ 1   │ [1, 2, 3, 4, 5, 6, 7, 8, 9, 10  …  16, 17, 18, 19, 20, 21, 22, 23, 24, 25] │
│ 2   │ "sdf"                                                                                       │
```

In short - the problem is complex so I will try to give it some thought and will open a separate issue and write down my recommendation. I will strip this PR from field width calculation changes and leave only improved display features.

nalimilan:
> if we omit " in strings how do we visually distinguish "NA" string from true NA (in R this is a problem with "<NA>")

Yes, that would be a problem unless we add a header with the eltype of each column. But we can keep the quotes for now and discuss that possibility later.

>    how do we handle columns of custom types (normal and Nullable - whatever they will be eventually called) - in particular when their representation is very long (should some truncation be applied); in particular current show has problem for calculation of width of such structures, e.g.:

Some truncation should probably be applied (I think you can do that by setting a property on `IOContext` now). But that can also be improved later, no need to fix everything in a single PR. It's not that common to have fields like that in DataFrames anyway.

nalimilan:
@bkamins I think you need to rebase on the latest master. If there are still unrelated commits, use `git rebase -i master` and remove them.

bkamins:
Agreed - that is why in this PR I have left only changes to `show` global behavior and left other changes for the future.

One question regarding git: I have made a merge not a rebase and now I can see the bad consequences that all the intermediate commits got included. Is there any simple and safe way to fix it?

nalimilan:
Hmm... I guess the easiest solution would be to start a new branch from master, cherry-pick your commits into it, and then force push to this branch using `git push --force bkamins :newshow`.

bkamins:
Just as a comment: I believe that the build failed on Julia latest is unrelated to this PR.

src/abstractdataframe/show.jl
```diff
@@ -543,32 +549,68 @@ end
...
 
-showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
+#' @exported
+#' @description
+#'
+#' Render a summary of the column names, column types and column missingness
+#' count.
+#'
+#' @param df::AbstractDataFrame An AbstractDataFrame.
+#' @param allcols::Bool If `false` (default), only a subset of columns
```
nalimilan: Maybe just call this `all`, since "col" is already clear from the function's name?

test/show.jl
```diff
@@ -1,19 +1,40 @@
...
+    df = DataFrame(A = 1:4, B = ["x\"", "∀ε⫺0: x+ε⫺x", "z\$", "AB\nC"], C = Float32[1.0, 2.0, 3.0, 4.0])
+    srand(1)
+    df_big = DataFrame(rand(50,50))
 
     io = IOBuffer()
     show(io, df)
     show(io, df, true)
     showall(io, df)
-    showall(io, df, true)
+    showall(io, df, false)
```
nalimilan: Could you test the actual output? You can use triple-quoted strings for that.

bkamins:
@nalimilan Inline comments got removed so I reply here:
* changed `Symbol("Row")` to `:Row`
* `@param values` comment string corrected
* all lines are below 92 chars
* I have removed `Symbol` in `:Values` formatting, but if we change the way `DataFrame` columns containing strings are printed it might have to be revised
* `allcols` changed to `all` (but one has to remember it clashes with `all` function)
* added testing of actual output in tests

src/abstractdataframe/show.jl
```diff
@@ -297,6 +297,8 @@ end
 #'        required to render each column.
 #' @param splitchunks::Bool Should the printing of the AbstractDataFrame
 #'        be done in chunks? Defaults to `false`.
+#' @param allcols::Bool Should only one chunk be printed if printing in
+#'        chunks? Defaults to `false`.
```
nalimilan: Defaults to `false`.

src/abstractdataframe/show.jl
```diff
@@ -315,21 +317,30 @@ function showrows(io::IO,
...
     if isempty(rowindices1)
+        if displaysummary
+            println(io, summary(df))
+        end
         return
     end
 
     rowmaxwidth = maxwidths[ncols + 1]
     chunkbounds = getchunkbounds(maxwidths, splitchunks, displaysize(io)[2])
     nchunks = length(chunkbounds) - 1
```
nalimilan: Would be clearer to do `nchunks = allcols ? length(chunkbounds) - 1 : min(nchunks, 1)`.

src/abstractdataframe/show.jl
```diff
@@ -543,32 +549,70 @@ end
...
+function showcols(io::IO, df::AbstractDataFrame, all::Bool = false,
+                  values::Bool = true) # -> Void
     println(io, summary(df))
     metadata = DataFrame(Name = _names(df),
                          Eltype = eltypes(df),
                          Missing = colmissing(df))
-    showall(io, metadata, true, Symbol("Col #"), false)
+    nrows, ncols = size(df)
+    if values && nrows > 0
+        # type of Values column is now String; it might need to be changed
```
nalimilan: I don't think this comment is needed: tests will (or should) catch this and people will figure out what needs to be changed anyway.

src/abstractdataframe/show.jl
nalimilan: This still applies, right?

src/abstractdataframe/show.jl
```diff
@@ -543,32 +549,70 @@ end
...
+        # if the way strings are printed in data frames changes
+        if nrows == 1
+            metadata[:Values] = [sprint(showcompact, df[1, i]) for i in 1:ncols]
```
nalimilan: Also, what about using `ourshowcompact`?

test/show.jl
```diff
@@ -1,19 +1,103 @@
...
+    df = DataFrame(A = 1:4, B = ["x\"", "∀ε⫺0: x+ε⫺x", "z\$", "ABC"],
+                   C = Float32[1.0, 2.0, 3.0, 4.0])
+    srand(1)
+    df_big = DataFrame(rand(50,50))
+
+    refstr = """
+    4×3 DataFrames.DataFrame
+    │ Row │ A │ B             │ C   │
+    ├─────┼───┼───────────────┼─────┤
+    │ 1   │ 1 │ x\"            │ 1.0 │
```
nalimilan: I suppose the fact that vertical lines are not aligned is a bug elsewhere? Then better leave a `TODO` somewhere to make it clear.

test/show.jl
```diff
@@ -1,19 +1,103 @@
...
+        str = String(take!(io))
+        @test str == refstr
+    end
 
     io = IOBuffer()
-    show(io, df)
-    show(io, df, true)
-    showall(io, df)
-    showall(io, df, true)
+    show(io, df_big)
```
nalimilan: Could you also test the output of these functions (even if that's verbose, it's fine)? Else the case when there are too many rows won't be covered. You should probably pass a custom `IOContext` to control the size of the display. Also better define `df_big` here rather than above, where it isn't used.

test/show.jl
```diff
@@ -1,19 +1,103 @@
 module TestShow
     using Base.Test, DataFrames
 
-    df = DataFrame(A = 1:3, B = ["x", "y", "z"])
+    # In the future newline characte \n should be added to this test case
+    df = DataFrame(A = 1:4, B = ["x\"", "∀ε⫺0: x+ε⫺x", "z\$", "ABC"],
+                   C = Float32[1.0, 2.0, 3.0, 4.0])
```
nalimilan: Could you add a `null` value somewhere so that this is covered (unless it's done elsewhere already)?

test/show.jl
bkamins: They are aligned when the string is printed, but `"` needs to be escaped in string literal which breaks alignment in the code.

test/show.jl
bkamins: It is already covered I believe in line:
```
df = DataFrame(Fish = ["Suzy", "Amir"], Mass = [1.5, null])
```
at the end of the file

test/show.jl
nalimilan: Ah, of course!

test/show.jl
nalimilan: Yes, but `showcols` isn't tested there. Would be worth adding a test.

test/show.jl
bkamins: Actually there is a problem - in the next line `│ 2   │ 2 │ ∀ε⫺0: x+ε⫺x │ 2.0 │` which is not aligned properly and it is a TODO do be added for `getmaxwidths` function. Sorry for confusion

test/show.jl
bkamins: added `showcols` test

bkamins:
@nalimilan I hope I have managed to clean up everything.

src/abstractdataframe/show.jl
bkamins: changed to `true`

src/abstractdataframe/show.jl
bkamins: changed (with a bit different code as `nchunks` is undefined before this line)

src/abstractdataframe/show.jl
bkamins: removed

src/abstractdataframe/show.jl
bkamins: Changed - but it creates problems in corner cases (described in TODO for `getmaxwidths`)

test/show.jl
```diff
@@ -41,27 +39,181 @@ module TestShow
     refstr = """
     4×3 DataFrames.DataFrame
```
nalimilan: Just a detail, but we probably don't need an empty line? That would be more consistent with the other format.

test/show.jl
```diff
@@ -41,27 +39,181 @@ module TestShow
...
+    │ 25  │ 0.751313 │ 0.884837 │ 0.550334 │
+    
+    │ Row │ x4       │ x5       │
+    ├─────┼──────────┼──────────┤
+    │ 1   │ 0.580782 │ 0.138763 │
+    ⋮
+    │ 24  │ 0.762276 │ 0.755415 │
+    │ 25  │ 0.339081 │ 0.649056 │"""
+
+    io = IOContext(IOBuffer(), :displaysize=>(10,40))
```
nalimilan: Maybe set the number of rows to a lower value in order to have smaller test and check what happens when not all rows can be shown in a single page? Can also be done in a later PR if you prefer.

test/show.jl
bkamins: removed

test/show.jl
bkamins: I believe this is what I check here. I assume 10 rows and 40 columns. And you can see the difference between `show` and `showall`. 
`show` limits the output to fit page height and `showall` does not do that.
They also differ in how they handle *wide* data (not fitting the screen vertically) and we set `allcols` to `true`: `show` does paging and `showall` prints full table ignoring `:displaysize` (which could useful, when e.g. we want to dump `DataFrame` show result to a file).

bkamins:
@cjprybol fixed the issue with tests on 32 bit machine.

nalimilan:
Thanks! Merging since Travis doesn't seem to be willing to run on Mac...

PR #997, opened 2016-06-14
ENH: Add sample function for DataFrames

scls19fr:
Might fix https://github.com/JuliaStats/StatsBase.jl/issues/170

### DIFF
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -70,6 +70,7 @@ export @~,
        readtable,
        rename!,
        rename,
+       sample,
        showcols,
        stack,
        stackdf,
@@ -111,7 +112,9 @@ for (dir, filename) in [
         ("statsmodels", "statsmodel.jl"),
 
         ("", "RDA.jl"),
-        ("", "deprecated.jl")
+        ("", "deprecated.jl"),
+
+        ("other", "sample.jl"),
     ]
 
     include(joinpath(dir, filename))
--- a/src/other/sample.jl
+++ b/src/other/sample.jl
@@ -18,6 +18,7 @@ julia> srand(1)
 julia> sample(iris, 5)
 5×5 DataFrames.DataFrame
 │ Row │ SepalLength │ SepalWidth │ PetalLength │ PetalWidth │ Species      │
+|-----|-------------|------------|-------------|------------|--------------|
 │ 1   │ 5.0         │ 2.0        │ 3.5         │ 1.0        │ "versicolor" │
 │ 2   │ 6.2         │ 2.9        │ 4.3         │ 1.3        │ "versicolor" │
 │ 3   │ 6.7         │ 3.1        │ 4.7         │ 1.5        │ "versicolor" │
--- a/src/other/sample.jl
+++ b/src/other/sample.jl
@@ -1,0 +1,30 @@
+import StatsBase: sample
+
+"""
+    sample(df[, n; replace=true, ordered=false])
+
+Draw a random sample of `n` rows from a data frame `df` and return the result as a data frame.
+
+# Arguments
+
+- `replace::Bool=true`: Should sampling be performed with replacement?
+- `ordered::Bool=false`: Should an ordered sample be taken?
+
+# Example
+```
+julia> using RDatasets
+julia> iris = dataset("datasets", "iris")
+julia> srand(1)
+julia> sample(iris, 5)
+5×5 DataFrames.DataFrame
+│ Row │ SepalLength │ SepalWidth │ PetalLength │ PetalWidth │ Species      │
+│ 1   │ 5.0         │ 2.0        │ 3.5         │ 1.0        │ "versicolor" │
+│ 2   │ 6.2         │ 2.9        │ 4.3         │ 1.3        │ "versicolor" │
+│ 3   │ 6.7         │ 3.1        │ 4.7         │ 1.5        │ "versicolor" │
+│ 4   │ 5.5         │ 2.3        │ 4.0         │ 1.3        │ "versicolor" │
+│ 5   │ 5.8         │ 2.7        │ 5.1         │ 1.9        │ "virginica"  │
+```
+"""
+function sample(df::AbstractDataFrame, n::Integer=1; replace::Bool=true, ordered::Bool=false)
+    df[sample(1:size(df, 1), n, replace=replace, ordered=ordered), :]
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -27,7 +27,8 @@ my_tests = ["utils.jl",
             "iteration.jl",
             "duplicates.jl",
             "show.jl",
-            "statsmodel.jl"]
+            "statsmodel.jl",
+            "sample.jl"]
 
 println("Running tests:")
 
--- a/test/sample.jl
+++ b/test/sample.jl
@@ -1,0 +1,10 @@
+module TestSample
+    using Base.Test
+    using DataFrames
+    
+    df = DataFrame(A=1:10,B=11:20)
+    srand(1)
+    df_sample = sample(df, 5)
+    @test df_sample[:A] == [1,8,7,4,2]
+    @test df_sample[:B] == [11,18,17,14,12]
+end


src/other/sample.jl
```diff
@@ -0,0 +1,28 @@
+import StatsBase: sample
+
+function sample(df::DataFrame)
+    sample(df, 1)
+end
+
+"""
+    sample(df[, N])
+
+Returns a (random) sample of a DataFrame
```
diegozea: I prefer something like *"Returns a (random) sample of rows from a DataFrame"* to be explicit about the selección of rows.

src/other/sample.jl
```diff
@@ -0,0 +1,28 @@
...
+5×5 DataFrames.DataFrame
+│ Row │ SepalLength │ SepalWidth │ PetalLength │ PetalWidth │ Species      │
+│ 1   │ 5.6         │ 2.9        │ 3.6         │ 1.3        │ "versicolor" │
+│ 2   │ 6.4         │ 3.1        │ 5.5         │ 1.8        │ "virginica"  │
+│ 3   │ 5.1         │ 3.8        │ 1.6         │ 0.2        │ "setosa"     │
+│ 4   │ 5.5         │ 2.4        │ 3.7         │ 1.0        │ "versicolor" │
+│ 5   │ 5.1         │ 3.8        │ 1.6         │ 0.2        │ "setosa"     │
+```
+"""
+function sample(df::DataFrame, N::Integer)
```
diegozea: The kw arg `replace` is missing.

Would be great to add to new kw args: `stratum` and `orderby` to allow [stratified](https://en.wikipedia.org/wiki/Stratified_sampling) and [systematic](https://en.wikipedia.org/wiki/Systematic_sampling) sampling respectively. That would be easy to implement using the `by` and `sort` functions i.e.:

```julia
function sample(df::DataFrame, N::Integer; stratum=nothing, orderby=...)
...
if stratum !== nothing
    by(df, stratum) do subset
        if sort !== nothing
            sorted = sort(subset, cols=orderby)
            ...sample...
...
``` 
If that's a complicated API, two new functions `systematic_sample` and `stratified_sample` can be added instead of the kw args.

diegozea:
Thanks for doing this @scls19fr ! :) Test are missing, It would be great to have some tests for this functions.

src/other/sample.jl
nalimilan: +1, but say "of `N` rows".

src/other/sample.jl
```diff
@@ -0,0 +1,28 @@
...
+"""
+    sample(df[, N])
+
+Returns a (random) sample of a DataFrame
+
+# Example
+```
+julia> using RDatasets
+julia> iris = dataset("datasets", "iris")
+julia> sample(iris, 5)
```
nalimilan: Please add a call to `srand(1)` to ensure the results are always the same.

src/other/sample.jl
nalimilan: That sounds interesting (and would better justify adding this function which is otherwise quite trivial). But I don't see the point of the `orderby` argument: just sort the data frame manually. OTOH, a useful extension would be to be able to pass several variables as strata, and to give a number `N` or a sampling probability for each. Would you make a PR?

Regarding systematic sampling, do you think is really useful when you have access to all of the data like here? Anyway, it should be implemented for simple vectors in StatsBase first.

nalimilan: @scls19fr Could you also add the `replace::Bool=true` and `ordered::Bool=false` keyword arguments?

nalimilan: Ah, and the signature should use `AbstractDataFrame`.

src/DataFrames.jl
```diff
@@ -111,7 +111,9 @@ for (dir, filename) in [
         ("statsmodels", "statsmodel.jl"),
 
         ("", "RDA.jl"),
-        ("", "deprecated.jl")
+        ("", "deprecated.jl"),
+
+        ("other", "sample.jl"),
```
nalimilan: Please add this to "utils.jl", no need for a separate file.

nalimilan:
Thanks. Please add tests for the new feature, and add the function to the list of exports.

scls19fr:
I haven't add sample to `utils.jl` but to `sample.jl` because `AbstractDataFrame` must be defined before using them

src/other/sample.jl
```diff
@@ -0,0 +1,29 @@
+import StatsBase: sample
+
+function sample(df::AbstractDataFrame; replace::Bool=true, ordered::Bool=false)
```
nalimilan: Just do `N::Integer=1` in the definition below, and you'll get this one for free.

src/other/sample.jl
```diff
@@ -0,0 +1,29 @@
...
+    sample(df, 1, replace=replace, ordered=ordered)
+end
+
+"""
+    sample(df[, N])
+
+Returns a (random) sample of N rows from a DataFrame
```
nalimilan: Backquotes around `N`, which should actually be lower-case `n` for consistency with StatsBase.

Also, at https://github.com/JuliaStats/StatsBase.jl/pull/172, the wording is "Draw a random sample of elements from an array `a`. So probably better be consistent and say "Draw a random sample of `N` rows from a data frame `df` and return the result as a data frame."

Finally, you need to document the keyword arguments:
```
# Arguments

- `replace::Bool=true`: Should sampling be performed with replacement?
 • `ordered::Bool=false`: Should an ordered sample be taken?
```

src/other/sample.jl
```diff
@@ -0,0 +1,30 @@
+import StatsBase: sample
+
+"""
+    sample(df[, n])
+
+Draw a random sample of `n` rows from a data frame `df` and return the result as a data frame
```
nalimilan: Missing ending dot.

src/other/sample.jl
```diff
@@ -0,0 +1,30 @@
+import StatsBase: sample
+
+"""
+    sample(df[, n])
```
nalimilan: Need to add keyword arguments after a `;`.

nalimilan:
Any opinions about the opportunity of adding this function?

src/other/sample.jl
```diff
@@ -0,0 +1,31 @@
...
+│ Row │ SepalLength │ SepalWidth │ PetalLength │ PetalWidth │ Species      │
+|-----|-------------|------------|-------------|------------|--------------|
+│ 1   │ 5.0         │ 2.0        │ 3.5         │ 1.0        │ "versicolor" │
+│ 2   │ 6.2         │ 2.9        │ 4.3         │ 1.3        │ "versicolor" │
+│ 3   │ 6.7         │ 3.1        │ 4.7         │ 1.5        │ "versicolor" │
+│ 4   │ 5.5         │ 2.3        │ 4.0         │ 1.3        │ "versicolor" │
+│ 5   │ 5.8         │ 2.7        │ 5.1         │ 1.9        │ "virginica"  │
+```
+"""
+function sample(df::AbstractDataFrame, n::Integer=1; replace::Bool=true, ordered::Bool=false)
```
ararslan: Should extend the function from StatsBase, i.e. `function StatsBase.sample(...)`

quinnj:
I think this could be useful, but needs a rebase.

bkamins:
I am closing this since we removed StatsBase.jl dependency in DataFrames.jl.

Given the relation of cost of materializing a `DataFrame` vs. cost of sampling I think it is easy enough to use `df[sample(something selecting rows), :]` even if this is suboptimal performance wise (the difference will not be noticeable).

Please reopen if you disagree.

nalimilan:
It would still be kind of nice to be able to do `sample(df, n)` at some point. Maybe we could define `sample` in DataAPI, just like we do for `describe`.

bkamins:
Exactly, `DataFrame` is not specific here and we will delegate all the work to StatsBase.jl `sample` anyway.

anandijain:
What about for groupby objects?

`groups = groupby(df, :A)`
I think it would be nice to be able to `sample(groups, 3)` and get three of the groups.

bkamins:
There are two issues here:

1. The question is we want `GroupedDataFrame` to be a subtype of `AbstractVector`.
@nalimilan - what is your current status of thinking about it?

2. In general such sampling is not allowed as it would possibly produce duplicates in `sample(groups, 3)` and `GroupedDataFrame` does not allow for duplicate groups (and I think it should not allow for this).

In short to sample subgroups now you should write `sample(collect(groups), 3)` and probably this will stay this way.

nalimilan:
IIRC we concluded that subtyping `AbstractVector` wasn't a good idea at least at this point. And indeed you're right that duplicating groups would be problematic.

bkamins:
Thank you. I just wanted to make sure with `AbstractVector` status (and I agree).

PR #998, opened 2016-06-14
remove linebreak that made README render badly

KristofferC:


nalimilan:
There's the same problem with the line above ("with the").

nalimilan:
BTW, could you add a line break between the stable and devel documentation? That would make it much clearer.

KristofferC:
Could add doc badges like in https://github.com/JuliaDocs/Documenter.jl

KristofferC:
made a suggestion how it could look

nalimilan:
Thanks!

PR #1001, opened 2016-06-25
Setting Dataframe from dataframe

mdavezac:
Previously, when setting a dataframe from a dataframe, the following would fail

~~~Julia
df = DataFrame(A = 1:3, B = [:a, :b, :c], C=["a", 1, "2"])
other = DataFrame(B = [:A, :B, :C], A = 4:6, D=5:7)
df[2, [:A, :B]] = other
@test df[:A] == [1, 4, 3]
@test df[:B] == [:a, :A, :c]
@test df[:C] == ["a", 1, "2"]
~~~

because the columns in `other` where chosen depending on their integer index in `df`. So the indexing would fail whenever the order of the columns did not match in setter and settee.
This pull request allows indexing the columns in the setter and settee using the same type of index  in both.

It also adds a number of tests for indexing (when there were none before). Its incomplete, but something is better than nothing?


nalimilan:
Did you intend to close this?

mdavezac:
Yes, I belatedly realised it's a feature, not a bug. The settee dataframe is considered a matrix, rather than a dataframe. It's coherent with how row indices work.

PR #1004, opened 2016-06-28
Remove "stagedfunction" from reserved words

nalimilan:
It only existed for a short period in the 0.4 development cycle. This
made the test fail on 0.5.


nalimilan:
Failure on 32-bit Windows is in the RDA test.

Issue #955, opened 2016-05-12
Can we remove src/RDA.jl?


dmbates:
This code to read a subset of the `.rda` (R data archive) format was used to import saved R `data.frame` objects.  The `RCall` package is much better at this.  I suggest deprecating `read_rda` while doing the conversion for julia v0.5  I believe this will also remove the dependency on the `Gzip`  package.

nalimilan:
I was wondering about this too. +1

garborg:
I'm for this, too.

The only downside I see is that in [RDatasets.jl](https://github.com/johnmyleswhite/RDatasets.jl), DataFrames with legitimate categorical columns (not just because strings were often pooled in R) are stored as .rda files.

Maybe move the file into RDatasets.jl in the short term and move those files to feather or .jld longer term? c.c. @johnmyleswhite

nalimilan:
At the very least, we could move `read_rda` to RDatasets waiting for a larger refactoring.

johnmyleswhite:
+1

nalimilan:
See also https://github.com/JuliaStats/DataFrames.jl/issues/958, which moving to RCall.jl would likely fix.

ararslan:
@alyst What's the state of your RData package? Is it such that it could remove the RDA stuff in here entirely and direct users to your package?

For the others, see https://github.com/alyst/RData.jl

alyst:
@ararslan I'm using it regularly and it works for me. It's an extended version of `DataFrames.read_rda()` that supports all R data types (so it does not crash on an arbitrary RData file, although not every R type is converted to Julia counterpart). So, yes, I think you can recommend it to the users. I suspect the first users might run into issues, because `RData` received limited testing so far, but I will try to fix the bugs as they are discovered.

nalimilan:
Great. So let's deprecate RDA.jl ASAP in favor of RData.jl. A good stress test for it is porting RDatasets.jl and checking that all data sets can be loaded correctly.

ararslan:
First RData needs to be registered in METADATA. Should also set up Travis and coverage.

ararslan:
@alyst If you want you could transfer ownership of RData.jl to the JuliaStats organization for more rapid development and maintenance, but that's totally up to you.

alyst:
@ararslan If you think that making `RData` a part of JuliaStats would be beneficial for the community, I am all for it. Is there anything required from my side?

ararslan:
[...]

PR #1006, opened 2016-06-29
Fix failure of RDA test by working around a bug with NaN bit patterns

nalimilan:
The failure only affected Julia 0.5 since on 0.4.6 the corruption
happened everywhere: the bit pattern was consistent and comparisons
worked. See JuliaLang/julia#17195.

-------
This should completely fix the tests on 0.5. Therefore, I'd like to tag a new release after merging this, even if there still are deprecation warnings.

### DIFF
--- a/src/RDA.jl
+++ b/src/RDA.jl
@@ -56,14 +56,17 @@ const SXPtab = @compat Dict(      # Defined in Rinternals.h
 ##############################################################################
 ##
 ## Constants used as NA patterns in R.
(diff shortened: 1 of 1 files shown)


dmbates:
Note also issue #955 

I think this facility can be removed from `DataFrames`

nalimilan:
Yes, it should definitely be moved to another package, but we need a deprecation period in which it continues to work.

PR #850, opened 2015-08-07
Enhance joining and grouping

alyst:
PR addresses 4 limitations of current joining and grouping implementation:
- the way row groups are indexed for multicolumn joining/grouping is very sparse (i.e. many usable indices have no rows assigned) and can lead to math overflows even for medium size frames
- rows order in the left-joined or outer-joined frame doesn't match the rows order in the left frame (which is usually nice to maintain)
- for any join kind, including `:inner`, `:semi`, `:anti` full matching of left and right frames is done, which is not necessary
- poor performance for frames containing `PooledDataVector` with large pools due to slow `setindex!(PooledDataArray)`.

Here's the simple benchmarking scripts to test the performance.

``` julia
using Distributions, DataFrames

function random_frame(nrow::Int, col_values::Dict{Symbol, Any})
  DataFrame(Any[isa(col_values[key], PooledDataArray) ?
                @pdata(sample(col_values[key], nrow)) :
                @data(sample(col_values[key], nrow)) for key in keys(col_values)],
            keys(col_values) |> collect)
end

function random_join(kind::Symbol, nrow_left::Int, nrow_right::Int,
                     on_col_values::Dict{Symbol, Any},
                     left_col_values::Dict{Symbol, Any},
                     right_col_values::Dict{Symbol, Any})
  dfl = random_frame(nrow_left, merge(on_col_values, left_col_values))
  dfr = random_frame(nrow_right, merge(on_col_values, right_col_values))
  join(dfl, dfr, on = keys(on_col_values) |> collect, kind = kind)
end

function f(n::Int)
  for i in 1:n
    r = random_join(:outer, 1000, 2000,
                Dict{Symbol,Any}(:A => 1:10, :B => @data([:A, :B, :C, :D]),
                                 :C => 1:10, :D => 1:10),
                Dict{Symbol,Any}(:E => 1:10, :F => @data([:A, :B, :C, :D])),
                Dict{Symbol,Any}(:G => 1:10, :H => @data([:A, :B, :C, :D])))
  end
end

f(1)

@time f(100)
```

For the simple cases PR has almost the same join performance as the current implementation (increased GC overhead is likely due to the allocation of additional arrays that store the rows order in the resulting frame):

Current times:

```
  7.196765 seconds (60.15 M allocations: 1.920 GB, 3.94% gc time)
  7.134183 seconds (60.16 M allocations: 1.920 GB, 4.01% gc time)
  7.180465 seconds (60.15 M allocations: 1.920 GB, 3.95% gc time)
[...]

PR #1008, opened 2016-07-03
Port to NullableArrays and CategoricalArrays

nalimilan:
Completely remove support for DataArrays.

Still rough, but passes the tests on Julia 0.5.


datnamer:
Does this improve speed and type stability?

nalimilan:
Type-stability: yes. Speed: haven't measured anything yet, but it should.

ViralBShah:
I did write a very simple sum benchmark using NullableArrays and DataArrays yesterday. NullableArrays was within 2x of Arrays, and 10x faster than DataArrays. I would expect this to carry over, but of course, much better and systematic benchmarking is necessary.

quinnj:
Word. I'm pumped for this. I'll try to find some time in the next few days to pull the branch and play around.

src/abstractdataframe/join.jl
```diff
@@ -76,21 +89,127 @@ function join_idx(left, right, max_groups)
      right_sorter[right_indexer], right_sorter[rightonly_indexer])
 end
 
-function DataArrays.PooledDataVecs(df1::AbstractDataFrame,
-                                   df2::AbstractDataFrame)
+# FIXME: rename this, and possibly move to CategoricalArrays.jl
+function PooledDataVecs{S,N}(v1::NullableCategoricalArray{S,N},
```
quinnj: Yeah, I think it would be good to rename this as it's clearly a reference to PooledDataArrays. The operation here is a join!/union!/merge!, right?

src/abstractdataframe/reshape.jl
```diff
@@ -200,23 +205,28 @@ function unstack(df::AbstractDataFrame, colkey::Int, value::Int)
...
+    rowkey = zeros(Int, size(df, 1))
     for i in 1:length(groupidxs)
         rowkey[groupidxs[i]] = i
     end
-    keycol = PooledDataArray(df[colkey])
+    keycol = NullableNominalArray(df[colkey])
     valuecol = df[value]
     df1 = df[g.idx[g.starts], g.cols]
-    keys = unique(keycol)
+    keys = unique(keycol) # FIXME: why not levels?
```
quinnj: There shouldn't be a difference right?

quinnj: Can there be cases where you have "unused levels" in CategoricalArrays? So unique would have to ensure that each level was actually used in the array?

src/dataframe/dataframe.jl
```diff
@@ -183,7 +185,7 @@ function DataFrame{D <: Associative}(ds::Vector{D}, ks::Vector)
     df = DataFrame(col_eltypes, ks, length(ds))
     for (i,d) in enumerate(ds)
         for (j,k) in enumerate(ks)
-            df[i,j] = get(d, k, NA)
+            df[i,j] = get(d, k, Nullable())
         end
     end
```
quinnj: There are a couple references to `AbstractDataVector` in lines 217-232

src/dataframe/dataframe.jl
```diff
@@ -234,7 +236,9 @@ function Base.getindex(df::DataFrame, col_ind::ColumnIndex)
 end
 
 # df[MultiColumnIndex] => (Sub)?DataFrame
-function Base.getindex{T <: ColumnIndex}(df::DataFrame, col_inds::AbstractVector{T})
+function Base.getindex{T <: ColumnIndex}(df::DataFrame,
+                                         col_inds::Union{AbstractVector{T},
+                                                         AbstractVector{Nullable{T}}})
```
quinnj: I'm guessing there are other operations that somehow rely on the fact that indexing with a Nullable works? Otherwise, this seems a little overly-permissive and could just be `AbstractVector{T}`.

src/dataframe/dataframe.jl
```diff
@@ -759,7 +792,8 @@ end
 ##
 ##############################################################################
 
-pool(a::AbstractVector) = compact(PooledDataArray(a))
+pool(a::AbstractVector) = compact(NominalArray(a))
+pool{T<:Nullable}(a::AbstractVector{T}) = compact(NullableNominalArray(a))
 
 function pool!(df::DataFrame, cname::@compat(Union{Integer, Symbol}))
```
quinnj: Is there a reason `cname` isn't `ColumnIndex` here?

src/dataframe/io.jl
```diff
@@ -801,7 +801,7 @@ function readtable(io::IO,
                    separator::Char = ',',
                    quotemark::Vector{Char} = ['"'],
                    decimal::Char = '.',
-                   nastrings::Vector = ["", "NA"],
+                   nastrings::Vector = ["", "NULL", "NA"],
```
quinnj: Is this strictly necessary? Do we actually write out `NULL` now in writetable? If we do, I'd say we shouldn't and just follow the typical CSV/DBMS tradition of writing nothing for null fields. i.e `col1,col2,col3\n,,\n`.

src/dataframerow/dataframerow.jl
```diff
@@ -50,25 +50,26 @@ function Base.hash(r::DataFrameRow, h::UInt)
...
-
 # comparison of DataFrame rows
 # only the rows of the same DataFrame could be compared
 # rows are equal if they have the same values (while the row indices could differ)
+function Base.:(==)(r1::DataFrameRow, r2::DataFrameRow)
+    if r1.df !== r2.df
+        throw(ArgumentError("Comparing rows from different frames not supported"))
+    end
+    if r1.row == r2.row
+        return Nullable(true)
```
quinnj: Why do we return `Nullable(true)` instead of just `true`? Is there something inherently `Nullable` about a DataFrameRow?

test/RDA.jl
```diff
@@ -1,62 +0,0 @@
-module TestRDA
```
quinnj: What exactly is the RDA stuff? R's binary dataframe format? Should we have a stand-alone package for RDA instead of just dropping it?

REQUIRE
```diff
@@ -1,5 +1,6 @@
 julia 0.4
-DataArrays 0.3.4
+NullableArrays 0.0.5
+CategoricalArrays 0.0.2
```
quinnj: Not exactly sure if these are right or not, but `Pkg` seemed to be unhappy with my state of things when I checked out this branch.

src/DataFrames.jl
```diff
@@ -75,7 +75,61 @@ export @~,
        stackdf,
        unique!,
        unstack,
-       writetable
+       writetable,
+       # FIXME: unexport, these should go in Base or nowhere
```
quinnj: Has there been a larger discussion on this? DataFrames seems like a natural place for these....

src/abstractdataframe/join.jl
nalimilan: Yeah, I'm not sure about the best name. Maybe `join!`. Ideally, we would also avoid accessing the `refs` field from the `PooledDataVecs(::AbstractDataFrame, ::AbstractDataFrame)` method below. But I'm even less sure what the API should look like in CategoricalArrays.jl. Could still improve this in steps.

src/abstractdataframe/reshape.jl
nalimilan: Yes, unused levels aren't removed, in particular for efficiency. Also, `unique` returns the levels in their order in appearance I think (which we could stop doing for efficiency if we don't think it's useful). But I left the FIXME because I didn't want to check whether it would be an issue to keep unused levels, nor to find the best way to check this. Now that tests pass it's time to investigate...

src/dataframe/dataframe.jl
nalimilan: Ah, good catch!

src/dataframe/dataframe.jl
nalimilan: I guess I added these because it triggered failures in tests. The problem is that any `DataArray` was accepted by this signature, but that's no longer the case with `NullableArrays`. This means you can't index a data frame based on a condition on one of its columns. I think that's a showstopper.

src/dataframe/dataframe.jl
nalimilan: No idea, but you're probably right that it's an oversight. Care to make a PR to fix this on master too?

src/dataframe/io.jl
nalimilan: Well, empty fields can also mean `""` (empty string) for `String` columns, so that cannot replace `NULL` in all cases. It's also a good idea to provide some flexibility here, e.g. to read files produced by R. And yes, we do write `NULL` for missing values by default.

test/RDA.jl
nalimilan: Yes, that's https://github.com/JuliaStats/DataFrames.jl/issues/955. I removed these files to get them out of the way, but we should move them to another package ASAP and deprecate them here.

REQUIRE
nalimilan: What's the error? I tagged these releases a few days ago precisely to make this branch pass the tests.

src/DataFrames.jl
nalimilan: Why would they be limited to data frames? In R, these functions work on any vector, returning the first/last elements. There's been some discussion in Base about that, a possible solution was to extend `first` and `last`.

It shouldn't be related to this PR in theory, but these methods are defined in DataArrays, that's why I had to do something about them. But I can keep them in DataFrames for now and tackle that problem later.

src/dataframe/io.jl
davidanthoff: I agree with @quinnj here, I would much prefer if ``NULL`` values would not write out a ``NULL`` but just nothing, and if the readtable function wouldn't add ``NULL`` to the ``nastrings`` vector by default. These default ``nastrings`` are quite a problem already: I think they make sense for non-string columns, but having them there by default for string columns is not ideal at all, after all both "NA" or "NULL" are valid values for a string. And e.g. according to the iso alpha-2 code "NA" is the country code for Namibia, and there is a *lot* of data out there that uses that classification.

johnmyleswhite: I think it's worth noting that Hive prints out `NULL` by default and that R uses `NA` by default. If people are unhappy with the defaults, I think the only defensible solution is to have no default and require people to specify things manually.

ViralBShah: I prefer following R's lead here.

src/dataframe/io.jl
```diff
@@ -517,7 +517,7 @@ function builddf(rows::Integer,
             values = Array(o.eltypes[j], rows)
         end
 
-        missing = falses(rows)
+        missing = fill(false, rows)
```
ararslan: Why change `falses`?

src/DataFrames.jl
```diff
@@ -12,7 +12,8 @@ using Compat
 import Compat.String
 using Reexport
 @reexport using StatsBase
-@reexport using DataArrays
+@reexport using NullableArrays
```
ararslan: Are the `@reexport`s necessary? Assuming I didn't miss anything reading through this, you're only overloading one NullableArrays function and two CategoricalArrays functions. Could we just export the modified functions rather than reexporting the entire modules?

dmbates:
With the RDA stuff removed you can probably avoid `using GZip` in `src/DataFrames.jl`.  I included that specifically for reading compressed .rda files.

ViralBShah:
What holds up the merging here? If a few things need to be moved out, that can be handled in a different PR. I say we merge this if basically this does the right things.

ararslan:
Milan is on vacation, which I assume is at least part of the holdup.

ViralBShah:
I missed that. Thanks.

dmbates:
I find myself stumbling over DataArrays in doing some analyses and was wondering how this branch is coming along.  Are there things that others can do to help?

nalimilan:
This branch needs some cleanup before merging, but I think we should also wait until we provide convenience APIs for lifting semantics. Without these, the port will be a major regression in terms of usability. I would certainly be useful if people could experiment in that direction in particular in DataFramesMeta, jplyr, NullableArrays and NullableUtils, as @davidagold is already doing.

dmbates:
@nalimilan I have been working on an analysis in Julia of some benchmarking results, which is why I found myself tripping over DataArrays.  The analysis is in [TimingResults](https://github.com/dmbates/JSM/blob/master/TimingResults.ipynb)  As you can see I used `DataFramesMeta` in addition to `DataFrames`.  I started using `Gadfly` for plots but still find it difficult to implement the plot I am thinking of.  I ended up reverting to lattice graphics accessed through RCall, because I know lattice graphics quite well.

My conclusion from this exercise is that the single most effective change we can make regarding DataFrames is not to create a NullableArray or DataArray unless it is necessary.  It is fine to consider lifting semantics, etc. but, when there are no missing data in a column, creating a missing-data-enabled type then sweating over how to handle missing data values that are never present is what J. Edwards Deming called "burning the toast and then scraping it".  Why go to all that trouble and, in the process, slow down the computation if you don't need to?   Functions such as `readtable` should allow for missing data, of course, but should return `Array`s when there are no missing data values in a column. Reading a feather file using the Feather package should do the same thing.  RCall conversions of R `data.frame`s should do this too.  I would go one step further and suggest adding a function (I haven't decided on a good name for it) that takes a DataFrame and converts any `NullableArray{T}` columns to `Array{T}` if there are no nulls.  Same for `NullableNominalArray`, etc. The `completecases` and `completecases!` functions should also convert nullable structures to plain arrays.

By the way, @nalimilan, I applaud your decision to create separate types `NominalArray` and `NullableNominalArray`, etc. in the `CategoricalArrays` package.  This is exactly the distinction we should make.

As an exercise, consider the proportion of columns with missing data in the data sets you encounter.  For me, at least, columns with missing data are the exception rather than the rule.

Regarding `DataFramesMeta` I took a while to learn it and now find it quite useful.  Is there a reason to also mention `jplyr`?  Anyone is free to create their own package - this is open source software - but the JuliaStats group should try to settle on one approach to a given problem, not multiple incompatible approaches.  (The selection of verbs in `DataFramesMeta` follows the LINQ/SQL tradition whereas `jplyr` follows `dplyr`).  Unless there is something wonderful in `jplyr` that cannot be implemented in `DataFramesMeta` the approach should be to improve `DataFramesMeta`, not create another incompatible package.  As an aside, @davidagold , a name like `jplyr`, while informative to those who know `dplyr`, removes flexibility.  You are bound by the name to emulate the approach of `dplyr`, whether it fits in well with Julia or not, and you are always playing catch up in the implementation. 

So, what am I missing?  Should I be creating my plots using the `Plots` package or ...?

johnmyleswhite:
@dmbates: I think the problem is that DataFramesMeta doesn't fully solve the problems you're describing as the moment, but somewhat exacerbates them: its existing commitment to non-lazy DAG accumulation means that we can't use it to generate SQL -- which is the main example we have of backend-independence. We could change those semantics in that package, but it would completely break all existing code using that package.

David and I have begun preparing a document we'll share soon about the semantics we hope all tables will agree upon. Assuming we can actually get everyone on board, we could fix the situation by ensuring that every downstream package works only with a standardized abstract interface that multiple backends could implement. Part of the purpose of this document is to point out that we can implement the dplyr semantics with very high performance and fully automatic lifting regardless of historical Julia backends: we could make the current DataFrames fast or make the new version with NullableArrays fast. The only thing we really need is to maintain the tradition of always keeping the values array and is-null mask array separate. In cases where we intend to keep a non-nullable column in a table, I think we're better off not using `Array` objects, but instead adding boolean flags to every `NullableArray` that specify `allownulls` and `anynull`. That would simplify implementations, which will need to have access to the internal representation of columns. We can always work around that, but I don't think it will buy us much.

As for Gadfly, I think the best thing we could do there is remove the canonical interface and focus on the heretical one, which would assume that inputs are specified as `AbstractArray` objects. Then the tabular data interface issue just becomes one of ensuring that a table can always provide access to its columns as `AbstractArray` objects.

tshort:
I agree with @dmbates on preferring `Arrays` where possible. Towards that end, I included a `PassThrough` type with abbreviation `P` in DataFramesMeta to try to avoid creation of `DataArrays` where they aren't needed.

With regards to jplyr, it's quite new and promising. I wouldn't want to discourage development or exploration. It's approach has advantages over DataFramesMeta. DataFramesMeta was always meant to be a playground.

dmbates:
@nalimilan Thanks for the clarification on how `DataFramesMeta` can be improved and is being improved in `jplyr`.  I appreciate the comments by @tshort too.

My comments on the name `jplyr` still hold.  It is a name that will be recognized by those with `dplyr` experience but it is not informative to others.  (`plyr` is not meaningful to those outside the HadleyVerse.) Furthermore, it seems to claim that `jplyr` is `dplyr` in Julia - nothing more, nothing less.  That's a difficult expectation to fulfill and also an unnecessary constraint.  Please consider renaming it.

I still think it is important to prefer `Array`s when possible.  I will read your document with interest when it is available but also, I'm sorry to say, with a bit of skepticism.  There is no free lunch.  Saying that all columns of a DataFrame must be of a missing-value-enabled type but perhaps with an extra tag that indicates they don't really have any missing data means that they are no longer arrays and you are back at the lifting semantics.  To me that still seems like burning the toast and scraping it.  And, it seems to me, it is unnecessary.  Why take an `Array` and make it into something that is no longer an `Array` and must be massaged to get it back to being an `Array` so you can operate on it?

I am not infallible by any means and the whole Julia project has already pulled a very big rabbit out of the hat by creating a language with dynamic types where methods can be compiled to machine code.  So I could be pleasantly surprised and I hope I will be.  But if your objective is to create fully automatic lifting that is as fast as not doing any lifting at all, that seems to be a big ask.

davidagold:
@dmbates The name alludes to Hadley's notion of dplyr as a "grammar of data manipulation", rather than a specific design spec. It's just a working project title -- I intend to change it. Does anybody have any suggestions? One thought I've had is Select.jl.

dmbates:
@davidagold Before there was `dplyr` there was `plyr`, which was a generalization of the various `*apply` functions in R and a play on the word "plyer".  But functions like `lapply` and friends are called `map` in other languages, including Julia.

I regret that I don't have a good suggestion for a name at this point but I will certainly speak up if something comes to mind.

dmbates:
I just noticed that I replied to a comments by @johnmyleswhite as if they were by @nalimilan.  Sorry for the confusion.

I must admit to being discouraged that, after over 4 years of development, DataFrames still has some pretty glaring problems.  I will be giving a presentation at the Joint Statistical Meetings on Sunday and, for all the good aspects of Julia, I can't yet tell the audience "yes, you really should be using Julia for data analysis".  Maybe the grand vision thing will work out.  Maybe it will be another case of "the best is the enemy of the good".  I don't think that potential users will wait forever to find out.

JeffBezanson:
@dmbates I agree with you that nullability should not be in the driver's seat. The natural factoring is to have a container like DataFrame or SomeArray think as little as possible about what goes into it, and treat Nullable as just another data type that user code needs to know how to handle. I'd provide a function like the following:

```
function nlift(f)
    lifted(x) = (isnull(x) ? x : Nullable(f(x)))
    lifted(x, y) = (isnull(x) ? x : isnull(y) ? y : Nullable(f(x,y)))
    return lifted
end
```

and then a user can pass e.g. `nlift(+)` instead of `+`. I'll even let you use `return_type` in there :)
This may seem less convenient at first, but I believe going along with the natural, functional-programmer-approved factoring is usually better in the long run.

milktrader:
Select.jl is an elegant and descriptive name. I've got some clever names in mind but this topic doesn't seem like clever will do.

ViralBShah:
I think Select.jl could mean too many things to different people. Perhaps discussion about new names for jplyr should be moved to that repo. 

I agree with the point here that nullability should not be in the driver's seat. 

@dmbates I think rather than saying that DataFrames.jl has glaring problems after 4 years, it might be better to frame this as the challenges the Julia community has faced working with heterogeneous and missing data, and how we are addressing the problem. It doesn't stop there, but the stuff we are doing goes beyond tabular data with multi-dimensional sparse storage - eg. NDSparseData.jl., or the work with graphs and data in Pranav's ParallelGraphs.jl (needs renaming). There's lots of interesting developments happening in this space in Julia right now, and while it is taking long, I find this to be quite a rich and rewarding experience. Perhaps there will be people at JSM that may appreciate this.

davidanthoff:
And to add another option to the mix: I started a package [LINQ.jl](https://github.com/davidanthoff/LINQ.jl/tree/dev) a while ago. It is far less advanced than jplyr.jl, but it might have some ideas worth pursuing. The core idea is to use iterators that produce NamedTuples for most everything.

matthieugomez:
@dmbates for me, columns with missing data are the rule, not the exception. I think it holds true for applied researchers in social science.

johnmyleswhite:
David and I have made a lot of progress on testing out the ideas in jplyr. I've created a repo with my notes (as I follow along with David's work it's helpful to keep a separate repo that doesn't have to stay in sync).

I did some benchmarking of the approach to automatic lifting I've been envisioning, which can seen in the README at https://github.com/johnmyleswhite/jplyr_notes.jl. Basically there's a noticeable fixed cost to adding in more abstractions to make these things smoother, but the asymptotic performance relative to working with arrays without nullability is about a 50% loss in performance. To me, that's a pretty minor factor: Julia itself is about 50% slower than C often.

It remains to be seen if we can sustain that small performance gap as we add more functionality, but I now believe that the anonymous function changes in 0.5 (and Jeff's newfound tolerance of calling `return_types`) puts us in a position to implement an API that will solve both our performance concerns and the burdens of manually lifting arithmetic expressions.

dmbates:
@johnmyleswhite Thanks for creating the comparisons.  May I suggest using [`RCall`](https://github.com/JuliaStats/RCall.jl) to create the comparison plot directly from the Julia benchmarking code?

johnmyleswhite:
Ok. I'll sit down and finally learn RCall before making the next graphs.

quinnj:
I've been mulling @dmbates thoughts over for the last few days and I'm inclined to agree that we should try to have plain Arrays where possible. I think it fits well in the entire Julia paradigm of performance and abstraction where needed.

My one concern, and what I think we need to have some more discussion/thought around is the idea of what formally constitutes the "Column" interface. I think in addition to the ideas that have been discussed around an `AbstractTable` type, we also need to think carefully about the `AbstractColumn` interface so that whatever gets put in an `AbstractTable` can be consistently extracted, looped over, etc. I think `AbstractArray` obviously is a close cousin here, but also not quite the same, since we want to probably define something like `countnulls` or `removenulls`, etc.

I think getting some definitions here will help in the long-term so that one could theoretically use any kind of "column" type as the internal storage for a DataFrame, including a `SparseVector` or any other kind of exotic datastructure.

johnmyleswhite:
I suspect I'm in a minority here, but I would discourage there being any kind of column-level interface. I think it's ok to ask for a DB to materialize a column into an array, but I think you don't want to ever get into a situation in which you assume that DB's internally have access to columns that behave anything like arrays. (Neither Hive nor Presto could support such a thing and you'd OOM-kill any local machine that attempted to materialize a single column into an array.) From the perspective of relational algebra, I don't think columns exist as an abstraction that theory knows how to talk about: there are just tables composed of tuples with only one element.

datnamer:
Can't we just have a columned table subtype?

davidanthoff:
Should we move this discussion somewhere else? I think it has moved long away from what this PR here covers. Maybe a new issue here in DataFrames would be best? That way people would at least find it. There seem to be a number of repos around now with ideas for implementations, but I think having the discussion here at DataFrames makes most sense.

PallHaraldsson:
I see: dataframesjl.readthedocs.io/en/latest/introduction.html

"The DataFrames package (and its sibling, DataArrays)"

This needs to change if not already in commit (that I see now, is WIP).

[I see speed improvements, up in discussions I just scanned. Does this close feature and performance gap with R (at least concerning NA)?]

davidanthoff:
Bump, what is the status with this PR?

ararslan:
I think Milan is away again but IIRC I think he should be back fairly soon?

nalimilan:
Right, I'll try to start working on this again soon.

davidanthoff:
Btw., what is the transition story here? I assume this will break almost any code that currently uses a ``DataFrame``?

quinnj:
Hey @nalimilan, how is `DataFrame` equality affected here? That's been a problematic thing as I've been a heavy DF + NullableArray user for a few weeks now. The problem is that I think a DataFrame just checks that each cell is equal to each cell of DF2, yet with NullableArrays you end up with `TypeError: non-boolean (Nullable{Bool}) used in boolean context`.

The story here is probably more centered around `Nullable` equality. I would think there are two notions of equality: one where we of course want to account for null values and have the result be `Nullable{Bool}` (and this should be the default), but it seems like we also need a notion where if only one value is null then `false`, if both are false then `true`, otherwise the actual values are compared. This would allow a more natural "DataFrame equality" where we really want to compare the *structure*, do we have the same column names, types, and then do we have the same values, with missing cells meaning two DataFrames are matching.

Happy to entertain other reasonings/thoughts here.

davidagold:
@quinnj I've always used `isequal` for that purpose, e.g. in testing:

```julia
julia> using NullableArrays

julia> A = NullableArray([1,2,3], [true, true, false]);

julia> B = NullableArray([1,1,3], [true, true, false]);

julia> isequal(A, B)
true
```

There's also https://github.com/JuliaStats/NullableArrays.jl/pull/84 which I suppose could (perhaps even ought to be) merged.

src/DataFrames.jl
nalimilan: I think the idea was that people shouldn't have to call `using` to be able to use methods from these packages, since one can hardly work with DataFrames without also loading these packages.

nalimilan:
With the latest round of commits, I think this is ready for a serious review. Any volunteers? To get tests passing, you'll need the latest CategoricalArrays.jl, as well as the `nl/local` branch of NullableArrays.jl (which includes my recent, still unmerged PRs).

quinnj:
I'll take some time to dig in today.

nalimilan:
OK, tests now pass except for an ugly crash during precompilation on 0.4. Probably better drop support for 0.4 at this point than wasting time on bisecting the fix.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -231,42 +232,33 @@ Base.ndims(::AbstractDataFrame) = 2
...
 ##
 ## Equality
 ##
 ##############################################################################
 
-function Base.isequal(df1::AbstractDataFrame, df2::AbstractDataFrame)
-    size(df1, 2) == size(df2, 2) || return false
-    isequal(index(df1), index(df2)) || return false
+# Imported in DataFrames.jl for compatibility across Julia 0.4 and 0.5
+function @compat(Base.:(==))(df1::AbstractDataFrame, df2::AbstractDataFrame)
```
davidagold: I'm still not sold on returning a `Nullable{Bool}` here. `DataFrame` is not a `Nullable` type, and there's nothing that says that it has to inherit the logic of `==` to which its stored values subscribe. cf. https://github.com/JuliaStats/DataFrames.jl/pull/1008#issuecomment-240612693

quinnj:
Are you running into https://github.com/JuliaLang/julia/pull/16745 ? What version of 0.4 are you running? I'd rather keep support for 0.4 if at all possible, especially since 0.5 isn't even out yet. There's a broader discussion around whether it might make for a cleaner transition to only support this change on 0.5 forward; anyone have any strong feelings on that?

nalimilan:
I have no idea what's the issue I'm hitting as it's the first time I see it. That's on 0.4.6. I had previously seen https://github.com/JuliaLang/julia/issues/18299 (for which I added a workaround) on 0.5 too. Anyway, we can investigate it if we still want to support 0.4. OTOH, as you said, it would make sense to only enable this port on 0.5 to avoid breaking people's workflow too badly: basically all operations are going to break...

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -285,10 +277,10 @@ Base.isempty(df::AbstractDataFrame) = ncol(df) == 0
 ##
 ##############################################################################
 
-DataArrays.head(df::AbstractDataFrame, r::Int) = df[1:min(r,nrow(df)), :]
-DataArrays.head(df::AbstractDataFrame) = head(df, 6)
-DataArrays.tail(df::AbstractDataFrame, r::Int) = df[max(1,nrow(df)-r+1):nrow(df), :]
-DataArrays.tail(df::AbstractDataFrame) = tail(df, 6)
+head(df::AbstractDataFrame, r::Int) = df[1:min(r,nrow(df)), :]
```
davidagold: What happened to that proposal to put `head`/`tail` in Base? I can't find it.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Yeah, but let's continue discussing this at https://github.com/JuliaStats/NullableArrays.jl/issues/84 to avoid making this PR too messy. I guess I'll switch it to returning a `Bool` since it seems that's the general preference.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -404,32 +392,33 @@ function StatsBase.describe(io, df::AbstractDataFrame)
         println(io, )
     end
 end
-StatsBase.describe(dv::AbstractArray) = describe(STDOUT, dv)
-function StatsBase.describe{T<:Number}(io, dv::AbstractArray{T})
-    if all(isna(dv))
+StatsBase.describe(nv::AbstractArray) = describe(STDOUT, nv)
+function StatsBase.describe{T<:Number}(io, nv::AbstractArray{T})
+    if all(_isnull, nv)
         println(io, " * All NA * ")
```
davidagold: `" * All null * "`?

davidagold: `" * All null * "`?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -518,27 +526,28 @@ function Base.convert{T}(::Type{Matrix{T}}, df::AbstractDataFrame)
     n, p = size(df)
     res = Array(T, n, p)
     idx = 1
-    for col in columns(df)
-        anyna(col) && error("DataFrame contains NAs")
-        copy!(res, idx, data(col))
+    for (name, col) in zip(names(df), columns(df))
+        anynull(col) && error("cannot convert a DataFrame containing null values to array (found for column $name)")
```
davidagold: Not specifically related to this PR, but this error message may give the (possibly false) impression that only the column mentioned in the error has null values. Sorry if this is pedantic.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Good catch. I should run a series of `git grep` to detect remnants of DataArrays. I haven't ported the documentation and benchmarks either.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Yeah, touching many parts of the code allows spotting many different bugs Please file issues for any improvements you can think of so that we don't forget them.

src/abstractdataframe/abstractdataframe.jl
nalimilan: For now I've decided to keep these functions in DataFrames, and try migrating them to Base in a subsequent phase. The PR is already complex enough.

quinnj:
I just bumped the CategoricalArrays requirement, let's see what travis thinks now

nalimilan:
It shouldn't make any difference, as it uses the latest version. I still need to fix failures which are not related to the previous crash.

davidanthoff:
Really great to see all the progress here!

The one thing I'm still a bit worried about is the transition story. Will there just be a release that breaks almost every user of DataFrames? If so, it might make sense to have a defined time window during which these changes are on master, but not in a tagged version, and during which we try *very* hard to convince at least the packages registered in METADATA that use DataFrames to make them compatible with the new version before it gets tagged.

johnmyleswhite:
FWIW, I think we need to announce this change and then wait at least a month before tagging this release. It will be massively breaking, but I suspect the only solutions are either pre-announcements so that people know what's coming or just renaming the package.

tlnagy:
As someone who's just found out about this, what are the main breaking changes to expect? I want to make sure that Gadfly doesn't get broken by this.

quinnj:
I think the transition looks like:

* Merge this PR to master with a bump to REQUIRE to `julia 0.5-`
* julia-users announcement about DataFrames change to master that will be tagged for julia 0.5 in the next few weeks
* Allow a few weeks or so for various packages to test on DataFrames master
* Tag a new DataFrames release in METADATA, again for `julia 0.5-`
* Help rest of the ecosystem tag upper bounds for package that aren't ready to support this
* Help transition as many packages to support this with new tagged releases

nalimilan:
I agree with that plan as a whole. One problem with supporting only 0.5, though, is that it will force all dependencies to drop support for 0.4, or to maintain parallel branches. So I'm not sure what's the best approach.

@tlnagy It really depends on what Gadfly is doing with DataFrames, but one of the main changes is that e.g. `sin(df[1])` will no longer work without a special invocation (like a macro or `map`). The convenience APIs to replace the current system aren't completely defined yet (the porting period should help identify the needs and the best solutions).

quinnj:
Not sure if this is specific to this PR, or it might live in NullableArrays or even Base, but it seems to be a show alignment issue related to #NULL values.
```julia
│ Row │ CHARACTER_SET_NAME │ COLLATION_NAME     │ COLUMN_TYPE     │ COLUMN_KEY │
├─────┼────────────────────┼────────────────────┼─────────────────┼────────────┤
│ 1   │ #NULL              │ #NULL              │ "bigint(20)"    │ ""        │
│ 2   │ #NULL              │ #NULL              │ "bit(1)"        │ ""        │
│ 3   │ #NULL              │ #NULL              │ "decimal(10,0)" │ ""        │
│ 4   │ #NULL              │ #NULL              │ "int(11)"       │ ""        │
│ 5   │ #NULL              │ #NULL              │ "decimal(10,0)" │ ""        │
│ 6   │ #NULL              │ #NULL              │ "smallint(6)"   │ ""        │
│ 7   │ #NULL              │ #NULL              │ "mediumint(9)"  │ ""        │
│ 8   │ #NULL              │ #NULL              │ "tinyint(4)"    │ ""        │
│ 9   │ #NULL              │ #NULL              │ "float"         │ ""        │
│ 10  │ #NULL              │ #NULL              │ "double"        │ ""        │
│ 11  │ #NULL              │ #NULL              │ "date"          │ ""        │
│ 12  │ #NULL              │ #NULL              │ "datetime"      │ ""        │
│ 13  │ #NULL              │ #NULL              │ "timestamp"     │ ""        │
│ 14  │ #NULL              │ #NULL              │ "time"          │ ""        │
│ 15  │ #NULL              │ #NULL              │ "year(4)"       │ ""        │
│ 16  │ "utf8"            │ "utf8_general_ci" │ "char(1)"       │ ""        │
│ 17  │ "utf8"            │ "utf8_general_ci" │ "varchar(16)"   │ ""        │
│ 18  │ #NULL              │ #NULL              │ "binary(2)"     │ ""        │
│ 19  │ #NULL              │ #NULL              │ "varbinary(16)" │ ""        │
│ 20  │ #NULL              │ #NULL              │ "tinyblob"      │ ""        │
│ 21  │ #NULL              │ #NULL              │ "blob"          │ ""        │
│ 22  │ #NULL              │ #NULL              │ "mediumblob"    │ ""        │
│ 23  │ #NULL              │ #NULL              │ "longblob"      │ ""        │
│ 24  │ "utf8"            │ "utf8_general_ci" │ "tinytext"      │ ""        │
│ 25  │ "utf8"            │ "utf8_general_ci" │ "text"          │ ""        │
│ 26  │ "utf8"            │ "utf8_general_ci" │ "mediumtext"    │ ""        │
│ 27  │ "utf8"            │ "utf8_general_ci" │ "longtext"      │ ""        │

```

tcovert:
I think the recent update in CategoricalArrays, which removed the Nominal and Ordinal distinctions, has broken this branch.  I get the following error when I try to `use DataFrames`:

`julia> using DataFrames
INFO: Precompiling module DataFrames.
ERROR: LoadError: LoadError: UndefVarError: NominalArray not defined
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /usr/local/Cellar/julia/0.5.0/lib/julia/sys.dylib:?
 in macro expansion; at /Users/tcovert/.julia/v0.5/DataFrames/src/DataFrames.jl:130 [inlined]
 in anonymous at ./<missing>:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /usr/local/Cellar/julia/0.5.0/lib/julia/sys.dylib:?
 in macro expansion; at ./none:2 [inlined]
 in anonymous at ./<missing>:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Any) at /usr/local/Cellar/julia/0.5.0/lib/julia/sys.dylib:?
 in process_options(::Base.JLOptions) at ./client.jl:239
 in _start() at ./client.jl:318
 in _start() at /usr/local/Cellar/julia/0.5.0/lib/julia/sys.dylib:?
while loading /Users/tcovert/.julia/v0.5/DataFrames/src/other/utils.jl, in expression starting on line 147
while loading /Users/tcovert/.julia/v0.5/DataFrames/src/DataFrames.jl, in expression starting on line 99
ERROR: Failed to precompile DataFrames to /Users/tcovert/.julia/lib/v0.5/DataFrames.ji.
 in compilecache(::String) at ./loading.jl:593
 in require(::Symbol) at ./loading.jl:422
 in require(::Symbol) at /usr/local/Cellar/julia/0.5.0/lib/julia/sys.dylib:?`

quinnj:
Hey @nalimilan, how are you feeling about this? Obviously needs a rebase and an update for CategoricalArrays, but are there any other lingering PRs out there this depends on? If not, I'd say let's do one more comb over the code and merge. We can then let things shake out on master before doing a release. Are we still thinking the next release will be 0.5-only? I think it's the best way to help transition forward.

ViralBShah:
0.5 only seems ok.

nalimilan:
> I think the recent update in CategoricalArrays, which removed the Nominal and Ordinal distinctions,

Yes, I'll fix this shortly. Then I'll push updates of the docs, and it should be OK for final review and hopefully merging.

I'm not sure about supporting 0.4 or not. Actually, I'm not even sure we should tag a release with the port until jplyr is more or less usable. Else, I'm afraid it's going to be quite painful. Anyway I'd recommend letting people test how the port works for their use cases on master for a few weeks before releasing it: we're likely to discover lacking features in the `Nullable` story.

davidagold:
I actually intend to tag a 0.0.1 release of jplyr.jl today. Note that jplyr.jl itself is just a query framework; the backend support ("collection machinery") is intended to live elsewhere. We've been developing collection machinery that's generic over in-memory Julia data structures that look like `DataFrame`s [here](https://github.com/davidagold/_AbstractTables.jl/tree/master/src/column_indexable/query). I can just re-use all that code for supporting `DataFrame`s in particular. The only question is whether that code should live in the present package (which would then need to require jplyr.jl) or in an interfacing package that would require both jplyr.jl and DataFrames.jl -- maybe a package called DataQueries.jl or something like that.

davidanthoff:
Note also that [Query.jl](https://github.com/davidanthoff/Query.jl) is functional and tagged and works with DataFrames (and lots of other sources). I'm just finishing the documentation and then need to make sure I'm up-to-date with the recent DataStreams changes and then I will announce the package widely on mailing lists etc. Query right now works with the tagged version of DataFrames, but I had a version that worked with this branch here as well, so it should be very simple to make the switch once this is merged.

nalimilan:
OK, tests pass again on 0.5 and 0.6.

quinnj:
Alright, diving in

quinnj:
Not sure if this is just an old doc example or affected by this PR
```julia
using DataFrames, RDatasets
iris = dataset("datasets", "iris")
iris[:id] = 1:size(iris, 1)  # this makes it easier to unstack
d = stack(iris, [1:4])
```
produces
```julia
julia> d = stack(iris, [1:4])
ERROR: MethodError: no method matching getindex(::DataFrames.Index, ::Array{UnitRange{Int64},1})
 in stack(::DataFrames.DataFrame, ::Array{UnitRange{Int64},1}) at /Users/jacobquinn/.julia/v0.5/DataFrames/src/abstractdataframe/reshape.jl:93
```

quinnj:
This example from the docs doesn't seem to work, how do you modify in-place DataFrame columns to make them CategoricalArrays?
```julia
df = DataFrame(A = [1, 1, 1, 2, 2, 2],
               B = ["X", "X", "X", "Y", "Y", "Y"])
categorical!(df, [:A, :B])
```

quinnj:
Pushed a few doc updates.

quinnj:
Ok, pushed a few more updates (mostly examples, references to NA, etc.).

I think they main things I'd like to see are:

* Make a final decision on the return type of `==(df, df2)`. `Bool`? `Nullable{Bool}`? I know there's been a lot of back and forth on this in a few different places. My preference would be for `Bool`. We also need to make `DataFrameRowIterator` consistent
* Fix a couple of the examples I noted above, where we're running into `*(::Nullable{Int64}, ::Int64)` type errors. We're in the Nullable world now, so as many examples/tools we can share with users, the better.
* Figure out what to do about 0.4. We either need to figure out something to do for CategoricalArrays precompilation on 0.4 (maybe disabling another constructor or two?), or drop support for 0.4 altogether. I think preferences are leaning towards the latter.

ararslan:
+1 for `==` being `Bool` and +1 for dropping 0.4 support in the tag that includes this

docs/src/man/getting_started.md
```diff
@@ -95,8 +95,8 @@ ncols = size(df, 2)
 We can also look at small subsets of the data in a couple of different ways:
 
 ```julia
-head(df)
-tail(df)
+DataFrames.head(df)
```
nalimilan: Why would we need this?

docs/src/man/pooling.md
```diff
@@ -19,21 +19,21 @@ A companion type, `NullableCategoricalArray`, allows storing missing values in t
 In addition to representing repeated data efficiently, the `CategoricalArray` type allows us to determine efficiently the allowed levels of the variable at any time using the `levels` function (note that levels may or may not be actually used in the data):
 
 ```julia
-levels(cv)
+CategoricalArrays.levels(cv)
```
nalimilan: Same here. We can't reasonably expect users to fully qualify the function name on every use.

docs/src/man/pooling.md
```diff
@@ -19,21 +19,21 @@ A companion type, `NullableCategoricalArray`, allows storing missing values in t
...
 ```julia
-cv = compact(cv)
+cv = CategoricalArrays.compact(cv)
 ```
 
 Often, you will have factors encoded inside a DataFrame with `Array` or `NullableArray` columns instead of `CategoricalArray` or `NullableCategoricalArray` columns. You can do conversion of a single column using the `categorical` function:
 
 ```julia
-cv = categorical(v)
+cv = CategoricalArray(v)
```
nalimilan: I contemplated adding this function as a way to create either a `CategoricalArray` or a `NullableCategoricalArray` depending on whether the input is an `AbstractArray{<:Nullable}`. Does that sound like a good idea? Using the `CategoricalArray` constructor doesn't sound right, as it should always return a `CategoricalArray` object for clarity.

docs/src/man/getting_started.md
quinnj: It doesn't seem that `head` is exported, so this is the only way this line works

docs/src/man/pooling.md
quinnj: if `levels` isn't exported from CategoricalArrays, then what other option do we have?

nalimilan:
`categorical!` doesn't exist yet, but we should probably add it if the old `pool!` method is deemed useful. Same for `categorical`. Thoughts?

As regards `==` I'd vote for deprecating it for now in favor of `isequal` until we solve that issue. (If we don't, then let's return a `Bool`.)

As regards 0.4, I think we could work around the crashes if needed. So the decision shouldn't be based on that, but on more general plans for the transition.

Finally, I'll have a look at the failing examples if nobody beats me to it.

docs/src/man/getting_started.md
quinnj: oh, weird. For some reason it wasn't working for me earlier unless it was fully qualified. I assumed this was because it wasn't exported, but that doesn't actually seem to be the case. Carry on.

docs/src/man/pooling.md
quinnj: Same here. I'm not sure what was weird with my setup before, but nothing seemed to be working unless it was fully qualified.

quinnj:
Ok, I just pushed everything I was aware of, including:

* Doc updates, make sure all examples work
* Create `categorize` and `categorize!` which replace `pool` and `pool!`
* Return a `Bool` from `==` for DataFrame and DataFrameRow instead of `Nullable{Bool}`
* Bump the julia requirement to 0.5-

Let's see what travis says.

davidanthoff:
@quinnj Small point, but shouldn't we bump the julia requirement to 0.5? Why support pre-release versions at this point?

quinnj:
Why not? There was nothing breaking in the RCs that we'd need to be careful of here, at least that I'm aware of.

davidanthoff:
Doesn't ``0.5-`` refer to any 0.5 pre-release, i.e. also any pre-RC build?

quinnj:
No, you're right @davidanthoff, for some reason I was thinking it only applied to release candidates. Just pushed a few more cleanups that hopefully should make travis happy.

quinnj:
Alls green here. At this point, I feel confident it's ready to go, but I'm not opposed to going through another round or two of review if anyone else is up for it or would like to. For me, I feel like we've reached a point of "merge and let things shake out".

I say we leave this open for another day or two and anyone who'd like to review or just try it out locally and post feedback, we'll try and get any more changes/tweaks made.

FYI, instructions to try this out locally are:

* From the command line, navigate to `~/.julia/v0.5/DataFrames` directory
* run `git fetch origin`
* Then run `git checkout -b nl/nullable origin/nl/nullable`

This will put your DataFrames package on this PR's code and you can start up julia and run `using DataFrames` to start playing around. Once you've had your fun, just run `git checkout master` or `git checkout v0.8.3` to get back to current master or the last tagged version, respectively.

dmbates:
I think the usual approach for trying this out is
```jl
Pkg.checkout("DataFrames")
Pkg.checkout("DataFrames", "nl/nullable")
```
followed by
```jl
Pkg.free("DataFrames")
```
to revert.

src/dataframe/io.jl
nalimilan: I realize I missed that comment. That was because `NullableArray` stores missingness as a `Vector{Bool}`, not a `BitArray`.

nalimilan:
I've added a few commits tweaking the changes you made, @quinnj. What do you think? If that's OK, I agree it should be good to merge (after tagging a new CategoricalArrays release, cf. https://github.com/nalimilan/CategoricalArrays.jl/pull/19).

nalimilan:
Looks like we're good to go now. Who wants to click the green button? :-)

src/deprecated.jl
```diff
@@ -4,17 +4,13 @@ import Base: @deprecate
...
-@deprecate array(df::AbstractDataFrame) convert(Array, df)
-@deprecate array(r::DataFrameRow) convert(Array, r)
-if VERSION < v"0.4.0-"
-    @deprecate DataArray(df::AbstractDataFrame) convert(DataArray, df)
-end
-@deprecate DataArray(df::AbstractDataFrame, T::DataType) convert(DataArray{T}, df)
-
 @deprecate read_rda(args...) FileIO.load(args...)
+
+@deprecate pool categorize
```
alyst: Should it be `categorical`?

src/deprecated.jl
```diff
@@ -4,17 +4,13 @@ import Base: @deprecate
...
+@deprecate pool! categorize!
```
alyst: `categorical!`?

dmbates:
I seem to have a mismatch between version 0.0.6 of `CategoricalArrays` and this branch
```
julia> Pkg.status("CategoricalArrays")
 - CategoricalArrays             0.0.6

julia> Pkg.add("DataFrames")
INFO: No packages to install, update or remove
INFO: Package database updated

julia> Pkg.checkout("DataFrames")
INFO: Checking out DataFrames master...
INFO: Pulling DataFrames latest master...
INFO: No packages to install, update or remove

julia> Pkg.checkout("DataFrames", "nl/nullable")
INFO: Checking out DataFrames nl/nullable...
WARNING: Cannot perform fast-forward merge.
INFO: Pulling DataFrames latest nl/nullable...
WARNING: Cannot perform fast-forward merge.
INFO: No packages to install, update or remove

julia> using DataFrames
INFO: Precompiling module DataFrames.
ERROR: LoadError: LoadError: UndefVarError: NominalArray not defined
 in include_from_node1(::String) at ./loading.jl:488
 in macro expansion; at /mnt/resource/analytics/doug/.julia/v0.5/DataFrames/src/DataFrames.jl:130 [inlined]
 in anonymous at ./<missing>:?
 in include_from_node1(::String) at ./loading.jl:488
 in macro expansion; at ./none:2 [inlined]
 in anonymous at ./<missing>:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in process_options(::Base.JLOptions) at ./client.jl:239
 in _start() at ./client.jl:318
while loading /mnt/resource/analytics/doug/.julia/v0.5/DataFrames/src/other/utils.jl, in expression starting on line 147
while loading /mnt/resource/analytics/doug/.julia/v0.5/DataFrames/src/DataFrames.jl, in expression starting on line 99
ERROR: Failed to precompile DataFrames to /mnt/resource/analytics/doug/.julia/lib/v0.5/DataFrames.ji.
 in compilecache(::String) at ./loading.jl:593
 in require(::Symbol) at ./loading.jl:422
```

nalimilan:
@dmbates The problem is with your `nl/nullable` branch in DataFrames. Try with `git reset --hard origin` (be careful to do `git checkout nl/nullable` first).

quinnj:
I think we're good to go here. Any last objections to merging?

quinnj:
The deed is done.

tlnagy:
Gadfly tests don't pass with this branch checked out:

```
julia> Pkg.checkout("DataFrames", "nl/nullable")
INFO: Checking out DataFrames nl/nullable...
INFO: Pulling DataFrames latest nl/nullable...
INFO: Upgrading CategoricalArrays: v0.0.4 => v0.0.6
INFO: Downgrading DataStreams: v0.0.15 => v0.0.14
INFO: Building Rmath

julia> Pkg.test("Gadfly")
INFO: Computing test dependencies for Gadfly...
INFO: No packages to install, update or remove
INFO: Testing Gadfly
WARNING: Base.LambdaStaticData is deprecated, use LambdaInfo instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Utilities.jl:122
WARNING: symbol is deprecated, use Symbol instead.
 in depwarn(::String, ::Symbol) at ./deprecated.jl:64
 in symbol(::SubString{String}, ::Vararg{SubString{String},N}) at ./deprecated.jl:30
 in collect(::Base.Generator{Array{SubString{String},1},Docile.Utilities.##2#4}) at ./array.jl:307
 in @H_str(::Any) at /Users/tamasnagy/.julia/v0.5/Docile/src/Utilities.jl:17
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in macro expansion; at ./none:2 [inlined]
 in anonymous at ./<missing>:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Any) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in process_options(::Base.JLOptions) at ./client.jl:239
 in _start() at ./client.jl:318
 in _start() at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
while loading /Users/tamasnagy/.julia/v0.5/Docile/src/Runner/lookup.jl, in expression starting on line 95
WARNING: symbol is deprecated, use Symbol instead.
 in depwarn(::String, ::Symbol) at ./deprecated.jl:64
 in symbol(::SubString{String}, ::Vararg{SubString{String},N}) at ./deprecated.jl:30
 in collect_to!(::Array{DataType,1}, ::Base.Generator{Array{SubString{String},1},Docile.Utilities.##2#4}, ::Int64, ::Int64) at ./array.jl:340
 in collect(::Base.Generator{Array{SubString{String},1},Docile.Utilities.##2#4}) at ./array.jl:308
 in @H_str(::Any) at /Users/tamasnagy/.julia/v0.5/Docile/src/Utilities.jl:17
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in macro expansion; at ./none:2 [inlined]
 in anonymous at ./<missing>:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Any) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in process_options(::Base.JLOptions) at ./client.jl:239
 in _start() at ./client.jl:318
 in _start() at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
while loading /Users/tamasnagy/.julia/v0.5/Docile/src/Runner/lookup.jl, in expression starting on line 122
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/utilities.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/utilities.jl:481
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.writemime is deprecated.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:49

WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.

WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.

WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:480
WARNING: Base.writemime is deprecated.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Collector/types.jl:151
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/types.jl:14
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/types.jl:38
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/interface.jl:1
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/interface.jl:1
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/interface.jl:1
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/interface.jl:1
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/interface.jl:1
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Cache/interface.jl:1
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Legacy/types.jl:53
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Legacy/types.jl:53
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Legacy/types.jl:53
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/tamasnagy/.julia/v0.5/Docile/src/Legacy/types.jl:53
WARNING: symbol is deprecated, use Symbol instead.
 in depwarn(::String, ::Symbol) at ./deprecated.jl:64
 in symbol(::String, ::Vararg{String,N}) at ./deprecated.jl:30
 in @META_str(::String, ::Vararg{String,N}) at /Users/tamasnagy/.julia/v0.5/Docile/src/Formats/metaparse.jl:52
 in docm(::String, ::Expr, ::Bool) at ./docs/Docs.jl:605 (repeats 2 times)
 in @doc(::String, ::Vararg{Any,N}) at ./boot.jl:337
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in macro expansion; at ./none:2 [inlined]
 in anonymous at ./<missing>:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Any) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in process_options(::Base.JLOptions) at ./client.jl:239
 in _start() at ./client.jl:318
 in _start() at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
while loading /Users/tamasnagy/.julia/v0.5/Docile/src/Extensions/Extensions.jl, in expression starting on line 16
WARNING: Method definition ==(Base.Nullable{#S1<:Any}, Base.Nullable{#S2<:Any}) in module NullableArrays at /Users/tamasnagy/.julia/v0.5/NullableArrays/src/operators.jl:121 overwritten in module DataFrames at /Users/tamasnagy/.julia/v0.5/DataFrames/src/DataFrames.jl:107.
WARNING: Method definition ==(Base.Nullable, Base.Nullable) in module Base at nullable.jl:76 overwritten in module DataFrames at /Users/tamasnagy/.julia/v0.5/DataFrames/src/DataFrames.jl:117.
ERROR: LoadError: LoadError: UndefVarError: NominalArray not defined
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in macro expansion; at /Users/tamasnagy/.julia/v0.5/DataFrames/src/DataFrames.jl:170 [inlined]
 in anonymous at ./<missing>:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in macro expansion; at ./none:2 [inlined]
 in anonymous at ./<missing>:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Any) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in process_options(::Base.JLOptions) at ./client.jl:239
 in _start() at ./client.jl:318
 in _start() at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
while loading /Users/tamasnagy/.julia/v0.5/DataFrames/src/other/utils.jl, in expression starting on line 147
while loading /Users/tamasnagy/.julia/v0.5/DataFrames/src/DataFrames.jl, in expression starting on line 140
ERROR: LoadError: LoadError: Failed to precompile DataFrames to /Users/tamasnagy/.julia/lib/v0.5/DataFrames.ji.
 in compilecache(::String) at ./loading.jl:593
 in require(::Symbol) at ./loading.jl:393
 in require(::Symbol) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Any) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in require(::Symbol) at ./loading.jl:415
 in require(::Symbol) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in include_from_node1(::String) at ./loading.jl:488
 in include_from_node1(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in process_options(::Base.JLOptions) at ./client.jl:262
 in _start() at ./client.jl:318
 in _start() at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
while loading /Users/tamasnagy/.julia/v0.5/RDatasets/src/RDatasets.jl, in expression starting on line 5
while loading /Users/tamasnagy/.julia/v0.5/Gadfly/test/runtests.jl, in expression starting on line 3
===================================================[ ERROR: Gadfly ]====================================================

failed process: Process(`/Applications/Julia-0.5.app/Contents/Resources/julia/bin/julia -Ccore2 -J/Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib --compile=yes --depwarn=yes --check-bounds=yes --code-coverage=none --color=yes --compilecache=yes /Users/tamasnagy/.julia/v0.5/Gadfly/test/runtests.jl`, ProcessExited(1)) [1]

========================================================================================================================
INFO: No packages to install, update or remove
ERROR: Gadfly had test errors
 in #test#61(::Bool, ::Function, ::Array{AbstractString,1}) at ./pkg/entry.jl:740
 in (::Base.Pkg.Entry.#kw##test)(::Array{Any,1}, ::Base.Pkg.Entry.#test, ::Array{AbstractString,1}) at ./<missing>:0
 in (::Base.Pkg.Dir.##2#3{Array{Any,1},Base.Pkg.Entry.#test,Tuple{Array{AbstractString,1}}})() at ./pkg/dir.jl:31
 in cd(::Base.Pkg.Dir.##2#3{Array{Any,1},Base.Pkg.Entry.#test,Tuple{Array{AbstractString,1}}}, ::String) at ./file.jl:59
 in #cd#1(::Array{Any,1}, ::Function, ::Function, ::Array{AbstractString,1}, ::Vararg{Array{AbstractString,1},N}) at ./pkg/dir.jl:31
 in (::Base.Pkg.Dir.#kw##cd)(::Array{Any,1}, ::Base.Pkg.Dir.#cd, ::Function, ::Array{AbstractString,1}, ::Vararg{Array{AbstractString,1},N}) at ./<missing>:0
 in #test#3(::Bool, ::Function, ::String, ::Vararg{String,N}) at ./pkg/pkg.jl:258
 in test(::String, ::Vararg{String,N}) at ./pkg/pkg.jl:258
```

nalimilan:
@tlnagy I suspect you still have an old version of this branch. Anyway, now it doesn't exist anymore: use the master branch instead.

That said, Gadfly will probably need some changes. Please file issues there, linking back to this one so that we can track them.

davidanthoff:
There will probably be dozens if not 100s of packages that will need updates to work with this new version of ``DataFrames``. Next step is to send out an email on julia-users that people should make their packages work with ``DataFrames`` master, right?

I think it would be really helpful if there was a firm timeline that indicates when these changes in ``DataFrames`` will show up in a tagged version. Maybe 1 month? And maybe the email that anounces this change could literally say "We will tag DataFrames with NullableArrays support on 10/26, if you don't deal with this until then, your package will break."

nalimilan:
Yes. One month sounds like a reasonable time frame, but I wouldn't give a precise deadline.

BTW, it would be great if you @davidanthoff and @davidagold could add instructions to the manual about how to use Query.jl and StructureQuery.jl with data frames. I feel like many of the current examples should be replaced with these more powerful frameworks.

davidanthoff:
I think the main issue with no timeline is that as a package developer I would want to know at which point folks will have the new ``DataFrames`` bits. For example, say we send out an announcement without a timeline, then I as a package dev might enthusiastically upgrade my package on my master to work with the new version. But if it then takes 8 weeks until ``DataFrames`` get tagged, I might get really annoyed, because essentially I can't tag a new version of my package until ``DataFrames`` has been tagged. Much easier for me to plan things if I know an exact timeline for ``DataFrames``.

I'm just finishing up the last bits of documentation for [Query](https://github.com/davidanthoff/Query.jl), and then I'll announce it widely, and yes, happy to add a link to the ``DataFrames`` documentation at that point. You can take a look at the current work-in-progress documentation [here](http://www.david-anthoff.com/Query.jl/latest/).

ararslan:
I'm with Milan regarding the deadline. If it were me, I would just say,
> Hey everybody, DataFrames master now uses NullableArrays rather than DataArrays. Check out master, play around with it, report any issues, and in about a month we'll tag a new release of DataFrames. We anticipate this to be a breaking change, so if you haven't modified your DataFrames-dependent code to work with this update by then, you may have to set an upper bound on DataFrames in your package's REQUIRE file.

Gives folks an estimate of a timeframe without holding us to a hard deadline, plus it encourages use and tells people how to deal with the change if they don't adapt.

quinnj:
I think we can give a pretty good idea of when a new release will go out. Even if we tell users that it'll be 2-4 weeks out, that seems reasonable.

I'm most interested in getting an actual feel for breakage in the package ecosystem. If we had a better idea of what all might break, that would help focus efforts and we could even pro-actively do upper bounds on packages that might not be ready in time.

ararslan:
@tkelman Is there a way we could check out DataFrames master and run PkgEval to see what all this will break?

quinnj:
I actually pinged Tony about this already, he said

> probably the easiest thing to do is make a metadata fork that has a test tag, and modify pkgeval to use that metadata fork for a test run

ararslan:
Ah, okay. Sounds like a good plan. (Sorry for the unnecessary second ping, Tony!)

davidanthoff:
I assume the version with NullableArrays will be 0.9.0? I just asking because then I can put the correct upper limits on some of my packages already that don't support the NullableArrays stuff.

nalimilan:
Yeah, that sounds like the best version number. I don't think anybody will make the case that we are 1.0 already. :-)

ViralBShah:
Now that this is merged, is it worthwhile to have an announcement on the Julia blog?

nalimilan:
That would be nice, but I'm not sure we have a sexy enough story to tell users yet. Maybe wait until we can show a user-friendly API with query macros?

alyst:
I'm working on rebasing #850 and updating `RData.jl`, I think that helps fine-tuning what nullable/categorical arrays provide. I don't know whether to postpone the announcement until the new query macros are ready, but waiting for a few key packages to be updated seems reasonable.

ViralBShah:
Sounds like a plan.

davidanthoff:
I would wait with a blog until a) this is tagged (a couple of weeks away) and b) the the rest of the package scene has caught up. It would be really bad PR if there was a blog post saying "cool new stuff", and then people try it and nothing works. 

I could write a blog post about Query.jl now, though. It is ready and works with the existing DataFrames story.

davidanthoff:
I now have a branch in [Query.jl](https://github.com/davidanthoff/Query.jl) that works with the current ``DataFrames`` ``master``, i.e. with the ``NullableArrays`` version. See davidanthoff/Query.jl#62.

nalimilan:
@davidanthoff How does Query.jl work as regards `Nullable` lifting? Can you e.g. compute a mean of a `NullableArray` conveniently? Select observations based on conditions on `NullableArrays`?

nalimilan:
BTW, there's now a [blog post on StructuredQueries.jlg](http://julialang.org/blog/2016/10/StructuredQueries), but it's more about the design than a simple introduction for users. We also need to address the issue of lifting to offer a decent replacement for the current DataFrames features.

davidagold:
StructuredQueries (SQ) occupies a strange position in that it provides a user-facing interface but no backend support. The former is intended to be extended by packages that define a particular semantics. I don't intend for users to do `using StructuredQueries`. I expect for them to do, say `using X`, where X reexports SQ and implements a backend support ("collection machinery" in my lingo). I'm still working out how this will look in practice. I think if users want a ready-to-go data manipulation framework *right now*, they're better off using DataFramesMeta or Query.

johnmyleswhite:
Can we plan out a timeline for tagging this release? Given the roughness of our API offering, can we hold off for, say, 2-3 months? I think this is clearly the future, but I think we want to make sure that we're ready to tell everyone to query Query.jl or DataFramesMeta.jl or StructuredQueries.jl.

davidanthoff:
@nalimilan I don't want Query to provide any special lifting support, it is meant to just pick up whatever there is in base. Right now I [define](https://github.com/davidanthoff/Query.jl/blob/master/src/operators.jl) a whole bunch of methods that provide lifted implementations for things like ``+`` etc., but I want to remove those in the future once we have that in base.

tkelman:
Which is known as type piracy and changes the behavior of unrelated code. Packages should not be doing that.

davidanthoff:
@tkelman Totally agreed, that stuff should all be moved to base. Both NullableArrays and Query are guilty of this right now, as far as I can tell.

nalimilan:
Yeah, we need to move these into Base, and have a package for 0.4 and 0.5 which only provides lifted operators, so that it's clear to the user what happens.

But without automatic lifting for other functions, I'm afraid working with Query.jl is going to be quite painful. Time will tell...

davidanthoff:
> Yeah, we need to move these into Base, and have a package for 0.4 and 0.5 which only provides lifted operators, so that it's clear to the user what happens.

Could we just back port these things to 0.5, and just not support this stuff on 0.4?

tkelman:
We don't backport behavior changes, only bug fixes.

johnmyleswhite:
I entirely agree that we should not bother supporting 0.4.

ararslan:
DataFrames master doesn't, so we can't anyway. ¯\\\_(ツ)\_/¯

davidanthoff:
Who is going to create the package that hosts the nullable methods for julia 0.5? Is it going to be hosted under JuliaStats?

Probably easier if we start with adding these methods in the package, and then once that is done we can move them all into julia base master in one go, right?

Would be great if someone could create the package so that we can start contributing, I'd be eager to get this moving.

ararslan:
As Tony said, packages shouldn't be engaging in type piracy, so perhaps we should just starting making base PRs?

davidanthoff:
@ararslan I believe there was agreement (including from @tkelman) that in this case, for julia 0.5 compat only, there would be one package that would have lifted methods and engage in type piracy. That package would only be used on julia 0.5, and the same methods would then be included in julia base for 0.6.

Essentially that package would just get all the type piracy that is currently going on in NullableArrays and in Query.

If we don't do that, I really don't see how we can possibly tag the new DataFrames before julia 0.6 is released.

ararslan:
Oh sorry, I didn't realize that such an agreement had been reached.

Looks like this is what we're after, then: https://github.com/davidagold/NullableOps.jl

nalimilan:
> Probably easier if we start with adding these methods in the package, and then once that is done we can move them all into julia base master in one go, right?

AFAIK everything is done since I filed https://github.com/JuliaLang/julia/pull/18758. I just need to open a new PR with non-boolean operators and see how the discussion goes.

davidagold:
@nalimilan I don't think that's the PR you mean. Where's your PR that introduces lifted operators?

nalimilan:
Indeed, I've fixed the link!

davidanthoff:
I'm currently working on a macro that might make adding these lifted methods much easier. The general idea is this:
````julia
@lift function log(x)
````
would for example add a lifted method. The whole thing would work much more general:
````julia
@lift function +(x::Number, y::Number)
@lift function +{T<:Number}(x::T,y::T)
```
etc. should all work.

This macro could be used to then implement another macro called ``@lifted`` that could be used like the ``@generated`` macro to easily add lifted versions of a function at the function def site:
````julia
@lifted foo(x::Float64)
    return blabla
end
````
would automatically add lifted methods to ``foo``.

It is not done, I'm just mentioning this here in case someone else is working on a similar thing, then we should merge effort.

tkelman:
Doing lifting through selective manual method extension seems like you're repeating the dot-operator mistakes. It's also fragile due to the global side-effects of modifying method tables. The approach taken in https://github.com/JuliaLang/julia/pull/18758 seems safer, more general, and will have more localized effects.

Wilfred:
This was merged in September 2016, but as far as I can see version 0.10.1 (August 2017) doesn't have this. Is there a plan for releasing?

nalimilan:
This PR has actually been merged, then moved to DataTables, and then replaced with a port to Nulls.jl before being "backported" to DataFrames. See [this Discourse post](https://discourse.julialang.org/t/announcement-an-update-on-dataframes-future-plans/4646) for the full story. We're currently preparing a release.

PR #1013, opened 2016-07-15
Use vec to avoid indexing with a matrix

ararslan:
This fixes an issue with a test in StatsBase wherein `Bool[x != 0 for x in sum(ff, 2)]` returns a vector on 0.4 but a matrix on 0.5. Using `vec` ensures that we always get a vector here, as desired.


ararslan:
Closing in favor of #1014. Andreas is awesome.

PR #1015, opened 2016-07-19
Versioning specificity

tkelman:


ararslan:
Looks like the OS X nightly failure is https://github.com/JuliaLang/julia/issues/17484 and the AppVeyor failure is because Rmath and Windows are mortal enemies.

Thanks for doing this, looks good to me.

tkelman:
Rmath on Windows is next on my list, really.

src/abstractdataframe/io.jl
```diff
@@ -8,7 +8,7 @@ function escapedprint(io::IO, x::Any, escapes::AbstractString)
     print(io, x)
 end
 
-@static if VERSION < v"0.5.0-"
+@static if VERSION < v"0.5.0-dev+4354"
```
bicycle1885: Is this `@static` macro call needed at the top level?

tkelman: no, probably not

PR #1017, opened 2016-07-19
Clean up and document ModelMatrix logic

dmbates:
Cf. https://github.com/JuliaStats/DataFrames.jl/issues/1012.

### DIFF
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -27,7 +27,8 @@ end
 type Terms
     terms::Vector
     eterms::Vector        # evaluation terms
-    factors::Matrix{Int8} # maps terms to evaluation terms
+    factors::Matrix{Bool} # maps terms to evaluation terms
+# order can probably be dropped.  It is vec(sum(factors, 1))
     order::Vector{Int}    # orders of rhs terms
     response::Bool        # indicator of a response, which is eterms[1] if present
     intercept::Bool       # is there an intercept column in the model matrix?
@@ -200,20 +201,8 @@ function Terms(f::Formula)
     end
     ev = unique(vcat(etrms...))
     sets = [Set(x) for x in etrms]
-    facs = Int8[t in s for t in ev, s in sets]
+    facs = Bool[t in s for t in ev, s in sets]
     Terms(tt, ev, facs, oo, haslhs, !any(noint))
-end
-
-function remove_response(t::Terms)
-    # shallow copy original terms
-    t = Terms(t.terms, t.eterms, t.factors, t.order, t.response, t.intercept)
-    if t.response
-        t.order = t.order[2:end]
-        t.eterms = t.eterms[2:end]
-        t.factors = t.factors[2:end, 2:end]
-        t.response = false
-    end
-    return t
 end
 
 ## Default NA handler.  Others can be added as keyword arguments
@@ -248,7 +237,7 @@ ModelFrame(ex::Expr, d::AbstractDataFrame) = ModelFrame(Formula(ex), d)
 
 function StatsBase.model_response(mf::ModelFrame)
     mf.terms.response || error("Model formula one-sided")
-    convert(Array, mf.df[round(Bool, mf.terms.factors[:, 1])][:, 1])
+    convert(Array, mf.df[mf.terms.eterms[1]])
 end
 
 function contr_treatment(n::Integer, contrasts::Bool, sparse::Bool, base::Integer)
@@ -261,7 +250,7 @@ end
 contr_treatment(n::Integer,contrasts::Bool,sparse::Bool) = contr_treatment(n,contrasts,sparse,1)
 contr_treatment(n::Integer,contrasts::Bool) = contr_treatment(n,contrasts,false,1)
 contr_treatment(n::Integer) = contr_treatment(n,true,false,1)
-cols(v::PooledDataVector) = contr_treatment(length(v.pool))[v.refs,:]
+cols(v::PooledDataVector) = contr_treatment(length(v.pool))[v.refs, :]
 cols(v::DataVector) = hcat(convert(Vector{Float64}, v.data))
 cols(v::Vector) = hcat(convert(Vector{Float64}, v))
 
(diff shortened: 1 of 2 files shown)


src/statsmodels/formula.jl
```diff
@@ -246,78 +235,145 @@ end
 ModelFrame(f::Formula, d::AbstractDataFrame) = ModelFrame(Terms(f), d)
 ModelFrame(ex::Expr, d::AbstractDataFrame) = ModelFrame(Formula(ex), d)
 
+asMatrix(a::AbstractMatrix) = a
+asMatrix(v::AbstractVector) = reshape(v, (length(v), 1))
+
+"""
+    StatsBase.model_response(mf::ModelFrame)
+Extract the response column, if present.  `DataVector` or
```
ararslan: Need a line break after the indented line for the markdown to render properly

src/statsmodels/formula.jl
```diff
@@ -246,78 +235,145 @@ end
...
+    end
 end
 
+"""
+    contr_treatment(n::Integer, contrasts::Bool, sparse::Bool, base::Integer)
+Create a sparse or dense identity of size `n`.  Return the identity if `contrasts`
+is false.  Otherwise drop the `base` column.
+"""
 function contr_treatment(n::Integer, contrasts::Bool, sparse::Bool, base::Integer)
     if n < 2 error("not enought degrees of freedom to define contrasts") end
```
ararslan: It'd be nice to use explicitly typed exceptions, e.g.
```julia
n < 2 && throw(ArgumentError("Not enough degrees of freedom to define contrasts"))
```
Also you have a trailing "t" on "enough."

src/statsmodels/formula.jl
```diff
@@ -246,78 +235,145 @@ end
...
     contr = sparse ? speye(n) : eye(n) .== 1.
-    if !contrasts return contr end
-    if !(1 <= base <= n) error("base = $base is not allowed for n = $n") end
-    contr[:,vcat(1:(base-1),(base+1):end)]
+    if !contrasts
+        contr
+    elseif !(1 <= base <= n)
+        error("base = $base is not allowed for n = $n")
```
ararslan: Could use `ArgumentError` here too

src/statsmodels/formula.jl
```diff
@@ -246,78 +235,145 @@ end
 ModelFrame(f::Formula, d::AbstractDataFrame) = ModelFrame(Terms(f), d)
 ModelFrame(ex::Expr, d::AbstractDataFrame) = ModelFrame(Formula(ex), d)
 
+asMatrix(a::AbstractMatrix) = a
```
ararslan: This doesn't follow the Julia function naming convention of all lowercase or lower snake case.

src/statsmodels/formula.jl
```diff
@@ -246,78 +235,145 @@ end
...
+cols(v::Vector) = asMatrix(convert(Vector{Float64}, v))
+
+"""
+    expandcols(trm::Vector)
+Create pairwise products of columns from a vector of matrices
+"""
 function expandcols(trm::Vector)
     if length(trm) == 1
-        return convert(Array{Float64}, trm[1])
+        asMatrix(convert(Array{Float64}, trm[1]))
```
ararslan: `trm` is already a vector, why convert to an array?

src/statsmodels/formula.jl
```diff
@@ -246,78 +235,145 @@ end
...
     end
-    n
 end
 
+"""
+    dropResponse(trms::Terms)
+Drop the response term, `trms.eterms[1]` and the first row and column
+of `trms.factors` if `trms.response` is true.
+"""
+dropResponse(trms::Terms) = !trms.response ? trms :
```
ararslan: `dropresponse` or `drop_response`

ararslan: Why not just modify `remove_response`?

src/statsmodels/formula.jl
```diff
@@ -246,78 +235,145 @@ end
...
-    n = 1
-    for x in trm
-        n *= size(x, 2)
+"""
+    dropRanefTerms(trms::Terms)
+Expressions of the form `(a|b)` are "random-effects" terms and are not
+incorporated in the ModelMatrix.  This function checks for such terms and,
+if any are present, drops them from the `Terms` object.
+"""
+function dropRanefTerms(trms::Terms)
```
ararslan: Besides the camel case, this function name is kind of confusing. Even `droprandomeffects` would be better IMO.

ararslan:
I dunno why Travis is mad about `view`. If it's been a while you might want to rebase from master.

src/statsmodels/formula.jl
dmbates: Actually you don't need a line break.  A line with a 4 space indent followed by a line with no indent causes a new markdown block.

src/statsmodels/formula.jl
dmbates: Thanks for catching that.

By the way, I think the
```
    if condition
        throw(ArgumentError(...))
    end
```
is preferred to the more compact
```
    condition && throw(...
```
for checking code coverage.  If the check and error throwing are on the same line you can't tell if the error is ever thrown in the tests.

src/statsmodels/formula.jl
dmbates: Thanks.

src/statsmodels/formula.jl
dmbates: `CategoricalArrays` has also used the `drop` verb in `droplevels!` so I will use `dropresponse!`.  It should have the `!` in the name.

src/statsmodels/formula.jl
ararslan: Oh neat I didn't know that! Thanks for letting me know. :)

ararslan:
This is great work. Thanks so much for doing this, Doug! You rock! 🎸 :metal: 👏

src/statsmodels/formula.jl
ararslan: Oh that's good to know, thanks.

kleinschmidt:
This is great, thanks for doing it Doug. I think the only place where there's substantive overlap with #870 is in the ModelMatrix code itself. In order to construct columns for categorical variables, it's necessary to check whether any lower-order terms are present (e.g., an intercept column for a main effect `A`, or a main effect `A` for an interaction `A&B`). But I think that should be pretty simple to rebase on top of this PR.

dmbates:
@ararslan Thanks.  I appreciate your careful reading of the changes.

@kleinschmidt I like your approach to allowing more general contrasts.  I can look into merging those changes in the next couple of days.  It would be worthwhile refactoring Terms in general.  I unabashedly patterned it after the R terms object but I think it would be much easier to do better than that now that I understand Julia a bit better.

The Appveyor failure seems unrelated to this PR.  It failed on only one of the 3 configurations and that was in the tests for RDA.

I merge this tomorrow unless someone else gets there ahead of me.

kleinschmidt:
You might also be interested in looking at https://github.com/kleinschmidt/Formulas.jl

I played around a bit with a representation of terms that uses type parameters and multiple dispatch. It substantially simplifies the rewrite rules. The code is pretty ugly as it stands (lots of splatting) but I think it's a clearer and more expressive approach than the R-style one.

src/statsmodels/formula.jl
nalimilan: Still, please add a line break for consistency with other docstrings, in particular in Julia Base.

src/statsmodels/formula.jl
kleinschmidt: I find the `!` confusing. I thought it was idiomatically used for things that modified their first argument, but `dropresponse!` doesn't (it returns a new `Terms` instance).

nalimilan:
@dmbates You merged without fixing some naming issues highlighted during the review. Could you fix them in another PR?

nalimilan:
Nevermind, got confused while viewing an older version of the diff.

tkelman:
This was just released in a tagged version yesterday, and I think it caused problems in StatsBase, FixedEffectModels, and MixedModels
http://pkg.julialang.org/logs/StatsBase_0.4.log
http://pkg.julialang.org/logs/FixedEffectModels_0.4.log
http://pkg.julialang.org/logs/MixedModels_0.4.log

ararslan:
Actually I think it was #870 that caused the issues, at least the one in StatsBase.

tkelman:
Whoops. I'll leave it to people who know what they're talking about. Do we want to do the same thing we did with Distributions, and make a new less-breaking patch tag off a branch from before either or both of these large recent changes and with just the deprecation fixes cherry-picked?

ararslan:
In the interest of fire extinguishing, that's probably a good idea. (If we're quick we could even get to it before PkgEvaluator does.)

tkelman:
We've got an hour before it starts, though I can tweak my cronjob and push it back a bit if that's too tight.

ararslan:
Shouldn't be necessary. https://github.com/JuliaLang/METADATA.jl/pull/5893

matthieugomez:
This commit introduced a pretty bad error in my package (i.e. a function in my package would return a wrong numerical result). It was caught by the tests in my package as noticed by @tkelman , althought it was hidden by another test failure due to #1035.

The issue is that ModelMatrix becomes view on the underlying dataframe in the case where is only one vector
```julia
using RDatasets
df = dataset("plm", "Cigar")
mf = ModelFrame(Sales ~ 0 + Price, df)
X = ModelMatrix(mf).m
fill!(X, 0.0)
mf.df[:Price]
```

nalimilan:
@matthieugomez So this could simply be fixed by ensuring we copy the column? Would you make a PR?

PR #975, opened 2016-05-24
Docile is not needed with julia 0.4

dmbates:
Now that the package depends on julia 0.4, Docile is no longer needed.  In Julia v0.5-dev Docile causes many warnings to be printed.

PR #1027, opened 2016-08-03
Remove dependency on Docile

davidanthoff:
This PR replaces #975. It is minimal, and should merge cleanly into current master.

It would be great if this could be merged quickly and then a new version would be tagged, to get the deprecation warning story on julia 0.5 in order.


ararslan:
Windows sure hates that RDA stuff. Failures unrelated. Thanks!

PR #1030, opened 2016-08-09
Use cleaned version of terms, not original

dmbates:
Some whitespace changes crept in here too.  The only non-whitespace change is using `terms.terms`, not `mf.terms.terms`.  The local variable `terms` has random-effects and response terms dropped.


ararslan:
Looks good to me, pending green CI.

dmbates:
Looks like the appveyor builds are failing in RDA.jl and are unrelated to these changes.

ararslan:
Yep. Looking forward to getting RDA out of here.

kleinschmidt:
Sorry for the whitespace mess and thanks for catching the `mf.terms`. Probably should also add some tests with random effects to make sure that they're dropped as expected.

nalimilan:
Could you rebase (and merge) this?

PR #1034, opened 2016-08-11
In coefnames, iterate over the terms after dropping, not the full set

ararslan:
This fixes and tests an issue that showed up in MixedModels, wherein `coefnames` was attempting to index into an empty array because it was iterating over the full model frame terms, not the terms after dropping the response and random effects.


ararslan:
As usual, AppVeyor failures are RDA.

kleinschmidt:
Looks good to me.

PR #1036, opened 2016-08-12
Address tty_size deprecation warning

AndyGreenwell:
Replacing call to tty_size with displaysize(io) to address deprecation warning for Julia 0.5.  This call is triggered when displaying a DataFrame in a Jupyter notebook via IJulia.


ararslan:
As usual, the AppVeyor failures are unrelated (RDA). Thanks, looks good to me!

PR #1038, opened 2016-08-17
Make ModelMatrix throw an error when it can't be constructed

### LINKED ISSUE #1035: ModelMatrix fails for mixed effects models without an intercept
Currently if you construct a mixed effects model with no other terms or an intercept, `reduce` throws an error. For example,
```julia
df = DataFrame(x = [1,2,3], y = [4,5,6])
mf = ModelFrame(y ~ 0 + (1 | x), df)
ModelMatrix(mf)
```
currently produces
```julia
 in _mapreduce(::Base.#identity, ::Base.#hcat, ::Base.LinearFast, ::Array{Array{Float64,2},1}) at ./reduce.jl:148
 in DataFrames.ModelMatrix{T<:Union{Float32,Float64}}(::DataFrames.ModelFrame) at /Users/ararslan/.julia/DataFrames/src/statsmodels/formula.jl:482
```
This happens because when you get the set of terms as `droprandomeffects(dropresponse!(mf))` and there's no intercept, the current logic still lets the case get through to the end, but without adding anything to a vector of blocks, which gets `reduce`d at the end.

Before @dmbates' overhaul of model matrices, the above code produced `DataFrames.ModelMatrix{Float64}(,Int64[])`.

I'm not sure what we actually want to do in this situation since there isn't really a model matrix to construct, but if it should be an error then we should make it an explicit one.

Side note: It appears a test in FixedEffectModels relies on something like this working. See [here](https://github.com/matthieugomez/FixedEffectModels.jl/blob/master/test/reg.jl#L38)

ararslan:
Fixes #1035


ararslan:
AppVeyor failure is RDA, as usual.

src/statsmodels/formula.jl
```diff
@@ -553,7 +557,7 @@ function coefnames(mf::ModelFrame)
         push!(term_names, expandtermnames(names))
     end
 
-    reduce(vcat, term_names)
+    isempty(term_names) ? Vector{Compat.UTF8String}() : reduce(vcat, term_names)
```
andreasnoack: Why not just `String` instead of `Compat.UTF8String` here? What are the possible output types here? Is it always an array of strings? In that case, it might be cleaner with just `reduce(vcat, String[], term_names)` which doesn't fail on empty input.

ararslan: Because `Compat.String` is `ByteString` on 0.4, whereas `Compat.UTF8String` is `UTF8String` on 0.4 and `String` on 0.5. If I'm not mistaken, the only possible output type from `coefnames` is `Vector{Compat.UTF8String}`, which is why I did that here. I like your `reduce` suggestion, I'll implement that later today (unless someone wants to push that to this branch in the meantime).

Issue #614, opened 2014-06-06
Sparse Model.Matrix 


gragusa:
Are there any plans to make Model.Matrix return a Sparse matrix? Something similar to ```sparse.model.matrix``` implemented in the R package ```Matrix``` written by @dmbates. I wanted to give it a try, but I wanted to know first whether there are issues blocking such an attempt.

GordStephen:
Has this situation evolved in the past few years? This is something I'd be very interested in and would be happy to help contribute to.

nalimilan:
I don't think anybody has worked on it, but we'd be happy to help you if you want to make a pull request.

nalimilan:
As for the implementation/interface, I guess we should just make the `T` parameter of `ModelMatrix` include the container type and not only the element type. Then a default constructor would use `Matrix`, but you could also specify `SparseMatrixCSC`.

GordStephen:
Ok, thanks, I'll take a look and come back with any questions.

GordStephen:
Ok, thanks, I'll take a look and come back with any questions.

GordStephen:
Ok, thanks, I'll take a look and come back with any questions.

GordStephen:
OK, so if I'm understanding things properly, what would need to be done is:

1. Provide a sparse equivalent to [`cols`](https://github.com/JuliaStats/DataFrames.jl/blob/d3851e823ae08f928a7857db71e9910379d7080a/src/statsmodels/formula.jl#L264) - shouldn't be difficult since the existing `PooledDataArray` case just calls `contr_treatment`, which already has a sparse option, and the other cases are trivial
2. Provide a sparse equivalent to [`expandcols`](https://github.com/JuliaStats/DataFrames.jl/blob/d3851e823ae08f928a7857db71e9910379d7080a/src/statsmodels/formula.jl#L274). Also should be fairly straightforward.

With that, `hcat` should abstract away the rest. Does that seem sane?

nalimilan:
Actually, I would be a shame to base your work on the current code while there's a large rework in the tubes at https://github.com/JuliaStats/DataFrames.jl/pull/870. Better base your work on that PR. You should look at the `contrasts_matrix` functions, and add version taking a type parameter (see my previous comment).

> Provide a sparse equivalent to cols - shouldn't be difficult since the existing PooledDataArray case just calls contr_treatment, which already has a sparse option, and the other cases are trivial

[...]

PR #1040, opened 2016-08-20
RFC: Sparse ModelMatrix support

GordStephen:
As discussed in #614 - coming back to this now that #870 and #1017 have been merged.

### DIFF
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -374,12 +374,11 @@ end
     expandcols(trm::Vector)
 Create pairwise products of columns from a vector of matrices
 """
-function expandcols(trm::Vector)
+function expandcols{T<:ModelMatrixContainer}(trm::Vector{T})
     if length(trm) == 1
-        asmatrix(Matrix{Float64}, convert(Array{Float64}, trm[1]))
+        trm[1]
     else
-        a = convert(Array{Float64}, trm[1])
-        b = expandcols(trm[2 : end])
+        a, b = trm[1], expandcols(trm[2 : end])
         reduce(hcat, [broadcast(*, a, Compat.view(b, :, j)) for j in 1 : size(b, 2)])
     end
 end
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -352,7 +352,7 @@ modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::DataVector) = convert(T, res
 modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::Vector) = convert(T, reshape(v, length(v), 1))
 
 """
-    modelmat_cols(T::Type{AbstractFloatMatrix}, v::PooledDataVector, contrast::ContrastsMatrix)
+    modelmat_cols(::Type{T}, v::PooledDataVector, contrast::ContrastsMatrix)
 
 Construct `ModelMatrix` columns of type `T` based on specified contrasts, ensuring that
 levels align properly.
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -48,8 +48,8 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-type ModelMatrix{T <: @compat(Union{Float32, Float64})}
-    m::Matrix{T}
+type ModelMatrix{T <: @compat(Union{Matrix{Float32}, Matrix{Float64}, SparseMatrixCSC{Float32,Int}, SparseMatrixCSC{Float64,Int}})}
+    m::T
     assign::Vector{Int}
 end
 
@@ -479,7 +479,7 @@ function ModelMatrix(mf::ModelFrame)
         append!(assign, fill(i_term, size(blocks[end], 2)))
     end
 
-    ModelMatrix{Float64}(reduce(hcat, blocks), assign)
+    ModelMatrix{Matrix{Float64}}(reduce(hcat, blocks), assign)
 end
 
 
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -48,7 +48,11 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-type ModelMatrix{T <: @compat(Union{Matrix{Float32}, Matrix{Float64}, SparseMatrixCSC{Float32,Int}, SparseMatrixCSC{Float64,Int}})}
+modelmatrixcontainertypes = [Matrix{Float32}, Matrix{Float64},
+                         SparseMatrixCSC{Float32,Int},
+                         SparseMatrixCSC{Float64,Int}]
+
+type ModelMatrix{T <: Union{modelmatrixcontainertypes...}}
     m::T
     assign::Vector{Int}
 end
@@ -437,21 +441,21 @@ If there is an intercept in the model, that column occurs first and its
 Mixed-effects models include "random-effects" terms which are ignored when
 creating the model matrix.
 """
-function ModelMatrix(mf::ModelFrame)
+function ModelMatrix(T::Union{map(t->Type{t}, modelmatrixcontainertypes)...}, mf::ModelFrame)
     dfrm = mf.df
     terms = droprandomeffects(dropresponse!(mf.terms))
 
-    blocks = Matrix{Float64}[]
+    blocks = T[]
     assign = Int[]
     if terms.intercept
-        push!(blocks, ones(size(dfrm, 1), 1))  # columns of 1's is first block
-        push!(assign, 0)                        # this block corresponds to term zero
+        push!(blocks, convert(T, ones(size(dfrm, 1), 1)))  # columns of 1's is first block
+        push!(assign, 0)                                   # this block corresponds to term zero
     end
 
     factors = terms.factors
 
     ## Map eval. term name + redundancy bool to cached model matrix columns
-    eterm_cols = @compat Dict{Tuple{Symbol,Bool}, Array{Float64}}()
+    eterm_cols = @compat Dict{Tuple{Symbol,Bool}, T}()
     ## Accumulator for each term's vector of eval. term columns.
 
     ## TODO: this method makes multiple copies of the data in the ModelFrame:
@@ -462,7 +466,7 @@ function ModelMatrix(mf::ModelFrame)
     ## "promoted" full-rank versions of categorical columns for non-redundant
     ## eval. terms:
     for (i_term, term) in enumerate(terms.terms)
-        term_cols = Matrix{Float64}[]
+        term_cols = T[]
         ## Pull out the eval terms, and the non-redundancy flags for this term
         ff = Compat.view(factors, :, i_term)
         eterms = Compat.view(terms.eterms, ff)
@@ -479,8 +483,9 @@ function ModelMatrix(mf::ModelFrame)
         append!(assign, fill(i_term, size(blocks[end], 2)))
     end
 
-    ModelMatrix{Matrix{Float64}}(reduce(hcat, blocks), assign)
+    ModelMatrix{T}(reduce(hcat, blocks), assign)
 end
+ModelMatrix(mf::ModelFrame) = ModelMatrix(Matrix{Float64}, mf)
 
 
 """
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -352,7 +352,7 @@ modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::DataVector) = convert(T, res
 modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::Vector) = convert(T, reshape(v, length(v), 1))
 
 """
-    modelmat_cols(::Type{T}, v::PooledDataVector, contrast::ContrastsMatrix)
+    modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::PooledDataVector, contrast::ContrastsMatrix)
 
 Construct `ModelMatrix` columns of type `T` based on specified contrasts, ensuring that
 levels align properly.
@@ -368,7 +368,7 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::PooledDataVector, c
 end
 
 """
-    expandcols(trm::Vector)
+    expandcols{T<:AbstractFloatMatrix}(trm::Vector{T})
 Create pairwise products of columns from a vector of matrices
 """
 function expandcols{T<:AbstractFloatMatrix}(trm::Vector{T})
@@ -423,8 +423,9 @@ end
 
 
 """
-    ModelMatrix(mf::ModelFrame)
-Create a `ModelMatrix` from the `terms` and `df` members of `mf`
+    ModelMatrix{T<:AbstractFloatMatrix}(mf::ModelFrame)
+Create a `ModelMatrix` of type `T` (default `Matrix{Float64}`) from the
+`terms` and `df` members of `mf`.
 
 This is basically a map-reduce where terms are mapped to columns by `cols`
 and reduced by `hcat`.  During the collection of the columns the `assign`
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -48,9 +48,9 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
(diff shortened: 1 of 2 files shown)


src/statsmodels/formula.jl
```diff
@@ -48,8 +48,8 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-type ModelMatrix{T <: @compat(Union{Float32, Float64})}
-    m::Matrix{T}
+type ModelMatrix{T <: @compat(Union{Matrix{Float32}, Matrix{Float64}, SparseMatrixCSC{Float32,Int}, SparseMatrixCSC{Float64,Int}})}
```
ararslan: We could drop the `@compat` here since the package no longer supports Julia 0.3.

src/statsmodels/formula.jl
```diff
@@ -437,21 +441,21 @@ If there is an intercept in the model, that column occurs first and its
...
-        push!(assign, 0)                        # this block corresponds to term zero
+        push!(blocks, convert(T, ones(size(dfrm, 1), 1)))  # columns of 1's is first block
+        push!(assign, 0)                                   # this block corresponds to term zero
     end
 
     factors = terms.factors
 
     ## Map eval. term name + redundancy bool to cached model matrix columns
-    eterm_cols = @compat Dict{Tuple{Symbol,Bool}, Array{Float64}}()
+    eterm_cols = @compat Dict{Tuple{Symbol,Bool}, T}()
```
GordStephen: I couldn't find any reason not to restrict the Array dimension here. Did I miss something?

src/statsmodels/formula.jl
```diff
@@ -48,8 +48,12 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-type ModelMatrix{T <: @compat(Union{Float32, Float64})}
-    m::Matrix{T}
+modelmatrixcontainertypes = [Matrix{Float32}, Matrix{Float64},
```
nalimilan: Maybe we should accept any subtype of `Union{AbstractMatrix{Float32}, AbstractMatrix{Float32}}`? Or is there anything specific that only `Matrix` and `SparseMatrixCSC`?

src/statsmodels/formula.jl
```diff
@@ -437,21 +441,21 @@ If there is an intercept in the model, that column occurs first and its
...
     dfrm = mf.df
     terms = droprandomeffects(dropresponse!(mf.terms))
 
-    blocks = Matrix{Float64}[]
+    blocks = T[]
     assign = Int[]
     if terms.intercept
-        push!(blocks, ones(size(dfrm, 1), 1))  # columns of 1's is first block
-        push!(assign, 0)                        # this block corresponds to term zero
+        push!(blocks, convert(T, ones(size(dfrm, 1), 1)))  # columns of 1's is first block
```
nalimilan: Since `blocks` is already typed, conversion should happen automatically without changing anything AFAIK.

src/statsmodels/formula.jl
nalimilan: The only issue I can think of is the case where a single-column term would give a column vector instead of a one-column matrix. But conversion will probably happen automatically, and tests should catch this. Have you run the tests of GLM.jl on the modified package?

GordStephen:
Should the new constructor be `ModelMatrix{Matrix{Float64}(mf)` instead of `ModelMatrix(Matrix{Float64}, mf)`? The latter isn't type-stable (right?), but I wasn't sure how to implement the former.

src/statsmodels/formula.jl
GordStephen: From what I could tell `modelmat_cols` handles the matrix conversion. The GLM.jl tests pass as well, so hopefully this is OK.

src/statsmodels/formula.jl
GordStephen: Not that I'm aware of?

ararslan: Could we even skip the `Union` and do `AbstractMatrix{AbstractFloat}`?

nalimilan:
Both are type-stable, but the former is probably more idiomatic. You can implement it via the `(::Type{ModelMatrix{T}}){T<:AbstractMatrix{...}}(...) = ...` syntax.

src/statsmodels/formula.jl
nalimilan: No idea. @andreasnoack?

src/statsmodels/formula.jl
```diff
@@ -374,12 +374,11 @@ end
     expandcols(trm::Vector)
 Create pairwise products of columns from a vector of matrices
 """
-function expandcols(trm::Vector)
+function expandcols{T<:ModelMatrixContainer}(trm::Vector{T})
     if length(trm) == 1
-        asmatrix(Matrix{Float64}, convert(Array{Float64}, trm[1]))
+        trm[1]
     else
-        a = convert(Array{Float64}, trm[1])
```
GordStephen: As far as I can tell, the conversions here are redundant since elements of `trm` are always created by `modelmat_cols`, which already handles this?

src/statsmodels/formula.jl
```diff
@@ -339,33 +339,35 @@ function StatsBase.model_response(mf::ModelFrame)
...
-function modelmat_cols(v::PooledDataVector, contrast::ContrastsMatrix)
+function modelmat_cols{T<:ModelMatrixContainer}(::Type{T}, v::PooledDataVector, contrast::ContrastsMatrix)
     ## make sure the levels of the contrast matrix and the categorical data
     ## are the same by constructing a re-indexing vector. Indexing into
     ## reindex with v.refs will give the corresponding row number of the
     ## contrast matrix
     reindex = [findfirst(contrast.levels, l) for l in levels(v)]
-    return contrast.matrix[reindex[v.refs], :]
+    contrastmatrix = convert(T, contrast.matrix)
+    return contrastmatrix[reindex[v.refs], :]
```
GordStephen: This array creation can be extremely slow for sparse `T` and large datasets... but I'm not sure if there's a faster way to do it without creating dense columns first?

GordStephen:
Ok, this should be functionally complete now. I essentially just generalized the existing model matrix creation logic to abritrary types beyond `Matrix{Float64}`, of which `SparseMatrixCSC` is one possible option. There may be a more type-specific approach that's faster...  Not sure.

src/statsmodels/formula.jl
nalimilan: No idea. Why is it slow? Indexing rows shouldn't be a problem for sparse matrices AFAIK.

src/statsmodels/formula.jl
```diff
@@ -48,8 +48,10 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-type ModelMatrix{T <: @compat(Union{Float32, Float64})}
-    m::Matrix{T}
+typealias ModelMatrixContainer{T<:AbstractFloat} AbstractMatrix{T}
```
nalimilan: I think it would be clearer if it was called `AbstractFloatMatrix` or something like that.

src/statsmodels/formula.jl
```diff
@@ -321,8 +323,8 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
 end
 setcontrasts!(mf::ModelFrame; kwargs...) = setcontrasts!(mf, Dict(kwargs))
 
-asmatrix(a::AbstractMatrix) = a
-asmatrix(v::AbstractVector) = reshape(v, (length(v), 1))
+asmatrix(T::Type, a::AbstractMatrix) = convert(T, a)
```
nalimilan: Since this function is only used in `modelmat_cols`, probably better remove it altogether. You can call `reshape(a, size(a, 1), size(a, 2)))` instead, which works for both vectors and matrices.

src/statsmodels/formula.jl
```diff
@@ -337,45 +339,46 @@ function StatsBase.model_response(mf::ModelFrame)
...
 """
-function modelmat_cols(v::PooledDataVector, contrast::ContrastsMatrix)
+function modelmat_cols{T<:ModelMatrixContainer}(::Type{T}, v::PooledDataVector, contrast::ContrastsMatrix)
     ## make sure the levels of the contrast matrix and the categorical data
     ## are the same by constructing a re-indexing vector. Indexing into
     ## reindex with v.refs will give the corresponding row number of the
     ## contrast matrix
     reindex = [findfirst(contrast.levels, l) for l in levels(v)]
-    return contrast.matrix[reindex[v.refs], :]
+    contrastmatrix = convert(T, contrast.matrix)
```
nalimilan: In what cases can `T` be different from `contrast.matrix`'s type?

src/statsmodels/formula.jl
```diff
@@ -437,21 +440,22 @@ If there is an intercept in the model, that column occurs first and its
 Mixed-effects models include "random-effects" terms which are ignored when
 creating the model matrix.
 """
-function ModelMatrix(mf::ModelFrame)
+@compat function (::Type{ModelMatrix{T}}){T<:ModelMatrixContainer}(mf::ModelFrame)
+    sparsemm = T <: AbstractSparseMatrix
```
nalimilan: This variable doesn't seem to be used anywhere.

src/statsmodels/formula.jl
GordStephen: I'm not an expert on sparse matrix indexing, but it seems to spend a lot of time sorting... Truncated profile output from a million-row reference vector and 5-column constrast matrix:

```
 3594 ./event.jl:68; (::Base.REPL.##3#4{Base.REPL.REPLBackend})()
 3594 ./REPL.jl:95; macro expansion
  3594 ./REPL.jl:64; eval_user_input(::Any, ::Base.REPL.REPLBackend)
   3594 ./boot.jl:234; eval(::Module, ::Any)
    3594 ./<missing>:?; anonymous
     3594 ./profile.jl:16; macro expansion;
      3594 ./sparse/sparsematrix.jl:2099; getindex(::SparseMatrixCSC{Float64,Int64}, ::Array{Int64,1}, :...
       4    ./sparse/sparsematrix.jl:2437; getindex(::SparseMatrixCSC{Float64,Int64}, ::Array{Int64,1}, :...
        3 ./reduce.jl:371; extrema(::Array{Int64,1})
        1 ./reduce.jl:372; extrema(::Array{Int64,1})
       3590 ./sparse/sparsematrix.jl:2448; getindex(::SparseMatrixCSC{Float64,Int64}, ::Array{Int64,1}, :...
        3547 ./sparse/sparsematrix.jl:2422; getindex_general(::SparseMatrixCSC{Float64,Int64}, ::Array{In...
         1    ./sort.jl:451; #sortperm#11(::Base.Sort.QuickSortAlg, ::Function, ::Function...
         4    ./sort.jl:452; #sortperm#11(::Base.Sort.QuickSortAlg, ::Function, ::Function...
         3542 ./sort.jl:454; #sortperm#11(::Base.Sort.QuickSortAlg, ::Function, ::Function...
          3542 ./sort.jl:404; sort!(::Array{Int64,1}, ::Base.Sort.QuickSortAlg, ::Base.Ord...
```

src/statsmodels/formula.jl
GordStephen: `contrast.matrix` is always a `Matrix{Float64}`, while `T` is whatever your desired output type is (any `AbstractFloatMatrix` subtype).

src/statsmodels/formula.jl
nalimilan: Hmm... You could ask on the mailing list for advice about the best algorithm to do this for sparse matrices. I guess working column by column (for `SparseMatrixCSC`) would make more sense.

src/statsmodels/formula.jl
```diff
@@ -337,45 +339,46 @@ function StatsBase.model_response(mf::ModelFrame)
...
 """
-function expandcols(trm::Vector)
+function expandcols{T<:ModelMatrixContainer}(trm::Vector{T})
     if length(trm) == 1
-        asmatrix(convert(Array{Float64}, trm[1]))
+        trm[1]
     else
-        a = convert(Array{Float64}, trm[1])
-        b = expandcols(trm[2 : end])
+        a, b = trm[1], expandcols(trm[2 : end])
```
nalimilan: Keep the assignments on two lines.

src/statsmodels/formula.jl
```diff
@@ -337,44 +336,46 @@ function StatsBase.model_response(mf::ModelFrame)
...
+        modelmat_cols(T, mf.df[name])
     end
 end
 
+modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::DataVector) = convert(T, reshape(v.data, length(v), 1))
+modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::Vector) = convert(T, reshape(v, length(v), 1))
+
 """
-    modelmat_cols(v::PooledDataVector, contrast::ContrastsMatrix)
+    modelmat_cols(T::Type{AbstractFloatMatrix}, v::PooledDataVector, contrast::ContrastsMatrix)
```
nalimilan: Signature is incorrect, follow the actual one from the code.

test/formula.jl
```diff
@@ -386,58 +403,68 @@ d[:n] = 1.:8
...
 
 # two two-way interactions, with no lower-order term. both are promoted in
 # first (both x and y), but only the old term (x) in the second (because
 # dropping x gives z which isn't found elsewhere, but dropping z gives x
 # which is found (implicitly) in the promoted interaction x&y).
 mf = ModelFrame(n ~ 0 + x&y + x&z, d, contrasts=cs)
-@test ModelMatrix(mf).m == [1 0 0 0 -1  0
+mm = ModelMatrix(mf)
+@test mm.m == [1 0 0 0 -1  0
                             0 1 0 0  0 -1
```
nalimilan: Preserve indentation with other lines.

nalimilan: (Same in other places.)

nalimilan:
LGTM apart from the details I just commented.  The test failures are due to RDA. Anything else to do before merging? Could you also test that MixedModels tests pass?

GordStephen:
MixedModels tests pass locally on 0.4 and 0.5 - GLM tests fail on 0.4 on both `master` and this PR (with the same error... not anything to do with DataFrames as far as I can tell). GLM tests pass on v0.5.

I think that's it then... I'll look into the sparse indexing performance. For now waiting a bit during large sparse model matrix creation is certainly preferable to an `OutOfMemoryError`.

src/statsmodels/formula.jl
```diff
@@ -352,7 +352,7 @@ modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::DataVector) = convert(T, res
 modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, v::Vector) = convert(T, reshape(v, length(v), 1))
 
 """
-    modelmat_cols(T::Type{AbstractFloatMatrix}, v::PooledDataVector, contrast::ContrastsMatrix)
+    modelmat_cols(::Type{T}, v::PooledDataVector, contrast::ContrastsMatrix)
```
nalimilan: You should also mention the restriction `T<:AbstractFloatMatrix` as a type parameter.

src/statsmodels/formula.jl
```diff
@@ -337,44 +336,46 @@ function StatsBase.model_response(mf::ModelFrame)
...
     ## reindex with v.refs will give the corresponding row number of the
     ## contrast matrix
     reindex = [findfirst(contrast.levels, l) for l in levels(v)]
-    return contrast.matrix[reindex[v.refs], :]
+    contrastmatrix = convert(T, contrast.matrix)
+    return contrastmatrix[reindex[v.refs], :]
 end
 
 """
     expandcols(trm::Vector)
```
nalimilan: This signature should also be updated to mention the restriction on the element type.

GordStephen:
Ok, docstrings updated.

nalimilan:
Thanks!

jeffwong:
In case anyone is googling and cannot find the right way to call this functionality, you can use

```
X = ModelMatrix{SparseMatrixCSC{Float64, Integer}}(ModelFrame(@formula(y ~ x), df)).m
```

Note that SparseMatrixCSC requires two types, while Matrix only requires one, e.g. Matrix{Float64}

nalimilan:
Or rather `X = ModelMatrix{SparseMatrixCSC{Float64, Int}}(ModelFrame(@formula(y ~ x), df)).m`, since the abstract `Integer` will make the code quite slower. It would make sense to allow skipping the second type parameter and default to `Int` just like `sparse` does.

PR #1042, opened 2016-08-28
Add compatibility with pre-contrasts ModelFrame constructor

GordStephen:
Some downstream packages (like FixedEffectModels) call the direct `ModelFrame` constructor rather than use the outer constructor functions. This PR provides compatibility with the pre-#870 signature for that constructor method.

### DIFF
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -285,11 +285,7 @@ end
 
 const DEFAULT_CONTRASTS = DummyCoding
 
-function ModelFrame(trms::Terms, d::AbstractDataFrame;
-                    contrasts::Dict = Dict())
-    df, msng = na_omit(DataFrame(map(x -> d[x], trms.eterms)))
-    names!(df, convert(Vector{Symbol}, map(string, trms.eterms)))
-    for c in eachcol(df) dropunusedlevels!(c[2]) end
+function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
 
     ## Set up contrasts:
     ## Combine actual DF columns and contrast types if necessary to compute the
@@ -303,6 +299,16 @@ function ModelFrame(trms::Terms, d::AbstractDataFrame;
                                                 DEFAULT_CONTRASTS(),
                                                 col)
     end
+    return evaledContrasts
+end
+
+function ModelFrame(trms::Terms, d::AbstractDataFrame;
+                    contrasts::Dict = Dict())
+    df, msng = na_omit(DataFrame(map(x -> d[x], trms.eterms)))
+    names!(df, convert(Vector{Symbol}, map(string, trms.eterms)))
+    for c in eachcol(df) dropunusedlevels!(c[2]) end
+
+    evaledContrasts = evalcontrasts(df, contrasts)
 
     ## Check for non-redundant terms, modifying terms in place
     check_non_redundancy!(trms, df)
@@ -310,6 +316,7 @@ function ModelFrame(trms::Terms, d::AbstractDataFrame;
     ModelFrame(df, trms, msng, evaledContrasts)
 end
 
+ModelFrame(df::AbstractDataFrame, term::Terms, msng::BitArray) = ModelFrame(df, term, msng, evalcontrasts(df))
 ModelFrame(f::Formula, d::AbstractDataFrame; kwargs...) = ModelFrame(Terms(f), d; kwargs...)
 ModelFrame(ex::Expr, d::AbstractDataFrame; kwargs...) = ModelFrame(Formula(ex), d; kwargs...)
 
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -285,12 +285,11 @@ end
 
 const DEFAULT_CONTRASTS = DummyCoding
 
+## Set up contrasts:
+## Combine actual DF columns and contrast types if necessary to compute the
+## actual contrasts matrices, levels, and term names (using DummyCoding
+## as the default)
 function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
-
-    ## Set up contrasts:
-    ## Combine actual DF columns and contrast types if necessary to compute the
-    ## actual contrasts matrices, levels, and term names (using DummyCoding
-    ## as the default)
     evaledContrasts = Dict()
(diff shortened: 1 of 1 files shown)


kleinschmidt:
LGTM

nalimilan:
Just curious, why do you need to do that?

@kleinschmidt You're the most qualified to review this now that you've rewritten most of that code, so feel free to merge after leaving a chance for others to comment.

GordStephen:
Not sure why FixedEffectModels needs the raw constructor specifically, @matthieugomez would be the one to ask. My guess would be that it's to support explicitly specifying a subset of the input dataframe rows for use in the model?

matthieugomez:
Tbh, I'm a little bit confused about what I did and what has changed here. 

I do need to remove more observations than what `ModelFrame` usual removes:
missing obs for variables used in fixed effects (not part of the formula given to ModelFrame  — they are treated separtely), missing obs for variables used in standard errors (like clustered standard errors), missing obs for weight. Also I have a subset option.

src/statsmodels/formula.jl
```diff
@@ -303,13 +299,24 @@ function ModelFrame(trms::Terms, d::AbstractDataFrame;
                                                 DEFAULT_CONTRASTS(),
                                                 col)
     end
+    return evaledContrasts
```
nalimilan: The convention in Julia code is not to use camelCase for variable names.

src/statsmodels/formula.jl
```diff
@@ -285,11 +285,7 @@ end
 
 const DEFAULT_CONTRASTS = DummyCoding
 
-function ModelFrame(trms::Terms, d::AbstractDataFrame;
-                    contrasts::Dict = Dict())
-    df, msng = na_omit(DataFrame(map(x -> d[x], trms.eterms)))
-    names!(df, convert(Vector{Symbol}, map(string, trms.eterms)))
-    for c in eachcol(df) dropunusedlevels!(c[2]) end
+function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
```
nalimilan: Remove empty line at top of function.

src/statsmodels/formula.jl
nalimilan: Sorry, I just realized this variable already exists. Carry on then.

src/statsmodels/formula.jl
nalimilan: And BTW, move the comments below before the signature, as they clearly document the function.

nalimilan:
@matthieugomez When you'll have the time to look at this more deeply, it would be good to find a higher-level interface which would work for your use case. With the recent changes, we support arbitrary contrasts (not just dummy coding), but your custom evaluation function probably doesn't.

matthieugomez:
I used this constructor because I already had a dataframe without NA (after doing a bunch of operations, in particular including missing weights and standard errors options) and did not want to do a new copy of the dataframe with the line 
`df, msng = na_omit(DataFrame(map(x -> d[x], trms.eterms)))`
I do think that `ModelFrame` and `ModelMatrix` are doing excessive amount of copies (in particular it's impossible to do regressions with datasets more than 1/10 my RAM).
Now, this particular line is just one more copy and I'm not sure it changes anything.

nalimilan:
If that's the only problem, then we can probably change this line (or `drop_na`) to avoid making a copy when there are no missing values. (That code is know to be particularly inefficient WRT copying, and there are some low-hanging fruits waiting for someone to fix.)

matthieugomez:
Honestly it's ok. There will probably be a better fix someday (using subdataframes). A temporary fix based on the number of missing values would just complicate the function (`df` would sometimes be a reference to the original dataframes, sometimes not).

matthieugomez:
Sorry. I do need a way to specify a `msng` vector that is not restricted to the missing obs for variables in the formula. I think other may use it too. What about splitting the function `ModelFrame`
```julia
ModelFrame(trms::Terms, d::AbstractDataFrame; contrasts::Dict = Dict())
df, msng = na_omit(DataFrame(map(x -> d[x], trms.eterms)))
end

ModelFrame(trms::Terms, df::AbstractDataFrame, msng ; contrasts::Dict = Dict())
names!(df, convert(Vector{Symbol}, map(string, trms.eterms)))
    for c in eachcol(df) dropunusedlevels!(c[2]) end

    ## Set up contrasts:
    ## Combine actual DF columns and contrast types if necessary to compute the
    ## actual contrasts matrices, levels, and term names (using DummyCoding
    ## as the default)
    evaledContrasts = Dict()
    for (term, col) in eachcol(df)
        is_categorical(col) || continue
        evaledContrasts[term] = ContrastsMatrix(haskey(contrasts, term) ?
                                                contrasts[term] :
                                                DEFAULT_CONTRASTS(),
                                                col)
    end

    ## Check for non-redundant terms, modifying terms in place
    check_non_redundancy!(trms, df)

    ModelFrame(df, trms, msng, evaledContrasts)
```

nalimilan:
Yeah, makes sense as a public API. Please make a PR (with tests for the new method). Though I'm not sure `DataFrame(map(x -> d[x], trms.eterms))` is the most efficient approach: why not index `d` with a vector of symbols?

Also, would this replace or complement the present PR?

matthieugomez:
Sorry I'm not familiar with the package, why isn't this efficient?
```julia
DataFrame(map(x -> d[x], trms.eterms))
```

nalimilan:
@GordStephen Thanks. Unfortunately, recent changes on master have introduced conflicts. Could you rebase? (I would have done it myself, but I don't have the rights to push to your branch...)

@matthieugomez I'm not sure it really inefficient, but if the simpler `d[trms.eterms]` works, better use it. Anyway, please open a PR an we can discuss that.

GordStephen:
Ah, sorry. I'll check that "Allow edits from maintainers" box. Rebased now.

nalimilan:
Thanks!

matthieugomez:
For reference, I ended up using the basic ModelFrame function, replacing
the msng field by a custom vector at the end.
https://github.com/matthieugomez/FixedEffectModels.jl/blob/f1487fd013e9b5ac27d261ccaf3ff8c754c294e0/src/utils/formula.jl#L100

nalimilan:
Can you make a PR to add this method? You shouldn't have to carry things like that in your package.

PR #1045, opened 2016-08-31
fix issue from #1017

matthieugomez:


test/formula.jl
```diff
@@ -535,4 +535,11 @@ df = DataFrame(x = [1,2,3], y = [4,5,6])
 mf = ModelFrame(y ~ 1 + (1 | x), df)
 @test coefnames(mf) == ["(Intercept)"]
 
+
+# Ensure X is not a view on df column
+df = DataFrame(x = [1.0,2.0,3.0], y = [4.0,5.0,6.0])
+mf = ModelFrame(y ~ 0 + x, df)
+X = ModelMatrix(mf).m
+X[1] = 0.0
+@assert isapprox(mf.df[1, :x], 1.0)
```
ararslan: Should be `@test` rather than `@assert`. Also this test is causing the failure on all platforms.

nalimilan:
Actually, I don't think using `reduce` here was a good idea at all: it's going to create as many temporary copies as the number of columns, minus 1. What we should do is allocate the matrix, then fill it with a loop (`hcat(blocks...)` would work, but splatting isn't recommended with a variable number of arguments).

matthieugomez:
I don't understand the code so I'm not going to do that.

nalimilan:
You don't need to understand the code to do that (and actually I don't really understand it either).

matthieugomez:
How to compute the number of columns ex ante?

matthieugomez:
Oh. Ok I'll just do a loop afterwards.

src/statsmodels/formula.jl
```diff
@@ -481,7 +481,16 @@ creating the model matrix.
...
 
-    ModelMatrix{T}(reduce(hcat, blocks), assign)
+    I = size(dfrm, 1)
+    J = mapreduce(x -> size(x, 2), +, blocks)
+    X = zeros(eltype(T), I, J)
+    i = 1
+    for block in blocks
+        len = size(block, 2)
+        X[:, i:(i + len - 1)] = block
+        i += len
```
ararslan: What's this `i` doing? Doesn't appear to be used after the loop.

matthieugomez: it keeps track of the current column in X. What do you suggest?

ararslan: WOW, I'm dumb! I don't know how I missed it being used literally in the line above where I commented. Sorry about that...

src/statsmodels/formula.jl
```diff
@@ -481,7 +481,16 @@ creating the model matrix.
         append!(assign, fill(i_term, size(blocks[end], 2)))
     end
 
-    ModelMatrix{T}(reduce(hcat, blocks), assign)
+    I = size(dfrm, 1)
+    J = mapreduce(x -> size(x, 2), +, blocks)
+    X = zeros(eltype(T), I, J)
```
nalimilan: @GordStephen Is this OK as regards your recent work on sparse matrices? Or should we use `similar` here?

test/formula.jl
```diff
@@ -535,4 +535,11 @@ df = DataFrame(x = [1,2,3], y = [4,5,6])
 mf = ModelFrame(y ~ 1 + (1 | x), df)
 @test coefnames(mf) == ["(Intercept)"]
 
+
+# Ensure X is not a view on df column
+df = DataFrame(x = [1.0,2.0,3.0], y = [4.0,5.0,6.0])
+mf = ModelFrame(y ~ 0 + x, df)
+X = ModelMatrix(mf).m
+X[1] = 0.0
+@test isapprox(mf.df[1, :x], 1.0)
```
nalimilan: Do you really need `isapprox` instead of `==` here? (Note you can also `≈` instead.)

matthieugomez:
I'm sorry but dont have time to work more on that. Feel free to use to use the code in another pull request.

nalimilan:
See https://github.com/JuliaStats/DataFrames.jl/pull/1048.

matthieugomez:
Thanks

PR #1048, opened 2016-09-03
fix issue from #1017

nalimilan:
Continuation of https://github.com/JuliaStats/DataFrames.jl/pull/1045.

### DIFF
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -487,7 +487,7 @@ creating the model matrix.
 
     I = size(dfrm, 1)
     J = mapreduce(x -> size(x, 2), +, blocks)
-    X = Array(eltype(T), I, J)
+    X = similar(blocks[1], I, J)
     i = 1
     for block in blocks
         len = size(block, 2)
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -485,7 +485,16 @@ creating the model matrix.
         error("Could not construct model matrix. Resulting matrix has 0 columns.")
     end
 
-    ModelMatrix{T}(reduce(hcat, blocks), assign)
+    I = size(dfrm, 1)
+    J = mapreduce(x -> size(x, 2), +, blocks)
+    X = zeros(eltype(T), I, J)
+    i = 1
+    for block in blocks
+        len = size(block, 2)
+        X[:, i:(i + len - 1)] = block
+        i += len
+    end
+    ModelMatrix{T}(X, assign)
 end
 ModelMatrix(mf::ModelFrame) = ModelMatrix{Matrix{Float64}}(mf)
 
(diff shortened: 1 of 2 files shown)


nalimilan:
@dmbates @kleinschmidt Does that sound OK to you?

src/statsmodels/formula.jl
```diff
@@ -485,7 +485,16 @@ creating the model matrix.
         error("Could not construct model matrix. Resulting matrix has 0 columns.")
     end
 
-    ModelMatrix{T}(reduce(hcat, blocks), assign)
+    I = size(dfrm, 1)
+    J = mapreduce(x -> size(x, 2), +, blocks)
+    X = Array(eltype(T), I, J)
```
GordStephen: Following up on the question in #1045, it would be good to avoid preallocating a dense matrix that's potentially large and expected to be sparse. Uninitialized sparse arrays are tricky though... `spzeros` is probably the best option, but that doesn't generalize well for the dense case. The only thing I can think of off the top of my head would be something not-so-elegant like

```julia
T <: AbstractSparseArray ? spzeros(Tv, Ti, I, J) : Array(Tv, I, J)
```

nalimilan: Does it happen that `blocks[1]` is a sparse matrix though? Also, is the  code below, which fills the matrix column by column, reasonably efficient for sparse matrices?

ararslan: I think the `Array{eltype(T)}(I, J)` is preferred

GordStephen: `blocks` is a `Vector{T}` so `blocks[1]` will always be a `T`, i.e. correctly sparse or dense depending on what was requested.

At least in my use cases, the reason for wanting a `ModelMatrix{SparseMatrixCSC{Float64,Int}}` is because a `ModelMatrix{Matrix{Float64}}` wouldn't fit in memory... Thus an `Array{eltype(T)}(I, J)` wouldn't fit in memory either, defeating the point of supporting sparse ModelMatrix construction in the first place.

nalimilan: > blocks is a Vector{T} so blocks[1] will always be a T, i.e. correctly sparse or dense depending on what was requested.

OK. I thought that part of the code hadn't been adapted to use sparse matrices, even when requested.

> At least in my use cases, the reason for wanting a ModelMatrix{SparseMatrixCSC{Float64,Int}} is because a ModelMatrix{Matrix{Float64}} wouldn't fit in memory... Thus an Array{eltype(T)}(I, J) wouldn't fit in memory either, defeating the point of supporting sparse ModelMatrix construction in the first place.

Sure, that's really essential.

kleinschmidt:
Looks good to me. 

Tangentially, it might be possible to be even more clever about this to avoid making at least some of the copies involved in generating `blocks` (e.g., copying only once into the model matrix and using that to generate higher order terms). But that's going to require a pretty substantial re-organization. Just wanted to make a note of it here.

nalimilan:
Yeah, that code is clearly sub-optimal WRT performance, but I didn't want to dive into that right now.

PR #1056, opened 2016-09-05
Fix depwarn of read_rda

andreasnoack:


andreasnoack:
We need a new version of `Compat.jl` before this will pass. https://github.com/JuliaLang/METADATA.jl/pull/6263

ararslan:
This is awesome. Thanks so much, and sorry for the mess!

ararslan:
> We need a new version of Compat.jl before this will pass.

Once that's tagged we should set 0.8.9 as the lower bound on Compat in REQUIRE.

andreasnoack:
I'll do it as part of this PR.

ararslan:
Right, that's what I meant

test/REQUIRE
```diff
@@ -1 +1,4 @@
+Compat 8.9
```
ararslan: Should be 0.8.9, right?

andreasnoack: Oops.

ararslan:
We should tag a new patch version. I can do that later today if no one beats me to it.

tkelman:
Please do tag, the MethodError for `depwarn(::String)` broke multiple packages

test/deprecated.jl
```diff
@@ -0,0 +1,8 @@
+using DataFrames, FileIO, RDatasets, Compat
+fn = tempname()
+open(fn, "w") do f
+    redirect_stderr(f) do
+        dataset("datasets", "iris")
+    end
+end
+@test contains(readstring(fn), "WARNING: read_rda(args...) is deprecated, use FileIO.load(args...) instead.")
```
tkelman: you should delete the file when this is done

Issue #1054, opened 2016-09-05
RDatasets broken by recent deprecations in DataFrames


tlnagy:
Gadfly's tests are broken (https://github.com/dcjones/Gadfly.jl/issues/890) on the latest release of `DataFrames` due to https://github.com/JuliaStats/DataFrames.jl/pull/1031 being merged and breaking `RDatasets`. Thought I would post this here since `RDatasets` is pretty low traffic.

@ararslan @nalimilan

Mattriks:
There's a PR to fix this [johnmyleswhite/RDatasets.jl#43](https://github.com/johnmyleswhite/RDatasets.jl/pull/43)

nalimilan:
Thanks for the heads up. Let's close this since there's a PR.

PR #1057, opened 2016-09-05
Fill existing arrays with scalars

### LINKED ISSUE #1051: julia dataframe changing one cell changes entire row

TotalVerb:
Closes #1053 (replacing it). Fixes #1051. Closes (maybe?) #1054, pending possible further discussion.

This implements @nalimilan and @andreasnoack's idea of filling arrays with scalars instead of creating new ones.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -365,10 +365,13 @@ function Base.setindex!(df::DataFrame,
 end
 
 # df[SingleColumnIndex] = Single Item (EXPANDS TO NROW(DF) if NCOL(DF) > 0)
-function Base.setindex!(df::DataFrame,
-                v::Any,
-                col_ind::ColumnIndex)
-    insert_single_column!(df, upgrade_scalar(df, v), col_ind)
+function Base.setindex!(df::DataFrame, v, col_ind::ColumnIndex)
+    if haskey(index(df), col_ind)
+        fill!(df[col_ind], v)
+    else
+        insert_single_column!(df, upgrade_scalar(df, v), col_ind)
+    end
+    return df
 end
 
 # df[MultiColumnIndex] = DataFrame
@@ -397,7 +400,7 @@ function Base.setindex!{T <: ColumnIndex}(df::DataFrame,
                                   col_inds::AbstractVector{T})
     dv = upgrade_vector(v)
     for col_ind in col_inds
-        insert_single_column!(df, dv, col_ind)
+        df[col_ind] = dv
     end
     return df
 end
@@ -411,9 +414,8 @@ end
 function Base.setindex!{T <: ColumnIndex}(df::DataFrame,
                                   val::Any,
                                   col_inds::AbstractVector{T})
-    dv = upgrade_scalar(df, val)
     for col_ind in col_inds
-        insert_single_column!(df, dv, col_ind)
+        df[col_ind] = val
     end
     return df
 end
--- a/test/index.jl
+++ b/test/index.jl
@@ -56,4 +56,12 @@ for name in names(i)
   i2[name] # Issue #715
 end
 
+#= Aliasing & Mutation =#
+
+# columns should not alias if scalar broadcasted
+df = DataFrame(A=[0],B=[0])
+df[1:end] = 0.0
+df[1,:A] = 1.0
+@test df[1,:B] === 0
+
 end


src/dataframe/dataframe.jl
```diff
@@ -368,7 +380,7 @@ end
 function Base.setindex!(df::DataFrame,
                 v::Any,
                 col_ind::ColumnIndex)
-    insert_single_column!(df, upgrade_scalar(df, v), col_ind)
+    fill_single_column!(df, v, col_ind)
```
andreasnoack: Couldn't `setindex!` for the column type handle this, e.g. something like
```jl
df[:col_ind] = v
```
?

TotalVerb: This is `setindex!` for the column type.

TotalVerb: But you're right in that the `fill_single_column!` function is just `setindex!`. I suppose it would be easier just to inline this function.

andreasnoack: Sorry. What I meant was to use `setindex` for the `DataVector`/`NullableVector` which would probably be something like
```df
df[:col_ind][:] = v
```

TotalVerb: Good point. Unfortunately `df[:col_ind] .= v` does not work (even on v0.6) due to `broadcast!` being a little particular with what data types it considers "vectors", such as `String`. Also, I recall there was a discussion about `setindex!(A, v, Colon())` still being slower than `fill!(A, v)`, though I don't know if this issue has been addressed by now.

TotalVerb: It looks like https://github.com/JuliaLang/julia/pull/16310 has closed the performance difference, although that means it still exists on v0.4. I think keeping `fill!` might be better for now.

TotalVerb: I've inlined the `fill_single_column` function now into `setindex!`.

andreasnoack:
Great. Seems much simpler. Don't the tests fit into `index.jl`?

TotalVerb:
There used to be a few more tests for vector `setindex!` aliasing, but that's not fixed here (needs a more copying solution like #1054). I can put this one in index.jl.

TotalVerb:
OK, test merged into index.jl now.

src/dataframe/dataframe.jl
```diff
@@ -365,10 +365,13 @@ function Base.setindex!(df::DataFrame,
...
 
 # df[SingleColumnIndex] = Single Item (EXPANDS TO NROW(DF) if NCOL(DF) > 0)
-function Base.setindex!(df::DataFrame,
-                v::Any,
-                col_ind::ColumnIndex)
-    insert_single_column!(df, upgrade_scalar(df, v), col_ind)
+function Base.setindex!(df::DataFrame, v, col_ind::ColumnIndex)
+    if haskey(index(df), col_ind)
+        j = index(df)[col_ind]
+        fill!(df.columns[j], v)
```
nalimilan: Doesn't `fill!(df[col_ind], v)` work here?

src/dataframe/dataframe.jl
```diff
@@ -406,14 +409,13 @@ end
 function Base.setindex!(df::DataFrame,
                 val::Any,
                 col_inds::AbstractVector{Bool})
-    setindex!(df, val, find(col_inds))
+    df[find(col_inds)] = val
```
nalimilan: Since you change this one, there are other occurrences of this above.

TotalVerb: Oops, this change is not correct. I will revert it.

src/dataframe/dataframe.jl
TotalVerb: Fixed..

src/dataframe/dataframe.jl
nalimilan: Why? It's the same thing AFAICT.

TotalVerb: The return value of `setindex!` is the dataframe; the return value of `x[] = y` is the value assigned.

nalimilan: Ah, good catch.

dmbates:
Is this ready to be merged now?

PR #1060, opened 2016-09-05
Allow arguments to displaysize on 0.4

### LINKED ISSUE #1058: writemime error
On the current master (and tagged version), Julia 0.4.5, in IJulia

```julia
> using DataFrames
> DataFrame(x=[1,2], y=[3,4])

UndefVarError: displaysize not defined

 in writemime at /Users/cedric/.julia/v0.4/DataFrames/src/abstractdataframe/io.jl:181
 in sprint at iostream.jl:206
 in display_dict at /Users/cedric/.julia/v0.4/IJulia/src/execute_request.jl:38
 in execute_request at /Users/cedric/.julia/v0.4/IJulia/src/execute_request.jl:195
 in eventloop at /Users/cedric/.julia/v0.4/IJulia/src/IJulia.jl:138
 in anonymous at task.jl:447
```

Seems to come from [https://github.com/JuliaLang/julia/pull/14580](Julia/14580). Probably needs an `@compat`?

ararslan:
Fixes #1058

### DIFF
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -92,6 +92,12 @@ export @~,
 ##
 ##############################################################################
 
+if VERSION < v"0.5.0-dev+2023"
+    _displaysize(x...) = Base.tty_size()
+else
+    const _displaysize = Base.displaysize
+end
+
 for (dir, filename) in [
         ("other", "utils.jl"),
         ("other", "index.jl"),
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -178,7 +178,7 @@ end
         write(io, "<th>$column_name</th>")
     end
     write(io, "</tr>")
-    tty_rows, tty_cols = Base.displaysize(io)
+    tty_rows, tty_cols = _displaysize(io)
     mxrow = min(n,tty_rows)
     for row in 1:mxrow
         write(io, "<tr>")
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -165,11 +165,6 @@ function getprintedwidth(maxwidths::Vector{Int}) # -> Int
     return totalwidth
 end
 
-# For Julia < 0.5
-if VERSION < v"0.5.0-dev+2023"
-    displaysize(io::IO) = Base.tty_size()
-end
-
 #' @description
 #'
 #' When rendering an AbstractDataFrame to a REPL window in chunks, each of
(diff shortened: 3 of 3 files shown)


src/abstractdataframe/show.jl
```diff
@@ -165,11 +165,6 @@ function getprintedwidth(maxwidths::Vector{Int}) # -> Int
     return totalwidth
 end
 
-# For Julia < 0.5
-if VERSION < v"0.5.0-dev+2023"
```
nalimilan: That function doesn't exist in Compat, only the one without any arguments (I didn't include it there since it's a bit a cheating: `io` isn't used at all).

src/DataFrames.jl
```diff
@@ -92,6 +92,8 @@ export @~,
 ##
 ##############################################################################
 
+const displaysize(io::IO) = Compat.displaysize()
```
nalimilan: Does `const` actually have a meaning here?

src/abstractdataframe/show.jl
ararslan: Ah, good catch, thanks. Fixed in another commit.

src/DataFrames.jl
nalimilan: Oh, and the problem with defining this even on 0.5 is that we won't use `io` even there. So you really need to keep the original `if`.

ararslan: Not really, no. Only that it won't change. It's just what I've seen done in Compat so I mimicked it here. I can take it out if that'd be better.

nalimilan: > Not really, no. Only that it won't change. It's just what I've seen done in Compat so I mimicked it here. I can take it out if that'd be better.

The difference in Compat.jl is that the code created a new constant binding to the function. Here, you define a new function which calls the other one in its body.

ararslan: Yeah, I realized that after saying it. I wish I knew what was wrong with my brain today...

src/abstractdataframe/io.jl
```diff
@@ -178,7 +178,7 @@ end
         write(io, "<th>$column_name</th>")
     end
     write(io, "</tr>")
-    tty_rows, tty_cols = Base.displaysize(io)
+    tty_rows, tty_cols = Compat.displaysize()
```
nalimilan: No, you need to pass `io` (at least where possible), as it is the recommended way in Julia 0.5.

ararslan: Oh, okay. I'm surprised the `io` thing didn't make it into Compat then. I'll revert the most recent commit. Was it okay prior to this commit?

ararslan:
The 0.6 AppVeyor failure is both unrelated and strange.

nalimilan:
Ah, that's https://github.com/JuliaLang/julia/issues/17932. We need to investigate this.

(I've changed the title, since arguments were already allowed, just the function definition failed.)

PR #1061, opened 2016-09-05
Do a better job deprecating read_rda

ararslan:
Alternative to #1056 (and borrows some code) cc @andreasnoack

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,4 +5,4 @@ GZip
 SortingAlgorithms
 Reexport
 Compat 0.8.4
-FileIO 0.1.2
+RData 0.0.4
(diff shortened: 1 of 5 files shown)


andreasnoack:
I've just made the same change so this shouldn't be necessary.

ararslan:
Oh sweet, thanks! Then the Compat version bump shouldn't be necessary, right?

andreasnoack:
It is necessary for the new `redirect_stderr` method.

ararslan:
Oh, okay. Thanks again for cleaning up the mess I made. I really appreciate it!

ararslan:
Have you not pushed your changes to your branch?

andreasnoack:
Not yet. I trying to be gentle to the Travis queue. In particular, OSX seems to be a problem and since `Compat` hasn't been tagged yet, I thought it would be better to wait.

vchuravy:
;) I love the branch name.

ararslan:
@andreasnoack Yeah, good call. We had a bunch of PRs all at once and I think Travis got a little overwhelmed.

PR #1062, opened 2016-09-06
🚮 delete temp file when test/deprecated.jl is done

tkelman:


ararslan:
+1 for the change and for the commit message.

tkelman:
There are several in `test/io.jl` that are also littering, they delete the file if it exists immediately after creating it but don't delete it before overwriting the variable that contained the filename.

test/deprecated.jl
```diff
@@ -1,8 +1,9 @@
 using DataFrames, FileIO, RDatasets, Compat
-fn = tempname()
-open(fn, "w") do f
-    redirect_stderr(f) do
-        dataset("datasets", "iris")
+mktemp() do fn
```
ararslan: Looks like the syntax or something is off, maybe? This line is causing failures on all versions and platforms.

ararslan:
> There are several in test/io.jl that are also littering

If you wanted to tackle those as well in this PR, that would be much appreciated :wink:

test/deprecated.jl
tkelman: oops, sorry for wasting your ci time. danger of web-ui PR's

andreasnoack: Didn't know about `mktemp`. Anyway, I had to close the file before I could call `readstring` so `mktemp` might not work in this case.

PR #1065, opened 2016-09-09
Remove deprecation test

### LINKED ISSUE #1064: Deprecation tests are failing
We're relying on RDatasets' use of `DataFrames.read_rda` to test for a deprecation warning given by `read_rda` by redirecting STDERR and looking for the deprecation message. However, a recently merged RDatasets PR (that was subsequently tagged) fixes the deprecation warning, so there's no message for us to read in the test.

I see two possible courses of action here:

1. Add an RDA file and manually read it using `read_rda` to continue to test the existence of a deprecation warning.

2. We know the deprecation works since the test had been passing until RDatasets fixed it on their end, and `read_rda` here won't change (it's only referenced in a `@deprecate`), we could just remove the test.

Thoughts?

ararslan:
Fixes #1064

CI should be green once this is tagged, and PkgEval should be as well if we tag afterward.

### DIFF
--- a/test/deprecated.jl
+++ b/test/deprecated.jl
@@ -1,9 +1,0 @@
-using DataFrames, FileIO, RDatasets, Compat
-mktemp() do fn, fio
-    open(fn, "w") do f
-        redirect_stderr(f) do
-            dataset("datasets", "iris")
-        end
-    end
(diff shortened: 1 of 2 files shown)


ararslan:
Linux nightly failure is unrelated (can't load LibGit2)

PR #1070, opened 2016-09-18
Faster sparse model matrix construction

GordStephen:
Addresses slow row-wise `SparseMatrixCSC` indexing when building sparse model matrix columns by transposing the contrast matrix before reindexing column-wise, then transposing the result (rather than just reindexing row-wise).


nalimilan:
Cc: @kleinschmidt

PR #1071, opened 2016-09-22
Fixes a bug in writetable's append parameter.

deyandyankov:
Previous to the fix, the following resulted in an error:
using DataFrames
df = DataFrames.DataFrame(a = @data([1, 2]))
filename = "testme.csv"
isfile(filename) && rm(filename)
writetable(filename, df, append = true)
writetable(filename, df, append = true)

ERROR: LoadError: KeyError: key "Column names don't match names in file"
not found

The changes in abstractdataframe/io.jl resolve the issue.
A test for this particular case has been added in test/io.jl

### DIFF
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -94,6 +94,7 @@ writetable(filename, df, [keyword options])
 * `separator::Char` -- The separator character that you would like to use. Defaults to the output of `getseparator(filename)`, which uses commas for files that end in `.csv`, tabs for files that end in `.tsv` and a single space for files that end in `.wsv`.
 * `quotemark::Char` -- The character used to delimit string fields. Defaults to `'"'`.
 * `header::Bool` -- Should the file contain a header that specifies the column names from `df`. Defaults to `true`.
+* `append::Bool` -- If `filename` already exists, the dataframe is only appended to it. `append` = true turns automatically sets `header` = false but it makes sure that the names of the DataFrame being appended to the file correspond to the header in the file.
 * `nastring::AbstractString` -- What to write in place of missing data. Defaults to `"NA"`.
 
 ### Result
@@ -108,6 +109,8 @@ writetable("output.csv", df)
 writetable("output.dat", df, separator = ',', header = false)
 writetable("output.dat", df, quotemark = '\', separator = ',')
 writetable("output.dat", df, header = false)
+writetable("output.dat", df, header = true)
+writetable("output.dat", df, append = true)
 ```
 """
 function writetable(filename::AbstractString,
@@ -124,7 +127,6 @@ function writetable(filename::AbstractString,
 
     if append && isfile(filename) && filesize(filename) > 0
         file_df = readtable(filename, header = false, nrows = 1)
-
         # Check if number of columns matches
         if size(file_df, 2) != size(df, 2)
             throw(DimensionMismatch("Number of columns differ between file and DataFrame"))
@@ -133,8 +135,8 @@ function writetable(filename::AbstractString,
         # When 'append'-ing to a nonempty file,
         # 'header' triggers a check for matching colnames
         if header
-            if any(i -> @compat(Symbol(file_df[1, i])) != index(df)[i], 1:size(df, 2))
-                throw(KeyError("Column names don't match names in file"))
+            if any(i -> @compat(Symbol(file_df[1, i])) != names(df)[i], 1:size(df, 2))
+                throw(KeyError("Column names in dataframe don't match column names from header in file " * filename))
             end
 
             header = false
--- a/test/io.jl
+++ b/test/io.jl
@@ -366,10 +366,10 @@ module TestIO
 
     # Overwrites file if append == false
     writetable(tf, df1)
-    writetable(tf, df2)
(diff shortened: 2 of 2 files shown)


deyandyankov:
The tests from `statsmodel.jl` fail which was unaffected by my changes. Any comments/reviews welcome.

src/abstractdataframe/io.jl
```diff
@@ -94,6 +94,7 @@ writetable(filename, df, [keyword options])
 * `separator::Char` -- The separator character that you would like to use. Defaults to the output of `getseparator(filename)`, which uses commas for files that end in `.csv`, tabs for files that end in `.tsv` and a single space for files that end in `.wsv`.
 * `quotemark::Char` -- The character used to delimit string fields. Defaults to `'"'`.
 * `header::Bool` -- Should the file contain a header that specifies the column names from `df`. Defaults to `true`.
+* `append::Bool` -- If `filename` already exists, the dataframe is only appended to it. `append` = true turns automatically sets `header` = false but it makes sure that the names of the DataFrame being appended to the file correspond to the header in the file.
```
deyandyankov: perhaps should scrap everything after the first sentence

src/abstractdataframe/io.jl
```diff
@@ -108,6 +109,8 @@ writetable("output.csv", df)
 writetable("output.dat", df, separator = ',', header = false)
 writetable("output.dat", df, quotemark = '\', separator = ',')
 writetable("output.dat", df, header = false)
+writetable("output.dat", df, header = true)
```
nalimilan: These should be added to tests rather than to examples.

src/abstractdataframe/io.jl
```diff
@@ -124,7 +127,6 @@ function writetable(filename::AbstractString,
 
     if append && isfile(filename) && filesize(filename) > 0
         file_df = readtable(filename, header = false, nrows = 1)
-
```
nalimilan: Unrelated change.

test/io.jl
```diff
@@ -366,10 +366,10 @@ module TestIO
 
     # Overwrites file if append == false
     writetable(tf, df1)
-    writetable(tf, df2)
-    @test readtable(tf) == df2
+    writetable(tf, df3)
```
nalimilan: ?

test/io.jl
```diff
@@ -366,10 +366,10 @@ module TestIO
...
+    @test readtable(tf) == df3
 
-    # Enforces matching column names iff append == true && header == true
+    # Enforces matching column names if append == true && header == true
```
nalimilan: Is this change intended? "Iff" usually means "if and only if", so that's not a typo.

test/io.jl
```diff
@@ -379,6 +379,14 @@ module TestIO
     @test_throws DimensionMismatch writetable(tf, df3b, header = false, append = true)
     rm(tf)
 
+    # create a new file if one does not exist. then, append to it.
```
nalimilan: "if one does not exist" sounds weird, since you specifically remove it before creation.

src/abstractdataframe/io.jl
```diff
@@ -133,8 +135,8 @@ function writetable(filename::AbstractString,
         # When 'append'-ing to a nonempty file,
         # 'header' triggers a check for matching colnames
         if header
-            if any(i -> @compat(Symbol(file_df[1, i])) != index(df)[i], 1:size(df, 2))
-                throw(KeyError("Column names don't match names in file"))
+            if any(i -> @compat(Symbol(file_df[1, i])) != names(df)[i], 1:size(df, 2))
+                throw(KeyError("Column names in dataframe don't match column names from header in file " * filename))
```
nalimilan: While you're at it, could you print the name of the header which triggered the error? That's really really useful when trying to understand what's going wrong.

src/abstractdataframe/io.jl
nalimilan: No, being explicit sounds fine to me. Though there's a typo in "turns automatically sets", and backticks should be around the whole `append = true` and  `header = false` expressions.

pfarndt:
Would be nice if this bug could be fixed soon ...

deyandyankov:
apologies @pfarndt I did not realise that @nalimilan had reviewed this and had forgotten about the issue.

after reading @nalimilan's reviews it seems that I'll need to get back to the code and make improvements before I resubmit. @nalimilan is this codebase still relevant given the grand changes to DataFrames? I'll need to fetch the latest changes and see if the issue persists. I should be able to look into it during the weekend.

deyandyankov:
According to https://github.com/JuliaData/DataTables.jl/issues/10 it seems like `writetable` is being replaced with `CSV.writetable` - @pfarndt, would `CSV.jl` work for you?

pfarndt:
`CSV.jl` works for me - however `writeable` as some extra code that in case one appends rows to a file  

- checks the identity and order of the column names
- automatically outputs a header in case the file is empty

This becomes handy if one aggregates results in a cvs/tsv file. Would be nice to have this in `CSV` as well.

nalimilan:
Please file an issue in CSV.jl then.

quinnj:
`readtable` `writetable` are now deprecated.

PR #1072, opened 2016-09-23
Only sort duplicated columns once

gustafsson:
```
> df = DataFrame(a=[2,1]);
> df[:b] = df[:a];
> sort!(df, cols=:a)
2×2 DataFrames.DataFrame
│ Row │ a │ b │
├─────┼───┼───┤
│ 1   │ 2 │ 2 │
│ 2   │ 1 │ 1 │
```
The column `:a` is not sorted here. This happens because `:b` and `:a` refer to the same array which gets permuted twice. By skipping duplicated columns in the permutation we get the correct result.
```
2×2 DataFrames.DataFrame
│ Row │ a │ b │
├─────┼───┼───┤
│ 1   │ 1 │ 1 │
│ 2   │ 2 │ 2 │
```

### DIFF
--- a/src/dataframe/sort.jl
+++ b/src/dataframe/sort.jl
@@ -13,7 +13,10 @@ function Base.sort!(df::DataFrame, a::Base.Sort.Algorithm, o::Base.Sort.Ordering
     p = sortperm(df, a, o)
     pp = similar(p)
     c = columns(df)
-    uc = [!any(j->c[i] ≡ c[j], 1:i-1) for i=1:length(c)]
+
+    # Check that columns that shares the same underlying array are only permuted once PR#1072
+    uc = [!any(j->c[i] === c[j], 1:i-1) for i=1:length(c)]
+
     for col in c[uc]
         copy!(pp,p)
         Base.permute!!(col, pp)
--- a/src/dataframe/sort.jl
+++ b/src/dataframe/sort.jl
@@ -12,7 +12,9 @@ end
 function Base.sort!(df::DataFrame, a::Base.Sort.Algorithm, o::Base.Sort.Ordering)
     p = sortperm(df, a, o)
     pp = similar(p)
-    for col in columns(df)
+    c = columns(df)
+    uc = [!any(j->c[i] ≡ c[j], 1:i-1) for i=1:length(c)]
+    for col in c[uc]
         copy!(pp,p)
         Base.permute!!(col, pp)
     end
--- a/src/dataframe/sort.jl
+++ b/src/dataframe/sort.jl
@@ -15,7 +15,7 @@ function Base.sort!(df::DataFrame, a::Base.Sort.Algorithm, o::Base.Sort.Ordering
     c = columns(df)
 
     for (i,col) in enumerate(c)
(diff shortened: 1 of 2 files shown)


test/sort.jl
```diff
@@ -41,5 +41,8 @@ module TestSort
 
     @test df == ds
 
-
+    df = DataFrame(a=[2,1])
```
nalimilan: Please add a comment explaining what you're testing here, and refer to this PR.

src/dataframe/sort.jl
```diff
@@ -12,7 +12,9 @@ end
 function Base.sort!(df::DataFrame, a::Base.Sort.Algorithm, o::Base.Sort.Ordering)
     p = sortperm(df, a, o)
     pp = similar(p)
-    for col in columns(df)
+    c = columns(df)
+    uc = [!any(j->c[i] ≡ c[j], 1:i-1) for i=1:length(c)]
```
nalimilan: It would be slightly more efficient to compute this directly in the loop, and `continue` for columns which have already been sorted.

Also, I don't think the `≡` is used anywhere in the code base, better keep the ASCII version for now. Finally, a comment would be useful.

gustafsson: === it is then

I moved it inside the for-loop

src/dataframe/sort.jl
```diff
@@ -12,7 +12,14 @@ end
 function Base.sort!(df::DataFrame, a::Base.Sort.Algorithm, o::Base.Sort.Ordering)
     p = sortperm(df, a, o)
     pp = similar(p)
-    for col in columns(df)
+    c = columns(df)
+
+    for (i,col) in enumerate(c)
+        # Check if this column has been sorted already. PR#1072
```
nalimilan: No PR names in the code, `git blame` is there for this. (I know, it's a bit confusing that we keep them in tests, but...)

nalimilan:
Have you bumped into this while doing real work? Just curious. FWIW, we'll still get weird results if a data frame contains two subarrays sharing the same parent. I don't think Julia handles aliasing very well except for obvious cases like this one anyway.

gustafsson:
Yes I noticed this during "real work". `sort()` produced the right result but `sort!()` didn't.

In pre-processing data I had `d[:a]=f(d[:b])` but in some cases `f(x)` would simply just return its input. I solved it by returning `x*1` instead in those cases to avoid aliasing.

How can a DataFrame contain a SubArray? `upgrade_vector(adv::SubArray)` doesn't exist and not even `upgrade_vector(adv::AbstractArray)` so you'd have to collect the subarray before assigning it anyways. We don't have a `SubDataArray` unless I'm mistaken.

If `sort!` relied on `Base.permute!` any nested structures (such as a future `SubDataArray`) could implement `Base.permute!(a::SubDataArray, p)` and shuffle its indices instead of modifying the parent.

nalimilan:
Thanks, I'm somewhat relieved we don't support sub arrays. Not sure whether there might be a use case for them at some point.

gustafsson:
Thanks @ararslan

ararslan:
Thank you for the contribution!

PR #1075, opened 2016-09-27
Fix test failures on master

nalimilan:
Use vcat() instead of collect() in colwise(), and identity() instead of abs(),
since the latter do not work with Nullable.

--------------

I introduced test failures when merging https://github.com/JuliaStats/DataFrames.jl/pull/939, as the fact that test passed when the PR was opened does not indicate that it still works today (even if there are no merge conflicts). We should probably close and reopen all PRs filed before the port before merging them.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -304,12 +304,12 @@ colwise(sum, groupby(df, :a))
 ```
 
 """
-colwise(f::Function, d::AbstractDataFrame) = Any[collect(f(d[idx])) for idx in 1:size(d, 2)]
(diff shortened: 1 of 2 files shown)


gustafsson:
Per request; yes, `identity` here serves the same purpose as `abs` did.

And `vcat` serves the same purpose as `collect` of keeping array input as an array but creating a 1-element array from scalars.

The version with `collect` supported iterators, but more accidentally than intentionally (`vcat` doesn't collect iterators). But since iterators were never explicitly supported that's not an issue [grouping.jl#389](https://github.com/JuliaStats/DataFrames.jl/blob/master/src/groupeddataframe/grouping.jl#L389)

lgtm

PR #1076, opened 2016-09-27
Misc minor enhancements

alyst:
Several changes that looked too small and innocent for a separate PR:
  * more tests for corner-case joins/groupings/assignments
  * implement `AbstractVector` interface for `Cols` (mostly required to allow `map!(..., columns(df))` for `SubDataFrame`)
  * some optimizations for `GroupedDataFrame` (avoiding array reallocation, improving type inference)

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -79,13 +79,17 @@ abstract AbstractDataFrame
 ##
 ##############################################################################
 
-immutable Cols{T <: AbstractDataFrame}
+immutable Cols{T <: AbstractDataFrame} <: AbstractVector{Any}
     df::T
 end
 Base.start(::Cols) = 1
 Base.done(itr::Cols, st) = st > length(itr.df)
 Base.next(itr::Cols, st) = (itr.df[st], st + 1)
 Base.length(itr::Cols) = length(itr.df)
+Base.size(itr::Cols, ix) = ix==1 ? length(itr) : throw(ArgumentError("Incorrect dimension"))
+Base.size(itr::Cols) = (length(itr.df),)
+Base.linearindexing{T}(::Type{Cols{T}}) = Base.LinearFast()
+Base.getindex(itr::Cols, inds...) = getindex(itr.df, inds...)
 
 # N.B. where stored as a vector, 'columns(x) = x.vector' is a bit cheaper
 columns{T <: AbstractDataFrame}(df::T) = Cols{T}(df)
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -201,14 +201,7 @@ eltypes(df)
 ```
 
 """
-function eltypes(df::AbstractDataFrame)
-    ncols = size(df, 2)
-    res = Array(Type, ncols)
-    for j in 1:ncols
-        res[j] = eltype(df[j])
-    end
-    return res
-end
+eltypes(df::AbstractDataFrame) = map!(eltype, Vector{Type}(size(df,2)), columns(df))
 
 Base.size(df::AbstractDataFrame) = (nrow(df), ncol(df))
 function Base.size(df::AbstractDataFrame, i::Integer)
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -78,25 +78,23 @@ function stack(df::AbstractDataFrame, measure_vars::Vector{Int}, id_vars::Vector
                   [Compat.repeat(df[c], outer=N) for c in id_vars]...],      # id_var columns
               cnames)
 end
-function stack(df::AbstractDataFrame, measure_vars::Int, id_vars::Int)
-    stack(df, [measure_vars], [id_vars])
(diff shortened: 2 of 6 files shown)


src/abstractdataframe/reshape.jl
```diff
@@ -78,26 +78,24 @@ function stack(df::AbstractDataFrame, measure_vars::Vector{Int}, id_vars::Vector
...
-function stack(df::AbstractDataFrame, measure_vars::Int, id_vars::Vector{Int})
-    stackdf(df, [measure_vars], id_vars)
+function stack(df::AbstractDataFrame, measure_var::Int, id_vars::Vector{Int})
+    stackdf(df, [measure_var], id_vars)
 end
 stack(df::AbstractDataFrame, measure_vars, id_vars) =
     stack(df, index(df)[measure_vars], index(df)[id_vars])
-function stack(df::AbstractDataFrame, measure_vars)
+# no vars specified, by default select only numeric columns
+numeric_vars(df::AbstractDataFrame) = [T <: AbstractFloat || (T <: Nullable && eltype(T) <: AbstractFloat)
```
nalimilan: Numeric corresponds to `Number`, doesn't it? I don't know why the original code did that.

src/groupeddataframe/grouping.jl
```diff
@@ -203,15 +203,14 @@ Not meant to be constructed directly, see `groupby` abnd
...
+    vals::Vector{T}
 
-    function GroupApplied(gd, vals)
-        if length(gd) != length(vals)
-            error("GroupApplied requires keys and vals be of equal length.")
-        end
-        new(gd, vals)
+    @compat function (::Type{GroupApplied})(gd::GroupedDataFrame, vals::Vector)
+        length(gd) == length(vals) ||
+            throw(DimensionMismatch("GroupApplied requires keys and vals be of equal length."))
```
nalimilan: Could you add `(got $(length(gd)) and $(length(vals))` to the message?

src/groupeddataframe/grouping.jl
```diff
@@ -203,15 +203,14 @@ Not meant to be constructed directly, see `groupby` abnd
...
+        new{eltype(vals)}(gd, vals)
```
nalimilan: Probably more idiomatic (and maybe more efficient too) to make the eltype a parameter of the method.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -197,14 +201,7 @@ eltypes(df)
...
 """
-function eltypes(df::AbstractDataFrame)
-    ncols = size(df, 2)
-    res = Array(Type, ncols)
-    for j in 1:ncols
-        res[j] = eltype(df[j])
-    end
-    return res
-end
+eltypes(df::AbstractDataFrame) = map!(eltype, Vector{Type}(size(df,2)), columns(df))
```
nalimilan: Wouldn't it be more readable as a comprehension? And isn't type inference able to choose the best element type?

src/groupeddataframe/grouping.jl
alyst: I thought that type parameters create some overhead for dispatch. Here there are no inter-argument type constraints, so it's not necessary. In terms of type inference and the resulting code both versions should be equivalent. But I can change it, if using type parameters is the recommended way.

src/abstractdataframe/abstractdataframe.jl
alyst: I have tried several variants, including `map()` and comprehensions. Unfortunately, `eltype()` returns `Any`, so it results in "reds" for `@code_warntype eltypes(..)`. The proposed version gives the shortest "no-reds" code.

src/abstractdataframe/reshape.jl
alyst: I can rename it, but I don't have a better name idea. I agree that the default behaviour of stack() is somewhat inconsistent with e.g. joins requiring all joining columns to be specified, and I would rather remove this default. Just made the change to squeeze some duplicated code.

src/abstractdataframe/abstractdataframe.jl
ararslan: So `Type[eltype(col) for col in columns(df)]` has element type `Any`, even with the type annotation in the front?

alyst: `eltype(col)` has result type `Any`, and IIRC `eltype(col)::Type` doesn't help getting the type of array inferred automatically. Probably it would be as fast as `map!()`, I just wanted to have a version without any `Any` :)

src/groupeddataframe/grouping.jl
nalimilan: No idea...

src/abstractdataframe/reshape.jl
nalimilan: Agreed, let's deprecate this later.

src/groupeddataframe/grouping.jl
alyst: done

alyst:
@nalimilan Thanks for the review. I've just added fix for the `groupby()` doc. With this PR it would be easier for me to rebase #850

nalimilan:
Should we merge this now?

alyst:
@nalimilan Yes, if nobody objects.

Issue #818, opened 2015-06-17
ModelFrame and SubDataFrame


matthieugomez:
It seems like regressors are duplicated twice before fitting a linear model - once to return a `ModelFrame`, once to return a `ModelMatrix`.

Could [ModelFrame](https://github.com/JuliaStats/DataFrames.jl/blob/master/src/statsmodels/formula.jl#L237) return a `SubDataFrame` rather than a deep copy of the original `DataFrame`? 
One way to do that would be to replace `df[complete_cases(df)]` by `sub(df, complete_cases(df))` and move the `dropUnusedLevel` to `ModelMatrix`?

kleinschmidt:
I noticed the same thing, and I don't see any reason why not.  Probably want to restrict to the columns that are actually required to generate the `ModelMatrix` (the `etrms` field of the `Terms`).

kleinschmidt:
(and evaluate `complete_cases` with respect to only those columns, too)

nalimilan:
+1

johnmyleswhite:
I'm also in favor of this.

Any chance some of you will be at JuliaCon? I'd love to just sit down with all of the stakeholders and decide on the definitive semantics for transforming DataFrames into matrices.

nalimilan:
Sorry, I won't be there (but I haven't really thought about this issue so I'm not sure I would have been very helpful anyway...).

kleinschmidt:
I won't either, unfortunately.  I'd love to skype into a meeting if you end up getting a quorum though.

johnmyleswhite:
Maybe we should set up a Skype meeting post JuliaCon instead?

kleinschmidt:
Sure! Let me know when would work, my schedule is pretty open in the summer.

johnmyleswhite:
How about tentatively setting a meeting for sometime on July 15th? That gives us plenty of room to organize.

kleinschmidt:
👍

matthieugomez:
So for now, ModelMatrix(ModelFrame()) does 3+ copies of the data:
- once in ModelFrame via `na_omit`
- twice in ModelMatrix:  1 in `cols` (to convert to `Vector{Float64}`), once in `hcat` (to create the matrix).
In case of interactions, even more copies are created (in `expandcols`).

[...]

PR #1081, opened 2016-09-29
RFC: model matrix construction w/ fewer copies

kleinschmidt:
This makes it slightly more efficient to construct a `ModelMatrix` by constructing the matrix in place where possible, and `view`ing the columns that correspond to "evaluation terms" that get used in constructing higher order (interaction terms). 

Previously, each evaluation term was stored in a separate array, and only later were they all combined into a single matrix.

For model matrices with only main effects (which covers some non-trivial percentage of use cases I've seen), this means that the `ModelMatrix` constructor itself only copies the data once, into the matrix.


kleinschmidt:
Also, I suspect this could be made even more efficient by turning `expandcols` into some kind of `view` itself. The mapping from columns of the expanded block to the columns of the lower-order terms is the same as mapping from linear to cartesian indexing. But this might require some more substantial changes to the way that terms are handled in the first place.

src/statsmodels/formula.jl
```diff
@@ -444,59 +455,79 @@ creating the model matrix.
...
+    num_cols = convert(Int, terms.intercept)
+
+    for (i_term, term) in enumerate(terms.terms)
+        f = Compat.view(factors, :, i_term)
+        any(f) || continue
+
+        block_sizes = map((et,nr) -> nc(et, mf, non_redundant=nr),
+                          Compat.view(terms.eterms, f),
+                          Compat.view(terms.is_non_redundant, f, i_term))
+        num_term_cols = reduce(*, block_sizes)
```
ararslan: Is there a reason to prefer `reduce(*, x)` over `prod(x)`?

kleinschmidt: Nope. I think I first tried to write this as a `mapreduce` but `prod` makes more sense now that you mention it.

src/statsmodels/formula.jl
```diff
@@ -444,59 +455,79 @@ creating the model matrix.
...
+    factors = terms.factors
+
+    ## Step 1:
+    ## compute which columns correspond to which terms in order to pre-allocate
+    ## model matrix and set up views later
+    term_col_inds = []
+    num_cols = convert(Int, terms.intercept)
+
+    for (i_term, term) in enumerate(terms.terms)
+        f = Compat.view(factors, :, i_term)
```
nalimilan: `using Compat` at the top of the file would avoid the need to fully qualify names, and would have the advantage of being the actual Julia 0.5 syntax (so no updates needed in the future).

kleinschmidt: Actually, didn't we bump the required Julia to 0.5, making these unnecessary? (I discovered in trying to test this under 0.4.6...)

nalimilan: Right.

nalimilan:
@dmbates Anything to say about this?

@kleinschmidt I guess this fixes https://github.com/JuliaStats/DataFrames.jl/issues/818?

kleinschmidt:
I fixes (approximately) one of those copies. There's still the issue of copying the underlying data frame when constructing the `ModelFrame`. One of the proposals in #818 was to use a `SubDataFrame` instead of copying, but there are some mutating changes that happen (e.g. dropping unused levels from categorical columns) so I'm not really sure what the best way to go about that is. I suppose we could just check whether there are any mutating changes we need to make and if not use a `SubDataFrame`, and otherwise make a copy. Or, have a "mutating" constructor `ModelFrame!` that has permission to modify the underlying DataFrame, and throw an error to tell people to use that if there is any mutation required. Of course, that would introduce additional overhead of checking whether a mutation is required (by, for instance, checking that all levels of categorical variables are present).

nalimilan:
Do we need to actually drop the levels in the data frame? Can't we defer that to the last step (when building the model matrix)?

kleinschmidt:
It's necessary for turning the contrasts into `ContrastMatrix`es. We could check for missing levels in the `ContrastMatrix` constructor, but the contrasts code is written in a way that assumes that `levels(col)` accurately reflects that levels that are present (e.g., for checking whether there are missing or extra levels in a data column relative to a contrasts matrix). We'd have to replace calls to `levels(col)` with something that iterates over the elements.

nalimilan:
> We'd have to replace calls to levels(col) with something that iterates over the elements.

I was precisely thinking that we should define `unique(x::CategoricalArray)` to return the same thing as `droplevels!(x::CategoricalArray); levels(x)` (but of course without modifying `x`). Would that help?

kleinschmidt:
Yes, that should do it. Looking forward to a backend-independent implementation, we'll just need to make sure that `unique` is defined for the relevant fields.

kleinschmidt:
I _thought_ this would make things more efficient in constructing `ModelMatrix`es, but I did a bit of benchmarking and it doesn't seem to be the case: 20% slower than master and actually uses _more_ memory and makes more allocs. I'm at a bit of a loss to understand why.

https://gist.github.com/kleinschmidt/0125c14dd6d5b285e7bb82031e602ea1

nalimilan:
That's indeed puzzling. The number of allocations isn't that much higher either, so it doesn't seem to come from type instability. Have you tried to spot where the allocations happen `--track-allocation=user`?

src/statsmodels/formula.jl
```diff
@@ -444,59 +455,79 @@ creating the model matrix.
...
+        non_redundants = view(terms.is_non_redundant, ff, i_term)
+        if length(eterms) == 1 && term == eterms[1]
+            ## handle special case where term == eterm (e.g., main effect
+            ## terms). in this case, we can store the columns generated for the
+            ## eterm directly in the model matrix and just hold onto a view of
+            ## those columns for any higher-order terms that need them
+            et = eterms[1]
+            nr = non_redundants[1]
+            mm[:, term_col_inds[i_term]] = modelmat_cols(T, et, mf, non_redundant=nr)
+            eterm_cols[(et, nr)] = view(mm, :, term_col_inds[i_term])
```
kleinschmidt: It looks like the extra allocations are happening here: by one benchmark with 24Mb of allocations for the actual model matrix, there are 16Mb of allocations on this line.

ararslan:
The commits here got rekt after master switched back to DataArrays. If you think this is still something we should pursue, would you mind updating the PR, @kleinschmidt?

kleinschmidt:
Oof. I've switched my efforts on a tuple iterator based approach in StatsModels that makes zero copies and works with general tables, so I'm inclined to abandon this. In any case, we'd want to move it to StatsModels anyway and probably easier to make a clean start at this point.

ararslan:
Makes sense. Thanks for your efforts here nonetheless!

Issue #1082, opened 2016-09-30
Compact display of Nullables and IJulia


amellnik:
The [show method for Nullables ](https://github.com/JuliaLang/julia/blob/b36141f4799d73113b3597ee94d5925929b2ccd5/base/nullable.jl#L30) has two different displays-- a compact one which shows only the value or `#NULL`, and a longer one which also includes the type information as well as the word `Nullable`.  In general, the compact display is used, but one notable exception is with DataFrames in IJulia.  This leads to ungainly situations like the following on DataFrames master:

![image](https://cloud.githubusercontent.com/assets/7132414/19008542/f43d8540-8720-11e6-93d8-5257a8b198fe.png)

Is there any way to specify in DataFrames (or possibly IJulia) that Nullables should be displayed compactly, even when using the nicer HTML tables from IJulia?

johnmyleswhite:
This seems like it's just an oversight and that the HTML printing should be updated. Care to make a PR?

nalimilan:
If that helps, one needs to call `showcompact` or `sprint(showcompact, x)` to get the short output (instead of `show`, `print`, `string` or `repr`).

PR #1084, opened 2016-10-01
Better display of Nullables

amellnik:
As per #1082

### DIFF
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -232,8 +232,8 @@ function Base.show(io::IO, ::MIME"text/latex", df::AbstractDataFrame)
     write(io, "\t& ")
     header = join(map(c -> latex_escape(string(c)), cnames), " & ")
     write(io, header)
-    write(io, "\\\\ \n")
-    write(io, "\t\\hline \n")
+    write(io, "\\\\\n")
+    write(io, "\t\\hline\n")
     for row in 1:nrows
         write(io, "\t")
         write(io, @sprintf("%d", row))
@@ -249,7 +249,7 @@ function Base.show(io::IO, ::MIME"text/latex", df::AbstractDataFrame)
                 end
             end
         end
-        write(io, " \\\\ \n")
+        write(io, " \\\\\n")
     end
     write(io, "\\end{tabular}\n")
 end
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -184,7 +184,7 @@ end
         write(io, "<tr>")
         write(io, "<th>$row</th>")
         for column_name in cnames
-            cell = string(df[row, column_name])
+            cell = sprint(ourshowcompact, df[row, column_name])
             write(io, "<td>$(html_escape(cell))</td>")
         end
         write(io, "</tr>")
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -62,8 +62,9 @@ end
 #' ourshowcompact(STDOUT, "abc")
 #' ourshowcompact(STDOUT, 10000)
 ourshowcompact(io::IO, x::Any) = showcompact(io, x) # -> Void
-ourshowcompact(io::IO, x::AbstractString) = showcompact(io, x) # -> Void
+ourshowcompact(io::IO, x::AbstractString) = print(io, x) # -> Void
 ourshowcompact(io::IO, x::Symbol) = print(io, x) # -> Void
+ourshowcompact(io::IO, x::Nullable{String}) = isnull(x) ? showcompact(io, x) : print(io, get(x)) # -> Void
 
 #' @description
 #'
--- a/test/io.jl
+++ b/test/io.jl
@@ -498,17 +498,17 @@ module TestIO
                    C = [L"\alpha", L"\beta", L"\gamma", L"\sum_{i=1}^n \delta_i"],
                    D = [1.0, 2.0, Nullable(), 3.0]
                    )
-    @test isequal(reprmime(MIME("text/latex"), df),
-            """
-            \\begin{tabular}{r|cccc}
-            \t& A & B & C & D\\\\
-            \t\\hline
-            \t1 & 1 & \\\$10.0 & \$\\alpha\$ & 1.0 \\\\
-            \t2 & 2 & M\\&F & \$\\beta\$ & 2.0 \\\\
-            \t3 & 3 & A\\textasciitilde{}B & \$\\gamma\$ &  \\\\
-            \t4 & 4 & \\textbackslash{}alpha & \$\\sum_{i=1}^n \\delta_i\$ & 3.0 \\\\
-            \\end{tabular}
-            """)
+    str = """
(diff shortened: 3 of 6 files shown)


amellnik:
As per #1082

src/abstractdataframe/io.jl
```diff
@@ -184,8 +184,13 @@ end
         write(io, "<tr>")
         write(io, "<th>$row</th>")
         for column_name in cnames
-            cell = string(df[row, column_name])
+          if isnull(df[row, column_name])
```
nalimilan: You don't need this. Just call [`ourshowcompact`](https://github.com/amellnik/DataFrames.jl/blob/baeb0206fa3e0ac87fecd04d4fd48c5161e0d854/src/abstractdataframe/show.jl#L64) just like (`show`)[https://github.com/amellnik/DataFrames.jl/blob/baeb0206fa3e0ac87fecd04d4fd48c5161e0d854/src/abstractdataframe/show.jl#L255] does. Note in particular the `try...catch` block for undef entries.

Also, please use four-space indents just like in the rest of the code.

amellnik: I tried `ourshowcompact` but I think I might be missing something as I'm still seeing the quotes:
image

nalimilan:
Yeah, that's how it's printed at the REPL. I think it's good to be consistent, but we can discuss whether we should print the quotes at the REPL too. Looks like it's common in data base tools, including R and Pandas data frames, to omit quotes around strings. And indeed our docs for `outshowcompact` say that quotes are supposed to be omitted. I think you can just change the method for `AbstractString` to use `print` instead of `showcompact`, and it should be good. Be sure this is tested too.

I'll need to find how to make this work well for `CategoricalArray{String}`, but that's for another PR.

amellnik:
Both `print` and `showcompact` include the quotation marks, so I added stripping them to a `Nullable{String}` method for `ourshowcompact`.  If this approach is OK let me know, and I'll add tests and squash.

nalimilan:
Please avoid special-casing as much as possible. For this, make your code as similar as possible to `show` (see the link I provided above). The different printing methods shouldn't diverge unless there's a real reason to. So please get rid of the `isnull` branch, which isn't needed since `ourshowcompact` works for null values already.

src/abstractdataframe/show.jl
```diff
@@ -64,6 +64,7 @@ end
 ourshowcompact(io::IO, x::Any) = showcompact(io, x) # -> Void
 ourshowcompact(io::IO, x::AbstractString) = showcompact(io, x) # -> Void
 ourshowcompact(io::IO, x::Symbol) = print(io, x) # -> Void
+ourshowcompact(io::IO, x::Nullable{String}) = isnull(x) ? print(io, "NA") : print(io, replace(get(x),  r"^\"|\"$", "")) # -> Void
```
nalimilan: Much simpler and cleaner: `isnull(x) ? showcompact(io, x) : print(io, get(x))`.

src/abstractdataframe/show.jl
```diff
@@ -64,6 +64,7 @@ end
 ourshowcompact(io::IO, x::Any) = showcompact(io, x) # -> Void
 ourshowcompact(io::IO, x::AbstractString) = showcompact(io, x) # -> Void
```
nalimilan: Need to change `showcompact` to `print` if you want to remove quotes.

test/io.jl
```diff
@@ -490,4 +490,11 @@ module TestIO
     # Need to wrap macro call inside eval to prevent the error from being
     # thrown prematurely
     @test_throws ArgumentError eval(:(csv"foo,bar"a))
+
+    #Test HTML output for IJulia and similar
+    df = DataFrame(Fish = ["Suzy", "Amir"], Mass = [1.5, Nullable{Float64}()])
+    io = IOBuffer()
+    show(io, "text/html", df)
+    str = takebuf_string(io)
+    @test str == "<table class=\"data-frame\"><tr><th></th><th>Fish</th><th>Mass</th></tr><tr><th>1</th><td>Suzy</td><td>1.5</td></tr><tr><th>2</th><td>Amir</td><td>#NULL</td></tr></table>"
```
nalimilan: Can you break this into one line per row via `"""`?

test/show.jl
```diff
@@ -35,4 +36,11 @@ module TestShow
     show(io, A)
     A = DataFrames.RepeatedVector([1, 2, 3], 1, 5)
     show(io, A)
+
+    #Test show output for REPL and similar
+    df = DataFrame(Fish = ["Suzy", "Amir"], Mass = [1.5, Nullable{Float64}()])
+    io = IOBuffer()
+    show(io, df)
+    str = takebuf_string(io)
+    @test str == "2×2 DataFrames.DataFrame\n│ Row │ Fish │ Mass  │\n├─────┼──────┼───────┤\n│ 1   │ Suzy │ 1.5   │\n│ 2   │ Amir │ #NULL │"
```
nalimilan: Same here.

test/io.jl
```diff
@@ -490,4 +490,11 @@ module TestIO
     # Need to wrap macro call inside eval to prevent the error from being
     # thrown prematurely
     @test_throws ArgumentError eval(:(csv"foo,bar"a))
+
+    #Test HTML output for IJulia and similar
+    df = DataFrame(Fish = ["Suzy", "Amir"], Mass = [1.5, Nullable{Float64}()])
```
nalimilan: FWIW, you don't actually need to specify `{Float64}` since promotion does that for you.

test/io.jl
amellnik: On this one, the HTML output doesn't contain any linebreaks.  Is there any easy way to compare the strings ignoring `\n`?  The REPL version doesn't have this issue.

nalimilan: Ah, indeed. Don't worry about it then. Just rebase and it should be good to go.

nalimilan:
Thanks, but you still need to rebase against latest master.

nalimilan:
Test failed because of an extraneous ending `\n`.

amellnik:
Fixed my test.  The Latex test was failing because the Latex output was writing lines with trailing whitespace.  When I edited io.jl in Atom this whitespace was automatically trimmed, leading the test to fail.  I don't think the trailing whitespace was intentional, so I removed it and tests now pass.

src/abstractdataframe/io.jl
```diff
@@ -232,8 +232,8 @@ function Base.show(io::IO, ::MIME"text/latex", df::AbstractDataFrame)
     write(io, "\t& ")
     header = join(map(c -> latex_escape(string(c)), cnames), " & ")
     write(io, header)
-    write(io, "\\\\ \n")
```
nalimilan: @maximerischard Was the space intended?

nalimilan:
Thanks, sorry for the delay.

I suspect we'll have to add another special case for CategoricalArray printing. If you want to give it a look...

src/abstractdataframe/io.jl
maximerischard: I think it was just to make it a bit more readable than having 5 backslashes in a row.

PR #1086, opened 2016-10-01
handle `A ~ B - 1` and add tests

### LINKED ISSUE #574: Error in `Terms` when no intercept is requested
This does not work
```
df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"]);
ModelMatrix(ModelFrame(A~B-1, df))
ERROR: no method convert(Type{Expr}, Array{Any,1})
 in unshift! at array.jl:511
 in Terms at /Users/andreasnoackjensen/.julia/v0.3/DataFrames/src/formula/formula.jl:145
 in ModelFrame at /Users/andreasnoackjensen/.julia/v0.3/DataFrames/src/formula/formula.jl:174
```
but this does
```
julia> ModelMatrix(ModelFrame(A~B+(-1), df))
ModelMatrix{Float64}(4x1 Array{Float64,2}:
 1.0
 0.0
 0.0
 1.0,[1])
```

kleinschmidt:
Fixes #574.

### DIFF
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -40,8 +40,7 @@ type Terms
     intercept::Bool       # is there an intercept column in the model matrix?
 end
 
-import Base.==
-==(t1::Terms, t2::Terms) = all( getfield(t1, f)==getfield(t2, ) for f in fieldnames(t1))
+Base.:(==)(t1::Terms, t2::Terms) = all(getfield(t1, f)==getfield(t2, f) for f in fieldnames(t1))
 
 type ModelFrame
     df::AbstractDataFrame
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -40,6 +40,9 @@ type Terms
     intercept::Bool       # is there an intercept column in the model matrix?
 end
 
+import Base.==
+==(t1::Terms, t2::Terms) = all( getfield(t1, f)==getfield(t2, ) for f in fieldnames(t1))
+
 type ModelFrame
     df::AbstractDataFrame
     terms::Terms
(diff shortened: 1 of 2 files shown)


src/statsmodels/formula.jl
```diff
@@ -40,6 +40,9 @@ type Terms
     intercept::Bool       # is there an intercept column in the model matrix?
 end
 
+import Base.==
+==(t1::Terms, t2::Terms) = all( getfield(t1, f)==getfield(t2, ) for f in fieldnames(t1))
```
nalimilan: We don't seem to `import` functions in this file.

kleinschmidt: It's not strictly necessary, just a convenience for testing `Terms`. I could move it to the test file?

nalimilan: I just meant that you should write `Base.:(==)(...) = ...`. Though I just noticed `f` seems to be missing in `getfield(t2, )`?

src/statsmodels/formula.jl
```diff
@@ -85,19 +87,26 @@ function dospecials(ex::Expr)
...
-    if a1 != :* return excp end
-    aa = excp.args
-    a2 = aa[2]
-    a3 = aa[3]
-    if length(aa) > 3
-        excp.args = vcat(a1, aa[3:end])
-        a3 = dospecials(excp)
+    if a1 == :-
+        a2, a3 = excp.args[2:3]
+        a3 == 1 || error("invalid expression $ex; subtraction only supported or -1")
```
nalimilan: or -> for

test/formula.jl
```diff
@@ -53,9 +53,7 @@ module TestFormula
     @test t.intercept == false
     @test t.terms == [:x1, :x2]
 
-    t = Terms(y ~ -1 + x1 + x2)
-    @test t.intercept == false
-    @test t.terms == [:x1, :x2]
+    @test t == Terms(y ~ -1 + x1 + x2) == Terms(y ~ x1 - 1 + x2) == Terms(y ~ x1 + x2 -1)
```
nalimilan: Maybe also test for the error case?

src/statsmodels/formula.jl
kleinschmidt: That would explain the test failures :)  I tried it without the `import` first but didn't realize that `Base.:(==)` was the right way to refer to `==` in that context, thanks.

PR #1090, opened 2016-10-04
JSON Read/Write for DataFrames

joshbode:
Adds `readjson()` and a `JSON.lower` representation for `DataFrames`

Serialises/Deserialises data frames to/from JSON in the form:

```
{"names": [...], "types": [...], "columns": [[...], ..., [...]]}
```

where:
- `names` is the column names (in order)
- `types` is the type name for each column (in order), e.g. `Int64` -> `"Int64"`.
  
  Permits JSONable types, e.g. subtypes of `String`, `Char`, `Number`, `Date` and `DateTime`
- `columns` is the columns (in order).

(required completed PR: JuliaIO/JSON.jl#174 -> v0.8.0 which added JSON representations for `Char` and `Nullable`)


nalimilan:
Thanks, looks nice. Though we're rather trying to move data import/export functions *out* of this package, like `readtable` -> CSV.jl. Do you think you could take inspiration from the latter package, and write something which would integrate with the DataStreams.jl framework? That could be a separate package under the JuliaData organization, or it could remain inside JSON.jl.

joshbode:
Thanks! Guessing DataFrames.jl is getting too big :)
I'll take a look at CSV.jl and DataStreams.jl and see what I can do!
Not sure whether it would be appropriate to go in JSON.jl or not, but I'll find out.

ararslan:
Since JSON.jl is relied upon pretty heavily by the web stack packages and DataFrames.jl is somewhat of a heavyweight dependency, I think it would make more sense for JSON/DataFrame interoperability to be in a separate, lighter-weight package.

nalimilan:
> Since JSON.jl is relied upon pretty heavily by the web stack packages and DataFrames.jl is somewhat of a heavyweight dependency, I think it would make more sense for JSON/DataFrame interoperability to be in a separate, lighter-weight package.

It wouldn't depend on DataFrames, just on DataStreams. But yeah, until we have optional dependencies, that's a bit annoying.

ararslan:
> It wouldn't depend on DataFrames, just on DataStreams.

Even so, I tend to be rather purist when it comes to adding dependencies to widely used packages. 😜 Every time Travis submits coverage, it would have to download and precompile DataStreams as one of the JSON dependencies, which seems unfortunate. That's why I think this would be better as a separate package. Though if the JSON maintainers want that in there, that's their prerogative. This is just my two cents.

TotalVerb:
I don't think this is appropriate for JSON.jl. A separate package would be great.

joshbode:
Agree - I'm starting work on a new package using JSON.jl and DataStreams.jl, which I think I'll call JSONStreams.jl (unless someone has a better name!)

Issue #1018, opened 2016-07-21
DataFrames seems a little bloated with modeling-specific functionality


kleinschmidt:
There's lots of functionality here that's not specific about data frames, like the formula language, contrast coding (if #870 is merged), model matrix construction, etc. Is it time to refactor the modeling-oriented bits out into one or more JuliaStats packages? I'd suggest something like
- Formulas: model specification DSL
- Contrasts: converting categorical data into numerical matrices for modeling, in a way that's agnostic to the underlying data type (e.g., PooledDataArray, CategoricalArray, etc.)
- DataFramesModels: `ModelFrame`, `ModelMatrix`, `DataFrameRegressionModel`, etc. types.

Alternatively, we could just have one big DataFramesModels package that has all the modeling stuff that's currently in DataFrames.jl. It's not immediately clear to me how to cleanly separate the `ModelMatrix` and `ModelFrame` logic from the formula and contrasts logic, but that might be doable.

nalimilan:
Let's move these to StatsBase?

ararslan:
Then wouldn't StatsBase have a dependency on DataFrames? That seems a little weird to me; IMO StatsBase should be agnostic of how the underlying data is stored. Unless I'm misunderstanding you, @nalimilan.

(Welcome back from vacation, by the way! ☀️)

quinnj:
What we should really do is build the Formulas, Contracts, and Models stuff to all be based on an AbstractDataFrame type (or as yet unannounced, [AbstractTable](https://github.com/JuliaData/AbstractTables.jl)). 

With all the "AbstractTable" code able to live on it's own, it makes things like Formulas/Models much easier to split out because they only have a dependency on the AbstractTable (small, simple definitions of the Table interface) instead of all DataFrames (which would be a full implementation of the AbstractTable interface).

ararslan:
So if someone has data stored in an actual matrix, as in `Array{Whatever,2}`, they'll have to convert their data to a table type for use in modeling? Hm, I guess that does make sense, otherwise you don't have a clear way of referring to specific columns in the data for modeling purposes. (Well, you *do*; you have the column's position. But specifying a model in terms of matrix column positions sounds like a disaster.)

ararslan:
Btw @quinnj, I'd love to help out with JuliaData stuff. 😄

kleinschmidt:
[...]

PR #1097, opened 2016-10-06
Update StatsBase.df to dof

ararslan:
why is this even defined here

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.5
 NullableArrays 0.0.8
 CategoricalArrays 0.0.6
-StatsBase 0.8.3
+StatsBase 0.11.0
 GZip
 SortingAlgorithms
 Reexport
--- a/src/statsmodels/statsmodel.jl
+++ b/src/statsmodels/statsmodel.jl
@@ -62,7 +62,7 @@ typealias DataFrameModels @compat(Union{DataFrameStatisticalModel, DataFrameRegr
 @delegate DataFrameModels.model [StatsBase.coef, StatsBase.confint,
                                  StatsBase.deviance, StatsBase.nulldeviance,
                                  StatsBase.loglikelihood, StatsBase.nullloglikelihood,
-                                 StatsBase.df, StatsBase.df_residual, StatsBase.nobs,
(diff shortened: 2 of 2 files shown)


ararslan:
We should tag a patch release once this is merged, but we'll have to cherry-pick around the nullable change.

simonster:
This is defined here for models fit on DataFrames because we wrap them in a `DataFrameStatisticalModel`/`DataFrameRegressionModel`. We do this to allow models to be fit on DataFrames without requiring explicit package support, but while still keeping track of the mapping between coefficients and columns.

ararslan:
@simonster Oh okay, makes sense. Thanks for the explanation!

nalimilan:
Yes, these should be moved to a separate package together with the model frame code.

simonster:
Maybe? From a usability perspective, it seems like it may be hard for people to figure out that to fit a GLM on a DataFrame, you need `GLM`, `DataFrames`, and then some other package that makes them talk to each other.

nalimilan:
We could still reexport it by default. The idea is 1) to shrink the large code base of Data Frames, and 2) allow fitting models from alternative data sources (TypedTables, databases...).

ararslan:
Or perhaps GLM could reexport it?

nalimilan:
Possibly, but that would mean adding this to every modeling package, which are more numerous than data sources.

ararslan:
Ah yeah. Well, I can start the process of migration... Repo name? StatModels.jl, in keeping with the name of the file in which this stuff is defined here? That's pretty general though.

nalimilan:
The name needs to be general since that's a general framework. StatModels isn't bad.

Though before doing that work, make sure everybody agrees with that plan by asking for comments somewhere. I think I remember some opposition.

Cc: @dmbates @kleinschmidt @andreasnoack

ararslan:
Oh sure, I wouldn't take anything out of here before getting comments. Also cc @johnmyleswhite.

If we're going to do this I also think it would be a good time to rethink how formulas are specified. It would be nice if base Julia could stop making a special case to parse ~ as an infix macro for the sake of formulas. Doug brought up alternative syntax on the mailing list, wherein he proposed using pairs instead. Anyway, something to think about when we get there.

kleinschmidt:
Might make sense to discuss this under #1018...

Issue #1098, opened 2016-10-10
Conversion to html?


jw3126:
I have a dataframe, which is large but not huge (few hundreds of rows) and I need to convert it to html. Is there a recommended way to do this? 
```
show(io, MIME{Symbol("text/html")}(), df)
```
used to work a long time ago (see #501), but now truncates the frame.

nalimilan:
@johnmyleswhite Maybe we should have a mechanism for printing the full data frame?

nalimilan:
I think we should check the `limit` property of the `IOContext` before skipping some rows. @jw3126 Would you make a pull request?

jw3126:
Sure!

PR #1099, opened 2016-10-11
limit attribute of IOContext is used for html generation

jw3126:
See #1098

### DIFF
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -180,7 +180,7 @@ end
     haslimit = get(io, :limit, true)
     n = size(df, 1)
     if haslimit
-        tty_rows, tty_cols = displaysize(io)
+        tty_rows, tty_cols = _displaysize(io)
         mxrow = min(n,tty_rows)
     else
         mxrow = n
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -169,7 +169,6 @@ function html_escape(cell::AbstractString)
 end
 
 @compat function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame)
-    n = size(df, 1)
     cnames = _names(df)
     write(io, "<table class=\"data-frame\">")
     write(io, "<tr>")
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/io.jl
```diff
@@ -178,8 +177,14 @@ end
         write(io, "<th>$column_name</th>")
     end
     write(io, "</tr>")
-    tty_rows, tty_cols = _displaysize(io)
-    mxrow = min(n,tty_rows)
+    haslimit = get(io, :limit, true)
+    n = size(df, 1)
+    if haslimit
+        tty_rows, tty_cols = displaysize(io)
```
nalimilan: Use `_displaysize` for compatibility with 0.4.

test/io.jl
```diff
@@ -516,4 +516,15 @@ module TestIO
     show(io, "text/html", df)
     str = takebuf_string(io)
     @test str == "<table class=\"data-frame\"><tr><th></th><th>Fish</th><th>Mass</th></tr><tr><th>1</th><td>Suzy</td><td>1.5</td></tr><tr><th>2</th><td>Amir</td><td>#NULL</td></tr></table>"
+
+    # test limit attribute of IOContext is used
+    df = DataFrame(a=collect(1:1000))
+    ioc = IOContext(IOBuffer(), displaysize=(10, 10), limit=false)
+    show(ioc, MIME{Symbol("text/html")}(), df)
```
nalimilan: I think `MIME"text/html"` is equivalent and shorter.

test/io.jl
```diff
@@ -516,4 +516,15 @@ module TestIO
...
+    @test ioc.io |> takebuf_string |> length > 10000
```
nalimilan: Use standard syntax for nested function calls, as it's used elsewhere in the code. Also, maybe you could use `countlines` for a more precise check?

jw3126: I just checked, there are currently no linebreaks in the html anyway.

jw3126:
The default is the same as before (e.g. coupled to displaysize). Only if you set IOContext limit=false the output will be unlimited.

src/abstractdataframe/io.jl
nalimilan: Yeah, but adding an underscore doesn't hurt, and we could imagine backporting things. If we remove 0.4 support, we should remove `_displaysize` too, not just stop using it.

ararslan: Right

PR #1108, opened 2016-10-17
Loosen constructor for a DataFrame

andyferris:
It seems to be a little more difficult than it should be to pass a container of columns to the `DataFrame`'s constructor. This change lets you, e.g.

```julia
DataFrame([[1,2,3]], [:a])
```

Before this would complain because `[[1,2,3]]` is of type `Vector{Vector{Int}}` which is not a subtype of (concrete) type `Vector{Any}` (type parameters aren't convariant). This will automatically loosen the type parameter for the column container, if necessary.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -104,7 +104,7 @@ end
 
 function DataFrame(columns::AbstractVector,
                    cnames::AbstractVector{Symbol} = gennames(length(columns)))
-    return DataFrame(convert(Vector{Any}, columns), Index(cnames))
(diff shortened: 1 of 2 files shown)


nalimilan:
Why not, though this triggers an ambiguity error that should be fixed. Could you add a test too?

nalimilan:
Thanks. Still needs a test though.

andyferris:
OK. I'm done now. Sorry, I've only had a few seconds here and there.

andyferris:
Hmm now `==` for DataFrame doesn't seem to do what I expected... (it is false if the eltypes are the same... I was expecting the same behaviour as `Array` (`[1,2,3] == [1.0,2.0,3.0]`)

nalimilan:
Thanks!

PR #1109, opened 2016-10-18
Use the tagged version of Documenter

davidanthoff:
@MichaelHatherly You put in the ``clone`` originally, just double checking that you are ok with this change here.


MichaelHatherly:
Thanks, that was probably there from before the package was registered I guess.

PR #1112, opened 2016-10-22
fix typo in Nullable holding 1 example

bkamins:


nalimilan:
Thanks! I thought I had fixed this one already, but apparently not...

nalimilan:
Ah, that was in https://github.com/JuliaStats/DataFrames.jl/pull/1077.

PR #990, opened 2016-06-08
vcat should expand pooled columns when needed

gustafsson:
`pool!` creates compact pooled dataframes that doesn't have room for more levels than needed.

When vcat-ing together such pooled dataframes the resulting pool might be need to be larger. This PR checks which levels that are needed prior to copying the data.

Canonical example

```
a = DataFrame(x=pool(1:200))
b = DataFrame(x=pool(100:300))
vcat(a,b)
```

A similar problem with DataArrays occurs for `vcat(pool(1:200), pool(100:300))` but that is not addressed by this PR.

PR #1118, opened 2016-11-01
[RFC] Base.vcat AbstractDataFrame should rely on Base.vcat for columns

gustafsson:
This creates unnecessary copies of columns that doesn't exist in all concatenated dataframes. But uses Base.vcat to let each array type decide what vcat means.

Related #990


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,22 +685,32 @@ Base.vcat(df::AbstractDataFrame) = df
...
-    for j in 1:length(colnams)
-        colnam = colnams[j]
-        col = similar(similars[j], coltyps[j], Nrow)
-
-        i = 1
-        for df in dfs
-            if haskey(df, colnam)
-                copy!(col, i, df[colnam])
+    nrows = sum(nrow, dfs)
+    for colnam in unique([(names(e) for e in dfs)...;])
```
nalimilan: Avoid splatting on an arbitrary number of arguments when possible. Here, better use ` unique(Base.flatten(names.(df)))`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,22 +685,32 @@ Base.vcat(df::AbstractDataFrame) = df
...
+        k = Bool[haskey(e, colnam) for e in dfs]
+        c = vcat((e[colnam] for e in view(dfs, k))...)
```
nalimilan: `view` is probably not worth it as `dfs` is just a short vector of references. Could use a `if` guard instead.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,22 +685,32 @@ Base.vcat(df::AbstractDataFrame) = df
...
+        if all(k)
+            col = c
+        else
+            col = if _isnullable(c)
```
nalimilan: Better assign to `col` separately within each branch. Also, wrong indentation below.

test/grouping.jl
```diff
@@ -74,4 +74,19 @@ module TestGrouping
     @test isequal(gd[2], DataFrame(Key1="B", Key2="A", Value=3))
     @test isequal(gd[3], DataFrame(Key1="A", Key2="B", Value=2))
     @test isequal(gd[4], DataFrame(Key1="A", Key2="A", Value=1))
+
+    a = DataFrame(x=categorical(1:200))
+    b = DataFrame(x=categorical(100:300))
+    a[:x] = compact(a[:x])
+    b[:x] = compact(b[:x])
+    r = vcat(a,b)
+    @test isequal(r, DataFrame(x=[categorical(1:200);categorical(100:300)]))
```
nalimilan: Please add spaces after commas and semicolons here and below.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,22 +685,32 @@ Base.vcat(df::AbstractDataFrame) = df
...
-        colnam = colnams[j]
-        col = similar(similars[j], coltyps[j], Nrow)
-
-        i = 1
-        for df in dfs
-            if haskey(df, colnam)
-                copy!(col, i, df[colnam])
+    nrows = sum(nrow, dfs)
+    for colnam in unique([(names(e) for e in dfs)...;])
+        k = Bool[haskey(e, colnam) for e in dfs]
```
nalimilan: Could find a more explicit/logical names than `k`, `e` and `c`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,22 +685,32 @@ Base.vcat(df::AbstractDataFrame) = df
...
+            col = if _isnullable(c)
+              similar(c, nrows)
+            else
+              NullableArray{eltype(c)}(nrows)
+            end
+
+            i = 1
+            j = 1
+            for df in dfs
+                if haskey(df, colnam)
```
nalimilan: Could reuse `k` here.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,22 +685,32 @@ Base.vcat(df::AbstractDataFrame) = df
...
+                    copy!(col, i, view(c, j:j+nrow(df)-1))
```
nalimilan: You don't need `view` as `copy!` can take offsets and number of elements to copy directly.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,22 +685,32 @@ Base.vcat(df::AbstractDataFrame) = df
...
-                copy!(col, i, df[colnam])
+    nrows = sum(nrow, dfs)
+    for colnam in unique([(names(e) for e in dfs)...;])
+        k = Bool[haskey(e, colnam) for e in dfs]
+        c = vcat((e[colnam] for e in view(dfs, k))...)
+        if all(k)
+            col = c
+        else
+            col = if _isnullable(c)
+              similar(c, nrows)
```
nalimilan: Choosing the column type that way isn't correct. Until https://github.com/JuliaLang/julia/issues/18472 is fixed, I think we should be able to find out what's the most appropriate return type by calling `Base.return_types` on `vcat` and the types of the input columns. If inference fails, we could fall back to allocating an empty `Array` of the required length when the column is missing, and calling `vcat` on all columns; it would be wasteful, but that would only happen with broken types (and until we have a better mechanism).

src/abstractdataframe/abstractdataframe.jl
gustafsson: I don't follow, could you show me?

src/abstractdataframe/abstractdataframe.jl
gustafsson: Do you mean that I should assign to `res[colnam] ` separately within each branch?

src/abstractdataframe/abstractdataframe.jl
nalimilan: Something like `vcat((dfs[i][colnam] for i in 1:length(dfs) if k[i])...)`.

src/abstractdataframe/abstractdataframe.jl
gustafsson: I don't follow. `vcat` decides the column type, I'm just allocating a larger array. If vcat settled on something that couldn't cope with nulls I add that. Maybe I'm missing the point and you're actually disagreeing with the whole wasteful approach of allocating columns "twice"?

src/abstractdataframe/abstractdataframe.jl
gustafsson: Oh I didn't know about `Base.flatten`. And I get to use the new cool `names.`-notation!

src/abstractdataframe/abstractdataframe.jl
nalimilan: Ah, I misread the code. So indeed that's correct, just suboptimal. You could use `Base.return_types` as I described to avoid doing this when inference works, and only fall back to it when it fails.

src/abstractdataframe/abstractdataframe.jl
gustafsson: Oh, I didn't know you could use `if` in a `for`-generator

src/abstractdataframe/abstractdataframe.jl
nalimilan: Yeah, more or less, though in the second branch, I think I'd call `vcat` on existing columns as you do now for consistency. Also, you'll have to take the first result from `return_types`, and `T<:Union` should be `isleaftype(T)`.

gustafsson: If you do `r = vcat(cols)` in both branches of `if isleafttype(T)` they become identical so you don't need the branch in the first place. No?

I thought the problem was that you wanted some other way to allocate an array with a nullable eltype? Is that related? What benefit does `return_types` have over `similar`?

I don't know what you want to achieve so my questions become rather random.

nalimilan: Sorry if I wasn't clear. The goal is to avoid an unneeded allocation. My idea was that if `isleaftype(T)`, then you can create the (empty) final array, and fill it directly with `copy!`. To be sure you can have missing values, add `NullableArray{promote_type(eltypes.(c)...)}` to the types of existing columns.

If `!isleaftype(T)`, then the only way to find out the type of that array is to create it by actually calling `vcat`. Since you need to do that, you can as well create empty `NullableArrays` of the needed length and add them to the `vcat` call so that it gives the final array directly. But honestly, I don't expect this branch to happen often in practice; we could almost raise an error.

Does that make sense?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,78 +685,66 @@ Base.vcat(df::AbstractDataFrame) = df
...
+    nrows = sum(nrow, dfs)
+    for colnam in unique(Base.flatten(names.(dfs)))
+        k = Bool[haskey(df, colnam) for df in dfs]
+        C = Base.return_types(vcat, [typeof(dfs[i][colnam]) for i in 1:length(dfs) if k[i]])
+
+        if length(C)==1 && isleaftype(C[1])
+            if _isnullable(C[1])
+              NC = C[1]
+            else
+              NC = NullableArray{eltype(C[1])}
```
gustafsson: I used the value from `return_types` instead of `NullableArray{promote_type(eltypes(c)}` as @nalimilan suggested.

gustafsson:
Implemented your comments, thanks. You're right that the return type is mostly stable. Meaning we'll only use `vcat` to find the array type, but not for concatenating data.

I only found one example in the unit tests where this fails though and `return_types(vcat, c)` doesn't give a leaf type. From [cat.jl:121](https://github.com/JuliaStats/DataFrames.jl/blob/master/test/cat.jl#L121) `Base.return_types` get's called like this:
```
Base.return_types(vcat, (NullableCategoricalArray{Int64,1,UInt32}, NullableArray{Int64,1}))
1-element Array{Any,1}:
 CategoricalArrays.NullableCategoricalArray{T,N,R<:Integer}
```

This is an example where the return type of `vcat` depend on the order of arguments. In this case it returns a `NullableCategoricalArray{Int64,1,UInt32}`.

gustafsson:
This PR depends on https://github.com/JuliaStats/NullableArrays.jl/pull/152

nalimilan:
Thanks, will give a longer look later.

> This is an example where the return type of vcat depend on the order of arguments. In this case it returns a NullableCategoricalArray{Int64,1,UInt32}.

Sounds like we should change this to be type-stable. I guess the problem is in the method defined in CategoricalArrays.jl?

gustafsson:
@nalimilan the example where the return type of vcat depend on the order of arguments comes from [`Base.typed_vcat`](https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L1025) which uses similar on its first argument.

gustafsson:
Related; CategoricalArrays might need a special `vcat(V1::CategoricalArray,V::AbstractVector...)` that either creates a regular `Array` or arrange so that [a[pos:p1] = Vk](https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L1030) in `typed_vcat` uses the equivalent of `CategoricalArrays.copy!` somehow. But's that's really an issue for `CategoricalArrays` not `DataFrames`.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Still could use more explicit names.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,78 +685,52 @@ Base.vcat(df::AbstractDataFrame) = df
...
+        if all(k)
+            res[colnam] = vcat((dfs[i][colnam] for i in 1:length(dfs))...)
+            continue
         end
 
-        res[colnam] = col
-    end
-    res
-end
+        c = ((typeof(dfs[i][colnam]) for i in 1:length(dfs) if k[i])...,)
```
nalimilan: Trailing comma isn't needed, right?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,78 +685,52 @@ Base.vcat(df::AbstractDataFrame) = df
...
-                    # Needed on Julia 0.4 since e.g.
-                    # promote_type(Nullable{Int}, Nullable{Float64}) gives Nullable{T},
-                    # which is not a usable type: fall back to Nullable{Any}
-                    if VERSION < v"0.5.0-dev" &&
-                       coltyps[idx] <: Nullable && !isa(coltyps[idx].types[2], DataType)
-                        coltyps[idx] = Nullable{Any}
-                    end
+        if length(C)==1 && isleaftype(C[1])
+            if _isnullable(C[1])
+              NC = C[1]
```
nalimilan: Four-space indent.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,78 +685,52 @@ Base.vcat(df::AbstractDataFrame) = df
...
-    end
-    colnams = _names(colindex)
 
-    coltyps, colnams, similars
+        res[colnam] = col
+    end
+    res
 end
 
+_isnullable{T}(::AbstractArray{T}) = T <: Nullable
```
nalimilan: Better define these above since they're used there.

nalimilan:
> @nalimilan the example where the return type of vcat depend on the order of arguments comes from Base.typed_vcat which uses similar on its first argument.

OK, I guess that's good enough for now. Looks like it's going to be changed pretty soon by https://github.com/JuliaLang/julia/pull/16740/files#diff-2264bb51acec4e7e2219a3cb1c733651R1186.

Though we'll still need a mechanism to promote the input types (https://github.com/JuliaLang/julia/issues/18472). Since you're now familiar with `vcat`, would you experiment with a PR against Base? One approach would be to use `promote` for that; you could add a `promote_rule` method for `AbstractArray` (similar to what I recently did for `Pair` at https://github.com/JuliaLang/julia/pull/19171).

> Related; CategoricalArrays might need a special vcat(V1::CategoricalArray,V::AbstractVector...) that either creates a regular Array or arrange so that a[pos:p1] = Vk in typed_vcat uses the equivalent of CategoricalArrays.copy! somehow. But's that's really an issue for CategoricalArrays not DataFrames.

It seems that indeed `a[pos:p1] = Vk` should be equivalent to `copy!`, shouldn't it? But could you elaborate on how it's different? As regards handling the levels?

CategoricalArray also likely need specific methods so that concatenation with `AbstractArray{<:Nullable}` gives a `NullableCategoricalArray`. Ideally it would be done via the `promote` mechanism, but currently it seems we need a `vcat` method.

gustafsson:
Since we don't handle `setindex!(::CategoricalArray, ::AbstractArray, ::AbstractArray)` the call to `a[pos:p1] = Vk` will call the default implementation in base which in turn will call `setindex!` for each element. This is doesn't use the efficient concatenation implemented in `copy!` and `vcat` for CategoricalArray. Might just be a simple fix of implementing a `setindex!` that just calls `copy!`. I'm not sure if there are any side-effects to consider.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,76 +685,49 @@ Base.vcat(df::AbstractDataFrame) = df
...
-        col = similar(similars[j], coltyps[j], Nrow)
-
-        i = 1
-        for df in dfs
-            if haskey(df, colnam)
-                copy!(col, i, df[colnam])
-            end
-            i += size(df, 1)
+    nrows = sum(nrow, dfs)
+    for colnam in unique(Base.flatten(names.(dfs)))
```
nalimilan: `flatten` lives under `Base.Iterators` on Julia 0.6. You can just test this using `isdefined` and do `using Base: flatten` or `using Base.Iterators: flatten` at the top of the file.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -685,76 +685,49 @@ Base.vcat(df::AbstractDataFrame) = df
...
-                push!(colindex, cn)
-                push!(coltyps, ct)
-                push!(similars, col)
-                push!(nonnull_ct, !_isnullable(col))
+                j += nrow(dfs[i])
             end
-        end
-    end
+        else
+            # warn("Unstable return types: ", C, " from vcat of ", [typeof(dfs[i][colnam]) for i in 1:length(dfs) if k[i]])
```
nalimilan: Can you remove this once it's ready?

nalimilan:
OK. I'd rather implement `copy!` in terms of `setindex!`, though it doesn't really matter.

There's still a failure which appears to be real (besides the `flatten` problem on 0.6).

nalimilan:
Is there anything blocking here?

ararslan:
@nalimilan If you mean blocking a merge, your review still says "requested changes" and all checks failed on all platforms...

nalimilan:
I just wondered whether there were hard issues to tackle to get this to a mergeable state.

nalimilan:
@gustafsson Can you revise this so that we merge it?

nalimilan:
Actually, this cannot be merged until Base provides a mechanism for `cat` promotion (https://github.com/JuliaLang/julia/pull/20815). See also discussion at https://github.com/JuliaData/DataTables.jl/pull/30/files#r105499372.

nalimilan:
@cjprybol Do you think we still need this, or did you implement it in DataTables? (See the last four commits.)

cjprybol:
I think everything here should be covered by what we merged in https://github.com/JuliaData/DataTables.jl/pull/45. And until something like https://github.com/JuliaLang/julia/pull/20815 is implemented in Base, we can't rely on Base.vcat anyway, so this looks like it follows similar logic to what we have implemented currently.

cjprybol:
Thanks for your efforts here @gustafsson! Sorry it got left behind while we were experimenting in the DataTables package

PR #1121, opened 2016-11-02
Juno Display

MikeInnes:
This patch adds an extremely lightweight dependency on Juno.jl, and implements rich display inside of Juno, like so:

screenshot 2016-11-02 19 11 33

In future I'd like to improve on this by having access to the entire dataframe rather than truncating it, but this is something for now.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,4 +6,4 @@ SortingAlgorithms
 Reexport
 Compat 0.8.4
 FileIO 0.1.2
-Juno
+Juno 0.2.4
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,3 +6,4 @@ SortingAlgorithms
 Reexport
 Compat 0.8.4
 FileIO 0.1.2
+Juno
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -581,7 +581,7 @@ end
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
 
 using Juno
-import Juno: Inline, Tree, Table, Row, strong
+import Juno: Inline, LazyTree, Table, Row, strong
 
 const SIZE = 25
 
@@ -591,6 +591,6 @@ const SIZE = 25
   header = map(x->strong(string(x)), names(frame)[1:width]')
   body = hcat([frame[1:height,i] for i = 1:width]...)
   view = Table(vcat(header, body))
-  Tree(Row(typeof(frame), text" ", Juno.dims(size(frame)...)),
-       [view])
+  LazyTree(Row(typeof(frame), text" ", Juno.dims(size(frame)...)),
+           () -> [view])
 end
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -581,14 +581,14 @@ end
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
 
 using Juno
-using Juno: Inline, LazyTree, Table, Row, strong, undefs
+using Juno: Inline, LazyTree, Table, Row, strong
 
 const SIZE = 25
 
 function tomat(df::AbstractDataFrame)
     res = Array{Any}(size(df))
     for (j, col) in enumerate(columns(df)), i = 1:length(col)
-      isassigned(col, i) && (res[i, j] = col[i])
+        isassigned(col, i) && (res[i, j] = col[i])
     end
     return res
 end
@@ -597,7 +597,7 @@ end
     width = min(size(df, 2), SIZE)
     height = min(size(df, 1), SIZE)
     header = map(x->strong(string(x)), names(df)[1:width]')
-    body = undefs(tomat(df))[1:height, 1:width]
+    body = Juno.undefs(tomat(df))[1:height, 1:width]
     view = Table(vcat(header, body))
     LazyTree(Row(typeof(df), text" ", Juno.dims(size(df)...)),
              () -> [view])
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -589,7 +589,7 @@ const SIZE = 25
     width = min(size(frame, 2), SIZE)
     height = min(size(frame, 1), SIZE)
     header = map(x->strong(string(x)), names(frame)[1:width]')
-    body = hcat([frame[1:height,i] for i = 1:width]...)
+    body = Array(frame[1:height, 1:width])
     view = Table(vcat(header, body))
     LazyTree(Row(typeof(frame), text" ", Juno.dims(size(frame)...)),
              () -> [view])
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -586,11 +586,11 @@ import Juno: Inline, LazyTree, Table, Row, strong
 const SIZE = 25
 
 @render Inline frame::AbstractDataFrame begin
-  width = min(size(frame, 2), SIZE)
-  height = min(size(frame, 1), SIZE)
-  header = map(x->strong(string(x)), names(frame)[1:width]')
-  body = hcat([frame[1:height,i] for i = 1:width]...)
-  view = Table(vcat(header, body))
-  LazyTree(Row(typeof(frame), text" ", Juno.dims(size(frame)...)),
-           () -> [view])
+    width = min(size(frame, 2), SIZE)
+    height = min(size(frame, 1), SIZE)
+    header = map(x->strong(string(x)), names(frame)[1:width]')
+    body = hcat([frame[1:height,i] for i = 1:width]...)
+    view = Table(vcat(header, body))
+    LazyTree(Row(typeof(frame), text" ", Juno.dims(size(frame)...)),
+             () -> [view])
 end
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -581,16 +581,16 @@ end
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
 
 using Juno
-import Juno: Inline, LazyTree, Table, Row, strong
+using Juno: Inline, LazyTree, Table, Row, strong
 
 const SIZE = 25
 
-@render Inline frame::AbstractDataFrame begin
-    width = min(size(frame, 2), SIZE)
-    height = min(size(frame, 1), SIZE)
-    header = map(x->strong(string(x)), names(frame)[1:width]')
-    body = Array(frame[1:height, 1:width])
+@render Inline df::AbstractDataFrame begin
+    width = min(size(df, 2), SIZE)
+    height = min(size(df, 1), SIZE)
+    header = map(x->strong(string(x)), names(df)[1:width]')
+    body = Array(df[1:height, 1:width])
     view = Table(vcat(header, body))
-    LazyTree(Row(typeof(frame), text" ", Juno.dims(size(frame)...)),
+    LazyTree(Row(typeof(df), text" ", Juno.dims(size(df)...)),
              () -> [view])
 end
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -585,7 +585,7 @@ using Juno: Inline, LazyTree, Table, Row, strong
 
 const SIZE = 25
 
-function tomat(df::AbstractDataFrame)
+function to_matrix(df::AbstractDataFrame)
     res = Array{Any}(size(df))
     for (j, col) in enumerate(columns(df)), i = 1:length(col)
         isassigned(col, i) && (res[i, j] = col[i])
@@ -597,7 +597,7 @@ end
     width = min(size(df, 2), SIZE)
     height = min(size(df, 1), SIZE)
     header = map(x->strong(string(x)), names(df)[1:width]')
-    body = Juno.undefs(tomat(df))[1:height, 1:width]
+    body = Juno.undefs(to_matrix(df))[1:height, 1:width]
     view = Table(vcat(header, body))
     LazyTree(Row(typeof(df), text" ", Juno.dims(size(df)...)),
              () -> [view])
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -581,15 +581,23 @@ end
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
 
 using Juno
-using Juno: Inline, LazyTree, Table, Row, strong
+using Juno: Inline, LazyTree, Table, Row, strong, undefs
 
 const SIZE = 25
+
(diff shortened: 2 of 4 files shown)


nalimilan:
This is nice,  but why does DataFrames need to depend on Juno for that? I'd rather put this into a separate package that Juno would use. It doesn't sound right to add dependencies on each GUI which will be developed in the future, especially since DataFrames is a dependency of a lot of packages.

src/abstractdataframe/show.jl
```diff
@@ -579,3 +579,18 @@ function showcols(io::IO, df::AbstractDataFrame) # -> Void
...
+using Juno
+import Juno: Inline, LazyTree, Table, Row, strong
+
+const SIZE = 25
+
+@render Inline frame::AbstractDataFrame begin
+  width = min(size(frame, 2), SIZE)
+  height = min(size(frame, 1), SIZE)
+  header = map(x->strong(string(x)), names(frame)[1:width]')
+  body = hcat([frame[1:height,i] for i = 1:width]...)
```
nalimilan: Splatting vectors of arbitrary length is generally not a good idea as it will trigger compilation of specialized methods without a clear gain. I think you can just use `Array(frame[1:height, 1:width])` here.

MikeInnes:
See for example the discussion over at Gadfly: https://github.com/GiovineItalia/Gadfly.jl/pull/902.

To a large extent this is a stop gap; we'll be able to remove these kinds of dependencies once conditional modules are working. Until then, the only realistic way for two large modules A and B to communicate, without one depending on the other, is to implement a small interface package that both depend on.

As long as the package is a small pure-Julia dependency there's very little downside. We're also working on making these interfaces more general (see e.g. WebDisplay.jl which Juno will move towards) so that a single package can provide an interface to multiple frontends.

MikeInnes:
Actually, there is an alternative, which is for DataFrames to provide the minimal interface rather than Juno. If there was a DataFramesCore.jl package which exported a minimal set of definitions (`AbstractDataFrame` alone may be sufficient in this case) without the functionality and dependencies, then Juno could depend on that. This could also be useful for packages like Gadfly and reduce the number of packages with hard dependencies on DataFrames generally.

ViralBShah:
I love the rich display!

nalimilan:
Actually that interface is precisely being developed right now at https://github.com/davidagold/AbstractTables.jl So I'd rather wait for it to be ready; anyway we won't make a new release without it (except maybe for bug fixes).

MikeInnes:
That package seems pretty substantial to me. If it provides a bulk of functionality (i.e. NullableArrays and StructuredQueries), as opposed to just the minimal interface, it's not what I mean, though it may be the right direction.

Aside from that, it doesn't seem like the DataFrames overhaul will be ready for some time. Can we figure something out for the existing release? If you have concrete concerns about the effects of this patch I can try to address those.

davidagold:
If the dependency is lightweight and Julia only, then I support this PR. 

I know that in principle we should be looking to remove dependencies from DF rather than add them. However, I consider the following mitigating points in this case: (i) integration of `DataFrame` and the Julia IDE is something I think we should encourage; (ii) there’s a clear path towards removing the dependency eventually (via `AbstractTables` or conditional module requirements); and (iii) there is evidently a plan to prevent the proliferation of such dependencies via WebDisplays.jl.

Re: AbstractTables.jl. This package could be a solution for this and related interfacing issues, but I've been questioning lately whether or not `AbstractTable` is the right abstraction. It serves functionality by means of inheritance (`AbstractTable` being an abstract type), but arguably an interface would serve better. For instance, ideally `CSV.Source`s would be able to rendered in Juno using the same rich display, but it's unnecessarily restrictive to require that `CSV.Source <: AbstractTable`. This will take time to think about and resolve. So for now, I think the present proposal is appropriate.

src/abstractdataframe/show.jl
```diff
@@ -579,3 +579,18 @@ function showcols(io::IO, df::AbstractDataFrame) # -> Void
 end
 
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
+
+using Juno
+import Juno: Inline, LazyTree, Table, Row, strong
```
nalimilan: Do you really need `import` rather than `using` here?

src/abstractdataframe/show.jl
```diff
@@ -579,3 +579,18 @@ function showcols(io::IO, df::AbstractDataFrame) # -> Void
...
+
+const SIZE = 25
+
+@render Inline frame::AbstractDataFrame begin
```
nalimilan: `frame` is more commonly called `df`.

src/abstractdataframe/show.jl
```diff
@@ -579,3 +579,18 @@ function showcols(io::IO, df::AbstractDataFrame) # -> Void
 end
 
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
+
+using Juno
+import Juno: Inline, LazyTree, Table, Row, strong
+
+const SIZE = 25
```
nalimilan: Don't you want to allow users to tweak this somewhere in Juno?

nalimilan:
OK, let's go with that then until we have optional dependencies. Though please add tests for the new feature.

src/abstractdataframe/show.jl
MikeInnes: Possibly, but we'd have to implement a Juno.jl API for that. I think a better approach would be to extend this view to be capable of showing the entire dataframe at once.

quinnj:
FWIW, I'm not jazzed about this, purely for the dependency reasons. I'd much rather create a separate package for now that can ship w/ a Juno bundle or Atom bundle by default that provides the functionality and requires DataFrames and Juno.

That and I think we need to at least play the part of "squeaky wheel" in Base for conditional modules and/or help in any way we can to push that forward. Bloating one of the most popular Julia packages with a "requirement" on optional stuff just seems like a cop out.

src/abstractdataframe/show.jl
nalimilan: OK. Though showing the full data frame could be quite problematic if it contains one million rows and one hundred columns. Better take a subset first.

MikeInnes:
It's inherently somewhat tricky to test conditional dependencies (with language support or not), since by definition the functionality is not available unless both packages are loaded. For now, I've added a simple test that Juno will be able to call the method; if you want I can add a test dependency on Atom.jl and add more comprehensive testing.

nalimilan:
If we add this feature, we should bite the bullet and test it thoroughly even if that means adding more requirements. Test dependencies are not so much a concern as runtime dependencies anyway.

MikeInnes:
> That and I think we need to at least play the part of "squeaky wheel" in Base for conditional modules and/or help in any way we can to push that forward.

FWIW, supporting the use cases for this feature is a great way to do that. Having packages in existence where the feature provides a clear immediate benefit (fewer deps) is far more compelling than a few people saying "wouldn't it be nice if...".

MikeInnes:
Ok, better test added.

nalimilan:
Thanks. The test failure is real, though.

src/abstractdataframe/show.jl
MikeInnes: Not if we lazy-load it as needed

src/abstractdataframe/show.jl
```diff
@@ -581,15 +581,23 @@ end
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
 
 using Juno
-using Juno: Inline, LazyTree, Table, Row, strong
+using Juno: Inline, LazyTree, Table, Row, strong, undefs
 
 const SIZE = 25
 
+function tomat(df::AbstractDataFrame)
```
nalimilan: This function is essentially `convert(Matrix, ...)`, except that it doesn't throw when undefined entries are present. Better improve the existing method then.

MikeInnes: `convert(Matrix, ...)` also throws on `NA`s so this would involve a change of semantics – are you sure that's what you want?

MikeInnes:
Tests will require https://github.com/JuliaLang/METADATA.jl/pull/6973 to pass

REQUIRE
```diff
@@ -6,3 +6,4 @@ SortingAlgorithms
 Reexport
 Compat 0.8.4
 FileIO 0.1.2
+Juno
```
ararslan: If the tests require a specific version of Juno to be able to pass, that should be set as the minimum required Juno version.

src/abstractdataframe/show.jl
nalimilan: Ah, indeed. Let's keep it that way, then, but please remove `tomat` to something more explicit (`to_matrix`?), as this one reads too much like 🍅.

src/abstractdataframe/show.jl
```diff
@@ -579,3 +579,26 @@ function showcols(io::IO, df::AbstractDataFrame) # -> Void
...
+
+using Juno
+using Juno: Inline, LazyTree, Table, Row, strong, undefs
+
+const SIZE = 25
+
+function tomat(df::AbstractDataFrame)
+    res = Array{Any}(size(df))
+    for (j, col) in enumerate(columns(df)), i = 1:length(col)
+      isassigned(col, i) && (res[i, j] = col[i])
```
tkelman: wrong indent

src/abstractdataframe/show.jl
```diff
@@ -579,3 +579,26 @@ function showcols(io::IO, df::AbstractDataFrame) # -> Void
 end
 
 showcols(df::AbstractDataFrame) = showcols(STDOUT, df) # -> Void
+
+using Juno
```
tkelman: not needed, redundant with following line

test/show.jl
```diff
@@ -35,4 +38,7 @@ module TestShow
     show(io, A)
     A = DataFrames.RepeatedVector([1, 2, 3], 1, 5)
     show(io, A)
+
+    render(Juno.Editor(), df)
```
tkelman: these should try to verify the result is correct instead of just that the method doesn't error

nalimilan:
Any idea why tests failed only on 0.4?

src/abstractdataframe/show.jl
MikeInnes: Fair enough, will go with a less fruity name :)

src/abstractdataframe/show.jl
MikeInnes: no

MikeInnes:
Ok, added a test guard for 0.4 – Juno only calls this method on 0.5 anyway.

test/show.jl
```diff
@@ -16,8 +16,10 @@ module TestShow
     showall(io, subdf)
     showall(io, subdf, true)
 
-    using Atom, Juno
-    render(Juno.Editor(), df)
+    if VERSION > v"0.5-"
+      using Atom, Juno
+      render(Juno.Editor(), df)
```
ararslan: Wrong indent here

ararslan:
Tony had requested--and I agree--that more descriptive commit messages be used, but the two most recent messages are "four word commit message" and just ".". The commits should be squashed on merge, but in case someone mistakenly merges without squashing, it would be nice if the commit messages at least vaguely reflected the changes implemented. I've been guilty of this as well, but I've been trying to be better about it. Anyway, just something to consider. 😉

MikeInnes:
The commit log in its current form is not a good way to convey intent regardless of the messages, because there's little ordering and it's mixed with lots of noise from small fixups. If the DataFrames maintainers particularly value having a pretty commit log (perhaps they want to hang it on their wall or something?) I'm more than happy to rebase once the patch is ready to go.

ViralBShah:
I would like to hang the julia commit log on my wall. :-)

test/show.jl
nalimilan: @MikeInnes Is there any way of doing this?

MikeInnes: The return value of this is a JSON representation designed to be consumed by the frontend; in other words it's an implementation detail that's not part of the API. I'm assuming you don't want the DataFrames tests to break if/when Juno's implementation changes.

tkelman: doesn't have to compare the exact output, but anything that the code in this package should be adding to the representation should be tested for as a sanity check. otherwise it could easily be broken by accident and regress here.

MikeInnes: I don't really see why. The important thing on the DataFrames side is that it's using Juno.jl's API correctly, and calling the function tests that. All of the actual functionality and logic that could regress is on the Juno side and can be tested there.

MikeInnes: Ok I've added a test for the implementation. Given that there's no logic it's equivalent to checking the source code hasn't changed, but whatever, it's a test.

MikeInnes:
On reflection – if this integration is barely worth a few extra characters in require, then it's certainly not worth the hours of time I'm now sinking into it. I would have liked to use this to start prototyping things like spreadsheet views, but that's not even a feature I'd use much personally, so if it's all the same to you I'm happy to focus on more productive things and stick with plain text output here indefinitely.

nalimilan:
Hey, keep cool, looks like the PR is ready now. :-) Sorry if the lack of support for optional dependencies made it more painful than it should.

@quinnj Are you OK with this living in DataFrames until we have a better solution?

quinnj:
I already "cast my vote", so to speak in saying I'd rather not, but I also realize that it's just one vote. Happy to go along if more people want this in.

nalimilan:
Not many people have voted, and I'm myself torn about this. But well, I think I'd rather have this feature leave here for now rather than not exist at all.

StefanKarpinski:
I vote for having it here since having it merged will keep it from bitrotting and its presence will prod people to figure out a better way to do it. The perfect is the enemy of the good and all that.

nalimilan:
Thanks @MikeInnes. Could you make the same PR against master so that we don't lose support in the next release?

MikeInnes:
Will do, thanks @nalimilan and others for reviewing and pushing this through.

src/abstractdataframe/show.jl
```diff
@@ -579,3 +579,28 @@ function showcols(io::IO, df::AbstractDataFrame) # -> Void
...
+    for (j, col) in enumerate(columns(df)), i = 1:length(col)
+        isassigned(col, i) && (res[i, j] = col[i])
+    end
+    return res
+end
+
+function _render(df::AbstractDataFrame)
+    width = min(size(df, 2), SIZE)
+    height = min(size(df, 1), SIZE)
+    header = map(x->strong(string(x)), names(df)[1:width]')
```
pfitzseb: [Here as well.](https://github.com/JuliaStats/DataFrames.jl/pull/1125/files#r88341867)

KristofferC: Transpose is recursive. 

https://github.com/JuliaLang/julia/issues/17834

pfitzseb: `reshape(names(df)[1:width], 1, 2)` works, btw.

PR #1122, opened 2016-11-03
Add thead and tbody tags

phillc73:
Added thead and tbody tags to table output for stricter HTML compliance and to support other formatting options e.g. datatables.net


ararslan:
Test failures appear to be unrelated.

PR #1126, opened 2016-11-14
Query docs backport

davidanthoff:
This backports #1105. I've left this as the same commits right now (some slightly modified to match the doc structure here), so that it is clearer where they come from. Might make sense to just squash everything...


ararslan:
I don't anticipate any bisecting issues here so we may as well squash on merge.

davidanthoff:
I think the travis failure is just a timeout, plus only on julia nightly.

nalimilan:
It's now online, but there's a display bug: http://juliastats.github.io/DataFrames.jl/stable/man/querying_frameworks/

nalimilan:
BTW, it looks like you forgot to add the new page to the table of contents (in both branches).

davidanthoff:
@MichaelHatherly Any idea why [this](http://juliastats.github.io/DataFrames.jl/stable/man/querying_frameworks/) mangels up the display? If I build the release-0.8 branch doc locally, everything looks good.

I think there are two problems: 1) that the first ``@setup`` block is not properly detected, and 2) that warnings seem to be printed in the output. Aren't warnings printed to stderr? Documenter should probably not put things on stderr in the final document.

davidanthoff:
> BTW, it looks like you forgot to add the new page to the table of contents (in both branches).

Thanks! #1128 and #1129.

MichaelHatherly:
> Any idea why this mangels up the display? If I build the release-0.8 branch doc locally, everything looks good.

Not sure, looks fine for me as well locally. I'd think it's probably some kind of transient failure, but it's pretty difficult to track down through the travis build logs. The next tagged release should probably fix it, or someone can build locally and push the resulting build to `gh-pages` manually to resolve it now instead.

Of more concern is how this folder (https://github.com/JuliaStats/DataFrames.jl/tree/gh-pages/builddocs) ended up on `gh-pages` &mdash; *really* weird. That should only really happen if someone manually adjusts the config in `docs/make.jl` though I've found nothing noticeable that could be the cause. It could possibly be due to `Pkg.clone`ing Documenter from my personal fork perhaps, that should be changed to `Pkg.add("Documenter")` as soon as possible in any case.

> warnings seem to be printed in the output. Aren't warnings printed to stderr? Documenter should probably not put things on stderr in the final document.

Yes, warnings are printed to `STDERR`, so are things like `info` output. This behaviour isn't going change since `@example`, `@repl`, and `jldoctest` are meant to mimic the normal Julia REPL as much as possible; capturing `STDERR` is part of that. To avoid getting those kind of deprecation and method overwriting warnings should just require doing the imports in the `@setup` block, or even in `docs/make.jl` if possible.

nalimilan:
> . It could possibly be due to Pkg.cloneing Documenter from my personal fork perhaps, that should be changed to Pkg.add("Documenter") as soon as possible in any case.

I've just changed this, let's see whether it improves things.

nalimilan:
Looks like it did the trick. There are still some surprising column alignment issues in the output, but that's less of an issue. Maybe the font isn't really fixed-width?

MichaelHatherly:
>  There are still some surprising column alignment issues in the output, but that's less of an issue. Maybe the font isn't really fixed-width?

The `DataFrame` output with the unicode borders? On Firefox it's all lined up correctly for me, but on Chrome the column alignment is indeed messed up.

MichaelHatherly:
Also, would be good to switch to the new SSH deploy keys (https://juliadocs.github.io/Documenter.jl/latest/man/hosting.html#SSH-Deploy-Keys-1) when someone with admin access as some time since the GitHub token based approach is still being used &mdash; it's being deprecated over the next few releases.

nalimilan:
> The DataFrame output with the unicode borders? On Firefox it's all lined up correctly for me, but on Chrome the column alignment is indeed messed up.

I'm on Firefox too (on Fedora 24). That's probably a font issue then, so not something we can fix.

I'll have a look at the SSH keys.

nalimilan:
Done to key change. Looks like it worked.

PR #1131, opened 2016-11-21
Update joins doc to include rename!

dx034:
The instruction to rename columns if they don't match was part of the docs previously (http://dataframesjl.readthedocs.io/en/latest/joins_and_indexing.html). I adapted the syntax to avoid using {}.

### DIFF
--- a/docs/src/man/joins.md
+++ b/docs/src/man/joins.md
@@ -53,15 +53,22 @@ join(a, b, kind = :cross)
 
 In order to join data frames on keys which have different names, you must first rename them so that they match. This can be done using rename!:
 
-
+```julia
+a = DataFrame(ID = [1, 2], Name = ["A", "B"])
+b = DataFrame(IDNew = [1, 2], Job = ["Doctor", "Lawyer"])
+rename!(b, :IDNew, :ID)
+join(a, b, on = :ID, kind = :inner)
 ```
-rename!(df3, :old_name, :new_name)
-```
-
 
 Or renaming multiple columns at a time:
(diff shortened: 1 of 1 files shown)


docs/src/man/joins.md
```diff
@@ -50,3 +50,18 @@ Cross joins are the only kind of join that does not use a key:
 ```julia
 join(a, b, kind = :cross)
 ```
+
+In order to join data frames on keys which have different names, you must first rename them so that they match. This can be done using rename!:
+
+
+```
```
nalimilan: Need `julia` after backquotes.

docs/src/man/joins.md
```diff
@@ -50,3 +50,18 @@ Cross joins are the only kind of join that does not use a key:
 ```julia
 join(a, b, kind = :cross)
 ```
+
+In order to join data frames on keys which have different names, you must first rename them so that they match. This can be done using rename!:
+
```
nalimilan: One empty line is enough.

docs/src/man/joins.md
```diff
@@ -50,3 +50,18 @@ Cross joins are the only kind of join that does not use a key:
...
+
+```
+rename!(df3, :old_name, :new_name)
```
nalimilan: Could you give a real example, i.e. create a data frame `c` with a different column name, rename the column and then join? Would also be nice to create two columns so that you can make the example below work too.

dx034:
Comitted a new version with working code snippets.

PR #1135, opened 2016-12-15
Added latin encoding support

amgad-naiem:


nalimilan:
Could you explain what this fixes? Please also add a test which failed previously but passes with this change. Thanks.

nalimilan:
Bump.

nalimilan:
The rebase didn't work. Since master was replaced with release-0.8, you need to rebase on the new master and ensure you don't include any commits from the old one.

amgad-naiem:
It is actually not UTF-8 problem. After further investigation, I found out that the files I have issue with are latin format. So I added latin1 support. I'm changing the pull request name to this.

Also I've added a test for it

nalimilan:
Sorry, but I don't think we want to support custom encodings like this. People should rather use StringEncodings to convert the file contents before passing them to `readtable`. Your solution using `transcode` won't work in general for any encoding.

amgad-naiem:
Okay.

nalimilan:
Thanks for making a PR though.

PR #1136, opened 2016-12-16
Fix Warning: the no-op `transpose` fallback is deprecated warning.

Nectarineimp:
replace line 599 of show.jl
header = map(x->strong(string(x)), names(df)[1:width]')

with
header = map(x->strong(string(x)), reshape(names(df), 1, width))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,6 @@ script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia --check-bounds=yes -e 'Pkg.clone(pwd()); Pkg.build("DataFrames"); Pkg.test("DataFrames"; coverage=true)'
 after_success:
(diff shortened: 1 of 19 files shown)


ararslan:
Hi @Nectarineimp! Did you mean to open this PR against master rather than release-0.8? It would be great if you could do master instead, that way we can cherry-pick the commit(s) to the release-0.8 branch as needed to make a patch release in the 0.8 series.

Nectarineimp:
This is my first ever pull request, and it didn't enable to button with the default master/master dropdowns. I'd love to do this against the master since its as simple fix and eliminates something being deprecated. I will see if I can modify the pull.

ararslan:
I don't think it's possible to modify which branch you're submitting a pull request to, so I think your best bet will be to close this one and open a new PR against master.

Nectarineimp:
Closing so I can resubmit the correct way.

PR #1137, opened 2016-12-16
Warning: the no-op 'transpose' fallback is deprecated

Nectarineimp:
fixes "Warning: the no-op `transpose` fallback is deprecated" warning.

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -589,7 +589,7 @@ end
 function _render(df::AbstractDataFrame)
     width = min(size(df, 2), SIZE)
     height = min(size(df, 1), SIZE)
-    header = map(x->strong(string(x)), names(df)[1:width]')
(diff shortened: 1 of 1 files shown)


Nectarineimp:
Hopefully this will work in spite of my inexperience with git's pull mechanism.

nalimilan:
Thanks, but looks like GitHub experienced a bug when computing the patch: it removes lots of lines to re-add them later. Any chance you could do that locally instead?

Nectarineimp:
I just realized it may be that the file base format (LF vs CRLF) may have changed in my editor. If you haven't already committed the change, I will redo the file locally. That will prevent the issue of every line being replaced from happening again.

ararslan:
The changes have not yet been committed; the pull request is still open. You could run something like `dos2unix` on the file and then push to the branch and see if that fixes it.

ararslan:
Closing this one since you opened a new PR for this: #1141

PR #1138, opened 2016-12-28
Avoid closing IO unless responsible for opening

omus:


ararslan:
Nightly failures are unrelated.

Issue #1029, opened 2016-08-08
v0.5 Julia removed fallback transpose and alternative does not work for DataFrames


tlnagy:
Since https://github.com/JuliaLang/julia/issues/13171 went through, doing something like the following no longer works:

```julia
julia> versioninfo()
Julia Version 0.5.0-rc0+0
Commit 0030eec* (2016-07-26 20:22 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin15.4.0)
  CPU: Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, haswell)

julia> Pkg.checkout("DataFrames")
INFO: Checking out DataFrames master...
INFO: Pulling DataFrames latest master...
INFO: No packages to install, update or remove

julia> using DataFrames
INFO: Recompiling stale cache file /Users/tamasnagy/.julia/lib/v0.5/DataFrames.ji for module DataFrames.

julia> a = DataFrame(rand(10,3))
10×3 DataFrames.DataFrame
│ Row │ x1        │ x2       │ x3        │
├─────┼───────────┼──────────┼───────────┤
│ 1   │ 0.104009  │ 0.288695 │ 0.0858412 │
│ 2   │ 0.161572  │ 0.535473 │ 0.7343    │
│ 3   │ 0.905977  │ 0.541049 │ 0.675175  │
│ 4   │ 0.459802  │ 0.19878  │ 0.049148  │
│ 5   │ 0.469835  │ 0.029731 │ 0.258061  │
│ 6   │ 0.0575359 │ 0.494867 │ 0.808683  │
│ 7   │ 0.327723  │ 0.225381 │ 0.14324   │
│ 8   │ 0.628949  │ 0.726375 │ 0.831952  │
│ 9   │ 0.883181  │ 0.808421 │ 0.803585  │
│ 10  │ 0.927315  │ 0.777683 │ 0.287885  │

julia> a'
WARNING: the no-op `transpose` fallback is deprecated, and no more specific `transpose` method for DataFrames.DataFrame exists. Consider `permutedims(x, [2, 1])` or writing a specific `transpose(x::DataFrames.DataFrame)` method if appropriate.
[...]

PR #1141, opened 2017-01-10
Update show.jl

Nectarineimp:
My last attempt at this caused a full file difference, when  only one line was actually changed. This time I did the file change in github. I think the last time what happened was the editor or the git process converted the file from LF to CRLF style, which would account for every line in both files appearing different to diff.

This fixes issue #1137

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -589,7 +589,7 @@ end
 function _render(df::AbstractDataFrame)
     width = min(size(df, 2), SIZE)
     height = min(size(df, 1), SIZE)
-    header = map(x->strong(string(x)), names(df)[1:width]')
+    header = map(x->strong(string(x)), reshape(names(df), 1, width))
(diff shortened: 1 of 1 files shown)


ararslan:
I don't the issue you reference is the right one... Did you mean #1029?

nalimilan:
Can you add a test for the previously buggy behavior?

Nectarineimp:
Sorry, my fault. #1137


Sent via the Samsung Galaxy Tab® 4, an AT&T 4G LTE tablet

-------- Original message --------

Nectarineimp:
The bug is deprecated usage. I'm not sure how to test for the warning. Any suggestions?


Sent via the Samsung Galaxy Tab® 4, an AT&T 4G LTE tablet

-------- Original message --------

nalimilan:
> The bug is deprecated usage. I'm not sure how to test for the warning. Any suggestions?

Ah, OK, so that's fine.

PR #1143, opened 2017-01-13
ENH: stack: allow naming of measure/value column names

kdmurray91:
In R's `reshape2::melt`  it is possible to use a custom name for the newly created stacked columns. This adds two optional keyword arguments that allow this behaviour (demonstrated below).

I have added tests, but I would like to see if I can simplify the diff. I'd like to avoid having to define these kwargs for all the various `stack` and `melt` invocations (suggestions welcome).

## TODO:

- [x] ~Decide on final name for keyword arguments: `variable_name` vs `measure_name`.~ Decision: `variable_name`

```julia
julia> d = DataFrame(red=rand(2), blue=rand(2), green=rand(2), sample=["A", "B"])
2×4 DataFrames.DataFrame
│ Row │ red      │ blue     │ green    │ sample │
├─────┼──────────┼──────────┼──────────┼────────┤
│ 1   │ 0.134251 │ 0.668584 │ 0.463817 │ A      │
│ 2   │ 0.784527 │ 0.46851  │ 0.634924 │ B      │

julia> ds = stack(d, [:red, :blue, :green])
6×3 DataFrames.DataFrame
│ Row │ variable │ value    │ sample │
├─────┼──────────┼──────────┼────────┤
│ 1   │ red      │ 0.134251 │ A      │
│ 2   │ red      │ 0.784527 │ B      │
│ 3   │ blue     │ 0.668584 │ A      │
│ 4   │ blue     │ 0.46851  │ B      │
│ 5   │ green    │ 0.463817 │ A      │
│ 6   │ green    │ 0.634924 │ B      │

julia> dsn = stack(d, [:red, :blue, :green], variable_name=:colour,
                   value_name=:mean_pixel_value)
6×3 DataFrames.DataFrame
│ Row │ colour │ mean_pixel_value │ sample │
├─────┼────────┼──────────────────┼────────┤
│ 1   │ red    │ 0.134251         │ A      │
│ 2   │ red    │ 0.784527         │ B      │
│ 3   │ blue   │ 0.668584         │ A      │
│ 4   │ blue   │ 0.46851          │ B      │
│ 5   │ green  │ 0.463817         │ A      │
│ 6   │ green  │ 0.634924         │ B      │
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,3 @@ test/data/bigrams.tsv
 docs/build/
 docs/site/
 .DS_Store
-# Vim/vi
-.*.sw?
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ test/data/bigrams.tsv
 docs/build/
 docs/site/
 .DS_Store
+# Vim/vi
+.*.sw?
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -18,16 +18,10 @@ Stacks a DataFrame; convert from a wide to long format
 
 
 ```julia
-stack(df::AbstractDataFrame, measure_vars, id_vars;
-      measure_name=:variable, value_name=:value)
-stack(df::AbstractDataFrame, measure_vars;
-      measure_name=:variable, value_name=:value)
-stack(df::AbstractDataFrame;
-      measure_name=:variable, value_name=:value)
-melt(df::AbstractDataFrame, id_vars, measure_vars;
-     measure_name=:variable, value_name=:value)
-melt(df::AbstractDataFrame, id_vars;
-     measure_name=:variable, value_name=:value)
+stack(df::AbstractDataFrame, [measure_vars], [id_vars];
+      measure_name::Symbol=:variable, value_name::Symbol=:value)
+melt(df::AbstractDataFrame, [id_vars], [measure_vars];
+     measure_name::Symbol=:variable, value_name::Symbol=:value)
 ```
 
 ### Arguments
@@ -81,7 +75,8 @@ d1s_name = melt(d1, [:a, :b, :e], measure_name=:somemeasure)
 
 """
 function stack(df::AbstractDataFrame, measure_vars::Vector{Int},
-               id_vars::Vector{Int}; measure_name=:variable, value_name=:value)
+               id_vars::Vector{Int}; measure_name::Symbol=:variable,
+               value_name::Symbol=:value)
     N = length(measure_vars)
     cnames = names(df)[id_vars]
     insert!(cnames, 1, value_name)
@@ -92,22 +87,22 @@ function stack(df::AbstractDataFrame, measure_vars::Vector{Int},
               cnames)
 end
 function stack(df::AbstractDataFrame, measure_var::Int, id_var::Int;
-               measure_name=:variable, value_name=:value)
+               measure_name::Symbol=:variable, value_name::Symbol=:value)
     stack(df, [measure_var], [id_var];
           measure_name=measure_name, value_name=value_name)
 end
 function stack(df::AbstractDataFrame, measure_vars::Vector{Int}, id_var::Int;
-               measure_name=:variable, value_name=:value)
+               measure_name::Symbol=:variable, value_name::Symbol=:value)
     stack(df, measure_vars, [id_var];
           measure_name=measure_name, value_name=value_name)
 end
 function stack(df::AbstractDataFrame, measure_var::Int, id_vars::Vector{Int};
-               measure_name=:variable, value_name=:value)
+               measure_name::Symbol=:variable, value_name::Symbol=:value)
     stackdf(df, [measure_var], id_vars;
             measure_name=measure_name, value_name=value_name)
 end
 function stack(df::AbstractDataFrame, measure_vars, id_vars;
-               measure_name=:variable, value_name=:value)
+               measure_name::Symbol=:variable, value_name::Symbol=:value)
     stack(df, index(df)[measure_vars], index(df)[id_vars];
           measure_name=measure_name, value_name=value_name)
 end
@@ -117,7 +112,7 @@ numeric_vars(df::AbstractDataFrame) =
      for T in eltypes(df)]
 
 function stack(df::AbstractDataFrame, measure_vars = numeric_vars(df);
-               measure_name=:variable, value_name=:value)
+               measure_name::Symbol=:variable, value_name::Symbol=:value)
     mv_inds = index(df)[measure_vars]
     stack(df, mv_inds, _setdiff(1:ncol(df), mv_inds);
           measure_name=measure_name, value_name=value_name)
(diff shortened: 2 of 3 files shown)


src/abstractdataframe/reshape.jl
```diff
@@ -38,9 +38,14 @@ melt(df::AbstractDataFrame, id_vars)
   stacking, a normal column indexing type; for `stack` defaults to all
   variables that are not `measure_vars`
 
+
```
nalimilan: Unrelated.

src/abstractdataframe/reshape.jl
```diff
@@ -68,43 +73,49 @@ d1m = melt(d1, [:a, :b, :e])
 ```
 
 """
-function stack(df::AbstractDataFrame, measure_vars::Vector{Int}, id_vars::Vector{Int})
+function stack(df::AbstractDataFrame, measure_vars::Vector{Int}, id_vars::Vector{Int}; measure_name=:variable, value_name=:value)
```
nalimilan: Please break lines so as not to exceed 92 chars.

src/abstractdataframe/reshape.jl
```diff
@@ -18,11 +18,11 @@ Stacks a DataFrame; convert from a wide to long format
 
 
 ```julia
-stack(df::AbstractDataFrame, measure_vars, id_vars)
-stack(df::AbstractDataFrame, measure_vars)
-stack(df::AbstractDataFrame)
-melt(df::AbstractDataFrame, id_vars, measure_vars)
-melt(df::AbstractDataFrame, id_vars)
+stack(df::AbstractDataFrame, measure_vars, id_vars; measure_name=:variable, value_name=:value)
```
nalimilan: `melt` from data.table calls the first argument `variable_name`. I wonder which one is best. (AFAICT only `value_name` exists in reshape2.)

src/abstractdataframe/reshape.jl
```diff
@@ -38,9 +38,14 @@ melt(df::AbstractDataFrame, id_vars)
   stacking, a normal column indexing type; for `stack` defaults to all
   variables that are not `measure_vars`
 
+
 If neither `measure_vars` or `id_vars` are given, `measure_vars`
 defaults to all floating point columns.
 
+The names of the resulting "stacked" columns can be controlled with keyword
```
nalimilan: Should use a bullet for each argument. You can move the paragraph above to the `measure_vars` bullet so that it's not out of place.

src/abstractdataframe/reshape.jl
```diff
@@ -68,43 +73,49 @@ d1m = melt(d1, [:a, :b, :e])
...
-    stack(df, mv_inds, _setdiff(1:ncol(df), mv_inds))
+    stack(df, mv_inds, _setdiff(1:ncol(df), mv_inds);
+          measure_name=measure_name, value_name=value_name)
 end
 
 """
 Stacks a DataFrame; convert from a wide to long format; see
 `stack`.
 """
 melt(df::AbstractDataFrame, id_vars::@compat(Union{Int,Symbol})) = melt(df, [id_vars])
```
nalimilan: Should add the keyword arguments here too (and below). Also please add tests for them (which are apparently missing else you'd have noticed their lack).

src/abstractdataframe/reshape.jl
kdmurray91: I chose `measure_name` as we use `measure_vars` through this code.  I'm open to change this though.

nalimilan: Yeah, but positional argument names can be changed without affecting existing code, which isn't the case for keyword arguments. So now is the time to make a decision. What do others think about it?

kdmurray91:
@nalimilan 1ff78cb should fix most of your concerns.

src/abstractdataframe/reshape.jl
```diff
@@ -65,49 +76,74 @@ d1 = DataFrame(a = repeat([1:3;], inner = [4]),
...
+          measure_name=measure_name, value_name=value_name)
 end
 
 """
 Stacks a DataFrame; convert from a wide to long format; see
 `stack`.
 """
-melt(df::AbstractDataFrame, id_vars::@compat(Union{Int,Symbol})) = melt(df, [id_vars])
-function melt(df::AbstractDataFrame, id_vars)
+function melt(df::AbstractDataFrame, id_vars::@compat(Union{Int,Symbol});
```
kdmurray91: I felt the concise function definition form was less readable in many cases once the new keyword arguments were added. I'm happy to revert this if others disagree.

src/abstractdataframe/reshape.jl
```diff
@@ -420,40 +456,58 @@ d1m = meltdf(d1, [:a, :b, :e])
...
-    stackdf(df, [measure_var], [id_var])
+function stackdf(df::AbstractDataFrame, measure_var::Int, id_var::Int;
+                 measure_name::Symbol=:variable, value_name::Symbol=:value)
+    stackdf(df, [measure_var], [id_var]; measure_name=measure_name,
+            value_name=value_name)
 end
-function stackdf(df::AbstractDataFrame, measure_vars, id_var::Int)
-    stackdf(df, measure_vars, [id_var])
+function stackdf(df::AbstractDataFrame, measure_vars, id_var::Int;
+                 measure_name::Symbol=:variable, value_name::Symbol=:value)
```
kdmurray91: Do `measure_name` and `value_name` need this type annotation? I seem to have inconsistently used it. If there are no type annotations, would that affect type stability, or would an error be caused [here](https://github.com/JuliaStats/DataFrames.jl/pull/1143/files#diff-15c11ad91975e38974d256d28f21db51R465) in any case?

nalimilan: The annotation is only useful to ensure an error is printed immediately rather than later when the code attempts doing something which only works for symbols. In general it's better to strictly type arguments when it makes sense for clarity, and it makes the documentation clearer.

src/abstractdataframe/reshape.jl
```diff
@@ -18,11 +18,16 @@ Stacks a DataFrame; convert from a wide to long format
 
 
 ```julia
-stack(df::AbstractDataFrame, measure_vars, id_vars)
-stack(df::AbstractDataFrame, measure_vars)
-stack(df::AbstractDataFrame)
-melt(df::AbstractDataFrame, id_vars, measure_vars)
-melt(df::AbstractDataFrame, id_vars)
+stack(df::AbstractDataFrame, measure_vars, id_vars;
```
nalimilan: Please write `[measure_vars], [id_vars]` and get rid of the second definition. Same for `melt`. This is getting too verbose for my taste.

src/abstractdataframe/reshape.jl
nalimilan: Since @andreasnoack's is the only one to have commented and he supports `variable_name`, could you change the (positional and keyword) arguments? Then it should be good to merge. Thanks.

src/abstractdataframe/reshape.jl
kdmurray91: i.e. condense the definition to a single one with optional measure/id vars?

src/abstractdataframe/reshape.jl
kdmurray91: OK, I shall add these annotations to all changed function declarations.

src/abstractdataframe/reshape.jl
kdmurray91: Whoops, github hid this comment. Fixed.

src/abstractdataframe/reshape.jl
kdmurray91: @nalimilan I believe this was fixed, could you double check that your requirement was met?

kdmurray91:
@nalimilan Think it's good to go now.

.gitignore
```diff
@@ -5,3 +5,5 @@ test/data/bigrams.tsv
 docs/build/
 docs/site/
 .DS_Store
+# Vim/vi
```
nalimilan: Unrelated, please open another PR if you want to have this merged.

nalimilan:
Thanks, looks good except for the unrelated addition to `.gitignore`.

nalimilan:
> EDIT: Is documenation some place other the docstrings required/wanted?

AFAIK no other changes are needed (only when adding new functions).

.gitignore
kdmurray91: Ah crap, that was accidental, due to `git add .`. Reveted

nalimilan:
Thanks!

src/abstractdataframe/reshape.jl
```diff
@@ -65,49 +70,75 @@ d1 = DataFrame(a = repeat([1:3;], inner = [4]),
...
+
+function stack(df::AbstractDataFrame, measure_vars = numeric_vars(df);
+               variable_name::Symbol=:variable, value_name::Symbol=:value)
     mv_inds = index(df)[measure_vars]
-    stack(df, mv_inds, _setdiff(1:ncol(df), mv_inds))
+    stack(df, mv_inds, _setdiff(1:ncol(df), mv_inds);
+          variable_name=variable_name, value_name=value_name)
 end
 
 """
```
nalimilan: If you feel like it, it would be nice to remove this docstring in favor of the one from above, splitting `stack` and `melt`. Another option would be to use exactly the same docstring for both, I think some doc tricks allow doing this.

PR #1147, opened 2017-01-15
Fix handling of string columns in model matrix

nalimilan:
All non-real columns are now considered as categorical, since conversion
to a float column will likely fail. Types which should be converted to float
will have to define is_categorical(::T) = false. Before this, Vector{String}
and NullableVector{String} columns triggered an error.

Replace ContrastsMatrix(c::ContrastsMatrix, x::CategoricalArray) with
ContrastsMatrix(c::ContrastsMatrix, levels::AbstractVector), offering
equivalent functionality: it's more consistent with the ContrastsMatrix()
constructor, and more general since only the levels are actually needed.

Fixes https://github.com/JuliaStats/DataFrames.jl/issues/1085.

----------

The `_levels` custom function should probably be included in CategoricalArrays. I'd like to hear your opinions about it, though, especially since it treats `Nullable` in a special fashion (which is consistent with how it works with `NullableCategoricalArray`). On the one hand, that's really useful in many circumstances; on the other, it breaks the `AbstractArray` interface by not always following its element type. The element type of `CategoricalArray` is `CategoricalValue{T}` anyway, so maybe that's not a concern. That's one of the fundamental design questions with categorical arrays.

Of course this will also have to be applied to StatsModels.


src/statsmodels/formula.jl
```diff
@@ -283,6 +287,20 @@ end
 
 const DEFAULT_CONTRASTS = DummyCoding
 
+_levels(x::Union{CategoricalArray, NullableCategoricalArray}) = levels(x)
+
+function _levels{T<:Nullable}(x::AbstractArray{T})
+    levs = [get(l) for l in unique(x) if !isnull(l)]
+    try; sort!(levs); end
```
ararslan: Why is the `try` necessary here?

nalimilan: Because not all types support `isless`. And `method_exists` is actually not enough since it doesn't work with `Array{Any}`, and `isless` might even fail when it's defined (e.g. with unordered `CategoricalValue`).

ararslan: Ah, okay, that makes sense. Thanks for the explanation.

src/statsmodels/formula.jl
```diff
@@ -283,6 +287,20 @@ end
 
 const DEFAULT_CONTRASTS = DummyCoding
 
+_levels(x::Union{CategoricalArray, NullableCategoricalArray}) = levels(x)
```
nalimilan: Yes. One issue is that `unique(::NullableCategoricalArray)` has to return `Nullable` elements to be consistent with the generic `AbstractArray` method. But of course we can unwrap these in our internal method.

nalimilan:
Closing in favor of https://github.com/JuliaStats/StatsModels.jl/pull/13.

PR #1152, opened 2017-01-25
Fix join() on key mixing Array and NullableArray

nalimilan:
Previously failed when trying to index Dict{T} with Nullable{T} key,
since isequal(::T, Nullable{T}) returns false even if convert() works.

Fixes https://github.com/JuliaStats/DataFrames.jl/issues/1151.

(Failure on 0.6 is still due to CategoricalArrays.)


nalimilan:
OK to merge?

ararslan:
What's the stack overflow about on nightly, any idea? I'm inclined to assume it isn't related but it'd be good to figure out why it's happening.

nalimilan:
Closing in favor of https://github.com/JuliaData/DataTables.jl/pull/4.

PR #1155, opened 2017-01-29
Add VI/VIM swap files to gitignore

kdmurray91:
Accidentally added (and reverted) during #1143.

Cheers,
Kevin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@ test/data/bigrams.tsv
 docs/build/
 docs/site/
 .DS_Store
+
+# vi/vim temp files
+.*.sw?


kmsquire:
I'm not very active in this repo right now, so I might be overridden.  That said, my general understanding is that, by convention, .gitignore inside a repo is usually used for files produced as part of a standard build or run of the repo (e.g., compiled .o files, python's .pyc files, etc.). 

If editor files were included there, one would be adding files from lots of different editors, and the file would get quite unwieldy. 

The general recommendation I've heard is that developers add these types of files to their own global gitignore file.

kdmurray91:
@kmsquire Consider this PR a question regarding this repo's policy :smile:

ararslan:
I fully agree with @kmsquire. These types of files should be in a user's global gitignore, not in every possible repository.

ararslan:
The .DS_Store actually shouldn't be in here either but I don't think that's really worth changing at this point.

kdmurray91:
An understandable position. I'll close this PR then :)

ararslan:
Your effort for the contribution is appreciated nonetheless!

PR #1160, opened 2017-02-12
support for confidence intervals in predict

mkborregaard:
There is now support in GLM.jl for generating confidence intervals from the `predict` function for linear models. I added a method here to allow the user to supply newX as a DataFrame.

Some comments:
1. I didn't write any tests, as it seems like all essential functionality is already tested for the existing predict function. If there is something that needs to be tested I wouldn't mind writing one.
2. There has to be two predict methods to ensure type stability. I copied a lot of the code, as extracting the common code into a function called by both methods seemed like it would create less clear resulting code.

### DIFF
--- a/src/statsmodels/statsmodel.jl
+++ b/src/statsmodels/statsmodel.jl
@@ -86,7 +86,7 @@ function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame; 
     return(out)
 end
 
-# Predict function with confidence intervars
+# Predict function with confidence intervals
 function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame, interval_type::Symbol, level::Real = 0.95; kwargs...)
     newTerms = dropresponse!(mm.mf.terms)
     mf = ModelFrame(newTerms, df; contrasts = mm.mf.contrasts)
--- a/src/statsmodels/statsmodel.jl
+++ b/src/statsmodels/statsmodel.jl
@@ -86,6 +86,16 @@ function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame; 
     return(out)
 end
 
+# Predict function with confidence intervars
+function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame, interval_type::Symbol, level::Real = 0.95; kwargs...)
+    newTerms = dropresponse!(mm.mf.terms)
+    mf = ModelFrame(newTerms, df; contrasts = mm.mf.contrasts)
+    newX = ModelMatrix(mf).m
+    yp = predict(mm, newX, interval_type, level; kwargs...) # this and the below two are different from the non-confint method above
+    out = NullableArray(eltype(yp), size(df, 1), 3)
+    out[mf.msng, :] = yp
+    return(out)
+end
 
 
 # coeftable implementation


mkborregaard:
Requested here https://github.com/JuliaStats/GLM.jl/issues/174#issuecomment-278958185

src/statsmodels/statsmodel.jl
```diff
@@ -86,6 +86,16 @@ function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame;
     return(out)
 end
 
+# Predict function with confidence intervars
```
nalimilan: "intervals"

src/statsmodels/statsmodel.jl
```diff
@@ -86,6 +86,16 @@ function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame;
...
+function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame, interval_type::Symbol, level::Real = 0.95; kwargs...)
```
nalimilan: You could merge the two methods by having `interval_type::Union{Symbol, Void}=nothing`. Then you would check  `interval_type == nothing` inside the body (ideally the method in `StatsBase` would also accept `nothing` so that you can even pass it directly).

nalimilan:
Even if the `Matrix` method is tested in StatsBase, it would be good to add some tests to `test/statsmodel.jl`. Should be fairly easy since similar tests are already present.

src/statsmodels/statsmodel.jl
mkborregaard: I am not sure - because the method with interval_type returns a Matrix, the version without returns a Vector. Having separate methods is necessary for type stability, is it not?

mkborregaard:
I tried writing some tests based on the tests already in there, but was not sure what to test for. I couldn't think of something to test that wasn't already covered?

src/statsmodels/statsmodel.jl
mkborregaard: ah.

src/statsmodels/statsmodel.jl
```diff
@@ -86,6 +86,16 @@ function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame;
...
+    newTerms = dropresponse!(mm.mf.terms)
+    mf = ModelFrame(newTerms, df; contrasts = mm.mf.contrasts)
+    newX = ModelMatrix(mf).m
+    yp = predict(mm, newX, interval_type, level; kwargs...) # this and the below two are different from the non-confint method above
+    out = NullableArray(eltype(yp), size(df, 1), 3)
```
mkborregaard: I see that the `predict` method without confidence intervals uses`DataArray` not `NullableArray`. Which is right?

ararslan: All vestiges of NullableArrays should be purged from the repo at some point, since the version of DataFrames that uses NullableArrays instead of DataArrays has been separated into https://github.com/JuliaData/DataTables.jl. So whatever uses DataArrays here is right.

nalimilan: Yeah, the situation is really messy ATM. We will have to apply these changes to DataTables and StatsModels too, or maybe just to the latter if we are able to move DataTables to it soon...

nalimilan:
> I tried writing some tests based on the tests already in there, but was not sure what to test for. I couldn't think of something to test that wasn't already covered?

Well, testing the new method for example? :-p

mkborregaard:
Ha ha. Sure but it cannot call GLM, right? So I can test a dummy method similar to the current predict test, I guess. I don't mind writing the test (so I'll do it when I get a moment), am just a little unsure it will test something not already covered.

src/statsmodels/statsmodel.jl
mkborregaard: Does that mean these methods will be left behind when DataFrames move to use Nullables, or that I should make a PR to DataTables (/StatModels) too?

mkborregaard: Does that mean these methods will be left behind when DataFrames move to use Nullables, or that I should make a PR to DataTables (/StatModels) too?

andreasnoack:
I think it would be okay to add GLM as a test dependency for this.

src/statsmodels/statsmodel.jl
nalimilan: I think we'll have to make PRs everywhere unfortunately. But the code hasn't changed much, so it shouldn't be too hard.

mkborregaard:
GLM has DataFrames as a test dependency and tests both pure-GLM and GLM-DataFrames functionality. Would you suggest duplicating tests across repos, or keeping the GLM-DataFrames tests in GLM and identifying something that is specific to DataFrames alone for here?

nalimilan:
I think for this PR it's OK to just add a test using `DummyMod` like the existing ones here. Then we can improve tests in general later (in StatsModels).

nalimilan:
The rebasing didn't work correctly. The unrelated commits shouldn't appear in this PR.

mkborregaard:
Sorry. I was faffing around with a new git client and did something I shouldn't (didn't Force Push after the Rebase, to be exact). I'll kill the branch and open up a new PR with the same changes?

nalimilan:
No, just `git rebase -i master` and removing the unrelated commits should be enough (followed by another force push).

mkborregaard:
Thanks, sorry for the noise.

nalimilan:
No problem. Can you add simple tests based on `DummyMod` for the new method?

mkborregaard:
Absolutely, I will. I am completely strung up until April 5th, though - after that I will finish up the PRs.

mkborregaard:
And open the same PR on DataTables if you like?

ararslan:
All of the modeling stuff has been removed from DataTables (or at least it should be if it hasn't already been removed)

mkborregaard:
Cool. OK let me finish this and then let me know if I need to take any action wrt StatsModels (if that's where it is gone now).

ararslan:
Bump. Have you had the opportunity to revisit this, @mkborregaard?

mkborregaard:
Thanks for the bump - I promise to have it fixed within a week :-) (am on a family holiday the next 5 days but may be able to steal time)

ararslan:
No rush, I was just wondering. Enjoy your holiday!

BenjaminBorn:
I stumbled across this pull request when trying to run the OLS example in the GLM.jl readme. 

Would it make sense to adjust the readme until this PR is merged?

mkborregaard:
It would. But - I should really just do that test and get it merged already. I'll do it after the ERC deadline on Tuesday.

mkborregaard:
I am aware that 6 months ago I also promised to do this within a week... :-O

BenjaminBorn:
Hehe, thanks for the quick reply, then I'll just wait. An ERC grant is more important than this for sure.

nalimilan:
Modeling code doesn't live in DataFrames anymore.

PR #1161, opened 2017-02-14
Test fixes

nalimilan:
Backport/sideport of https://github.com/JuliaData/DataTables.jl/pull/2.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -8,6 +8,7 @@ anyerrors = false
 
 using Base.Test
 using DataFrames
+using Compat
 
 my_tests = ["utils.jl",
             "cat.jl",
@@ -50,7 +51,9 @@ if anyerrors
     throw("Tests failed")
 end
 
-stdin = joinpath(dirname(@__FILE__), "stdin.sh")
-ENV2 = copy(ENV)
-ENV2["JULIA_HOME"] = JULIA_HOME
-run(setenv(`bash $stdin`, ENV2))
+if !is_windows()
+    stdin = joinpath(dirname(@__FILE__), "stdin.sh")
+    ENV2 = copy(ENV)
+    ENV2["JULIA_HOME"] = JULIA_HOME
+    run(setenv(`sh $stdin`, ENV2))
+end
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,7 +34,7 @@ install:
 build_script:
 # Need to convert from shallow to complete for Pkg.clone to work
   - IF EXIST .git\shallow (git fetch --unshallow)
-  - C:\projects\julia\bin\julia -F -e "versioninfo();
+  - C:\projects\julia\bin\julia -e "versioninfo();
       Pkg.clone(pwd(), \"DataFrames\"); Pkg.build(\"DataFrames\")"
 
 test_script:
--- a/test/stdin.sh
+++ b/test/stdin.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 TESTDIR=$(dirname $0)
 cat $TESTDIR/data/utf8/utf8.csv | $JULIA_HOME/julia $TESTDIR/stdin.jl


test/runtests.jl
```diff
@@ -50,7 +50,9 @@ if anyerrors
     throw("Tests failed")
 end
 
-stdin = joinpath(dirname(@__FILE__), "stdin.sh")
-ENV2 = copy(ENV)
-ENV2["JULIA_HOME"] = JULIA_HOME
-run(setenv(`bash $stdin`, ENV2))
+if !is_windows()
```
ararslan: This will need a `@compat` (or `using Compat: is_windows`) if we want to keep supporting 0.4.

ararslan: Actually, we won't be able to support 0.4 (or even 0.5) soon, since the DataArrays that works on Julia 0.6 will be 0.6 only.

andreasnoack: `DataFrames` can just use older versions of `DataArrays` on 0.4 and 0.5.

ararslan: Right, we just need to apply upper bounds in METADATA. I mean going forward on master here.

nalimilan: Good catch. I've changed it to use Compat.

andreasnoack: > I mean going forward on master here.

I'm don't think I understand. As long as we don't use any new features of `DataArrays` there is not reason to require the latest version of `DataArrays` here, right?

> Right, we just need to apply upper bounds in METADATA.

...or just tell people to `Pkg.update()`.

ararslan: To support 0.6 here, we need the (forthcoming) new version of DataArrays that supports 0.6 as the minimum DataArrays version requirement for the package. Since that's 0.6-only, this will similarly have to be made 0.6-only. Right?

Older versions of the package should be upper bounded at Julia 0.6 as a precaution, since the current state of affairs is known not to work on 0.6 due to DataArrays issues. That's what we had talked about with Tony in triage.

andreasnoack: > Since that's 0.6-only, this will similarly have to be made 0.6-only. Right?

I don't think it is necessary to enforce this in packages that depend on `DataArrays`. As long as we don't use new features added to `DataArrays` it is up to `DataArrays` to figure out which version can be installed. Which problem would be avoided by bumping the version requirement for `DataArrays`?

> Older versions of the package should be upper bounded at Julia 0.6 as a precaution, since the current state of affairs is known not to work on 0.6 due to DataArrays issues. That's what we had talked about with Tony in triage.

I know that Tony wants this but I'm not convinced that it matters much in practice. Probably my view will change when things are more mature but right now I can see two scenarios. If the upper bounds are added at the same time as a new version of the package is made available for, say 0.6, then when running `Pkg.update()`, you'll receive the upper bounds in METADATA **and** the updated package. In that situation, not having the bounds would have given the same result. The other scenario is a situation where the package doesn't have a working version for 0.6. Then adding the upper bounds trades a broken package for a complaining package manager. Most people I've discussed this with would prefer the broken package because of `Pkg`.

nalimilan: Shouldn't we add upper bounds to DataArrays only? Then everything should work fine.

tkelman: Not having the bounds does not give the same result, if you've applied any other bounds for the sake of reproducibility. Products depend on this.

tkelman: There's also the parsing of `~` which has changed in 0.6 relative to what every past version of this package expects.

ararslan: > Which problem would be avoided by bumping the version requirement for DataArrays?

Yes I think I'm wrong about this. Sorry!

andreasnoack: > Not having the bounds does not give the same result, if you've applied any other bounds for the sake of reproducibility. Products depend on this.

It sounds like a different scenario. Could you give a concrete example where it would matter?

> There's also the parsing of `~` which has changed in 0.6 relative to what every past version of this package expects.

I've wondered about this one as well. My current thinking is that we should just remove all the formula code here. It is now in `StatsModels`. If we do that, it might be possible not to bump the version requirement since `~` wouldn't be used in the package anymore.

ararslan: The problem is that StatsModels is based on NullableArrays and CategoricalArrays rather than DataArrays, which makes it incompatible with this package.

ararslan: Perhaps StatsBase could provide an abstraction for arrays with missing data and/or categorical values, then StatsModels could be written against that, assuming NullableArrays and DataArrays can be made to adhere to whatever common interface would be required for such an abstraction. Or we can just copy some of the code from StatsModels over...

andreasnoack: Good point. I haven't looked at the formula code in `StatsModels`. Couldn't it be written for `AbstractArray` and avoid specific Nullable features or do you think that would be impossible/impractical?

ararslan: IIRC the model matrix code takes advantage of the categorical-ness of `CategoricalArray`s (in StatsModels) and `PooledDataArray`s (in the formula code here), so I don't think StatsModels could go fully against `AbstractArray`.

andreasnoack: Yes. That makes sense. Then we have a problem.

tkelman: The issue with bounds is that all past versions assume `~` is parsed as a macro. If every single one of those versions is therefore broken on Julia 0.6, then we should upper bound those versions. If you try to apply a lower upper bound on DataFrames with Julia 0.6 than whatever first version gets released based on StatsModels, then you aren't going to be able to get a working copy of DataFrames. If old versions of DataFrames are able to load on 0.6 while using 0.6-compatible versions of DataArrays, just with broken formula functionality, then maybe that would be okay. Will have to test.

nalimilan: Yet we could have an abstract interface for `iscategorical`, `isnull` and `dropnull`. We've discussed this in another issue recently. I guess it could go into AbstractTables.

PR #1163, opened 2017-02-15
Fix stack() example in manual

nalimilan:
This example has not been updated after [1:4] stopped returning a Vector{Int}
in Julia 0.5.

Fixes https://github.com/JuliaStats/DataFrames.jl/issues/1162.


nalimilan:
This fix is actually already included in DataTables via a2ae0ca1.

PR #1164, opened 2017-02-15
RFC: Preparation for Julia 0.6

andreasnoack:
The plan is to remove that formula and modeling functionality and hopefully `StatsModels` will be able to handle that part. For now, I've commented out the formula and modeling related tests. To make this easier, I've loosened the requirement on the `Vector` input to the `DataFrame` constructors such that the element type doesn't have to be `Any`. With that change, the `ModelFrame` method in `StatsModels` works.

There are also various routine syntax updates that should be uncontroversial.

Finally, I plan to remove the dependency on Juno. It was controversial when added and has caused problems since added. E.g., right now `Juno` breaks the REPL and this is not captured by `Juno`'s tests. Getting nice printing of `DataFrames` in `Juno` must be handled elsewhere.


nalimilan:
Too bad I just made the same changes in DataTables (https://github.com/JuliaData/DataTables.jl/pull/9, https://github.com/JuliaData/DataTables.jl/pull/8; FWIW the latter has a few more fixes). Maintaining the two repos is parallel leads to some waste of time...

One issue is whether StatsModels is going to be able to support both DataFrames and DataTables at the same time. We likely need the three of them to use a common AbstractTable interface for that. Until then, I'd rather keep the modeling functions in DataFrames.

Is the Juno dep a problem on 0.6?

src/subdataframe/subdataframe.jl
```diff
@@ -51,7 +51,7 @@ immutable SubDataFrame{T <: AbstractVector{Int}} <: AbstractDataFrame
     parent::DataFrame
     rows::T # maps from subdf row indexes to parent row indexes
 
-    function SubDataFrame(parent::DataFrame, rows::T)
+    @compat function SubDataFrame{T}(parent::DataFrame, rows::T) where T
```
nalimilan: Is `T` really needed here?

nalimilan: Nevermind, I just found the notes in https://github.com/JuliaLang/julia/pull/20308/files.

nalimilan: Yet, shouldn't `T<:AbstractVector{Int}`? Not sure what would happen without it, probably an error a bit later. In his PR Jeff added these (duplicated) constraints.

andreasnoack: Not sure. I just followed the deprecation warning. Shouldn't harm to make that restriction though.

andreasnoack:
> Too bad I just made the same changes in DataTables (JuliaData/DataTables.jl#9, JuliaData/DataTables.jl#8; FWIW the latter has a few more fixes). Maintaining the two repos is parallel leads to some waste of time...

I agree. That is why I've been asking all the questions about `DataArrays`/`NullableArrays` lately. Ideally, we didn't maintain two almost identical repos in parallel but I find `DataArrays` so much more convenient to work with that I'm willing to contribute time for now. I've tried to find all the best arguments for `NullableArrays` to convince myself that moving over is a good idea but so far I'm not convinced that it is an improvement relative to `DataArrays`.

> Until then, I'd rather keep the modeling functions in DataFrames.

I concluded that it wasn't really feasible because `~` is not a macro anymore. Fixing the formula code here would therefore require reimplementing the formula code from `StatsModels`. I tried to use `StatsModels` with `DataFrames` and it mostly worked. Many of the methods can dispatch on `DataFrame`/`DataTable` and do the right thing. However, I need your PRs merged into `DataTables` before I can test this further.

> Is the Juno dep a problem on 0.6?

Yes. If you want to use the REPL 😄 (see my last comment). Juno is not sufficiently tested to capture these bugs so it is up to us to detect that `using Juno` breaks the REPL. But even if that was fixed I think it is the wrong dependency direction. If Juno doesn't want to depend on this package then we'd be better off with a `JunoDataFrames`. When we have conditional modules, that package could be integrated into Juno.

nalimilan:
> I tried to use StatsModels with DataFrames and it mostly worked. Many of the methods can dispatch on DataFrame/DataTable and do the right thing. However, I need your PRs merged into DataTables before I can test this further.

That's cool, but we still need to define the functions in a separate interface package so that StatsModels can use them without DataFrames and DataTables conflicting. AbstractTables is the most logical candidate, but I'm not sure how stable is its current API.

I'm not really in favor of keeping the Juno code, I just wanted to know whether it was one of the reasons for the test failures on 0.6.

test/join.jl
```diff
@@ -18,9 +18,9 @@ module TestJoin
                       Job = @data(["Lawyer", "Doctor", "Florist", NA, "Farmer"]))
 
     # (Tests use current column ordering but don't promote it)
-    right = outer[!isna(outer[:Job]), [:Name, :ID, :Job]]
-    left = outer[!isna(outer[:Name]), :]
-    inner = left[!isna(left[:Job]), :]
+    right = outer[.!isna(outer[:Job]), [:Name, :ID, :Job]]
+    left = outer[.!isna(outer[:Name]), :]
```
nalimilan: Will that work on 0.5? I would try with `(!).(...)`.

andreasnoack: No it won't. `(!).(...)` is indeed necessary. Just realized that while fixing another package.

andreasnoack:
Updated. I've dropped 0.4 support, deleted the statsmodels and formula code but kept the contrasts code. However, the contrasts tests rely on `ModelFrame`s for now, so I've disabled the tests. Depending on the StatsModels solution the contrasts could be dropped from this package or we could change the tests to avoid `ModelFrames`.

src/subdataframe/subdataframe.jl
```diff
@@ -51,14 +51,14 @@ immutable SubDataFrame{T <: AbstractVector{Int}} <: AbstractDataFrame
     parent::DataFrame
     rows::T # maps from subdf row indexes to parent row indexes
 
-    @compat function SubDataFrame{T}(parent::DataFrame, rows::T) where T
+    function (::Type{SubDataFrame{T}}){T}(parent::DataFrame, rows::T)
```
nalimilan: Is this the new recommended syntax for inner constructors? I'm still confused about this, and the manual doesn't help.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -490,19 +490,19 @@ complete_cases!(df::AbstractDataFrame)
...
 df = DataFrame(i = 1:10, x = rand(10), y = rand(["a", "b", "c"], 10))
 df[[1,4,5], :x] = NA
 df[[9,10], :y] = NA
-complete_cases!(df)
+completecases!(df)
 ```
 
 """
-complete_cases!(df::AbstractDataFrame) = deleterows!(df, find(!complete_cases(df)))
+completecases!(df::AbstractDataFrame) = deleterows!(df, find(!completecases(df)))
```
nalimilan: Better do `find(!, completecases(df))` as in DataTables to avoid allocating a copy.

src/abstractdataframe/io.jl
```diff
@@ -181,7 +181,7 @@ end
     write(io, "</tr>")
     write(io, "</thead>")
     write(io, "<tbody>")
-    tty_rows, tty_cols = _displaysize(io)
+    tty_rows, tty_cols = Base.displaysize(io)
```
nalimilan: No need for `Base.` here and below.

src/statsmodels/contrasts.jl
```diff
@@ -292,7 +292,6 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
     mat = mat[[baseind; 1:(baseind-1); (baseind+1):end], :]
     return mat
 end
-
```
nalimilan: ?

src/abstractdataframe/show.jl
```diff
@@ -580,27 +580,27 @@ end
...
-    height = min(size(df, 1), SIZE)
-    header = map(x->strong(string(x)), names(df)[1:width]')
-    body = Juno.undefs(to_matrix(df))[1:height, 1:width]
-    view = Table(vcat(header, body))
-    LazyTree(Row(typeof(df), text" ", Juno.dims(size(df)...)),
-             () -> [view])
-end
-
-@render Inline df::AbstractDataFrame _render(df)
+# using Juno
```
nalimilan: Just remove this altogether, that's easy to find in the git history if needed.

nalimilan:
I remove all of the code related to modeling, or keep all of it. I don't like commenting out code.

Other than that, looks good to me.

src/subdataframe/subdataframe.jl
andreasnoack: I'm sure it is recommended but it works without warnings across 0.5 and 0.6 so I think it the right solution until we drop 0.5 support.

test/runtests.jl
```diff
@@ -16,17 +16,16 @@ my_tests = ["utils.jl",
...
             "sort.jl",
             "grouping.jl",
             "join.jl",
             "iteration.jl",
             "duplicates.jl",
-            "show.jl",
-            "statsmodel.jl",
-            "contrasts.jl"]
+            "show.jl"#,
+            # "contrasts.jl"
```
nalimilan: Remove this line and I would say it's good to go.

nalimilan: (And remove the corresponding file...)

ararslan:
Until StatsModels works flawlessly with DataFrames, I think we should just copy the definition of `@formula` from StatsModels here. None of the other modeling code in this package would need to change, just the definition of formulas.

andreasnoack:
The right solution might then be to make an update for 0.5 before these fixes where we deprecate the use of `~` and clearly write how to transition, i.e. use StatsModels. However, it requires that we make StatsModels work generically first. Hopefully soon.

ararslan:
That sounds good to me. We should start an issue to track everything that's needed to get StatsModels working generically, or at least across DataFrames and DataTables.

src/deprecated.jl
```diff
@@ -9,12 +9,10 @@ import Base: keys, values, insert!
...
-@deprecate array(df::AbstractDataFrame) convert(Array, df)
-@deprecate array(r::DataFrameRow) convert(Array, r)
-if VERSION < v"0.4.0-"
-    @deprecate DataArray(df::AbstractDataFrame) convert(DataArray, df)
-end
 @deprecate DataArray(df::AbstractDataFrame, T::DataType) convert(DataArray{T}, df)
 
 @deprecate read_rda(args...) FileIO.load(args...)
+
+@deprecate complete_cases(df)  complete_cases(df)
```
cjprybol:
```julia
@deprecate complete_cases(df)  completecases(df)
```

src/deprecated.jl
```diff
@@ -9,12 +9,10 @@ import Base: keys, values, insert!
...
+@deprecate complete_cases!(df) complete_cases!(df)
```
cjprybol:
```julia
@deprecate complete_cases!(df) completecases!(df)
```

cjprybol: given https://github.com/JuliaData/DataTables.jl/pull/6#issue-207690988 and linked issues could this be changed to deprecate `complete_cases!(df)` to `dropna!(df)`?

would need `dropna` too I guess https://github.com/JuliaData/DataTables.jl/pull/6#issuecomment-280668091

nalimilan:
> That sounds good to me. We should start an issue to track everything that's needed to get StatsModels working generically, or at least across DataFrames and DataTables.

https://github.com/JuliaStats/StatsModels.jl/issues/14

ararslan:
Heh, whoops. Missed that. Thanks!

davidanthoff:
Bump, what is the status here? Would be great if we had a tagged version of DataFrames that works on julia 0.6. I'm trying to move ExcelReaders (and then Query) to 0.6, but my CI builds won't pass until there is a tagged version of DataFrames that works on julia 0.6.

ararslan:
DataArrays master works on 0.6 so once that's tagged, I think that version of DataArrays + this PR = a complete working DataFrames on 0.6.

nalimilan:
Though model formula support still isn't there AFAIK. I guess we could still tag a beta version without it to fix packages which don't need formulas (most of them).

andreasnoack:
I've been traveling and busy but I hope to have time work a little on this again next week.

davidanthoff:
I think it would be great if there was a version tagged with a REQUIRE line ``julia 0.6-`` really soon, even if that version still has some known issues. Right now a simple ``using DataFrames`` breaks on julia 0.6, and that pretty much means that a huge number of packages can't even start to work on 0.6 compatibility without going through a lot of hoops.

ararslan:
I've dealt with the formulas situation temporarily in #1170 by copying over the requisite definitions from StatsModels. That way we have something here that will work in the interim while StatsModels is made to work with DataFrames.

src/statsmodels/contrasts.jl
ararslan: Removes trailing whitespace. Note that the diff shows that this line had the whitespace removed in favor of an empty line below.

PR #1166, opened 2017-02-19
Precised that by can use multiple columns to split

sylvaticus:
While in aggregate function it is explicitly stated column (or columns) that was not the  case with the description of the `by` function.

### DIFF
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -24,7 +24,7 @@ by(iris, :Species) do df
 end
 ```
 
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks! I would rather phrase it as "one or more columns", though.

nalimilan:
Bump.

nalimilan:
Thanks! Could you apply the same fix to DataTables?

ararslan:
Who, me or @sylvaticus?

nalimilan:
Either! :-)

PR #1170, opened 2017-03-10
Replace bare tilde with formula macro

ararslan:
This copies over the definition of `@formula` from StatsModels and deprecates `@~` for Julia versions where `~` parses as an infix macro rather than a regular call.


cjprybol:
Looks good! Failures due to `sub -> view` deprecation? https://travis-ci.org/JuliaStats/DataFrames.jl/jobs/209917401#L1184

ararslan:
Failures due to everything fixed by #1164.

PR #1173, opened 2017-03-14
Suppress deprecation warning in v0.6

dmbates:
julia v0.6 deprecates `!retrms` in favor of `.!retrms` but that syntax is not available in v0.5 or through the `Compat` package


src/statsmodels/formula.jl
```diff
@@ -419,7 +419,7 @@ function droprandomeffects(trms::Terms)
     else
         # the rows of `trms.factors` correspond to `eterms`, the columns to `terms`
         # After dropping random-effects terms we drop any eterms whose rows are all false
-        ckeep = !retrms                 # columns to retain
+        ckeep = map(!, retrms)                 # columns to retain
```
ararslan: Nice, I like this better than the alternative `(!).(retrms)`, which also works but looks bizarre.

PR #1175, opened 2017-03-17
Add spreadsheet-like pivot() function

sylvaticus:
Addded spreadsheet-like pivot function to pivot and optionally filter and sort in a single function.

pivot(df::AbstractDataFrame, rowFields, colField, valuesField; <keyword arguments>)

See: https://discourse.julialang.org/t/a-spreadsheet-like-pivot-function/2701

(ps: this pull request doesn't include the export of the function.. where is that defined ?)

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -308,7 +308,7 @@ function pivot(df::AbstractDataFrame, rowfields, colfield::Symbol, valuesfield::
     catfields::AbstractVector{Symbol} = cat(1,rowfields, colfield)
     dfs  = DataFrame[]
     opsv =[]
-    if(isa(ops, Array))
+    if isa(ops, Array)
         opsv = ops
     else
         push!(opsv,ops)
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -224,6 +224,110 @@ end
 
 unstack(df::AbstractDataFrame) = unstack(df, :id, :variable, :value)
 
+##############################################################################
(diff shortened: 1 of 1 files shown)


nalimilan:
I'm not sure this kind of function should really live in this package. We're rather trying to make it smaller, concentrated on core features. Ideally that function would be very generic and could support any data source, including e.g. SQL tables.

At some point I wanted to extend the scope of my FreqTables package to any kind of pivot table, but I never found the time to do it. However, I would have returned the result as a `NamedArray` or a similar structure, not as a data frame. Do we really need a function for that, given that query frameworks make this quite easy? For example, Query makes this kind of thing quite straightforward: http://www.david-anthoff.com/Query.jl/stable/querycommands.html#Grouping-1 A future dplyr-like framework would make this even less verbose.

OTOH, there's previous art in for this kind of function in Pandas: http://pandas.pydata.org/pandas-docs/stable/reshaping.html#pivot-tables

ararslan:
We have stack and join operations in here, perhaps a pivot wouldn't be so out of place, at least for now.

ararslan:
@sylvaticus It would be great if you could take care of the following points:
* Add tests
* Use lowercase instead of camel case (e.g. `rowField` -> `rowfield`)
* Use 4 space indents everywhere (you mostly have)
* No parentheses around `if` conditions
* Model the docstring formatting after others in this repo (i.e. signature first and indented, etc.)

sylvaticus:
Sorry for the mess, I couldn't update this pull request, so I opened a new one (#1181) implementing the required changes.

PR #1177, opened 2017-03-26
Fix handling of non-Int row indices

joshbode:
Non-`Int64` row-indices were not being correctly converted to `Int64`.

- explicit conversion to `Vector{Int64}`
- test-cases added

Interestingly, this came up only when DataFrames was precompiled into a userimg (see JuliaLang/julia#21164) and only after the rename of `sub()` to `view()`

### DIFF
--- a/src/subdataframe/subdataframe.jl
+++ b/src/subdataframe/subdataframe.jl
@@ -62,16 +62,13 @@ immutable SubDataFrame{T <: AbstractVector{Int}} <: AbstractDataFrame
     end
 end
 
-function SubDataFrame{T <: AbstractVector{Int}}(parent::DataFrame, rows::T)
-    return SubDataFrame{T}(parent, rows)
+function SubDataFrame{T <: Integer}(parent::DataFrame, rows::AbstractVector{T})
+    rows = convert(Vector{Int}, rows)
+    return SubDataFrame{typeof(rows)}(parent, rows)
 end
 
 function SubDataFrame(parent::DataFrame, row::Integer)
     return SubDataFrame(parent, [row])
-end
-
-function SubDataFrame{S <: Integer}(parent::DataFrame, rows::AbstractVector{S})
-    return view(parent, Int(rows))
 end
 
 
--- a/test/data.jl
+++ b/test/data.jl
@@ -66,6 +66,10 @@ module TestData
     @test size(sdf6a) == (1,3)
     sdf6d = view(df6, [1,3], :B)
     @test size(sdf6d) == (2,1)
+    sdf6e = view(df6, 0x01)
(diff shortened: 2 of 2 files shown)


src/subdataframe/subdataframe.jl
```diff
@@ -71,7 +71,7 @@ function SubDataFrame(parent::DataFrame, row::Integer)
 end
 
 function SubDataFrame{S <: Integer}(parent::DataFrame, rows::AbstractVector{S})
-    return view(parent, Int(rows))
+    return SubDataFrame{Vector{Int}}(parent, convert(Vector{Int}, rows))
```
nalimilan: We should only convert to `AbstractVector{Int}` since that's what the type accepts. I suspect the `Int(rows)` expression comes from an automated replacement which was done after `int(x)` was deprecated. This was effectively equivalent to `convert(AbstractVector{Int}, rows)`.

Also it seems it would more sense to remove this method and put this line into the generic constructor with `rows::T` above.

nalimilan:
Thanks. Note that `Int` is not necessarily `Int64`, it's `Int32` on 32-bit.

joshbode:
Good points - I'll try to merge this into the earlier constructor and come back

src/subdataframe/subdataframe.jl
```diff
@@ -62,18 +62,14 @@ immutable SubDataFrame{T <: AbstractVector{Int}} <: AbstractDataFrame
     end
 end
 
-function SubDataFrame{T <: AbstractVector{Int}}(parent::DataFrame, rows::T)
-    return SubDataFrame{T}(parent, rows)
+function SubDataFrame{S <: Integer}(parent::DataFrame, rows::AbstractVector{S})
```
nalimilan: Minor point, but using `T` is more common.

src/subdataframe/subdataframe.jl
```diff
@@ -62,18 +62,14 @@ immutable SubDataFrame{T <: AbstractVector{Int}} <: AbstractDataFrame
...
+    return SubDataFrame{AbstractVector{Int}}(parent, convert(AbstractVector{Int}, rows))
```
nalimilan: `{AbstractVector{Int}}` isn't correct here: it should be the actual concrete type to which `rows` was converted. So you need to store the result of the conversion in an object first, and call `typeof` on it.

joshbode: OK - thanks, that makes sense

ararslan:
Thanks!

iamed2:
This also happened with `--compilecache=no` (see #1182)

joshbode:
All my favourite Julia bugs seem to tie back to precompilation :)

PR #1179, opened 2017-03-30
Fix typo in example

fabianlischka:
ID in example output was duplicated. Make ID distinct from row number, to avoid confusion. Align examples.


nalimilan:
@fabianlischka Do you feel like making the same change in DataTables?

PR #1185, opened 2017-05-04
Remove describe code moved to DataArrays and StatsBase

cjprybol:
Follows https://github.com/JuliaStats/StatsBase.jl/pull/247 and https://github.com/JuliaStats/DataArrays.jl/pull/248, addresses https://github.com/JuliaStats/DataFrames.jl/issues/1184

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6-
-DataArrays 0.3.4
+DataArrays 0.5
 StatsBase 0.11.0
 GZip
 SortingAlgorithms
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -404,35 +404,6 @@ function StatsBase.describe(io, df::AbstractDataFrame)
         println(io, )
     end
 end
-StatsBase.describe(dv::AbstractArray) = describe(STDOUT, dv)
-function StatsBase.describe{T<:Number}(io, dv::AbstractArray{T})
-    if all(isna, dv)
-        println(io, " * All NA * ")
-        return
-    end
-    filtered = float(dropna(dv))
-    qs = quantile(filtered, [0, .25, .5, .75, 1])
-    statNames = ["Min", "1st Qu.", "Median", "Mean", "3rd Qu.", "Max"]
-    statVals = [qs[1:3]; mean(filtered); qs[4:5]]
-    for i = 1:6
-        println(io, string(rpad(statNames[i], 8, " "), " ", string(statVals[i])))
-    end
-    nas = count(isna, dv)
-    println(io, "NAs      $nas")
-    println(io, "NA%      $(round(nas*100/length(dv), 2))%")
-    return
-end
-function StatsBase.describe{T}(io, dv::AbstractArray{T})
-    ispooled = isa(dv, PooledDataVector) ? "Pooled " : ""
-    nmiss = count(isna, dv)
-    # if nothing else, just give the length and element type and NA count
-    println(io, "Length  $(length(dv))")
-    println(io, "Type    $(ispooled)$(string(eltype(dv)))")
-    println(io, "NAs     $nmiss")
-    println(io, "NA%     $(round(nmiss*100/length(dv), 2))%")
-    println(io, "Unique  $(length(unique(dv)))")
-    return
-end
 
 ##############################################################################
 ##
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -58,4 +58,52 @@ module TestUtils
     else
         @test DataFrames._fnames(funs) == ["mean", "sum", "var", string(funs[end])]
     end
+
+    @testset "describe" begin
+        io = IOBuffer()
+        df = DataFrame(Any[collect(1:4), DataArray(Int, 4),
+                           PooledDataArray(collect(3:6)),
+                           PooledDataArray(Int, 4)],
+                       [:array, :naarray, :pooledarray, :napooledarray])
+        describe(io, df)
+        @test String(take!(io)) ==
+            """
+            array
+            Summary Stats:
+            Mean:           2.500000
+            Minimum:        1.000000
+            1st Quartile:   1.750000
+            Median:         2.500000
+            3rd Quartile:   3.250000
+            Maximum:        4.000000
+            Length:         4
+            Type:           $Int
+
+            naarray
+            Summary Stats:
+            Type:           $Int
+            Number Missing: 4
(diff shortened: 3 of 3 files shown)


cjprybol:
Had to adjust some `isna(Vector)` calls which no longer work in commit 1 before removing the describe code that now lives in StatsBase and DataArrays in commit 2. Everything looks good in 0.6 but as DataArrays is 0.6 only, none of the 0.5 tests will run

REQUIRE
```diff
@@ -1,5 +1,5 @@
 julia 0.5
-DataArrays 0.3.4
+DataArrays 0.5
```
ararslan: If you revert this part do the tests pass on 0.5 and 0.6? The code you've changed here is necessary for DataArrays on 0.6 but should also work on 0.5 (the older DataArrays code just isn't optimized for that style), so we might be able to get away with letting dependency resolution do its thing in this case.

cjprybol: If we remove the `describe` code here then we need DataArrays 0.5, otherwise `describe` won't be able to handle NAs.

What if instead of removing the `describe` code currently in DataFrames outright, use an `include_string` macro to only use it when the user isn't on Julia v0.6+ and DataArrays v0.5+? e.g.
```julia
if Pkg.installed("DataArrays") >= v"0.5"
    @include_string("""
                    ... current describe code for DataArrays in DataFrames ...
                   )
else
    do nothing, since DataArrays >= v"0.5" has the new describe code
end
```

I can add version specific tests against the current style of describe in DataFrames and the udpated style in DataArrays v0.5+, and I think everything else can remain untouched.

cjprybol: And with https://github.com/JuliaStats/DataArrays.jl/pull/252 we don't need to change `isna(::DataArray)` to `isna.(::DataArray)` yet as that will now throw a depwarn, right?

ararslan: The problem is that DataArrays only supports Julia 0.6, but DataFrames still supports 0.5. That's why the 0.5 CI builds are failing.

cjprybol:
Ok, I think this should work. If the user has the version of DataArrays installed that contains the `describe` functions (v0.5+), then DataFrames will use the code in DataArrays. Otherwise it will load the functions which I copied and pasted in an `include_string` block.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -404,33 +404,35 @@ function StatsBase.describe(io, df::AbstractDataFrame)
...
-    ispooled = isa(dv, PooledDataVector) ? "Pooled " : ""
-    # if nothing else, just give the length and element type and NA count
-    println(io, "Length  $(length(dv))")
-    println(io, "Type    $(ispooled)$(string(eltype(dv)))")
-    println(io, "NAs     $(sum(isna(dv)))")
-    println(io, "NA%     $(round(sum(isna(dv))*100/length(dv), 2))%")
-    println(io, "Unique  $(length(unique(dv)))")
-    return
+# drop when DataFrames is julia v0.6 only (can support DataArrays >= v0.5)
+if Pkg.installed("DataArrays") < v"0.5"
```
ararslan: `Pkg` functions shouldn't be used within package code, as it can have issues with precompilation (e.g. when a package becomes available after the current one has already been precompiled). Let's ask @tkelman, master of all dependency-related things, what the best course of action is.

nalimilan: Maybe just `!method_exists(StatsBase.describe, Tuple{AbstractDataVector})`?

cjprybol: `method_exists` is returning `true`, I think because the function is going by `AbstractDataVector <: AbstractVector`.

I could do a manual comparison against the signatures?
```julia
julia> [m.sig for m in methods(StatsBase.describe)]
5-element Array{DataType,1}:
 Tuple{StatsBase.#describe,AbstractArray}
 Tuple{StatsBase.#describe,IO,AbstractArray{T<:Real,N}}
 Tuple{StatsBase.#describe,IO,AbstractArray}
 Tuple{StatsBase.#describe,DataFrames.AbstractDataFrame}
 Tuple{StatsBase.#describe,Any,DataFrames.AbstractDataFrame}
```

on Julia 0.5 w/ DataArrays 0.3.12
```julia
julia> [string(m.sig) == "Tuple{StatsBase.#describe,IO,DataArrays.AbstractDataArray{T,1} where T}" for m in methods(StatsBase.describe)]
5-element Array{Bool,1}:
 false
 false
 false
 false
 false
```

Julia 0.6 w/ DataArrays 0.5.1
```julia
julia> [string(m.sig) == "Tuple{StatsBase.#describe,IO,DataArrays.AbstractDataArray{T,1} where T}" for m in methods(StatsBase.describe)]
8-element Array{Bool,1}:
 false
 false
 false
  true
 false
 false
 false
 false
```

and then we could use
```julia
!any([string(m.sig) == "Tuple{StatsBase.#describe,IO,DataArrays.AbstractDataArray{T,1} where T}" for m in methods(StatsBase.describe)])
```

in place of
```julia
!method_exists(StatsBase.describe, Tuple{AbstractDataVector})
```

ararslan: Any idea whether method tables are saved as part of the precompiled cache? I would think so, in which case I don't think checking against the methods explicitly will be any less problematic than using `Pkg`, but I could be wrong.

nalimilan: Hmm, I guess that could work but it's a bit fragile. Isn't there another method which only exists in the new version?

tkelman: using Pkg wouldn't work correctly if the package is loaded from somewhere else on LOAD_PATH, like on juliabox

nalimilan: Looks like the best solution is something like this:
```julia
x = methods(describe, Tuple{DataArray})
x.ms[1].sig.parameters[2] <: DataArray
```

ararslan: That's kind of hilariously hacky. I kind of like it but if we're going to go that far into the weeds, we might as well just drop 0.5 support.

nalimilan: We may as well do that if we need it for https://github.com/JuliaStats/DataFrames.jl/pull/1186.

cjprybol:
Ok I've rebased this onto the changes introduced by https://github.com/JuliaStats/DataFrames.jl/pull/1186 and tests look good!

cjprybol:
Thanks!

PR #1186, opened 2017-05-12
Fix isna, allna and anyna deprecations

nalimilan:
Also fix some uses of all(isna(x)) and any(isna(x)), and avoid computing
the number of missing values twice in describe().

@ararslan I'm not sure whether I should bump the dependency on DataArrays. AFAICT this form worked before the deprecation was added, since it's possible since Julia 0.5, right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 
 language: julia
 julia:
-  - 0.5
+  - 0.6
   - nightly
 os:
   - linux
--- a/README.md
+++ b/README.md
@@ -1,9 +1,8 @@
 DataFrames.jl
 =============
 
-[![0.4](http://pkg.julialang.org/badges/DataFrames_0.4.svg)](http://pkg.julialang.org/?pkg=DataFrames)
-[![0.5](http://pkg.julialang.org/badges/DataFrames_0.5.svg)](http://pkg.julialang.org/?pkg=DataFrames)
 [![0.6](http://pkg.julialang.org/badges/DataFrames_0.6.svg)](http://pkg.julialang.org/?pkg=DataFrames)
+[![0.7](http://pkg.julialang.org/badges/DataFrames_0.7.svg)](http://pkg.julialang.org/?pkg=DataFrames)
 
 [![Coverage Status](https://coveralls.io/repos/JuliaStats/DataFrames.jl/badge.svg?branch=master&service=github)](https://coveralls.io/github/JuliaStats/DataFrames.jl?branch=master)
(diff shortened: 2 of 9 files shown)


nalimilan:
Mmm, unfortunately this doesn't work with 0.3.12 (on Julia 0.5):
```julia
julia> isna.(DataArray(1:2))
2-element DataArrays.DataArray{Int64,1}:
 0
 0
```

So we'll either have to make a new 0.3.x release for DataArrays (but then we cannot mark this as a dependency for DataFrames), or drop support for Julia 0.5.

src/abstractdataframe/reshape.jl
```diff
@@ -212,7 +212,7 @@ function unstack(df::AbstractDataFrame, colkey::Int, value::Int)
         j = Int(keycol.refs[k])
         i = rowkey[k]
         if i > 0 && j > 0
-            if nowarning && !isna(df2[j][i])
+            if nowarning && !isna.(df2[j][i])
```
ararslan: `df2[j][i]` should be a scalar, right? Otherwise unary `!` would throw a depwarn on 0.6 (not to mention it wouldn't work in this `Bool`-only context).

ararslan: Same above

ararslan:
> So we'll either have to make a new 0.3.x release for DataArrays

I tried add 0.5 support back to DataArrays but it was too difficult and I decided I didn't care enough. Since there's a release candidate now for 0.6 it's probably fine to drop 0.5 here.

src/abstractdataframe/reshape.jl
nalimilan: Good catch. I probably shouldn't have replaced all occurrences and reverted those who triggered test failures. These lines do not appear to be covered by tests, I think Cameron added some in DataTables recently.

nalimilan:
Let's drop 0.5 support then. I've changed the PR to do that.

src/dataframe/dataframe.jl
```diff
@@ -168,7 +168,7 @@ function DataFrame{D <: Associative}(ds::Vector{D}, ks::Vector)
     for d in ds
         for (i,k) in enumerate(ks)
             # TODO: check for user-defined "NA" values, ala pandas
-            if haskey(d, k) && !isna(d[k])
+            if haskey(d, k) && !isna.(d[k])
```
ararslan: The `broadcast` here still shouldn't be needed, right?

nalimilan:
Let's merge this?

PR #1189, opened 2017-05-22
Restore an orphan commit from the v0.4.3 tag

tkelman:
Doing some scraping of metadata, found that the v0.4.3 tag of DataFrames is the below orphaned commit, it's not in the history of any branches and wasn't pushed as a git tag. I don't actually want anyone to merge this, but just so this commit can be checked out without getting a "fatal: reference is not a tree" error, it would be good if someone with commit access could do
```
git remote add tkelman https://github.com/tkelman/DataFrames.jl
git fetch tkelman
git checkout 84523937447b336dfcbf1747b123e9d1da1308c6
git tag v0.4.3
git push origin v0.4.3
```

That is all, if someone does that, we can go ahead and close this. Thanks!

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
-julia 0.2-
+julia 0.2- 0.3-
 DataArrays
-Stats
+StatsBase 0.3.8-
 GZip
 Blocks
 SortingAlgorithms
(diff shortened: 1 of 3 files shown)


ararslan:
Done. Can you confirm that things are in the correct state before closing?

tkelman:
Yep. The commit now has the tag label, and after a fetch it can be checked out again. Thanks!

PR #1194, opened 2017-06-14
Handling multiple text encodings

mbeltagy:
This as a expansion of the attempt at #1135 to cover any kind of encoding supported by [StringEncodings](https://github.com/nalimilan/StringEncodings.jl). Now the encoding keyword is no longer a symbol, it is a string.


src/DataFrames.jl
```diff
@@ -84,6 +85,7 @@ export @~,
        unique!,
        unstack,
        writetable,
+       encodings,
```
nalimilan: I don't think it's worth exporting this function. Just use `StringEncodings.encodings` in the error message. Also better not add `using StringEncodings` just for two calls: use the fully qualified names in these places.

src/dataframe/io.jl
```diff
@@ -814,11 +814,11 @@ function readtable(io::IO,
                    skipstart::Integer = 0,
                    skiprows::AbstractVector{Int} = Int[],
                    skipblanks::Bool = true,
-                   encoding::Symbol = :utf8,
+                   encoding::String = "UTF-8",
                    allowescapes::Bool = false,
                    normalizenames::Bool = true)
-    if encoding != :utf8
-        throw(ArgumentError("Argument 'encoding' only supports ':utf8' currently."))
+    if !any(encodings().==encoding)
```
nalimilan: Just `!(encoding in encodings())`. But maybe we don't need to check this at all, and leave `StringDecoder` throw the error (if it's explicit enough)?

mbeltagy:
I have made all the changes you requested. However, I am not so sure about using   `::Union{AbstractString, Symbol}` for `encoding`.  Since `:latin1` encoding never worked before, it is unlikely that anyone would have used it, or even felt the need to explicitly specify `:utf8`.

nalimilan:
Makes sense, I had forgotten it didn't work at all before. Feel free to drop support for `Symbol` then.

mbeltagy:
Removed `Symbol` support.

src/dataframe/io.jl
```diff
@@ -848,7 +846,13 @@ function readtable(io::IO,
                      allowescapes, normalizenames)
 
     # Use the IO stream method for readtable()
-    df = readtable!(p, io, nrows, o)
+    if encoding == "UTF-8"
+      df = readtable!(p, io, nrows, o)
```
nalimilan: Should use four-space indent.

src/dataframe/io.jl
```diff
@@ -848,7 +846,13 @@ function readtable(io::IO,
...
+    else # we wrap the IO stream in a decoder
+      d_io = StringEncodings.StringDecoder(io, encoding, "UTF-8")
```
nalimilan: `"UTF-8"` isn't needed here (even if it doesn't really hurt).

src/dataframe/io.jl
mbeltagy: Will fix

src/dataframe/io.jl
mbeltagy: Fixed.

test/io.jl
```diff
@@ -324,6 +324,12 @@ module TestIO
     io = IOBuffer(abnormal*",%_B*\tC*,end\n1,2,3\n")
     @test names(readtable(io, normalizenames=false)) == [Symbol(abnormal),Symbol("%_B*\tC*"),:end]
 
+    # Testing reading files with none utf8 encoding
```
nalimilan: "non-UTF8"

mbeltagy: Fixed.

mbeltagy:
I removed the strict dependency StringEncodings. So now it will use it only if already installed. Otherwise it will tell the user that non-utf8 decoding is not possible without the said package.

nalimilan:
> I removed the strict dependency StringEncodings. So now it will use it only if already installed. Otherwise it will tell the user that non-utf8 decoding is not possible without the said package.

Unfortunately, I don't think that's a recommended approach. Even if the user installs StringEncodings, DataFrames won't be recompiled, the precompiled image will have to be removed too.

mbeltagy:
In error message we could suggest that that the user will need to rebuild
after installing StringEncodings.

On Jun 17, 2017 4:35 PM, "Milan Bouchet-Valat" <notifications@github.com>
wrote:

I removed the strict dependency StringEncodings. So now it will use it only
if already installed. Otherwise it will tell the user that non-utf8
decoding is not possible without the said package.

Unfortunately, I don't think that's a recommended approach. Even if the
user installs StringEncodings, DataFrames won't be recompiled, the
precompiled image will have to be removed too.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/JuliaStats/DataFrames.jl/pull/1194#issuecomment-309218779>,
or mute the thread

mbeltagy:
Now `StringEncodings` will be loaded if installed, without the need to do compilation as that part now is inside the `__init__` function.
 `DataFrames` will on need to built compiled once.  
I can not think of proper way to validate  that in the unit tests. But I have tried adding and removing  `StringEncodings`  while running small test on DataFrames, and it works just fine.

src/dataframe/io.jl
```diff
@@ -861,6 +862,16 @@ function readtable(io::IO,
     return df
 end
 
+# Import  "StringEncodings" if available and set up decoder
+ENCODING_PACKAGE_READY = false
+try Pkg.installed("StringEncodings")
```
ararslan: This isn't a good way to implement optional dependencies, since it causes issues with precompilation.

mbeltagy: I agree.  I fixed that in my latest commit.

src/DataFrames.jl
```diff
@@ -128,4 +128,18 @@ for (dir, filename) in [
     include(joinpath(dir, filename))
 end
 
+# Import  "StringEncodings" if available and set up decoder
+_decoder=nothing
+ENCODING_PACKAGE_READY = false
+function __init__()
+    try Pkg.installed("StringEncodings")
```
tkelman: this is not a good idea, it will interact badly with precompilation and not work with packages installed on LOAD_PATH

ararslan: If we need a dependency on StringEncodings at all it should be a hard dependency. There isn't really a legitimate way to implement optional dependencies right now, though hopefully there will be first-class support for that in the language at some point.

tkelman: https://github.com/JuliaLang/julia/pull/21743

mbeltagy: Can you kindly explain why will it interact badly with pre-compilation? the `__init__` function block will not be precompiled, everything is else will. At least that is what  I think is happening. 
Also `StringEncdoings` will not be on the `LOAD_PATH`.

nalimilan: @ararslan How do you feel about adding a dependency on StringEncodings? It shouldn't be an issue since the package is very small, and it only relies on iconv which is already installed on all major platforms. But I'm biased so I'll let you decide.

yuyichao: For a start, using `Pkg.installed` here is wrong. I'm not sure what do you mean by `StringEncodings` will not be on `LOAD_PATH`.

Also, doing this kind of thing in `__init__` is basically strictly worse than doing it at compile time. If no "optional" dependencies are added or removed, the two methods have the same effect with `__init__` being slower. If optional dependencies are changed, doing it at compile time will require a manual recompilation only when an optional dependency is added, while doing it in `__init__` will break every packages depening on this one, since you suddenly starts to include dependencies that the precompile system did not see before.

tkelman: having to manually manage the precompile cache for correct behavior defeats the purpose of marking a package as opting in to precompilation though.

yuyichao: The point is that what you end up doing by putting it in `__init__` is automatically break all other package, not automatically manages the precompile cache.

mbeltagy: Thank you @yuyichao to trying to clarify. I meant that the `StringEncodings` will be inside `~/.julia/`, not in the  default `LOAD_PATH`  array, in my system it is something like

```julia
julia> LOAD_PATH
2-element Array{Any,1}:
 "/home/someusr/temp/juliaExpr/julia0.6/usr/local/share/julia/site/v0.6"
 "/home/someusr/temp/juliaExpr/julia0.6/usr/share/julia/site/v0.6"      
```
I might be missing something here as to the meaning of `LOAD_PATH` . 

> If optional dependencies are changed, doing it at compile time will require a manual recompilation only when an optional dependency is added, 

I agree. 
>while doing it in __init__ will break every packages depening on this one, since you suddenly starts to include dependencies that the precompile system did not see before.

Just to make sure, you are saying the packages the require `DataFrames` will broken. right? Are you saying a precompiled package with dependencies will not be calling the __init__ of its dependent packages when loading.  

Or, more concretely, if a package `X` depends on  `DataFrames` that was initially installed without `StringEncodings`. After installing `StringEncodings`, package `X` will still not be able to use feature in `DataFrames` that `StringEncodings` enable. Right?

yuyichao: > I might be missing something here as to the meaning of `LOAD_PATH`.

You probably don't have any packages in `LOAD_PATH`, it doesn't mean others don't.

> Just to make sure, you are saying the packages the require DataFrames will broken. right?

Yes.

> Are you saying a precompiled package with dependencies will not be calling the init of its dependent packages when loading.

It will, and that's the problem, since that's a unpredictable action during precompilation.

> Or, more concretely, if a package X depends on DataFrames that was initially installed without StringEncodings. After installing StringEncodings, package X will still not be able to use feature in DataFrames that StringEncodings enable. Right?

The precise symptom will not be very simple. It can cause this for sure in some cases.

tkelman: If DataFrames were to get precompiled including optional functionality but then that optional package were later removed, the behavior prior to 0.5.1 would be unable to load the precompiled copy. At least that bug should be fixed now on 0.5.2 and 0.6.

mbeltagy: Thanks @tkelman I tested on ` 0.6.0-rc3.0` and it worked fine. 

I will be reverting all the optional loading functionality hacks. Hopeful we can fix that when there is first class support for it at some point.

mbeltagy:
All optional dependency  commits are now reverted. It should be ready for merge if community is happy with `StringEncodings` hard dependency.

ararslan:
> @ararslan How do you feel about adding a dependency on StringEncodings?

@nalimilan It's not ideal but as long as we have this kind of I/O functionality in this package then it makes sense to me. Plus I happen to know that the author of that package makes great software and is attentive with maintenance. 😉

tkelman:
Adding the iconv binary dependency on windows is a little inconvenient, but as long as things are working okay on appveyor...

nalimilan:
StringEncodings uses win_iconv on Windows, which is a very small wrapper around the native Windows charset APIs. GNU iconv isn't used there.

src/dataframe/io.jl
```diff
@@ -848,7 +846,13 @@ function readtable(io::IO,
                      allowescapes, normalizenames)
 
     # Use the IO stream method for readtable()
-    df = readtable!(p, io, nrows, o)
+    if encoding == "UTF-8"
+        df = readtable!(p, io, nrows, o)
+    else # we wrap the IO stream in a decoder
+        d_io = StringEncodings.StringDecoder(io, encoding)
+        df = readtable!(p, d_io, nrows, o)
```
nalimilan: It would be good to use `try.. finally` to be sure the handle is closed even in case of error.

mbeltagy:
@nalimilan Done! 
It makes me think that it might also be a good idea to wrap the original `io` handle in a `try..finally`.

tkelman:
win_iconv still needs to be downloaded, it's a binary dependency where previously none was needed

mbeltagy:
If we want pure Julia, we might then opt to resurrect https://github.com/JuliaStats/DataFrames.jl/pull/1135. It will cover only `latin1` however.

quinnj:
This would be better contributed to CSV or TextParse.

nalimilan:
Sorry @mbeltagy, as I had noted above `readtable` was planned to be deprecated.

PR #1201, opened 2017-07-20
make column Vector always of type Any so arbitrary vectors always can be pushed to it

KristofferC:
Ref https://discourse.julialang.org/t/add-a-column-of-ints-to-a-dataframe-of-strings/4976

Not sure about performance implications.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -70,7 +70,7 @@ size(df1)
 
 """
 type DataFrame <: AbstractDataFrame
-    columns::Vector
+    columns::Vector{Any}
     colindex::Index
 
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks, but at this point changes should be made in DataTables until we backport everything. Maybe the bug is already fixed there?

ararslan:
Can't DataTables already store vectors of any type? So whenever we get around to backporting everything this functionality will be available?

KristofferC:
Looks like https://github.com/JuliaData/DataTables.jl/blob/38750208cc2c81ebfe6946583f52ff193182d8ed/src/datatable/datatable.jl#L73 has it fixed

nalimilan:
Cool, thanks for checking!

PR #1203, opened 2017-07-21
Fixed Compat UTF8String deprecation warnings.

### LINKED ISSUE #1202: WARNING: Compat.UTF8String is deprecated, use String instead.
There are some dozens of warnings like this in `dataframe/io.jl`, could you please fix it in master and perhaps create a new tag?

rofinn:
Closes #1202.

Replaced all instances of Compat.UTF8String with String.
NOTE: Compat deprecated UTF8String when it dropped julia-0.4 support.

### DIFF
--- a/src/dataframe/io.jl
+++ b/src/dataframe/io.jl
@@ -621,7 +621,7 @@ function builddf(rows::Integer,
                     continue
                 else
                     is_bool = false
-                    values = Vector{Compat.UTF8String}(rows)
+                    values = Vector{String}(rows)
                     i = 0
                     continue
                 end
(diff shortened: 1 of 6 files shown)


ararslan:
Since this package is 0.6-only now do we even need the Compat dependency anymore?

rofinn:
Probably not, but that would require a minor (vs patch) release and I imagine we'll need it again as the syntax changes in 0.7.

ararslan:
Adding and subtracting dependencies doesn't require a minor release so long as the APIs don't change. I think it's unlikely anything that this package does will need Compat in the future, and we can always add it back if/when it's needed. In the meantime don't worry about that for this PR.

rofinn:
> Adding and subtracting dependencies doesn't require a minor release so long as the APIs don't change.

Okay, for some reason I thought changing dependencies was a minor release thing. Alright, I'll make a separate PR for dropping Compat then.

PR #1204, opened 2017-07-21
Dropped Compat dependency.

rofinn:


Keno:
If the REQUIRE had said `julia 0.6` instead of `julia0.6-`, FemtoCleaner would have stripped all of these `@compat`s for you.

ararslan:
Good call, thanks. FemtoCleaner did do a pass here.

Issue #1206, opened 2017-07-25
With Julia master, DataFrames precompile fails with range expression argument


colbec:
In 
```
Julia Version 0.7.0-DEV.1106
Commit d328a827da (2017-07-25 04:45 UTC)
```
(checked out latest master of DataFrames, no difference, I think this happened in last 24/48 hours or so) precompile fails with 
```
ERROR: LoadError: LoadError: syntax: missing last argument in "+(baseind,1):" range expression 
Stacktrace:
 [1] include_relative(::Module, ::String) at ./loading.jl:464
 [2] include at ./sysimg.jl:14 [inlined]
 [3] include(::String) at /home/colin/.julia/v0.7/DataFrames/src/DataFrames.jl:3
 [4] macro expansion at /home/colin/.julia/v0.7/DataFrames/src/DataFrames.jl:128 [inlined]
 [5] anonymous at ./<missing>:?
 [6] include_relative(::Module, ::String) at ./loading.jl:464
 [7] include(::Module, ::String) at ./sysimg.jl:14
 [8] anonymous at ./<missing>:2
while loading /home/colin/.julia/v0.7/DataFrames/src/statsmodels/contrasts.jl, in expression starting on line 292
while loading /home/colin/.julia/v0.7/DataFrames/src/DataFrames.jl, in expression starting on line 97
ERROR: Failed to precompile DataFrames to /home/colin/.julia/lib/v0.7/DataFrames.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:598
 [2] _require(::Symbol) at ./loading.jl:387
 [3] require(::Symbol) at ./loading.jl:318

```

ararslan:
I think this was broken by https://github.com/JuliaLang/julia/pull/22919, since we're using `(baseind+1):end` as part of a `cat` expression.

ararslan:
Fixed in Julia master

PR #1207, opened 2017-07-25
Pair Constructor

omus:
Created a `DataFrame` constructor that uses pairs. Works similarly to the keyword argument constructor but allows you to use column names that are Julia keywords:

```julia
julia> DataFrame(type=[])
ERROR: syntax: unexpected "="

julia> DataFrame(:type => [])
0×1 DataFrames.DataFrame
```

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -9,6 +9,7 @@ particularly a Vector or CategoricalVector.
 ```julia
 DataFrame(columns::Vector, names::Vector{Symbol})
 DataFrame(kwargs...)
+DataFrame(pairs::Pair{Symbol}...)
 DataFrame() # an empty DataFrame
 DataFrame(t::Type, nrows::Integer, ncols::Integer) # an empty DataFrame of arbitrary size
 DataFrame(column_eltypes::Vector, names::Vector, nrows::Integer)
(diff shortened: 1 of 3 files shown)


ararslan:
Nightly failure is #1206.

src/dataframe/dataframe.jl
```diff
@@ -102,6 +103,14 @@ function DataFrame(; kwargs...)
     return result
 end
 
+function DataFrame(pairs::Pair{Symbol}...)
```
nalimilan: Looks like you could rewrite the keyword arguments constructor above using the new constructor, right?

nalimilan:
Why not, that would be consistent with `Dict`. But let's wait until the DataTables changes are backported, else this change is going to be lost.

src/dataframe/dataframe.jl
omus: Yes we could. It would look something like:
```julia
function DataFrame(; kwargs...)
    function pair(t) 
        k, v = t
        k => v
    end
    DataFrame(map(pair, kwargs)...)
end
```
Look good?

omus:
> But let's wait until the DataTables changes are backported, else this change is going to be lost.
Sounds good to me

src/dataframe/dataframe.jl
nalimilan: Yeah, but you can just use a generator rather than defining `pair` and calling `map`:
```julia
DataFrame((k => v for t in kwargs)...)
```

src/dataframe/dataframe.jl
```diff
@@ -9,6 +9,7 @@ particularly a Vector, DataVector, or PooledDataVector.
 ```julia
 DataFrame(columns::Vector, names::Vector{Symbol})
 DataFrame(kwargs...)
+DataFrame(Pair{Symbol}...)
```
nalimilan: Need `::` (and check that the result is OK locally by running `julia docs/make.jl`).

omus: Thanks for catching this. I've corrected the documentation. Note that looking at the documentation locally doesn't help much as the output format is still markdown.

omus:
Thanks @nalimilan. I'll squash my commits and then let this gather dust.

quinnj:
Ready for a rebase and then we can merge.

omus:
@quinnj anything else to do here? Nightly looks borked for DataFrames in general

rofinn:
@omus Yeah, the failure on nightly is expected. This looks good to me.

PR #1209, opened 2017-07-30
RFC: Adjust to eltype changes in DataArrays where the function now returns Union{T,NAtype}

andreasnoack:
These are the adjustments required for DataFrames's tests to pass after the changes in https://github.com/JuliaStats/DataArrays.jl/pull/280

### DIFF
--- a/src/statsmodels/formula.jl
+++ b/src/statsmodels/formula.jl
@@ -62,7 +62,7 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-const AbstractFloatMatrix{T<:AbstractFloat} = AbstractMatrix{T}
+const AbstractFloatMatrix{T<:DataArrays.Data{AbstractFloat}} = AbstractMatrix{T}
 
 type ModelMatrix{T <: AbstractFloatMatrix}
     m::T
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -439,9 +439,8 @@ completecases(df)
 
 """
 function completecases(df::AbstractDataFrame)
-    ## Returns a Vector{Bool} of indexes of complete cases (rows with no NA's).
-    res = (!).(isna.(df[1]))
-    for i in 2:ncol(df)
+    res = trues(size(df, 1))
+    for i in 1:ncol(df)
         res .&= (!).(isna.(df[i]))
     end
     res
@@ -480,7 +479,7 @@ function Base.convert(::Type{Array}, df::AbstractDataFrame)
     convert(Matrix, df)
 end
 function Base.convert(::Type{Matrix}, df::AbstractDataFrame)
-    T = reduce(typejoin, eltypes(df))
+    T = mapreduce(DataArrays.extractT, typejoin, eltypes(df))
     convert(Matrix{T}, df)
 end
 function Base.convert{T}(::Type{Array{T}}, df::AbstractDataFrame)
@@ -502,7 +501,7 @@ function Base.convert(::Type{DataArray}, df::AbstractDataFrame)
     convert(DataMatrix, df)
 end
 function Base.convert(::Type{DataMatrix}, df::AbstractDataFrame)
-    T = reduce(typejoin, eltypes(df))
+    T = mapreduce(DataArrays.extractT, typejoin, eltypes(df))
     convert(DataMatrix{T}, df)
 end
 function Base.convert{T}(::Type{DataArray{T}}, df::AbstractDataFrame)
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -94,7 +94,7 @@ function stack(df::AbstractDataFrame, measure_vars)
     stack(df, mv_inds, _setdiff(1:ncol(df), mv_inds))
 end
 function stack(df::AbstractDataFrame)
-    idx = [1:length(df);][[t <: AbstractFloat for t in eltypes(df)]]
+    idx = [1:length(df);][[t <: DataArrays.Data{AbstractFloat} for t in eltypes(df)]]
     stack(df, idx)
 end
 
@@ -102,7 +102,7 @@ end
 Stacks a DataFrame; convert from a wide to long format; see
 `stack`.
 """
-melt(df::AbstractDataFrame, id_vars::Union{Int,Symbol}) = melt(df, [id_vars])
+melt(df::AbstractDataFrame, id_vars::DataArrays.Data{Int}) = melt(df, [id_vars])
 function melt(df::AbstractDataFrame, id_vars)
     id_inds = index(df)[id_vars]
     stack(df, _setdiff(1:ncol(df), id_inds), id_inds)
@@ -441,7 +441,7 @@ function stackdf(df::AbstractDataFrame, measure_vars)
     stackdf(df, m_inds, _setdiff(1:ncol(df), m_inds))
 end
 function stackdf(df::AbstractDataFrame)
-    idx = [1:length(df);][[t <: AbstractFloat for t in eltypes(df)]]
+    idx = [1:length(df);][[t <: DataArrays.Data{AbstractFloat} for t in eltypes(df)]]
     stackdf(df, idx)
 end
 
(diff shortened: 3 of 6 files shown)


src/statsmodels/formula.jl
```diff
@@ -62,7 +62,7 @@ type ModelFrame
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-@compat const AbstractFloatMatrix{T<:AbstractFloat} = AbstractMatrix{T}
+@compat const AbstractFloatMatrix{T<:Union{AbstractFloat,NAtype}} = AbstractMatrix{T}
```
nalimilan: The name of the alias is not completely correct with this change. Maybe you could just remove that variable and use `ModelMatrix{S<:Union{AbstractFloat,NAtype}, T<:AbstractMatrix{S}}` below now that it's supported?

quinnj:
No longer relevant?

nalimilan:
Only relevant if somebody wants to make another (last?) release based on DataArrays.

ararslan:
Probably not worth it at this point. The DataArrays tag that introduces the `eltype` change still hasn't been merged, as it's expected to be breaking and folks wanted to know the extent of the breakage first. If we're moving away from DataArrays entirely, our last DataArrays-based release should be with DataArrays as it is now, perhaps upper bounded on 0.7.0. That would also make this PR obsolete.

ararslan:
I'm not sure whether the changes introduced in DataArrays 0.7.0 will be user-facing after this PR, but if so, it's likely not worth the churn to then change it again as we adopt Nulls.

nalimilan:
We could still imagine replacing `NA` with `null` in DataArrays, and then tag a release of both the new DataArrays and DataFrames based on it, at the same time as a release based on master. That way people who need the performance of `DataArray` on Julia 0.6 could have it, while starting to port some of their code to Nulls.

Of course that's also a lot of work for an unclear gain, so I guess that depends on somebody being directly interested in it (maybe @andreasnoack is).

ararslan:
I believe @iamed2 was particularly interested in `Union` element types in DataArrays (i.e. Andreas' "stop lying" PR), so perhaps it would be good to hear from Eric as well from a user's perspective.

iamed2:
This isn't important if the ecosystem can stabilize around Nulls soon. And we only care about DataArrays insofar as they're the array type for DataFrames. If they stop being that, they're irrelevant to us.

ararslan:
Great, thanks for the input. I'm not sure it's worth churning DataArrays if we aren't going to use them here anymore. But the upside of no longer using them here is that we can update DataArrays independently of DataFrames, so we can move to Nulls there later if we want. (Assuming DataArrays will still be relevant once things settle on Nulls, which I imagine it won't be.)

andreasnoack:
Fine with me but has anybody made any benchmarks comparing operations on `Array{Union{T,Null},N}` to `DataArray{T,N}`? Just curious.

andreasnoack:
Just two small examples for now. (I'm testing `DataArrays` on 0.6 since `DataStructures` appears to be broken on master)

### `sum`
```julia
# Union{Float64,Null}
julia> x = Union{Float64,Null}[rand() for i in 1:10^7];

julia> @time sum(x);
  0.065898 seconds (65.54 k allocations: 1.000 MiB)

# DataArrays
julia> y = @data rand(10^7);

julia> @time sum(x);
  0.007327 seconds (9 allocations: 320 bytes)
```

### `broadcast`
```julia
# Union{Float64,Null}
julia> x = Union{Float64,Null}[rand() for i in 1:10^7];

julia> @time broadcast(+, x, 1);
  0.447139 seconds (20.00 M allocations: 381.479 MiB, 35.30% gc time)

julia> x[end] = Null()
null

julia> @time broadcast(+, x, 1);
  1.849494 seconds (40.00 M allocations: 762.941 MiB, 27.76% gc time)

# DataArrays
julia> y = @data rand(10^7);

julia> @time broadcast(+, y, 1);
  0.057529 seconds (61 allocations: 78.681 MiB, 20.38% gc time)

julia> y[end] = NA
NA

julia> @time broadcast(+, y, 1);
  0.054045 seconds (61 allocations: 78.681 MiB, 19.61% gc time)
```

ararslan:
Brutal

quinnj:
Having a brief glance, I imagine most of the cost is https://github.com/JuliaLang/julia/issues/23338, i.e. nothing getting inlined anywhere so the dynamic dispatch is killing performance.

rofinn:
@quinnj It seems to be allocating a lot of memory too though.
```julia
julia> using BenchmarkTools

julia> using Nulls

julia> using DataArrays

julia> x = Union{Float64,Null}[rand() for i in 1:10^7];

julia> y = @data rand(10^7);

julia> @benchmark sum($x)
BenchmarkTools.Trial:
  memory estimate:  763.94 MiB
  allocs estimate:  50065533
  --------------
  minimum time:     494.707 ms (5.28% GC)
  median time:      505.274 ms (5.27% GC)
  mean time:        506.101 ms (5.28% GC)
  maximum time:     526.756 ms (5.40% GC)
  --------------
  samples:          10
  evals/sample:     1

julia> @benchmark sum($y)
BenchmarkTools.Trial:
  memory estimate:  112 bytes
  allocs estimate:  2
  --------------
  minimum time:     5.421 ms (0.00% GC)
  median time:      5.887 ms (0.00% GC)
  mean time:        6.083 ms (0.00% GC)
  maximum time:     13.873 ms (0.00% GC)
  --------------
  samples:          819
  evals/sample:     1
```
Couldn't we just repurpose DataArrays to operate on `Union{T, Null}` instead?

quinnj:
Oh yes, it definitely will on 0.6; isbits Union arrays having their elements stored inline is 0.7 only.

nalimilan:
`Union{Null, T}` performance is 10x better on Julia 0.7 compared with 0.6 for `sum`, and 2x better for `broadcast` without nulls (no clear difference when there's a null value). So the difference with `DataArray` is already less of an issue there, and could probably be reduced even more. But it would still be problematic to release the new Nulls-based DataFrames on Julia 0.6 given the poor performance.

andreasnoack:
Just to avoid confusion, my timings of `Union{Null, T}` were on 0.7 (from yesterday). Only the `DataArrays` timings were made on 0.6.

nalimilan:
Ah, OK. So that would be even worse on 0.6. :-/

nalimilan:
A possible intermediate solution would be to port DataArrays to `Nulls` (really, this just implies deprecating `NA` in favor of `Null`; `PooledDataArray` could also be deprecated in favor of `CategoricalArray`), and check that they are supported in DataFrames master, at least for most operations. Since DataFrames is now column type agnostic, it should already work, except that when a new column is created (concatenation, merge) an `Array{Union{Null, T}}` is created. We could add some temporary promotion rules so that `DataArray` wins. Then people could keep using DataArrays if they care about performance.

nalimilan:
See https://github.com/JuliaStats/DataArrays.jl/pull/288 for a DataArrays port to Nulls.

nalimilan:
For reference, things have of course improved a lot in 1.0, but we've still regressed a lot. We should probably file issues to track progress for these specific cases.

On Julia 0.6:
```julia
julia> x = DataArray(Union{Float64,Missing}[rand() for i in 1:10^7]);

julia> @btime sum(x);
  5.518 ms (2 allocations: 128 bytes)

julia> @btime broadcast(+, x, 1);
  53.678 ms (70 allocations: 78.68 MiB)

julia> x[end] = missing
missing

julia> @btime broadcast(+, x, 1);
  54.128 ms (70 allocations: 78.68 MiB)

julia> x[1] = missing
missing

julia> @btime broadcast(+, x, 1);
  52.828 ms (70 allocations: 78.68 MiB)
```
On Julia 1.0:
```julia
julia> x = Union{Float64,Missing}[rand() for i in 1:10^7];

julia> @btime sum(x);
  31.034 ms (1 allocation: 16 bytes)

julia> @btime broadcast(+, x, 1);
  127.155 ms (8 allocations: 76.29 MiB)

julia> x[end] = missing
missing

julia> @btime broadcast(+, x, 1);
  485.791 ms (19999498 allocations: 467.29 MiB)

julia> x[1] = missing
missing

julia> @btime broadcast(+, x, 1);
  734.313 ms (9 allocations: 85.83 MiB)
```

Closing anyway since the solution won't come from this PR.

PR #1215, opened 2017-08-22
Integrate with TableTraits.jl

davidanthoff:
This moves the iterable table traits implementation from IterableTables.jl (where it lived behind a ``@requires`` clause) into this package.

In terms of sequencing, it would be fantastic if we could do a DataFrames.jl release that includes this PR but not all the major redesign things that are going on right now in various PRs, so that we have one tagged version of DataFrames.jl that is integrated with TableTraits.jl pre DataFrame-geddon.

Maybe one option would be to merge this to ``master`` and then tag a release based on that commit?

### DIFF
--- a/src/other/tabletraits.jl
+++ b/src/other/tabletraits.jl
@@ -1,0 +1,154 @@
+import TableTraits
+import DataValues
+import NamedTuples
+
+# T is the type of the elements produced
+# TS is a tuple type that stores the columns of the DataFrame
+immutable DataFrameIterator{T, TS}
+    df::DataFrame
+    # This field hols a tuple with the columns of the DataFrame.
+    # Having a tuple of the columns here allows the iterator
+    # functions to access the columns in a type stable way.
+    columns::TS
+end
+
+TableTraits.isiterable(x::DataFrame) = true
+TableTraits.isiterabletable(x::DataFrame) = true
+
+function TableTraits.getiterator(df::DataFrame)
+    col_expressions = Array{Expr,1}()
+    df_columns_tuple_type = Expr(:curly, :Tuple)
+    for i in 1:length(df.columns)
+        if isa(df.columns[i], DataArray)
+            push!(col_expressions, Expr(:(::), names(df)[i], DataValues.DataValue{eltype(df.columns[i])}))
+        else
+            push!(col_expressions, Expr(:(::), names(df)[i], eltype(df.columns[i])))
+        end
+        push!(df_columns_tuple_type.args, typeof(df.columns[i]))
+    end
+    t_expr = NamedTuples.make_tuple(col_expressions)
+
+    t2 = :(DataFrameIterator{Float64,Float64})
+    t2.args[2] = t_expr
+    t2.args[3] = df_columns_tuple_type
+
+    t = eval(t2)
+
+    e_df = t(df, (df.columns...))
+
+    return e_df
+end
+
+function Base.length{T,TS}(iter::DataFrameIterator{T,TS})
+    return size(iter.df,1)
+end
+
+function Base.eltype{T,TS}(iter::DataFrameIterator{T,TS})
+    return T
+end
+
+function Base.start{T,TS}(iter::DataFrameIterator{T,TS})
+    return 1
+end
+
+@generated function Base.next{T,TS}(iter::DataFrameIterator{T,TS}, state)
+    constructor_call = Expr(:call, :($T))
+    for i in 1:length(iter.types[2].types)
+        if iter.parameters[1].parameters[i] <: DataValues.DataValue
+            push!(constructor_call.args, :(isna(columns[$i],i) ? $(iter.parameters[1].parameters[i])() : $(iter.parameters[1].parameters[i])(columns[$i][i])))
+        else
+            push!(constructor_call.args, :(columns[$i][i]))
+        end
+    end
+
+    quote
+        i = state
+        columns = iter.columns
+        a = $constructor_call
+        return a, state+1
+    end
+end
+
+function Base.done{T,TS}(iter::DataFrameIterator{T,TS}, state)
+    return state>size(iter.df,1)
+end
+
+# Sink
+
+@generated function _tabletraits_fill_df(columns, enumerable)
+    n = length(columns.types)
+    push_exprs = Expr(:block)
+    for i in 1:n
+        if columns.parameters[i] <: DataArray
+            ex = :( push!(columns[$i], isnull(i[$i]) ? DataArrays.NA : get(i[$i])) )
+        else
+            ex = :( push!(columns[$i], i[$i]) )
+        end
+        push!(push_exprs.args, ex)
+    end
+
+    quote
+        for i in enumerable
+            $push_exprs
+        end
+    end
+end
+
+function _construct_dataframe_from_iterabletable(x)
+    iter = TableTraits.getiterator(x)
+
+    T = eltype(iter)
+    if !(T<:NamedTuples.NamedTuple)
+        error("Can only collect a NamedTuple iterator into a DataFrame")
+    end
+
+    column_types = TableTraits.column_types(iter)
+    column_names = TableTraits.column_names(iter)
+
+    columns = []
+    for t in column_types
+        if isa(t, TypeVar)
+            push!(columns, Array{Any}(0))
+        elseif t <: DataValues.DataValue
+            push!(columns, DataArray(t.parameters[1],0))
+        else
+            push!(columns, Array{t}(0))
+        end
+    end
+    df = DataFrame(columns, fieldnames(T))
+    _tabletraits_fill_df((df.columns...), iter)
+    return df
+end
+
+function DataFrame(x::Array{T,1}) where {T<:NamedTuples.NamedTuple}
+    return _construct_dataframe_from_iterabletable(x)
+end
+
+function DataFrame(x)
+    if TableTraits.isiterabletable(x)
+        return _construct_dataframe_from_iterabletable(x)        
+    else
+        return convert(DataFrame, x)
+    end
+end
+
+function ModelFrame(f::Formula, source; kwargs...)
+    TableTraits.isiterabletable(source) || error()
+    return ModelFrame(f, DataFrame(source); kwargs...)
+end
+
+function StatsBase.fit{T<:StatsBase.StatisticalModel}(::Type{T}, f::Formula, source, args...; contrasts::Dict = Dict(), kwargs...)
+    TableTraits.isiterabletable(source) || error()
+    mf = ModelFrame(f, source, contrasts=contrasts)
+    mm = ModelMatrix(mf)
+    y = model_response(mf)
+    DataFrameStatisticalModel(fit(T, mm.m, y, args...; kwargs...), mf, mm)
+end
+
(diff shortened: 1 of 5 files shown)


rofinn:
If this PR was just adding some methods that allowed it to better integrate with the TableTraits / IndexableTables ecosystem then I'd support including it, but adding 3 new dependencies right before we start backporting DataTables and begin refactoring seems like it'll just create more work and confusion for us.

quinnj:
I agree w/ @rofinn. This is really the perfect candidate for "optional dependencies" if there ever was one. It doesn't make sense for this to live in IterableTables.jl/Query.jl, which would require dependencies on a crazy # of packages. It also doesn't feel great here since we're actively working on _removing_ as many dependencies as possible for DataFrames. And as @rofinn mentioned, it makes it harder for the planned improvements/development if there's this "optional" code that is layered on top of the core functionality.

I think a better option would be to have a DataFramesTableTraits.jl package that can pin against the current release version of DataFrames and will be able to update once all the changes are made here.

I also think we should really push for a true "optional dependency" solution in core julia. Ideally, there'd be a way to express in IterableTables/Query.jl that if DataFrames is/will be loaded, the interface implementation code should also be loaded along w/ DataFrames. That allows the code to live where it was developed (IterableTables) while not requiring a hard dependency.

davidanthoff:
I think there are two questions here: 1) should this code live in DataFrames.jl?, and 2) if yes, should it be moved in now or later.

On 1), I think it should be in DataFrames.jl, not in some third package. The reason is simple: if someone writes ``using DataFrames, TypedTables``, the machinery here should kick in and enable a smooth interop. If this code lived in a package like DataFramesTableTraits.jl we will *always* end up with type piracy, plus people then need to know about TableTraits.jl and that they need to load extra packages to make things work smoothly. It just adds friction for users in my mind. Same with the whole File IO story described [here](https://discourse.julialang.org/t/ann-file-io-for-tabular-data/4712), I think it would be better if that worked if folks just do ``using FileIO, DataFrames``. The three dependencies are harmless, in my mind: NamedTuples.jl will go away in any case once julia 1.0 is out (plus Jeff seems to maintain it, so it is really high quality), TableTraits.jl is really super, super minimal and DataValues.jl also does nothing fancy. Plus, isn't this exactly the same situation as with the DataStreams.jl integration? Or are there plans to also remove them from the package?

The question of timing might be more tricky and I agree we should plan that carefully. I don't see how the extra dependencies would make any of the new plans more difficult. But, one would of course have to change the table traits integration code as the definition for ``DataArray`` changes and then once things move to ``Array{Union{T,Null},1}``. That will be extra work. I'd be happy to take care of those code changes, they really should be very straightforward.

davidanthoff:
Ah wait, DataStreams.jl integration is not part of this? I thought that was moved into this package?

rofinn:
The DataStreams.jl is currently part of the DataTables.jl backport, but it is currently commented out and I'm not sure if that'll stay. However, if it does it, it should already support `Union{T, Nulls}` and probably won't require any other dependencies.

quinnj:
> 1) should this code live in DataFrames.jl?

I really don't think so. We already went through this process with the `Juno.jl` dependency; it was added because "it's just a lightweight thing to provide some interop functionality". We ended up having to rip it out because of complications down the road. And to be clear, it's definitely not because of the code itself: I'm sure DataValues.jl/NamedTuples.jl/etc are probably fine, but the idea is that we need a core "DataFrames" package that is lightweight and truly avoids any non-core dependencies.

I think we'd be better served in the short-term (until there's a better optional dependency solution), to split DataFrames.jl up into a DataFramesType.jl package and a DataFrames.jl package that could hold all the "extra" code like this.

ararslan:
> I really don't think so. We already went through this process with the Juno.jl dependency; it was added because "it's just a lightweight thing to provide some interop functionality". We ended up having to rip it out because of complications down the road. And to be clear, it's definitely not because of the code itself: I'm sure DataValues.jl/NamedTuples.jl/etc are probably fine, but the idea is that we need a core "DataFrames" package that is lightweight and truly avoids any non-core dependencies.

👍 

> I think we'd be better served in the short-term (until there's a better optional dependency solution), to split DataFrames.jl up into a DataFramesType.jl package and a DataFrames.jl package that could hold all the "extra" code like this.

👎

davidanthoff:
So in my mind this really comes down to whether there should be a core table interop story or not. If the iterable tables story is to be the most basic, primitive idea of a table interface, then table packages will have to take a dependency on it. Given that one of the things in this PR is a constructor ``function DataFrame(data)`` (i.e. a constructor that catches anything that didn't get caught by a more specific constructor), it really is not good to have that in a separate package. That would just be the worst kind of type piracy. If there was a split a la DataFramesType.jl and DataFrames.jl, the iterable tables stuff should live in DataFramesType.jl because it adds that constructor to ``DataFrame``.

I'd be interested to hear what the problems with Juno.jl were, certainly we wouldn't want to make the same mistakes again. But it is a bit hard for me to judge without more details what went wrong there (although, looking at the runtest.jl file in Juno.jl, I can kind of see that it would cause trouble ;).

rofinn:
FWIW, here is how I see this fitting in with the existing steps/objectives discussed elsewhere:

1. Get all the changes we want backported from DataTables.
2. Deprecate DataTables, so we have one less Tables package to worry about ;)
3. Switch to using `Union{T, Null}`. It seems like this will likely happen at the same time as step 1 to avoid migrating between a 3rd storage type on master. Unfortunately, this requires some cleanup on DataTables and DataStreams first.
3. Work on simplifying the existing default `DataFrame` type. I imagine this will eventually look like a mixture of @quinnj's [NTTables](https://github.com/quinnj/NTTables.jl) and [TableFrames](https://github.com/rofinn/TableFrames.jl) (e.g., a minimal Table type which provides operations on any associative of abstract vectors where the default backend will use a `NTDict` like `NTTables`). This might make sense in its own package that is re-export from DataFrames.jl
4. **Add the `TableTraits` interface** if we're already depending on `NamedTuples` and the `TableTraits` interface no longer has a required dependence on `DataValues`.
5. Start factoring out other table operations like (e.g., `join`, `aggregate`, `dropnull`, `printtable`)

I expect DataFrames.jl will eventually be a wrapper package that re-exports functionality from a collection of packages.

nalimilan:
I'd be in favor of defining a common interface in a small package (be it TableTraits.jl or the AbstractTables.jl package we've been talking about for ages) and adding a dependency on it in DataFrames as long as it does not add any indirect dependency. Then any package could use it to support DataFrames, including Query.jl and DataStreams.jl. But DataFrames really cannot depend on DataValues: only one representation of missing values should be used.

davidanthoff:
My roadmap for TableTraits.jl use of DataValues.jl is this: there won't be a change until julia 1.0 is out. I need something that is a) fast, b) compatible with the Query.jl design and c) easy to use. ``DataValue`` is the only way to get all three on julia 0.6, as far as I can tell. ``Nullable`` has a) and b) but not c) and Nulls.jl has only c) on julia 0.6. Once julia 1.0 is out I'll take another look, if the approach in Nulls.jl can deliver on all three requirements at that point I will most likely change TableTraits.jl over to that approach.

Note though that even though the iterable tables interface right now says that missing values have to be represented using a ``DataValue`` *in the interface*, it does not impose ``DataValue`` on the types that interact with iterable tables. For example, if you start with a ``DataFrame`` that has ``DataArray``s today, convert it to a ``DataTable`` via the iterable tables interface, you'll end up with a ``DataTable`` that uses ``NullableArray`` for storing missing values. If you go the other way around, same story, you'll end up with a ``DataFrame`` that uses ``DataArray``s. So the design really tries to respect the choices that individual packages make on how they want to deal with missing values. In practice that means that even if DataFrames.jl were to take a dependency on TableTraits.jl and DataValues.jl now to integrate with the iterable tables interface, users would literally never know, see or experience a ``DataValue``, given that the iterable tables interface is really a low level interface. If the interface migrates to Nulls.jl in the future, users would also not even notice.

I think at the end of the day you guys, the maintainers of this package, need to just decide whether you want this PR or not, I'm not going to fight for it. Here is what is going to happen on my end re integrations that aren't migrated into the packages: I will keep them in IterableTables.jl, but packages like Query.jl, CSVFiles.jl, FeatherFiles.jl, ExcelFiles, StatFiles.jl and soon ParquetFiles.jl (almost there!) won't have a dependency on IterableTables.jl, instead they are going to rely on TableTraits.jl only. So they will bring in the minimal interface definition, but they won't automatically load the integration code itself. For users that means if they want to use packages that don't have the iterable tables integration in the package itself, they will have to manually load ``using IterableTables`` to be able to use a package like DataFrames.jl with the ecosystem around iterable tables. I don't think it is ideal and as a user I would prefer things to just work, but it is also not the end of the world.

nalimilan:
I see the logic of your design and that's great that it works, but I think it's clear that the dependency on DataValues is a non-starter for DataFrames. Hopefully we'll have a unified framework to represent missing values for Julia 1.0.

quinnj:
Yeah, I think keeping the implementation in IterableTables.jl will honestly be the easiest for everyone right there. There's just a ton of churn/dev going on between DataTables/DataFrames right now, so it'll just be simpler to keep things separate.

src/other/tabletraits.jl
```diff
@@ -0,0 +1,154 @@
...
+    df = DataFrame(columns, fieldnames(T))
+    _tabletraits_fill_df((df.columns...), iter)
+    return df
+end
+
+function DataFrame(x::Array{T,1}) where {T<:NamedTuples.NamedTuple}
+    return _construct_dataframe_from_iterabletable(x)
+end
+
+function DataFrame(x)
```
quinnj: I feel like this is one of the most unfortunate parts of the IterableTables required interface; table types should be able to define their own catchall constructor without having to be required for IterableTables functionality. I think it'd be better to transition this to an `IterableTables.IterableTable(::Type{DataFrame}, x)` method that packages can implement.

The catchall constructor ends up getting called from [here](https://github.com/davidanthoff/Query.jl/blob/ca1b176326840c9e4c622fb3e068294fbdf88e27/src/sinks/sink_type.jl#L2), right? I think it'd be better to not have to rely on the catchall constructor there.

davidanthoff: This is not part of the required interface of IterableTables.jl, I only recommend it (I think that is pretty clear in the documentation). If sinks don't want to do this, they can do whatever else they want. The whole user API won't be as smooth for them, but they can still participate in the iterable tables story.

The question of how to integrate with Query.jl is to some extent distinct. In general you don't have to terminate a query with a ``@collect`` statement, and then you can just use whatever method you want to materialize things. But yes, the ``@collect T`` pattern requires that there is a constructor for ``T`` that can deal with whatever the query returns. The new piping syntax also only works if that is the case: ``df |> @where(_.a>2) |> DataFrame`` obviously requires a ``DataFrame`` constructor that can handle whatever is returned by ``@where``.

quinnj: C'mon, that's not really fair. Basically, if you don't, you can't use full Query functionality w/ your type, oh, but it's still optional? It really shouldn't be hard to change it to be a proper part of the interface, instead of requiring this "constructor coercion".

davidanthoff: Probably best to discuss this in an issue over in Query.jl, do you want to open an issue there?

nalimilan:
This is now obsolete now that we integrate with Tables.jl (https://github.com/JuliaData/DataFrames.jl/pull/1495), right?

quinnj:
Yes, #1495 added TableTratis integration

PR #1217, opened 2017-08-23
Added return type annotation to `completecases(df)`.

### LINKED ISSUE #1212: Cannot getindex
Here in versions >= 0.10.0 `getindex` cannot be applied to `cc` which is `DataArray` or `Pooled...` here.

https://github.com/JuliaStats/DataFrames.jl/blob/2b8a2bd4e58384f9993ff3884ba04cdd68e3d0e3/src/statsmodels/formula.jl#L236

rofinn:
Fixes #1212

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -438,8 +438,7 @@ completecases(df)
 ```
 
 """
-function completecases(df::AbstractDataFrame)
-    ## Returns a Vector{Bool} of indexes of complete cases (rows with no NA's).
+function completecases(df::AbstractDataFrame)::AbstractVector{Bool}
     res = (!).(isna.(df[1]))
     for i in 2:ncol(df)
         res .&= (!).(isna.(df[i]))
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -438,8 +438,8 @@ completecases(df)
 ```
 
 """
-function completecases(df::AbstractDataFrame)::AbstractVector{Bool}
-    res = (!).(isna.(df[1]))
+function completecases(df::AbstractDataFrame)
+    res::AbstractVector{Bool} = (!).(isna.(df[1]))
     for i in 2:ncol(df)
         res .&= (!).(isna.(df[i]))
     end
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -316,4 +316,8 @@ module TestDataFrame
     df4[2,:Mass] = NA
     @test isequal(df2, df4)
 
+
+    # completecases fix for #1212
+    df = DataFrame(A = Any[1, 2, 3, 4], B = Any[true, false, true, false])
+    @test isa(completecases(df), AbstractVector{Bool})
 end


ararslan:
Can you add an example of the failing behavior in #1212 as a test?

rofinn:
I suppose I could also test `DataFrames.na_omit(df)` which just tries to index with the result of `completecases`, but that seems unnecessary

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -438,8 +438,7 @@ completecases(df)
 ```
 
 """
-function completecases(df::AbstractDataFrame)
-    ## Returns a Vector{Bool} of indexes of complete cases (rows with no NA's).
+function completecases(df::AbstractDataFrame)::AbstractVector{Bool}
```
nalimilan: Maybe the type annotation should be added to `res` inside the function instead? We don't want an internal type instability to hurt performance.

rofinn:
@nalimilan I think the issue is that `isna` doesn't always return a `Bool`, so julia can't infer the type in the broadcast.
```julia
julia> code_typed(isna)
6-element Array{Any,1}:
 CodeInfo(:(begin
        return $(QuoteNode(true))
    end))=>Bool
 CodeInfo(:(begin
        return (Core._apply)(DataArrays.getindex, (Core.tuple)((Core.getfield)(da, :na)::BitArray{N} where N, I)::Tuple{BitArray{N} where N,Real}, Is)
    end))=>Any
 CodeInfo(:(begin
        return (DataArrays.getindex)((Core.getfield)(pda, :refs)::Array{R,N} where N where R<:Integer, i)::Integer == 0
    end))=>Any
 CodeInfo(:(begin  # line 57:
        $(Expr(:invoke, MethodInstance for depwarn(::String, ::Symbol), :(Base.depwarn), :($(Expr(:invoke, MethodInstance for string(::String, ::Vararg{String,N} where N), :(Base.string), "isna(x::AbstractArray)", " is deprecated, use ", "isna.(x)", " instead."))), :(DataArrays.#89###oldmtname#663))) # line 59:
        return (Base.broadcast)(DataArrays.isna, x)
    end))=>Any
 CodeInfo(:(begin
        unless (DataArrays.NAtype <: $(Expr(:static_parameter, 1)))::Bool goto 3
        return ((DataArrays.getindex)(a, i) isa DataArrays.NAtype)::Bool
        3:
        return false
    end))=>Bool
 CodeInfo(:(begin
        return false
    end))=>Bool

julia> methods(isna)
# 6 methods for generic function "isna":
isna(x::DataArrays.NAtype) in DataArrays at /Users/rory/.playground/share/dataframes/packages/v0.6/DataArrays/src/natype.jl:60
isna(da::DataArrays.DataArray, I::Real, Is::Real...) in DataArrays at /Users/rory/.playground/share/dataframes/packages/v0.6/DataArrays/src/dataarray.jl:252
isna(pda::DataArrays.PooledDataArray, i::Real) in DataArrays at /Users/rory/.playground/share/dataframes/packages/v0.6/DataArrays/src/pooleddataarray.jl:232
isna(x::AbstractArray) in DataArrays at deprecated.jl:56
isna(a::AbstractArray{T,N} where N, i::Real) where T in DataArrays at /Users/rory/.playground/share/dataframes/packages/v0.6/DataArrays/src/abstractdataarray.jl:51
isna(x) in DataArrays at /Users/rory/.playground/share/dataframes/packages/v0.6/DataArrays/src/natype.jl:61
```
This shouldn't be an issue once the two methods that return `Any` are deprecated, but until then annotating it in this code is probably the best option.

src/abstractdataframe/abstractdataframe.jl
rofinn: I like having the return type as part of the function definition cause then it's also self documenting and I didn't notice much of a difference benchmarking the two choices. That said, annotating `res` directly might result in more consistent performance in situations I didn't test against.

andreasnoack: Notice that there is only a single assignment because of the `.&=`. However, it might make sense just to initialize `res` to be the type we'd like and the extend the loop range to include `1`.

rofinn: Preallocating `res` does seem cleaner, but it is 1.4x slower.

rofinn: Alright, I decided to go with @nalimilan's approach since it would at least cause an earlier failure conditions if for some reason the output can't be converted to `AbstractVector{Bool}` and might have better performance in some situations.

PR #1219, opened 2017-08-30
Update example in documentation for unstack

JonathanAnderson:
the `long = stack(wide)` didn't work for me as I was working through the docs. adding the fields, fixed it for me. (I think)


nalimilan:
Thanks for catching this. I think both forms should be equivalent, so there's probably a bug somewhere.

nalimilan:
Mmm, this works perfectly fine here: `stack(wide, [:c, :d]) == stack(wide)` gives `true`. Can you post the error you get and version information?

PR #1222, opened 2017-09-03
Clean up isless(::DataFrameRow, ::DataFrameRow)

nalimilan:
Rewrite the code logic to match that used for Tuple and make it more natural.
Fix tests changed by 6035da8, and which should not use < and ==
(even if the latter falls back to isless).


ararslan:
Fun fact: When I view a page in a language other than English, my browser automatically adds a spam filter for the language. When I viewed the diff here just a moment ago, my browser automatically added an Icelandic spam filter. 🤔

PR #1223, opened 2017-09-07
Remove unused src/dataframe/io.jl file

nalimilan:
This file has been removed in DataTables, but was missed when backporting
changes.


quinnj:
Good catch.

nalimilan:
I wish I would have noticed it when rebasing. Apparently, that's the only one.

Issue #1200, opened 2017-07-17
length(::DataFrame) returns number of columns


omus:
Currently calling `length` on a `DataFrame` returns the number of columns. This is strange as `length` usually returns the number of elements.

omus:
cc: @ararslan @andyferris

ararslan:
Yeah this is really weird. We probably shouldn't define `length` at all.

kmsquire:
There was debate about this when it was first added, mostly between those coming from an R background (to whom, I think, the current definition made sense) and those coming from a Pandas background (where length is the number of rows).  So what makes the most sense probably depends on what you've used before.

ararslan:
Having it be inconsistent between languages is another reason not to define it here, IMO. Then it confuses no one. 🙂

andyferris:
If we want to think of a dataframe in the relational algebra sense (as a collection of named tuples, i.e. rows), then iterating over rows and having `length` for the number of rows makes sense to me.

There has been a lot of discussion about this surrounding Jeff's `NamedTuple` pull request (partly because it is infrastructure for making such iteration fast).

rofinn:
Given that more descriptive methods such as `size`, `nrow` and `ncol` exist (could be better documented though) I don't really see a need to keep `length`.

andyferris:
It goes with iteration, so if you can't iterate a `DataFrame` then you shouldn't have a `length`.

rofinn:
I'm not sure `length` even needs to go with iteration. For example, we can iterate over a `Channel` which doesn't provide a `length` method either.

ararslan:
`length` is an optional part of the iteration protocol, per the documentation. I guess we could have `length` defined on the `EachRow` or whatever iterator types we define for rows/columns, though it doesn't really seem useful there.

nalimilan:
`length(df)` is consistent with the fact that `df[1]` returns the first column. We could remove both and require writing `df[:, 1]`.

`nrow` and `ncol` should probably be deprecated too, cf. https://github.com/JuliaStats/DataFrames.jl/issues/406.

rofinn:
> `length(df)` is consistent with the fact that `df[1]` returns the first column. We could remove both and require writing `df[:, 1]`.

Yeah, I recall that confusing me the first time I used dataframes cause I figured scalar indexing into a dataframe would return the first row.

ararslan:
Okay, so the plan as I understand it:
[...]

Issue #406, opened 2013-11-11
Remove nrow/ncol


johnmyleswhite:
We should remove `nrow` and `ncol`. Everything should be done with `size(df)`.

ararslan:
I'm surprised this hasn't been done in the past 3 years since this issue was brought up. I definitely agree that we should be using `size` everywhere rather than `nrow`/`ncol`.

andreasnoack:
I'm not sure about this. I don't think it is a good idea to make `DataFrames` too similar to arrays. It's better to think of them as a table and therefore we might not want to follow the AbstractArray interface.

ararslan:
Hm, that's a very good point.

nalimilan:
OTOH, `nrow` and `ncol` clearly use the matrix vocabulary. I think we should either move to `size`, or use a database vocabulary like `nobs` and `nvar`.

phaverty:
nrow and ncol will be very familiar to people coming from R, who are likely
to make up a large proportion of the DataFrames users. I would recommend
leaving nrow an ncol in.

Pete

____________________
Peter M. Haverty, Ph.D.
Genentech, Inc.
phaverty@gene.com

nalimilan:
@phaverty While we certainly take it into account, the fact that something exists in a language is never a sufficient argument by itself for including something in Julia. The most important question is whether it makes for a good and consistent design in Julia.

ararslan:
`nobs`/`nvar` seems like a good compromise to me between avoiding array-specific terminology and allowing clear, familiar names for accessing these dimensions.

quinnj:
I'm not a fan of `nobs` personally. I personally like `size` because of it's julian heritage at this point, I think there are enough differentiators from AbstractArray that DataFrames won't be confusing. I like that size also encourages getting the # of rows/columns in a single call as well. I vote we deprecate/remove nrow/ncol, and just keep size.

johnmyleswhite:
+1 for Jacob's proposal. Adding new names seems superfluous.

ararslan:
I'd be fine with that as well.

milktrader:
Also not in favor af new names at this point. Just because ncol/nrow are used in R doesn't mean we have to be different. I prefer it slightly to size because it is more natural to a table vs array interface.

andreasnoack:
[...]

PR #1224, opened 2017-09-08
Deprecate length, nrow, and ncol on DataFrames in favor of size. Fixe…

quinnj:
…s #1200


nalimilan:
+1 for deprecating `nrow` and `ncol`, but I'm not so sure about `length`. As noted in #1200, it's consistent with `df[i]`, so as long as we support doing that I'd rather keep `length`.

garborg:
I'd have thought deprecating size would have been more likely at this point.

I thought the move to treat DataFrames as matrix-like had fizzled out. length and ncols, for example, would be more of a fit for DataFrames that are conceptually a collection of namedtuples.

quinnj:
I think there's been more consensus recently to go with `size` as a more "Julian" idiom rather than as a reason to treat DataFrames as matrix-like, though there's also https://github.com/JuliaLang/julia/issues/22665, where it was debated to change `size` to `shape`. We'd probably want to switch to `shape` if that ends up happening in Base (though it's unclear if anything will get changed there).

I don't see what the big fuss is about deprecating `length(df)`; at this point, it's just the more confusing thing between `length(df)` and `df[1]`, mainly, IMO, because `df[:col]` has been such a staple in DataFrames for a long time. It can also be the "step 1" in transitioning DataFrame to be more of a "bag of namedtuples" representation rather than the bag-of-columns type of definition it's always had.

nalimilan:
> I don't see what the big fuss is about deprecating length(df); at this point, it's just the more confusing thing between length(df) and df[1], mainly, IMO, because df[:col] has been such a staple in DataFrames for a long time. It can also be the "step 1" in transitioning DataFrame to be more of a "bag of namedtuples" representation rather than the bag-of-columns type of definition it's always had.

I'm not sure what you mean here. Can you elaborate?

src/dataframe/dataframe.jl
```diff
@@ -173,9 +173,11 @@ end
 index(df::DataFrame) = df.colindex
 columns(df::DataFrame) = df.columns
 
-# TODO: Remove these
-nrow(df::DataFrame) = ncol(df) > 0 ? length(df.columns[1])::Int : 0
-ncol(df::DataFrame) = length(index(df))
+Base.size(df::DataFrame) = ((length(df.columns) > 0 ? length(df.columns[1]) : 0),
+                            length(index(df)))
+Base.size(df::DataFrame, i::Int) = i == 1 ?
```
nalimilan: Would be cleaner with an `if... else` block (with a long-form function definition), the indentation a bit weird here. BTW, could print "$i" in the error message, that's always useful to understand what went wrong.

garborg:
@quinnj I agree with the issues you brought up about `length` being problematic as a replacement for `nrows`. Take back my suggestion of using `ncols` and `length`, and replace it with using `ncols` and `nrows` (or whatever else).

`size` would be more julian if it wasn't a pun, but I don't see `size(::Collection{NamedTuple}, 2)` returning the number of fields the elements have for other collections of records. 

Leaning solely on a couple (maybe gross) non-base names (like `ncols`), would make the next step easier if it turns out that this change is is not the end state.

Six in one hand half, dozen in the other -- just wanted to point out that `size` seems really punny to me, as I was MIA during #1200 and missed pointing out then that the punniness is why #406 never happened in the couple years after the issue was opened.

rofinn:
FWIW, I like just having `size` cause it's simple and intuitive. I've found `nrow` and `eachrow` hard to remember in the julia ecosystem (e.g., `length(rows(df))` always seemed more intuitive to me).

garborg:
`size` is consistent with how indexing into a cell with row & col happens, too, to argue against my previous comments.

Nectarineimp:
I haven't done much but lurk and submit one PR, but on this issue I will say this - nrow/ncol are nice, but separate operations. A data frame is not a matrix or an array. It has area, so to speak. size() gets you everything you need in an easy to parse result. From my experience, simpler and complete usually wins in design. Compatibility with other packages and aspects of Julia are, of course, major considerations. I am not sure there is enough concern here about compatibility to disqualify size, no?

nalimilan:
@quinnj How about merging the deprecation of `nrow` and `ncol` now? Then we can discuss whether to keep `length` or not.

quinnj:
Meh, if JuliaLang/julia#22665 settles on `size` vs. `shape` then I think it's worth deprecating, but until that's settled, I'm not sure it's worth another potential deprecation later.

nalimilan:
Doesn't seem `size` is going to be deprecated. The only option that appears to be seriously considered is to replace `indices` with  `shape`.

randyzwitch:
Might be late to the party on this, but I'm a fan of `nrow`, one because R does it, but more conceptually, almost every day I sample a DataFrame (or limit in select query), so I quite frequently think of "n(umber of) rows". Whereas I've never cared the number of columns, because they are named. 

`size` to me feels much more of memory usage than the (# rows, #cols) tuple.

bkamins:
Now we have `length` deprecated.

So the decision is do we:
1. retain `nrow` and `ncol`
2. deprecate them in favor of `size`?

In both cases this PR should probably be closed, but if we want to deprecate them I would open a new PR only with this deprecation.

nalimilan:
I'd rather keep them. Since we've decided data frames are collections of rows, their dimensions are not really symmetric like matrices, so it would be weird to have to use `size`.

bkamins:
Fine with me.

bkamins:
So close this?

PR #1225, opened 2017-09-08
Remove dependencies on StatsBase & SortingAlgorithms

quinnj:


quinnj:
Anybody have any over-arching concerns w/ this? cc: @omus, @nalimilan, @kleinschmidt, @ararslan, @rofinn, @cjprybol, @andreasnoack? It's been discussed that once a solution for conditional dependencies exist, it'd be fine to re-enable these kinds of integrations, but for now, it feels more important to keep DataFrames as small/concise as possible as such an important upstream package.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -11,7 +11,7 @@ type in that it allows indexing by a key (the columns).
 
 The following are normally implemented for AbstractDataFrames:
 
-* [`describe`](@ref) : summarize columns
+* [`summarize`](@ref) : summarize columns
```
nalimilan: Ah, a proof that we choose a good name for this!

src/abstractdataframe/sort.jl
```diff
@@ -223,29 +223,9 @@ ordering(df::AbstractDataFrame, cols::Tuple, args...) = ordering(df, [cols...],
 # Default sorting algorithm
 ###########################
 
-# TimSort is fast for data with structure, but only if the DataFrame is large enough
```
nalimilan: Please run at least one reasonable benchmark to check that this doesn't totally kill performance. Also file an issue linking to that code so that we don't forget it.

davidanthoff:
I generally think no functionality should be removed from DataFrames.jl until there is a replacement somewhere. I understand the desire to make DataFrames small/concise, but I think that needs to be balanced with the fact that DataFrames is currently essentially the only really usable data stack package in the julia ecosystem and is used by a lot of folks in their day-to-day work. I'm all for merging something like this PR, *after* someone has moved the functionality into some other package so that end-users can just continue to use these features.

nalimilan:
Note that this doesn't really remove features: `describe` is replaced with `summarize`, and sorting might be slower (need to check).

@quinnj We need a deprecation for `describe`.

quinnj:
How do we deprecate `describe` when it was being overloaded for StatsBase? Optional deperecation? (i.e if StatsBase is defined?)

nalimilan:
That's a tough issue, and I'm afraid there's no solution except keeping the dependency on StatsBase for one more release.

quinnj:
That's fine.

davidanthoff:
> Note that this doesn't really remove features: describe is replaced with summarize, and sorting might be slower (need to check).

Ah, sorry for the noise, I didn't get that :)

rofinn:
FWIW, I'm also inclined to keep statsbase as a dependency for a few reasons.

1. We'll need it for that `describe` deprecation cycle as has already been mentioned.
2. I'd prefer to focus on flushing out AbstractTables.jl and extracting the base DataFrame type (e.g., DataFramesBase.jl).
3. StatsBase doesn't seem like an unreasonable dependency (it only depends on DataStructures.jl and SpecialFunctions.jl).

PR #1226, opened 2017-09-09
Move from JuliaStats to JuliaData

nalimilan:


nalimilan:
Tests passed on Julia 0.6 Linux and Windows, merging so that the Documenter manual is deployed to the new GitHub website (since the old one is no longer available).

PR #1231, opened 2017-09-12
`isempty(df)` should return true if either dimension == 0.

### LINKED ISSUE #1230: `isempty` checks number of columns, rather than number of rows
Related to #1200.

Compare:

```julia
julia> arr = Array{Any}((0,2))
0×2 Array{Any,2}

julia> isempty(arr)
true
```

with

```julia
julia> df = DataFrame(col1=[], col2=[])
0×2 DataFrames.DataFrame


julia> isempty(df)
false
```

It seems clear to me that a `DataFrame` with zero rows should be considered empty.

FWIW, @quinnj's PR from a few days ago https://github.com/JuliaData/DataFrames.jl/pull/1224 fixes this as well, but this particular issue seems less controversial than how we define `length`.

rofinn:
Fixes #1230.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -254,7 +254,7 @@ end
 
 Base.haskey(df::AbstractDataFrame, key::Any) = haskey(index(df), key)
 Base.get(df::AbstractDataFrame, key::Any, default::Any) = haskey(df, key) ? df[key] : default
-Base.isempty(df::AbstractDataFrame) = ncol(df) == 0
+Base.isempty(df::AbstractDataFrame) = size(df, 2) == 0 || size(df, 1) == 0
 
 ##############################################################################
 ##
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -254,7 +254,7 @@ end
 
 Base.haskey(df::AbstractDataFrame, key::Any) = haskey(index(df), key)
 Base.get(df::AbstractDataFrame, key::Any, default::Any) = haskey(df, key) ? df[key] : default
-Base.isempty(df::AbstractDataFrame) = size(df, 2) == 0 || size(df, 1) == 0
+Base.isempty(df::AbstractDataFrame) = size(df, 1) == 0 || size(df, 2) == 0
 
 ##############################################################################
 ##
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -93,6 +93,7 @@ module TestDataFrame
     @test empty!(df) === df
     @test isempty(df.columns)
     @test isempty(df)
+    @test isempty(DataFrame(a=[], b=[]))
 
     df = DataFrame(a=Union{Int, Null}[1, 2], b=Union{Float64, Null}[3.0, 4.0])
     @test_throws BoundsError insert!(df, 5, ["a", "b"], :newcol)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -254,7 +254,7 @@ end
 
 Base.haskey(df::AbstractDataFrame, key::Any) = haskey(index(df), key)
 Base.get(df::AbstractDataFrame, key::Any, default::Any) = haskey(df, key) ? df[key] : default
-Base.isempty(df::AbstractDataFrame) = ncol(df) == 0
+Base.isempty(df::AbstractDataFrame) = size(df, 2) == 0 || size(df, 1) == 0
```
nalimilan: Do you actually need to check `size(df, 2) == 0`? If there are no columns, there can't be any rows, so the second condition will hold already (and I think `size` already has a special case for that which returns 0).

rofinn: @nalimilan Currently, I don't think there is a a way to construct a dataframe where there are rows without columns, but I could imagine several ways that assumption might be broken in the future. I personally like that this is very explicit and makes as few assumption about the dataframe. I'm not aware of any special cases for `size`, is this a special case for Dataframes specifically?

rofinn: FWIW, base julia does this by checking that the length (# of element) equals 0, but that isn't a reasonable option given the debate around what `length(df)` should return ( #1200).

nalimilan: See the definition of `nrows`, which is used by `size`:
```julia
nrow(df::DataFrame) = ncol(df) > 0 ? length(df.columns[1])::Int : 0
```
So you don't need to check the number of columns again.

rofinn: Again, while I agree that the extra col check in `isempty` currently isn't needed, I'd like to minimize the number of assumptions being made about the structure of the dataframe moving forward.

rofinn: I could see an argument for reversing the order though if we want to short circuit on nrows? That would currently bypass the extra check while remaining explicit and less dependent on the underlying structure.

rofinn: FWIW, the addition of the extra col check seems fairly minimal compared to checking nrows.
```julia
julia> @benchmark size($df, 2) == 0
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.684 ns (0.00% GC)
  median time:      6.690 ns (0.00% GC)
  mean time:        6.896 ns (0.00% GC)
  maximum time:     29.170 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark size($df, 1) == 0
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     39.164 ns (0.00% GC)
  median time:      39.681 ns (0.00% GC)
  mean time:        40.402 ns (0.00% GC)
  maximum time:     68.512 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     991

julia> @benchmark size($df, 1) == 0 || size($df, 2) == 0
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     44.666 ns (0.00% GC)
  median time:      47.513 ns (0.00% GC)
  mean time:        48.973 ns (0.00% GC)
  maximum time:     68.798 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     988
```

nalimilan: These are probably going to be inlined and optimized out anyway, I was just thinking  in terms of consistency of the code. Honestly, I don't see how in the future a data frame could be allowed to have no columns and yet contain multiple rows, but if you're afraid of that, why not...

PR #1233, opened 2017-09-18
Fixes for 0.7

quinnj:


test/utils.jl
```diff
@@ -159,7 +159,8 @@ module TestUtils
                            CategoricalArray{Union{Int, Null}}(4:7)],
                        [:arr, :nullarr, :cat, :nullcat])
         describe(io, df)
-        @test String(take!(io)) ==
+        # Julia 0.7
```
cjprybol: Is this diff accurate? I feel like I've seen this same change pattern on multiple PRs since the backport. Excluding the déjà vu, this looks good to me.

nalimilan: But `nullfirst` and `nullsecond` aren't used anywhere, right? BTW, if would be cool to wrap this inside `if VERSION < ...` blocks so that Femtocleaner will remove the < 0.7 part automatically when we drop support for it.

quinnj: fixed.

nalimilan: Actually, I just realized it would be much simpler to do `$(Union{Int, Null})` inside the string block, wouldn't it?

PR #1237, opened 2017-09-19
Improve show in corner cases

bkamins:
Incrementally improving `show` for `DataFrames`.
In this PR I propose to indicate `null` with a star. This way one can distinguish `null` from `"null"` and `:null` as can be seen in test cases.
Additionally as `*` is put in an unused visual column it helps to spot null values in a data frame.

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -253,9 +253,10 @@ function showrowindices(io::IO,
         for _ in 1:padding
             write(io, ' ')
         end
-        print(io, " │ ")
+        print(io, " │")
         # Print DataFrame entry
         for j in leftcol:rightcol
+            print(io, isnull(df[i,j]) ? "*" : " ")
             strlen = 0
             try
                 strlen = ourstrwidth(df[i, j])
@@ -275,7 +276,7 @@ function showrowindices(io::IO,
                     print(io, " │\n")
                 end
             else
-                print(io, " │ ")
+                print(io, " │")
             end
         end
     end
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -57,8 +57,6 @@ end
 ourshowcompact(io::IO, x::Any) = showcompact(io, x) # -> Void
 ourshowcompact(io::IO, x::AbstractString) = escape_string(io, x, "") # -> Void
 ourshowcompact(io::IO, x::Symbol) = ourshowcompact(io, string(x)) # -> Void
-ourshowcompact(io::IO, x::CategoricalValue{<:AbstractString}) =
-    ourshowcompact(io, String(x)) # -> Void
 
 #' @description
 #'
@@ -251,7 +249,7 @@ function showrowindices(io::IO,
                 s = df[i, j]
                 strlen = ourstrwidth(s)
                 if isnull(s)
-                    print_with_color(:light_yellow, io, s)
+                    print_with_color(:light_black, io, s)
                 else
                     ourshowcompact(io, s)
                 end
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -35,14 +35,8 @@ let
     function ourstrwidth(x::Any) # -> Int
         truncate(io, 0)
         ourshowcompact(io, x)
-        return position(io)
+        strwidth(String(take!(io)))
     end
-    ourstrwidth(x::AbstractString) = strwidth(x) # -> Int
-    ourstrwidth(s::Symbol) =
-        Int(ccall(:u8_strwidth,
-                  Csize_t,
-                  (Ptr{UInt8}, ),
-                  Base.unsafe_convert(Ptr{UInt8}, s)))
 end
 
 #' @description
@@ -61,10 +55,10 @@ end
 #' ourshowcompact(STDOUT, "abc")
 #' ourshowcompact(STDOUT, 10000)
 ourshowcompact(io::IO, x::Any) = showcompact(io, x) # -> Void
-ourshowcompact(io::IO, x::AbstractString) = print(io, x) # -> Void
-ourshowcompact(io::IO, x::Symbol) = print(io, x) # -> Void
+ourshowcompact(io::IO, x::AbstractString) = escape_string(io, x, "") # -> Void
+ourshowcompact(io::IO, x::Symbol) = ourshowcompact(io, string(x)) # -> Void
 ourshowcompact(io::IO, x::CategoricalValue{<:AbstractString}) =
-    print(io, String(x)) # -> Void
+    ourshowcompact(io, string(x)) # -> Void
 
 #' @description
 #'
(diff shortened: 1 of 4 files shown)


cjprybol:
Thanks for the proposal!

+1 for disambiguation, but -1 on the `*`. I think this has been talked about before on several issues and it's commonly regarded as annoying to have quotes around strings and colons before symbols in the DataFrames show output due to the extra screen space it takes up, but if we need to disambiguate then I think it makes less sense to introduce a novel disambiguation syntax than it does to just show the values in the same way base Julia shows them.

```julia
julia> null
null

julia> "null"
"null"

julia> :null
:null

julia> DataFrame(x = null, y = "null", z = :null)
1×3 DataFrames.DataFrame
│ Row │ x    │ y    │ z    │
├─────┼──────┼──────┼──────┤
│ 1   │ null │ null │ null │

# why not
1×3 DataFrames.DataFrame
│ Row │ x    │ y      │ z     │
├─────┼──────┼────────┼───────┤
│ 1   │ null │ "null" │ :null │
```

Adding type declarations at the top of the column to help disambiguate values of different types that were printed the same was also proposed ([see examples from Tibbles](https://cran.r-project.org/web/packages/tibble/vignettes/tibble.html)) but since users can now use whatever element types they want in columns, I can imagine that non-base types with long names will quickly eat up screen real estate more quickly than the quotes and colons would.

To help with quick visual recognition of nulls, has anyone tried printing nulls with `print_with_color`? A column full of bright red `null`s would very quickly allow recognition of missing values.

At least that's my 2¢

nalimilan:
I don't really like adding stars, it's not obvious that it means "missing value". @cjprybol's suggestion of using a different color sounds better (though I'd rather use grey than red for nulls). Or if we're really afraid of ambiguities, maybe adding quotes around strings would be OK.

bkamins:
Thank you. Adding color actually required fixing some more issues - so now the PR is a bit more complex.
What it changes:
* prints `null` in light yellow (I am open to change it :) - it simply looked best on my terminal)
* makes characters like `\n`, `\r` be printed as escaped strings (and not ruining display); the cost is that `\` is also escaped;
* make `ourshowcompact` better calculate printed witdth in corner cases
* in `ourstringwidth` cast `Symbol` and `CategoricalValue` to string (instead of custom printing method)

bkamins:
Sorry - I have messed up one thing on push. I am fixing it now

nalimilan:
Hmm, light yellow is really flashy (at least in my terminal). Nulls should be almost hidden, not highlighted. What about one of the available greys? Not sure how consistent is their look across terminals, but color `8` doesn't look too bad (there are other choices, see `foreach(x->print_with_color(x, "null"), 1:256)`).

src/abstractdataframe/show.jl
```diff
@@ -61,10 +55,10 @@ end
 #' ourshowcompact(STDOUT, "abc")
 #' ourshowcompact(STDOUT, 10000)
 ourshowcompact(io::IO, x::Any) = showcompact(io, x) # -> Void
-ourshowcompact(io::IO, x::AbstractString) = print(io, x) # -> Void
-ourshowcompact(io::IO, x::Symbol) = print(io, x) # -> Void
+ourshowcompact(io::IO, x::AbstractString) = escape_string(io, x, "") # -> Void
+ourshowcompact(io::IO, x::Symbol) = ourshowcompact(io, string(x)) # -> Void
 ourshowcompact(io::IO, x::CategoricalValue{<:AbstractString}) =
```
nalimilan: `CategoricalValue <: AbstractString` in the current master, so I could as well tag a release so that we can depend on it and remove this method.

bkamins:
Of greys `:light_black` looks best on my terminal.

src/abstractdataframe/show.jl
nalimilan: See https://github.com/JuliaLang/METADATA.jl/pull/11290.

test/show.jl
```diff
@@ -285,6 +290,38 @@ module TestShow
     │ 1     │ Fish │ String                     │ 0       │ Suzy  …  Amir │
     │ 2     │ Mass │ Union{Float64, Nulls.Null} │ 1       │ 1.5  …  null  │"""
 
+    # Test showing null
+    df = DataFrame(A=[:Symbol, null, :null],
```
nalimilan: Spaces around `=`.

src/abstractdataframe/show.jl
bkamins: I understand we wait till CategoricalValue is merged? I have implemented what you have requested but the changes will fail on CI now.

nalimilan: Yes, hopefully that won't take too much time.

nalimilan: Release has been tagged now.

nalimilan: Woops, making `CategoricalValue{T} <: AbstractString` creates issues when `T` is not an `AbstractString` (e.g. here it's a `Char`). The problem is that dispatch uses a method which works only for strings, and it fails later because the data isn't actually a string. We could fix this particular case, but this sounds like a fragile approach in general. Maybe we'd better restrict `T <: AbstractString` for `CategoricalArray{T}`. Let's discuss this at https://github.com/JuliaData/CategoricalArrays.jl/pull/77.

nalimilan: The easiest solution for now is probably to stop creating a `CategoricalArray{Char}` and use a standard array instead.

bkamins: fixed

bkamins:
@nalimilan I have properly merged the change now.

cjprybol:
Anyone have any ideas what the nightly errors are? [link](https://ci.appveyor.com/project/nalimilan/dataframes-jl/build/1.0.428/job/idv2icg2hl0emafm#L1949)

```
Unreachable reached at 26EBC55E
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ILLEGAL_INSTRUCTION at 0x26ebc55e -- 
```
Once we figure that out let's merge, this looks great!

nalimilan:
I don't think we should worry about 0.7 right now, there are known failure related to `Union{T, Null}` (@quinnj can confirm).

bkamins:
This looks similar to https://github.com/JuliaLang/julia/issues/23675 (the same error and also related to type inference) and it should have been already fixed on master.
Could you please trigger CI again once Julia nightly build is updated to incorporate the relevant commit? If the problem persists probably an additional issue to Julialang should be submitted.

bkamins:
Travis is still not updated with the commit (nightly is from 2017-09-13).

Appveyor produces error:
```
Exception: EXCEPTION_ACCESS_VIOLATION at 0x0 -- unknown function (ip: 0000000000000000)
```
which seems caused by some recursion (the log is almost 6MB) that is unrelated.

I push a small fix to the `show.jl` code to remove deprecation warning on 0.7.

PR #1239, opened 2017-09-20
Add allocate(::Type{Null})

alyst:
Otherwise `allocate(::Type{T}, rows, ref) where {T <: Union{WeakRefString, Null}}` method gets called resulting in an error.


nalimilan:
Is there a way to test this?

alyst:
Originally I've encountered it while trying to `CSV.read()` a table with all-nulls columns. It fails in `DataFrame(schema)` call that should create a `DataFrameStream()` object.
That's what I've added to the tests. But since this API is new to me, I just don't know whether it's a valid code pattern or the user code triggering this path should look different.

PR #1241, opened 2017-09-28
Adapt to new == behavior with nulls

nalimilan:
With Nulls 0.0.7, `1 == null` and `null == null` return `null`. Prepare
tests to this new behavior.

### DIFF
--- a/test/cat.jl
+++ b/test/cat.jl
@@ -1,5 +1,6 @@
 module TestCat
     using Base.Test, DataFrames
+    const ≅ = isequal
 
     #
     # hcat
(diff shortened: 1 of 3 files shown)


nalimilan:
CI failures are only on nightlies and unrelated.

cjprybol:
Are we going to be promoting `≅` for use generally as a replacement for `isequal` or is that just to make managing/updating the internal tests easier?

nalimilan:
There have been discussions about introducing a new operator in Julia itself (in particular ≅), but for now it's just a local convention to make the code easier to read.

Issue #1240, opened 2017-09-21
Formally deprecate readtable


randyzwitch:
I just got caught in an upgrade to DataFrames master where `readtable` just becomes undefined, where it works in DataFrames 0.10.1. 

Per the issue below, it appears that it's moving to CSV.jl, but it feels like a deprecation warning is needed, I have to imagine this function was used quite a bit.

https://github.com/JuliaData/DataFrames.jl/issues/1133

Example, works on 0.10.1:

```julia
julia> using DataFrames, HTTP
julia> f = HTTP.get("http://goo.gl/UDv12g")
julia> satData = readtable(f.body)
```
Doesn't work on master

```julia
julia> satData = readtable(f.body)
ERROR: UndefVarError: readtable not defined
```

nalimilan:
The problem is that CSV depends on DataFrames (since it uses it as its default table type), so DataFrames cannot depend on CSV. We need to first tag a release of the Nulls-base DataFrames which doesn't depend on CSV, then we will be able to tag a version of the Nulls-base CSV which depends on (the Nulls-base) DataFrames, and finally we will be able to tag a release of DataFrames which depends on CSV. We could do that in the same day so that users do not notice.

randyzwitch:
Ok, just bringing it up here in case it was unexpected. But it sounds like there's a plan, so no problem :)

nalimilan:
See https://github.com/JuliaData/DataFrames.jl/issues/1240.

PR #1243, opened 2017-10-01
Reintroduce readtable(), writetable() and string-literal macros, keeping them deprecated

nalimilan:
This will ease the transition to CSV, without introducing a dependency on it
and without breakage due to behavior changes.

I tested this locally, but it's not covered by the test suite.

Fixes https://github.com/JuliaData/DataFrames.jl/issues/1240.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -51,7 +51,7 @@ writetable(filename, df, [keyword options])
 df = DataFrame(A = 1:10)
 writetable("output.csv", df)
 writetable("output.dat", df, separator = ',', header = false)
-writetable("output.dat", df, quotemark = '\', separator = ',')
+writetable("output.dat", df, quotemark = '\'', separator = ',')
 writetable("output.dat", df, header = false)
 ```
 """
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -17,3 +17,1252 @@ import Base: keys, values, insert!
 @deprecate complete_cases completecases
 
 @deprecate sub(df::AbstractDataFrame, rows) view(df, rows)
+
+
+## write.table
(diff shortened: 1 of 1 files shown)


src/deprecated.jl
```diff
@@ -17,3 +17,1252 @@ import Base: keys, values, insert!
...
+
+* `::DataFrame`
+
+### Examples
+
+```julia
+df = DataFrame(A = 1:10)
+writetable("output.csv", df)
+writetable("output.dat", df, separator = ',', header = false)
+writetable("output.dat", df, quotemark = '\', separator = ',')
```
cjprybol: I think quotemark here needs to be escaped, `quotemark = '\'` -> `quotemark = '\\'`, although I couldn't find any errors in the CI to confirm that

cjprybol:
This is just copied and pasted from earlier source code, right? If so, then I don't think we should worry about adding the tests back as well.

I don't love the idea of pointing people towards CSV while https://github.com/JuliaData/CSV.jl/issues/109 is unresolved.

nalimilan:
> This is just copied and pasted from earlier source code, right? If so, then I don't think we should worry about adding the tests back as well.

Yeah, that's copy/pasted, except for the place where `DataArray` columns where created (but changes were very limited). 

> I don't love the idea of pointing people towards CSV while JuliaData/CSV.jl#109 is unresolved.

We're not going to tag a release immediately anyway, so that leaves time to fix that nasty bug.

src/deprecated.jl
nalimilan: Indeed, it must have been there for a very long time. Fixed (I think the correct value is `'\''`).

Issue #1093, opened 2016-10-06
Update docstrings to new conventions


nalimilan:
We should improve our docstrings by adapting them to new conventions [from the Julia manual](http://docs.julialang.org/en/latest/manual/documentation/). In particular (not all of these steps need to be done at the same time, though):
- give the signature first with a four-space indent
- merge the `Details` section with the general description (in a separate paragraph if necessary)
- only use an `# Arguments` section when there are many of them; mention the meaning of arguments in the description for simple cases
- only use a `# Result` section when really useful
- transform examples into doctests
- document return types as `::T` in the signature only when useful, and remove `# -> T` comments in the code

bkamins:
@nalimilan - I think it is done (or I have missed something?)

nalimilan:
Maybe some details could still be cleaned up, but I guess you did most of it, yes!

PR #1246, opened 2017-10-09
Modify aggregate for efficiency and decide its future

cjprybol:
Bypassing map and combine for aggregate speeds up code and reduces
memory allocations by ~1-2 orders of magnitude. The `By` function
still uses map and combine to allow more complex anonymous functions
that can return DataFrames and do blocks. Functions for naming new
columns were inlined, and their Julia 0.4 Compat removed. Added tests.

Replaces https://github.com/JuliaData/DataTables.jl/pull/65. Should be re-benchmarked, and the benchmarks should be made permanent with PkgBenchmark.jl. [We should try using master](https://discourse.julialang.org/t/a-living-post-of-julia-vs-rs-data-manipulation-tasks-speeds/6316/16)

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -320,9 +320,6 @@ aggregate(gd::GroupedDataFrame, fs)
 * `fs` : a function or vector of functions to be applied to vectors
   within groups; expects each argument to be a column vector
 
-Each `fs` should return a value or vector. All returns must be the
-same length.
-
 ### Returns
 
 * `::DataFrame`
@@ -342,16 +339,23 @@ df |> groupby(:a) |> [sum, x->mean(Nulls.skip(x))]   # equivalent
 """
 aggregate(d::AbstractDataFrame, fs::Function; sort::Bool=false) = aggregate(d, [fs], sort=sort)
 function aggregate(d::AbstractDataFrame, fs::Vector{T}; sort::Bool=false) where T<:Function
-    headers = _makeheaders(fs, _names(d))
-    _aggregate(d, fs, headers, sort)
+    headers = [Symbol(c, "_", f) for f in fs for c in names(d)]
+    res = DataFrame(Any[f(d[c]) for f in fs for c in names(d)], headers)
+    sort && sort!(res)
+    res
 end
 
 # Applies aggregate to non-key cols of each SubDataFrame of a GroupedDataFrame
 aggregate(gd::GroupedDataFrame, f::Function; sort::Bool=false) = aggregate(gd, [f], sort=sort)
 function aggregate(gd::GroupedDataFrame, fs::Vector{T}; sort::Bool=false) where T<:Function
-    headers = _makeheaders(fs, setdiff(_names(gd), gd.cols))
-    res = combine(map(x -> _aggregate(without(x, gd.cols), fs, headers), gd))
-    sort && sort!(res, cols=headers)
+    res = gd.parent[gd.idx[gd.starts], gd.cols]
+    cols = setdiff(names(gd.parent), gd.cols)
+    for f in fs
+        for c in cols
+            res[Symbol(c, "_", f)] = [f(g[c]) for g in gd]
+        end
+    end
+    sort && sort!(res)
     res
 end
 
@@ -364,15 +368,4 @@ function aggregate(d::AbstractDataFrame,
                    fs::Union{T, Vector{T}};
                    sort::Bool=false) where {S<:ColumnIndex, T <:Function}
     aggregate(groupby(d, cols, sort=sort), fs)
-end
-
-function _makeheaders(fs::Vector{T}, cn::Vector{Symbol}) where T<:Function
-    fnames = _fnames(fs) # see other/utils.jl
-    [Symbol(colname,'_',fname) for fname in fnames for colname in cn]
-end
-
-function _aggregate(d::AbstractDataFrame, fs::Vector{T}, headers::Vector{Symbol}, sort::Bool=false) where T<:Function
-    res = DataFrame(Any[vcat(f(d[i])) for f in fs for i in 1:size(d, 2)], headers)
-    sort && sort!(res, cols=headers)
-    res
 end
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -141,18 +141,3 @@ function countnull(a::CategoricalArray)
     end
     return res
 end
-
-# Gets the name of a function. Used in groupeDataFrame/grouping.jl
-function _fnames(fs::Vector{T}) where T<:Function
-    λcounter = 0
-    names = map(fs) do f
-        name = string(f)
-        if name == "(anonymous function)" # Anonymous functions with Julia < 0.5
-            λcounter += 1
-            name = "λ$(λcounter)"
-        end
-        name
-    end
-    names
-end
-
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -181,4 +181,40 @@ module TestGrouping
     @test gd[2] == DataFrame(Key1="A", Key2="B", Value=2)
     @test gd[3] == DataFrame(Key1="B", Key2="A", Value=3)
     @test gd[4] == DataFrame(Key1="B", Key2="B", Value=4)
+
+    @testset "aggregate" begin
+        # test converting functions to valid column names
+        @test Symbol.([mean, sum]) == [:mean, :sum]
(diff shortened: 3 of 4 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -342,16 +339,23 @@ df |> groupby(:a) |> [sum, x->mean(Nulls.skip(x))]   # equivalent
...
+    res = DataFrame(Any[f(d[c]) for f in fs for c in names(d)], headers)
+    sort && sort!(res)
+    res
 end
 
 # Applies aggregate to non-key cols of each SubDataFrame of a GroupedDataFrame
 aggregate(gd::GroupedDataFrame, f::Function; sort::Bool=false) = aggregate(gd, [f], sort=sort)
 function aggregate(gd::GroupedDataFrame, fs::Vector{T}; sort::Bool=false) where T<:Function
-    headers = _makeheaders(fs, setdiff(_names(gd), gd.cols))
-    res = combine(map(x -> _aggregate(without(x, gd.cols), fs, headers), gd))
```
nalimilan: Actually, I missed this in the first review, but the current code indeed allows returning a vector for each group, and uses `combine` to turn the result into a single vector. What's annoying is that it's going to slow down everything, but maybe we can make `combine` efficient when the returned value is a scalar (maybe via inference?)?

We could also imagine having a different function for non-scalar operations. I guess we should check what Pandas and dplyr do.

cjprybol: [You caught it in the first review too](https://github.com/JuliaData/DataTables.jl/pull/65#discussion_r118842011), but your proposed solution was simpler last time : ). Checking how this is handled elsewhere is a good idea

nalimilan: Yeah, but I think I was wrong in my previous review, since I hadn't noticed that call to `combine`. (AFAICT).

nalimilan: dplyr offers `summarize`, which only allows functions to return a single value, and errors otherwise. Maybe we should provide the same function for simple cases like that. Currently people use `groupby` or `aggregate`, which are powerful but slow (slow because powerful?).

Or, as an interesting Julian challenge, we could try using inference and see whether it can allow us to find out whether a function is going to return a scalar. In that case we could use a fast path.

nalimilan: Maybe we should drop support for returning vectors. That's what Pandas does, see https://discourse.julialang.org/t/stack-overflow-in-dataframes-group-by/6357/8.

cjprybol: Thanks for looking into this and discussing it with ExpandingMan! Offering multiple functions of varying levels of capability/efficiency sounds like the most straightforward way to support all use cases and keep everyone happy in terms of performance. I'd be happy to clarify the distinctions between the functions as part of [Doctoberfest](https://github.com/JuliaData/DataFrames.jl/issues/1093) to make sure users understand how to use them effectively and to which use-case each applies.

nalimilan: Nevertheless, I think we should investigate whether inference could allow supporting vector results efficiently. The less different functions we need, the easier our API will be, and it would be nice to be more flexible than Pandas while still choosing the most efficient approach automatically.

Since `aggregate` works column-wise, we can use inference, so in theory it would be possible to take a fast path when we detect the function returns a scalar for all columns. We only need to check this once, since the type of the columns is the same across groups. If inference fails it's fine to go back to the current slow approach.

quinnj:
I don't imagine this is relevant anymore, given all the work that has gone into performance in other efforts. Good to close @nalimilan?

nalimilan:
`aggregate` doesn't take advantage of the new efficient methods AFAICT since it uses the type-unstable method. The approach from this PR might still be faster anyway since you can just iterate over columns. (BTW, I'd also like to deprecate `aggregate` in favor of something like `by(df, :key, AllCols() .=> f)`.)

bkamins:
@nalimilan - do you have an opinion what we should do about this PR?

nalimilan:
I think we should see how we can deprecate `aggregate` in favor of `by(df, :key, AllCols() .=> f)`, and then we'll see whether a separate implementation from `combine` is needed.

bkamins:
Working on this I have encountered the following error, that makes me uneasy:
```
julia> using DataFrames

julia> df = DataFrame(rand(3,4))
3×4 DataFrame
│     │ Float64  │ Float64  │ Float64  │ Float64  │
│ 1   │ 0.651508 │ 0.231298 │ 0.574306 │ 0.941665 │
│ 2   │ 0.168641 │ 0.25154  │ 0.452554 │ 0.811189 │
│ 3   │ 0.198076 │ 0.637567 │ 0.973212 │ 0.484695 │

julia> by(df, :x1, names(df) .=> sin)
Internal error: encountered unexpected error in runtime:
MethodError(f=typeof(Base.string)(), args=(Expr(:<:, :t, :r),), world=0x0000000000000eec)
```
there seems to be some problem with the internal design of `by` (I understand the reason for the error, but stack trace indicates that something bad is going on there).

If we pass a function that accepts vectors all is OK:
```
julia> by(df, :x1, names(df) .=> sum)
3×5 DataFrame
│ Row │ x1       │ x1_sum   │ x2_sum   │ x3_sum   │ x4_sum   │
│     │ Float64  │ Float64  │ Float64  │ Float64  │ Float64  │
├─────┼──────────┼──────────┼──────────┼──────────┼──────────┤
│ 1   │ 0.651508 │ 0.651508 │ 0.231298 │ 0.574306 │ 0.941665 │
│ 2   │ 0.168641 │ 0.168641 │ 0.25154  │ 0.452554 │ 0.811189 │
│ 3   │ 0.198076 │ 0.198076 │ 0.637567 │ 0.973212 │ 0.484695 │

julia> by(df, :x1, names(df) .=> x -> sin.(x))
3×5 DataFrame
│ Row │ x1       │ x1_function │ x2_function │ x3_function │ x4_function │
│     │ Float64  │ Float64     │ Float64     │ Float64     │ Float64     │
├─────┼──────────┼─────────────┼─────────────┼─────────────┼─────────────┤
│ 1   │ 0.651508 │ 0.606386    │ 0.229241    │ 0.543252    │ 0.808539    │
│ 2   │ 0.168641 │ 0.167843    │ 0.248896    │ 0.437264    │ 0.725107    │
│ 3   │ 0.198076 │ 0.196783    │ 0.595242    │ 0.826697    │ 0.465938    │
```
(so it seems that we essentially have it already)

nalimilan:
Interesting. With a recent Julia master, I get a proper `MethodError`, so it looks like it's been fixed recently.

Regarding the design, you're right that using `names(df)` we could already deprecate `aggregate` without even changing `by`. Then we could add support for `All` later. Let's do that?

bkamins:
Yes - adding `All` support can be for later.
And yes - we can drop `aggregate` (let us just make sure that we cover all possibilities of using `aggregate`) since we are in a clean-up mode.

In general (now this is opinion - not a recommendation 😄) - I do not find `aggregate` very useful anyway, as it does not give you control over column names so it is mostly interactive use function.

bkamins:
@nalimilan - what is your state of thinking about the future of `aggregate` 😄.
I think we should make this decision before 1.0.

nalimilan:
I'll have a look. `All` is waiting on https://github.com/JuliaData/DataAPI.jl/pull/10 but we could use `names(df)` until then.

bkamins:
Thank you.

With https://github.com/JuliaData/DataAPI.jl/pull/10 I think everyone was on board with the change (is there something that is stopping that PR)?

nalimilan:
I just wanted to get Matt's input, as the InvertedIndices author and broadcasting expert.

bkamins:
We should decide what to do with `aggregate` before 1.0 release.
For this I understand the crucial thing is `Not` broadcasting.
@nalimilan - I understand you do not have any feedback on it yet?

nalimilan:
Unfortunately not. Maybe we could go ahead anyway, in the worst case people could do `All(Not(...)) .=>`.

bkamins:
I have opened https://github.com/mbauman/InvertedIndices.jl/issues/15.

I think then let us go with `All(Not(...` and we can just simplify things later if `Not` starts supporting broadcasting.

bkamins:
I am closing this, as even if we want to work on `aggregate` in the future this PR would have to be completely rewritten.

bkamins:
Of course please reopen if you feel I am wrong.

PR #1252, opened 2017-10-11
Update documentation

cjprybol:
Goals:

- [x] Add missing exported functions and types to online documentation
- [x] Move documentation examples to doctests
    - [x] Make sure the broken examples get fixed before merging


README.md
```diff
@@ -2,7 +2,7 @@ DataFrames.jl
 =============
 
 [![0.6](http://pkg.julialang.org/badges/DataFrames_0.6.svg)](http://pkg.julialang.org/?pkg=DataFrames)
-[![0.7](http://pkg.julialang.org/badges/DataFrames_0.7.svg)](http://pkg.julialang.org/?pkg=DataFrames)
+<!-- [![0.7](http://pkg.julialang.org/badges/DataFrames_0.7.svg)](http://pkg.julialang.org/?pkg=DataFrames) -->
```
nalimilan: Why change this?

docs/src/index.md
```diff
@@ -1,21 +1,40 @@
-# DataFrames Documentation Outline
+# DataFrames.jl
+
+Welcome to the DataFrames documentation! This resource aims to teach you everything you need
+to know to get up and running with tabular data manipulation using the DataFrames.jl package
+and the Julia language. If there is something you expect DataFrames to be capable of, but
+cannot figure out how to do, please reach out with questions in Domains/Data on
+[Discourse](https://discourse.julialang.org/new-topic?title=[DataFrames%20Question]:%20&body=%23%20Question:%0A%0A%23%20Dataset%20(if%20applicable):%0A%0A%23%20MWE%20(if%20applicable):%0A&category=Domains/Data&tags=question).
```
nalimilan: "MWE" could be spelled in full.

docs/src/index.md
```diff
@@ -1,21 +1,40 @@
...
+You can also give feedback and suggest features or improvements by
```
nalimilan: "give feedback" is too general, better say "report bugs". Also "suggest features" is likely to attract many little useful posts which we will have to close, creating frustration on both sides. Better leave people discuss their problems on Discourse.

docs/src/lib/functions.md
```diff
@@ -0,0 +1,86 @@
+```@meta
+CurrentModule = DataFrames
+```
+
+# Functions
```
nalimilan: Isn't everything listed here a function?

docs/src/lib/functions.md
```diff
@@ -0,0 +1,86 @@
...
+
+```@docs
+melt
+stack
+unstack
+stackdf
+meltdf
+```
+
+## Data Manipulation
```
nalimilan: Maybe "Split-Apply-Combine"? Though `colwise` is different.

docs/src/lib/functions.md
```diff
@@ -0,0 +1,86 @@
...
+
+```@docs
+aggregate
+by
+colwise
+groupby
+```
+
+## DataFrame Manipulation
```
nalimilan: `combine` should be in the previous section. I'm not sure how to describe this set of functions, but the current title isn't descriptive enough IMHO. I find it hard to find meaningful groupings for all functions.

docs/src/lib/functions.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+## Iteration
+
+```@docs
+eachcol
+eachrow
+```
+
+## Descriptive
```
nalimilan: `completecases` and `nonunique` shouldn't be listed here.

docs/src/lib/functions.md
```diff
@@ -0,0 +1,86 @@
...
+ncol
+nrow
+showcols
+size
+```
+
+## IO
+
+```@docs
+head
```
nalimilan: `head` and `tail` are not I/O. `printtable` should probably not be exported. `show` should be with `showcols`.

docs/src/lib/functions.md
```diff
@@ -0,0 +1,86 @@
...
+printtable
+tail
+show
+```
+
+## Sorting and Ordering
+
+```@docs
+sort
```
nalimilan: We have `sort!`, right?

docs/src/lib/functions.md
```diff
@@ -0,0 +1,86 @@
...
+eachrow
+```
+
+## Descriptive
+
+```@docs
+completecases
+describe
+eltypes
+names
```
nalimilan: Should be with `names!`.

docs/src/man/categorical.md
```diff
@@ -2,52 +2,176 @@
...
 ```
 
 In addition to representing repeated data efficiently, the `CategoricalArray` type allows us to determine efficiently the allowed levels of the variable at any time using the `levels` function (note that levels may or may not be actually used in the data):
 
-```julia
-levels(cv)
+```jldoctest
+julia> using CategoricalArrays, Nulls
+
+julia> cv = CategoricalArray(["Group A", null, "Group A",
```
nalimilan: If you give a name to the previous test and use the same here, you don't need to repeat these commands.

docs/src/man/getting_started.md
```diff
@@ -13,7 +13,7 @@ Throughout the rest of this tutorial, we will assume that you have installed the
 
 To get started, let's examine the `Null` type. `Null` is a type implemented by the [Nulls.jl](https://github.com/JuliaData/Nulls.jl) package to represent missing data. `null` is an instance of the type `Null` used to represent a missing value.
 
-```julia
+```jldocttest
```
nalimilan: `jldoctest`. Same below.

docs/src/man/getting_started.md
```diff
@@ -111,61 +123,202 @@ julia> nulls(Int, 1, 3)
...
+julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
+4×2 DataFrames.DataFrame
+│ Row │ A │ B │
+├─────┼───┼───┤
+│ 1   │ 1 │ M │
+│ 2   │ 2 │ F │
+│ 3   │ 3 │ F │
+│ 4   │ 4 │ M │
+
+julia> df == DataFrame(:A => 1:4, :B => ["M", "F", "F", "M"])
```
nalimilan: Better just create the data frame and let the user see it gives the same result (no need to define `df` in either of these lines).

docs/src/man/getting_started.md
```diff
@@ -111,61 +123,202 @@ julia> nulls(Int, 1, 3)
...
+│ 5   │ 5 │ F │
+│ 6   │ 6 │ M │
+│ 7   │ 7 │ M │
+│ 8   │ 8 │ F │
+
 ```
 
-The `DataFrame` we build in this way has 8 rows and 2 columns. You can check this using `size` function:
+The `DataFrame` we build in this way has 8 rows and 2 columns. You can check this using the
+`size`, `nrow`, and `ncol` functions:
```
nalimilan: Better not mention `nrow` and `ncol` as there have been lots of discussions about removing these.

docs/src/man/getting_started.md
```diff
@@ -111,61 +123,202 @@ julia> nulls(Int, 1, 3)
...
+julia> median(df[:A])
+4.5
+
 ```
 
 We can also apply a function to each column of a `DataFrame` with the `colwise` function. For example:
 
-```julia
-df = DataFrame(A = 1:4, B = randn(4))
-colwise(c->cumsum(Nulls.skip(c)), df)
```
nalimilan: That example was quite weird: better apply a summary statistic like the mean or sum.

docs/src/man/joins.md
```diff
@@ -2,23 +2,42 @@
...
+julia> names = DataFrame(ID = [20, 40], Name = ["John Doe", "Jane Doe"]);
 
-Output:
+julia> jobs = DataFrame(ID = [20, 40], Job = ["Lawyer", "Doctor"]);
 
-| Row | ID | Name       | Job      |
-|-----|----|------------|----------|
-| 1   | 20 | "John Doe" | "Lawyer" |
-| 2   | 40 | "Jane Doe" | "Doctor" |
+julia> joined = join(names, jobs, on = :ID)
```
nalimilan: No need to define `joined`.

docs/src/man/joins.md
```diff
@@ -34,41 +53,171 @@ There are seven kinds of joins supported by the DataFrames package:
 
 You can control the kind of join that `join` performs using the `kind` keyword argument:
```
nalimilan: Since the output is now longer, it would be useful to list the supported values before illustrating them. Maybe link to a description of the joins terminology (e.g. on Wikipedia) if that's not already the case.

docs/src/man/querying_frameworks.md
```diff
@@ -12,59 +12,111 @@ A query is started with the `@from` macro and consists of a series of query comm
 
 A simple example of a query looks like this:
 
-```@setup 1
```
nalimilan: This kind of annotation is precisely what allows not repeating `using` in every example. :-)

docs/src/man/querying_frameworks.md
```diff
@@ -12,59 +12,111 @@ A query is started with the `@from` macro and consists of a series of query comm
...
+│ Row │ name  │ age  │ children │
+├─────┼───────┼──────┼──────────┤
+│ 1   │ John  │ 54.0 │ 0        │
+│ 2   │ Sally │ 34.0 │ 2        │
+│ 3   │ Roger │ 79.0 │ 4        │
+
+julia> q2 = @from i in df begin
+            @where i.age > 40
+            @select {number_of_children=i.children, i.name}
+       end;
```
nalimilan: Is it so ugly that we can't print it? That would be clearer, even if it's not really useful.

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -2,18 +2,91 @@
...
-using CSV
-iris = CSV.read(joinpath(Pkg.dir("DataFrames"), "test/data/iris.csv"), DataFrame)
-iris[:id] = 1:size(iris, 1)  # this makes it easier to unstack
-d = stack(iris, 1:4)
+```jldoctest
+julia> using DataFrames, CSV
+
+julia> iris = CSV.read(joinpath(Pkg.dir("DataFrames"), "test/data/iris.csv"));
+
+julia> head(iris)
```
nalimilan: If we just call `iris`, don't we get the first and last rows already?

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -22,41 +95,187 @@ The stacked DataFrame that results includes all of the columns not specified to
...
 
 `melt` is an alternative function to reshape from wide to long format. It is based on `stack`, but it prefers specification of the id columns as:
 
-```julia
-d = melt(iris, :Species)
-```
+```jldoctest
+julia> using DataFrames, CSV
 
-All other columns are assumed to be measured variables (they are stacked).
```
nalimilan: Why remove these two sentences?

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -22,41 +95,187 @@ The stacked DataFrame that results includes all of the columns not specified to
...
 
-```julia
-d = stackdf(iris)
+```jldoctest
+julia> using DataFrames, CSV
+
+julia> iris = CSV.read(joinpath(Pkg.dir("DataFrames"), "test/data/iris.csv"));
+
+julia> d = stackdf(iris);
+ERROR: StackOverflowError:
```
nalimilan: :-)

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -72,14 +291,51 @@ This repeats the original columns N times where N is the number of columns stack
...
+│ Row │ variable    │ value │ Species │
+├─────┼─────────────┼───────┼─────────┤
+│ 1   │ SepalLength │ 5.1   │ setosa  │
+│ 2   │ SepalLength │ 4.9   │ setosa  │
+│ 3   │ SepalLength │ 4.7   │ setosa  │
+│ 4   │ SepalLength │ 4.6   │ setosa  │
+│ 5   │ SepalLength │ 5.0   │ setosa  │
+│ 6   │ SepalLength │ 5.4   │ setosa  │
+
+julia> x = by(d, [:variable, :Species], df -> DataFrame(vsum = mean(parse.(Float64, df[:value]))));
```
nalimilan: Add a line break. To work around the CSV.jl parsing bug, probably better go back to latest version for now.

src/other/utils.jl
```diff
@@ -82,18 +82,12 @@ function make_unique(names::Vector{Symbol}; allow_duplicates=true)
...
-#'
-#' @returns names::Vector{Symbol} A vector of standardized column names.
-#'
-#' @examples
-#'
-#' DataFrames.gennames(10)
+"""
+    gennames(n::Integer)
+
+Generate standardized names for columns of a DataFrame. The first name will be :x1, the
```
nalimilan: `` `:x1` ``

src/other/utils.jl
```diff
@@ -103,17 +97,11 @@ function gennames(n::Integer)
...
-#'
-#' @returns count::Int The number of null values in `a`.
-#'
-#' @examples
-#'
-#' DataFrames.countnull([1, 2, 3])
+"""
+    countnull(a::AbstractArray)
+
+Count the number of null values in an array.
```
nalimilan: `` `null` ``

README.md
cjprybol: It doesn't render a badge on the README. I think because v0.7 doesn't have tests on pkg.julialang.org [link](https://pkg.julialang.org/?pkg=DataFrames#DataFrames). I'd also like to change the CI tests to allow failures on nightly until a release candidate comes out for v0.7. 👍  or 👎 ? These can all be moved to another PR too

docs/src/lib/functions.md
cjprybol: Yes, everything on that page is a function, I'm not sure this H1 header adds anything. We could remove it and only show the H2 headers of the function-groups (and bump them up to being H1 headers). But before I act on any of your other comments on where/how functions should be grouped, what are your thoughts on just listing all of the functions alphabetically in a single section? It might make navigation difficult since the functions wouldn't be grouped by concept, but the conceptual groupings are pretty arbitrary and no matter how we group them I'm sure we won't be able to make the groups intuitive to everyone. It will probably be easier to keep these docstrings up to date if we just list the functions in the same way they are exported in the `DataFrames.jl` source file.

docs/src/man/querying_frameworks.md
cjprybol: I'll update these annotations with the syntax that allows testing and preserving state between blocks [Documenter documentation link](https://juliadocs.github.io/Documenter.jl/latest/man/doctests/#Preserving-definitions-between-blocks-1)

docs/src/man/reshaping_and_pivoting.md
cjprybol: Yes, it's just distractingly long. We can probably come up with something better than chaining `head` and `tail` calls though.

cjprybol:
I agree that the other goals can be changed in different PRs, I'll go with that approach rather than turning this into a monolithic behemoth of code churn

docs/src/man/reshaping_and_pivoting.md
nalimilan: Maybe we can find a way to set the `displaysize`?

docs/src/lib/functions.md
nalimilan: Actually, due to the presence of the isolated `meltdf`, I had misread this as a H2 heading. I'm OK with keeping it as H1, `meltdf` should simply be moved.

I'd say grouping is a better approach, but we don't need a lot of groups. Maybe one for basics, and one for grouping/joining & co.?

README.md
nalimilan: Ah, OK. Just remove it then.

cjprybol: Ok, I'm going to remove the change here in favor of a badge-specific PR in https://github.com/JuliaData/DataFrames.jl/pull/1253

docs/src/man/reshaping_and_pivoting.md
cjprybol: I've also removed the example they describe, since it doesn't work and I'm not sure what it was trying to achieve that isn't covered by the other examples

cjprybol:
Ok I think that's most of it. Right now the docstrings aren't rendering for the `types` and `functions` API man pages, but I think that's because not every type/function in those sections have docstrings. That'll be a followup PR

src/DataFrames.jl
```diff
@@ -49,7 +49,6 @@ export AbstractDataFrame,
        nrow,
        nullable!,
        order,
-       printtable,
```
cjprybol: This will stop exporting the `printtable` function as suggested. If we go with that then we need to fix the tests to call `DataFrames.printtable()` https://travis-ci.org/JuliaData/DataFrames.jl/jobs/287109648#L606

docs/src/man/reshaping_and_pivoting.md
cjprybol: I couldn't get our ideas to work on this. setting `ENV["ROWS"]` and `ENV["LINES"]` [from here](https://github.com/JuliaLang/julia/blob/903644385b91ed8d95e5e3a5716c089dd1f1b08a/base/stream.jl#L351-L353) don't appear to have any effect on our show methods. The `IOContext` approach works as long as you explicitly call `show(IOContext, df)`, but I couldn't find a way to set it in a way that would globally change the DataFrame printing. I think we should try and make `ENV["LINES"]` apply to printing (the show + IOContext didn't seem like a useful thing to show in the introductory doctests). It seems that we're already accounting for `ENV["COLUMNS"]` [here](https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/show.jl#L182) & [here](https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/show.jl#L330)

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -72,14 +276,181 @@ This repeats the original columns N times where N is the number of columns stack
...
+      4: Symbol PetalWidth
+    inner: Int64 150
+    outer: Int64 1  value: DataFrames.StackedVector
+    components: Array{Any}((4,))
+      1: Array{Float64}((150,)) [5.1, 4.9, 4.7, 4.6, 5.0, 5.4, 4.6, 5.0, 4.4, 4.9  …  6.7, 6.9, 5.8, 6.8, 6.7, 6.7, 6.3, 6.5, 6.2, 5.9]
+      2: Array{Float64}((150,)) [3.5, 3.0, 3.2, 3.1, 3.6, 3.9, 3.4, 3.4, 2.9, 3.1  …  3.1, 3.1, 2.7, 3.2, 3.3, 3.0, 2.5, 3.0, 3.4, 3.0]
+      3: Array{Float64}((150,)) [1.4, 1.4, 1.3, 1.5, 1.4, 1.7, 1.4, 1.5, 1.4, 1.5  …  5.6, 5.1, 5.1, 5.9, 5.7, 5.2, 5.0, 5.2, 5.4, 5.1]
+      4: Array{Float64}((150,)) [0.2, 0.2, 0.2, 0.2, 0.2, 0.4, 0.3, 0.2, 0.2, 0.1  …  2.4, 2.3, 1.9, 2.3, 2.5, 2.3, 1.9, 2.0, 2.3, 1.8]  Species: DataFrames.RepeatedVector{CategoricalArrays.CategoricalValue{String,UInt32}}
+    parent: CategoricalArrays.CategoricalArray{String,1,UInt32,String,Union{}}
+      refs: Array{UInt32}((150,)) UInt32[0x00000001, 0x00000001, 0x00000001, 0x00000001, 0x00000001, 0x00000001, 0x00000001, 0x00000001, 0x00000001, 0x00000001  …  0x00000003, 0x00000003, 0x00000003, 0x00000003, 0x00000003, 0x00000003, 0x00000003, 0x00000003, 0x00000003, 0x00000003]
```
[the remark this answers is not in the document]
cjprybol: It won't pass the doctests, but if we just want to truncate the output then that's fine. To be honest I'm not sure I see the value of showing `dump` in the documentation (unless we add an advanced section to the documentation, since this is very useful but not something I'd show to a newcomer)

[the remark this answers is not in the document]
nalimilan: Yeah, let's just remove this.

docs/src/man/reshaping_and_pivoting.md
nalimilan: Too bad. I think that's because `displaysize(STDOUT)` ends up calling `uv_tty_get_winsize` instead of relying on `ENV`. A possible hack would be to override `Base.show(df::AbstractDataFrame, allcols::Bool = false)` to call something like `show(IOContext(STDOUT, :displaysize=>(10, 80), df, allcols))`.

docs/src/man/categorical.md
```diff
@@ -91,57 +91,90 @@ julia> cv = compress(cv)
...
+`CategoricalArray` columns. You can convert one or more columns of the DataFrame using the
+`categorical!` function, which modifies the input DataFrame in-place.
 
 ```jldoctest categorical
 julia> using DataFrames
 
-julia> df = DataFrame(A = [1, 1, 1, 2, 2, 2],
-                      B = ["X", "X", "X", "Y", "Y", "Y"])
+julia> df = DataFrame(A = ["A", "B", "C", "D", "D", "A"],
+                             B = ["X", "X", "X", "Y", "Y", "Y"])
```
nalimilan: Bug?

docs/src/man/categorical.md
```diff
@@ -91,57 +91,90 @@ julia> cv = compress(cv)
...
-│ 5   │ 2 │ Y │
-│ 6   │ 2 │ Y │
+│ 1   │ A │ X │
+│ 2   │ B │ X │
+│ 3   │ C │ X │
+│ 4   │ D │ Y │
+│ 5   │ D │ Y │
+│ 6   │ A │ Y │
+
+julia> allcols = deepcopy(df); bothcols = deepcopy(df); onecol = deepcopy(df)
```
nalimilan: It's confusing to do all that setup. Maybe just show first that you can specify a single column, and then call the function without an argument and show that the other column was also converted? Then people can read the docstring to find out more.

nalimilan:
Anything left to do here?

cjprybol:
> Anything left to do here?

Not anymore!

nalimilan:
Doctests are failing on CI (this is hidden by default, need to unroll the section). I think you'll have to install CSV in the same line that installs Documenter.

cjprybol:
`Unsatisfiable requirements detected for package WeakRefStrings` 🙁. Not sure if we can do anything about that until we tag the 11.0 release here since CSV, WeakRefStrings, and DataStreams are upstream. Thoughts on merging and cleaning up any remaining issues post version tag? Everything looks good locally, only issues I think we might still have are on the variability of the # of columns omitted during DataFrame printing and the ordering of union printing (`Union{Null, T}` and `Union{T, Null}`) being different on 0.6 and 0.7

Issue #1250, opened 2017-10-11
Show for columns of CategoricalVector{Int}


cjprybol:
The column is showing `#undef` when printed as a DataFrame even though the underlying data is still intact.

```julia
julia> using DataFrames

julia> df = DataFrame(A = [1, 1, 1, 2, 2, 2], B = ["X", "X", "X", "Y", "Y", "Y"])
6×2 DataFrames.DataFrame
│ Row │ A │ B │
├─────┼───┼───┤
│ 1   │ 1 │ X │
│ 2   │ 1 │ X │
│ 3   │ 1 │ X │
│ 4   │ 2 │ Y │
│ 5   │ 2 │ Y │
│ 6   │ 2 │ Y │

julia> categorical!(df, [:A, :B])
6×2 DataFrames.DataFrame
│ Row │ A      │ B │
├─────┼────────┼───┤
│ 1   │ #undef │ X │
│ 2   │ #undef │ X │
│ 3   │ #undef │ X │
│ 4   │ #undef │ Y │
│ 5   │ #undef │ Y │
│ 6   │ #undef │ Y │

julia> df[:A]
6-element CategoricalArrays.CategoricalArray{Int64,1,UInt32}:
 1
 1
 1
 2
 2
 2

julia> df
6×2 DataFrames.DataFrame
│ Row │ A      │ B │
├─────┼────────┼───┤
│ 1   │ #undef │ X │
│ 2   │ #undef │ X │
│ 3   │ #undef │ X │
│ 4   │ #undef │ Y │
│ 5   │ #undef │ Y │
│ 6   │ #undef │ Y │
```

nalimilan:
That's because `CategoricalValue <: AbstractString` (https://github.com/JuliaData/CategoricalArrays.jl/pull/77). I think we should discourage using non-string element types with `CategoricalArray` (maybe print a deprecation warning on construction?).

nalimilan:
Now I see that you must have bumped into this while calling `categorical!(df, [:A, :B])`. We should change that function to only turn string columns into categorical columns.

cjprybol:
See https://github.com/JuliaData/DataFrames.jl/pull/1254

nalimilan:
Fixed in CategoricalArrays master by https://github.com/JuliaData/CategoricalArrays.jl/pull/87:
```julia
julia> categorical!(df, [:A, :B])
6×2 DataFrames.DataFrame
│ Row │ A │ B │
[...]

PR #1254, opened 2017-10-11
Deprecate use of categorical! function on non-AbstractString columns

cjprybol:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1250

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -740,20 +740,28 @@ end
 ##############################################################################
 
 function categorical!(df::DataFrame, cname::Union{Integer, Symbol})
-    df[cname] = CategoricalVector(df[cname])
+    if Nulls.T(eltype(df[cname])) <: AbstractString
+        df[cname] = CategoricalVector(df[cname])
+    else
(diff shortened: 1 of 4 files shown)


src/dataframe/dataframe.jl
```diff
@@ -740,20 +740,28 @@ end
 ##############################################################################
 
 function categorical!(df::DataFrame, cname::Union{Integer, Symbol})
-    df[cname] = CategoricalVector(df[cname])
+    if Nulls.T(eltype(df[cname])) <: AbstractString
```
nalimilan: Instead of adding this deprecation here, I'd rather do that directly in CategoricalArrays. That way we have only one deprecation to remove once we decide the deprecation period is over.

cjprybol:
Waiting on (post-merge) discussions on what types to support in CategoricalArrays here -> https://github.com/JuliaData/CategoricalArrays.jl/pull/77

nalimilan:
With https://github.com/JuliaData/CategoricalArrays.jl/pull/87, I guess we can keep the current behavior which converts all columns to `CategoricalArray`?

cjprybol:
Yes! Very exciting to see https://github.com/JuliaData/CategoricalArrays.jl/pull/87 get merged

PR #1260, opened 2017-10-14
WIP: reintroduce DataArrays-based tests from 0.10.1

nalimilan:
This PR is a bit special. Its main purpose it to check that the transition from DataFrames 0.10.1 doesn't introduce unintended breakage (that is, on top of intended breakage) and that `DataArray` columns are still supported. The final state is relatively clean and could be merged so that `DataArray` columns continues to be tested for as long as we support them. But each intermediate commit has value on its own since it performs a stricter check which should be useful until we tag 0.11.

The first commit copies the tests from DataFrames 0.10.1 without changing anything.

The second commit adds a hack in the `DataFrame` constructor to force all columns to be `DataArray` or `PooledDataArray`, to be sure these column types are tested. It also updates tests so that they pass. 

The third commit removes this hack and adapts tests to pass again. It gives a cleaner code, but doesn't always test `DataArray` columns since they are not created by default.

The fourth commit fixes all deprecation warnings, which again gives a cleaner code, but doesn't test the transition from the old API to the new one.

Of course all of this won't pass CI until a Nulls-based DataArrays is tagged (https://github.com/JuliaStats/DataArrays.jl/pull/288), but it works locally.


nalimilan:
Yes, I hope that I some point we'll no longer need DataArrays...

Issue #1257, opened 2017-10-13
`join()` doesn't preserve categorical levels order


alyst:
```
julia> using DataFrames
julia> A = DataFrame(a = [1,2,3], b = ["a", "b", "c"]);
julia> B = DataFrame(b = ["a", "b", "c"], c = levels!(categorical(["a", "b", "b"]), ["b", "a"]));
julia> levels(B[:c])
2-element Array{String,1}:
 "b"
 "a"

julia> C = join(A, B, on=:b)
3×3 DataFrames.DataFrame
│ Row │ a │ b │ c │
├─────┼───┼───┼───┤
│ 1   │ 1 │ a │ a │
│ 2   │ 2 │ b │ b │
│ 3   │ 3 │ c │ b │

julia> levels(C[:c])
2-element Array{String,1}:
 "a"
 "b"
```

nalimilan:
Good catch. I think this happens because we call `similar` and fill the resulting array. It's kind of unfortunate since the old code (in 0.10.1) used to rely only on `vcat`, which already implements the required logic in CategoricalArrays. We changed that in https://github.com/JuliaData/DataTables.jl/pull/17 and https://github.com/JuliaData/DataTables.jl/pull/44 for performance reasons (see also [your comment](https://github.com/JuliaData/DataTables.jl/pull/17/files#r104328002)). After https://github.com/JuliaData/DataTables.jl/pull/17, `resize!` was used rather than `similar`, but IIRC there were bugs (which may not have been related directly to this question) which were fixed in https://github.com/JuliaData/DataTables.jl/pull/44. Using `resize!` isn't perfect either since `vcat` includes some logic to merge levels when concatenating two `CategoricalArray`s, which would have to be reimplemented.

If we can't find an efficient solution which doesn't rely on `similar`, we'll have to add a special `fillcolumn!` method for `CategoricalArray` columns which would call `CategoricalArrays.mergelevels` and set levels based on that. I don't like the idea of adding another special case, though.

Would you be willing to investigate this? I'm pretty swamped at the moment.

Cc: @cjprybol

alyst:
To me copying the pool into the new array when doing `similar(A::CategoricalArray)` seems the most straightforward solution.
Similar problem would have to be solved for Query.jl (when it would properly handle categorical arrays), which grows the result row by row, so the order of the levels in the new vector may change.
IMO there are very few situations when `similar()` doesn't need to copy/share the pool.

[...]

PR #1266, opened 2017-10-17
Ensure levels are maintained when joining df's with Categorical cols

cjprybol:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1257

cc @alyst

### DIFF
--- a/test/join.jl
+++ b/test/join.jl
@@ -325,4 +325,50 @@ module TestJoin
         @test all(isa.(o(on).columns,
                        [CategoricalVector{Union{T, Missing}} for T in (Int, Float64)]))
     end
+
+    @testset "maintain CategoricalArray levels ordering on join - non-`on` cols" begin
+        A = DataFrame(a = [1, 2, 3], b = ["a", "b", "c"])
+        B = DataFrame(b = ["a", "b", "c"], c = CategoricalVector(["a", "b", "b"]))
+        levels!(B[:c], ["b", "a"])
+        @test levels(join(A, B, on=:b, kind=:inner)[:c]) == ["b", "a"]
+        @test levels(join(B, A, on=:b, kind=:inner)[:c]) == ["b", "a"]
+        @test levels(join(A, B, on=:b, kind=:left)[:c]) == ["b", "a"]
+        @test levels(join(A, B, on=:b, kind=:right)[:c]) == ["b", "a"]
+        @test levels(join(A, B, on=:b, kind=:outer)[:c]) == ["b", "a"]
+        @test levels(join(B, A, on=:b, kind = :semi)[:c]) == ["b", "a"]
+    end
+
+    @testset "maintain CategoricalArray levels ordering on join - ordering conflicts" begin
+        A = DataFrame(a = [1, 2, 3, 4], b = CategoricalVector(["a", "b", "c", "d"]))
+        levels!(A[:b], ["d", "c", "b", "a"])
+        B = DataFrame(b = CategoricalVector(["a", "b", "c"]), c = [5, 6, 7])
+        @test levels(join(A, B, on=:b, kind=:inner)[:b]) == ["d", "c", "b", "a"]
+        @test levels(join(B, A, on=:b, kind=:inner)[:b]) == ["a", "b", "c"]
+        @test levels(join(A, B, on=:b, kind=:left)[:b]) == ["d", "c", "b", "a"]
+        @test levels(join(B, A, on=:b, kind=:left)[:b]) == ["a", "b", "c"]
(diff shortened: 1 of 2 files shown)


cjprybol:
Updated to include the `mergelevels` steps from `copy!` in CategoricalArrays. I think it's probably possible to refactor this to use `copy!` directly, but if we did that then we would need to add a few other steps that I think would ultimately make that less efficient than what we already have here. For a `copy!`-based approach we would need to copy, then (potentially) resize, then (potentially) promote the categorical eltype from T to Union{T, Null}, and then we'd still have to call our custom `fillcolumn!` function to make sure the indices from the original DataFrames get placed in the correct indices in the output DataFrame when the values in the resized column should have gaps inserted between them (to accept values from right outer join). Checking if the original columns were categorical, copying over the levels, and using `mergelevels` to handle the right outer join case where we need to bring in values from the right table that are missing from the left table seems like a sensible way to get the same result.

If the gaps that we insert to accept outer join values are always at the end, then I'm wrong here and `copy!` might allow us to get rid of `fillcolumn!`, but I don't think that's the case (see examples from before we fixed the ROJ bug [here](https://github.com/JuliaData/DataTables.jl/issues/34#issuecomment-287203032))

I've added a note to the docstring for `join` to clarify that the levels ordering of the left table takes precedence over the ordering of the right table

test/join.jl
```diff
@@ -325,4 +325,50 @@ module TestJoin
...
+        @test levels(join(B, A, on=:b, kind=:left)[:b]) == ["a", "b", "c"]
+        @test levels(join(A, B, on=:b, kind=:right)[:b]) == ["d", "c", "b", "a"]
+        @test levels(join(B, A, on=:b, kind=:right)[:b]) == ["a", "b", "c", "d"]
+        @test levels(join(B, A, on=:b, kind=:outer)[:b]) == ["a", "b", "c", "d"]
+        @test levels(join(A, B, on=:b, kind=:outer)[:b]) == ["d", "c", "b", "a"]
+        @test levels(join(A, B, on=:b, kind = :semi)[:b]) == ["d", "c", "b", "a"]
+        @test levels(join(B, A, on=:b, kind = :semi)[:b]) == ["a", "b", "c"]
+    end
+
+    @testset "maintain Categorical levels ordering on join - only 1 is categorical" begin
```
nalimilan: "Only LHS is categorical" (IIUC).

test/join.jl
```diff
@@ -325,4 +325,50 @@ module TestJoin
         @test all(isa.(o(on).columns,
                        [CategoricalVector{Union{T, Null}} for T in (Int, Float64)]))
     end
+
+    @testset "maintain Categorical levels ordering on join - non-`on` cols" begin
```
nalimilan: `CategoricalArray`

test/join.jl
```diff
@@ -325,4 +325,50 @@ module TestJoin
...
+        A = DataFrame(a = [1, 2, 3], b = ["a", "b", "c"])
+        c = levels!(categorical(["a", "b", "b"]), ["b", "a"])
+        B = DataFrame(b = ["a", "b", "c"], c = c)
+        @test levels(join(A, B, on=:b, kind=:inner)[:c]) == ["b", "a"]
```
nalimilan: It would be nice to test `isordered` too, since that's easy to get wrong. While you're at it, testing the actual contents of the resulting column wouldn't hurt either, if that's not too much additional code.

cjprybol:
Ignore what I said before, a combination of `permute!` plus an off-by-one error I made during my first attempt at using `copy!` confused me into thinking something completely different was going on here. This now uses `copy!`.

I think I remember why we went with `fillcolumn!` instead of `copy!` the first time though. We make a copy of the original column when we subset/reorder and pass it as the source into the `copy!` function. One work around would be to `view(col, all_orig_left/right_indices)` rather than `col[all_orig_left/right_indices]` which would achieve the same goal and avoid the copy. We'd have to extend https://github.com/JuliaData/CategoricalArrays.jl/pull/92 to accept `SubArray{CategoricalArray}`. Alternatively, we might be able to just permute the source column in-place (within the DataTableJoiner object).

nalimilan:
Ah, great, that's really the most natural and generic API. Using `view` would also sound like the simplest solution. Can you extend JuliaData/CategoricalArrays.jl#92  to accept `SubArray{CategoricalArray}`?

cjprybol:
This is working for me locally with https://github.com/JuliaData/CategoricalArrays.jl/pull/97. We'll need to merge that, tag a new release and update this PR with a new minimum requirement on CategoricalArrays

nalimilan:
CI won't pass until we move to Missings. I'm preparing a PR for that.

nalimilan:
Thanks! I really like how you have been able to fix a bug while removing code.

Issue #1249, opened 2017-10-10
Decide what to do with null values in grouping


nalimilan:
On current master, and since https://github.com/JuliaData/DataTables.jl/pull/17, we create separate groups for `null` values:
```julia
julia> groupby(df, :x)
DataFrames.GroupedDataFrame  2 groups with keys: Symbol[:x]
First Group:
1×1 DataFrames.SubDataFrame{Array{Int64,1}}
│ Row │ x    │
├─────┼──────┤
│ 1   │ null │
⋮
Last Group:
1×1 DataFrames.SubDataFrame{Array{Int64,1}}
│ Row │ x │
├─────┼───┤
│ 1   │ 0 │
```

The old code used to discard nulls silently, without an option to retain them (just like Pandas does). Since the behavior changes so radically, we should decide what's the best default before tagging a release. I think keeping nulls is a good default as it's consistent with what we do everywhere. But we should probably provide an argument to skip them so that people can get the old behavior easily. This is not so easy to implement when there are multiple grouping variables, each of which can be null independently from others.

A second issue is that by default we no longer sort groups, which means that null groups are not necessarily at the end. Maybe we should enable sorting by default. IIRC it isn't very costly it is in terms of performance, and I suspect it would be a more useful default.

Note that when using `N` grouping variables with nulls, there will be `(2^N)-1` groups with a `null`. We could imagine providing an option to collapse all of these into a single "at least one `null`" group, but probably not by default so that can wait.

cjprybol:
I'd prefer requiring explicit handling of nulls and treat them as a group just like the other values. The behavior you describe of:
> The old code used to discard nulls silently, without an option to retain them

is the kind of thing (my) data analysis nightmares are made of. Groups with missing data could be one of the most informative or important groups in the dataset, I think it makes sense to support them by default as first-class citizens. Removing them by default would be at best an annoyance and at worst a paper retraction waiting to happen.

I think the best argument in favor of handling nulls in `groupby` is that we still don't have great support for dropping nulls on subsets of the DataFrame as was brought up in https://github.com/JuliaData/DataTables.jl/issues/38, but I think we'd be better served by implementing something like @mkborregaard's proposal of allowing columns to be specified.

e.g. `groupby(dropnull(df, :x), :x)` would achieve the same purpose but allows more flexibility (because the functions can be used independently) and better re-use of the DataFrames method library than implementing something like `groupby(df, :x, dropnull=true)` would.

[...]

PR #1267, opened 2017-10-17
Add skipnull argument to groupby()

nalimilan:
The new argument allows skipping values with at least one null in one
of the grouping columns. Doing this in `groupby()` is more efficient
and more convenient than dropping them before calling `groupby()`.

Fixes https://github.com/JuliaData/DataFrames.jl/issues/1249.

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -22,15 +22,25 @@ end
 
 # "kernel" functions for hashrows()
 # adjust row hashes by the hashes of column elements
-function hashrows_col!(h::Vector{UInt}, v::AbstractVector)
+function hashrows_col!(h::Vector{UInt},
+                       n::Vector{Bool},
+                       v::AbstractVector{T}) where T
     @inbounds for i in eachindex(h)
-        h[i] = hash(v[i], h[i])
+        el = v[i]
+        h[i] = hash(el, h[i])
+        if T >: Null && length(n) > 0
+            # el isa Null should be redundant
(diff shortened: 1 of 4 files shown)


nalimilan:
Here's a fix for https://github.com/JuliaData/DataFrames.jl/issues/1249. I still need to run benchmarks to check that the performance impact is reasonable (and almost non-existent with `skipnull=false`, which remains the default).

nalimilan:
I've tested the benchmarks from https://github.com/JuliaData/DataTables.jl/pull/17, and the overhead due to this PR with `skip=false` and `Union{T, Null}` columns with no nulls (i.e. the worst case) is generally around 10-14%. In theory this shouldn't happen, but codegen is really brittle currently in the presence of `Union`s. Indeed, with `CategoricalArray` columns, where we work directly with integer codes, the overhead is almost zero. Given that it's supposed to improve and that this is an essential feature, I'd say it's OK. We should add a `DataArray` method similar to the `CategoricalArray` one to have the same benefits until `Union` is correctly handled.

cjprybol:
I agree that the value of this feature > minor performance slowdown. I assume it'll improve back to current performance as Union improvements continue to land in base and get shipped out with future versions.

> We should add a DataArray method similar to the CategoricalArray one

^What method do you mean here? `hashrows_col!` with `v::DataVector{T}`?

nalimilan:
> ^What method do you mean here? hashrows_col! with v::DataVector{T}?

Yes.

PR #1269, opened 2017-10-27
Improve rename API

nalimilan:
Using pairs is more convenient, more flexible and more consistent.


nalimilan:
It's so nice to have tests passing on 0.7 again...

PR #1274, opened 2017-11-09
 Fix deprecations

Nosferican:


nalimilan:
```
@isdefined not defined
```

That funny message is due to the fact that it doesn't exist on 0.6.

src/deprecated.jl
```diff
@@ -460,10 +460,10 @@ for allowcomments in tf, skipblanks in tf, allowescapes in tf, wsv in tf
     end
 end
 
-function bytematch{T <: String}(bytes::Vector{UInt8},
-                                    left::Integer,
-                                    right::Integer,
-                                    exemplars::Vector{T})
+function bytematch(bytes::Vector{UInt8},
+                       left::Integer,
```
nalimilan: Incorrect indentation. BTW, no need to define `T`.

src/deprecated.jl
```diff
@@ -480,16 +480,16 @@ function bytematch{T <: String}(bytes::Vector{UInt8},
...
-                                  falsestrings::Vector{P} = P[])
+function bytestotype(::Type{N},
+                     bytes::Vector{UInt8},
+                     left::Integer,
+                     right::Integer,
+                     nastrings::Vector{T},
+                     wasquoted::Bool = false,
+                     truestrings::Vector{P} = P[],
+                     falsestrings::Vector{P} = P[]) where {N <: Integer,
+                                                         T <: String,
```
nalimilan: Incorrect indentation. Same below.

Nosferican:
As for `isdefined` I saw it was mainly as a catch... I remember reading a while back that a best practice was to always classify variables...
```
import DataFrames
myfunc(df::DataFrames.AbstractDataFrame)
```
rather than
```
using DataFrames
myfunc(df::DataFrame)
```
would that be a better style? If so, the `isdefined` wouldn't be needed.
I could open an issue to change the syntax to the first method and make a PR if it is good.

nalimilan:
I don't understand the link with the `@isdefined` problem. That code is just there to handle Julia 0.6 where `Base.unique!` does not exist. There's nothing to qualify since there's no function at all.

ararslan:
`@isdefined` doesn't exist in Base nor in Compat on 0.6

src/deprecated.jl
nalimilan: Still incorrect indentation.

src/deprecated.jl
nalimilan: Same here.

src/deprecated.jl
```diff
@@ -528,16 +528,16 @@ end
...
-                                      falsestrings::Vector{P} = P[])
+    function bytestotype(::Type{N},
+                         bytes::Vector{UInt8},
+                         left::Integer,
+                         right::Integer,
+                         nastrings::Vector{T},
+                         wasquoted::Bool = false,
+                         truestrings::Vector{P} = P[],
+                         falsestrings::Vector{P} = P[]) where {N <: AbstractFloat,
+                                                             T <: String,
```
nalimilan: Same here.

src/deprecated.jl
Nosferican: What's the proper indentation?
```
function func(obj::Vector{T,R},
    left::T,
    right::T) where {T <: Any,
    R <: Any}
```
Always four spaces for long lines?
The other functions in the package have it running past 80 characters and matching previous line start position.

nalimilan: As for all functions in this project, align arguments vertically with the first argument. Align type parameters with the first one.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -596,7 +596,7 @@ end
 nonunique(df::AbstractDataFrame, cols::Union{Real, Symbol}) = nonunique(df[[cols]])
 nonunique(df::AbstractDataFrame, cols::Any) = nonunique(df[cols])
 
-if isdefined(:unique!)
+if @isdefined unique!
```
nalimilan: Don't use `@isdefined`.

src/deprecated.jl
Nosferican: Thanks!

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -596,12 +596,8 @@ end
 nonunique(df::AbstractDataFrame, cols::Union{Real, Symbol}) = nonunique(df[[cols]])
 nonunique(df::AbstractDataFrame, cols::Any) = nonunique(df[cols])
 
-if isdefined(:unique!)
```
nalimilan: No, you need to keep `isdefined` since `unique!` does not exist on Julia 0.6.

src/deprecated.jl
```diff
@@ -528,16 +528,17 @@ end
...
-    function bytestotype{N <: AbstractFloat,
+    function bytestotype(::Type{N},
+                         bytes::Vector{UInt8},
+                         left::Integer,
+                         right::Integer,
+                         nastrings::Vector{T},
+                         wasquoted::Bool = false,
+                         truestrings::Vector{P} = P[],
+                         falsestrings::Vector{P} = P[]) where {
+                         N <: AbstractFloat,
```
nalimilan: Use the same style as above for consistency.

src/deprecated.jl
```diff
@@ -558,16 +559,16 @@ let out = Vector{Float64}(1)
...
-                                  falsestrings::Vector{P} = P[])
+function bytestotype(::Type{N},
+                     bytes::Vector{UInt8},
+                     left::Integer,
+                     right::Integer,
+                     nastrings::Vector{T},
+                     wasquoted::Bool = false,
+                     truestrings::Vector{P} = P[],
+                     falsestrings::Vector{P} = P[]) where {N <: Bool,
+                                                         T <: String,
```
nalimilan: Still not aligned correctly here and below.

Nosferican:
Thanks for all the feedback!

nalimilan:
> Is there any reason that the AppVeyor nightlies are pointing to 0.6.1 instead of 0.7-? Both sets of windows tests are using Julia commit 0d7248e2ff*

I think they are broken right now but that people are working on it.

@Nosferican We have approved the PR, but it fails due to `textwidth` not being defined. I guess it's not in Compat yet?

Nosferican:
It was merged to the master yesterday [411](https://github.com/JuliaLang/Compat.jl/pull/411). Are you aware of any restrictions on using the latest Compat? If not, it should be fine as soon as the version gets tagged and we can up the Compat version to include it. As an aside, for the Julia 0.6 version, will it still use the Missings rather than Nullable (I thought `Nullable` was better in Julia 0.6 for the compiler and `Union{Missing,T}` a viable alternative on Julia 0.7+).

Nosferican:
This should be ready to be merged.

PR #1283, opened 2017-11-23
Use skipmissing() instead of Missings.skip()

nalimilan:
The latter is deprecated.

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -44,21 +44,21 @@ true
 
 ```
 
-`missing` values can be excluded when performing operations by using `Missings.skip`, which returns a memory-efficient iterator.
+`missing` values can be excluded when performing operations by using `skipmissing`, which returns a memory-efficient iterator.
 
 ```jldoctest missings
-julia> Missings.skip(x)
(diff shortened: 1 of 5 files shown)


docs/src/man/getting_started.md
```diff
@@ -44,21 +44,21 @@ true
...
 ```
 
-`missing` values can be excluded when performing operations by using `Missings.skip`, which returns a memory-efficient iterator.
+`missing` values can be excluded when performing operations by using `skipmissing`, which returns a memory-efficient iterator.
 
 ```jldoctest missings
-julia> Missings.skip(x)
-Base.Generator{Base.Iterators.Filter{Missings.##4#6,Array{Union{Int64, Missings.Missing},1}},Missings.##3#5}(Missings.#3, Base.Iterators.Filter{Missings.##4#6,Array{Union{Int64, Missings.Missing},1}}(Missings.#4, Union{Int64, Missings.Missing}[1, 2, missing]))
+julia> skipmissing(x)
+Missings.EachSkipMissing{Array{Union{$Int, Missings.Missing},1}}(Union{$Int, Missings.Missing}[1, 2, missing])
```
nalimilan: Looking at the online manual, it doesn't look like `$Int` was correct here.

@cjprybol Do you know how we're supposed to handle this kind of situation?

cjprybol: The `$Int` is necessary to get tests to pass on 32 & 64 bit, but it won't work because the `$Int` is within the triple-backtick code block? Is that the issue?

nalimilan:
Yes. Do you know of any solution?

cjprybol:
Couldn't think of anything better than just switching the tests to Floats, but I guess that'll do the job, right? See https://github.com/JuliaData/DataFrames.jl/pull/1299

PR #1289, opened 2017-11-26
Switch to CodecZlib from GZip

alyst:
I'm not sure whether `writetable()` is already at its end-of-life stage, but if it stays for a while, it might be better to switch to `CodecZlib.jl` from `GZip.jl` as the latter is deprecated and not actively maintained.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,5 +6,5 @@ SortingAlgorithms
 Reexport
 WeakRefStrings 0.4.0
 DataStreams 0.3.0
-GZip
+CodecZlib 0.4
 Compat 0.36.0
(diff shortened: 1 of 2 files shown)


cjprybol:
Have we set an end-of-life date for `read/writetable` functions? DataFrames 0.12? Julia 1.0? Maybe each version release (0.11.2, 0.11.3, etc.) we add a `sleep(10)` call until the delay becomes so unbearable that everyone voluntary converts their code to use CSV.jl?

nalimilan:
I don't think we're in a hurry to remove them, they're not a big burden. It would be logical to remove them in the next pseudo-major release, so 0.12 I guess.

PR #1290, opened 2017-11-27
add join pipelining

bicycle1885:
This is a simple trick to pipeline joins in the same way as dplyr.

```
julia> name |> join(job, on=:ID)
2×3 DataFrames.DataFrame
│ Row │ ID │ Name     │ Job    │
├─────┼────┼──────────┼────────┤
│ 1   │ 1  │ John Doe │ Lawyer │
│ 2   │ 2  │ Jane Doe │ Doctor │

```

I think somebody has ever tried the same idea, but I couldn't find related discussions.

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -300,6 +300,11 @@ function Base.join(df1::AbstractDataFrame,
     end
 end
 
+function Base.join(right::AbstractDataFrame; on=Symbol[], kind=:inner)
+    # for pipelining with |>
+    return left -> join(left, right; on=on, kind=kind)
+end
+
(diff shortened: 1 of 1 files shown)


nalimilan:
Hmm, I'm not sure we want to go that route. If we start adding convenience functions like this, where are we going to stop? I'd rather use dedicated macros like DataFramesMeta and Query.

bicycle1885:
Me, neither. I made a pull request because I found `groupby` supports similar pipelining. If you'd like to reject this pull request, I think we should remove all such interfaces from this package.

cjprybol:
replaced by https://github.com/JuliaData/DataFrames.jl/pull/1296

PR #1296, opened 2017-12-01
deprecate pipelining with |>

### LINKED ISSUE #1295: Removing pipeline API from DataFrames.jl
I opened #1290 to add pipeline API to `join` dataframes. This is because I found `group_by` supports the way to combine operations with `|>`: https://github.com/JuliaData/DataFrames.jl/blob/74eb8902ebccaa5ce413ca5e4c75c00b0e296694/src/groupeddataframe/grouping.jl#L365-L366

After a while, I started to think this kind of API should not be present in DataFrames.jl because it is too opinionated to have in this basic package. It should be offered by other high-level utility packages like Query.jl or DataFramesMeta.jl as @nalimilan suggested. Please note that I don't say I dislike pipeline interfaces like dplyr, I just say it can be placed anywhere else.

bicycle1885:
This fixes #1295.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -235,7 +235,7 @@ df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
                b = repeat([2, 1], outer=[4]),
                c = randn(8))
 colwise(sum, df)
-colwise([sum, lenth], df)
+colwise([sum, length], df)
 colwise((minimum, maximum), df)
 colwise(sum, groupby(df, :a))
 ```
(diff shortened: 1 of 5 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -239,7 +229,7 @@ df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
                b = repeat([2, 1], outer=[4]),
                c = randn(8))
 colwise(sum, df)
-colwise([sum, lenth], df)
```
cjprybol: 💯

Issue #1294, opened 2017-12-01
allowmissing!() should preserve CategoricalArray


nalimilan:
Currently:
```julia
julia> df = DataFrame(a=categorical(["a"]))
1×1 DataFrames.DataFrame
│ Row │ a │
├─────┼───┤
│ 1   │ a │

julia> allowmissing!(df)
1×1 DataFrames.DataFrame
│ Row │ a │
├─────┼───┤
│ 1   │ a │

julia> df[1]
1-element Array{Union{CategoricalArrays.CategoricalString{UInt32}, Missings.Missing},1}:
 "a"
```

See https://github.com/JuliaData/Missings.jl/issues/60#issuecomment-348454257 about the more general problem about supporting any custom array type.

nalimilan:
As noted at https://github.com/JuliaData/Missings.jl/issues/60, we could just do `convert(AbstractArray{Union{eltype(x), Missing}}, col)`, which already works for `CategoricalArray`, but also for other special cases like `BitArray`.

Issue #1291, opened 2017-11-27
`similar(::DataFrame, 0)` changes column types


omus:
Using DataFrames v0.11.0:
```julia
julia> row = DataFrame(a="foo")
1×1 DataFrames.DataFrame
│ Row │ a   │
├─────┼─────┤
│ 1   │ foo │

julia> df = similar(row, 0)
0×1 DataFrames.DataFrame


julia> append!(df, row)
ERROR: Column eltypes do not match
Stacktrace:
 [1] append!(::DataFrames.DataFrame, ::DataFrames.DataFrame) at /tmp/julia/v0.6/DataFrames/src/dataframe/dataframe.jl:765

julia> DataFrames.columns(row)
1-element Array{Any,1}:
 String["foo"]

julia> DataFrames.columns(df)
1-element Array{Any,1}:
 Union{Missings.Missing, String}[]
```

omus:
Maybe we should update the [`append!`](https://github.com/JuliaData/DataFrames.jl/blob/master/src/dataframe/dataframe.jl#L765) element check to use subtype checks:
```julia
all(issubtype(el2, el1) for (el1, el2) in zip(eltypes(df1), eltypes(df2)))
```

nalimilan:
I guess we should change `similar` to respect column eltypes.

Using subtype checks in `append!` also make sense, but note that in your example it wouldn't help since it's the first eltype which is a subtype of the second one. A more general solution would be to attempt a conversion (i.e. call `copy!` and see what happens), but it will be tricky to handle when conversion fails mid-way.

cjprybol:
@omus The behavior of `similar` should now be correct on master, thank you for the report. I'll leave this open until we address `append!`

JonWel:
When loading a file with [CSV.jl ](https://github.com/JuliaData/CSV.jl), it happens quite often that the columns of the dataframe are a union with missing.
This make it difficult to append a new line.

How am I expected to create a dataframe which columns are `Union{Float64, Missings.Missing}`?
```Julia
julia> eltypes(DataFrame(c = 0.1::Union{Float64, Missings.Missing}))
1-element Array{Type,1}:
 Float64
```
Seems I'll have to copy a line from the dataframe, change it's value, then maybe it will append ...

Would be much better to be able to insert a dataframe with `Float64` in a dataframe composed of `Union{Float64, Missings.Missing}`.

omus:
[...]

PR #1298, opened 2017-12-01
Small improvements to missing value support

cjprybol:
- change behavior of `similar` on DataFrames to not auto-enable missing
  value support and instead use same column eltypes as parent df
- add docstrings for `similar` and `allowmissing!`
- change behavior of `allowmissing!` on CategoricalArrays to preserve
  correct Array type (which is a CategoricalArray)
- add `similar` to docs

Should fix https://github.com/JuliaData/DataFrames.jl/issues/1294 and https://github.com/JuliaData/DataFrames.jl/issues/1291, although it does not address the `append!` ideas presented in https://github.com/JuliaData/DataFrames.jl/issues/1291.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -233,8 +233,17 @@ Base.ndims(::AbstractDataFrame) = 2
 ##
 ##############################################################################
 
-Base.similar(df::AbstractDataFrame, dims::Int) =
-    DataFrame(Any[similar_missing(x, dims) for x in columns(df)], copy(index(df)))
+"""
+    similar(df::DataFrame[, rows::Int])
+
+Create a new `DataFrame` with the same column names and column element types
+as `df`. An optional second argument can be provided to request a number of rows
+that is different than the number of rows present in `df`.
+"""
+function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
+    rows < 0 && throw(ArgumentError("the number of rows must be positive"))
+    DataFrame(Any[similar(x, rows) for x in columns(df)], copy(index(df)))
+end
 
 ##############################################################################
 ##
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -234,7 +234,7 @@ Base.ndims(::AbstractDataFrame) = 2
 ##############################################################################
 
 """
-    similar(df::DataFrame[, rows::Int])
+    similar(df::DataFrame[, rows::Integer])
 
 Create a new `DataFrame` with the same column names and column element types
 as `df`. An optional second argument can be provided to request a number of rows
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -721,11 +721,29 @@ Base.hcat(df1::DataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame...) = h
 ## Missing values support
 ##
 ##############################################################################
+"""
+    allowmissing!(df::DataFrame)
+
+Convert all columns of a `DataFrame` from element type `T` to
+`Union{T, Missing}` to support missing values.
+
+    allowmissing!(df::DataFrame, col::Union{Integer, Symbol})
+
+Convert a single column of a `DataFrame` from element type `T` to
+`Union{T, Missing}` to support missing values.
+
+    allowmissing!(df::DataFrame, cols::AbstractVector{<:Union{Integer, Symbol}})
+
+Convert multiple columns of a `DataFrame` from element-type `T` to
+`Union{T, Missing}` to support missing values.
+"""
+function allowmissing! end
 
 function allowmissing!(df::DataFrame, col::ColumnIndex)
-    df[col] = Vector{Union{eltype(df[col]), Missing}}(df[col])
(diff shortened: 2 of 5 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -233,8 +233,30 @@ Base.ndims(::AbstractDataFrame) = 2
...
-    DataFrame(Any[similar_missing(x, dims) for x in columns(df)], copy(index(df)))
+"""
+    similar(df::DataFrame[, rows::Int])
+
+Create a new DataFrame with the same column names and column element-types
+as `df`. An optional second argument can be provided to request a number of rows
+that is different than the number of rows present in `df`.
+"""
+function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
+    @assert rows >= 0 "the number of rows must be positive"
```
nalimilan: Throw an `ArgumentError` instead.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -233,8 +233,30 @@ Base.ndims(::AbstractDataFrame) = 2
...
+    DataFrame(Any[similar(x, rows) for x in columns(df)], copy(index(df)))
+end
+
+"""
+    similar_missing(df::DataFrame[, rows::Int])
```
nalimilan: I'd prefer if we avoided adding this function until we're completely certain it's needed. People can use `allowmissing!`, and hopefully the compiler will be able to make this efficient at some point.

src/dataframe/dataframe.jl
```diff
@@ -721,11 +721,25 @@ Base.hcat(df1::DataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame...) = h
 ## Missing values support
 ##
 ##############################################################################
+"""
+    allowmissing!(df::DataFrame, col::ColumnIndex)
 
+Convert a single column of a DataFrame from element-type `T` to
```
nalimilan: `` `DataFrame` `` and "element type".

I think you could merge the two docstrings into a single one, with one line for each signature (or maybe three, with a line for the single-argument method). Attach it to `function allowmissing! end`.

src/dataframe/dataframe.jl
```diff
@@ -721,11 +721,25 @@ Base.hcat(df1::DataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame...) = h
...
+    #TODO use allowmissing from Missings or Base here when possible
+    df[col] = convert(AbstractArray{Union{eltype(df[col]), Missing}}, df[col])
     df
 end
+"""
+    allowmissing!(df::DataFrame[, cols::AbstractVector{<: ColumnIndex}])
+
+Convert columns of a DataFrame in-place from element-type `T` to
+`Union{T, Missing}`. Defaults to converting all columns of the DataFrame,
+although users can manually specify a subset of columns to convert using the
```
nalimilan: I don't really like talking about "users" here, it's weird for the "user" who is reading the docstring. :-)

test/dataframe.jl
```diff
@@ -439,12 +439,40 @@ module TestDataFrame
...
+        @test eltypes(df)[1] <: Union{CategoricalValue{Int}, Missing}
+        @test eltypes(df)[2] <: Union{CategoricalString, Missing}
+    end
+
+    @testset "similar" begin
+        df = DataFrame(a = ["foo"],
+                       b = CategoricalArray(["foo"]),
+                       c = [0.0],
+                       d = CategoricalArray([0.0]))
+        @test typeof.(df.columns) == typeof.(similar(df).columns)
```
nalimilan: Also test the size of the result, and passing a valid number or rows different from the original?

test/dataframe.jl
```diff
@@ -79,9 +79,9 @@ module TestDataFrame
                    b = Union{String, Missing}["b"],
                    c = CategoricalArray{Union{Float64, Missing}}([3.3]))
     missingdf = DataFrame(a = missings(Int, 2),
-                       b = missings(String, 2),
-                       c = CategoricalArray{Union{Float64, Missing}}(2))
-    @test missingdf ≅ similar(df, 2)
+                          b = missings(String, 2),
+                          c = CategoricalArray{Union{Float64, Missing}}(2))
+    # @test missingdf ≅ similar(df, 2)
```
cjprybol: https://github.com/JuliaData/Missings.jl/issues/66

nalimilan: Better leave a comment pointing to that issue, and in the meantime check the types and size of the result.

src/dataframe/dataframe.jl
```diff
@@ -721,11 +721,29 @@ Base.hcat(df1::DataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame...) = h
 ## Missing values support
 ##
 ##############################################################################
+"""
+    allowmissing!(df::DataFrame)
+
+Convert all columns of a `DataFrame` from element type `T` to
+`Union{T, Missing}` to support missing values.
+
+    allowmissing!(df::DataFrame, col::ColumnIndex)
```
nalimilan: `ColumnIndex` is not exported, so I guess we should use `Union{Integer, Symbol}`. Not sure what we do in other docstrings.

cjprybol:
Thanks for tagging a new release of Missings @nalimilan! Closing to re-run tests

PR #1299, opened 2017-12-01
Handle different platforms better with doctests

cjprybol:
- Convert some Int tests to Float64 to avoid issues running doctests on
  32 bit machines
- use `showall` in `aggregate` doctests to force showing all columns


nalimilan:
Clever! We should probably get in touch with Documenter developers to find more general solutions with both 32- vs 64-bit and screen width, but for now that's clearly better than failing tests. Do you want to reenable them BTW?

cjprybol:
We can re-enable everything but the Query doctests, those aren't working for me locally. I see that's why the doctests were temporarily disabled in https://github.com/JuliaData/DataFrames.jl/pull/1285. Happy to change the commit by re-enabling the doctests and disabling the query tests or we can leave it as is, let me know which you prefer

Issue #1301, opened 2017-12-02
Possible improvement of hcat


bkamins:
Consider the following code:
```
 x = DataFrame([(i,j) for i in 1:3, j in 1:4])
hcat(x, [1,2,3])
```
works OK, but:
```
hcat([1,2,3], x)
```
fails.

Additionally:
```
hcat([1 2 3], x)
```
appends `DataFrame` to `[1 2 3]` in fourth position.

Is this intended, or should it be fixed.

bkamins:
Also this is probably something we should avoid (`x` is a DataFrame):
```
hcat(x, "bad mistake")
```
which produces StackOverflow.

nalimilan:
I don't think that's really intended, but it's difficult to fix in full generality since AFAIK there's no way to define a method which is called when at least one of the arguments is a `DataFrame`. See https://github.com/JuliaLang/julia/pull/20815. I guess we could add methods to handle the two-argument case at least, since it's likely to be the most common one. PR welcome.

Regarding the stack overflow, it seems it could be fixed by adding a two-argument `hcat!` method to stop the recursive call of `hcat!(a::DataFrame, b, c...)`.

bkamins:
This is what I meant. I will make a PR (I was not so much into DataFrames development earlier so I do not know well what already has been considered and decided).

bkamins:
closing as the PR is merged.

PR #1303, opened 2017-12-02
Improve hcat for DataFrames

bkamins:
Following the discussion in #1301:

* `hcat` now accepts `AbstarctVector` as first argument if the second argument is `DataFrame` and produces a `DataFrame` in result;
* `hcat` with `DataFrame` throws an error if something else than `DataFrame` or `AbstractVector` is passed as a second argument.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -714,10 +714,10 @@ end
 hcat!(df::DataFrame, x::AbstractVector) = hcat!(df, DataFrame(Any[x]))
 hcat!(x::AbstractVector, df::DataFrame) = hcat!(DataFrame(Any[x]), df)
 function hcat!(x, df::DataFrame)
-    throw(ArgumentError("x must be AbstaractVector or AbstaractDataFrame"))
+    throw(ArgumentError("x must be AbstaractVector or AbstractDataFrame"))
 end
-hcat!(df::DataFrame, x)
-    throw(ArgumentError("x must be AbstaractVector or AbstaractDataFrame"))
+function hcat!(df::DataFrame, x)
+    throw(ArgumentError("x must be AbstaractVector or AbstractDataFrame"))
 end
 
 # hcat! for 1-n arguments
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -713,8 +713,12 @@ end
 
 hcat!(df::DataFrame, x::AbstractVector) = hcat!(df, DataFrame(Any[x]))
 hcat!(x::AbstractVector, df::DataFrame) = hcat!(DataFrame(Any[x]), df)
-hcat!(x, df::DataFrame) = throw(ArgumentError("x must be a vector or a data frame"))
-hcat!(df::DataFrame, x) = throw(ArgumentError("x must be a vector or a data frame"))
+function hcat!(x, df::DataFrame)
+    throw(ArgumentError("x must be AbstaractVector or AbstaractDataFrame"))
+end
+hcat!(df::DataFrame, x)
+    throw(ArgumentError("x must be AbstaractVector or AbstaractDataFrame"))
+end
 
 # hcat! for 1-n arguments
 hcat!(df::DataFrame) = df
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -705,6 +705,12 @@ function hcat!(df1::DataFrame, df2::AbstractDataFrame)
     end
     return df1
 end
+
+# definition required to avoid hcat! ambiguity
+function hcat!(df1::DataFrame, df2::DataFrame)
+    invoke(hcat!, Tuple{DataFrame, AbstractDataFrame}, df1, df2)
+end
+
 hcat!(df::DataFrame, x::AbstractVector) = hcat!(df, DataFrame(Any[x]))
 hcat!(x::AbstractVector, df::DataFrame) = hcat!(DataFrame(Any[x]), df)
 hcat!(x, df::DataFrame) = throw(ArgumentError("x must be a vector or a data frame"))
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -714,10 +714,10 @@ end
 hcat!(df::DataFrame, x::AbstractVector) = hcat!(df, DataFrame(Any[x]))
 hcat!(x::AbstractVector, df::DataFrame) = hcat!(DataFrame(Any[x]), df)
 function hcat!(x, df::DataFrame)
-    throw(ArgumentError("x must be AbstaractVector or AbstractDataFrame"))
+    throw(ArgumentError("x must be AbstractVector or AbstractDataFrame"))
 end
 function hcat!(df::DataFrame, x)
-    throw(ArgumentError("x must be AbstaractVector or AbstractDataFrame"))
+    throw(ArgumentError("x must be AbstractVector or AbstractDataFrame"))
 end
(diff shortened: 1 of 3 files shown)


src/dataframe/dataframe.jl
```diff
@@ -697,15 +697,24 @@ end
...
 end
+
+# definition required to avoid hcat! ambiguity
+function hcat!(df1::DataFrame, df2::DataFrame)
+    invoke(hcat!, Tuple{DataFrame, AbstractDataFrame}, df1, df2)
+end
+
 hcat!(df::DataFrame, x::AbstractVector) = hcat!(df, DataFrame(Any[x]))
+hcat!(x::AbstractVector, df::DataFrame) = hcat!(DataFrame(Any[x]), df)
+hcat!(x, df::DataFrame) = throw(ArgumentError("x must be a vector or a data frame"))
```
nalimilan: I'd be more precise and say `AbstractVector` or `AbstractDataFrame`.

test/cat.jl
```diff
@@ -62,6 +62,15 @@ module TestCat
         @test df[2] == collect(1:10)
     end
 
+    @testset "hcat ::Vectors" begin
+        df = DataFrame()
+        df2 = hcat(DataFrames.CategoricalVector{Union{Int, Missing}}(1:10), df)
```
nalimilan: Why the `DataFrames.` prefix?

Also test that the result contains a single column, and check its name, that doesn't hurt.

test/cat.jl
```diff
@@ -62,6 +62,15 @@ module TestCat
         @test df[2] == collect(1:10)
     end
 
+    @testset "hcat ::Vectors" begin
```
nalimilan: Either merge with previous block, or find a different name?

test/cat.jl
```diff
@@ -62,6 +62,15 @@ module TestCat
...
+        df = DataFrame()
+        df2 = hcat(DataFrames.CategoricalVector{Union{Int, Missing}}(1:10), df)
+        @test df2[1] == collect(1:10)
+        df3 = hcat(11:20, df2)
+        @test df3[1] == collect(11:20)
+        @test_throws ArgumentError hcat("a", df)
```
nalimilan: This test should live in the previous block (or merge them).

src/dataframe/dataframe.jl

test/cat.jl
bkamins: There were earlier such duplicate blocks that is why I left it. Now I merge them all.

test/cat.jl
bkamins: prefix - removed.
additional checks added

test/cat.jl
bkamins: merged

src/dataframe/dataframe.jl
```diff
@@ -713,8 +713,12 @@ end
 
 hcat!(df::DataFrame, x::AbstractVector) = hcat!(df, DataFrame(Any[x]))
 hcat!(x::AbstractVector, df::DataFrame) = hcat!(DataFrame(Any[x]), df)
-hcat!(x, df::DataFrame) = throw(ArgumentError("x must be a vector or a data frame"))
-hcat!(df::DataFrame, x) = throw(ArgumentError("x must be a vector or a data frame"))
+function hcat!(x, df::DataFrame)
+    throw(ArgumentError("x must be AbstaractVector or AbstaractDataFrame"))
```
nalimilan: "Abstract"

test/cat.jl
```diff
@@ -55,13 +47,29 @@ module TestCat
...
         DataFrames.hcat!(df, CategoricalVector{Union{Int, Missing}}(1:10))
         @test df[1] == CategoricalVector(1:10)
         DataFrames.hcat!(df, collect(1:10))
         @test df[2] == collect(1:10)
-    end
 
+        df = DataFrame()
+        df2 = hcat(CategoricalVector{Union{Int, Missing}}(1:10), df)
+        @test df2[1] == collect(1:10)
+        @test names(df2) == ["x1"]
```
nalimilan: This will fail unless you use symbols AFAIK

src/dataframe/dataframe.jl

test/cat.jl
bkamins: sure. fixed

test/cat.jl
```diff
@@ -55,13 +47,29 @@ module TestCat
     end
 
     @testset "hcat ::Vectors" begin
+        df = DataFrame()
+        DataFrames.hcat!(df, CategoricalVector{Union{Int, Missing}}(1:10))
```
nalimilan: Am I missing something, or this duplicated below?

src/dataframe/dataframe.jl
```diff
@@ -697,15 +697,28 @@ end
...
 end
+
+# definition required to avoid hcat! ambiguity
+function hcat!(df1::DataFrame, df2::DataFrame)
+    invoke(hcat!, Tuple{DataFrame, AbstractDataFrame}, df1, df2)
+end
+
 hcat!(df::DataFrame, x::AbstractVector) = hcat!(df, DataFrame(Any[x]))
+hcat!(x::AbstractVector, df::DataFrame) = hcat!(DataFrame(Any[x]), df)
+function hcat!(x, df::DataFrame)
```
nalimilan: "Abstract". Same below.

bkamins: I would leave it as is. I thought it was a design decision to have `hcat!` handle `DataFrame` and that is why it is in *dataframe.jl*.
`hcat!` is not exported and those two cases actually will be called by those functions in *abstractdataframe.jl*:
```
Base.hcat(df::AbstractDataFrame, x) = hcat!(df[:, :], x)
Base.hcat(x, df::AbstractDataFrame) = hcat!(x, df[:, :])
```
which make sure to create a `DataFrame`. Any comment on it (I do not know why it was designed like this and I do not want to break the original intention)?

test/cat.jl
bkamins: It was old code that I simply moved. I have cleaned it up now.

src/dataframe/dataframe.jl
nalimilan: No, I meant you should fix "AbstaractVector" on the next line, not sure why it ended up on this line.

test/cat.jl
```diff
@@ -57,11 +49,24 @@ module TestCat
     @testset "hcat ::Vectors" begin
         df = DataFrame()
         DataFrames.hcat!(df, CategoricalVector{Union{Int, Missing}}(1:10))
+        @test df[1] == collect(1:10)
         @test df[1] == CategoricalVector(1:10)
```
nalimilan: This line is redundant with the previous one (or vice-versa).

test/cat.jl
```diff
@@ -57,11 +49,24 @@ module TestCat
...
-        DataFrames.hcat!(df, collect(1:10))
+        DataFrames.hcat!(df, 1:10)
```
nalimilan: As the name of the set says "::Vectors", you should keep `collect` here and below (or change the name).

src/dataframe/dataframe.jl
bkamins: ah - those typos. fixed

test/cat.jl
bkamins: it was not my code so I did not want to mess with it, fixed to `AbstractVector` test as this is what we really want to test

test/cat.jl
bkamins: fixed - these were old tests.

PR #1308, opened 2017-12-05
Deprecation warning when using insert! with duplicate column name

bkamins:
This is a small proposal of improvement and breaking but probably `insert!` was not used much as it is not documented. We can change `allow_duplicates` to `true` by default if we want a non-breaking change.

In general the idea is to make sure that no functions in DataFrames would allow to make duplicate column names by default.


src/dataframe/dataframe.jl
```diff
@@ -596,17 +596,22 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
...
+    end
+
     insert!(index(df), col_ind, name)
     insert!(df.columns, col_ind, item)
     df
 end
 
-function Base.insert!(df::DataFrame, col_ind::Int, item, name::Symbol)
-    insert!(df, col_ind, upgrade_scalar(df, item), name)
+function Base.insert!(df::DataFrame, col_ind::Int, item, name::Symbol; allow_duplicates=true)
```
cjprybol: `allow_duplicates=true` -> `allow_duplicates=false`?

bkamins: of course - I have changed my mind to propose a breaking change and forgotten to update it here. Fixed.

nalimilan:
Makes sense, but I'd rather keep the current behavior and print a warning using `Base.depwarn`. Then we can switch to the new behavior in the next breaking release.

bkamins:
@nalimilan I have never used `Base.depwarn` before. What would be a recommended way to do it here? Thanks for explaining.

nalimilan:
Just something like `Base.depwarn("your message", :insert!)`. There are examples in `base/deprecations.jl`. The point of that method is that warnings are printed with a stacktraces, and shown only once for each call site.

src/dataframe/dataframe.jl
```diff
@@ -596,17 +596,25 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
 
 Base.empty!(df::DataFrame) = (empty!(df.columns); empty!(index(df)); df)
 
-function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol)
+function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector,
+                      name::Symbol; allow_duplicates=true)
+    Base.depwarn(string("Currently insert! allows duplicate column names.",
```
nalimilan: The idea with deprecations is that they should only be printed when people use the old feature, not all the time. Here, it should only be printed when `allow_duplicates=true` and that a duplicate column name is encountered.

Actually, it's not clear we need to add the `allow_duplicates` argument: people should stop creating columns with duplicate names anyway so that their code continues to work in the next major release. And passing `allow_duplicates=false` isn't really essential, since it will just turn a deprecation warning into an error.

bkamins: > people should stop creating columns with duplicate names anyway

Agreed, but then I would remove `allow_duplicates` argument from `names!` also. If this is OK I will change this PR to only throw `depwarn` in `names!` and `insert!` now when duplicates are encountered (I understand this is the proper way to implement it) - this PR would be merged now. And then have a second PR removing `allow_duplicates` argument from `names!` and making `insert!` disallow duplicates that would be merged after the next DataFrames release.

Is this the way you see we should go?

bkamins: Yes. I guess there is no *real* use case. The only situation I have encountered is that when source data has duplicate columns - but it was always a bug in source data, and actually catching it was what was the proper way to go.

I guess we should treat `DataFrame` as `Dict` if indexed by column name. This is an important invariant. So duplicates should not be allowed.

I will adjust this PR accordingly and make this next PR on top of it to be merged later.

bkamins:
Changed according to the discussion. Actually in the end I came to the conclusion that `names!` can work as it works now:
* throw an error by default;
* generate new unique names wthen `allow_duplicates` is `true`.

For `insert!` I think that `allow_duplicates` will be not needed. Here we insert one column (not do mass rename as in `names!`) so it is enough to throw an error (this will be done in another PR when this one is marged).

nalimilan:
OK. Though I've realized tibbles still allow columns with duplicate column names on concatenation. Not sure why. What happens after this PR when calling e.g. `hcat(df, df)`?

bkamins:
This PR will not change anything.

We have two functionalities in Julia now:
* `hcat`: will generate new unique name (and do not warn);
* `merge!` will overwrite one column with the other (and also do not warn);
Both are reasonable given we document it.

If we wanted we could (this is something a minimal change):
* make `hcat` print a warning when duplicates are encountered (this is not a problem as `hcat` is not in place operation);
* make `merge!` either: (a) also throw a warning and go on overwriting the columns

A bigger change would be to add `allow_duplicates` to `hcat` and `merge!` that would work exactly the same as it works now for `names!` (which is a good solution - if `false` it is an error to have a duplicate, if `true` generate a new unique name).

But all this is a separate PR from `insert!` functionality.

nalimilan:
Hmm... I would find it kind of inconsistent to print a deprecation warning from `insert!`, but still allow duplicates in `hcat`. So we should find a global strategy and change everything from the same PR.

I'd say that if we continue to automatically add suffixes in some cases, all functions should support an `allow_duplicates` argument (which should arguably use a more explicit name like `make_unique`), and they should throw an error by default.

OTOH, `merge!` is OK as-is since it's part of the `Associative` interface. We could actually do the same for `insert!`: that would be [consistent with DataStructures](http://datastructuresjl.readthedocs.io/en/latest/sorted_containers.html?highlight=insert!`). Then we wouldn't have to bother about duplicate names. :-)

See also some discussions at https://github.com/tidyverse/dplyr/issues/2401 and related issues.

nalimilan:
Thanks for writing this down, we should always have this kind of general plan before making changes. Sounds very reasonable to me. I've posted a comment on the dplyr issue to get more information about their general rule, to be sure we're not missing a use case where duplicates are useful (they are generally not afraid of breaking compatibility with classic R where it makes sense).

bkamins:
👍
So let us wait for the response and then I can go on with an appropriate PR.

nalimilan:
OK, Hadley replied that they would like to raise an error everytime duplicate column names are found. So let's go ahead with your plan.

cjprybol:
To second what @nalimilan said, thank you for [the blueprint](https://github.com/JuliaData/DataFrames.jl/pull/1308#issuecomment-349752647) @bkamins. This clarifies a lot of confusion I had with these functions when working with DataTables. Once the functionality is implemented, we should add those descriptions along with code examples to the documentation.

bkamins:
OK. Tests pass so this should be ready for a review.
One thing I want to raise is what should be the default value for `make_unique`.
I have made it `false` except `Index`, where I kept it `true` as it was till now.

What is the consequence:
```
DataFrame(Any[[1,2], [1,2], [:x,:x])
```
goes through without an error. If we made `make_unique` equal to `false` by default this would be an error. The consequence is that we then probably should add `make_unique` keyword argument to `DataFrame`, `join` etc. or otherwise they would throw errors on duplicates.

I see the following base options (some variants of them are possible - any opinion on what would be best?):
1. make `make_unique` equal to `true` by default in every function (after seeing what happens I would opt for this - users should then be warned that by default DataFrames tries to add columns generating new names, but in some functions - `insert!`, `hcat` and `names!` this default behavior can be switched to a safe approach when an error is thrown on duplicate);
2. remove `make_unique` keyword argument (old `allow_duplicates`) altogether and always dynamically generate new names for variables (a user should independently check before an operation if there are duplicates in names).
3. keep the solution as it is implemented now (it is `false` by default except `Index` where it is `true` by default) - then `insert!`, `hcat` and `names!` are strict by default and throw an error and all other functions generate unique duplicate names by default (a bit inconsistent but maybe a good compromise);
4. make `make_unique` equal to `false` by default (it is the safest approach but very breaking and probably should be followed by adding `make_unique` keyword argument to some more functions).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -129,8 +129,12 @@ names!(df, [:a, :b, :a], allow_duplicates=true)  # renames second :a to :a_1
 ```
 
 """
-function names!(df::AbstractDataFrame, vals; allow_duplicates=false)
-    names!(index(df), vals; allow_duplicates=allow_duplicates)
+function names!(df::AbstractDataFrame, vals; allow_duplicates=false, make_unique=false)
```
nalimilan: While you're at it, add `::Bool`.

src/dataframe/dataframe.jl
```diff
@@ -596,10 +596,34 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
 
 Base.empty!(df::DataFrame) = (empty!(df.columns); empty!(index(df)); df)
 
-function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol)
+# TODO: add docstring of make_unique after 0.4 release of DataFrames
```
nalimilan: 0.4? And why not document `make_unique` immediately? It's also fine to avoid documenting deprecated arguments.

src/dataframe/dataframe.jl
```diff
@@ -609,6 +633,31 @@ function Base.insert!(df::DataFrame, col_ind::Int, item, name::Symbol)
     insert!(df, col_ind, upgrade_scalar(df, item), name)
 end
 
+"""
+Merge `DataFrame`s
```
nalimilan: It would be nice to give more details about what this does, and in particular to mention `join`, because people coming from R will expect `merge` to be the same as `join`.

src/dataframe/dataframe.jl
```diff
@@ -596,10 +596,34 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
...
 Base.empty!(df::DataFrame) = (empty!(df.columns); empty!(index(df)); df)
 
-function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol)
+# TODO: add docstring of make_unique after 0.4 release of DataFrames
+function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol;
+                      make_unique::Bool=false)
     0 < col_ind <= ncol(df) + 1 || throw(BoundsError())
     size(df, 1) == length(item) || size(df, 1) == 0 || error("number of rows does not match")
 
+    if name in _names(df)
```
nalimilan: I wonder whether there isn't a more efficient way of checking this by using the `Index` type, which stores a dict mapping column names to their indices.

src/dataframe/dataframe.jl
```diff
@@ -698,7 +747,16 @@ end
 ##############################################################################
 
 # hcat! for 2 arguments, only a vector or a data frame is allowed
-function hcat!(df1::DataFrame, df2::AbstractDataFrame)
+function hcat!(df1::DataFrame, df2::AbstractDataFrame; make_unique::Bool=false)
+    common = intersect(_names(df1), _names(df2))
+    if !make_unique && length(common) > 0
```
nalimilan: Shouldn't `make_unique` be passed when inserting columns? Else a warning will be printed downstream?

src/dataframe/dataframe.jl
```diff
@@ -698,7 +747,16 @@ end
...
+        # TODO: after 0.4 release of DataFrames remove depwarn and uncomment ArgumentError below
+        Base.depwarn("Inserting duplicate column names is deprecated.", :hcat!)
+        # msg = """Duplicate variable names $(common).
+        #      Pass make_unique=true to make them unique using a suffix automatically."""
+        # throw(ArgumentError(msg))
+
```
nalimilan: Remove empty line.

test/dataframe.jl
```diff
@@ -106,6 +106,14 @@ module TestDataFrame
...
     @test df[:newcol] == ["a", "b"]
 
+    @test insert!(df, 1, ["a1", "b1"], :newcol, make_unique=true) == df
+    @test names(df) == [:newcol_1, :newcol, :a, :b]
+    @test df[:a] == [1, 2]
+    @test df[:b] == [3.0, 4.0]
+    @test df[:newcol] == ["a", "b"]
+    @test df[:newcol_1] == ["a1", "b1"]
+
+
```
nalimilan: A single empty line is enough.

nalimilan:
I'd print deprecation warnings everywhere we currently silently make column names unique, adding a `make_unique` keyword argument to all of these functions which will default to `false` after the deprecation period.

BTW, maybe we should call the argument `makeunique`? We should choose a convention and stick to it, and it seems that the trend is to get rid of underscores (e.g. we have `skipmissing`).

bkamins:
OK - I will go this way with the PR.

bkamins:
The problem is that `DataFrame` constructor:

    function DataFrame(; kwargs...)

uses keyword arguments to give names to columns so we would have to redefine it as:

    function DataFrame(makeunique::Bool=false; kwargs...)

so it would be a different style (positional rather than keyword argument) than any other function using `makeunique`. What do you think about it?

nalimilan:
Argh. No, a positional argument would be really weird. If you define `makeunique` as a keyword argument in the first position, I guess it will take precedence? That means you won't be able to create columns called `makeunique`, but I'm not sure that's a serious issue in practice...

bkamins:
It will not take precedence, e.g.:
```
julia> f(;a=true,b...) = println((a,b))

julia> f(b=1,a=10)
(10, Any[(:b, 1)])
```
The consequence will be exactly as you say - that it is impossible to create `makeunique` column.

The other approach could be to leave `DataFrame(makeunique::Bool=false; kwargs...)` without the possibility of `makeunique` and always make it assume `makeunique=false` (this constructor is probably only used when hardcoding the columns so it should not be problematic?).

nalimilan:
Yeah, it's still possible to use the other constructor taking a vector of names if you want to pass them programmatically. Let's just deprecate duplicate names without a new argument and see whether people complain. It's easy to add the argument if it's needed.

src/abstractdataframe/abstractdataframe.jl

src/dataframe/dataframe.jl
bkamins: What release number should it be then? I will try to add documentations (although this is massive and I might miss something).

src/dataframe/dataframe.jl
bkamins: fixed: `name in keys(index(df))` is faster

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl
bkamins: changed. now it will as `add_names` is downstream and I have to make it accept `makeunique` anyway given other comments.

src/dataframe/dataframe.jl

test/dataframe.jl

src/dataframe/dataframe.jl
nalimilan: I wouldn't mention the name of the version here, it's easy to find out the history from git anyway.

src/dataframe/dataframe.jl
bkamins: actually `haskey(df, name)` is probably best

bkamins:
The current state of this PR should cover all the relevant functions in DataFrames by adding `makeunique`. In some functions, like `stack`, `melt`, `unstack` I did not add this parameter as they are higher level. Appropriate annotations in comments are added.

bkamins:
Also `getindex` on `DataFrames` after depreciation period will throw an error in cases like `df[[1,1,1,1]]` (currently it deduplicated column names).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -172,6 +177,9 @@ rename(f::Function, df::AbstractDataFrame)
 
 * `::AbstractDataFrame` : the updated result
 
+New names are processed sequentially. A new name must not exist in the `DataFrame`
```
nalimilan: "must not already exist"?

src/abstractdataframe/reshape.jl
```diff
@@ -84,7 +84,7 @@ function stack(df::AbstractDataFrame, measure_vars::Vector{Int},
     DataFrame(Any[repeat(_names(df)[measure_vars], inner=nrow(df)),   # variable
                   vcat([df[c] for c in measure_vars]...),             # value
                   [repeat(df[c], outer=N) for c in id_vars]...],      # id_var columns
-              cnames)
+              cnames) # TODO update this comment later: now it allows duplicate names, but after deprecation it will not
```
nalimilan: Which comment? Or do you mean "update this line"?

src/dataframe/dataframe.jl
```diff
@@ -596,10 +605,33 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
 
 Base.empty!(df::DataFrame) = (empty!(df.columns); empty!(index(df)); df)
 
-function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol)
+# TODO: add docstring of makeunique after 0.4 release of DataFrames
```
nalimilan: As noted elsewhere, the docstring should be added now.

src/dataframe/dataframe.jl
```diff
@@ -596,10 +605,33 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
...
+                nn = Symbol("$(name)_$k")
+                if !haskey(df, nn)
+                    name = nn
+                    break
+                end
+                k += 1
+            end
+        else
+            # TODO: after 0.4 release of DataFrames remove depwarn and uncomment ArgumentError below
+            Base.depwarn("Inserting duplicate column name is deprecated.", :insert!)
```
nalimilan: The message should already recommend using `makeunique=true`. Also remove mention to release 0.4.

src/groupeddataframe/grouping.jl
```diff
@@ -365,7 +365,7 @@ function _makeheaders(fs::Vector{T}, cn::Vector{Symbol}) where T<:Function
 end
 
 function _aggregate(d::AbstractDataFrame, fs::Vector{T}, headers::Vector{Symbol}, sort::Bool=false) where T<:Function
-    res = DataFrame(Any[vcat(f(d[i])) for f in fs for i in 1:size(d, 2)], headers)
+    res = DataFrame(Any[vcat(f(d[i])) for f in fs for i in 1:size(d, 2)], headers) # after deprecation period this will disallow duplicate names
```
nalimilan: I don't think this kind of comment is useful. On the contrary, we could forget to remove it, and it would be confusing.

src/other/index.jl
```diff
@@ -22,11 +24,16 @@ Base.isequal(x::Index, y::Index) = isequal(x.lookup, y.lookup) && isequal(x.name
 # Imported in DataFrames.jl for compatibility across Julia 0.4 and 0.5
 Base.:(==)(x::Index, y::Index) = isequal(x, y)
 
-function names!(x::Index, nms::Vector{Symbol}; allow_duplicates=false)
+# TODO: after deprecation period remove allow_duplicates
+function names!(x::Index, nms::Vector{Symbol}; allow_duplicates=false, makeunique::Bool=false)
+    if allow_duplicates
+        Base.depwarn("Keyword allow_duplicates is deprecated. Use makeunique.", :names!)
```
nalimilan: "Keyword argument". Same below.

src/other/index.jl
```diff
@@ -131,6 +142,15 @@ function add_names(ind::Index, add_ind::Index)
         name = u[i]
         in(name, seen) ? push!(dups, i) : push!(seen, name)
     end
+    if length(dups) > 0
+        if makeunique
+            Base.depwarn("Keyword makeunique will be false by default.", :add_names)
```
nalimilan: `makeunique` should default to `false` here, and the deprecation should only be printed when it's `false`. Else people have no way to silence these warnings. Same below.

src/other/utils.jl
```diff
@@ -49,7 +49,8 @@ function makeidentifier(s::AbstractString)
     return String(take!(res))
 end
 
-function make_unique(names::Vector{Symbol}; allow_duplicates=true)
+# TODO: change makeunique to false after deprecation period
+function _makeunique(names::Vector{Symbol}; makeunique::Bool=true)
```
nalimilan: Why rename the function?

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/reshape.jl
bkamins: I meant the comment "now it allows duplicate names".
I do not want to change the code, but leave a comment for someone reading the source in the future that `stack` will not de-duplicate names of variables.

This relates to the following case:
```
julia> df = DataFrame(variable='a':'c', x=rand(3), y=rand(3))
3×3 DataFrames.DataFrame
│ Row │ variable │ x        │ y        │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 'a'      │ 0.858719 │ 0.961591 │
│ 2   │ 'b'      │ 0.648647 │ 0.205988 │
│ 3   │ 'c'      │ 0.588156 │ 0.166744 │

julia> stack(df)
6×3 DataFrames.DataFrame
│ Row │ variable │ value    │ variable_1 │
├─────┼──────────┼──────────┼────────────┤
│ 1   │ x        │ 0.858719 │ 'a'        │
│ 2   │ x        │ 0.648647 │ 'b'        │
│ 3   │ x        │ 0.588156 │ 'c'        │
│ 4   │ y        │ 0.961591 │ 'a'        │
│ 5   │ y        │ 0.205988 │ 'b'        │
│ 6   │ y        │ 0.166744 │ 'c'        │
```
After deprecation period this will throw an error (and it should I think).

src/dataframe/dataframe.jl
bkamins: added

src/dataframe/dataframe.jl

src/groupeddataframe/grouping.jl
bkamins: removed

src/other/index.jl

src/other/index.jl
bkamins: fixed - although I was not sure if what I did is what you have intended.

src/other/utils.jl
bkamins: I renamed it when we had `make_unique` keyword. Now it is `makeunique` so there would be no name clash.

I understood that `_` signals that the function is for internal use only. So based on my understanding `_makeunique` should be an OK name. I can change it if there are some guidelines for this.

src/groupeddataframe/grouping.jl
bkamins: actually I changed it to target comment - OK?

src/abstractdataframe/reshape.jl
bkamins: I have changed it to a target comment

src/other/index.jl
nalimilan: Yeah, but the message should be changed to something like `Duplicate variable names are deprecated: pass makeunique=true to add a suffix automatically`.

src/other/utils.jl
nalimilan: I'd better keep it as is, as there's no strong reason to change it and I don't think we systematically use the `_` prefix for internal variables.

src/groupeddataframe/grouping.jl
nalimilan: Not sure, it's kind of weird to say this here as there is no reason to think there would be duplicate names. Why would the reader care? I'd remove these comments and keep them only in place where there's a special reason to mention this fact.

src/dataframe/dataframe.jl
```diff
@@ -596,19 +605,118 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
 
 Base.empty!(df::DataFrame) = (empty!(df.columns); empty!(index(df)); df)
 
-function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol)
+"""
+Inserts a column into a `DataFrame` in place.
```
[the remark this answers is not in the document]

src/other/index.jl

src/other/utils.jl
bkamins: fixed

src/groupeddataframe/grouping.jl

src/dataframe/dataframe.jl
```diff
@@ -698,36 +806,48 @@ end
 ##############################################################################
 
 # hcat! for 2 arguments, only a vector or a data frame is allowed
-function hcat!(df1::DataFrame, df2::AbstractDataFrame)
-    u = add_names(index(df1), index(df2))
+# TODO: after deprecation period change all to makeunique::Bool=false
+function hcat!(df1::DataFrame, df2::AbstractDataFrame; makeunique::Bool=true)
```
nalimilan: Shouldn't we set `makeunique=false` immediately? Else, now deprecation warning will ever be printed, right?

src/other/utils.jl
```diff
@@ -58,11 +59,15 @@ function make_unique(names::Vector{Symbol}; allow_duplicates=true)
...
     end
 
-    if !allow_duplicates && length(dups) > 0
-        d = unique(names[dups])
-        msg = """Duplicate variable names: $d.
-                 Pass allow_duplicates=true to make them unique using a suffix automatically."""
-        throw(ArgumentError(msg))
+    if length(dups) > 0
+        if makeunique
+            Base.depwarn("Keyword makeunique will be false by default.", :_makeunique)
```
nalimilan: "keyword argument"

bkamins: fixed

src/dataframe/dataframe.jl
bkamins: Now I get what you are getting at :).
I will set `makeunique` to `false` everywhere but it will change the code so that it has no consequences for now except printing deprecation. Right?

nalimilan: Right!

bkamins:
I wanted to add one more thing for consideration.
Given code
```
x = DataFrame(a=1:3)
y = x[[1,1]]
```
will fail after deprecation period (now it prints a warning).
Maybe in `getindex` we should allow duplicates by default (possibly with printing a warning that column aliases are being created?)?

nalimilan:
> Maybe in getindex we should allow duplicates by default (possibly with printing a warning that column aliases are being created?)?

I don't like printing warnings, especially when there's no way to silence them. I find it better to throw an error and provide a way to avoid it. In the present case, I admit it would be a bit weird to require using `getindex` with a keyword argument. But maybe better wait and see whether somebody has a use case for this.

bkamins:
Thank you for the fixes. As for `getindex` I agree and that is why I have implemented it as is, but I wanted to be explicit. In summary I understand the reasons are:
* if we do not allow it now and start allowing it later it is not breaking;
* as such an operation creates an alias I do not see much use for it (and I would rather discourage it as error prone - in base R it creates a copy so it is less risky); if someone wants an alias it is always possible with something like `df[:y] = df[:x]`, but this is explicit.

src/abstractdataframe/io.jl
```diff
@@ -307,5 +307,5 @@ DataFrame(sink, sch::Data.Schema, ::Type{S}, append::Bool;
     append!(sink.columns[col], column)
 end
     
-
-Data.close!(df::DataFrameStream) = DataFrame(collect(Any, df.columns), Symbol.(df.header))
+Data.close!(df::DataFrameStream, makeunique::Bool=false) =
```
nalimilan: Do we really need this? This implements a particular DataStreams interface which is supposed to be completely generic, so the caller should not have to adapt to `DataFrame` specificities. We should follow the rule adopted by DataStreams in general with regard to duplicate column names.

@quinnj Should we automatically deduplicate column names, or throw an error?

src/dataframe/dataframe.jl
```diff
@@ -596,19 +600,118 @@ Base.setindex!(df::DataFrame, x::Void, col_ind::Int) = delete!(df, col_ind)
...
+```
+
+For every column `c` with name `n` in `others` sequentially perform `df[n] = c`.
+This behavior is identical with how `merge!` works for any `Associative` type.
+Use `join` if you want to join two `DataFrame`s.
+
+**Arguments**
+
+* `df` : the DataFrame to merge into
+* `others` : `AbstractDataFrame`s to be merged into `df`
```
nalimilan: `makeunique` should be mentioned.

src/abstractdataframe/join.jl
```diff
@@ -272,10 +274,10 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner)
+                   kind::Symbol = :inner, makeunique::Bool=false)
```
nalimilan: Why is this needed? By definition, joins match columns with identical names, so that shouldn't be a problem? Anyway it should be in the docstring.

src/abstractdataframe/io.jl
bkamins: I wanted to be explicit rather than implicitly assume one type of behavior without thinking about it. Of course if @quinnj has a clear opinion let us implement it.

src/abstractdataframe/join.jl
bkamins: It is needed. Join matches on `on` keyword argument not on columns having identical names. And we can have columns with identical names on which we do not join.
See the example as it works now:
```
julia> x = DataFrame(rand(3,3))
3×3 DataFrames.DataFrame
│ Row │ x1       │ x2       │ x3       │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 0.97283  │ 0.200961 │ 0.591387 │
│ 2   │ 0.391169 │ 0.532865 │ 0.630848 │
│ 3   │ 0.090204 │ 0.320481 │ 0.775537 │

julia> y = DataFrame(rand(3,3))
3×3 DataFrames.DataFrame
│ Row │ x1       │ x2       │ x3       │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 0.470765 │ 0.468833 │ 0.199083 │
│ 2   │ 0.303053 │ 0.992384 │ 0.140139 │
│ 3   │ 0.284372 │ 0.816792 │ 0.675941 │

julia> x[:id] = 1:3
1:3

julia> y[:id] = 1:3
1:3

julia> join(x, y, on=:id)
3×7 DataFrames.DataFrame
│ Row │ x1       │ x2       │ x3       │ id │ x1_1     │ x2_1     │ x3_1     │
├─────┼──────────┼──────────┼──────────┼────┼──────────┼──────────┼──────────┤
│ 1   │ 0.97283  │ 0.200961 │ 0.591387 │ 1  │ 0.470765 │ 0.468833 │ 0.199083 │
│ 2   │ 0.391169 │ 0.532865 │ 0.630848 │ 2  │ 0.303053 │ 0.992384 │ 0.140139 │
│ 3   │ 0.090204 │ 0.320481 │ 0.775537 │ 3  │ 0.284372 │ 0.816792 │ 0.675941 │
```
I will improve the docstring.

src/dataframe/dataframe.jl

src/abstractdataframe/io.jl
nalimilan: OK, I've just tested, and it turns out CSV.jl will happily return duplicated column names. So let's always pass `makeunique=true` to `DataFrames`, and remove the keyword argument to `close!`.

src/dataframe/dataframe.jl
bkamins: Sorry - `merge` does not have `makeunique` argument. It follows `Associative` interface, i.e. it overwrites duplicate columns. I have added an additional explanation.

bkamins: But I have added information about `makeunique` to `DataFrame` docstring.

nalimilan: Funny, not sure how I got confused.

src/dataframe/dataframe.jl
```diff
@@ -692,6 +701,8 @@ merge!(df::DataFrame, others::AbstractDataFrame...)
 ```
 
 For every column `c` with name `n` in `others` sequentially perform `df[n] = c`.
+In particular, if there are duplicate column names present in `df` and `others`
+the last encountered columnwill be retained.
```
nalimilan: "columnwill"? :-)

src/abstractdataframe/join.jl
```diff
@@ -241,6 +241,11 @@ Join two `DataFrame` objects
   - `:cross` : a full Cartesian product of the key combinations; every
     row of `df1` is matched with every row of `df2`
 
+* `makeunique` : how to handle columns with duplicate names other than `on` in joined tables:
```
nalimilan: Can you make this more concise? A bullet list sounds too much for a mere boolean. Why not use a description similar to that used for `names!`?

Also, I know I used it, but I don't think "deduplicate" is a great term for an official documentation. "Make unique" sounds better.

src/abstractdataframe/io.jl
bkamins: OK

src/abstractdataframe/join.jl
bkamins: OK - fixing

src/abstractdataframe/join.jl
```diff
@@ -241,10 +241,11 @@ Join two `DataFrame` objects
   - `:cross` : a full Cartesian product of the key combinations; every
     row of `df1` is matched with every row of `df2`
 
-* `makeunique` : how to handle columns with duplicate names other than `on` in joined tables:
 
-  - `false` : throw an error if duplicate column names are present
-  - `true` : duplicate column names in `df2` will be deduplicated by adding a suffix
+* `makeunique` : if `false` (the default), an error will be raised
+  if duplicate names are found in columns not joined on;
+  if `true`, duplicate names will be suffixed with `_i`
```
nalimilan: Am I right that `i` can only be `1` in this case, since we don't allow duplicated column names in either of the sources?

bkamins: No - it can be a bigger value. See:
```
julia> x = DataFrame(rand(3,3))
3×3 DataFrames.DataFrame
│ Row │ x1       │ x2       │ x3       │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 0.791145 │ 0.193656 │ 0.238548 │
│ 2   │ 0.756677 │ 0.195111 │ 0.929581 │
│ 3   │ 0.736347 │ 0.234449 │ 0.499932 │

julia> y = DataFrame(rand(3,3))
3×3 DataFrames.DataFrame
│ Row │ x1       │ x2        │ x3       │
├─────┼──────────┼───────────┼──────────┤
│ 1   │ 0.220774 │ 0.0541193 │ 0.735224 │
│ 2   │ 0.147731 │ 0.104057  │ 0.144468 │
│ 3   │ 0.282832 │ 0.0247473 │ 0.188086 │

julia> x[:id] = 1:3
1:3

julia> y[:id] = 1:3
1:3
julia> names!(y, [:x1, :x1, :x1, :id], allow_duplicates=true)
3×4 DataFrames.DataFrame
│ Row │ x1       │ x1_1      │ x1_2     │ id │
├─────┼──────────┼───────────┼──────────┼────┤
│ 1   │ 0.220774 │ 0.0541193 │ 0.735224 │ 1  │
│ 2   │ 0.147731 │ 0.104057  │ 0.144468 │ 2  │
│ 3   │ 0.282832 │ 0.0247473 │ 0.188086 │ 3  │

julia> join(x, y, on=:id)
3×7 DataFrames.DataFrame
│ Row │ x1       │ x2       │ x3       │ id │ x1_3     │ x1_1      │ x1_2     │
├─────┼──────────┼──────────┼──────────┼────┼──────────┼───────────┼──────────┤
│ 1   │ 0.791145 │ 0.193656 │ 0.238548 │ 1  │ 0.220774 │ 0.0541193 │ 0.735224 │
│ 2   │ 0.756677 │ 0.195111 │ 0.929581 │ 2  │ 0.147731 │ 0.104057  │ 0.144468 │
│ 3   │ 0.736347 │ 0.234449 │ 0.499932 │ 3  │ 0.282832 │ 0.0247473 │ 0.188086 │
```

nalimilan: Ah, right. It would make sense to have a `left_suffix`/`right_suffix` argument as in R to choose a deduplication suffix to add to columns from the LHS and RHS. We would need to look at what other software does first.

If we improve this before the next release, we could remove `makeunique` without a deprecation.

nalimilan:
Thanks!

bkamins:
@quinnj, @nalimilan  following #1495 the reasoning is that I want to use DataFrames.jl in production, not only in interactive mode.

In interactive mode probably the default of `makeunique` to `true` is OK. In production I would hate this and prefer error than silently renaming the columns as the code downstream might rely on column names.

Anyway there are two points:
1. We have `makeunique` parameter that allows both modes and we make sure that it is consistently handled in all DataFrames.jl ecosystem;
2. The question is what should be the default value of `makeunique` - now we think it should be `false` (so duplicates will throw an error) - this is identical behavior to `NamedTuple` - it throws an error on duplicate in constructor. But I am open to the discussion to change the default to `true`. Personally I even in REPL prefer `makeunique` default to `false`, but maybe for other people it is too annoying.

quinnj:
I lean towards having `makeunique=true` be the default; my basis for that would be the assumption that most users are using Julia interactively, from the REPL or scripts, and it's much more convenient to not have things throw errors all over the place. When running a production app, it's also easier to add in the single keyword argument `makeunique=false` and add `try-catch` blocks as necessary.

bkamins:
I can see the benefits of this decision and I will not oppose going this way.

If we decide to go this way then after #1495 is merged I can:
* cleanup whole code base to support it;
* can remove deprecation warnings (as they are not needed);
* make it clear in documentation and my tutorial what are the consequences (so that people are aware to use `makeunique` in scripts)

nalimilan:
I still like the `makeunique=false` default. As I've noted above, even dplyr is leaning towards throwing errors in these cases (https://github.com/tidyverse/dplyr/issues/2401#issuecomment-349919904). Even when working interactively, if you combine two datasets which have column `:x`, and then do `df[:x]`, you can get totally incorrect results if you expected to access the column from the second data frame. For scientific work in particular, preventing this kind of mistake is essential. And in general throwing an error early helps debugging.

@quinnj Do you have a precise scenario in mind where it would be annoying?

PR #1309, opened 2017-12-05
Refactor unstack

bkamins:
Changes:
1) allow multiple columns in rowkeys
2) remove incorrect and slower implementation for single column in rowkey

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -150,8 +150,12 @@ melt(df::AbstractDataFrame; variable_name::Symbol=:variable, value_name::Symbol=
 Unstacks a DataFrame; convert from a long to wide format
 
 ```julia
-unstack(df::AbstractDataFrame, rowkey, colkey, value)
-unstack(df::AbstractDataFrame, colkey, value)
+unstack(df::AbstractDataFrame, rowkeys::Union{Symbol, Integer},
+        colkey::Union{Symbol, Integer}, value::Union{Symbol, Integer})
+unstack(df::AbstractDataFrame, rowkeys::AbstractVector{<:Union{Symbol, Integer}},
+        colkey::Union{Symbol, Integer}, value::Union{Symbol, Integer})
+unstack(df::AbstractDataFrame, colkey::Union{Symbol, Integer},
+        value::Union{Symbol, Integer})
 unstack(df::AbstractDataFrame)
 ```
 
@@ -159,7 +163,7 @@ unstack(df::AbstractDataFrame)
 
 * `df` : the AbstractDataFrame to be unstacked
 
-* `rowkey` : the column with a unique key for each row, if not given,
+* `rowkeys` : the column(s) with a unique key for each row, if not given,
   find a key by grouping on anything not a `colkey` or `value`
 
 * `colkey` : the column holding the column names in wide format,
@@ -171,6 +175,10 @@ unstack(df::AbstractDataFrame)
 
 * `::DataFrame` : the wide-format DataFrame
 
+If `colkey` contains `missing` values then they will be skipped and a warning will be printed.
+
+If combination of `rowkeys` and `colkey` contains duplicate entries then last `value` will
+be retained and a warning will be printed.
 
 ### Examples
 
@@ -185,39 +193,70 @@ long = stack(wide)
 wide0 = unstack(long)
 wide1 = unstack(long, :variable, :value)
 wide2 = unstack(long, :id, :variable, :value)
+wide3 = unstack(long, [:id, :a], :variable, :value)
 ```
 Note that there are some differences between the widened results above.
-
 """
 function unstack(df::AbstractDataFrame, rowkey::Int, colkey::Int, value::Int)
-    # `rowkey` integer indicating which column to place along rows
-    # `colkey` integer indicating which column to place along column headers
-    # `value` integer indicating which column has values
-    refkeycol = CategoricalArray{Union{eltype(df[rowkey]), Missing}}(df[rowkey])
-    # make sure we report only levels actually present in rowkey column
-    # this is consistent with how the other unstack method based on groupby works
+    refkeycol = categorical(df[rowkey])
     droplevels!(refkeycol)
+    keycol = categorical(df[colkey])
+    droplevels!(keycol)
     valuecol = df[value]
-    keycol = CategoricalArray{Union{eltype(df[colkey]), Missing}}(df[colkey])
+    _unstack(df, rowkey, colkey, value, keycol, valuecol, refkeycol)
+end
+
+function _unstack(df::AbstractDataFrame, rowkey::Int,
+                  colkey::Int, value::Int, keycol, valuecol, refkeycol)
     Nrow = length(refkeycol.pool)
     Ncol = length(keycol.pool)
-    payload = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
-    nowarning = true
+    unstacked_val = [similar_missing(valuecol, Nrow) for i in 1:Ncol]
+    hadmissing = false # have we encountered missing in refkeycol
+    mask_filled = falses(Nrow+1, Ncol) # has a given [col,row] entry been filled?
+    warned_dup = false # have we already printed duplicate entries warning?
+    warned_missing = false # have we already printed missing in keycol warning?
+    keycol_order = Vector{Int}(CategoricalArrays.order(keycol.pool))
+    refkeycol_order = Vector{Int}(CategoricalArrays.order(refkeycol.pool))
     for k in 1:nrow(df)
-        j = Int(CategoricalArrays.order(keycol.pool)[keycol.refs[k]])
-        i = Int(CategoricalArrays.order(refkeycol.pool)[refkeycol.refs[k]])
-        if i > 0 && j > 0
-            if nowarning && !ismissing(payload[j][i])
-                warn("Duplicate entries in unstack.")
-                nowarning = false
+        kref = keycol.refs[k]
+        if kref <= 0 # we have found missing in colkey
+            if !warned_missing
+                warn("Missing value in variable $(_names(df)[colkey]) at row $k. Skipping.")
+                warned_missing = true
             end
-            payload[j][i]  = valuecol[k]
+            continue # skip processing it
         end
+        j = keycol_order[kref]
+        refkref = refkeycol.refs[k]
+        if refkref <= 0 # we have found missing in rowkey
+            if !hadmissing # if it is the first time we have to add a new row
+                hadmissing = true
+                # we use the fact that missing is greater than anything
+                for i in eachindex(unstacked_val)
+                    push!(unstacked_val[i], missing)
+                end
+            end
+            i = length(unstacked_val[1])
+        else
+            i = refkeycol_order[refkref]
+        end
+        if !warned_dup && mask_filled[i, j]
+            warn("Duplicate entries in unstack at row $k for key "*
+                 "$(refkeycol[k]) and variable $(keycol[k]).")
+            warned_dup = true
+        end
+        unstacked_val[j][i] = valuecol[k]
+        mask_filled[i, j] = true
     end
     levs = levels(refkeycol)
-    col = similar_missing(df[rowkey], length(levs))
-    insert!(payload, 1, copy!(col, levs), _names(df)[rowkey])
+    # we have to handle a case with missings in refkeycol as levs will skip missing
+    col = similar(df[rowkey], length(levs) + hadmissing)
+    copy!(col, levs)
+    hadmissing && (col[end] = missing)
+    df2 = DataFrame(unstacked_val, map(Symbol, levels(keycol)))
+    insert!(df2, 1, col, _names(df)[rowkey])
 end
+
 unstack(df::AbstractDataFrame, rowkey::ColumnIndex,
         colkey::ColumnIndex, value::ColumnIndex) =
     unstack(df, index(df)[rowkey], index(df)[colkey], index(df)[value])
@@ -226,34 +265,61 @@ unstack(df::AbstractDataFrame, rowkey::ColumnIndex,
 unstack(df::AbstractDataFrame, colkey::ColumnIndex, value::ColumnIndex) =
     unstack(df, index(df)[colkey], index(df)[value])
 
-function unstack(df::AbstractDataFrame, colkey::Int, value::Int)
-    # group on anything not a key or value:
-    rowkeys = setdiff(_names(df), _names(df)[[colkey, value]])
+# group on anything not a key or value
+unstack(df::AbstractDataFrame, colkey::Int, value::Int) =
+    unstack(df, setdiff(_names(df), _names(df)[[colkey, value]]), colkey, value)
+
+unstack(df::AbstractDataFrame, rowkeys, colkey::ColumnIndex, value::ColumnIndex) =
+    unstack(df, rowkeys, index(df)[colkey], index(df)[value])
+
+unstack(df::AbstractDataFrame, rowkeys::AbstractVector{<:Real}, colkey::Int, value::Int) =
+    unstack(df, names(df)[rowkeys], colkey, value)
+
+function unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Symbol}, colkey::Int, value::Int)
     length(rowkeys) == 0 && throw(ArgumentError("No key column found"))
+    length(rowkeys) == 1 && return unstack(df, rowkeys[1], colkey, value)
     g = groupby(df, rowkeys, sort=true)
+    keycol = categorical(df[colkey])
+    droplevels!(keycol)
+    valuecol = df[value]
+    _unstack(df, rowkeys, colkey, value, keycol, valuecol, g)
+end
+
+function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Symbol},
+                  colkey::Int, value::Int, keycol, valuecol, g)
     groupidxs = [g.idx[g.starts[i]:g.ends[i]] for i in 1:length(g.starts)]
     rowkey = zeros(Int, size(df, 1))
     for i in 1:length(groupidxs)
         rowkey[groupidxs[i]] = i
     end
-    keycol = CategoricalArray{Union{eltype(df[colkey]), Missing}}(df[colkey])
-    valuecol = df[value]
-    df1 = allowmissing!(df[g.idx[g.starts], g.cols], g.cols)
+    df1 = df[g.idx[g.starts], g.cols]
     Nrow = length(g)
     Ncol = length(levels(keycol))
-    df2 = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
-    nowarning = true
+    unstacked_val = [similar_missing(valuecol, Nrow) for i in 1:Ncol]
+    mask_filled = falses(Nrow, Ncol)
+    warned_dup = false
+    warned_missing = false
+    keycol_order = Vector{Int}(CategoricalArrays.order(keycol.pool))
     for k in 1:nrow(df)
-        j = Int(CategoricalArrays.order(keycol.pool)[keycol.refs[k]])
-        i = rowkey[k]
-        if i > 0 && j > 0
-            if nowarning && !ismissing(df2[j][i])
-                warn("Duplicate entries in unstack at row $k.")
-                nowarning = false
+        kref = keycol.refs[k]
+        if kref <= 0
+            if !warned_missing
+                warn("Missing value in variable $(_names(df)[colkey]) at row $k. Skipping.")
+                warned_missing = true
             end
-            df2[j][i]  = valuecol[k]
+            continue
         end
+        j = keycol_order[kref]
+        i = rowkey[k]
+        if !warned_dup && mask_filled[i, j]
+            warn("Duplicate entries in unstack at row $k for key "*
+                 "$(tuple((df[1,s] for s in rowkeys)...)) and variable $(keycol[k]).")
+            warned_dup = true
+        end
+        unstacked_val[j][i] = valuecol[k]
+        mask_filled[i, j] = true
     end
+    df2 = DataFrame(unstacked_val, map(Symbol, levels(keycol)))
     hcat(df1, df2)
 end
 
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -212,7 +212,7 @@ function _unstack(df::AbstractDataFrame, rowkey::Int,
     Ncol = length(keycol.pool)
     unstacked_val = [similar_missing(valuecol, Nrow) for i in 1:Ncol]
     hadmissing = false # have we encountered missing in refkeycol
-    mask_filled = falses(Nrow+1, Ncol) # has a given [col,row] entry been filled?
+    mask_filled = falses(Nrow+1, Ncol) # has a given [row,col] entry been filled?
     warned_dup = false # have we already printed duplicate entries warning?
     warned_missing = false # have we already printed missing in keycol warning?
     keycol_order = Vector{Int}(CategoricalArrays.order(keycol.pool))
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -299,12 +299,15 @@ module TestDataFrame
 
     #Check the output of unstack
     df = DataFrame(Fish = CategoricalArray{Union{String, Missing}}(["Bob", "Bob", "Batman", "Batman"]),
-                   Key = Union{String, Missing}["Mass", "Color", "Mass", "Color"],
+                   Key = CategoricalArray{Union{String, Missing}}(["Mass", "Color", "Mass", "Color"]),
                    Value = Union{String, Missing}["12 g", "Red", "18 g", "Grey"])
     # Check that reordering levels does not confuse unstack
     levels!(df[1], ["XXX", "Bob", "Batman"])
+    levels!(df[2], ["YYY", "Color", "Mass"])
     #Unstack specifying a row column
     df2 = unstack(df, :Fish, :Key, :Value)
+    @test levels(df[1]) == ["XXX", "Bob", "Batman"] # make sure we did not mess df[1] levels
+    @test levels(df[2]) == ["YYY", "Color", "Mass"] # make sure we did not mess df[2] levels
     #Unstack without specifying a row column
     df3 = unstack(df, :Key, :Value)
     #The expected output, XXX level should be dropped as it has no rows with this key
@@ -321,6 +324,52 @@ module TestDataFrame
     #This changes the expected result
     df4[1,:Mass] = missing
     @test df2 ≅ df4
+
+    #The same as above but without CategoricalArray
+    df = DataFrame(Fish = ["Bob", "Bob", "Batman", "Batman"],
+                   Key = ["Mass", "Color", "Mass", "Color"],
+                   Value = ["12 g", "Red", "18 g", "Grey"])
+    #Unstack specifying a row column
+    df2 = unstack(df, :Fish, :Key, :Value)
+    #Unstack without specifying a row column
+    df3 = unstack(df, :Key, :Value)
+    #The expected output, XXX level should be dropped as it has no rows with this key
+    df4 = DataFrame(Fish = ["Batman", "Bob"],
+                    Color = ["Grey", "Red"],
+                    Mass = ["18 g", "12 g"])
+    @test df2 ≅ df4
+    @test typeof(df2[:Fish]) <: Vector{String}
+    # first column stays as CategoricalArray in df3
+    @test df3 == df4
+    #Make sure unstack works with missing values at the start of the value column
+    allowmissing!(df, :Value)
+    df[1,:Value] = missing
+    df2 = unstack(df, :Fish, :Key, :Value)
+    #This changes the expected result
+    allowmissing!(df4, :Mass)
+    df4[2,:Mass] = missing
+    @test df2 ≅ df4
+
+    # test empty set of grouping variables
+    @test_throws ArgumentError unstack(df, Int[], :Key, :Value)
+    @test_throws ArgumentError unstack(df, Symbol[], :Key, :Value)
+    @test_throws KeyError unstack(stack(DataFrame(rand(10, 10))))
+
+    # test missing value in grouping variable
+    mdf = DataFrame(id=[missing,1,2,3], a=1:4, b=1:4)
+    @test unstack(melt(mdf, :id))[1:3,:] == sort(mdf)[1:3,:]
+    @test unstack(melt(mdf, :id))[:,2:3] == sort(mdf)[:,2:3]
+
+    # test more than one grouping column
+    wide = DataFrame(id = 1:12,
+                     a  = repeat([1:3;], inner = [4]),
+                     b  = repeat([1:4;], inner = [3]),
+                     c  = randn(12),
+                     d  = randn(12))
+
+    long = stack(wide)
+    wide3 = unstack(long, [:id, :a], :variable, :value)
+    @test wide3 == wide[:, [1, 2, 4, 5]]
 
     df = DataFrame(A = 1:10, B = 'A':'J')
     @test !(df[:,:] === df)
@@ -350,7 +399,8 @@ module TestDataFrame
         @test udf == DataFrame(Any[Union{Int, Missing}[1, 2], Union{Int, Missing}[1, 5],
                                    Union{Int, Missing}[2, 6], Union{Int, Missing}[3, 7],
                                    Union{Int, Missing}[4, 8]], [:id, :a, :b, :c, :d])
-        @test all(isa.(udf.columns, Vector{Union{Int, Missing}}))
+        @test isa(udf[1], Vector{Int})
+        @test all(isa.(udf.columns[2:end], Vector{Union{Int, Missing}}))
         df = DataFrame(Any[categorical(repeat(1:2, inner=4)),
                            categorical(repeat('a':'d', outer=2)), categorical(1:8)],
                        [:id, :variable, :value])
@@ -359,18 +409,22 @@ module TestDataFrame
         @test udf == DataFrame(Any[Union{Int, Missing}[1, 2], Union{Int, Missing}[1, 5],
                                    Union{Int, Missing}[2, 6], Union{Int, Missing}[3, 7],
                                    Union{Int, Missing}[4, 8]], [:id, :a, :b, :c, :d])
-        @test all(isa.(udf.columns, CategoricalVector{Union{Int, Missing}}))
+        @test isa(udf[1], CategoricalVector{Int})
+        @test all(isa.(udf.columns[2:end], CategoricalVector{Union{Int, Missing}}))
     end
 
     @testset "duplicate entries in unstack warnings" begin
-        df = DataFrame(id=Union{Int, Missing}[1, 2, 1, 2], variable=["a", "b", "a", "b"], value=[3, 4, 5, 6])
+        df = DataFrame(id=Union{Int, Missing}[1, 2, 1, 2],
+                       id2=Union{Int, Missing}[1, 2, 1, 2], 
+                       variable=["a", "b", "a", "b"], value=[3, 4, 5, 6])
         @static if VERSION >= v"0.6.0-dev.1980"
-            @test_warn "Duplicate entries in unstack." unstack(df, :id, :variable, :value)
-            @test_warn "Duplicate entries in unstack at row 3." unstack(df, :variable, :value)
+            @test_warn "Duplicate entries in unstack at row 3 for key 1 and variable a." unstack(df, :id, :variable, :value)
+            @test_warn "Duplicate entries in unstack at row 3 for key (1, 1) and variable a." unstack(df, :variable, :value)
         end
         a = unstack(df, :id, :variable, :value)
+        @test a ≅ DataFrame(id = [1, 2], a = [5, missing], b = [missing, 6])
         b = unstack(df, :variable, :value)
-        @test a ≅ b ≅ DataFrame(id = [1, 2], a = [5, missing], b = [missing, 6])
+        @test b ≅ DataFrame(id = [1, 2], id2 = [1, 2], a = [5, missing], b = [missing, 6])
 
         df = DataFrame(id=1:2, variable=["a", "b"], value=3:4)
         @static if VERSION >= v"0.6.0-dev.1980"
@@ -380,6 +434,42 @@ module TestDataFrame
         a = unstack(df, :id, :variable, :value)
         b = unstack(df, :variable, :value)
         @test a ≅ b ≅ DataFrame(id = [1, 2], a = [3, missing], b = [missing, 4])
+
+        df = DataFrame(variable=["x", "x"], value=[missing, missing], id=[1,1])
+        @test_warn "Duplicate entries in unstack at row 2 for key 1 and variable x." unstack(df, :variable, :value)
+        @test_warn "Duplicate entries in unstack at row 2 for key 1 and variable x." unstack(df)
+    end
+
+    @testset "missing values in colkey" begin
+        df = DataFrame(id=[1, 1, 1, missing, missing, missing, 2, 2, 2],
+                       variable=["a", "b", missing, "a", "b", "missing", "a", "b", "missing"],
+                       value=[missing, 2.0, 3.0, 4.0, 5.0, missing, 7.0, missing, 9.0])
+        @test_warn "Missing value in variable variable at row 3. Skipping." unstack(df)
+        udf = unstack(df)
+        @test names(udf) == [:id, :a, :b, :missing]
+        @test udf[:missing] ≅ [missing, 9.0, missing]
+        df = DataFrame(id=[1, 1, 1, missing, missing, missing, 2, 2, 2],
+                       id2=[1, 1, 1, missing, missing, missing, 2, 2, 2],
+                       variable=["a", "b", missing, "a", "b", "missing", "a", "b", "missing"],
+                       value=[missing, 2.0, 3.0, 4.0, 5.0, missing, 7.0, missing, 9.0])
+        @test_warn "Missing value in variable variable at row 3. Skipping." unstack(df, 3, 4)
+        udf = unstack(df, 3, 4)
+        @test names(udf) == [:id, :id2, :a, :b, :missing]
+        @test udf[:missing] ≅ [missing, 9.0, missing]
+    end
+
+    @testset "stack-unstack correctness" begin
+        x = DataFrame(rand(100, 50))
+        x[:id] = [1:99; missing]
+        x[:id2] = string.("a", x[:id])
+        x[:s] = [i % 2 == 0 ? randstring() : missing for i in 1:100]
+        allowmissing!(x, :x1)
+        x[1, :x1] = missing
+        y = melt(x, [:id, :id2])
+        z = unstack(y)
+        @test all(isequal(z[n], x[n]) for n in names(z))
+        z = unstack(y, :variable, :value)
+        @test all(isequal(z[n], x[n]) for n in names(x))
     end
 
     @testset "rename" begin
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 Missings 0.2.3
-CategoricalArrays 0.3.0
+CategoricalArrays 0.3.2
 StatsBase 0.11.0
 SortingAlgorithms
 Reexport


bkamins:
@nalimilan @cjprybol
Just for a sake of completeness of refactoring `unstack` let me mention that currently we do not handle the following case 100% correct:
```
julia> y = DataFrame(variable=["x","x"], value=[missing,missing], id=[1,1])
2×3 DataFrames.DataFrame
│ Row │ variable │ value   │ id │
├─────┼──────────┼─────────┼────┤
│ 1   │ x        │ missing │ 1  │
│ 2   │ x        │ missing │ 1  │

julia> unstack(y)
1×2 DataFrames.DataFrame
│ Row │ id │ x       │
├─────┼────┼─────────┤
│ 1   │ 1  │ missing │
```
There is a duplicate in `y` that is not reported because it is `missing` (and `missing` is used to indicate yet unfilled data).

The second thing is that `unstack` makes all columns accept missing even if we are not sure that we have to (this means that stack-unstack procedure changes types of columns).

The solution would be to use a temporary data structure that would hold binary indicators if a position is filled (so we have no duplicate `missing` issue). Later those indicators can be also used to decide if we have to make a column accept `Missing` or not.

But I am not sure if you would feel it is worth the effort as it will introduce some memory and computational overhead.

bkamins:
tests fail because the fix of `allowmissing!` is not tagged yet so we have rerun CI after the update of dependencies.

src/abstractdataframe/reshape.jl
```diff
@@ -159,7 +159,7 @@ unstack(df::AbstractDataFrame)
 
 * `df` : the AbstractDataFrame to be unstacked
 
-* `rowkey` : the column with a unique key for each row, if not given,
+* `rowkeys` : the column or columns with a unique key for each row, if not given,
```
nalimilan: Just "column(s)".

src/abstractdataframe/reshape.jl
```diff
@@ -150,7 +150,7 @@ melt(df::AbstractDataFrame; variable_name::Symbol=:variable, value_name::Symbol=
 Unstacks a DataFrame; convert from a long to wide format
 
 ```julia
-unstack(df::AbstractDataFrame, rowkey, colkey, value)
+unstack(df::AbstractDataFrame, rowkeys, colkey, value)
```
nalimilan: I would be useful to mention accepted types too. I guess you could have one line with `::Union{Symbol, Integer}`, and another one with `::AbstractVector{<:Union{Symbol, Integer}}`.

src/abstractdataframe/reshape.jl
```diff
@@ -185,50 +185,32 @@ long = stack(wide)
...
-            payload[j][i]  = valuecol[k]
-        end
-    end
-    levs = levels(refkeycol)
-    col = similar_missing(df[rowkey], length(levs))
-    insert!(payload, 1, copy!(col, levs), _names(df)[rowkey])
-end
+unstack(df::AbstractDataFrame, rowkey::Int, colkey::Int, value::Int) =
+    unstack(df, [rowkey], colkey, value)
 unstack(df::AbstractDataFrame, rowkey::ColumnIndex,
```
nalimilan: Add a line break since you do the same below.

src/abstractdataframe/reshape.jl
```diff
@@ -185,50 +185,32 @@ long = stack(wide)
 wide0 = unstack(long)
 wide1 = unstack(long, :variable, :value)
 wide2 = unstack(long, :id, :variable, :value)
+wide3 = unstack(long, [:id, :a], :variable, :value)
 ```
 Note that there are some differences between the widened results above.
 
 """
-function unstack(df::AbstractDataFrame, rowkey::Int, colkey::Int, value::Int)
```
nalimilan: Can you explain why you find it natural that the `groupby` approach is more efficient than the approach used here? AFAIK it should be quite faster, especially if the original column is a `CategoricalArray` (since we don't have an optimized `groupby` implementation for `CategoricalArray` at the moment, and it's not trivial to do).

Since the code exists, I'm reluctant to remove it since we have sometimes found out that some trivial optimizations could be applied which made a dramatic difference.

nalimilan:
> There is a duplicate in y that is not reported because it is missing (and missing is used to indicate yet unfilled data).
> 
> The second thing is that unstack makes all columns accept missing even if we are not sure that we have to (this means that stack-unstack procedure changes types of columns).
> 
> The solution would be to use a temporary data structure that would hold binary indicators if a position is filled (so we have no duplicate missing issue). Later those indicators can be also used to decide if we have to make a column accept Missing or not.
> 
> But I am not sure if you would feel it is worth the effort as it will introduce some memory and computational overhead.

Why not, I guess the overhead could be kept relatively low. With compiler improvements we should be able to convert an `Array{Union{Missing, T}` to an `Array{T}` very efficiently, and keeping track of whether an entry is missing or not shouldn't be too costly, if you're willing to do the work. Among other things, `ismissing(payload[j][i])` is probably quite slow currently (waiting for more compiler optimizations), and could be much faster if we used a boolean array.

It would make sense to check what other implementations do, though.

test/dataframe.jl
```diff
@@ -319,6 +319,27 @@ module TestDataFrame
     df4[1,:Mass] = missing
     @test df2 ≅ df4
 
+    # test empty set of grouping variables
+    @testthrows ArgumentError unstack(df, Int[], :Key, :Value)
```
nalimilan: `@test_throws`.

src/abstractdataframe/reshape.jl
bkamins: See the benchmark:
```
julia> x = DataFrame(rand(1000, 10));

julia> x[:id] = 1:1000;

julia> y = stack(x);

julia> z = copy(y);

julia> categorical!(z, :id);

julia> eltypes(y)
3-element Array{Type,1}:
 Symbol
 Float64
 Int64

julia> eltypes(z)
3-element Array{Type,1}:
 Symbol
 Float64
 CategoricalArrays.CategoricalValue{Int64,UInt32}

julia> @benchmark unstack(y)
BenchmarkTools.Trial:
  memory estimate:  2.01 MiB
  allocs estimate:  78752
  --------------
  minimum time:     7.898 ms (0.00% GC)
  median time:      8.700 ms (0.00% GC)
  mean time:        8.998 ms (3.72% GC)
  maximum time:     59.264 ms (84.15% GC)
  --------------
  samples:          556
  evals/sample:     1

julia> @benchmark unstack(y, :variable, :value)
BenchmarkTools.Trial:
  memory estimate:  1.85 MiB
  allocs estimate:  57636
  --------------
  minimum time:     6.008 ms (0.00% GC)
  median time:      6.769 ms (0.00% GC)
  mean time:        6.965 ms (3.64% GC)
  maximum time:     58.657 ms (87.87% GC)
  --------------
  samples:          717
  evals/sample:     1

julia> @benchmark unstack(z)
BenchmarkTools.Trial:
  memory estimate:  1.65 MiB
  allocs estimate:  68695
  --------------
  minimum time:     10.947 ms (0.00% GC)
  median time:      11.326 ms (0.00% GC)
  mean time:        11.677 ms (2.53% GC)
  maximum time:     64.046 ms (81.62% GC)
  --------------
  samples:          428
  evals/sample:     1

julia> @benchmark unstack(z, :variable, :value)
BenchmarkTools.Trial:
  memory estimate:  1.92 MiB
  allocs estimate:  56270
  --------------
  minimum time:     9.071 ms (0.00% GC)
  median time:      9.890 ms (0.00% GC)
  mean time:        10.179 ms (2.94% GC)
  maximum time:     65.845 ms (84.51% GC)
  --------------
  samples:          491
  evals/sample:     1
```

The difference is because of the lines:
```
i = Int(CategoricalArrays.order(refkeycol.pool)[refkeycol.refs[k]]) # more expensive
```
and
```
i = rowkey[k] # cheaper
```

Do you see any area for optimization? If single-column version is not significantly faster I would remove it as now:
1. we have a duplicate implementation of the same feature (which increases package complexity and maintenance cost);
2. the single-column version has a bug when unstacking on id-column that has missing values.

nalimilan: Yeah, the second line is cheaper, but it only comes after a call to `groupby` which should take much more time. So it looks like something is going wrong with the other method. Have you tried profiling it?

I agree with you that removing duplicated methods is generally a good idea. But I don't want to do that until we are certain to understand the situation, and here it looks like something doesn't work as I would expect it to. For example, we have recently realized that building a `CategoricalArray` was much slower than it should have been. If we could fix a similar issue here, lots of places would benefit from the fix (even if in the end we remove the method).

A possibility here is that constructing a `CategoricalArray` is relatively slow since IDs are all unique, and `CategoricalArray` isn't efficient for this kind of data. Or maybe the compiler is not able to move the `CategoricalArrays.order(keycol.pool)` call outside of the loop, i.e. we should manually store the order in a variable?

src/abstractdataframe/reshape.jl
bkamins: fixed (although a bit messy)

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl

test/dataframe.jl

bkamins:
I have fixed correct handling of missing. I have decided to retain that all columns accept missing - this would introduce performance overhead to control for it and I think it is not crucially problematic.

src/abstractdataframe/reshape.jl
bkamins: I did the profiling - I will post it below.

But actually observe having all unique IDs is a point of working `stack`/`unstack` (i.e. they are unique after `unstack` but are not unique after `stack`).

bkamins:
First part - timing of new vs old code (new code uses `mask` Boolean test):
```
# Old code
julia> @benchmark unstack(y)
  memory estimate:  2.01 MiB
  mean time:        8.176 ms (3.61% GC)
julia> @benchmark unstack(y, :variable, :value)
  memory estimate:  1.85 MiB
  mean time:        6.263 ms (3.67% GC)
julia> @benchmark unstack(z)
  memory estimate:  1.65 MiB
  mean time:        11.022 ms (2.53% GC)
julia> @benchmark unstack(z, :variable, :value)
  memory estimate:  1.92 MiB
  mean time:        10.190 ms (2.86% GC)

# The proposed code with mask array
julia> @benchmark unstack(y)
  memory estimate:  1.85 MiB
  mean time:        6.001 ms (4.54% GC)
julia> @benchmark unstack(y, :variable, :value)
  memory estimate:  1.85 MiB
  mean time:        5.891 ms (3.69% GC)
julia> @benchmark unstack(z)
  memory estimate:  1.92 MiB
  mean time:        9.367 ms (3.08% GC)
julia> @benchmark unstack(z, :variable, :value)
  memory estimate:  1.92 MiB
  mean time:        9.391 ms (2.90% GC)
```

bkamins:
I performed the profiling and actually it seems that creation of `groupby`:
```
g = groupby(df, setdiff(_names(df), _names(df)[[colkey, value]]), sort=true)
```
is faster than:
```
keycol = CategoricalArray{Union{eltype(df[colkey]), Missing}}(df[colkey])
refkeycol = CategoricalArray{Union{eltype(df[rowkey]), Missing}}(df[rowkey])
```
if they have to be run. This means that `unstack` without `groupby` would be faster if `colkey` and `rowkey` column would be categorial already.

In consequence the question is if we want to make `stack` turn `:variable` column into `Categorical` (it will make `stack` slower but `unstack` faster).

nalimilan:
> I performed the profiling and actually it seems that creation of groupby:
> 
> g = groupby(df, setdiff(_names(df), _names(df)[[colkey, value]]), sort=true)
> 
> is faster than:
> 
> keycol = CategoricalArray{Union{eltype(df[colkey]), Missing}}(df[colkey])
> refkeycol = CategoricalArray{Union{eltype(df[rowkey]), Missing}}(df[rowkey])
> 
> if they have to be run.

Why not `g = groupby(df, rowkeys, sort=true)` as in the code? Though in this case it should be equivalent?

Also, it's not fair to include `keycol`, since it needs to be created in both cases, right?

> This means that unstack without groupby would be faster if colkey and rowkey column would be categorial already.

But is that indeed the case in benchmarks? In my quick tests, it wasn't, which is really weird.

Anyway, if `groupby` than creating a `CategoricalArray`, that means that hashing a column first, and only then perform dictionary lookups based on these hashes is faster that hashing and doing the lookups on the fly at the same time. Indeed that's the main difference I can think of between the two approaches.

bkamins:
I was testing on the implementation before my changes (that is the reason for code differences). I have used @profile - those are the lines that came up (and actually single `Categorical` creation was slower).

However, it is weird indeed - I will investigate more into it. What bothers me is:
```
using BenchmarkTools

df = DataFrame(rand(1000,10))
df[:id] = string.("a", 1:1000)
sdf = stack(df)
sdf = sdf[randperm(nrow(sdf)), :]

f1(s) = groupby(s, [:id], sort=true)
f2(s) = CategoricalArray{Union{eltype(s[:id]), Missing}}(s[:id])

@benchmark f1($sdf)
@benchmark f2($sdf)
```

gives:
```
julia> @benchmark f1($sdf)
BenchmarkTools.Trial:
  memory estimate:  601.81 KiB
  allocs estimate:  12377
  --------------
  minimum time:     2.283 ms (0.00% GC)
  median time:      2.422 ms (0.00% GC)
  mean time:        2.528 ms (3.56% GC)
  maximum time:     55.176 ms (94.73% GC)
  --------------
  samples:          1975
  evals/sample:     1

julia> @benchmark f2($sdf)
BenchmarkTools.Trial:
  memory estimate:  756.02 KiB
  allocs estimate:  15795
  --------------
  minimum time:     1.448 ms (0.00% GC)
  median time:      1.549 ms (0.00% GC)
  mean time:        1.693 ms (7.64% GC)
  maximum time:     55.446 ms (97.12% GC)
  --------------
  samples:          2940
  evals/sample:     1
```
which is in line with your intuition and against the results I report above.

bkamins:
Actually running `@code_warntype` on both implementations is all red. I will check if we can improve something here.

bkamins:
OK. I have it. We will leave separate implementations for single and multiple columns.
The problem is solved in a standard way by barrier function:
```
function unstack(df::AbstractDataFrame, rowkey::Int, colkey::Int, value::Int)
    refkeycol = CategoricalArray{Union{eltype(df[rowkey]), Missing}}(df[rowkey])
    keycol = CategoricalArray{Union{eltype(df[colkey]), Missing}}(df[colkey])
    _unstack(df, rowkey, colkey, value, refkeycol, keycol)
end
```
(it is a fast writeup - I will push a clean solution when I have it but roughly it is 2x speedup on the test I report above)

src/abstractdataframe/reshape.jl
```diff
@@ -150,16 +150,20 @@ melt(df::AbstractDataFrame; variable_name::Symbol=:variable, value_name::Symbol=
 Unstacks a DataFrame; convert from a long to wide format
 
 ```julia
-unstack(df::AbstractDataFrame, rowkey, colkey, value)
-unstack(df::AbstractDataFrame, colkey, value)
+unstack(df::AbstractDataFrame, rowkeys::Union{Symbol, Integer},
```
cjprybol: `Union{Symbol, Integer}` -> `ColumnIndex`. Currently, `ColumnIndex` is actually `Union{Symbol, Real}`, but I've had a PR on my to-do list for a while to change it to `Union{Symbol, Integer}`. Let's stay consistent and hopefully we can further restrict the types of all column-accepting functions at once by just changing `ColumnIndex` to `Union{Symbol, Integer}`

nalimilan: The problem is that it's not exported currently, and even if it was I'm not sure it's a good idea to use it in docs since it's less explicit than `Union`. Ideally we could interpolate it using `$ColumnIndex`.

bkamins: I have `Union{Symbol, Integer}` everywhere in my local repo added. I will probably push this during the weekend when I finish tuning the performance.

cjprybol: If we use 4-space indent rather than backticks the variable interpolation works, just tried it out

bkamins:
In this commit I have concentrated on functionality (some testing appreciated):
1) the speedup of execution should be around 2x;
2) correctly and consistently handle `missing` in `:id` column - in `groupby` and `CategoricalArray` implementation;
3) correctly handle a situation with `missing` in `:variable` column (now it throws a warning - I can change it to an error if you feel it would be a better approach).

bkamins:
Here are the benchmarks (I have cut the output to leave what is important - timing is given after `unstack` is compiled for given arguments).
The conclusion is that we are 2x faster with this implementation and 0.7 is faster than 0.6 (and keep in mind that the new implementation does more work as it has to handle missing values in `rowkeys` and `colkey`).

Data setup code:
```
x = DataFrame(rand(10000, 1000))
x[:id] = 1:10000
x[:id2] = string.("a", x[:id])
x[:s] = [randstring() for i in 1:10000]
y = melt(x, [:id, :id2])
```

Summary of benchmark results:
```
Julia 0.6

OLD
julia> @time unstack(y);
 10.974269 seconds (121.05 M allocations: 2.254 GiB, 22.68% gc time)
julia> @time unstack(y, :variable, :value);
 10.412161 seconds (73.60 M allocations: 1.859 GiB, 30.93% gc time)

 categorical!(y, [:id, :id2])
julia> @time unstack(y);
 10.393829 seconds (111.06 M allocations: 1.919 GiB, 22.76% gc time)
julia> @time unstack(y, :variable, :value);
 11.879770 seconds (83.64 M allocations: 2.010 GiB, 31.39% gc time)

NEW 
julia> @time unstack(y);
  4.848957 seconds (34.63 M allocations: 1.006 GiB, 42.45% gc time)
julia> @time unstack(y, :variable, :value);
  8.192743 seconds (24.48 M allocations: 1.128 GiB, 56.92% gc time)

 categorical!(y, [:id, :id2])
julia> @time unstack(y);
  4.182978 seconds (24.64 M allocations: 687.415 MiB, 46.12% gc time)
julia> @time unstack(y, :variable, :value);
  9.793470 seconds (34.52 M allocations: 1.280 GiB, 55.00% gc time)

Julia 0.7

OLD
julia> @time unstack(y);
 11.376244 seconds (121.07 M allocations: 2.254 GiB, 11.02% gc time)
julia> @time unstack(y, :variable, :value);
  9.536922 seconds (73.59 M allocations: 1.859 GiB, 22.01% gc time)

 categorical!(y, [:id, :id2])
julia> @time unstack(y);
 10.165981 seconds (111.06 M allocations: 1.919 GiB, 11.03% gc time)
julia> @time unstack(y, :variable, :value);
 10.698634 seconds (83.64 M allocations: 2.010 GiB, 22.48% gc time)

NEW
julia> @time unstack(y);
  4.928271 seconds (34.56 M allocations: 1.005 GiB, 41.12% gc time)
julia>  @time unstack(y, :variable, :value);
  5.809069 seconds (24.48 M allocations: 1.128 GiB, 42.01% gc time)

 categorical!(y, [:id, :id2])
julia> @time unstack(y);
  3.382329 seconds (24.55 M allocations: 686.091 MiB, 28.77% gc time)
julia> @time unstack(y, :variable, :value);
  6.865709 seconds (34.53 M allocations: 1.280 GiB, 40.00% gc time)
```

src/abstractdataframe/reshape.jl
```diff
@@ -171,6 +175,10 @@ unstack(df::AbstractDataFrame)
 
 * `::DataFrame` : the wide-format DataFrame
 
+If `colkey` contains `missing` values then they will be skipped and warning will be printed.
```
nalimilan: "a warning". Same below.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
-    nowarning = true
+    df2 = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
+    _unstack(df, rowkeys, colkey, value, keycol, valuecol, df2, refkeycol)
+end
+
+function _unstack(df::AbstractDataFrame, rowkeys::Int,
+                  colkey::Int, value::Int, keycol, valuecol, df2, refkeycol)
+    Nrow = length(refkeycol.pool)
+    Ncol = length(keycol.pool)
+    hadmissing = false # have we encounered missing in refkeycol
```
nalimilan: "encountered".

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+    mask_filled = falses(Ncol, Nrow) # has a given [col,row] entry been filled?
+    nowarning = true # do we print duplicate entries warning
+    nowarning2 = true # do we print missing in keycol
+    keycol_pool = Vector{Int}(CategoricalArrays.order(keycol.pool))
```
nalimilan: Better call this `keycol_order`. Same below.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+end
+
+function _unstack(df::AbstractDataFrame, rowkeys::Int,
+                  colkey::Int, value::Int, keycol, valuecol, df2, refkeycol)
+    Nrow = length(refkeycol.pool)
+    Ncol = length(keycol.pool)
+    hadmissing = false # have we encounered missing in refkeycol
+    mask_filled = falses(Ncol, Nrow) # has a given [col,row] entry been filled?
+    nowarning = true # do we print duplicate entries warning
+    nowarning2 = true # do we print missing in keycol
```
nalimilan: Better call this `nowarning_missing`, or maybe `warned_missing` (and revert it). Same for `nowarning`.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+    keycol_pool = Vector{Int}(CategoricalArrays.order(keycol.pool))
+    refkeycol_pool = Vector{Int}(CategoricalArrays.order(refkeycol.pool))
     for k in 1:nrow(df)
-        j = Int(CategoricalArrays.order(keycol.pool)[keycol.refs[k]])
-        i = Int(CategoricalArrays.order(refkeycol.pool)[refkeycol.refs[k]])
-        if i > 0 && j > 0
-            if nowarning && !ismissing(payload[j][i])
-                warn("Duplicate entries in unstack.")
-                nowarning = false
+        keycol_refs = keycol.refs[k]
```
nalimilan: Plural is weird since it's a single value. Same for `refkeycol_ref`. Could almost be `kref` and `rkref`.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+        if keycol_refs == 0 # we have found missing in colkey
+            if nowarning2
+              warn("Missing value in colkey variable at row $k. Skipping.")
```
nalimilan: Indentation issue. Would also be nice to print the name of the variable while you're at it.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+            nowarning = false
+        end
+        df2[j][i] = valuecol[k]
+        mask_filled[j, i] = true
     end
     levs = levels(refkeycol)
-    col = similar_missing(df[rowkey], length(levs))
-    insert!(payload, 1, copy!(col, levs), _names(df)[rowkey])
+    # we have to handle a case with missings in refkeycol as levs will skip missing
+    if hadmissing
```
nalimilan: It would be slightly more efficient to copy the levels into `col2` and set the missing value at the end, without creating a temporary array.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+    valuecol = df[value]
     Nrow = length(refkeycol.pool)
     Ncol = length(keycol.pool)
-    payload = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
-    nowarning = true
+    df2 = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
+    _unstack(df, rowkeys, colkey, value, keycol, valuecol, df2, refkeycol)
+end
+
+function _unstack(df::AbstractDataFrame, rowkeys::Int,
```
nalimilan: Singular `rowkey`.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+            continue # skip processing it
+        end
+        j = keycol_pool[keycol_refs]
+        refkeycol_refs = refkeycol.refs[k]
+        if refkeycol_refs == 0 # we have found missing in rowkeys
+            if !hadmissing # if it is the first time we have to add a new row
+                hadmissing = true
+                # we use the fact that missing is greater than anything
+                push!(df2, missings(ncol(df2)))
+                mask_filled = hcat(mask_filled, falses(ncol(df2)))
```
nalimilan: Better allocate the column for missing values unconditionally, even if it's not used. Also doesn't the data frame column need a name?

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+        if keycol_refs == 0 # we have found missing in colkey
+            if nowarning2
+              warn("Missing value in colkey variable at row $k. Skipping.")
+              nowarning2 = false
+            end
+            continue # skip processing it
+        end
+        j = keycol_pool[keycol_refs]
+        refkeycol_refs = refkeycol.refs[k]
+        if refkeycol_refs == 0 # we have found missing in rowkeys
```
nalimilan: Use `<= 0`, I have possible plans to use negative values to support multiple types of missing values.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+                push!(df2, missings(ncol(df2)))
+                mask_filled = hcat(mask_filled, falses(ncol(df2)))
             end
-            payload[j][i]  = valuecol[k]
+            i = nrow(df2)
+        else
+            i = refkeycol_pool[refkeycol_refs]
         end
+        if nowarning && mask_filled[j, i]
+            warn("Duplicate entries in unstack at row $k.")
```
nalimilan: Would also be nice to print the value and the name of the column.

src/abstractdataframe/reshape.jl
```diff
@@ -226,33 +266,59 @@ unstack(df::AbstractDataFrame, rowkey::ColumnIndex,
...
-    # group on anything not a key or value:
-    rowkeys = setdiff(_names(df), _names(df)[[colkey, value]])
+# group on anything not a key or value
+unstack(df::AbstractDataFrame, colkey::Int, value::Int) =
+    unstack(df, setdiff(_names(df), _names(df)[[colkey, value]]), colkey, value)
+
+unstack(df::AbstractDataFrame, rowkeys, colkey::ColumnIndex, value::ColumnIndex) =
+    unstack(df, rowkeys, index(df)[colkey], index(df)[value])
+
+unstack(df::AbstractDataFrame, rowkeys::AbstractVector{T}, colkey::Int, value::Int) where T<:Real =
```
nalimilan: Use `AbstractVector{<:Real}`.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,71 @@ long = stack(wide)
...
+    refkeycol = copy(CategoricalArray{Union{eltype(df[rowkeys]), Missing}}(df[rowkeys]))
     droplevels!(refkeycol)
-    valuecol = df[value]
     keycol = CategoricalArray{Union{eltype(df[colkey]), Missing}}(df[colkey])
+    valuecol = df[value]
     Nrow = length(refkeycol.pool)
     Ncol = length(keycol.pool)
-    payload = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
-    nowarning = true
+    df2 = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
```
nalimilan: It would be more natural to build the returned data frame inside `_unstack`.

test/dataframe.jl
```diff
@@ -377,6 +398,36 @@ module TestDataFrame
         a = unstack(df, :id, :variable, :value)
         b = unstack(df, :variable, :value)
         @test a ≅ b ≅ DataFrame(id = [1, 2], a = [3, missing], b = [missing, 4])
+
+        df = DataFrame(variable=["x","x"], value=[missing,missing], id=[1,1])
```
nalimilan: Use spaces around `==` and after commas for consistency.

bkamins:
> Great! Though, why is unstack(y) slightly faster than unstack(y, :variable, :value)?

Because `unstack(y)` uses single column `:id` (and does not use `groupby`) and `unstack(y, :variable, :value)` works on `[:id, :id2]` (and uses `groupby`).

> You should be able to make it even dramatically more efficient by passing a tuple of columns to _unstack rather than a DataFrame. See what I did at JuliaData/DataTables.jl#79.

This is essentially what I do. I pass `df` only to get names for rows. But I can create `df2` later (and work on an array when assigning values) - I will check if it helps.

bkamins:
@nalimilan good advice - delaying creation of `df2` is another 2x speedup.

nalimilan:
> This is essentially what I do. I pass df only to get names for rows. But I can create df2 later (and work on an array when assigning values) - I will check if it helps.

Oh, right, that's for `df2` that type unstability still hurts. Anyway it makes sense to create that data frame inside `_unstack`.

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl
bkamins: fixed (and improving performance as reported below)

src/abstractdataframe/reshape.jl
bkamins: fixed in all signatures assuming a single column (I could decide myself if it was better when developing the code so I am OK with this change).

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl
bkamins: fixed (actually I had similar names originally, but thought them to be less informative)

src/abstractdataframe/reshape.jl
bkamins: fixed both for `kref` and `refkref`.

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl
bkamins: fixed `mask_filled`. As for `push!` - no it adds a row not a column (now it is changed anyway as I do create a `DataFrame` later in the code).

src/abstractdataframe/reshape.jl
bkamins: I skip this. It gets messy as I would have to say something like

*duplicate value for index "something which can be multiple columns" and variable "something"*

I feel it is easy enough to to check the contents of the reported column anyway.

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl
bkamins: OK. Out of curiosity - why it is a preferred style?

nalimilan: I think because it's shorter, you can read it directly instead of looking for `T` at the end of the line, and it makes it clear `T` isn't used in the body.

bkamins: 👍

src/abstractdataframe/reshape.jl
nalimilan: Sure, you can check it manually, but I find it really makes the difference between software which works and software which is pleasant and efficient to work with. For example, we print the index and the dimensions of the indexed array with `BoundsError`, and it makes a big difference from R where you need to start the debugger to find it out.

Don't do it if that's too much work, but if that's possible it would be really nice, even if the output isn't completely pretty (e.g. it's fine to print a one-element tuple when there's only one column, though `join`'s `last` argument is very useful to format this properly).

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
-
 """
 function unstack(df::AbstractDataFrame, rowkey::Int, colkey::Int, value::Int)
-    # `rowkey` integer indicating which column to place along rows
-    # `colkey` integer indicating which column to place along column headers
-    # `value` integer indicating which column has values
-    refkeycol = CategoricalArray{Union{eltype(df[rowkey]), Missing}}(df[rowkey])
-    # make sure we report only levels actually present in rowkey column
-    # this is consistent with how the other unstack method based on groupby works
+    refkeycol = copy(CategoricalArray{Union{eltype(df[rowkey]), Missing}}(df[rowkey]))
```
nalimilan: I don't think you need `copy`, IIRC I ensured that `CategoricalArray` constructors always make a copy. BTW, that means that if you don't need a copy for `refkeycol`, better use `convert` to avoid it if the input is already a `CategoricalArray`.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
     droplevels!(refkeycol)
-    valuecol = df[value]
     keycol = CategoricalArray{Union{eltype(df[colkey]), Missing}}(df[colkey])
+    valuecol = df[value]
+    Nrow = length(refkeycol.pool)
+    Ncol = length(keycol.pool)
+    df2m = [similar_missing(valuecol, Nrow) for i in 1:Ncol]
```
nalimilan: Why create this here? Don't you have all the needed information inside `_unstack`?

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
+
+function _unstack(df::AbstractDataFrame, rowkey::Int,
+                  colkey::Int, value::Int, keycol, valuecol, df2m, refkeycol)
     Nrow = length(refkeycol.pool)
     Ncol = length(keycol.pool)
-    payload = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
-    nowarning = true
+    hadmissing = false # have we encountered missing in refkeycol
+    mask_filled = falses(Ncol, Nrow+1) # has a given [col,row] entry been filled?
+    warned_missing_1 = false # do we print duplicate entries warning
```
nalimilan: My suggestion was to call this `warned_dup` or something like that, to get rid of 1 and 2. :-)

And the comment would be "have we already printed..."

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
-        j = Int(CategoricalArrays.order(keycol.pool)[keycol.refs[k]])
-        i = Int(CategoricalArrays.order(refkeycol.pool)[refkeycol.refs[k]])
-        if i > 0 && j > 0
-            if nowarning && !ismissing(payload[j][i])
-                warn("Duplicate entries in unstack.")
-                nowarning = false
+        kref = keycol.refs[k]
+        if kref <= 0 # we have found missing in colkey
+            if !warned_missing_2
+                warn("Missing value in '$(_names(df)[colkey])' variable at row $k. Skipping.")
```
nalimilan: I've just checked, the convention used in the `DataFrame` constructor is to print column names without quotes. Maybe clearer with "variable X" (rather than "X variable").

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
+            continue # skip processing it
+        end
+        j = keycol_order[kref]
+        refkref = refkeycol.refs[k]
+        if refkref <= 0 # we have found missing in rowkey
+            if !hadmissing # if it is the first time we have to add a new row
+                hadmissing = true
+                # we use the fact that missing is greater than anything
+                for i in eachindex(df2m)
+                    push!(df2m[i], missing)
```
nalimilan: How common do we expect this to happen? Unless that reflects malformed data, we could allocate `Nrow+1` entries and resize to `Nrow` when creating `df2m`, so that this line does not have to make any copy.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
+                end
             end
-            payload[j][i]  = valuecol[k]
+            i = length(df2m[1])
+        else
+            i = refkeycol_order[refkref]
+        end
+        if (!warned_missing_1) && mask_filled[j, i]
```
nalimilan: No need for parentheses.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
+            warn("Duplicate entries in unstack at row $k.")
+            warned_missing_1 = true
         end
+        df2m[j][i] = valuecol[k]
+        mask_filled[j, i] = true
     end
     levs = levels(refkeycol)
+    # we have to handle a case with missings in refkeycol as levs will skip missing
     col = similar_missing(df[rowkey], length(levs))
```
nalimilan: Use `length(levs) + hadmissing` for the length, then you can do `hadmissing && col[end] = missing` below. That way you avoid a copy.

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,72 @@ long = stack(wide)
...
+            i = length(df2m[1])
+        else
+            i = refkeycol_order[refkref]
+        end
+        if (!warned_missing_1) && mask_filled[j, i]
+            warn("Duplicate entries in unstack at row $k.")
+            warned_missing_1 = true
         end
+        df2m[j][i] = valuecol[k]
+        mask_filled[j, i] = true
```
nalimilan: `[j, i]` is less natural than `[i, j]`, is this to optimize memory access patterns? Do we expect the data to be sorted (and therefore filled) by rows or by columns? Given the layout of data frames, it would be optimal to fill one column at a time. I admit I don't really understand whether that's the case currently or not.


We should also ensure that's consistent with what `stack` outputs.

test/dataframe.jl
```diff
@@ -377,6 +398,36 @@ module TestDataFrame
...
         b = unstack(df, :variable, :value)
         @test a ≅ b ≅ DataFrame(id = [1, 2], a = [3, missing], b = [missing, 4])
+
+        df = DataFrame(variable=["x", "x"], value=[missing, missing], id=[1,1])
+        @test_warn "Duplicate entries in unstack at row 2." unstack(df, :variable, :value)
+        @test_warn "Duplicate entries in unstack at row 2." unstack(df)
+    end
+
+    @testset "missing values in colkey" begin
+        df = DataFrame(id=[1, 1, 1, missing, missing, missing, 2, 2, 2],
```
nalimilan: Spaces around `=`.

src/abstractdataframe/reshape.jl
bkamins: done!

src/abstractdataframe/reshape.jl
bkamins: I need a copy as I `droplevels!` in the next line. Done. I add test to check that we actually make a copy.

src/abstractdataframe/reshape.jl
bkamins: bacause it is ~5% faster this way. I guess Julia does not like functions with large bodies and `_unstack` is already heavy. I do not know why in detail - those are benchmark results.

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl

bkamins: although now we print "Missing value in variable variable at row"

src/abstractdataframe/reshape.jl
bkamins: I guess this should not happen normally.
This will take place only if we have `missing` in index column (I have even considered throwing an error in this case but in the end I think we can accept `missing` as index).
Therefore I wanted to have a typical-case clean and use `push!` only as an exception.

src/abstractdataframe/reshape.jl
bkamins: I would leave them - I have learned from my students 😄 that they are often confused about operator precedence.
But if you feel that it is cleaner to remove them I will do it.

src/abstractdataframe/reshape.jl
bkamins: `stack` sorts by variable not by index. So if we `unstack` something that was stacked `j` is changing faster than `i` so I guess this layout is better. Yes?

src/abstractdataframe/reshape.jl

test/dataframe.jl
bkamins: I thought the style without spaces is recommended by Julia Manual. Also you do not have spaces in this very file (see e.g. at its beginning - most of calls to `DataFrame` do not use spaces here).
I can fix it - but let us decide - do we want spaces or no spaces everywhere?
(adding spaces will change much more lines)

src/abstractdataframe/reshape.jl
nalimilan: But do you need a copy for `keycol` too?

src/abstractdataframe/reshape.jl
nalimilan: 5% doesn't sound worth it to me. And this kind of difference can change with compiler improvements. Better prioritize the clearer design.

src/abstractdataframe/reshape.jl
nalimilan: I don't think we use parentheses with `!` in any of the Julia codebases I know. I agree precedence can be confusing at times, but in this case it's pretty clear, so please remove it.

test/dataframe.jl
nalimilan: OK, it's already inconsistent, so let's leave it that way.

src/abstractdataframe/reshape.jl
nalimilan: BTW, better not call it `df` since it's not a data frame.

src/abstractdataframe/reshape.jl
nalimilan: If that's the case (haven't checked), then it's not optimal: we need `i` to change the fastest, since accessing one column at a time is more efficient for data frames. Maybe we should change `stack`? Though we should check what R/dplyr/Pandas do first.

src/abstractdataframe/reshape.jl
bkamins: Moved and changed name to `unstacked_val`

src/abstractdataframe/reshape.jl
bkamins: We do (and there was another bug related to this that I have fixed if `keycol` has nonexistent levels).

However, I have a question here - why we want `keycol` to allow `Missing` even if originally it does not have `missing`s? (this is the original design and I respect it, but I do not understand the reason for such an approach)

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl
bkamins: Two points (contradictory though):
* I was wrong we fill the data column-by-column (so `i` is changing faster)
* However, both - result of `stack` and then random permutation of result of `stack` are minimally faster when we have the current order (i.e. `mask_filled = falses(Ncol, Nrow)`). I do not understand why.

Anyway this does not matter much - it is again the case of minimal difference in performance. So I would leave it as is.

src/abstractdataframe/reshape.jl
nalimilan: The original design was probably based on `DataArray` everywhere, so it makes sense to change it to return arrays which do not allow for missing where possible.

src/abstractdataframe/reshape.jl
nalimilan: OK, thanks for checking. That's really weird, but well... For maximal performance, we could also use an `Array{Bool}`, but probably not worth the increased memory use.

nalimilan: BTW, have you tried adding `@inbounds` before indexing expressions where we are sure indices are correct?

test/dataframe.jl
```diff
@@ -296,12 +296,14 @@ module TestDataFrame
 
     #Check the output of unstack
     df = DataFrame(Fish = CategoricalArray{Union{String, Missing}}(["Bob", "Bob", "Batman", "Batman"]),
-                   Key = Union{String, Missing}["Mass", "Color", "Mass", "Color"],
+                   Key = CategoricalArray{Union{String, Missing}}["Mass", "Color", "Mass", "Color"],
```
nalimilan: Do we still have similar tests with non-categorical columns?

bkamins: added

src/abstractdataframe/reshape.jl
bkamins: the code is complex and easy to make mistake in reasoning. I would leave out `@inbounds` until it is well tested by users. I will add a TODO in the comment.

nalimilan: No need for a TODO: either you know it's OK or you don't.

At least it should be fine when you index `refkeycol` and `keycol`.

src/abstractdataframe/reshape.jl
bkamins: ok. I removed allowing missing if not needed (I hope it is ok - you know `Missings` best 😄):
* I create categorical using `categorical`;
* I do not force `col` in single index case to allow missings;
* I do not force `df1` to allow missings;

src/abstractdataframe/reshape.jl
bkamins: I know 😄, but conditional on what you have written, so I would not risk. I have checked that the gains form `@inbounds` are minimal <1%.

bkamins:
@nalimilan This is probably a bug in `CategoricalArrays` which causes me problems with `unstack` 😄:
```
julia> x = categorical([1,2,3])
3-element CategoricalArrays.CategoricalArray{Int64,1,UInt32}:
 1
 2
 3

julia> levels!(x, [4,3,2,1])
3-element CategoricalArrays.CategoricalArray{Int64,1,UInt32}:
 1
 2
 3

julia> y = categorical(x)
3-element CategoricalArrays.CategoricalArray{Int64,1,UInt32}:
 1
 2
 3

julia> droplevels!(y)
3-element CategoricalArrays.CategoricalArray{Int64,1,UInt32}:
 1
 2
 3

julia> x
3-element CategoricalArrays.CategoricalArray{Int64,1,UInt32}:
 3
 2
 1
```

The reason is that `categorical` creates new `pool` but `refs` is the same in `x` and `y`.

bkamins:
@nalimilan Temporarily I have wrapped `categorical` in `deepcopy` to make the tests pass, but this should be removed when `categocial` is fixed or some other decision in https://github.com/JuliaData/CategoricalArrays.jl/issues/107 is made.

test/dataframe.jl
```diff
@@ -347,7 +396,7 @@ module TestDataFrame
         @test udf == DataFrame(Any[Union{Int, Missing}[1, 2], Union{Int, Missing}[1, 5],
                                    Union{Int, Missing}[2, 6], Union{Int, Missing}[3, 7],
                                    Union{Int, Missing}[4, 8]], [:id, :a, :b, :c, :d])
-        @test all(isa.(udf.columns, Vector{Union{Int, Missing}}))
+        @test all(isa.(udf.columns[2:end], Vector{Union{Int, Missing}}))
```
nalimilan: Also test the type of the first column. Same below.

test/dataframe.jl
```diff
@@ -377,6 +429,36 @@ module TestDataFrame
...
         @test a ≅ b ≅ DataFrame(id = [1, 2], a = [3, missing], b = [missing, 4])
+
+        df = DataFrame(variable=["x", "x"], value=[missing, missing], id=[1,1])
+        @test_warn "Duplicate entries in unstack at row 2 for key 1 and variable x." unstack(df, :variable, :value)
+        @test_warn "Duplicate entries in unstack at row 2 for key 1 and variable x." unstack(df)
+    end
+
+    @testset "missing values in colkey" begin
+        df = DataFrame(id=[1, 1, 1, missing, missing, missing, 2, 2, 2],
+                       variable=["a", "b", missing, "a", "b", "missing", "a", "b", "missing"],
```
nalimilan: Why `"missing"`? Doesn't seem to be tested specifically here.

src/abstractdataframe/reshape.jl
```diff
@@ -150,16 +150,20 @@ melt(df::AbstractDataFrame; variable_name::Symbol=:variable, value_name::Symbol=
 Unstacks a DataFrame; convert from a long to wide format
 
 ```julia
-unstack(df::AbstractDataFrame, rowkey, colkey, value)
-unstack(df::AbstractDataFrame, colkey, value)
+unstack(df::AbstractDataFrame, rowkeys::Union{Symbol, Integer},
+        colkey::Union{Symbol, Integer}, value::Union{Symbol, Integer})
+unstack(df::AbstractDataFrame, rowkeys::Union{AbstractVector{<:Union{Symbol, Integer}}},
```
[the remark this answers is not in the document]

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,70 @@ long = stack(wide)
...
+function _unstack(df::AbstractDataFrame, rowkey::Int,
+                  colkey::Int, value::Int, keycol, valuecol, refkeycol)
     Nrow = length(refkeycol.pool)
     Ncol = length(keycol.pool)
-    payload = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
-    nowarning = true
+    unstacked_val = [similar_missing(valuecol, Nrow) for i in 1:Ncol]
+    hadmissing = false # have we encountered missing in refkeycol
+    mask_filled = falses(Ncol, Nrow+1) # has a given [col,row] entry been filled?
+    warned_dup = false # hawe we already printed duplicate entries warning?
```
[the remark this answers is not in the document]

src/abstractdataframe/reshape.jl
```diff
@@ -185,39 +193,70 @@ long = stack(wide)
...
+
+function _unstack(df::AbstractDataFrame, rowkey::Int,
+                  colkey::Int, value::Int, keycol, valuecol, refkeycol)
     Nrow = length(refkeycol.pool)
     Ncol = length(keycol.pool)
-    payload = DataFrame(Any[similar_missing(valuecol, Nrow) for i in 1:Ncol], map(Symbol, levels(keycol)))
-    nowarning = true
+    unstacked_val = [similar_missing(valuecol, Nrow) for i in 1:Ncol]
+    hadmissing = false # have we encountered missing in refkeycol
+    mask_filled = falses(Ncol, Nrow+1) # has a given [col,row] entry been filled?
```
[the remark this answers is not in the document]
bkamins: ok, fixed (the difference is minimal)

test/dataframe.jl

test/dataframe.jl
bkamins: It is intentional. What I test:
* if I pass `missing` a warning should be printed;
* if I pass `"missing"` a column should be created with that name;

I added lines to test this.

bkamins:
build errors seem unrelated

bkamins:
`@inbounds` - minimal benefit and the code is complex.

nalimilan:
I've added a dependency on CategoricalArrays 0.3.2 since tests wouldn't pass with older versions.

nalimilan:
Can you suggest a commit message, or squash and add a message to the commit?

bkamins:
I have squashed and rebased. It ended up quite big - many things required fixing.

nalimilan:
Thanks!

PR #1310, opened 2017-12-06
Fix readtable() in the presence of missing values

nalimilan:
It failed because of a name conflict introduced when porting from Null to Missing.

See https://discourse.julialang.org/t/how-to-read-a-table-with-missing-values-using-csv/7556/4.


nalimilan:
Yeah, even too accurate apparently... :-)

Issue #1297, opened 2017-12-01
Join on columns of different name


cjprybol:
Transfer of https://github.com/JuliaData/DataTables.jl/issues/78

idea summary: Use a 2-element tuple to support joining DataFrames on columns that do not have the same names. Currently, we require renaming columns so that they match before columns can be used for joining

```julia
join(a, b, on=(:a_name, :b_name))
join(a, b, on=[(:a_name1, :b_name1), (:a_name2, :b_name2)])
```

[Pandas does this more explicitly by having `left_on` `right_on` keywords](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.merge.html#pandas-dataframe-merge), and R's Data.Table has `by.x` and `by.y` ([jump to page 61 in docs](https://cran.r-project.org/web/packages/data.table/data.table.pdf))

ararslan:
I'd suggest using pairs, e.g. `join(a, b, on=[:a_name1 => :b_name1, :a_name2 => :b_name2])`

nalimilan:
I'm hesitant about using `Pair`. That kind of suggests a direction or a movement, but here the two column names are treated symmetrically. I would make sense if the second column name was retained in the result, but using the first one would sound more natural (that's what R does at least). OTOH, pairs make it clearer that matching columns need to be grouped, rather than grouping by (left, right). Without them it's really hard to remember.

Anyway, if we use `first` and `last` internally, we can support both tuples and pairs, so we just need to choose which syntax we promote in the docs.

cjprybol:
closed via https://github.com/JuliaData/DataFrames.jl/pull/1312

PR #1312, opened 2017-12-08
Support joining dataframes on columns with different left/right names

cjprybol:
Banishes https://github.com/JuliaData/DataFrames.jl/issues/1297 to [**Closed**](https://github.com/JuliaData/DataFrames.jl/issues?q=is%3Aissue+is%3Aclosed)

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -6,21 +6,34 @@
 similar_missing(dv::AbstractArray{T}, dims::Union{Int, Tuple{Vararg{Int}}}) where {T} =
     fill!(similar(dv, Union{T, Missing}, dims), missing)
 
+const OnType = Union{Symbol, NTuple{2,Symbol}, Pair{Symbol,Symbol}}
+
 # helper structure for DataFrames joining
 struct DataFrameJoiner{DF1<:AbstractDataFrame, DF2<:AbstractDataFrame}
     dfl::DF1
     dfr::DF2
     dfl_on::DF1
     dfr_on::DF2
-    on_cols::Vector{Symbol}
+    left_on::Vector{Symbol}
+    right_on::Vector{Symbol}
 
-    function DataFrameJoiner{DF1, DF2}(dfl::DF1, dfr::DF2, on::Union{Symbol,Vector{Symbol}}) where {DF1, DF2}
-        on_cols = isa(on, Symbol) ? [on] : on
-        new(dfl, dfr, dfl[on_cols], dfr[on_cols], on_cols)
+    function DataFrameJoiner{DF1, DF2}(dfl::DF1, dfr::DF2,
+                                       on::Union{<:OnType, AbstractVector{<:OnType}}) where {DF1, DF2}
+
+        on_cols = isa(on, Vector) ? on : [on]
+        if eltype(on_cols) == Symbol
+            left_on = on_cols
+            right_on = on_cols
+        else
+            left_on = [first(x) for x in on_cols]
+            right_on = [last(x) for x in on_cols]
+        end
+        new(dfl, dfr, dfl[left_on], dfr[right_on], left_on, right_on)
     end
 end
 
-DataFrameJoiner(dfl::DF1, dfr::DF2, on::Union{Symbol,Vector{Symbol}}) where {DF1<:AbstractDataFrame, DF2<:AbstractDataFrame} =
+DataFrameJoiner(dfl::DF1, dfr::DF2, on::Union{<:OnType, AbstractVector{<:OnType}}) where
+    {DF1<:AbstractDataFrame, DF2<:AbstractDataFrame} =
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/join.jl
```diff
@@ -6,21 +6,34 @@
 similar_missing(dv::AbstractArray{T}, dims::Union{Int, Tuple{Vararg{Int}}}) where {T} =
     fill!(similar(dv, Union{T, Missing}, dims), missing)
 
+# OnType
+const OT = Union{Symbol, NTuple{2,Symbol}, Pair{Symbol,Symbol}}
```
nalimilan: Better spell it in full: `OnType`.

src/abstractdataframe/join.jl
```diff
@@ -182,23 +195,21 @@ function update_row_maps!(left_table::AbstractDataFrame,
...
 
 ### Arguments
 
 * `df1`, `df2` : the two AbstractDataFrames to be joined
 
 ### Keyword Arguments
 
-* `on` : a Symbol or Vector{Symbol}, the column(s) used as keys when
-  joining; required argument except for `kind = :cross`
+* `on` : A column, or vector of columns to join df1 and df2 on; If the column(s)
```
nalimilan: Use period rather than semicolon. Also add backticks around `(left, right)` and `(left => right)`, and "Tuple" should either be in backticks and in singular, or be in lowercase.

Could you also add an example?

test/join.jl
```diff
@@ -371,4 +371,21 @@ module TestJoin
...
+            DataFrame(id = 3:7, sid = string.(3:7), SID = string.(3:7))
+        @test join(left, right, on = (:id => :ID)) ==
+            DataFrame(id = 3:7, sid = string.(3:7), SID = string.(3:7))
+        @test join(left, right, on = [(:id, :ID)]) ==
+            DataFrame(id = 3:7, sid = string.(3:7), SID = string.(3:7))
+        @test join(left, right, on = [(:id => :ID)]) ==
+            DataFrame(id = 3:7, sid = string.(3:7), SID = string.(3:7))
+        @test join(left, right, on = [(:id, :ID), (:sid, :SID)]) ==
+            DataFrame(id = 3:7, sid = string.(3:7))
+        @test join(left, right, on = [(:id => :ID), (:sid => :SID)]) ==
```
nalimilan: Parentheses aren't needed, right?

test/join.jl
```diff
@@ -371,4 +371,21 @@ module TestJoin
         @test levels(join(A, B, on=:b, kind = :semi)[:b]) == ["d", "c", "b", "a"]
         @test levels(join(B, A, on=:b, kind = :semi)[:b]) == ["a", "b", "c"]
     end
+
+    @testset "join on columns with different left/right names" begin
```
nalimilan: Can you also tests different kinds of joins?

cjprybol: Great idea. Your suspicion
> Looks mostly good, but I wonder how it can work given that left_on isn't used anywhere...

was very astute : )

src/abstractdataframe/join.jl
```diff
@@ -182,23 +195,21 @@ function update_row_maps!(left_table::AbstractDataFrame,
...
-Join two DataFrames
+    join(df1, df2; on = Symbol[], kind = :inner)
 
-```julia
-join(df1::AbstractDataFrame,
-     df2::AbstractDataFrame;
-     on::Union{Symbol, Vector{Symbol}} = Symbol[],
-     kind::Symbol = :inner)
-```
+Join two DataFrames
```
nalimilan: `data frames` or `` `DataFrame` objects``

src/abstractdataframe/join.jl
```diff
@@ -6,21 +6,34 @@
...
+            left_on = [first(x) for x in on_cols]
+            right_on = [last(x) for x in on_cols]
+        end
+        new(dfl, dfr, dfl[left_on], dfr[right_on], left_on, right_on)
     end
 end
 
-DataFrameJoiner(dfl::DF1, dfr::DF2, on::Union{Symbol,Vector{Symbol}}) where {DF1<:AbstractDataFrame, DF2<:AbstractDataFrame} =
+DataFrameJoiner(dfl::DF1, dfr::DF2, on::Union{<:OnType, AbstractVector{<:OnType}}) where
+    {DF1<:AbstractDataFrame, DF2<:AbstractDataFrame} =
```
nalimilan: I guess that's the indentation convention for this kind of situation? I'm not used to it yet.

cjprybol: I'm not sure if there is a convention, but this only works if the line break is after the `where` clause. If you line break before the `where` then it results in an error

nalimilan: Yeah, that's annoying. But I was wondering whether we should use four spaces, or align with arguments. Hard to decide.

PR #1314, opened 2017-12-12
fix typo in `isdone`: df -> source

jrklasen:


nalimilan:
Thanks! Do you have a minimal example to reproduce the bug? We should add it to the tests so that this kind of mistake doesn't happen again.

jrklasen:
unfortunately not. I spotted this by reading the code in order to familiarize myself with DataStreams.

jrklasen:
:), agreed

nalimilan:
> Yeah, to test you could just do @test(Data.isdone(DataFrame(a=[1,2], b=[3,4])), 3, 3)

Can you make a PR to add it?

PR #1315, opened 2017-12-12
Simplify source files inclusion

alyst:
The current code calls `include()` in a loop. It confuses `Revise`, so it cannot handle `DataFrames` package in v0.6.
v0.7 should solve such problems, but since the loop doesn't really simplify or reduce the code, let's just remove it.


nalimilan:
So you've been hit by this too... :-)

PR #1316, opened 2017-12-13
Preserve non-missingness during non-inner joins

alyst:
Currently, if you do a few joins, most of the columns in your data frame become "missable" (`>: Missing`).
That's because so far `join` preserves non-missingness only for inner joins:
 * inner join doesn't introduce new missing values => non-missingness of columns from both left and right frames should be preserved.

This PR adds the other logical rules:
  * left join doesn't introduce new missing values in the left columns => non-missingness of the left frame should be preserved
  * right join doesn't introduce new missing values in the right columns => non-missingness of the right frame should be preserved
  * joins do not introduce missing values to the on-columns => non-missingness of the on columns should be preserved if both left-on and right-on columns are non-missing

The rules above are "type stable", so to say. I.e. they don't depend on the contents of the frames.
But I thought it also makes sense to introduce data-dependent rules (there's no performance penalty in checking them):
   * if all right rows have matching left rows (i.e. no missing values are introduced to the left columns) => non-missingness of the left columns is preserved
   * if all left rows have matching right rows (i.e. no missing values are introduced to the right columns) => non-missingness of the right columns is preserved

As a reasonable side effect of the rules implementation, when doing the right join, the eltype (and levels, for categorical arrays) of the right on-column has priority over the left on-column.

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -85,25 +85,41 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
     @assert nrow == length(all_orig_right_ixs) + loil
     ncleft = ncol(joiner.dfl)
     cols = Vector{Any}(ncleft + ncol(dfr_noon))
-    _similar = kind == :inner ? similar : similar_missing
+    # inner and left joins preserve non-missingness of the left frame
+    _similar_left = kind == :inner || kind == :left ? similar : similar_missing
     for (i, col) in enumerate(columns(joiner.dfl))
-        cols[i] = _similar(col, nrow)
+        cols[i] = _similar_left(col, nrow)
         copy!(cols[i], view(col, all_orig_left_ixs))
     end
+    # inner and right joins preserve non-missingness of the right frame
+    _similar_right = kind == :inner || kind == :right ? similar : similar_missing
     for (i, col) in enumerate(columns(dfr_noon))
-        cols[i+ncleft] = _similar(col, nrow)
+        cols[i+ncleft] = _similar_right(col, nrow)
         copy!(cols[i+ncleft], view(col, all_orig_right_ixs))
         permute!(cols[i+ncleft], right_perm)
     end
     res = DataFrame(cols, vcat(names(joiner.dfl), names(dfr_noon)))
 
     if length(rightonly_ixs.join) > 0
-        # some left rows are missings, so the values of the "on" columns
+        # some left rows are missing, so the values of the "on" columns
         # need to be taken from the right
         for (on_col_ix, on_col) in enumerate(joiner.left_on)
             # fix the result of the rightjoin by taking the nonmissing values from the right table
             offset = nrow - length(rightonly_ixs.orig) + 1
             copy!(res[on_col], offset, view(joiner.dfr_on[on_col_ix], rightonly_ixs.orig))
+        end
+    end
+    if kind ∈ (:right, :outer) && !isempty(rightonly_ixs.join)
+        # At this point on-columns of the result allow missing values, because
+        # right-only rows were filled with missing values when processing joiner.dfl
+        # However, when the right on-column (plus the left one for the outer join)
+        # does not allow missing values, the result should also disallow them.
+        for (on_col_ix, on_col) in enumerate(joiner.left_on)
+            LT = eltype(joiner.dfl_on[on_col_ix])
+            RT = eltype(joiner.dfr_on[on_col_ix])
+            if !(RT >: Missing) && (kind == :right || !(LT >: Missing))
+                res[on_col] = disallowmissing(res[on_col])
+            end
         end
     end
     return res
--- a/test/join.jl
+++ b/test/join.jl
@@ -186,17 +186,17 @@ module TestJoin
                                 fid = [1, 3, 5],
                                 fid_1 = [1, 3, missing])
         @test typeof.(l(on).columns) ==
-            [Vector{Union{T, Missing}} for T in (Int, Float64, Float64)]
+            [Vector{Int}, Vector{Float64}, Vector{Union{Float64, Missing}}]
         @test r(on) ≅ DataFrame(id = [1, 3, 0, 2, 4],
                                 fid = [1, 3, missing, missing, missing],
                                 fid_1 = [1, 3, 0, 2, 4])
         @test typeof.(r(on).columns) ==
-            [Vector{Union{T, Missing}} for T in (Int, Float64, Float64)]
+            [Vector{Int}, Vector{Union{Float64, Missing}}, Vector{Float64}]
         @test o(on) ≅ DataFrame(id = [1, 3, 5, 0, 2, 4],
                                 fid = [1, 3, 5, missing, missing, missing],
                                 fid_1 = [1, 3, missing, 0, 2, 4])
         @test typeof.(o(on).columns) ==
-            [Vector{Union{T, Missing}} for T in (Int, Float64, Float64)]
+            [Vector{Int}, Vector{Union{Float64, Missing}}, Vector{Union{Float64, Missing}}]
 
         on = :fid
         @test i(on) == DataFrame(Any[[1, 3], [1.0, 3.0], [1, 3]], [:id, :fid, :id_1])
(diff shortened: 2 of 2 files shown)


alyst:
Travis Mac CI stalled during Julia image download. other tests ok

test/join.jl
```diff
@@ -346,8 +343,8 @@ module TestJoin
         @test levels(join(B, A, on=:b, kind=:inner)[:b]) == ["a", "b", "c"]
         @test levels(join(A, B, on=:b, kind=:left)[:b]) == ["d", "c", "b", "a"]
         @test levels(join(B, A, on=:b, kind=:left)[:b]) == ["a", "b", "c"]
-        @test levels(join(A, B, on=:b, kind=:right)[:b]) == ["d", "c", "b", "a"]
-        @test levels(join(B, A, on=:b, kind=:right)[:b]) == ["a", "b", "d", "c"]
+        @test levels(join(A, B, on=:b, kind=:right)[:b]) == ["a", "b", "c"]
```
cjprybol: Do these changes here and below to Categorical levels adhere to the ordering requests made in https://github.com/JuliaData/DataFrames.jl/issues/1257? I don't see why restricting the auto promotion of columns should affect these

cjprybol:
Last time we considered this we voted against it because we were still using NullableArrays and we wanted to consistently interact with the columns using `get` and `f.()` dot-broadcasting, right? But now we're using the `Missings` approach and will interact with columns the same way whether they have missing data or not. And there will be downstream performance gains (at least in the near future) if we can avoid introducing the `Union{T, Missing}` and just keep `T`, so the extra work in `join` probably makes sense in many situations. Thanks for putting together this PR!

src/abstractdataframe/join.jl
```diff
@@ -85,27 +85,53 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
     @assert nrow == length(all_orig_right_ixs) + loil
     ncleft = ncol(joiner.dfl)
     cols = Vector{Any}(ncleft + ncol(dfr_noon))
-    _similar = kind == :inner ? similar : similar_missing
+    # inner and left joins preserve non-missingness of the left frame
+    # it is also preserved if all right rows have left matches
+    _similar_left = kind == :inner || kind == :left || length(rightonly_ixs.join) == 0 ? similar : similar_missing
```
nalimilan: Since the condition is very long, better use `if ...; _similar_left = ...`

src/abstractdataframe/join.jl
```diff
@@ -85,27 +85,53 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
...
     for (i, col) in enumerate(columns(joiner.dfl))
-        cols[i] = _similar(col, nrow)
-        copy!(cols[i], view(col, all_orig_left_ixs))
+        on_col_ix = findfirst(joiner.left_on, names(joiner.dfl)[i])
+        if on_col_ix > 0 && kind == :right
+            # if right join, construct the on-column
+            # using the right frame to preserve missingness and cat.levels
```
nalimilan: What's "cat.levels"?

src/abstractdataframe/join.jl
```diff
@@ -85,27 +85,53 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
...
+            rcol = joiner.dfr_on[on_col_ix]
+            cols[i] = similar(rcol, nrow)
+            copy!(cols[i], view(rcol, all_orig_right_ixs))
+            permute!(cols[i], right_perm)
```
nalimilan: Isn't there an alternative solution which doesn't involve calling `permute!`? I'm concerned about the performance impact compared with the previous code which simply called `copy!`.

alyst: This is the case only for the right joins. Previously it was handled later in the code (`if length(rightonly_ixs) > 0` block).
Actually, now it should be faster, because we compose the resulting column using the right frame directly, instead of initializing it using the left frame and fixing it later.

src/abstractdataframe/join.jl
alyst: categorical levels :) I'll expand it

alyst:
> I'm not a fan of changing the returned type depending on the data. Apart from making the behavior hard to predict, it won't work with a fully type-stable variant of DataFrame that I've considered supporting as an alternative. Do you really think it's common to make a left/right join without any missing values? The point of these operations is that you expect some rows to be missing from the second table, or you can use an inner join instead.

The problem of the inner join is that it "swallows" the rows that do not have matches, and checking for that is more expensive (performance- and code size-wise) than the introduction of missing values and checking the resulting column type.
I'm also a little bit concerned about "type-instability". I was thinking whether we can introduce some switch to `join()` that specifies the missingness behaviour (don't have a good name for it ATM):
 * type-stable: no data-dependent rules
 * with data-dependent rules
 * strict: the original non-missingness is preserved, so if we are doing e.g. left join and some left rows have no matching right rows, but some right columns don't allow missing values, `join()` throws `MissingException` (so that's inner join with no "swallowed" rows allowed)

nalimilan:
> The problem of the inner join is that it "swallows" the rows that do not have matches, and checking for that is more expensive (performance- and code size-wise) than the introduction of missing values and checking the resulting column type.

It's not an issue if we use an `Array{Union{T, Missing}}` internally, as long as we take care of calling `disallowmissing` before returning it.

> I'm also a little bit concerned about "type-instability". I was thinking whether we can introduce some switch to join() that specifies the missingness behaviour (don't have a good name for it ATM):
> 
>     type-stable: no data-dependent rules
>     with data-dependent rules
>     strict: the original non-missingness is preserved, so if we are doing e.g. left join and some left rows have no matching right rows, but some right columns don't allow missing values, join() throws MissingException (so that's inner join with no "swallowed" rows allowed)

Given that we have `disallowmissing!`, I don't really see the advantage of providing a keyword argument to achieve the same result. And the "strict" case sounds unusual to me, as left/right joins are supposed to generate missing values anyway. I don't think any other software supports this kind of option, which could be confusing.

alyst:
> It's not an issue if we use an `Array{Union{T, Missing}}` internally, as long as we take care of calling `disallowmissing` before returning it.

I'm not sure I follow. By "internally" you mean within `join()`?
Or do you mean calling `disallowmissing()` *after* `join()` (from the user code)?
That's what I'm doing right now (after the left join) in my scripts to verify that each left row has right row match.
That's fine for the check (it requires just one right column), but if I want all non-missing right columns to stay non-missing, that's a few lines of dataschema-specific code.

> Given that we have `disallowmissing!`, I don't really see the advantage of providing a keyword argument to achieve the same result. 

See the usecase above, the keyword is equivalent to applying `disallowmissing` to all relevant columns (that were not of type `Union{T,Missing}` in the source frame and do not contain `missing` in the result).
I was thinking whether it could be implemented as a separate method, but it would require all 3 frames (left, right, result), so using it will look just like ugly join postprocessing.

> And the "strict" case sounds unusual to me, as left/right joins are supposed to generate missing values anyway. I don't think any other software supports this kind of option, which could be confusing.

In SQL you have "INSERT INTO tbl SELECT ...", which checks that the result of SELECT fits the constraints of "tbl".
In data frame world we don't have data schemas, so the `strict` option is just a poor man's replacement (the result fits the constraints imposed by the source frames).
But I agree it would be confusing. Actually, since it's just inner join + the check that each row has a match, an alternative is to provide it as a new join kind (`kind=:strict`).

alyst:
I've updated the commit removing the data-dependent rules.
It looks like there is consensus regarding the rest of the PR, so doesn't make sense to block it.
Data-dependent rules deserve a separate PR (I'll submit it as soon as this one is merged) and separate discussion.

alyst:
CI passes on 0.6

src/abstractdataframe/join.jl
```diff
@@ -85,27 +85,51 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
...
-        # need to be taken from the right
+        # need to be taken from the right (unless already done when processing the left frame)
         for (on_col_ix, on_col) in enumerate(joiner.left_on)
             # fix the result of the rightjoin by taking the nonmissing values from the right table
             offset = nrow - length(rightonly_ixs.orig) + 1
             copy!(res[on_col], offset, view(joiner.dfr_on[on_col_ix], rightonly_ixs.orig))
         end
     end
+    if kind == :outer && !isempty(rightonly_ixs.join)
+        # some non-missing on-columns may have become missing
```
nalimilan: "missing column" can be confusing as it's not clear whether it's the column which might be missing. Better say something like "columns allowing for missing" even if it's a bit long.

src/abstractdataframe/join.jl
```diff
@@ -85,27 +85,51 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
...
+        # when constructing res, because there are both left-only and right-only rows
+        for (on_col_ix, on_col) in enumerate(joiner.left_on)
+            LT = eltype(joiner.dfl_on[on_col_ix])
+            RT = eltype(joiner.dfr_on[on_col_ix])
+            if Missings.T(LT) === LT && Missings.T(RT) === RT
```
nalimilan: Couldn't this just be `!(LT >: Missing || RT >: Missing)`? That's more explicit IMHO. It could also be worth adding a special case for `Any` columns, for which there's no point in calling `disallowmissing` since it won't have any effect.

test/join.jl
nalimilan: Why does this change? The docstring for `join` gives guarantees about the levels, so either it wasn't correct before or it might need to be updated.

src/abstractdataframe/join.jl
alyst: I did it this way because `Any >: Missing`. But you're right, in this context we don't need to fix `Any` columns. Will fix.

nalimilan: But `Missings.T(Any) === Any` too.

src/abstractdataframe/join.jl
alyst: Will fix. In the long run would be nice to have a term for it (even the ugly internal one, which would be used in cases like this).

test/join.jl
alyst: In this PR the types and categorical levels of the right on-columns have priority over the left ones.
I think it's a logical thing. Also it's easier to implement.
So I can either patch the PR to the previous behaviour or update the docstring.

src/abstractdataframe/join.jl
alyst: Exactly, so the code would have called `disallowmissing`, which is wrong.
Since `!(Any >: Missing || T >: Missing) === false`, it would not be called.

test/join.jl
nalimilan: As long as it's not too hard to implement, I'd choose the rule which is the simplest to explain in the docstring, which is a good indication that it's easy to understand. IIUC, the new behavior you propose is to give priority to the left column, except for `kind=:right`? I'm not too fan of exceptions, even if I see the justification.

alyst: I've reverted to the old category levels ordering.

src/abstractdataframe/join.jl
alyst: I've added a testset for joins with Any eltypes.

alyst:
squashed, in case you want to merge it preserving the commits

PR #1318, opened 2017-12-13
Fix hash()

nalimilan:
Types should always implement the two-argument version, which is the most general method.
Follow the same approach as the AbstractArray implementation in Base.

See https://discourse.julialang.org/t/a-question-on-hashes/7742.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -786,16 +786,17 @@ end
 ##
 ## Hashing
 ##
-## Make sure this agrees with isequals()
-##
 ##############################################################################
 
-function Base.hash(df::AbstractDataFrame)
-    h = hash(size(df)) + 1
+const hashdf_seed = UInt == UInt32 ? 0xfd8bb02e : 0x6215bada8c8c46de
+
+function Base.hash(df::AbstractDataFrame, h::UInt)
+    h += hashdf_seed
+    h += hash(size(df))
     for i in 1:size(df, 2)
         h = hash(df[i], h)
     end
-    return UInt(h)
+    return h
 end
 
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -786,16 +786,17 @@ end
 ##
 ## Hashing
 ##
-## Make sure this agrees with isequals()
-##
 ##############################################################################
 
-function Base.hash(df::AbstractDataFrame)
-    h = hash(size(df)) + 1
+const hashdf_seed = UInt == UInt32 ? 0xfd8bb02e : 0x6215bada8c8c46de
```
ararslan: It'd be nice to use the general convention of all uppercase for constants.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -786,16 +786,17 @@ end
...
+
+function Base.hash(df::AbstractDataFrame, h::UInt)
+    h += hashdf_seed
+    h += hash(size(df))
```
ararslan: `h +=  hashdf_seed + hash(size(df))`?

src/abstractdataframe/abstractdataframe.jl
nalimilan: Actually I've followed what Base does with `AbstractArray` and in other places. I'd rather be consistent with that pattern.

ararslan: Okay, that's fine

src/abstractdataframe/abstractdataframe.jl
ararslan: Works for me

PR #1320, opened 2017-12-17
0.7 updates

quinnj:


src/DataFrames.jl
```diff
@@ -1,6 +1,10 @@
 __precompile__(true)
 module DataFrames
 
+if !isdefined(Base, :pairs)
+    pairs(x) = x
```
ararslan: This seems like a dangerous definition to have lying around, but I guess it's okay until we stop supporting 0.6.

nalimilan: Yeah. Shouldn't Compat provide something like that, since many packages will have the same problem?

BTW, is Femtocleaner able to remove this kind of `isdefined` check? Or should we use `VERSION` checks instead?

ararslan: I'm not sure about FemtoCleaner, but `pairs` will be available in Compat as of 0.40.0: https://github.com/JuliaLang/METADATA.jl/pull/12530

nalimilan: So do we want to use `Compat.pairs` here rather than a custom method?

ararslan: I'd say so, now that it's available in Compat. Just have to bump the REQUIRE to 0.40.0.

quinnj: now we might as well wait till Compat gets Unicode since DataFrames show totally breaks w/o it.

ararslan: Nah, I'd say just add
```julia
if VERSION >= v"0.7.0-DEV.2915"
    using Unicode
end
```

ararslan:
WOW that 0.7 failure. :scream: Reported at https://github.com/JuliaLang/julia/issues/25176

bkamins:
I believe that this line:
https://github.com/JuliaData/DataFrames.jl/commit/cbb54c792dc52ee3ea99da4f366fdb87a38c381b#diff-0aa1ff52d56badabf7d1511d3dc3f279R119
is not working correctly with Julia 0.6 and all `DataFrames` tests will fail:

```
julia> using Compat.pairs

julia> function test(; kwargs...)
           collect((k => v for (k,v) in pairs(kwargs)))
       end
test (generic function with 1 method)

julia> test(a=1,b=2)
2-element Array{Pair{Int64,Tuple{Symbol,Int64}},1}:
 1=>(:a, 1)
 2=>(:b, 2)
```
or I am missing something?

nalimilan:
Did I approve the changes despite CI not passing? I'm too used to seeing 0.7 fail, I didn't even check whether 0.6 passed...

ararslan:
Yeah 0.6 was failing; I was surprised this got merged as-is.

nalimilan:
See https://github.com/JuliaData/DataFrames.jl/pull/1329.

PR #1323, opened 2017-12-21
added a test

bramtayl:


nalimilan:
Thanks, but would you mind adding a whole series of tests at the same time rather than opening a PR for each?

PR #1325, opened 2017-12-22
improve DataFrame constructors and conversions for Vector and Matrix

bkamins:
Following discussion in https://discourse.julialang.org/t/how-do-dataframes-jl-compare-to-rs-and-interoperability-between-r-and-julia/7387/14 I propose to:
1. Add a constructor of `DataFrame` from a `Matrix` that can take column names argument;
2. Add a method to `convert` allowing conversion from `Vector` to `DataFrame` (so that both the constructor and conversion work and work the same way);

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -131,7 +131,7 @@ end
 
 function DataFrame(columns::AbstractVector, cnames::AbstractVector{Symbol};
                    makeunique::Bool=false)::DataFrame
-    if !any(col -> isa(col, AbstractVector), columns)
+    if !all(col -> isa(col, AbstractVector), columns)
         # change to throw(ArgumentError("columns argument must be a vector containing at least one vector"))
         Base.depwarn("passing columns argument containing only scalars is deprecated", :DataFrame)
     end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -132,8 +132,8 @@ end
 function DataFrame(columns::AbstractVector, cnames::AbstractVector{Symbol};
                    makeunique::Bool=false)::DataFrame
     if !all(col -> isa(col, AbstractVector), columns)
-        # change to throw(ArgumentError("columns argument must be a vector containing at least one vector"))
-        Base.depwarn("passing columns argument containing only scalars is deprecated", :DataFrame)
+        # change to throw(ArgumentError("columns argument must be a vector of AbstractVector objects"))
+        Base.depwarn("passing columns argument with non-AbstractVector entries is deprecated", :DataFrame)
     end
     return DataFrame(convert(Vector{Any}, columns), Index(convert(Vector{Symbol}, cnames),
                      makeunique=makeunique))
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -8,6 +8,7 @@ particularly a Vector or CategoricalVector.
 
 ```julia
 DataFrame(columns::Vector, names::Vector{Symbol}; makeunique::Bool=false)
+DataFrame(columns::Matrix, names::Vector{Symbol}; makeunique::Bool=false)
 DataFrame(kwargs...)
 DataFrame(pairs::Pair{Symbol}...; makeunique::Bool=false)
 DataFrame() # an empty DataFrame
@@ -20,7 +21,7 @@ DataFrame(ds::Vector{Associative})
 
 **Arguments**
 
-* `columns` : a Vector with each column as contents
+* `columns` : a Vector with each column as contents or a Matrix
 * `names` : the column names
 * `makeunique` : if `false` (the default), an error will be raised
   if duplicates in `names` are found; if `true`, duplicate names will be suffixed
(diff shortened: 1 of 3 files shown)


src/dataframe/dataframe.jl
```diff
@@ -802,14 +809,8 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame)
...
-function Base.convert(::Type{DataFrame}, A::AbstractMatrix)
-    n = size(A, 2)
-    cols = Vector{Any}(n)
-    for i in 1:n
-        cols[i] = A[:, i]
-    end
-    return DataFrame(cols, Index(gennames(n)))
-end
+Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
+Base.convert(::Type{DataFrame}, V::AbstractVector) = DataFrame(V)
```
nalimilan: Note that this method is very different from the previous one as it expects the entries to be columns. I'm not sure we should support it, as it could be confusing compared with a 1-column matrix.

src/dataframe/dataframe.jl
```diff
@@ -125,6 +126,12 @@ function DataFrame(columns::AbstractVector,
     return DataFrame(convert(Vector{Any}, columns), Index(convert(Vector{Symbol}, cnames)))
 end
 
+# if #1308 is merged this sould be refactor to include makeunique
+function DataFrame(columns::AbstractMatrix, cnames::AbstractVector{Symbol} = gennames(size(columns, 2)))
+    DataFrame(Any[columns[:, i] for i in 1:size(columns, 2)], cnames)
+end
+
+
```
nalimilan: One blank line is enough.

src/dataframe/dataframe.jl
```diff
@@ -125,6 +126,12 @@ function DataFrame(columns::AbstractVector,
     return DataFrame(convert(Vector{Any}, columns), Index(convert(Vector{Symbol}, cnames)))
 end
 
+# if #1308 is merged this sould be refactor to include makeunique
+function DataFrame(columns::AbstractMatrix, cnames::AbstractVector{Symbol} = gennames(size(columns, 2)))
```
nalimilan: Better use the short function syntax for one-liners like this.

bkamins:
In R a matrix is actually our `NamedArray` - `data.frame` uses names of columns.
Actually this is something that could be considered do be added in `NamedArrays`.

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl
bkamins: To say the truth I do not like `DataFrame(::Vector)` constructor, especially something like `DataFrame([1,2,3])` is very non-intuivite. I pass a column vector and get a row of data.

But if we allow it then it is consistent to provide `convert` also.

Maybe a solution is to restrict the implementation of `DataFrame(columns::AbstractVector` to check internally:
```
all(col -> isa(col, AbstractVector), columns) || throw(ArgumentError("columns must be vectors"))
```

bkamins:
I decided to recommend allowing `DataFrames` constructor (and thus `convert`) from `AbstractVector` only if it contains `AbstractVector`. This is breaking so I put deprecation for the time being.

Other issue is that tests on 0.7 fail (for different reasons - I am not sure if this is something to worry about, but in one case Julia crashed trying to precompile `CategoricalArrays`)

src/dataframe/dataframe.jl
```diff
@@ -122,9 +123,17 @@ end
 
 function DataFrame(columns::AbstractVector,
                    cnames::AbstractVector{Symbol} = gennames(length(columns)))
+    if !all(col -> isa(col, AbstractVector), columns)
+        # change to throw(ArgumentError("columns argument mutst be a vector of vectors"))
```
nalimilan: "must"

src/dataframe/dataframe.jl
```diff
@@ -122,9 +123,17 @@ end
...
+        Base.depwarn("passing columns argument other than a vector of vectors is deprecated", :DataFrame)
+    end
     return DataFrame(convert(Vector{Any}, columns), Index(convert(Vector{Symbol}, cnames)))
 end
 
+# if #1308 is merged this sould be refactor to include makeunique
```
nalimilan: Let's merge #1308 first.

src/dataframe/dataframe.jl
nalimilan: Yeah, it's a bit weird. Maybe we could also require specifying the column names when passing a vector, given that this is a rather low-level constructor.

bkamins: agreed. this means that `convert` should be removed. I add deprecation to calling *vector of vectors* `DataFrame` without passing names.

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl
bkamins: OK. I am pushing the as-is version to have it on GitHub, but I will rebase it when 1308 is merged.

src/dataframe/dataframe.jl
```diff
@@ -120,11 +121,23 @@ function DataFrame(; kwargs...)
     end
 end
 
-function DataFrame(columns::AbstractVector,
-                   cnames::AbstractVector{Symbol} = gennames(length(columns)))
+function DataFrame(columns::AbstractVector)
```
nalimilan: That function should be moved to deprecated.jl.

bkamins: moved

bkamins:
rebased, added `makeunique` to `Matrix` constructor and in `Vector` constructor changed that it is required that at least one element is a vector (not all as was earlier).

nalimilan:
I'd rather require all entries to be `AbstractVector` as you did originally, not just one of them. That's simpler, and since that's a low-level constructor is not a big deal to require the caller to create a vector.

bkamins:
OK - fixed

PR #1326, opened 2017-12-22
make boolean indexing of Index strict

### LINKED ISSUE #1321: Selecting columns by boolean index
Currently the following is OK:
```
df = DataFrame(rand(3,5))
df[[true, false]]
```
Usually it is expected that when using boolean selection its length should be exactly the length of the selection.
I propose to have the same strict checking in this case.

The current behavior is error prone. E.g. consider what happens when we write `df[df[:x1] .< 2]` on the above `DataFrame` (this is a mistake by the developer, but highly probable in practice) - probably the intended condition was `df[df[:x1] .< 2, :]`, but the above code will go through without any problem.

bkamins:
Fixes #1321.
Additionally corrects signatures of `getindex`.

### DIFF
--- a/src/other/index.jl
+++ b/src/other/index.jl
@@ -112,7 +112,10 @@ Base.getindex(x::Index, idx::Symbol) = x.lookup[idx]
 Base.getindex(x::AbstractIndex, idx::Real) = Int(idx)
 Base.getindex(x::AbstractIndex, idx::AbstractVector{Union{Bool, Missing}}) =
     getindex(x, collect(Missings.replace(idx, false)))
-Base.getindex(x::AbstractIndex, idx::AbstractVector{Bool}) = find(idx)
+function Base.getindex(x::AbstractIndex, idx::AbstractVector{Bool})
+    length(x) == length(idx) || Base.throw_boundserror(x, idx)
+    find(idx)
+end
 Base.getindex(x::AbstractIndex, idx::AbstractVector{T}) where {T >: Missing} =
     getindex(x, collect(skipmissing(idx)))
 Base.getindex(x::AbstractIndex, idx::AbstractRange) = [idx;]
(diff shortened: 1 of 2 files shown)


src/other/index.jl
```diff
@@ -112,7 +112,10 @@ Base.getindex(x::Index, idx::Symbol) = x.lookup[idx]
 Base.getindex(x::AbstractIndex, idx::Real) = Int(idx)
 Base.getindex(x::AbstractIndex, idx::AbstractVector{Union{Bool, Missing}}) =
     getindex(x, collect(Missings.replace(idx, false)))
-Base.getindex(x::AbstractIndex, idx::AbstractVector{Bool}) = find(idx)
+function Base.getindex(x::AbstractIndex, idx::AbstractVector{Bool})
+    length(x) == length(idx) || Base.throw_boundserror(x, idx)
```
nalimilan: I'd rather not call an internal Base function, especially since it's so simple. If you're afraid of stopping the compiler from inlining this method, just copy the definition here.

src/other/index.jl
```diff
@@ -112,7 +112,10 @@ Base.getindex(x::Index, idx::Symbol) = x.lookup[idx]
 Base.getindex(x::AbstractIndex, idx::Real) = Int(idx)
 Base.getindex(x::AbstractIndex, idx::AbstractVector{Union{Bool, Missing}}) =
     getindex(x, collect(Missings.replace(idx, false)))
```
nalimilan: This constructor needs the same check, right? Though I guess we can leave it that way if you remove it soon.

bkamins: this one will call the one I changed so it should be safe (and as you say I plan to remove it anyway).
Does removing of handling of `Missing` require deprecation period?

nalimilan: Ah, indeed. Dropping support for missing values should indeed go through a deprecation, but that's a different issue, right?

bkamins: yes

bkamins:
OK - back on track after @nalimilan fixing `pairs` 👍.

nalimilan:
Thanks! BTW, you might be interested in hanging around on Slack's #data channel.

PR #1329, opened 2017-12-23
Fix failure due to calling pairs() on keyword arguments on Julia 0.6

nalimilan:
Introduced by https://github.com/JuliaData/DataFrames.jl/pull/1320.


bkamins:
Looks OK (I do not know how to check the exact 0.7 version to condition on; the PR that introduced `pairs` is https://github.com/JuliaLang/julia/pull/22907).
Checked that on 0.6.2 the change works.

The only thing we have is that in the other constructor we have `pairs` as argument name, which in 0.7 conflicts with function in Base. Probably this is not a top priority to change it.

ararslan:
> I do not know how to check the exact 0.7 version to condition on

directions

Issue #43, opened 2012-07-26
replace/filter interfaces for DataFrames


HarlanH:
The existing nafilter/Filter and similar methods, and their flags in the DataVecs, seem limited when it comes to working with DataFrames. I like the way that different columns have different behaviors (replace/filter modes), but it's not clear how to combine them. For example, if building a model matrix for an OLS model, do you do a complete_cases operation? The naFilter iterator generator doesn't really work usefully in that context. 

One option would be to have a filter_nas() method that generates a SubDataFrame without any rows that contained an NA in a column with filtering mode set. The result could then be iterated over row-wise, with NAs being replaced in any columns in replace mode. Other options and variations are certainly possible.

See also #4.

Issue #323, opened 2013-07-14
Remove seemingly equivalent ways of producing subsets of DataFrames


johnmyleswhite:
We have too many ways of producing things that seem like subsets of a `DataFrame`. After I added `filter`, we now have four ways that seem superficially equivalent, but reflect two different algorithms:

```
julia> df[:(A % 2 .== 0 && B % 4 .== 0), :]
ERROR: type: non-boolean (DataArray{Bool,1}) used in boolean context
 in anonymous at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1457
 in with at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1458
 in getindex at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:428

julia> sub(df, :(A % 2 .== 0 && B % 4 .== 0))
ERROR: type: non-boolean (DataArray{Bool,1}) used in boolean context
 in anonymous at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1457
 in with at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1458
 in sub at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1014

julia> subset(df, :(A % 2 .== 0 && B % 4 .== 0))
ERROR: type: non-boolean (DataArray{Bool,1}) used in boolean context
 in anonymous at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1457
 in with at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1458
 in sub at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1014

julia> filter(:(A % 2 .== 0 && B % 4 .== 0), df)
ERROR: type: non-boolean (DataArray{Bool,1}) used in boolean context
 in anonymous at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1457
 in with at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1458
 in filter at /Users/johnmyleswhite/.julia/DataFrames/src/dataframe.jl:1023
```

I think we should:
- Return `SubDataFrame` objects everywhere when accessing all columns of a DataFrame, which is what `sub` does, but direct indexing does not do.
- Rename `sub` to `filter` after flipping the argument order. We will then remove `sub` and `subset` completely.

[...]

PR #1330, opened 2017-12-23
Add filter() and filter!() methods

nalimilan:
Currently these cannot be made efficient as the column types are not encoded in `DataFrameRow`, but this can be improved in the future with `NamedTuples`.

Closes https://github.com/JuliaData/DataFrames.jl/issues/43. See also https://github.com/JuliaData/DataFrames.jl/issues/323 and a previous `filter` function (with a different interface) which was removed in https://github.com/JuliaData/DataFrames.jl/commit/e47accbac6e9a92ca5f8b4776c337216deb3e7a2.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -539,6 +539,63 @@ dropmissing!(df)
 """
 dropmissing!(df::AbstractDataFrame) = deleterows!(df, find(!, completecases(df)))
 
+"""
+    filter(function, df::AbstractDataFrame)
+
+Return a copy of data frame `df` containing only rows for which `function`
+returns `true`. The function is passed a `DataFrameRow` as its only argument.
+
+# Examples
+```
+julia> df = DataFrame(x = [3, 1, 2, 1], y = ["b", "c", "a", "b"])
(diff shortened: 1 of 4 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -539,6 +539,63 @@ dropmissing!(df)
...
+
+julia> filter(row -> row[:x] > 1, df)
+2×2 DataFrames.DataFrame
+│ Row │ x │ y │
+├─────┼───┼───┤
+│ 1   │ 3 │ b │
+│ 2   │ 2 │ a │
+```
+"""
+Base.filter(f, df::AbstractDataFrame) = df[collect(f(r)::Bool for r in eachrow(df)), :]
```
ararslan: Is `collect`ing the generator necessary? Seems like `map(f, eachrow(df))` would suffice?

nalimilan: `map` works too, and AFAIK is essentially equivalent to collecting a generator. The advantage of the generator is that we can enforce `::Bool`, which can help for type stability and gives a clearer warning when a missing value is returned (which will likely be a common mistake).

PR #1332, opened 2017-12-29
check categorical argument length

bkamins:
a proposal of a small fix checking that `categorical` argument in `DataFrames` constructor has the right length.


bkamins:
thanks

src/dataframe/dataframe.jl
```diff
@@ -152,6 +152,10 @@ function DataFrame(column_eltypes::AbstractVector{T}, cnames::AbstractVector{Sym
                    categorical::Vector{Bool}, nrows::Integer) where T<:Type
     # upcast Vector{DataType} -> Vector{Type} which can hold CategoricalValues
     updated_types = convert(Vector{Type}, column_eltypes)
+    if length(categorical) != length(column_eltypes)
+        throw(ArgumentError("arguments column_eltypes and categorical must have the same length " *
```
ararslan: Isn't `DimensionMismatch` typically used rather than `ArgumentError` for cases such as this?

bkamins: right - I will fix it

PR #1341, opened 2018-01-08
Unbreak deprecated readtable(GZip)

andreasnoack:
I.e. use the `CodecZLib` API since we don't load `GZip` anymore. This broke the `RDatasets` tests.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1050,45 +1050,45 @@ function readtable(pathname::AbstractString;
     Base.depwarn("readtable is deprecated, use CSV.read from the CSV package instead",
                  :readtable)
 
+    _r(io) = readtable(io,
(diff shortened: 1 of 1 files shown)


src/deprecated.jl
```diff
@@ -1050,24 +1050,7 @@ function readtable(pathname::AbstractString;
...
-        error("BZip2 decompression not yet implemented")
-    # (4) Path is an uncompressed file
-    else
-        io = open(pathname, "r")
-        nbytes = filesize(pathname)
-    end
-
-    return readtable(io,
+    _r(io) = readtable(io,
                      nbytes,
```
nalimilan: Need to change indentation.

andreasnoack: Fixed

Issue #1280, opened 2017-11-21
Issues with Split-Apply and Missing (failing for combine under conditions)


Nosferican:
Let `obj` be some data frame which can be used in a split-apply procedure.
```
obj = DataFrame(x = repeat(1:2, inner = 5), t = repeat(1:5, outer = 2), y = 1:10)
```
If the result of the split-apply process yields any `Vector{Missing}` the combine process fails.
For example,
```
output = by(obj, :x) do subdf
    output = DataFrame()
    for col ∈ eachcol(subdf)
        output[col[1]] = mean(col[2])
    end
    output[:,1] = [missing]
    return output
end
---
type TypeofBottom has no field parameters

Stacktrace:
 [1] promote_col_type(...) at /.julia/v0.7/DataFrames/src/abstractdataframe/abstractdataframe.jl:686
 [2] _vcat(::Array{DataFrames.DataFrame,1}) at /.julia/v0.7/DataFrames/src/abstractdataframe/abstractdataframe.jl:763
 [3] combine(::DataFrames.GroupApplied{DataFrames.DataFrame}) at /.julia/v0.7/DataFrames/src/groupeddataframe/grouping.jl:203
 [4] #by#74(::Bool, ::Function, ::DataFrames.DataFrame, ::Symbol, ::Function) at /jbsc/.julia/v0.7/DataFrames/src/groupeddataframe/grouping.jl:303
 [5] (::getfield(DataFrames, Symbol("#kw##by")))(::Array{Any,1}, ::typeof(DataFrames.by), ::DataFrames.DataFrame, ::Symbol, ::Function) at ./<missing>:0
 [6] #by#75(::Bool, ::Function, ::Function, ::DataFrames.DataFrame, ::Any) at /.julia/v0.7/DataFrames/src/groupeddataframe/grouping.jl:305
 [7] by(::Function, ::DataFrames.DataFrame, ::Symbol) at /jbsc/.julia/v0.7/DataFrames/src/groupeddataframe/grouping.jl:305
```
Using the same process, but ensuring a dummy `Union` allows the desired output,
```
output = by(obj, :x) do subdf
    output = DataFrame()
    for col ∈ eachcol(subdf)
        output[col[1]] = mean(col[2])
    end
    output[:,1] = Vector{Union{Missing,Int64}}([missing])
    return output
end
---
	x	x_1	t	y
1	1	missing	3.0	3.0
2	2	missing	3.0	8.0
```
[...]

PR #1344, opened 2018-01-17
Fix combining DataFrames with a column of Missings

nicoleepp:
Fix for issue #1280 

I came across the same issue when using `vcat` to combine two DataFrames with a column of exclusively Missings so I figured I would just submit the fix.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -763,7 +763,7 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
 
 @generated function promote_col_type(cols::AbstractVector...)
     T = mapreduce(x -> Missings.T(eltype(x)), promote_type, cols)
-    if T <: CategoricalValue
+    if T <: CategoricalValue && T !== Union{}
         T = T.parameters[1]
     end
     if any(col -> eltype(col) >: Missing, cols)
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -764,7 +764,7 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
 @generated function promote_col_type(cols::AbstractVector...)
     T = mapreduce(x -> Missings.T(eltype(x)), promote_type, cols)
     if CategoricalArrays.iscatvalue(T)
-        T = T.parameters[1]
+        T = CategoricalArrays.leveltype(T)
     end
     if any(col -> eltype(col) >: Missing, cols)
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -763,7 +763,7 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
 
 @generated function promote_col_type(cols::AbstractVector...)
     T = mapreduce(x -> Missings.T(eltype(x)), promote_type, cols)
-    if T <: CategoricalValue
+    if T <: CategoricalValue && T !== Union{}
```
ararslan: You don't have to change this, I just want to note that this can be written as `Union{} <: T <: CategoricalValue`. Kind of fun. :slightly_smiling_face:

alyst: Hmm, on Julia 0.6.2:
```
julia> using CategoricalArrays

julia> Union{} <: Union{} <: CategoricalString
true

julia> Union{} <: CategoricalString{UInt32} <: CategoricalValue
false
```

Note the last comparison. It's confusing and unfortunate, but there could be categorical values types that are not `<: CategoricalValue` (in fact, `CategoricalString` is more frequent than `CategoricalValue`).
So the proper check is `CategoricalArrays.iscatvalue(T)`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -763,7 +763,7 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
 
 @generated function promote_col_type(cols::AbstractVector...)
     T = mapreduce(x -> Missings.T(eltype(x)), promote_type, cols)
-    if T <: CategoricalValue
+    if CategoricalArrays.iscatvalue(T)
         T = T.parameters[1]
```
alyst: Sorry, I haven't looked at the whole function when I gave the last comment.
This line would give incorrect result for e.g. `CategoricalString{UInt8}`.
The proper way should be `T = CategoricalArrays.leveltype(T)`

nalimilan: Good catch. Note that REQUIRE should also be raised to the latest CategoricalArrays version.

nicoleepp: @nalimilan It seems like REQUIRE already requires the latest CategoricalArrays (version 0.3.3)?

nalimilan:
@alyst Merge if it's OK for you.

nicoleepp:
Can a new version of DataFrames be tagged?

nalimilan:
Sure, see https://github.com/JuliaLang/METADATA.jl/pull/12965

PR #1354, opened 2018-02-15
Fix some Julia 0.7 deprecations

nalimilan:
Based on running FemtoCleaner locally.

### DIFF
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -287,7 +287,7 @@ module TestDataFrame
 
     # describe
     #suppress output and test that describe() does not throw
-    devmissing = is_unix() ? "/dev/null" : "nul"
+    devmissing = Compat.Sys.isunix() ? "/dev/null" : "nul"
     open(devmissing, "w") do f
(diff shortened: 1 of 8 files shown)


test/dataframe.jl
```diff
@@ -287,7 +287,7 @@ module TestDataFrame
 
     # describe
     #suppress output and test that describe() does not throw
-    devmissing = is_unix() ? "/dev/null" : "nul"
+    devmissing = Compat.Sys.isunix() ? "/dev/null" : "nul"
```
ararslan: Not sure why we aren't using `DevNull` here but that's an aside.

test/utils.jl
```diff
@@ -16,11 +16,11 @@ module TestUtils
 
     # Check that reserved words are up to date
 
-    f = "$JULIA_HOME/../../src/julia-parser.scm"
+    f = "$Sys.BINDIR/../../src/julia-parser.scm"
```
ararslan: I think this needs to be `Compat.Sys.BINDIR`.

nalimilan: Woops, due to the absence of parentheses this just interpolated `Sys`...

Issue #1359, opened 2018-02-23
Bug in view with colon


cstjean:
```julia
> @view df[1:10, :]
ArgumentError: idx[1] has type Colon; DataFrame only supports indexing columns with integers, symbols or boolean vectors
```

It's really minor, since `@view df[1:10]` and `@view df[1:10, 1:end]` work fine.

nalimilan:
Probably just a missing method somewhere.

bkamins:
In #1360 I have proposed missing methods.

bkamins:
@cstjean OK to close?

cstjean:
Yes, thank you.

PR #1360, opened 2018-02-24
allow colon for rows and columns in DataFrame view

bkamins:
PR filling missing functionality indicated in #1359.

### DIFF
--- a/src/subdataframe/subdataframe.jl
+++ b/src/subdataframe/subdataframe.jl
@@ -68,6 +68,10 @@ function SubDataFrame(parent::DataFrame, rows::T) where {T <: AbstractVector{Int
     return SubDataFrame{T}(parent, rows)
 end
 
+function SubDataFrame(parent::DataFrame, rows::Colon)
+    return SubDataFrame(parent, 1:nrow(parent))
+end
+
 function SubDataFrame(parent::DataFrame, row::Integer)
     return SubDataFrame(parent, [Int(row)])
 end
@@ -84,6 +88,10 @@ function SubDataFrame(sdf::SubDataFrame, rowinds::Union{T, AbstractVector{T}}) w
     return SubDataFrame(sdf.parent, sdf.rows[rowinds])
 end
 
+function SubDataFrame(sdf::SubDataFrame, rowinds::Colon)
+    return sdf
+end
+
 function Base.view(adf::AbstractDataFrame, rowinds::AbstractVector{T}) where {T >: Missing}
     # Vector{>:Missing} need to be checked for missings
     any(ismissing, rowinds) && throw(MissingException("missing values are not allowed in indices"))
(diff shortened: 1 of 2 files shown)


src/subdataframe/subdataframe.jl
```diff
@@ -94,12 +102,12 @@ function Base.view(adf::AbstractDataFrame, rowinds::Any)
...
 end
 
-function Base.view(adf::AbstractDataFrame, rowinds::Any, colinds::AbstractVector)
+function Base.view(adf::AbstractDataFrame, rowinds::Any, colinds::Union{Colon, AbstractVector})
     return SubDataFrame(adf[colinds], rowinds)
 end
 
-function Base.view(adf::AbstractDataFrame, rowinds::Any, colinds::Any)
-    return SubDataFrame(adf[[colinds]], rowinds)
+function Base.view(adf::AbstractDataFrame, rowinds::Any, colind::ColumnIndex)
```
nalimilan: Why change this? Would be good to document it in the commit message.

bkamins: This only changes the moment when an error is caught when column index is invalid as I thought it would be easier to understand error message for the user as actually we only allow `ColumnIndex` to be wrapped in a vector.

E.g. now we get the following error:
```
julia> view(df, 1:2, true)
ERROR: BoundsError: attempt to access DataFrames.Index(Dict(:x2=>2,:x3=>3,:x1=>1), Symbol[:x1, :x2, :x3])
  at index [true]
Stacktrace:
 [1] getindex(::DataFrames.Index, ::Array{Bool,1}) at src/other/index.jl:131
 [2] getindex(::DataFrames.DataFrame, ::Array{Bool,1}) at src/dataframe/dataframe.jl:246
 [3] view(::DataFrames.DataFrame, ::UnitRange{Int64}, ::Bool) at src/subdataframe/subdataframe.jl:102
```
which I guess would be unclear for a regular user. Now `MethodError` when trying to call `view` will be raised instead.

It can be changed to `Any` if you feel it is better. I understand that the benefit you see is that we keep column indexing logic in one place. However, we already have it spread out as we use `ColumnIndex` not only in `src/other/index.jl` but also in `src/dataframe/dataframe.jl`.

bkamins: Actually thinking about this more I find this change necessary as in a corner case of `Bool` for 1-column `DataFrame` it is significant. If we do not fix it we have the following problem:

```
julia> x = DataFrame(a=1:3)
3×1 DataFrames.DataFrame
│ Row │ a │
├─────┼───┤
│ 1   │ 1 │
│ 2   │ 2 │
│ 3   │ 3 │

julia> view(x, 1:3, true)
3×1 DataFrames.SubDataFrame{UnitRange{Int64}}
│ Row │ a │
├─────┼───┤
│ 1   │ 1 │
│ 2   │ 2 │
│ 3   │ 3 │

julia> x[1:3, true]
WARNING: Indexing with Bool values is deprecated except for Vector{Bool}
Stacktrace:
 [1] depwarn(::String, ::Symbol) at .\deprecated.jl:70
 [2] getindex at D:\Software\JULIA_PKG\v0.6\DataFrames\src\deprecated.jl:1295 [inlined]
 [3] getindex(::DataFrames.DataFrame, ::UnitRange{Int64}, ::Bool) at D:\Software\JULIA_PKG\v0.6\DataFrames\src\dataframe\dataframe.jl:269
 [4] eval(::Module, ::Any) at .\boot.jl:235
 [5] eval_user_input(::Any, ::Base.REPL.REPLBackend) at .\REPL.jl:66
 [6] macro expansion at .\REPL.jl:97 [inlined]
 [7] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at .\event.jl:73
while loading no file, in expression starting on line 0
3-element Array{Int64,1}:
 1
 2
 3
```
and `view` goes through silently although it should not.

nalimilan: OK. I just wanted to know the reason for the change, in general it should be documented in the commit message, or at least in the PR.

bkamins: Agreed 👍.

PR #1363, opened 2018-02-26
Fix tests for makeunique keyword and deprecated column indexing

bkamins:
This is a maintenance PR of test files adding:
* `makeunique` keyword where needed (except for one place with `getindex` taking the same column twice - this test case should be changed to throw an error after deprecation);
* cases of indexing into `Index` tests that are deprecated are removed and a list of cases to test for error after deprecation period is introduced; added tests for indices being `Integer` other than `Int`;
* in *join.jl* two parts of code are wrapped in `@testset` as they defined the same function twice.

### DIFF
--- a/src/other/index.jl
+++ b/src/other/index.jl
@@ -135,14 +135,16 @@ end
 function Base.getindex(x::AbstractIndex, idx::AbstractVector{Union{Bool, Missing}})
     if any(ismissing, idx)
         # TODO: this line should be changed to throw an error after deprecation
+        # throw(ArgumentError("missing values for column indexing are not alloved"))
         Base.depwarn("using missing in column indexing is deprecated", :getindex)
     end
     getindex(x, collect(Missings.replace(idx, false)))
 end
 
 function Base.getindex(x::AbstractIndex, idx::AbstractVector{<:Integer})
-    # TODO: this line should be changed to throw an error after deprecation
     if any(v -> v isa Bool, idx)
+        # TODO: this line should be changed to throw an error after deprecation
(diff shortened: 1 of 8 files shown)


test/dataframe.jl
```diff
@@ -58,6 +58,7 @@ module TestDataFrame
 
     z = vcat(v, x)
 
+    # TODO: change test to throw an error after deprecation
```
nalimilan: Without more details we may well miss it when removing the deprecation. Same below. Would be good to add a reference to the PR which introduced the deprecation, and a comment in deprecated.jl saying to look for them. Even better, prepare the code and keep it commented out while it's fresh in your mind, so that we know what to do later.

bkamins: Fixed - now deprecations are thrown in both cases so when we are past deprecation period they will not be missed as tests will throw an error.

Additionally in each case I have added expected error thrown in source (to be uncommented when removing deprecation) and accompanying test for this error.

Finally I removed two methods from *deprecated.jl* as they are covered in other places with a proper deprecation.

bkamins:
This PR additionally specifies the expected behavior after deprecation (currently commented in the code).

src/other/index.jl
```diff
@@ -135,14 +135,16 @@ end
 function Base.getindex(x::AbstractIndex, idx::AbstractVector{Union{Bool, Missing}})
     if any(ismissing, idx)
         # TODO: this line should be changed to throw an error after deprecation
+        # throw(ArgumentError("missing values for column indexing are not alloved"))
```
nalimilan: "allowed". I'd also put "for column indexing" at the end. Same below.

Issue #1347, opened 2018-01-30
Vcat automatic recognition of same columns name


DarioSarra:
I tried to look a lot for this method and it feels like in a previous version this was working, I hope I am not missing the existence of it. 
However:
I am trying to use vcat to combine different DataFrames, which have the same columns but in different orders. Wouldn't be nice if vcat could access the name list and mutually combine the columns based on the names instead of looking ony for the same exact order?
minimal example

```julia
a=DataFrame(x = [1,2,3], y = ["a","b","c"])
b=DataFrame( y = ["d","e","f"], x = [4,5,6])
vcat(a,b)
```

nalimilan:
That kind of makes sense (and that's how R and dplyr work with `rbind`). Though there's a tension between treating data frames as a vector of columns with an order or as a matrix, and treating them like a database. `vcat` is essentially a vector/matrix API, so we could also provide a different function to combine columns by name.

AFAICT the error was introduced by https://github.com/JuliaData/DataTables.jl/pull/45, but it doesn't seem we've discussed the opportunity of changing the behavior. Cc: @cjprybol

tbeason:
I also ran into this recently. I think it makes sense to provide a fallback for if the `df` columns are in a different order, perhaps using a permutation on `names`? Might be slightly slower, but could just document that it will be much faster if the columns are similarly ordered on the two (or many) dataframes.

For example, ignoring the necessary error-checking, something like the below would do the trick. 
```julia
permcols(dfa,dfb) = collect( findfirst(equalto(nma),names(dfb)) for nma in names(dfa))
``` 
Then `dfb[permcols(dfa,dfb)]` results in `dfb` having the same column order as `dfa`

nalimilan:
It's really not hard to do and performance should be the same, it's more of a design question.

DarioSarra:
I solved the problem reorganising the second dataframe in the order of the first one with a simple function

`concat_data!(a, b) = append!(a, b[:, names(a)])`

I guess this could be add as an alternative method if the function checks:
1 mismatch in col names
2 presence of the same names

nalimilan:
@bkamins Would you be willing to work on this?

bkamins:
I can do it, but first let us agree on the contract we want to follow comprehensively for those functions. Then we can decide on the implementation plan.

We have two functions that combine `DataFrames`:
* `vcat` creates a new `DataFrame`, can join more than two `DataFrames`, promotes types of columns;
[...]

PR #1366, opened 2018-02-28
vcat allowing column reordering

bkamins:
Implementation following discussion in #1347.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -787,13 +787,30 @@ end
 """
     vcat(dfs::AbstractDataFrame...)
 
-Vertically concatenate `AbstractDataFrames` that have the same column names in
-the same order.
+Vertically concatenate `AbstractDataFrames`.
+
+Column names in all passed data frames must be the same, but they can have
+different order. In such cases the order of names in the first passed
+`DataFrame` is used.
 
 # Example
 ```jldoctest
 julia> df1 = DataFrame(A=1:3, B=1:3);
 julia> df2 = DataFrame(A=4:6, B=4:6);
+julia> vcat(df1, df2)
+6×2 DataFrames.DataFrame
+│ Row │ A │ B │
+├─────┼───┼───┤
+│ 1   │ 1 │ 1 │
+│ 2   │ 2 │ 2 │
+│ 3   │ 3 │ 3 │
+│ 4   │ 4 │ 4 │
+│ 5   │ 5 │ 5 │
+│ 6   │ 6 │ 6 │
+
+
+julia> df1 = DataFrame(A=1:3, B=1:3);
+julia> df2 = DataFrame(B=4:6, A=4:6);
 julia> vcat(df1, df2)
 6×2 DataFrames.DataFrame
 │ Row │ A │ B │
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -784,8 +784,11 @@ end
 """
     vcat(dfs::AbstractDataFrame...)
 
-Vertically concatenate `AbstractDataFrames` that have the same column names in
-the same order.
+Vertically concatenate `AbstractDataFrames`.
+
+Column names in all passed `DataFrames` must be the same, but they can have
```
nalimilan: Better say "data frame" since they are not necessarily of the `DataFrame` type. "case" -> "cases" (I think)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -827,29 +843,25 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
...
-            data = [df[i] for df in dfs]
-            lens = map(length, data)
-            cols[i] = promote_col_type(data...)(sum(lens))
-            offset = 1
-            for j in 1:length(data)
-                copy!(cols[i], offset, data[j])
-                offset += lens[j]
-            end
+        # this should not happen
+        throw(ArgumentError("Unexpected error - please report steps allowing to reproduce it"))
```
nalimilan: Better do `@assert !isempty(coldiff)`, as `ArgumentError` is supposed to reflect an error made by the caller. Or just remove the check, as it could only happen if there was a bug in Base, and we can't check all possible bugs... ;-)

test/cat.jl
```diff
@@ -120,11 +120,11 @@ module TestCat
     @test size(vcat(df, df, df)) == (size(df, 1) * 3, size(df, 2))
 
     alt_df = deepcopy(df)
-    vcat(df, alt_df)
+    @test size(vcat(df, alt_df)) == (8, 3)
```
nalimilan: It would be even better to check the actual result, right? It's weird that the existing code didn't...

src/abstractdataframe/abstractdataframe.jl

test/cat.jl
bkamins: added (I have added a similar test earlier elsewhere also).

src/abstractdataframe/abstractdataframe.jl
bkamins: Finally I have made a bit different implementation based on `coldiff` without using `Set`s which should be safe.

bkamins:
So it remains to be decided if we want to add `reorder` keyword argument to `vcat`. I am not sure we really need it as typically you would call `vcat` by `[df1; df2]` where it is not possible to pass a keyword argument anyway.

But if you feel that it is worth to have it I can add it.

test/cat.jl
```diff
@@ -120,11 +120,13 @@ module TestCat
...
     alt_df = deepcopy(df)
-    vcat(df, alt_df)
+    @test vcat(df, alt_df) == DataFrame([[3.0,2.0,3.0,3.0,3.0,2.0,3.0,3.0],
+                                         [2.0,2.0,1.0,3.0,2.0,2.0,1.0,3.0],
+                                         [2,2,2,3,2,2,2,3]])
 
     # Don't fail on non-matching types
     df[1] = zeros(Int, nrow(df))
-    vcat(df, alt_df)
+    @test size(vcat(df, alt_df)) == (8, 3)
```
nalimilan: Could you also test the result here? That would also make sense below even if that's cumbersome (maybe we don't need to test all long combinations if they are redundant). Else coverage makes it look like we test some code while we actually only check that it doesn't throw an error, which is misleading.

nalimilan:
I don't think we really need to add a keyword argument. At least there's no regression if we don't add it, and it's easy to add if somebody requests it.

test/cat.jl
bkamins: added

bkamins:
Agreed

nalimilan:
How about checking contents for other tests too? :-p

bkamins:
I have not added them because the tests have to be rewritten as now the contents would not tell us anything (I have already added one meaningful test because of this).
I will rewrite them.

bkamins:
Added

bkamins:
Rebased and should be good to merge (however, recent Julia 0.7 upgrade PR made significant changes so it would not hurt to have a quick look before merging).

Issue #1372, opened 2018-03-04
Coverage regression between 0.11.3 and 0.11.4


nalimilan:
It appears that some lines are not covered by tests anymore, while they used to be in 0.11.3. For example, compare this line in 0.11.3:
https://coveralls.io/builds/14794235/source?filename=src%2Fdataframe%2Fdataframe.jl#L401
with 0.11.4:
https://coveralls.io/builds/14956234/source?filename=src%2Fdataframe%2Fdataframe.jl#L403

@bkamins I guess it could be related to your recent indexing PRs, but I couldn't identify the source of the drop in coverage (CI is green for all commits). For reference, the commits between these two versions are https://github.com/JuliaData/DataFrames.jl/compare/v0.11.3...v0.11.4

bkamins:
There are minimal reasons from my PR (e.g. we do not test adding the same variable to data frame more than two times like `DataFrame([[1], [2], [3]], [:x, :x, :x])` or we are not testing all deprecated paths).

I will simply add tests to things that I can catch from the listing you gave:
https://coveralls.io/builds/14956234/source?filename=src%2Fdataframe%2Fdataframe.jl#L403

bkamins:
I will split my changes into several PRs as apparently it is not only about improving test coverage :) and actually some bugs have to be fixed in the process, see #1373.

bkamins:
Some additional tests require deciding if `setindex!`:
1. taking `AbstractVector{Bool}` should ensure that that the length of the vector is correct (it does not do that now);
2. `AbstractVector{Union{Missing, Bool}}` should be accepted if actually it contains only `Bool` values;

This is how `getindex` now works.

Additionally if we decide on any as `yes` then should it go through deprecation or simply implement new functionality (treating current different behavior as a bug).

nalimilan:
I'd say yes to both. What currently happens in case 1? I'm not sure we need a deprecation given that it's really a weird corner case.

bkamins:
`setindex!` does not check the length of passed `AbstractVector{Bool}` which means:
1. if it is too short it is simply accepted silently;
2. if it is too long it is simply accepted silently if all not inbounds values are `false` or throws an error if any of not inbounds values is `true`;

In general I would say that if someone passes such a vector as index it is a mistake (there is no reason to do something like this) so I would simply throw an error in such cases without deprecating the above behavior.

nalimilan:
Agreed, I can't imagine a valid use of that behavior.

[...]

PR #1375, opened 2018-03-05
improve push! and categorical! test coverage

bkamins:
Part three of implementation of #1372.

Changes:
* full coverage of `push!`
* full coverage of `categorical!`
* removed `push!` method that was not needed (duplicate in special case with no extra functionality and the same performance)
* cleaned up deprecated `Associative` references
* corrected documentation of `DataFrame(ds::AbstractDict)` constructor (it does not accept a vector but a dictionary)

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -16,7 +16,7 @@ DataFrame(t::Type, nrows::Integer, ncols::Integer) # an empty DataFrame of arbit
 DataFrame(column_eltypes::Vector, names::Vector, nrows::Integer; makeunique::Bool=false)
 DataFrame(column_eltypes::Vector, cnames::Vector, categorical::Vector, nrows::Integer;
           makeunique::Bool=false)
-DataFrame(ds::Vector{AbstractDict})
+DataFrame(ds::AbstractDict)
 ```
 
 **Arguments**
@@ -33,7 +33,7 @@ DataFrame(ds::Vector{AbstractDict})
 * `column_eltypes` : elemental type of each column
 * `categorical` : `Vector{Bool}` indicating which columns should be converted to
                   `CategoricalVector`
-* `ds` : a vector of Associatives
+* `ds` : `AbstractDict`
 
 Each column in `columns` should be the same length.
 
@@ -620,7 +620,7 @@ Base.setindex!(df::DataFrame, x::Nothing, col_ind::Int) = delete!(df, col_ind)
 
 ##############################################################################
 ##
-## Mutating Associative methods
+## Mutating AbstractDict methods
 ##
 ##############################################################################
 
@@ -721,7 +721,7 @@ merge!(df::DataFrame, others::AbstractDataFrame...)
 For every column `c` with name `n` in `others` sequentially perform `df[n] = c`.
 In particular, if there are duplicate column names present in `df` and `others`
 the last encountered column will be retained.
-This behavior is identical with how `merge!` works for any `Associative` type.
+This behavior is identical with how `merge!` works for any `AbstractDict` type.
 Use `join` if you want to join two `DataFrame`s.
 
 **Arguments**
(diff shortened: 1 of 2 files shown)


src/dataframe/dataframe.jl
```diff
@@ -33,7 +33,7 @@ DataFrame(ds::Vector{AbstractDict})
 * `column_eltypes` : elemental type of each column
 * `categorical` : `Vector{Bool}` indicating which columns should be converted to
                   `CategoricalVector`
-* `ds` : a vector of Associatives
+* `ds` : `AbstractDict`
```
nalimilan: Specify "of columns" or something like that?

src/dataframe/dataframe.jl
```diff
@@ -952,23 +952,6 @@ Base.convert(::Type{DataFrame}, d::AbstractDict) = DataFrame(d)
 ##
 ##############################################################################
 
-function Base.push!(df::DataFrame, associative::AbstractDict{Symbol,Any})
```
nalimilan: AFAICT this method was provided separately because it only works with `Symbol` keys, while the one below only works with `AbstractString` keys. The latter is unusual since I think we generally require symbols, so it could be deprecated (and its `get` call is really weird).

test/dataframe.jl
```diff
@@ -405,6 +427,18 @@ module TestDataFrame
     @test findfirst(c -> typeof(c) <: CategoricalVector{Union{Int, Missing}},
                     categorical!(deepcopy(df), 1).columns) == 1
 
+    @testset "categorical!" begin
+        df = DataFrame([["a", "b"], ['a', 'b'], [true, false], 1:2, ["x", "y"]])
+        @test eltypes(categorical!(df)) == [CategoricalArrays.CategoricalString{UInt32},
```
nalimilan: Better not hardcode `UInt32` as the default could change in the future. I think in other tests we used another approach based on `isa.(eltypes(...), [CategoricalValue{Char}, ...])`.

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl
bkamins: OK. But - as noted above - this `get` has the same performance as `getindex` and `push!` for `:Symbol` was implemented in a bad way as it required `Any` value. I will revert and fix it and deprecate the other one.

test/dataframe.jl

src/dataframe/dataframe.jl
bkamins: I am removing the tests for deprecated version of the method.

bkamins: Actually I had to use other approach for deprecation as we can have `Dict{Any}` or `Dict{Union{Symbol, Missing}}` that actually contains only `Symbol` as keys.

src/dataframe/dataframe.jl
```diff
@@ -952,28 +952,15 @@ Base.convert(::Type{DataFrame}, d::AbstractDict) = DataFrame(d)
...
-        i += 1
-    end
-end
-
 function Base.push!(df::DataFrame, associative::AbstractDict)
     i = 1
     for nm in _names(df)
         try
-            val = get(() -> associative[string(nm)], associative, nm)
+            val = get(() -> (Base.depwarn("push!(::DataFrame, ::AbstractDict) with"*
```
ararslan: Should add a space at the end of the string, otherwise with the concatenation you'll get `withAbstractDict`.

test/dataframe.jl
bkamins: actually this pattern did not go through so I used `all(map(<:, eltypes(...), [...]))` approach.

src/dataframe/dataframe.jl
```diff
@@ -952,28 +952,15 @@ Base.convert(::Type{DataFrame}, d::AbstractDict) = DataFrame(d)
...
-        i += 1
-    end
-end
-
 function Base.push!(df::DataFrame, associative::AbstractDict)
     i = 1
     for nm in _names(df)
         try
-            val = get(() -> associative[string(nm)], associative, nm)
+            val = get(() -> (Base.depwarn("push!(::DataFrame, ::AbstractDict) with "*
```
nalimilan: Won't this warning be printed when `associative` doesn't contain a key for one of `df`'s columns? Then it will be confusing to print this before an error.

BTW, you can use the `do` syntax. Also `associative` could be renamed to `d` or `dict`.

bkamins: good catch. will fix

Issue #1370, opened 2018-03-02
missing `lastindex` method


ExpandingMan:
`AbstractDataFrame` is missing the `lastindex(::DataFrame, ::Integer)` method.  Without it you can't do e.g. `df[1:end, 1]`.

bkamins:
Actually also `axes` method is missing also. I will make a PR adding both.

bkamins:
See PR #1376. @ExpandingMan tests, especially under 0.7 are welcome (it passes on my version of Julia 0.7, but the ecosystem evolves ...).

bkamins:
close?

ExpandingMan:
Great, thanks.

PR #1376, opened 2018-03-06
make end work under Julia 0.7

bkamins:
following issue #1370

### DIFF
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -582,4 +582,32 @@ module TestDataFrame
         df[4] = [1,2,3]
         @test names(df) == [:x3, :x3_1, :x3_2, :x4]
     end
+
+    @testset "handling of end in indexing" begin
+        z = DataFrame(rand(4,5))
+        for x in [z, view(z, 1:4)]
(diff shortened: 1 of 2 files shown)


test/dataframe.jl
```diff
@@ -582,4 +582,32 @@ module TestDataFrame
         df[4] = [1,2,3]
         @test names(df) == [:x3, :x3_1, :x3_2, :x4]
     end
+
+    @testset "handling of end in indexing" begin
+        z = DataFrames.DataFrame(rand(4,5))
```
nalimilan: Why the `DataFrames` prefix? Same below.

Otherwise, looks good.

bkamins: leftover from initial testing. fixed.

ExpandingMan:
Would be great if we could merge this as currently missing these methods is breaking all sorts of things in 0.7 :wink:.

bkamins:
@ExpandingMan I would propose that you make a separate PR for `STDOUT` as it is an unrelated change to `end` so it will be better to keep it as a different commit to master. And you do not have to wait till this PR is merged.

Issue #1369, opened 2018-03-02
should `getindex` type semantics be cleaned up?


ExpandingMan:
Currently `getindex` methods, particularly for `Index` accept a variety of container types and in many cases they iterate over the contents of the container to check for correctness rather than simply dispatching on the type.  For example, one can do `getindex(::Index, ::Vector{Union{Int,Missing}})`.  I have no idea what the consensus is on this (if one exists) but this issue is to open up discussion with the hope of cleaning up the methods.  The main concern here I think is programatically generated indices, i.e. from data frame operations themselves.

My own attitude toward this is that we should be fairly restrictive.  I suspect that a major concern will be that many users are probably extremely familiar with pandas, and the type situation in pandas is basically total chaos for obvious reasons.  However, as people get more familiar with Julia I have a feeling they tend to have more type-correct code, and it seems like a small ask to convert columns when they are being used for indexing.  In every case I can think of, if the conversion throws an error, so will the indexing operation.

nalimilan:
As I noted at https://github.com/JuliaData/DataFrames.jl/pull/1364#discussion_r171851466: Yes, I think it makes sense to ignore the type of the array and look only at the actual type of the elements. Else it can be cumbersome given that Julia is a dynamic language. That's how indexing arrays works.

bkamins:
Current implementation follows @nalimilan comment - i.e. it is fast for "clean" types like `AbstractVector{Int}`, `AbstractVector{Symbol}`, `AbstractVector{Bool}` and is slower but working for all nasty types that can be passed to `getindex` as long as the data inside those types is correct.

@ExpandingMan do I understand your proposal correctly to remove support for all those "unclean" types? As @nalimilan noted this is not how Base works, e.g.:
```
julia> x = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> x[Any[1, 3]]
2-element Array{Int64,1}:
 1
 3
```

the execption is that we handle `Union{Bool, Missing}` and base does not accept it:
```
julia> x[Union{Bool,Missing}[true, true, true]]
┌ Warning: indexing with Bool values is deprecated. Convert the index to an integer first with `Int(i)`.
│   caller = macro expansion at multidimensional.jl:657 [inlined]
└ @ Core multidimensional.jl:657
3-element Array{Int64,1}:
 1
 1
 1
```
Maybe then we should remove this support? (or change the way Base works)

ExpandingMan:
[...]

PR #1380, opened 2018-03-08
WIP: improve DataFrame indexing

bkamins:
Following #1369 and #1372 I have proposed changes to `setindex!` (and a small fixe in `getindex` depwarn) to make it more consistent.
It does not have a full test coverage yet and I have added comments in the code where the behavior is unexpected.

Finally I am not sure if we want `df[1] = nothing` to remove columns from a `DataFrame` (this is the current behavior which makes `df[1]=x` inconsistent (it has different behavior if `x=nothing` than when e.g. `x=missing` or `x="abracadabra"`))? It is also inconsistent with `AbstractDict` interface.

When we decide if we want what I have implemented (or something else) I will update this PR and add appropriate tests.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -339,6 +339,12 @@ function insert_single_column!(df::DataFrame,
             push!(index(df), col_ind)
             push!(df.columns, dv)
         else
+            if !(typeof(col_ind) <: Integer)
+                throw(ArgumentError("Only Integer numbers accepted for column indexing"))
+            end
+            if isa(col_ind, Bool)
+                throw(ArgumentError("Bool values are not allowed for column indexing"))
+            end
             if ncol(df) + 1 == Int(col_ind)
                 push!(index(df), nextcolname(df))
                 push!(df.columns, dv)
@@ -351,11 +357,17 @@ function insert_single_column!(df::DataFrame,
 end
 
 function insert_single_entry!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnIndex)
+    if !(typeof(col_ind) <: Union{Integer,Symbol})
+        throw(ArgumentError("Only Integer numbers accepted for column indexing"))
+    end
+    if isa(col_ind, Bool)
+        throw(ArgumentError("Bool values are not allowed for column indexing"))
+    end
     if haskey(index(df), col_ind)
         df.columns[index(df)[col_ind]][row_ind] = v
         return v
     else
-        error("Cannot assign to non-existent column: $col_ind")
+        throw(ArgumentError("Cannot assign to non-existent column: $col_ind"))
     end
 end
 
@@ -383,11 +395,18 @@ end
 # df[SingleColumnIndex] = AbstractVector
 function Base.setindex!(df::DataFrame, v::AbstractVector, col_ind::ColumnIndex)
     insert_single_column!(df, v, col_ind)
+    return df
 end
 
 # df[SingleColumnIndex] = Single Item (EXPANDS TO NROW(df) if NCOL(df) > 0)
 function Base.setindex!(df::DataFrame, v, col_ind::ColumnIndex)
     if haskey(index(df), col_ind)
+        if !(typeof(col_ind) <: Union{Integer,Symbol})
+            throw(ArgumentError("Only Integer numbers accepted for column indexing"))
+        end
+        if isa(col_ind, Bool)
+            throw(ArgumentError("Bool values are not allowed for column indexing"))
+        end
         fill!(df[col_ind], v)
     else
         insert_single_column!(df, upgrade_scalar(df, v), col_ind)
@@ -396,12 +415,15 @@ function Base.setindex!(df::DataFrame, v, col_ind::ColumnIndex)
 end
 
 # df[MultiColumnIndex] = DataFrame
-function Base.setindex!(df::DataFrame, new_df::DataFrame, col_inds::AbstractVector{Bool})
+function Base.setindex!(df::DataFrame, new_df::DataFrame,
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, new_df, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         col_inds::AbstractVector{<:ColumnIndex})
+    # it is allowed that new_df has more columns than required
     for j in 1:length(col_inds)
         insert_single_column!(df, new_df[j], col_inds[j])
     end
@@ -409,12 +431,17 @@ function Base.setindex!(df::DataFrame,
 end
 
 # df[MultiColumnIndex] = AbstractVector (REPEATED FOR EACH COLUMN)
-function Base.setindex!(df::DataFrame, v::AbstractVector, col_inds::AbstractVector{Bool})
+function Base.setindex!(df::DataFrame, v::AbstractVector,
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, v, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
                         v::AbstractVector,
                         col_inds::AbstractVector{<:ColumnIndex})
+    # this allows to add arbitrarily many columns to a DataFrame, e.g.
+    # x = DataFrame(); x[1:10] = rand(3)
+    # will create 10 new columns in x
     for col_ind in col_inds
         df[col_ind] = v
     end
@@ -424,29 +451,35 @@ end
 # df[MultiColumnIndex] = Single Item (REPEATED FOR EACH COLUMN; EXPANDS TO NROW(df) if NCOL(df) > 0)
 function Base.setindex!(df::DataFrame,
                         val::Any,
-                        col_inds::AbstractVector{Bool})
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, val, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame, val::Any, col_inds::AbstractVector{<:ColumnIndex})
+    # this allows to add arbitrarily many columns to a DataFrame, e.g.
+    # x = DataFrame(); x[1:10] = 3
+    # will create 10 new columns in x
     for col_ind in col_inds
         df[col_ind] = val
     end
     return df
 end
 
-# df[:] = AbstractVector or Single Item
+# df[:] = DataFrame, AbstractVector or Single Item
 Base.setindex!(df::DataFrame, v, ::Colon) = (df[1:size(df, 2)] = v; df)
 
 # df[SingleRowIndex, SingleColumnIndex] = Single Item
 function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnIndex)
     insert_single_entry!(df, v, row_ind, col_ind)
+    df
 end
 
 # df[SingleRowIndex, MultiColumnIndex] = Single Item
 function Base.setindex!(df::DataFrame,
                         v::Any,
                         row_ind::Real,
-                        col_inds::AbstractVector{Bool})
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, v, row_ind, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
@@ -463,13 +496,15 @@ end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         row_ind::Real,
-                        col_inds::AbstractVector{Bool})
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, new_df, row_ind, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         row_ind::Real,
                         col_inds::AbstractVector{<:ColumnIndex})
+    nrow(new_df) == 1 || throw(ArgumentError("Source DataFrame contains multiple rows"))
     for j in 1:length(col_inds)
         insert_single_entry!(df, new_df[j][1], row_ind, col_inds[j])
     end
@@ -510,7 +545,8 @@ end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         row_inds::AbstractVector{Bool},
-                        col_inds::AbstractVector{Bool})
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, new_df, findall(row_inds), findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
@@ -522,7 +558,8 @@ end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         row_inds::AbstractVector{<:Real},
-                        col_inds::AbstractVector{Bool})
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, new_df, row_inds, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
@@ -530,7 +567,7 @@ function Base.setindex!(df::DataFrame,
                         row_inds::AbstractVector{<:Real},
                         col_inds::AbstractVector{<:ColumnIndex})
     for j in 1:length(col_inds)
-        insert_multiple_entries!(df, new_df[:, j], row_inds, col_inds[j])
+        insert_multiple_entries!(df, new_df[j], row_inds, col_inds[j])
     end
     return df
 end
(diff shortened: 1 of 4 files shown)


bkamins:
Having though more about it I also do not like that:

```
df = DataFrame(x=1,y=2)
df[3:4] = 3
```
works. I guess it is copied from R, but it actually is not intuitive for me. Any thoughts on this?

src/dataframe/dataframe.jl
```diff
@@ -396,25 +415,33 @@ function Base.setindex!(df::DataFrame, v, col_ind::ColumnIndex)
...
-function Base.setindex!(df::DataFrame, v::AbstractVector, col_inds::AbstractVector{Bool})
+function Base.setindex!(df::DataFrame, v::AbstractVector,
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, v, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
                         v::AbstractVector,
                         col_inds::AbstractVector{<:ColumnIndex})
+    # this allows to add arbitrarily many columns to a DataFrame, e.g.
```
nalimilan: That doesn't sound too bad. I'm not sure there are similar structures in Julia where it would make sense to use the same approach (dicts can't because ranges are valid keys).

src/dataframe/dataframe.jl
```diff
@@ -463,13 +496,15 @@ end
...
-                        col_inds::AbstractVector{Bool})
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, new_df, row_ind, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         row_ind::Real,
                         col_inds::AbstractVector{<:ColumnIndex})
+    nrow(new_df) == 1 || throw(ArgumentError("Source DataFrame contains multiple rows"))
```
nalimilan: Better say "should contain a single row" (and no initial uppercase).

test/dataframe.jl
```diff
@@ -617,6 +617,28 @@ module TestDataFrame
...
+        df[Union{Bool,Missing}[true, false, false, false]] = DataFrame([[100,100,100]])
+        @test df[1] == [100,100,100]
+        df[Union{Bool,Missing}[true, false, false, false]] = [0,0,0]
+        @test df[1] == [0,0,0]
+        df[Union{Bool,Missing}[true, false, false, false]] = 1000
+        @test df[1] == [1000,1000,1000]
+
+        df1 = DataFrame()
+        df1[1:10] = rand(3)
+        @test size(df1) == (3, 10)
```
nalimilan: Better also test contents (with a smaller number of columns).

nalimilan:
> Any comments on this?

I don't find it too bad, especially since there is no simple alternative. Not sure.

nalimilan:
> Finally I am not sure if we want df[1] = nothing to remove columns from a DataFrame (this is the current behavior which makes df[1]=x inconsistent (it has different behavior if x=nothing than when e.g. x=missing or x="abracadabra"))? It is also inconsistent with AbstractDict interface.

Agreed, this would better be deprecated (I wonder when and why it was added).

src/dataframe/dataframe.jl
```diff
@@ -339,6 +339,12 @@ function insert_single_column!(df::DataFrame,
             push!(index(df), col_ind)
             push!(df.columns, dv)
         else
+            if !(typeof(col_ind) <: Integer)
```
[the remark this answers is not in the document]

src/dataframe/dataframe.jl
```diff
@@ -339,6 +339,12 @@ function insert_single_column!(df::DataFrame,
...
+                throw(ArgumentError("Only Integer numbers accepted for column indexing"))
+            end
+            if isa(col_ind, Bool)
+                throw(ArgumentError("Bool values are not allowed for column indexing"))
```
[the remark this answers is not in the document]

src/dataframe/dataframe.jl
```diff
@@ -383,11 +395,18 @@ end
 # df[SingleColumnIndex] = AbstractVector
 function Base.setindex!(df::DataFrame, v::AbstractVector, col_ind::ColumnIndex)
     insert_single_column!(df, v, col_ind)
+    return df
```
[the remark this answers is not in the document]
bkamins: fixed, and changed `return df` in the function below.

src/dataframe/dataframe.jl
```diff
@@ -396,25 +415,33 @@ function Base.setindex!(df::DataFrame, v, col_ind::ColumnIndex)
...
-function Base.setindex!(df::DataFrame, new_df::DataFrame, col_inds::AbstractVector{Bool})
+function Base.setindex!(df::DataFrame, new_df::DataFrame,
+                        col_inds::AbstractVector{<:Union{Bool,Missing}})
+    length(col_inds) != ncol(df) && throw(BoundsError(df, col_inds))
     setindex!(df, new_df, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         col_inds::AbstractVector{<:ColumnIndex})
+    # it is allowed that new_df has more columns than required
```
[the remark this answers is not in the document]
bkamins: Actually we do not allow columns in different orders on RHS now (RHS is treated as if it were a matrix ignoring column names on RHS):
```
julia> x = DataFrame([1:3, 4:6, 7:9])
3×3 DataFrames.DataFrame
│ Row │ x1 │ x2 │ x3 │
├─────┼────┼────┼────┤
│ 1   │ 1  │ 4  │ 7  │
│ 2   │ 2  │ 5  │ 8  │
│ 3   │ 3  │ 6  │ 9  │

julia> y = DataFrame([11:13, 14:16, 17:19])
3×3 DataFrames.DataFrame
│ Row │ x1 │ x2 │ x3 │
├─────┼────┼────┼────┤
│ 1   │ 11 │ 14 │ 17 │
│ 2   │ 12 │ 15 │ 18 │
│ 3   │ 13 │ 16 │ 19 │

julia> x[[:x3]] = y
3×3 DataFrames.DataFrame
│ Row │ x1 │ x2 │ x3 │
├─────┼────┼────┼────┤
│ 1   │ 11 │ 14 │ 17 │
│ 2   │ 12 │ 15 │ 18 │
│ 3   │ 13 │ 16 │ 19 │

julia> x
3×3 DataFrames.DataFrame
│ Row │ x1 │ x2 │ x3 │
├─────┼────┼────┼────┤
│ 1   │ 1  │ 4  │ 11 │
│ 2   │ 2  │ 5  │ 12 │
│ 3   │ 3  │ 6  │ 13 │
```
so I am adding a deprecation for wrong length and this behavior should be documented.

src/dataframe/dataframe.jl
bkamins: OK - leaving it, but it should be documented.

src/dataframe/dataframe.jl

test/dataframe.jl

bkamins:
In general we should have a documentation of `getindex`/`setindex!` somewhere.
There are several corner cases that are non-obvious in my opinion.
What is the right place?

I can put the details in https://github.com/bkamins/Julia-DataFrames-Tutorial of course, but I guess it should be rather in main package?

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: OK, so this follows what happens in R. It's kind of weird not to check the column names,  but well...

nalimilan:
Why not put them in docstrings?

src/dataframe/dataframe.jl
```diff
@@ -388,33 +398,51 @@ end
...
     setindex!(df, new_df, findall(col_inds))
 end
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         col_inds::AbstractVector{<:ColumnIndex})
+    if length(col_inds) != ncol(new_df)
+        Base.depwarn("Number of columns in RHS DataFrame should be equal to number of assignments.", :insert!)
+    end
+    # after deprecation replace above warning by
+    # length(col_inds) != ncol(new_df) && throw(ArgumentError("Number of columns in RHS DataFrame must be equal to number of assignments"))
```
nalimilan: No initial uppercase (also elsewhere). Also it would be nice to print the actual number of columns on both sides, that makes it much easier to understand what has gone wrong.

bkamins:
I have added docstring, fixed warning/error messages and made correct range checking for row indexing by `AbstractVector{Bool}`.

bkamins:
I have made a lot of small changes. If you prefer push patches to this commit (especially if there are language - related comments as the text is complex and I am not a native). Of course you can leave comments if you prefer

src/dataframe/dataframe.jl
```diff
@@ -394,65 +393,78 @@ end
...
-* if `v` is `AbstractVector` its length must be equal to number of rows selected by `row_ind`
-  and its elements must be convertible to a type of elements in the column assigned to;
-  otherwise `v` is treated as scalar and broadcasted to all entries of a target column except
-  when it is a `DataFrame` in which case the same rules as for column assignment of `DataFrame`
-  apply with an additional restriction that `v` must have the same number of rows as selected
-  by `row_ind`;
-* if `row_ind` is `:` then it behaves exactly like `setindex!` without row selection except
-  when `v` is a `DataFrame` in which case `df[:]` and `df[:,:]` performs a copy of `v` to `df`
-
-In all above rules if `DataFrame` is mentioned on right hand side of an assignment it may not
```
nalimilan: I've removed this paragraph since this sounds like a bug to me, isn't it? If so we should file an issue.

src/dataframe/dataframe.jl
```diff
@@ -394,65 +393,78 @@ end
...
-Otherwise `v` is used as a column vector (no data copying is performed).
-`col_ind` can be either a `Symbol` or `Integer` (except `Bool`).
-If `col_ind` is present in a `DataFrame` then:
-* if `v` is an abstract vector then it replaces `df[col_ind]`
-* if `v` i not an abstract vector then existing `df[col_ind]` is filled with `v`
+Assign `v` to column `col_ind` of data frame `df`.
+
+If column `col_ind` is already present in `df` then:
+* if `v` is an `AbstractVector` then it replaces `df[col_ind]` (no copy is made)
+* otherwise, the existing `df[col_ind]` vector is filled with `v`
```
nalimilan: Is this behavior really desirable? `df[col_ind]` accesses the column vector, so it sounds like assigning to this should replace the vector, not fill it with values.

Often the best rules are the ones which are simple to describe. Here it would be simpler if we didn't have to distinguish the case where the column does not exist and the case where it does (except to describe how and where new columns are created).

What do you think?

src/dataframe/dataframe.jl
```diff
@@ -394,65 +393,78 @@ end
...
+If `v` is not a data frame then assignment rules follow single-column `setindex!` .
 
-You can create new columns in `df` this way and the rules are the same as for `setindex!`
-on a single column.
+If `v` is a data frame then each column in `v` is assigned sequentially to the column
+at the corresponding position in `col_inds`, disregarding column names.
+`df` and `v` must have the same number of rows, and the number of columns in `v`
+must be equal to the length of `col_inds`.
 
-If `col_inds<:AbstractVector{Union{Bool, Missing}}` then it must only contain `Bool` values
```
nalimilan: I've removed this because it would be long to explain that we don't care about the actual type of vectors, only about their contents, except for logical indexing, and because mentioning `Missing` could be misinterpreted as saying that missing values are accepted. Probably better simply mention the type of the actual elements.

src/dataframe/dataframe.jl
```diff
@@ -326,10 +327,11 @@ end
 # Will automatically add a new column if needed
 function insert_single_column!(df::DataFrame,
                                dv::AbstractVector,
-                               col_ind::ColumnIndex)
+                               col_ind::Union{Integer, Symbol})
```
nalimilan: I think I had noted this before: wouldn't it be better to change `ColumnIndex` once for all?

bkamins:
Thank you for the comments. Having read your modifications - and reflecting a bit more on `setindex!/getindex` I would reimplement the whole functionality to simplify the design and make it consistent.

We have bugs like:

```
julia> x = DataFrame([1:3,4:6,7:9])
3×3 DataFrames.DataFrame
│ Row │ x1 │ x2 │ x3 │
├─────┼────┼────┼────┤
│ 1   │ 1  │ 4  │ 7  │
│ 2   │ 2  │ 5  │ 8  │
│ 3   │ 3  │ 6  │ 9  │

julia> y = view(x, 1:2, 1:2)
2×2 DataFrames.SubDataFrame{UnitRange{Int64}}
│ Row │ x1 │ x2 │
├─────┼────┼────┤
│ 1   │ 1  │ 4  │
│ 2   │ 2  │ 5  │

julia> y[1:3] = 100
ERROR: Cannot assign to non-existent column: 3
Stacktrace:
 [1] insert_multiple_entries!(::DataFrames.DataFrame, ::Int64, ::UnitRange{Int64}, ::Int64) at \dataframe\dataframe.jl:355
 [2] setindex!(::DataFrames.DataFrame, ::Int64, ::UnitRange{Int64}, ::UnitRange{Int64}) at \dataframe\dataframe.jl:576
 [3] setindex!(::DataFrames.SubDataFrame{UnitRange{Int64}}, ::Int64, ::UnitRange{Int64}) at \subdataframe\subdataframe.jl:126

julia> x
3×3 DataFrames.DataFrame
│ Row │ x1  │ x2  │ x3 │
├─────┼─────┼─────┼────┤
│ 1   │ 100 │ 100 │ 7  │
│ 2   │ 100 │ 100 │ 8  │
│ 3   │ 3   │ 6   │ 9  │
```
(i.e. `setindex!` for `SubDataFrame` is not atomic)

and probably more, because the design of `setindex!/getindex` is complex.

I can write down the proposal setting all the rules so that they are consistent and implement them later in a generic way (now e.g `setindex!` functionality is 400 lines of code because there is a lot of boilerplate code which additionally introduces unexpected corner cases). In general the new rules for normal usage would be exactly the same, but would handle those special cases consistently (e.g I find not accepting `AbstractDataFrame` on RHS strange, or `df[:,:] = df2` overwriting `df` non intuitive, we do not match columns by name when `RHS` is a `DataFrame` etc.). However, the problem is that we will get some breaking changes that will be hard to track with deprecation warnings (as the code would be rewritten). I do not know if this would be acceptable (we could e.g. write a info message every time DataFrames is loaded stating that there are new `setindex!/getindex` rules). Would this be OK?

ExpandingMan:
I actually like the `df[:col] = nothing` thing for deleting columns.  It's already been established that `nothing` has some different behaviors than other types (e.g. it can't be use as a `Dict` key).  Aside from this, I don't even know if there is an easier way of dropping columns now than `df[filter(x -> x != col, names(df))]`.

nalimilan:
> I can write down the proposal setting all the rules so that they are consistent and implement them later in a generic way (now e.g setindex! functionality is 400 lines of code because there is a lot of boilerplate code which additionally introduces unexpected corner cases). In general the new rules for normal usage would be exactly the same, but would handle those special cases consistently (e.g I find not accepting AbstractDataFrame on RHS strange, or df[:,:] = df2 overwriting df non intuitive, we do not match columns by name when RHS is a DataFrame etc.). However, the problem is that we will get some breaking changes that will be hard to track with deprecation warnings (as the code would be rewritten). I do not know if this would be acceptable (we could e.g. write a info message every time DataFrames is loaded stating that there are new setindex!/getindex rules). Would this be OK?

@bkamins I'd say that breaking corner cases is fine if that makes the behavior and implementation simpler. Though for special cases which are not completely corner cases, it would be nice to have a deprecation first. It depends on each particular case I guess. Anyway it would be really useful if you could list cases which appear to break the rules.

> I actually like the df[:col] = nothing thing for deleting columns. It's already been established that nothing has some different behaviors than other types (e.g. it can't be use as a Dict key). Aside from this, I don't even know if there is an easier way of dropping columns now than df[filter(x -> x != col, names(df))].
> 
> Actually this seems to not work anyway, and we have delete! (we could probably use better docs of the miscellaneous functions).

@ExpandingMan `nothing` can absolutely be used as a dict key, it just cannot be found using the default `findfirst` method. For DataFrames, we just need to document `delete!`. It's much easier to discover than the `nothing` assignment trick (how did you find out about this?!).

ExpandingMan:
Funny story, today I started having to do some much more obnoxious data mangling than I've had to do in a while and I thought "Wait, I don't actually know a good way of deleting a column" and then I remembered this PR :laughing: .  Anyway, it's not important, `delete!` is fine.

bkamins:
Initial writeup of a target behavior (I will update it when I experiment with this a bit):

`getindex` and `setindex!` LHS:

row idnexing:
* '<:Integer'
* collection of `<:Integer` other than `Bool`; collection eltype is not important, only values
* collection of `Bool` of length exactly equal to number of rows in data frame
* `:`
* no index: will return original columns (no copying)

column indexing:
* `<:Integer` or `Symbol` (in this case `Vector` is returned, otherwise `DataFrame`)
* collection of `<:Integer` other than `Bool`; collection eltype is not important, only values
* collection of `Symbol` collection eltype is not important, only values
* collection of `Bool` of length exactly equal to number of rows in data frame
* `:`

`setindex!` RHS:
* if it is a `DataFrame`
    * assign to LHS what is in RHS in respective row and column indices (in particular: if symbols perform name matching)
* if it is a `Matrix`
    * shape of RHS and LHS regions must match exactly
    * broadcasting should do in place change of existing columns if whole columns are selected
    * do we want to broadcast over multiple rows/columns like in matrix assignment?
* if it is a `Vector`
    * number of rows of RHS and LHS regions must match exatly
    * broadcasting does not expand over multiple columns
* if it is a scalar
    * only allowed when single colmn and row is selected on LHS
    * when broadcasting allowed to broadcast to multiple rows or columns

nalimilan:
>     * do we want to broadcast over multiple rows/columns like in matrix assignment?

Maybe at some point, but I'd rather start with the basics, it's easy to add more fancy behaviors later if they are errors now.

bkamins:
After #1866 this is deprecated.

Issue #1327, opened 2017-12-23
issorted(df, :col) does not work


nalimilan:
`issorted` does not work when passed a single column, either an index or as a one-element tuple/array:
```julia
julia> df = DataFrame(x = [3, 1, 2, 1], y = ["b", "c", "a", "b"])
4×2 DataFrames.DataFrame
│ Row │ x │ y │
├─────┼───┼───┤
│ 1   │ 3 │ b │
│ 2   │ 1 │ c │
│ 3   │ 2 │ a │
│ 4   │ 1 │ b │

julia> issorted(df, :x)
ERROR: MethodError: no method matching lt(::Base.Order.Perm{Base.Order.ForwardOrdering,Array{Int64,1}}, ::DataFrames.DataFrameRow{DataFrames.DataFrame}, ::DataFrames.DataFrameRow{DataFrames.DataFrame})
Closest candidates are:
  lt(::Base.Order.ForwardOrdering, ::Any, ::Any) at ordering.jl:45
  lt(::Base.Order.ReverseOrdering, ::Any, ::Any) at ordering.jl:46
  lt(::Base.Order.By, ::Any, ::Any) at ordering.jl:47
  ...
Stacktrace:
 [1] issorted(::DataFrames.DFRowIterator{DataFrames.DataFrame}, ::Base.Order.Perm{Base.Order.ForwardOrdering,Array{Int64,1}}) at ./sort.jl:54
 [2] #issorted#185(::Array{Any,1}, ::Function, ::Function, ::Bool, ::Base.Order.ForwardOrdering, ::Function, ::DataFrames.DataFrame, ::Symbol) at /home/milan/.julia/DataFrames/src/abstractdataframe/sort.jl:275
 [3] issorted(::DataFrames.DataFrame, ::Symbol) at /home/milan/.julia/DataFrames/src/abstractdataframe/sort.jl:270
 [4] macro expansion at ./REPL.jl:97 [inlined]
 [5] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
```

This appears to be due to the fact that `lt` only accepts a `Perm` as first argument when the two following arguments are integers. That's not a problem for `sort` and `sortperm`, since these operate on row indices. But `issorted` operates on `DataFrameRow` objects, for which no method exists. When several columns are passed, a custom `DFPerm` object is passed as the first argument to `lt`, so a custom function is used.

I'm not sure what's the appropriate solution to fix this.

[...]

PR #1391, opened 2018-04-04
fix issorted bug

bkamins:
Resolves both problems from #1327.
In general `lt` function was broken when subsetting of `DataFrame` for `issorted` was performed.

@nalimilan CI will fail because of `CategoricalArray` bug. Need to rebase this PR after that one is merged, but currently it should be good for a review.

### DIFF
--- a/src/abstractdataframe/sort.jl
+++ b/src/abstractdataframe/sort.jl
@@ -73,14 +73,11 @@ DFPerm(o::O, df::DF) where {O<:Ordering, DF<:AbstractDataFrame} = DFPerm{O,DF}(o
 col_ordering(o::DFPerm{O}, i::Int) where {O<:Ordering} = o.ord
 col_ordering(o::DFPerm{V}, i::Int) where {V<:AbstractVector} = o.ord[i]
 
-Base.@propagate_inbounds Base.getindex(o::DFPerm, i::Int, j::Union{Int,Symbol}) = o.df[i, j]
-Base.@propagate_inbounds Base.getindex(o::DFPerm, a::DataFrameRow, j::Union{Int,Symbol}) = a[j]
-
 function Sort.lt(o::DFPerm, a, b)
-    @inbounds for (i,n) = enumerate(names(o.df))
+    @inbounds for i in 1:ncol(o.df)
         ord = col_ordering(o, i)
-        va = o[a, n]
-        vb = o[b, n]
+        va = o.df[a, i]
+        vb = o.df[b, i]
         lt(ord, va, vb) && return true
         lt(ord, vb, va) && return false
     end
@@ -272,9 +269,12 @@ function Base.issorted(df::AbstractDataFrame, cols_new=[]; cols=[],
                      :issorted)
         cols_new = cols
     end
-    cols_new isa Union{Real, Symbol} && return issorted(df[cols_new], lt=lt, by=by, rev=rev, order=order)
-    length(cols_new) == 1 && return issorted(df[cols_new[1]], lt=lt, by=by, rev=rev, order=order)
-    issorted(eachrow(df), ordering(df, cols_new, lt, by, rev, order))
+    if cols_new isa ColumnIndex
+        return issorted(df[cols_new], lt=lt, by=by, rev=rev, order=order)
+    elseif length(cols_new) == 1
+        return issorted(df[cols_new[1]], lt=lt, by=by, rev=rev, order=order)
+    end
+    issorted(1:nrow(df), ordering(df, cols_new, lt, by, rev, order))
 end
 
 # sort and sortperm functions
--- a/src/abstractdataframe/sort.jl
+++ b/src/abstractdataframe/sort.jl
@@ -73,14 +73,14 @@ DFPerm(o::O, df::DF) where {O<:Ordering, DF<:AbstractDataFrame} = DFPerm{O,DF}(o
 col_ordering(o::DFPerm{O}, i::Int) where {O<:Ordering} = o.ord
 col_ordering(o::DFPerm{V}, i::Int) where {V<:AbstractVector} = o.ord[i]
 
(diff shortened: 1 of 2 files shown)


bkamins:
CI passes. The required design is a bit slower than the old one but cleaning it up would require a general redesign of sorting functionality (which I would mark as a to-do for future anyway as there were comments in other places that this needs improvement).

src/abstractdataframe/sort.jl
```diff
@@ -73,14 +73,14 @@ DFPerm(o::O, df::DF) where {O<:Ordering, DF<:AbstractDataFrame} = DFPerm{O,DF}(o
 col_ordering(o::DFPerm{O}, i::Int) where {O<:Ordering} = o.ord
 col_ordering(o::DFPerm{V}, i::Int) where {V<:AbstractVector} = o.ord[i]
 
-Base.@propagate_inbounds Base.getindex(o::DFPerm, i::Int, j::Int) = o.df[i, j]
-Base.@propagate_inbounds Base.getindex(o::DFPerm, a::DataFrameRow, j::Int) = a[j]
+Base.@propagate_inbounds Base.getindex(o::DFPerm, i::Int, j::Union{Int,Symbol}) = o.df[i, j]
```
nalimilan: Maybe use `ColIndex` here? Same below.

src/abstractdataframe/sort.jl
```diff
@@ -272,6 +272,8 @@ function Base.issorted(df::AbstractDataFrame, cols_new=[]; cols=[],
                      :issorted)
         cols_new = cols
     end
+    cols_new isa Union{Real, Symbol} && return issorted(df[cols_new], lt=lt, by=by, rev=rev, order=order)
```
nalimilan: Given that there are three branches here, I'd find it clearer to use `if... elseif... else`.

src/abstractdataframe/sort.jl
```diff
@@ -73,14 +73,14 @@ DFPerm(o::O, df::DF) where {O<:Ordering, DF<:AbstractDataFrame} = DFPerm{O,DF}(o
...
-Base.@propagate_inbounds Base.getindex(o::DFPerm, i::Int, j::Int) = o.df[i, j]
-Base.@propagate_inbounds Base.getindex(o::DFPerm, a::DataFrameRow, j::Int) = a[j]
+Base.@propagate_inbounds Base.getindex(o::DFPerm, i::Int, j::Union{Int,Symbol}) = o.df[i, j]
+Base.@propagate_inbounds Base.getindex(o::DFPerm, a::DataFrameRow, j::Union{Int,Symbol}) = a[j]
 
 function Sort.lt(o::DFPerm, a, b)
-    @inbounds for i = 1:ncol(o.df)
+    @inbounds for (i,n) = enumerate(names(o.df))
         ord = col_ordering(o, i)
-        va = o[a, i]
```
nalimilan: Just curious: why doesn't `i` work here?

src/abstractdataframe/sort.jl
bkamins: fixed

src/abstractdataframe/sort.jl
bkamins: fixed

src/abstractdataframe/sort.jl
bkamins: This is why I say that it is messy.

The reason is line:
```
issorted(eachrow(df), ordering(df, cols_new, lt, by, rev, order))
```
which calls `ordering` which calls `DFPerm` like
```
DFPerm(ords, df[newcols])
```

So we see that in `lt` value of `o.df` is a data frame with `newcols` but `a` and `b` are data frames with all columns from `df` because they are produced by `eachrow(df)`.

However, given your comment I have decided to start cleaning it up a bit and simplified the code (now it does not use `eachrow`).

bkamins:
Additionally rebased.
As a side effect current implementation avoiding `eachrow` is up to ~2x faster on large `DataFrames`.

src/abstractdataframe/sort.jl
```diff
@@ -272,7 +269,12 @@ function Base.issorted(df::AbstractDataFrame, cols_new=[]; cols=[],
                      :issorted)
         cols_new = cols
     end
-    issorted(eachrow(df), ordering(df, cols_new, lt, by, rev, order))
+    if cols_new isa ColumnIndex
+        return issorted(df[cols_new], lt=lt, by=by, rev=rev, order=order)
+    elseif length(cols_new) == 1
+        return issorted(df[cols_new[1]], lt=lt, by=by, rev=rev, order=order)
+    end
+    issorted(1:nrow(df), ordering(df, cols_new, lt, by, rev, order))
```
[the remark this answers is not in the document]
bkamins: fixed. now I think I get what you meant - that we can remove `return` statements with such a construct.

test/sort.jl
```diff
@@ -42,6 +42,17 @@ module TestSort
 
     @test df == ds
 
+    df = DataFrame(x = [3, 1, 2, 1], y = ["b", "c", "a", "b"])
+    @test issorted(df, :x) == false
```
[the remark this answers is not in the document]
bkamins: fixed

test/sort.jl
```diff
@@ -52,4 +63,47 @@ module TestSort
     sort!(x, :y)
     @test x[:y] == [1,2,3,4]
     @test x[:x] == [1,3,2,4]
+
+    @test_throws ArgumentError sort(x, by=:x)
+
+    srand(1)
+    df_rand = DataFrame(rand(100, 4))
+    for n1 in names(df_rand)
+        @test sort(df_rand, n1)[n1] == sort(df_rand[n1])
```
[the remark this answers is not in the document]
bkamins: nice idea. fixed.

bkamins:
@nalimilan good to merge?

Issue #1388, opened 2018-04-04
Inconsistency of handling DataFrame creation and assignment


bkamins:
When `DataFrame` is created it converts all arguments to `Vector`. However, assignment allows other structures (like `Range`) as columns in `DataFrame`. I knew about it fore some time, but it constantly bothers me so here is an issue 😄 .
Is it an intended feature or should we fix it?

Here is an example of the difference:
```
julia> using DataFrames

julia> DataFrame(x=1:3)
3×1 DataFrames.DataFrame
│ Row │ x │
├─────┼───┤
│ 1   │ 1 │
│ 2   │ 2 │
│ 3   │ 3 │

julia> df1 = DataFrame(x=1:3)
3×1 DataFrames.DataFrame
│ Row │ x │
├─────┼───┤
│ 1   │ 1 │
│ 2   │ 2 │
│ 3   │ 3 │

julia> df2 = DataFrame()
0×0 DataFrames.DataFrame


julia> df2[:x] = 1:3
1:3

julia> df1[:x]
3-element Array{Int64,1}:
 1
 2
 3

julia> df2[:x]
1:3
```

nalimilan:
Good catch. I'm not sure what's the best choice. I'm slightly inclined to avoid converting ranges to vectors, given that it's easy to call `collect`, but that there's no way around if we make the conversion automatically.

nalimilan:
Yes, it was introduced on purpose when we moved away from `DataArray` by default. Good point about the breakage when removing or adding rows, I hadn't thought about that. So maybe better keep that behavior and change assignment to do the same. After all the gain in memory use should be quite limited compared with the size of the full set of columns.

bkamins:
OK - will make a PR and update docs accordingly.

bkamins:
PR in #1392

bkamins:
fixed by #1392

PR #1392, opened 2018-04-05
Always collect AbstractRange

bkamins:
Following #1388 this PR always collects `AbstractRange` before assignment to a `DataFrame`. In this way we have consistency between constructor and `setindex!`.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -40,14 +40,17 @@ Each column in `columns` should be the same length.
 **Notes**
 
 A `DataFrame` is a lightweight object. As long as columns are not
-manipulated, creation of a DataFrame from existing AbstractVectors is
+manipulated, creation of a `DataFrame` from existing AbstractVectors is
 inexpensive. For example, indexing on columns is inexpensive, but
 indexing by rows is expensive because copies are made of each column.
-An exception is assignment of `AbstractRange` as a column of
-`DataFrame` in which case it is collected to a vector.
 
-Because column types can vary, a DataFrame is not type stable. For
-performance-critical code, do not index into a DataFrame inside of
+If a column is passed to a `DataFrame` constructor or is assigned as a whole
+using `setindex!` then its reference is stored in the `DataFrame`. An exception
+to this rule is assignment of `AbstractRange` as a column in which case it is
+collected to a vector.
+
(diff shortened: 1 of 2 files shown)


bkamins:
@nalimilan any opinion on this?

src/dataframe/dataframe.jl
```diff
@@ -43,6 +43,8 @@ A `DataFrame` is a lightweight object. As long as columns are not
 manipulated, creation of a DataFrame from existing AbstractVectors is
 inexpensive. For example, indexing on columns is inexpensive, but
 indexing by rows is expensive because copies are made of each column.
+An exception is assignment of `AbstractRange` as a column of
```
nalimilan: I don't think "exception" is correct, as the previous sentence does not refer to columns passed to the constructor. Better add a separate paragraph, which would also document that column vectors are not copied in general.

I guess `setindex!` should be changed too, and the docstring be adapted?

bkamins: Second try to clarify things.

`setindex!` was changed already (but no `setindex!` methods needed changing directly). It does not have its docstring so I have left it as is for now (hopefully in the near future I will have time to go back to #1380 as it is a major redesign and then I will write an updated docstring).

nalimilan: Has it? I see this:
```julia
julia> x = DataFrame(x=1:2)
2×1 DataFrames.DataFrame
│ Row │ x │
├─────┼───┤
│ 1   │ 1 │
│ 2   │ 2 │

julia> x[:x] = 2:3
2:3

julia> x[:x]
2:3

```

bkamins: Try testing with this PR checked out. You will see:
```
julia> x[:x] = 2:3
2:3

julia> x[:x]
2-element Array{Int64,1}:
 2
 3
```

nalimilan: OK, that's what I realized after posting. But then it should be tested, right?

bkamins: This is exactly what lines 624 and 625 test.

nalimilan: Damn, let's say it was too late in the evening... :-/

bkamins:
@nalimilan thanks!

PR #1395, opened 2018-04-11
Add `permutecols!(df, p)` to allow column reordering

### LINKED ISSUE #1394: Add `permute!(df, p)` to allow column reordering
There are plenty of cases (especially after a `join`) where the columns of a `DataFrame` end up in an undesirable order. In these cases, I've found a simple `permute!` function (e.g., the one I included in [KeyedFrames.jl](https://github.com/invenia/KeyedFrames.jl)) to be quite helpful.

I'll issue a PR to add this function shortly. Comments would be most welcome, of course.

spurll:
Closes #1394


ararslan:
I like this in general, my only problem is the name `permute`. In this case it's permuting columns, but I would naively have assumed that the function permutes rows. Perhaps we could call this `permutecols!`?

spurll:
Makes sense to me.

nalimilan:
We need to make a systematic review of the API and decide whether functions should apply to rows or to columns by default. I think JuliaDB chose rows, but R data frames are collections of columns and the current API inherits from that approach in many cases (but not in all, e.g. `sort`).

spurll:
`length` has certainly confounded me in the past.

nalimilan:
See https://github.com/JuliaData/DataFrames.jl/issues/1200.

src/dataframe/reshape.jl
```diff
@@ -0,0 +1,54 @@
...
+├─────┼───┼───┼───┤
+│ 1   │ 3 │ 1 │ 2 │
+│ 2   │ 4 │ 2 │ 3 │
+│ 3   │ 5 │ 3 │ 4 │
+│ 4   │ 6 │ 4 │ 5 │
+│ 5   │ 7 │ 5 │ 6 │
+```
+"""
+function permutecols!(df::DataFrame, p::AbstractVector)
+    length(p) == size(df, 2) && isperm(p) || error("$p is not a valid column permutation")
```
ararslan: Perhaps use `throw(ArgumentError())` instead of `error()`?

test/dataframe.jl
```diff
@@ -662,4 +662,104 @@ module TestDataFrame
...
+            expected = DataFrame(; c=c, b=b, a=a)
+            permutecols!(df, [:c, :b, :a])
+            @test df == expected
+            df = deepcopy(original)
+            permutecols!(df, [3, 2, 1])
+            @test df == expected
+
+            # Invalid
+            df = DataFrame(; a=a, b=b, c=c)
+            @test_throws Exception permutecols!(df, [:a, :b])
```
ararslan: Should be `@test_throws ArgumentError ...` once the exception type is changed above.

src/dataframe/reshape.jl
```diff
@@ -0,0 +1,54 @@
+"""
+    permutecols!(df::DataFrame, p::AbstractVector)
+
+Permute the columns of `df` in-place, according to permutation `p`.
+
+### Examples
+
+```julia
+julia> df = DataFrame(; a=1:5, b=2:6, c=3:7)
```
nalimilan: `;` isn't needed. Same below.

src/DataFrames.jl
```diff
@@ -92,6 +93,7 @@ include("dataframerow/utils.jl")
 include("abstractdataframe/iteration.jl")
 include("abstractdataframe/join.jl")
 include("abstractdataframe/reshape.jl")
+include("dataframe/reshape.jl")
```
nalimilan: Why put this in a file called `reshape`? Also, does it actually need a separate file?

src/dataframe/reshape.jl
```diff
@@ -0,0 +1,54 @@
+"""
+    permutecols!(df::DataFrame, p::AbstractVector)
+
+Permute the columns of `df` in-place, according to permutation `p`.
```
nalimilan: Should mention what kind of permutations are accepted.

src/dataframe/reshape.jl
```diff
@@ -0,0 +1,54 @@
...
+function permutecols!(df::DataFrame, p::AbstractVector)
+    length(p) == size(df, 2) && isperm(p) || error("$p is not a valid column permutation")
+    permute!(DataFrames.columns(df), p)
+    df.colindex = DataFrames.Index(
+        Dict(names(df)[j] => i for (i, j) in enumerate(p)), [names(df)[j] for j in p]
+    )
+end
+
+function permutecols!(df::DataFrame, p::AbstractVector{Symbol})
+    permutecols!(df, indexin(p,names(df)))
```
nalimilan: I think there's a more efficient way to compute column indices, by indexing into the data frame index. Also, space after comma.

src/dataframe/reshape.jl
```diff
@@ -0,0 +1,54 @@
...
+│ 2   │ 4 │ 2 │ 3 │
+│ 3   │ 5 │ 3 │ 4 │
+│ 4   │ 6 │ 4 │ 5 │
+│ 5   │ 7 │ 5 │ 6 │
+```
+"""
+function permutecols!(df::DataFrame, p::AbstractVector)
+    length(p) == size(df, 2) && isperm(p) || error("$p is not a valid column permutation")
+    permute!(DataFrames.columns(df), p)
+    df.colindex = DataFrames.Index(
```
nalimilan: The `DataFrames` prefix isn't needed here nor above. To construct the `Index`, can't you just call `Index(permute!(names(df), p))`?

test/dataframe.jl
```diff
@@ -662,4 +662,104 @@ module TestDataFrame
...
+    @testset "permutecols!" begin
+        column_sets = [
+            (1:5, 2:6, 3:7),
+            (1.0:5.0, 2.0:6.0, 3.0:7.0),
+            ('a':'e', 'b':'f', 'c':'g'),
+            (["foo", "bar", "baz"], ["this", "that", "the other"], ["who", "what", "when"]),
+            ([:who, :what, :when], 'a':'c', ["foo", "bar", "baz"]),
+        ]
+
+        for (a, b, c) in column_sets
```
nalimilan: Is it really useful to test all these combinations of columns? The exercized code paths are the same. Also you can directly test the three columns case, the one- and two-column cases aren't that special AFAICT.

spurll: I have a habit of overdoing it a bit on the tests. I'll scale these back.

spurll:
Thanks for the feedback @nalimilan. I'll incorporate your suggestions.

test/dataframe.jl
ararslan: FWIW I don't particularly mind having a lot of tests even if they're somewhat redundant, unless they contribute significantly to the overall time to run the test suite.

nalimilan: I'm not concerned about the runtime, more about the unnecessary complexity/verbosity. I'm all for testing all possible situations where it makes sense, but here there's no reason to test multiple number of columns AFAICT.

ararslan: That's fair.

src/dataframe/reshape.jl
spurll: Yep. It needs a broadcast because the indices are stored in a `Dict`, but benchmarking still shows that it's much better than `indexin`.

src/dataframe/reshape.jl
spurll: Absolutely! I missed the single argument constructor. In fact we can just index into `names(df)` with `p` and skip the `permute!` here.

spurll:
Okay, I believe I addressed all of the points @nalimilan brought up. (And since a merge went in over the weekend I also rebased onto master.)

src/dataframe/dataframe.jl
```diff
@@ -1029,3 +1029,64 @@ function Base.push!(df::DataFrame, iterable::Any)
...
+├─────┼───┼───┼───┤
+│ 1   │ 3 │ 1 │ 2 │
+│ 2   │ 4 │ 2 │ 3 │
+│ 3   │ 5 │ 3 │ 4 │
+│ 4   │ 6 │ 4 │ 5 │
+│ 5   │ 7 │ 5 │ 6 │
+```
+"""
+function permutecols!(df::DataFrame, p::AbstractVector)
+    length(p) == size(df, 2) && isperm(p) || error("$p is not a valid column permutation")
```
ararslan: It would be great if you could make this `throw(ArgumentError("..."))` instead of `error`.

test/dataframe.jl
```diff
@@ -662,4 +662,66 @@ module TestDataFrame
...
+        expected = DataFrame(c=c, b=b, a=a)
+        permutecols!(df, [:c, :b, :a])
+        @test df == expected
+        df = deepcopy(original)
+        permutecols!(df, [3, 2, 1])
+        @test df == expected
+
+        # Invalid
+        df = DataFrame(a=a, b=b, c=c)
+        @test_throws Exception permutecols!(df, [:a, :b])
```
ararslan: Once `error` is changed, this should be `ArgumentError` instead of `Exception`.

spurll: Sure thing.

Not sure why the tests are failing. They're passing locally. I'll do a `Pkg.update()` and see if something's changed in one of the requirements that might be affecting things. The error appears to be happening before Travis even gets to the `permutecols!` tests. (Line 376 of test/dataframe.jl, looks like some stack/unstack tests.)

spurll: After updating everything, I'm now able to replicate the error locally. I'm still not sure what's causing it, however. It's appearing as part of an `isequals` check earlier in the tests. I checked out master and it appears to be an issue on master as well, so I don't think that it's related to my changes.

spurll:
CI failure appears to be due to [this change to CategoricalArrays.jl](https://github.com/JuliaData/CategoricalArrays.jl/pull/139), now tagged as v0.3.8, which seems to (understandably) be resulting in changes to column ordering after `unstack`. I'll open an issue.

ararslan:
@nalimilan What do you think, good to go?

PR #1398, opened 2018-04-18
Fix push!(df, a, b, ...)

cstjean:
[This method](https://github.com/JuliaLang/julia/blob/d386e40c17d43b79fc89d3e579fc04547241787c/base/abstractarray.jl#L1940) assumes that `push!(df, x)` returns `df`.


nalimilan:
Thanks. Test failures are unrelated and will be fixed by https://github.com/JuliaLang/METADATA.jl/pull/14324.

Does the new test you added cover both changes?

cstjean:
Thank you for the cleanup.

PR #1399, opened 2018-04-18
Fix error with undefined values in HTML printing

### LINKED ISSUE #1397: Error when preallocating dataframe with strings
```
DataFrame([String],[:F],2)
```
works on `0.10.1` but breaks on `0.11.5`

nalimilan:
Also change try... catch blocks to isassigned calls in show.jl, which is more correct and can avoid hiding bugs.

Fixes #1397.


ararslan:
Huh, the fact that Travis is failing on all versions and platforms but Windows is passing is odd.

nalimilan:
AFAIK that was because the push build wasn't restarted by closing/reopening. It's all green now.

PR #1405, opened 2018-04-27
Fix tests on Julia 0.7

nalimilan:
The remaining failure is fixed with DataStreams master.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -606,7 +606,7 @@ julia> df
 ```
 """
 Base.filter!(f, df::AbstractDataFrame) =
-    deleterows!(df, findall(!f, eachrow(df)))
+    deleterows!(df, findall(collect(!f(r)::Bool for r in eachrow(df))))
 
 function Base.convert(::Type{Array}, df::AbstractDataFrame)
     convert(Matrix, df)
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -2,10 +2,13 @@ import Base: isidentifier, is_id_start_char, is_id_char
 
 const RESERVED_WORDS = Set(["begin", "while", "if", "for", "try",
     "return", "break", "continue", "function", "macro", "quote", "let",
-    "local", "global", "const", "abstract", "typealias", "type", "bitstype",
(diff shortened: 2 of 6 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -606,7 +606,7 @@ julia> df
 ```
 """
 Base.filter!(f, df::AbstractDataFrame) =
-    deleterows!(df, findall(!f, eachrow(df)))
+    deleterows!(df, findall(collect(!f(r)::Bool for r in eachrow(df))))
```
ararslan: Why is this change necessary? Because `eachrow` isn't indexable and `findall` requires an indexable collection?

src/other/utils.jl
```diff
@@ -2,10 +2,13 @@ import Base: isidentifier, is_id_start_char, is_id_char
 
 const RESERVED_WORDS = Set(["begin", "while", "if", "for", "try",
     "return", "break", "continue", "function", "macro", "quote", "let",
-    "local", "global", "const", "abstract", "typealias", "type", "bitstype",
+    "local", "global", "const", "type",
     "immutable", "do", "module", "baremodule", "using", "import", "struct",
     "export", "importall", "end", "else", "elseif", "catch", "finally"])
 
+if VERSION < v"0.7.0-DEV.3220"
```
ararslan: Didn't this happen in 0.6-dev?

src/abstractdataframe/abstractdataframe.jl
nalimilan: Yes. That's kind of annoying, IIRC in the PR it was mentioned that a wrapper could be introduced in Base to use linear indices in such situations. But here it's probably faster to allocate a vector anyway (https://github.com/JuliaLang/julia/pull/25879), and we do that in `deleterows`.

src/other/utils.jl
nalimilan: No, or the tests would fail on 0.6. This commit appears to be the one which removed these deprecated identifiers.

bjarthur:
tests are failing again on 0.7-beta.  perhaps 0.7 should be added to travis.yml?

nalimilan:
Sure.

PR #1406, opened 2018-04-28
Override getproperty, setproperty! and propertynames for AbstractDataFrame and DataFrameRow

### LINKED ISSUE #1336: Add column lookup with Base.getproperty?
Once this package requires 0.7 at a minimum, we can define column lookup in terms of `getproperty` in addition to `getindex`, so that one can access a column `x` of a `DataFrame` `df` as `df.x` in addition to `df[:x]`.

We could consider a full replacement, but continuing to allow `df[:x]` permits fetching multiple columns at a time via `df[:, [:x, :y, :z]]`. I'm not sure how that would translate to the `getproperty` syntax.

nalimilan:
Provides a nice shorthand syntax for simple getindex and setindex! uses, with autocompletion,
and makes `DataFrameRow` more like `NamedTuple`.
This requires replacing all direct field accesses with calls to accessor functions.
Do not mention this feature in the manual yet since it is not available on a released Julia
version. Most uses of `df[:col]` will better be changed to `df.col` when Julia 0.7 is released.

(CI failure on 0.7 is fixed with DataStreams master.)

Fixes #1336.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -236,6 +236,13 @@ Compat.axes(df, i) = axes(df)[i]
 
 Base.ndims(::AbstractDataFrame) = 2
 
+if VERSION >= v"0.7.0-DEV.3067"
+    Base.getproperty(df::AbstractDataFrame, col_ind::Symbol) = getindex(df, col_ind)
+    Base.setproperty!(df::AbstractDataFrame, col_ind::Symbol, x) = setindex!(df, x, col_ind)
+    # Private fields are never exposed since they can conflict with column names
+    Base.propertynames(df::AbstractDataFrame, private::Bool=false) = names(df)
+end
+
 ##############################################################################
 ##
 ## Similar
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -218,13 +218,11 @@ struct DataFrameStream{T}
     columns::T
     header::Vector{String}
 end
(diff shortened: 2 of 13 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -236,6 +236,13 @@ Compat.axes(df, i) = axes(df)[i]
 
 Base.ndims(::AbstractDataFrame) = 2
 
+if VERSION >= v"0.7.0-DEV.3067"
+    Base.getproperty(df::AbstractDataFrame, col_ind::Symbol) = getindex(df, col_ind)
+    Base.setproperty!(df::AbstractDataFrame, col_ind::Symbol, x) = setindex!(df, x, col_ind)
+    # Private fields are never exposed since they can conflict with column names
+    Base.propertynames(df::AbstractDataFrame, private::Bool=false) = names(df)
```
nalimilan: I'm not sure what's the best solution here. This situation is going to arise for other types where field names are dynamic. We could also throw an error when `private == true`.

ararslan:
Fixes #1336?

nalimilan:
Thanks, I had forgotten about this!

> How does the . notation work with non-standard column names. What I mean by this is a name like Symbol("Column 1"). It is accessible via the method call I guess - right? But can it be accessed via . somehow or it is not possible? I guess those things should be eventually mentioned in the docs.

I'm not sure, but indeed I think you'd need to call `getproperty`. These clearly need docs, but until 0.7/1.0 are released it would be misleading to use them in the manual.

PR #1409, opened 2018-05-12
Make `describe` return a DataFrame

pdeffebach:
This pull request is the result of [this](https://discourse.julialang.org/t/a-proposal-for-describe-of-a-dataframe/10254) discussion on Discourse about making `describe(df::AbstractDataFrame)` return a tabular object for easier reading. 

The `describe` function now works as follows:

- Allows for a keyword argument `colstats` so that the user can select which summary statistics are reported
- It reports `nothing` in the cell where the summary statistic is not defined. For example, the mean of a `String` column is simply `nothing`. 
- It reports the number of missing values, similarly reporting `nothing` if the column does not allow missing values. 

The `describe(x::AbstractVector)` functions are not modified by this pull request.  This is the behavior that simply prints the results of `StatsBase.summarize` (with some additions for non-numeric vectors that DataFrames had already defined.

I updated the tests and docstrings to this new behavior. 

I believe this PR will result in more friendly behavior for Data Exploration. It is much easier to read the printout of a DataFrame than the current behavior, where one has to scroll a good deal to see the numbers they want.


REQUIRE
```diff
@@ -1,11 +1,11 @@
...
-CategoricalArrays 0.3.6
-StatsBase 0.11.0
-SortingAlgorithms
-Reexport
-WeakRefStrings 0.4.0
-DataStreams 0.3.0
-CodecZlib 0.4
-TranscodingStreams
-Compat 0.59.0
+julia 0.6
```
nalimilan: I'm not sure what you changed here but since it only affects whitespace, please leave this file alone.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -374,28 +374,35 @@ end
 Summarize the columns of an AbstractDataFrame
 
 ```julia
-describe(df::AbstractDataFrame)
-describe(io, df::AbstractDataFrame)
+describe(df::AbstractDataFrame; colstats = [:mean, :min, :median, :max, :Nmissing, :datatype])
+describe(io, df::AbstractDataFrame; colstats = [:mean, :min, :median, :max, :Nmissing, :datatype])
```
nalimilan: I'd use `stats`, `:nmissing` and `:eltype`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -374,28 +374,35 @@ end
...
 ```
 
 **Arguments**
 
 * `df` : the AbstractDataFrame
 * `io` : optional output descriptor
+* `colstats` a vector of symbols representing the summaristatistics you wish to report
```
```
`colstats::AbstractVector{Symbol}`: the summary statistics to report

File: src/abstractdataframe/abstractdataframe.jl [hunk 4]
```diff
@@ -374,28 +374,35 @@ end
...
 
 **Result**
 
-* nothing
+* a DataFrame where each row represents a variable of your input DataFrame and each
```
```
A `DataFrame` where each row represents a variable and each column a summary statistic.
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -374,28 +374,35 @@ end
...
 
-If the column's base type derives from Number, compute the minimum, first
-quantile, median, mean, third quantile, and maximum. Missings are filtered and
-reported separately.
+If the column's base type derives from Number, compute the mean, standard 
+deviation, minimum, first quantile, median, third quantile, and maximum. If 
+a column is not numeric, these statistics are populated with `nothing`s. 
 
-For boolean columns, report trues, falses, and missings.
+For variables of *all* types, `describe` can also report the type of the
```
nalimilan: "`describe` can also report" -> "also report". No need to highlight "all".

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -374,28 +374,35 @@ end
...
+If the column's base type derives from Number, compute the mean, standard 
+deviation, minimum, first quantile, median, third quantile, and maximum. If 
+a column is not numeric, these statistics are populated with `nothing`s. 
 
-For boolean columns, report trues, falses, and missings.
+For variables of *all* types, `describe` can also report the type of the 
+variable and the number of unique values. 
 
-For other types, show column characteristics and number of missings.
+Missings are filtered in the calculation of all statistics, however the optional
```
nalimilan: "Missings" -> "Missing values".
Also the description of `Nmissing` is incorrect.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
-function StatsBase.describe(io, df::AbstractDataFrame)
-    for (name, col) in eachcol(df)
-        println(io, name)
-        describe(io, col)
-        println(io, )
+StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
+function StatsBase.describe(io, df::AbstractDataFrame; colstats = [:mean, :min, :median, :max, :Nmissing, :datatype])
+    # Check that people don't specify the wrong fields. 
+    allowed_fields = [:mean, :sd, :min, :q25, :median, :q75, :max, :datatype, :Nunique, :Nmissing]
+    for i in colstats
```
nalimilan: You can do this more simply with `issubset`. Then it would be nice to print the name of problematic fields using `setdiff(colstats, allowed_fields)`. Also no need to use a new line for each name.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
-function StatsBase.describe(io::IO, X::AbstractVector{Union{T, Missing}}) where T
-    missingcount = count(ismissing, X)
-    pmissing = 100 * missingcount/length(X)
-    if pmissing != 100 && T <: Real
-        show(io, StatsBase.summarystats(collect(skipmissing(X))))
-    else
-        println(io, "Summary Stats:")
+    # Define 4 functions for getting summary statistics 
+    # use a dict because we dont know which measures the user wants
+    function get_stats(col::AbstractArray{T} where T <: Real)
```
nalimilan: `AbstractArray{<:Real}` is enough.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+        stats = summarystats(col)
+        stats_dict = Dict(
+            :mean => stats.mean,
+            :sd => sqrt(var(col)),
```
nalimilan: Just use `std`. Since the mean is already known, also do `std(col, mean=stats.mean)`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+            :median => stats.median,
+            :q75 => stats.q75,
+            :max => stats.max,
+            :datatype=> eltype(col),
+            :Nmissing => nothing,
+            :Nunique => nothing
+        )
+    end
+    
+    function get_stats(col::AbstractArray{Union{T, Missing}} where T <: Real)
```
nalimilan: `AbstractArray{<:Union{Real, Missing}}`

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+        stats_dict = Dict(
+            :mean => stats.mean,
+            :sd => sqrt(var(collect(skipmissing(col)))),
+            :min => stats.min ,
+            :q25 => stats.q25,
+            :median => stats.median,
+            :q75 => stats.q75,
+            :max => stats.max,
+            :datatype=> Missings.T(eltype(col)),
+            :Nmissing => count(ismissing(col)),
```
nalimilan: `count(ismissing, col)`. Above, missing space before `=>`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+            :Nmissing => nothing,
+            :Nunique => nothing
+        )
+    end
+    
+    function get_stats(col::AbstractArray{Union{T, Missing}} where T <: Real)
+        stats = summarystats(collect(skipmissing(col)))
+        stats_dict = Dict(
+            :mean => stats.mean,
+            :sd => sqrt(var(collect(skipmissing(col)))),
```
nalimilan: No need to call `collect`. But since it's needed for `summarystats` above, better store the result and reuse it here too.

BTW, not related, but we should really fix `summarystats` so that `collect` isn't needed at all.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+            :Nunique => length(unique(col))
+        )    
+    end 
+    # Takes in a column and returns a row vector of the statistics
+    function spread(col)
+        d = get_stats(col) # a Dict
+        row = [d[f] for f in colstats]
+        reshape(row, 1, length(row))
+    end
+    # Append the summary stats into a matrix
```
nalimilan: Creating a matrix is wasteful since what we need in the end is a `DataFrame`. So better collect values directly into a `DataFrame`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+    data = Array{Any}(size(df,2),length(colstats))
+    i = 1
```
nalimilan: `i` isn't used.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+            :median => nothing,
+            :q75 => nothing,
+            :max => nothing,
+            :datatype=> eltype(col),
+            :Nmissing => nothing,
+            :Nunique => length(unique(col))
+        )   
+    end
+    
+    function get_stats(col:: AbstractArray{Union{T, Missing}} where T <: Any)
```
nalimilan: Just `col::AbstractArray`, since `Union{Any, T} == Any`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +412,109 @@ describe(df)
...
+            :max => stats.max,
+            :datatype=> eltype(col),
+            :Nmissing => nothing,
+            :Nunique => nothing
+        )
+    end
+    
+    function get_stats(col::AbstractArray{Union{T, Missing}} where T <: Real)
+        stats = summarystats(collect(skipmissing(col)))
+        stats_dict = Dict(
```
nalimilan: No need to give a name to this object since it's returned immediately.

test/dataframe.jl
```diff
@@ -325,25 +325,48 @@ module TestDataFrame
...
-        @test nothing == describe(f, Any["1", "2", missing])
-        @test nothing == describe(f, ["1", "2", missing])
-        @test nothing == describe(f, CategoricalArray(["1", "2", missing]))
-    end
+    
+    # Test that the describe output handles all values and missings properly
+    # construct the output dataframe
+    struct MySuperLongNameForAStructIsThisToolong end
+    @testset "describe(df)" begin
+        variable = Vector{Any}([:number, :number_missing, :non_number, :non_number_missing, :long_struct])
```
nalimilan: Just create the `DataFrame` directly, no need to store columns in temporary objects. Also I don't think we care about the type of the vector, it will be chosen automatically. Same below.

test/dataframe.jl
```diff
@@ -325,25 +325,48 @@ module TestDataFrame
...
+        df = DataFrame(number = vec_number, 
+            number_missing = vec_number_missing, 
+            non_number = vec_non_number, 
+            non_number_missing = vec_non_number_missing,
+            long_struct = long_struct)
+        
+        @test describe(df) == describe_output
+    
+        # Test that the keyword arguments works
+        describe_output_mean = describe_output[[:variable, :mean]]
```
nalimilan: No need to create temporary variable.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Thanks for all the feedback. I'm working on changing the code now. Do you have any idea what the best way to do this would be? 

```
df = DataFrame(rand(10,10))
df[1,:] = rand(10)' # normal vector doesn't work either

>MethodError: Cannot `convert` an object of type Array{Float64,1} to an object of type Float64
```

It looks like you can replace a row in a DataFrame with a number, which then propagates across the whole row, but you can't replace it with a row. The expected behavior was that of a `Matrix`, where you can do

```
x = rand(10,10)
x[1,:] = ones(10) # no transpose needed
```

REQUIRE
pdeffebach: Sorry, i thought I had fixed this. This is my first time doing a PR for anything, and my first time using Git, too.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Fixed all this, and changed `Nunique` to `nunique` as well.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Added. Thank you for helping me with this.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Added. Thank you.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Fixed. "Will also report..."

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Thank you. i had forgotten the broadcats `count(ismissing.(col))`

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Fixed. thank you for this.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Thank you, I have updated the type signatures.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: I store the result and re-use it now.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Fixed, changed the broadcast to `count(ismissing, col)`

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Could you please clarify this? How else can we dispatch on the presence of missing values? Since we want it to return `nothing` sometimes and the number of missing values other times.

pdeffebach:
PS, sorry about all the commits. I was stuck in package lock with DataFrames on my computer so I was using Travis CI to test. I re-downloaded Julia and can now compile my local module.

test/dataframe.jl
```diff
@@ -325,25 +325,31 @@ module TestDataFrame
...
-        @test nothing == describe(f, Any["1", "2", missing])
-        @test nothing == describe(f, ["1", "2", missing])
-        @test nothing == describe(f, CategoricalArray(["1", "2", missing]))
-    end
+    
+    # Test describe
+    struct MySuperLongNameForAStructIsThisToolong end
+    @testset "describe(df)" begin
+        describe_output = DataFrame(
+            variable = Vector{Any}([:number, :number_missing, :non_number, :non_number_missing, :long_struct]),
```
nalimilan: Still no need to call `Vector{Any}` here nor below.

test/dataframe.jl
```diff
@@ -325,25 +325,31 @@ module TestDataFrame
...
+            min = [1.0, 1.0, nothing, nothing, nothing],
+            median = [2.5, 2.0, nothing, nothing, nothing],
+            max = [4.0, 3.0, nothing, nothing, nothing],
+            nmissing = [nothing, 1, nothing, 1, nothing],
+            eltype= Vector{Any}([Int64, Int64, String, String, MySuperLongNameForAStructIsThisToolong])
+        )
+    
+        # Construct the test dataframe
+        df = DataFrame(number = [1, 2, 3, 4],
+            number_missing = [1,2, 3, missing],
```
nalimilan: Wrong alignment.

It would also be nice to test the column types, and to have a test mixing e.g. numbers and Dates: the min and max will have heterogeneous types.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -371,31 +371,36 @@ end
...
 **Arguments**
 
 * `df` : the AbstractDataFrame
 * `io` : optional output descriptor
+* `stats::AbstractVector{Symbol}`: the summary statistics to report
 
 **Result**
 
-* nothing
+* A `DataFrame` where each row represents a variable and each column a summary statistic.
```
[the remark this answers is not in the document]
pdeffebach: I think it was a list originally.

test/dataframe.jl
pdeffebach: This is the start of a larger conversation about edge-case types, I think. the `Date` type is not a subtype of number, so it acts like `Any`. 

Additionally, the behavior i currently have returns `nothing` for `nunique` on numeric types. On one hand, numeric variables are likely to have many values, making `nunique` an expensive operation. On the other hand, if you are using a small amount of integers, `nunique` makes sense. On the third hand, people might store `Point{Float64,Float64}` in a DataFrame, so `nunique` becomes even more expensive. 

Not to even mention factor variables. A numeric vector that is then coded as a categorical variable ceases to be a numeric. This might be annoying if you want to explore survey results coded 1-5, but then use them as categorical data in `GLM`. I'm sure there are tons more types that are in between `Numeric` and not, and I'm not entirely sure how to approach all of those cases.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +410,100 @@ describe(df)
...
+            :q75 => nothing,
+            :max => nothing,
+            :eltype => Missings.T(eltype(col)),
+            :nmissing => count(ismissing, col),
+            :nunique => length(unique(col))
+        )    
+    end 
+
+    # Put the summary stats into the return dataframe
+    data = DataFrame(Any, size(df,2), length(stats) + 1)
```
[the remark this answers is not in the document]
pdeffebach: re-did as described. let me know if its correct.

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +410,100 @@ describe(df)
...
+            :median => nothing,
+            :q75 => nothing,
+            :max => nothing,
+            :eltype => eltype(col),
+            :nmissing => nothing,
+            :nunique => length(unique(col))
+        )   
+    end
+    
+    function get_stats(col:: AbstractArray{Union{T, Missing}} where T <: Any)
```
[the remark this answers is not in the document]
pdeffebach: added.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +410,100 @@ describe(df)
...
-        show(io, StatsBase.summarystats(collect(skipmissing(X))))
-    else
-        println(io, "Summary Stats:")
+    
+    function get_stats(col::AbstractArray{<:Union{Real, Missing}})
+        nomissing = collect(skipmissing(col))
+        sumstats = summarystats(nomissing)
+        Dict(
+            :mean => sumstats.mean,
+            :std => std(nomissing),
```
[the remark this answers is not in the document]
pdeffebach: added.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +410,100 @@ describe(df)
...
+function StatsBase.describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, :eltype])
+     # Check that people don't specify the wrong fields. 
+    allowed_fields = [:mean, :std, :min, :q25, :median, :q75, :max, :eltype, :nunique,:nmissing] 
+    if !issubset(stats, allowed_fields) 
+        disallowed_fields = setdiff(stats, allowed_fields)
+       error("Not an allowed field. Allowed fields are:mean, :min, :q25, 
+        :median, :q75, :max, :eltype, :nmissing, :nunique. The field(s)
+        you entered which are not allowed are $disallowed_fields") 
+   end
+    # Define 4 functions for getting summary statistics
```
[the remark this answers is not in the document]
pdeffebach: fixed the indentation.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +410,100 @@ describe(df)
...
-        println(io, name)
-        describe(io, col)
-        println(io, )
+StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
+function StatsBase.describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, :eltype])
+     # Check that people don't specify the wrong fields. 
+    allowed_fields = [:mean, :std, :min, :q25, :median, :q75, :max, :eltype, :nunique,:nmissing] 
+    if !issubset(stats, allowed_fields) 
+        disallowed_fields = setdiff(stats, allowed_fields)
+       error("Not an allowed field. Allowed fields are:mean, :min, :q25,
```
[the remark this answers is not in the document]
pdeffebach: Added the `throw(ArgumentError(...))` but I kept the list of allowed fields for the user to see what they probably mis-typed (`sd` instead of `std`, for example).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +410,100 @@ describe(df)
...
 """
-StatsBase.describe(df::AbstractDataFrame) = describe(stdout, df)
-function StatsBase.describe(io, df::AbstractDataFrame)
-    for (name, col) in eachcol(df)
-        println(io, name)
-        describe(io, col)
-        println(io, )
+StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
+function StatsBase.describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, :eltype])
+     # Check that people don't specify the wrong fields.
```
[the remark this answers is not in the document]

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -371,31 +371,36 @@ end
...
 
-For boolean columns, report trues, falses, and missings.
+Will also report the type of the 
+variable and the number of unique values. 
 
-For other types, show column characteristics and number of missings.
+Missing values are filtered in the calculation of all statistics, however the column
+`nmissing` will report the number of missing values of that variable. 
+If the column does not allow missing values, `nothing` is returned. 
+Consequently, `nmissing = 0` (and not nothing) indicates that the column allows
```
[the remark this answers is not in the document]

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -371,31 +371,36 @@ end
...
-* nothing
+* A `DataFrame` where each row represents a variable and each column a summary statistic.
+
 
 **Details**
 
-If the column's base type derives from Number, compute the minimum, first
-quantile, median, mean, third quantile, and maximum. Missings are filtered and
-reported separately.
+If the column's base type derives from Number, compute the mean, standard
```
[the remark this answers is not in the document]
pdeffebach: Fixed.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -371,31 +371,36 @@ end
...
 **Details**
 
-If the column's base type derives from Number, compute the minimum, first
-quantile, median, mean, third quantile, and maximum. Missings are filtered and
-reported separately.
+If the column's base type derives from Number, compute the mean, standard 
+deviation, minimum, first quantile, median, third quantile, and maximum. If 
+a column is not numeric, these statistics are populated with `nothing`s. 
 
-For boolean columns, report trues, falses, and missings.
```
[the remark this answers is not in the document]
pdeffebach: Added, however this would change the behavior, as I mentioned in another comment, it doesn't currently get unique values for `Number` types, but maybe it should.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +412,92 @@ describe(df)
...
+        # for each statistic, loop through the columns array to find values
+        data[stat] = [columns[i][stat] for i in 1:size(df, 2)]
     end
-    println(io, "Length:         $(length(X))")
-    println(io, "Type:           $(eltype(X))")
-    !(T <: Real) && println(io, "Number Unique:  $(length(unique(X)))")
-    println(io, "Number Missing: $(missingcount)")
-    @printf(io, "%% Missing:      %.6f\n", pmissing)
-    return
+   return data
```
nalimilan: Incorrect indentation.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +412,92 @@ describe(df)
...
+
+    # Put the summary stats into the return dataframe
+    data = DataFrame()
+    data[:variable] = names(df)
+
+    # An array of Dicts for summary statistics
+    columns = [get_stats(col) for (name,col) in eachcol(df)] 
+    for stat in stats
+        # for each statistic, loop through the columns array to find values
+        data[stat] = [columns[i][stat] for i in 1:size(df, 2)]
```
nalimilan: Minor stylistic point: `[col[stat] for col in columns]` would be simpler. BTW, the comment could be moved just above `for`, and it could mention "letting the comprehension choose the appropriate type" (which is not immediately obvious).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +412,92 @@ describe(df)
...
+            :nunique => length(unique(col))
+        )   
+    end
+
+    # Put the summary stats into the return dataframe
+    data = DataFrame()
+    data[:variable] = names(df)
+
+    # An array of Dicts for summary statistics
+    columns = [get_stats(col) for (name,col) in eachcol(df)]
```
nalimilan: You can also do `for col in columns(df)`.

test/dataframe.jl
```diff
@@ -325,25 +325,28 @@ module TestDataFrame
...
-        @test nothing == describe(f, [1, 2, 3])
-        @test nothing == describe(f, [1, 2, 3])
-        @test nothing == describe(f, CategoricalArray([1, 2, 3]))
-        @test nothing == describe(f, Any["1", "2", missing])
-        @test nothing == describe(f, ["1", "2", missing])
-        @test nothing == describe(f, CategoricalArray(["1", "2", missing]))
-    end
+    
+    # Test describe
+    struct MySuperLongNameForAStructIsThisToolong end
```
nalimilan: Actually it's not very useful here to have a type with a long name, as the printing is done using the standard data frame printing system. Better use e.g. `Date`, and if you want to test printing with long types, add a test in show.jl which wouldn't use `describe` at all.

Also, please add a `CategoricalArray` column as that's a common type.

test/dataframe.jl
```diff
@@ -325,25 +325,28 @@ module TestDataFrame
...
+    @testset "describe(df)" begin
+        describe_output = DataFrame(variable = [:number, :number_missing, :non_number, :non_number_missing, :long_struct],
+                                    mean = [2.5, 2.0, nothing, nothing, nothing],
+                                    min = [1.0, 1.0, nothing, nothing, nothing],
+                                    median = [2.5, 2.0, nothing, nothing, nothing],
+                                    max = [4.0, 3.0, nothing, nothing, nothing],
+                                    nmissing = [nothing, 1, nothing, 1, nothing],
+                                    eltype= [Int64, Int64, String, String, MySuperLongNameForAStructIsThisToolong])
```
nalimilan: Add space before `=`.

test/dataframe.jl
nalimilan: Of course, I forgot this was limited to numbers. Yet there can be some "weird" number types, so it's not a bad idea to use a comprehension so that the element type is chosen automatically.

Good point about the fact that the number of unique values could become too high even for non-`Number` types (`Date` being a case in point). At the extreme, the computer could run out of memory, which would be really bad. I guess the best we can do would be to use a custom `nunique` function which would replicate `unique`, but stop after a large number of unique values has been found.

Regarding numeric columns with a small number of unique values, I'm not sure there's much we can do. People should probably use `CategoricalArray` if they want these variables to be treated as categorical.

As you suggest, it could make sense to use `try... catch` (and/or `applicable`). That's not hard, just verbose. Probably better leave that for another PR anyway.

test/dataframe.jl
nalimilan: Ah, and use `Int` rather than `Int64` (this explains the 32-bit Windows failure).

test/dataframe.jl
pdeffebach: I added a try-catch for `get_stats` when working with `Any` and `Union{Any, Missing}`. This way the user can define a custom type and all the expected overloadings of functions will work. I am not sure of the performance, however. I assume that `try catch` is O(1) since it doesn't take very long to see if there's an error. But I don't know how the compiler works.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +412,92 @@ describe(df)
...
-    for (name, col) in eachcol(df)
-        println(io, name)
-        describe(io, col)
-        println(io, )
+StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
+function StatsBase.describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, :eltype])
+     # Check that people don't specify the wrong fields. 
+    allowed_fields = [:mean, :std, :min, :q25, :median, :q75, :max, :eltype, :nunique, :nmissing] 
+    if !issubset(stats, allowed_fields) 
+                 disallowed_fields = setdiff(stats, allowed_fields)
```
nalimilan: Use four spaces for indentation.

nalimilan:
Let's go with the current list of `get_stats` methods for now, then we can improve this in another PR. As long as there's no regression compared to the current code, it's OK.

test/dataframe.jl
pdeffebach: Submitting this PR now, but will open up another one soon with `try..catch`. Should be very simple since it's just `get_stats` that's being changed.

pdeffebach:
Okay, I think its ready for another review. Everything should be fixed. I did a force push though, which I'm pretty sure was a dumb thing to do. I wouldn't mind a run-down of all the workflow-related things I did wrong in this PR, you can DM me on discourse to go over it.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -368,34 +368,42 @@ function Base.dump(io::IO, df::AbstractDataFrame, n::Int, indent)
...
+For non-numeric types, `describe` can also report the number of missing values 
+with `nunique` specified in the `stats` keyword argument. For numeric types, 
+`nunique` will return `nothing`s.
 
-For other types, show column characteristics and number of missings.
+Missing values are filtered in the calculation of all statistics, however the column
+`nmissing` will report the number of missing values of that variable. 
+If the column does not allow missing values, `nothing` is returned. 
+Consequently, `nmissing = 0` indicates that the column allows
+missing values, but does not contain any at the time.
```
nalimilan: I'm not a native English speaker, but "at the time" sounds weird. Maybe "currently"?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +413,94 @@ describe(df)
...
+            :median => nothing,
+            :q75 => nothing,
+            :max => nothing,
+            :nmissing => count(ismissing, col),
+            :nunique => length(unique(col)),
+            :eltype => Missings.T(eltype(col))
+        )    
+    end 
+    
+    function get_stats(col)
```
nalimilan: Drop this method, it will never be called due to the `::AbstractArray{>:Missing}` one above.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +413,94 @@ describe(df)
...
+            :eltype => eltype(col)
+        )   
+    end
+
+    # Put the summary stats into the return dataframe
+    data = DataFrame()
+    data[:variable] = names(df)
+
+    # An array of Dicts for summary statistics
+    columns = [get_stats(col) for col in df.columns]
```
nalimilan: Better use `columns(df)`, as on 0.7 `df.columns` doesn't work due to overriding field access. But then the variable `columns` needs to use another name.

test/dataframe.jl
```diff
@@ -325,25 +325,44 @@ module TestDataFrame
...
+    
+    # Test describe
+    struct CustomStruct end
+    @testset "describe(df)" begin
+        # Construct the test dataframe
+        df = DataFrame(number = [1, 2, 3, 4],
+                       number_missing = [1,2, 3, missing],
+                       non_number = ["a", "b", "c", "d"],
+                       non_number_missing = ["a", "b", "c", missing],
+                       custom_struct = [CustomStruct() for i in 1:4],
```
nalimilan: As I said I don't think it's useful to use `CustomStruct` here. Better test `CategoricalArray`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +413,94 @@ describe(df)
...
 
 """
-StatsBase.describe(df::AbstractDataFrame) = describe(stdout, df)
-function StatsBase.describe(io, df::AbstractDataFrame)
-    for (name, col) in eachcol(df)
-        println(io, name)
-        describe(io, col)
-        println(io, )
+StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
+function StatsBase.describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, :eltype])
```
nalimilan: `io::IO` and `stats::AbstractVector{Symbol}`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,29 +413,94 @@ describe(df)
...
-function StatsBase.describe(io::IO, X::AbstractVector{Union{T, Missing}}) where T
-    missingcount = count(ismissing, X)
-    pmissing = 100 * missingcount/length(X)
-    if pmissing != 100 && T <: Real
-        show(io, StatsBase.summarystats(collect(skipmissing(X))))
-    else
-        println(io, "Summary Stats:")
+    # Define 4 functions for getting summary statistics 
+    # use a dict because we dont know which measures the user wants
+    function get_stats(col::AbstractArray{<:Real})
```
nalimilan: Use a different name for the `col` argument, since it triggers an error in Julia 0.7 (apparently due to the conflict with the variable in the parent function).

nalimilan: Actually it's totally unrelated to the existence of a `col` variable in the function. Can you move the definitions of `get_stats` out of the function for now to work around the bug in Julia 0.7?

test/dataframe.jl
```diff
@@ -325,25 +325,44 @@ module TestDataFrame
...
+                       dates  = Date.([2000, 2001, 2003, 2004]))
+
+        describe_output = DataFrame(variable = [:number, :number_missing, :non_number, :non_number_missing, :custom_struct, :dates],
+                                    mean = [2.5, 2.0, nothing, nothing, nothing, nothing],
+                                    min = [1.0, 1.0, nothing, nothing, nothing, nothing],
+                                    median = [2.5, 2.0, nothing, nothing, nothing, nothing],
+                                    max = [4.0, 3.0, nothing, nothing, nothing, nothing],
+                                    nmissing = [nothing, 1, nothing, 1, nothing, nothing],
+                                    eltype= [Int64, Int64, String, String, CustomStruct, Date])
```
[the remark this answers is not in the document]
pdeffebach: Fixed. I thought i had fixed this earlier but am bad at git.

test/dataframe.jl
```diff
@@ -325,25 +325,44 @@ module TestDataFrame
...
+                                              nunique = [nothing, nothing, 4, 4, 1, 4],
+                                              nmissing = [nothing, 1, nothing, 1, nothing, nothing],
+                                              eltype= [Int64, Int64, String, String, CustomStruct, Date])
+
+        
+        # Test that it works as a whole, without keyword arguments
+        @test describe_output == describe(df)
+
+        # Test that it works with all keyword arguments
+        @test describe_output_all_stats == describe(df, stats = [:mean, :std, :min, :q25, :median, :q75, :max, :nunique, :nmissing, :eltype])
```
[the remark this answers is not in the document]
pdeffebach: Should I do this for every line beyond 92 characters? Could you point me to a guide for how to have these line breaks conform with good style?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -368,34 +368,42 @@ function Base.dump(io::IO, df::AbstractDataFrame, n::Int, indent)
     end
 end
 
-# summarize the columns of a df
-# TODO: clever layout in rows
+
 """
-Summarize the columns of an AbstractDataFrame
+Report descriptive statistics for a DataFrame
```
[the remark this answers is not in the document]
pdeffebach: Fixed.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -368,34 +368,42 @@ function Base.dump(io::IO, df::AbstractDataFrame, n::Int, indent)
...
+describe(df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, :datatype])
+describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, :datatype])
 ```
 
 **Arguments**
 
 * `df` : the AbstractDataFrame
 * `io` : optional output descriptor
+* `stats::AbstractVector{Symbol}`: the summary statistics to report. Allowed 
+  fields are `:mean, :std, :min, :q25, :median, :q75, :max, :eltype, :nunique,
```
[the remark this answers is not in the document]
pdeffebach: Fixed.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -368,34 +368,42 @@ function Base.dump(io::IO, df::AbstractDataFrame, n::Int, indent)
...
-If the column's base type derives from Number, compute the minimum, first
-quantile, median, mean, third quantile, and maximum. Missings are filtered and
-reported separately.
+If the column's base type derives from `Number`, compute the mean, standard 
+deviation, minimum, first quantile, median, third quantile, and maximum. If 
+a column is not numeric, these statistics are populated with `nothing`s. 
 
-For boolean columns, report trues, falses, and missings.
+For non-numeric types, `describe` can also report the number of missing values 
+with `nunique` specified in the `stats` keyword argument. For numeric types,
```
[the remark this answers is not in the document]
pdeffebach: Fixed this. 

I changed the sentence. 

When `:nunique` is contained `stats` vector, `describe` will report the 
number of unique values in a column. If a column's base type derives from `Number`,
`:nunique` will return `nothing`s.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: It is definitely bad english. Changed to currently.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Fixed.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Are you sure? I think the "any" method is getting called based on the tests. If it weren't being called, we would see `:nmissing` be `0` for the `non_number` column. But instead we get `nothing`.

test/dataframe.jl
pdeffebach: Changed to categorical Array

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
pdeffebach: I moved all of them out of the `describe` function but kept the `col` argument name.

test/dataframe.jl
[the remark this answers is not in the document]
nalimilan: Yes, CONTRIBUTING.md in the Julia repo gives that rule. Unfortunately there aren't many details about formatting conventions, so the best solution is to imitate surrounding code. In general we align arguments, so here I'd rather put `describe` on the next line, and if it's still necessary also keep the break in the middle of the array as you did.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -368,34 +368,42 @@ function Base.dump(io::IO, df::AbstractDataFrame, n::Int, indent)
...
 
-If the column's base type derives from Number, compute the minimum, first
-quantile, median, mean, third quantile, and maximum. Missings are filtered and
-reported separately.
+If the column's base type derives from `Number`, compute the mean, standard 
+deviation, minimum, first quantile, median, third quantile, and maximum. If 
+a column is not numeric, these statistics are populated with `nothing`s. 
 
-For boolean columns, report trues, falses, and missings.
+When `:nunique` is contained `stats` vector, `describe` will report the
```
nalimilan: "When `stats` contains `nunique`".

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -405,31 +413,99 @@ describe(df)
...
 
 """
-StatsBase.describe(df::AbstractDataFrame) = describe(stdout, df)
-function StatsBase.describe(io, df::AbstractDataFrame)
-    for (name, col) in eachcol(df)
-        println(io, name)
-        describe(io, col)
-        println(io, )
+StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
+function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector{Symbol} = [:mean, :min, :median, :max, :nmissing, :eltype])
```
nalimilan: Also break at 92 chars (after `;`).

test/dataframe.jl
```diff
@@ -325,25 +325,45 @@ module TestDataFrame
...
+                                              mean = [2.5, 2.0, nothing, nothing, nothing, nothing],
+                                              std = [Compat.std(df[:number]), 1.0, nothing, nothing, nothing, nothing],
+                                              min = [1.0, 1.0, nothing, nothing, nothing, nothing],
+                                              q25 = [1.75, 1.5, nothing, nothing, nothing, nothing],
+                                              median = [2.5, 2.0, nothing, nothing, nothing, nothing],
+                                              q75 = [3.25, 2.5, nothing, nothing, nothing, nothing],
+                                              max = [4.0, 3.0, nothing, nothing, nothing, nothing],
+                                              nunique = [nothing, nothing, 4, 4, 4, 2],
+                                              nmissing = [nothing, 1, nothing, 1, nothing, nothing],
+                                              eltype= [Int, Int, String, String, Date, eltype(df[:catarray])])
```
nalimilan: Add space before `=`.

test/dataframe.jl
```diff
@@ -325,25 +325,45 @@ module TestDataFrame
...
-    end
+    
+    # Test describe
+    @testset "describe(df)" begin
+        # Construct the test dataframe
+        df = DataFrame(number = [1, 2, 3, 4],
+                       number_missing = [1,2, 3, missing],
+                       non_number = ["a", "b", "c", "d"],
+                       non_number_missing = ["a", "b", "c", missing],
+                       dates  = Date.([2000, 2001, 2003, 2004]),
```
nalimilan: Add `using Compat.Dates` at the top of the file to fix failure on Julia 0.7.

bkamins:
Thank you for the PR. I was thinking about this for some time and it is great to have it on the plate 👍.

I have one suggestion to consider (it might be covered by what you write in point 1, but I could not find them). The special type we have is ordered categorical vector. In this case it makes sense to calculate quantiles.

Also documentation mentions `Number` while implementation uses `Real` if I am not mistaken. The difference is (apart from being precise) for complex numbers for which, e.g. it makes sense to calculate mean.

nalimilan:
> The special type we have is ordered categorical vector. In this case it makes sense to calculate quantiles.

AFAIK we don't support `quantile` for non-numeric types because we need to call `middle` when the number of values is even. There's been a thread recently about that.

bkamins:
> There's been a thread recently about that.

Ah - correct.

Could you please pass the link so that I can have a look (as under some definitions of empirical quantile it is well defined)?

pdeffebach:
> Also documentation mentions Number while implementation uses Real if I am not mistaken. The difference is (apart from being precise) for complex numbers for which, e.g. it makes sense to calculate mean.

Thanks for the feedback. I suppose a complex number is a number, but certainly isn't Real. I change the docs in a future pull request. 

`try...catch` and other behavior for arbitrary types is something I will start a PR for soon. It should be easy sense it just requires modifying the `get_stats` function. 

I will also submit a PR to StatsBase for a quantile-like function that returns the lower of the two in an even tie. @nalimilan does this seem like a good idea to you? [Here](https://discourse.julialang.org/t/functions-for-median-of-ordinal-data/10964/6) is a discourse thread about the subject.

nalimilan:
> I will also submit a PR to StatsBase for a quantile-like function that returns the lower of the two in an even tie. @nalimilan does this seem like a good idea to you? Here is a discourse thread about the subject.

Why not, at least as an option. Better file an issue in Julia to discuss that first (it's not defined in StatsBase).

nalimilan:
It seems the docs need updating. There's at least one example presenting the old output of `describe`: https://github.com/JuliaData/DataFrames.jl/blob/369e0fa33d5dda6e15553554874961bbec5f420b/docs/src/man/getting_started.md

pdeffebach:
opened up #1442 for this

PR #1415, opened 2018-06-04
Move Missing to it's own page in the Docs

oxinabox:
It really does not belong right at the top of getting started.
It doesn't belong on that page at all -- everything else is concerned with creating dataframes,
or basic manipulation of them.

The missing section doesn't even mention them.

It is not the *first* thing someone getting started wants to see.

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -171,6 +171,33 @@ julia> colwise(sum, df)
 
 ```
 
+### Data frames with Missing values
+If your dataset has missing values, most functions will require you to remove them
+beforehand. Here we will replace all odd-numbered rows in the first column with missing data
+to show how to handle the above example when missing values are present in your dataset.
+For more information see [Missing Data](@ref).
+
+```jldoctest dataframe
+julia> df[:A] = [isodd(i) ? missing : value for (i, value) in enumerate(df[:A])];
+
+julia> df
+8×2 DataFrames.DataFrame
+│ Row │ A       │ B │
+├─────┼─────────┼───┤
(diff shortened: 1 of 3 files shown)


docs/src/man/getting_started.md
```diff
@@ -344,6 +245,9 @@ julia> head(iris)
 
 ```
 
+
```
nalimilan: These three line breaks aren't consistent with the rest of the file AFAICT.

oxinabox:
> Woops, can you also move the remaining mention of missing from the main page to the dedicated section?

I kind of left it intentionally, because I think missing should get a mention at least on the getting started.

I just took a look at what it would look-like if I moved it over to the missing data page,
but it doesn't look great,  there are plenty of clearer examples their already,
and this one uses the same dataframe as the rest of the examples on getting started.

So in the last commit I have added a link as a "For more info see Missing Data".

nalimilan:
Maybe just drop that example if it's redundant with that we have on the Missing values page. I don't think it makes sense to say "If your data frame has missing values" without any explanation of what missing values are nor how they are represented. Better have just a short paragraph telling the reader to look the the "Missing values" section for information about missing values.

oxinabox:
Are you sure? It has some virtue

nalimilan:
What virtue? Since `missing` hasn't been presented, it can only confuse people.

oxinabox:
I guess I was thinking as a kind of "taster" that shows `missing` exists,
followed by a "See more" that would un-confuse them.

oxinabox:
Is this good now?

PR #1418, opened 2018-06-06
Make `describe` more type-agnostic

pdeffebach:
This PR splits the behavior of the `get_stats` functions for `describe` into three ways

- Real numbers, same as before
- Strings, same as before (mostly `nothings`)
- Any
  - I added `try...catch` for determining the statistics when we we are working with other types
  - That way, aslong as, say `isless` is defined, we can return `minimum` and `maximum` for that column. 

In particular, this addresses concerns about `Date` types and ordered categorical array types, but obviously will apply to far more types.


bkamins:
I recommend to add in this PR:
* deprecation message for `showcols` stating that `describe` is recommended;
* add `first` (first element of the vector), `last` (last element of the vector) and `unique`  (number of unique values in the vector) statistics; to be discussed if they should be shown by default; I recommend them to be shown (or at least `unique` to be shown) - otherwise `describe` is not very useful for non-numeric data; I think it is not a problem that they are added as showing a produced `DataFrames` by default will drop displaying those columns to fit the screen width (and if someone wants to see all one can use `showall`); the order could be `unique` after `max` but before `nmissing` and `fist` and `last` after `eltype` (so that they are not displayed if the output is very wide).

pdeffebach:
Thanks for the feedback, I'm working on it now. 

I hit a roadblock when writing tests and filed an issue #1420 about comparisons with missing values. This happened because `last` will return missing for some of the columns in the test dataframe. 

PS I also added an keyword argument `allstats = false` that when `true` returns all the allowed fields.

pdeffebach:
I think this is ready for review again. Again, sorry about all the commits. I just now started using Revise.jl and working in the package directory...

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -499,7 +502,7 @@ function get_stats(col::AbstractArray{>:Missing})
     )    
 end 
 
-function get_stats(col)
+function get_stats(col::AbstractArray{String})
```
nalimilan: Do we really need these special `String` methods? AFAICT they give the same result as the fallback except for minimum and maximum, right? I'd rather not add a special method just for that, and printing "minimum" and "maximum" strings can make sense (e.g. for ordered categories). At the very least it doesn't really hurt and it's not completely unintuitive for users to use lexicographic ordering.

test/dataframe.jl
```diff
@@ -332,43 +332,48 @@ module TestDataFrame
...
         @test describe_output == describe(df)
 
+        # Test that it works with one stats argument
+        @test describe_output[[:variable, :mean]] == describe(df, stats = [:mean])
+
         # Test that it works with all keyword arguments
-        @test describe_output_all_stats == 
-              describe(df, stats = [:mean, :std, :min, :q25, :median, :q75, :max, 
-                       :nunique, :nmissing, :eltype])
+        # Use isqual because we have `missing`s in this dataframe
```
nalimilan: Use `≅` (`\congr[tab]`, see rest of the file), and no need for a comment.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -422,17 +425,20 @@ describe(df)
 """
 StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
 function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector{Symbol} = 
-                            [:mean, :min, :median, :max, :nmissing, :eltype])
+                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype], 
+                            allstats = false)
```
nalimilan: Why not use a special name like `:all` instead?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -390,20 +390,23 @@ describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissi
 * `io` : optional output descriptor
 * `stats::AbstractVector{Symbol}`: the summary statistics to report. Allowed 
   fields are `:mean`, `:std`, `:min`, `:q25`, `:median`, `:q75`, `:max`, `:eltype`, 
-  `:nunique`, and `:nmissing`
+  `:nunique`, `first', `last`, and `:nmissing`
```
nalimilan: Should be `:first` and `:last`. Also the first closing backtick isn't a real one.

pdeffebach:
This is ready for another review. My git skills are improving, I think. 

I'm not sure I did the deprecation correctly. Should I also modify `depracated.jl`?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -390,20 +390,23 @@ describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissi
...
 **Result**
 
 * A `DataFrame` where each row represents a variable and each column a summary statistic.
 
 **Details**
 
-If the column's base type derives from `Number`, compute the mean, standard 
-deviation, minimum, first quantile, median, third quantile, and maximum. If 
-a column is not numeric, these statistics are populated with `nothing`s. 
+* If the column's base type derives from `Real`, compute the mean, standard
```
nalimilan: "a column". Could also just say "For `Real` columns" on so on below. But maybe you could just say that all these stats are computed if possible, as you say for the fallback?

Actually, do we really need special methods for `Real` now? They should give the same result, there are just additional `try... catch` checks which don't really hurt, right?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -390,20 +390,23 @@ describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissi
...
 * A `DataFrame` where each row represents a variable and each column a summary statistic.
 
 **Details**
 
-If the column's base type derives from `Number`, compute the mean, standard 
-deviation, minimum, first quantile, median, third quantile, and maximum. If 
-a column is not numeric, these statistics are populated with `nothing`s. 
+* If the column's base type derives from `Real`, compute the mean, standard 
+  deviation, minimum, first quantile, median, third quantile, and maximum. 
+* If a column's type derives from `String`, these statistics are populated with `nothing`s.
```
nalimilan: This doesn't apply anymore right?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -422,17 +425,20 @@ describe(df)
 """
 StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
 function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector{Symbol} = 
-                            [:mean, :min, :median, :max, :nmissing, :eltype])
+                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype], 
+                            all = false)
```
nalimilan: What about my suggestion to use an `:all` pseudo-stat instead?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -422,17 +425,20 @@ describe(df)
...
+                      :max, :nunique, :nmissing, :first, :last, :eltype] 
     if !issubset(stats, allowed_fields) 
         disallowed_fields = setdiff(stats, allowed_fields)
         not_allowed = "Field(s) not allowed: $disallowed_fields. "
         allowed = "Allowed fields are: $allowed_fields."
         throw(ArgumentError(not_allowed * allowed)) 
     end
 
+    # See if the user wants to show all summary statistics
+    all == true ? stats = allowed_fields : nothing
```
nalimilan: Better use an `if` block or `&&`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -448,68 +454,80 @@ function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector
...
-        :mean => nothing,
-        :std => nothing,
-        :min => nothing,
-        :q25 => nothing,
-        :median => nothing,
-        :q75 => nothing,
-        :max => nothing,
+        :mean => try mean(nomissing) catch end,
+        :std => try Compat.std(nomissing) catch end,
+        :min => try minimum(nomissing) catch end,
```
nalimilan: Would be nice to use `extrema` to avoid one additional pass over the data. If that fails, we can return `nothing` for both minimum and maximum. Same below.

src/abstractdataframe/show.jl
```diff
@@ -569,6 +569,7 @@ end
 #' showcols(df)
 function showcols(io::IO, df::AbstractDataFrame, all::Bool = false,
                   values::Bool = true) # -> Nothing
+    Base.depwarn("showcols has been deprecated. Use `describe(df, stats = [:eltype, :nmissing, :first, :last]` instead", :showcols)
```
nalimilan: The standard approach for deprecations is to completely remove this function and add this to deprecated.jl:
```julia
@deprecate showcols(df::AbstractDataFrame, all::Bool=false, values::Bool=true) describe(df, stats = [:eltype, :nmissing, :first, :last])
@deprecate showcols(io::IO, df::AbstractDataFrame, all::Bool=false, values::Bool=true) show(io, describe(df, stats = [:eltype, :nmissing, :first, :last]), all)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -422,17 +425,20 @@ describe(df)
 """
 StatsBase.describe(df::AbstractDataFrame; kwargs...) = describe(stdout, df; kwargs...)
 function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector{Symbol} =
```
nalimilan: I had missed this in the previous PR, but the `io` argument isn't used now. It should be removed (and the two methods merged), with a deprecation like this in deprecated.jl:
```julia
 @deprecate StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector{Symbol}=[:mean, :min, :median, :max, :nunique, :nmissing, :eltype]) show(io, describe(df, stats=stats))
```

src/abstractdataframe/abstractdataframe.jl
pdeffebach: The only difference would be `unique()`, which might be slow. I will do some benchmarking to see if it really matters

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Oh sorry I didn't understand the first time. Will add.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: But that's not julian! Only multiple dispatch!

I will make that addition. It definitely makes things simpler.

src/abstractdataframe/show.jl
pdeffebach: Okay, so showcols no longer exists and the result just calls `describe`?

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
pdeffebach: I have edited the docstring.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Have added it, and changed the test.

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/show.jl
pdeffebach: I deleted all the `showcols` code and added the above code to `deprecated`. I think this works.

pdeffebach: I also purged the tests of `showcols`. I hope this wasn't too extreme

src/abstractdataframe/abstractdataframe.jl
pdeffebach: So we should have just one function definition, `StatsBase.describe(df::AbstractDataFrame, stats...)`?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -422,17 +424,23 @@ describe(df)
...
+                      :max, :nunique, :nmissing, :first, :last, :eltype, :all] 
     if !issubset(stats, allowed_fields) 
         disallowed_fields = setdiff(stats, allowed_fields)
         not_allowed = "Field(s) not allowed: $disallowed_fields. "
         allowed = "Allowed fields are: $allowed_fields."
         throw(ArgumentError(not_allowed * allowed)) 
     end
 
+    # See if the user wants to show all summary statistics
+    if contains(==, stats, :all)
```
nalimilan: `:all in stats` is enough. The comment isn't really useful.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -448,68 +456,58 @@ function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector
...
-        :max => sumstats.max,
-        :nmissing => count(ismissing, col),
-        :nunique => nothing,
-        :eltype => Missings.T(eltype(col))
-    )
-end
+    
+    q = try quantile(nomissing, [.25, .5, .75]) catch [nothing, nothing, nothing] end
+    ex = try extrema(nomissing) catch (nothing, nothing) end
+    if typeof(nomissing) <: AbstractArray{<:Real}
```
nalimilan: `eltype(nomissing) <: Real` would be simpler, and the `?` ternary syntax should be fine here.

test/dataframe.jl
nalimilan: Still no need for the comment (and there's a typo in it).

src/abstractdataframe/abstractdataframe.jl
nalimilan: Yes.

pdeffebach: Fixed this on the second commit below. Should be ready, then.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -420,19 +422,23 @@ describe(df)
...
-                      :max, :nunique, :nmissing, :eltype] 
+                      :max, :nunique, :nmissing, :first, :last, :eltype, :all] 
     if !issubset(stats, allowed_fields) 
         disallowed_fields = setdiff(stats, allowed_fields)
         not_allowed = "Field(s) not allowed: $disallowed_fields. "
         allowed = "Allowed fields are: $allowed_fields."
         throw(ArgumentError(not_allowed * allowed)) 
     end
 
+    if contains(==, stats, :all)
```
nalimilan: `:all in stats` would still be enough.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -448,68 +454,58 @@ function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector
...
     Dict(
-        :mean => nothing,
-        :std => nothing,
-        :min => nothing,
-        :q25 => nothing,
-        :median => nothing,
-        :q75 => nothing,
-        :max => nothing,
+        :mean => try mean(nomissing) catch end,
+        :std => try Compat.std(nomissing) catch end,
```
nalimilan: It would be nice to keep `Compat.std(col, mean = m)`. If `mean` could not be computed, it will be `nothing` and the call will fail, which is OK since it's unlikely that we would be able to compute the standard deviation but not the mean. Same in the other method below.

test/dataframe.jl
```diff
@@ -332,43 +332,47 @@ module TestDataFrame
...
         @test describe_output == describe(df)
 
+        # Test that it works with one stats argument
+        @test describe_output[[:variable, :mean]] == describe(df, stats = [:mean])
+
         # Test that it works with all keyword arguments
-        @test describe_output_all_stats == 
-              describe(df, stats = [:mean, :std, :min, :q25, :median, :q75, :max, 
-                       :nunique, :nmissing, :eltype])
+        @test describe_output_all_stats ≅ describe(df, stats = [:all])
```
nalimilan: Looking at this, it just occurred to me that it would be nice to allow passing `:all` directly. Actually, it would make even more sense to reject `:all` when it's inside an array, since it's redundant with other entries. So we could just check that if `stats` is a `Symbol` it's equal to `:all`.

pdeffebach: Made this change and added it to the docs and errors.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: made this change

src/abstractdataframe/abstractdataframe.jl
pdeffebach: changed it so a symbol.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -390,21 +390,23 @@ describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissi
 * `io` : optional output descriptor
 * `stats::AbstractVector{Symbol}`: the summary statistics to report. Allowed 
   fields are `:mean`, `:std`, `:min`, `:q25`, `:median`, `:q75`, `:max`, `:eltype`, 
-  `:nunique`, and `:nmissing`
+  `:nunique`, `:first`, `:last`, `:nmissing`, and `:all`
```
nalimilan: `:all` is no longer allowed as a field. Also should use `stats::Union{Symbol,AbstractVector{Symbol}}`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -390,21 +390,23 @@ describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissi
...
 **Result**
 
 * A `DataFrame` where each row represents a variable and each column a summary statistic.
 
 **Details**
 
-If the column's base type derives from `Number`, compute the mean, standard 
-deviation, minimum, first quantile, median, third quantile, and maximum. If 
-a column is not numeric, these statistics are populated with `nothing`s. 
+For `Real` columns, compute the mean, standard eviation, minimum, first quantile, median,
```
nalimilan: "deviation"

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -390,21 +390,23 @@ describe(io, df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissi
...
+third quantile, and maximum. If a column does not derive from `Real`, `describe` will 
+attempt to calculate all statistics, using `nothing` as a fall-back in the case of an error.
 
 When `stats` contains `:nunique`, `describe` will report the 
-number of unique values in a column. If a column's base type derives from `Number`,
-`:nunique` will return `nothing`s.
+number of unique values in a column. If a column's base type derives from `Real`,
+`:nunique` will return `nothing`s. 
+
+When `stats` equals `:all`, `describe` will report all summary statistics.
```
nalimilan: Better mention this when documenting `stats`.

pdeffebach: okay hopefully its ready to merge now.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -380,31 +380,30 @@ end
...
-a column is not numeric, these statistics are populated with `nothing`s. 
+For `Real` columns, compute the mean, standard deviation, minimum, first quantile, median, 
+third quantile, and maximum. If a column does not derive from `Real`, `describe` will 
+attempt to calculate all statistics, using `nothing` as a fall-back in the case of an error.
 
 When `stats` contains `:nunique`, `describe` will report the 
-number of unique values in a column. If a column's base type derives from `Number`,
-`:nunique` will return `nothing`s.
+number of unique values in a column. If a column's base type derives from `Real`,
+`:nunique` will return `nothing`s.
```
bkamins: Please remove space at the end of the line.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -417,22 +416,32 @@ missing values, but does not currently contain any.
...
-                            [:mean, :min, :median, :max, :nmissing, :eltype])
+function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector{Symbol}} = 
+                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
     # Check that people don't specify the wrong fields. 
     allowed_fields = [:mean, :std, :min, :q25, :median, :q75, 
-                      :max, :nunique, :nmissing, :eltype] 
-    if !issubset(stats, allowed_fields) 
+                      :max, :nunique, :nmissing, :first, :last, :eltype] 
+    
+    if typeof(stats) == Symbol && stats != :all
```
bkamins: This throws error when a call like `describe(df, stats=:mean)` is passed. I thought we want to allow passing a single stat not inside a vector. Also The printed error looks e.g. like:

```
ERROR: ArgumentError: Field(s) not allowed: mean. Allowed fields are: Symbol[:mean, :std, :min, :q25, :median, :q75, :max, :nunique, :nmissing, :first, :last, :eltype], or `stats = :all`
```
I would not print `Symbol` in front of the bracket (but it is a minor issue).

Additionally in this case you know that one field was passed so `Field not allowed:` would be better correct (also a minor issue).

If we do not want to allow passing a single symbol then I would recommend to make an error message clearer that the passed list of fields must be a vector. Now the error message is not very clear.

bkamins: No matter what functionality is finally implemented I would test passing a single `Symbol`, e.g. `:mean`, as `stats` (to test that it either throws an error or produces a correct result).

pdeffebach: I was under the impression it was either a a vector `[:mean, :min, :max...]` or the symbol `:all`. 

If we want `stats = :mean` to be allowed, we could add something like 

`isa(stats, Symbol) ? stats = [stats] : nothing`

bkamins: For me it is OK to require a vector or `:all` but then I recommend to make the error message clearer.

The fix should rather be to add after line 430 something like:

```
if stats in allowed_fields
    stats = [stats]
else
... here go on with the error message we currently have
end
```

pdeffebach: I think I will do the above, because if someone just wants the means of all their columns, it makes sense for them to write `stats = :mean`. An alternative way to do that is a bit more cumbersome.

pdeffebach: Here is what I have now. I try to have two separate error messages, for if someone put in the wrong symbol and if they put in the wrong vector of symbols. 

With regards to the awkward `Symbol[:whatever]` in the error message, this is because of the escaping `$`. I'm not sure how to avoid this without writing it out manually, which I can do obviously. 

Should I push this change and then you can tweak the error message if you want? 

```
    # Check that people don't specify the wrong fields. 
    allowed_fields = [:mean, :std, :min, :q25, :median, :q75, 
                      :max, :nunique, :nmissing, :first, :last, :eltype] 
    if stats == :all
        stats = allowed_fields 
    end

    if typeof(stats) == Symbol 
        if !(stats in allowed_fields)
            throw(ArgumentError("$stats not allowed. Allowed fields are " * 
                ":mean, :std, :min, :q25, :median, :q75, " *
                ":max, :nunique, :nmissing, :first, :last, :eltype"))
        else 
            stats = [stats]
        end
    end

    if !issubset(stats, allowed_fields)
        disallowed_fields = setdiff(stats, allowed_fields)
        not_allowed = "Field(s) not allowed: $disallowed_fields. "
        allowed = "Allowed fields are: $allowed_fields"
        throw(ArgumentError(not_allowed * allowed)) 
    end
```

bkamins: Great. As for error message I would go for something like:

```
 # Check that people don't specify the wrong fields. 
    allowed_fields = [:mean, :std, :min, :q25, :median, :q75, 
                      :max, :nunique, :nmissing, :first, :last, :eltype]
    allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
    if stats == :all
        stats = allowed_fields 
    end

    if typeof(stats) == Symbol 
        if !(stats in allowed_fields)
            throw(ArgumentError("$stats not allowed." * allowed_msg))
        else 
            stats = [stats]
        end
    end

    if !issubset(stats, allowed_fields)
        disallowed_fields = setdiff(stats, allowed_fields)
        not_allowed = "Field(s) not allowed: :" * join(disallowed_fields, ", :") * "."
        throw(ArgumentError(not_allowed * allowed_msg)) 
    end
```
does it look good to you?

pdeffebach: Addded and pushed. I add a ": " somewhere so the error message reflects a symbol.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -380,57 +380,69 @@ end
...
-function StatsBase.describe(io::IO, df::AbstractDataFrame; stats::AbstractVector{Symbol} = 
-                            [:mean, :min, :median, :max, :nmissing, :eltype])
+function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector{Symbol}} = 
+                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
     # Check that people don't specify the wrong fields. 
     allowed_fields = [:mean, :std, :min, :q25, :median, :q75, 
-                      :max, :nunique, :nmissing, :eltype] 
-    if !issubset(stats, allowed_fields) 
+                      :max, :nunique, :nmissing, :first, :last, :eltype]
+    allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
```
nalimilan: Better call this only in case of errors. It's not a big deal if you repeat this twice.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -380,57 +380,69 @@ end
...
+    if stats == :all
+        stats = allowed_fields 
+    end
+
+    if typeof(stats) == Symbol
```
nalimilan: Minor point, but you can do `stats isa Symbol`.

pdeffebach: pushed!

nalimilan:
Thanks!

Issue #1412, opened 2018-05-25
Feature request for `join`


nickeubank:
[ Carrying over from [Discourse](https://discourse.julialang.org/t/two-feature-requests-for-merge) ] 

I'd like to propose two added functions to the `join` function to be offered as optional keywords:

- `validate`: duplicates the functionality of `validate` keyword in the `pandas` [merge](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.merge.html) function. Accepts `"1:1"`, `"1:m"`, `"m:1"`, and `"m:m"`, and raises an exception if the merge is not 1 to 1, 1 to many, many to 1, or many to many (respectively). 
- `indicator`: duplicates functionality of `indicator` keyword in `pandas` [merge] function. If `True`, adds column to returned object which records whether resulting row has data from `both` datasets, from the `left_only`, or the `right_only` (presumably stored as categoricals? Or something else if there are performance issues.)

(Both are actually replications of behavior from [Stata](https://www.stata.com/manuals13/dmerge.pdf))

Personally, I find these exceedingly value when working with real world data, as there's no place problems become more evident than in merges, and it gets exhausting writing code the replicates these functionalities every time I merge (especially the `indicator` command). 

As noted in the Discourse discussion, there are ways of getting at these with existing tools, but (at least in my experience) they're so used I think a little syntactic sugar is warranted.

I’m an empirical social scientist, and work with data from lots of sources (census, various government agencies, other researchers, etc.). Those data sources are often supposed to relate in certain ways (Everyone in dataset A should also be in B, but there should be people in B not in A), but they almost inevitably don’t. Just doing a Left or Right join would get only the people who fit the promised relationship, but I (a) want to know how many people didn’t merge right (so I usually do an Outer join, then check the merge-state, and (b) when things go wrong, I want to look at the errors (so it’s nice to quickly query the people who are in A but not in B, in the example above).

I recognize this is not a problem that many CS people (with nice, single-source datasets) or people in the private sector (working with data from a single source, or from a SQL database that has inbuilt checks for these kinds of things) deal with much, but I think this is super common among social scientists.

bkamins:
I do not see a problem with adding both (but maybe other people would have some objections). Please just specify the desired design.

* for `incicator` keyword argument:
    * do you want it to be `Bool` or something more general like in Pandas?
    * what should be a column name (or should there be one column or two columns - one for left and one for right)?
[...]

PR #1421, opened 2018-06-06
first pass at key uniqueness checking

nickeubank:
Half of #1412 

Can anyone advise on how to handle the `on` argument for stranger cases (like dictionaries)? Looks like there's a system in place (`OnType`), but can't quite figure out.

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -213,7 +213,8 @@ function update_row_maps!(left_table::AbstractDataFrame,
 end
 
 """
-    join(df1, df2; on = Symbol[], kind = :inner, makeunique = false)
+    join(df1, df2; on = Symbol[], kind = :inner, makeunique = false,
+         indicator = nothing, validate = (false, false))
 
 Join two `DataFrame` objects
 
@@ -252,6 +253,12 @@ Join two `DataFrame` objects
    only `df2` (`"right_only"`) or in both (`"both"`). If `Symbol` is already in use,
    the column name will be modified if `makeunique=true`.
 
+* `validate` : whether to check that columns passed as the `on` argument
+   define unique keys in each input data frame (according to [`isequal`](@ref)).
+   Can be a tuple or a pair, with the first element indicating whether to
+   run check for `df1` and the second element for `df2`.
+   By default no check is performed.
+
 For the three join operations that may introduce missing values (`:outer`, `:left`,
 and `:right`), all columns of the returned data table will support missing values.
 
@@ -286,7 +293,8 @@ function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
                    kind::Symbol = :inner, makeunique::Bool=false,
-                   indicator::Union{Nothing, Symbol} = nothing)
+                   indicator::Union{Nothing, Symbol} = nothing,
+                   validate::Union{Pair{Bool, Bool}, Tuple{Bool, Bool}}=(false, false))
     if indicator !== nothing
         indicator_cols = ["_left", "_right"]
         for i in 1:2
@@ -296,8 +304,8 @@ function Base.join(df1::AbstractDataFrame,
                  indicator_cols[i] *= 'X'
             end
          end
-        df1[Symbol(indicator_cols[1])] = trues(nrow(df1))
-        df2[Symbol(indicator_cols[2])] = trues(nrow(df2))
+         df1[Symbol(indicator_cols[1])] = trues(nrow(df1))
+         df2[Symbol(indicator_cols[2])] = trues(nrow(df2))
     end
 
     if kind == :cross
@@ -308,6 +316,26 @@ function Base.join(df1::AbstractDataFrame,
     end
 
     joiner = DataFrameJoiner(df1, df2, on)
+
+    # Check merge key validity
+    left_invalid = validate[1] ? any(nonunique(joiner.dfl, joiner.left_on)) : false
+    right_invalid = validate[2] ? any(nonunique(joiner.dfr, joiner.right_on)) : false
+
+    if left_invalid && right_invalid
+        first_error_df1 = findfirst(nonunique(joiner.dfl, joiner.left_on))
+        first_error_df2 = findfirst(nonunique(joiner.dfr, joiner.right_on))
+        throw(ArgumentError("Merge key(s) are not unique in both df1 and df2. " *
+                            "First duplicate in df1 at $first_error_df1. " *
+                            "First duplicate in df2 at $first_error_df2"))
+    elseif left_invalid
+        first_error = findfirst(nonunique(joiner.dfl, joiner.left_on))
+        throw(ArgumentError("Merge key(s) in df1 are not unique. " *
+                            "First duplicate at row $first_error"))
+    elseif right_invalid
+        first_error = findfirst(nonunique(joiner.dfr, joiner.right_on))
+        throw(ArgumentError("Merge key(s) in df2 are not unique. " *
+                            "First duplicate at row $first_error"))
+    end
 
     if kind == :inner
         joined = compose_joined_table(joiner, kind, update_row_maps!(joiner.dfl_on, joiner.dfr_on,
--- a/test/join.jl
+++ b/test/join.jl
@@ -578,6 +578,7 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
+
     @testset "indicator columns" begin
         outer_indicator = DataFrame(ID = [1, 2, 2, 3, 4],
                                     Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
@@ -608,6 +609,7 @@ module TestJoin
         @test join(name2, job2, on = :ID, kind = :outer, indicator=:_left,
                    makeunique=true) ≅ outer_indicator
     end
+
     @testset "test checks of merge key uniqueness" begin
         @test_throws ArgumentError join(name, job, on=:ID, validate=(false, true))
         @test_throws ArgumentError join(name, job, on=:ID, validate=(true, true))
--- a/test/join.jl
+++ b/test/join.jl
@@ -578,7 +578,6 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
-
     @testset "indicator columns" begin
         outer_indicator = DataFrame(ID = [1, 2, 2, 3, 4],
                                     Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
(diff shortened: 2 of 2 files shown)


nickeubank:
OK, think I fixed the `OnType` issue...

src/abstractdataframe/join.jl
```diff
@@ -247,6 +273,11 @@ Join two `DataFrame` objects
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 
+* `check_uniqueness` : checks for whether merge keys are unique before merging.
```
nalimilan: I suggest something like (feel free to improve it if you have ideas):
"whether to check that columns passed as the `on` argument define unique keys in each input data frame. Can be a tuple or a pair, with the first element indicating whether to run check for `df1` and the second element for `df2`. By default no check is performed."

Also the name of the argument isn't great, as we already have `makeunique` which refers to column names. Not sure what's the best name. Maybe `validate` as in Pandas? I agree it's not very specific.

src/abstractdataframe/join.jl
```diff
@@ -280,7 +311,9 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner, makeunique::Bool=false)
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   check_uniqueness::Tuple{Bool,Bool}=(false, false))
```
nalimilan: Would be nice to also allow `Pair{Bool,Bool}` as we do for `on`.

Also don't add space below.

src/abstractdataframe/join.jl
```diff
@@ -212,8 +212,34 @@ function update_row_maps!(left_table::AbstractDataFrame,
     return to_bimap(left_ixs), to_bimap(leftonly_ixs), to_bimap(right_ixs), to_bimap(rightonly_ixs)
 end
 
+function check_uniqueness_func(check_uniqueness, joiner)
```
nalimilan: Given that it's only called from one place, there's no point in defining  a separate function.

src/abstractdataframe/join.jl
```diff
@@ -212,8 +212,34 @@ function update_row_maps!(left_table::AbstractDataFrame,
...
+
+    if check_uniqueness == (true, false) && any(nonunique(joiner.dfl, joiner.left_on))
```
nalimilan: I'd simplify this code by handling `df1` and `df2` separately (if `check_uniqueness` is false then just do as if the check was OK), and only then have an `if` with three branches determining the error message to print.

Since you compute `nonunique`, it would also be nice to give the name of the first row for which the problem was found. It's very useful when trying to find out the problem.

test/join.jl
```diff
@@ -578,4 +578,23 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
+
+    @testset "test checks of merge key uniqueness" begin
+
```
nalimilan: No need for all these empty lines here nor below (see rest of the file for examples).

test/join.jl
```diff
@@ -578,4 +578,23 @@ module TestJoin
...
+        @test_throws ArgumentError join(name, job, on=:ID,
+                     check_uniqueness=(false, true))
```
nalimilan: Align arguments with parentheses (or maybe they even fit on the previous line?).

test/join.jl
```diff
@@ -578,4 +578,23 @@ module TestJoin
...
+        @test_throws ArgumentError join(name, job, on=:ID,
+                      check_uniqueness=(true, true))
+        @test_throws ArgumentError join(job, name, on=:ID,
+                      check_uniqueness=(true, false))
+        @test_throws ArgumentError join(job, name, on=:ID,
+                      check_uniqueness=(true, true))
+
+        @test join(name, job, on=:ID, check_uniqueness=(true, false)) ==  inner
+        @test join(name, job, on = :ID, kind = :inner, check_uniqueness=(false, false)) == inner
```
nalimilan: No spaces around `=` for consistency.

test/join.jl
```diff
@@ -578,4 +578,23 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
+
+    @testset "test checks of merge key uniqueness" begin
```
nalimilan: Can you also test what happens when the keys contain missing values? What are we supposed to do in that case (and what does Stata do)?

src/abstractdataframe/join.jl
nickeubank: I split it out to minimize computational costs. This way we don't ever run a `nonunique` unnecessarily. If I do `df1` and `df2` up front, we run it on both even if we only need to check one. 

> Since you compute nonunique, it would also be nice to give the name of the first row for which the problem was found. It's very useful when trying to find out the problem.

Good idea. Will put in error message.

src/abstractdataframe/join.jl
nickeubank: I know I don't have to, but there is a point: good decomposition / style! It's a lot of clutter to leave in the core function definition, no?

test/join.jl
nickeubank: test added

src/abstractdataframe/join.jl
nalimilan: Yes but you can do something like:
```julia
left_valid = check_uniqueness[1] ? any(nonunique(joiner.dfl, joiner.left_on)) : true
right_valid = check_uniqueness[2] ? any(nonunique(joiner.dfr, joiner.right_on)) : true
```

src/abstractdataframe/join.jl
nalimilan: I think with my suggestion you should be able to make it shorter.

src/abstractdataframe/join.jl
nickeubank: ok

nickeubank:
fixed, rebased, and squashed

bkamins:
I am not sure if my comments are visible - at least I do not see them (sorry - no experience in reviewing yet). I will sum up what I suggested other than @nalimilan:

1. clearly write that `isequal` is used to check for uniqueness in the documentation.
2. Add tests not only for `missing` but also for `NaN` and `0.0`/`-0.0`.

Sorry for the mess

nalimilan:
> Is that what's used for nonunique? I'm just using nonunique. Is it better to cite that?

Yes, it uses the same grouping code as `groupby`, which is based on the `Dict` code, which in turn uses `isequal`. It's probably clearer to mention `isequal`, which people are more likely to be familiar with.

bkamins:
This is the point; `isequal` is a standard contract defined in Base.

nickeubank:
ok, think that hits all the issues raised!

src/abstractdataframe/join.jl
```diff
@@ -247,6 +248,12 @@ Join two `DataFrame` objects
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 
+* `validate` : whether to check that columns passed as the on argument
```
nalimilan: `` `on` ``

src/abstractdataframe/join.jl
```diff
@@ -247,6 +248,12 @@ Join two `DataFrame` objects
...
+   define unique keys in each input data frame. Can be a tuple or a pair,
+   with the first element indicating whether to run check for `df1` and
+   the second element for `df2`. By default no check is performed.
+   Merge key equality it tested using `isequal`.
```
nalimilan: "Merge" isn't the terminology we use. I'd just add this mention this after "in each input data frame", like this: ``(according to [`isequal`](@ref)) ``.

src/abstractdataframe/join.jl
```diff
@@ -290,6 +298,26 @@ function Base.join(df1::AbstractDataFrame,
...
+
+    if left_invalid && !right_invalid
+        first_error = findfirst(nonunique(joiner.dfl, joiner.left_on))
+        throw(ArgumentError("Merge key(s) in df1 are not unique. " *
+                            "First duplicate at row $first_error"))
+    elseif !left_invalid && right_invalid
+        first_error = findfirst(nonunique(joiner.dfr, joiner.right_on))
+        throw(ArgumentError("Merge key(s) in df2 are not unique. " *
+                            "First duplicate at row $first_error"))
+    elseif left_invalid && right_invalid
```
nalimilan: If you put this branch first, you can avoid the `!left_invalid` and `!right_invalid` checks.

test/join.jl
```diff
@@ -578,4 +578,54 @@ module TestJoin
...
+        @test_throws ArgumentError join(job, name, on=:ID,
+                                        validate=(true, true))
+        @test_throws ArgumentError join(job, job, on=:ID,
+                                        validate=(true, true))
+
+
+        @test join(name, job, on=:ID, validate=(true, false)) ==  inner
+        @test join(name, job, on=:ID, kind=:inner, validate=(false, false)) == inner
+
+        # Make sure ok with various weird characters
```
nalimilan: These are not "characters". And "weird" could just be "special".

Below, it would be useful to include a `0` in the values of `ID`, to check that it's treated as different from `-0.0`. Then you can drop `0.0` from the list here. Without this, the code doesn't actually test that `0.0` and `-0.0` are considered as different.

test/join.jl
```diff
@@ -578,4 +578,54 @@ module TestJoin
...
+        for weird in [missing, NaN, 0.0, -0.0]
+            name_w_missing = DataFrame(ID = [1, 2, 3, weird],
+                                       Name = Union{String, Missing}["John Doe", "Jane Doe",
```
nalimilan: `Union{String, Missing}` isn't useful here, right? Same below.

test/join.jl
```diff
@@ -578,4 +578,54 @@ module TestJoin
...
+                                                                     "Joe Blogs", "Maria Tester"])
+            @test join(name_w_missing, job, on=:ID, validate=(true, false)) ==  inner
+
+            # Make sure duplicated missings still an exception
```
nalimilan: Comment and variable names should be adjusted since you test other values than missing.

test/join.jl
```diff
@@ -578,4 +578,54 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
+
+    @testset "test checks of merge key uniqueness" begin
+        @test_throws ArgumentError join(name, job, on=:ID,
+                                        validate=(false, true))
```
nalimilan: Better put this on the previous line, it's well below 92 chars.

test/join.jl
```diff
@@ -578,4 +578,54 @@ module TestJoin
...
+                                Name = Union{String, Missing}["John Doe", "Jane Doe", "Joe Blogs"])
+         job_multi = DataFrame(ID1 = Union{Int, Missing}[1, 2, 2, 4],
+                               ID2 = ["a", "b", "b", "c"],
+                               Job = Union{String, Missing}["Lawyer", "Doctor", "Florist", "Farmer"])
+         outer_multi = DataFrame(ID1 = [1, 1, 2, 2, 2, 4],
+                                 ID2 = ["a", "b", "a", "b", "b", "c"],
+                                 Name = ["John Doe", "Jane Doe", "Joe Blogs", missing, missing, missing],
+                                 Job = ["Lawyer", missing, missing, "Doctor", "Florist",  "Farmer"])
+
+         @test isequal(join(name_multi, job_multi, on=[:ID1, :ID2], kind=:outer, validate=(true, false)),
```
nalimilan: Use `≅`.

test/join.jl
nalimilan: Julia's CONTRIBUTING.md mentions 92 chars, that's usually the target we try to follow.

test/join.jl
nalimilan: These type specifications have been preserved in existing tests when moving from `DataArray`, to make sure we still test such arrays. But in new code it's generally not needed.

nickeubank: k

test/join.jl
nickeubank: wow that is so much more breathing room

nickeubank:
ok I think that covers everything. Thanks for all the help @nalimilan @bkamins

src/abstractdataframe/join.jl
```diff
@@ -290,6 +298,26 @@ function Base.join(df1::AbstractDataFrame,
...
     joiner = DataFrameJoiner(df1, df2, on)
 
+    # Check merge key validity
+    left_invalid = validate[1] ? any(nonunique(joiner.dfl, joiner.left_on)) : false
+    right_invalid = validate[2] ? any(nonunique(joiner.dfr, joiner.right_on)) : false
+
+    if left_invalid && right_invalid
+        first_error_df1 = findfirst(nonunique(joiner.dfl, joiner.left_on))
+        first_error_df2 = findfirst(nonunique(joiner.dfr, joiner.right_on))
+        throw(ArgumentError("Merge key(s) are not unique in df1 or df2. " *
```
nalimilan: "or" -> "and"?

test/join.jl
```diff
@@ -578,4 +578,64 @@ module TestJoin
...
+                                  Name = ["John Doe", "Jane Doe",
+                                          "Joe Blogs", "Maria Tester",
+                                          "Jill Jillerson"],
+                                  Name_1 = ["John Doe", "Jane Doe",
+                                            "Joe Blogs", "Maria Tester",
+                                            "Jill Jillerson"])
+
+        @test join(name_w_zeros, name_w_zeros, on=:ID, validate=(true, true)) ≅ name_w_zeros2
+
+         # Check for multiple-column merge keys
```
nalimilan: Incorrect alignment.

nickeubank:
any more open issues here?

src/abstractdataframe/join.jl
```diff
@@ -290,6 +298,26 @@ function Base.join(df1::AbstractDataFrame,
 
     joiner = DataFrameJoiner(df1, df2, on)
 
+    # Check merge key validity
+    left_invalid = validate[1] ? any(nonunique(joiner.dfl, joiner.left_on)) : false
```
bkamins: I do not remember if it was discussed but `nonunique(joiner.dfl, joiner.left_on)` and `nonunique(joiner.dfr, joiner.right_on)` could be saved in a variable to avoid making the call again on error.

src/abstractdataframe/join.jl
```diff
@@ -290,6 +298,26 @@ function Base.join(df1::AbstractDataFrame,
...
+    right_invalid = validate[2] ? any(nonunique(joiner.dfr, joiner.right_on)) : false
+
+    if left_invalid && right_invalid
+        first_error_df1 = findfirst(nonunique(joiner.dfl, joiner.left_on))
+        first_error_df2 = findfirst(nonunique(joiner.dfr, joiner.right_on))
+        throw(ArgumentError("Merge key(s) are not unique in either df1 or df2. " *
```
bkamins: Should it rather be `not unique in df1 and df2` as we know that there are duplicates in both `DataFrame`s?

bkamins:
I have left some small comments (but in general PR looks good to merge).

nickeubank:
Rebased and corrected message. Open to suggestion on `nonunique` issue noted above. 

Note this PR now includes a patch for #1424  -- I realized on change we made resulted in modification of the input datasets. In particular, the correction is now at Lines 307-308 (using `hcat` instead of direct column insertion). Tests to ensure datasets not modified added as well.

bkamins:
I am OK to leave `nonunique` as is, so it is good to merge for me. Thanks!

bkamins:
In the future the 307-308 patch would be better to be a separate commit (or separate PR) to keep history or changes.

nickeubank:
@bkamins Oh, ok -- sorry about that!  It came up here (that error broke the tests for this PR) so I just patched, but in the future I'll keep them separate.

nalimilan:
Yes, please move that commit to a different PR, we already have enough changes on our plate here.

test/join.jl
```diff
@@ -578,14 +578,17 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
-
```
nalimilan: Keep this line, and add a break before the `@testset` you added.

test/join.jl
```diff
@@ -578,14 +578,17 @@ module TestJoin
...
     @testset "indicator columns" begin
         outer_indicator = DataFrame(ID = [1, 2, 2, 3, 4],
                                     Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
                                     Job = ["Lawyer", "Doctor", "Florist", missing, "Farmer"],
                                     _merge = ["both", "both", "both", "left_only", "right_only"])
+        pre_join_name = copy(name)
```
nalimilan: I don't understand why this is needed.

nickeubank:
OK, split out in #1434 . Note these tests will fail until that patch is merged.

nalimilan:
Thanks!

PR #1424, opened 2018-06-10
add merge indicator function

### LINKED ISSUE #1412: Feature request for `join`

nickeubank:
Together with #1421 closes #1412 .

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -247,6 +247,11 @@ Join two `DataFrame` objects
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 
+* `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
+   column named `Symbol` for whether a row appeared in only `df1` (`"left_only"`),
+   only `df2` (`"right_only"`) or in both (`"both"`). If `Symbol` is already in use,
+   the column name will be modified if `makeunique=true`.
+
 For the three join operations that may introduce missing values (`:outer`, `:left`,
 and `:right`), all columns of the returned data table will support missing values.
 
@@ -280,7 +285,21 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner, makeunique::Bool=false)
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   indicator::Union{Nothing, Symbol} = nothing)
+    if indicator !== nothing
+        indicator_cols = ["_left", "_right"]
+        for i in 1:2
+            while (haskey(index(df1), Symbol(indicator_cols[i])) ||
+                   haskey(index(df2), Symbol(indicator_cols[i])) ||
+                   Symbol(indicator_cols[i]) == indicator)
+                 indicator_cols[i] *= 'X'
+            end
+         end
+        df1[Symbol(indicator_cols[1])] = trues(nrow(df1))
+        df2[Symbol(indicator_cols[2])] = trues(nrow(df2))
+    end
+
     if kind == :cross
         (on == Symbol[]) || throw(ArgumentError("Cross joins don't use argument 'on'."))
         return crossjoin(df1, df2, makeunique=makeunique)
@@ -291,25 +310,25 @@ function Base.join(df1::AbstractDataFrame,
     joiner = DataFrameJoiner(df1, df2, on)
 
     if kind == :inner
-        compose_joined_table(joiner, kind, update_row_maps!(joiner.dfl_on, joiner.dfr_on,
-                                                            group_rows(joiner.dfr_on),
-                                                            true, false, true, false)...,
-                                                            makeunique=makeunique)
+        joined = compose_joined_table(joiner, kind, update_row_maps!(joiner.dfl_on, joiner.dfr_on,
+                                                                     group_rows(joiner.dfr_on),
+                                                                     true, false, true, false)...,
+                                      makeunique=makeunique)
     elseif kind == :left
-        compose_joined_table(joiner, kind, update_row_maps!(joiner.dfl_on, joiner.dfr_on,
+        joined = compose_joined_table(joiner, kind, update_row_maps!(joiner.dfl_on, joiner.dfr_on,
                                                             group_rows(joiner.dfr_on),
                                                             true, true, true, false)...,
-                                                            makeunique=makeunique)
+                                      makeunique=makeunique)
     elseif kind == :right
-        compose_joined_table(joiner, kind, update_row_maps!(joiner.dfr_on, joiner.dfl_on,
+        joined = compose_joined_table(joiner, kind, update_row_maps!(joiner.dfr_on, joiner.dfl_on,
                                                             group_rows(joiner.dfl_on),
                                                             true, true, true, false)[[3, 4, 1, 2]]...,
-                                                            makeunique=makeunique)
+                                      makeunique=makeunique)
     elseif kind == :outer
-        compose_joined_table(joiner, kind, update_row_maps!(joiner.dfl_on, joiner.dfr_on,
+        joined = compose_joined_table(joiner, kind, update_row_maps!(joiner.dfl_on, joiner.dfr_on,
                                                             group_rows(joiner.dfr_on),
                                                             true, true, true, true)...,
-                                                            makeunique=makeunique)
+                                      makeunique=makeunique)
     elseif kind == :semi
         # hash the right rows
         dfr_on_grp = group_rows(joiner.dfr_on)
@@ -323,7 +342,7 @@ function Base.join(df1::AbstractDataFrame,
                 push!(left_ixs, l_ix)
             end
         end
-        return joiner.dfl[left_ixs, :]
+        joined = joiner.dfl[left_ixs, :]
     elseif kind == :anti
         # hash the right rows
         dfr_on_grp = group_rows(joiner.dfr_on)
@@ -337,10 +356,24 @@ function Base.join(df1::AbstractDataFrame,
                 push!(leftonly_ixs, l_ix)
             end
         end
-        return joiner.dfl[leftonly_ixs, :]
+        joined = joiner.dfl[leftonly_ixs, :]
     else
         throw(ArgumentError("Unknown kind of join requested: $kind"))
     end
+
+    if indicator !== nothing
+        left = joined[Symbol(indicator_cols[1])]
+        right = joined[Symbol(indicator_cols[2])]
+
+        refs = UInt8[coalesce(l, false) + 2 * coalesce(r, false) for (l, r) in zip(left, right)]
+        indicatorcol = CategoricalArray{String,1}(refs, CategoricalPool{String,UInt8}(["left_only", "right_only", "both"]))
+        joined = hcat(joined, DataFrame(indicator => indicatorcol), makeunique=makeunique)
+
+        delete!(joined, Symbol(indicator_cols[1]))
+        delete!(joined, Symbol(indicator_cols[2]))
+    end
+
+    return joined
 end
 
 function crossjoin(df1::AbstractDataFrame, df2::AbstractDataFrame; makeunique::Bool=false)
--- a/test/join.jl
+++ b/test/join.jl
@@ -578,4 +578,29 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
+
+    @testset "indicator columns" begin
+        outer_indicator = DataFrame(ID = [1, 2, 2, 3, 4],
+                                    Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
+                                    Job = ["Lawyer", "Doctor", "Florist", missing, "Farmer"],
+                                    _merge = ["both", "both", "both", "left_only", "right_only"])
+        @test join(name, job, on = :ID, kind = :outer, indicator=:_merge,
+                   makeunique=true) ≅ outer_indicator
+
+        # Works with conflicting names
+        name2 = DataFrame(ID = [1, 2, 3], Name = ["John Doe", "Jane Doe", "Joe Blogs"],
+                         _left = [1, 1, 1])
+        job2 = DataFrame(ID = [1, 2, 2, 4], Job = ["Lawyer", "Doctor", "Florist", "Farmer"],
+                        _left = [1, 1, 1, 1])
+
+        outer_indicator = DataFrame(ID = [1, 2, 2, 3, 4],
+                                    Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
+                                    _left = [1, 1, 1, 1, missing],
+                                    Job = ["Lawyer", "Doctor", "Florist", missing, "Farmer"],
+                                    _left_1 = [1, 1, 1, missing, 1],
+                                    _left_2 = ["both", "both", "both", "left_only", "right_only"])
+
+        @test join(name2, job2, on = :ID, kind = :outer, indicator=:_left,
+                   makeunique=true) ≅ outer_indicator
+    end
 end


src/abstractdataframe/join.jl
```diff
@@ -212,6 +212,7 @@ function update_row_maps!(left_table::AbstractDataFrame,
     return to_bimap(left_ixs), to_bimap(leftonly_ixs), to_bimap(right_ixs), to_bimap(rightonly_ixs)
 end
 
+
```
nalimilan: Unrelated?

src/abstractdataframe/join.jl
```diff
@@ -247,6 +248,13 @@ Join two `DataFrame` objects
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 
+* `indicator` : Default: `false`. If `true`, adds categorical indicator
```
nalimilan: I thought we agreed at #1412 that it would be better to have two separate keywords allowing the user the specify the name of the columns to create, with a default to `nothing`.

nickeubank: I thought we agreed we didn't want a column of booleans, but I took the thumbs up on my last post suggesting this categorical solution (same as pandas and stata) as confirmation that made sense. I can revert them to separate bit columns, but I feel like users will usually want the interaction of the two, jumping to a categorical with all cases seemed logical to me.

src/abstractdataframe/join.jl
nickeubank: not sure how that snuck in. Removed for next commit.

src/abstractdataframe/join.jl
nalimilan: I'm not sure what's the most practical for users. Do you have examples of operations people may want to do with these columns? It sounds simpler to me to do `df[:left]` than `df[:merge] .== "left_only"` (and it's also more efficient and takes less memory).

nickeubank: So `df[:left]` and `df[:merge] .== "left_only"` are different operations. The first will get you all observations that were in `df1`. The second will get you observations that were in `df1` but not in `df2`. 

Personally, I use this a lot for doing things like `@assert nrow(df[df[:_merge] .== "left_only"]) == 0` to validate that I successfully found a match for all keys in the left dataset in the right dataset.

nalimilan: OK. `nrow(df[df[:_merge] .== "left_only"]) == 0` can be written more simply as `!any(df[:_merge] .== "left_only"])` or `!("left_only" in df[:_merge])`. With two Boolean columns it would be `!any(df[:left] .& .!df[:right])`. Personally I find the latter a bit cleaner (no need to remember whether it"s `"left_only"`, `"left only"`, `"only left"` or any other variation), but while it should be faster it's also slightly more convoluted.

But since you say you use it as a validation check, aren't there better ways of achieving this? For example, doing an inner join and then checking `nrow(joined) == nrow(df1)`? Or maybe we should add a keyword argument to do that?

nickeubank: > But since you say you use it as a validation check, aren't there better ways of achieving this? For example, doing an inner join and then checking nrow(joined) == nrow(df1)? Or maybe we should add a keyword argument to do that?

The indicator column gives you both sides immediately, and only requires one merge (nice computationally). For example, say I think the data should have all keys match (so `outer` should be same as `inner`), but I want to be sure everything works, so I do an `outer` join. I could then do a `right` join (to check for unmatched `df2` keys) and a `left` join (to check for unmatched `df1` keys) and compare all of those. But that's three merges. Here it's one column from one merge. 

It's also the case that I often leave those simple tests in place in my code for defensive purposes, but when a problem comes up I want to query those bad matches, for which `nrow` is not enough. So if my test fails, I'd want to look at the `df[:_merge] .== "left_only"]`. 

And yes, I'm of the view that `df[:_merge] .== "left_only"]` is more readable / like english than `df[df[:_left] .& .!df[:_right]` to me.

nalimilan: > The indicator column gives you both sides immediately, and only requires one merge (nice computationally). For example, say I think the data should have all keys match (so outer should be same as inner), but I want to be sure everything works, so I do an outer join. I could then do a right join (to check for unmatched df2 keys) and a left join (to check for unmatched df1 keys) and compare all of those. But that's three merges. Here it's one column from one merge.

My example also only requires one merge. If you do an inner join and get as many columns as `df1`, that means all rows matched.

Anyway I'm not saying we shouldn't have the ability to create that new column, but maybe we also want a convenience argument to ensure all rows matched?

@bkamins Opinions?

nickeubank: > My example also only requires one merge. If you do an inner join and get as many columns as df1, that means all rows matched.

That's true, sorry -- was thinking ahead to the case of wanting to see which observations failed if it didn't work.

bkamins: As I have stated in the issue. I would prefer two columns of boolean indicators as this will be a faster and more memory efficient operation.

But @nickeubank knows his usecase best, and if one column is much more convenient for him I would be OK with this. The only question is if we want to use `String` or `Symbol` (I guess `Symbol` like `:left_only`, `:right_only` and `:both` should be a bit faster and less memory hungry as `Symbol`s are interned - this would have to be tested).

Finally I have a strong preference for a calling convention. Instead of `indicator` being `Bool` I would like it to be `Union{Nothing, Symbol}` and `nothing` being default. In such a situation if some user wants `indicator` column she would pass the column name that suits her best. Then if duplicate column name is encountered it would be handled following value of `makeunique`.

(the same would be recommended if we went for two columns of `Bool`)

nalimilan: > But @nickeubank knows his usecase best, and if one column is much more convenient for him I would be OK with this. The only question is if we want to use String or Symbol (I guess Symbol like :left_only, :right_only and :both should be a bit faster and less memory hungry as Symbols are interned - this would have to be tested).

We should definitely use a `CategoricalArray` for that, it would be much more efficient (with a `UInt8` reference type it would use 8× less memory than a 64-bit pointer). Then a `String` (actually`CategoricalString`) element type sounds the most natural.

BTW, `"right"` and `"left"` sound better than `"right_only"` and `"left_only"`, don't they?

> Finally I have a strong preference for a calling convention. Instead of indicator being Bool I would like it to be Union{Nothing, Symbol} and nothing being default. In such a situation if some user wants indicator column she would pass the column name that suits her best. Then if duplicate column name is encountered it would be handled following value of makeunique.

Agreed.

bkamins: `CategoricalArray` is indeed most natural

nickeubank: > Finally I have a strong preference for a calling convention. Instead of indicator being Bool I would like it to be Union{Nothing, Symbol} and nothing being default. In such a situation if some user wants indicator column she would pass the column name that suits her best. Then if duplicate column name is encountered it would be handled following value of makeunique.

Happy to do that. Will update. 

> BTW, "right" and "left" sound better than "right_only" and "left_only", don't they?

I think `right_only` is more descriptively accurate. All observations in `both` are in the right dataset, and also in the left dataset. The point is the observations labeled `right_only` are, in fact, only in the right dataset. 

Also one more thought on an earlier point: the other limitation to checking using something like  `nrow(joined) == nrow(df1)` is that it only works for a 1:1 merge. If the merge is 1:many, that test becomes uninformative. :/

nickeubank:
@bkamins that closer to what you wanted for argument syntax?

src/abstractdataframe/join.jl
```diff
@@ -280,7 +286,16 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner, makeunique::Bool=false)
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   indicator::Union{Nothing, Symbol} = nothing)
+
+    if !( indicator == nothing )
```
bkamins: typically comparison to `nothing` is done using `===` AFAIK.

src/abstractdataframe/join.jl
```diff
@@ -337,10 +352,27 @@ function Base.join(df1::AbstractDataFrame,
...
+    if !(indicator == nothing)
+        left = joined[indicator_cols[1]]
+        left = Int.(coalesce.(left, false))
+
+        right = joined[indicator_cols[2]]
+        right = Int.(coalesce.(right, false)) * 2
+
+        _merge = DataFrame(indicator => categorical(left + right))
+        _merge[indicator] = recode(_merge[indicator], 1 => "left_only", 2 => "right_only", 3 => "both")
+        joined = hcat(joined, _merge, makeunique=true)
```
bkamins: This is inconsistent with the documentation, where you state that you use the value of `makeunique` passed by the user. Here you override it by setting it to `true`. I thought you want to use what the user has passed (this is probably not relevant in interactive use but is relevant when used programmatically).

src/abstractdataframe/join.jl
```diff
@@ -337,10 +352,27 @@ function Base.join(df1::AbstractDataFrame,
                 push!(leftonly_ixs, l_ix)
             end
         end
-        return joiner.dfl[leftonly_ixs, :]
+        joined = joiner.dfl[leftonly_ixs, :]
     else
         throw(ArgumentError("Unknown kind of join requested: $kind"))
     end
+
+    if !(indicator == nothing)
```
bkamins: use of `indicator_cols` can cause problems when `_left` or `_right` name was used in the "other" table. Then with `makeunique=true` they would get renamed and you would:

1) remove the wrong column below
2) create `left` and `right` using a wrong column

A quick fix would be to use column number not column name.

src/abstractdataframe/join.jl
bkamins: also spaces at `(` and `)` are not needed I guess

bkamins:
Thank you 👍 . Yes - this is what I thought of. I have left some comments regarding corner cases of the functionality.

src/abstractdataframe/join.jl
bkamins: Actually you have to create column names in a better way if I am not mistaken anyway. So this code can stay, but column creation above should be fixed. The reason is that with `makeunique=false` if there is a duplicate column `_left` or `_right` in the other `DataFrame` join will fail as it will detect a duplicate.

(if I have missed something and you handle this correctly then please let me know, it would be best if an appropriate test covering all corner cases with offending column names were created)

src/abstractdataframe/join.jl
nickeubank: @bkamins is there a PEP8 for Julia yet about this kind of spacing stuff?

src/abstractdataframe/join.jl
nickeubank: OH! I didn't mean to suggest I'd obey the user's argument passed to `makeunique` -- I meant that renaming takes place using `makeunique` norms (i.e. the way it updates column names when a conflict occurs is using `makeunique`). Will fix docs.

src/abstractdataframe/join.jl
nickeubank: I think I'm ok with this actually -- since when I `hcat` in `_left` and `_right`, I have `makeunique=true`, there's no risk of name collisions. Then on line 296 I pull the resulting column names out so if a conflict and rename occurred, I'd get the updated column name (`_left_x` or whatever). It's those pulled names I used in the later delete. 

But the obvious answer to check when I fill out the tests, which I was just holding off till we agreed on syntax. :) Will work on those shortly.

src/abstractdataframe/join.jl
bkamins: For both comments above this is just style. But if you grep Julia sources in Base you will see that this is a predominant style (to use `===` to compare to `nothing` and not to put spaces after `(` and before `)`). Anyway this is a minor issue.

src/abstractdataframe/join.jl
bkamins: Actually I agreed with the docs. Why it is important in my opinion to be consistent with what user passes in `makeunique` is the following scenario.

Assume we are using `indicator` keyword argument in a program (not interactive use). A program then wants to access the column named after `indicator`. If you always silently rename the column you may run into problems trying to access the wrong column. It is better to throw an error if `makeunique` is set to `false` by the user in such a case. The user can always pass `makeunique=ture` if she wants column to be renamed.

nickeubank: Right, super fair. Will adapt in that direction then!

src/abstractdataframe/join.jl
nickeubank: OH! now I see, thanks for explaining. Will use column numbers in next commit.

bkamins: Even with column numbers you still need careful column names as temporary names could clash on join with normal names in the other table (and cause an error if `makeunique=false` or cause an unnecessary rename of columns if `makeunique=true`).

The simplest that comes to my mind is to create a `Set` of existing column names in both data frames and check if `_left` is in it. If yes add `X` at the end to get `_leftX` and so on till you get a unique name. The reason I recommend to add `X` is that we know that when DataFrames.jl tries to make unique column names it adds `_{number}` suffix which will not clash with `X` character (so you can use whatever you like as long as it is not a number 😄). The same with `_right`.

nickeubank: @bkamins Oh, I thought by "column numbers" you meant positional column numbers. Those are pretty safe, no? In the final result, can't I count on the first helper column (currently `_left`) being in position `ncol(df1) + 1` (where `df1` is initial input dataframe -- just `ncol(df1)` after I add the column), and the second (`_right`) in `ncol(joined) - 2`?

nickeubank:
@bkamins kinda fixed, but see #1425 ...

src/abstractdataframe/join.jl
bkamins: Now it looks good. @nalimilan - can you please have a look?

src/abstractdataframe/join.jl
```diff
@@ -280,7 +286,22 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner, makeunique::Bool=false)
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   indicator::Union{Nothing, Symbol} = nothing)
+
+    if !(indicator === nothing)
```
nalimilan: Use `!==`. Also, no empty line above.

src/abstractdataframe/join.jl
```diff
@@ -247,6 +247,12 @@ Join two `DataFrame` objects
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 
+* `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
+   column for whether a row appeared in only `df1` ("left_only"),
```
nalimilan: Should put backticks around string values.

src/abstractdataframe/join.jl
```diff
@@ -247,6 +247,12 @@ Join two `DataFrame` objects
...
+   only `df2` ("right_only") or in both ("both").
+   By default column will use the passed `Symbol` as its name unless
```
nalimilan: I find this a bit confusing. Maybe better say "add categorical indicator column with the specified name" in the previous sentence. Then you can explain what happens in case of conflict: an error is raised, unless `makeunique` is `true`.

src/abstractdataframe/join.jl
```diff
@@ -280,7 +286,22 @@ join(name, job2, on = :ID => :identifier)
...
+    if !(indicator === nothing)
+        names_in_use = Set([names(df1); names(df2); [indicator]])
+        indicator_cols = ["_left", "_right"]
+        for i in 1:2
+            while Symbol(indicator_cols[i]) ∈ names_in_use
+                 indicator_cols[i] = indicator_cols[i] * "X"
+            end
+         end
+        indicator_cols = Symbol.(indicator_cols)
+        df1 = hcat(df1, DataFrame(Dict(indicator_cols[1] => [true for i in 1:nrow(df1)] )))
```
nalimilan: Wouldn't this be enough (and use less memory)?
```julia
df1[indicator_cols[1]] = trues(nrow(df1))
```
Also better call `Symbol` when indexing rather than replacing `indicator_cols` (type instability!).

src/abstractdataframe/join.jl
```diff
@@ -280,7 +286,22 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner, makeunique::Bool=false)
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   indicator::Union{Nothing, Symbol} = nothing)
+
+    if !(indicator === nothing)
+        names_in_use = Set([names(df1); names(df2); [indicator]])
```
nalimilan: The `DataFrame` type already includes a dictionary of column names. Just do `c =  Symbol(indicator_cols[i]); haskey(index(df1), c) || haskey(index(df2), c) || c == indicator` in the loop.

src/abstractdataframe/join.jl
```diff
@@ -280,7 +286,22 @@ join(name, job2, on = :ID => :identifier)
...
+        indicator_cols = ["_left", "_right"]
+        for i in 1:2
+            while Symbol(indicator_cols[i]) ∈ names_in_use
+                 indicator_cols[i] = indicator_cols[i] * "X"
```
nalimilan: `indicator_cols[i] *= 'X'`

src/abstractdataframe/join.jl
```diff
@@ -337,10 +358,27 @@ function Base.join(df1::AbstractDataFrame,
...
+
+    if !(indicator == nothing)
+        left = joined[indicator_cols[1]]
+        left = Int.(coalesce.(left, false))
+
+        right = joined[indicator_cols[2]]
+        right = Int.(coalesce.(right, false)) * 2
+
+        _merge = DataFrame(indicator => categorical(left + right))
+        _merge[indicator] = recode(_merge[indicator], 1 => "left_only", 2 => "right_only", 3 => "both")
```
nalimilan: You can make all of this more efficient:
- instead of allocating the result in `left` and `right`, loop over both columns at the same time in a comprehension and allocate directly the final vector `refs`, with a `UInt8` eltype
- use the low-level `CategoricalArray` constructor `CategoricalArray{String,1}(refs, CategoricalPool{String,UInt8}(["left_only", "right_only", "both"]))`
- only create the `DataFrame` at that point

test/join.jl
```diff
@@ -578,4 +578,30 @@ module TestJoin
         @test eltypes(join(l, r, on=:b, kind=:outer, makeunique=true)) ==
             [Union{Int, Missing}, CS, Union{Int, Missing}]
     end
+
+    @testset "indicator columns" begin
+        outer_indicator = DataFrame(ID = [1, 2, 2, 3, 4],
+                                Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
+                                Job = ["Lawyer", "Doctor", "Florist", missing, "Farmer"],
+                                _merge = categorical(["both", "both", "both", "left_only", "right_only"])
+                                );
```
nalimilan: Close this on the previous line, no semicolon, and align with first argument (not with parenthesis). Same below.

test/join.jl
```diff
@@ -578,4 +578,30 @@ module TestJoin
...
+        job2 = DataFrame(ID = Union{Int, Missing}[1, 2, 2, 4],
+                        Job = Union{String, Missing}["Lawyer", "Doctor", "Florist", "Farmer"],
+                        _left = [1, 1, 1, 1])
+
+        outer_indicator = DataFrame(ID = [1, 2, 2, 3, 4],
+                                    Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
+                                    _left = [1, 1, 1, 1, missing],
+                                    Job = ["Lawyer", "Doctor", "Florist", missing, "Farmer"],
+                                    _left_1 = [1, 1, 1, missing, 1],
+                                    _left_2 = categorical(["both", "both", "both", "left_only", "right_only"])
```
nalimilan: No need to make this categorical.

src/abstractdataframe/join.jl
nalimilan: Looks good now, but the only way to be sure is to have tests. AFAICT they currently cover this corner case, right?

src/abstractdataframe/join.jl
nickeubank: So, re: #1425 , the problem is at the moment this doesn't raise an error. It ignores the command...

nickeubank:
Thanks all! I think I integrated all those. The only odd one is the question of how to document behavior when there's a naming conflict, since I'm not really clear what's supposed to happen / whether the current behavior is the expected behavior (see #1425 ).

src/abstractdataframe/join.jl
nickeubank: I'm afraid I can't get this to work -- when I try the low-level construction I get: 

```
CategoricalArray{String,1,UInt8}(refs, CategoricalPool{String,UInt8}(["left_only", "right_only", "both"]))
ERROR: MethodError: no method matching CategoricalArray{String,1,UInt8,V,C,U} where U where C where V(::Array{Int64,1}, ::CategoricalPool{String,UInt8,CategoricalString{UInt8}})
Closest candidates are:
  CategoricalArray{String,1,UInt8,V,C,U} where U where C where V(::AbstractArray; ordered) where {T, N, R} at /Users/Nick/.julia/packages/CategoricalArrays/AxkM/src/array.jl:170
  CategoricalArray{String,1,UInt8,V,C,U} where U where C where V(::Any) where T<:AbstractArray at abstractarray.jl:22
```

And I can't seem to find documentation for the kind of construction you're suggesting...

src/abstractdataframe/join.jl
```diff
@@ -247,6 +247,11 @@ Join two `DataFrame` objects
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 
+* `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
+   column named `Symbol` for whether a row appeared in only `df1` (`left_only`),
```
nalimilan: Should be `` `"left_only"` `` and so on.

src/abstractdataframe/join.jl
```diff
@@ -280,7 +285,21 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner, makeunique::Bool=false)
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   indicator::Union{Nothing, Symbol} = nothing)
+
```
nalimilan: Remove empty line.

src/abstractdataframe/join.jl
```diff
@@ -247,6 +247,11 @@ Join two `DataFrame` objects
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 
+* `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
+   column named `Symbol` for whether a row appeared in only `df1` (`left_only`),
+   only `df2` (`right_only`) or in both (`both`). If `Symbol` is already in use,
+   the column name will be modified if `makeunique=true`.
```
nalimilan: So what happens when `makeunique=false`? Can we check that explicitly and throw an error if the name is already used?

src/abstractdataframe/join.jl
nalimilan: Yeah it's hard to find the right combination of parameters which works (these are low-level, so not for general consumption and not really documented). But have you tried the exact one I suggested? In my tests, it worked.

nalimilan: Looking at the code in the PR, the problem is that you should do `refs = UInt8[...]`

src/abstractdataframe/join.jl
```diff
@@ -337,10 +356,28 @@ function Base.join(df1::AbstractDataFrame,
...
         end
-        return joiner.dfl[leftonly_ixs, :]
+        joined = joiner.dfl[leftonly_ixs, :]
     else
         throw(ArgumentError("Unknown kind of join requested: $kind"))
     end
+
+    if indicator !== nothing
+        left = joined[Symbol(indicator_cols[1])]
+        left = Int.(coalesce.(left, false))
```
nalimilan: This allocates two copies of the column, and introduces a type instability by changing the type of `left`. You don't actually need to do this here: just call `coalesce(left[i], false)` inside the comprehension (`Int` isn't needed if you speficy `UInt8[...]` before the comprehension).

src/abstractdataframe/join.jl
nickeubank: if `makeunique=false`, it ignores it and just acts like `makeunique=true`. Again, see #1425

src/abstractdataframe/join.jl
nickeubank: OK, but I don't think that's a type instability (or what is generally meant by "type instability" in Julia). Isn't a "type instability" a situation in which the type of a variable depends on the values of data. When that happens, it causes problems for the compiler because the compiler is only conditioning its code on function input types, not input values, so it can't make a clean type inference. e.g.

```
function x(a::Integer)
    if a % 2  == 0
        return a
    else
       return "a is odd"
end
```
Is type unstable because if a is even , the function returns an integer, if its odd it return a string. And the compiler doesn't condition on the *value* of a, just it's type (here, `Integer`). 

Here, by contrast, the type of the vector to which `left` is bound changes, but in an entirely transparent way. The compiler should have no problems with this. 

Also, I think you may be underestimating the compiler -- it knows not to double allocate: 

```
julia> function test_basic(left)
           return UInt8.(coalesce.(left, false))
           end
julia> function test_fancy(left)
           return UInt8[coalesce(left[i], false) for i in 1:length(left)]
           end
julia> left = rand([true, missing], 10000)

julia> @btime test_fancy(left)
  77.040 μs (1 allocation: 9.94 KiB)

julia> @btime test_basic(left)
  80.942 μs (1 allocation: 9.94 KiB)
```

nickeubank:
@nalimilan github won't let me reply directly to this, so putting here:

> Yeah it's hard to find the right combination of parameters which works (these are low-level, so not for general consumption and not really documented). But have you tried the exact one I suggested? In my tests, it worked.

No, it didn't, but I'm using the last releases of CategoricalArray, not the `master` branch. :/

src/abstractdataframe/join.jl
nickeubank:
```
julia> a = DataFrame(col = [1, 2, 3])
3×1 DataFrame
│ Row │ col │
├─────┼─────┤
│ 1   │ 1   │
│ 2   │ 2   │
│ 3   │ 3   │

julia> b = DataFrame(col = [1, 2, 3])
3×1 DataFrame
│ Row │ col │
├─────┼─────┤
│ 1   │ 1   │
│ 2   │ 2   │
│ 3   │ 3   │

julia> hcat(a, b, makeunique=false)
3×2 DataFrame
│ Row │ col │ col_1 │
├─────┼─────┼───────┤
│ 1   │ 1   │ 1     │
│ 2   │ 2   │ 2     │
│ 3   │ 3   │ 3     │
```

src/abstractdataframe/join.jl
bkamins: You are over-estimating the compiler here I guess. Here is a short example that this is type unstable:

```
julia> using DataFrames

julia> x = DataFrame(rand(3,4))
3×4 DataFrames.DataFrame
│ 1   │ 0.825846 │ 0.719652 │ 0.968676 │ 0.892195 │
│ 2   │ 0.46385  │ 0.51642  │ 0.229933 │ 0.675513 │
│ 3   │ 0.790047 │ 0.811756 │ 0.635098 │ 0.836908 │

julia> function f(df)
       left = df[1]
       left = floor.(Int, left)
       end
f (generic function with 2 methods)

julia> f(x)
3-element Array{Int64,1}:
 0
 0
 0

julia> @code_warntype f(x)
Variables:
  #self# <optimized out>
  df::DataFrames.DataFrame
  left::Any
  selected_column <optimized out>

Body:
  begin
      left::Any = (Base.arrayref)((Core.getfield)(df::DataFrames.DataFrame, :columns)::Array{T,1} where T, 1)::Any # line 3:
      SSAValue(0) = (Base.broadcast)(Main.floor, Main.Int, left::Any)::Any
      left::Any = SSAValue(0)
      return SSAValue(0)
  end::Any
```

Sometimes compiler can optimize out such code, but it is not guaranteed.

In general type-stability is a situation when type of a value bound to a variable is unpredictable or may change (even if predictably) I would say.

src/abstractdataframe/join.jl
nalimilan: OK. But I guess a deprecation warning is printed? If not, better throw an error manually.

src/abstractdataframe/join.jl
```diff
@@ -337,10 +356,27 @@ function Base.join(df1::AbstractDataFrame,
...
+
+    if indicator !== nothing
+        left = joined[Symbol(indicator_cols[1])]
+        left_ints = UInt8.(coalesce.(left, false))
+
+        right = joined[Symbol(indicator_cols[2])]
+        right_ints = UInt8.(coalesce.(right, false)) * 2
+
+        _merge = DataFrame(indicator => categorical(left_ints + right_ints))
+        _merge[indicator] = recode(_merge[indicator], 1 => "left_only", 2 => "right_only", 3 => "both")
```
nalimilan: Here's what I suggested:
```julia
refs = UInt8[coalesce(l, false) + 2 * coalesce(r, false) for (l, r) in zip(left, right)]
indicatorcol = CategoricalArray{String,1}(refs, CategoricalPool{String,UInt8}(["left_only", "right_only", "both"]))
joined = hcat(joined, DataFrame(indicator => indicatorcol), makeunique=makeunique)
```

src/abstractdataframe/join.jl
```diff
@@ -280,7 +285,21 @@ join(name, job2, on = :ID => :identifier)
 function Base.join(df1::AbstractDataFrame,
                    df2::AbstractDataFrame;
                    on::Union{<:OnType, AbstractVector{<:OnType}} = Symbol[],
-                   kind::Symbol = :inner, makeunique::Bool=false)
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   indicator::Union{Nothing, Symbol} = nothing)
+    if indicator !== nothing
+        indicator_cols = ["_left", "_right"]
+        for i in 1:2
+            while (haskey(index(df1), Symbol(indicator_cols[i])) ||
```
nalimilan: What was the problem with `c = Symbol(indicator_cols[i])`? It was more readable.

BTW, you can move `df1[Symbol(indicator_cols[1])] = trues(nrow(df1))` and `df2[Symbol(indicator_cols[2])] = trues(nrow(df2))` inside the `for`.

nalimilan:
> No, it didn't, but I'm using the last releases of CategoricalArray, not the master branch. :/

That shouldn't make any difference.

src/abstractdataframe/join.jl
bkamins: The warning is printed and an error will be thrown after deprecation period.

nalimilan: OK, let's say that's fine then.

src/abstractdataframe/join.jl
nickeubank: The `c = Symbol(indicator_cols[i])` had to be written twice -- once before the loop and once inside the loop again (or the change didn't propagate). DIdn't like have the code duplicated.

nickeubank: Putting them inside the `for` requires some tricks to make `df1` and `df2` indexable. I think more readable with just two explicit lines.

nickeubank: e.g. had to be:

```
c = Symbol(indicator_cols[i])
while (haskey(index(df1), c) || haskey(index(df2), c) || Symbol(indicator_cols[i]) == c)
    indicator_cols[i] *= 'X'
    c = Symbol(indicator_cols[i])
end
```

or something like that.

src/abstractdataframe/join.jl
nickeubank: FYI, re: https://discourse.julialang.org/t/why-is-changing-type-of-variable-problematic/11722/17 : Looks like this is getting much better in 0.7. 

Re your example @bkamins , I think that still has problems in 0.7, but that's because of the general issues with type inference and `DataFrames`. Doing the same with regular vectors seems fine:

```
function f2(m)
              onecol = m[:,1]
              onecol = floor.(Int, onecol)
end

julia> m = rand(3,4)
3×4 Array{Float64,2}:
 0.841133  0.613176  0.145906   0.526924
 0.388555  0.709457  0.0679149  0.633461
 0.858938  0.835699  0.599265   0.814602

julia> f2(m)
3-element Array{Int64,1}:
 0
 0
 0

julia> @code_warntype f2(m)
Body::Array{Int64,1}
2 1 ── %1   = Base.arraysize(%%m, 1)::Int64 getindex
  │    %2   = Base.arraysize(%%m, 2)::Int64  to_indices
  │    %3   = Base.slt_int(%1, 0)::Bool       axes
  │    %4   = Base.ifelse(%3, 0, %1)::Int64    map
  │    %5   = new(Base.OneTo{Int64}, %4)::Base.OneTo{Int64}
  │    %6   = Base.slt_int(%2, 0)::Bool          Type
  │           Base.ifelse(%6, 0, %2)              max
  │    %8   = new(Base.Slice{Base.OneTo{Int64}}, %5)::Base.Slice{Base.OneTo{Int64}}
  └───        goto 6 if not true             _getindex
  2 ── %10  = Core.tuple(%8, 1)::Tuple{Base.Slice{B

[lots more but no warnings...]
```

nalimilan: Good to know. However in many situations it's clearer to avoid using the same name for values of different types.

src/abstractdataframe/join.jl
nickeubank: so just remembered why I was using `hcat` -- doing it the way proposed leads to modification of the input dataset. Fixing in #1421

PR #1429, opened 2018-06-15
A second attempt at DataFrames Metadata

pdeffebach:
This is another attempt at adding metadata support to DataFrames, modeled after #1413. The ultimate api for metadata between this approach and that one is the same. Users will be able to create arbitrary metadata fields, with a special case for a `label` that hopefully other packages, like plotting, will be exposed to. 

However the implementation of metadata differs substantially. A `MetaData` type is just a wrapper for a `Dict` from `Symbol` to `Vector{String}`. The goal is for `df.metadata` to behave exactly like `df.columns`. This means that the `index.jl` code does not need to be changed at all. `getindex` and `setindex` will work the same for `df.metadata` as for `df.columns`, enabling code that looks like this: 

```
function insert(df, col_ind, item)
    insert!(index(df), col_ind, name)
    insert!(columns(df), col_ind, item)
    insert!(metadata(df), col_ind) # the only change that had to be made to the insert method. 
end
```

I have deliberately touched only a small amount of code for this PR. 

* Added a new file `src/other/metadata/jl` which defines basic operations like `getindex`, `setindex` and `append` for metadata. 
* Added *only a single new constructor* to `dataframes`, such that a new dataframe always creates a dataframe with empty metadata. 
    * This is probably not desirable, but new constructors can be added later. 
* Implemented `getindex`, `setindex` etc. such that any subsetting, adding, and merging will work and preserve metadata. 
* Added *only three functions* exposed to the user. `addlabel!`, `showlabel` and `showlabels`. While adding arbitrary metadata fields (`:unit` etc) is feasible under the current system, I didn't want to complicate the api while we sort out how the interface in general might work. 
* Metadata is *only* strings, and an empty metadata is just an empty string `""`. When a new columns gets added to the dataframe, I just push `""` onto the end of each vector in the `metadata` dictionary. 
* global metadata, i.e. metadata that is tied to a dataframe as a whole and not just a column is not supported, because this will presumably be easy to implement in the future. 

This just a stab at one implementation, and if people decide metadata should be implemented differently, that's fine and there can be another PR for another method.

Appreciate any feedback, thanks.

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -27,7 +27,9 @@ julia> DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
-It is also possible to construct a `DataFrame` in stages:
+### Constructing Column by Column
+
+It is also possible to construct a `DataFrame` one column at a time.
 
 ```jldoctest dataframe
 julia> df = DataFrame()
@@ -63,8 +65,8 @@ julia> df
 
 ```
 
-The `DataFrame` we build in this way has 8 rows and 2 columns. You can check this using the
-`size` function:
+The `DataFrame` we build in this way has 8 rows and 2 columns.
+You can check this using the `size` function:
 
 ```jldoctest dataframe
 julia> size(df, 1) == 8
@@ -77,6 +79,46 @@ julia> size(df) == (8, 2)
 true
 
 ```
+
+### Constructing Row by Row
+
+It is also possible to construct a `DataFrame` row by row.
+
+First a `DataFrame` with empty columns is constructed:
+
+```jldoctest dataframe
+julia> df = DataFrame(A = Int[], B = String[])
+0×2 DataFrames.DataFrame
+```
+
+Rows can then be added as `Vector`s, where the row order matches the columns order:
+
+```jldoctest dataframe
+julia> push!(df, [1, "M"])
+1×2 DataFrames.DataFrame
+│ Row │ A │ B │
+├─────┼───┼───┤
+│ 1   │ 1 │ M │
+```
+
+Rows can also be added as `Dict`s, where the dictionary keys match the column names:
+
+```jldoctest dataframe
+julia> push!(df, Dict(:B => "F", :A => 2))
+2×2 DataFrames.DataFrame
+│ Row │ A │ B │
+├─────┼───┼───┤
+│ 1   │ 1 │ M │
+│ 2   │ 2 │ F │
+```
+
+Note that constructing a `DataFrame` row by row is significantly less performant than
+constructing it all at once, or column by column. For many use-cases this will not matter,
+but for very large `DataFrame`s  this may be a consideration.
+
+## Working with Data Frames
+
+### Taking a Subset
 
 We can also look at small subsets of the data in a couple of different ways:
 
@@ -113,6 +155,8 @@ julia> df[1:3, :]
 
 ```
 
+### Summarizing with `describe`
+
 Having seen what some of the rows look like, we can try to summarize the entire data set using `describe`:
 
 ```jldoctest dataframe
@@ -136,6 +180,8 @@ julia> var(df[:A]) ==  var(df[1]) == 6.0
 true
 
 ```
+
+### Column-Wise Operations
 
 We can also apply a function to each column of a `DataFrame` with the `colwise` function. For example:
 


src/dataframe/dataframe.jl
```diff
@@ -749,6 +765,10 @@ merge!(df, df2)  # column z is added, column id is overwritten
 """
 function Base.merge!(df::DataFrame, others::AbstractDataFrame...)
     for other in others
+        notinother = setdiff(names(other), names(df))
```
nalimilan: It would probably be cleaner to define `merge!(::MetaData, ::MetaData...)`.

src/dataframe/dataframe.jl
```diff
@@ -1094,9 +1114,56 @@ function permutecols!(df::DataFrame, p::AbstractVector)
...
+
+
+##############################################################################
+##
+## Set and Get MetaData 
+##
+##############################################################################
+
+"""
+    addlabel!(df::DataFrame, var::Symbol, label::String)
```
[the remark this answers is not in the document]
pdeffebach: Is this just for now, or do you envision this in the future? Because an idiomatic way to set labels, particularly with chaining in mind, first argument is a dataframe, returns a dataframe, seems important.

src/other/metadata.jl
```diff
@@ -0,0 +1,125 @@
...
+	nothing
+end
+
+function Base.permute(x::MetaData, p::AbstractVector)
+	new_metadata = copy(x)
+	permute!(new_metadata, p)
+end
+
+function newfield!(x::MetaData, ncol::Int, field::Symbol,)
+	x.columndata[field] = ["" for i in 1:ncol]
```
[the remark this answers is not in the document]
pdeffebach: added. `Vector{Any}([nothing for i in 1:N])` this way the user can include arbitrary things. However I am kind of worried that the ability to include arbitrary objects in metadata will cause people to abuse metadata to make it hold actual (non-meta) data.

src/other/metadata.jl
```diff
@@ -0,0 +1,125 @@
+# Defining behavior for DataFrames metadata
+abstract type AbstractMetaData end
+
+mutable struct MetaData <: AbstractMetaData
+	columndata::Dict{Symbol, Vector{String}}
```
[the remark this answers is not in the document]
pdeffebach: fixed my sublime settings.

src/dataframe/dataframe.jl
pdeffebach: I was trying to make all the code in `metadata.jl` not know anything about the dataframe attached to it. Since this code is really about working with the distinct names of both dataframes, i put it here.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
pdeffebach: Another thing with this approach is that `metadata` doesn't know anything about the symbols and the `colindex` of the dataframe. So any function would have to include the `dataframe` in it, right?

src/dataframe/dataframe.jl
```diff
@@ -709,6 +724,7 @@ function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::S
     end
     insert!(index(df), col_ind, name)
     insert!(columns(df), col_ind, item)
+    insert!(metadata(df), col_ind)
```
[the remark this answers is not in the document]
pdeffebach: Done. And I think I removed all references to making things "String". 

Though thinking about the broader uses of labels, like with `@df` in statsplots, there is value in at least a more tightly controlled `:label` field, which must be a string. This was other packages can interface with `:label` and know what they are getting.

src/dataframe/dataframe.jl
```diff
@@ -292,7 +302,8 @@ end
 function Base.getindex(df::DataFrame, row_inds::AbstractVector, col_inds::AbstractVector)
     selected_columns = index(df)[col_inds]
     new_columns = Any[dv[row_inds] for dv in columns(df)[selected_columns]]
-    return DataFrame(new_columns, Index(_names(df)[selected_columns]))
+    new_metadata = metadata(df)[selected_columns]
```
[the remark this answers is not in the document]
pdeffebach: I think I fixed this by calling a constructor with `new_metadata`

pdeffebach:
To clarify my above comments, if we had the user use 

```
setfield!(metadata(df), :columndata, info...
```

We would run into the problem where `metadata` objects don't know about the names of the columns in the dataframe. `getfield(metadata(df), :columndata)` will only return a `Dict` that is a bunch of arrays. So `info` in the above argument would have to be a vector of the right length, and with all the existing metadata just right. 

Perhaps we should have a setup similar to a `dataframerow`, so that `metadata` can see the dataframe it is attached to? But this hurts us because it means `metadata(df)` behaves less like `columndata(df)`.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: I don't understand: `new_metadata` still isn't used (and that's fine, just remove it).

src/dataframe/dataframe.jl
```diff
@@ -116,8 +117,13 @@ mutable struct DataFrame <: AbstractDataFrame
                 throw(DimensionMismatch("columns must be 1-dimensional"))
             end
         end
-        new(columns, colindex)
+        new(columns, colindex, MetaData())
     end
+
+    function DataFrame(columns::Vector{Any}, colindex::Index, metadata::MetaData)
```
nalimilan: Replace this with `metadata::MetaData=MetaData()` in the constructor above. Here you're bypassing all consistency checks done by the existing constructor.

src/other/metadata.jl
```diff
@@ -0,0 +1,117 @@
...
+# deleting columns is handled by get_index? 
+
+function getmeta(x::MetaData, field::Symbol, col_ind::Int)
+    if haskey(x.columndata, field)
+        return x.columndata[field][col_ind]
+    else
+        return "Field does not exist"
+    end
+end
+
```
nalimilan: Remove empty lines.

src/other/metadata.jl
```diff
@@ -0,0 +1,117 @@
...
+    vcat!(newmeta, rightmeta)
+end
+
+# deleting columns is handled by get_index? 
+
+function getmeta(x::MetaData, field::Symbol, col_ind::Int)
+    if haskey(x.columndata, field)
+        return x.columndata[field][col_ind]
+    else
+        return "Field does not exist"
```
nalimilan: Throw an error?

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: > Is this just for now, or do you envision this in the future? Because an idiomatic way to set labels, particularly with chaining in mind, first argument is a dataframe, returns a dataframe, seems important.

I don't know yet, that's why I'd rather start with the strictly minimal API.

src/dataframe/dataframe.jl
```diff
@@ -344,10 +355,13 @@ function insert_single_column!(df::DataFrame,
         if typeof(col_ind) <: Symbol
             push!(index(df), col_ind)
             push!(columns(df), dv)
+            addcolumn!(metadata(df)) # adds an empty string to all vectors of metadata
```
nalimilan: Should push `nothing` instead.

src/other/metadata.jl
```diff
@@ -0,0 +1,117 @@
...
+end
+
+function addcolumn!(x::MetaData)
+    for key in keys(x.columndata)
+        push!(x.columndata[key], "")
+    end
+end
+
+# For creating a new column
+function addcolumn!(x::MetaData)
```
nalimilan: Duplicated method. Also as noted above it should push `nothing`.

src/other/metadata.jl
```diff
@@ -0,0 +1,117 @@
...
+    nothing
+end
+
+function Base.permute(x::MetaData, p::AbstractVector)
+    new_metadata = copy(x)
+    permute!(new_metadata, p)
+end
+
+function newfield!(x::MetaData, ncol::Int, field::Symbol,)
+    x.columndata[field] = Vector{Any}([nothing for i in 1:ncol])
```
nalimilan: Rather than using `Any`, I think `Union{eltype(info), Nothing}` would be more appropriate. That would be more efficient, and it's probably flexible enough for meta-data. We can always add API to choose a different type later if it turns out to be useful.

src/other/metadata.jl
```diff
@@ -0,0 +1,117 @@
...
+    columndata::Dict{Symbol, Vector}
+end
+    
+MetaData() = MetaData(Dict{Symbol,Vector}())
+
+Base.isequal(x::MetaData, y::MetaData) = isequal(x.columndata, y.columndata)
+Base.:(==)(x::MetaData, y::MetaData) = isequal(x, y)
+
+Base.copy(x::MetaData) = MetaData(copy(x.columndata))
+Base.deepcopy(x::MetaData) = MetaData(copy(x.columndata)) # field is immutable
```
nalimilan: What's immutable?

src/other/metadata.jl
```diff
@@ -0,0 +1,117 @@
+# Defining behavior for DataFrames metadata
+abstract type AbstractMetaData end
+
+mutable struct MetaData <: AbstractMetaData
+    columndata::Dict{Symbol, Vector}
+end
+    
+MetaData() = MetaData(Dict{Symbol,Vector}())
+
+Base.isequal(x::MetaData, y::MetaData) = isequal(x.columndata, y.columndata)
```
nalimilan: Do we really need these definitions?

src/dataframe/dataframe.jl
nalimilan: OK. I'm a bit concerned about the fact that this allocates, even when one isn't using meta-data at all. We should really ensure there's a minimal or zero overhead in that case. Maybe we can handle that by checking whether `metadata` is empty first?

In terms of code organization, maybe this could be moved to a function taking `Index` objects for two data frames, and it could also be used for `join` operations?

src/other/metadata.jl
pdeffebach: Now I throw an error. I'll need help on error type though.

src/other/metadata.jl
pdeffebach: fixed

src/other/metadata.jl
pdeffebach: I thought they were standard boilerplate for new structs. they are deleted now.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
pdeffebach: I also changed `getmeta` and `showmeta` to be more generic. `showmeta` now returns a dataframe, but its probably a bad idea cause dataframes aren't that readable for long strings. But it was easy to write and not a horrible idea.

src/dataframe/dataframe.jl
```diff
@@ -279,7 +284,8 @@ end
 function Base.getindex(df::DataFrame, row_ind::Real, col_inds::AbstractVector)
     selected_columns = index(df)[col_inds]
     new_columns = Any[dv[[row_ind]] for dv in columns(df)[selected_columns]]
-    return DataFrame(new_columns, Index(_names(df)[selected_columns]))
+    # no subsetting required for metadata cause rows dont matter
```
nalimilan: This comment doesn't sound very useful, meta-data just works as the index.

src/dataframe/dataframe.jl
```diff
@@ -749,6 +759,20 @@ merge!(df, df2)  # column z is added, column id is overwritten
 """
 function Base.merge!(df::DataFrame, others::AbstractDataFrame...)
     for other in others
+        d = diff_indices(index(other), index(df))
+        #=
```
nalimilan: This should go to the docstring for `append!`. Also, better find another name since it doesn't follow the signature of the generic `append!`.

src/other/index.jl
```diff
@@ -96,6 +96,14 @@ function Base.delete!(x::Index, idx::Integer)
     return x
 end
 
+"""
+Returns returns the indices of the columns in x that are not in y. 
+"""
+function diff_indices(x::Index, y::Index)
```
nalimilan: Sorry, when I suggested having a separate function for this, I was thinking about a `MetaData`-aware function which would avoid calling `setdiff` if the meta-data is empty. But it can't live in index.jl, which is only about `Index`. Since the function is just `x[setdiff(names(x), names(y))]`, it's not very useful. Maybe just move all that stuff into `append!` and pass it the two `Index` objects. That way it can be a no-op when meta-data is empty. I guess it depends on what can be shared with the `join` code.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: As I said, let's start with a minimal implementation. We can always add convenience method later if that's useful.

I'd also rather rename `getmeta` to `metadata` and `addmeta!` to `setmetadata!` or `metadata!`.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
...
+function newfield!(x::MetaData, ncol::Int, field::Symbol, info)
+    x.columndata[field] = Vector{Union{typeof(info), Nothing}}([nothing for i in 1:ncol])
+end
+
+function addmeta!(x::MetaData, col_ind::Int, ncol::Int, field::Symbol, info)
+    if !haskey(x.columndata, field)
+        newfield!(x, ncol, field, info)
+    end
+    x.columndata[field][col_ind] = info
+    return nothing
```
nalimilan: Could as well remove this and return `info `, that can be useful for chaining.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
...
+
+function Base.append!(leftmeta::MetaData, rightmeta::MetaData, ncol_left::Int, ncol_right::Int)
+    notonleft = setdiff(keys(rightmeta.columndata), keys(leftmeta.columndata))
+    notonright = setdiff(keys(leftmeta.columndata), keys(rightmeta.columndata))
+
+    for field in notonleft
+        newfield!(leftmeta, ncol_left, field, nothing)
+    end
+
+    for field in notonright
```
nalimilan: It's not OK to modify `rightmeta` in this function. More generally, it would be more efficient to create directly vectors of the final length in the loop below, rather than creating vectors for each input data frame, and then concatenate them.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
...
+function append(leftmeta::MetaData, rightmeta::MetaData)
+    append!(copy(leftmeta), rightmeta)
+end
+
+# deleting columns is handled by get_index? 
+function getmeta(x::MetaData, col_ind::Int, field::Symbol)
+    if haskey(x.columndata, field)
+        return x.columndata[field][col_ind]
+    else
+        error("The field does not exist")
```
nalimilan: Use `KeyError`, and print the name of the field.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
...
+end
+
+function Base.permute!(x::MetaData, p::AbstractVector)
+    for key in keys(x.columndata)
+        x.columndata[key] = permute!(x.columndata[key], p)
+    end 
+    nothing
+end
+
+function Base.permute(x::MetaData, p::AbstractVector)
```
nalimilan: This function isn't used AFAICT.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
...
+        newfield!(rightmeta, ncol_right, field, nothing)
+    end
+
+    for key in keys(leftmeta.columndata)
+        leftmeta.columndata[key] = 
+            vcat(leftmeta.columndata[key], rightmeta.columndata[key])
+    end
+end
+
+function append(leftmeta::MetaData, rightmeta::MetaData)
```
nalimilan: Not used.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
...
+        new_columndata[key] = new_columndata[key][col_inds]
+    end 
+    MetaData(new_columndata)
+end
+
+function Base.permute!(x::MetaData, p::AbstractVector)
+    for key in keys(x.columndata)
+        x.columndata[key] = permute!(x.columndata[key], p)
+    end 
+    nothing
```
nalimilan: `permute!` from Base returns the `p`, better do the same.

src/dataframe/dataframe.jl
```diff
@@ -1100,9 +1124,59 @@ function permutecols!(df::DataFrame, p::AbstractVector)
...
+##############################################################################
+
+"""
+    addlabel!(df::DataFrame, var::Symbol, label::String)
+
+Adds a label to a DataFrame. Does not add other metadata.
+"""
+function addmeta!(df::DataFrame, var::Symbol, field::Symbol, info)
+    ind = index(df)[var]
+    # pass the number of columns to the function so that it can create a new array of
```
nalimilan: That sounds OK to me (and no need for a comment).

BTW, these functions would be turned into one-liners if you skip creating the `ind` variable.

src/dataframe/dataframe.jl
```diff
@@ -147,7 +149,8 @@ function DataFrame(; kwargs...)
 end
 
 function DataFrame(columns::AbstractVector,
-                   cnames::AbstractVector{Symbol}=gennames(length(columns));
+                   cnames::AbstractVector{Symbol}=gennames(length(columns)),
+                   metadata = MetaData();
```
nalimilan: Unused argument. Anyway for now `MetaData` is purely internal, like `Index`, so it shouldn't appear here.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
...
+end
+
+function Base.permute(x::MetaData, p::AbstractVector)
+    new_metadata = copy(x)
+    permute!(new_metadata, p)
+end
+
+
+function newfield!(x::MetaData, ncol::Int, field::Symbol, info)
+    x.columndata[field] = Vector{Union{typeof(info), Nothing}}([nothing for i in 1:ncol])
```
nalimilan: `Union{typeof(info), Nothing}[nothing for i in 1:ncol]` avoids a copy.

src/other/metadata.jl
```diff
@@ -0,0 +1,91 @@
+# Defining behavior for DataFrames metadata
+abstract type AbstractMetaData end
+
+mutable struct MetaData <: AbstractMetaData
+    columndata::Dict{Symbol, Vector}
```
nalimilan: `struct` would be enough, right?

Also, `columndata` is a bit of a weird name for this, since it sounds like there are other fields with non-column data. Maybe just `dict`?

src/dataframe/dataframe.jl
```diff
@@ -263,7 +267,8 @@ end
 function Base.getindex(df::DataFrame, col_inds::AbstractVector)
     selected_columns = index(df)[col_inds]
     new_columns = columns(df)[selected_columns]
-    return DataFrame(new_columns, Index(_names(df)[selected_columns]))
+    new_metadata = metadata(df)[selected_columns]
+    return DataFrame(new_columns, Index(_names(df)[selected_columns]), new_metadata)
```
nalimilan: Better use the same pattern as elsewhere and drop the `new_metadata` variable.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
pdeffebach: I understand what you are saying more about `setindex`. Might be nice to write 

`metadata(df, :x1, :label) = "A variable label"` work.

src/dataframe/dataframe.jl
pdeffebach: fixed

src/other/metadata.jl
pdeffebach: done.

src/dataframe/dataframe.jl
pdeffebach: okay this part is still a bit confusing for me. but i think this makes sense.

src/other/metadata.jl
pdeffebach: changed.

src/other/metadata.jl
pdeffebach: changed.

src/dataframe/dataframe.jl
pdeffebach: fixed

pdeffebach:
I think I got my push and pulls confused for this, somehow making me add the new changes to master... let me know what to do to sort it out, because I'm  not sure what to do in this situation. I think you reject these and I re-submit?


I responded to all the changes, but there are a few issues to work on. 

- allocating for `merge!` . It seems intuitive that that the new metadata should be a combination of the two in the merge, I'm not sure what a non-allocating version would look like, since the current implementation needs full vectors in the `columndata` `Dict` (now `dict` `Dict`). 
- I like the idea of having `metadata(df, :x1, :field) = "my information`. This requires overloading `setindex` and `getindex` right?
- I think that automatically making a new vector if you add metadata to a field that doesn't exist yet is dangerous, because then typos can allocate new fields silently. But this is small and can be addressed later. 

Major issue still to come is `join`.

nalimilan:
> I think I got my push and pulls confused for this, somehow making me add the new changes to master... let me know what to do to sort it out, because I'm not sure what to do in this situation. I think you reject these and I re-submit?

Better continue the conversation in this PR. You should be able to fix this with `git fetch && git rebase -i origin/master`, and removing lines which correspond to unrelated commits. Then if everything it OK, do a `git push --force`. (In general, better work in special branches and keep `master` in sync with `origin`.)

pdeffebach:
@nalimilan i re-organized everthing with git to try and fix this. In the process I guess I closed this branch. If it's okay, can I submit another PR? I have all the code still, and it is now up to date with current master. I can go through and add comments where you left off too, to make the transition easier.

nalimilan:
You should have been able to push your branch to this PR even if locally it has a different name, but now that the PR has been closed GitHub won't leave us reopening it anyway for strange reasons, so you'll have to file another one.

PR #1432, opened 2018-06-18
Relax column type requirements in append! and improve behavior on error

nalimilan:
Strictly checking column types was particularly problematic now that we support columns with both `Union{T,Missing}` and `T` element types.

See https://github.com/JuliaData/DataFrames.jl/issues/1291#issuecomment-397868499.

### DIFF
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -452,7 +452,14 @@ module TestDataFrame
     df = DataFrame(A = 1:10, B = 'A':'J')
     @test !(df[:,:] === df)
 
-    @test append!(DataFrame(A = 1:2, B = 1:2), DataFrame(A = 3:4, B = 3:4)) == DataFrame(A=1:4, B = 1:4)
+    df = DataFrame(A = 1:2, B = 1:2)
+    df2 = DataFrame(A=1:4, B = 1:4)
+    @test append!(df, DataFrame(A = 3:4, B = [3.0, 4.0])) == df2
+    @test_throws InexactError append!(df, DataFrame(A = 3:4, B = [3.5, 4.5]))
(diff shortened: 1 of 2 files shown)


test/dataframe.jl
```diff
@@ -452,7 +452,14 @@ module TestDataFrame
...
     @test !(df[:,:] === df)
 
-    @test append!(DataFrame(A = 1:2, B = 1:2), DataFrame(A = 3:4, B = 3:4)) == DataFrame(A=1:4, B = 1:4)
+    df = DataFrame(A = 1:2, B = 1:2)
+    df2 = DataFrame(A=1:4, B = 1:4)
+    @test append!(df, DataFrame(A = 3:4, B = [3.0, 4.0])) == df2
+    @test_throws InexactError append!(df, DataFrame(A = 3:4, B = [3.5, 4.5]))
+    @test df == df2
+    @test_throws MethodError append!(df, DataFrame(A = 3:4, B = ["a", "b"]))
+    @test df == df2
```
omus: Would be good to add a test which includes `missing`:
```julia
using DataFrames

df = DataFrame(A=[missing, 1])
df2 = DataFrame(A=3:4)
append!(df, df2)
```

nalimilan: There's really nothing special about that example, `Int` is a subtype of `Union{Int,Missing}`. So I don't really see why we'd need to test it.

nalimilan:
> Looks OK. Additionally it is actually better than append! in Base, which grows the vector even on error (and garbage is left in the resized vector - not sure if this should be documented in Base).

I think there's a reason for this and I remember a discussion about it, but I forgot the details.

PR #1434, opened 2018-06-19
patch bug in indicator that modified input datasets

nickeubank:
Previous version of indicator function would modify input datasets. This corrects that behavior and adds tests.

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -296,8 +296,8 @@ function Base.join(df1::AbstractDataFrame,
                  indicator_cols[i] *= 'X'
             end
          end
-        df1[Symbol(indicator_cols[1])] = trues(nrow(df1))
-        df2[Symbol(indicator_cols[2])] = trues(nrow(df2))
+         df1 = hcat(df1, DataFrame(Dict(indicator_cols[1] => trues(nrow(df1)))))
+         df2 = hcat(df2, DataFrame(Dict(indicator_cols[2] => trues(nrow(df2)))))
     end
 
     if kind == :cross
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/join.jl
```diff
@@ -296,8 +296,8 @@ function Base.join(df1::AbstractDataFrame,
                  indicator_cols[i] *= 'X'
             end
          end
-        df1[Symbol(indicator_cols[1])] = trues(nrow(df1))
-        df2[Symbol(indicator_cols[2])] = trues(nrow(df2))
+         df1 = hcat(df1, DataFrame(Dict(indicator_cols[1] => [true for i in 1:nrow(df1)] )))
```
nalimilan: Better keep using `trues`.

test/join.jl
```diff
@@ -584,8 +584,13 @@ module TestJoin
                                     Name = ["John Doe", "Jane Doe", "Jane Doe", "Joe Blogs", missing],
                                     Job = ["Lawyer", "Doctor", "Florist", missing, "Farmer"],
                                     _merge = ["both", "both", "both", "left_only", "right_only"])
+
+        pre_join_name = copy(name)
```
nalimilan: As noted in the other PR, why is this needed?

nickeubank: Because `name` is being modified by `join`. that's the purpose of this patch. So to test and make sure that isn't happening in the future, we need a copy of `name` from before the join to compare to the post-join version of join.

src/abstractdataframe/join.jl
nickeubank: patched

test/join.jl
nalimilan: OK. These cases are difficult to decide: virtually all functions could modify their inputs, and we don't typically check this for all of them. Anyway if you prefer adding this please add a comment.

nickeubank: My philosophy if that if a bug is found, when you patch it you add a test to ensure that the bug would be caught if it weren't fixed. Otherwise how do you know if your patch works? I'll add a comment.

nickeubank:
thanks!

PR #1437, opened 2018-06-29
improve iteration protocol and fix assignment in combine

### LINKED ISSUE #1431: Fixes needed in grouping.jl code
I am adding this as a TODO as maybe someone would be willing to make a PR (I can go back to it in a few weeks if this is not fixed).

The following line:
https://github.com/JuliaData/DataFrames.jl/blob/master/src/groupeddataframe/grouping.jl#L199
needs '.=' assignment as writing only `=` is deprecated in Julia 0.7.

Also code starting from line:
https://github.com/JuliaData/DataFrames.jl/blob/master/src/groupeddataframe/grouping.jl#L97
needs correction as it uses an old iteration protocol.

### LINKED ISSUE #1441: dropmissing should not allow missing values in columns
It seems like `dropmissing` should change the type of columns from `Union{T, Missing}` to `T`, at least by default. That's more efficient, but also safer since missing values cannot be introduced silently in the resulting data frame.

bkamins:
This resolves #1431.

@nalimilan I was not sure which version of Julia should be a cutoff (or is there some other canonical way of supporting old and new iteration protocol).

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -94,11 +94,22 @@ groupby(d::AbstractDataFrame, cols;
         sort::Bool = false, skipmissing::Bool = false) =
     groupby(d, [cols], sort = sort, skipmissing = skipmissing)
 
-Base.start(gd::GroupedDataFrame) = 1
-Base.next(gd::GroupedDataFrame, state::Int) =
-    (view(gd.parent, gd.idx[gd.starts[state]:gd.ends[state]]),
-     state + 1)
-Base.done(gd::GroupedDataFrame, state::Int) = state > length(gd.starts)
+if VERSION < v"0.7.0-DEV.5126"
+    Base.start(gd::GroupedDataFrame) = 1
+    Base.next(gd::GroupedDataFrame, state::Int) =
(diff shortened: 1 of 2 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -94,11 +94,22 @@ groupby(d::AbstractDataFrame, cols;
         sort::Bool = false, skipmissing::Bool = false) =
     groupby(d, [cols], sort = sort, skipmissing = skipmissing)
 
-Base.start(gd::GroupedDataFrame) = 1
-Base.next(gd::GroupedDataFrame, state::Int) =
-    (view(gd.parent, gd.idx[gd.starts[state]:gd.ends[state]]),
-     state + 1)
-Base.done(gd::GroupedDataFrame, state::Int) = state > length(gd.starts)
+if VERSION < v"0.7.0-alpha"
```
nalimilan: AFAICT the exact commit is 0.7.0-DEV.5126.

bkamins: Fixed. Thank you for being precise (the issue came out independently today https://discourse.julialang.org/t/maintaining-iterator-protocol-for-both-0-6-and-0-7/12084/4 😄).

PR #1440, opened 2018-07-06
Added Pkg3 toml files

ExpandingMan:
This should make it significantly easier to work on in 0.7 as it will allow you to use updated packages, and update them further using Pkg3.

Note that here I used the latest tagged versions of all packages, I figure that due to the upgrading frenzy we need *at least* that.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,160 @@
+[[Base64]]
+uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+
+[[BinaryProvider]]
+deps = ["Compat", "Pkg", "SHA"]
+git-tree-sha1 = "94dac52c662ca793008fb689e3daf626b6139943"
+uuid = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
+version = "0.3.3"
+
+[[CategoricalArrays]]
+deps = ["Compat", "Future", "JSON", "Missings", "Printf", "Reexport"]
+git-tree-sha1 = "6a78981b00a954d850d923cae8500801647d843e"
+uuid = "324d7699-5711-5eae-9e2f-1d82baa6b597"
+version = "0.3.10"
+
+[[CodecZlib]]
+deps = ["BinaryProvider", "Compat", "Test", "TranscodingStreams"]
+git-tree-sha1 = "82742ef572290f566c6fc6d977d87586a0d3d5f1"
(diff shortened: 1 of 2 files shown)


ararslan:
I don't think Manifest files are supposed to be checked into repositories.

ExpandingMan:
This is one of several aspects of Pkg3 that I'm rather confused about.  Looking at the existing [documentation](https://docs.julialang.org/en/latest/stdlib/Pkg/) I see that the manifest file is "optional", but I don't see it say anything about whether it belongs it the git repo.  Since it lists dependencies further down the dependency graph, I'm assuming that you're right and it shouldn't be included.

Ok, I've removed it.

quinnj:
I think it comes back to whether your code should be considered as an "application" or a "library": for a "library" like this DataFrames package, we _don't_ want to include the Manifest, because the Manifest is more strict than the Project; i.e. it includes the exact versions of dependencies that must be used. The Project.toml, however, includes the list of dependencies and any necessary compat, while still trying to allow as many versions to co-exist as possible. Basically, we should try to allow any version of DataFrames to work w/ as many versions of everything else as possible. Updates to the library or its dependencies can be managed by the end user if they need to update a package here or there, which would update the Manifest.toml in whatever environment they're working in.

In the context of an "application", however, the end goal is just to ensure that the application _runs_; in this case, it's much more appropriate the include the Manifest.toml file to ensure exact reproducibility by anyone who wishes to run the application. By running `Pkg.instantiate()` within the application directory, the exact versions of all required dependencies will be installed so ensure proper running of the application. Future updates won't be managed by the end user, but the application developer by updating the Manifest.toml file.

ExpandingMan:
[This](https://discourse.julialang.org/t/where-does-pkg-3-find-version-numbers/12330) thread on discourse made things a lot clearer for me.  It seems we *definitely* do not want to include the `Manifest.toml`.

Are there other issues?  @nalimilan could you clarify what you think might go wrong?

nalimilan:
I don't know, I'd just like somebody knowledgeable to confirm it's OK.

ararslan:
@KristofferC?

KristofferC:
This is not really good, it is missing `name`, `uuid` and `compat` from the REQUIRE file. I will put https://github.com/JuliaLang/Pkg.jl/pull/392 into my fork of PkgDev, use it to generate the REQUIRE file and make a PR against DataFrames.

PR #1442, opened 2018-07-08
Update docs for new `describe`

pdeffebach:


nalimilan:
Thanks!

This reminds me it would be nice to print `nothing` as the empty string to make this even nicer.

pdeffebach:
For all of dataframes? Yeah I think thats a good idea.

bkamins:
This should be clearly documented as otherwise there is a risk of user confusion (a milder version would be to print `nothing` using a different color like for `missing`).

pdeffebach:
At the very least, it's super simple to just ad `try mean(col) catch "" end` in `describe`. That would definitely make the output prettier.

bkamins:
Ah - now I understand. I agree with replacing `nothing` with `""` in describe as @pdeffebach explained.
I thought of a general printing of `nothing` when displaying `DataFrame` (but this is clearly a separate issue).

bkamins:
Reflecting REPL behavior is convincing. 👍 for general behavior then (but let us document it).

pdeffebach:
What about `missing`, then? We could do colored `m` and `n` for missing and nothing, respectively. Though I would agree that a blank being interpreted as `nothing` is more natural than, say, having another method of printing for `missing`

pdeffebach:
Opened up a PR for printing an empty string instead of nothings. #1444 . But I think there is a better strategy than what I just wrote.

nalimilan:
We've considered using various symbols to represent `missing` (in Slack discussions and at https://github.com/JuliaData/Missings.jl/pull/51#issuecomment-343761689), but it would be confusing to use a custom symbol if we don't defined the corresponding object equal to `missing`. Maybe we could agree on a standard symbol -- at the time of that PR, printing wasn't the most pressing issue.

PR #1444, opened 2018-07-08
make REPL printing of `nothing` an `empty string`

pdeffebach:
Also changes the name of `nastring` to `missingstring`

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -57,6 +57,7 @@ end
 ourshowcompact(io::IO, x::Any) = show(IOContext(io, :compact=>true), x) # -> Void
 ourshowcompact(io::IO, x::AbstractString) = escape_string(io, x, "") # -> Void
 ourshowcompact(io::IO, x::Symbol) = ourshowcompact(io, string(x)) # -> Void
+ourshowcompact(io::IO, x::Nothing) = ourshowcompact(io, "") # -> Void
 
 #' @description
 #'
@@ -250,9 +251,6 @@ function showrowindices(io::IO,
                 strlen = ourstrwidth(s)
                 if ismissing(s)
                     printstyled(io, s, color=:light_black)
-                elseif s == nothing
-                    ourshowcompact(io, "")
-                    strlen = 0
                 else
                     ourshowcompact(io, s)
                 end
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -250,6 +250,9 @@ function showrowindices(io::IO,
                 strlen = ourstrwidth(s)
                 if ismissing(s)
                     printstyled(io, s, color=:light_black)
+                elseif s == nothing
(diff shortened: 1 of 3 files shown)


pdeffebach:
Actually, I think now that the better way to go about this is to add a method for `ourshowcompact`, but let me know if that's the way forward.

pdeffebach:
Actually, why didn't we just overload `ourshowcompact` to have the `prettyprint` call for `missing` too?

pdeffebach:
Okay now the overall diff should just be one line: an added `ourshowcompact` method.

bkamins:
Maybe one thing - as test coverage has dropped - is this change covered by test cases?

pdeffebach:
Just added some tests.

src/abstractdataframe/show.jl
```diff
@@ -57,6 +57,7 @@ end
 ourshowcompact(io::IO, x::Any) = show(IOContext(io, :compact=>true), x) # -> Void
 ourshowcompact(io::IO, x::AbstractString) = escape_string(io, x, "") # -> Void
 ourshowcompact(io::IO, x::Symbol) = ourshowcompact(io, string(x)) # -> Void
+ourshowcompact(io::IO, x::Nothing) = ourshowcompact(io, "") # -> Void
```
nalimilan: Couldn't we just do nothing and return `nothing`? Calling `ourshowcompact` on an empty strings sounds overkill for a no-op.

test/show.jl
```diff
@@ -205,6 +205,21 @@ module TestShow
...
+    str = String(take!(io))
+    @test str == """
+    3×2 DataFrames.DataFrame
+    │ Row │ A   │ B         │
+    ├─────┼─────┼───────────┤
+    │ 1   │ 1.0 │ something │
+    │ 2   │ 2.0 │ g         │
+    │ 3   │ 3.0 │           │"""
+
+
```
nalimilan: One empty line is enough.

src/abstractdataframe/show.jl
pdeffebach: That's a good idea. We don't need to print anything, just define `strlen` to be zero.

src/abstractdataframe/show.jl
```diff
@@ -249,7 +249,10 @@ function showrowindices(io::IO,
                 s = df[i, j]
                 strlen = ourstrwidth(s)
                 if ismissing(s)
-                    printstyled(io, s, color=:light_black)
+                    printstyled(io, "m", color=:light_black)
```
nalimilan: This changes the printing of `missing`, right? I'd rather keep it as-is, since `m` isn't a real Julia object.

pdeffebach: sorry, that was just me playing around. can you reject that change or should I submit another PR?

nalimilan:
Again, we forgot that the docs should be updated... :-)

Issue #1400, opened 2018-04-19
Scalar indexing by row should return a DataFrameRow


rofinn:
Example)
```julia
julia> df
10×3 DataFrames.DataFrame
│ Row │ A  │ B         │ C     │
├─────┼────┼───────────┼───────┤
│ 1   │ 1  │ 0.778923  │ true  │
│ 2   │ 2  │ 0.292706  │ false │
│ 3   │ 3  │ 0.8056    │ true  │
│ 4   │ 4  │ 0.868616  │ false │
│ 5   │ 5  │ 0.788371  │ false │
│ 6   │ 6  │ 0.0806976 │ true  │
│ 7   │ 7  │ 0.909549  │ true  │
│ 8   │ 8  │ 0.609525  │ true  │
│ 9   │ 9  │ 0.0454008 │ true  │
│ 10  │ 10 │ 0.214475  │ false │

julia> obs = df[1, :]
DataFrameRow (row 1)
A  1
B  0.7789232193182367
C  true

julia> obs[:A]
1
```

NOTE: Unfortunately, this breaks a lot of existing functionality (e.g., `DataFrameRow`s depend on the existing `df[1, :]` behaviour), so this may result in a larger discussion around replacing DataFrameRows with a `Dict` or `NamedTuple`. Similarly, we may want to move in the direction where a `DataFrame`s is just an alias for a `Dict{Symbol, <:AbstractVector}` or `NamedTuple`.

nalimilan:
+1

Regarding the fact that `df[1, :]` fails, it shouldn't be hard to fix since the code for `getindex` already exists, it just needs to be moved from `DataFrame` to `DataFrameRow` methods.

We've discussed replacing `DataFrameRow` with `NamedTuple` before at https://github.com/JuliaData/DataFrames.jl/issues/1337, but unfortunately it's not really possible since you couldn't do things like `row[:x] += 1`. So we should probably keep `DataFrameRow`, but have it behave like a `NamedTuple` as much as possible, in particular via field overloading. Also a `Dict` would really be too slow and that's not the right structure for this (the parent data frame already includes a dict to lookup columns).

ararslan:
If we express a `DataFrame` as something like
```julia
const DataFrame = NamedTuple{names,<:NTuple{N,<:AbstractVector}} where {names,N}
```
then we would get mutable rows trivially, since to get a row you'd need to get a column first, which would expose the underlying vector. However, indexing then becomes much less convenient...

[...]

PR #1449, opened 2018-07-11
Deprecate iterating over DataFrameRow in favor of pairs()

nalimilan:
This will allow changing iteration to return values instead of pairs, which is consistent with `NamedTuple` and `Vector`.

First step to fix https://github.com/JuliaData/DataFrames.jl/issues/1400. See also https://github.com/JuliaData/DataFrames.jl/pull/1439.

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -41,14 +41,22 @@ Base.length(r::DataFrameRow) = size(parent(r), 2)
 
 Compat.lastindex(r::DataFrameRow) = size(parent(r), 2)
 
-Base.collect(r::DataFrameRow) = Tuple{Symbol, Any}[x for x in r]
+function Base.iterate(r::DataFrameRow)
+    Base.depwarn("iteration over DataFrameRow will return values in the future:" *
+                 "use pairs(r::DataFrameRow) to get the current behavior",
+                 :start)
+    iterate(r, 1)
+end
 
-function Base.iterate(r::DataFrameRow, st=1)
+function Base.iterate(r::DataFrameRow, st)
     st > length(r) && return nothing
     return ((_names(r)[st], r[st]), st + 1)
 end
 
 Base.convert(::Type{Array}, r::DataFrameRow) = convert(Array, parent(r)[row(r),:])
+
+Base.keys(r::DataFrameRow) = names(parent(r))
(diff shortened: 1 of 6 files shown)


pdeffebach:
We should do this for `eachcol` too. It currently also returns a tuple so something like 
`reduce(+, eachcol(df)` doesn't work.

bkamins:
Having reviewed this PR I have the following question.

Why `keys` and `values` for `DataFrame` are deprecated? (this also makes `pairs` not work on `DataFrame`). Probably there is some good reason but I did not participate in those discussions.

In particular as `values` are deprecated why do we recommend using `DataFrame.columns` instead which is not exported. Is `DataFrame.columns` part of official API (though not exported)?

bkamins:
Regarding `keys`/`values` issue: JuliaDB.jl returns rows as `values`, and `keys` are not supported at least for now. I am not sure if rows or columns approach would be better to take. I would slightly prefer column approach as it was before, as then we could deprecate `eachcol` for `pairs` as they are essentially the same and `values` would be used instead of `DataFrame.columns`.

Even if we do not make a decision here probably a public way of getting raw columns (`DataFrame.columns` now) would be good to have. Any opinion?

pdeffebach:
@nalimilan I just checked out this branch and I was playing around with it. I just did 

```
t = DataFrameRow(df, 1)
mean(t)
> Throws an method error about adding Symbols
```

A named tuple wouldn't throw an error here. Is there anything we can do to this PR to add the behavior to make a `DataFrameRow` more vector like?

pdeffebach:
I apologize. i didn't realize this was *just* the deprecation PR.

bkamins:
The rationale for `DataFrameRow` over `NamedTuple` is that it allows you to mutate parent `DataFrame` I guess.

Regarding the percentage of `eachcol` vs `eachrow` - it is hard to tell for me. What I know for sure is that often I wished to have access to `df.columns` and used `last.(eachcol(x))` which is not that bad but could be more direct. I am not 100% sure though.

nalimilan:
Yes, this is just the deprecation part unfortunately...

I'm not completely convinced we should define `pairs` to iterate over columns. I think I'd like to add a `columns` function to iterate over columns, and/or maybe deprecate `eachcol`. At any rate it would be nice to have symmetric `rows`/`columns` or `eachrow`/`eachcol` instead of the inconsistency we currently have.

PR #1450, opened 2018-07-11
fix docs for describe printing nothings

pdeffebach:


docs/src/man/getting_started.md
```diff
@@ -161,10 +161,11 @@ Having seen what some of the rows look like, we can try to summarize the entire
 
 ```jldoctest dataframe
 julia> describe(df)
+2×8 DataFrames.DataFrame
 │ Row │ variable │ mean    │ min │ median  │ max │ nunique │ nmissing │ eltype │
 ├─────┼──────────┼─────────┼─────┼─────────┼─────┼─────────┼──────────┼────────┤
-│ 1   │ A        │ 4.5     │ 1   │ 4.5     │ 8   │ nothing │ nothing  │ Int64  │
-│ 2   │ B        │ nothing │ F   │ nothing │ M   │ 2       │ nothing  │ String │
+│ 1   │ A        │ 4.5     │ 1   │ 4.5     │ 8   │         │          │ Int64  │
+│ 2   │ B        │         │ F   │         │ M   │ 2       │          │ String │
```
nalimilan: I'm surprised this column isn't less wide now. I guess you just replaced `nothing` with 7 spaces, right? :-)

pdeffebach: no i copied and pasted! 

It's a call to `ourstrwidth` that's causing this. `ourstrwidth` prints the value, then uses `take` to get the length. I'll submit a PR to change this behavior so that `ourstrwidth(x::Nothing) = 0`

pdeffebach: Actually, modifying `ourshowcompact` might be the best thing to do here, because `ourstrwidth` calls `ourshowcompact` when it prints.

ararslan:
Wth, `nothing` prints as blank space? That's pretty bad.

pdeffebach:
I had the idea the other day of printing to match `maxcolwidth` so we would have like 

```
| A | a_var | a_longer_var |
|--------------------------|
| n | noth  | nothing      |
```

We could maybe do the same with missings.

nalimilan:
> Wth, nothing prints as blank space? That's pretty bad.

Not as a blank space: as nothing (as at the REPL). See https://github.com/JuliaData/DataFrames.jl/pull/1409.

pdeffebach:
With these commits I added a method to `ourshowcompact(io::IO, x::Nothing) = print("")`. 

The reason for this is that the way `show` calculates the width of a column is to call `ourshowcompact` on each element of a column and uses `take` to get the width. So it was either add if statements in both the `getmaxwidth` functions and the printing function or add a method to `ourshowcompact`.

nalimilan:
`print("")` isn't correct, it will print to `stdout`. Also since it's a no-op, just define the function as `= nothing`.

BTW, looks like you rebased on the wrong branch.

pdeffebach:
I hope this fix gets it back on track.

pdeffebach:
wait no i messed this up to. Can we close this and I can do it right this time.

PR #1453, opened 2018-07-13
Clean up warnings in tests

dmbates:


src/DataFrames.jl
```diff
@@ -7,19 +7,12 @@ module DataFrames
...
 using Compat: @warn
 
-if VERSION >= v"0.7.0-DEV.2738"
-    const kwpairs = pairs
-else
-    kwpairs(x::AbstractArray) = (first(v) => last(v) for v in x)
-    using Compat.IOBuffer
-end
-
+const kwpairs = pairs
```
nalimilan: You should be able to remove this too.

src/abstractdataframe/io.jl
```diff
@@ -134,26 +134,8 @@ end
...
-        if char == "\\"
-            return "\\textbackslash{}"
-        elseif char == "~"
-            return "\\textasciitilde{}"
-        else
-            return string("\\", char)
-        end
-    end
+function latex_char_escape(char::Char)
+    char == '\\' ? "\\textbackslash{}" : char == '~' ? "\\textasciitilde{}" : string('\\', char)
```
nalimilan: The old version seemed more readable to me, even if it's longer.

src/other/utils.jl
```diff
@@ -19,12 +15,12 @@ function identifier(s::AbstractString)
 end
 
 function makeidentifier(s::AbstractString)
-    i = start(s)
-    done(s, i) && return "x"
+    i = firstindex(s)
```
nalimilan: This is not completely correct AFAIK, as the iteration state isn't required to be the index. We should use `iterate(s)` instead, and do `i === nothing && return "x"`. Then below you also need to check `i === nothing` rather than `i <= ncodeunits(s)`.

test/dataframe.jl
```diff
@@ -350,7 +346,7 @@ module TestDataFrame
                                                           :string, :string_missing,
                                                           :dates, :catarray],
                                               mean = [2.5, 2.0, nothing, nothing, nothing, nothing],
-                                              std = [Compat.std(df[:number]), 1.0, nothing, 
+                                              std = [Statistics.std(df[:number]), 1.0, nothing,
```
nalimilan: No need to prefix with the module, right?

test/duplicates.jl
```diff
@@ -20,7 +20,7 @@ module TestDuplicates
     @test pdf ≅ updf
 
     @testset "missing" begin
-        df = DataFrame(A = 1:12, B = repeat('A':'C', inner=4))
-        @test DataFrames.colmissing(df) == [0, 0]
+        df1 = DataFrame(A = 1:12, B = repeat('A':'C', inner=4))
```
nalimilan: I don't think we should change this. There are many more warnings like this which are totally spurious, better just wait until the deprecations are removed.

test/show.jl
```diff
@@ -149,69 +149,32 @@ module TestShow
...
-
+                   
     # Test showing nothing
     df_nothing = DataFrame(A = [1.0, 2.0, 3.0], B = ["something", "g", nothing])
     io = IOBuffer()
     show(io, df_nothing)
     str = String(take!(io))
     @test str == """
-    3×2 DataFrames.DataFrame
+    3×2 $DataFrame
```
nalimilan: Is the `$` really needed? Better check the exact output we expect if possible.

nalimilan:
Ah, you also need to drop Julia 0.6 from the Travis and AppVeyor config files, and add 0.7 instead (in addition to nightlies).

src/other/utils.jl
dmbates: These are the changes suggested by the deprecation messages.  I tried to alter the code according to your suggestions and was unable to get a version that passed tests.  It may be better to revert to the original version as I am not confident in my changes.  I don't have a clear idea of the logic in this code.

test/dataframe.jl
dmbates: Removed `Statistics.`

src/abstractdataframe/io.jl
dmbates: I reverted to the earlier version.

src/DataFrames.jl
dmbates: Done

test/duplicates.jl
dmbates: I think the issue is that some of the tests are outside of `@testset` and some are within.  I agree that it is probably not a good idea to fix a few of these in isolation.

src/other/utils.jl
nalimilan: Ah, you're right that the deprecation suggests this. But I'm not sure whether it's supposed to be a general solution, or whether it's just because there's no better deprecation for `start(::AbstractString)` as a temporary measure. Anyway, we really shouldn't need to use `firstindex` here.

Doesn't something like this work?
```julia
i = iterate(s)
i === nothing && return "x"
c, i = iterate(s, i)
...
while i !== nothing
    c, i = iterate(s, i)

File: test/duplicates.jl (hunk 5 above)
Comment (nalimilan) [reply test/duplicates.jl]: Yeah, but we really don't care whether the variable is global or local here anyway since we generally (always?) overwrite it.

File: src/other/utils.jl (hunk 3 above)
Comment (dmbates) [reply src/other/utils.jl]: On thinking about it some more, it isn't `i` that gets checked for `=== nothing`.  I believe it is more like
```julia
(iresult = iterate(s)) === nothing && return "x"
(c, i) = iresult
...
while (iresult = iterate(s, i)) !== nothing
    (c, i) = iresult
```
I'll try that out

src/subdataframe/subdataframe.jl
```diff
@@ -158,3 +158,5 @@ end
 Base.map(f::Function, sdf::SubDataFrame) = f(sdf) # TODO: deprecate
 
 without(sdf::SubDataFrame, c) = view(without(parent(sdf), c), rows(sdf))
+# Resolve a method ambiguity
+without(sdf::SubDataFrame, c::Vector{<:Integer}) = view(without(parent(sdf), c), rows(sdf))
```
nalimilan: Doesn't the same ambiguity exist for `c::Int`?

test/grouping.jl
```diff
@@ -171,18 +171,18 @@ module TestGrouping
     @test gd[4] == DataFrame(Key1="B", Key2="B", Value=4)
 
     @testset "grouping with missings" begin
-        df = DataFrame(Key1 = ["A", missing, "B", "B", "A"],
-                       Key2 = CategoricalArray(["B", "A", "A", missing, "A"]),
-                       Value = 1:5)
+        global df = DataFrame(Key1 = ["A", missing, "B", "B", "A"],
```
nalimilan: I don't think this is right. `df` shouldn't be used outside of this testset.

test/cat.jl
```diff
@@ -74,109 +75,76 @@ module TestCat
...
-                                         [2.0,2.0,1.0,3.0,2.0,2.0,1.0,3.0],
-                                         [2,2,2,3,2,2,2,3]])
-
-    # Don't fail on non-matching types
-    df[1] = zeros(Int, nrow(df))
-    @test vcat(df, alt_df) == DataFrame([[0.0,0.0,0.0,0.0,3.0,2.0,3.0,3.0],
-                                         [2.0,2.0,1.0,3.0,2.0,2.0,1.0,3.0],
-                                         [2,2,2,3,2,2,2,3]])
-
-    dfr = vcat(df4, df4)
```
nalimilan: This block is missing in the new code.

test/constructors.jl
```diff
@@ -1,73 +1,74 @@
...
     #
     # DataFrame
     #
+    @testset "constructors" begin
+        df = DataFrame()
+        @test columns(df) == Any[]
+        @test index(df) == Index()
+
+        df = DataFrame(Any[CategoricalVector{Union{Float64, Missing}}(zeros(3)),
+                        CategoricalVector{Union{Float64, Missing}}(ones(3))],
```
nalimilan: Incorrect indentation. Please check other places as I've noted similar issues elsewhere.

test/grouping.jl
dmbates: The problem is the name.  Several other `testset`s use a `DataFrame` called `df` which is defined at the module level.  I know this dataframe is not used outside this testset but I am just trying to silence the warning.  In another file I did that by changing the name `df` -> `df1` in the testset and, in your review, you said not to do that.  In this case I added the `global` and you say not to do that.  What do you want done?

nalimilan: Nothing. ;-)

I don't care about these warnings, they will go away in 1.0 anyway. It would be too bad if they prompted us to make our code worse.

dmbates:
I really don't care anymore.  Probably best to delete this branch and have someone else start over from scratch.

nalimilan:
What? Everything is ready, except for the tests you must have accidentally deleted. `git diff -b` ignores whitespaces and will show you the lines which have been removed.

dmbates:
It's not ready because it won't past tests.  There is still the problem of
```julia
describe: Error During Test at /home/bates/.julia/dev/DataFrames/test/dataframe.jl:364
  Test threw exception MethodError(convert, (CategoricalValue{Int64,UInt32}, 1), 0x0000000000006be5)
  Expression: describe_output == describe(df)
  MethodError: Cannot `convert` an object of type Int64 to an object of type CategoricalValue{Int64,UInt32}
  Closest candidates are:
    convert(::Type{T<:(Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R)}, !Matched::T<:(Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R)) where T<:(Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R) at /home/bates/.julia/packages/CategoricalArrays/HgFP/src/value.jl:65
    convert(::Type{S}, !Matched::Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R) where S at /home/bates/.julia/packages/CategoricalArrays/HgFP/src/value.jl:67
    convert(::Type{T}, !Matched::T) where T at essentials.jl:123
    ...
```
which I have been asking you to take a look at for over a week now.

I think the best course is to close this PR, delete the branch and then when you finally do look at this failure on comparing `CategoricalValue` you can fix the other issues.  That way you can be sure they are fixed the way you want them to be.

nalimilan:
I'm in a conference right now, so it's a bit hard to find time for this, but I'll try. Better keep the PR open anyway, anybody can add push additional commits. Deleting the branch would waste all of your work. :-/

nalimilan:
Turns out the failure is due to a small bug in `convert` for `CategoricalValue`. See https://github.com/JuliaData/CategoricalArrays.jl/pull/153. Feel free to merge and tag a new version once CI has run.

ararslan:
https://github.com/JuliaLang/METADATA.jl/pull/15912

nalimilan:
I've restarted CI after tagging version. The failure is fixed, but the result of `describe` still needs to be adjusted.

ararslan:
Could you push the appropriate fix, @nalimilan?

dmbates:
Thanks for the fix in `CategoricalArrays`, @nalimilan Later in that same test script another test failure occurs related to the display of `CategoricalValue{Int64,UInt32}`.  Could you take a look at that, please?

dmbates:
Thanks for reviewing this.

PR #1459, opened 2018-07-19
Add rowwise function

pdeffebach:
Iterates through all rows in a dataframe and makes an array of length `nrow(df)`.


nalimilan:
Now that iteration over `DataFrameRow` yields values, this can be written as `[mean(r) for r in eachrow(df)]`, so I don't think we need this. We could also get rid of `colwise`. See https://github.com/JuliaData/DataFrames.jl/issues/1595.

PR #1466, opened 2018-07-30
Define getproperty for DataFrame only, not AbstractDataFrame

### LINKED ISSUE #1465: Stack overflow when defining getindex for AbstractDataFrame subtypes
Consider this example:
```
julia> using DataFrames

julia> mutable struct DataFrameWrapper <: AbstractDataFrame
           df::DataFrame
       end

julia> Base.getindex(dfw::DataFrameWrapper, col::Symbol) = getindex(dfw.df, col)

julia> dfw = DataFrameWrapper(DataFrame(A=[1,2,3], B=[4,5,6]));

julia> dfw[:A]
ERROR: StackOverflowError:
Stacktrace:
 [1] getindex(::DataFrameWrapper, ::Symbol) at ./REPL[4]:1
 [2] getproperty(::DataFrameWrapper, ::Symbol) at /Users/alex/.julia/packages/DataFrames/WM3yT/src/abstractdataframe/abstractdataframe.jl:239
 ... (the last 2 lines are repeated 39998 more times)
 [79999] getindex(::DataFrameWrapper, ::Symbol) at ./REPL[4]:1
```
The line in DataFrames it's pointing to is
```julia
Base.getproperty(df::AbstractDataFrame, col_ind::Symbol) = getindex(df, col_ind)
```
I am not (knowingly, at least) calling `getproperty` anywhere in the above code. I thought maybe this was telling me I needed to define a `getproperty` method. Turns out that didn't do it either:
```
julia> Base.getproperty(dfw::DataFrameWrapper, col::Symbol) = getproperty(dfw.df, col)

julia> dfw[:A]
ERROR: StackOverflowError:
Stacktrace:
 [1] getproperty(::DataFrameWrapper, ::Symbol) at ./REPL[7]:1 (repeats 80000 times)
```

ararslan:
Fixes #1465.

This is incredibly annoying and difficult to debug for `AbstractDataFrame` subtypes that contain a `DataFrame` as a field, as in the linked issue. Rather than forcing everyone to embrace this `.column` access, we should limit it to `DataFrame` (edit: and `SubDataFrame`) only and let other `AbstractDataFrame`s decide whether they want to do it.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -236,11 +236,6 @@ Compat.axes(df, i) = axes(df)[i]
 
 Base.ndims(::AbstractDataFrame) = 2
 
-Base.getproperty(df::AbstractDataFrame, col_ind::Symbol) = getindex(df, col_ind)
-Base.setproperty!(df::AbstractDataFrame, col_ind::Symbol, x) = setindex!(df, x, col_ind)
-# Private fields are never exposed since they can conflict with column names
-Base.propertynames(df::AbstractDataFrame, private::Bool=false) = names(df)
-
 ##############################################################################
 ##
 ## Similar
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -626,6 +626,17 @@ Base.setindex!(df::DataFrame, x::Nothing, col_ind::Int) = delete!(df, col_ind)
 
 ##############################################################################
 ##
+## getproperty/setproperty!
+##
+##############################################################################
+
+Base.getproperty(df::DataFrame, col_ind::Symbol) = getindex(df, col_ind)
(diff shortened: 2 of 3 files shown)


vtjnash:
I think if `.`-access is an approved access API for DataFrame-like objects, then doing this as a getproperty overload on the abstract type make sense as an appropriate choice.

ararslan:
I don't think anyone ever said that it's an "approved API" for _all_ `DataFrame`-like objects. That would be incredibly unfortunate.

nalimilan:
I agree with @vtjnash that it should be defined on `AbstractDataFrame`, since it's part of the API just like `getindex`. It would be weird to define it for all concrete `AbstractDataFrame` types we define in the package (which we do now), but not make it part of the API.

bkamins:
I would close it.

nalimilan:
Sorry @ararslan! ;-)

Issue #1474, opened 2018-08-09
Explicitly add conversion to Matrix


kescobo:
Constructors no longer fall back to `convert()` on 0.7, so `Matrix(df)` now throws an error. 

julia0.6:

```
julia> using DataFrames

julia> df = DataFrame(a=[1,2], b=[3,4])
2×2 DataFrames.DataFrame
│ Row │ a │ b │
├─────┼───┼───┤
│ 1   │ 1 │ 3 │
│ 2   │ 2 │ 4 │

julia> Matrix(df)
2×2 Array{Int64,2}:
 1  3
 2  4
```

julia0.7

```
julia> using DataFrames

julia> df = DataFrame(a=[1,2], b=[3,4])
2×2 DataFrame
│ Row │ a │ b │
├─────┼───┼───┤
│ 1   │ 1 │ 3 │
│ 2   │ 2 │ 4 │

julia> Matrix(df)
ERROR: MethodError: no method matching Array{T,2} where T(::DataFrame)
#...
```

This can be replaced with `convert(Matrix, df)`, but it might be worth adding this functionality back (as discussed with @nalimilan on Slack).

From the REPL, it seems like it might be as easy as 

```
julia> import Base.Matrix

julia> @inline Matrix(df::DataFrame) = convert(Matrix, df)
Array{T,2} where T

julia> Matrix(df)
2×2 Array{Int64,2}:
 1  3
 2  4
```

If so, I'm happy to submit a PR if that would be helpful (I'd just want to know what the best place in the code base for the method would be). But if it's more complicated it's perhaps more efficiently handled by someone that knows more.

bkamins:
I think it is OK to add `Matrix` and `Matrix{T}` as they are both possibly convenient. I would just add them next to `convert` methods.

nalimilan:
A PR would be welcome. No need for `@inline` though.

kescobo:
>I think it is OK to add Matrix and Matrix{T} as they are both possibly convenient. I would just add them next to convert methods.

@bkamins is the latter necessary? Just now in 0.7 I did:

```
julia> using DataFrames
[ Info: Recompiling stale cache file /Users/kev/.julia/compiled/v0.7/DataFrames/AR9oZ.ji for DataFrames [a93c6f00-e57d-5684-b7b6-d8193f3e46c0]
┌ Warning: `using A.B` will only be allowed for modules, not single bindings. Use `using A: B` instead
│   caller = ip:0x0
[...]

PR #1475, opened 2018-08-09
Matrix conversion method

kescobo:
Addresses #1474.

There are a few failing tests and a bunch of deprecation warnings, but a quick glance suggests that they're the same as on Master.

Please let me know if there's anything else I can do.

cc @nalimilan @bkamins

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -721,7 +721,7 @@ function Base.convert(::Type{Matrix{T}}, df::AbstractDataFrame) where T
     return res
 end
 Base.Matrix(df::AbstractDataFrame) = Base.convert(Matrix, df)
-Base.Matrix{T}(df::AbstractDataFrame) where T = Base.convert(Matrix{T}, df)
+Base.Matrix{T}(df::AbstractDataFrame) where {T} = Base.convert(Matrix{T}, df)
 
 """
 Indexes of duplicate rows (a row that is a duplicate of a prior row)
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -387,8 +387,8 @@ describe(df::AbstractDataFrame; stats = [:mean, :min, :median, :max, :nmissing, 
 **Arguments**
 
 * `df` : the AbstractDataFrame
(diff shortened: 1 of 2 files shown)


kescobo:
Odd - it looks like my text editor got cute and removed a bunch of spaces at the end of lines. let me know if you'd like me to strip out those changes.

src/DataFrames.jl
```diff
@@ -8,7 +8,7 @@ module DataFrames
 
 using Reexport, StatsBase, SortingAlgorithms, Compat, Statistics, Unicode, Printf
 @reexport using CategoricalArrays, Missings
-using Base.Sort, Base.Order
+using Base.Sort, Base.Order, Base.Matrix
```
bkamins: I do not think you need this line

src/DataFrames.jl
```diff
@@ -57,6 +57,7 @@ export AbstractDataFrame,
        head,
        tail,
        permutecols!,
+       Matrix,
```
bkamins: Similar with this line.

bkamins:
It would probably be better to split removal of the spaces to another commit but it is good that it is cleaned up 😄.

src/DataFrames.jl
kescobo: Right you are.

kescobo:
@bkamins is there an easy way to go back and move it to a new commit? I can start over and just blow away the current commits (not like there's a lot of work involved)

bkamins:
You can (I am not a git wizard but this is what I do in such cases):
* revert all 3 commits in git in your local repo using `git reset` - this will still leave them in your working directory (alternatively if you add `--hard` you will remove all changes from your working directory and will have to rewrite the change anew)
* then make two commits - one for the change and one for stripping endline white spaces (here you might need to commit selected lines in a file - git allows for this)
* `git push --force`

But I guess we can leave it as is as if you have problems with this as it is not hugely confusing.

kescobo:
No, I've done stuff like that before, just wasn't sure how you felt about rewriting history :-). Will get that done now.

kescobo:
I think this should be good to go, unless you want me to rebase off of the latest master.

nalimilan:
Thanks!

kescobo:
My pleasure! Thanks for speedy review

PR #1477, opened 2018-08-11
Tables integration

mdavezac:
Adds some glue to inter operate with other tabular formats via JuliaData/Tables.jl#1


.gitignore
```diff
@@ -5,3 +5,4 @@ test/data/bigrams.tsv
 docs/build/
 docs/site/
 .DS_Store
+Manifest.toml
```
bkamins: Do we need this change in this PR? I would focus on adding Tables.jl support only here.

Project.toml
```diff
@@ -0,0 +1,19 @@
+name = "DataFrames"
```
bkamins: Do we need this change in this PR? I would focus on adding Tables.jl support only here.

src/other/tables.jl
```diff
@@ -0,0 +1,11 @@
+using Tables
```
bkamins: I would add a dependency to REQUIRE for the time being.

src/other/tables.jl
```diff
@@ -0,0 +1,11 @@
...
+
+Tables.schema(df::DataFrame) = NamedTuple{Tuple(names(df)), Tuple{eltypes(df)...}}
+Tables.AccessStyle(df::DataFrame) = Tables.ColumnAccess()
+Tables.rows(df::DataFrame) = eachrow(df)
+Tables.columns(df::DataFrame) = df
+
+DataFrame(x::Tables.RowTable) = DataFrame([collect(u) for u in Tables.columns(x)],
+                                          collect(Tables.names(Tables.schema(x))))
+DataFrame(x::Any) = DataFrame([collect(u) for u in Tables.columns(x)],
```
bkamins: Can you please explain the rationale behind this catchall definition? Why it would be required (a general type might not support Tables.jl interface)?

test/tables.jl
```diff
@@ -0,0 +1,42 @@
+module TestTables
```
bkamins: Could you please add tests for the following things (also in combination):

* columns containing `missing` like `[1, missing]`;
* columns having a mixed type, e.g. `Union{String, Char, Missing}["A", 'a', missing]`, `Union{String, Char, Missing}["A", "B", "C"]`

I would think it would be good to test against three things:

* if we have proper column names (this is done I think);
* if we have proper values;
* if we have proper column types after conversion and back.

Also it would be good to test conversions to `rowtable` and `columntable` in both ways.

src/other/tables.jl
bkamins: OTOH I would add `Tables.ColumnTable` constructor.

bkamins:
And could you also please add docstrings (or improve existing docstrings) for new methods?

src/other/tables.jl
nalimilan: I agree this duck-typing doesn't sound ideal. Would there be a trait to ensure we only try to do this on structures for which it makes sense? Otherwise I'm afraid it will throw confusing errors for users.

test/tables.jl
```diff
@@ -0,0 +1,42 @@
...
+    @testset "Row-style" begin
+        bare_rows = rowtable(df)
+        for (actual, expected) in zip(bare_rows, eachrow(df))
+            @test actual.a == expected.a
+            @test actual.b == expected.b
+        end
+
+        and_back = DataFrame(rowtable(df))
+        @test and_back isa DataFrame
+        @test all(DataFrames.names(and_back) .== (:a, :b))
```
nalimilan: Just `names(and_back) == [:a, :b]`? Same below.

src/other/tables.jl
bkamins: That is what I am trying to convince to @quinnj here: https://github.com/JuliaData/Tables.jl/issues/5.

mdavezac:
thanks, I'll take a look. It might take a few weeks for me to get around to this.

quinnj:
I'll go ahead and pick this up and work on it; w/ Tables.jl registered today, I'm eager to get things rolling here.

.gitignore
quinnj: This was just to help spread the fun at the JuliaCon hackathon; I'll make sure to remove in the final commit.

src/other/tables.jl
quinnj: This is really meant to replace the type-piracy currently in [IterableTables.jl](https://github.com/queryverse/IterableTables.jl/blob/45f95e18023058410434f21ed3faa7db66364f3d/src/integrations/dataframes.jl#L20) that defines it's own DataFrames catchall. I think we can come up w/ a simple trait to make this all work nicely too.

quinnj:
Ok, I've addressed all the current feedback in https://github.com/JuliaData/DataFrames.jl/pull/1495

PR #1479, opened 2018-08-11
fix show for DataFrameRow

bkamins:
Fix implementation of `show` for `DataFrameRow` which threw an error when `nothing` was passed. Now nothing (empty string) is printed - the same as when displaying `DataFrame`.


bkamins:
Sure - I will update the other places 😄. I have fought to separate `something` form `coalesce` in Base so I agree of course (simply I have not used it enough to have it in muscle memory). Thank you. I merge it then.

PR #1482, opened 2018-08-15
Fix a bug in unstack(df)

bkamins:
Currently `unstack(df)` defaults to `:id` being a rowkey. This is not documented, and actually against documentation (which says that we take all columns except `colkey` and `value` as `rowkeys`) and not intuitive (why `:id` should be special?). Therefore I fix it as a bug.

However, this is breaking as in some cases a resulting data frame will contain more columns than before (it is not very painful I guess because we produce more not less but still breaking).

Therefore we might also go through deprecation period (or change the behavior but print a warning if we produce more than `:id` column in this case). Not sure what to do

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -323,7 +323,7 @@ function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Symbol},
     hcat(df1, df2)
 end
 
-unstack(df::AbstractDataFrame) = unstack(df, :id, :variable, :value)
+unstack(df::AbstractDataFrame) = unstack(df, :variable, :value)
 
 
 ##############################################################################
--- a/test/data.jl
+++ b/test/data.jl
@@ -209,7 +209,8 @@ module TestData
         d1us3 = unstack(d1s2, :variable, :value)
         @test d1us[:a] == d1[:a]
         @test d1us2[:d] == d1[:d]
-        @test d1us2[:3] == d1[:d]
+        @test d1us2[6] == d1[:d]
+        @test d1us2 == d1us3
 
         # test unstack with exactly one key column that is not passed
         df1 = melt(DataFrame(rand(10,10)))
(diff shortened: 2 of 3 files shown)


nalimilan:
Good catch. With this change, the new behavior matches the docstring?
> * `rowkeys` : the column(s) with a unique key for each row, if not given,
  find a key by grouping on anything not a `colkey` or `value`

I'm not sure which behavior is the best one, I guess I'd look at what tidyr/Pandas do. Anyway a deprecation would be nice.

bkamins:
This is exactly the point that the proposed implementation matches documentation.

It works exactly like `spread` in `dplyr` (which actually always uses all columns and does not allow you to select a subset AFAIK).

In Pandas you `unstack` an index so this is not directly comparable.

Regarding deprecation I can do two things:

do a standard deprecation: do not change the behavior but do not change the functionality.

or

change the behavior (as I would say it is a bug) but print a warning that the behavior has changed if there is more than one `rowkeys` column or if `id` column does not exist (currently if `id` column does not exist `unstack(df)` throws an error).

Also note that now `usntack(df)` and `unstack(df, :variable, :value)` produce a different result and after this PR they will be equivalent. That is why I say that current `unstack(df)` behavior seems to be a bug.

nalimilan:
OK, cool. However, I think we should first deprecate the current behavior, and only in a second phase change it. That's more painful, but now that Julia 1.0 has been released we should be really careful when breaking compatibility. You can prepare both PRs at the same time so that it's easy to make the change later.

nalimilan:
> BTW: we currently have many deprecations (e.g. with makeunique) - when do you think we should actually make the transition with the functionality?

In theory they should probably be removed in the following major release, but in practice this would have been 0.12.0 (since before 1.0 we use the minor release number for that). I guess we could remove deprecations at any time if we want, at least for those which prevent us from changing the behavior. OTOH there's no harm in keeping deprecations which do not block anything.

bkamins:
Closing this as we have #1629 that is properly rebased.

PR #1487, opened 2018-08-18
fixed overzealous type assertion

ExpandingMan:
... that was causing failures on 32-bit machines.  See discussion and appveyor link [here](https://github.com/JuliaLang/METADATA.jl/pull/17015).

This problem seems a little silly at first glance, so I fear I may be missing some part of what's going on here.

If this is indeed a proper fix, ideally it should be applied all the way back to 0.11, though to be honest I'm not sure what the easiest way of doing that would be (we could tag off new branches, but that seems somehow excessive).

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -225,7 +225,7 @@ index(df::DataFrame) = getfield(df, :colindex)
 columns(df::DataFrame) = getfield(df, :columns)
 
 # TODO: Remove these
-nrow(df::DataFrame) = ncol(df) > 0 ? length(columns(df)[1])::Int : 0
+nrow(df::DataFrame) = ncol(df) > 0 ? length(columns(df)[1])::Integer : 0
 ncol(df::DataFrame) = length(index(df))
 
 ##############################################################################
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -225,7 +225,7 @@ index(df::DataFrame) = getfield(df, :colindex)
 columns(df::DataFrame) = getfield(df, :columns)
 
 # TODO: Remove these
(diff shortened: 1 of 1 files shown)


ExpandingMan:
The only reason I didn't remove it completely is that it seemed strange to me that it was there in the first place, so I thought perhaps maybe there was some bizarre reason for it.

As far as I can tell, the thing that was breaking was not that `nrow` was returning the wrong type, but rather that the `length` function retruned a `Int64` and was failing that assertion.

ExpandingMan:
Honestly I don't know for sure if this will fix Feather.jl.  It was failing on a 32-bit windows (:fearful:) appveyor test.  That's not something I can test here in 64-bit linux world.

Regardless, this type assertion seemed dangerous to me, so I thought I'd remove it hoping that resolves the issue, and hopefully make DataFrames.jl a tiny, tiny bit more stable in the process.

Of course, it would be crazy to tag something over this, so I don't know what we are going to do about the tests. We'll have to see if the other maintainers are that worried about failing Windows tests.

bkamins:
Actually we should reopen it as without the assertion `nrow` is not type stable as compiler has problem with type inference here.

We should have something like:

```
nrow(df::DataFrame)::Int = ncol(df) > 0 ? length(columns(df)[1]) : 0
```

unless Feather.jl has some other requirements here. @ExpandingMan what should be a proper return type?

ExpandingMan:
Probably the simplest and safest thing to do would be
```julia
nrow(df::DataFrame)::Int = ncol(df) > 0 ? Int(length(columns(df)[1])) : zero(Int)
```

bkamins:
I think we do not need both - either return type annotation on annotations in body should be enough.

Would you care to make a PR fixing this? And the lesson learned is to put `@inferred` for `nrow` in tests (so we are sure all is OK).

Thanks!

PR #1489, opened 2018-08-18
add DataFrame constructor accepting a DataFrame

bkamins:
Now when constructors do not fall back to conversion we need to define `DataFrame(df::DataFrame)` constructor.

In general it should return a copy of a `DataFrame` but it is a breaking change, so I guess we should first keep the old behavior and return the same object and throw a deprecation - right?


ExpandingMan:
That does seem to be consistent with the behavior of `Array` (`convert` gives the same object and `Array` gives a copy), so I like it. :+1:

PR #1491, opened 2018-08-26
Bump julia version in docs/make.jl

mortenpi:
The documentation hasn't actually deployed to gh-pages since 0.6 was dropped on Travis.


mortenpi:
Note that I did not update the version selector (`versions.js`) in #1492, so it might be best to merge that before this. Once this is merged, the `master` build will sort of the version selector.

PR #1493, opened 2018-08-29
revenge of the overzealous type assertion

ExpandingMan:
This form should be both safe and unambiguous to the compiler.

@bkamins

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -231,8 +231,8 @@ index(df::DataFrame) = getfield(df, :colindex)
 columns(df::DataFrame) = getfield(df, :columns)
 
 # TODO: Remove these
-nrow(df::DataFrame) = ncol(df) > 0 ? length(columns(df)[1]) : 0
-ncol(df::DataFrame) = length(index(df))
+nrow(df::DataFrame)::Int = ncol(df) > 0 ? Int(length(columns(df)[1])) : zero(Int)
+ncol(df::DataFrame)::Int = Int(length(index(df)))
 
 ##############################################################################
 ##


bkamins:
Thanks!. Can you add `@inferred` in tests for both functions. And I think that we do not need to doubly make sure all is OK (especially if we test it with `@inferred`).

bkamins:
I have tested that it is enough to do:

```
nrow(df::DataFrame)::Int = ncol(df) > 0 ? length(columns(df)[1]) : 0
ncol(df::DataFrame)::Int = length(index(df))
```

so maybe let us use this form?

The reason is that type assertion in a method definition translates to two operations:
1. conversion
2. assertion

Here is an example:
```
julia> f(x)::Char = x

julia> f(100)
'd': ASCII/Unicode U+0064 (category Ll: Letter, lowercase)

julia> @code_lowered f(100)
CodeInfo(
1 1 ─ %1 = (Base.convert)(Main.Char, x)                                                                                              │
  │   %2 = (Core.typeassert)(%1, Main.Char)                                                                                          │
  └──      return %2                                                                                                                 │
)
```

(for `ncol` it is not strictly needed now but I guess we can have it just to indicate the expected return type)

ExpandingMan:
Cool, I hadn't realized that assertions on function signatures were also conversions.

nalimilan:
I'm a bit late, but I'm not really convinced by the changes made at https://github.com/JuliaData/DataFrames.jl/pull/1487 and here. `length(::Vector)` will always return `Int`, so I don't see how it could trigger a failure in Feather.jl. The new code is less efficient since it needs to handle the possibility of a conversion. Also, `zero(Int)` is just `0`. Can we go back to the old version (keeping the new tests)?

bkamins:
You are right. Actually I think that the code is less efficient because it makes a dynamic dispatch to `convert`. So in short we can revert it back to the original. I will push a PR.

PR #1495, opened 2018-08-29
Tables.jl integration

quinnj:
Supersedes @mdavezac's PR #1477


src/other/tables.jl
```diff
@@ -0,0 +1,25 @@
...
+function _isiterabletable(x::T) where {T}
+    Base.isiterable(x) || return false
+    return Base.IteratorEltype(x) === Base.HasEltype() ? eltype(x) <: NamedTuple : false
+end
+
+function DataFrame(x::T) where {T}
+    if Tables.istable(T)
+        DataFrame([collect(u) for u in Tables.columns(x)], collect(Tables.names(Tables.schema(x))))
+    elseif _isiterabletable(x)
+        _DataFrame(x)
```
quinnj: @davidanthoff, I still need to do thorough testing, but I think this might do the trick for IterableTables. Essentially, we first do a check for `Tables.istable` (table types have to explicitly opt-in after https://github.com/JuliaData/Tables.jl/pull/9), then we'll do the `isiterabletable` check and if so, call `_DataFrame` that is defined in IterableTables. That should make it safe to remove the catchall definition from IterableTables.

quinnj: Ok, had to refactor the initial attempt a bit. But it seems to be working now. Companion PR to IterableTables is here: https://github.com/queryverse/IterableTables.jl/pull/84

davidanthoff: I think the ``Tables.istable`` helps a lot!

The ``_isiterabletable`` though won't work. It might seem to work now, but it will break with the next Query.jl release. I have two suggestions: first, I think it would be more efficient to call ``isiterabletable`` from TableTraits.jl directly here. The ``getiterator`` call can be costly and do a lot of stuff, and the ``isiterabletable`` call cann be made directly on ``T``, not the return value of ``getiterator``. It would also force us to come up with a solution for the case where ``isiterabletable`` returns ``missing``, which is the case where we don't know whether it is an iterable table until we start iterating it. That will complicate things, but I don't see any way around it, unfortunately... I think maybe the general strategy would be: if ``isiterabletable`` returns ``true``, everything is easy. If it returns ``false``, also easy, we just go to the ``convert`` call right away. If it returns ``missing``, we next can check whether ``T`` is an iterable with ``IteratorInterfaceExtensions.isiterable``. If that returns ``false``, we go to the ``convert`` case. If that returns ``true``, we will have to start iterating, look at the first element and then decide whether this is an iterable table or not... Ugly, ugly, but I can't think of anything to avoid that.

src/other/tables.jl
```diff
@@ -0,0 +1,32 @@
...
+Tables.istable(::Type{DataFrame}) = true
+Tables.rows(df::DataFrame) = eachrow(df)
+Tables.columns(df::DataFrame) = df
+
+function _isiterabletable(x::T) where {T}
+    Base.isiterable(T) || return false
+    return Base.IteratorEltype(x) === Base.HasEltype() ? eltype(x) <: NamedTuple : false
+end
+
+function DataFrame(x::T) where {T}
```
bkamins: I guess if we decide to add this catch-all definition we should also update docstring of `DataFrame` constructor explaining all the logic that happens here.

src/other/tables.jl
```diff
@@ -0,0 +1,32 @@
...
+    if Tables.istable(T)
+        return DataFrame([collect(u) for u in Tables.columns(x)],
+                          collect(Tables.names(Tables.schema(x))))
+    end
+    y = IteratorInterfaceExtensions.getiterator(x)
+    if _isiterabletable(y)
+        return DataFrame([collect(u) for u in Tables.buildcolumns(eltype(y), y)],
+                          collect(Tables.names(eltype(y))))
+    else
+        convert(DataFrame, x)
```
bkamins: I guess Base moved out from default fallback of constructors to `convert`. I do not know what would be best - trying to `convert` or throw an error immediately.

If we want to go with trying a conversion then in particular in general `convert` could create something else than a `DataFrame` so maybe it would be cleaner to return `x` here and above add type assertion:
```
function DataFrame(x::T)::DataFrame where {T}
```

nalimilan: `convert(DataFrame, x)` must always return a `DataFrame`, but I'm not sure why we would call `convert` from the constructor, since Base has removed that fallback. Types should define their custom constructors if needed.

nalimilan: Yes, but people aren't supposed to do that so we don't care what happens if they do. :-)

bkamins: I agree - but I just want to say, that if we want to allow it it is better to do return type assertion I think.

But as you have said - maybe we should error then and not try doing conversion.

src/other/tables.jl
```diff
@@ -0,0 +1,48 @@
...
+            return DataFrame([collect(u) for u in columns],
+                        collect(Tables.schema(columns).names))
+        end
+    end
+    throw(ArgumentError("unable to construct DataFrame from $T"))
+end
+
+Base.append!(df::DataFrame, x) = append!(df, DataFrame(x))
+
+# This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
```
bkamins: Actually the conflicting constructor was handling this case in deprecated mode for long enough that I think given this change we should finish deprecation period and restrict that other constructor signature to allow only vectors of vectors.

src/other/tables.jl
```diff
@@ -0,0 +1,48 @@
...
+    y = IteratorInterfaceExtensions.getiterator(x)
+    yT = typeof(y)
+    if Base.isiterable(yT)
+        if Base.IteratorEltype(yT) === Base.HasEltype() && eltype(y) <: NamedTuple
+            columns = Tables.buildcolumns(Tables.Schema(eltype(y)), y)
+            return DataFrame([collect(u) for u in columns],
+                        collect(Tables.schema(columns).names))
+        else
+            # non-NamedTuple or UnknownEltype
+            columns = Tables.buildcolumns(Tables.Schema(propertynames(first(x)), nothing), y)
```
davidanthoff: Hm, I think the call to ``first`` here is quite problematic. For some query operators, iterating the first element is costly (i.e. the whole query essentially is executed for that first iteration), so we really want to avoid calling ``iterate`` twice on a query. I think that would happen here: it would once get called for the ``first`` call, and then again later for the materialization, right? I think with this implementation, essentially for many queries we would essentially execute them twice to materialize them into a ``DataFrame``...

davidanthoff:
Oh, and one meta comment: this all seems quite involved now... Just for comparison, the model I have with TableTraits.jl/TableTraitsUtils.jl (where there are some separate helper functions that make it easier to consume things) now looks like this for the entire DataFrames.jl/TableTraits.jl integration:
````julia
# Source

IteratorInterfaceExtensions.isiterable(x::DataFrames.DataFrame) = true
TableTraits.isiterabletable(x::DataFrames.DataFrame) = true

function IteratorInterfaceExtensions.getiterator(df::DataFrames.DataFrame)
    return TableTraitsUtils.create_tableiterator(getfield(df, :columns), names(df))
end

# Sink

function DataFrames.DataFrame(x::AbstractVector{T}) where {T<:NamedTuple}
    cols, names = create_columns_from_iterabletable(x, na_representation=:missing)

    return DataFrames.DataFrame(cols, names)
end

function DataFrames.DataFrame(x)
    cols, names = create_columns_from_iterabletable(x, na_representation=:missing)

    return DataFrames.DataFrame(cols, names)
end
````
It seems to me that a lot of the code that is in the constructor of ``DataFrames`` here in this PR right now is boilerplate code that could be moved to a similar helper function as the ``create_tableiterator`` that I have in TableTraitsUtils.jl.

File: src/other/tables.jl (hunk 5 above)
Comment (nalimilan) [reply src/other/tables.jl]: I also noticed this at https://github.com/JuliaData/Tables.jl/pull/10#discussion_r214523261.

**Comment by quinnj** [comment]
Alright, I've cleaned up the PR here again; I'll close/re-open once the new [Tables.jl tag](https://github.com/JuliaLang/METADATA.jl/pull/17752) is merged so we can get a clean signal. I think this is getting in pretty good shape with the new support for unknown schemas in Tables.jl.

File: docs/src/man/getting_started.md [hunk 6]
```diff
@@ -116,6 +116,27 @@ Note that constructing a `DataFrame` row by row is significantly less performant
...
+interacting with tabular data. This means that a DataFrame can be used as a "source"
+to any package that expects a Tables.jl interface input, (file format packages,
+data manipulation packages, etc.). A DataFrame can also be a sink for any Tables.jl
+interface input. Some example uses are:
+
+```julia
+df = DataFrame(a=[1, 2, 3], b=[:a, :b, :c])
+
+# write DataFrame out to csv file
+df |> CSV.write("dataframe.csv")
```
nalimilan: Should add `using CSV, SQLite, Query`.

BTW, in the paragraph above, no comma before parenthesis.

src/dataframe/dataframe.jl
```diff
@@ -34,6 +35,8 @@ DataFrame(ds::AbstractDict)
 * `categorical` : `Vector{Bool}` indicating which columns should be converted to
                   `CategoricalVector`
 * `ds` : `AbstractDict` of columns
+* `table`: a "catchall" constructor that accepts any type that implements the
```
nalimilan: I wouldn't say "a "catchall" constructor" since you describe an argument rather than a function.

davidanthoff:
I just ran a performance comparison of a simple query with the existing integration in IterableTables.jl and the integration here in this PR/Tables.jl etc. I'm seeing really significant performance regressions on that. For the existing integration code I get:
````
BenchmarkTools.Trial:
  memory estimate:  2.01 MiB
  allocs estimate:  132
  --------------
  minimum time:     1.323 ms (0.00% GC)
  median time:      1.884 ms (0.00% GC)
  mean time:        2.060 ms (7.54% GC)
  maximum time:     50.265 ms (96.13% GC)
  --------------
  samples:          2423
  evals/sample:     1
````
For the new Tables.jl based integration code I get:
````
BenchmarkTools.Trial:                          
  memory estimate:  11.39 MiB                  
  allocs estimate:  599079                     
  --------------                               
  minimum time:     22.590 ms (0.00% GC)       
  median time:      24.612 ms (2.16% GC)       
  mean time:        25.177 ms (3.11% GC)       
  maximum time:     76.341 ms (64.36% GC)      
  --------------                               
  samples:          199                        
  evals/sample:     1                          
````
The code I use to test this is [here](https://github.com/davidanthoff/performance_replication). Essentially you run the ``benchmark_code.jl`` file once in the environment defined in the ``existing`` folder, and then again in the ``new_tables_jl`` folder.

**Comment by quinnj** [comment]
Thanks for the report @davidanthoff; I'll take a look.

File: src/other/tables.jl [hunk 8]
```diff
@@ -0,0 +1,39 @@
...
+fromcolumns(x) = DataFrame(Any[getvector(c) for c in Tables.eachcolumn(x)], collect(propertynames(x)))
+
+function DataFrame(x::T) where {T}
+    if Tables.istable(T)
+        return fromcolumns(Tables.columns(x))
+    end
+    y = IteratorInterfaceExtensions.getiterator(x)
+    yT = typeof(y)
+    if Base.isiterable(yT)
+        Base.depwarn("constructing a DataFrame from an iterator is deprecated; $T should support the Tables.jl interface", nothing)
```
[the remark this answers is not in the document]
nalimilan: I thought the goal was to converge towards a single common interface in Tables.jl?

Anyway it's going to be hard or impossible to support two interfaces given that we cannot dispatch on a particular type: this method will be called as a fallback for all kinds of tables.

src/other/tables.jl
```diff
@@ -0,0 +1,39 @@
...
+getvector(x) = collect(x)
+fromcolumns(x) = DataFrame(Any[getvector(c) for c in Tables.eachcolumn(x)], collect(propertynames(x)))
+
+function DataFrame(x::T) where {T}
+    if Tables.istable(T)
+        return fromcolumns(Tables.columns(x))
+    end
+    y = IteratorInterfaceExtensions.getiterator(x)
+    yT = typeof(y)
+    if Base.isiterable(yT)
```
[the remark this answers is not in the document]
quinnj: It'd be ideal to avoid the dependency on TableTraits.jl. I can't see any difference between what's here and what's in TableTraits?

davidanthoff:
Here is another idea. What if we remove all TableTraits.jl-Tables.jl integration code from Tables.jl, and change the constructor of ``DataFrame`` here to the following:
````julia
function queryversehook(x::T, ignore) where {T}
    throw(ArgumentError("unable to construct DataFrame from $T"))
end

function DataFrame(x::T) where {T}
    if Tables.istable(T)
        return fromcolumns(Tables.columns(x))
    end
    return queryversehook(x, Val(:hook))
end
````
I would then replace the integration code for ``DataFrame`` in IterableTables.jl with this:
````julia
function DataFrames.queryversehook(x::T, ::Val{:hook}) where {T}
    y = create_columns_from_iterabletable(x, na_representation=:missing)
    if y===nothing
        throw(ArgumentError("unable to construct DataFrame from $T"))
    else
        return DataFrames.DataFrame(y[1], y[2])
    end
end
````
Wouldn't that be significantly simpler than the current code? It would also get rid of the type piracy in IterableTables.jl, and instead use an "approved" hook. Yes, that part is not super elegant, but is that so bad?

I see a couple of other benefits: 1) as I'm adding more optional interfaces that speed up collection from queries, we don't have to coordinate that stuff with Tables.jl, but the two interfaces could just evolve independently on their own schedule, 2) we don't have to spend time digging into the performance issues I pointed out above, 3) in general, it seems to me we reduce complexity and dependencies between packages, which always seems like a good idea to me.

Yes, I know it would be giving up on the idea that Tables.jl essentially implements the TableTraits.jl interface under the hood. I was all for that when I though the typical sink implementation would look like
````julia
DataFrame(x) = Tables.columns(x)
````
and it would just handle everything, but the current code in this PR seems quite far from that and seems to introduce quite complex logic when we could have this quite a bit easier?

**Comment by quinnj** [comment]
I'm opposed because I think there are significant maintenance and contributor advantages to having all the code in one place. I've already got a good handle on the performance issue and will have a fix shortly, so no worries there. This code is no more complex than what is currently in IterableTables.jl; you might not be as familiar yet, but there's really not anything crazy going on here; happy to help clarify anything that doesn't seem to make sense or be overly complex.

File: src/other/tables.jl [hunk 10]
```diff
@@ -0,0 +1,39 @@
...
+Tables.columns(df::AbstractDataFrame) = df
+Tables.rowaccess(::Type{DataFrame}) = true
+Tables.rows(df::DataFrame) = eachrow(df)
+
+Tables.schema(df::AbstractDataFrame) = Tables.Schema(names(df), eltypes(df))
+Tables.schema(df::DFRowIterator) = Tables.schema(df.df)
+
+getvector(x::AbstractVector) = x
+getvector(x) = collect(x)
+fromcolumns(x) = DataFrame(Any[getvector(c) for c in Tables.eachcolumn(x)], collect(propertynames(x)))
```
bkamins: Here and below in `DataFrame` method we do not provide `makeunique` argument. This means that if there are duplicate column names the constructor will fail in the future (and in general Tables.jl do not guarantee that column names are unique now if I understand the current design correctly).

Maybe this is what we want, but in other constructors we allow the user decide what to do with `makeunique` keyword argument.

docs/src/man/getting_started.md
ararslan: Please don't advocate piping for such a simple operation in the documentation...

quinnj:
Hmmmm, so the performance issue wasn't quite as straightforward as I originally thought. The quick fix I found was for `columntable`, but w/ that fix, the DataFrames case was still about 10x slower. I spent an hour or two digging in and trying some things, but it turns out we can't quite avoid that pesky `df.columns[col]` allocation. Let me clarify: the approach I was trying avoided any generated functions and just using the plain `eachrow(df)` iterator. I do think the compiler will get there where pulling out those columns will be hoisted and all will be well, but I can't seem to find a way to get that last getindex to inline + hoist. The solution then is to take the approach that David was taking in TableTraitsUtils, and define `Tables.rows(df::DataFrame) = Tables.rows(columntable(df))`, which brings performance back to what's expected. I've pushed that change here.

src/other/tables.jl
```diff
@@ -0,0 +1,39 @@
...
+    end
+    throw(ArgumentError("unable to construct DataFrame from $T"))
+end
+
+Base.append!(df::DataFrame, x) = append!(df, DataFrame(x))
+
+# This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
+DataFrame(x::Vector{T}) where {T <: NamedTuple} = fromcolumns(Tables.columns(x))
+
+IteratorInterfaceExtensions.getiterator(df::DataFrame) = Tables.datavaluerows(df)
```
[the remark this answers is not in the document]
quinnj: Added the `isiterable` overload

src/other/tables.jl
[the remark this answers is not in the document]
quinnj: I agree that we don't need to be so hasty in deprecating, especially with Tables.jl not even used anywhere yet. I'll comment out the depwarn and we can see how things play out over the next little while; it would certainly be annoying for lots of users to see that depwarn without anyway to get rid of it.

src/other/tables.jl
quinnj: I'm open to suggestions here; how about we go w/ this for now and address feedback as it comes in?

bkamins: @nalimilan - any opinion? I can live with what we have now as adding `makeunique` keyword argument later is non breaking.

The only scenario when I find it is really relevant when you create a `DataFrame` from a file on a disk that has duplicate column names and want to pipe the data through some interface that supports Tables.jl interface and then you have no way to allow `DataFrame` creation with renamed columns using this API.

nalimilan: It would indeed be nice to support `makeunique` as for other constructors. That should be really simple: `makeunique` just has to be passed to the lower-level `DataFrame` constructor.

quinnj: I've added support for `makeunique`.

src/dataframe/dataframe.jl
quinnj: Fixed.

src/other/tables.jl
```diff
@@ -9,23 +9,28 @@ Tables.rows(df::DataFrame) = Tables.rows(columntable(df))
...
+    return v
+end
+
 getvector(x::AbstractVector) = x
 getvector(x) = collect(x)
-fromcolumns(x) = DataFrame(Any[getvector(c) for c in Tables.eachcolumn(x)], collect(propertynames(x)))
+fromcolumns(x, makeunique) = DataFrame(Any[getvector(c) for c in Tables.eachcolumn(x)], collect(propertynames(x)); makeunique=makeunique)
 
-function DataFrame(x::T) where {T}
+function DataFrame(x::T; makeunique::Bool=false) where {T}
```
nalimilan: The docstring should also be updated.

src/other/tables.jl
nalimilan: Let's discuss this at https://github.com/JuliaData/DataFrames.jl/pull/1308.

src/other/tables.jl
quinnj: Updated.

quinnj:
In terms of timeline here, I'd like @davidanthoff to chime in on his status upgrading Queryverse to 1.0. I definitely don't want to push a release that might be disruptive right as he's releasing his stuff (I think it's close?). In any case, we'll want to coordinate the merge of this PR & the [IterableTables.jl](https://github.com/queryverse/IterableTables.jl/pull/84) PR to ensure anyone tracking masters sees a smooth transition.

If everyone is ok, I'd like to merge into master soonish and then hold off on making a new release for a week or two. Having the changes in master allows easier development of related packages and for sorting out any tweaks without pushing it on everyone. For reference w/ Tables.jl, we're up to about 6-7 packages implementing the interface, w/ most of the implementations in PRs.

davidanthoff:
I've chimed in on the next steps at https://github.com/JuliaData/Tables.jl/issues/7, might be easier to discuss there? Long story short, the Tables.jl/TableTraits.jl integration right now is not right, and I wouldn't want to release the current version on the world.

src/other/tables.jl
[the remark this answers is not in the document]
davidanthoff: Lots of sources in the Queryverse add a method to ``TableTraits.isiterabletable``, and you won't catch that with your own implementation.

Would it be so bad to take a dependency on TableTraits.jl? The package is specifically designed to not cause any problems, the whole code is https://github.com/queryverse/TableTraits.jl/blob/master/src/TableTraits.jl. I'm also making a guarantee that a) I will never ever make a breaking release for that package (unless julia 2.0.0 forces me to) and b) I won't take any further dependencies on. I can (maybe, a little bit) understand a reluctance to take a dependency on DataValues.jl, but TableTraits.jl I think is harmless. The package also has a long track record of not breaking anything.

I'm also realizing that the call to ``IteratorInterfaceExtensions.getiterator(x)`` is probably not what you want here. In general, the sequence ``y = IteratorInterfaceExtensions.getiterator(x); yT = typeof(y); Base.isiterable(yT)`` should be written as ``IteratorInterfaceExtensions.isiterable(x)`` for a much more efficient story. But even that is not really good, because really the way to properly hook into the Querverse there is to use ``TableTraits.isiterabletable``.

src/other/tables.jl
```diff
@@ -0,0 +1,45 @@
...
+    throw(ArgumentError("unable to construct DataFrame from $T"))
+end
+
+Base.append!(df::DataFrame, x) = append!(df, DataFrame(x))
+
+# This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
+DataFrame(x::Vector{T}; makeunique::Bool=false) where {T <: NamedTuple} = fromcolumns(Tables.columns(x), makeunique)
+
+IteratorInterfaceExtensions.getiterator(df::DataFrame) = Tables.datavaluerows(df)
+IteratorInterfaceExtensions.isiterable(x::DataFrame) = true
```
davidanthoff: I think we should move ``TableTraits.isiterabletable(x::DataFrames.DataFrame) = true`` here as well. Right now it still lives in ``IterableTables.jl``, but I think if we generally move the TableTraits.jl integration to here and Tables.jl, we should just remove it entirely from IterableTables.jl. Otherwise the integration will be strung out over Tables.jl, IterableTables.jl and here, which seems a bit convoluted.

quinnj: Ok, moved `TableTraits.isiterabletable` definition here.

src/other/tables.jl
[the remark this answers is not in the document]
quinnj: Ok, I'm not quite sure I understand why doing `TableTraits.isiterabletable` explicitly is really better than what we had (which was doing the exact same fallback checks); are there queryverse sources that overload `isiterabletable` but wouldn't satisfy the fallback definition? Doesn't that seem troubling from a "clean interface" perspective? Anyway, not really the point here.

I've added the TableTraits.jl dependency; I agree that it's a very small dependency and simplifies things a bit w/ the queryverse integration. That said, the feeling among DataFrames admins is that DataFrames should really only need to support a single table interface long-term. The hope is that Tables.jl support continues to grow and given that it would be about 4 lines of code for any queryverse source to implement the interface, it would be great to be able to standardize.

So with this PR, DataFrames will have dependencies on DataStreams, Tables, and TableTraits, but the long-term goal is that only Tables.jl will be needed.

quinnj:
Alright, I think this is ready to go: @davidanthoff, can we coordinate a merge of https://github.com/queryverse/IterableTables.jl/pull/84 w/ here? Ideally, we could merge these soon (today/tomorrow) and let things sit on master for a bit before doing releases.

src/other/tables.jl
```diff
@@ -0,0 +1,39 @@
...
+        # non-NamedTuple or EltypeUnknown
+        return fromcolumns(Tables.buildcolumns(nothing, Tables.DataValueUnwrapper(y)), makeunique)
+    end
+    throw(ArgumentError("unable to construct DataFrame from $T"))
+end
+
+Base.append!(df::DataFrame, x) = append!(df, DataFrame(x))
+
+# This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
+DataFrame(x::Vector{T}; makeunique::Bool=false) where {T <: NamedTuple} = fromcolumns(Tables.columns(x), makeunique)
```
davidanthoff: There is one test in the old IterableTables.jl integration story that now fails, namely [this](https://github.com/queryverse/IterableTables.jl/blob/45f95e18023058410434f21ed3faa7db66364f3d/test/test_integration_dataframes.jl#L42) one here. That is because this method here doesn't hook the datavalues->missing conversion story. Could that be hooked in, if DataValues.jl is loaded?

quinnj: Ah, good catch. The failure is related to the fact that `T = Vector{NamedTuple{(:a, :b), Tuple{DataValue{Int}, DataValue{Float64}}}}` is `Tables.istable(T)`, so it's actually good we have a separate method to catch this case so we can wrap it in `Tables.DataValueUnwrapper`. I've pushed a fix.

bkamins:
@quinnj Can you please comment what was the reason behind removing `makeunique`?

nalimilan:
The idea is that Tables.jl now requires unique names, so it cannot have any effect here.

quinnj:
@nalimilan is right, and sorry for not commenting when I pushed the commit. The Tables.jl interface now requires column names to be unique, so this shouldn't be an issue.

quinnj:
Hmmm, a lot of CI flakiness here; I've restarted the travis jobs, but it looks like @nalimilan would have to restart the appveyor one (though I'm not sure a) it's worth it since it was only one build that failed and 2) if appveyor even let's you restart a single build).

KristofferC:
So If I have a Table that is an `AbstractVector` then I hit this constructor

https://github.com/JuliaData/DataFrames.jl/blob/ef7233aee2d816ea3922945107d0024c1bbfcde5/src/dataframe/dataframe.jl#L158

which doesnt work.

What to do?

quinnj:
Wait for https://github.com/JuliaData/DataFrames.jl/pull/1527 to be merged

PR #1500, opened 2018-09-01
Add eltype to eachrow and eachcol

bkamins:
It might help inference in some cases. For `eltype` of `eachcol` I left the second element of the tuple to be `Any` as this is what we have in `DataFrame` implementation, but maybe `AbstractVector` would be better (the performance should be the same though I think).

### DIFF
--- a/test/iteration.jl
+++ b/test/iteration.jl
@@ -9,6 +9,7 @@ module TestIteration
 
     @test size(eachrow(df)) == (size(df, 1),)
     @test eachrow(df)[1] == DataFrameRow(df, 1)
+    @test typeof(collect(eachrow(df))) == Vector{DataFrameRow{DataFrame}}
     for row in eachrow(df)
         @test isa(row, DataFrameRow)
         @test (row[:B] - row[:A]) == 1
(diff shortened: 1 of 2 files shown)


test/iteration.jl
```diff
@@ -9,6 +9,7 @@ module TestIteration
 
     @test size(eachrow(df)) == (size(df, 1),)
     @test eachrow(df)[1] == DataFrameRow(df, 1)
+    @test typeof(collect(eachrow(df))) == Vector{DataFrameRow{DataFrame}}
```
nalimilan: AFAIK this test should pass even without the change, right? Anyway it would be nice to test `eltype` too. BTW, you can use `isa` for a more compact code.

bkamins: No it would fail. Both `eachrow` and `eachcol` now produce `Vector{Any}`. I will correct the style.

bkamins: I have also corrected one more line which used `==` for testing type and I was following that style.

test/iteration.jl
```diff
@@ -9,16 +9,20 @@ module TestIteration
 
     @test size(eachrow(df)) == (size(df, 1),)
     @test eachrow(df)[1] == DataFrameRow(df, 1)
+    @test collect(eachrow(df)) isa Vector{DataFrameRow{DataFrame}}
+    @test eltype(collect(eachrow(df))) == DataFrameRow{DataFrame}
```
nalimilan: This test is redundant with the previous one, I meant calling `eltype(eachrow(df))`.

bkamins: OK - now I get it. Will fix it.

PR #1505, opened 2018-09-05
only allow vector of vectors in DataFrame constructor

bkamins:
To complement #1495 I propose to finish deprecation error of vector of vectors constructor.

Now what Tables.jl integration proposed works like this:
```
julia> DataFrame([(a=1,b=2), (a=3, b=4)])
┌ Warning: passing columns argument with non-AbstractVector entries is deprecated
│   caller = top-level scope at none:0
└ @ Core none:0
1×2 DataFrame
│ Row │ x1             │ x2             │
├─────┼────────────────┼────────────────┤
│ 1   │ (a = 1, b = 2) │ (a = 3, b = 4) │
```
so the change in #1495 is breaking (although probably no one would do this 😄).

After this the above will be disallowed and Tables.jl will fill this gap.


bkamins:
Agreed

PR #1509, opened 2018-09-15
identifier function and related code to deprecated

bkamins:
This PR cleans up some deprecated functionality and moves it to deprecated section.

@nalimilan also `colmissing` and `countmissing` functions are not used in DataFrames.jl except one place in tests and they are not exported. Do we want to keep them?


bkamins:
The drop in CI is purely due to moving code around.

nalimilan:
> @nalimilan also `colmissing` and `countmissing` functions are not used in DataFrames.jl except one place in tests and they are not exported. Do we want to keep them?

Not, let's remove them if they aren't used.

bkamins:
CI failures seem unrelated.

src/deprecated.jl
```diff
@@ -1,3 +1,4 @@
+import Base: isidentifier, is_id_start_char, is_id_char
```
nalimilan: I hadn't noticed this. Better fully qualify these, so that we can remove the deprecations separately without having unrelated code at the top of the file.

bkamins: fixed

PR #1510, opened 2018-09-15
implemented string getproperty

ExpandingMan:
I don't see any ambiguity in what the behavior of e.g. `df."abc"` should be so I thought we should have it.  This is also helpful for getting columns with "irregular" symbol characters (such as spaces), whether or not it's a good idea to name columns that way in the first place (it probably isn't).

Thanks to @KristofferC for the suggestion.

(As I think about this a bit more, I am slightly worried it will put us on a slippery slope to allowing `String` for indexing, which I'm not quite so sure is a good idea, so I'm not feeling completely great about this being merged.  Let's see what everyone thinks.)

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -238,6 +238,7 @@ Compat.axes(df, i) = axes(df)[i]
 Base.ndims(::AbstractDataFrame) = 2
 
 Base.getproperty(df::AbstractDataFrame, col_ind::Symbol) = getindex(df, col_ind)
+Base.getproperty(df::AbstractDataFrame, col_ind::AbstractString) = getproperty(df, Symbol(col_ind))
 Base.setproperty!(df::AbstractDataFrame, col_ind::Symbol, x) = setindex!(df, x, col_ind)
 # Private fields are never exposed since they can conflict with column names
 Base.propertynames(df::AbstractDataFrame, private::Bool=false) = names(df)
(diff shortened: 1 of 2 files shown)


pdeffebach:
maybe `df.$x` for escaping? I'm not sure if that is idiomatic anywhere else...

KristofferC:
> And given that they are necessarily slower than symbols, I'm not sure it's a good idea to encourage their use.

Do you have a benchmark where this matters?

nalimilan:
> Do you have a benchmark where this matters?

No. One would need to make the change first to be able to measure that. All I can say is that looking up a string in a dict is slower than looking up a symbol (about 2× slower in a simple test I just did).

It shouldn't matter when doing vectorized operations, which are the only efficient approach currently. But if you iterate over rows, I imagine it could make a big difference. Anyway I'm not saying that should be a decisive factor since these approaches will be slow anyway.

ExpandingMan:
I think we should aim to make iterating over rows better, not worse.  I also ran a couple of simple tests, I've not seen string `Dict` lookups that are twice as slow, but I do see them as being 3/2 as slow.  I think that's a good enough reason for me to not start supporting strings.

In another context my attitude would probably be "let's support strings when it's more convenient (for whatever reason) and the default use should be symbols", but I think for this package in particular it might lead us down a bad path.  I recall seeing a number of discourse posts in which users coming from pandas are trying to use strings to index dataframes.  I am worried that if we allow strings, there will be a sizable subset of users who *always* use strings, and we probably don't want that.

I also wonder if, somewhere down the road, using `Symbol`s will help us take advantage of constant propagation.  Currently it does nothing for us, but at some point we are going to have to figure out how to have some semblance of type stability.  I don't really know whether `Symbol` vs `String` is relevant to that, but I have a feeling it might be.

Anyway, thanks again, this was a nice idea, but I think we should shelve this for now at least.

PR #1511, opened 2018-09-15
Improve docs

nalimilan:
Add documentation about indexing, improve various things. Replace `DataFrames.DataFrame` with `DataFrame`. Use `df.col` instead of `df[:col]` where possible.


docs/src/man/subsets.md
```diff
@@ -1,109 +0,0 @@
-# Subsets
```
nalimilan: This file was missing from the ToC. It's redundant with the new contents, except for `filter` and `filter!`, which don't really like advertising given that they are quite slow currently.

docs/src/man/getting_started.md
```diff
@@ -227,30 +427,4 @@ df = DataFrame(x = 1, y = 2)
 CSV.write(output, df)
 ```
 
-The behavior of CSV functions can be adapted via keyword arguments. For more information, use the REPL [help-mode](http://docs.julialang.org/en/stable/manual/interacting-with-julia/#help-mode) or checkout the online [CSV.jl documentation](https://juliadata.github.io/CSV.jl/stable/).
-
-## Loading a Classic Data Set
```
nalimilan: I've removed this as it feels out of place (it's not used in this page). Other pages include code to load it anyway.

docs/src/man/getting_started.md
```diff
@@ -118,68 +162,223 @@ but for very large `DataFrame`s  this may be a consideration.
...
+   9
+  11
+   ⋮
+ 991
+ 993
+ 995
+ 997
+ 999
+
+julia> df[:A] === df[:, :A]
```
bkamins: This is going to be deprecated (also above), only `==` will pass in the future as we want to be consistent with indexing in Base, where `:` creates a copy.

docs/src/man/joins.md
```diff
@@ -6,14 +6,14 @@ We often need to combine two or more data sets together to provide a complete pi
 julia> using DataFrames
 
 julia> names = DataFrame(ID = [20, 40], Name = ["John Doe", "Jane Doe"])
-2×2 DataFrames.DataFrame
+2×2 DataFrame
```
bkamins: Here and below need to regenerate tables to include column eltype

docs/src/man/missing.md
```diff
@@ -54,9 +52,13 @@ julia> collect(skipmissing(x))
 
 ```
 
-`missing` elements can be replaced with other values via `Missings.replace`.
+The [Missings.jl](https://github.com/JuliaData/Missings.jl) package provides a few convenience functions to work with missing values.
+
+The function `Missings.replace` replaces `missing` elements with another value:
```
bkamins: Why do we recommend it over `coalesce.(x,1)` (I know it is lazy, but is it that useful?)

docs/src/man/missing.md
```diff
@@ -69,32 +71,34 @@ The function `Missings.T` returns the element-type `T` in `Union{T, Missing}`.
 
 ```jldoctest missings
 julia> eltype(x)
-Union{Int64, Missings.Missing}
+Union{Int64, Missing}
```
bkamins: And here the same question against `Base.nonmissingtype`?

docs/src/man/querying_frameworks.md
```diff
@@ -15,8 +16,10 @@ A simple example of a query looks like this:
 ```jldoctest query
 julia> using DataFrames, Query
 
-julia> df = DataFrame(name=["John", "Sally", "Roger"], age=[54., 34., 79.], children=[0, 2, 4])
-3×3 DataFrames.DataFrame
+julia> df = DataFrame(name=["John", "Sally", "Roger"],
+                      age=[54., 34., 79.],
+                      children=[0, 2, 4])
+3×3 DataFrame
```
bkamins: tables missing eltype information need regeneration below

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -8,7 +8,7 @@ julia> using DataFrames, CSV
 julia> iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../test/data/iris.csv"));
 
 julia> head(iris)
-6×5 DataFrames.DataFrame
+6×5 DataFrame
```
bkamins: Same here with eltype and below

docs/src/man/sorting.md
```diff
@@ -10,7 +10,7 @@ julia> iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../test/data/iris.
 julia> sort!(iris);
 
 julia> head(iris)
-6×5 DataFrames.DataFrame
+6×5 DataFrame
```
bkamins: eltype problem with this file also

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -651,7 +651,7 @@ returns `true`. The function is passed a `DataFrameRow` as its only argument.
 # Examples
 ```
 julia> df = DataFrame(x = [3, 1, 2, 1], y = ["b", "c", "a", "b"])
-4×2 DataFrames.DataFrame
+4×2 DataFrame
```
bkamins: add eltype information

docs/src/man/missing.md
nalimilan: Good idea. I now present `coalesce` first.

docs/src/man/missing.md
nalimilan: `Base.nonmissingtype` isn't part of the public API, and initially it has only been added for internal use. Hopefully we'll be able to support something like `typesubtract(T, Missing)` at some point. For now I prefer recommending `Missings.T`, which is public even if it's not exported.

nalimilan:
Thanks. I've included changes in the commits. Regarding the eltypes, better fix that separately.

BTW, have you considered moving some of the material from Julia-DataFrames-Tutorial here?

docs/src/man/missing.md
bkamins: OK

bkamins:
OK - so let us merge it. I will update `eltypes` in a second run of clean up when I read all the material after merging on the website.

Issue #1460, opened 2018-07-19
check for preserved index column during combine


kleinschmidt:
When using `by` with a function that returns a dataframe with the grouping columns still present, `combine` interprets it as a duplicate column and renames it with a warning.  It would be nicer if we could check for whether the grouping column is present and all elements equal to the group values, and if so ignore it.

current behavior:
```
julia> by(d, :c1) do df
           filter(row -> row[:c2] == maximum(df[:c2]), df)
       end
WARNING: Duplicate variable names are deprecated: pass makeunique=true to add a suffix automatically.
...
2×3 DataFrames.DataFrame
│ Row │ c1 │ c1_1 │ c2 │
├─────┼────┼──────┼────┤
│ 1   │ a  │ a    │ 3  │
│ 2   │ b  │ b    │ 4  │
```

desired behavior:
```
2×2 DataFrames.DataFrame
│ Row │ c1 │ c2 │
├─────┼────┼────┤
│ 1   │ a  │ 3  │
│ 2   │ b  │ 4  │
```

bkamins:
@nalimilan I see two approaches here:
1. do what you propose (silently overwrite the columns)
2. add a `keepgrouping` keyword argument to `combine` and `by`; if it is set to `true` (the default) we keep the current behavior; if it is set to `false` we always drop them (this might be useful in some cases in general)

If we picked option 2., then at the same time we should add `skipmissing` keyword argument to `by` to clean up the API. What would you prefer?

bkamins:
By "drop them" I mean that we do not append grouping columns to the result.

bkamins:
Ah - right. We also should keep column order in mind (when we add grouping columns they come first).

Given the new split-apply-combine API (not using a `SubDataFrame` by default) I think that in practice it is better to have a keyword argument like in #1555, but set to keep the grouping columns by default (as I have proposed in option 2).

Also if we agree on this scenario I would not overwrite the columns if the user wants to keep grouping columns and there are duplicate column names, but keep the current behavior. But I am open to other opinions.

CC @ExpandingMan

nalimilan:
I'm fine with adding an argument, but I don't really like the current behavior with duplicate columns.

[...]

PR #1520, opened 2018-09-20
Improve performance of by() using NamedTuples

nalimilan:
Remove `GroupApplied` and deprecate combine in favor of `map(f, ::GroupedDataFrame)`. This avoids storing a copy of the per-group data returned by the user-provided function. Take advantage of this by allowing that function to return a `NamedTuple`. Introduce two completely different code paths depending on whether the first returned object is `DataFrame` or a `NamedTuple`, as the latter allows for more efficient operation by assuming that it represents a single row. Use the same progressive eltype widening approach as `Base.map` so that we fill column vectors whose types are known inside the kernel functions. This does not eliminate the type unstability due to the fact that the user-provided function takes a `DataFrame`, but ensuring type stability for half of the operations still improves performance significantly.

Also parameterize `GroupedDataFrame` on the type of data frame it wraps, and make its column index
have a concrete type. Deprecate an old map method for `SubDataFrame`. Fix a type unstability in hcat!.

The new approach is about 10× faster for a simple sum when the number of groups is large. There's a slight slowdown the first time `map` is called, but it disappears for the second call even for a different function (anyway each anonymous function is unique).
```julia
               b = repeat(20000:-1:1, outer=[40]),
               c = randn(800000))
gd = groupby(df, :a)
using BenchmarkTools, Statistics

# Master

julia> @time map(d -> mean(d[:c]), gd); # Warm up using a different function
  0.605958 seconds (2.64 M allocations: 167.579 MiB, 16.16% gc time)

julia> @time map(d -> sum(d[:c]), gd);
  0.396017 seconds (1.99 M allocations: 135.049 MiB, 47.55% gc time)

julia> @btime map(d -> sum(d[:c]), gd);
  270.617 ms (1918984 allocations: 131.52 MiB)

julia> @time map(d -> (s=sum(d[:c]), m=mean(d[:b]), std=std(d[:b])), gd);
  0.765082 seconds (3.00 M allocations: 199.228 MiB, 32.78% gc time)

julia> @btime map(d -> (s=sum(d[:c]), m=mean(d[:b]), std=std(d[:b])), gd);
  312.121 ms (2238984 allocations: 159.59 MiB)

julia> @time map(d -> DataFrame(x=sum(d[:c])), gd);
  0.457975 seconds (2.14 M allocations: 139.817 MiB, 34.31% gc time)

julia> @btime map(d -> DataFrame(x=sum(d[:c])), gd);
  273.170 ms (1998984 allocations: 132.74 MiB)

# This PR
julia> @time map(d -> sum(d[:c]), gd); # Warm up using a different function
  0.877008 seconds (3.16 M allocations: 156.162 MiB, 6.27% gc time)

julia> @time map(d -> sum(d[:c]), gd);
  0.114020 seconds (621.35 k allocations: 32.728 MiB, 5.25% gc time)

julia> @btime map(d -> sum(d[:c]), gd);
  29.794 ms (479034 allocations: 25.93 MiB)

julia> @time map(d -> (s=sum(d[:c]), m=mean(d[:b]), std=std(d[:b])), gd);
  0.567517 seconds (2.06 M allocations: 113.389 MiB, 4.26% gc time)

julia> @btime map(d -> (s=sum(d[:c]), m=mean(d[:b]), std=std(d[:b])), gd);
  65.593 ms (799045 allocations: 50.34 MiB)

julia> @time map(d -> DataFrame(x=sum(d[:c])), gd);
  0.540145 seconds (2.85 M allocations: 177.991 MiB, 5.39% gc time)

julia> @btime map(d -> DataFrame(x=sum(d[:c])), gd);
  147.087 ms (2039599 allocations: 138.41 MiB)
```

CC: @quinnj @piever who are familiar with these eltype widening tricks


src/groupeddataframe/grouping.jl
```diff
@@ -115,96 +114,204 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
+        return similar(gd.parent, 0)[gd.cols]
     end
-    hcat!(gd.parent[idx, gd.cols], valscat)
 end
 
+function _combine(first::NamedTuple, f::Function, gd::GroupedDataFrame)
+    m = length(first)
+    n = length(gd)
+    idx = Vector{Int}(undef, n)
+    initialcols = ntuple(i -> Vector{typeof(first[i])}(undef, n), m)
```
quinnj: What about CategoricalValues here? Don't we want to use some kind of `similar` construct? In Tables.jl, we have [`allocatecolumn`](https://github.com/JuliaData/Tables.jl/blob/master/src/fallbacks.jl#L47) which can be overloaded by custom scalars to provide an alternative AbstractVector type. It'd be nice possibly have something like `vectortype(scalar::T) = Vector{T}` in Base eventually so this could be used throughout the package ecosystem.

src/groupeddataframe/grouping.jl
```diff
@@ -115,96 +114,204 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
+    cols = _combine!(first, initialcols, idx, 1, 1, f, gd, propertynames(first))
+    valscat = DataFrame(collect(cols), collect(propertynames(first)))
+    idx, valscat
+end
+
+function _combine(first::AbstractDataFrame, f::Function, gd::GroupedDataFrame)
+    m = size(first, 2)
+    idx = Vector{Int}()
```
quinnj: `Int[]`?

src/groupeddataframe/grouping.jl
nalimilan: Right. It's ironic I missed that point. :-)

I've added a commit to use `similar` for the data frame case, but as you note it can't be solved for the scalar case without calling `allocatecolumn`. I guess we can wait until your Tables.jl PR is merged before merging this one. Do you think it would be OK for Tables.jl to implement the necessary method? It could use Require.jl to avoid a hard dependency on CategoricalArrays.

I also agree it would be useful to have something like this in Base.

src/groupeddataframe/grouping.jl
nalimilan: I'd rather keep it that way for consistency with the method above.

src/groupeddataframe/grouping.jl
quinnj: Yeah, Tables.jl already defines the [right methods](https://github.com/JuliaData/Tables.jl/blob/master/src/Tables.jl#L11) for both CategoricalValue & WeakRefString.

piever:
I haven't looked in depth to this specific case but I wanted to say that I would generally be in favor of having "universally useful" methods in Tables and relying or them (to avoid code duplication).

In particular there are `buildcolumns(::Nothing, itr)` in Tables (which is an improved version of `collect_columns` from IndexedTables) and `collect_colums_flattened` [here](https://github.com/JuliaComputing/IndexedTables.jl/blob/master/src/collect.jl#L82) in IndexedTables to collect an iterable of iterables of `NamedTuples` into a table with columnar storage. I wonder if some of that could be useful here. It'd be nicer if at least for basic things (collecting grouped data into normal data) the machinery was shared between here and IndexedTables. Vice versa, if the method here is superior, it be nice to keep the "basic tools" in Tables so that IndexedTables can rely on them (to the extend possible of course).

nalimilan:
Yeah, I've thought about reusing code too, but AFAICT the situation is too specific here. Note that I couldn't even share the code between the data frame method and the named tuple method. Also the user-provided function doesn't operate on a single row, but on a whole table, and it can return multiple types. Maybe this function could be moved to Tables.jl or another generic package, but it would remain specialized for grouping operations (even if it worked on any kind of table).

src/groupeddataframe/grouping.jl
nalimilan: Perfect. So I can update the PR once you've merged yours.

nalimilan:
One point I should have mentioned is the choice to replace `combine` with `map` for `GroupedDataFrame`s. I think it makes sense since mapping a function to each group in a `GroupedDataFrame` cannot return a `GroupedDataFrame`. It could return a vector with the results, but that's not very useful.

src/groupeddataframe/grouping.jl
nalimilan: Done. That allowed simplifying constructors a bit too.

nalimilan:
I've pushed two new commits. The first one allows using names in a different order from one group to the other. This is currently supported even if it affects performance a bit.

The second one reintroduces `combine`, and changes `map` to return a `GroupedDataFrame` (where it currently returns a `GroupApplied`). I think this makes sense as the grouping information is preserved, allowing for further operations if one wishes. The code is mostly the same between the two functions anyway, it's more a matter of API. In the longer term, we could imagine making `GroupedDataFrame` an `AbstractDataFrame`, which would allow behaviors similar to dplyr.

Unfortunately, the new code crashes Julia with "Illegal instruction" (https://github.com/JuliaLang/julia/issues/29430).

bkamins:
@nalimilan Is this ready for a review (I am asking because tests are failing)

bkamins:
@nalimilan maybe you will want to comment more on https://github.com/h2oai/db-benchmark/issues/30

src/groupeddataframe/grouping.jl
```diff
@@ -14,9 +14,9 @@ view into the AbstractDataFrame grouped by rows.
 
 Not meant to be constructed directly, see `groupby`.
 """
-mutable struct GroupedDataFrame
-    parent::AbstractDataFrame
-    cols::Vector         # columns used for sorting
+mutable struct GroupedDataFrame{T<:AbstractDataFrame}
```
bkamins: What is the benefit of `GroupedDataFrame` being mutable (and does it have performance impact)?

(as a side note - I have just noticed the same question applies to `DataFrame`)

src/groupeddataframe/grouping.jl
```diff
@@ -87,7 +90,7 @@ function groupby(df::AbstractDataFrame, cols::Vector;
         permute!(df_groups.starts, group_perm)
         Base.permute!!(df_groups.stops, group_perm)
     end
-    GroupedDataFrame(df, cols, df_groups.rperm,
+    GroupedDataFrame(df, DataFrames.index(df)[cols], df_groups.rperm,
```
bkamins: is `DataFrames.` needed here?

bkamins: Also `index(df)` assumes that `cols` are symbols and they can be integers or vector of bools.

src/groupeddataframe/grouping.jl
```diff
@@ -115,71 +118,102 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
-end
-function Base.map(f::Function, ga::GroupApplied)
-    GroupApplied(ga.gd, [wrap(f(df)) for df in ga.vals])
-end
-
 wrap(df::AbstractDataFrame) = df
+wrap(nt::NamedTuple) = nt
 wrap(A::Matrix) = convert(DataFrame, A)
-wrap(s::Any) = DataFrame(x1 = s)
+wrap(s::Union{AbstractVector, Tuple}) = DataFrame(x1 = s)
```
bkamins: what is the idea of wrapping a tuple like this?

As the code gives us:
```
julia> DataFrame(x=(1,2,3))
1×1 DataFrame
│ Row │ x         │
│     │ Tuple…    │
├─────┼───────────┤
│ 1   │ (1, 2, 3) │
```
(why do we prefer `DataFrame` to `NamedTuple` in this case?)

src/groupeddataframe/grouping.jl
```diff
@@ -115,71 +118,102 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
+
+For each group in `gd`, `f` is passed a `SubDataFrame` view holding the corresponding rows.
+`f` can return a single value, a named tuple, a vector, or a data frame.
+This determines the shape of the resulting data frame:
+- A single value gives a data frame with a single column and one row per group.
+- A named tuple gives a data frame with one column for each field in the named tuple
+  and one row per group.
+- A vector gives a data frame with a single column and as many rows
+  for each group as the length of the returned vector for that group.
+- A data frame gives a data frame with the same columns and as many rows
```
bkamins: by data frame I guess we also cover `SubDataFrame` - right? What about handling of `DataFrameRow`?

src/groupeddataframe/grouping.jl
nalimilan: Probably none. I'm surprised we haven't noticed this before. Let's change this in another PR.

src/groupeddataframe/grouping.jl
```diff
@@ -115,71 +118,102 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
+  for each group as the rows returned for that group.
+
+In all cases, the resulting `GroupedDataFrame` contains all the grouping columns in addition
```
bkamins: Do we want to mention what do we do in case of duplicate columns? Or maybe even add `makeunique` kwarg?

src/groupeddataframe/grouping.jl
```diff
@@ -115,71 +118,102 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
+to those listed above. Note that `f` must always return the same type of object for
+all groups, and (if a named tuple or data frame) with the same fields or columns.
+Returning a single value or a named tuple is significantly faster than
```
bkamins: This comment is not immediately clear as it is obvious that in the former cases you have less rows to process so this should be faster (but I guess you mean something else?).

src/groupeddataframe/grouping.jl
nalimilan: AFAICT integers and Bools are accepted. Good point about the unnecessary prefix.

src/groupeddataframe/grouping.jl
```diff
@@ -115,71 +118,102 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
-Combine a GroupApplied object (rudimentary)
+    map(f::Function, gd::GroupedDataFrame)
+
+Apply a function to each group of rows and return a `GroupedDataFrame`.
+
+For each group in `gd`, `f` is passed a `SubDataFrame` view holding the corresponding rows.
+`f` can return a single value, a named tuple, a vector, or a data frame.
+This determines the shape of the resulting data frame:
+- A single value gives a data frame with a single column and one row per group.
+- A named tuple gives a data frame with one column for each field in the named tuple
```
bkamins: What happens in the case of named tuple of vectors?
His is relevant because currently:

```
julia> DataFrame((a=1:5, b=5:-1:1))
5×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 5     │
│ 2   │ 2     │ 4     │
│ 3   │ 3     │ 3     │
│ 4   │ 4     │ 2     │
│ 5   │ 5     │ 1     │
```

src/groupeddataframe/grouping.jl
nalimilan: Hmm. I think I incorrectly assumed that the existing code treated tuples like vectors, but clearly it doesn't. I'll change that to use `NamedTuple` as it's more efficient.

src/groupeddataframe/grouping.jl
```diff
@@ -115,71 +118,102 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Bool}) =
...
+  and one row per group.
+- A vector gives a data frame with a single column and as many rows
```
bkamins: The same question for a vector of named tuples as:

```
julia> DataFrame([(a=1,b=2), (a=3,b=4)])
2×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │
│ 2   │ 3     │ 4     │
```

In general - we do not have to handle this but I want to be clear what we want.

Also what happens in case a matrix is passed?

bkamins:
I have left design comments as discussed. Also the general question is what if `f` in `map` or `combine` returns a valid Tables.jl value?

src/groupeddataframe/grouping.jl
nalimilan: Yeah, I tend to use "data frame" as a nicer replacement for `AbstractDataFrame`, and `DataFrame` when we really require that type. `DataFrameRow` should be handled like `NamedTuple`, but I don't think I've tested that. Will have a look.

src/groupeddataframe/grouping.jl
nalimilan: I'd like to tackle this separately as part of https://github.com/JuliaData/DataFrames.jl/issues/1460.

src/groupeddataframe/grouping.jl
nalimilan: I guess I could change this to "a single-row data frame" (vectors shouldn't be mentioned here as they only apply when there are several rows).

src/groupeddataframe/grouping.jl
nalimilan: Currently:
```julia
julia> by(d -> (a=[1, 2],), DataFrame(x=[1, 1, 2, 3]), :x)
3×2 DataFrame
│ Row │ x     │ x1            │
│     │ Int64 │ NamedTup…     │
├─────┼───────┼───────────────┤
│ 1   │ 1     │ (a = [1, 2],) │
│ 2   │ 2     │ (a = [1, 2],) │
│ 3   │ 3     │ (a = [1, 2],) │
```

This PR:
```julia
julia> by(d -> (a=[1, 2],), DataFrame(x=[1, 1, 2, 3]), :x)
3×2 DataFrame
│ Row │ x     │ a      │
│     │ Int64 │ Array… │
├─────┼───────┼────────┤
│ 1   │ 1     │ [1, 2] │
│ 2   │ 2     │ [1, 2] │
│ 3   │ 3     │ [1, 2] │
```

Neither behavior sounds very useful. We could throw an error for now in case we want to treat named tuples of vectors as lightweight data frames. I guess it could be more efficient (since the column types can be inferred). But it would be annoying if you really want to store a vector in a cell (what alternative would you have?).

src/groupeddataframe/grouping.jl
nalimilan: Behavior is unchanged by the PR:
```julia
julia> by(d -> [(a=1,), (a=2,)], DataFrame(x=[1, 1, 2, 3]), :x)
6×2 DataFrame
│ Row │ x     │ x1        │
│     │ Int64 │ NamedTup… │
├─────┼───────┼───────────┤
│ 1   │ 1     │ (a = 1,)  │
│ 2   │ 1     │ (a = 2,)  │
│ 3   │ 2     │ (a = 1,)  │
│ 4   │ 2     │ (a = 2,)  │
│ 5   │ 3     │ (a = 1,)  │
│ 6   │ 3     │ (a = 2,)  │
```

If a matrix is returned, it's passed to the `DataFrame` constructor, which creates `x1`... `xn` columns. I'll document this since we have a special method for it.

nalimilan:
> I have left design comments as discussed. Also the general question is what if `f` in `map` or `combine` returns a valid Tables.jl value?

You mean, something which isn't an `AbstractDataFrame` but nevertheless a `Table`? We should probably treat it as a data frame, but that's really not high in the priority list IMO.

src/groupeddataframe/grouping.jl
bkamins: Right - I do not know why it threw an error when I checked this.

src/groupeddataframe/grouping.jl
bkamins: You would wrap it in a vector like this `(a=[[1,2,3]])`. Anyway we can leave it for later when we decide if we want to detect support of Tables.jl interface in general.

nalimilan:
I've pushed a commit which should address your comments, and another one which disables the crashing tests. Unfortunately I haven't been able to find a workaround at this point.

src/groupeddataframe/grouping.jl
```diff
@@ -202,23 +237,182 @@ df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
...
+    len = length(gd)
+    # Handle first group
+    j = fill_row!(first, cols, rowstart, colstart, colnames)
+    @assert j === nothing # eltype is guaranteed to match
+    idx[rowstart] = gd.idx[gd.starts[rowstart]]
+    # Handle remaining groups
+    @inbounds for i in rowstart+1:len
+        row = wrap(f(gd[i]))
+        j = fill_row!(row, cols, i, 1, colnames)
+        if j !== nothing # Need to widen column type
```
pdeffebach: Should this be `!(j  === nothing)`?

nalimilan: That's the same thing.

src/groupeddataframe/grouping.jl
```diff
@@ -14,9 +14,9 @@ view into the AbstractDataFrame grouped by rows.
 
 Not meant to be constructed directly, see `groupby`.
 """
-struct GroupedDataFrame
-    parent::AbstractDataFrame
-    cols::Vector         # columns used for sorting
+struct GroupedDataFrame{T<:AbstractDataFrame}
+    parent::T
+    cols::Vector{Int}    # columns used for sorting
```
bkamins: Is this sorting or grouping?

src/groupeddataframe/grouping.jl
```diff
@@ -50,14 +50,17 @@ groupby(cols; sort = false, skipmissing = false)
 
 An iterator over a `GroupedDataFrame` returns a `SubDataFrame` view
 for each grouping into `d`. A `GroupedDataFrame` also supports
-indexing by groups and `map`.
+indexing by groups, `map` (which applies a function to each group)
+and `combine` (which applies a function to each group
+and combines the result into a data frame).
```
bkamins: Can `combine` return something else than `DataFrame`?

src/groupeddataframe/grouping.jl
```diff
@@ -72,8 +75,8 @@ vcat([g[:b] for g in gd]...)
 for g in gd
     println(g)
 end
-map(d -> mean(skipmissing(d[:c])), gd)   # returns a GroupApplied object
-combine(map(d -> mean(skipmissing(d[:c])), gd))
+map(d -> mean(skipmissing(d[:c])), gd)
```
bkamins: `using Statistics` is required for this example to work.

src/groupeddataframe/grouping.jl
```diff
@@ -87,7 +90,7 @@ function groupby(df::AbstractDataFrame, cols::Vector;
         permute!(df_groups.starts, group_perm)
         Base.permute!!(df_groups.stops, group_perm)
     end
-    GroupedDataFrame(df, cols, df_groups.rperm,
+    GroupedDataFrame(df, index(df)[cols], df_groups.rperm,
```
bkamins: We could write `intcols = index(df)[cols]` at the start of this method. Then `sdf = df[intcols]` will be faster and we do not need the second `groupby` method below` and can omit `::Vector` qualifier in this method signature for `cols`.

In short I would pass checking validity of `cols` to `index(df)[cols]` call and later work on what it returns. In this way we have a single point where column checking is done (in `getindex` for `index(df)`).

src/groupeddataframe/grouping.jl
```diff
@@ -129,71 +132,103 @@ end
...
-    GroupApplied(ga.gd, [wrap(f(df)) for df in ga.vals])
-end
-
 wrap(df::AbstractDataFrame) = df
+wrap(nt::NamedTuple) = nt
+wrap(r::DataFrameRow) = r
 wrap(A::Matrix) = convert(DataFrame, A)
-wrap(s::Any) = DataFrame(x1 = s)
+wrap(s::AbstractVector) = DataFrame(x1 = s)
+wrap(s::Any) = (x1 = s,)
```
bkamins: Can you add an internal comment what is the purpose of `wrap` function for future developers?

src/groupeddataframe/grouping.jl
```diff
@@ -129,71 +132,103 @@ end
...
 
 """
-Combine a GroupApplied object (rudimentary)
+    map(f::Function, gd::GroupedDataFrame)
+
+Apply a function to each group of rows and return a `GroupedDataFrame`.
+
+For each group in `gd`, `f` is passed a `SubDataFrame` view holding the corresponding rows.
+`f` can return a single value, a named tuple, a vector, or a data frame.
```
bkamins: can it return a `DataFrameRow` or a `Matrix`?

bkamins: Also below - `Matrix` should be added to the description where we say about a data frame.

src/groupeddataframe/grouping.jl
```diff
@@ -202,23 +237,185 @@ df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
                b = repeat([2, 1], outer=[4]),
                c = randn(8))
 gd = groupby(df, :a)
-combine(map(d -> mean(skipmissing(d[:c])), gd))
+map(d -> sum(skipmissing(d[:c])), gd)
```
bkamins: `combine` instead of `map`?

src/groupeddataframe/grouping.jl
```diff
@@ -129,71 +132,103 @@ end
...
+### See also
 
-* `ga` : a GroupApplied
+`combine(f, gd)` returns a `DataFrame` rather than a `GroupedDataFrame`
 
-### Returns
+"""
+function Base.map(f::Function, gd::GroupedDataFrame)
+    if length(gd) > 0
+        idx, valscat = _combine(wrap(f(gd[1])), f, gd)
```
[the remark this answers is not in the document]
bkamins: Again - I see it is OK (sorry, but the PR is long and I am reading it in linear fashion).

src/groupeddataframe/grouping.jl
```diff
@@ -276,16 +473,29 @@ by(f::Function, d::AbstractDataFrame, cols; sort::Bool = false)
...
 
-`f` can return a value, a vector, or a DataFrame. For a value or
-vector, these are merged into a column along with the `cols` keys. For
-a DataFrame, `cols` are combined along columns with the resulting
-DataFrame. Returning a DataFrame is the clearest because it allows
-column labeling.
+### Details
+
+For each group in `gd`, `f` is passed a `SubDataFrame` view with the corresponding rows.
+`f` can return a single value, a named tuple, a vector, or a data frame.
```
bkamins: same comments as for `map`

test/grouping.jl
```diff
@@ -86,89 +85,237 @@ module TestGrouping
...
+    @testset "by and groupby" begin
+        Random.seed!(1)
+        df = DataFrame(a = repeat(Union{Int, Missing}[1, 3, 2, 4], outer=[2]),
+                       b = repeat(Union{Int, Missing}[2, 1], outer=[4]),
+                       c = Vector{Union{Float64, Missing}}(randn(8)))
+
+        cols = [:a, :b]
+        f(df) = DataFrame(cmax = maximum(df[:c]))
+        g(df) = (cmax = maximum(df[:c]),)
+        h(df) = maximum(df[:c])
```
bkamins: Do we test against: `Vector`, `Matrix`, `SubDataFrame` and `DataFrameRow`?
Also testing for more than one column might be good.
Do we test against a case when what `f` returns produces a duplicate column name?

bkamins:
@nalimilan I have left some comments. Thank you for a massive work (sometimes I left a question rather than a comment/suggestion to make sure we are OK).

src/groupeddataframe/grouping.jl
nalimilan: Indeed that's grouping.

src/groupeddataframe/grouping.jl
nalimilan: I've modified the wording a bit to avoid repetition and mention matrices.

src/groupeddataframe/grouping.jl
bkamins: This is what I understand - so we should write `DataFrame` not `data frame` right?

test/grouping.jl
nalimilan: `SubDataFrame` is tested via `identity`. I've added tests for other cases. Duplicated names were exercised by `identity` too, but I've adapted the code to really check the names of duplicated columns.

src/groupeddataframe/grouping.jl
bkamins: also please mention `DataFrameRow` (I understand you have not pushed the changes yet as I do not see them).

src/groupeddataframe/grouping.jl
nalimilan: I'm not sure it really matters. It wouldn't be unreasonable for a custom `AbstractDataFrame` type to return something other than a `DataFrame`, even if currently that's not the case.

src/groupeddataframe/grouping.jl
```diff
@@ -129,71 +132,103 @@ end
...
+wrap(s::AbstractVector) = DataFrame(x1 = s)
+wrap(s::Any) = (x1 = s,)
 
 """
-Combine a GroupApplied object (rudimentary)
+    map(f::Function, gd::GroupedDataFrame)
+
+Apply a function to each group of rows and return a `GroupedDataFrame`.
+
+For each group in `gd`, `f` is passed a `SubDataFrame` view holding the corresponding rows.
```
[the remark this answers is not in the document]
nalimilan: We always use promotion to choose the element type for the column. I'm not sure it's worth mentioning given that it's a general pattern in Julia (like `map`).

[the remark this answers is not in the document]
bkamins: ok

src/groupeddataframe/grouping.jl
nalimilan: Good idea, but the second method is needed AFAICT, otherwise we extract a single vector instead of a data frame.

bkamins: Right.

src/groupeddataframe/grouping.jl
```diff
@@ -202,23 +238,185 @@ df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
...
-    idx = Vector{Int}(undef, size(valscat, 1))
-    j = 0
-    @inbounds for (start, val) in zip(gd.starts, vals)
-        n = size(val, 1)
-        idx[j .+ (1:n)] .= gd.idx[start]
-        j += n
+function combine(f::Function, gd::GroupedDataFrame)
+    if length(gd) > 0
+        idx, valscat = _combine(wrap(f(gd[1])), f, gd)
+        return hcat!(gd.parent[idx, gd.cols], valscat, makeunique=true)
```
nalimilan: I had to put `makeunique=true` to get some of the tests passing, in particular `by(identity, ...)`. I guess that's OK for now until we tackle #1460 more directly. Indeed, I'm afraid it would be too annoying not to allow people to return the same data frame as the input.

bkamins: Agreed, at least we are specific now, so it is easier to track cases like this.

bkamins:
`map` should also accept `Type` not only `Function`.

nalimilan:
Fixed. Good to go now?

src/groupeddataframe/grouping.jl
```diff
@@ -121,71 +125,102 @@ end
...
+# by the user-provided function lacks them
+wrap(x::Union{AbstractDataFrame, NamedTuple, DataFrameRow}) = x
+wrap(x::AbstractMatrix) = convert(DataFrame, x)
+wrap(x::AbstractVector) = DataFrame(x1 = x)
+wrap(x::Any) = (x1 = x,)
 
 """
-The result of a `map` operation on a GroupedDataFrame; mainly for use
-with `combine`
+    map(f::Function, gd::GroupedDataFrame)
```
bkamins: maybe update the docs to `map(f::Union{Function, Type}`?

Issue #1338, opened 2018-01-05
Docstrings aren't expanded.


bicycle1885:
A large part of docstrings found in <https://juliadata.github.io/DataFrames.jl/stable/lib/functions.html#Basics-1> is not expanded, which is annoying since function links (e.g. <https://juliadata.github.io/DataFrames.jl/stable/lib/functions.html#DataFrames.rename!>) do not go anywhere.

![image](https://user-images.githubusercontent.com/905683/34595052-1985d042-f218-11e7-94a3-bed568c85835.png)

nalimilan:
@cjprybol Any idea why docstrings aren't expanded?

jlapeyre:
I have run into blocks of function names that were not expanded in other packages. In one case, qualifying them, i.e. `ModuleName.func` caused them to be processed properly. These were unexported symbols.

nalimilan:
May be due to one of these errors, as Documenter avoids expanding a block if one error happens:
```
 !! No docs found for 'DataFrameRow'. [src/lib/types.md]
 !! No docs found for 'categorical!'. [src/lib/functions.md]
 !! No docs found for 'deleterows!'. [src/lib/functions.md]
 !! No docs found for 'eachcol'. [src/lib/functions.md]
 !! No docs found for 'names'. [src/lib/functions.md]
 !! No docs found for 'order'. [src/lib/functions.md]
 !! No docs found for 'show'. [src/lib/functions.md]
 !! No docs found for 'showcols'. [src/lib/functions.md]
 !! No docs found for 'size'. [src/lib/functions.md]
Documenter: building cross-references.
 !! No doc found for reference '[`isequal`](@ref)'. [src/lib/functions.md]
```

jlapeyre:
> May be due to one of these errors, as Documenter avoids expanding a block if one error happens:

Exactly this happened to me a couple of day ago. An undocumented function (it really had no docscrtring) prevented subsequent doc strings from being rendered. I more-or-less bisected to discover this.

Issue #1265, opened 2017-10-17
Functions that need docstrings


cjprybol:
!! No docs found for 'DataFrameRow'. [src/lib/types.md]
 !! No docs found for 'categorical!'. [src/lib/functions.md]
 !! No docs found for 'deleterows!'. [src/lib/functions.md]
 !! No docs found for 'eachcol'. [src/lib/functions.md]
 !! No docs found for 'names'. [src/lib/functions.md]
 !! No docs found for 'nullable!'. [src/lib/functions.md]
 !! No docs found for 'order'. [src/lib/functions.md]
 !! No docs found for 'show'. [src/lib/functions.md]
 !! No docs found for 'showcols'. [src/lib/functions.md]
 !! No docs found for 'size'. [src/lib/functions.md]
 !! No docs found for 'sort'. [src/lib/functions.md]
 !! No docs found for 'sort!'. [src/lib/functions.md]

quipa:
I am willing to work on this!

Nosferican:
Probably just write the docstring for those functions according to the standards (refer to the documentation for Julia and examples from the JuliaData ecosystem such as this one).

Nosferican:
Related #1338.

nalimilan:
Remaining functions to document (at least):
```
categorical!
deleterows!
eachcol
names
order
show
similar
size
```

ajwheeler:
Still remaining:
   
    names
    order
    size

PR #1521, opened 2018-09-20
Fix expansion of docstrings in manual

nalimilan:
Functions with no docstrings should not be listed as they trigger an error which blocks
expansion of the whole block. Remove them until we write proper docstrings.
Also fix closing backticks in a few docstrings.

Fixes https://github.com/JuliaData/DataFrames.jl/issues/1338.


nalimilan:
Yes, I've put it there: https://github.com/JuliaData/DataFrames.jl/issues/1265#issuecomment-423277880

PR #1527, opened 2018-09-21
tighten DataFrame of vector signature

piever:
This is needed to allow Tables integration to work with `AbstractVector` of `NamedTuples`. For example with this PR:

```julia
julia> DataFrame([(a=1, b=2)])
1×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │

julia> using StructArrays

julia> DataFrame(StructArray([1+3im]))
1×2 DataFrame
│ Row │ re    │ im    │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 3     │
```

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -927,7 +927,7 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
 
     header = allheaders[1]
     length(header) == 0 && return DataFrame()
-    cols = Vector{Any}(undef, length(header))
+    cols = Vector{AbstractVector}(undef, length(header))
     for (i, name) in enumerate(header)
         data = [df[name] for df in dfs]
         lens = map(length, data)
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -310,5 +310,5 @@ DataFrame(sink, sch::Data.Schema, ::Type{S}, append::Bool;
 end
 
 Data.close!(df::DataFrameStream) =
-    DataFrame(collect(Any, df.columns), Symbol.(df.header), makeunique=true)
+    DataFrame(collect(AbstractVector, df.columns), Symbol.(df.header), makeunique=true)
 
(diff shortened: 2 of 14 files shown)


bkamins:
As commented on Slack:

> The difference is that the container can be `Vector{Any}` and contain abstract vectors. But I guess we also can dump the support of it. The question is if then we should change `columns` field of `DataFrame` to `Vector{<:AbstractVector}`?

piever:
I think it makes sense to use a `Vector{AbstractVector}` if that's the only thing it can contain anyway.

bkamins:
OK - I will change it in a separate PR then (CC @nalimilan). In this PR it is enough to fix the tests.

piever:
Ok, so the storage was left `Vector` and the inner constructor was also untouched and tests should be passing now.

bkamins:
Yes - I want to change the internal storage type of `DataFrame` and constructors without rushing to rethink the consequences.

bkamins:
Actually I think that what we should do (and it is non-breaking) is to add in line 29 of `other/tables.jl` file a test if:

* `x` is an `AbstractVector`;
* all its entries as `AbstractVectors`

and if yes perform the conversion to appropriate `eltype`. In this way the change we propose will be minimally breaking (i.e. it will be breaking only where the logic in lines 16 to 28 of `other/tables.jl` kicks in and overwrites the old behavior).

(that is why I have said that I have to think it over 😄). I will push a PR with this proposed change for a review in minutes to make sure we are in sync.

quinnj:
How did this break user code? We've been printing a deprecation for a while and recently we started throwing an error. All this change does was avoid the code path that thru the error; I don't see how we could be breaking user code that wasn't already broken.

bkamins:
You can check out what tests had to be updated - those were all user code examples that were broken by this PR.

Explicitly e.g. `Any[1:3, 1:3]` used to work (without deprecation), but now it is broken.

See #1529 for a fix (fortunately it is not very problematic).

quinnj:
Got it; thanks for clarifying.

PR #1528, opened 2018-09-22
Avoid aliasing columns when assigning vector

nalimilan:
But avoid making a redundant copy for the first column.
Also move related test out of test/index.jl, which is about testing the Index type
rather than indexing.

Replaces https://github.com/JuliaData/DataFrames.jl/pull/1052, continuing https://github.com/JuliaData/DataFrames.jl/pull/1057.

It's a bit weird that only the first column would actually be `===` to the passed vector. Would it be better to make copies for all columns?

### DIFF
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -735,12 +735,20 @@ module TestDataFrame
 
         # columns should not alias if vector assigned
         df = DataFrame(A=[0], B=[0])
-        df[1:end] = [0.0]
+        x = [0.0]
+        df[1:end] = x
+        x[1] = 1.0
+        @test df[1, :A] === 0.0
+        @test df[1, :B] === 0.0
         df[1, :A] = 1.0
         @test df[1, :B] === 0.0
 
         df = DataFrame(A=[0], B=[0])
-        df[:, 1:end] = [0.0]
+        x = [0.0]
+        df[:, 1:end] = x
+        x[1] = 1.0
+        @test df[1, :A] === 0.0
+        @test df[1, :B] === 0.0
         df[1, :A] = 1.0
         @test df[1, :B] === 0.0
     end
--- a/test/dataframe.jl
+++ b/test/dataframe.jl
@@ -721,6 +721,30 @@ module TestDataFrame
         end
     end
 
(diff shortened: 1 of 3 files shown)


nalimilan:
> Would it be better to make copies for all columns?

So you say no? :-)

bkamins:
I think the first column should not be copied, but aliased as this is what would happen in `df.column = vector`. The following columns should be copied (in general copying them is against DataFrames.jl design - as `for col in cols df[col] = vector end ` would create aliases, but in this case it is what users probably expect so I am OK to change it this way). So I have approved your proposed implementation 😄.

Whatever we do - we should document it (as whatever choice we make it is not obvious). My current working plan is:

* get `getindex` right
* document `getindex`
* get `setindex` right (this will be a hard part to get all corner cases right - and actually this is why I think we are not ready for DataFrames.jl 1.0)
* document `setindex`

nalimilan:
> I think the first column should not be copied, but aliased as this is what would happen in `df.column = vector`.

The hesitation I have is that `df[column]` is different from `df[[column]]` since in the latter case you index one *or more* columns. So it would be OK to have a different behavior even for the first column.

I don't really like the difference in treatment between the first and the other columns. OTOH that's the only way to avoid making a possibly redundant copy, so...

I agree documenting the whole thing would be useful. Let's hope we can achieve consistency so that it makes sense.

bkamins:
Well - you have convinced me. `df[[1]]` is a useful pattern to *force a copy* - let us eventually document it and make copies everywhere :). If someone wants aliasing it is easy enough with loops.

nalimilan:
OK, done!

test/dataframe.jl
```diff
@@ -721,6 +721,30 @@ module TestDataFrame
...
+        df[1:end] = 0.0
+        df[1, :A] = 1.0
+        @test df[1, :B] === 0
+
+        df = DataFrame(A=[0], B=[0])
+        df[:, 1:end] = 0.0
+        df[1, :A] = 1.0
+        @test df[1, :B] === 0
+
+        # columns should not alias if vector assigned
```
bkamins: Those tests do not check for aliasing of the first column.

PR #1529, opened 2018-09-22
Make AbstractVector change in constructor clean and non deprecating

bkamins:
This follows #1527. What I do:
1. make the that change as non-breaking as possible
2. change other constructors to follow the expected behavior
3. update tests (so that we have coverage of all cases) and clean-up tests where `AbstractVector` qualification is not needed
4. Make `DataFrame` internally use `Vector{AbstractVector}`.
5. Fix some indentation issues introduced (not sure if I spotted 100% cases)

The thing I was unsure of was what type `columns` should be. I decided for `Vector{AbstractVector}` not `Vector{<:AbstractVector}` as the latter could e.g. allow `Vector{Vector}` to sneak in, and then we would not be able to add other `AbstractVectors` to such a `DataFrame` (CC @nalimilan - any comment on this as you have also written `Vector{<:AbstractVector}` earlier).

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -156,6 +156,15 @@ function DataFrame(; kwargs...)
     end
 end
 
+function DataFrame(columns::AbstractVector, cnames::AbstractVector{Symbol};
+                   makeunique::Bool=false)::DataFrame
+    if !all(col -> isa(col, AbstractVector), columns)
+        throw(ArgumentError("columns argument must be a vector of AbstractVector objects"))
+    end
+    return DataFrame(convert(Vector{AbstractVector}, columns),
+                     Index(convert(Vector{Symbol}, cnames), makeunique=makeunique))
+end
+
 function DataFrame(columns::AbstractVector{<:AbstractVector},
                    cnames::AbstractVector{Symbol}=gennames(length(columns));
                    makeunique::Bool=false)::DataFrame
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -83,7 +83,7 @@ size(df1)
 
 """
 mutable struct DataFrame <: AbstractDataFrame
-    columns::Vector
+    columns::Vector{AbstractVector}
     colindex::Index
 
     function DataFrame(columns::Vector{Any}, colindex::Index)
@@ -119,7 +119,7 @@ mutable struct DataFrame <: AbstractDataFrame
                 throw(DimensionMismatch("columns must be 1-dimensional"))
             end
         end
-        new(columns, colindex)
+        new(Vector{AbstractVector}(columns), colindex)
     end
 end
 
@@ -149,7 +149,7 @@ end
 
 function DataFrame(; kwargs...)
     if isempty(kwargs)
-        DataFrame(Any[], Index())
+        DataFrame(AbstractVector[], Index())
     else
         DataFrame(pairs(kwargs)...)
     end
@@ -158,7 +158,8 @@ end
 function DataFrame(columns::AbstractVector{<:AbstractVector},
                    cnames::AbstractVector{Symbol}=gennames(length(columns));
                    makeunique::Bool=false)::DataFrame
-    return DataFrame(convert(Vector{Any}, columns), Index(convert(Vector{Symbol}, cnames),
+    return DataFrame(Vector{AbstractVector}(columns),
+                     Index(convert(Vector{Symbol}, cnames),
                      makeunique=makeunique))
 end
 
@@ -169,7 +170,7 @@ DataFrame(columns::AbstractMatrix, cnames::AbstractVector{Symbol} = gennames(siz
 # Initialize an empty DataFrame with specific eltypes and names
 function DataFrame(column_eltypes::AbstractVector{T}, cnames::AbstractVector{Symbol},
                    nrows::Integer; makeunique::Bool=false)::DataFrame where T<:Type
-    columns = Vector{Any}(undef, length(column_eltypes))
+    columns = Vector{AbstractVector}(undef, length(column_eltypes))
     for (j, elty) in enumerate(column_eltypes)
         if elty >: Missing
             if Missings.T(elty) <: CategoricalArrays.CatValue
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -86,7 +86,7 @@ mutable struct DataFrame <: AbstractDataFrame
     columns::Vector{AbstractVector}
     colindex::Index
 
-    function DataFrame(columns::Vector{Any}, colindex::Index)
+    function DataFrame(columns::Vector, colindex::Index)
         if length(columns) == length(colindex) == 0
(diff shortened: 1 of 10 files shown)


src/dataframe/dataframe.jl
```diff
@@ -83,10 +83,10 @@ size(df1)
...
 """
 mutable struct DataFrame <: AbstractDataFrame
-    columns::Vector
+    columns::Vector{AbstractVector}
     colindex::Index
 
-    function DataFrame(columns::Vector{Any}, colindex::Index)
+    function DataFrame(columns::Vector, colindex::Index)
         if length(columns) == length(colindex) == 0
             return new(Vector{Any}(undef, 0), Index())
```
nalimilan: This one needs to be changed (`AbstractVector[]` would be shorter BTW).

src/dataframe/dataframe.jl
```diff
@@ -119,7 +119,7 @@ mutable struct DataFrame <: AbstractDataFrame
                 throw(DimensionMismatch("columns must be 1-dimensional"))
             end
         end
-        new(columns, colindex)
+        new(Vector{AbstractVector}(columns), colindex)
```
nalimilan: Would `convert` be OK here or would aliasing be a problem? Else, this change may have a performance impact when creating lots of data frames, like in `by`.

Same elsewhere.

test/dataframe.jl
```diff
@@ -491,21 +491,21 @@ module TestDataFrame
...
 
     @testset "unstack promotion to support missing values" begin
-        df = DataFrame(AbstractVector[repeat(1:2, inner=4), repeat('a':'d', outer=2), collect(1:8)],
+        df = DataFrame([repeat(1:2, inner=4), repeat('a':'d', outer=2), collect(1:8)],
                        [:id, :variable, :value])
         udf = unstack(df, :variable, :value)
         @test udf == unstack(df, :variable, :value) == unstack(df, :id, :variable, :value)
-        @test udf == DataFrame(AbstractVector[Union{Int, Missing}[1, 2], Union{Int, Missing}[1, 5],
+        @test udf == DataFrame([Union{Int, Missing}[1, 2], Union{Int, Missing}[1, 5],
                                    Union{Int, Missing}[2, 6], Union{Int, Missing}[3, 7],
```
nalimilan: Still incorrect indentation. Same below.

src/dataframe/dataframe.jl
bkamins: see comment below

src/dataframe/dataframe.jl
bkamins: also see a comment below

test/dataframe.jl

bkamins:
@nalimilan Having given it some thought I have decided to leave `Vector{Any}` for now. The reason is what you have commented about performance.

nalimilan:
Actually, AFAICT it's better for performance to use `::Vector{AbstractVector}`, since we can avoid copying the input if the user provides a `Vector{AbstractVector}`. OTC, with `::Vector{Any}`, we *always* have to make a copy since we require users to pass an `AbstractVector{<:AbstractVector}`.

One case where that's not true is when a scalar is passed, since in that case the input cannot be a `Vector{AbstractVector}`. But since we need to allocate the column vectors already, it's probably not a big deal to also copy that vector.

bkamins:
The scalar case was exactly the reason. I reverted `Vector{AbstractVector}` and adjusted the signature of the inner constructor accordingly. Actually we could have two inner constructors but hopefully the compiler can optimize-out the checks that are not needed in `Vector{AbstractVector}` case (but if you feel it would be good to split the union into two methods I can add it).

nalimilan:
I guess we could move the scalar recycling logic to an outer constructor, but it should be OK as-is. The overhead should be negligible when providing a `Vector{AbstractVector}`.

bkamins:
Next round of changes. Actually it seems that we need to have another two parameter constructor to avoid deprecation.

src/dataframe/dataframe.jl
```diff
@@ -156,6 +156,15 @@ function DataFrame(; kwargs...)
     end
 end
 
+function DataFrame(columns::AbstractVector, cnames::AbstractVector{Symbol};
```
nalimilan: Isn't this supposed to be deprecated?

bkamins: We can deprecate it, but it was not deprecated. This works cleanly now:
```
julia> DataFrame(Any[1:3, [1,2,3]], [:a,:b])
3×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │
│ 2   │ 2     │ 2     │
│ 3   │ 3     │ 3     │
```
Why would you deprecate it?

bkamins: By now I mean on currently tagged version of DataFrames.jl

bkamins: Actually there is one call, namely: `DataFrames(Any[[1,2,3], [2,3,4], makeunique=true)` that used to work and will not work now, but I guess no one would use it anyway.

bkamins: Also with this the tests go through cleanly and they failed without it (seems it is good to have tests catching breaking changes 😄).

bkamins:
CI goes through, but I guess it is better to think over the changes carefully before merging them.

src/dataframe/dataframe.jl
nalimilan: Wasn't it the point of https://github.com/JuliaData/DataFrames.jl/pull/1527? I must be missing something.

bkamins: To my understanding the point of #1527 was to make `DataFrame` constructor in sync with Tables.jl (as without it integration with Tables.jl was broken).

I do not see the benefit of deprecating this constructor (although it is only useful for `AbstractVector{Any}` containing `AbstractVectors`), as Tables.jl passes one positional and zero keyword arguments so it is not conflicting, but if you see it we can add `depwarn` to it (but at least for the time being we should keep it because removing it is breaking).

nalimilan: I find it confusing that `Vector{Any}` would be accepted when there are two arguments, but not when there's a single one. Deprecating both methods sounds clearer.

bkamins: Which two methods do you mean to deprecate?

Single `Vector{Any}` is accepted after this PR as well as `Vector{Any}` and column names. Also both were accepted in the past (without deprecation)

nalimilan: OK, sorry. I thought this PR only added the old one-argument constructor, but deprecated. Now I see we can perfectly keep it without conflicting with the Tables constructor. Fine then.

bkamins: No problem :) - those things are like "communicating vessels" problem (and I tried to be as non-breaking as reasonable) - that is why I really appreciate a careful review 👍.

bkamins:
@nalimilan do we merge it?

Issue #1533, opened 2018-09-24
Behavior of `df[1, cols]` and `DataFrameRow`


bkamins:
@nalimilan This is speculative, but I want to discuss it before we finish `getindex` cleanup.

1. I think we should keep `DataFrameRow` structure (there were some discussion of returning `NamedTuple` from `eachrow` but it would not allow for mutation);
2. Currently `getindex[1, :]` returns a `DataFrame` but maybe it should return a `DataFrameRow`? Also then maybe `DataFrameRow` could accept only a selection of columns to make `getindex[1, columns]` work consistently and return `DataFrameRow` (this would be breaking and I am not 100% convinced we want it)?
3. If we do not go for point 2) maybe it would be reasonable to expose some function that allows to get a row of a `DataFrame` as `DataFrameRow` directly (something like `getrow(df, 1)`)?

nalimilan:
That definitely makes sense, especially if we decide that iterating over a data frame gives rows (and therefore `DataFrameRow` objects). That would also be much more efficient than creating a new `DataFrame`. Overall I wouldn't expect this to break a lot of code, but who knows. Anyway we should go through a deprecation period first, and we can always go back if it turns out to be a bad idea.

bkamins:
Before I move on can you clarify me the following design issues (those happened before I started contributing):

* why `DataFrameRow` is not a subtype of `AbstractDataFrame`?
* why `SubDataFrame` does not allow arbitrary `AbstractDataFrame` as a parent?
* why `DataFrameRow` cannot be removed and replaced by `SubDataFrame` (assuming that the questions above do not have a strong reason to unify them - I guess all the mechanics where `DataFrameRow` is used in DataFrames.jl ecosystem could be rewritten for `SubDataFrame` and the mechanics allowing to iterate over `DataFrameRow` could be removed).

Probably you see what it is getting at: remove `DataFrameRow` in favor of `SubDataFrame`. Then `df[1, :]` would return a `DataFrame` still but `@view df[1,:]` would get a view of a row - and the same would be returned by `eachrow` iterator.

We could even retain `DataFrameRow` internally in DataFrames.jl if it is really needed but do not expose it to the user but instead expose publicly only `SubDataFrame`, e.g. in `eachrow`.

nalimilan:
>     * why `DataFrameRow` is not a subtype of `AbstractDataFrame`?

[...]

Issue #142, opened 2012-12-29
Make getindex(::SubDataFrame, ::Union{Integer, Symbol}) return a view of column vector


GlenHertz:
When cleaning up data or doing signal processing, element by element, the user may need to modify individual values of a dataframe (df), subdataframes (sdf), and datavec (dv).  So that assignments can be made like:

``` Julia
sdf["B"][n] = (sdf["B"][n-1] + sdf["B"][n+1])/2 + sdf["C"][n-1]

#For Arrays and SubArrays:
a[:,1] #is a vector reference
a[2,3] #is a element reference
a[:,3][2] #is an element copy

#For DataFrames
df["A"] #is a dv reference
df[1,2] #is a element reference

#For SubDataFrames
sdf["A"] #is a dv copy
sdf[1,2] #is not valid

#For DataVecs
dv[3] #is an element reference
```

Let me know if I made any errors.

I would propose making `sdf["A"]` return a `SubDataVec` reference.  For consistency I think `sdf[1,2]` should be valid but since df's use named columns I don't see this being that popular.

Glen

quinnj:
I don't think this is particularly relevant anymore, `view` is pretty standard and it's clear that it returns a "view" instead of copying, while all other indexing generally match Base array indexing behavior. The exception, I guess, is `df[3]`, which just returns the underlying column itself, but I think that's fair given the columnar-representation of DataFrames.

nalimilan:
I think there's still something to improve: `sdf[1]` makes a copy of the first column with only the relevant indices. It would make more sense to return a view, since users would create a `SubDataFrame` precisely to avoid making copies.

nalimilan:
@bkamins This is relevant for your `getindex` consistency improvements.

bkamins:
I was not aware about this issue, but I thought the current functionality was intended.

Actually I think that there is a problem with `SubDataFrame` and `view` methods not `getindex`. We should reimplement `SubDataFrame` and `view` so that they take two arguments in general like `getindex` and passing one argument selects a column not a row.

Here is the example of the issue:

```
julia> sdf = view(DataFrame(x = 1:10, y = 1.0:10.0), 1:5)
5×2 SubDataFrame{UnitRange{Int64}}
│ Row │ x     │ y       │
│     │ Int64 │ Float64 │
├─────┼───────┼─────────┤
│ 1   │ 1     │ 1.0     │
│ 2   │ 2     │ 2.0     │
[...]

PR #1534, opened 2018-09-24
WIP: clean up getindex: part 2

bkamins:
This PR was hard for me. Therefore I split it into implementation (pushing now) and tests (which I will add when we decide this is the functionality we want).

It implements discussion in #1190, #142 and #1533 and some minor clean-ups.

A particular difficulty was what to deprecate and what to change. @nalimilan - an initial review is appreciated (I will work on it further if we decide this is a right direction).


src/subdataframe/subdataframe.jl
```diff
@@ -101,24 +113,28 @@ Base.parent(sdf::SubDataFrame) = getfield(sdf, :parent)
...
-    return SubDataFrame(adf, convert(Vector{Missings.T(T)}, rowinds))
+function Base.view(adf::AbstractDataFrame, rowinds)
+    Base.depwarn("view(adf, x) will create a `SubDataFrame` containing columns `x` and all rows in the future.", :view)
+    return SubDataFrame(adf, rowinds, :)
 end
 
-function Base.view(adf::AbstractDataFrame, rowinds::Any)
-    return SubDataFrame(adf, rowinds)
+function Base.view(adf::AbstractDataFrame, rowinds, colinds)
+    return SubDataFrame(adf[colinds], rowinds)
```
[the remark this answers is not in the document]
bkamins: I was thinking about it when I asked the same about `DataFrameRow` earlier. I came to the conclusion that it is impossible to consistently handle columns. The reason is that we allow dual column indexing: by their numbers and by their names. Later we can rename columns or move them breaking one type of indexing or the other and if we break number-name mapping in the parent it is not clear what we should do in the view.

Finally `SubDataFrame` in general is supposed to be only used when original `DataFrame` is not mutated in terms of rows/columns (e.g. removing rows in source `DataFrame` can break `SubDataFrame`) - this is something we should clearly indicate in the documentation.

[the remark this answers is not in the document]
nalimilan: > I was thinking about it when I asked the same about `DataFrameRow` earlier. I came to the conclusion that it is impossible to consistently handle columns. The reason is that we allow dual column indexing: by their numbers and by their names. Later we can rename columns or move them breaking one type of indexing or the other and if we break number-name mapping in the parent it is not clear what we should do in the view.

I'm not sure that's really an issue. Anyway as you note one isn't supposed to change the structure of the parent. What bothers me more is the inconsistency between passing `:` for columns and passing a vector. Though in practice it shouldn't really matter.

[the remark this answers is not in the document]
bkamins: To sum up: we should leave not creating a copy when passing `:` for performance reasons. I will add a note in the docs when `parent` is copied and when not.

In fact - this is consistent with the thinking that `DataFrame` is a collection of rows.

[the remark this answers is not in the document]
nalimilan: > To sum up: we should leave not creating a copy when passing `:` for performance reasons.

Yes, I agree, what I meant is that if `SubDataFrame` also stored the indices of a subset of columns we wouldn't need to make a copy when not passing `:` either. I'm not saying we should definitely do this, even less than it should be done in this PR, but maybe something to keep in mind.

bkamins:
Before implementing, based on #1533 and trying to be consistent with Base behavior I have added the documentation of target `getindex`/`view` behavior to this PR before moving forward.

This is complex and subtly breaking in several places (I would recommend that during a review each proposal is verified against current behavior - at least this is what I had to do and sometimes were surprised by the result).

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* a vector, later denoted as `cols`:
+    * a vector of `Symbol` (does not have to be a subtype of `AbstractVector{Symbol}`);
+    * a vector of `Integer` other than `Bool` (does not have to be a subtype of `AbstractVector{<:Integer}`);
+    * a vector of `Bool` that has to be a subtype of `AbstractVector{Bool}`.
+    * a colon.
+
+The rules for a valid type of index into a row are the following:
+* a value, later denoted as `row`:
+    * an `Integer` that is not `Bool`;
+* a value, later denoted as `rows`:
```
nalimilan: value -> vector

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+    * a vector of `Integer` other than `Bool` (does not have to be a subtype of `AbstractVector{<:Integer}`);
+    * a vector of `Bool` that has to be a subtype of `AbstractVector{Bool}`;
+    * a colon.
+
+The following list specifies return types of `getindex` operations depending on argument types. In the summary `df` represents a `DataFrame`, `sdf` is a `SubDataFrame` and `dfr` is a `DataFrameRow`.
+
+`DataFrame`:
+* `df[col]` -> an alias of a vector contained in column `col`;
```
nalimilan: "a alias" is a bit weird. Maybe just say "the vector", it should be clear it's not a copy since you mention "a copy" when that's the case. You could also mention this before or after the list.

In other places too, it would be better to say "the vector" or "the value" rather than "a", since there's only one vector or value which corresponds to the indices.

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* `df[cols]` -> a freshly allocated `DataFrame` containing aliases of vectors contained in columns `cols`;
+* `df[row, col]` -> a value contained in row `row` of column `col`;
+* `df[row, cols]` -> a `NamedTuple` containing data from row `row` in columns `cols`;
+* `df[rows, col]` -> a copy of a vector `df[col]` with only rows `rows` selected;
```
nalimilan: Maybe "with only the entries corresponding to rows `` `rows` ``"?

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* `df[rows, cols]` -> a `DataFrame` containing columns `cols` and `df[rows, col]` as a vector in each `col` in `cols`.
```
nalimilan: I don't find this one very clear. The fact that all columns are copied should be mentioned explicitly. Maybe drop the recursive definition in terms of `df[rows, col]`.

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* `@view df[col]` -> an alias of a vector contained in column `col`;
```
nalimilan: Should mention that it's identical to `df[col]`.

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* `@view df[cols]` -> a `SubDataFrame` with parent `df` if `cols` is a colon and `df[cols]` otherwise;
```
nalimilan: Maybe the special treatment of colon could be problematic, and we should also use `df[:]` as the parent in that case? Same in similar cases.

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* `@view df[row, col]` -> translates to `view(df[col], row)` (a `0`-dimensional view into `df[col]`);
```
nalimilan: If you mention the translation to vector `view`, you could also do the same above for `getindex`. That could indeed be simpler to explain (not sure).

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* `@view df[row, cols]` -> a `DataFrameRow` with parent `df` if `cols` is a colon and `df[cols]` otherwise;
+* `@view df[rows, col]` -> translates to `view(df[col], rows)` (a standard view into `df[col]` vector);
+* `@view df[rows, cols]` -> a `SubDataFrame` with parent `df` if `cols` is a colon and `df[cols]` otherwise.
+
+`SubDataFrame`:
+* `sdf[col]` -> a view of vector contained in column `col` of `parent(sdf)` with `DataFrames.rows(sdf)` as selector;
+* `sdf[cols]` -> a freshly allocated `DataFrame` containing copies of vectors contained in columns `cols` of `sdf`;
```
nalimilan: This seems inconsistent with `sdf[col]` and `df[cols]`, which don't make copies. Better return a `SubDataFrame`.

docs/src/lib/functions.md
```diff
@@ -48,4 +48,62 @@ sort!
...
+* `df[col]` -> an alias of a vector contained in column `col`;
+* `df[cols]` -> a freshly allocated `DataFrame` containing aliases of vectors contained in columns `cols`;
+* `df[row, col]` -> a value contained in row `row` of column `col`;
+* `df[row, cols]` -> a `NamedTuple` containing data from row `row` in columns `cols`;
+* `df[rows, col]` -> a copy of a vector `df[col]` with only rows `rows` selected;
+* `df[rows, cols]` -> a `DataFrame` containing columns `cols` and `df[rows, col]` as a vector in each `col` in `cols`.
+* `@view df[col]` -> an alias of a vector contained in column `col`;
+* `@view df[cols]` -> a `SubDataFrame` with parent `df` if `cols` is a colon and `df[cols]` otherwise;
+* `@view df[row, col]` -> translates to `view(df[col], row)` (a `0`-dimensional view into `df[col]`);
+* `@view df[row, cols]` -> a `DataFrameRow` with parent `df` if `cols` is a colon and `df[cols]` otherwise;
```
nalimilan: So this is going to require creating a new `DataFrame` for each row? That sounds really bad. Maybe we could change `DataFrameRow` to also take a subset of rows? That's really a corner case though, so it could be fixed later.

docs/src/lib/functions.md
bkamins: Retaining `df` will be faster and I guess it will be important for `groupby` when you have a lot of small groups.

This is a case that currently was not on the table as `@view df[x]` treats `x` as rows, but it is going to be changed to ensure consistency.

As long as `df` is not mutated this should not matter in practice.

docs/src/lib/functions.md
bkamins: I was not sure what is best here myself.

In Base such an operation destroys a view. Also some code in DataFrames.jl relies on the fact that `sdf[cols]` returns a `DataFrame` if I remember correctly.

I will leave as is in the next iteration, but let us discuss it further.

docs/src/lib/functions.md
bkamins: We have the same situation with `SubDataFrame` as current `SubDataFrame` and `DaraFrameRow` were designed with the assumption that only row-indexing is performed.

I would leave it for a separate PR (we will have to handle colon and subset of columns separately and support a mapping from column numbers of a view to column numbers of a parent in the latter). In the worst case user should first create a parent that has correct columns and then run a `view` on all of them. I will open an issue for this to keep track of it.

bkamins:
I have reviewed the comments. The only problematic thing is what `sdf[cols]` should produce.

docs/src/lib/functions.md
nalimilan: But in Base `x[inds]` always makes a copy, so it's consistent that it's also the case for views. For `DataFrame`, `df[cols]` doesn't copy the column vectors, so it would be consistent to do that for `SubDataFrame`. The goal is that for consistency if you mutate the result, the original `AbstractDataFrame` is changed in both cases.

docs/src/lib/functions.md
nalimilan: `https://github.com/JuliaData/DataFrames.jl/issues/1557`

docs/src/lib/indexing.md
```diff
@@ -0,0 +1,77 @@
...
+
+The following list specifies return types of `getindex` operations depending on argument types. In all operations
+copying vectors is avoided where possible. If it is performed a description explicitly mentions that the data is *copied*.
+
+`DataFrame`:
+* `df[col]` -> the vector contained in column `col`;
+* `df[cols]` -> a freshly allocated `DataFrame` containing the vectors contained in columns `cols`;
+* `df[row, col]` -> the value contained in row `row` of column `col`, the same as `df[col][row]`;
+* `df[row, cols]` -> a `NamedTuple` containing data from row `row` in columns `cols`;
+* `df[rows, col]` -> a copy of the vector `df[col]` with only the entries corresponding to `rows` selected, the same as df[col][rows];
```
nalimilan: Missing backticks.

docs/src/lib/functions.md
nalimilan: "As long as `df` is not mutated" is what bothers me. :-)

Why do you expect `view(df, :)` to be used in `groupby`? Presumably if the overhead of calling `df[cols]` is OK when `cols != :`, it's also OK when `cols == :`? Or is the latter more common than the former?

bkamins:
> Why do you expect `view(df, :)` to be used in groupby? Presumably if the overhead of calling `df[cols]` is OK when `cols != :`, it's also OK when `cols == :`? Or is the latter more common than the former?

Currently we have:
```
Base.getindex(gd::GroupedDataFrame, idx::Int) =
    view(gd.parent, gd.idx[gd.starts[idx]:gd.ends[idx]])
```
which essentially will have to be rewritten as `view(gd.parent, gd.idx[gd.starts[idx]:gd.ends[idx]], :)` given this PR (after the deprecation period).

So, in short, because we rewrite `view(df, i)` to mean `view(df, i, :)` after this PR the form with a colon will be most common in users code. In the current code `view(df, i)` is a very popular pattern (I would say the only that is encountered in practice given current design).

Also note that currently `view(df, r, c)` always makes a copy of `df` while `view(df, r)` never makes a copy.

docs/src/lib/functions.md
bkamins: OK, I have rewritten it to:
> a `SubDataFrame`, with parent `parent(sdf)` if `cols` is a colon and `parent(sdf)[cols]` otherwise

nalimilan:
OK, I see. The best fix is really to make `SubDataFrame` store column indices (#1557), or to make it really cheap to construct a new `DataFrame`. Until then, better keep `view(df, rows, :)` efficient I guess.

bkamins:
The good thing is that we will go through a deprecation period when this would not matter much 😄, and in the mean time I will try to make #1557 efficient.

bkamins:
Actually there is a lot to improve in the implementation, apart from the earlier comments, once we have decided on the functionality.

Anyway this PR will be mostly deprecations given what we have discussed.

src/other/index.jl
```diff
@@ -121,47 +121,42 @@ function Base.insert!(x::Index, idx::Integer, nm::Symbol)
...
 end
 
 # catch all method handling cases when type of idx is not narrowest possible, Any in particular
 # also it handles passing missing values in idx
 function Base.getindex(x::AbstractIndex, idx::AbstractVector)
-    idxs = filter(!ismissing, idx)
-    if length(idxs) != length(idx)
-        throw(ArgumentError("missing values are not allowed for column indexing"))
-    end
+    idxs = disallowmissing(idx)
```
nalimilan: Is this call really useful now?

src/subdataframe/subdataframe.jl
```diff
@@ -72,26 +72,47 @@ function SubDataFrame(parent::DataFrame, rows::Colon)
...
+    if !all(x -> (x isa Integer) && !(x isa Bool), rows)
+        throw(ArgumentError("only `Integer` indices are accepted in `rows`"))
+    end
+    return SubDataFrame(parent, convert(Vector{Int}, rows))
 end
 
-function SubDataFrame(sdf::SubDataFrame, rowinds::Colon)
-    return sdf
+SubDataFrame(parent::DataFrame, rows) =
+    SubDataFrame(parent, (1:nrow(parent))[rows])
```
nalimilan: What's this trick?

src/subdataframe/subdataframe.jl
```diff
@@ -101,23 +122,48 @@ Base.parent(sdf::SubDataFrame) = getfield(sdf, :parent)
...
+# TODO: implement
+# `@view df[col]` -> the vector contained in column `col` (this is equivalent to `df[col]`)
+# `@view df[cols]` -> a `SubDataFrame` with parent `df` if `cols` is a colon and `df[cols]` otherwise
+# `sdf[col]` -> a view of the vector contained in column `col` of `parent(sdf)` with `DataFrames.rows(sdf)` as a selector;
+# `sdf[cols]` -> a `SubDataFrame`, with parent `parent(sdf)` if `cols` is a colon and `parent(sdf)[cols]` otherwise;
+# after deprecation period
 
-function Base.view(adf::AbstractDataFrame, rowinds::Any)
+function Base.view(adf::AbstractDataFrame, rowinds)
+    Base.depwarn("`view(adf, x)` will be translated to `view(adf, x, :)` in the future.", :view)
```
nalimilan: Better tell people that it's the new syntax they should use, that's clearer.

src/subdataframe/subdataframe.jl
```diff
@@ -101,23 +122,48 @@ Base.parent(sdf::SubDataFrame) = getfield(sdf, :parent)
...
 
-function Base.view(adf::AbstractDataFrame, rowinds::Any, colinds::Union{Colon, AbstractVector})
+function Base.view(adf::AbstractDataFrame, rowind::Integer, colind::ColumnIndex)
+    Base.depwarn("`view(adf, rowind, colind)` will create a 0-dimensional view into `adf[colind]` in the future." *
+                 " Use `view(adf, rowind:rowind, [colind])` instead.", :view)
+    return SubDataFrame(adf[colind], rowind)
+end
+
+function Base.view(adf::AbstractDataFrame, rowind::Integer, colinds)
+    Base.depwarn("`view(adf, x)` will create a `DataFrameRow` in the future." *
```
nalimilan: Deprecation doesn't match the called method.

src/subdataframe/subdataframe.jl
```diff
@@ -130,10 +176,22 @@ index(sdf::SubDataFrame) = index(parent(sdf))
...
+    return parent(sdf)[rows(sdf), colinds]
+end
+
 function Base.getindex(sdf::SubDataFrame, colinds)
+    Base.depwarn("`sdf[colinds]` will create a `SubDataFrame` in the future." *
+                 " Use `sdf[:, colinds]` to get a `DataFrame`.", :getindex)
     return parent(sdf)[rows(sdf), colinds]
 end
 
+function Base.getindex(sdf::SubDataFrame, rowind::Integer, colind::ColumnIndex)
```
nalimilan: This is redundant with the method below, right?

src/dataframerow/dataframerow.jl
```diff
@@ -37,7 +40,13 @@ Base.setproperty!(r::DataFrameRow, idx::Symbol, x::Any) = setindex!(r, x, idx)
 # Private fields are never exposed since they can conflict with column names
 Base.propertynames(r::DataFrameRow, private::Bool=false) = names(r)
 
-Base.view(r::DataFrameRow, c) = DataFrameRow(parent(r)[[c]], row(r))
+function Base.view(r::DataFrameRow, col::ColumnIndex)
+    Base.depwarn("view(dfr, col] a `0`-dimensional view in the future.", :getindex)
+    DataFrameRow(parent(r)[[col]], row(r))
+end
+
+Base.view(r::DataFrameRow, cols) = DataFrameRow(parent(r)[cols], row(r))
```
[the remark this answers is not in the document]
bkamins: This is the current functionality and we do not change anything here and we want to keep it (i.e. `view` of a `DataFrameRow` is to remain `DataFrameRow` except when a single column is requested).

I have only added methods that were problematic:
* when single column was passed there was a bug in the old code;
* when a colon was passed a copy of a parent was done, now I avoid it.

src/dataframerow/dataframerow.jl
```diff
@@ -13,15 +12,19 @@ Return the parent data frame of `r`.
...
-    return DataFrameRow(parent(r)[idx], row(r))
-end
-
 function Base.getindex(r::DataFrameRow, idx::ColumnIndex)
     return parent(r)[row(r), idx]
 end
 
-Base.getindex(r::DataFrameRow, ::Colon) = r
+function Base.getindex(r::DataFrameRow, idxs::AbstractVector)
+    Base.depwarn("dfr[idxs] will create a `NamedTuple` in the future.", :getindex)
```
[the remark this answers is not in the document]
bkamins: This is what we have agreed upon, i.e. that `DataFrameRow` should be turned to `NamedTuple` if multiple columns are selected.

So let us confirm if we want it. This is consistent with the behavior `df[row, cols]` is a `NamedTuple` so `view(df, row, :)[cols]` should be the same `NamedTuple`.

This is the consequence of the discussion in #1533 in my opinion. But we can change it.

Also if we agree to keep that `NamedTuple` is a materialized counterpart of `DataFrameRow` then I will add a `copy` method to `DataFrameRow` that converts it to a `NamedTuple` for consistency.

src/other/index.jl
```diff
@@ -121,47 +121,42 @@ function Base.insert!(x::Index, idx::Integer, nm::Symbol)
...
     length(idxs) == 0 && return Int[] # special case of empty idxs
     if idxs[1] isa Real
```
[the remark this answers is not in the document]
bkamins: I left test for `Real` as it gives a more specific error message than the one at the bottom.

Additionally `disallowmissing` was not needed anyway so I have removed it after reviewing.

Also I did not remove any specific methods AFAIK (apart from `Union{Bool, Missing}` case after your comment). But I have checked that it is as type stable as possible.

src/subdataframe/subdataframe.jl
bkamins: The idea is that:
* we are sure that we get `Vector{Int}` as a result;
* `(1:nrow(parent))[rows]` will use machinery from Base to catch any problematic values in `rows`, OTOH anything that is valid in Base will be valid here.

If not for performance and error clarity reasons this could be the only definition we use. I leave it because it is future proof - if anything outside from the definitions in other methods gets allowed in Base for indexing we will allow it.

But I can drop it as it is not strictly needed for now.

src/dataframerow/dataframerow.jl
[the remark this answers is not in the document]
nalimilan: I guess I was just confused (it was a bit late when I read that... :-).

src/subdataframe/subdataframe.jl
nalimilan: We don't use that approach elsewhere, do we? I'd rather be consistent. Also the error thrown will be probably be unclear for users.

src/subdataframe/subdataframe.jl
bkamins: I have forgotten to add depwarn in one of them. Fixed

bkamins:
OK. I tried to clean up everything (thank you for the patience - this PR is complex unfortunately).

We are left with a `NamedTuple` debate. But as you have said - we will see how it goes. A user always can get a `DataFrame` by selecting `[row]` instead of `row`. The upside is that after this it gets type stable. The downside is immutability.

But I guess in #1533 we investigated all the options and this one had least downsides. How well this goes also hinges on https://github.com/JuliaLang/julia/issues/29417 decision, but I guess it did not get a lot of attention there.

bkamins:
Just to add (is it is better to be sure what we are doing) - I would be also OK, if `df[row, columns]` produced a `DataFrame` not a `NamedTuple` but I guess the majority thinking is that we should drop a dimension if we select a single row only (it has an upside that we have a convenient way to create a `NamedTuple` from a row in a `DataFrame` which could be convenient for users).

bkamins:
OK - I dug through getting this PR green on CI.
Probably it is worth to have another look at it.

Especially as I came to the conclusion that:
* it is very cheap to create `DataFrameRow`;
* I am not sure I know how to create` NamedTuple` from a row of `DataFrame` cheap - @nalimilan have you experimented with that?

src/dataframe/dataframe.jl
```diff
@@ -264,28 +264,24 @@ function Base.getindex(df::DataFrame, row_ind::Integer, col_ind::ColumnIndex)
...
-    throw(ArgumentError("invalid row index: $row_ind of type Bool"))
-end
-
 # df[SingleRowIndex, MultiColumnIndex] => DataFrame (will be NamedTuple)
 function Base.getindex(df::DataFrame, row_ind::Integer, col_inds::AbstractVector)
-    Base.depwarn("Selecting a single row from a `DataFrame` will return a `NamedTuple` in future.", :getindex)
+    if row_ind isa Bool
+        throw(ArgumentError("invalid row index: $row_ind of type Bool"))
+    end
+    Base.depwarn("Selecting a single row from a `DataFrame` will return a `NamedTuple` in the future.", :getindex)
```
nalimilan: There are still several messages where it would be useful to give the syntax which allows avoiding the warning and getting the same result as currently in the future.

nalimilan:
> * I am not sure I know how to create`NamedTuple` from a row of `DataFrame` cheap - @nalimilan have you experimented with that?

I would say it's quite simple, just iterate over columns and extract the requested entry. It should be as fast as we can hope -- which unfortunately should be relatively slow, but not slower than iterating over all entries in a `DataFrameRow`. The main question is whether the compiler will be able to avoid actually accessing all columns when only some of them are used. I guess we should do some benchmarking before switching.

Overall this PR looks OK to me. But these things are so complex than we may well have missed a few cases.

bkamins:
Hopefully I have cleaned up all depwarn messages.

src/dataframe/dataframe.jl
```diff
@@ -302,30 +296,28 @@ end
...
+    new_columns = AbstractVector[dv[row_inds] for dv in columns(df)[selected_columns]]
     return DataFrame(new_columns, Index(_names(df)[selected_columns]))
 end
 
 # df[:, SingleColumnIndex] => AbstractVector
 # df[:, MultiColumnIndex] => DataFrame
 function Base.getindex(df::DataFrame, row_ind::Colon, col_inds)
-    Base.depwarn("indexing with colon as row will create a copy in the future" *
-                 " use df[col_inds] to get the columns without copying", :getindex)
+    Base.depwarn("Sndexing with colon as row will create a copy in the future. " *
```
ararslan: I love Sndexing into data frames

bkamins: This is a new operation I am working on. A combination of "Selecting" and "Indexing" in one go with a bit more of the latter than the former.

nalimilan:
> I would say it's quite simple, just iterate over columns and extract the requested entry. It should be as fast as we can hope -- which unfortunately should be relatively slow, but not slower than iterating over all entries in a `DataFrameRow`. The main question is whether the compiler will be able to avoid actually accessing all columns when only some of them are used. I guess we should do some benchmarking before switching.

I've played a bit with creating tuples for each row and it appears that even when passing a tuple of `N` columns to a function, the compiler isn't able to avoid the creating of a `N`-tuple for each row when we use only a single row (see [this gist](https://gist.github.com/nalimilan/a98ca41e0d5933e5a42281d5bd442135)). Of course the same applies to named tuples. And inside a function taking a data frame, the type instability makes this even harder for the compiler.

In the end `DataFrameRow` is much faster than `Tuple` (I didn't even test `NamedTuple`) when naively iterating over rows: see [this gist](https://gist.github.com/nalimilan/e69d361e1a5982c98587ed59c5600583). So maybe we should make an exception an have `getindex` return a `DataFrameRow`. Even if that's inconsistent with `AbstractArray`, in practice it shouldn't matter much.

@quinnj I guess you've considered these issues a lot when designing Tables.jl?

bkamins:
That is exactly what my experience was (I used something similar to #1564 to test it against `NamedTuples`). Also - as I note in #1564 there - is an issue that tuples you produce in your gist are covariant (a better situation) while named tuples as planned are invariant which makes things even worse.

nalimilan:
Why worse? Can it actually be worse than that anyway? :-)

bkamins:
@nalimilan Is there a conclusion on the `NamedTuple` case and how does it affect this PR? (should we keep it or change `NamedTuple` to `DataFrameRow`?)

nalimilan:
I'd say use `DataFrameRow` for now instead of `NamedTuple`.

bkamins:
Good. One last thing before I clean it up.

As `DataFrameRow` is a kind of view are we sure that we want `df[1, cols]` return `DataFrameRow` or should it be `DataFrame` (this is current behavior).

I am OK with both (given the discussion we had before both alternatives have pros and cons) - but prefer to double check before moving forward.

nalimilan:
I'd go with `DataFrameRow`, as returning a `DataFrame` is really slow (probably even slower than `NamedTuple`).

pdeffebach:
Is it too late to just make our own two types `<: AbstractVector`? That way we can make sure that `DataFrameRow` (attached to a dataframe) behaves exactly like `DataFrameRow` (not attached to a dataframe).

bkamins:
@pdeffebach Can you explain what you mean in a bit more detail please because I am not sure what you mean (in general I guess it is not too late for changes - we try to clean up DataFrames.jl as much as possible before 1.0 release).

pdeffebach:
Something like `AttachedDataFrameRow` and `DetachedDataFrameRow` (without the awkward names)

```
struct AttachedDataFrameRow <: AbstractVector # or similar
    parent::DataFrame
    row::Int
end

Base.getindex(... work with integer using parent Index)
Base.getindex(... work with symbol using parent Index)
function row(dfrow::AttachedDataFrameRow) ...
# Define array interface 
```

Then there is `DetachedDataFrameRow` that tries to keep the exact same API

```
struct DetachedDataFrameRow <: AbstractVector # or similar
    data::Vector
    colindex::Index
    row::Int
end

function DetachedDataFrameRow(df::DataFrame, row::Int)
    data = [df[row, name] for name in names(df)]
    DetachedDataFrameRow(data, getfield(df, :colindex), row)
end

Base.copy(dfrow::AttachedDataFrameRow) DetachedDataFrameRow(parent(dfrow), row(dfrow))

Base.getindex(... work with integer)
Base.getindex(... work with symbol)
function row(dfrow::DetachedDataFrameRow) ...
# Define array interface  
```

I like this plan because it means we don't have to worry about losing behavior when a `DataFrameRow`-like object is attached verses a `copy`, even down to both using a `Index` object. However we also don't have to make compromises: we can index with mutiple symbols `dfrow[:x1, :x2, :x3]` (benefit of `DataFrame`), yet still have `mean(dfrow)` work (benefit of `NamedTuple`. 

On the other hand, I don't like this plan because its one more type we have to export and maintain.

nalimilan:
The problem is that creating a new `Vector` to hold the data is slow, so `DetachedDataFrameRow` wouldn't really be usable as a replacement for `DataFrameRow`.

bkamins:
I have improved the documentation to explain design considerations.

nalimilan:
Thanks. It this good to go then?

nalimilan:
Great work!

nalimilan:
There are lots of deprecation warnings when running tests. Some seem to be potentially problematic for users. For example:
```
┌ Warning: `sdf[colind]` will create a view of `parent(sdf)[colind]` in the future. Use sdf[:, [colind]]` to get a `DataFrame`.
│   caller = isequal(::SubDataFrame{Array{Int64,1}}, ::DataFrame) at abstractdataframe.jl:283
└ @ DataFrames ~/.julia/dev/DataFrames/src/abstractdataframe/abstractdataframe.jl:283
```
This seems to happen when calling `isequal` with a `SubDataFrame`. Maybe we should provide a temporary `isequal` method in deprecated.jl until we remove the deprecation? I guess this applies to other functions.

bkamins:
That was the issue I was afraid of when asking about cleaning up deprecations process.
I will go through the tests of DataFrames.jl, try to clean up and submit a PR today.

PR #1536, opened 2018-09-26
Make setindex! and hcat use column alias where possible

bkamins:
A small PR that prepares for `df[:, cols]` to make a copy of columns. It should introduce no functional changes as of now.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -844,7 +844,7 @@ without(df::AbstractDataFrame, c::Any) = without(df, index(df)[c])
 Base.hcat(df::AbstractDataFrame, x; makeunique::Bool=false) =
     hcat!(copy(df), x, makeunique=makeunique)
 Base.hcat(x, df::AbstractDataFrame; makeunique::Bool=false) =
-    hcat!(x, copy(df), makeunique=makeunique)
+    hcat!(x, df, makeunique=makeunique)
 Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame; makeunique::Bool=false) =
     hcat!(copy(df1), df2, makeunique=makeunique)
 Base.hcat(df::AbstractDataFrame, x, y...; makeunique::Bool=false) =
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -842,11 +842,11 @@ without(df::AbstractDataFrame, c::Any) = without(df, index(df)[c])
 # catch-all to cover cases where indexing returns a DataFrame and copy doesn't
 
 Base.hcat(df::AbstractDataFrame, x; makeunique::Bool=false) =
-    hcat!(df[:, :], x, makeunique=makeunique)
+    hcat!(df[:], x, makeunique=makeunique)
 Base.hcat(x, df::AbstractDataFrame; makeunique::Bool=false) =
-    hcat!(x, df[:, :], makeunique=makeunique)
+    hcat!(x, df[:], makeunique=makeunique)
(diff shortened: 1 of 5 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -842,11 +842,11 @@ without(df::AbstractDataFrame, c::Any) = without(df, index(df)[c])
 # catch-all to cover cases where indexing returns a DataFrame and copy doesn't
 
 Base.hcat(df::AbstractDataFrame, x; makeunique::Bool=false) =
-    hcat!(df[:, :], x, makeunique=makeunique)
+    hcat!(df[:], x, makeunique=makeunique)
```
nalimilan: Using `copy` would be clearer? That's the intended effect here, right?

Maybe also add tests to ensure we don't change this behavior inadvertently?

bkamins: OK added tests and changed to `SubDataFrame`. Had to add a definition of `copy ` to `SubDataFrame`, but I guess it should have been implemented anyway.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -842,11 +842,11 @@ without(df::AbstractDataFrame, c::Any) = without(df, index(df)[c])
 # catch-all to cover cases where indexing returns a DataFrame and copy doesn't
 
 Base.hcat(df::AbstractDataFrame, x; makeunique::Bool=false) =
-    hcat!(df[:, :], x, makeunique=makeunique)
+    hcat!(copy(df), x, makeunique=makeunique)
 Base.hcat(x, df::AbstractDataFrame; makeunique::Bool=false) =
-    hcat!(x, df[:, :], makeunique=makeunique)
+    hcat!(x, copy(df), makeunique=makeunique)
```
nalimilan: This one wasn't and isn't needed, right?

test/subdataframe.jl
```diff
@@ -1,6 +1,14 @@
 module TestSubDataFrame
     using Test, DataFrames
 
+    @testset "copy - SubDataFrame" begin
+        df = DataFrame(x = 1:10, y = 1.0:10.0)
+        sdf = view(df, 1:2, 1:1)
+        @test sdf isa SubDataFrame
+        @test copy(sdf) isa DataFrame
+        @test isequal(sdf, copy(sdf))
```
nalimilan: `==` is enough.

src/subdataframe/subdataframe.jl
```diff
@@ -155,6 +155,8 @@ end
 ##
 ##############################################################################
 
+copy(sdf::SubDataFrame) = sdf[:]
```
nalimilan: You could add `deepcopy` while you're at it.

bkamins: I think not if we want to be consistent with Base where: `deepcopy` of an array view is an array view, while `copy` of an array view is an array.

test/subdataframe.jl
bkamins: But is is a problem? `isequal` is stricter and handles `missing` (of course in this case it makes no difference, but I want to understand the preferred style).

src/abstractdataframe/abstractdataframe.jl
bkamins: Good point.

test/subdataframe.jl
nalimilan: That's just a matter of style. In general I prefer `==` since it's more readable and `isequal` makes it sound like there's something special going on. Also in most files we used `const ≅= isequal`.

src/subdataframe/subdataframe.jl
nalimilan: Funny, I hadn't realized that. That array view method returns a view on a copied array... :-/

Issue #1443, opened 2018-07-08
Make `describe` work for a grouped dataframe


pdeffebach:
Reporting this to remind myself to make this PR. 

Since `GroupedDataFraemes` seem to support broadcasting, this should be easy. 

Though I think `map(f, x::GroupedDataFrame)` does some weird unnecessary printing, so I might track that down too.

pdeffebach:
```
df = DataFrame(rand(5,5))
df[:gender] = ["m", "m", "m", "f", "f"]
gd = groupby(df, :gender)
```

I don't get how broadcasting works with grouped data frames. I was under the impression that `describe` would just work with a broadcast. 

```
println.(gd) # works as expected
describe.(gd)
> ERROR: MethodError: no method matching describe(::DataFrames.GroupedDataFrame)
```

*but* `describe(gd[1])` works, since `gd[1]` is a subdataframe.

pdeffebach:
Using `by` is a bit weird too

```
> by(df, :gender, describe)
│ Row │ gender │ variable │ mean     │ min       │ median    │ max      │ nunique │ nmissing │ eltype  │
├─────┼────────┼──────────┼──────────┼───────────┼───────────┼──────────┼─────────┼──────────┼─────────┤
│ 1   │ m      │ x1       │ 0.334892 │ 0.0531118 │ 0.0600073 │ 0.891557 │ nothing │ nothing  │ Float64 │
│ 2   │ m      │ x2       │ 0.484557 │ 0.351354  │ 0.424645  │ 0.677673 │ nothing │ nothing  │ Float64 │
│ 3   │ m      │ x3       │ 0.561228 │ 0.0260729 │ 0.775023  │ 0.882588 │ nothing │ nothing  │ Float64 │
│ 4   │ m      │ x4       │ 0.41416  │ 0.0112837 │ 0.605188  │ 0.626008 │ nothing │ nothing  │ Float64 │
│ 5   │ m      │ x5       │ 0.596377 │ 0.454716  │ 0.573004  │ 0.761409 │ nothing │ nothing  │ Float64 │
│ 6   │ m      │ gender   │ nothing  │ m         │ nothing   │ m        │ 1       │ nothing  │ String  │
│ 7   │ f      │ x1       │ 0.328554 │ 0.266763  │ 0.328554  │ 0.390345 │ nothing │ nothing  │ Float64 │
[...]

PR #1539, opened 2018-09-26
Add describe for GroupedDataFrame

pdeffebach:
Resolves issue #1443 by using `combine(map(describe, g)) |> groupby`. 

I added a test as well.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -372,3 +372,9 @@ function _aggregate(d::AbstractDataFrame, fs::Vector{T}, headers::Vector{Symbol}
     sort && sort!(res, headers)
     res
 end
+
+# A describe function for a GroupedDataFrame, returning a GroupedDataFrame
+function StatsBase.describe(g::GroupedDataFrame; kwargs...)
+       des = combine(map(d -> describe(d; kwargs...), g)) # a dataframe
+       groupby(des, g.cols) # group it again
+end
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -240,4 +240,16 @@ module TestGrouping
             @test size(v) == (2,2)
         end
     end
+
+    @testset "gouped describe" begin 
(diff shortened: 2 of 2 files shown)


bkamins:
@nalimilan - how does it combine with #1520. especially - I understand that `combine` stays and does exactly the same thing?

nalimilan:
See discussion at https://github.com/JuliaData/DataFrames.jl/issues/1443.

nalimilan:
Is this still needed now that we have `describe.(gd)`?

pdeffebach:
If your grouping column isn't one of the first or last 10 columns, then it will get cut from default `DataFrame` printing so you won't be able to know which group you are looking at each time. Other than that, it's good.

nalimilan:
Wait, I was confused, I meant `map(describe, gd)` or `combine(describe, gd)`. In these cases the grouping column is always listed first AFAICT (I've filed https://github.com/JuliaData/DataFrames.jl/issues/1680 about the difference between the two). Am I right?

It could still make sense to add `describe(::GroupedDataFrame)` in order to remove the row describing the grouping key, which doesn't add any useful information. That would also make `GroupedDataFrame` closer to the `AbstractDataFrame` interface. But that's relatively minor.

pdeffebach:
I think it is good. 

1) it works best when `allgroups = true)` is given, because then intermediate groups aren't omitted. 
2) I agree with Milan that we should omit the header each time we print a grouped DataFrame so it's easier to read.

nalimilan:
OK. So do you want to close this, or adjust it (like I suggested in my last comment)?

bkamins:
@pdeffebach 
bump (I am trying to clean up the GitHub repo for DataFrames.jl so please let me know how you want to proceed with this PR). Thank you!

pdeffebach:
I will close this now. This has come across my mind a bit in the past few months and none of the solutions seems that appetizing. I will ponder other methods and open a new PR if I think of anything i like.

PR #1540, opened 2018-09-27
Fix GroupedDataFrame printing

### LINKED ISSUE #1538: Weird printing of grouped DataFrame on Master
On master, if you use a grouped DataFrame, you get the following: 

I think this was introduced in #1506 

https://github.com/JuliaData/DataFrames.jl/blob/b02ad3c94a3c8f149bab83f44f6d2e54db475c34/src/groupeddataframe/show.jl#L10. 
```
 keys = join(':' .* string.([:A, :B]), ", ")
```
should be 
```
 keys = join(':' .* string.($g.cols), ", ")
```

```
julia> g
GroupedDataFrame with 2 groups based on keys: :A, :B
First Group: 6 rows
│ Row │ x1       │ x2       │ a    │
│     │ Float64  │ Float64  │ Bool │
├─────┼──────────┼──────────┼──────┤
│ 1   │ 0.958328 │ 0.933681 │ true │
│ 2   │ 0.615396 │ 0.48649  │ true │
│ 3   │ 0.894144 │ 0.239123 │ true │
│ 4   │ 0.632327 │ 0.209647 │ true │
│ 5   │ 0.737108 │ 0.148122 │ true │
│ 6   │ 0.612745 │ 0.696692 │ true │
⋮
Last Group: 6 rows
│ Row │ x1       │ x2       │ a     │
│     │ Float64  │ Float64  │ Bool  │
├─────┼──────────┼──────────┼───────┤
│ 1   │ 0.284062 │ 0.694668 │ false │
│ 2   │ 0.175051 │ 0.563256 │ false │
│ 3   │ 0.449609 │ 0.688743 │ false │
│ 4   │ 0.396613 │ 0.237781 │ false │
```

nalimilan:
Actually print grouping columns instead of always `:A, :B`. Also use singular or plural depending on number of groups.

Fixes #1538.

### DIFF
--- a/src/groupeddataframe/show.jl
+++ b/src/groupeddataframe/show.jl
@@ -6,9 +6,10 @@ function Base.show(io::IO, gd::GroupedDataFrame;
                    rowlabel::Symbol = :Row,
                    summary::Bool = true)
     N = length(gd)
-    keystr = N > 1 ? "keys" : "key"
-    keys = join(':' .* string.([:A, :B]), ", ")
-    summary && print(io, "$(typeof(gd)) with $N groups based on $keystr: $keys")
+    keys = join(':' .* string.(gd.cols), ", ")
+    keystr = length(gd.cols) > 1 ? "keys" : "key"
+    groupstr = N > 1 ? "groups" : "group"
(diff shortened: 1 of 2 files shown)


nalimilan:
I really wonder how these two names ended up in my code...

PR #1548, opened 2018-09-30
Small improvements to convert

nalimilan:
Make error message more accurate, avoid an unnecessary conversion and check, and cleanup tests.


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -718,8 +718,16 @@ function Base.convert(::Type{Matrix{T}}, df::AbstractDataFrame) where T
     res = Matrix{T}(undef, n, p)
     idx = 1
     for (name, col) in zip(names(df), columns(df))
-        !(T >: Missing) && any(ismissing, col) && error("cannot convert a DataFrame containing missing values to array (found for column $name)")
-        copyto!(res, idx, convert(Vector{T}, col))
+        try
+            copyto!(res, idx, col)
+        catch err
+            if err isa MethodError && err.f == convert &&
+               !(T >: Missing) && any(ismissing, col)
```
bkamins: Is this special case really helpful? If you try to convert `Missing` to some other concrete type you will get essentially the same message e.g.:
```
MethodError: Cannot `convert` an object of type Missing to an object of type Float64
```
and in other cases we do not provide a column information anyway. Maybe simply print information which column was offending, what was its `eltype` and rethrow the error?

nalimilan: Yeah, it's a useful special case to have since you may have lots of `Union{T,Missing}` columns, but few missing values. That's quite different from other columns which are typically homogeneous, in which case you just need to check the eltypes of columns.

I'm not sure how I could rethrow the error after adding information about the column name. Just printing a message to stdout would be ugly and wouldn't work with `try... catch`. I guess we'd need to define a special exception type which would contain the information from the `MethodError` plus the name of the column. Maybe not worth it, at least at this stage.

nalimilan:
> in general I understand that `copyto!` tries to convert what it copies anyway - right?

Yes, like any assignment. I suspect this `convert` was needed at some point when we used `DataArray`.

Issue #1543, opened 2018-09-28
DataFrameRow needs to support 'haskey'


LeoK987:
since its elements can be accessed via keys, it better allows 'haskey'.  Just a request.

LeoK987:
PR made.

bkamins:
Close?

PR #1550, opened 2018-10-01
added haskey in DataFrameRow

LeoK987:


LeoK987:
Issue #1543

nalimilan:
Thanks. Can you add a test too?

LeoK987:
test added

LeoK987:
Didn't know about negative test.  Thank you @nalimilan

bkamins:
I guess we can merge it for now, but in the long term `haskey` for `DataFrame` should be removed (after we switch to row-based major approach), so maybe better to write `haskey(index(parent(df)), key)`?

OTOH it is useful to be able to efficiently check if a `DataFrame` has some column and `col in names(df)` is not efficient. @nalimilan Any thoughts on this?

nalimilan:
> I guess we can merge it for now, but in the long term `haskey` for `DataFrame` should be removed (after we switch to row-based major approach), so maybe better to write `haskey(index(parent(df)), key)`?
> 
> OTOH it is useful to be able to efficiently check if a `DataFrame` has some column and `col in names(df)` is not efficient. @nalimilan Any thoughts on this?

Yes, that's a tough decision. We could keep `haskey` even if that's not consistent with the view of data frames as collections of rows. At least it's consistent with `getindex(df, col)`.

PR #1552, opened 2018-10-03
fix a corner case of showing types

bkamins:
@nalimilan Fixes of by one error per discussion in #1551.

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -42,7 +42,7 @@ function compacttype(T::Type, maxwidth::Int=8)
         sT = string(T)
         suffix = "⍰"
         # handle the case when after removing Missing union type name is short
-        length(sT) < 8 && return sT * suffix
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/show.jl
```diff
@@ -42,7 +42,7 @@ function compacttype(T::Type, maxwidth::Int=8)
         sT = string(T)
         suffix = "⍰"
         # handle the case when after removing Missing union type name is short
-        length(sT) < 8 && return sT * suffix
+        length(sT) ≤ 8 && return sT * suffix
```
nalimilan: Shoudn't this use `maxwidth`?

bkamins: Earlier we have agreed on the rule:

> if whole type name fits `maxwidth` then use it otherwise always truncate to 8 to have consistency when printing types.

This is especially important for narrow columns (e.g. when `maxwidth` is 1 or 2)

PR #1556, opened 2018-10-05
Remove obsolete special paths for Missing

nalimilan:
These no longer make any difference to generated code.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -528,18 +528,17 @@ end
 ##############################################################################
 
 function _nonmissing!(res, col)
-    # workaround until JuliaLang/julia#21256 is fixed
-    eltype(col) >: Missing || return
-
     @inbounds for (i, el) in enumerate(col)
         res[i] &= !ismissing(el)
     end
+    return nothing
 end
 
 function _nonmissing!(res, col::CategoricalArray{>: Missing})
     for (i, el) in enumerate(col.refs)
         res[i] &= el > 0
     end
+    return nothing
 end
 
 
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -62,11 +62,13 @@ Base.values(r::DataFrameRow) = ntuple(col -> parent(r)[col][row(r)], length(r))
 
 # hash column element
 Base.@propagate_inbounds hash_colel(v::AbstractArray, i, h::UInt = zero(UInt)) = hash(v[i], h)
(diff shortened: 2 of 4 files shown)


src/dataframerow/dataframerow.jl
```diff
@@ -62,11 +62,13 @@ Base.values(r::DataFrameRow) = ntuple(col -> parent(r)[col][row(r)], length(r))
 
 # hash column element
 Base.@propagate_inbounds hash_colel(v::AbstractArray, i, h::UInt = zero(UInt)) = hash(v[i], h)
-Base.@propagate_inbounds hash_colel(v::AbstractCategoricalArray, i, h::UInt = zero(UInt)) =
```
bkamins: why is this not needed any more?

src/dataframerow/dataframerow.jl
```diff
@@ -62,11 +62,13 @@ Base.values(r::DataFrameRow) = ntuple(col -> parent(r)[col][row(r)], length(r))
...
-    hash(CategoricalArrays.index(v.pool)[v.refs[i]], h)
 Base.@propagate_inbounds function hash_colel(v::AbstractCategoricalArray{>: Missing}, i, h::UInt = zero(UInt))
     ref = v.refs[i]
-    ref == 0 ? hash(missing, h) : hash(CategoricalArrays.index(v.pool)[ref], h)
+    if eltype(v) >: Missing && ref == 0
+        hash(missing, h)
+    else
+        hash(CategoricalArrays.index(v.pool)[ref], h)
```
bkamins: I understand that here `ref` will never be `0` right?

nalimilan: Yes (at least if the array isn't corrupt). This is really equivalent to the previous code, except that the type check is inside the function now.

src/dataframerow/dataframerow.jl
nalimilan: Woops, I forgot to widen the signature of the following method.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -528,18 +528,17 @@ end
 ##############################################################################
 
 function _nonmissing!(res, col)
-    # workaround until JuliaLang/julia#21256 is fixed
-    eltype(col) >: Missing || return
-
     @inbounds for (i, el) in enumerate(col)
         res[i] &= !ismissing(el)
     end
+    return nothing
```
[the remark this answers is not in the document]
nalimilan: Yes (I've checked that it's indeed the case).

PR #1558, opened 2018-10-07
Improve GroupedDataFrame API

bkamins:
@nalimilan A parallel effort to `getindex` cleanup for `AbstractDataFrame` but in this case I guess it is more clear cut what is the desired behavior.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -117,20 +117,11 @@ Base.getindex(gd::GroupedDataFrame, idxs::Colon) =
 function Base.:(==)(gd1::GroupedDataFrame, gd2::GroupedDataFrame)
     res = gd1.parent == gd2.parent
     ismissing(res) && return missing
-    res &&
-    (gd1.cols == gd2.cols) &&
-    (gd1.idx == gd2.idx) &&
-    (gd1.starts == gd2.starts) &&
-    (gd1.ends == gd2.ends)
+    res && gd1.cols == gd2.cols
 end
 
-function Base.isequal(gd1::GroupedDataFrame, gd2::GroupedDataFrame)
-    isequal(gd1.parent, gd2.parent) &&
-    isequal(gd1.cols, gd2.cols) &&
-    isequal(gd1.idx, gd2.idx) &&
-    isequal(gd1.starts, gd2.starts) &&
-    isequal(gd1.ends, gd2.ends)
-end
+Base.isequal(gd1::GroupedDataFrame, gd2::GroupedDataFrame) =
+    isequal(gd1.parent, gd2.parent) && isequal(gd1.cols, gd2.cols)
 
 Base.names(gd::GroupedDataFrame) = names(gd.parent)
 _names(gd::GroupedDataFrame) = _names(gd.parent)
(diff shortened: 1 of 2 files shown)


test/grouping.jl
```diff
@@ -240,4 +240,19 @@ module TestGrouping
             @test size(v) == (2,2)
         end
     end
+
+    @testset "getindex" begin
+        df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
+                       b = repeat([2, 1], outer=[4]))
+        gd = groupby(df, :a)
+        @test gd[1] isa SubDataFrame
```
nalimilan: While you're at it, better test that the result is equal to what you expect. Same below.

bkamins: OK. :). I have pushed a commit with some additional related features.

src/groupeddataframe/grouping.jl
```diff
@@ -107,10 +107,30 @@ Compat.lastindex(gd::GroupedDataFrame) = length(gd.starts)
...
+    GroupedDataFrame(gd.parent, gd.cols, gd.idx, gd.starts, gd.ends)
+
+function Base.:(==)(gd1::GroupedDataFrame, gd2::GroupedDataFrame)
+    res = gd1.parent == gd2.parent
+    ismissing(res) && return missing
+    res &&
+    (gd1.cols == gd2.cols) &&
+    (gd1.idx == gd2.idx) &&
+    (gd1.starts == gd2.starts) &&
+    (gd1.ends == gd2.ends)
```
nalimilan: If I'm not mistaken, checking `cols` is enough.

We could also decide that what matters is the actual grouping, even if it's based on different columns, in which case checking `idx` and `starts` would be necessary and sufficient. But that's more work without a clear advantage.

nalimilan: At https://github.com/JuliaData/DataFrames.jl/pull/1520 I force `cols` to contain integers. You could apply that change here.

bkamins: OK. So I will leave only `cols` check and after #1520 all will be guaranteed to be consistent. Right?

nalimilan:
CI fails due to deprecated methods not throwing errors yet.

bkamins:
I had a typo. Fixed.
However, I think it is time to clean up some old deprecations as we are adding to add many new ones soon and currently CI is swamped by deprecations.

nalimilan:
Yes, once we are done with adding all the planned deprecations, we should tag a new minor release and get rid of all old deprecations that are blocking us from making changes. (Though I prefer keeping deprecations which don't hurt since examples using them can often be found online.)

bkamins:
Agreed, but IMHO there should be some point, when we move to DataFrames 1.0.0, when all deprecations are removed.

bkamins:
Now I have an interesting question. CI passes on 1.0, but fails on 0.7, because 0.7 accepted `x[true]` when `x` is a vector, and 1.0 does not accept it any more (and this is expected by the tests).

Shall I handle 0.7 or we concentrate on 1.0?

nalimilan:
I'd just make the test conditional on the version.

bkamins:
@nalimilan I leave it up to you if we want to merge it now or later (merging will create conflicts with #1520 which will need rebasing).

PR #1559, opened 2018-10-07
Remove mutable structs

bkamins:
This PR removes `mutable` from struct definitions where possible.
Exceptions are: `GroupedDataFrame` (this is handled in #1558) and `Index` (which is mutated internally with `names!` - but maybe we can decide to reimplement `names!` and avoid `mutable` there also).

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -82,7 +82,7 @@ size(df1)
 ```
 
 """
-mutable struct DataFrame <: AbstractDataFrame
+struct DataFrame <: AbstractDataFrame
     columns::Vector{AbstractVector}
     colindex::Index
 
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1123,7 +1123,7 @@ function permutecols!(df::DataFrame, p::AbstractVector)
         throw(ArgumentError("$p is not a valid column permutation for this DataFrame"))
     end
     permute!(columns(df), p)
-    permutecols!(getfield(df, :colindex), p)
+    @inbounds permute!(index(df), p)
     df
 end
 
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1123,14 +1123,10 @@ function permutecols!(df::DataFrame, p::AbstractVector)
         throw(ArgumentError("$p is not a valid column permutation for this DataFrame"))
     end
     permute!(columns(df), p)
-    newindex = Index(names(df)[p])
-    colindex = getfield!(df, :colindex)
-    colindex.names = newindex.names
-    colindex.lookup = newindex.lookup
+    permutecols!(getfield(df, :colindex), p)
     df
 end
 
 function permutecols!(df::DataFrame, p::AbstractVector{Symbol})
-    lu = index(df).lookup
-    permutecols!(df, [lu[x] for x in p])
+    permutecols!(df, index(df)[p])
 end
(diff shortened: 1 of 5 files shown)


src/dataframe/dataframe.jl
```diff
@@ -1123,7 +1123,10 @@ function permutecols!(df::DataFrame, p::AbstractVector)
         throw(ArgumentError("$p is not a valid column permutation for this DataFrame"))
     end
     permute!(columns(df), p)
-    setfield!(df, :colindex, Index(names(df)[p]))
+    newindex = Index(names(df)[p])
```
nalimilan: This isn't very clean as it breaks the `Index` abstraction. Better implement `permute!(::Index, ...)`.

src/other/index.jl
```diff
@@ -61,6 +61,15 @@ rename!(f::Function, x::Index) = rename!(x, [(x=>f(x)) for x in x.names])
 rename(x::Index, args...) = rename!(copy(x), args...)
 rename(f::Function, x::Index) = rename!(f, copy(x))
 
+function permutecols!(x::Index, p::AbstractVector)
+    # to be callsed by permutecols! for DataFrame
```
nalimilan: I'd rather call this `permute!` (just like I've kept `merge!`), as there's no ambiguity with rows vs. columns.

Also, instead of assuming that the permutation is valid, you can use `@boundscheck`, mark the function with `@propagate_inbounds`, and use `@inbounds` when calling it.

Finally, I think you could make this slightly faster by permuting `x.names`, and updating `x.lookup` in place.

src/dataframe/dataframe.jl
```diff
@@ -1123,11 +1123,10 @@ function permutecols!(df::DataFrame, p::AbstractVector)
         throw(ArgumentError("$p is not a valid column permutation for this DataFrame"))
     end
     permute!(columns(df), p)
-    setfield!(df, :colindex, Index(names(df)[p]))
+    permutecols!(getfield(df, :colindex), p)
```
nalimilan: `index(df)` is shorter.

src/other/index.jl
bkamins: OK. In the process I thought then that we can make `Index` immutable by rewriting `names!` in a similar fashion.

bkamins:
CI passes apart from a standard failure.
The question is if the changes to `Index` making it immutable are OK.

nalimilan:
Making `Index` immutable can't hurt, and reusing the `lookup` dict makes sense since the number of entries is the same and most of the names won't have changed, so the structure should be similar.

bkamins:
OK. Let us wait a bit if there any more comments on this and merge if none. Thanks!

bkamins:
OK - I know that unconsciously bothered me with this PR.

It will be breaking with serialization, JLD2 and BSON (and we would get a similar issue as with `Char` between Julia 0.6 and 1.0). What would be the best adoption path? Do not merge it yet and print a warning on package load that this is going to change?

nalimilan:
JLD2 didn't support data frames a few weeks ago, and AFAIK there are still lots of problems with `missing` support. So I wouldn't worry too much. We'll tag a new "major" (minor) release -- better break this now before things stabilize. People would better use Feather to store data sets at this point. 

But why would making something immutable break restoring files? Maybe that could be fixed?

bkamins:
OK. The problem is that e.g. JLD2 uses a different strategy to load mutable and immutable types (this is the part of code I have fixed there for Julia 1.0).

bkamins:
But as you have said - let us change it.

bkamins:
So - do we merge it? I prefer to double check because this is a significant change.

Issue #1514, opened 2018-09-18
Review row vs. column orientation of API


nalimilan:
It's been noted several times that we don't have a consistent view on whether a data frame is a collection of rows or of columns. We should decide which one it is so that all exported functions and collection functions from Base we implement operate either on rows or on columns. Then functions which operate on the other dimension should mention it explicitly in their names, e.g. with the suffix `rows` or `cols` (as in IndexedTables).

See also: https://github.com/JuliaData/DataFrames.jl/issues/406, https://github.com/JuliaData/DataFrames.jl/issues/1200, https://github.com/JuliaData/DataFrames.jl/pull/1459, https://github.com/JuliaData/DataFrames.jl/issues/1513, https://github.com/JuliaData/DataFrames.jl/issues/1377.

Below is the list of all functions which are either row- or column-oriented. Those marked with * could be interpreted/useful both as row-oriented or column-oriented, and are therefore the most problematic (i.e. the ones to make consistent).

Row-oriented functions:
`append!`*
`filter`/`filter!`*
`head` & `tail`*

`sort`/`sort!` & `sortperm` & `issorted`
`unique` & `nonunique`
`completecases`
`dropmissing`/`dropmissing!`

Column-oriented functions:
`delete!`*
`insert!` (inconsistent with `append!` and `push!`)*
`merge!`*
`haskey` & `get` (old `Associative` interface)*
`length`*
`getindex` (with single argument)*

`allowmissing`/`allowmissing!`
`disallowmissing`/`disallowmissing!`
`categorical!`
`names` & `rename`
`eltypes`
`describe`

Both row- and column-oriented functions:
`empty!`

Functions with explicit name:
`deleterows!` (compare with `delete!`)
`permutecols!`
`nrow`/`ncol`
`hcat`/`vcat`
`colwise`
`eachrow`/`eachcol`

Complex:
`by`/`groupby` -> mostly row-oriented
`aggregate` -> mostly column-oriented (akin to `colwise`)

--------

Overall, it seems it would be easier and more natural to get rid of *column*-oriented functions marked with *, and consider that a data frame is a collection of rows. Many of the problematic functions are either inherited from the time when `DataFrame` implemented the `Associative` interface, which isn't very useful in practice. Others (like `insert!`) are in conflict with similar functions which operate on rows. I suggest we rename the ones which are deemed useful to add the `cols` suffix. Functions with `rows` in their names could drop that suffix. This means:
[...]

PR #1560, opened 2018-10-07
Deprecate delete!, insert! and merge!

nalimilan:
This implements the two least controversial and most useful changes from https://github.com/JuliaData/DataFrames.jl/issues/1514:

Rename delete! to deletecols! and insert! to insertcol! to fit the definition of data frames
as collections of rows. This will allow renaming deleterows! to delete! after the deprecation
period.

Remove merge! since it is trivial to write by hand, and it is likely to confuse users coming
from languages where joins are performed using function "merge".

Also deprecate `df[col] = nothing` in favor of `deletecols!(df, col)`: this is an R-ism which doesn't make sense in Julia and isn't more convenient than the more standard function call form.


src/dataframe/dataframe.jl
```diff
@@ -700,8 +697,8 @@ julia> insert!(d, 1, 'a':'c', :b)
 ```
 
 """
-function Base.insert!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol;
-                      makeunique::Bool=false)
+function insertcol!(df::DataFrame, col_ind::Int, item::AbstractVector, name::Symbol;
+                    makeunique::Bool=false)
```
bkamins: When updating this you might want to fix the following bug:
```
julia> df = DataFrame(a=Int[], b=[], c=[])
0×3 DataFrame


julia> insert!(df, 2, [1,2,3], :d)
0×4 DataFrame


julia> df
0×4 DataFrame


julia> df[:a]
0-element Array{Int64,1}

julia> df[:d]
3-element Array{Int64,1}:
 1
 2
 3
```

bkamins:
Additionally - do you care reviewing internal methods defined in `other/index.jl` that do similar things, e.g. `merge!`. This is not strictly needed, but would unify codebase.

nalimilan:
I've seen that method, but it sounds OK since we do call `merge` on `Index` objects and it makes sense there (since there's no ambiguity between rows and columns).

src/dataframe/dataframe.jl
nalimilan: Funny. I don't understand why the code explicitly allows a non-matching length when `size(df, 1) == 0`. Any idea?

bkamins: My idea was that it is a bug 😄. I assume the intention was `size(df, 2) == 0` so that we allow to add a column to a `DataFrame` that has `0` columns.

Actually that is why I prefer `nrows` and `ncols` to `size` (although I know that there are comments that those should be removed).

nalimilan: Sounds plausible. I've pushed a commit to this that.

nalimilan:
I've realized the API for `insertcol` was weird since it didn't match `insert!` and the name of the column appeared at the end. I've pushed a commit to use pairs or keyword arguments instead, tell me what you think. Something I wonder is whether we should call the function `insertcols!` and allow inserting several columns in the future. The name would then match the suffix of other functions.

bkamins:
Very nice. Actually I would go for `insertcols!` with multiple columns (this should not be that complex as you can just add column one by one).

It is amazing that whatever part of DataFrames.jl API is touched there is something that can be improved.

nalimilan:
OK, I've pushed a commit renaming `insertcol!` to `insertcols`. I'll leave it for later to support adding several columns.

> It is amazing that whatever part of DataFrames.jl API is touched there is something that can be improved.

Yeah, but I'm not sure that's a good sign... :-/

bkamins:
> Yeah, but I'm not sure that's a good sign... :-/

My fear is that we are cleaning up for the future users at the risk of breaking things for current users. However, the alternative is what R has with currently three implementations of data frame like object for legacy reasons.

PR #1562, opened 2018-10-08
fix == for GroupedDataFrame

bkamins:
@nalimilan That is how later we have to fix things 😄.
A small fix to `==` because when `cols` are different we know that `GroupedDataFrame`s do not match even if their parents have `missing`s.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -115,13 +115,11 @@ Base.getindex(gd::GroupedDataFrame, idxs::Colon) =
     GroupedDataFrame(gd.parent, gd.cols, gd.idx, gd.starts, gd.ends)
 
 function Base.:(==)(gd1::GroupedDataFrame, gd2::GroupedDataFrame)
-    res = gd1.parent == gd2.parent
-    ismissing(res) && return missing
-    res && gd1.cols == gd2.cols
+    gd1.cols == gd2.cols && gd1.parent == gd2.parent
 end
 
 Base.isequal(gd1::GroupedDataFrame, gd2::GroupedDataFrame) =
-    isequal(gd1.parent, gd2.parent) && isequal(gd1.cols, gd2.cols)
+    isequal(gd1.cols, gd2.cols) && isequal(gd1.parent, gd2.parent)
 
 Base.names(gd::GroupedDataFrame) = names(gd.parent)
 _names(gd::GroupedDataFrame) = _names(gd.parent)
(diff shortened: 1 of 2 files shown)


bkamins:
Actually there is one more tricky case here. What do you think we should do with the following cases (now I remember what I had in mind initially with all those checks):
1. `sort` was used so two grouped data frames can have the same `cols` but different order;
2. `skipmissing` was set to `true` so source can have missing values but they get removed by this option;
3. `skipmissing` was set to `false` but `missing` only appears in grouping columns (so implicitly the user says that they should be treated as equal);

It seems that we should compare `cols` and if they are equal actually we should iterate through the `SubDataFrames` generated and compare them for equality.

This will not cover case 3 - i.e. `==` will return `missing` when `missing`s are present (I guess it is OK, but I wanted to list this case so that we are clear what we want).

A completely different approach would be to say that `==` and `isequal` for `GroupedDataFrame` is the same and we check if `paent`s are `isequal` and that have the same `cols`, `starts` and `ends`. So essentially we say that `GroupedDataFrame` is based on the fact that it represents the same grouping disregarding the contents. This would be kind of similar behavior to:
```
julia> Dict(missing=>1) == Dict(missing=>1)
true
```

nalimilan:
OK. Checking each `SubDataFrame` sounds like the best semantics and the simplest implementation. I guess there's a faster way of doing that when parents are `===`, but we can always add optimizations later if it matters.

bkamins:
If `DataFrames` are `===` this is still tricky because of possibility of `missing`.

As a side note. Actually this is an interesting case. If `x` and `y` are collections and `x === y`, but they contain `missing` then `x == y` is `missing` although theoretically we know that this unknown `missing` must have the same value (it is unknown but it must be the same as the object is the same). I do not want to say we should change this, but I never thought about it before.

bkamins:
OK - based on what I have pushed it is, in general, possible that two `GroupedDataFrame`s have different parents but compare as equal.

The cases are:
* they have different parents, but sorting makes `GroupedDataFrame`s look the same;
* they have different parents, but `skipmissing` makes `GroupedDataFrame`s look the same;

I think that it is actually OK (i.e. as long as `GroupedDataFrame`s produce the same `SubDataFrame`s and have the same grouping columns they should compare as equal).

src/groupeddataframe/grouping.jl
```diff
@@ -115,13 +115,16 @@ Base.getindex(gd::GroupedDataFrame, idxs::Colon) =
     GroupedDataFrame(gd.parent, gd.cols, gd.idx, gd.starts, gd.ends)
 
 function Base.:(==)(gd1::GroupedDataFrame, gd2::GroupedDataFrame)
-    res = gd1.parent == gd2.parent
-    ismissing(res) && return missing
-    res && gd1.cols == gd2.cols
+    gd1.cols == gd2.cols &&
+    length(gd1) == length(gd2) &&
```
nalimilan: Add indentation.

bkamins: fixed (if this is what you meant)

bkamins:
CI is good.

PR #1568, opened 2018-10-15
fix show methods for csv/tsv

pfitzseb:


nalimilan:
Thanks! Can you add tests so that these functions are not broken again?

pfitzseb:
Done. Should I squash or will you do that on merge?

PR #1571, opened 2018-10-16
Extend setindex api

### LINKED ISSUE #1569: Creating an empty DataFrame is unwieldy and has unexpected behavior
The canonical way of creating an empty `DataFrame` according to [the docs](https://juliadata.github.io/DataFrames.jl/stable/man/getting_started.html#The-DataFrame-Type-1) (see the **Constructing Row by Row** header) is a bit unwieldy and has unexpected behavior.

The constructor that creates columns based on keyword arguments seems to be mostly for demo purposes, so I'll use the constructor that takes in a dictionary (or a vararg parameter of pairs) in my examples.

Suppose I want to create two `DataFrame`s to hold measurements before and after some event. Getting the data for each row is not trivial, so the overhead of creating a dataframe row by row is not an issue for me. It's the most convenient way to do it and won't have a big impact on the runtime of my program, so I'll create both dataframes row by row. Since they have the same columns, I'll make one column dictionary, and create two dataframes with the same dictionary, as shown in the following example.
```
columnnames = ["x", "y", "z"]
columns = [Symbol(col) => Float64[] for col in columnnames]
df1 = DataFrame(columns...)
df2 = DataFrame(columns...)
```
Then, if I add a row to `df1`, since the same `Float64` vectors in `columns` are used in the underlying data storage of both dataframes, it also adds the row to `df2`. This is very unexpected, and took me a little while to figure out what exactly was going on.

The problem could be fixed simply by changing line 100 of `src/dataframe/dataframe.jl` from `return new(columns, colindex)` to `return new(map(copy, columns), colindex)`, however, I think these constructors are still awkward to use.

The best current alternative is using `DataFrame([Float64, Float64, Float64], [:x, :y, :z], 0)` but I still don't like this, for two reasons. First, the names of the columns are separated from the types, which I'm not a fan of--I'd prefer it if I could pass in a dictionary mapping column names to column types (if you try to do that now it creates a column with the type itself in the first row), and second, the argument for the number of rows seems superfluous--given the current API of DataFrames, I don't see what the use cases would be for initializing an empty dataframe with a given number of rows. It makes sense if you know beforehand what size your data will be, but then (as far as I can tell) there's no convenient way to set a row to something. So I'd like it if there were some `setindex!` methods that paralleled the available `push!` methods (e.g. `setindex!(df::DataFrame, row::Union{AbstractDict, NamedTuple}, row_ind::Real, col_ind::Colon)`).

So in conclusion, I think this should be fixed. We could just add a `copy` when dataframes get constructed with empty vectors, to protect the user, or we could extend the API to make it more intuitive. I'd propose adding constructors that take in a dictionary or a list of pairs where the column values are types instead of vectors, with an optional size parameter, and also adding methods to `setindex!` so that it can be used the same way `push!` can be, to set rows in empty dataframes of a predetermined size so that creating a dataframe row by row isn't quite as slow.

I'm happy to help implement any of my suggestions, but I just didn't want to start working on them before knowing that they would likely be approved.

coreywoodfield:
This resolves #1569 

I wasn't sure if we want the second commit, but it was pretty easy to add so I just did it (and currently, trying to make such a call interprets the iterable as a scalar and broadcasts it to all the columns in the row - that may be desirable when the columns hold collections, but that seems like an uncommon use case, if the columns don't hold collections it just throws an error and doesn't do anything). I'm happy to remove it though if we don't want it.

One thing that I considered that could be added to this is that perhaps we want to allow updating just some values in a row with a named tuple (or a dictionary), I just wasn't sure what syntax we would want. It'd be simplest to allow using a named tuple missing some columns in the assignment `df[1, :] = tuple`, and only update the columns that the tuple has, but it'd be more clear if we required `df[1, keys(tuple)] = tuple`. I'm happy to add either one to this pull request if we want.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -589,14 +589,13 @@ function Base.setindex!(df::DataFrame,
                         row_ind::Real,
                         col_inds::AbstractVector{<:ColumnIndex})
     checkdimensions(new_df, row_ind, col_inds)
-    if usenames(df, new_df, col_inds)
-        for col_name in _names(new_df)
-            insert_single_entry!(df, new_df[col_name][1], row_ind, col_name)
-        end
-    else # use column order
-        for (j, col_ind) in enumerate(col_inds)
-            insert_single_entry!(df, new_df[j][1], row_ind, col_ind)
-        end
+    if !usenames(df, new_df, col_inds)
+        msg = "Column names in inserted dataframe don't match the columns " *
+              "being inserted into"
+        throw(ArgumentError(msg))
+    end
+    for col_name in _names(new_df)
+        insert_single_entry!(df, new_df[col_name][1], row_ind, col_name)
     end
     df
 end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -569,12 +569,24 @@ function checkdimensions(new_df::AbstractDataFrame,
     end
 end
 
-function usenames(df::DataFrame,
+function checknames(df::DataFrame,
                   new_df::AbstractDataFrame,
                   col_inds::AbstractVector{<:ColumnIndex})
     new_names = _names(new_df)
     subset = all(name -> haskey(df, name), new_names)
-    subset && issetequal(index(df)[new_names], index(df)[col_inds])
+    if !subset || !issetequal(index(df)[new_names], index(df)[col_inds])
+        msg = "Column names in inserted dataframe don't match the columns " *
+              "being inserted into"
+        throw(ArgumentError(msg))
+    end
+end
+
+function checkpreconditions(df::DataFrame,
+                            new_df::AbstractDataFrame,
+                            row_inds::Any,
+                            col_inds::Any)
+    checkdimensions(new_df, row_inds, col_inds)
+    checknames(df, new_df, col_inds)
 end
 
 # df[SingleRowIndex, MultiColumnIndex] = 1-Row DataFrame
@@ -588,12 +600,7 @@ function Base.setindex!(df::DataFrame,
                         new_df::AbstractDataFrame,
                         row_ind::Real,
                         col_inds::AbstractVector{<:ColumnIndex})
-    checkdimensions(new_df, row_ind, col_inds)
-    if !usenames(df, new_df, col_inds)
-        msg = "Column names in inserted dataframe don't match the columns " *
-              "being inserted into"
-        throw(ArgumentError(msg))
-    end
+    checkpreconditions(df, new_df, row_ind, col_inds)
     for col_name in _names(new_df)
         insert_single_entry!(df, new_df[col_name][1], row_ind, col_name)
     end
@@ -623,8 +630,9 @@ function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
                         row_inds::AbstractVector{<:Real},
                         col_inds::AbstractVector{<:ColumnIndex})
-    for j in 1:length(col_inds)
-        insert_multiple_entries!(df, new_df[j], row_inds, col_inds[j])
+    checkpreconditions(df, new_df, row_inds, col_inds)
+    for col_name in _names(new_df)
+        insert_multiple_entries!(df, new_df[col_name], row_inds, col_name)
     end
     return df
 end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -500,6 +500,34 @@ function Base.setindex!(df::DataFrame,
     return df
 end
 
+# df[SingleRowIndex, :] = Union{AbstractDict, NamedTuple}
+function Base.setindex!(df::DataFrame,
+                        row::Union{AbstractDict, NamedTuple},
+                        row_ind::Real,
+                        ::Colon)
+    old_row = df[row_ind, :]
+    for nm in _names(df)
+        try
+            # after deprecation replace this call with `val = row[nm]`
+            val = get(row, nm) do
+                v = row[string(nm)]
+                msg = "setindex!(::DataFrame, ::AbstractDict, ::Real, " *
+                      "::Colon) with AbstractDict keys other than Symbol " *
+                      "is deprecated"
+                Base.depwarn(msg, :setindex!)
+                v
+            end
+            insert_single_entry!(df, val, row_ind, nm)
+        catch
+            # clean up partial row
+            df[row_ind, :] = old_row
+            msg = "Error adding value to column :$nm."
+            throw(ArgumentError(msg))
+        end
+    end
+    df
+end
+
 # df[MultiRowIndex, SingleColumnIndex] = AbstractVector
 function Base.setindex!(df::DataFrame,
                         v::AbstractVector,
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -527,6 +527,29 @@ function Base.setindex!(df::DataFrame,
     end
     df
 end
+# df[SingleRowIndex, MultiColumnIndex] = iterable
+function Base.setindex!(df::DataFrame,
+                        iterable::AbstractVector,
+                        row_ind::Real,
+                        col_inds::AbstractVector{<:ColumnIndex})
+    if length(iterable) != length(col_inds)
+        msg = "Length of iterable does not match DataFrame column count."
+        throw(ArgumentError(msg))
+    end
+    old_row = df[row_ind, col_inds]
+    for (val, col_ind) in zip(iterable, col_inds)
+        try
+            insert_single_entry!(df, val, row_ind, col_ind)
+        catch
+            #clean up partial row
+            df[row_ind, col_inds] = old_row
+            msg = "Error adding $val to column :$(_names(df)[col_ind])." *
+                  " Possible type mis-match."
+            throw(ArgumentError(msg))
+        end
+    end
+    df
+end
 
 # df[MultiRowIndex, SingleColumnIndex] = AbstractVector
 function Base.setindex!(df::DataFrame,
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -466,23 +466,6 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
     insert_single_entry!(df, v, row_ind, col_ind)
 end
 
-# df[SingleRowIndex, MultiColumnIndex] = 1-Row DataFrame
-function Base.setindex!(df::DataFrame,
-                        new_df::DataFrame,
-                        row_ind::Real,
-                        col_inds::AbstractVector{Bool})
-    setindex!(df, new_df, row_ind, findall(col_inds))
-end
-function Base.setindex!(df::DataFrame,
-                        new_df::DataFrame,
-                        row_ind::Real,
-                        col_inds::AbstractVector{<:ColumnIndex})
-    for j in 1:length(col_inds)
-        insert_single_entry!(df, new_df[j][1], row_ind, col_inds[j])
-    end
-    return df
-end
-
 # df[SingleRowIndex, :] = Union{AbstractDict, NamedTuple}
 function Base.setindex!(df::DataFrame,
                         row::Union{AbstractDict, NamedTuple},
@@ -572,6 +555,50 @@ function Base.setindex!(df::DataFrame,
                         col_ind::ColumnIndex)
     insert_multiple_entries!(df, v, row_inds, col_ind)
     return df
+end
+
+function checkdimensions(new_df::AbstractDataFrame,
+                         row_inds::Any,
+                         col_inds::Any)
+    new_rows, new_cols = size(new_df)
+    # note that length(::Number) == 1
+    if (new_rows, new_cols) != (length(row_inds), length(col_inds))
+        msg = "Tried to assign $new_rows×$new_cols dataframe to " *
+              "$(length(row_inds))×$(length(col_inds)) destination"
+        throw(DimensionMismatch(msg))
+    end
+end
+
+function usenames(df::DataFrame,
+                  new_df::AbstractDataFrame,
+                  col_inds::AbstractVector{<:ColumnIndex})
+    new_names = _names(new_df)
+    subset = all(name -> haskey(df, name), new_names)
+    subset && issetequal(index(df)[new_names], index(df)[col_inds])
+end
+
+# df[SingleRowIndex, MultiColumnIndex] = 1-Row DataFrame
+function Base.setindex!(df::DataFrame,
+                        new_df::AbstractDataFrame,
+                        row_ind::Real,
+                        col_inds::AbstractVector{Bool})
+    setindex!(df, new_df, row_ind, findall(col_inds))
+end
+function Base.setindex!(df::DataFrame,
+                        new_df::AbstractDataFrame,
+                        row_ind::Real,
+                        col_inds::AbstractVector{<:ColumnIndex})
+    checkdimensions(new_df, row_ind, col_inds)
+    if usenames(df, new_df, col_inds)
+        for col_name in _names(new_df)
+            insert_single_entry!(df, new_df[col_name][1], row_ind, col_name)
+        end
+    else # use column order
+        for (j, col_ind) in enumerate(col_inds)
+            insert_single_entry!(df, new_df[j][1], row_ind, col_ind)
+        end
+    end
+    df
 end
 
 # df[MultiRowIndex, MultiColumnIndex] = DataFrame
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -466,23 +466,6 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
     insert_single_entry!(df, v, row_ind, col_ind)
 end
 
-# df[SingleRowIndex, MultiColumnIndex] = Single Item
-function Base.setindex!(df::DataFrame,
-                        v::Any,
-                        row_ind::Real,
-                        col_inds::AbstractVector{Bool})
-    setindex!(df, v, row_ind, findall(col_inds))
-end
-function Base.setindex!(df::DataFrame,
-                        v::Any,
-                        row_ind::Real,
-                        col_inds::AbstractVector{<:ColumnIndex})
-    for col_ind in col_inds
-        insert_single_entry!(df, v, row_ind, col_ind)
-    end
-    return df
-end
-
 # df[SingleRowIndex, MultiColumnIndex] = 1-Row DataFrame
 function Base.setindex!(df::DataFrame,
                         new_df::DataFrame,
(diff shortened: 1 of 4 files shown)


test/dataframe.jl
```diff
@@ -707,6 +707,35 @@ module TestDataFrame
...
     end
 
+    @testset "setindex! sets row with dict/iterable" begin
+        df = DataFrame(a=[0, 0], b=['a', 'a'])
+        dfa = DataFrame(a=0:1, b='a':'b')
+        df[2, :] = Dict(:a => 1, :b => 'b')
+        @test df == dfa
+
+        df = DataFrame(a=[0, 0], b=['a', 'a'])
+        df[2, :] = (a=1, b='b')
```
nalimilan: Put `b` first to check that order doesn't matter.
```suggestion
        df[2, :] = (b='b', a=1)
```

test/dataframe.jl
```diff
@@ -707,6 +707,35 @@ module TestDataFrame
...
+        @test df == dfa
+
+        df = DataFrame(a=[0, 0], b=['a', 'a'], c=[0, 0])
+        dfa = DataFrame(a=0:1, b=['a', 'a'], c=0:1)
+        df[2, [:a, :c]] = [1, 1]
+        @test df == dfa
+
+        df = DataFrame(a=[0, 0], b=['a', 'a'], c=[0, 0])
+        dfc = deepcopy(df)
+        @test_throws ArgumentError df[2, :] = [1, 1]
```
nalimilan: Can you also test a case that works with `:`?

test/dataframe.jl
```diff
@@ -707,6 +707,35 @@ module TestDataFrame
...
+        df[2, :] = Dict(:a => 1, :b => 'b')
+        @test df == dfa
+
+        df = DataFrame(a=[0, 0], b=['a', 'a'])
+        df[2, :] = (a=1, b='b')
+        @test df == dfa
+
+        df = DataFrame(a=[0, 0], b=['a', 'a'])
+        dfc = deepcopy(df)
+        @test_throws ArgumentError df[2, :] = Dict(:a => 1)
```
nalimilan: Also test what happens in case of conversion error (same below).

test/dataframe.jl
coreywoodfield: Unless I misunderstand your request, there already is one, on 726

src/dataframe/dataframe.jl
```diff
@@ -500,6 +500,57 @@ function Base.setindex!(df::DataFrame,
...
+            df[row_ind, :] = old_row
+            msg = "Error adding value to column :$nm."
+            throw(ArgumentError(msg))
+        end
+    end
+    df
+end
+# df[SingleRowIndex, MultiColumnIndex] = iterable
+function Base.setindex!(df::DataFrame,
+                        iterable::AbstractVector,
```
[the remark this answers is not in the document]
coreywoodfield: I changed it to any and it spits out a deprecation warning if a single value is passed in. `df[1,:] .= 1` isn't working and I'm looking into that

src/dataframe/dataframe.jl
```diff
@@ -500,6 +500,57 @@ function Base.setindex!(df::DataFrame,
     return df
 end
 
+# df[SingleRowIndex, :] = Union{AbstractDict, NamedTuple}
+function Base.setindex!(df::DataFrame,
+                        row::Union{AbstractDict, NamedTuple},
```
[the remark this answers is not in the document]
coreywoodfield: `DataFrameRow` isn't defined until after these functions are, how should I deal with that? I could move the definition of `DataFrameRow` up, or add a definition of `setindex!` after `DataFrameRow` gets defined. Both of those don't seem like great options to me, but I'm leaning towards adding a definition of `setindex!` in dataframerow.jl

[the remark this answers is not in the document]
coreywoodfield: Alternatively I could take in an `AbstractDataFrame` and then verify the size, that might be best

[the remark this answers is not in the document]
nalimilan: Yes, sounds like the best approach. BTW, there's a method for `DataFrame` above which doesn't look great (it uses the column order rather than names, and doesn't handle errors).

test/dataframe.jl
nalimilan: I mean one that doesn't throw.

coreywoodfield: (What is currently) 728 through 733 is as follows:
```
        df[2, :] = [1, 'b']
        @test df == dfa

        df = DataFrame(a=[0, 0], b=['a', 'a'])
        df[2, :] = (1, 'b')
        @test df == dfa
```
Is that not what you're looking for?

nalimilan: Sorry, not sure why I missed it.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
coreywoodfield: So should I use column names strictly? I could see a few different use cases for `setindex!` where it takes in a dataframe, which gives us a few different options.

* We could just use column names and throw an error if every column name doesn't match
* We could check to see if the column names match, use them if they do, or just use column order if they don't
* We could only use column order

I'm not sure which one is really the behavior that people would expect (if someone is just thinking of the dataframe as a fancy matrix, then all they're worrying about is the dimensions), but I think the second option is the most forgiving of different expectations, so I'd think we should probably do that. If we do do that, it seems appropriate that we should also change the code for `df[MultiRowIndex, MultiColumnIndex] = DataFrame` around line 580 to have the same behavior. It seems like we should be able to expect that if we were to delegate one to the other that they'd give the same behavior

coreywoodfield:
So I've got setting a row in a dataframe with a 1-row dataframe in a place where I'm happy with it, but there may be a few things left to do, and it may be a few days before I can try and tackle them. I added two `@test_skip` tests in the `setindex!` tests for things that were mentioned in the discussion here but aren't working yet, those are broadcasted assignment and setting a row with a `DataFrameRow` (which we sort of decided not to do, I still think it would make sense to have though). Beyond that I think it'd be good to add checks on calling `setindex!` with a `DataFrame`, and I've done some work on that that I have locally on my machine, but a lot of it was overlapping with the methods that take in `Colon`s so I'd need to dedicate a bit more time to make sure I understand how each of those is delegating before I'd be confident that I had a working method for setting from a dataframe.

Here's what I have left to do:
- [ ] Get broadcasted assignment working (or decide to not deprecate implicit broadcasting of scalar values--I think my check [here](https://github.com/JuliaData/DataFrames.jl/pull/1571/files#diff-0aa1ff52d56badabf7d1511d3dc3f279R501) is pretty robust, and will do what the user expects)
- [ ] Add a method to `setindex!` that takes in a `DataFrameRow`, or decide not to do it, as it'd have to be after the definition of `DataFrameRow`, and thus separated from the rest of the `setindex!` definitions
- [ ] Add error checking to the `setindex!` methods that take in a `DataFrame`, and update them to use column names rather than column order when applicable (I gathered from @nalimilan's comments that that would be preferred)

If I could get guidance on what exactly we want on these, I should be able to get them done within the next week or two

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: What we do with `push!` and `vcat` is to use names and ignore order, so we should just do that here too for consistency.

coreywoodfield:
@nalimilan I implemented the third point, so that column names are now used instead of order. I had to change a few tests so that they didn't break, and it seems like this change could affect a lot of existing workflows, so I was thinking we should probably have a deprecation period, where we allow the old behavior if column names don't match, but also spit out a deprecation warning. What do you think?

src/dataframerow/dataframerow.jl
```diff
@@ -4,6 +4,15 @@ struct DataFrameRow{T <: AbstractDataFrame}
     row::Int
 end
 
+# this is defined here because it has to be after the DataFrameRow definition
```
nalimilan: I don't really like this, it breaks the current organization of the indexing code. Can't you change the inclusion order of files so that `DataFrameRow` is defined before `DataFrame`? If that's not possible, we could move all type definitions to a separate file which would be included first.

src/dataframe/dataframe.jl
```diff
@@ -530,6 +557,56 @@ function Base.setindex!(df::DataFrame,
...
+    new_names = _names(new_df)
+    subset = all(name -> haskey(df, name), new_names)
+    if !subset || !issetequal(index(df)[new_names], index(df)[col_inds])
+        msg = "Column names in inserted dataframe don't match the columns " *
+              "being inserted into"
+        throw(ArgumentError(msg))
+    end
+end
+
+function checkpreconditions(df::DataFrame,
```
nalimilan: Better call `checknames` and `checkdimensions` directly, defining this helper method actually requires more lines of code.

src/dataframe/dataframe.jl
```diff
@@ -530,6 +557,56 @@ function Base.setindex!(df::DataFrame,
...
+    # note that length(::Number) == 1
+    if (new_rows, new_cols) != (length(row_inds), length(col_inds))
+        msg = "Tried to assign $new_rows×$new_cols dataframe to " *
+              "$(length(row_inds))×$(length(col_inds)) destination"
+        throw(DimensionMismatch(msg))
+    end
+end
+
+function checknames(df::DataFrame,
+                  new_df::AbstractDataFrame,
```
nalimilan: Please align arguments (here and elsewhere).

src/dataframe/dataframe.jl
```diff
@@ -530,6 +557,56 @@ function Base.setindex!(df::DataFrame,
     return df
 end
 
+function checkdimensions(new_df::AbstractDataFrame,
+                         row_inds::Any,
+                         col_inds::Any)
+    new_rows, new_cols = size(new_df)
+    # note that length(::Number) == 1
+    if (new_rows, new_cols) != (length(row_inds), length(col_inds))
+        msg = "Tried to assign $new_rows×$new_cols dataframe to " *
```
nalimilan: Should be "data frame" (here and elsewhere).

test/cat.jl
```diff
@@ -108,9 +108,9 @@ module TestCat
         df[:x3] = 2
 
         # assignment of subtables
-        df[1, 1:2] = df[2, 2:3]
-        df[1:2, 1:2] = df[2:3, 2:3]
-        df[[true,false,false,true], 2:3] = df[1:2,1:2]
+        df[1, 1:2] = df[2, 1:2]
```
nalimilan: Do we have checks that an error is thrown when names don't match?

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
 end
 function Base.setindex!(df::DataFrame,
-                        new_df::DataFrame,
+                        v::Any,
                         row_ind::Real,
                         col_inds::AbstractVector{<:ColumnIndex})
-    for j in 1:length(col_inds)
-        insert_single_entry!(df, new_df[j][1], row_ind, col_inds[j])
+    # remove this check after implicit scalar broadcasting is deprecated
+    if applicable(length, v) && Base.IteratorSize(v) != Base.HasShape{0}()
```
nalimilan: Maybe add a special case for `AbstractString` too, since these are `HasLength` but they act as scalar when broadcasting, and are quite common in data frames uses. Also put the `IteratorSize` call first since it can be resolved statically (using `isa HasShape{0}` or `!==` would help), but currently `applicable` cannot.

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
+                      " Possible type mis-match."
+                throw(ArgumentError(msg))
+            end
+        end
+    else
+        msg = "Implicitly broadcasting setindex! with a scalar " *
+              "across columns is deprecated, use .= instead, i.e. " *
+              "df[row, :] .= val"
+        Base.depwarn(msg, :setindex!)
+        for col_ind in col_inds
```
nalimilan: The `try... catch` logic is also needed here, right? Same below.

nalimilan:
Yes, it would be nice to print deprecations to avoid breaking code. Also, gave you looked at how broadcasting could be implemented for `setindex`?

This is likely to have conflicts with https://github.com/JuliaData/DataFrames.jl/pull/1534.

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
+# df[SingleRowIndex, :] = Union{AbstractDict, NamedTuple}
 function Base.setindex!(df::DataFrame,
-                        v::Any,
-                        row_ind::Real,
-                        col_inds::AbstractVector{Bool})
-    setindex!(df, v, row_ind, findall(col_inds))
-end
-function Base.setindex!(df::DataFrame,
-                        v::Any,
+                        row::Union{AbstractDict, NamedTuple},
```
bkamins: `row_ind` must be `Integer`, but not `Bool`.

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
                         row_ind::Real,
-                        col_inds::AbstractVector{<:ColumnIndex})
-    for col_ind in col_inds
-        insert_single_entry!(df, v, row_ind, col_ind)
+                        ::Colon)
+    old_row = df[row_ind, :]
```
bkamins: This will create `DataFrameRow` so we cannot use it to handle cleanup.

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
+    for nm in _names(df)
```
bkamins: I would think that we should be sure that `names(df)` matches exactly names of `NamedTuple`/`AbstractDict` and not is a subset. Why do we want to accept the subset.

Whatever way we choose I would check this condition instead of `try/catch` and then cleanup.

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
 
-# df[SingleRowIndex, MultiColumnIndex] = 1-Row DataFrame
+# df[SingleRowIndex, MultiColumnIndex] = iterable (or scalar, being deprecated)
 function Base.setindex!(df::DataFrame,
-                        new_df::DataFrame,
+                        v::Any,
                         row_ind::Real,
                         col_inds::AbstractVector{Bool})
-    setindex!(df, new_df, row_ind, findall(col_inds))
+    setindex!(df, v, row_ind, findall(col_inds))
```
bkamins: we should check here if `col_inds` has a correct length.

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
 end
 function Base.setindex!(df::DataFrame,
-                        new_df::DataFrame,
+                        v::Any,
                         row_ind::Real,
                         col_inds::AbstractVector{<:ColumnIndex})
```
bkamins: We should handle the case when `col_inds` have duplicates also `AbstractVector{<:ColumnIndex}` allows mixing numbers and symbols which should not be allowed.

src/dataframe/dataframe.jl
```diff
@@ -466,38 +466,65 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Real, col_ind::ColumnInd
...
-    for j in 1:length(col_inds)
-        insert_single_entry!(df, new_df[j][1], row_ind, col_inds[j])
+    # remove this check after implicit scalar broadcasting is deprecated
+    if applicable(length, v) && Base.IteratorSize(v) != Base.HasShape{0}()
+        if length(v) != length(col_inds)
+            msg = "Length of iterable does not match DataFrame column count."
+            throw(ArgumentError(msg))
+        end
+        old_row = df[row_ind, col_inds]
```
bkamins: This will create a `DataFrameRow` so we cannot use it as a fallback

src/dataframe/dataframe.jl
```diff
@@ -530,6 +557,56 @@ function Base.setindex!(df::DataFrame,
     return df
 end
 
+function checkdimensions(new_df::AbstractDataFrame,
+                         row_inds::Any,
+                         col_inds::Any)
+    new_rows, new_cols = size(new_df)
+    # note that length(::Number) == 1
+    if (new_rows, new_cols) != (length(row_inds), length(col_inds))
```
bkamins: do we handle the case that `row_inds` or `col_inds` have duplicates?
Also do we avoid calling this method if either of them is `AbstractVector{Bool}`?

src/dataframe/dataframe.jl
```diff
@@ -530,6 +557,56 @@ function Base.setindex!(df::DataFrame,
...
+    checkdimensions(new_df, row_inds, col_inds)
+    checknames(df, new_df, col_inds)
+end
+
+# df[SingleRowIndex, MultiColumnIndex] = 1-Row DataFrame
+function Base.setindex!(df::DataFrame,
+                        new_df::AbstractDataFrame,
+                        row_ind::Real,
+                        col_inds::AbstractVector{Bool})
+    setindex!(df, new_df, row_ind, findall(col_inds))
```
bkamins: we should check the length of `col_inds` here.

src/dataframe/dataframe.jl
```diff
@@ -530,6 +557,56 @@ function Base.setindex!(df::DataFrame,
...
+end
+function Base.setindex!(df::DataFrame,
+                        new_df::AbstractDataFrame,
+                        row_ind::Real,
+                        col_inds::AbstractVector{<:ColumnIndex})
```
bkamins: this allows mixing integers and symbols.

src/dataframerow/dataframerow.jl
```diff
@@ -4,6 +4,15 @@ struct DataFrameRow{T <: AbstractDataFrame}
...
+# df[SingleRowIndex, MultiColumnIndex] = DataFrameRow
+function Base.setindex!(df::DataFrame,
+          val::DataFrameRow,
+          row_ind::Real,
```
bkamins: it can be only `Integer` other than `Bool`.

bkamins:
I would also recommend, after rebasing it against #1534 (when it is merged) to add a specification of `setindex!` to docs/lib/indexing.md so that we can check if what is implemented is the same as what is indented (and ideally this specification should cover both `=` and `.=` assignment options like `getindex` also covers `view` now).

src/dataframe/dataframe.jl
nalimilan: Indeed. So I guess the best solution is to call `copy`?

src/dataframe/dataframe.jl
nalimilan: Agreed. Though `try... catch` is still needed to handle conversion issues AFAICT.

src/dataframe/dataframe.jl
bkamins: Right point.

bkamins:
@coreywoodfield, I am aware that the number of things I have outline is large. Therefore, if you do not have time to work on all of that I think that the minimal I would recommend is:
* rebase against #1534 as it is now merged;
* concentrate in code only on the features you want to change (not fixing all other things) - but in those changes be consistent with #1534;
* add a specification of RHS of `setindex!` in the cases you cover to the documentation page where `getindex` is now documented (especially outlining how data is interpreted in RHS in cases when in can contain column names and what is the contract assumed).
* LHS of `setindex!` should strictly conform the LHS specification of `getindex`.

The rest can be done later (I can work on it since #1534 is finished, but of course you are also welcome - let us coordinate the efforts 😄). Thank you!

coreywoodfield:
Sorry I'm taking so long on this, I'm still intending to finish it, but I've been swamped with school and other things. I wanted to tackle all of the `setindex!` stuff, but at this point I'd like to just get the `Dict` code working, and then I might try and contribute more depending on how much time I have

bkamins:
Thank you for your contribution. Very soon (after #1603 is merged) similar work has to be done for `setindex!` (I have outlined the key steps above).

Given your last comment I would recommend the following approach:
1. you finish `Dict` code (and concentrate only on this case)
2. I plan to write down rules for `setindex!` in the next few weeks in `docs/src/lib/indexing.md` for `DataFrame`, `SubDataFrame` and `DataFrameRow` - when the plan is proposed we should discuss it (if you are willing to do it please comment)
3. Then a PR or set of PRs with deprecations only will have to be implemented
4. Next we have to release DataFrames.jl
5. Finally we need a PR with the target functionality

bkamins:
@coreywoodfield After 0.19 I will implement the rules set in https://juliadata.github.io/DataFrames.jl/latest/lib/indexing/ soon. So I think this PR can be closed (but please comment).

@coreywoodfield + @nalimilan The only question is if we should allow `AbstractDict` as RHS (I planned to disallow it, but we allow it in `push!` so maybe we also can allow it for `setindex!`). A comment would be appreciated.

nalimilan:
Why not allow dicts, but that sounds like low priority.

bkamins:
I will write the code soon anyway - I can add dicts then as this is simple

coreywoodfield:
Sounds good, thanks!

I think that generally it makes sense for the allowed signatures for `push!` to match those of `setindex!`. If we allow things to be done with `push!` that we don't allow with `setindex!`, then it seems to me like we're promoting bad practices (namely, creating an empty dataframe and filling it row by row rather than allocating a dataframe with the appropriate size and then filling it). If someone already has `Dict`s, and they can `push!` `Dict`s, it only makes sense to me that they should also be able to `setindex!` with `Dict`s. So in my view, we should either allow it in both places or disallow it in both places (but preferably allow it)

bkamins:
Agreed. This is what we will do.

PR #1575, opened 2018-10-26
convert for DataFrameRow

bkamins:
PR following discussion in #1564.

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -57,7 +57,18 @@ function Base.iterate(r::DataFrameRow, st)
     return ((_names(r)[st], r[st]), st + 1)
 end
 
-Base.convert(::Type{Array}, r::DataFrameRow) = convert(Array, parent(r)[row(r),:])
+function Base.convert(::Type{Array}, r::DataFrameRow)
+    Base.depwarn("Conversion of `DataFrameRow` to a `Matrix` is deprecated. " *
+                 "`Vector` will be returned in the future.", :convert)
+    convert(Array, parent(r)[row(r),:])
+end
+Base.convert(::Type{Vector}, dfr::DataFrameRow) =
+    [dfr[i] for i in 1:length(dfr)]
+Base.convert(::Type{Vector{T}}, dfr::DataFrameRow) where T =
+    T[dfr[i] for i in 1:length(dfr)]
+Base.Vector(dfr::DataFrameRow) = Base.convert(Vector, df)
+Base.Vector{T}(dfr::DataFrameRow) where T = Base.convert(Vector{T}, df)
+
 
 Base.keys(r::DataFrameRow) = names(parent(r))
 Base.values(r::DataFrameRow) = ntuple(col -> parent(r)[col][row(r)], length(r))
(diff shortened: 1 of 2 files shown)


src/dataframerow/dataframerow.jl
```diff
@@ -57,7 +57,18 @@ function Base.iterate(r::DataFrameRow, st)
...
+function Base.convert(::Type{Array}, r::DataFrameRow)
+    Base.depwarn("Conversion of `DataFrameRow` to a `Matrix` is deprecated. " *
+                 "`Vector` will be returned in the future.", :convert)
+    convert(Array, parent(r)[row(r),:])
+end
+Base.convert(::Type{Vector}, dfr::DataFrameRow) =
+    [dfr[i] for i in 1:length(dfr)]
+Base.convert(::Type{Vector{T}}, dfr::DataFrameRow) where T =
+    T[dfr[i] for i in 1:length(dfr)]
+Base.Vector(dfr::DataFrameRow) = Base.convert(Vector, df)
```
nalimilan: No need for `Base.`.

test/dataframerow.jl
```diff
@@ -100,5 +100,12 @@ module TestDataFrameRow
     io = IOBuffer()
     show(io, DataFrameRow(df, 1))
     @test String(take!(io)) == "DataFrameRow (row 1)\na  \nb  1"
+
+    df = DataFrame(a=[1, missing], b=[2.0, 3.0])
+    dfr = DataFrameRow(df, 1)
+    @test Vector(dfr) == Real[1, 2.0]
+    @test eltype(Vector(dfr)) == Real
+    @test Vector{Int}(dfr) == [1, 2]
+    @test eltype(Vector{Int}(dfr)) == Int
```
nalimilan: Also test `convert` directly.

nalimilan:
Actually, I'm a bit hesitant about whether `convert` should return a `Vector` or a `Matrix`. `Vector` sounds more logical if we consider `DataFrameRow` as equivalent to a (named) tuple, but tuples don't support conversion to vectors. `Matrix` is more consistent if we consider that `vcat(convert(Array, DataFrameRow(df, 1)), convert(Array, DataFrameRow(df, 2)))` should be equivalent to `convert(Array, df[1:2,:])`.

src/dataframerow/dataframerow.jl
bkamins: Interesting - I did not think earlier abut such case. Where is this documented?

bkamins:
OK - we can leave it as is.

test/dataframerow.jl
bkamins: added

src/dataframerow/dataframerow.jl
nalimilan: Well, you never need to qualify calls to exported functions from other modules. You don't write `Base.length` either, do you? :-)

nalimilan: Also, the CI failure is due to `df` instead of `dfr`.

bkamins: Sure - I thought you wanted to strip `Base.` from `Base.Vector` and got confused.

Issue #1276, opened 2017-11-14
column names of aggregated DataFrame with anonymous functions


pfarndt:
When using an anonymous functions to generate an aggregated DataFrame the column names are not reproducible and inconvenient for further usage:

```
using DataFrames

d = DataFrame(g = [1,1,2,2], v=1:4)

println(names(aggregate(d, :g, x->sum(x))))
println(names(aggregate(d, :g, x->sum(x))))
```

This code produces:
```
Symbol[:g, Symbol("v_#1")]
Symbol[:g, Symbol("v_#3")]
```
although the last two commands are identical.

The documentation is promising names like `v_\lambda1` at this point. I am using Julia v0.6.0.

IMO the function `_fnames` in [https://github.com/JuliaData/DataFrames.jl/blob/master/src/other/utils.jl](url) should be adjusted for new versions of julia.

cjprybol:
> not reproducible

I was able to reproduce this
```julia
julia> using DataFrames

julia> d = DataFrame(g = [1,1,2,2], v=1:4)
4×2 DataFrames.DataFrame
│ Row │ g │ v │
├─────┼───┼───┤
│ 1   │ 1 │ 1 │
│ 2   │ 1 │ 2 │
│ 3   │ 2 │ 3 │
│ 4   │ 2 │ 4 │

julia> println(names(aggregate(d, :g, x->sum(x))))
Symbol[:g, Symbol("v_#1")]

julia> println(names(aggregate(d, :g, x->sum(x))))
Symbol[:g, Symbol("v_#3")]
```

The names of the columns are the function identifiers. Here's another fresh session to show the identifiers of the anonymous functions, which you'll see match the column names, and are again reproducible. 

```julia
julia> x->sum(x)
(::#1) (generic function with 1 method)

julia> x->sum(x)
(::#3) (generic function with 1 method)
```
> inconvenient for further usage

The previously used lambda syntax has no relation (aside from order) to the actual functions that were used to create the data and hence was removed.  Unfortunately, it is not yet possible to extract the original code of an anonymous function by its identifier (see https://github.com/JuliaLang/julia/issues/2625#issuecomment-306270254), although in principle if that is added as a language feature then using the identifiers of the anonymous functions will provide both a stable identifier as well as a way to recover the function associated with that identifier. Currently, only the "stable identifier" part is supported while the lambda syntax cannot support either.

[...]

PR #1576, opened 2018-10-27
improve naming of anonymous functions in aggregate

bkamins:
October is about to end, so I proposed to solve #1276.

I use standard `DataFrame` constructor machinery to generate names. The default suffix for lambda function is `λ`.

Also added `makeunique` keyword argument as it was missing (as here I guess we should set it to `true` it by default).

I do not know a better approach to test if a function is anonymous than by checking if first character of its name is `#` (maybe there is a cleaner solution to this?).

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -376,13 +376,13 @@ function aggregate(d::AbstractDataFrame,
     aggregate(groupby(d, cols, sort=sort), fs)
 end
 
-function dfnameof(f)
+function funname(f)
     n = nameof(f)
-    String(n)[1] == '#' ? :λ : n
+    String(n)[1] == '#' ? :function : n
 end
 
 _makeheaders(fs::Vector{<:Function}, cn::Vector{Symbol}) =
-    [Symbol(colname, '_', dfnameof(f)) for f in fs for colname in cn]
+    [Symbol(colname, '_', funname(f)) for f in fs for colname in cn]
 
 function _aggregate(d::AbstractDataFrame, fs::Vector{T}, headers::Vector{Symbol}, sort::Bool=false) where T<:Function
     res = DataFrame(AbstractVector[vcat(f(d[i])) for f in fs for i in 1:size(d, 2)], headers, makeunique=true)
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks. I guess detecting anonymous functions that way is fine.

I'm not sure using `λ` for this is a good idea given that it cannot be typed directly at the keyboard. We always provide ASCII variants when using Unicode identifiers, and here there is no ASCII way to access that column. Why not use `f` instead (or any other ASCII name)?

src/groupeddataframe/grouping.jl
```diff
@@ -376,11 +376,16 @@ function aggregate(d::AbstractDataFrame,
     aggregate(groupby(d, cols, sort=sort), fs)
 end
 
+function dfnameof(f)
```
nalimilan: Why `df`? Sounds like "data frame".

bkamins:
I use `λ` because:
* this reduces the probability of function name clash;
* it is short (e.g. when printed);
* it produces a valid identifier that can be typed in in any Julia editor and Julia REPL.

If we want ASCII, then maybe `lambda` would be good (I do not like `f` because of the name clash possibility)?

src/groupeddataframe/grouping.jl
bkamins: This is the reason because I want to get a name of `f` for internal DataFrames.jl purposes. Do you have any suggestion what would be a good name.

I also can change it to an anonymous function call in line 385 if this is preferred.

nalimilan:
I'm not a fan of `λ`/`lambda` as it's not the most common/explicit term for anonymous functions, especially for people not coming from CS. AFAICT we almost always say "anonymous function" in Julia.

A few ideas: `fun`, `function`, `agg`

src/groupeddataframe/grouping.jl
nalimilan: Maybe `funname` or something like that? Just bikeshedding.

bkamins: fixed

bkamins:
`function` is OK, as no function will have this name, `fun` and `agg` re pretty common when you check students' homework 😄.

quinnj:
FWIW, I like lambda.

bkamins:
I merge `function` version, as it guarantees not to conflict with an actual function name.

PR #1577, opened 2018-10-27
Enforce makeunique

bkamins:
Finish deprecation period for `makeunique` keyword argument. It has been for ~1 year now in this state.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -728,8 +728,8 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
                 k += 1
             end
         else
-            msg = """Duplicate variable name $(name).
-                 Pass makeunique=true to make it unique using a suffix automatically."""
+            msg = "Duplicate variable name $(name). Pass makeunique=true" *
+                  " to make it unique using a suffix automatically."
             throw(ArgumentError(msg))
         end
     end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -728,7 +728,7 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
                 k += 1
             end
         else
-            msg = "Duplicate variable name $(name). Pass makeunique=true" *
+            msg = "Duplicate variable name $name. Pass makeunique=true" *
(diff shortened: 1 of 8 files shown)


bkamins:
@nalimilan The only question with removing the deprecation is with Tables.jl integration as earlier the tests allowed for duplicate names in the source, but I understood for the earlier discussion that we should not allow them (so now it always throws an error).

src/dataframe/dataframe.jl
```diff
@@ -728,12 +728,9 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
...
             end
         else
-            # TODO: remove depwarn and call and uncomment ArgumentError below
-            Base.depwarn("Inserting duplicate column name is deprecated, use makeunique=true.", :insertcols!)
-            insertcols!(df, col_ind, name => item; makeunique=true) # temporary fix to avoid duplicates
-            # msg = """Duplicate variable name $(name).
-            #      Pass makeunique=true to make it unique using a suffix automatically."""
-            # throw(ArgumentError(msg))
+            msg = """Duplicate variable name $(name).
+                 Pass makeunique=true to make it unique using a suffix automatically."""
```
nalimilan: We don't generally use line breaks in messages, do we? Same below.

quinnj:
Yeah, unique names are required, so I think we're good here.

src/other/index.jl
```diff
@@ -195,11 +192,9 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
     end
     if length(dups) > 0
         if !makeunique
-            Base.depwarn("Duplicate variable names are deprecated: pass makeunique=true to add a suffix automatically.", :add_names)
-            # TODO: uncomment the lines below after deprecation period
-            # msg = """Duplicate variable names: $(u[dups]).
-            #          Pass makeunique=true to make them unique using a suffix automatically."""
-            # throw(ArgumentError(msg))
+            msg = "Duplicate variable names: $(u[dups]). Pass makeunique=true" *
```
nalimilan: AFAICT this will print as `Symbol[:col]`. I think you can get something really nice using `$(join(string.(':', u[dups]), ", ", " and "))`. Same in other places (including in untouched code above).

bkamins: OK - fixed. I have also made it to avoid printing the same name of an offending column more than once.

PR #1579, opened 2018-10-29
Add support for another TableTraits.jl interface

davidanthoff:
I've had something like this lingering around in branches for a very long time, but now finally finished it upstream.

With this PR, the current performance overhead that [CSVFiles.j](https://github.com/queryverse/CSVFiles.jl) has over a raw [TextParse.jl](https://github.com/JuliaComputing/TextParse.jl) disappears entirely. So in the benchmarks shown [here](https://discourse.julialang.org/t/textparse-jl-is-fast-again/16664/10), [CSVFiles.j](https://github.com/queryverse/CSVFiles.jl) gets the same (excellent) performance that one can see in raw [TextParse.jl](https://github.com/JuliaComputing/TextParse.jl) if used with ``DataFrame``.

We have a whole bunch of other plans for the [Queryverse.jl](https://github.com/queryverse/Queryverse.jl) that will lead to more and more sources that will utilize this interface as well, so there will be more benefits down the road. That includes the other file IO packages, but also [Query.jl](https://github.com/queryverse/Query.jl) itself (although that latter part is not going to happen overnight).

Things still to do:
- [x] Add a test.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -10,5 +10,5 @@ CodecZlib 0.4
 TranscodingStreams
 Compat 0.59.0
 Tables
-IteratorInterfaceExtensions
-TableTraits
+IteratorInterfaceExtensions 0.1.1
+TableTraits 0.4.0
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -16,6 +16,10 @@ function DataFrame(x)
     if Tables.istable(typeof(x))
         return fromcolumns(Tables.columns(x))
     end
+    if TableTraits.supports_get_columns_copy_using_missing(x)
+        y = TableTraits.get_columns_copy_using_missing(x)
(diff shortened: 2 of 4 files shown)


davidanthoff:
> But with this approach, you'd end up with a DataFrame with DataValueArray columns, right?

No, you get ``Vector{Union{T,Missing}}``s. The philosophy of [TableTraits.jl](https://github.com/queryverse/TableTraits.jl) in general is to always respect the default missing value story of each source and sink.

quinnj:
That's good to know. My only concern then is how long we plan on supporting TableTraits.jl in general in DataFrames.jl. It was added here to ease the transition from IterableTables.jl to here, but the original thoughts were to support the Tables.jl interface long-term.

davidanthoff:
Thanks!

Issue #1573, opened 2018-10-18
Support for construction from a named tuple?


jlperla:
I think this is related to #1569 but gives a very specific suggestion.

Existing DataFrames seem to play very well with named tuples.  FOr example,
```julia
julia> df = DataFrame(a = Float64[], b = Int64[])
0×2 DataFrame

julia> push!(df, (a = 4.0, b = 2))
1×2 DataFrame
│ Row │ a       │ b     │
│     │ Float64 │ Int64 │
├─────┼─────────┼───────┤
│ 1   │ 4.0     │ 2     │
```
What is irritating is that you cannot create the DataFrame directly from the named tuple, and instead need to specify types yourself.

There is a workaround with splatting, which can rely entirely on inference!
```julia
julia> t = (a = 4.0, b = 2)
(a = 4.0, b = 2)

julia> df = DataFrame(;t...)
1×2 DataFrame
│ Row │ a       │ b     │
│     │ Float64 │ Int64 │
├─────┼─────────┼───────┤
│ 1   │ 4.0     │ 2     │
```
So my question is: (1) is there any real flaw in the splatting interface
and (2) would you be interested in a PR that makes this more explicit.  That is,  it would add in a constructor for named tuples and support 
```
julia> t = (a = 4.0, b = 2)
(a = 4.0, b = 2)

julia> df = DataFrame(t) # not valid, but could be
1×2 DataFrame
│ Row │ a       │ b     │
│     │ Float64 │ Int64 │
├─────┼─────────┼───────┤
│ 1   │ 4.0     │ 2     │
```

nalimilan:
Splatting should be OK. You can also do `DataFrame([(a = 4.0, b = 2)])`.

Whether or not constructors should allow this also depends on whether we want to allow this for Tables.jl. Cc: @quinnj

quinnj:
I don't think Tables.jl would do something like this; _technically_, `(a = 4.0, b = 2)` is a NamedTuple of "iterables", but only because `4.0` and `2` themselves are iterable (though this has been debated a lot and most think scalars _shouldn't_ iterate). So for the time being at least, valid "tables" from Tables.jl perspective include a vector of namedtuples or a namedtuple of abstractvectors, but not plain namedtuples.

jlperla:
OK, I didn't know that would work.  Solatting is pretty tough for new users, but:

SO you are saying that
```julia
julia>  t = (a = 4.0, b = 2)
[...]

PR #1580, opened 2018-10-30
Construction from a named tuple documentation

ajozefiak:
Addition of an example to the documentation of constructing a data frame from an array of named tuples. As suggested in #1573

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -124,7 +124,7 @@ It is also possible to fill a `DataFrame` row by row. Let us construct an empty 
 julia> df = DataFrame(A = Int[], B = String[])
 0×2 DataFrame
 ```
-Additionally, a `DataFrame` object can be constructed from a named tuple, or an arbitrary number of named tuples. The order and types of the elements in the named tuples speciefy the order of the columns and their type:
+Additionally, a `DataFrame` object can be constructed from an array of named tuples, where each tuple of the array corresponds to a row of the new data frame. The order and types of the elements in each of the named tuples specify the order of the columns and their types:
 
 ```jldoctest dataframe
 julia> r1 = (A = 1, B = "M")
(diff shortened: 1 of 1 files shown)


docs/src/man/getting_started.md
```diff
@@ -124,37 +124,63 @@ It is also possible to fill a `DataFrame` row by row. Let us construct an empty
 julia> df = DataFrame(A = Int[], B = String[])
 0×2 DataFrame
 ```
+Additionally, a `DataFrame` object can be constructed from an array of named tuples, where each tuple of the array corresponds to a row of the new data frame. The order and types of the elements in each of the named tuples specify the order of the columns and their types:
```
bkamins: I guess it should be

> vector of named tuples, where each named tuple in the vector"

(not native speaker though).

bkamins:
```suggestion
Additionally, a `DataFrame` object can be constructed from any object that supports Tables.jl interface, for example a vector of `NamedTuple`s:
```

docs/src/man/getting_started.md
```diff
@@ -124,37 +124,63 @@ It is also possible to fill a `DataFrame` row by row. Let us construct an empty
...
+
+```jldoctest dataframe
+julia> r1 = (A = 1, B = "M")
+(A = 1, B = "M")
+
+julia> r2 = (A = 2, B = "N")
+(A = 2, B = "N")
+
+julia> df = DataFrame([r1, r2])
```
bkamins:
```suggestion
julia> df = DataFrame([(A = 1, B = "M"),
                       (A = 2, B = "N")])
```

bkamins: I think we do not need to show creation of `r1` and `r2` explicitly. Can you also please remove the respective lines above?

bkamins:
@ajozefiak I have left some comments. I think this PR is good to merge when they are incorporated. Thank you!

bkamins:
I have opened #2077 instead of this PR.

PR #1582, opened 2018-10-31
rename example

### LINKED ISSUE #1578: same order for parameter of rename (and rename!) function
Hello,

I wonder why `rename` and `rename!` function

http://juliadata.github.io/DataFrames.jl/latest/lib/functions.html#DataFrames.rename

don't have alway `df` as first parameter.

That's odd. But maybe I missed something

Kind regards

femtotrader:
rename with function and do block
Closes #1578

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -189,6 +189,9 @@ rename(df, :i => :A, :x => :X)
 rename(df, [:i => :A, :x => :X])
 rename(df, Dict(:i => :A, :x => :X))
 rename(x -> Symbol(uppercase(string(x))), df)
+rename(df) do x
+    x |> string |> lowercase |> Symbol
+end
 rename!(df, Dict(:i =>: A, :x => :X))
 ```
 


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -189,6 +189,9 @@ rename(df, :i => :A, :x => :X)
 rename(df, [:i => :A, :x => :X])
 rename(df, Dict(:i => :A, :x => :X))
 rename(x -> Symbol(uppercase(string(x))), df)
+rename(df) do x
+    x |> string |> lowercase |> Symbol
```
bkamins: I guess you wanted `uppercase` not `lowercase` here. Right?

nalimilan: Also I'd rather not use `|>` here to keeps things simple for newcomers.

nalimilan:
Thanks. Can you add the same example to `rename!`?

src/abstractdataframe/abstractdataframe.jl
femtotrader: I think if noone shows how to use `|>`, newcomers won't understand how to use it.
This example being just after

    rename(x -> Symbol(uppercase(string(x))), df)

I think it helps to understand both `|>` and `do` block

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -189,7 +189,13 @@ rename(df, :i => :A, :x => :X)
 rename(df, [:i => :A, :x => :X])
 rename(df, Dict(:i => :A, :x => :X))
 rename(x -> Symbol(uppercase(string(x))), df)
+rename(df) do x
+    x |> string |> uppercase |> Symbol
+end
 rename!(df, Dict(:i =>: A, :x => :X))
+rename!(df) do x
+    x |> string |> uppercase |> Symbol
```
bkamins: here it maybe better to use `lowercase` 😄 as otherwise this is a no-op as we used `rename!` in the line above?

PR #1583, opened 2018-11-01
Cleanup of documentation code

bkamins:
Fixes:

* generation of doc page of indexing that was missing;
* hopefully makes http://juliadata.github.io/DataFrames.jl/latest/lib/types.html#Types-1 correctly contain documentation of types (not only their names)

### DIFF
--- a/docs/src/lib/types.md
+++ b/docs/src/lib/types.md
@@ -11,23 +11,26 @@ Pages = ["types.md"]
 
 ## Type hierarchy design
 
-`AbstractDataFrame` is an abstract type that provides an interface for working with tabular data.
+`AbstractDataFrame` is an abstract type that provides an interface for data frame types.
+It is not intended as a fully generic interface for working with tabular data, which is the role of
+interfaces defined by [Tables.jl](https://github.com/JuliaData/Tables.jl/) instead.
 
-A key subtype of the `AbstractDataFrame` type is a `DataFrame` type that uses columnar data storage.
+`DataFrame` is the most fundamental subtype of `AbstractDataFrame`, which stores a set of columns
+as `AbstractVector` objects.
 
-The package provides a `SubDataFrame` type that is also `AbstractDataFrame` subtype.
-The `SubDataFrame` type is a view into `DataFrame`. It stores only a reference to parent `DataFrame`
-and information about which rows from the parent are selected. Typically it is created by using the `view`
-function or is returned by indexing into a `GroupedDataFrame` object.
+`SubDataFrame` is an `AbstractDataFrame` subtype representing a view into a `DataFrame`.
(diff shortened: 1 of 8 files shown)


bkamins:
and adds `insertcols!` to documentation generation.

src/dataframerow/dataframerow.jl
```diff
@@ -1,4 +1,8 @@
-# Container for a DataFrame row
+"""
+    DataFrameRow
```
nalimilan: Should mention the type parameter. Same below.

bkamins: Fixed. I have also added supertype informatin where applicable (here I think it is especially relevant as many users think that `DataFrameRow` a subtype of `AbstractDataFrame`).

bkamins:
I have also added a description of types in DataFrames.jl (and removed `GroupApplied` from the list as I understand it will soon be gone).

bkamins:
@pszufe - can you have a look at the descriptions in this PR (if they are clear to you and without English errors, especially explanation of types) before we merge it?

docs/src/lib/types.md
```diff
@@ -9,11 +9,33 @@ CurrentModule = DataFrames
 Pages = ["types.md"]
 ```
 
+## Type hierarchy design
+
+`AbstractDataFrame` is an abstract type that provides database-like interface.
```
pszufe: ...interface for representing data in relational format.

Database-like seems too broad.

docs/src/lib/types.md
```diff
@@ -9,11 +9,33 @@ CurrentModule = DataFrames
...
+
+A key subtype of the `AbstractDataFrame` type is a `DataFrame` type that uses columnar data storage.
+
+The package provides a `SubDataFrame` type that is also `AbstractDataFrame` subtype.
+Is a view into `DataFrame`. It stores only a reference to parent `DataFrame` and information about
```
pszufe: The `SubDataFrame` type is a view...

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1,5 +1,6 @@
-
 """
+    AbstractDataFrame
+
```
pszufe: ...interface for representing data in relational format (or column-oriented data).

Database-like seems too broad.

bkamins:
Fixed comments.

nalimilan:
I've proposed a few changes, let me know what you think. In particular, I wanted to avoid possible confusion about `AbstractDataFrame`, which is an abstract mostly useful for internal use and not really documented: people better use Tables.jl when possible.

bkamins:
Thank you. I have merged this to see how the documentation renders.

PR #1585, opened 2018-11-01
Fix depwarns of getindex

bkamins:
I am in the process of cleaning up after `getindex` revision. I will continue as this is not finished (but it gives 
 the idea of changes).

A thing to discuss from what I have already changed is for `hcat!` - if we want a target functionality to put `views` or copies if RHS of `hcat!` is a `SubDataFrame`.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1040,7 +1040,7 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
         # the code below assumes that only DataFrame and SubDataFrame
         # are subtypes of AbstractDataFrame
         # it should be removed ASAP after deprecation
-        data = [df isa DataFrame ? df[name] : view(parent(df)[colind], rows(df)) for df in dfs]
+        data = [df isa DataFrame ? df[name] : view(parent(df)[name], rows(df)) for df in dfs]
         lens = map(length, data)
         T = mapreduce(eltype, promote_type, data)
         cols[i] = Tables.allocatecolumn(T, sum(lens))
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1037,7 +1037,10 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
     for (i, name) in enumerate(header)
         # TODO: replace with commented out code after getindex deprecation
         # data = [df[name] for df in dfs]
-        data = [df isa DataFrame ? df[name] : df[:, name] for df in dfs]
+        # the code below assumes that only DataFrame and SubDataFrame
+        # are subtypes of AbstractDataFrame
+        # it should be removed ASAP after deprecation
+        data = [df isa DataFrame ? df[name] : view(parent(df)[colind], rows(df)) for df in dfs]
         lens = map(length, data)
         T = mapreduce(eltype, promote_type, data)
         cols[i] = Tables.allocatecolumn(T, sum(lens))
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -85,7 +85,14 @@ struct Cols{T <: AbstractDataFrame} <: AbstractVector{AbstractVector}
 end
 function Base.iterate(itr::Cols, st=1)
     st > length(itr.df) && return nothing
-    return (itr.df[st], st + 1)
+    # TODO: remove the condition and replace with code commented out
+    # after getindex deprecation period
+    # (itr.df[st], st + 1)
+    if itr.df isa SubDataFrame
+        (itr.df[:, st], st + 1)
+    else
+        (itr.df[st], st + 1)
+    end
 end
 Base.length(itr::Cols) = length(itr.df)
 Base.size(itr::Cols, ix) = ix==1 ? length(itr) : throw(ArgumentError("Incorrect dimension"))
@@ -296,6 +303,8 @@ end
 
 Base.haskey(df::AbstractDataFrame, key::Any) = haskey(index(df), key)
 Base.get(df::AbstractDataFrame, key::Any, default::Any) = haskey(df, key) ? df[key] : default
+# TODO: remove this method after getindex deprecation period
+Base.get(df::SubDataFrame, key::Any, default::Any) = haskey(df, key) ? df[:, key] : default
 Base.isempty(df::AbstractDataFrame) = size(df, 1) == 0 || size(df, 2) == 0
 
 ##############################################################################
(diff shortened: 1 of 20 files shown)


bkamins:
@nalimilan Before I move forward I would like to discuss one design issue. In the *transition period* I need to write specialized code for `SubDataFrame` before it is defined in the package (e.g. in src/abstractdataframe/abstractdataframe.jl`). What do you think would be the best approach to handle this?

nalimilan:
I'd rather define duplicated methods in deprecated.jl, so that the main code remains clean and they are easy to remove. That should also ensure `SubDataFrame` has already been defined.

bkamins:
Good point - thanks.
So - for now I am left with the question about `hcat!`. The status is that currently `hcat!` for `DataFrame` passes a column without copying, but it does copy `SubDataFrame` columns.

The question is if we want to keep it
or
we should pass a view of a vector constituting the column of a `SubDataFrame` (so we would be consistent, in that `hcat!` produces an object that allows for modification of source data frames, but the downside is that we would have columns that are views not normal vector).

@nalimilan - do you have any opinion on this?

nalimilan:
Tricky question! If we make copies automatically then there will be no (simple) way of concatenating a `SubDataFrame` while preserving views; conversely, if we don't make copies automatically, people can easily and efficiently call `copy` manually. As you said, it would also be consistent in that modifying the result would modify the inputs. So these are two arguments in favor of preserving views.

OTOH I'm not sure it's a very common need, and it makes things a bit more complex (creating `SubArray` columns). In terms of performance, preserving views is faster for one-off operations, but slower for repeated operations. In particular, if people use `hcat!` with `by`, we'd better preserve views or performance will be terrible.

Overall I'd lean towards preserving views, but it's a tough call.

bkamins:
I am slowly moving forward. I make `hcat!` as you suggested.
Now the other point for you to look at is `groupby` where we make:
```
    sdf = df[cols]
```
selection. The tricky part is if we want to have `sdf` to be a `SubDataFrame` if `df` is a `SubDataFrame` (this is what would happen if I did not add a check as currently in this PR). This is related to `groupby` performance on `SubDataFrame` (maybe it is better that `sdf` is a `DataFrame` always - this is what is currently implemented in this PR).

test/dataframe.jl
```diff
@@ -133,59 +133,93 @@ module TestDataFrame
...
-        @test all(ismissing, df[:, 2])
-        @test all(ismissing, df[:, 3])
+        @test typeof(df[1]) == Vector{Union{Int, Missing}}
+        @test typeof(df[2]) == Vector{Union{Float64, Missing}}
+        @test typeof(df[3]) <: CategoricalVector{Union{String, Missing}}
+        @test all(ismissing, df[1])
+        @test all(ismissing, df[2])
+        @test all(ismissing, df[3])
+        # TODO: enable those tests after getindex deprecation period
+        # @test typeof(df[:, 1]) == Vector{Union{Int, Missing}}
```
nalimilan: Can you use `@test_broken` for these?

nalimilan:
I'm not sure it really matters whether we make a copy or not in `groupby`. Since we go over the grouping columns twice, copying could actually be a bit faster. Anyway I expect the difference to be negligible compared with the cost of grouping and combining (note that this is orthogonal to the fact that we pass a `SubDataFrame` to the user-provided function). So I'd go with the simplest solution.

test/dataframe.jl
bkamins: This is not broken - the test passes, but prints warnings (massively) and I understood that you wanted to avoid printing unnecessary warnings.

nalimilan: I think that's OK if the tests print warnings, as long as they are correct in the long term. Better have good tests that are temporarily verbose than having to enable them manually later.

bkamins: So I understand we go the following route:
* it is not a problem if tests print warnings;
* however, we want to avoid printing warnings by library functions in normal usage.

Yes?

nalimilan: Right.

nalimilan:
The amount of code that needs duplicating is terrifying. I wonder whether it wouldn't be simpler even for users to change the behavior without a deprecation. Indeed in most situations the deprecation will force people to write `sdf[:, col]` even if they don't actually need a copy. It's probably relatively rare to rely on the copying behavior. Though of course it could break code without notice. What do you think?

bkamins:
I propose that I finish this PR (I will remove WIP then).
Then we can have a look how much copying was performed.
Actually - given your proposal - the code that is not needed is not that hard to track 😄.

However, I would retain deprecation - even for a short period to simplify debugging (even something what Julia 0.7 and 1.0 had - actually it worked well from my experience).

The big question is if we wait for `setindex!` changes and improved performance of groupby before the release (I would wait for it, but I am not sure what is the status of https://github.com/JuliaData/DataFrames.jl/pull/1571).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1035,7 +1035,9 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
     length(header) == 0 && return DataFrame()
     cols = Vector{AbstractVector}(undef, length(header))
     for (i, name) in enumerate(header)
-        data = [df[name] for df in dfs]
+        # TODO: replace with commented out code after getindex deprecation
+        # data = [df[name] for df in dfs]
+        data = [df isa DataFrame ? df[name] : df[:, name] for df in dfs]
```
nalimilan: Wouldn't it be better to always use `view` here (even if that's temporary)? Same for `completecases` and maybe other places.

bkamins: We could do this, but this would be breaking.
That is: currently `df[name]` when `df` is a `SubDataFrame` produces a copy and so I used `df[:, name]` to still make a copy temporarily (it will be a view after `getindex` deprecation).

Do you think making a `view` now would be better?

nalimilan: But here it's an implementation detail which shouldn't have any user-visible consequences, right?

bkamins: Ah - right. I mixed it up with `hcat!` (this PR is really long). I will push a PR changing copy to `view` in places where it is not user visible.

bkamins: I have pushed a PR implementing what you have requested. Actually it is a good way to test if something will not break in the future. The only limitation we have is with `vcat` with which you have started (I have noted it in the code).

bkamins:
@nalimilan This should be cleaned up now for the next round of checking.

The only method that is left problematic is `Base.getindex(itr::Cols, inds...) = getindex(itr.df, inds...)` as it is sometimes called internally and occasionally can throw a warning, but this should not be much.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1035,7 +1035,12 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
     length(header) == 0 && return DataFrame()
     cols = Vector{AbstractVector}(undef, length(header))
     for (i, name) in enumerate(header)
-        data = [df[name] for df in dfs]
+        # TODO: replace with commented out code after getindex deprecation
+        # data = [df[name] for df in dfs]
+        # the code below assumes that only DataFrame and SubDataFrame
+        # are subtypes of AbstractDataFrame
+        # it should be removed ASAP after deprecation
+        data = [df isa DataFrame ? df[name] : view(parent(df)[name], rows(df)) for df in dfs]
```
nalimilan: It shouldn't matter much since it's temporary, but better call `view` only for `SubDataFrame`, so that the fallback works for any type. Then you don't need the scary comment.

bkamins: But I cannot do this here because `SubDataFrame` is not defined at this stage of DataFrames.jl loading. And I would have to move the whole function much later in the loading process to make it work, which I think should not be done since it is only a temporary change.

nalimilan: Argh. Well, let's say we will remove this soon anyway...

bkamins:
Good to merge (I do not want to do it myself, as it is too delicate 😄)?

nalimilan:
So scary...

PR #1588, opened 2018-11-06
cleanup map function

bkamins:
Changes in this PR:
* add `Type` in union with `Function` as a first argument to `map` (this omission produced inconsistencies in behavior of code as if we passed `Type` as a first argument a mathod of `map` from Base was called instead of custom `map`) - I left out writing tests for `map` on `groupby` in this case as it should be done in #1520 right (@nalimilan)?
* remove `map` for `DFRowIterator` as - according to my understanding it just re-implemented functionality from Base (i.e. without this definition all should work the same) - please correct me if I am wrong
* deprecate `map` for `SubDataFrame`

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,10 +1,5 @@
 import Base: @deprecate
 
-function Base.map(f::Function, sdf::SubDataFrame)
-    Base.depwarn("Calling map on SuBDataFrame is deprecated", :map)
-    f(sdf)
-end
-
 @deprecate by(d::AbstractDataFrame, cols, s::Vector{Symbol}) aggregate(d, cols, map(eval, s))
 @deprecate by(d::AbstractDataFrame, cols, s::Symbol) aggregate(d, cols, eval(s))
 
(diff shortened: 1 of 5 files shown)


src/deprecated.jl
```diff
@@ -1,5 +1,10 @@
 import Base: @deprecate
 
+function Base.map(f::Function, sdf::SubDataFrame)
+    Base.depwarn("Calling map on SuBDataFrame is deprecated", :map)
```
ararslan: Erroneously capitalized B there. Also, could this not just be `@deprecate map(f::Function, sdf::SubDataFrame) f(sdf)`?

bkamins: Actually now - after looking at #1520 I can see that @nalimilan included this deprecation there so I will remove it from this PR.

nalimilan: Better keep the function here then, and remove it when adding the deprecation at #1520.

bkamins: This is what I did if I understand you correctly

nalimilan: What I mean is that you shouldn't remove this line: https://github.com/JuliaData/DataFrames.jl/pull/1588/files#diff-7b74440e6f375ca303e4c5ef15117c26L220

bkamins: Right - sorry for forgetting this. Fixed

Issue #1586, opened 2018-11-06
Do we need columns and eachcol


bkamins:
We currently have `columns` (not exported) function and `eachcol` (exported) that essentially serve the same purpose. Do we need to keep them both and what should be the direction in design here?

CC @nalimilan

nalimilan:
`eachcol` is different from `columns` as it yields `(name, vector)` tuples. But indeed I've wondered for some time whether we shouldn't deprecate `eachcol` and export `columns`. We could also keep both, it doesn't really hurt and it might be a bit faster than doing `df[name]` repeatedly (only significant when there are only a few rows, though).

bkamins:
I am aware that they are a bit different, but I meant that `eachcol` covers all use cases of `columns` and should not be that much slower so actually `columns` is not needed if we keep `eachcol`.

For me it only hurts that it makes the design of DataFrames.jl package more complex to have two types serve almost the same purpose: `Cols` and `DFColumnIterator` (it is complex enough already which I have learned while working on `getindex` 😄).

nalimilan:
OK. But `eachcol` is also inconvenient if you just want to apply a function to all columns in a data frame.

bkamins:
Additional points to gather the facts (the first of them was discussed earlier):
* `columns` conflicts with `Tables.columns` (I know that there it is not exported, but nevertheless it might be confusing)
* we have an implementation which effectively creates `Cols` object for `SubDataFrame` and accesses field `columns` for `DataFrame` - so it is not a unified interface;
* I have checked the usage of `columns` in the package (which is enough as this function is internal) and actually there are not many places where it is used.

pdeffebach:
+1 to the `SubDataFrame` concern. Since `eachcol` is an iterator, it can be used with a subdataframe easily. But `columns` returns a vector of vectors, making (i think) it infeasible for a subdataframe.

bkamins:
@pdeffebach Note that is that currently `columns(sdf)` where `sdf` is a `SubDataFrame` is defined and returns object of type `Cols`.

pdeffebach:
Ah that makes sense. Sorry for the confusion. Yeah it makes sense that `Cols` is a bit too one-off to be user-facing.

nalimilan:
Makes sense.

bkamins:
[...]

PR #1590, opened 2018-11-08
Unify eachcol and columns functions

bkamins:
Implementing what was discussed in #1586.


bkamins:
This is still WIP as I have to think if we want to make `DFRowIterator` and `DFColumnIterator` subtypes of `AbstractVector` and if yes what is the best way to do it.

src/abstractdataframe/iteration.jl
```diff
@@ -27,25 +27,32 @@ with each row represented as a `DataFrameRow`.
 """
 eachrow(df::AbstractDataFrame) = DFRowIterator(df)
 
+Base.size(itr::DFRowIterator) = (size(itr.df, 1), )
+Base.size(itr::DFRowIterator, ix) =
```
nalimilan: This method shouldn't be needed if you declare that `DFRowIterator <: AbstractVector{DataFrameRow}` according to https://docs.julialang.org/en/latest/manual/interfaces/#man-interface-array-1. Then you won't need `firstindex`, `lastindex`, `eltype` and `iterate` either.

src/abstractdataframe/iteration.jl
```diff
@@ -27,25 +27,32 @@ with each row represented as a `DataFrameRow`.
...
 # Iteration by columns
 """
-    DFColumnIterator{<:AbstractDataFrame}
+    DFColumnIterator{<:AbstractDataFrame, C}
 
 Iterator over columns of an `AbstractDataFrame`.
-Each returned value is a tuple consisting of column name and column vector.
-
-A value of this type is returned by the [`eachcol`](@link) function.
+If `C` is `true` (a value returned by the [`eachcol`](@link) function)
```
```suggestion
If `C` is `true` (which is the case when calling [`eachcol`](@link))
```
(same below)

src/abstractdataframe/iteration.jl
```diff
@@ -55,10 +62,6 @@ end
 Return a `DFColumnIterator` that iterates an `AbstractDataFrame` column by column.
 Iteration returns a tuple consisting of column name and column vector.
 
-`DFColumnIterator` has a custom implementation of the `map` function which
```
nalimilan: This feature sounds useful, why remove it? People can use comprehensions if they want an array.

src/abstractdataframe/iteration.jl
```diff
@@ -72,29 +75,73 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
-    res = DataFrame()
-    for (n, v) in eachcol(dfci.df)
-        res[n] = f(v)
+
+Base.eltype(::DFColumnIterator{<:AbstractDataFrame,true}) =
+    Tuple{Symbol, AbstractVector}
+
+function Base.getindex(itr::DFColumnIterator{<:AbstractDataFrame,true}, j)
+    # TODO: change to the way getindex for false is defined below afted deprecation
+    if !(j isa Integer) || j isa Bool
```
nalimilan: Why print the warning only for these types?

src/abstractdataframe/iteration.jl
```diff
@@ -72,29 +75,73 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
+        Base.depwarn("calling getindex on DFColumnIterator{<:AbstractDataFrame,true} " *
+                     " object will only accept integer indexing and will return " *
```
nalimilan: Would be nice to advise using `columns` instead of `eachcol`. `DFColumnIterator` is an implementation detail which doesn't need to be mentioned. Also, remove leading space.

src/abstractdataframe/iteration.jl
```diff
@@ -27,25 +27,32 @@ with each row represented as a `DataFrameRow`.
...
-Each returned value is a tuple consisting of column name and column vector.
-
-A value of this type is returned by the [`eachcol`](@link) function.
+If `C` is `true` (a value returned by the [`eachcol`](@link) function)
+then each returned value is a tuple consisting of column name and column vector.
+If `C` is `false` (a value returned by the [`columns`](@link) function)
+then each returned value is a column vector.
 """
-struct DFColumnIterator{T <: AbstractDataFrame}
+struct DFColumnIterator{T<:AbstractDataFrame, C}
```
nalimilan: Instead of `C` being a Boolean, it could be either `AbstractVector` or `Tuple{Symbol,AbstractVector}`. That way you could do `<: AbstractArray{C}` here, and you wouldn't need to define `eltype` and such below.

BTW, it's a good occasion to move from a tuple to a pair. That shouldn't affect most uses but it's more natural here.

src/abstractdataframe/iteration.jl
```diff
@@ -72,29 +75,73 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
+ (:y, [11, 12, 13, 14])
+```
+"""
+eachcol(df::T) where T<: AbstractDataFrame = DFColumnIterator{T, true}(df)
+
+"""
+    columns(df::AbstractDataFrame)
+
+Return a `DFColumnIterator` that iterates an `AbstractDataFrame` column by column.
+Iteration returns a column vector.
```
```suggestion
, yielding column vectors.
```

src/abstractdataframe/iteration.jl
nalimilan: That said, if we keep the special behavior of `map` which returns a `DataFrame`, maybe we'd better not make this iterator an `AbstractVector`. Same for the row iterator if we want to do something similar in the future.

src/abstractdataframe/iteration.jl

src/abstractdataframe/iteration.jl

bkamins:
Having thought about it, I will go on with what I have suggested and then it lend under review.

src/abstractdataframe/iteration.jl
bkamins: To clarify - an old implementation of `getindex` for `eachcol` had a bug and allowed indexing with vectors which produced a `DataFrame` as a result - we do not want to allow for such indexing.

src/abstractdataframe/iteration.jl
nalimilan: OK, so we need to choose between these two solutions:
1. Define `map(f, columns(df))` to return a data frame. We could add `map(f, rows(df))` later if we want and that would be consistent. People have to use comprehensions to get a vector instead.
2. Keep the standard behavior of `map(f, columns(df))`, i.e. have it return a vector. This is basically `colwise`, but with a more standard name (and we could deprecate it, the broadcasting form `f.(columns(df))` is even more compact). We need another way to apply a function to each column of a data frame and get a data frame back: maybe `mapcols(f, df)`, and `mapcols!(f, df)` for the in-place variant (the naming would be consistent with other column-wise functions, cf. https://github.com/JuliaData/DataFrames.jl/issues/1514). Then `map(f, df)` would apply a function to each row (this is what JuliaDB [supports](http://juliadb.org/latest/api/selection.html#Map-1)).

> Also, I do not see much sense of doing the same functionality on `eachrow` (we would not know how to transform it to a `DataFrame` later).

That would be quite simple AFAICT: require the user-provided function to return a named tuple, and fill a data frame with that like we now do in `map(::Function, ::GroupedDataFrame)`.

Cc: @piever to discuss the consistency with JuliaDB?

bkamins:
I have rewritten the PR to implement subtype of `AbstractVector` (with renamed types) and added `mapcols` function. The design is significantly better, but it is breaking with `getindex` for `eachcol` (this is what I was afraid about). We cannot avoid this breakage without significant complication of code. I would go for this breaking change as now we get all good things `Base` implements for `AbstractVector`s.

bkamins:
> Don't you think we could add a deprecated getindex method to ease the transition?

The problem is that in the old design iteration protocol and `getindex` are inconsistent:
* iteration protocol returns a tuple `(name, value)`; 
* `getindex` returns a value.

If we switch to `AbstractVector` interface then we do not need to use iteration as it uses `getindex`. In order to ease the transition we would have to introduce this inconsistency back (and possibly in some cases some mechanisms from Base could break as they assume that `getindex` and iteration protocol are consistent).

I would assume that people did not use `getindex` on `eachcol` return value much (if at all) - the past design encouraged using it only for iteration.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -218,7 +202,7 @@ eltypes(df)
 ```
 
 """
-eltypes(df::AbstractDataFrame) = map!(eltype, Vector{Type}(undef, size(df,2)), columns(df))
+eltypes(df::AbstractDataFrame) = eltype.(columns(df))
```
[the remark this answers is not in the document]
bkamins: I think that having `eltypes` does not hurt and it is a useful function.
`eltype.(columns(df))` is not very verbose but I think it would not obvious for newcomers.

What are the arguments for deprecating it?

src/abstractdataframe/io.jl
```diff
@@ -276,26 +276,25 @@ function DataFrame(sch::Data.Schema{R}, ::Type{S}=Data.Field,
...
             sch.rows = newsize
         end
         # take care of a possible reference from source by addint to WeakRefStringArrays
         if !isempty(reference)
             foreach(col-> col isa WeakRefStringArray && push!(col.data, reference),
-                sink.columns)
+                    rawcolumns(sink))
         end
-        sink = DataFrameStream(sink)
+        DataFrameStream(sink)
```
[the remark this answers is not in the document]
bkamins: fixed

src/abstractdataframe/iteration.jl
```diff
@@ -8,57 +8,49 @@
...
-function Base.iterate(itr::DFRowIterator, i=1)
-    i > size(itr.df, 1) && return nothing
-    return (DataFrameRow(itr.df, i), i + 1)
-end
-Base.eltype(::DFRowIterator{T}) where {T} = DataFrameRow{T}
-Base.size(itr::DFRowIterator) = (size(itr.df, 1), )
-Base.length(itr::DFRowIterator) = size(itr.df, 1)
-Base.getindex(itr::DFRowIterator, i) = DataFrameRow(itr.df, i)
+Base.size(itr::DFRowVector) = (size(itr.df, 1), )
+Base.IndexStyle(::Type{<:DFRowVector}) = Base.IndexLinear()
```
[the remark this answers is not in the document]
bkamins: OK

src/abstractdataframe/iteration.jl
```diff
@@ -8,57 +8,49 @@
 
 # Iteration by rows
 """
-    DFRowIterator{<:AbstractDataFrame}
+    DFRowVector{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow{T}}
```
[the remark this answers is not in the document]
bkamins: OK

src/abstractdataframe/iteration.jl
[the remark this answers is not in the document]
bkamins: Sorry - this was meant for the earlier comment. I guess it is not strictly required, but it implies how `getindex` method should be defined. For `IndexLinear` the definition is a bit simpler so I would leave it as is (see discussion in https://docs.julialang.org/en/latest/manual/interfaces/#man-interface-array-1).

src/abstractdataframe/iteration.jl
```diff
@@ -8,57 +8,49 @@
...
-    i > size(itr.df, 1) && return nothing
-    return (DataFrameRow(itr.df, i), i + 1)
-end
-Base.eltype(::DFRowIterator{T}) where {T} = DataFrameRow{T}
-Base.size(itr::DFRowIterator) = (size(itr.df, 1), )
-Base.length(itr::DFRowIterator) = size(itr.df, 1)
-Base.getindex(itr::DFRowIterator, i) = DataFrameRow(itr.df, i)
+Base.size(itr::DFRowVector) = (size(itr.df, 1), )
+Base.IndexStyle(::Type{<:DFRowVector}) = Base.IndexLinear()
+Base.getindex(itr::DFRowVector, i::Int) = DataFrameRow(itr.df, i)
```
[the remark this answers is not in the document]
bkamins: `::Int` follows `AbstractArray` interface for `LinearIndex` - and this is really nice, that all other indexing options are handled for free by Base.

I will add bounds checking and `@inbounds` annotation.

test/constructors.jl
```diff
@@ -1,14 +1,14 @@
...
+    using DataFrames: Index, rawcolumns, index
     const ≅ = isequal
 
     #
     # DataFrame
     #
     @testset "constructors" begin
         df = DataFrame()
-        @test columns(df) == Any[]
+        @test rawcolumns(df) == AbstractVector[]
```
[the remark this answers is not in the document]
bkamins: I know, but I wanted to keep the old test. I will rewrite it

nalimilan:
A temporary inconsistency isn't a problem.

bkamins:
But the problem is that then the following will produce different results:
`[f(col) for col in eachcol(df)]`
and
`[f(eachcol(df)[i]) for i in 1:ncol(df)]`

And now a question: what will `f.(eachcol(df))` produce? Unless you are an expert in how broadcasting is internally implemented you have to guess.

In consequence users really would not know what they would get when passing `eachrow(df)` to some function as it would depend how the function accesses `eachrow(df)`.

Also now `eachrow(df)[1]` returns a vector and `eachrow(df)[1:2]` returns a `DataFrame`. Do we really want to keep this mess - even for a short period (I would say that current behavior of `getindex` for `eachrow` is on the border of being a bug)?

src/abstractdataframe/iteration.jl
```diff
@@ -72,28 +64,85 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
-│ 1   │ 10    │ 50    │
+│ 1   │ 1     │ 11    │
+│ 2   │ 2     │ 12    │
+│ 3   │ 3     │ 13    │
+│ 4   │ 4     │ 14    │
+
+julia> collect(columns(df))
+2-element Array{AbstractArray{T,1} where T,1}:
+ [1, 2, 3, 4]
+ [11, 12, 13, 14]
```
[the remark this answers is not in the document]
bkamins: added

src/abstractdataframe/iteration.jl
```diff
@@ -72,28 +64,85 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
-end
-Base.eltype(::DFColumnIterator) = Tuple{Symbol, AbstractVector}
-Base.size(itr::DFColumnIterator) = (size(itr.df, 2), )
-Base.length(itr::DFColumnIterator) = size(itr.df, 2)
-Base.getindex(itr::DFColumnIterator, j) = itr.df[j]
-function Base.map(f::Union{Function,Type}, dfci::DFColumnIterator)
+Base.size(itr::DFColumnVector) = (size(itr.df, 2),)
+Base.IndexStyle(::Type{<:DFColumnVector}) = Base.IndexLinear()
+Base.getindex(itr::DFColumnVector{<:AbstractDataFrame,
+                                    Pair{Symbol, AbstractVector}}, j::Int) =
```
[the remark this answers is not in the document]
bkamins: This is inconsistent with `AbstractVector` interface, it should be `Int` here.

src/deprecated.jl
```diff
@@ -1740,7 +1731,10 @@ function hashrows(df::SubDataFrame, skipmissing::Bool)
     return (rhashes, missings)
 end
 
+Base.getproperty(df::SubDataFrame, col_ind::Symbol) = getindex(df, :, col_ind)
```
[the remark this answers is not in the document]
bkamins: OK - will change.

test/cat.jl
```diff
@@ -165,13 +164,13 @@ module TestCat
     @testset "vcat mixed coltypes" begin
         df = vcat(DataFrame([[1]], [:x]), DataFrame([[1.0]], [:x]))
         @test df == DataFrame([[1.0, 1.0]], [:x])
-        @test typeof.(columns(df)) == [Vector{Float64}]
+        @test map(typeof, columns(df)) == [Vector{Float64}]
```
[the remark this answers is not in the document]
bkamins: OK - I will revert all the changes

nalimilan:
We don't necessarily need to preserve all bugs, but if that's not too hard I'd keep at least a deprecation for the current behavior of `eachcol(df)[i]`. We can remove it soon if we want to get the more consistent behavior.

src/abstractdataframe/iteration.jl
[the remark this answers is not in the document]
nalimilan: Ah, right, conversion methods are provided automatically for `AbstractArray`.

src/abstractdataframe/abstractdataframe.jl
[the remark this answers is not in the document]
nalimilan: Nothing special, I just find `eltype.(columns(df))` more explicit and standard. But we can discuss this elsewhere.

bkamins:
OK - in https://github.com/JuliaData/DataFrames.jl/pull/1590/commits/52e90107dc62e16211c83b2865832c9ad3c66c84 you have what had to be changed to better support this intermediate deprecation.

nalimilan:
Hmm... I guess we can also override `collect` for now. Anyway that's temporary -- and you've already duplicated half of the code base to silence `SubDataFrame` deprecations. ;-)

src/abstractdataframe/iteration.jl
```diff
@@ -72,28 +67,111 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
+ [1, 2, 3, 4]
+ [11, 12, 13, 14]
+
+julia> sum.(columns(df))
+2-element Array{Int64,1}:
+ 10
+ 50
+
+julia> map(columns(df)) do col
+       maximum(col) - minimum(col)
```
```suggestion
           maximum(col) - minimum(col)
```

src/abstractdataframe/iteration.jl
```diff
@@ -72,28 +67,111 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
 ```
 """
-eachcol(df::AbstractDataFrame) = DFColumnIterator(df)
+columns(df::T) where T<: AbstractDataFrame =
+    DataFrameColumns{T, AbstractVector}(df)
+
+Base.size(itr::DataFrameColumns) = (size(itr.df, 2),)
+Base.IndexStyle(::Type{<:DataFrameColumns}) = Base.IndexLinear()
+
+function Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame,
```
nalimilan: `@propagate_inbounds` wouldn't hurt here and below.

src/abstractdataframe/iteration.jl
```diff
@@ -72,28 +67,111 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
-Base.size(itr::DFColumnIterator) = (size(itr.df, 2), )
-Base.length(itr::DFColumnIterator) = size(itr.df, 2)
-Base.getindex(itr::DFColumnIterator, j) = itr.df[j]
-function Base.map(f::Union{Function,Type}, dfci::DFColumnIterator)
+
+"""
+    mapcols(f::Union{Function,Type}, df::AbstractDataFrame)
+
+Return a `DataFrame` where each column of `df` is transformed using function `f`.
+Note that `f` must return values of consistent length.
```
```suggestion
`f` must return `AbstractVector` objects all with the same length.
```

test/grouping.jl
```diff
@@ -318,9 +318,13 @@ module TestGrouping
                        c = Vector{Union{Float64, Missing}}(randn(8)))
 
         cols = [:a, :b]
-        f(df) = DataFrame(cmax = maximum(df[:c]))
-        g(df) = (cmax = maximum(df[:c]),)
-        h(df) = maximum(df[:c])
+        f(df) = DataFrame(cmax = maximum(df[:, :c]))
```
nalimilan: https://github.com/JuliaData/DataFrames.jl/pull/1592 will fix this so better leave it out from this PR (thanks for spotting it!).

test/iteration.jl
```diff
@@ -19,18 +15,34 @@ module TestIteration
...
+    @test size(columns(df)) == (size(df, 2),)
     @test length(eachcol(df)) == size(df, 2)
-    @test eachcol(df)[1] == df[1]
-    @test collect(eachcol(df)) isa Vector{Tuple{Symbol, AbstractVector}}
-    @test eltype(eachcol(df)) == Tuple{Symbol, AbstractVector}
+    @test length(columns(df)) == size(df, 2)
+    @test eachcol(df)[1] == df[1] # this will be (:A => df[1]) after deprecation
+    @test columns(df)[1] == df[1]
+    @test collect(eachcol(df)) isa Vector{AbstractVector} # this will be Vector{Pair{Symbol, AbstractVector}} after deprecation
+    @test collect(eachcol(df)) == [:A => Union{Missing, Int64}[1, 2]
```
nalimilan: `Union{Missing, Int64}` isn't needed.

src/abstractdataframe/iteration.jl
bkamins: AFAIK it will not change anything significantly as we do boundschecking anyway when we try to access column number `j` from `df` (note that `df` is an `AbstractDataFrame` so we do not know at this point how it stores data).

src/abstractdataframe/iteration.jl
bkamins: `f` may also return scalars and this will be OK so your correction is not 100% exact. Do we leave my version or add an option of a scalar to your version of wording?

test/grouping.jl

test/iteration.jl
bkamins: OK. I will leave that out, but I thought it does not hurt either 😄.

src/abstractdataframe/iteration.jl
bkamins: thinking of it - I will just correct your correction.

bkamins:
I will override `collect`, but please note that some other function will have the same problem.

In summary: after we merge this, what other PRs would you want to finish before tagging a release. After you tag a release I will immediately submit a PRs (probably it is better to have several for different types of deprecations) cleaning up all this mess.

src/abstractdataframe/iteration.jl
piever: For me to understand, there are a `ColumnIterator` and `RowIterator` types, and the decision is whether one should overload `map` on them to return a `DataFrame` rather than a vector of columns or a vector of rows?

In JuliaDB `map` acts by default on the rows and returns a `table`. We also have a `map_rows` which basically would correspond to `map_rows(f, iter...) = DataFrame(f(i) for i in iter...)`. It is useful when `f` returns `NamedTuples` but `iter...` are not a `DataFrame` (they could be vectors or iterators), for example `map_rows((x, y) -> (s = x+y, d = x-y), rand(100), rand(100))`. Could this be added to `DataFrames` as well? In which case I think `mapcols` `map_rows` is a strange pair. What could be a better pair of names?

For consistency with JuliaDB, option 2 seems much better (also, to keep `map` and `filter` consistent). I thought there was a `colwise` or similar in JuliaDB but apparently I'm misremembering. It looks like currently one has to do:

```julia
julia> t = table((a = rand(100), b = randn(100)));

julia> f(v) = partialsort(v, 1:10)

julia> table(map(f, columns(t)))
Table with 10 rows, 2 columns:
a           b
────────────────────
0.00392291  -2.52979
0.0221383   -2.01489
0.029867    -1.92054
0.0462468   -1.89799
0.0720458   -1.5951
0.0824684   -1.31414
0.105048    -1.25503
0.108803    -1.19262
0.1204      -1.13395
0.136053    -1.13125
```

This I think is not so bad but maybe it'd be more awkward in DataFrames because `DataFrame(map(f, columns(t)))` would lose the column names?

pdeffebach:
> But the problem is that then the following will produce different results:
> `[f(col) for col in eachcol(df)]`
> and
> `[f(eachcol(df)[i]) for i in 1:ncol(df)]`
> 
> And now a question: what will `f.(eachcol(df))` produce? Unless you are an expert in how broadcasting is internally implemented you have to guess.

Note that `[f(eachcol(df)[i]) for i in 1:ncol(df)]` is very unlikely to be used very much because `f` has to take in a `(name => column)` pair. Because of that, I think that the divergence between `Base.broadcast`, `Base.map`, and `f(col) for col in eachcol` is probably fine.  And I agree that is a preferred solution fo `colwise`.

src/abstractdataframe/iteration.jl
nalimilan: Yes but propagating this is better anyway, and the data frame implementation could use `@boundscheck` at some point.

src/abstractdataframe/iteration.jl
nalimilan: Good point. Then we should also test that it works.

src/abstractdataframe/iteration.jl
nalimilan: > In JuliaDB `map` acts by default on the rows and returns a `table`. We also have a `map_rows` which basically would correspond to `map_rows(f, iter...) = DataFrame(f(i) for i in iter...)`. It is useful when `f` returns `NamedTuples` but `iter...` are not a `DataFrame` (they could be vectors or iterators), for example `map_rows((x, y) -> (s = x+y, d = x-y), rand(100), rand(100))`. Could this be added to `DataFrames` as well? In which case I think `mapcols` `map_rows` is a strange pair. What could be a better pair of names?

I don't really see the advantage of providing a `map_rows` function if `DataFrame(f(i) for i in iter...)` does the same thing TBH. And using a generator like that sounds more Julian (it's like creating dicts).

> For consistency with JuliaDB, option 2 seems much better (also, to keep `map` and `filter` consistent).

OK. That's good since that's the approach @bkamins has implemented. ;-)

But the example you show below goes a bit against with this statement AFAICT: `map(f, columns(t))` returns a `NamedTuple` in JuliaDB, not a vector (which is why names are lost). That's logical since `column(t)` gives a `NamedTuple`, but that would be different from data frames. That said, it's not a big deal. Maybe JuliaDB could support `mapcols` just for consistency and convenience?

src/abstractdataframe/iteration.jl
```diff
@@ -138,11 +138,17 @@ function Base.iterate(itr::DataFrameColumns{<:AbstractDataFrame,
     return (_names(itr.df)[j] => itr.df[j], j + 1)
 end
 
+# TODO: remove this after deprecation period of getindex of DataFrameColumns
```
nalimilan: Why not put this in deprecated.jl?

nalimilan:
I'd like to merge https://github.com/JuliaData/DataFrames.jl/pull/1591 and finish a PR to speed up grouping on categorical arrays. Then yes, it would be nice to clean up all this mess.

src/abstractdataframe/iteration.jl
piever: Yep, maybe JuliaDB could deprecate `map_rows` as "unnecessary". The easiest is probably to add a `mapcols(f, t; select = All())` method to JuliaDB. As I mentioned above, I'm afraid this creates a bit of cacophony on the JuliaDB side (`mapcols` and `map_rows` have different underscore style and are not related in the obvious way). The cleanest is probably to deprecate `map_rows` and rename `setcol`, `popcol`, `pushcol`... to `setcols`, `popcols`, `pushcols` (as they accept multiple columns at once) and add `mapcols` to the collections. But this PR shouldn't worry about that as all the changes to make things consistent should come from JuliaDB.

nalimilan: Makes sense. It's good to anticipate to what API we both want to converge in the long term.

src/abstractdataframe/iteration.jl
bkamins: I have opened a separate issue https://github.com/JuliaData/DataFrames.jl/issues/1594 for this as this PR is complex enough and I am not sure if it will help anything (performancewise - disabling checking bounds saves us nanoseconds but then we have a dynamic dispatch on the result which is orders of magnitude more expensive).

src/abstractdataframe/iteration.jl
bkamins: It will be simpler for me to clean it up later (and I assume I will do it just after the coming release).
deprecated.jl has 62kb and almost 2000 loc now and actually it is easy to forget to fix something there.

src/abstractdataframe/iteration.jl
nalimilan: I'm not saying it matters, but it's cheap to add, so...

bkamins:
@nalimilan - thank you for the patience with this PR. I believe the having a first-class views of columns and rows of a `DataFrame` as abstract vectors will be very useful.

I have additionally opened https://github.com/JuliaData/DataFrames.jl/issues/1595 to discuss the future of `colwise` and similarly https://github.com/JuliaData/DataFrames.jl/issues/1596 for `eltypes` (both things came up in the discussion and I think I would wait if people would want to comment on those issues before deciding how to move forward).

nalimilan:
You're welcome!

bkamins:
If there will be no additional comments to this PR I will merge it tomorrow.

PR #1591, opened 2018-11-10
Deprecate length(df::AbstractDataFrame) in favor of size(df, 2)

### LINKED ISSUE #1200: length(::DataFrame) returns number of columns

nalimilan:
This definition goes against the view of data frames as collections of rows.

Fixes #1200. Part of https://github.com/JuliaData/DataFrames.jl/issues/1514.

(Coverage drop seems to be spurious, since the `axes` method is called by the one above, which is covered.)

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -231,10 +231,9 @@ function Base.size(df::AbstractDataFrame, i::Integer)
     end
 end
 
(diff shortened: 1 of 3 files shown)


nalimilan:
> The `lastindex` definition goes a bit against the contract in Base, but it has to be as is to support usage of `end` in indexing - right?

Yes. Anyway it's much less confusing than `length`.

PR #1600, opened 2018-11-15
Add fast grouping method for CategoricalArray keys

nalimilan:
When grouping on a single `CategoricalArray` column, we don't need a `Dict` to get group codes.
This approach cannot be used when joining, as the `Dict` is needed.
`group_rows` now handles the sorting, which allows skipping this step if the grouping is guaranteed to be sorted (as is the case for `CategoricalArray`).
This mechanism can be extended in the future to handle multiple `CategoricalArray` keys, or integer keys (using radix sort).

This makes grouping about 4 times faster on simple benchmarks, even with the optimized code from https://github.com/JuliaData/DataFrames.jl/pull/1565 (which BTW is still used when grouping on several `CategoricalArray` keys). Of course grouping only represents a small part of the time taken by a split-apply-combine operation, so typically the win isn't that big.
```julia
df = DataFrame(rand(500_000, 10));
df.key = categorical(string.(rand(1:100, 500_000)));

# Master
julia> @btime groupby(df, :key);
  8.258 ms (51 allocations: 15.45 MiB)

# This PR
julia> @btime groupby(df, :key);
  2.284 ms (64 allocations: 7.64 MiB)

df = DataFrame(rand(500_000, 10));
df.key = categorical(string.(rand(1:10_000, 500_000)));

# Master
julia> @btime groupby(df, :key);
  27.451 ms (54 allocations: 15.68 MiB)

# This PR
julia> @btime groupby(df, :key);
  7.595 ms (69 allocations: 7.86 MiB)
```

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -4,8 +4,6 @@
 struct RowGroupDict{T<:AbstractDataFrame}
     "source data table"
     df::T
-    "number of groups (i.e. highest group index: some groups may be empty)"
-    ngroups::Int
     "row hashes (optional, can be empty)"
     rhashes::Vector{UInt}
     "hashindex -> index of group-representative row (optional, can be empty)"
@@ -98,7 +96,8 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
     @assert sz >= length(rhashes)
     szm1 = sz-1
     gslots = zeros(Int, sz)
-    # If missings are to be skipped, they will all go to group 1
+    # If missings are to be skipped, they will all go to group 1,
+    # which will be removed by group_rows
     ngroups = skipmissing ? 1 : 0
     @inbounds for i in eachindex(rhashes)
         # find the slot and group index for a row
@@ -129,7 +128,7 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
             groups[i] = gix
         end
     end
-    return ngroups, rhashes, gslots, false, true
+    return ngroups, rhashes, gslots, false, false
 end
 
 function row_group_slots(cols::Tuple{CategoricalVector},
@@ -139,7 +138,8 @@ function row_group_slots(cols::Tuple{CategoricalVector},
     col = cols[1]
     @assert groups === nothing || length(groups) == length(col)
 
-    # If missings are to be skipped, they will all go to group 1
+    # If missings are to be skipped, they will all go to group 1,
+    # which will be removed by group_rows
     ngroups = length(levels(col)) + (eltype(col) >: Missing)
 
     if groups !== nothing
@@ -164,7 +164,8 @@ end
 # - hash: whether row hashes should be computed (if false, the rhashes and gslots fields
 #   hold empty vectors)
 # - sort: whether groups should be sorted
-# - skipmissing: whether rows with missing values should be skipped rather than put into group 0
+# - skipmissing: whether rows with missing values should be skipped
+#   rather than put into a separate group
 function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false,
                     skipmissing::Bool = false)
     groups = Vector{Int}(undef, nrow(df))
@@ -228,7 +229,7 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
         Base.permute!!(stops, group_perm)
     end
 
-    return RowGroupDict(df, ngroups, rhashes, gslots, groups, rperm, starts, stops)
+    return RowGroupDict(df, rhashes, gslots, groups, rperm, starts, stops)
 end
 
 # Find index of a row in gd that matches given row by content, 0 if not found
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -4,11 +4,11 @@
 struct RowGroupDict{T<:AbstractDataFrame}
     "source data table"
     df::T
-    "number of groups"
+    "number of groups (i.e. highest group index: some groups may be empty)"
     ngroups::Int
-    "row hashes"
+    "row hashes (optional, can be empty)"
     rhashes::Vector{UInt}
-    "hashindex -> index of group-representative row"
+    "hashindex -> index of group-representative row (optional, can be empty)"
     gslots::Vector{Int}
     "group index for each row"
     groups::Vector{Int}
@@ -70,10 +70,11 @@ function hashrows_col!(h::Vector{UInt},
 end
 
 # Calculate the vector of `df` rows hash values.
-function hashrows(df::AbstractDataFrame, skipmissing::Bool)
-    rhashes = zeros(UInt, nrow(df))
-    missings = fill(false, skipmissing ? nrow(df) : 0)
-    for (i, col) in enumerate(columns(df))
+function hashrows(cols::Tuple{Vararg{AbstractVector}}, skipmissing::Bool)
+    len = length(cols[1])
+    rhashes = zeros(UInt, len)
+    missings = fill(false, skipmissing ? len : 0)
+    for (i, col) in enumerate(cols)
         hashrows_col!(rhashes, missings, col, i == 1)
     end
     return (rhashes, missings)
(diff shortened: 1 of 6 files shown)


src/dataframerow/utils.jl
```diff
@@ -4,11 +4,11 @@
 struct RowGroupDict{T<:AbstractDataFrame}
     "source data table"
     df::T
-    "number of groups"
+    "number of groups (i.e. highest group index: some groups may be empty)"
```
bkamins: nonempty group?

src/dataframerow/utils.jl
```diff
@@ -70,36 +70,30 @@ function hashrows_col!(h::Vector{UInt},
 end
 
 # Calculate the vector of `df` rows hash values.
-function hashrows(df::AbstractDataFrame, skipmissing::Bool)
-    rhashes = zeros(UInt, nrow(df))
-    missings = fill(false, skipmissing ? nrow(df) : 0)
-    for (i, col) in enumerate(columns(df))
+function hashrows(cols::Tuple{Vararg{AbstractVector}}, skipmissing::Bool)
+    len = length(cols[1])
```
bkamins: this assumes that `cols` is not `()` which is probably the case - asking to make sure that this corner case is not relevant.

src/dataframerow/utils.jl
```diff
@@ -70,36 +70,30 @@ function hashrows_col!(h::Vector{UInt},
...
-    row_group_slots(ntuple(i -> df[i], ncol(df)), rhashes, missings, groups, skipmissing)
-end
-
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
-                         rhashes::AbstractVector{UInt},
-                         missings::AbstractVector{Bool},
+                         hash::Val = Val(true),
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)
+                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool, Bool}
```
bkamins: why is this needed?

src/dataframerow/utils.jl
```diff
@@ -70,36 +70,30 @@ function hashrows_col!(h::Vector{UInt},
...
     @assert groups === nothing || length(groups) == length(cols[1])
-    # inspired by Dict code from base cf. https://github.com/JuliaData/DataFrames.jl/pull/17#discussion_r102481481
+    rhashes, missings = hashrows(cols, skipmissing)
```
bkamins: why is this needed and was not needed earlier?

src/dataframerow/utils.jl
```diff
@@ -70,36 +70,30 @@ function hashrows_col!(h::Vector{UInt},
...
-                         groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)
-    rhashes, missings = hashrows(df, skipmissing)
-    row_group_slots(ntuple(i -> df[i], ncol(df)), rhashes, missings, groups, skipmissing)
-end
-
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
-                         rhashes::AbstractVector{UInt},
-                         missings::AbstractVector{Bool},
+                         hash::Val = Val(true),
```
bkamins: I understand that you do not use this value now, but add it for the future. Right?

bkamins: Or it is to make sure that we can also call this method on a single `CategoricalVector` if we pass `Val{true}`?

src/dataframerow/utils.jl
```diff
@@ -135,14 +129,47 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+end
+
+function row_group_slots(cols::Tuple{CategoricalVector},
+                         hash::Val{false},
+                         groups::Union{Vector{Int}, Nothing} = nothing,
+                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool, Bool}
+    col = cols[1]
+    @assert groups === nothing || length(groups) == length(col)
+
+    # If missings are to be skipped, they will all go to group 1
```
bkamins: and this group will be skipped; otherwise missings to group `ngroups`. Right?

src/dataframerow/utils.jl
```diff
@@ -135,14 +129,47 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+    return ngroups, UInt[], Int[], true, true
 end
 
 # Builds RowGroupDict for a given DataFrame.
 # Partly uses the code of Wes McKinney's groupsort_indexer in pandas (file: src/groupby.pyx).
-function group_rows(df::AbstractDataFrame, skipmissing::Bool = false)
+# - hash: whether row hashes should be computed (if false, the rhashes and gslots fields
+#   hold empty vectors)
+# - sort: whether groups should be sorted
+# - skipmissing: whether rows with missing values should be skipped rather than put into group 0
```
bkamins: group `ngroups` not `0` unless I do not understand something.

src/dataframerow/utils.jl
```diff
@@ -170,11 +197,37 @@ function group_rows(df::AbstractDataFrame, skipmissing::Bool = false)
...
     if skipmissing
-        splice!(starts, 1)
-        splice!(stops, 1)
+        popfirst!(starts)
+        popfirst!(stops)
         ngroups -= 1
     end
 
+    # if row_group_slots can return empty groups, remove them
+    if empty
```
bkamins: is it possible that `empty` is `false`?

test/dataframerow.jl
```diff
@@ -45,7 +45,7 @@ module TestDataFrameRow
     @test hash(DataFrameRow(df, 2)) != hash(DataFrameRow(df, 6))
 
     # check that hashrows() function generates the same hashes as DataFrameRow
-    df_rowhashes, _ = DataFrames.hashrows(df, false)
+    df_rowhashes, _ = DataFrames.hashrows(tuple(DataFrames.columns(df)...), false)
```
bkamins: I would rather write `Tuple(columns(df))` to avoid splatting and you do not have to quality `columns` with `DataFrames.`.

src/dataframerow/utils.jl
nalimilan: I'm not sure I understand the question. Some of the "groups" counted here may be empty, so that just gives a value guaranteed to be as high or higher than the actual number of groups. (Actually this field doesn't seem to be used, so we could remove it.)

src/dataframerow/utils.jl
nalimilan: Yes, since you can't group on no key. :-)

src/dataframerow/utils.jl
nalimilan: Yes, that's just a way to dispatch to the right method: the optimized `CategoricalArray` method can only be called with a single `CategoricalArray` column and when `hash::Val{false}`; this one is the fallback, and can be called with any columns and `hash` value.

src/dataframerow/utils.jl
nalimilan: Yeah, that's annoying, it's because we call it from `group_rows`, which takes a `DataFrame`, and therefore doesn't know the types of the columns it passes to `row_group_slots`. Since all `row_group_slots` return the same argument types, the compiler should be able to figure the return type anyway, but that doesn't seem to work without the assertions.

src/dataframerow/utils.jl
nalimilan: It was called by the `row_group_slots` helper before, but now we call it only from this method, and not from the optimized `CategoricalArray` one.

src/dataframerow/utils.jl
nalimilan: Woops, yes, that's the case with the fallback method, but for some reason I put `true`...

src/dataframerow/utils.jl
bkamins: I asked if we count only non-empty groups. But if it is not used maybe better remove it?

src/dataframerow/utils.jl
bkamins: Yes, but `groupby(df, [])` throws now an obscure error (`BoundsError`) - maybe some clearer message should be printed.

src/dataframerow/utils.jl
nalimilan: I've expanded the comment a bit as that's confusing.

src/dataframerow/utils.jl
nalimilan: Let's remove it.

src/dataframerow/utils.jl
nalimilan: Actually, the group isn't necessarily the first nor the last one if `sort=false`.

bkamins:
Thank you. Looks good. One thing that might be added is to internally document the meaning of return value of `row_group_slots` somewhere for future developers (in a comment probably is enough).

bkamins:
Thank you. I think we are good to merge.

Do you have any benchmark how we stand now in the benchmark against other frameworks?

nalimilan:
> Do you have any benchmark how we stand now in the benchmark against other frameworks?

Not yet. I'd like to implement the type-stable `combine` here or in DataFramesMeta first.

nalimilan:
Though it shouldn't make a big difference for https://github.com/h2oai/db-benchmark/issues/30 when there are enough rows per group. So we could ask them to update the benchmark and run it again. But `SubDataFrame` deprecations will probably slow down the simplest syntax, so better drop them first.

bkamins:
They rerun them automatically on the latest release of benchmarked packages and they are updated regularly AFAIK, so we simply have to move forward with the package I guess 😄.

PR #1601, opened 2018-11-17
Support type-stable map and combine on GroupedDataFrame

nalimilan:
This is inspired by the [JuliaDB API](http://juliadb.org/latest/api/aggregation.html#IndexedTables.groupby). When columns are specified via `select`, the user-provided function is passed column vectors (as `SubArray`s) rather than a `SubDataFrame`. This gives fully type-stable code and dramatically improves performance.

This relatively limited change allows taking full advantage of the recent refactoring (https://github.com/JuliaData/DataFrames.jl/pull/1520). The performance gain is really incredible: about 200× for a simple sum after compilation.
```julia
               b = randn(800000))
gd = groupby(df, :a)

julia> @time combine(d -> sum(d.b), gd);
  4.657727 seconds (7.39 M allocations: 640.692 MiB, 2.92% gc time)

julia> @btime combine(d -> sum(d.b), gd);
  3.129 s (4399077 allocations: 494.68 MiB)

julia> @time combine(:b => sum, gd);
  0.545264 seconds (1.65 M allocations: 80.211 MiB, 3.91% gc time)

julia> @btime combine(:b => sum, gd);
  10.395 ms (399620 allocations: 16.78 MiB)
```

I still need to write tests, but otherwise the main question is the API. The discussion has already been started [on Discourse](https://discourse.julialang.org/t/dataframes-obtaining-the-subset-of-rows-by-a-set-of-values/15923/24?u=nalimilan). To sum up: while the `select` API illustrated above is reasonable, the slow variant is a bit shorter and more intuitive, in particular because it avoids repeating the column names in two different places. Yet we *really* don't want people to use the slow variant unless they really need to (for complex operations or when the number of columns isn't known in advance), since its terribly slow.

The solution I proposed to that problem is to detect whether the user-provided function expects a `SubDataFrame` or columns, assuming the argument names are those of columns. This would allow writing the example above as simply `combine(b -> sum(b), gd)`. Implementation-wise, it is actually quite easy to extract the argument names and match them to column names. The main problem is that when the function takes a single argument, we don't know whether it's supposed to be a `SubDataFrame` or a column. So far I haven't found a good solution to that, except writing `combine((b, args...) -> sum(b), gd)` or something silly like this.

Anyway, we don't necessarily need to decide this right now, as the `select` approach could be complementary to another more convenient approach: `select` can be useful to choose programmatically which column(s) to operate on, and a convenience syntax could be based on that for its implementation.

Cc: @bkamins, @piever, @pdeffebach

### DIFF
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -57,7 +57,7 @@ julia> by(iris, :Species, :PetalLength => mean)
 │ 2   │ versicolor │ 4.26             │
 │ 3   │ virginica  │ 5.552            │
 
-julia> by(iris, :Species, (N = :Species => length,)) # Chosen column is arbitrary
+julia> by(iris, :Species, N = :Species => length) # Chosen column is arbitrary
 3×2 DataFrame
 │ Row │ Species       │ N     │
 │     │ Categorical…⍰ │ Int64 │
@@ -67,7 +67,7 @@ julia> by(iris, :Species, (N = :Species => length,)) # Chosen column is arbitrar
 │ 3   │ virginica     │ 50    │
 ```
 
-julia> by(iris, :Species, (N = :Species => length, mean = :PetalLength => mean)) # Chosen column is arbitrary
+julia> by(iris, :Species, N = :Species => length, mean = :PetalLength => mean) # Column for length is arbitrary
 3×3 DataFrame
 │ Row │ Species    │ N     │ mean    │
 │     │ String⍰    │ Int64 │ Float64 │
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -6,7 +6,7 @@ The DataFrames package supports the split-apply-combine strategy through the `by
 1. a `col => function` pair indicating that `function` should be called with the vector of values for column `col`, which can be a column name or index
 2. a `cols => function` pair indicating that `function` should be called with a named tuple holding columns `cols`, which can be a tuple or vector of names or indices
 3. several such pairs, either as positional arguments or as keyword arguments (mixing is not allowed), producing each a single separate column; keyword argument names are used as column names
-4. equivalently, a (named) tuple or vector of such pairs, producing each a single separate column
+4. equivalently, a (named) tuple or vector of such pairs
 5. a function which will be called with a `SubDataFrame` corresponding to each group; this form should be avoided due to its poor performance
 
 In all of these cases, the function can return either a single row or multiple rows, with a single or multiple columns:
@@ -18,7 +18,7 @@ In all of these cases, the function can return either a single row or multiple r
 
 The kind of return value and the number and names of columns must be the same for all groups.
 
-As a special case, if a tuple or vector of pairs is passed (form 3 above), each function is required to return a single value or vector, which will produce each a separate column.
+As a special case, if multiple pairs or a tuple of vectors or pairs is passed (forms 3 and 4 above), each function is required to return a single value or vector, which will produce each a separate column.
 
 The name for the resulting column can be chosen either by passing a named tuple of pairs, or by returning a named tuple or a data frame. If no name is provided, it is generated automatically. For functions taking a single column (first form), the input column name is concatenated with the function name: for standard functions like `mean` this will produce columns with names like `SepalLength_mean`; for anonymous functions like `x -> sqrt(x)^e`, the produced columns will be `SepalLength_function`. For functions taking multiple columns (second form), names are `x1`, `x2`, etc.
 
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -13,7 +13,7 @@ In all of these cases, the function can return either a single row or multiple r
 - a named tuple or `DataFrameRow` produces a single row and one column per field
 - a vector produces a single column with one row per entry
 - a named tuple of vectors produces one column per field with one row per entry in the vectors
-- a `DataFrame` produces as many rows and columns as it contains; this return value should be avoided due to its poor performance
+- a `DataFrame` or a matrix produces as many rows and columns as it contains; note that returning a `DataFrame` should be avoided due to its poor performance when the number of groups is large
 
 The name for the resulting column can be chosen either by passing a named tuple of pairs, or by returning a named tuple or a data frame. If no name is provided, it is generated automatically. For functions taking a single column (first form), the input column name is concatenated with the function name: for standard functions like `mean` this will produce columns with names like `SepalLength_mean`; for anonymous functions like `x -> sqrt(x)^e`, the produced columns will be `SepalLength_function`. For functions taking multiple columns (second form), names are `x1`, `x2`, etc.
 
@@ -65,7 +65,6 @@ julia> by(iris, :Species, N = :Species => length) # Chosen column is arbitrary
 │ 1   │ setosa        │ 50    │
 │ 2   │ versicolor    │ 50    │
 │ 3   │ virginica     │ 50    │
-```
 
 julia> by(iris, :Species, N = :Species => length, mean = :PetalLength => mean) # Column for length is arbitrary
 3×3 DataFrame
@@ -87,7 +86,7 @@ julia> by(iris, :Species, [:PetalLength, :SepalLength] =>
 │ 3   │ virginica  │ 0.842744 │ 277.6   │
 ```
 
-The `by` function also support the `do` block form. However, as noted above, this form is slow and should therefore be avoided when performance matters.
+The `by` function also supports the `do` block form. However, as noted above, this form is slow and should therefore be avoided when performance matters.
 
 ```jldoctest sac
 julia> by(iris, :Species) do df
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -2,11 +2,12 @@
 
 Many data analysis tasks involve splitting a data set into groups, applying some functions to each of the groups and then combining the results. A standardized framework for handling this sort of computation is described in the paper "[The Split-Apply-Combine Strategy for Data Analysis](http://www.jstatsoft.org/v40/i01)", written by Hadley Wickham.
 
-The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split the `DataFrame` on, and (3) a specification of one or more functions to apply to each subset of the `DataFrame`. This specification can be of the following forms:
-- a `col => function` pair indicating that `function` should be called with the vector of values for column `col`, which can be a column name or index
-- a `cols => function` pair indicating that `function` should be called with a named tuple holding columns `cols`, which can be a tuple or vector of names or indices
-- a (named) tuple or vector of such pairs, producing each a single separate column
-- a function which will be called with a `SubDataFrame` corresponding to each group; this form should be avoided due to its poor performance
+The DataFrames package supports the split-apply-combine strategy through the `by` function, which is a shorthand for `groupby` followed by `map` and/or `combine`. `by` takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split the `DataFrame` on, and (3) a specification of one or more functions to apply to each subset of the `DataFrame`. This specification can be of the following forms:
+1. a `col => function` pair indicating that `function` should be called with the vector of values for column `col`, which can be a column name or index
+2. a `cols => function` pair indicating that `function` should be called with a named tuple holding columns `cols`, which can be a tuple or vector of names or indices
+3. several such pairs, either as positional arguments or as keyword arguments (mixing is not allowed), producing each a single separate column; keyword argument names are used as column names
+4. equivalently, a (named) tuple or vector of such pairs, producing each a single separate column
+5. a function which will be called with a `SubDataFrame` corresponding to each group; this form should be avoided due to its poor performance
 
 In all of these cases, the function can return either a single row or multiple rows, with a single or multiple columns:
 - a single value produces a single row and column per group
@@ -14,6 +15,10 @@ In all of these cases, the function can return either a single row or multiple r
 - a vector produces a single column with one row per entry
 - a named tuple of vectors produces one column per field with one row per entry in the vectors
 - a `DataFrame` or a matrix produces as many rows and columns as it contains; note that returning a `DataFrame` should be avoided due to its poor performance when the number of groups is large
+
+The kind of return value and the number and names of columns must be the same for all groups.
+
+As a special case, if a tuple or vector of pairs is passed (form 3 above), each function is required to return a single value or vector, which will produce each a separate column.
 
 The name for the resulting column can be chosen either by passing a named tuple of pairs, or by returning a named tuple or a data frame. If no name is provided, it is generated automatically. For functions taking a single column (first form), the input column name is concatenated with the function name: for standard functions like `mean` this will produce columns with names like `SepalLength_mean`; for anonymous functions like `x -> sqrt(x)^e`, the produced columns will be `SepalLength_function`. For functions taking multiple columns (second form), names are `x1`, `x2`, etc.
 
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -2,7 +2,20 @@
 
 Many data analysis tasks involve splitting a data set into groups, applying some functions to each of the groups and then combining the results. A standardized framework for handling this sort of computation is described in the paper "[The Split-Apply-Combine Strategy for Data Analysis](http://www.jstatsoft.org/v40/i01)", written by Hadley Wickham.
 
-The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a DataFrame, (2) one or more columns to split the DataFrame on, and (3) a function or expression to apply to each subset of the DataFrame.
+The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split the `DataFrame` on, and (3) a specification of one or more functions to apply to each subset of the `DataFrame`. This specification can be of the following forms:
+- a `col => function` pair indicating that `function` should be called with the vector of values for column `col`, which can be a column name or index
+- a `cols => function` pair indicating that `function` should be called with a named tuple holding columns `cols`, which can be a tuple or vector of names or indices
+- a (named) tuple or vector of such pairs, producing each a single separate column
+- a function which will be called with a `SubDataFrame` corresponding to each group; this form should be avoided due to its poor performance
+
+In all of these cases, the function can return either a single row or multiple rows, with a single or multiple columns:
+- a single value produces a single row and column per group
+- a named tuple or `DataFrameRow` produces a single row and one column per field
+- a vector produces a single column with one row per entry
+- a named tuple of vectors produces one column per field with one row per entry in the vectors
+- a `DataFrame` produces as many rows and columns as it contains; this return value should be avoided due to its poor performance
+
+The name for the resulting column can be chosen either by passing a named tuple of pairs, or by returning a named tuple or a data frame. If no name is provided, it is generated automatically. For functions taking a single column (first form), the input column name is concatenated with the function name: for standard functions like `mean` this will produce columns with names like `SepalLength_mean`; for anonymous functions like `x -> sqrt(x)^e`, the produced columns will be `SepalLength_function`. For functions taking multiple columns (second form), names are `x1`, `x2`, etc.
 
 We show several examples of the `by` function applied to the `iris` dataset below:
 
@@ -35,25 +48,16 @@ julia> tail(iris)
 │ 5   │ 6.2         │ 3.4        │ 5.4         │ 2.3        │ virginica     │
 │ 6   │ 5.9         │ 3.0        │ 5.1         │ 1.8        │ virginica     │
 
-julia> by(iris, :Species, size)
+julia> by(iris, :Species, :PetalLength => mean)
 3×2 DataFrame
-│ Row │ Species       │ x1      │
-│     │ Categorical…⍰ │ Tuple…  │
-├─────┼───────────────┼─────────┤
-│ 1   │ setosa        │ (50, 5) │
-│ 2   │ versicolor    │ (50, 5) │
-│ 3   │ virginica     │ (50, 5) │
+│ Row │ Species    │ PetalLength_mean │
+│     │ String⍰    │ Float64          │
+├─────┼────────────┼──────────────────┤
+│ 1   │ setosa     │ 1.462            │
+│ 2   │ versicolor │ 4.26             │
+│ 3   │ virginica  │ 5.552            │
 
-julia> by(iris, :Species, df -> mean(df.PetalLength))
-3×2 DataFrame
-│ Row │ Species       │ x1      │
-│     │ Categorical…⍰ │ Float64 │
-├─────┼───────────────┼─────────┤
-│ 1   │ setosa        │ 1.462   │
-│ 2   │ versicolor    │ 4.26    │
-│ 3   │ virginica     │ 5.552   │
-
-julia> by(iris, :Species, df -> DataFrame(N = size(df, 1)))
+julia> by(iris, :Species, (N = :Species => length,)) # Chosen column is arbitrary
 3×2 DataFrame
 │ Row │ Species       │ N     │
 │     │ Categorical…⍰ │ Int64 │
@@ -63,11 +67,31 @@ julia> by(iris, :Species, df -> DataFrame(N = size(df, 1)))
 │ 3   │ virginica     │ 50    │
 ```
 
-The `by` function also support the `do` block form:
+julia> by(iris, :Species, (N = :Species => length, mean = :PetalLength => mean)) # Chosen column is arbitrary
+3×3 DataFrame
+│ Row │ Species    │ N     │ mean    │
+│     │ String⍰    │ Int64 │ Float64 │
+├─────┼────────────┼───────┼─────────┤
+│ 1   │ setosa     │ 50    │ 1.462   │
+│ 2   │ versicolor │ 50    │ 4.26    │
+│ 3   │ virginica  │ 50    │ 5.552   │
+
+julia> by(iris, :Species, [:PetalLength, :SepalLength] =>
+              x -> (a=mean(x.PetalLength)/mean(x.SepalLength), b=sum(x.PetalLength)))
+3×3 DataFrame
+│ Row │ Species    │ a        │ b       │
+│     │ String⍰    │ Float64  │ Float64 │
+├─────┼────────────┼──────────┼─────────┤
+│ 1   │ setosa     │ 0.29205  │ 73.1    │
+│ 2   │ versicolor │ 0.717655 │ 213.0   │
+│ 3   │ virginica  │ 0.842744 │ 277.6   │
+```
+
+The `by` function also support the `do` block form. However, as noted above, this form is slow and should therefore be avoided when performance matters.
 
 ```jldoctest sac
 julia> by(iris, :Species) do df
-          DataFrame(m = mean(df.PetalLength), s² = var(df.PetalLength))
+          (m = mean(df.PetalLength), s² = var(df.PetalLength))
        end
 3×3 DataFrame
 │ Row │ Species       │ m       │ s²        │
@@ -78,7 +102,7 @@ julia> by(iris, :Species) do df
 │ 3   │ virginica     │ 5.552   │ 0.304588  │
 ```
 
-A second approach to the Split-Apply-Combine strategy is implemented in the `aggregate` function, which also takes three arguments: (1) a DataFrame, (2) one or more columns to split the DataFrame on, and (3) one or more functions that are used to compute a summary of each subset of the DataFrame. Each function is applied to each column that was not used to split the DataFrame, creating new columns of the form `$name_$function`. For named functions like `mean` this will produce columns with names like `SepalLength_mean`. For anonymous functions like `x -> sqrt(x)^e`, which Julia tracks and references by a numerical identifier e.g. `#12`, the produced columns will be `SepalLength_#12`. We show several examples of the `aggregate` function applied to the `iris` dataset below:
+A second approach to the Split-Apply-Combine strategy is implemented in the `aggregate` function, which also takes three arguments: (1) a DataFrame, (2) one or more columns to split the DataFrame on, and (3) one or more functions that are used to compute a summary of each subset of the DataFrame. Each function is applied to each column that was not used to split the DataFrame, creating new columns of the form `$name_$function` like with `by` (see above). We show several examples of the `aggregate` function applied to the `iris` dataset below:
 
 ```jldoctest sac
 julia> aggregate(iris, :Species, length)
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1430,6 +1430,12 @@ function Base.getindex(itr::DataFrameColumns{<:SubDataFrame,AbstractVector}, j)
     itr.df[:, j]
 end
 
+function Base.getindex(itr::DataFrameColumns{<:SubDataFrame,AbstractVector}, j::Int)
+    Base.depwarn("Indexing into a return value of columns on SubDataFrame will return a" *
+                 " view of column value", :getindex)
+    itr.df[:, j]
+end
+
 function Base.iterate(itr::DataFrameColumns{<:SubDataFrame,
                                             Pair{Symbol, AbstractVector}}, j=1)
     Base.depwarn("iterating over value of eachcol on SubDataFrame will return a" *
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -241,7 +241,7 @@ Transform a `GroupedDataFrame` into a `DataFrame`.
 If the last argument(s) consist(s) in one or more `cols => f` pair(s), or if
 `colname = cols => f` keyword arguments are provided, `cols` must be
 a column name or index, or a vector or tuple thereof, and `f` must be a callable.
-A tuple or a named tuple of pairs can also be provided as the first or last argument.
+A pair or a (named) tuple of pairs can also be provided as the first or last argument.
 If `cols` is a single column index, `f` is called with a `SubArray` view into that
 column for each group; else, `f` is called with a named tuple holding `SubArray`
 views into these columns.
@@ -664,7 +664,7 @@ based on grouping columns `keys`, and return a `DataFrame`.
 If the last argument(s) consist(s) in one or more `cols => f` pair(s), or if
 `colname = cols => f` keyword arguments are provided, `cols` must be
 a column name or index, or a vector or tuple thereof, and `f` must be a callable.
-A tuple or a named tuple of pairs can also be provided as the first or last argument.
+A pair or a (named) tuple of pairs can also be provided as the first or last argument.
 If `cols` is a single column index, `f` is called with a `SubArray` view into that
 column for each group; else, `f` is called with a named tuple holding `SubArray`
 views into these columns.
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -123,18 +123,18 @@ wrap(x::AbstractMatrix) =
     NamedTuple{Tuple(gennames(size(x, 2)))}(Tuple(x[:, i] for i in 1:size(x, 2)))
 wrap(x::Any) = (x1=x,)
 
-function wrap_call(f::Union{Function, Type}, gd::GroupedDataFrame,
-                   incols::Tuple, i::Integer)
+function do_call(f::Union{Function, Type}, gd::GroupedDataFrame,
+                 incols::Tuple, i::Integer)
     idx = gd.idx[gd.starts[i]:gd.ends[i]]
-    wrap(f(map(c -> view(c, idx), incols)...))
+    f(map(c -> view(c, idx), incols)...)
 end
-function wrap_call(f::Union{Function, Type}, gd::GroupedDataFrame,
-                   incols::NamedTuple, i::Integer)
+function do_call(f::Union{Function, Type}, gd::GroupedDataFrame,
+                 incols::NamedTuple, i::Integer)
     idx = gd.idx[gd.starts[i]:gd.ends[i]]
-    wrap(f(map(c -> view(c, idx), incols)))
+    f(map(c -> view(c, idx), incols))
 end
-wrap_call(f::Union{Function, Type}, gd::GroupedDataFrame, incols::Nothing, i::Integer) =
-    wrap(f(gd[i]))
+do_call(f::Union{Function, Type}, gd::GroupedDataFrame, incols::Nothing, i::Integer) =
+    f(gd[i])
 
 _nrow(df::AbstractDataFrame) = nrow(df)
 _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
@@ -149,11 +149,11 @@ _ncol(x::Union{NamedTuple, DataFrameRow}) = length(x)
 Apply a function to each group of rows and return a `GroupedDataFrame`.
 
 If the first argument is a `cols => f` pair, `cols` must be a column name or index, or
-a vector or tuple thereof, and `f` must be a function. If `cols` is a single column index,
-`f` is passed a `SubArray` view into that column for each group; else, `f` is passed a
-named tuple holding `SubArray` views into these columns.
+a vector or tuple thereof, and `f` must be a callable. If `cols` is a single column index,
+`f` is callbed with a `SubArray` view into that column for each group; else, `f` is called
+with a named tuple holding `SubArray` views into these columns.
 
-If the first argument is a function, it is passed a `SubDataFrame` view for each group,
+If the first argument is a callable, it is passed a `SubDataFrame` view for each group,
 and the returned `DataFrame` then consists of the returned rows plus the grouping columns.
 Note that this second form is much slower than the first one due to type instability.
 
@@ -168,17 +168,19 @@ determines the shape of the resulting data frame:
   with the same columns and as many rows for each group as the rows returned for that group.
 
 In all cases, the resulting `GroupedDataFrame` contains all the grouping columns in addition
-to those listed above. Note that `f` must always return the same type of object for
+to those listed above. Column names are automatically generated when necessary: for functions
+operating on a single column and returning a single value or vector, the function name is
+appended to the input colummn name; for other functions, columns are called `x1`, `x2`
+and so on.
+
+Note that `f` must always return the same type of object for
 all groups, and (if a named tuple or data frame) with the same fields or columns.
 Due to type instability, returning a single value or a named tuple is dramatically
 faster than returning a data frame.
 
-Note that due to type instability, returning a single value or a named tuple is dramatically
-faster than returning a data frame.
-
 ### Examples
 
-```julia
+```jldoctest
 julia> df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
                       b = repeat([2, 1], outer=[4]),
                       c = 1:8);
@@ -248,16 +250,17 @@ end
 
 """
     combine(gd::GroupedDataFrame)
+    combine(cols => f, gd::GroupedDataFrame)
     combine(f, gd::GroupedDataFrame)
 
 Transform a `GroupedDataFrame` into a `DataFrame`.
 
 If the first argument is a `cols => f` pair, `cols` must be a column name or index, or
-a vector or tuple thereof, and `f` must be a function. If `cols` is a single column index,
-`f` is passed a `SubArray` view into that column for each group; else, `f` is passed a
-named tuple holding `SubArray` views into these columns.
+a vector or tuple thereof, and `f` must be a callable. If `cols` is a single column index,
+`f` is called with a `SubArray` view into that column for each group; else, `f` is called
+with a named tuple holding `SubArray` views into these columns.
 
-If the first argument is a function, it is passed a `SubDataFrame` view for each group,
+If the first argument is a callable `f`, it is passed a `SubDataFrame` view for each group,
 and the returned `DataFrame` then consists of the returned rows plus the grouping columns.
 Note that this second form is much slower than the first one due to type instability.
 
@@ -272,7 +275,12 @@ determines the shape of the resulting data frame:
   with the same columns and as many rows for each group as the rows returned for that group.
 
 In all cases, the resulting data frame contains all the grouping columns in addition
-to those listed above. Note that `f` must always return the same type of object for
+to those listed above. Column names are automatically generated when necessary: for functions
+operating on a single column and returning a single value or vector, the function name is
+appended to the input colummn name; for other functions, columns are called `x1`, `x2`
+and so on.
+
+Note that `f` must always return the same type of object for
 all groups, and (if a named tuple or data frame) with the same fields or columns.
 Due to type instability, returning a single value or a named tuple is dramatically
 faster than returning a data frame.
@@ -282,7 +290,7 @@ call from which `gd` was constructed. Otherwise, ordering of rows is undefined.
 
 ### Examples
 
-```julia
+```jldoctest
 julia> df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
                       b = repeat([2, 1], outer=[4]),
                       c = 1:8);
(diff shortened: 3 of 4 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -145,27 +162,65 @@ determines the shape of the resulting data frame:
...
+function Base.map(f::Union{Function, Type}, gd::GroupedDataFrame;
+                  select::Union{Nothing, ColumnIndices}=nothing)
     if length(gd) > 0
-        idx, valscat = _combine(wrap(f(gd[1])), f, gd)
+        if select === nothing
+            incols = nothing
+        elseif select isa Symbol || select isa Integer
+            incols = (gd.parent[select],)
+        else
+            incols = Tuple(columns(gd.parent[collect(select)]))
```
bkamins: This is not optimal in general but I guess it is good enough as I would assume that this is not the costly operation relative to the whole work that has to be done and `select` is has few elements.

bkamins:
I think that keeping `select` even in the future is desirable as it reduces the burden on the compiler with wide data frames when we need only few columns.

When reading the code I realized that we can do yet another optimization. Probably for another PR, but related. We could check if a data frame is sorted on columns on which we group and if yes then use an implementation that takes advantage of the fact. I do not know if it was discussed, but even if we called `issorted` it should be relatively small overhead (optionally user could pass a keyword argument set to `true` (guaranteed sorted) `false` (guaranteed not sorted) or `nothing` (check if sorted)).

nalimilan:
Yes, `select` will only be efficient for a relatively small number of columns, but that's the typical case AFAICT. And indeed operations could be quite faster when data is sorted, we could just store the indices as a range in `GroupedDataFrame`.

Regarding the API, I realize JuliaDB supports two different syntaxes: `combine(sum, gd; select=:b)` and `combine(:b => sum, gd)`. I actually prefer the latter, since the position of arguments sounds more natural, and it's also more flexible: you can do `combine(gd, resvar = :b => sum)` to specify the name of the returned column, and `combine(gd, resvar = :b => sum, resvar2 = :c => mean)` to compute two summary statistics in one run. The equivalent syntax with `select` is  `combine((b, c) -> (resvar1=sum(b), resvar2=mean(c)), gd, select=(:b, :c))`, which is longer and repetitve. I'm not sure why JuliaDB supports both syntaxes since they completely overlap AFAICT. @piever Any ideas about that?

bkamins:
> And indeed operations could be quite faster when data is sorted, we could just store the indices as a range 

This is exactly what I mean and I suppose (have not tested) that it even justifies running `issorted` to check for it.

The `Pairs` syntax looks nice - I have not seen it before, but it seems to cover all use cases, so the question is what is the experience of JuliaDB with it 😄. Actually it would also solve the ambiguity problem as then we clearly see if we got a `Pair` or a `Callable` as an argument (even with keyword-argument approach they could be mixed).

pdeffebach:
Wow this is truly excellent. I'm excited to look this over more closely when I try and port this logic to DataFramesMeta. Thanks for the work on this!

One note about syntax, I think a lot of the DataFrames syntax is motivated by the fact that you don't want to ever do `f(:x)` because that's dishonest: `:x` is a symbol, not a column. This is the reason the API is more of less of the form `d -> f(d.x)`. While this is a noble goal, I am not concerned about the `Pairs` syntax. It maps closely with the 

```
collapse, (mean) `mean_vars' (sum) `sum_vars'
```

pattern that I like in Stata.  

I have in fact recently required `collapsing` by a large number of columns, and I will test out the ways that #1520 and this PR help with that issue, which caused memory overloads even on a slurm cluster in Stata and caused Julia to hang. But it's definitely a use case I have so I'd be interested to benchmark it.

piever:
Really nice job!

Concerning the JuliaDB API, as far as I can tell, both this approaches work in JuliaDB:

```julia
julia> IndexedTables.groupby(:mean => :SepalWidth => mean, t, :Species);

julia> IndexedTables.groupby(mean, t, :Species, select = :SepalWidth);
```

I think there was a API rewrite at some point and it was decided that most operations (`groupby`, `join`, `summarize`, `unstack`) should have a `by` and a `select` (defaulting to primary columns the `by` and to non primary columns the `select`). One motivation for the sugar in JuliaDBMeta is, as you pointed out, that this syntax is a bit clumsy when selecting more than one column.

The "pair" syntax is a consequence of the [selection](https://juliadb.org/latest/api/selection.html) API, so basically instead of passing a function you are passing a combination of a selection and a function.

In terms of syntactic sugar, JuliaDBMeta provides both a `@groupby` that would detect the `select` condition and the anonymous function:

```
@groupby t :Species (m = mean(:SepalLength), s = std(:SepalWidth))
```

and a `@=>` macro to generate the pair and the anonymous function from an expression with symbols. (as can be used in combination with `groupby`, `select`, etc...). Note also that `summarize` (DataFrames `aggregate`) is just syntactic sugar to generate the "pair" expression to feed to `groupby`: so an interesting idea for DataFrames would be to focus on this pair object and create some utility function to easily generate commonly useful patterns.

There is a difference however between JuliaDB and DataFrames here that I am only realizing now, in that passing a select with a tuple of symbols in JuliaDB is just an optimization, i.e. both of the following work:

```julia
julia> IndexedTables.groupby(v -> (m = mean(column(v, :SepalLength)), s = std(column(v, :SepalWidth))), t, :Species)

julia> IndexedTables.groupby(v -> (m = mean(column(v, :SepalLength)), s = std(column(v, :SepalWidth))), t, :Species, select = (:SepalLength, :SepalWidth))
```

Whereas from what I understand, here the anonymous function would take as many arguments as you pass with `select`? I guess this is somewhat inevitable for technical reasons (unlike in IndexedTables, here creating the `DataFrames` would recreate the type instability). If that's the case I find it somewhat awkward that a keyword argument should specify the number of arguments to your function and I would prefer a:

```julia
by((:SepalLength, :SepalWidth) => (x, y) -> f(x, y), t, :Species)
```

bkamins:
Just to expand on what the function should accept. After thinking of it I get to prefer `NamedTuple` of vectors over positional arguments even more. The reason is that you can easily alter this `NamedTuple` e.g. with `merge` and then return a new `NamedTuple` (e.g. with an additional column or a replaced column) to get an efficient return value to work with. And inside the function `NamedTuple` is easy enough to work with I think. We would have to weigh the performance cost of creating a `NamedTuple` vs splatting a `Tuple` (which would be relevant if we have many small groups).

piever:
> Just to expand on what the function should accept. After thinking of it I get to prefer `NamedTuple` of vectors over positional arguments even more. 

I'm also a bit sceptical of the "many arguments" approach. The `NamedTuple` of vectors is actually nice enough in that it can be converted easily to `DataFrames` by the user, one can still extract columns with `.` syntax or also rows using `Tables.rows(t)`.

bkamins:
> In an ideal world I imagine one could have two versions of `DataFrame`

Actually this is what I want to put on a table very soon (but first I have to get `getindex` and `setindex!` right, as it is crucial to have a stable API first before adding this type). My current thinking is that a minimal such structure can just be a wrapper around `NamedTuple` of vectors (then in this wrapper we can also keep e.g. metadata, which is another big thing that is in the works, but did not get to be merged yet).

nalimilan:
OK. Let's use the pair approach rather than `select` then. But I must say I'm not a fan of the `:newcol => :oldcol => fun` syntax, as it uses the same operator for two very different meanings. This API has been designed before named tuples existed, but we can now do `(newcol = :oldcol => fun,)`. Or we could use keyword arguments (which are now fast) for that, and put these pairs as the last arguments. That way one could write easily `by(df, :key, mean1 = :col1 => mean, sum2 = :col2 => sum`, or just `by(df, :key, :col1 => mean, :col2 => sum)` to get the default names (which would probably be something like `:mean_col1` and `:sum_col2`).

Regarding what the function should be passed, I'm not sure `NamedTuple` is the best approach:
- It's not very convenient when a single column is selected, which is the most common case: `by(df, :key, :col1 => mean)` becomes `by(df, :key, :col1 => x -> mean(x.col1))`, and we lose the ability to generate a nice column name for the function name in simple cases like this.
- It forces recompiling `_combine!` for each new variable name, which is a waste since it should be frequent to call the same function on different columns of the same type.
- Finally it's not too hard to create a named tuple manually if needed, and if that's not convenient enough we could add `select` and pass a `NamedTuple` rather than a `SubDataFrame` to the user-provided function. This would be more consistent with what JuliaDB does.

bkamins:
OK - let us move forward with what you propose. When it is implemented I will checkout the PR and test it to see how it goes 😄. Thanks

piever:
> Regarding what the function should be passed, I'm not sure `NamedTuple` is the best approach.

I think I understand the design a bit better now, and I believe there really are two separate concerns. There is the concern of a "typed selection" API, i.e. a way to select a fully typed subset of columns, which, to be consistent with JuliaDB, could be done as follows:

```julia
select(x, :SepalLength) # select with a symbol -> get a vector
select(x, (:SepalLength, :SepalWidth) # select with a Tuple -> get the NamedTuple of columns, or "type stable DataFrame"
```

And ideally many functions could accept `select` as a keyword argument to turn the argument into a type stable objects (even filter or map), for example:

```julia
map(f, df, select = ...) = map(f, DataFrames.select(df, select))
```

And then there is the issue of describing functions that operate on a few columns, which is orthogonal to this. So one could have a default:

```julia
apply(f, df) = f(df)
used_cols(f, df) = names(df)
```

and it could have many useful specializations. For example `f` could be `:SepalLength => mean`, or `:SepalLength => (mean, std)`, or `SepalLength => (new_name = mean,)`, or `(s_m = :SepalLength => mean, p_t = :PetalLength => std)` and so on (it's important to decide exactly what exactly is supported). In this scenario it's true that it's probably handier to use the multi-argument approach.

Then, for each of this overloads to `apply(f, df)` one could potentially also overload `used_cols(f, df)` if it's clear from `f` which columns it will use, in which case `by` can be optimized.

pdeffebach:
> way one could write easily `by(df, :key, mean1 = :col1 => mean, sum2 = :col2 => sum`, or just `by(df, :key, :col1 => mean, :col2 => sum)` to get the default names (which would probably be something like `:mean_col1` and `:sum_col2`).

I'm not a fan of this because `mean1` isn't a symbol and thus its difficult to put these operations in functions. DataFrames has done well so far to avoid non-standard evaluation, and it'd be nice to keep it that way. 

I also want to note that I find it very common to perform the same action on a variety of columns, something like 

```
by(df, :id, [:x1, :x2, :x3] => mean, [:x1, :x2, :x4] => sum)
```

Would be ideal.

bkamins:
A possibility would be that we mimick `getindex` on data frame behavior here. If we write e.g. `:a => mean` we pass a vector to `mean` (as in `df.a` getindex, LHS could be a `Symbol` or `Integer` other than `Bool`), if we write e.g. `1:2 => fun` then we pass a `NamedTuple` (in general anything else that a `Symbol` or `Integer` other than `Bool` would produce a `NamedTuple`) - exactly how `df[1:2]`. This would be unambiguous I think.

@nalimilan how do you see this?

bkamins:
Also we should provide users with ability to know source column names (in case they wanted to dynamically generate column names in return value) and without a named tuple (or some equivalent, e.g. splatted keyword arguments) how do you plan to allow for this?

nalimilan:
Clever. Actually it looks like that's what JuliaDB [does](http://juliadb.org/latest/api/selection.html). I need to think about it, but that's appealing, and it would replace `select` completely.

> I'm not a fan of this because `mean1` isn't a symbol and thus its difficult to put these operations in functions. DataFrames has done well so far to avoid non-standard evaluation, and it'd be nice to keep it that way.

You can create a keyword argument call from a symbol very easily, e.g. `x = :a; f(; x=>1)`.

> I also want to note that I find it very common to perform the same action on a variety of columns, something like
> ```
> by(df, :id, [:x1, :x2, :x3] => mean, [:x1, :x2, :x4] => sum)
> ```
> 
> Would be ideal. (though this doesn't address the issue of column names)

Hmm, that interpretation of this syntax would conflict with other proposals and would prevent specifying functions which take several argument. I think you can do that with `aggregate`, right? Or you can also define a broadcasting function like `[:x1, :x2, :x3] => x -> mean.(x)`.

pdeffebach:
> I think you can do that with aggregate, right? 

It wouldn't work if you have different subsets of columns you wanted to act on. Maybe a column has `mean` defined for it but not some other function. However `[:x1, :x2, :x3] => x -> mean.(x)` seems pretty solid. 



> You can create a keyword argument call from a symbol very easily, e.g. `x = :a; f(; x=>1)`

This is really interesting! I will have to look into how this could be done with DataFramesMeta as well

nalimilan:
OK, I've pushed a commit to switch to the `Pair` approach. It still needs tests.

The pair needs to be passed as the first argument, even if the `do` block syntax doesn't apply here. This prevents using keyword arguments to specify the output column names. One could use a named tuple for that (currently not supported), or we could support both syntaxes (which `by` already does). The alternative is to require the pair to be at the end.

nalimilan:
I think this is getting ready for a review. In particular, please have a look at the API.

src/deprecated.jl
```diff
@@ -1430,6 +1430,12 @@ function Base.getindex(itr::DataFrameColumns{<:SubDataFrame,AbstractVector}, j)
     itr.df[:, j]
 end
 
+function Base.getindex(itr::DataFrameColumns{<:SubDataFrame,AbstractVector}, j::Int)
```
bkamins: A side comment.
In the past we did not have `DataFrameColumns{<:SubDataFrame,AbstractVector}`, as we only had `DataFrameColumns{<:SubDataFrame, Pair{Symbol, AbstractVector}}` (under another name).

Thus `DataFrameColumns{<:SubDataFrame,AbstractVector}` can safely return a view - consistently with the target behavior.

But if for some reason you prefer to add it it is OK. I will remove it in `getindex` cleanup PR later.

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,111 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
...
 """
-    map(f::Union{Function, Type}, gd::GroupedDataFrame)
+    map(cols => f, gd::GroupedDataFrame)
+    map(f, gd::GroupedDataFrame)
 
 Apply a function to each group of rows and return a `GroupedDataFrame`.
 
-For each group in `gd`, `f` is passed a `SubDataFrame` view holding the corresponding rows.
+If the first argument is a `cols => f` pair, `cols` must be a column name or index, or
+a vector or tuple thereof, and `f` must be a function. If `cols` is a single column index,
```
bkamins: A corner case: can `cols` be a vector of `Bool` to indicate columns?

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,111 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
...
+`f` is passed a `SubArray` view into that column for each group; else, `f` is passed a
+named tuple holding `SubArray` views into these columns.
+
+If the first argument is a function, it is passed a `SubDataFrame` view for each group,
```
bkamins: it would be more precise to say "callable" than `"function".

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,111 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
...
 
 In all cases, the resulting `GroupedDataFrame` contains all the grouping columns in addition
 to those listed above. Note that `f` must always return the same type of object for
 all groups, and (if a named tuple or data frame) with the same fields or columns.
-Returning a single value or a named tuple is significantly faster than
-returning a vector or a data frame.
+Due to type instability, returning a single value or a named tuple is dramatically
+faster than returning a data frame.
+
+Note that due to type instability, returning a single value or a named tuple is dramatically
```
bkamins: duplicate (almost) of the sentence above.

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,111 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
...
 
 ### Examples
 
 ```julia
-df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
-               b = repeat([2, 1], outer=[4]),
-               c = randn(8))
-gd = groupby(df, :a)
-map(d -> sum(skipmissing(d[:c])), gd)
+julia> df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),
```
bkamins: maybe switch it to `jldoctest`?

src/groupeddataframe/grouping.jl
```diff
@@ -245,39 +331,64 @@ end
...
-    initialcols = ntuple(i -> Tables.allocatecolumn(typeof(first[i]), n), m)
-    cols = _combine!(first, initialcols, idx, 1, 1, f, gd, tuple(propertynames(first)...))
-    valscat = DataFrame(collect(cols), collect(propertynames(first)))
-    idx, valscat
+function _combine(f::Union{Function, Type, Pair}, gd::GroupedDataFrame)
+    if f isa Pair{<:Union{Symbol,Integer}}
+        incols = (gd.parent[first(f)],)
+        fun = last(f)
+    elseif f isa Pair
+        incols = NamedTuple{Tuple(first(f))}(columns(gd.parent[collect(first(f))]))
```
bkamins: here I see that we will accept vector of `Bool` as a column selector.

src/groupeddataframe/grouping.jl
```diff
@@ -234,9 +320,9 @@ combine(d -> sum(skipmissing(d[:c])), gd)
 of `combine(map(f, groupby(df, cols)))`.
 
 """
-function combine(f::Function, gd::GroupedDataFrame)
+function combine(f::Union{Function, Type, Pair}, gd::GroupedDataFrame)
```
bkamins: A general comment (although I know I have changed it earlier to `Union{Function, Type}`). Maybe we should remove type restriction on `f` entirely in all those methods - this is what Base does. Later we can only check if `f` is a `Pair`. In this way we would allow users to pass functors as `f` (which is currently disallowed).

src/groupeddataframe/grouping.jl
```diff
@@ -190,42 +248,70 @@ end
...
-- A named tuple or a `DataFrameRow` gives a data frame with one column for each field
-  and one row per group.
+- A named tuple of single values or a `DataFrameRow` gives a data frame with one column
+  for each field and one row per group.
 - A vector gives a data frame with a single column and as many rows
   for each group as the length of the returned vector for that group.
-- A data frame or a matrix gives a data frame with the same columns and as many rows
-  for each group as the rows returned for that group.
+- A data frame, a named tuple of vectors or a matrix gives a data frame
+  with the same columns and as many rows for each group as the rows returned for that group.
```
bkamins: In the cases where we generate column names we should somewhere in the documentation describe the rules for generating them.

nalimilan:
Yes, it should be easy to add convenience methods to handle `by([:b => mean, :c => mean, :b => length], df, :a)`.

> I think it is OK to move `Pairs` argument to the end to allow passing output column name in a keyword argument. However, what will the output column name mean in the case when function returns a named tuple not a vector? (if I understood your idea correctly)

I've wondered about that too. What IndexedTables currently does is to return the named tuple as-is in that case (i.e. use it as a value). But they don't support creating multiple columns from a single function by returning a named tuple like we do, so that's quite different. I guess it doesn't really matter what we do. We could simply throw an error.

I'm still hesitant as to whether we should put these arguments first or last. Oour docs for `by` currently recommend putting the function last, putting it first is presented as a secondary method to support `do`. And indeed it feels more logical to say what to group and how before giving the functions. Another argument in favor of putting them last is that one could write `by(df, :key, x2 = :x => mean)` instead of `by((x2 = :x => mean,), df, :key)`.

OTOH conflicts could arise with "real" keyword arguments (currently just `sort`, so that's not the end of the world). Also putting these last would be inconsistent with `map` and `combine`, but that's not a big issue given that these are quite different functions anyway.

src/deprecated.jl
nalimilan: I just added this to fix an ambiguity error I was getting. I guess I should just remove the `depwarn` call from it.

bkamins:
I think that keeping `by` and `combine` and `map` consistent is important (it would be great if users had to learn only one semantic convention).

Also I have just noticed that `groupby` has also `skipmissing` keyword argument and `by` does not have one. I guess we should make this consistent?

In the end I think that the best compromise is to say that user can use the form you have proposed or (exclusively so without mixing) the form with keyword arguments - for functions returning single columns as a convenience (i.e. we have one way of writing things, that is fully general, and then the argument goes first as we have it now and a special syntax for common simple case - and then you can use keyword arguments). Also in that case we do not strictly need the support for `by([:b => mean, :c => mean, :b => length], df, :a)` as it can be mostly handled by the keyword argument approach.

As for `sort` and `skipmissing` keyword arguments - the user can always do `groupby` and then `combine` which is easy enough (and personally I typically prefer to make those two steps as separate operations for clarity - like in dplyr), so we can reserve those two kwargs for different purposes than column names in `by`.

src/deprecated.jl
bkamins: Ah - in this case it can stay or would write `itr.df[j]` below and remove a depwarn (as then depwarn will be triggered by `getindex` in `itr.df[j]` call).

bkamins:
To sum up my understanding of the API those would be the accepted signatures:
* `map(::Any, ::GroupedDataFrame)`; where the first argument must be callable (function, type or functor); we pass `SubDataFrame` to the first argument;
* `map(::Pair, ::GroupedDataFrame)`; follows the rules for `Pairs` you @nalimilan  proposed; including auto-generation of column names;
* `map(::GroupedDataFrame; kwargs)`; every keyword argument must be a `Pair` returning a vector or scalar; we use keyword argument name as column name;
* `combine(::GroupedDataFrame)`; produces a data frame;
* `combine(::Any, ::GroupedDataFrame)`; the same rules as for `map`;
* `combine(::Pair, ::GroupedDataFrame)`; the same rules as for `map`;
* `combine(::GroupedDataFrame; kwargs)`; the same rules as for `map`;
* `by(::Any, ::GroupedDataFrame, cols; sort, skipmissing)`; the same rules as for `map`;
* `by(::Pair, ::GroupedDataFrame, cols; sort, skipmissing)`; the same rules as for `map`;
* `by(::GroupedDataFrame, cols; kwargs)`; the same rules as for `map`; treat `sort` and `skipmissing` as in other `by` methods (this is what I have proposed above - but probably some other rules are thinkable, e.g. check if `sort` is `Bool` or a `Pair` and decide what to do based on this, but this is messy)
* `by(::GroupedDataFrame, cols, ::Function; sort, skipmissing)`; I would deprecate this method to simplify API and keep consistency between `map`, `combine` and `by`;

(not all of this has to be implemented in one shot - but I like to have a roadmap for the design)

Also a small design issue we had in `aggregate` is to decide if we want to set `makeunique` to `true` (this is what we do in `aggregate`) or `false` when we produce the final data frame. I think that here the risk of duplicate column names is low, so this is a secondary issue.

Another design issue is to keep in mind is https://github.com/JuliaData/DataFrames.jl/issues/1460 (but this can be kept for later I guess).

piever:
> it would be nice to allow for something like:
>
> `by([:b => mean, :c => mean, :b => length], df, :a)`

IMO it'd be nice to follow the JulaDB idea of recursively defining this object. Meaning a pair `colname => func` would be a selection and return the function applied to the column. If there is no name (meaning we don't return a named tuple), name is given automatically after the function. A tuple of selection is a selection whose result is the `merge` of the results of the selections, so `by((:b => mean, :c => mean), df, :a)`. Finally a `NamedTuple` of selections is the merge of the results of the selections with names from the named tuple, so `(meanb = :b => mean, meanc = :c => mean)` for example is possible.

bkamins:
I have updated what I think after considering how this will work with the DataFramesMeta.jl package. Also we should consider (probably in the DataFramesMeta.jl package) how we handle `Symbol` arguments as normally they would get rewritten here by the macro system - so we have to make sure to update in sync the relevant macros there.

src/deprecated.jl
nalimilan: OK let's keep it that way, it's simpler if it's identical to the method above.

src/groupeddataframe/grouping.jl
nalimilan: See other thread. I'd rather not mention this explicitly here since that's relatively unusual and it would make the sentence really complex. It would be interesting to define and export a type alias and use it in signatures, like `ColumnIndex` but including vectors (JuliaDB uses `::Selection`, but doesn't export it).

nalimilan:
> Actually having `by(::AbstractDataFrame, cols, ::Function; sort, skipmissing)` (and other methods where `GroupedDataFrame` or `AbstractDataFrame` is the first argument) might be desirable as it improves interoperation with the DataFramesMeta.jl package `@linq` macro.

I guess this applies to `map` and `combine` then? Also the `::Function` method should really be discouraged, so if we support it in one use case we should also support `::Pair`. Maybe we should define both orders for all arguments.

> Also I have just noticed that `groupby` has also `skipmissing` keyword argument and `by` does not have one. I guess we should make this consistent?

Yes, probably. I'm surprised it doesn't.


> IMO it'd be nice to follow the JulaDB idea of recursively defining this object. Meaning a pair `colname => func` would be a selection and return the function applied to the column. If there is no name (meaning we don't return a named tuple), name is given automatically after the function. A tuple of selection is a selection whose result is the `merge` of the results of the selections, so `by((:b => mean, :c => mean), df, :a)`. Finally a `NamedTuple` of selections is the merge of the results of the selections with names from the named tuple, so `(meanb = :b => mean, meanc = :c => mean)` for example is possible.

@piever Isn't the first part what the PR does? We just need to support the second part, i.e. providing column names explicitly.

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,113 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
...
 
 In all cases, the resulting `GroupedDataFrame` contains all the grouping columns in addition
-to those listed above. Note that `f` must always return the same type of object for
+to those listed above. Column names are automatically generated when necessary: for functions
```
nalimilan: I'm not totally satisfied with using `x1` as the column name for functions taking several arguments. We could simply concatenate the names of all inputs with that of the function. In practice it's unlikely to get very long. But often functions taking multiple arguments will be anonymous functions, so suffix will be `_function`, which isn't very useful. At least that would be consistent. Thoughts?

src/groupeddataframe/grouping.jl
bkamins: OK - but let us at least internally be clear if we want to support `AbstractVector{Bool}` (see other comment where the implementation is provided).

bkamins:
> Maybe we should define both orders for all arguments.

This is what I meant + after we merge it we should open a PR for DataFramesMeta.jl to sync the behavior.

src/groupeddataframe/grouping.jl
bkamins: I think we can go for join of column names and function name although it is clumsy indeed.

I would assume that in this case anyway user will return a `NamedTuple` or `DataFrame` or something similar.

If someone wants to return only a vector probably one will use:
```
by(df, cols; colname = selector => function)
```
pattern as it is more convenient.

So the case when we have to generate a column name should be rare in practice (if we appropriately document the functionality to teach people the right patterns - which I intend to do anyway as we discussed on Discourse)

nalimilan:
I've added support for (named) tuples of pairs. This gives a really convenient syntax (see added docs). It's (currently) about 4 times slower than the single-pair form in the benchmark above (for a single function), but that's already much faster than in the last release. Not sure whether it can be improved.

I've made it quite restrictive for now: an error is thrown if the multiple-pair form is used and the function returns an object which would create multiple columns. We could imagine merging these columns, but that's complex and can be improved later.

I've also changed named tuples of vectors to be interpreted as a lightweight data frame (i.e. with one row per entry in the vectors), while before (only in master, not released) the vectors were taken as single values in cells. This sounds more useful and more consistent with the treatment of single scalars vs. single vectors, and it's necessary to allow the multiple-function form to support vector return values.

nalimilan:
More remarks?

An additional question: should we treat tuples like named tuples, generating automatic names for columns, or should we continue to treat them as scalars? It's not really related to this PR, but it makes sense to discuss this now.

bkamins:
I will have a detailed look at it in the comming days and give an approve is all is OK 😄.
Regarding tuples - I would treat them as scalars. I do not see much value in additional rule to the API here (in general - I would discourage use of auto-generated names in the API except in very simple cases).

If we are at it - I will check it when reviewing - but probably @nalimilan knows this so I can get this information faster: how do we handle `makeunique` argument for `DataFrame` in this PR (i.e. what happens if: a) a duplicate name is supplied by a user, b) a duplicate name is auto-generated).

bkamins:
`makeunique=false` is OK. The other corner case is when auto-generated name clashes with grouping name - but when this happens throwing an error is also OK.

The only problematic case is when user returns key column in the result - as discussed in https://github.com/JuliaData/DataFrames.jl/issues/1460. If we want to handle this I think it would be good to include this functionality in this PR. What do you think? If we do it then we should check that values in columns with duplicate names are the same.

docs/src/man/split_apply_combine.md
```diff
@@ -2,7 +2,20 @@
 
 Many data analysis tasks involve splitting a data set into groups, applying some functions to each of the groups and then combining the results. A standardized framework for handling this sort of computation is described in the paper "[The Split-Apply-Combine Strategy for Data Analysis](http://www.jstatsoft.org/v40/i01)", written by Hadley Wickham.
 
-The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a DataFrame, (2) one or more columns to split the DataFrame on, and (3) a function or expression to apply to each subset of the DataFrame.
+The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split the `DataFrame` on, and (3) a specification of one or more functions to apply to each subset of the `DataFrame`. This specification can be of the following forms:
```
bkamins: maybe add that `by` is a convinience for  `groupby` followed by `combine` as those two are elementary operations?

docs/src/man/split_apply_combine.md
```diff
@@ -2,7 +2,20 @@
...
+- a `col => function` pair indicating that `function` should be called with the vector of values for column `col`, which can be a column name or index
+- a `cols => function` pair indicating that `function` should be called with a named tuple holding columns `cols`, which can be a tuple or vector of names or indices
+- a (named) tuple or vector of such pairs, producing each a single separate column
+- a function which will be called with a `SubDataFrame` corresponding to each group; this form should be avoided due to its poor performance
+
+In all of these cases, the function can return either a single row or multiple rows, with a single or multiple columns:
```
bkamins: It would be good to be precise in which cases only a single column is allowed.

If a single column is allowed then I understand that any of the return-value options is valid as long as it produces one column. Right?

docs/src/man/split_apply_combine.md
```diff
@@ -2,7 +2,20 @@
...
+- a single value produces a single row and column per group
+- a named tuple or `DataFrameRow` produces a single row and one column per field
+- a vector produces a single column with one row per entry
+- a named tuple of vectors produces one column per field with one row per entry in the vectors
+- a `DataFrame` produces as many rows and columns as it contains; this return value should be avoided due to its poor performance
+
```
bkamins: maybe add a sentence that the return values should be consistent across groups (i.e. the same type and the same size in terms of columns and rows).

docs/src/man/split_apply_combine.md
```diff
@@ -2,7 +2,20 @@
 
 Many data analysis tasks involve splitting a data set into groups, applying some functions to each of the groups and then combining the results. A standardized framework for handling this sort of computation is described in the paper "[The Split-Apply-Combine Strategy for Data Analysis](http://www.jstatsoft.org/v40/i01)", written by Hadley Wickham.
 
-The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a DataFrame, (2) one or more columns to split the DataFrame on, and (3) a function or expression to apply to each subset of the DataFrame.
+The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split the `DataFrame` on, and (3) a specification of one or more functions to apply to each subset of the `DataFrame`. This specification can be of the following forms:
+- a `col => function` pair indicating that `function` should be called with the vector of values for column `col`, which can be a column name or index
```
bkamins: maybe a numbered list as later you reference to the number of the form.

docs/src/man/split_apply_combine.md
```diff
@@ -35,25 +48,16 @@ julia> tail(iris)
...
-
-julia> by(iris, :Species, df -> DataFrame(N = size(df, 1)))
+│ Row │ Species    │ PetalLength_mean │
+│     │ String⍰    │ Float64          │
+├─────┼────────────┼──────────────────┤
+│ 1   │ setosa     │ 1.462            │
+│ 2   │ versicolor │ 4.26             │
+│ 3   │ virginica  │ 5.552            │
+
+julia> by(iris, :Species, N = :Species => length) # Chosen column is arbitrary
```
bkamins: This form (i.e. keyword argument) is not covered by the rules listed above as this is not the same as passing a named tuple.

docs/src/man/split_apply_combine.md
```diff
@@ -63,11 +67,31 @@ julia> by(iris, :Species, df -> DataFrame(N = size(df, 1)))
 │ 3   │ virginica     │ 50    │
 ```
 
-The `by` function also support the `do` block form:
+julia> by(iris, :Species, N = :Species => length, mean = :PetalLength => mean) # Column for length is arbitrary
```
bkamins: same here - actually you can pass more than one keyword argument - this is not mentioned (and what is allowed to be passed multiple times). And I think it should be clarified if mixing styles is allowed.

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,111 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
 # Wrapping automatically adds column names when the value returned
 # by the user-provided function lacks them
 wrap(x::Union{AbstractDataFrame, NamedTuple, DataFrameRow}) = x
-wrap(x::AbstractMatrix) = convert(DataFrame, x)
-wrap(x::AbstractVector) = DataFrame(x1 = x)
-wrap(x::Any) = (x1 = x,)
+wrap(x::AbstractMatrix) =
+    NamedTuple{Tuple(gennames(size(x, 2)))}(Tuple(x[:, i] for i in 1:size(x, 2)))
```
bkamins: do we want to create a copy or a view here? I am referring to `x[:, i]`.

docs/src/man/split_apply_combine.md
bkamins: In this case (single column) - what do we do if there is a name clash, i.e. the returned value returns a single column e.g. as a `DataFrame` with one column having a different name than indicated in the call.

docs/src/man/split_apply_combine.md
```diff
@@ -2,7 +2,20 @@
...
 
-The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a DataFrame, (2) one or more columns to split the DataFrame on, and (3) a function or expression to apply to each subset of the DataFrame.
+The DataFrames package supports the Split-Apply-Combine strategy through the `by` function, which takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split the `DataFrame` on, and (3) a specification of one or more functions to apply to each subset of the `DataFrame`. This specification can be of the following forms:
+- a `col => function` pair indicating that `function` should be called with the vector of values for column `col`, which can be a column name or index
+- a `cols => function` pair indicating that `function` should be called with a named tuple holding columns `cols`, which can be a tuple or vector of names or indices
```
bkamins: is this so? In the code below it seems that the function is called with splatted positional arguments not with a named tuple.

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,111 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
...
+wrap(x::Any) = (x1=x,)
+
+function do_call(f::Any, gd::GroupedDataFrame, incols::Tuple, i::Integer)
```
bkamins: here it seems to me that we are splatting columns in the call to `f`.

src/groupeddataframe/grouping.jl
```diff
@@ -190,42 +248,76 @@ end
 
 """
     combine(gd::GroupedDataFrame)
-    combine(f::Function, gd::GroupedDataFrame)
+    combine(cols => f, gd::GroupedDataFrame)
```
bkamins: Why you do not want to provide the same API for `combine` as for `by`?

Actually using `groupby` and then `combine` is a common pattern IMHO. As you can perform grouping once, and then call `combine` many times later.

I can live with `map` not having the full API, as map is probably a bit less useful than `combine`, but we also could make them consistent.

If you agree with this approach we can decide to do it later as this will be non-breaking.

src/groupeddataframe/grouping.jl
```diff
@@ -245,39 +337,110 @@ end
...
+    # Use temporary names for columns, and rename after the fact where appropriate
+    tmpnms = Tuple(gennames(length(f)))::NTuple{length(f), Symbol}
+    fun = function(x)
+        tup = map(f2) do p
+            local incols = first(p)
+            res = incols isa Tuple ?
+                  last(p)(NamedTuple{incols}(map(c -> x[c], incols))) :
+                  last(p)(x[incols])
+            if res isa Union{AbstractDataFrame, NamedTuple, DataFrameRow, AbstractVector, AbstractMatrix}
+                throw(ArgumentError("a single value result is required when passing a vector or tuple " *
```
bkamins: Why we do not allow returning `AbstractVector`?

I can agree that we do not allow other forms (and this answers my question above).

src/groupeddataframe/grouping.jl
```diff
@@ -234,9 +326,9 @@ combine(d -> sum(skipmissing(d[:c])), gd)
 of `combine(map(f, groupby(df, cols)))`.
 
 """
-function combine(f::Function, gd::GroupedDataFrame)
+function combine(f::Any, gd::GroupedDataFrame)
     if length(gd) > 0
-        idx, valscat = _combine(wrap(f(gd[1])), f, gd)
+        idx, valscat = _combine(f, gd)
```
bkamins: Actually you do `makeunique=true` here - at least in `combine`.

src/groupeddataframe/grouping.jl
```diff
@@ -122,50 +119,111 @@ _names(gd::GroupedDataFrame) = _names(gd.parent)
...
 ### See also
 
 `combine(f, gd)` returns a `DataFrame` rather than a `GroupedDataFrame`
 
 """
-function Base.map(f::Union{Function, Type}, gd::GroupedDataFrame)
+function Base.map(f::Any, gd::GroupedDataFrame)
     if length(gd) > 0
-        idx, valscat = _combine(wrap(f(gd[1])), f, gd)
+        idx, valscat = _combine(f, gd)
```
bkamins: here `makeunique=false` but in `combine` it is `true` which is inconsistent.

src/groupeddataframe/grouping.jl
```diff
@@ -245,39 +337,110 @@ end
...
+        fun = last(f)
+    elseif f isa Pair
+        df = gd.parent[collect(first(f))]
+        incols = NamedTuple{Tuple(names(df))}(columns(df))
+        fun = last(f)
+    else
+        incols = nothing
+        fun = f
+    end
+    firstres = do_call(fun, gd, incols, 1)
```
bkamins: OK - I see here that you use the `Tuple` version of `do_call` only when tuple has one element - right?

src/groupeddataframe/grouping.jl
```diff
@@ -245,39 +337,110 @@ end
...
+    if first isa AbstractDataFrame
+        n = 0
+        eltys = eltypes(first)
+    elseif first isa NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}
+        n = 0
+        eltys = map(eltype, first)
+    elseif first isa DataFrameRow
+        n = length(gd)
+        eltys = eltypes(parent(first))
+    else
```
bkamins: maybe move `NamedTuple` condition as the last condition to make it clear when `else` is posible (I understand it is only a named tuple of non-vectors).

Here also a question is what if a function returns a NamedTuple mixing vectors and scalars (which is allowed in DataFrame constructors).

src/groupeddataframe/grouping.jl
```diff
@@ -443,71 +612,140 @@ colwise(fns::Union{AbstractVector, Tuple}, d::AbstractDataFrame) = [f(d[i]) for
 colwise(f, gd::GroupedDataFrame) = [colwise(f, g) for g in gd]
 
 """
-Split-apply-combine in one step; apply `f` to each grouping in `d`
-based on columns `col`
+    by(d::AbstractDataFrame, keys, cols => f...; sort::Bool = false)
```
bkamins: do you want to add this drop missing keyword argument that is present in `groupby` in a separate PR?

src/groupeddataframe/grouping.jl
```diff
@@ -443,71 +612,140 @@ colwise(fns::Union{AbstractVector, Tuple}, d::AbstractDataFrame) = [f(d[i]) for
...
+    by(d::AbstractDataFrame, keys; (colname = cols => f)..., sort::Bool = false)
+    by(d::AbstractDataFrame, keys, f; sort::Bool = false)
```
bkamins: missing `by` with `f` at front.

src/groupeddataframe/grouping.jl
```diff
@@ -190,42 +248,76 @@ end
...
-    combine(f::Function, gd::GroupedDataFrame)
+    combine(cols => f, gd::GroupedDataFrame)
+    combine(f, gd::GroupedDataFrame)
 
 Transform a `GroupedDataFrame` into a `DataFrame`.
-If a function `f` is provided, it is called for each group in `gd` with a `SubDataFrame` view
-holding the corresponding rows, and the returned `DataFrame` then consists of the returned rows
-plus the grouping columns.
+
+If the first argument is a `cols => f` pair, `cols` must be a column name or index, or
```
bkamins: If I understand it correctly `combine` can also ingest a named tuple as a first argument. Right?

src/groupeddataframe/grouping.jl
```diff
@@ -443,71 +612,140 @@ colwise(fns::Union{AbstractVector, Tuple}, d::AbstractDataFrame) = [f(d[i]) for
...
     combine(f, groupby(d, cols, sort = sort))
-by(f::Function, d::AbstractDataFrame, cols; sort::Bool = false) =
+by(f::Any, d::AbstractDataFrame, cols::Any; sort::Bool = false) =
     by(d, cols, f, sort = sort)
+by(d::AbstractDataFrame, cols::Any, f::Pair; sort::Bool = false) =
+    combine(f, groupby(d, cols, sort = sort))
+by(d::AbstractDataFrame, cols::Any, f::Pair...; sort::Bool = false) =
+    combine(f, groupby(d, cols, sort = sort))
+by(d::AbstractDataFrame, cols::Any; sort::Bool = false, f...) =
+    combine(values(f), groupby(d, cols, sort = sort))
```
bkamins: In this section of code I can see why you have implemented `combine` differently than `by` but, as stated above, I feel it would be better if all the funtions had the same user-facing API.

bkamins:
I have went through this. A massive work - thank you @nalimilan. The major concern I had is API inconsitency between `by` and `combine` (`map` could be considered different, but it also could be consistent IMHO).

docs/src/man/split_apply_combine.md
nalimilan: Right. Currently the only way to indicate a name in the call is to pass a named tuple of pairs, and in that case we only allow returning a single value or vector, so that there are no names which could clash. I've added a mention of this.

src/groupeddataframe/grouping.jl
nalimilan: Good catch. That's actually a legacy from the original design, but that function is only called with a one-element tuple, so it had no consequences. I've changed it to take a vector instead.

src/groupeddataframe/grouping.jl
nalimilan: I agree, I just wasn't sure about the inconsistency with `map`. I've changed `combine` to support the same options as `by`. Maybe we'll have to use another function in addition to `map`, e.g. `select` or `transform` (like in DataFramesMeta).

src/groupeddataframe/grouping.jl
nalimilan: Yeah, that's https://github.com/JuliaData/DataFrames.jl/issues/1460.

src/groupeddataframe/grouping.jl
nalimilan: Right. I've set `makeunique=true` here too, but let's discuss that at #1460.

src/groupeddataframe/grouping.jl
nalimilan: It's already the last condition. I've added a comment to make this clearer.

Currently if you mix vectors and scalars the array is treated as a scalar. I've turned it into an error so that we can decide later to recycle as in constructors if we want.

src/groupeddataframe/grouping.jl
nalimilan: I'd rather add it later, this PR is already messy enough. :-)

docs/src/man/split_apply_combine.md
bkamins: OK

src/groupeddataframe/grouping.jl
bkamins: Any opinion on this? I am OK to leave it as is, but want this to be an explicit decision 😄.

docs/src/man/split_apply_combine.md
bkamins: We can treat is as resolved as we have established elsewhere what are the accepted arguments.

src/groupeddataframe/grouping.jl
bkamins: I see this as an important problematic case - why don't you want to accept returning an `AbstractVector`?

src/groupeddataframe/grouping.jl
bkamins: This can be marked as resolved - right?

src/groupeddataframe/grouping.jl

src/groupeddataframe/grouping.jl

src/groupeddataframe/grouping.jl

src/groupeddataframe/grouping.jl
bkamins: Did you leave this pattern as accepted or drop it?

nalimilan: I'm not sure what to do about that. It's supported as a by-product of accepting `f::Any` in first position. I could write a special method just to throw an error, but that sounds a bit absurd. So I've added a mention that this is supported.

src/groupeddataframe/grouping.jl
nalimilan: Indeed views are more appropriate since we're going to make copies anyway when allocating the final column.

src/groupeddataframe/grouping.jl
nalimilan: Just to make my life simpler. :-)

But now that everything has settled, I've added that feature.

src/groupeddataframe/grouping.jl

nalimilan:
I agree with the agenda. But I don't think we need to write anywhere that the functions will continue to evolve, that's OK as long as we don't plan on breaking them. We can always mention it in the release post. Also I think I've covered several corner cases of mixing return values of different types, shapes or lengths.

I've pushed the commit with the fixes.

src/groupeddataframe/grouping.jl
```diff
@@ -190,42 +231,100 @@ end
 
 """
     combine(gd::GroupedDataFrame)
-    combine(f::Function, gd::GroupedDataFrame)
+    combine(gd::GroupedDataFrame, cols => f...)
+    combine(gd::GroupedDataFrame; (colname = cols => f)...)
+    combine(gd::GroupedDataFrame, f)
+    combine(f, gd::GroupedDataFrame)
```
bkamins: In line 237 `f` can be also a tuple or a named tuple. Right?

src/groupeddataframe/grouping.jl
```diff
@@ -443,71 +655,145 @@ colwise(fns::Union{AbstractVector, Tuple}, d::AbstractDataFrame) = [f(d[i]) for
...
-* `cols` : a column indicator (Symbol, Int, Vector{Symbol}, etc.)
-* `f` : a function to be applied to groups; expects each argument to
-  be an AbstractDataFrame
-* `sort`: sort row groups (no sorting by default)
-
-### Details
+    by(d::AbstractDataFrame, keys, cols => f...; sort::Bool = false)
+    by(d::AbstractDataFrame, keys; (colname = cols => f)..., sort::Bool = false)
+    by(d::AbstractDataFrame, keys, f; sort::Bool = false)
+    by(f, d::AbstractDataFrame, keys; sort::Bool = false)
```
bkamins: same here - `f` can be tuple or named tuple I think. Can it be a `Pair` (same question to `combine`)

src/groupeddataframe/grouping.jl
bkamins: So the question left if `f` can be a 'Pair'?

nalimilan: Ah, yes. I've fixed that.

PR #1603, opened 2018-11-18
Remove deprecation of getindex

bkamins:
This should be merged only after new release of DataFrames.jl is tagged but I wanted to push it out now to show the consequences.

This depends on first fixing https://github.com/JuliaData/CategoricalArrays.jl/issues/170. I have commented out a relevant test, but it should be re-enabled before merging. When this is fixed I will also write additional tests.

What does this PR change many inter-related features:
* finish `getindex` deprecation for `DataFrame`, `SubDataFrame`, `DataFrameRow`
* improve `StackedVector` and `RepeatedVector` documentation and make them `IndexLinear` to use mechanics from Base to properly handle `getindex`;
* finish deprecation of iterator for `DataFrameColumns` (now it returns `Pair`)
* `DataFrame(::DataFrame)` makes a copy
* `DataFrameRow` constructor performs bounds checking which can be disabled by `@inbounds`
* add `size` to `DataFrameRow` and finish `iterate` deprecation
* `SubDataFrame` constructor disallows integer row selector (vector is required)


src/dataframe/dataframe.jl
```diff
@@ -303,11 +276,16 @@ function Base.getindex(df::DataFrame, row_inds::AbstractVector, col_inds::Abstra
...
+
 # df[:, MultiColumnIndex] => DataFrame
-function Base.getindex(df::DataFrame, row_ind::Colon, col_inds)
-    Base.depwarn("Indexing with colon as row will create a copy in the future. " *
-                 "Use `df[col_inds]` to get the columns without copying", :getindex)
-    df[col_inds]
+function Base.getindex(df::DataFrame, row_ind::Colon, col_inds::AbstractVector)
+    selected_columns = index(df)[col_inds]
+    new_columns = AbstractVector[copy(dv) for dv in _columns(df)[selected_columns]]
+    return DataFrame(new_columns, Index(_names(df)[selected_columns]))
```
nalimilan: Isn't there a more efficient way to create a subset of an `Index` than recreating one from scratch? Just wondering.

src/dataframe/dataframe.jl
```diff
@@ -303,11 +276,16 @@ function Base.getindex(df::DataFrame, row_inds::AbstractVector, col_inds::Abstra
...
+end
+
 # df[:, MultiColumnIndex] => DataFrame
-function Base.getindex(df::DataFrame, row_ind::Colon, col_inds)
-    Base.depwarn("Indexing with colon as row will create a copy in the future. " *
-                 "Use `df[col_inds]` to get the columns without copying", :getindex)
-    df[col_inds]
+function Base.getindex(df::DataFrame, row_ind::Colon, col_inds::AbstractVector)
+    selected_columns = index(df)[col_inds]
+    new_columns = AbstractVector[copy(dv) for dv in _columns(df)[selected_columns]]
```
nalimilan: Maybe it would be worth avoiding two allocations by iterating over `col_inds` and calling `_columns(df)[index(df)[col]]` on the fly in the comprehension? Not sure.

src/dataframerow/dataframerow.jl
```diff
@@ -6,8 +6,22 @@ A view of one row of an AbstractDataFrame.
...
+        end
+        @boundscheck if row < 1 || row > nrow(df)
+            throw(BoundsError("attempt to access a data frame with $(nrow(df)) " *
+                              "rows at index $row"))
+        end
+        new{T}(df, row)
+    end
 end
 
+@inline DataFrameRow(df::T, row::Integer) where T<:AbstractDataFrame =
```
```suggestion
@inline DataFrameRow(df::T, row::Integer) where {T<:AbstractDataFrame} =
```

src/dataframe/dataframe.jl
bkamins: An alternative I can think of would be to `deepcopy` the old index and remove columns from it, but it does not seem that it will be faster neither when `col_inds` is small nor when it is large relative to the original number of columns. So I would leave it as is for now.

src/dataframe/dataframe.jl
bkamins: This will not be efficient when we pass `col_inds` as symbols because we would have to do the lookup twice.
The alternative is to have three separate implementations for indexing with `Bool`, integers and `Symbols` - it is possible to improve the performance then, but I would leave it for a separate PR as it is significantly complex (what we do now replicates what was used in the past and is safe as we delegate handling of different corner cases to `Index`, where I made sure that all is "correct" - hopefully 😄).

bkamins:
> Can you move the non-breaking parts about `RepeatedVector` and `StackedVector` to a separate PR?

This would be only documentation parts. I can move it out, but I thought they are pretty simple so I included them (I can exclude them if you prefer). Should I do it?

The `LinearIndex` part and changing the signature from `Integer` to `Int` is a part of this PR as it is strictly related to making all types in DataFrames.jl follow the indexing style introduced in Julia 1.0 (essentially we make sure that all the mechanics is handled by Base as in the recent PR for `eachrow`/`eachcol`/`columns`).

nalimilan:
OK, leave these here then, I thought the code changes could be moved too.

bkamins:
Most likely merging https://github.com/JuliaData/DataFrames.jl/pull/1613 will create merge conflicts here so let us first finish https://github.com/JuliaData/DataFrames.jl/pull/1613, and then I will go back to this and fix them to make this PR fully ready.

src/dataframerow/dataframerow.jl
```diff
@@ -16,18 +30,27 @@ Return the parent data frame of `r`.
...
-function Base.getindex(r::DataFrameRow, idx::ColumnIndex)
-    return parent(r)[row(r), idx]
-end
-
+Base.view(adf::AbstractDataFrame, rowind::Integer, ::Colon) =
+    DataFrameRow(adf, rowind)
+Base.view(adf::AbstractDataFrame, rowind::Integer, colinds::AbstractVector) =
+    DataFrameRow(adf[colinds], rowind)
+
+Base.getindex(adf::DataFrame, rowind::Integer, colinds::AbstractVector) =
```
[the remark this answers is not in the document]
bkamins: I do not see the fallback you are referring to.

src/subdataframe/subdataframe.jl
```diff
@@ -112,54 +106,14 @@ Base.parent(sdf::SubDataFrame) = getfield(sdf, :parent)
...
-
-function Base.view(adf::AbstractDataFrame, rowinds, colind::ColumnIndex)
-    Base.depwarn("`view(adf, rowinds, col::ColumnIndex)` will create `view(df[col], rowinds)` in the future." *
-                 " Use `view(adf, rowinds, [col])` instead.", :view)
-    return SubDataFrame(adf[[colind]], rowinds)
-end
-
-function Base.view(adf::AbstractDataFrame, rowinds, colind::Bool)
+Base.view(adf::AbstractDataFrame, colinds) = view(adf, :, colinds)
+Base.view(adf::AbstractDataFrame, rowinds, colind::ColumnIndex) =
```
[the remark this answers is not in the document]
bkamins: Again - I am not clear which fallback you are referring to.

bkamins:
Apart from fixing #1608 I have also changed how `dump` works as it produced illegible results with `view` (`dump` for subarray is messy).

Also I cleaned-up the implementation of `dump` as the old one simply did not work correctly (but I do not know the reason why this `invoke` trick was used - maybe some old Julia stuff - but now the function did not work correctly), which you can see in this example:

```
julia> df = DataFrame(a=1:3, b=4)
3×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 4     │
│ 2   │ 2     │ 4     │
│ 3   │ 3     │ 4     │

julia> dump(stdout, df, 8, "")
DataFrame  3 observations of 2 variables
  a: ERROR: MethodError: no method matching dump(::Base.TTY, ::Array{Int64,1}, ::Int64, ::String)
```

src/dataframerow/dataframerow.jl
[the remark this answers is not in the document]
nalimilan: I think I confused with `view`.

src/dataframerow/dataframerow.jl
```diff
@@ -11,10 +11,7 @@ struct DataFrameRow{T<:AbstractDataFrame}
         if row isa Bool
             throw(ArgumentError("invalid index: $row of type Bool"))
         end
-        @boundscheck if row < 1 || row > nrow(df)
-            throw(BoundsError("attempt to access a data frame with $(nrow(df)) " *
-                              "rows at index $row"))
-        end
+        @boundscheck checkbounds(axes(df, 1), row)
```
nalimilan: The old error was much more explicit, right? Maybe we should move it to a helper function.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -317,43 +317,14 @@ Get a data frame with the `n` last rows of `df`.
...
-**Examples**
-
-```julia
-df = DataFrame(i = 1:10, x = rand(10), y = rand(["a", "b", "c"], 10))
-dump(df)
-```
-
-"""
-function Base.dump(io::IO, df::AbstractDataFrame, n::Int, indent)
+function Base.dump(io::IOContext, df::AbstractDataFrame, n::Int, indent)
```
nalimilan: I think that's too restrictive, any `IO` object should be allowed. Anyway, can we handle this in a separate PR?

src/dataframerow/dataframerow.jl
bkamins: OK. I will add better messages.

src/abstractdataframe/abstractdataframe.jl
bkamins: Any `IO` is allowed via `Base` via `dump(io::IO, @nospecialize(x); maxdepth=DUMP_DEFAULT_MAXDEPTH)` (show.jl line 1717).
This is a low-level function that gets called there. As I have showed in the error message - we really never handled `IO` as trying to use it threw an error. With this change `dump(stdout, df)` should work just fine.

bkamins:
@nalimilan This PR is good for a final review.

I have added several small fixes in recent commits so you can track the changes.
I tried to make sure that all works as expected. In order to ensure this I have added a separate test file only for testing indexing so it is easy to get a feeling what we will have after this PR (in the future `setindex!` tests should be added there).

src/dataframerow/dataframerow.jl
```diff
@@ -32,12 +32,16 @@ row(r::DataFrameRow) = getfield(r, :row)
 
 Base.view(adf::AbstractDataFrame, rowind::Integer, ::Colon) =
     DataFrameRow(adf, rowind)
+Base.view(adf::SubDataFrame, rowind::Integer, ::Colon) =
+    DataFrameRow(parent(adf), rowind)
```
nalimilan: The index is incorrect AFAICT. Also needs a test then.

bkamins: Thank you. Good catch. Actually I was aware of this, but forgotten to fix 😭.

Moreover I have written a test for it, but I was silly enough to make a `SubDataFrame` from a `DataFrame` starting from take the first row, when actually this passes OK.

I will fix all the tests to be better.

bkamins: a similar bug is also below.

bkamins:
OK. I have merged this PR. @nalimilan - thank you.

We do not have to release DataFrames 0.16 very soon, but I think it would be good not to make major changes in this release so that DataFrames 0.16 does not throw many deprecation warnings.

I now start sketching `setindex!` cleanup, but I would make a PR with the deprecations after DataFrames 0.16 is released so that in 0.17 we have `setindex!` deprecations and in 0.18 they would go in.

PR #1607, opened 2018-11-24
Deprecate head and tail in favor of first and last

nalimilan:
Now that we have decided that data frames are collections of rows (https://github.com/JuliaData/DataFrames.jl/issues/1514), there is no point in defining separate `head` and `tail` functions, which are a legacy from the R API. Do not define one-argument methods defaulting to `n=6` since other `first` and `last` methods in Base default to `n=1`, which should return a single row (as a `NamedTuple` or `DataFrameRow`).


bkamins:
There are two points to discuss:

Issue 1. Do we really want to get rid of `head` and `tail`. I will not oppose removing them, but how much benefit do we get by removing them (probably there is some global logic behind it - similarly to one behind wanting to remove `nrow` and `ncol` - was it discussed earlier and what are the main arguments for such an approach).

Issue 2. Actually if we want to do this change we should also define `first(::AbstractDataFrame)` and `last(::AbstractDataFrame)` as now the former throws an error and the latter returns last column of the data frame which is confusing.

pdeffebach:
`first` and `last` imply iteration to me. `df[1]` would have to return the first row. 

Also note that in `R`, `head` is useful because unlike the default `show` method, it prints all column names and types. Although this is usually illegible. We couldn't do that, but maybe `head` could be `show(df, allcols = true)` to get more similar behavior. 

In stata when you first get a dataset you write `summarize`, which is more akin to `describe`. Which is probably preferred imo since you learn more about a dataset. We should tell people to do that instead when they read in data for the first time. 

But removing `head` would mean constantly reminding people that `head` doesn't exist. There's not much harm in keeping it.

nalimilan:
`head` and `tail` don't really harm, except that they work only for data frames, while Julia already has `first` and `last` to get the first/last elements in a collection. I expect the two-argument form to be supported for vectors in the future -- we can open an issue in Julia to discuss that before merging this PR if you want. Also, `Base.tail` already exists and does something different, and it could be exported at some point.

Good catch about `last(df)` returning the last column. Better define `first(df)` and `last(df)` to return the first/last row, using `df[i, :]`. I've added a commit to do that. I don't think the inconsistency with `df[i]` is a problem: `df[i]` is the inconsistent syntax since we have decided data frames are collections of rows.

> But removing `head` would mean constantly reminding people that `head` doesn't exist. There's not much harm in keeping it.

That's not generally an argument we have considered as legitimate on its own, as consistency with other languages would make Julia inconsistent with itself. When several options are mostly equivalent, taking the one which is consistent with other languages is a good idea, but when something already exists in Julia duplicating features just to help users of language X makes the API bigger, and what happens when somebody coming from language Y wants to add another name for that function?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -288,38 +288,33 @@ Base.isempty(df::AbstractDataFrame) = size(df, 1) == 0 || size(df, 2) == 0
...
 
-```julia
-head(df::AbstractDataFrame, r::Int = 6)
-tail(df::AbstractDataFrame, r::Int = 6)
-```
-
-**Arguments**
+Get the first row of `df` as a `DataFrameRow`.
+"""
+Base.first(df::AbstractDataFrame) = df[1, :]
```
bkamins: This will throw an error on 0-row data frame. Maybe some better error message would be in place. Note that `first(df, 1)` will work on `0`-row data frame. Same with `last`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -288,38 +288,33 @@ Base.isempty(df::AbstractDataFrame) = size(df, 1) == 0 || size(df, 2) == 0
...
 
-* `df` : the AbstractDataFrame
-* `r` : the number of rows to show
+"""
+    first(df::AbstractDataFrame, n::Integer)
 
-**Result**
+Get a data frame with the `n` first rows of `df`.
+"""
+Base.first(df::AbstractDataFrame, n::Integer) = df[1:min(n,nrow(df)), :]
```
bkamins: Maybe it would be better to be a bit breaking and return a `SubDataFrame`? This would be consistent with `first(df)` which also returns a view now. I do not have very strong preference here over `DataFrame` so @nalimilan - please decide.

test/dataframe.jl
```diff
@@ -665,10 +665,16 @@ module TestDataFrame
 
     @testset "description" begin
         df = DataFrame(A = 1:10)
-        @test head(df) == DataFrame(A = 1:6)
-        @test head(df, 1) == DataFrame(A = 1)
-        @test tail(df) == DataFrame(A = 5:10)
-        @test tail(df, 1) == DataFrame(A = 10)
+
+        @test first(df) == df[1, :]
+        @test last(df) == df[10, :]
```
bkamins: maybe better `end` than `10` to really show what we mean?

src/abstractdataframe/abstractdataframe.jl
nalimilan: I've followed what happens with vectors. I think that makes sense. The error message isn't too bad, it indicates you tried to access row 1 in a 0-row data frame.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Good question. `first(df)` returns a view only because `df[1, :]` does, so I'd say it's consistent to also use `df[1:i, :]` here, and therefore return a copy.

pdeffebach:
> That's not generally an argument we have considered as legitimate on its own, as consistency with other languages would make Julia inconsistent with itself. 

Yeah that's a good point. It's important not to keep too many legacy APIs for sure. It's also worth nothing that i've never found `head` useful, and would not miss it.

bkamins:
I think we can merge this

Issue #1611, opened 2018-11-29
Tables.jl uses `eachcolumn` and a new PR to base does too


pdeffebach:
https://github.com/JuliaLang/julia/pull/29749#issuecomment-442637088 seeks to implement an `eachcolumn` feature on `AbstractArray` types. 

There is a discussion about how to handle the backporting there, but we can't do `import eachcolumn` is our function is named `eachcol`. 

Tables.jl also calls their column iterator `eachcolumn.` 

Maybe we should follow suit for consistency and change the name? It's massively breaking, though.

pdeffebach:
Ah. It makes sense that it was designed to avoid confusion. 

Still a bit odd to have `(rows, eachcolumn)`, `(eachrow, eachcol)`, and `(eachrow, eachcolumn)`. But that's another issue.

nalimilan:
That PR implements `eachcol`, not `eachcolumn` AFAICT.

What worries me, though, is that `eachcol` currently returns a `name => column` pair, while `Base.eachcol` will return a view on the column. For consistency, it would be nice if we returned just the column (like `columns` does). We could deprecate `eachcol(df)` in favor of `eachcol(df, true)` or something like that. I wish we had considered this when discussing https://github.com/JuliaData/DataFrames.jl/issues/1586. @bkamins What do you think?

bkamins:
I was unaware of https://github.com/JuliaLang/julia/pull/29749 when making the changes (it was submitted exactly in parallel).

Acutally initially I have proposed something like `eachcol(df, true)` for current behavior if I remember correctly. The problem is that this way `eachcol` would not be type stable. We would have to use `eachcol(df, Val(true))` or somethng similar. The good thing is that the `columns` change was not tagged yet so we can change it.

Given `https://github.com/JuliaLang/julia/pull/29749` I think what we should do is:
* in this release:
    * do not export `columns(df)`
    * make `eachcol(df, Val(false))` call `columns(df)`
    * change `eachcol(df)` to `eachcol(df, Val(true))`
    * when `eachcol(df)` is called execute `eachcol(df, Val(true))` and throw depwarn
* in the next release
    * when `eachcol(df)` is called execute `eachcol(df, Val(false))`
    * get rid of internal `columns(df)` and move its implementation to `eachcol(df, Val(false))`

This is a simple fix and I can push a PR that does this.

[...]

PR #1612, opened 2018-11-30
Remove export of columns function

bkamins:
PR following #1611 so that we can have a look at the consequences.


docs/src/lib/types.md
```diff
@@ -36,18 +36,21 @@ In the future accessing a single row of a data frame via `getindex` or `view` wi
...
 serves as an iterator over rows of an `AbstractDataFrame`, returning `DataFrameRow` objects.
 
-Similarly, the `eachcol` and `columns` functions return a value of the `DataFrameColumns` type, which
+Similarly, the `eachcol` function returns a value of the `DataFrameColumns` type, which
 serves as an iterator over columns of an `AbstractDataFrame`.
-The difference between the return value of `eachcol` and `columns` is the following:
+The return value can have two concrete types:
 
-* The `eachcol` function returns a value of the `DataFrameColumns{<:AbstractDataFrame, true}` type, which is an
+* If the `eachcol` function is called with `names` argument set to `false` (currently the default,
```
```suggestion
* If the `eachcol` function is called with the `names` argument set to `true` (currently the default,
```

src/abstractdataframe/iteration.jl
```diff
@@ -38,21 +38,24 @@ end
 """
     DataFrameColumns{<:AbstractDataFrame, V} <: AbstractVector{V}
 
-Iterator over columns of an `AbstractDataFrame`.
-If `V` is `Pair{Symbol,AbstractVector}` (which is the case when calling
-[`eachcol`](@link)) then each returned value is a pair consisting of
-column name and column vector. If `V` is `AbstractVector` (a value returned by
-the [`columns`](@link) function) then each returned value is a column vector.
+Iterator over columns of an `AbstractDataFrame` constructed using
+[`eachcol`](@link) function. . If `V` is `Pair{Symbol,AbstractVector}` then each
```
```suggestion
[`eachcol(df, true)`](@link). If `V` is a `Pair{Symbol,AbstractVector}` then each
```

src/abstractdataframe/iteration.jl
```diff
@@ -38,21 +38,24 @@ end
...
+returned value is a pair consisting of column name and column vector. If `V` is
+`AbstractVector` (a value returned by the [`columns`](@link) function) then each
```
```suggestion
an `AbstractVector` (a value returned by [`eachcol(df, false)`](@link)) then each
```

src/abstractdataframe/iteration.jl
```diff
@@ -67,54 +70,46 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
            maximum(col) - minimum(col)
        end
 2-element Array{Int64,1}:
  3
  3
 ```
 """
-columns(df::T) where T<: AbstractDataFrame =
-    DataFrameColumns{T, AbstractVector}(df)
+eachcol(df::T, names::Bool) where T<: AbstractDataFrame =
```
nalimilan: I'd use the long function form, and add `@inline` just in case.

bkamins: OK

src/abstractdataframe/iteration.jl
```diff
@@ -38,21 +38,24 @@ end
 """
     DataFrameColumns{<:AbstractDataFrame, V} <: AbstractVector{V}
 
-Iterator over columns of an `AbstractDataFrame`.
-If `V` is `Pair{Symbol,AbstractVector}` (which is the case when calling
-[`eachcol`](@link)) then each returned value is a pair consisting of
-column name and column vector. If `V` is `AbstractVector` (a value returned by
-the [`columns`](@link) function) then each returned value is a column vector.
+Iterator over columns of an `AbstractDataFrame` constructed using
+[`eachcol(df, true)`](@link) if `V` is a `Pair{Symbol,AbstractVector}`. Then each
```
```suggestion
[`eachcol(df, true)`](@ref) if `V` is a `Pair{Symbol,AbstractVector}`. Then each
```
Right?

bkamins: I will check with documenter - it was your change so I retained it and I never remember for sure what should be used when.

bkamins: You are right. Fixed.

bkamins:
It should be good for a final check. After the release I will make a second step changing `eachcol` to return only vectors.

PR #1613, opened 2018-12-03
make eachcol default to false

bkamins:
This cleans up deprecations and respects https://github.com/JuliaLang/julia/pull/29749.

It would not hurt if someone checked if I had put right Julia version in the condition. Thanks!

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -202,7 +202,7 @@ eltypes(df)
 ```
 
 """
-eltypes(df::AbstractDataFrame) = eltype.(columns(df))
+eltypes(df::AbstractDataFrame) = eltype.(eachcol(df))
 
 Base.size(df::AbstractDataFrame) = (nrow(df), ncol(df))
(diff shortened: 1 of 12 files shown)


nalimilan:
This looks good to me, but I'm not sure we should merge it right now since it will break code that hasn't been ported yet, and there's no hurry to change this: old code that used `eachcol(df)` will have to do `eachcol(df, true)` anyway. AFAICT the only advantage of doing this (apart from cleanup) is to allow people to write `eachcol(df)` instead of `eachcol(df, false)`. Better make a release removing deprecations which people cannot turn off (in indexing code).

bkamins:
OK - let us put it on hold. I will keep it open to remember what has to be changed. How releases would you wait with fixing this?

bkamins:
Do we want to finish the deprecation for `eachcol(::DataFrame)` or we wait one more release?

bkamins:
@nalimilan - I think that enough time has passed for the deprecation. If you are OK to merge it now I will review this PR to make sure and make it ready for merging.

nalimilan:
Fine with me.

PR #1614, opened 2018-12-03
avoid depwarn in eachcol

bkamins:
@nalimilan This is a pretty urgent PR to merge and tag a minor release. It avoids pringing deprecation warning on eachcol when using `dump` or showing a data frame (which calls `getmaxwidths`).

### DIFF
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -36,8 +36,6 @@ export AbstractDataFrame,
        disallowmissing!,
        dropmissing,
        dropmissing!,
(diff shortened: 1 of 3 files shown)


bkamins:
I have also added a Julia version check here as it is not breaking. This PR should be merged and tagged before #1613 is merged.

nalimilan:
https://github.com/JuliaLang/METADATA.jl/pull/19865

bkamins:
Thank you!

Issue #1617, opened 2018-12-05
`deleterows!` inconsistency


bkamins:
We  now have an inconsistency in if `deleterows!` modifies a source column (which is related to implementing #1441 if we want to make it work right in all cases):
```
julia> x = [1,2]
2-element Array{Int64,1}:
 1
 2

julia> df = DataFrame(x=x)
2×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │

julia> deleterows!(df, 1)
1×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 2     │

julia> x
1-element Array{Int64,1}:
 2
```
and
```
julia> x = [1,2]
2-element Array{Int64,1}:
 1
 2

julia> df = DataFrame(x=x)
2×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │

julia> deleterows!(df, 1:1)
1×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 2     │

julia> x
1-element Array{Int64,1}:
 2
```
but
```
julia> x = [1,2]
2-element Array{Int64,1}:
 1
 2

julia> df = DataFrame(x=x)
2×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │

julia> deleterows!(df, [1])
1×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 2     │

julia> x
2-element Array{Int64,1}:
 1
 2
```

Also
```
julia> x = [1,2]
2-element Array{Int64,1}:
 1
 2

julia> df = DataFrame(x=x)
2×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │

julia> deleterows!(df, [true, false])
ERROR: MethodError: no method matching deleterows!(::DataFrame, ::Array{Bool,1})
```
fails and it should work.

And:
```
julia> df = DataFrame(x = [1,2,3,missing], y = 1:4)
4×2 DataFrame
│ Row │ x       │ y     │
│     │ Int64⍰  │ Int64 │
[...]

PR #1618, opened 2018-12-06
improve row-removal functions

bkamins:
This PR relates to #1441 and #1617. It depends on https://github.com/JuliaLang/julia/pull/30287 for correctly handling some corner cases (but it is OK to merge it without that PR as those are obscure cases).

What it changes:
* change signature of `unique!` and `filter!`, `dropmissing!` functions as they allow only `DataFrame`
* `dropmissing!` and `dropmissing` get a keyword argument `keepeltype` stating if we want to keep eltype of the original columns or change them to disallow `Missing`
* fix `deleterows!` to always create a new vector instead of modifying the source (note: we could do it the other way around - i.e. always update the source; please comment what would be a preferred solution)
* make `deleterows!` work consistently with `deleteat!` in base (in particular allow boolean indexing and expect that the rows to be deleted are sorted - in the past this requirement was not present so this is a bit breaking - I can add sorting or I can add depwarn, but now an error on unsorted data is thrown so I think it is OK go go forward as proposed)

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -28,10 +28,12 @@ The following are normally implemented for AbstractDataFrames:
 * `convert` : convert to an array
 * [`completecases`](@ref) : boolean vector of complete cases (rows with no missings)
 * [`dropmissing`](@ref) : remove rows with missing values
-* [`dropmissing!`](@ref) : remove rows with missing values in-place
+* [`dropmissing!`](@ref) : remove rows with missing values in-place (`DataFrame` only)
 * [`nonunique`](@ref) : indexes of duplicate rows
-* [`unique!`](@ref) : remove duplicate rows
+* [`unique!`](@ref) : remove duplicate rows (`DataFrame` only)
 * `similar` : a DataFrame with similar columns as `d`
+* `filter` : remove rows
+* `filter` : remove rows in-place (`DataFrame` only)
 
 **Indexing**
 
@@ -592,13 +594,16 @@ end
 completecases(df::AbstractDataFrame, cols::AbstractVector) =
     completecases(df[cols])
 
+# TODO: update docstrings after deprecation of keepeltype
 """
-    dropmissing(df::AbstractDataFrame)
-    dropmissing(df::AbstractDataFrame, cols::AbstractVector)
-    dropmissing(df::AbstractDataFrame, cols::Union{Integer, Symbol})
+    dropmissing(df::AbstractDataFrame; keepeltype::Bool=true)
+    dropmissing(df::AbstractDataFrame, cols::AbstractVector; keepeltype::Bool=true)
+    dropmissing(df::AbstractDataFrame, cols::Union{Integer, Symbol}; keepeltype::Bool=true)
 
 Return a copy of data frame `df` excluding rows with missing values.
 If `cols` is provided, only missing values in the corresponding columns are considered.
+
+In the future `keepeltype` will be `false` by default.
 
 See also: [`completecases`](@ref) and [`dropmissing!`](@ref).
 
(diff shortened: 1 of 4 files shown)


bkamins:
This should be good for a review.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -28,10 +28,12 @@ The following are normally implemented for AbstractDataFrames:
...
 * [`completecases`](@ref) : boolean vector of complete cases (rows with no missings)
 * [`dropmissing`](@ref) : remove rows with missing values
-* [`dropmissing!`](@ref) : remove rows with missing values in-place
+* [`dropmissing!`](@ref) : remove rows with missing values in-place (`DataFrame` only)
 * [`nonunique`](@ref) : indexes of duplicate rows
-* [`unique!`](@ref) : remove duplicate rows
+* [`unique!`](@ref) : remove duplicate rows (`DataFrame` only)
 * `similar` : a DataFrame with similar columns as `d`
+* `filter` : remove rows
+* `filter` : remove rows in-place (`DataFrame` only)
```
```suggestion
* `filter!` : remove rows in-place (`DataFrame` only)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -592,14 +594,17 @@ end
 completecases(df::AbstractDataFrame, cols::AbstractVector) =
     completecases(df[cols])
 
+# TODO: update docstrings after deprecation of keepeltype
 """
-    dropmissing(df::AbstractDataFrame)
-    dropmissing(df::AbstractDataFrame, cols::AbstractVector)
-    dropmissing(df::AbstractDataFrame, cols::Union{Integer, Symbol})
+    dropmissing(df::AbstractDataFrame; keepeltype::Bool=true)
```
nalimilan: I'd call the argument `disallowmissing` for consistency with the function with the same name.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -645,79 +658,16 @@ julia> dropmissing(df, [:x, :y])
...
-│     │ Int64 │ Int64⍰ │ String⍰ │
-├─────┼───────┼────────┼─────────┤
-│ 1   │ 2     │ 4      │ missing │
-│ 2   │ 4     │ 2      │ d       │
-│ 3   │ 5     │ 1      │ e       │
+function dropmissing(df::AbstractDataFrame,
+                     cols::Union{Integer, Symbol, AbstractVector}=1:size(df, 2);
+                     keepeltype::Bool=true)
+    if keepeltype
+        Base.depwarn("dropmissing will change eltype of cols to disallow Missing by default. " *
```
```suggestion
        Base.depwarn("dropmissing will change eltype of cols to disallow missing by default. " *
```
Should also give the syntax which allows getting the new behavior to turn off the warning.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -645,79 +658,16 @@ julia> dropmissing(df, [:x, :y])
 ```
 
 """
-dropmissing(df::AbstractDataFrame,
-            cols::Union{Integer, Symbol, AbstractVector}=1:size(df, 2)) =
-    deleterows!(copy(df), findall(!, completecases(df, cols)))
```
nalimilan: I'm not sure I like this change. Technically, a custom `AbstractDataFrame` implementation could implement `deleterows!` and automatically benefit from these definitions. If people call this on e.g. a `SubDataFrame`, they will get an error about `deleterows!`, which is relatively clear. If we want a more explicit error, we can define a `SubDataFrame` method which would print an error saying that it cannot be modified. We could also allow calling `deleterows!` on a  `SubDataFrame`, that would kind of make sense.

In terms of implementations, I also find it clearer to have the in-place methods defined at the same place as the copying methods.

src/dataframe/dataframe.jl
```diff
@@ -777,39 +777,144 @@ end
 deletecols!(df::DataFrame, c::Int) = deletecols!(df, [c])
 deletecols!(df::DataFrame, c::Any) = deletecols!(df, index(df)[c])
 
-function deleterows!(df::DataFrame, ind::Union{Integer, UnitRange{Int}})
+function deleterows!(df::DataFrame, ind)
+    # we require ind to be stored and unique like in Base
+    keep = deleteat!(collect(axes(df, 1)), ind)
```
nalimilan: I'd prefer removing in-place, that's one of the interests of having `deleterows!`. I guess we could also provide `deleterows` which would make a copy.

In terms of implementation, it will probably be more efficient to have a method for `AbstractVector{Bool}` which will call `findall`, to avoid computing that for each column. BTW, that's probably something we should also do for `getindex`.

bkamins:
I will put this PR on hang, as first #1619 should go in (as that PR is non breaking and should go into a minor release). Then I will rebase this one.

In general I am  OK with what you propose:
* call the keyword argument `disallowmissing`
* allow `AbstractDataFrame` and have specific methods for `SubDataFrame` that fail
* mutate the columns in place

bkamins:
Closing. I have opened #1630 that is an improved version of this one.

PR #1619, opened 2018-12-06
add proper fallbacks for dump, getmaxwidths and mapcols to avoid depwarns

bkamins:
Avoid depwarn messages when using `SubDataFrame`


bkamins:
This can go as a minor release (unless we get some more things to fix)

bkamins:
I have also added non-breaking changes in unique-family methods as the test of `unique!` was failing on nightly due to changes in `unique!` method signature in Base (I have changed others to be in sync).

In general the change in base is similar to what we do in DataFrames.jl - i.e. to allow functors in the call `unique!(f, a)` which resulted in method ambiguity.

Issue #1615, opened 2018-12-03
`by` with arrays of inputs and broadcasting


pdeffebach:
It looks like the splatting doesn't quite work with having vectors of columns you want to act on. 

```
using Statistics
df = DataFrame(x = repeat(1:2, 5), a=rand(10), b=rand(10));
by(df, :x, [:a, :b] .=> mean, :b => std) # doesn't work
#=
ERROR: MethodError: no method matching by(::DataFrame, ::Symbol, ::Array{Pair{Symbol,typeof(mean)},1}, ::Pair{Symbol,typeof(std)})
=#
by(df, :x, :a => mean, :b => std) # works
```

I'm done with grad apps now so I will see if I can poke around for a fix soon.

bkamins:
I thought that this was intentional, but we can add merging as @nalimilan writes. The rules would be to merge:
* with positional arguments:
    * `Pair`s and `Tuple`s of `Pair`s into a single `Tuple` (this is pretty clear that it should be allowed and is simple)
    * several `NamedTuple`s of `Pair`s into a single `NamedTuple` (this is pretty clear that it should be allowed and is simple)
* with keyword arguments: here it is problematic as we would have to allow mixing positional and keyword arguments (this will be more involved to get it fully right as there are many possible problematic combinations to safeguard against) - @pdeffebach is there a need to handle this case?

pdeffebach:
I agree that named-tuple for broadcasted arguments is not necessary. But I think that this should be allowed. 

Yes I think merging should do the trick, we can discuss more in depth once I make a PR/

bkamins:
@pdeffebach When you write a PR please take this behavior into account:
```
julia> merge((a=1, b=2), (b=10, c=20))
(a = 1, b = 10, c = 20)
```
so duplicates are silently overwritten. We do not want this I think and an error should be thrown in case of a duplicate.

bkamins:
This is done, so I am closing this. We are missing `All`, `Not` and `Between` broadcasting, but this is a separate issue (the mechanics is there). If you feel something needs to ba added here please reopen.

PR #1620, opened 2018-12-07
Add more splatting and appending of arguments in by

pdeffebach:
This is a first attempt at resolving #1615 

It just gets around this by assuming that if you have many different arguments we can `vcat` them together and pass it off to `combine`. `combine` has a pretty good infrastructure for sorting out arguments so I don't think this needs to happen at `by`.


pdeffebach:
I added the method for `f::Any...` rather than write a new method because `f::Union{Pair, Vector{Pair}}...` choose `Pair` OR `Vector{Pair}` but does not allow a collection of both. 

Unless my interpretation is mistaken, which is likely is.

bkamins:
A small comment (I have not analyzed the code yet): we want `combine` and `by` to have the same API so probably `combine should get an additional method also and then `by` implementeation would change.

bkamins:
> I think this should be done by `combine`

Just to make sure I understand you correctly. Do you mean what I have recommended - to have `combine` handle all required cases consistently with `by`? And then `by` would just pass what it gets to `combine` and leave all the work for `combine`?

pdeffebach:
Okay I added a method for `combine` to make this work. 

I changed the `by` method to have a splatted `::Union{Pair, AbstractVector{<:Pair}}` but I could presume Milan you want a splatted `::Any...` in `by`?

pdeffebach:
This needs more work. I'm going to see if I can get this to work while also allowing (non-vector) keyword arguments.

pdeffebach:
Okay my strategy for this will be 4 methods for each `combine` and `by`. 

```
combine(f::Any, g) # for do syntax with old version
combine(g, f::Any) # for non-do syntax with old version

combine(g, f::{Union{Pair, AbstractVector{<:Pair}, Tuple{<:Pair}}...; kwargs...)  

combine(g, f::NamedTuple) # the one that actually goes to _combine
```
Here I take the non-keyword arguments and collect them into a named tuple somehow, meaning I move the the autogeneration of variable names to this stage, also making sure I don't overwrite names from the `kwargs` named tuple. 

Does that make sense?

pdeffebach:
I have implemented the above strategy, but only for `combine`, since we will have to repeat the same logic with `by` and its easier to focus on just `combine` while we nail down the strategy. 

Rather than splatting a vector of `Pair` inputs, it uses a `reduce(vcat, args)` to get a nice vector of `Pair` arguments regardless of wither you have just a `Pair`, or a `Vector` of `Pair`s or a mixture of the two. 

I have a big commented out block that will handle the re-naming in the case of duplicate names, but I need to put more work into understanding the logic of it and what I need to change. The code originally came from the `makeunique = true` option somewhere. 

Any feedback is welcome.

nalimilan:
Makes sense, but note you'll have to handle the case where a single `Pair` is passed separately, since we support returning a `NamedTuple` with several fields (and then the names can only be known from `firstres`). Maybe keep special methods for when you already have an homogeneous named tuple, to avoid doing unnecessary work? Also, better base your work on https://github.com/JuliaData/DataFrames.jl/pull/1651 as it changes the logic a bit.

pdeffebach:
Rebased and revived this PR. However in the process the `_combine` step doesn't work. Thankfully after this we can delete all the code for `_combine` that works with anything other than a named tuple. Not sure how to fix this error and how to figure out which `_combine` code should be deleted.

src/groupeddataframe/grouping.jl
```diff
@@ -373,17 +374,53 @@ function combine(f::Any, gd::GroupedDataFrame)
...
+# pair syntax  
+function combine(gd::GroupedDataFrame, args::Union{Pair, AbstractVector{<:Pair}}...; kwargs...)
 
-function combine(gd::GroupedDataFrame; f...)
-    if length(f) == 0
-        Base.depwarn("combine(gd) is deprecated, use DataFrame(gd) instead", :combine)
-        combine(identity, gd)
-    else
-        combine(values(f), gd)
+    # handle pairs on their own so we can iterate
```
nalimilan: That's really too much code for a small feature IMHO. Is there any way to make this simpler? We don't have to allow mixing keyword and positional arguments.

pdeffebach:
> That's really too much code for a small feature IMHO. Is there any way to make this simpler? We don't have to allow mixing keyword and positional arguments.

If we don't mix positional and keyword arguments than things become simple again, and I doubt people will miss them.  I would prefer to either close this or do a force-push of master to start anew. 

It would be nice to allow splatting for multiple arrays in a function call. Then we wouldn't have to worry about supporting vector arguments at all. However this isn't allowed, right? 

```
x = [1, 2, 3]; y = [4, 5, 6]
foo(x..., y...) # foo(1, 2, 3, 4, 5, 6)
```

pdeffebach:
I have reset this `--hard` and rebased. Will get a much more minimal implementation together tomorrow. I will

a) make a function with `Union{Pair, AbstractVector{<:Pair}...` and then call `reduce(vcat)` on it. 
b) delete the method for `_combine` that this change would make dead code.

pdeffebach:
Okay this PR should be ready for a review! I promise its a lot less to review than last time. 

I want to deprecate one method, and thats `combine(df, :a, (:b => mean,))`. It's only allowed because `_combine` allows for `NTuple{Pair}`. But now that using many vectors is flexible, supporting a Tuple argument seems unnecessary.

bkamins:
> But now that using many vectors is flexible, supporting a Tuple argument seems unnecessary.

But tuple is lighter than vector so why would we want to prefer vectors over tuples? (but this is a corner case anyway - probably @nalimilan has a stronger opinion here what is right, because he spent many more hours than me on this).

> I want to deprecate one method

I could not find the code with the deprecation (but you probably can add it later - first we should decide if we want to drop support for it).

> `combine(df, :a, (:b => mean,))`

Did you rather mean `by(df, :a, (:b => mean,))`?

Finally - maybe you can add to tests the original reason for this PR, i.e. something like `by(df, :x, [:a, :b] .=> mean, :b => std)` + as always add negative tests that should fail 😄 (they are really important in long run when we change functionality).

pdeffebach:
thanks, I will add those tests, and try to think of a good failing test. 

`by` and `combine` should always have the same syntax, I think. 

I have not deprecated it and it works now, but is only type-specified at the `_combine` level, rather than the `combine()` level. i.e. it fits in as part of `combin(gd::AbstractDataFrame, f::Any)` method. 

I want to deprecate it because if we allow tuple, then we might as well have multiple tuples, along with tuple and scalar...

bkamins:
> `by` and `combine` should always have the same syntax, I think.

AFAIK `by` takes `AbstractDataFrame` and grouping columns to group by and `combine` takes `GroupedDataFrame` (and it seems in what you have written you pass a data frame and grouping column).

pdeffebach:
Oh! sorry. yes that was my mistake.

pdeffebach:
I added a tuple argument for expliciteness, however you can't do the following, combining a tuple with a scalar

```
combine(gd, (:b => mean, :b => first), :c => last) # fails!
```

If you want to mix a collection with a scalar, you have to use `Vector`s. This is because the splatting will cause an ambiguity in methods if we use the same approach with `Tuple`s as with `Vector`s.

bkamins:
Thanks! Similar to what I have recommended with `vcat` could you please update the docs so that I am clear what contract for the function you propose and then I will check:

* if the implementation follows the contract;
* if the contract is OK (in particular I am not fully clear why it is not possible for the last example to work and it seems inconsistent unless the contract will be that we allow `Vector{Pair}` or `Pair` - in other words - if we are unable to handle tuples uniformly I would disallow them altogether as IMO this will not be a performance problem to use only vectors here)

pdeffebach:
To avoid having an inconsistency between tuples and vectors, I have deprecated `combine(gd, (:b => mean, :b => std))`. You now have to use a vector, due to the method ambiguity mentioned above. 

Does this change merit a larger discussion about the way we treat tuples of arguments? I would be inclined to purge all tuples of arguments in favor of vectors, if possible. This will cause a schism between JuliaDB and DataFrames.

pdeffebach:
@nalimilan Can we deprecate `combine(gd, f::Any)` and mandate that people use the `do` syntax with `combine(f, gd)`? Just thinking of how many methods we are defining.

pdeffebach:
I took a stab at re-writing the docstring to have a separate bulle for each argument input type. Let me know your comments.

bkamins:
> I would be inclined to purge all tuples of arguments in favor of vectors

I also prefer vectors (although they allocate and are type unstable) as they are a bit more comfortable to type, and the performance hit should be negligible. But there is this JuliaDB consistency issue you notice so I am not sure what would be best.

> Can we deprecate `combine(gd, f::Any)`

I am OK to remove it as this is not a recommended style anyway, but would it simplify the code (if not we could as well keep it)?

Regarding the docs - yes this is more or less how I would see it (I guess it is not finished - right?)

pdeffebach:
> I also prefer vectors (although they allocate and are type unstable) as they are a bit more comfortable to type, and the performance hit should be negligible. 

I, like the trend of pairs being used for our API, and pairs are always cheap. While we should definitely keep an eye out for these problems, if your dataframe has 2000 columns, your performance bottlenecks are unlikely to be these function calls. 

> I am OK to remove it as this is not a recommended style anyway, but would it simplify the code (if not we could as well keep it)?

The main uses of this `any` method is for a callable, like with `do` syntax and `NamedTuple`s. So I will push a method to make the `NamedTuple` argument explicit at the very least.

nalimilan:
I'd rather keep `combine(gd, f::Any)` for consistency with other `combine(gd, ...)` methods. Same for tuples.

pdeffebach:
Should I make the tuple and namedtuple arguments explicit? Is `Any` just used as a catch-all for `callable`?

nalimilan:
Yes `Any` is used because `Callable` doesn't cover callable objects. But I'm not sure where you want to make tuples explicit.

src/groupeddataframe/grouping.jl
```diff
@@ -344,6 +348,20 @@ julia> combine(:c => sum, gd)
 │ 3   │ 3     │ 10    │
 │ 4   │ 4     │ 12    │
 
+julia> combine(gd, [:b, :c] => sum)
```
nalimilan: Can you show an example using `.=>` and mixing a single pair and a vector? I've removed that from the signature for simplicity.

src/groupeddataframe/grouping.jl
```diff
@@ -373,9 +391,21 @@ function combine(f::Any, gd::GroupedDataFrame)
...
+function combine(gd::GroupedDataFrame, f::Union{Pair, AbstractVector{<:Pair}}...) 
+    vec_of_pairs = reduce(vcat, f)
+    combine(vec_of_pairs, gd)
+end
+
+function combine(gd::GroupedDataFrame, f::NamedTuple)
+    combine(f, gd)
+end
+
+function combine(gd::GroupedDataFrame, f::Tuple{Vararg{<:Pair}})
```
nalimilan: Why not include this in the `Union` above? Also, use short function form.

src/groupeddataframe/grouping.jl
```diff
@@ -1037,12 +1068,15 @@ julia> by(df, :a, (:b, :c) => x -> (minb = minimum(x.b), sumc = sum(x.c)))
...
     combine(f, groupby(d, cols, sort = sort))
+
 by(f::Any, d::AbstractDataFrame, cols::Any; sort::Bool = false) =
     by(d, cols, f, sort = sort)
-by(d::AbstractDataFrame, cols::Any, f::Pair; sort::Bool = false) =
-    combine(f, groupby(d, cols, sort = sort))
-by(d::AbstractDataFrame, cols::Any, f::Pair...; sort::Bool = false) =
-    combine(f, groupby(d, cols, sort = sort))
+
+function by(d::AbstractDataFrame, cols::Any, f::Union{Pair, AbstractVector{<:Pair}}...; sort::Bool = false)
```
nalimilan: Also allow tuples of pairs (which are tested AFAICT).

src/groupeddataframe/grouping.jl
```diff
@@ -1037,12 +1068,15 @@ julia> by(df, :a, (:b, :c) => x -> (minb = minimum(x.b), sumc = sum(x.c)))
 """
```
nalimilan: Update docstring for `by`.

src/groupeddataframe/grouping.jl
pdeffebach: added.

src/groupeddataframe/grouping.jl
pdeffebach: Its added, I forgot to delete.

src/groupeddataframe/grouping.jl
pdeffebach: It's covered in the `::Any` methods. I only list the ones where the arguments need something done to them.

src/groupeddataframe/grouping.jl
pdeffebach: done.

src/groupeddataframe/grouping.jl
```diff
@@ -617,7 +655,8 @@ function do_f(f, x...)
     end
 end
 
-function _combine(f::Union{AbstractVector{<:Pair}, Tuple{Vararg{Pair}},
+function _combine(f::Union{AbstractVector{<:Pair}, 
+                           Tuple{Vararg{Pair}},
```
nalimilan: The old version was OK.

src/groupeddataframe/grouping.jl
```diff
@@ -901,29 +940,55 @@ colwise(fns::Union{AbstractVector, Tuple}, d::AbstractDataFrame) = [f(d[i]) for
 colwise(f, gd::GroupedDataFrame) = [colwise(f, g) for g in gd]
 
 """
-    by(d::AbstractDataFrame, keys, cols => f...; sort::Bool = false)
-    by(d::AbstractDataFrame, keys; (colname = cols => f)..., sort::Bool = false)
+    by(d::AbstractDataFrame, keys, cols => f; sort::Bool = false)
+    by(d::AbstractDataFrame, keys, [col1 => f1, col2 => f2], col3 => f3; sort::Bool = false)
+    by(d::AbstractDataFrame, keys, [:col1, :col2, :col3] .=> f; sort::Bool = false)
+    by(d::AbstractDataFrame, keys; sort::Bool = false, colname = cols => f...)
```
nalimilan: I preferred the old version with `sort` at the end. Also parens are needed since `...` doesn't apply to `f` but to the pair.

src/groupeddataframe/grouping.jl
```diff
@@ -901,29 +940,55 @@ colwise(fns::Union{AbstractVector, Tuple}, d::AbstractDataFrame) = [f(d[i]) for
 colwise(f, gd::GroupedDataFrame) = [colwise(f, g) for g in gd]
 
 """
-    by(d::AbstractDataFrame, keys, cols => f...; sort::Bool = false)
-    by(d::AbstractDataFrame, keys; (colname = cols => f)..., sort::Bool = false)
+    by(d::AbstractDataFrame, keys, cols => f; sort::Bool = false)
```
nalimilan: The signature is supposed to present supported methods, not to give particular examples (which are given at the end). Please use the same presentation strategy as the one I implemented for `combine` above.

src/groupeddataframe/grouping.jl
```diff
@@ -901,29 +940,55 @@ colwise(fns::Union{AbstractVector, Tuple}, d::AbstractDataFrame) = [f(d[i]) for
...
     by(d::AbstractDataFrame, keys, f; sort::Bool = false)
-    by(f, d::AbstractDataFrame, keys; sort::Bool = false)
+    by(f, keys, d::AbstractDataFrame)
 
 Split-apply-combine in one step: apply `f` to each grouping in `d`
 based on grouping columns `keys`, and return a `DataFrame`.
 
 `keys` can be either a single column index, or a vector thereof.
 
-If the last argument(s) consist(s) in one or more `cols => f` pair(s), or if
```
nalimilan: Please also use the same text as the one I wrote for `combine`.

src/groupeddataframe/grouping.jl
```diff
@@ -1037,12 +1114,15 @@ julia> by(df, :a, (:b, :c) => x -> (minb = minimum(x.b), sumc = sum(x.c)))
...
+
 by(f::Any, d::AbstractDataFrame, cols::Any; sort::Bool = false) =
     by(d, cols, f, sort = sort)
-by(d::AbstractDataFrame, cols::Any, f::Pair; sort::Bool = false) =
-    combine(f, groupby(d, cols, sort = sort))
-by(d::AbstractDataFrame, cols::Any, f::Pair...; sort::Bool = false) =
-    combine(f, groupby(d, cols, sort = sort))
+
+function by(d::AbstractDataFrame, cols::Any, f::Union{Pair, AbstractVector{<:Pair}}...; sort::Bool = false) 
+    vec_of_pairs = reduce(vcat, f)
```
nalimilan: `vec_of_pairs` takes almost as many characters as `reduce(vcat, f)`, so better use that expression directly in the line below, and switch the method definition to the short `=` form.

src/groupeddataframe/grouping.jl
pdeffebach: Added.

src/groupeddataframe/grouping.jl
pdeffebach: Added the same text for those bullets.

src/groupeddataframe/grouping.jl
pdeffebach: fixed.

src/groupeddataframe/grouping.jl
```diff
@@ -917,29 +955,54 @@ function _combine_with_first!(first::Union{AbstractDataFrame,
...
+  with the same columns and as many rows for each group as the rows returned for that group.
+
+As a special case, if a tuple or vector of pairs is passed as the first argument, each function
+is required to return a single value or vector, which will produce each a separate column.
+
+In all cases, the resulting data frame contains all the grouping columns in addition
+to those listed above. Column names are automatically generated when necessary: for functions
+operating on a single column and returning a single value or vector, the function name is
+appended to the input column name; for other functions, columns are called `x1`, `x2`
+and so on. The resulting data frame will be sorted if `sort=true` was passed to `by`
```
```suggestion
and so on. The resulting data frame will be sorted if `sort=true` is passed
```

src/groupeddataframe/grouping.jl
```diff
@@ -400,6 +404,34 @@ julia> combine(:c => sum, gd)
 │ 3   │ 3     │ 10    │
 │ 4   │ 4     │ 12    │
 
+julia> combine(gd, [:b, :c] => sum)
```
```suggestion
julia> combine(gd, [:b, :c] .=> sum)
```

src/groupeddataframe/grouping.jl
```diff
@@ -975,8 +1015,6 @@ When computing the `sum` or `mean` over floating point columns, results will be
 accurate than the standard [`sum`](@ref) function (which uses pairwise summation). Use
 `col => x -> sum(x)` to avoid the optimized method and use the slower, more accurate one.
 
-`by(d, cols, f)` is equivalent to `combine(f, groupby(d, cols))` and to the
```
nalimilan: Any reason to remove this?

pdeffebach: re-added.

pdeffebach:
Okay I fixed some outstanding things and I think this is ready to be merged

pdeffebach:
It looks like we added some weird behavior in a previous commit. by defining 

```
combine(gd) = combine(identity, gd)
```

Current behavior gives

```
julia> df = DataFrame(a=[1, 1, 2, 2, 2], b=1:5);

julia> gd = groupby(df, :a);

julia> combine(identity, gd) # same as combine(gd)
5×3 DataFrame
│ Row │ a     │ a_1   │ b     │
│ 1   │ 1     │ 1     │ 1     │
│ 2   │ 1     │ 1     │ 2     │
│ 3   │ 2     │ 2     │ 3     │
│ 4   │ 2     │ 2     │ 4     │
│ 5   │ 2     │ 2     │ 5     │
```

This doesn't seem right. In this PR we now have. Note that there won't be an extra `:a_1` column.

```
combine(gd) = DataFrame(gd)
```

I think that `combine(identity, gd)` *should* have the extra `:a_1` column, because it's doing what is described in the documentation. 

Basically I'm deleting the deprecations because behavior preserved by the deprecation was not correct.

bkamins:
> Basically I'm deleting the deprecations because behavior preserved by the deprecation was not correct.

The deprecation of `combine(df)` is needed because what you observe is the old behavior of `combine`. We decided to disallow `combine(df)` exactly because `combine(identity, gd)` (with an extra column) is doing what it should do. So it should not be replaced by `DataFrame(gd)` but rather throw an error (this was the plan).

I do not see why `combine(df)` should be the same as `DataFrame(df)`. Conceptually it rather could return a `DataFrame` consisting only of grouping columns with one row per group (but this is not very intuitive either so we planned to remove this option after the deprecation period).

pdeffebach:
Ah okay. I shouldn't have changed that, then. (I'm hoping it was Milan who told me to do this lol). I will revert and then we can merge.

bkamins:
Later we can discuss what behavior we want for `combine(gd)` after the deprecation period, but it should be a separate PR I think. Thank you.

src/groupeddataframe/grouping.jl
```diff
@@ -338,7 +338,8 @@ function combine(f::Any, gd::GroupedDataFrame)
     end
 end
 combine(gd::GroupedDataFrame, f::Any) = combine(f, gd)
-combine(gd::GroupedDataFrame, f::Pair...) = combine(f, gd)
```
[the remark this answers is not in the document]
nalimilan: That comment still applies.

src/groupeddataframe/grouping.jl
```diff
@@ -429,9 +461,13 @@ function combine(f::Any, gd::GroupedDataFrame)
...
 end
+
 combine(gd::GroupedDataFrame, f::Any) = combine(f, gd)
-combine(gd::GroupedDataFrame, f::Pair...) = combine(f, gd)
-combine(gd::GroupedDataFrame, f::Pair) = combine(f, gd)
+
+combine(gd::GroupedDataFrame, f::Union{Pair, AbstractVector{<:Pair}}...) =
+    combine(reduce(vcat, f), gd)
+
+combine(gd::GroupedDataFrame, f::NamedTuple) = combine(f, gd)
```
nalimilan: Sorry if I've asked before, but why is that method needed?

pdeffebach: It's not needed. The user can always splat the keyword arguments.

src/groupeddataframe/grouping.jl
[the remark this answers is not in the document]
pdeffebach: yes. It still works without this. It's deleted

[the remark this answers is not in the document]
pdeffebach: It turns out it is needed. `combine(f, (:c => sum, :d => mean))` breaks without it. 

I thought 

```
combine(gd::GroupedDataFrame, f::Union{Pair, AbstractVector{<:Pair}}...) 
```

would capture it, but it does not. If we wanted to avoid the dispatch we could add a method for `combine` which collects a tuple of pairs into a `Vector`.

pdeffebach:
Tests pass! Everything seems clean! Bogumil you had a hold placed on this until tests passed.

src/groupeddataframe/grouping.jl
```diff
@@ -315,26 +315,30 @@ function Base.map(f::Any, gd::GroupedDataFrame)
 end
 
 """
-    combine(gd::GroupedDataFrame, cols => f...)
+    combine(gd::GroupedDataFrame, (cols => f)...)
+    combine(gd::GroupedDataFrame, [cols1 => f1, cols2 => f2]...)
```
bkamins: This is a bit confusing (i.e. I am starting to read this PR and it is not clear for me what is the API). Typically we use a type signature. Do you mean here `::AbstractVector{<:Pair{Symbol, Callable}}`? Can multiple vectors be passed and can this vector be mixed with non vector `cols=>f` syntax? If for all question the answer is yes maybe write `combine(gd::GroupedDataFrame, args...)` and later explain what each entry of `args` can be?

bkamins: I can see below that we allow mixing, so probably the `args` approach would be cleanest.

src/groupeddataframe/grouping.jl
```diff
@@ -315,26 +315,30 @@ function Base.map(f::Any, gd::GroupedDataFrame)
...
+The last argument(s) in `combine` can be either:
 
-If the last argument is a callable `f`, it is passed a [`SubDataFrame`](@ref) view for each group,
-and the returned `DataFrame` then consists of the returned rows plus the grouping columns.
-Note that this second form is much slower than the first one due to type instability.
-A method is defined with `f` as the first argument, so do-block
-notation can be used.
+* One or several `cols => f` pairs, or vectors or tuples of such pairs (mixing is allowed). `cols`
+  must be a column name or index in `gd`, or a vector or tuple thereof. `f` must be callable.
+  If `cols` is a single column index, `f` is called with a `SubArray` view into that
```
bkamins: maybe: "`f` is passed a view into ..."?

src/groupeddataframe/grouping.jl
```diff
@@ -315,26 +315,30 @@ function Base.map(f::Any, gd::GroupedDataFrame)
...
+  column for each group; else, `f` is called with a named tuple holding `SubArray`
+  views into these columns.
+* A named tuple of `colname = cols => f` pairs or keyword arguments of such pairs,
```
bkamins: Is a single `NamedTuple` only allowed?

src/groupeddataframe/grouping.jl
```diff
@@ -315,26 +315,30 @@ function Base.map(f::Any, gd::GroupedDataFrame)
...
+  where `colname` indicates the name of the column to be created in the new `DataFrame`.
+  Pairs must obey the same rules as above.
+* A callable `f` taking a `SubDataFrame` view for each group. The returned `DataFrame` 
+  then consists of the returned rows plus the grouping columns.
+  Note that this form is much slower than the others due to type instability.
+
+A method is defined with `f` as the first argument, so do-block notation can be used.
+In that case `f` can also be a named tuple of pairs.
```
bkamins: why do we allow a `NamedTuple` of Pairs for this case?

src/groupeddataframe/grouping.jl
```diff
@@ -429,9 +461,11 @@ function combine(f::Any, gd::GroupedDataFrame)
...
 end
+
 combine(gd::GroupedDataFrame, f::Any) = combine(f, gd)
-combine(gd::GroupedDataFrame, f::Pair...) = combine(f, gd)
-combine(gd::GroupedDataFrame, f::Pair) = combine(f, gd)
+
+combine(gd::GroupedDataFrame, f::Union{Pair, AbstractVector{<:Pair}}...) =
+    combine(reduce(vcat, f), gd)
 
 function combine(gd::GroupedDataFrame; f...)
```
bkamins: It is not clear for me how you handle the case `combine(gd, (a=1=>length, b=2=>length))`.

src/groupeddataframe/grouping.jl
```diff
@@ -315,26 +315,30 @@ function Base.map(f::Any, gd::GroupedDataFrame)
...
+  where `colname` indicates the name of the column to be created in the new `DataFrame`.
+  Pairs must obey the same rules as above.
+* A callable `f` taking a `SubDataFrame` view for each group. The returned `DataFrame` 
+  then consists of the returned rows plus the grouping columns.
+  Note that this form is much slower than the others due to type instability.
+
+A method is defined with `f` as the first argument, so do-block notation can be used.
+In that case `f` can also be a named tuple of pairs.
 
 `f` can return a single value, a row or multiple rows. The type of the returned value
```
bkamins: The restriction is if we write `colname = cols=f` it must return a column. Right?

src/groupeddataframe/grouping.jl
bkamins: OK - I see this is not this case (it is `;` instead of `,` in the signature 😄)

bkamins:
Given the contract in the docstring the following methods should work assuming `gd` has column `:a`. Can you please make sure that these test pass (or update the documentation to reflect exactly what we accept). Thank you:
```
combine(gd, 1=>length, (1=>length,))
combine(gd, 1=>length, [1=>length,])
combine(gd, (1=>length,), [1=>length,])
combine(gd, [1=>length,], [1=>length,])
combine(gd, (1=>length,), (1=>length,))
combine(gd, 1=>length, (1=>length,), [1=>length,])
combine(gd, :a=>length, (1=>length,))
combine(gd, :a=>length, [1=>length,])
combine(gd, (:a=>length,), [1=>length,])
combine(gd, [:a=>length,], [1=>length,])
combine(gd, (:a=>length,), (1=>length,))
combine(gd, :a=>length, (1=>length,), [1=>length,])
combine(gd, 1=>length, ((1,2)=>length,))
combine(gd, 1=>length, [(1,2)=>length,])
combine(gd, (1=>length,), [(1,2)=>length,])
combine(gd, [1=>length,], [(1,2)=>length,])
combine(gd, (1=>length,), ((1,2)=>length,))
combine(gd, 1=>length, (1=>length,), [(1,2)=>length,])
```

src/groupeddataframe/grouping.jl
pdeffebach: I think `cols` can be either a `Symbol` or a `Tuple` of `Symbols`. Yes `f` should be `Callable`. But I thought Julia was no longer using the word `Callable`.

src/groupeddataframe/grouping.jl
pdeffebach: Changed the wording.

src/groupeddataframe/grouping.jl
pdeffebach: Okay this is hard for me to wrap my head around. 

1. `NamedTuple` for `combine(n::NamedTuple, gd)` works because we define the keyword argument method as follows:

```
function combine(gd::GroupedDataFrame; f...)
    if length(f) == 0
        Base.depwarn("combine(gd) is deprecated, use DataFrame(gd) instead", :combine)
        combine(identity, gd)
    else
        combine(values(f), gd)
    end
end
```

However now I see that because we splay the `f...` there for the keyword arguments, the following also works: 

```
combine(gd, (n = :b => first, ))
```

At least I think this is why it is behaving like that.

src/groupeddataframe/grouping.jl
bkamins: `Base.Callable` is defined still. But as I have written later probably it is best to write `agrs` and explain later the details.

src/groupeddataframe/grouping.jl
pdeffebach: For the `Pairs`, including named tuple stuff, you can only return a scalar or a vector. The vectors get stacked. All pairs must return something of the same length. 

Tuples of vectors do not get splatted. For instance `combine(gd, :b => x -> (x, x .+ 1))` does not create two columns.

src/groupeddataframe/grouping.jl
bkamins: Thanks for the explanation.

The core of my question is if we **want** to allow this. My intuitive understanding was that the form `combine(f, gd)` was intended only for the case where `f` is a function to allow `do` syntax (and if we did not want this syntax we would disallow putting `f` before `gd` as normally `gd` should be the first argument).

src/groupeddataframe/grouping.jl
bkamins: Thank you. This is my understanding, but I could not find where we specify this in the docstring.

bkamins:
A general comment - my "pushing" for precise docstring is guided by two reasons:
* the current design becomes very powerful, but this means that we get more corner cases, so the docstring should specify a precise contract
* all cases of the above contract should have appropriate tests (and it is hard to write good tests if we do not have a contract for the function lied out in the docstring - at least this is my experience :))

src/groupeddataframe/grouping.jl
pdeffebach: Update: the user does not need to splat the keyword arguments for this to work.  This is because the keyword arguments are splatted in the function call already.

pdeffebach:
This definitely does not pass the tests listed. Tuples don't play nice because they cannot be mixed. Scalars and vectors can be mixed, but we don't include tuples here. If you want to use a tuple, you can use it by itself and you can't use more than one tuple. 

The reason one tuple works is because of our splatting for our `combine(gd, f::Any)` function call. 

The reason you can't mix and match is because `vcat(1, (1, 2, 3))` returns a 2 element vector and does not promote to a vector of length 4. 

We could widen the type signature of 

```
combine(gd::GroupedDataFrame, f::Union{Pair, AbstractVector{<:Pair}}...) =
    combine(reduce(vcat, f), gd)
```

and expand upon `reduce(vcat, f)` if we want to. 

Thank you for being so thorough with this. It shows how important it is to finalize the contract before writing the code.

bkamins:
Actually I am even OK with the current functionality (as mixing tuples and vectors etc. is not something someone will want to do), but I would like to have it written down precisely in a docstring and then tested in two ways:
* that what needs to work works
* that what should not work throws an error 😄

pdeffebach:
Sounds good. 

There is some weird behavior that I just came across. I will file an issue after this is merged but here it is, from a test. 


```
combine(gd, (:b, :c) => x -> (b_sum=sum(x.b), c_sum=sum(x.c)))
```

This is weird behavior that is not documented. It's caused because when we pass a tuple of symbols as an input to a function, we actually 

1. Subset the data frame to just have the columns in the pair
2. Make a NamedTuple of those columns and pass it to the rest of the infrastructure

This is undocumented and potentially memory intensive. But can wait until later.

bkamins:
The current documentation says

> `f` is called with a named tuple holding `SubArray` views into these columns

in this case. I understand that you say that this is what happens but in the process a temporary `AbstractDataFrame` (presumably `SubDataFrame`) is created and this intermediate step is unnecessary?

pdeffebach:
Okay I understand. what that documentation means now. I trust that its efficiently implemented. I will add examples in the docs in a separate PR

src/groupeddataframe/grouping.jl
nalimilan: We can add `args...`, but having a relatively readable list of common cases sounds useful to me.

src/groupeddataframe/grouping.jl
nalimilan: Note that the docstring for `map` should be kept in sync.

src/groupeddataframe/grouping.jl
nalimilan: Yes, we probably don't need to allow named tuples here.

src/groupeddataframe/grouping.jl
nalimilan: That's mentioned below:
> As a special case, if a tuple or vector of pairs is passed as the first argument, each function
is required to return a single value or vector, which will produce each a separate column.

It's really hard to find a good way to organize all that stuff given how many possible combinations there are.

nalimilan:
> combine(gd, (:b, :c) => x -> (b_sum=sum(x.b), c_sum=sum(x.c)))

AFAICT this is completely expected, and it's implemented efficiently by passing a named tuple of views into the two columns. Am I missing something?

About mixing vectors and tuples, we don't necessarily need to allow that. It can always be added later anyway.

pdeffebach:
@bkamins here is my write up of the scenarios you outlined. 

```
combine(gd, :a=>length, (:a=>length,)) # bad! tested
combine(gd, :a=>length, [:a=>length,]) # good! tested
combine(gd, (:a=>length,), [:a=>length,]) # bad! tested
combine(gd, [:a=>length,], [:a=>length,]) # good! tested
combine(gd, (:a=>length,), (:a=>length,)) # bad! tested
combine(gd, :a=>length, (:a=>length,), [:a=>length,]) # bad! tested
combine(gd, :a=>length, (:a=>length,)) # bad! tested
combine(gd, :a=>length, [:a=>length,]) # good! tested
combine(gd, (:a=>length,), [:a=>length,]) # bad! tested
combine(gd, [:a=>length,], [:a=>length,]) # good! tested
combine(gd, (:a=>length,), (:a=>length,)) # bad! tested
combine(gd, :a=>length, (:a=>length,), [:a=>length,]) # bad! tested
combine(gd, :a=>length, ((:a,:b)=>length,)) # bad! tested
combine(gd, :a=>length, [(:a,:b)=>length,]) # bad, but for other reasons
combine(gd, (:a=>length,), [(:a,:b)=>length,]) # bad, but for other reasons
combine(gd, [:a=>length,], [(:a,:b)=>length,]) # bad, but for other reasons
combine(gd, (:a=>length,), ((:a,:b)=>length,)) # bad, but for other reasons
combine(gd, :a=>length, (:a=>length,), [(:a,:b)=>length,]) # bad, but for other reasons
```

There is some behavior that *should* work even without this PR, but doesn't. I will file a separate issue, or fix after I commit this.

bkamins:
Thank you for taking care to do such extensive tests! You have passed some duplicates (but it is ok). "bad, but for other reasons" is strange as something like this `combine(gd, :x2=>length, (:x2, :x3)=>length)` works.

pdeffebach:
Okay this PR broke that behavior and tests didn't catch it. I am searching for the fix.

pdeffebach:
Here is what happened: 

* With my change by collecting multiple separate arguments into an array first, `combine(gd, :b => first, (:b, :c) => length)` passes the following to `combine`:

```
f = Pair{Any,Function}[:b=>first, (:b, :c)=>length]
```

When we call `map` on `f`, *every argument* in `map` becomes `Pair{Any, Function}`, which messes with conditionals like 

```
        if agg isa AbstractAggregate && p isa Pair{<:Union{Symbol,Integer}}
```

in `_combine`. 

I fixed this with:

```
    res = map(f) do p
        # Narrow the type of p
        p = Pair(first(p), last(p))
```

pdeffebach:
I now have the naming working with my (somewhat hacky) solution. This solution seems more elegant than making the type of our `Pair` vector a small union of a bunch of different things. 

I also added some tests that catch what went wrong.

nalimilan:
I think you can just replace `p isa Pair{<:Union{Symbol,Integer}}` with `p isa Pair && first(p) isa Union{Integer,Symbol}`.

pdeffebach:
That's a good suggestion, and I have incorporated it. 

Things should be good now.

bkamins:
`Union{Integer,Symbol}` should be `ColumnIndex`.

pdeffebach:
Thanks for the catch. I've updated it.

pdeffebach:
Can this be merged even though some of the coverage tests didn't pass?

bkamins:
Coverage passes. The problem is that it needs to be merged with master again (sorry for inconvenience).
Also when you merge the CI failures on nightly will get fixed.

pdeffebach:
I think things are correct now?

bkamins:
Thank you for the update.

Let us wait for @nalimilan for a final review, as he is the "groupby" master 😄 (he is off for a few days, but we can go back to this PR very soon).

bkamins:
@nalimilan  - how do you see we should proceed with this (given 1.0 release plans?)

src/groupeddataframe/grouping.jl
```diff
@@ -673,19 +707,22 @@ function do_f(f, x...)
...
+        if agg isa AbstractAggregate && p isa Pair && first(p) isa ColumnIndex
             incol = gd.parent[!, first(p)]
             idx = gd.idx[gd.starts]
             outcol = agg(incol, gd)
             return idx, outcol
         else
             fun = do_f(last(p))
-            if p isa Pair{<:ColumnIndex}
+      
+             if p isa Pair && first(p) isa ColumnIndex
```
```suggestion
            if p isa Pair && first(p) isa ColumnIndex
```

nalimilan:
Sorry for the delay. There are still open discussions above. Can you comment on or close them as appropriate? At least the one about the `map` docstring still applies AFAICT.

pdeffebach:
Someone posted on discuss [here](https://discourse.julialang.org/t/bad-performance-of-group-by-of-dataframes/30061). I need to finish this PR

bkamins:
Would this PR avoid recompilation? (I have posted an answer how the problem on Discourse can be solved using API we have now)

pdeffebach:
No, but it would have avoided that complicated named tuple expression with repeated pairs to the same functions.

bkamins:
This PR might have to be dropped or significantly reviewed if we go for `:oldcol => f => :newcol` pattern as discussed with @nalimilan.

pdeffebach:
yeah i dropped the ball on this PR, and I apologize for that. 

This PR would fall into place really easily if we implemented #2016 and some infrastructure to handle this kind of thing. For example ideally 

```
by(df, :a, Between(:d, :m) => mean)
by(df, :a, [Between(:d, :m), :q] .=> mean)
```

Should both work via some generic code that is not unique to `by`. 

One benefit of Milan's suggestion to avoid named tuples is that it reduces the gymnastics with arguments, support for which were dropped from this PR because it would have added a lot of boiler-plate code for getting all the different argument formats to play nice.

nalimilan:
Note that I didn't really suggest dropping keyword arguments, rather adding another possible syntax.

bkamins:
> Note that I didn't really suggest dropping keyword arguments, rather adding another possible syntax.

This is how I understood this. My point is that it has to be rethought how broadcasting would come into play in `:oldcol => f => :newcol` syntax as quickly thinking about it ``@. `oldcols => f => newcols`` should be used (but I have not analyzed it in detail).

bkamins:
I am closing it as it is now implemented (except for broadcasting of `Between` etc. which is a separate issue)

bkamins:
@pdeffebach - thank you for raising this functionality. I think it helped a lot to improve the package.

PR #1622, opened 2018-12-08
Small dump fix

bkamins:
Additional `dump` ambiguity fix and improved `SubDataFrame` tests.

@nalimilan Afer this I think we can do the release.


bkamins:
OK - we can leave the discussion in #1623 for later (if we decide to change something there).

Issue #1623, opened 2018-12-08
split-apply-combine depwarns


bkamins:
Currently running `by` and similar prints depwarns on first run, e.g.:
```
julia> using DataFrames

julia> df = DataFrame(rand(4,5))
4×5 DataFrame
│ Row │ x1       │ x2       │ x3       │ x4       │ x5        │
│     │ Float64  │ Float64  │ Float64  │ Float64  │ Float64   │
├─────┼──────────┼──────────┼──────────┼──────────┼───────────┤
│ 1   │ 0.507518 │ 0.343065 │ 0.250251 │ 0.925915 │ 0.685284  │
│ 2   │ 0.194043 │ 0.223878 │ 0.119892 │ 0.113459 │ 0.205621  │
│ 3   │ 0.930018 │ 0.113958 │ 0.293055 │ 0.75793  │ 0.574463  │
│ 4   │ 0.694755 │ 0.907914 │ 0.61208  │ 0.345201 │ 0.0718849 │

julia> by(df, :x1, identity)
┌ Warning: Indexing into a return value of columns on SubDataFrame will return a view of column value
│   caller = _broadcast_getindex at broadcast.jl:540 [inlined]
└ @ Core .\broadcast.jl:540
┌ Warning: Indexing into a return value of columns on SubDataFrame will return a view of column value
│   caller = copyto_nonleaf!(::Array{DataType,1}, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1},Tuple{Base.OneTo{Int64}},typeof(eltype),Tuple{Base.Broadcast.Extruded{DataFrames.DataFrameColumns{SubDataFrame{Array{Int64,1}},AbstractArray{T,1} where T},Tuple{Bool},Tuple{Int64}}}}, ::Base.OneTo{Int64}, ::Int64, ::Int64) at broadcast.jl:540
└ @ Base.Broadcast .\broadcast.jl:540
┌ Warning: `sdf[colind]` will create a view of `parent(sdf)[colind]` in the future. Use sdf[:, colind]` to get a freshly allocated vector.
│   caller = append_rows!(::SubDataFrame{Array{Int64,1}}, ::NTuple{5,Array{Float64,1}}, ::Int64, ::NTuple{5,Symbol}) at grouping.jl:566
└ @ DataFrames C:\Users\bogum\.julia\packages\DataFrames\lsL3v\src\groupeddataframe\grouping.jl:566
[...]

PR #1626, opened 2018-12-09
WIP: add benchmarks

bkamins:
Following #1623. I will slowly add the tests. Now I have sketched what tests I think it would be good to have (this is a huge cartesian unfortunately - maybe we will have to thin it down).


bkamins:
I will open another PR when I am ready to do this.

PR #1627, opened 2018-12-09
Fix latex printing

bkamins:
This patch is mildly breaking but pretty relevant to Jupyter Notebook. It limits the output of LaTeX when printing `DataFrame`. Without it there are huge lags in Jupter Notebook for large `DataFrames`.


simonbyrne:
Won't this affect all io devices by default? Can we make it more specific so that it only applies to IJulia notebooks?

bkamins:
It is a fully general solution in line how `IO` properties are expected to be treated.

If you want to always print everything then set `:limit` to `false` in your `IOContext` like here in the tests https://github.com/JuliaData/DataFrames.jl/pull/1627/files#diff-be8288df383517cc1dabe74772b7535aR29.

Similarly if you want a specific truncation point set `:displaysize` in your `IOContext` and `:limit` to `true`.

simonbyrne:
Fair enough.

nalimilan:
AFAICT that's consistent with what we currently do for HTML, but doesn't it go against what we do for the standard two-argument `show` since https://github.com/JuliaData/DataFrames.jl/pull/1506?

bkamins:
Neither HTML nor LaTeX supported those advanced outputs AFAICT.
I would merge this (as it solves a critical problem with Jupyter Notebooks).
If someone is willing to implement all possible variants of `show for those targets it would be great, but it will be a bigger change so it would be good to have this PR till that happens.

nalimilan:
We can certainly merge this as a short-term measure, but I'm trying to figure out what's the best long-term behavior. The fundamental question to address is whether the default should be to print everything, or to limit output. At https://github.com/JuliaData/DataFrames.jl/pull/1506 IIRC we chose to limit output by default for the two-argument form, but not for the three-argument form. This differs from arrays in Base, which print everything unless `limit` is explicitly set. How do they behave in Jupyter?

bkamins:
1. I would say that 95% of use cases of LaTeX is path Jupyter to PDF (as it uses LaTeX internally).
2. What I have implemented is that LaTeX will contain the same data as HTML (so if you save Jupyter as PDF you get the same result - and I think it is important to keep it this way - except that in LaTeX we do not print column types - we could change this but this is also a separate PR in my opinion).
3. If you use other `show` with kwargs (`allrows` etc.) then we do not have custom HTML/LaTeX outuput methods defined at all. In Jupyter you get the dump of raw text from the console (I have checked to make sure).

This PR addresses point 2 (synchronizing current functionality between HTML and LaTeX) for .

What I say that we could implement point 3 for LaTeX and HTML, but it is a separate PR I think (adding new functionality for HTML and LaTeX `show` method).

nalimilan:
I agree the HTML and LaTeX methods need to be in sync. But how does Jupyter accommodate the fact that arrays are printed in full by default AFAICT? Why is there a problem only for DataFrames?

bkamins:
Arrays are not printed in full in Jupyter Notebook. They are printed like in REPL. See an example here:
image

bkamins:
This is the reason. In general I am not entirely clear where you see the problem.

nalimilan:
What worries me is that by default `show` for `Array` prints everything. So I don't understand why we need to use a different default for our functions.

bkamins:
1. IJulia essentially uses this function https://github.com/JuliaLang/IJulia.jl/blob/master/src/inline.jl#L31 for displaying; it sets `:limit` to `true`;
2. By defualut (and I understand that you mean `stdout` as default) `show` for array prints everything because `stdout` does not have `:limit` argument set;
3. However, in all `DataFrame` printing functions we have a rule that if `:limit` is missing we set it to `true` (and as far as I remember you have implemented this for `text/plain` 😄) to avoid what happens when you run `show(rand(1000,1000))`. We could change this rule and do not make `:limit` set to `true` by default, but would it improve anything? (it will not mater for IJulia as it sets `:limit` to `true` anyway - see the link above)

nalimilan:
Yes, I know I implemented a different behavior, that's why I wonder whether it's the right choice. It's good to hear that IJulia doesn't depend on the default. So let's merge this, and maybe change the default everywhere later. That wouldn't really "improve" anything, I just think the default should be consistent across Julia.

PR #1628, opened 2018-12-10
Add `if`s everywhere in describe

pdeffebach:
Currently `describe` calculates the standard deviation of all columns, even if you don't ask it to. 

This PR adds `if` statements around all the function calls so that it only calculates the values it needs to for reporting the descriptive statistics the user asks for. 

However the PR in its current form adds a lot of if statements, which may end up being costly. I doubt it, but will benchmark tomorrow.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -432,7 +432,29 @@ function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractV
     data[:variable] = names(df)
 
     # An array of Dicts for summary statistics
-    column_stats_dicts = [get_stats(col, stats) for col in columns(df)]
+    column_stats_dicts = map(columns(df)) do col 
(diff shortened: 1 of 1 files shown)


pdeffebach:
Added! relevant stats were `:nmissing`, `:first`, and `:last`

pdeffebach:
Any thoughts on making `eltype` a more informative name? Since its actually the eltype of the non-missing elements. 

I toyed with having it print `Int64⍰` but then the column type is a `String`, not `DataType`. 

:non_missing_eltype is too long and ugly, i'd say. 

An ideal fix would be to get `Int64⍰` merged upstream but that is unlikely to happen.

nalimilan:
I think `eltype` is fine, as the number of missing values is visible from another column by default (`0` vs `nothing` tells you whether there can be missing values). But I agree the ideal fix would be to improve printing. We could change DataFrames printing to transform `Union{T, Missing}` into `T⍰` if we want; not sure whether it's a good idea or not. It should be rarely used outside of `describe` anyway.

PR #1630, opened 2018-12-10
improve dropmissing and deleterows!

bkamins:
I opened a new PR instead of #1618 as there were too many changes and this way is cleaner.

What it changes:
* add `disallowmissing` kwarg for `dropmissing` and `dropmissing!`
* add documentation for `deletecols!` and `deleterows!`
* `deleterows!` manipulates columns "in-place" (I did not do not in-place modification as it can be achieved by `getindex` so it would duplicate)
* `deleterows!` for `SubDataFrame` has a nicer error as it could be accidentally called

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -658,10 +658,18 @@ julia> dropmissing(df, [:x, :y])
 ```
 
 """
-dropmissing(df::AbstractDataFrame,
-            cols::Union{Integer, Symbol, AbstractVector}=1:size(df, 2);
-            disallowmissing::Bool=false) =
-    dropmissing!(copy(df), cols, disallowmissing=disallowmissing)
+function dropmissing(df::AbstractDataFrame,
+                     cols::Union{Integer, Symbol, AbstractVector}=1:size(df, 2);
+                     disallowmissing::Bool=false)
+    newdf = df[completecases(df, cols), :]
+    if disallowmissing
+        disallowmissing!(newdf, cols)
+    else
+        Base.depwarn("dropmissing will change eltype of cols to disallow missing by default. " *
+                     "Use dropmissing(df, cols, disallowmissing=false) to retain missing.", :dropmissing)
+    end
+    newdf
+end
 
 """
     dropmissing!(df::AbstractDataFrame; disallowmissing::Bool=false)
@@ -743,12 +751,12 @@ julia> df3
 function dropmissing!(df::AbstractDataFrame,
                       cols::Union{Integer, Symbol, AbstractVector}=1:size(df, 2);
                       disallowmissing::Bool=false)
-    deleterows!(df, (!).completecases(df, cols))
+    deleterows!(df, (!).(completecases(df, cols)))
     if disallowmissing
         disallowmissing!(df, cols)
     else
         Base.depwarn("dropmissing! will change eltype of cols to disallow missing by default. " *
-                     "Use dropmissing(df, cols, disallowmissing=false) to retain missing.", :dropmissing!)
+                     "Use dropmissing!(df, cols, disallowmissing=false) to retain missing.", :dropmissing!)
     end
     df
 end
@@ -890,10 +898,10 @@ end
 nonunique(df::AbstractDataFrame, cols::Union{Integer, Symbol}) = nonunique(df[[cols]])
 nonunique(df::AbstractDataFrame, cols::Any) = nonunique(df[cols])
 
-unique!(df::AbstractDataFrame) = deleterows!(df, findall(nonunique(df)))
-unique!(df::AbstractDataFrame, cols::AbstractVector) =
+Base.unique!(df::AbstractDataFrame) = deleterows!(df, findall(nonunique(df)))
+Base.unique!(df::AbstractDataFrame, cols::AbstractVector) =
     deleterows!(df, findall(nonunique(df, cols)))
-unique!(df::AbstractDataFrame, cols::Union{Integer, Symbol, Colon}) =
+Base.unique!(df::AbstractDataFrame, cols::Union{Integer, Symbol, Colon}) =
     deleterows!(df, findall(nonunique(df, cols)))
 
 # Unique rows of an AbstractDataFrame.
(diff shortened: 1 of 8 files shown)


bkamins:
and minor changes: remove conditional import of `unique!` as we only support Julia that has it and added `find` and `find!` to documentation

bkamins:
We have to wait till https://github.com/JuliaData/CategoricalArrays.jl/pull/176 is implemented before moving forward.

src/dataframe/dataframe.jl
```diff
@@ -763,6 +763,45 @@ end
...
+    deletecols!(df::DataFrame, ind)
+
+Delete columns from a data frame in place.
+
+
+### Arguments
+
+* `df` : the DataFrame from which we want to remove a column
+
+* `ind` : a position from which we want to remove a column
```
nalimilan: "position" makes it sound like it can only be an integer, but it can also be a symbol (the signature should say so).

BTW, I think we should stop having separate Arguments and Result sections when the function is simple (as recommended in the Julia manual). Here a simple sentence or two would be as clear.

src/dataframe/dataframe.jl
```diff
@@ -777,36 +816,58 @@ end
...
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 4     │
+│ 2   │ 3     │ 6     │
+```
+
+"""
+function deleterows!(df::DataFrame, ind)
+    if !isempty(ind) && size(df, 2) == 0
+        throw(BoundsError(summary(df), ind))
```
nalimilan: It's weird to pass `summary` here: we're not trying to access the summary string.

test/subdataframe.jl
```diff
@@ -124,4 +124,10 @@ module TestSubDataFrame
                                     y: Array{Float64}((5,)) [2.0, 3.0, 4.0, 5.0, 6.0]
                                   """
     end
+
+    @testset "deleterows!" begin
+        y = collect(1.0:10.0)
```
nalimilan: We automatically collect ranges already.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -645,18 +658,29 @@ julia> dropmissing(df, [:x, :y])
...
-    deleterows!(copy(df), findall(!, completecases(df, cols)))
+function dropmissing(df::AbstractDataFrame,
+                     cols::Union{Integer, Symbol, AbstractVector}=1:size(df, 2);
+                     disallowmissing::Bool=false)
+    newdf = df[completecases(df, cols), :]
+    if disallowmissing
+        disallowmissing!(newdf, cols)
+    else
+        Base.depwarn("dropmissing will change eltype of cols to disallow missing by default. " *
+                     "Use dropmissing(df, cols, disallowmissing=false) to retain missing.", :dropmissing)
```
```suggestion
                     "Use dropmissing(df, cols, disallowmissing=false) to allow for missing values.", :dropmissing)
```

test/subdataframe.jl
bkamins: OK. I fixed all usages of `collect` also above.

src/dataframe/dataframe.jl
bkamins: fixed. I also a fixed a bug in `deletecols!` as it allowed duplicate column identifier to be passed.

src/dataframe/dataframe.jl
nalimilan: If the output is unreadable, we should fix the printing of `BoundsError` in Julia. There's currently a special case for `AbstractArray`, but we should probably use `summary` if it's defined for that type, or at least set `limit=true`.

src/dataframe/dataframe.jl
```diff
@@ -763,8 +763,47 @@ end
 ##
 ##############################################################################
 
+"""
+    deletecols!(df::DataFrame, ind)
+
+Delete columns specified by `ind` from a `DataFrame` df in place.
```
nalimilan: Missing backticks. I'd also just say "and return it".

src/dataframe/dataframe.jl
```diff
@@ -763,8 +763,47 @@ end
...
+│ 3   │ 6     │
+```
+
+"""
 function deletecols!(df::DataFrame, inds::Vector{Int})
-    for ind in sort(inds, rev = true)
+    sorted_inds = sort(inds, rev=true)
+    for i in 2:length(sorted_inds)
+        if sorted_inds[i] == sorted_inds[i-1]
+            throw(ArgumentError("Duplicate values in inds found at positions" *
```
nalimilan: Why not call `unique!` instead?

src/dataframe/dataframe.jl
bkamins: I will change it to `ArgumentError` for now. We can use `BoundsError` later as it is not sure if thus change to Base would be backported and when

bkamins: Or to `throw(BoundsError())` - we use similar approach above in the same file. This way we will not change error type in the future when we can print it nicely.

src/dataframe/dataframe.jl
nalimilan: The performance impact should be absolutely negligible, and it makes things simpler for us and less annoying for users.

bkamins: But do you mean that you want to allow duplicates and drop them?
I would prefer to throw an error on duplicate.

bkamins: `deleteat!` in Base throws an error on duplicates (and thus so does `deleterows!`). So I implemented this to be consistent with that function. I can drop it and document that we ignore duplicates if you prefer.

nalimilan: OK, that's fine.

test/subdataframe.jl
```diff
@@ -96,8 +96,8 @@ module TestSubDataFrame
     end
 
     @testset "getproperty, setproperty! and propertynames" begin
-        x = 1:10
-        y = 1.0:10.0
+        x = collect(1:10)
```
nalimilan: Is this actually needed? In my tests this doesn't make any difference.

bkamins: It does - as below `x` is mutated and we do `@test x == [1; 1:5; 7:10]` several lines below. The same for `y`.

PR #1632, opened 2018-12-12
Show group values when printing grouped dataframe

pdeffebach:
In reference to #1539 this PR now prints the values of the grouping column whenever it prints a grouped dataframe.


pdeffebach:
Now the printing looks like this (note the `:d => 1`) etc. 

```
df = DataFrame(a = repeat(1:4, outer = 5), b = randn(20), c = randn(20) .+ 1)
g = groupby(df, :a)
julia> g
GroupedDataFrame{DataFrame} with 4 groups based on key: :d
First Group: 5 rows
:d => 1
│ Row │ a     │ b        │ c         │ d     │
│     │ Int64 │ Float64  │ Float64   │ Int64 │
├─────┼───────┼──────────┼───────────┼───────┤
│ 1   │ 1     │ 0.365603 │ -0.164928 │ 1     │
│ 2   │ 1     │ 1.78441  │ 1.67254   │ 1     │
│ 3   │ 1     │ -1.4079  │ 0.0170639 │ 1     │
│ 4   │ 1     │ 0.421552 │ 1.20265   │ 1     │
│ 5   │ 1     │ -1.12468 │ 0.783903  │ 1     │
First Group: 5 rows
:d => 4
│ Row │ a     │ b         │ c        │ d     │
│     │ Int64 │ Float64   │ Float64  │ Int64 │
├─────┼───────┼───────────┼──────────┼───────┤
│ 1   │ 4     │ 0.244776  │ 1.30593  │ 4     │
│ 2   │ 4     │ 1.49806   │ 0.43346  │ 4     │
│ 3   │ 4     │ -0.372229 │ 1.52525  │ 4     │
│ 4   │ 4     │ 0.845078  │ 1.19766  │ 4     │
│ 5   │ 4     │ -0.668419 │ 0.286902 │ 4     │
```

nalimilan:
I'd print this on the same line as the number of rows, maybe like this: `First Group (5 rows): col=value` (with a comma between multiple keys).

You'll also need to update tests.

pdeffebach:
Let me know if you like this behavior (though the code can change obviously) and I will add tests

src/groupeddataframe/show.jl
```diff
@@ -6,31 +6,51 @@ function Base.show(io::IO, gd::GroupedDataFrame;
...
             rows = nrows > 1 ? "rows" : "row"
-            print(io, "\nGroup $i: $nrows $rows")
+
+            identified_groups = [':' * string(parent_names[col], " = ", 
+                                 first(gd[i][col])) for col in gd.cols]
+
+            print(io, "\nGroup $i ($nrows $rows): ")            
+            join(io, identified_groups, ", ", " and ")
+            
             show(io, gd[i], summary=false,
```
nalimilan: Maybe we shouldn't print the grouping columns, since they're listed above?

pdeffebach: I think it reads well.

src/groupeddataframe/show.jl
```diff
@@ -6,31 +6,51 @@ function Base.show(io::IO, gd::GroupedDataFrame;
...
     groupstr = N > 1 ? "groups" : "group"
     summary && print(io, "$(typeof(gd)) with $N $groupstr based on $keystr: $keys")
     if allgroups
         for i = 1:N
             nrows = size(gd[i], 1)
             rows = nrows > 1 ? "rows" : "row"
-            print(io, "\nGroup $i: $nrows $rows")
+
+            identified_groups = [':' * string(parent_names[col], " = ", 
+                                 first(gd[i][col])) for col in gd.cols]
```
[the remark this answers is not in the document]
pdeffebach: I couldn't find any examples of the right indentation for constructors. So I added more line breaks? Let me know the course of action.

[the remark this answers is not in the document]
nalimilan: I'm not sure there's a clear rule for comprehensions, but here `first` is inside `string`, so it should definitely not be aligned with the bracket. I'd do this:
```suggestion
             identified_groups = [':' * string(parent_names[col], " = ", first(gd[i][col]))
                                  for col in gd.cols]
```

src/groupeddataframe/show.jl
nalimilan: As you like. But just to make sure we're talking about the same thing: I was suggesting we could drop the grouping columns from the data frame we print below, since all values are equal within a given group.

docs/src/lib/functions.md
```diff
@@ -1,54 +0,0 @@
-```@meta
```
nalimilan: This shouldn't be removed.

pdeffebach:
Sorry about deleting that file, i did it on accident and I thought my `git reset ... HEAD` worked.

We should show the grouping columns, so that `gd[i][1]` behaves according to what the user sees in printing.

pdeffebach:
Okay I think this is ready to be merged.

src/groupeddataframe/show.jl
```diff
@@ -6,31 +6,51 @@ function Base.show(io::IO, gd::GroupedDataFrame;
...
     keystr = length(gd.cols) > 1 ? "keys" : "key"
     groupstr = N > 1 ? "groups" : "group"
     summary && print(io, "$(typeof(gd)) with $N $groupstr based on $keystr: $keys")
     if allgroups
         for i = 1:N
             nrows = size(gd[i], 1)
             rows = nrows > 1 ? "rows" : "row"
-            print(io, "\nGroup $i: $nrows $rows")
+
+            identified_groups = [':' * string(parent_names[col], " = ", first(gd[i][col]))
```
nalimilan: I've just realized that when the grouping column is a string or symbol, it's printed as e.g. `:x = a` rather than `:x = "a"` or `:x = :a`. I guess we should change this (using `repr` or `ourshowcompact`)?

nalimilan: https://github.com/JuliaData/DataFrames.jl/pull/1684

PR #1637, opened 2018-12-14
DataFrameRow: fix iterate, ndims, iterate

bkamins:
Clean-up of `DataFrameRow` API.

Notice that `collect(dfr)` will have `eltype` set to `Any` (this is the standard mechanics), while `Vector(dfr)` does `promote_type` (and possibly makes a conversion - this is a behavior inherited from eralier implementation of `convert` to `Array` for `DataFrame`).

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -81,7 +81,7 @@ Base.size(r::DataFrameRow) = (size(parent(r), 2),)
 Base.size(r::DataFrameRow, i) = size(r)[i]
 Base.length(r::DataFrameRow) = size(parent(r), 2)
 Base.ndims(r::DataFrameRow) = 1
-Base.ndims(::Type{DataFrameRow{T}}) where {T<:AbstractDataFrame} = 1
+Base.ndims(::Type{<:DataFrameRow}) = 1
 
 Base.lastindex(r::DataFrameRow) = size(parent(r), 2)
 
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -91,6 +91,8 @@ function Base.iterate(r::DataFrameRow, st)
     return (r[st], st + 1)
 end
 
+Base.eltype(r::DataFrameRow) = reduce(typejoin, eltypes(parent(r)))
+
 Base.convert(::Type{Array}, dfr::DataFrameRow) =
     [dfr[i] for j in 1:1, i in 1:length(dfr)]
 Base.convert(::Type{Vector}, dfr::DataFrameRow) =
@@ -99,7 +101,6 @@ Base.convert(::Type{Vector{T}}, dfr::DataFrameRow) where T =
     T[dfr[i] for i in 1:length(dfr)]
 Base.Vector(dfr::DataFrameRow) = convert(Vector, dfr)
 Base.Vector{T}(dfr::DataFrameRow) where T = convert(Vector{T}, dfr)
-
 
 Base.keys(r::DataFrameRow) = names(parent(r))
 Base.values(r::DataFrameRow) = ntuple(col -> parent(r)[col][row(r)], length(r))
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -80,15 +80,11 @@ index(r::DataFrameRow) = index(parent(r))
 Base.size(r::DataFrameRow) = (size(parent(r), 2),)
 Base.size(r::DataFrameRow, i) = size(r)[i]
 Base.length(r::DataFrameRow) = size(parent(r), 2)
+Base.ndims(r::DataFrameRow) = 1
 
-Compat.lastindex(r::DataFrameRow) = size(parent(r), 2)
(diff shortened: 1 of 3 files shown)


test/dataframerow.jl
```diff
@@ -116,4 +122,16 @@ module TestDataFrameRow
                    c=["A", "B", "C", "A", "B", missing])
     @test copy(DataFrameRow(df, 1)) == (a = 1, b = 2.0, c = "A")
     @test isequal(copy(DataFrameRow(df, 2)), (a = 2, b = missing, c = "B"))
+
+    # iteration and collect
+    ref = ["a", "b", "c"]
+    df = DataFrame(permutedims(ref))
+    dfr = df[1, :]
+    @test collect(dfr) == ref
```
nalimilan: Also test the type? When you say the eltype is `Any`, I guess this applies only to heterogeneous column types, right? It would be useful to cover that case too.

(BTW, `Vector` doesn't promote AFAICT, as it uses a comprehension.)

bkamins: I will fix it and test it. Actually it should be narrowest type (as in comprehensions).

What is problematic (and confused me) that in `Matrix(df)` we use `T = reduce(promote_type, eltypes(df))` and probably we should use `T = reduce(typejoin, eltypes(df))`. Should I fix it?

bkamins:
I also needed to add one more `ndims` method

test/dataframerow.jl
nalimilan: See discussion at https://github.com/JuliaData/DataFrames.jl/pull/1641. Maybe we should use promotion here too.

src/dataframerow/dataframerow.jl
```diff
@@ -80,21 +80,20 @@ index(r::DataFrameRow) = index(parent(r))
...
-    iterate(r, 1)
-end
+Base.iterate(r::DataFrameRow) = iterate(r, 1)
 
 function Base.iterate(r::DataFrameRow, st)
     st > length(r) && return nothing
     return (r[st], st + 1)
 end
 
+Base.eltype(r::DataFrameRow) = reduce(typejoin, eltypes(parent(r)))
```
nalimilan: Should we implement this function? It should be about as costly to compute as extracting the value themselves, so I wonder whether it won't make performance worse. I guess it depends in what cases `eltype` is used by Julia.

bkamins: `eltype` is used by Julia in collect to get type information of the target container. It is a part of *iterable* interface specification. Without it vanilla `collect(dfr)` returns a vector of type `Any`.

I agree it will slow things down, but then, by defalut `DataFrameRow` is seen by Julia as having type `Any` as a container (this was my original implementation produced `Vector{Any}`).

The standard fallback method is `eltype(::Type) = Any` in array.jl in Base.

nalimilan: > Without it vanilla `collect(dfr)` returns a vector of type `Any`.

Does it? I'd expect it to widen the eltype as needed using `promote_typejoin`. Maybe we need to define `Base.IteratorEltype(::DataFrameRow) = Base.EltypeUnknown()`.

bkamins: OK. The `Base.IteratorEltype(::DataFrameRow) = Base.EltypeUnknown()` is an alternative good solution. I will use it.

src/dataframerow/dataframerow.jl
```diff
@@ -80,21 +80,20 @@ index(r::DataFrameRow) = index(parent(r))
 Base.size(r::DataFrameRow) = (size(parent(r), 2),)
 Base.size(r::DataFrameRow, i) = size(r)[i]
 Base.length(r::DataFrameRow) = size(parent(r), 2)
+Base.ndims(r::DataFrameRow) = 1
+Base.ndims(::Type{DataFrameRow{T}}) where {T<:AbstractDataFrame} = 1
```
```suggestion
Base.ndims(::Type{<:DataFrameRow}) = 1
```

test/dataframerow.jl
bkamins: As discussed 😄 in #1641, if someone wants a promotion to `T` one should use `Vector{T}(dfr)` in my opinion.

nalimilan: So should we use `promote_type` here too for consistency with #1641?

bkamins: I thought that this was exactly the point we wanted to have a difference here.

Essentially this is the same as:

```
julia> collect((1, 1.0, true))
3-element Array{Real,1}:
    1
    1.0
 true

julia> [v for v in (1, 1.0, true)]
3-element Array{Real,1}:
    1
    1.0
 true
```

But I am not too attached to what we have now. But it we want to do the promotion here we should define `eltype` (and this is exactly what I have removed some 2 commits ago 😄).

nalimilan: Maybe we should keep `collect` as-is for consistency with tuples, but have `convert` use promotion to be consistent with data frames?

bkamins: This is what we do. But it is implemented in https://github.com/JuliaData/DataFrames.jl/pull/1640.

PR #1639, opened 2018-12-14
Add summary option to HTML and LaTeX show methods

juliohm:
This PR adds the option to not print the number of rows and columns in HTML and LaTeX output. This is useful in custom DataFrame types that have their own headers.

### DIFF
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -92,7 +92,7 @@ function html_escape(cell::AbstractString)
     return cell
 end
 
-function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary=true)
+function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary::Bool=true)
     cnames = _names(df)
     write(io, "<table class=\"data-frame\">")
     write(io, "<thead>")
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -92,7 +92,7 @@ function html_escape(cell::AbstractString)
     return cell
 end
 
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/io.jl
```diff
@@ -92,7 +92,7 @@ function html_escape(cell::AbstractString)
     return cell
 end
 
-function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame)
+function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary=true)
```
```suggestion
function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary::Bool=true)
```

src/abstractdataframe/io.jl
```diff
@@ -165,7 +167,7 @@ function latex_escape(cell::AbstractString)
     replace(cell, ['\\','~','#','$','%','&','_','^','{','}']=>latex_char_escape)
 end
 
-function Base.show(io::IO, ::MIME"text/latex", df::AbstractDataFrame)
+function Base.show(io::IO, ::MIME"text/latex", df::AbstractDataFrame; summary=true)
```
nalimilan: Sorry, I thought we printed the summary in this method too. Since we don't, there's no point in taking that argument.

juliohm:
Done @nalimilan , please let me know if something else is missing.

juliohm:
Thank you. Appreciate if this is merged and tagged.

bkamins:
OK - I have merged it. Can you use master (or this specific commit) for now as I am not sure when we will tag a new release?
There are significant breaking changes already merged sine 0.15.2 tag so we need to go for 0.16.

juliohm:
It is not a big problem for now as it is just a printing issue in my
package. I will wait for a new tag. Thank you.

PR #1640, opened 2018-12-14
deprecate convert(Array, ::DataFrameRow)

bkamins:
I propose to deprecate `convert(Array, ::DataFrameRow)` as it is easy enough to get this functionality with other means and it is not consistent with anything I am aware of since we treat `DataFrameRow` as 1-dimensional object.


bkamins:
This goes through OK, but for some strange reasons coverage dropped.

PR #1641, opened 2018-12-14
better AbstractDataFrame conversion

bkamins:
1. Make conversion to Matrix use `typejoin` not `propote_type`
2. Remove internal function `promote_col_type` that is not used anywhere (@nalimilan - do you know its use?)
3. Deprecate `convert(Arraty, ::AbstractDataFrame)` as I do not understand its value.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -800,14 +800,8 @@ julia> df
 Base.filter!(f, df::AbstractDataFrame) =
     deleterows!(df, findall(collect(!f(r)::Bool for r in eachrow(df))))
 
-function Base.convert(::Type{Array}, df::AbstractDataFrame)
-    convert(Matrix, df)
-end
 function Base.convert(::Type{Matrix}, df::AbstractDataFrame)
-    T = reduce(promote_type, eltypes(df))
-    convert(Matrix{T}, df)
-end
-function Base.convert(::Type{Array{T}}, df::AbstractDataFrame) where T
+    T = reduce(typejoin, eltypes(df))
     convert(Matrix{T}, df)
 end
 function Base.convert(::Type{Matrix{T}}, df::AbstractDataFrame) where T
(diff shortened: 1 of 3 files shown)


nalimilan:
Thanks. `promote_col` isn't used since we switched to `Tables.allocatecolumn`, but I forgot to remove it.

I'm not really convinced we should remove `convert(Array, ::AbstractDataFrame)`. It's not very useful, but it has a clear meaning, so why not support it? For example, you can do `convert(Array{Float64}, 1:3)` even though it's redundant with `convert(Vector{Float64}, 1:3)`.

nalimilan:
I forgot to discuss the `typejoin` change. That's a difficult question, but `promote_type` could make more sense since you typically want a homogeneous matrix when possible. In practice people should mostly use this when all columns have the same type, but if e.g. one column is `Int32` and the other `Union{Int64,Missing}`, `typejoin` will return `Any`, which is really suboptimal.

bkamins:
Regarding the `Array` conversion - the problem I have is that I do not feel it has a clear meaning. I prefer explicit `Matrix` here and `Vector` in `DataFrameRow`.

See that e.g. for `DataFrameRow` we have `covert(Array, dfr)` return a `1xn` matrix which was completely unintuitive (but shows that actually it is not obvious what conversion to `Array` should produce).

Similarly we do not have e.g. `covert(AbstractArray, ::AbstractDataFrame)`.

I will not argue very strongly to remove it as it is a matter of taste (and really I do not expect anyone uses this conversion) but I feel it is redundant.

nalimilan:
> See that e.g. for `DataFrameRow` we have `covert(Array, dfr)` return a `1xn` matrix which was completely unintuitive (but shows that actually it is not obvious what conversion to `Array` should produce).

Indeed this one is a bit problematic.

bkamins:
Anyway - I wanted to put something on the table. So the proposal is to deprecate conversion to `Array`.

If you prefer to keep it I we can drop the deprecation here but only fix the `DataFrameRow` conversion to return a `Vector` (it is in other PR).

bkamins:
Regarding:

> I forgot to discuss the `typejoin` change.

My thinking was that `Matrix(df)` should not perform conversions. If the user wants to get a specific type `T` then `Matrix{T}(df)` should be use.

Note that:
```
julia> typejoin(Int, Union{Int, Missing})
Union{Missing, Int64}
```
so actually the problem you describe would happen in the following case:
```
julia> typejoin(Union{Int64, Missing}, Int32)
Any
```
but i would say it is extremely rare (and if someone mixes types this means that it is an advanced user knowing how to use `Matrix{T}(df)`).

bkamins:
@nalimilan: also notice that if we go "your way" then in order to avoid conversions you have to write `Matrix{Any}(df)` or an appropriate supertype, but I guess that for novices this will be more complex to guess that this is possible when they do not want to convert. While writing e.g. `Matrix{Float64}(df)` is an IMHO obvious way, for someone knowing Base semantics, to force a conversion.

nalimilan:
I'm not really convinced:
- Situations where column types are different should be rare, but that's not an argument for either behavior, since the choice only applies to these situations.
- In both cases, `Matrix{T}(df)` can/needs to be used to get a different behavior.

Having mixed types in a data frame isn't something reserved to advanced users IMO. For example CSV.jl could return an `Int` column and a `Float64` column just because the CSV file is written that way, and a user could want to convert that to a homogeneous matrix. I think it makes much more sense to return a `Matrix{Float64}` than a `Matrix{Real}` in that case. That's even worse if one has a `Vector{String}` column and a `CategoricalArray{String}` column: in that case we really want a `Matrix{String}`, not a `Matrix{Any}` with some elements `CategoricalString`.

Also, why would somebody want to build a matrix from a data frame without any conversion? I can't imagine a situation where this would be useful. And I don't see any reason to think "basic" users would need this more than homogeneous matrices.

bkamins:
OK - I will revert it here and change `Vector(::DataFrameRow)` to be consistent with this.

PR #1644, opened 2018-12-15
Add parentindices and parent

bkamins:
In this PR:
* add `parent` to `AbstractDataFrame` as identity (similarly to `parent` for arrays that are not views)
* add `parentindices` to `AbstractDataFrame`, `SubDataFrame` and `DataFrameRow` - this is a standard function and we can avoid users calling `DataFrames.row` or `DataFrames.rows`; see e.g. https://discourse.julialang.org/t/dataframerow-row-number/18692
* remove docstrings for `parent` as I think they are not needed (it is clear that `SubDataFrame` and `DataFrameRow` are views)


bkamins:
Good point with `axes`, I will change it. I wonder why Based does not use it?

Issue #1645, opened 2018-12-16
`setindex!`/`broadcast!` design


bkamins:
Here is a review of the current API (#1571 not included, but accounted for in the comments). I have added *decision* fields - sometimes I am pretty sure what to do. Sometimes it is a question. The biggest question I have is if we actually want to go for support of broadcasting when LHS is a `DataFrame` at all. Maybe it is OK to leave the current mechanics we have as a `DataFrame` is not an `AbstractArray`. I am not really sure what is best. The benefit of adding broadcasting is that we are more consistent with Base in terms of notation (now when `setindex!` is called broadcasting is done implicitly).

`setindex!(df::DataFrame, v::AbstractVector, col_ind::ColumnIndex)`

    - set `df[col_ind]` to `v` in place
    - will add a column if it is a new name or `ncol(df)+1` index
    - decision: I really do not like allowing `ncol(df)+1` index
      and I would deprecate it. This is a work for `hcat!` which should
      be exported.

`setindex!(df::DataFrame, v, col_ind::ColumnIndex)`

    - set `df[col_ind]` to `[v]` if `ncol(df)==0`
    - set `df[col_ind]` to `fill(v, nrow(df))` if `ncol(df)>0`
    - decision: what do we do when `ncol(df)>0` but `nrow(df)==0`?
    - decision: do we deprecate it and replace with `broadcast!`?

`setindex!(df::DataFrame, new_df::DataFrame, col_inds::AbstractVector{Bool})`
`setindex!(df::DataFrame, new_df::DataFrame, col_inds::AbstractVector{<:ColumnIndex})`

    - insert columns `col_inds` from `new_df` to `df` in place as `col_inds`
    - handling of `Bool` and `Integer` indexing is weird, see below
    - `new_df` uses integer indexing left-to-right; no column name matching is performed
    - decision: fix `Bool` and `Integer` indexing to be strict
    - decision: do we want to allow `NamedTuple` and `AbstractDict` of vectors also here?
    - decision: do we want to allow `AbstractMatrix` here?
    - decision: do we want to do column name matching

`setindex!(df::DataFrame, v::AbstractVector, col_inds::AbstractVector{Bool})`
`setindex!(df::DataFrame, val::Any, col_inds::AbstractVector{Bool})`
`setindex!(df::DataFrame, val::Any, col_inds::AbstractVector{<:ColumnIndex})`

    - `copy` of `v` is always performed if it is a `vector`
[...]

Issue #1599, opened 2018-11-14
DataFrame constructors


bkamins:
I am planning a small cleanup of `DataFrame` constructors and one significant change I am considering is to add a constructor for `DataFrame(::AbstractVector{Pair{Symbol, T}};makeunique) where T<:AbstractVector` or `DataFrame(::AbstractVector{Pair{Symbol}};makeunique)` (I am not 100% sure which would be better - in the second we would cast the second element of pair to `AbstractVector` inside the constructor).

This change would complement the existing constructors:
* `DataFrame(pairs::Pair{Symbol,<:Any}...; makeunique)`
* `DataFrame(; kwargs...)` (actually this constructor would be made more efficient as we would avoid splatting inside it)

and would work nicely with the fact that now `eachcol` returns such an `AbstractVector`.

Similarly - do we thing that a constructor accepting `AbstractVector{<:DataFrameRow}` would be good? (this would complement `eachrow` and is similar to an `AbstractVector` of `NamedTuple`s)

CC @nalimilan

nalimilan:
Makes sense, but couldn't the Tables.jl default constructor handle this automatically? Cc: @quinnj

bkamins:
We could add appropriate methods following Tables.jl to handle them.

The only place where this is not possible is `AbstractVector{<:DataFrameRow}` where we would have to check if the rows come from the same data frame and if yes materialize a `SubDataFrame` to avoid allocations (selecting only appropriate rows from the source). Such a behavior needs a custom method.

bkamins:
Given a discussion in https://github.com/JuliaData/DataFrames.jl/pull/1646#discussion_r255357453 maybe we should actually have the following list of possible additional constructors of `DataFrame` accepting the following:

1. a collection of `AbstractVectors` (now we only accept a `Vector`, this would allow e.g. a tuple of vectors).
2. a collection of `DataFrameRows` (now we accept a collection of `NamedTuples` via Tables.jl and this would complement it).
3. a collection of `Pair{Symbol, T} where T<:AbstractVector`, this would complement the constructors we have for Pairs by not requiring pairs to be splatted.

@nalimilan - this is what I have on a long-list of possible new constructors. None of them is crucial (so I am OK with not adding any of them - then I would close this issue), but please comment on all of them so that we can stabilize the API here.

nalimilan:
[...]

PR #1646, opened 2018-12-19
WIP: implementation of new setindex! and broadcasting rules

bkamins:
@nalimilan Following #1645 I have written down the possible target rules, so that we can edit them.

I will add a specification for `SubDataFrame` and `DataFrameRow` when we settle `AbstractDataFrame`.

CC @coreywoodfield

### DIFF
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -106,6 +106,7 @@ If it is not performed a description explicitly mentions that the data is assign
 
 > `df[cols] = v`
 
+* passed columns are copied like in `DataFrame` constructor with `copycols=true`;
 * if `v` is a `DataFrame` or `NamedTuple` of vectors or `AbstractDict` of vectors:
     * `nrow(df)` must be equal to number of rows in `v` unless `ncol(df)==0` in
       which case it is enough that number of rows in each entry of `v` is constant;
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -107,33 +107,29 @@ If it is not performed a description explicitly mentions that the data is assign
 > `df[cols] = v`
 
 * passed columns are copied like in `DataFrame` constructor with `copycols=true`;
-* if `v` is a `DataFrame` or `NamedTuple` of vectors or `AbstractDict` of vectors:
+* if `v` is an `AbstractDataFrame` of vectors:
     * `nrow(df)` must be equal to number of rows in `v` unless `ncol(df)==0` in
       which case it is enough that number of rows in each entry of `v` is constant;
     * number of columns in `v` must be equal to length of `cols`
     * before making any assignment `cols` is transformed to `Symbols` using `names(df)`
       (which means that passing `Integer` or `Bool` will fail for nonexistent columns,
       but adding new columns as `Symbol` is allowed)
-    * then `df[col] = v[col]` is called for each column name `col`
-* if `v` is an iterable of vectors the same process is performed but
-  `df[col] = v[i]` for `(i, col) in enumerate(col)` is performed instead
-* if `v` is an `AbstractMatrix` the same process is performed but
-  `df[col] = v[:, i]` for `(i, col) in enumerate(col)` is performed instead
+    * then `df[col] = copy(v[col])` is called for each column name `col`
+* in the future allowing of `NamedTuple`, `AbstractDict` and `AbstractMatrix` as `v` is planned
 
 > `df[cols] .= v`
 
 * before making any assignment `cols` is transformed to `Symbols` using `names(df)`
   (which means that passing `Integer` or `Bool` will fail for missing columns,
   but adding new columns as `Symbol` is allowed)
-* if `v` is a `DataFrame` or `NamedTuple` or `AbstractDict`:
+* if `v` is an `AbstractDataFrame`:
     * number of columns in `v` must be equal to length of `cols`
     * before making any assignment `cols` is transformed to `Symbols` using `names(df)`
       (which means that passing `Integer` or `Bool` will fail for missing columns,
       but adding new columns as `Symbol` is allowed)
     * then `df[col] .= v[col]` for `col in cols` is performed
-* if `v` is 0- or 1- dimensional then it is broadcasted to call `df[col] .= v` for all `col in cols`
-* if `v` is 2-dimensional then its dimensions of length 1 are expanded if needed and
-  `df[col] .= v[:, i]` for `(i, col) in enumerate(col)` is performed
+* otherwise it is broadcasted to call `df[col] .= v` for all `col in cols`
+* in the future allowing of `NamedTuple`, `AbstractDict` and `AbstractMatrix` as `v` is planned
 
 > `df[row, col] = v`
 
@@ -145,28 +141,32 @@ If it is not performed a description explicitly mentions that the data is assign
 
 > `df[row, cols] = v`
 
-* if `v` is a `DataFrameRow` or `NamedTuple` or `AbstractDict` then
-    * `length(v)` must be equal to `length(cols)`
-    * before making any assignment `cols` is transformed to `Symbols` using `names(df)`
-    * column names in `v` must be the same as selected by `cols`
-    * an operation `df[row, col] = v[col]` for `col in cols` is performed
-* if `v` is a iterable:
-    * `length(v)` must be equal to `length(cols)`
-    * an operation `df[row, col] = v[i]` for `(i, col) in enumerate(cols)` is performed
-* otherwise an error is thrown
+* currently this functionality is deprecated
+* in the future the following functionality is planned
+    * if `v` is a `DataFrameRow` or `NamedTuple` or `AbstractDict` then
+        * `length(v)` must be equal to `length(cols)`
+        * before making any assignment `cols` is transformed to `Symbols` using `names(df)`
+        * column names in `v` must be the same as selected by `cols`
+        * an operation `df[row, col] = v[col]` for `col in cols` is performed
+    * if `v` is a iterable:
+        * `length(v)` must be equal to `length(cols)`
+        * an operation `df[row, col] = v[i]` for `(i, col) in enumerate(cols)` is performed
+    * otherwise an error is thrown
 
 > `df[row, cols] .= v`
 
-* if `v` is a `DataFrameRow` or `NamedTuple` or `AbstractDict` then
-    * number of columns in `v` must be equal to length(cols)
-    * column names in v must be the same as selected by `cols`
-    * an operation `df[row, col] .= v[col]` for `col in cols` is performed
-* if `v` is 0-dimensional:
-    * an operation `df[row, col] .= v` for `col in cols` is performed
-* if `v` is 1-dimensional:
-    * number of elements in `v` must be equal to length(cols)
-    * an operation `df[row, col] .= v[i]` for `(i, col) in enumerate(cols)` is performed
-* otherwise an error is thrown
+* currently this functionality is undefined
+* in the future the following functionality is planned
+    * if `v` is a `DataFrameRow` or `NamedTuple` or `AbstractDict` then
+        * number of columns in `v` must be equal to length(cols)
+        * column names in v must be the same as selected by `cols`
+        * an operation `df[row, col] .= v[col]` for `col in cols` is performed
+    * if `v` is 0-dimensional:
+        * an operation `df[row, col] .= v` for `col in cols` is performed
+    * if `v` is 1-dimensional:
+        * number of elements in `v` must be equal to length(cols)
+        * an operation `df[row, col] .= v[i]` for `(i, col) in enumerate(cols)` is performed
+    * otherwise an error is thrown
 
 > `df[rows, col] = v`
 
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -117,7 +117,7 @@ If it is not performed a description explicitly mentions that the data is assign
       (which means that passing `Integer` or `Bool` will fail for nonexistent columns,
       but adding new columns as `Symbol` is allowed)
     * then `df[col] = v[col]` is called for each column name `col`
-* if `v` is a vector of vectors or a tuple of vectors the same process is performed but
+* if `v` is a vector of vectors the same process is performed but
   `df[col] = v[i]` for `(i, col) in enumerate(col)` is performed instead
 * if `v` is an `AbstractMatrix` the same process is performed but
   `df[col] = v[:, i]` for `(i, col) in enumerate(col)` is performed instead
@@ -151,7 +151,7 @@ If it is not performed a description explicitly mentions that the data is assign
     * `length(v)` must be equal to `length(cols)`
     * column names in `v` must be the same as selected by `cols`
     * an operation `df[row, col] = v[col]` for `col in cols` is performed
-* if `v` is a vector or a tuple:
+* if `v` is a vector:
     * `length(v)` must be equal to `length(cols)`
     * an operation `df[row, col] = v[i]` for `(i, col) in enumerate(cols)` is performed
 * otherwise an error is thrown
@@ -172,22 +172,22 @@ If it is not performed a description explicitly mentions that the data is assign
 > `df[rows, col] = v`
 
 * The same rules as for `df[col] = v` but on selected rows and always with copying.
-* Empty data frames are not allowed and no column adding is possible.
+* Data frames with zero columns are not allowed and no column adding is possible.
 
 > `df[rows, col] .= v`
 
 * The same rules as for `df[col] .= v` but on selected rows and always with copying.
-* Empty data frames are not allowed and no column adding is possible.
+* Data frames with zero columns are not allowed and no column adding is possible.
 
 > `df[rows, cols] = v`
 
 * The same rules as for `df[cols] = v` but on selected rows and always with copying.
-* Empty data frames are not allowed and no column adding is possible.
+* Data frames with zero columns are not allowed and no column adding is possible.
 
 > `df[rows, cols] .= v`
 
 * The same rules as for `df[cols] .= v` but on selected rows and always with copying.
-* Empty data frames are not allowed and no column adding is possible.
+* Data frames with zero columns are not allowed and no column adding is possible.
 
 Additionally for all operations:
 * If `rows` is `:` then it gets expanded to `axes(df, 1)` before any action.
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -194,17 +194,14 @@ Additionally for all operations:
 ### `SubDataFrame`:
 
 If `sdf` is a `SubDataFrame` before any operation the following translations take place:
-* `sdf[col]` to `parent(sdf)[sdf.rows, Index(sdf)[col]]`
-* `sdf[cols]` to `parent(sdf)[sdf.rows, Index(sdf)[cols]]`
-* `sdf[row, col]` to `parent(sdf)[sdf.rows[row], Index(sdf)[col]]`
-* `sdf[rows, cols]` to `parent(sdf)[sdf.rows[rows], Index(sdf)[cols]]`
+* `sdf[colidx]` to `parent(sdf)[sdf.rows, parentcols(index(sdf), colidx)]`
+* `sdf[rowidx, colidx]` to `parent(sdf)[sdf.rows[rowidx], parentcols(index(sdf), colidx)]`
 
 After this translation rules for `DataFrame` assignement and broadcasting apply.
 
 ### `DataFrameRow`:
 
 If `dfr` is a `DataFrameRow` before any operation the following translations take place:
-* `dfr[col]` to `parent(dfr)[dfr.row, Index(dfr)[col]]`
-* `dfr[cols]` to `parent(dfr)[dfr.row, Index(dfr)[cols]]`
+* `dfr[idx]` to `parent(dfr)[dfr.row, parentcols(index(dfr), idx)]`
 
 After this translation rules for `DataFrame` assignement and broadcasting apply.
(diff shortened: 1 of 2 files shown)


docs/src/lib/indexing.md
```diff
@@ -76,6 +76,124 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+
+> `df[col] = v` and `df.col = v`
+
+* `v` must be an `AbstractVector`. If `ncol(df) > 0` then `length(v)` must be equal to `nrow(df)`.
+* `v` is assigned to `df` in place unless it is a range, when it is converted to a `Vector`.
+* `haskey(df, col)` must hold unless `col isa Symbol`, in which case a new column is added to `df`.
+
+> `df[col] .= v` and `df.col .= v`
+
+* `df.col .= v` throws an error.
```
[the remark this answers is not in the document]
bkamins: Right. I meant to put it lower (the case of the new column with symbol). Fixed.

docs/src/lib/indexing.md
```diff
@@ -76,6 +76,124 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+* `v` must be broadcastable.
+* if `haskey(df, col)` then standard broadcasted assignment of `v` to `df[col]` is performed.
+* if not `haskey(df, col)` and `col isa Symbol`
+    * if `ncol(df) > 0` then we create a new vector `c` of an appropriate type, having `nrow(df)` entries;
```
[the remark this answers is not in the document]

docs/src/lib/indexing.md
```diff
@@ -76,6 +76,124 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+    * an operation `df[row, col] .= v[]` for `col in cols` is performed
+* if `v` is 1-dimensional:
+    * number of elements in `v` must be equal to length(cols)
+    * an operation `df[row, col] .= v[i]` for `(i, col) in enumerate(cols)` is performed
+* otherwise an error is thrown
+
+> `df[rows, col] = v`
+
+* The same rules as for `df[col] = v` but on selected rows and always with copying.
+* Empty data frames are not allowed and no column adding is possible.
```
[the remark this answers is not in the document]
bkamins: We have `df[col] = v` and `df[col] .= v` for this case. It would render `df[:, col] = v` inconsistent.

And here I exactly try to fix the inconsistency in the current design of DataFrames.jl. See the example how `x` and `y` are treated differently:
```
julia> x = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> y = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> df = DataFrame(x=x,y=y)
3×2 DataFrame
│ Row │ x     │ y     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │
│ 2   │ 2     │ 2     │
│ 3   │ 3     │ 3     │

julia> df[:, :x] = [0,0,0]
3-element Array{Int64,1}:
 0
 0
 0

julia> df[1:3, :y] = [0,0,0]
3-element Array{Int64,1}:
 0
 0
 0

julia> df
3×2 DataFrame
│ Row │ x     │ y     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 0     │ 0     │
│ 2   │ 0     │ 0     │
│ 3   │ 0     │ 0     │

julia> x
3-element Array{Int64,1}:
 1
 2
 3

julia> y
3-element Array{Int64,1}:
 0
 0
 0

julia> df[:, :x] = 'a':'c'
'a':1:'c'

julia> df
3×2 DataFrame
│ Row │ x    │ y     │
│     │ Char │ Int64 │
├─────┼──────┼───────┤
│ 1   │ 'a'  │ 0     │
│ 2   │ 'b'  │ 0     │
│ 3   │ 'c'  │ 0     │

julia> df[1:3, :y] = 'a':'c'
'a':1:'c'

julia> df
3×2 DataFrame
│ Row │ x    │ y     │
│     │ Char │ Int64 │
├─────┼──────┼───────┤
│ 1   │ 'a'  │ 97    │
│ 2   │ 'b'  │ 98    │
│ 3   │ 'c'  │ 99    │
```

docs/src/lib/indexing.md
```diff
@@ -76,6 +76,124 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+
+> `df[rows, col] .= v`
+
+* The same rules as for `df[col] .= v` but on selected rows and always with copying.
+* Empty data frames are not allowed and no column adding is possible.
+
+> `df[rows, cols] = v`
```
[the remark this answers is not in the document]
bkamins: `df[row, cols]` is significantly different from `df[rows, cols]` (which is rather like ``df[cols]` but does not allow adding new columns). I would keep it as a separate description for an easier reference.

docs/src/lib/indexing.md
```diff
@@ -76,6 +76,127 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+
+* if `v` is a `DataFrame` or `NamedTuple` of vectors or `AbstractDict` of vectors:
+    * `nrow(df)` must be equal to number of rows in `v` unless `ncol(df)==0` in
+      which case it is enough that number of rows in each entry of `v` is constant;
+    * number of columns in `v` must be equal to length of `cols`
+    * before making any assignment `cols` is transformed to `Symbols` using `names(df)`
+      (which means that passing `Integer` or `Bool` will fail for nonexistent columns,
+      but adding new columns as `Symbol` is allowed)
+    * then `df[col] = v[col]` is called for each column name `col`
+* if `v` is a vector of vectors or a tuple of vectors the same process is performed but
```
bkamins: I have added this rule. I do not know if we want to keep it (or keep it only for vectors and not for tuples)

docs/src/lib/indexing.md
```diff
@@ -76,6 +76,127 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+
+* translates to `df[col][row] .= v`
+
+> `df[row, cols] = v`
+
+* if `v` is a `DataFrameRow` or `NamedTuple` or `AbstractDict` then
+    * `length(v)` must be equal to `length(cols)`
+    * column names in `v` must be the same as selected by `cols`
+    * an operation `df[row, col] = v[col]` for `col in cols` is performed
+* if `v` is a vector or a tuple:
```
bkamins: this additional rule is related to this rule.

docs/src/lib/indexing.md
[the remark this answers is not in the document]
nalimilan: I agree this is problematic, but I don't see the relationship with empty data frames (it's been a long time....).

docs/src/lib/indexing.md
nalimilan: Sounds fine. Am I right that this doesn't introduce any ambiguity if you want a column which is a vector of vectors (since you would have to wrap it in a vector/tuple/data frame anyway since this is a multi-column assignment)?

docs/src/lib/indexing.md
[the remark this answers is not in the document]
bkamins: The proposed rule is that if we use `df[:, col]` or `df[rows, col]` in general we write to an existing vector in the `DataFrame` in-place. So - naturally it has to fail in an empty `DataFrame`.

Or - from another angle - the proposed rule is that on LHS `df[:, col]` is rewritten as `df[axes(df, 1), col]`.

This approach is consistent with Base, where the manual states [here](https://docs.julialang.org/en/v1/manual/arrays/index.html#man-array-indexing-1):

> This includes `Colon` (`:`) to select all indices within the entire dimension

I would be OK to bend the rules in this case, if there were no other convenient way to assign a new column to a `DataFrame`, but there is one and it is `df[col] = v`.

But if you really think it is useful we can make an exception for an empty `DataFrame`.

docs/src/lib/indexing.md
bkamins: Right - you would have to wrap it. Single column assignment is another rule.

Actually this is very delicate. My idea is that `df[cols] = v` should ideally give the same result as `df[cols] = DataFrame(v)` just without having to materialize the intermediate `DataFrame`.

The same for `df[rows, cols] = v`.

But now I see that:

```
julia> x = ([1,2,3], [1,2,3])
([1, 2, 3], [1, 2, 3])

julia> DataFrame(x)
ERROR: ArgumentError: 'Tuple{Array{Int64,1},Array{Int64,1}}' iterates 'Array{Int64,1}' values, which don't satisfy the Tables.jl Row-iterator interface

julia> x = [[1,2,3], [1,2,3]]
2-element Array{Array{Int64,1},1}:
 [1, 2, 3]
 [1, 2, 3]

julia> DataFrame(x)
3×2 DataFrame
│ Row │ x1    │ x2    │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │
│ 2   │ 2     │ 2     │
│ 3   │ 3     │ 3     │
```

which is quite surprising (in general the Tables.jl & friends catch all `DataFrame` constructor always catches me off guard 😞). So I think we should drop tuple here and leave only the vector?

docs/src/lib/indexing.md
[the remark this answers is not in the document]
nalimilan: But if we have e.g.
```julia
df = DataFrame(x=[])
df[:, 1] = []
```
Then what's the problem? We've set all entries in `df.x` to the values from the RHS -- it's just that there aren't any of them. :-)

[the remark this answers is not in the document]
bkamins: There is no problem 😄. Your `df` is not empty, it has 0-rows, but it is a different story. An empty data frame is returned by the call `DataFrame()`.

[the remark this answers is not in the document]
bkamins: Note that this distinction of emptyness is very relevant in other cases.

This should work:
```
df = DataFrame()
df.a = [1,2,3]
``
while this should fail
```
df = DataFrame(x=[])
df.a = [1,2,3]
```

An interesting corner case is:
```
julia> df = DataFrame(x=[1,2])
2×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │

julia> df.x = [1,2,3,4]
ERROR: ArgumentError: New columns must have the same length as old columns
```
which fails now and I thing it should still fail.

The short rule - an empty data frame `DataFrame()` has an undefined number of rows (not `0`) while any non-empty data frame has a concrete number of rows.

Actually we could consider to make `nrow(DataFrame())` return `nothing` instead of `0`, this would be technically stricter, but I always thought it would not be very useful, so I have not proposed to add this change.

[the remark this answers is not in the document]
Comment (nalimilan) [reply docs/src/lib/indexing.md]: OK, then maybe say "with zero columns" to avoid any possible ambiguity.

> Actually we could consider to make `nrow(DataFrame())` return `nothing` instead of `0`, this would be technically stricter, but I always thought it would not be very useful, so I have not proposed to add this change.

Or even `missing`, since it's really unknown. ;-) But `0` is probably the most useful answer.

File: docs/src/lib/indexing.md (hunk 5 above)
Comment (nalimilan) [reply docs/src/lib/indexing.md]: Why doesn't Tables.jl accept a tuple here?

Comment (bkamins) [reply docs/src/lib/indexing.md]: It accepts it, but assumes that the contents is row-wise not col-wise, see e.g.:
```
julia> DataFrame(((a=1,b=2),(a=3,b=4)))
2×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │
│ 2   │ 3     │ 4     │
```
That is why I say that this Tables.jl API is tricky when combined with what we already had in DataFrames.jl.

Comment (bkamins) [reply docs/src/lib/indexing.md]: I will remove tuples here from the description.

File: docs/src/lib/indexing.md (hunk 3 above)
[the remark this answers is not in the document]
Comment (bkamins) [reply docs/src/lib/indexing.md]: OK - I will replace empty with zero columns everywhere to disambiguate.

File: docs/src/lib/indexing.md (hunk 5 above)
Comment (nalimilan) [reply docs/src/lib/indexing.md]: Then maybe it should do the same for vectors?

Comment (bkamins) [reply docs/src/lib/indexing.md]: I can remove them too, but actually we have a `DataFrame` constructor taking a vector of vectors, which treats them col-wise.

The general question is in the assignment:

```
df[rows, cols] = v
```
should `v` be treated row-wise or col-wise? And a natural answer is to follow what `DataFrame(v)` would do.

In general I am OK, to narrow down what is accepted now, and we can add the functionality later.

Comment (nalimilan) [reply docs/src/lib/indexing.md]: I agree. But it doesn't sound very consistent for Tables.jl to treat vectors differently from tuples. Or maybe we should special-case collections of named tuples. Anyway it's probably worth opening a separate issue to discuss that.

Comment (bkamins) [reply docs/src/lib/indexing.md]: There is no inconsistency. Tables.jl does not support vector of vectors at all (the same as tuple of vectors). It is only supported in DataFrames.jl specific `DataFrame` constructor.

Actually thinking of it we might want to accept a `Matrix` as `v` (it is also not supported by Tables.jl but is supported by `DataFrame` constructor).

Comment (nalimilan) [reply docs/src/lib/indexing.md]: OK. Though currently there seems to be a bug with this (but at least it should work since there's a special method for it in DataFrames):
```julia
julia> DataFrame([(a=1,b=2),(a=3,b=4)])
ERROR: UndefVarError: DataValueUnwrapper not defined
```

Regarding what we can do in DataFrames, why not add a constructor taking a tuple of vectors to parallel the one taking a vector of vectors?

> Actually thinking of it we might want to accept a `Matrix` as `v` (it is also not supported by Tables.jl but is supported by `DataFrame` constructor).

Fine with me. But this can be added at any time since that's an error AFAICT.

bkamins: 1. Following your suggestion I have commented on the constructors in a separate issue https://github.com/JuliaData/DataFrames.jl/issues/1599.
2. Are you on latest versions of all packages from Tables.jl & friends? For me it works:
```
julia> DataFrame([(a=1,b=2),(a=3,b=4)])
2×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │
│ 2   │ 3     │ 4     │
```
(and even if it did not work it would not be an error in DataFrames.jl 😀)

nalimilan: OK, actually I tested that on my PooledArrays branch which was behind master. It works now.

bkamins: Given the discussion in https://github.com/JuliaData/DataFrames.jl/issues/1599#issuecomment-462261917 I changed my mind and would leave only vector of vectors here for now.

The resason is that in general it is not clear if you have a collection if it should be treated row-wise or col-wise. We can add other values of `v` later if there is a need.

nalimilan: My suggestion was that the collection type should not matter to decide whether it's row- or column-wise: tuples and vectors should be treated the same. Two constructors would take tuples and vectors of vectors. Other collections would be assumed to be row-oriented and handled by Tables.jl.

bkamins: OK :). So I add tuples here and in the constructors thread.

richgoldberg:
Regarding the specification in `docs/src/lib/indexing.md`:
> df[rows, col] = v
> The same rules as for df[col] = v but on selected rows and always with copying.
> Empty data frames are not allowed and no column adding is possible.
> 
> df[rows, col] .= v
> The same rules as for df[col] .= v but on selected rows and always with copying.
> Empty data frames are not allowed and no column adding is possible.
 
This doesn't seem to be the case right now (maybe you know this?) even when there is no issue with empty DataFrames.  Here's an example where `df[rows, col] .= v`  does not behave like `df[col] .= v`.  Further, it doesn't throw any sort of error or warning to the user.

```
# Example:  Start with a DataFrame and update a column in-place

# initial DataFrame to modify
using DataFrames
df = DataFrame(a = [1,3,5], b = ["A", "B", "C"])
display(df)

#Case 1: df[col] = v works
df = DataFrame(a = [1,3,5], b = ["A", "B", "C"])
v = ["a", "b", "c"]
df[ :b] = v
display(df)

#Case 2: df[rows, col] = v works
df = DataFrame(a = [1,3,5], b = ["A", "B", "C"])
v = ["a", "b", "c"]
df[:, :b] = v
display(df)

#Case 3: df[col] .= v works
df = DataFrame(a = [1,3,5], b = ["A", "B", "C"])
v = ["a", "b", "c"]
df[ :b] .= v
display(df)

#Case 4:, df[rows, col] .= v doesn't work!
df = DataFrame(a = [1,3,5], b = ["A", "B", "C"])
v = ["a", "b", "c"]
df[:, :b] .= v
display(df)
```

Note:  Run in Julia 1.1.0 using DataFrames v. 0.17.1

nalimilan:
This PR describes what should happen, not what happens right now.

docs/src/lib/indexing.md
```diff
@@ -76,6 +76,127 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
 * `@view dfr[col]` -> a `0`-dimensional view into `parent(dfr)[DataFrames.row(dfr), col]`;
 * `@view dfr[cols]` -> a `DataFrameRow` with parent `parent(dfr)` if `cols` is a colon and `parent(dfr)[cols]` otherwise;
 
-## `setindex!`
+## `setindex!`, `setproperty!` and broadcasted assignment
+
+This is a planned functionality.
```
oxinabox:
```suggestion
```

Do not forget to delete this

bkamins:
When implementing the `setindex!` and broadcasting rules I started to have a radical thought that we should not allow assignment to a `DataFrame`, but only to `df[col]`, `df[row, col]`, `df[rows, col]`. So essentially to allow assignment to a single column, which is a vector and then the rules are clear.

What I mean is that assignment to `df[cols]`, `df[row, cols]`, `df[rows, cols]` is problematic in two ways:
* should we do matching by column name or column index (we could set some rules);
* when broadcasting how should we treat a right hand side. `df[cols]` and `df[rows, cols]` theoretically should be viewed as a collection of rows, but `df[row, cols]` should be a collection of 1-element columns;

Additionally I am not sure how useful having `df[cols]`, `df[row, cols]`, `df[rows, cols]` on LHS is in practice while if it is really needed then it can be explicitly handled by a simple iteration over columns.

@nalimilan, @oxinabox - do you have any thoughts on it (if it is not clear what I mean here please comment and I can expand on my thoughts).

bkamins:
Just to give more perspective to it I am not convinced we want to accept any of the following:

```
julia> df = DataFrame(a=1:3)
3×1 DataFrame
│ Row │ a     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │
│ 3   │ 3     │

julia> df2 = DataFrame(x=4:6)
3×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 4     │
│ 2   │ 5     │
│ 3   │ 6     │

julia> df[[:b]] = df2
3×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 4     │
│ 2   │ 5     │
│ 3   │ 6     │

julia> df
3×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 4     │
│ 2   │ 2     │ 5     │
│ 3   │ 3     │ 6     │

julia> df[[:c, :d]] = 10
10

julia> df
3×4 DataFrame
│ Row │ a     │ b     │ c     │ d     │
│     │ Int64 │ Int64 │ Int64 │ Int64 │
├─────┼───────┼───────┼───────┼───────┤
│ 1   │ 1     │ 4     │ 10    │ 10    │
│ 2   │ 2     │ 5     │ 10    │ 10    │
│ 3   │ 3     │ 6     │ 10    │ 10    │

julia> df[1, [1]] = df2
3×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 4     │
│ 2   │ 5     │
│ 3   │ 6     │

julia> df
3×4 DataFrame
│ Row │ a     │ b     │ c     │ d     │
│     │ Int64 │ Int64 │ Int64 │ Int64 │
├─────┼───────┼───────┼───────┼───────┤
│ 1   │ 4     │ 4     │ 10    │ 10    │
│ 2   │ 2     │ 5     │ 10    │ 10    │
│ 3   │ 3     │ 6     │ 10    │ 10    │

julia> df[3, [1]] = df2
3×1 DataFrame
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 4     │
│ 2   │ 5     │
│ 3   │ 6     │

julia> df
3×4 DataFrame
│ Row │ a     │ b     │ c     │ d     │
│     │ Int64 │ Int64 │ Int64 │ Int64 │
├─────┼───────┼───────┼───────┼───────┤
│ 1   │ 4     │ 4     │ 10    │ 10    │
│ 2   │ 2     │ 5     │ 10    │ 10    │
│ 3   │ 4     │ 6     │ 10    │ 10    │
```

and in general now I feel that allowing a data frame on the RHS of an assignment is problematic.

What we could allow for (but I am not sure how much useful that would be so that is why I am hesitant to propose these - we can always add them later) are:

* `df[row, cols] = DataFrameRow | AbstractDict | NamedTuple | iterable` (in general to have the same functionality as `push!`)
* `df[rows, cols] = matrix` (treating `df` as a 2-dimensional object)
* `df[row, cols] .= any 0-dimensional value in terms of broadcasting` and `df[rows, cols] .= any 0, 1 or 2-dimensional value in terms of broadcasting` (in general: adhering to standard broadcasting that assumes we have numerical axis and disallowing any other thing on RHS)

bkamins:
#1866 implements current rules

Issue #1647, opened 2018-12-19
Export function for getting row number of DataFrameRow


kescobo:
There's an internal function `DataFrames.row()` (defined [here](https://github.com/JuliaData/DataFrames.jl/blob/master/src/dataframerow/dataframerow.jl#L27)) that returns the index of a `DataFrameRow`, which I often need while iterating. It would be nice if that function were exported, though `row()` is probably too generic. Maybe `rownumber` or something? 

I'd be happy to submit a PR if others would find this useful.

bkamins:
As indicated on https://julialang.slack.com/archives/C674VR0HH/p1545250280227000 we currently define `parentindices` which is a standard function in Base to get this information (row number is the first entry in the returned tuple).
Is this enough?

nalimilan:
I don't really like the idea of exporting two more functions just for this. What's wrong with `enumerate(eachrow(df))`? `parentindices(r)` isn't too bad either when you need the column indices in the parent.

bkamins:
There is one important thing here to remember. `parentindices(dfr,2)` gives you indices to `parent(dfr)`, however `parent(dfr)` does not have to be the data frame you have used to create the `DataFrameRow`. The same applies to `SubDataFrame`.

Of course in the most common case of `eachrow(df)` this does not matter, but it will matter eg. for `df[row, cols]` when `cols` is not a `Colon`.

So the decision to make is do we want to hold the true *parent* of `DataFrameRow` and `SubDataFrame` along with indices into it? It should be cheap. Actually the strategy should probably be to hold a vector of integer indices or a range (thus we will have to add an additional parameter to the types).

I was hesitant to do it, but now I see how it can be efficiently implemented so I could submit a PR if all agree (actually something like `df[1, 1:3]` should become much faster). Of course it will work as long as parent columns are not mutated (current implementation makes a copy and thus does not reflect the changes to the parent). This means that we would be mildly breaking (but I guess no one would notice it so I would go for it right away).

kescobo:
I only understand ~60% of what you're saying, but sounds great! 😆

bkamins:
In short - currently `parent` of `SubDataFrame` and `DataFrameRow` is a copy of the original `AbstractDataFrame` if you do column selection on it. We can avoid it and at the same time improve the performance.

bkamins:
closing as we have implemented all that was discussed

bkamins:
if I missed something please reopen

PR #1648, opened 2018-12-21
better parent for view

bkamins:
This follows the discussion in #1647. I have not tested it yet, but have implemented all the the required changes (hopefully) so the design approach can be discussed.

### DIFF
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -8,7 +8,7 @@
 
 # Iteration by rows
 """
-    DataFrameRows <: AbstractVector{DataFrameRow}
+    DataFrameRows{D<:AbstractDataFrame,S<:SubIndex} <: AbstractVector{DataFrameRow{D,S}}
 
 Iterator over rows of an `AbstractDataFrame`,
 with each row represented as a `DataFrameRow`.
@@ -17,9 +17,9 @@ Currently supports `DataFrame` and `SubDataFrame`.
 
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
-struct DataFrameRows{S<:AbstractDataFrame,T<:SubIndex} <: AbstractVector{DataFrameRow{S,T}}
-    df::S
-    subindex::T
+struct DataFrameRows{D<:AbstractDataFrame,S<:SubIndex} <: AbstractVector{DataFrameRow{D,S}}
+    df::D
+    subindex::S
 end
 
 """
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -79,7 +79,7 @@ x  3
 eachrow(df::DataFrame) =
     DataFrameRows(df, SubIndex(index(df), :))
 eachrow(sdf::SubDataFrame) =
-    DataFrameRows(df, index(df))
+    DataFrameRows(sdf, index(sdf))
 
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
 Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -8,15 +8,16 @@
 
 # Iteration by rows
 """
-    DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow}
+    DataFrameRows <: AbstractVector{DataFrameRow}
 
 Iterator over rows of an `AbstractDataFrame`,
 with each row represented as a `DataFrameRow`.
 
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
-struct DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow}
-    df::T
+struct DataFrameRows{T} <: AbstractVector{DataFrameRow{SubIndex{Base.OneTo{Int},Base.OneTo{Int}}}}
+    df::DataFrame
+    rows::T
 end
 
 """
@@ -61,14 +62,17 @@ julia> copy.(eachrow(df))
  (x = 4, y = 14)
 ```
 """
-eachrow(df::AbstractDataFrame) = DataFrameRows(df)
+eachrow(df::DataFrame) = DataFrameRows(df, :)
+eachrow(sdf::SubDataFrame) =
+    DataFrameRows(parent(sdf)[parentcols(sdf, :)], rows(sdf))
 
-Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
-@inline function Base.getindex(itr::DataFrameRows, i::Int)
-    @boundscheck checkbounds(itr, i)
-    return DataFrameRow(itr.df, i, :)
-end
+
+Base.size(itr::DataFrameRows) = (length(itr.rows), )
+Base.size(itr::DataFrameRows{Colon}) = (size(itr.df, 2), )
+
+@inline Base.getindex(itr::DataFrameRows{Colon}, i::Int) = DataFrameRow(itr.df, i, :)
+@inline Base.getindex(itr::DataFrameRows, i::Int) = DataFrameRow(itr.df, itr.rows[i], :)
 
 # Iteration by columns
 """
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -8,7 +8,7 @@
 
 # Iteration by rows
 """
-    DataFrameRows{D<:AbstractDataFrame,S<:SubIndex} <: AbstractVector{DataFrameRow{D,S}}
+    DataFrameRows{D<:AbstractDataFrame,S<:AbstractIndex} <: AbstractVector{DataFrameRow{D,S}}
 
 Iterator over rows of an `AbstractDataFrame`,
 with each row represented as a `DataFrameRow`.
@@ -17,19 +17,19 @@ Currently supports `DataFrame` and `SubDataFrame`.
 
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
-struct DataFrameRows{D<:AbstractDataFrame,S<:SubIndex} <: AbstractVector{DataFrameRow{D,S}}
+struct DataFrameRows{D<:AbstractDataFrame,S<:AbstractIndex} <: AbstractVector{DataFrameRow{D,S}}
     df::D
-    subindex::S
+    index::S
 end
+
+Base.summary(dfrs::DataFrameRows) = "$(length(dfrs))-element DataFrameRows"
+Base.summary(io::IO, dfrs::DataFrameRows) = print(io, summary(dfrs))
 
 """
     eachrow(df::AbstractDataFrame)
 
 Return a `DataFrameRows` that iterates a data frame row by row,
 with each row represented as a `DataFrameRow`.
-
-Note that for performance reasons `eachrow(df::DataFrame)` assumes that the number
-of columns of `df` does not change after constructing the `DataFrameRows` object.
 
 **Examples**
 
@@ -45,7 +45,7 @@ julia> df = DataFrame(x=1:4, y=11:14)
 │ 4   │ 4     │ 14    │
 
 julia> eachrow(df)
-4-element DataFrames.DataFrameRows{DataFrame,DataFrames.SubIndex{Base.OneTo{Int64},Base.OneTo{Int64}}}:
+4-element DataFrameRows:
  DataFrameRow (row 1)
 x  1
 y  11
@@ -67,7 +67,7 @@ julia> copy.(eachrow(df))
  (x = 4, y = 14)
 
 julia> eachrow(view(df, [4,3], [2,1]))
-2-element DataFrames.DataFrameRows{Array{Int64,1}}:
+2-element DataFrameRows:
  DataFrameRow (row 4)
 y  14
 x  4
@@ -76,18 +76,16 @@ y  13
 x  3
 ```
 """
-eachrow(df::DataFrame) =
-    DataFrameRows(df, SubIndex(index(df), :))
-eachrow(sdf::SubDataFrame) =
-    DataFrameRows(sdf, index(sdf))
+eachrow(df::DataFrame) = DataFrameRows(df, index(df))
+eachrow(sdf::SubDataFrame) = DataFrameRows(sdf, index(sdf))
 
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
 Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
 
 @inline Base.getindex(itr::DataFrameRows{DataFrame}, i::Int) =
-    DataFrameRow(itr.df, itr.subindex, i)
+    DataFrameRow(itr.df, itr.index, i)
 @inline Base.getindex(itr::DataFrameRows{<:SubDataFrame}, i::Int) =
-    DataFrameRow(parent(itr.df), itr.subindex, rows(itr.df)[i])
+    DataFrameRow(parent(itr.df), itr.index, rows(itr.df)[i])
 
 # Iteration by columns
 """
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -29,7 +29,7 @@ Return a `DataFrameRows` that iterates a data frame row by row,
 with each row represented as a `DataFrameRow`.
 
 Note that for performance reasons `eachrow(df::DataFrame)` assumes that the number
-of columns of `df` does not change after materializing `DataFrameRows` object.
+of columns of `df` does not change after constructing the `DataFrameRows` object.
 
 **Examples**
 
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -17,22 +17,19 @@ Currently supports `DataFrame` and `SubDataFrame`.
 
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
-struct DataFrameRows{T} <: AbstractVector{DataFrameRow{SubIndex{Base.OneTo{Int},Base.OneTo{Int}}}}
-    df::DataFrame
-    rows::T
+struct DataFrameRows{S<:AbstractDataFrame,T<:SubIndex} <: AbstractVector{DataFrameRow{T}}
+    df::S
+    subindex::T
 end
 
 """
-    eachrow(df::Union{DataFrame, SubDataFrame})
+    eachrow(df::Union{DataFrame,SubDataFrame})
 
 Return a `DataFrameRows` that iterates a data frame row by row,
 with each row represented as a `DataFrameRow`.
 
-`eachrow(df::DataFrame)` references the `df` directly, but
-for performance considerations `eachrow(df::SubDataFrame)` eagerly copies
-relevant column information from `df` (if columns of parent of a `SubDataFrame`
-are renamed or reallocated after calling `eachrow` this change will not be
-reflected by returned `DataFrameRows`).
+Note that for performance reasons `eachrow(df::DataFrame)` assumes that the number
+of columns of `df` does not change after materializing `DataFrameRows` object.
 
 **Examples**
 
@@ -48,7 +45,7 @@ julia> df = DataFrame(x=1:4, y=11:14)
 │ 4   │ 4     │ 14    │
 
 julia> eachrow(df)
-4-element DataFrames.DataFrameRows{Colon}:
+4-element DataFrames.DataFrameRows{DataFrame,DataFrames.SubIndex{Base.OneTo{Int64},Base.OneTo{Int64}}}:
  DataFrameRow (row 1)
 x  1
 y  11
@@ -79,17 +76,18 @@ y  13
 x  3
 ```
 """
-eachrow(df::DataFrame) = DataFrameRows(df, :)
+eachrow(df::DataFrame) =
+    DataFrameRows(df, SubIndex(index(df), :))
 eachrow(sdf::SubDataFrame) =
-    DataFrameRows(parent(sdf)[parentcols(sdf, :)], rows(sdf))
+    DataFrameRows(df, index(df))
 
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
+Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
 
-Base.size(itr::DataFrameRows) = (length(itr.rows), )
-Base.size(itr::DataFrameRows{Colon}) = (size(itr.df, 1), )
-
-@inline Base.getindex(itr::DataFrameRows{Colon}, i::Int) = DataFrameRow(itr.df, i, :)
-@inline Base.getindex(itr::DataFrameRows, i::Int) = DataFrameRow(itr.df, itr.rows[i], :)
+@inline Base.getindex(itr::DataFrameRows{DataFrame}, i::Int) =
+    DataFrameRow(itr.df, itr.subindex, i)
+@inline Base.getindex(itr::DataFrameRows{SubDataFrame}, i::Int) =
+    DataFrameRow(parent(itr.df), itr.subindex, rows(itr.df)[i])
 
 # Iteration by columns
 """
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -13,8 +13,6 @@
 Iterator over rows of an `AbstractDataFrame`,
 with each row represented as a `DataFrameRow`.
 
-Currently supports `DataFrame` and `SubDataFrame`.
-
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
 struct DataFrameRows{D<:AbstractDataFrame,S<:AbstractIndex} <: AbstractVector{DataFrameRow{D,S}}
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -24,6 +24,42 @@ end
 
 Return a `DataFrameRows` that iterates an `AbstractDataFrame` row by row,
 with each row represented as a `DataFrameRow`.
+
+**Examples**
+
+```jldoctest
+julia> df = DataFrame(x=1:4, y=11:14)
+4×2 DataFrame
+│ Row │ x     │ y     │
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 11    │
+│ 2   │ 2     │ 12    │
+│ 3   │ 3     │ 13    │
+│ 4   │ 4     │ 14    │
+
+julia> eachrow(df)
+4-element DataFrames.DataFrameRows{DataFrame}:
+ DataFrameRow (row 1)
+x  1
+y  11
+ DataFrameRow (row 2)
+x  2
+y  12
+ DataFrameRow (row 3)
+x  3
+y  13
+ DataFrameRow (row 4)
+x  4
+y  14
+
+julia> copy.(eachrow(df))
+4-element Array{NamedTuple{(:x, :y),Tuple{Int64,Int64}},1}:
+ (x = 1, y = 11)
+ (x = 2, y = 12)
+ (x = 3, y = 13)
+ (x = 4, y = 14)
+```
 """
 eachrow(df::AbstractDataFrame) = DataFrameRows(df)
 
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -40,7 +40,7 @@ julia> df = DataFrame(x=1:4, y=11:14)
 │ 4   │ 4     │ 14    │
 
 julia> eachrow(df)
-4-element DataFrames.DataFrameRows{DataFrame}:
+4-element DataFrames.DataFrameRows{Colon}:
  DataFrameRow (row 1)
 x  1
 y  11
@@ -69,7 +69,7 @@ eachrow(sdf::SubDataFrame) =
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
 
 Base.size(itr::DataFrameRows) = (length(itr.rows), )
-Base.size(itr::DataFrameRows{Colon}) = (size(itr.df, 2), )
+Base.size(itr::DataFrameRows{Colon}) = (size(itr.df, 1), )
 
 @inline Base.getindex(itr::DataFrameRows{Colon}, i::Int) = DataFrameRow(itr.df, i, :)
 @inline Base.getindex(itr::DataFrameRows, i::Int) = DataFrameRow(itr.df, itr.rows[i], :)
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -8,14 +8,14 @@
 
 # Iteration by rows
 """
-    DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow{T}}
+    DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow}
 
 Iterator over rows of an `AbstractDataFrame`,
 with each row represented as a `DataFrameRow`.
 
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
-struct DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow{T}}
+struct DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow}
     df::T
 end
 
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -17,7 +17,7 @@ Currently supports `DataFrame` and `SubDataFrame`.
 
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
-struct DataFrameRows{S<:AbstractDataFrame,T<:SubIndex} <: AbstractVector{DataFrameRow{T}}
+struct DataFrameRows{S<:AbstractDataFrame,T<:SubIndex} <: AbstractVector{DataFrameRow{S,T}}
     df::S
     subindex::T
 end
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -76,15 +76,14 @@ y  13
 x  3
 ```
 """
-eachrow(df::DataFrame) = DataFrameRows(df, index(df))
-eachrow(sdf::SubDataFrame) = DataFrameRows(sdf, index(sdf))
+eachrow(df::AbstractDataFrame) = DataFrameRows(df, index(df))
 
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
 Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
 
-@inline Base.getindex(itr::DataFrameRows{DataFrame}, i::Int) =
+Base.@propagate_inbounds Base.getindex(itr::DataFrameRows, i::Int) =
     DataFrameRow(itr.df, itr.index, i)
-@inline Base.getindex(itr::DataFrameRows{<:SubDataFrame}, i::Int) =
+Base.@propagate_inbounds Base.getindex(itr::DataFrameRows{<:SubDataFrame}, i::Int) =
     DataFrameRow(parent(itr.df), itr.index, rows(itr.df)[i])
 
 # Iteration by columns
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -13,6 +13,8 @@
 Iterator over rows of an `AbstractDataFrame`,
 with each row represented as a `DataFrameRow`.
 
+Currently supports `DataFrame` and `SubDataFrame`.
+
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
 struct DataFrameRows{T} <: AbstractVector{DataFrameRow{SubIndex{Base.OneTo{Int},Base.OneTo{Int}}}}
@@ -21,10 +23,16 @@ struct DataFrameRows{T} <: AbstractVector{DataFrameRow{SubIndex{Base.OneTo{Int},
 end
 
 """
-    eachrow(df::AbstractDataFrame)
+    eachrow(df::Union{DataFrame, SubDataFrame})
 
-Return a `DataFrameRows` that iterates an `AbstractDataFrame` row by row,
+Return a `DataFrameRows` that iterates a data frame row by row,
 with each row represented as a `DataFrameRow`.
+
+`eachrow(df::DataFrame)` references the `df` directly, but
+for performance considerations `eachrow(df::SubDataFrame)` eagerly copies
+relevant column information from `df` (if columns of parent of a `SubDataFrame`
+are renamed or reallocated after calling `eachrow` this change will not be
+reflected by returned `DataFrameRows`).
 
 **Examples**
 
@@ -60,6 +68,15 @@ julia> copy.(eachrow(df))
  (x = 2, y = 12)
  (x = 3, y = 13)
  (x = 4, y = 14)
+
+julia> eachrow(view(df, [4,3], [2,1]))
+2-element DataFrames.DataFrameRows{Array{Int64,1}}:
+ DataFrameRow (row 4)
+y  14
+x  4
+ DataFrameRow (row 3)
+y  13
+x  3
 ```
 """
 eachrow(df::DataFrame) = DataFrameRows(df, :)
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -23,7 +23,7 @@ struct DataFrameRows{S<:AbstractDataFrame,T<:SubIndex} <: AbstractVector{DataFra
 end
 
 """
-    eachrow(df::Union{DataFrame,SubDataFrame})
+    eachrow(df::AbstractDataFrame)
 
 Return a `DataFrameRows` that iterates a data frame row by row,
 with each row represented as a `DataFrameRow`.
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -86,7 +86,7 @@ Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
 
 @inline Base.getindex(itr::DataFrameRows{DataFrame}, i::Int) =
     DataFrameRow(itr.df, itr.subindex, i)
-@inline Base.getindex(itr::DataFrameRows{SubDataFrame}, i::Int) =
+@inline Base.getindex(itr::DataFrameRows{<:SubDataFrame}, i::Int) =
     DataFrameRow(parent(itr.df), itr.subindex, rows(itr.df)[i])
 
 # Iteration by columns
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -22,7 +22,7 @@ view(df::AbstractDataFrame, row, cols)
 
 ### Notes
 
-A `DataFrameRow` supports iteration interface so you can pass it to functions
+A `DataFrameRow` supports the iteration interface and can therefore be passed to functions
 that expect a collection as an argument.
 
 Indexing is one-dimensional like specifying a column of a `DataFrame`.
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -107,7 +107,7 @@ Base.haskey(r::DataFrameRow, key::Integer) = 1 ≤ key ≤ size(r, 1)
 function Base.haskey(r::DataFrameRow, key::Symbol)
     haskey(parent(r), key) || return false
     pos = index(parent(r))[key]
-    remap = lazyremap!(r)
+    remap = lazyremap!(index(r))
     checkbounds(Bool, remap, pos) || return false
     remap[pos] > 0
 end
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -1,23 +1,24 @@
 """
-    DataFrameRow{<:AbstractDataFrame, <:SubIndex}
+    DataFrameRow{<:SubIndex}
 
 A view of one row of an `AbstractDataFrame`.
+Currently supports `DataFrame` and `SubDataFrame`.
 
 A `DataFrameRow` is constructed with `view` or `getindex` when one row and a
-selection of columns is requested.
-It is also returned when iterating the result of the call to the [`eachrow`](@ref) function.
+selection of columns is requested. It is also returned when iterating the result
+of the call to the [`eachrow`](@ref) function.
 
 ```julia
-view(df::AbstractDataFrame, row, cols)
 df[row, cols]
+view(df::AbstractDataFrame, row, cols)
 ```
 
 ### Arguments
 
-* `df` : an `AbstractDataFrame`
-* `row` : an `Integer` other than `Bool` indicating requested row number
+* `df` : a `DataFrame` or `SubDataFrame`
+* `row` : an `Integer` other than `Bool` indicating the requested row number
 * `cols` : any indexing type for columns, typically
-  `AbstractVector{Int}`, `AbstractVector{Bool}` or `AbstractVector{Symbol}` or a colon
+  a vector of `Int`, `Bool` or `Symbol`, or a colon
 
 ### Notes
 
@@ -26,8 +27,8 @@ that expect a collection as an argument.
 
 Indexing is one-dimensional like specifying a column of a `DataFrame`.
 
-It is possible to create a `DataFrameRow` with duplicate columns, but in such case
-an error will be thrown when one tries to access some column by name.
+It is possible to create a `DataFrameRow` with duplicate columns.
+All such columns will have a reference to the same entry in the parent `DataFrame`.
 """
 struct DataFrameRow{S<:SubIndex}
     df::DataFrame
@@ -89,8 +90,7 @@ end
 
 @inline Base.getindex(r::DataFrameRow, idx::ColumnIndex) =
     parent(r)[row(r), parentcols(r, idx)]
-@inline Base.getindex(r::DataFrameRow, idxs::Union{AbstractVector{<:Integer},
-                                           AbstractVector{Symbol}}) =
+@inline Base.getindex(r::DataFrameRow, idxs::AbstractVector}) =
     DataFrameRow(parent(r), row(r), parentcols(r, idxs))
 @inline Base.getindex(r::DataFrameRow, ::Colon) = r
 
@@ -100,6 +100,7 @@ end
 index(r::DataFrameRow) = getfield(r, :colindex)
 
 Base.names(r::DataFrameRow) = _names(parent(r))[parentcols(r, :)]
+_names(r::DataFrameRow) = view(_names(parent(r)), parentcols(r, :))
 
 Base.haskey(r::DataFrameRow, key::Bool) =
     throw(ArgumentError("invalid key: $key of type Bool"))
@@ -190,7 +191,7 @@ function Base.:(==)(r1::DataFrameRow, r2::DataFrameRow)
         index(r1).cols == index(r2).cols || return false
         row(r1) == row(r2) && return true
     else
-        names(r1) == names(r2) || return false
+        _names(r1) == _names(r2) || return false
     end
     all(((a, b),) -> a == b, zip(r1, r2))
 end
@@ -200,7 +201,7 @@ function Base.isequal(r1::DataFrameRow, r2::DataFrameRow)
         index(r1).cols == index(r2).cols || return false
         row(r1) == row(r2) && return true
     else
-        names(r1) == names(r2) || return false
+        _names(r1) == _names(r2) || return false
     end
     all(((a, b),) -> isequal(a, b), zip(r1, r2))
 end
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -43,7 +43,7 @@ end
                           "columns at indices $cols"))
     end
     # non existing mappings are either out range or invalid
-    remap = 1:last(cols) .- first(cols) .+ 1
+    remap = (1:last(cols)) .- first(cols) .+ 1
     DataFrameRow(df, row, cols, remap)
 end
 
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -50,6 +50,9 @@ DataFrameRow(df::DataFrame, colindex::SubIndex, row::Bool) =
     DataFrameRow(df, SubIndex(index(df), cols), row)
 end
 
+@inline function DataFrameRow(sdf::SubDataFrame, row::Bool, cols) =
+    throw(ArgumentError("invalid index: $row of type Bool"))
+
 @inline function DataFrameRow(sdf::SubDataFrame, row::Integer, cols)
     @boundscheck if !checkindex(Bool, axes(sdf, 1), row)
         throw(BoundsError("attempt to access a data frame with $(nrow(sdf)) " *
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -36,10 +36,10 @@ struct DataFrameRow{S<:SubIndex}
     row::Int
 end
 
-DataFrameRow(df::DataFrame, colindex::SubIndex, row::Integer) =
+@inline DataFrameRow(df::DataFrame, colindex::SubIndex, row::Integer) =
     DataFrameRow{typeof(colindex)}(df, colindex, row)
 
-DataFrameRow(df::DataFrame, colindex::SubIndex, row::Bool) =
+@inline DataFrameRow(df::DataFrame, colindex::SubIndex, row::Bool) =
     throw(ArgumentError("invalid index: $row of type Bool"))
 
 @inline function DataFrameRow(df::DataFrame, row::Integer, cols)
@@ -68,14 +68,14 @@ row(r::DataFrameRow) = getfield(r, :row)
 Base.parent(r::DataFrameRow) = getfield(r, :df)
 Base.parentindices(r::DataFrameRow) = (row(r), index(r).cols)
 
-Base.view(adf::AbstractDataFrame, rowind::Integer, ::Colon) =
+@inline Base.view(adf::AbstractDataFrame, rowind::Integer, ::Colon) =
     DataFrameRow(adf, rowind, :)
-Base.view(adf::AbstractDataFrame, rowind::Integer, colinds::AbstractVector) =
+@inline Base.view(adf::AbstractDataFrame, rowind::Integer, colinds::AbstractVector) =
     DataFrameRow(adf, rowind, colinds)
 
-Base.getindex(df::AbstractDataFrame, rowind::Integer, colinds::AbstractVector) =
+@inline Base.getindex(df::AbstractDataFrame, rowind::Integer, colinds::AbstractVector) =
     DataFrameRow(df, rowind, colinds)
-Base.getindex(df::AbstractDataFrame, rowind::Integer, ::Colon) =
+@inline Base.getindex(df::AbstractDataFrame, rowind::Integer, ::Colon) =
     DataFrameRow(df, rowind, :)
 
 @inline parentcols(r::DataFrameRow, idx::Union{Integer, AbstractVector{<:Integer}}) =
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -46,7 +46,7 @@ DataFrameRow(df::DataFrame, colindex::SubIndex, row::Bool) =
         throw(BoundsError("attempt to access a data frame with $(nrow(df)) " *
                           "rows at index $row"))
     end
-    DataFrameRow(df, SubIndex(df, cols), row)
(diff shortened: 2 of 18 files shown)


nalimilan:
Interesting approach. Sounds nice for `SubDataFrame`, but I'm concerned about performance for `DataFrameRow`. I think we need to distinguish two cases:
- The common case where all columns are selected. We really don't want to affect the performance of this case. Have you checked that performance doesn't regress there?
- The less common case where a subset of columns is selected. It's currently very slow already since we allocate a new data frame. But I'm not sure it makes a lot of sense to try to make things faster by allocating two (small) arrays rather than a data frame: anyway, if we allocate something for each row, it will be unreasonably slow. Though I guess benchmarking can tell whether the improvement is really significant. Ideally we would find a solution to make the allocation only once. In theory a combination of inlining and pure functions could do the trick, but `@pure` is probably not applicable here and we don't have an alternative yet. Some experimentation is in order.

bkamins:
> The common case where all columns are selected

This should be fast as in this case I allocate nothing (but I will benchmark it - it may be a few nanoseconds, which should be negligible). The difference is that now, when we make a `SubDataFrame` or a `DataFrameRow` and modify parent we have duality:
* if it was a `:` selection then all changes are reflected in a view
* if it was some different selection then no changes are reflected in a vieew

Under the proposed implementation we are consistent and track column indices selected from the parent at the time of construction (this means that things like column renaming are reflected in the view or adding columns to the parent when `:` was used does not change view).

The benchmarks I did just to make generally sure on the impact are:
```
julia> @benchmark 1:10
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.542 ns (0.00% GC)
  median time:      2.056 ns (0.00% GC)
  mean time:        2.015 ns (0.00% GC)
  maximum time:     23.646 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark zeros(Int, 100)
BenchmarkTools.Trial:
  memory estimate:  896 bytes
  allocs estimate:  1
  --------------
  minimum time:     74.024 ns (0.00% GC)
  median time:      86.362 ns (0.00% GC)
  mean time:        131.177 ns (23.88% GC)
  maximum time:     2.311 μs (86.59% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark DataFrame()
BenchmarkTools.Trial:
  memory estimate:  1.61 KiB
  allocs estimate:  16
  --------------
  minimum time:     348.389 ns (0.00% GC)
  median time:      470.199 ns (0.00% GC)
  mean time:        807.188 ns (35.75% GC)
  maximum time:     34.534 μs (98.05% GC)
  --------------
  samples:          10000
  evals/sample:     211
```

So in general allocating an empty `DataFrame` (which has to be done in ) is much slower than allocating a vector with 100 columns filled with zeros so we should see the slowdown only in the case: original data frame has ~1000 columns, we want to select from it a non-continuous range of columns that is very small (like 10), then it would be better to use `Dict` than array but I think that this should not be a very common case.

>  Ideally we would find a solution to make the allocation only once.

Yes, but what would be the use case of this? E.g. in `eachrow` we select all columns anyway. Is the case in `groupby` different? I thought we also return all columns there.

In general I could provide a direct unsafe constructor for `DataFrameRow` and `SubDataFrame` that allows passing all struct parameters directly so it is enough they are computed once if we know.

nalimilan:
I agree that this shouldn't make performance worse in any case, but better benchmark this directly to be sure. I'd also like to see whether the performance improvement for the case where we select a subset of columns is really large or not.

> Yes, but what would be the use case of this? E.g. in `eachrow` we select all columns anyway. Is the case in `groupby` different? I thought we also return all columns there.

Well, that applies to all situations where you'd do `view(df, row, cols)`, which is the use case justifying this PR. If we don't care about it, we may as well throw an error. (`groupby` doesn't need this, and we could select the subset of columns manually anyway.)

bkamins:
Here are the benchmarks. The only really bad timing I get is `@btime view($df, 20, :);`. And I do not understand it as `@btime view($df, 2:3, :);` is two times faster but essentially does the same job. I have checked that removing bounds checking inside a constructor drops time down to 6ns, but doing `@inbounds view($df, 20, :)` does not disable them (so this means I do not understand something here 😢).

**THIS PR**
```

julia> df = DataFrame(rand(50, 50));

julia> @btime view($df, 2:3, :);
  21.590 ns (1 allocation: 64 bytes)

julia> @btime view($df, 20:30, 20:30);
  20.562 ns (1 allocation: 64 bytes)

julia> @btime view($df, [2,3,4,5], [2,3,4,5]);
  132.041 ns (5 allocations: 832 bytes)

julia> @btime view($df, 20, :);
  39.068 ns (1 allocation: 64 bytes)

julia> @btime view($df, 20, 20:30);
  39.582 ns (1 allocation: 64 bytes)

julia> @btime view($df, 2, [2,3,4,5]);
  116.629 ns (3 allocations: 704 bytes)

julia> @btime DataFrameRow($df, 2, [1:50;]);
  225.789 ns (3 allocations: 1.11 KiB)

julia> @btime DataFrameRow($df, 2, [1:50;]);
  174.373 ns (3 allocations: 1.11 KiB)

julia> @btime $dfr[10];
  23.646 ns (1 allocation: 16 bytes)

julia> @btime $dfr[:x1];
  34.955 ns (1 allocation: 16 bytes)

julia> @btime $dfr[[10,11]];
  160.354 ns (5 allocations: 800 bytes)

julia> @btime $dfr[[:x10,:x11]];
  185.396 ns (6 allocations: 832 bytes)
```

**MASTER**

```

julia> df = DataFrame(rand(50, 50));

julia> @btime view($df, 2:3, :);
  23.132 ns (1 allocation: 32 bytes)

julia> @btime view($df, 20:30, 20:30);
  2.113 μs (28 allocations: 4.30 KiB)

julia> @btime view($df, [2,3,4,5], [2,3,4,5]);
  1.028 μs (26 allocations: 2.08 KiB)

julia> @btime view($df, 20, :);
  22.618 ns (1 allocation: 32 bytes)

julia> @btime view($df, 20, 20:30);
  2.113 μs (28 allocations: 4.30 KiB)

julia> @btime view($df, 2, [2,3,4,5]);
  1.028 μs (25 allocations: 1.97 KiB)

julia> @btime DataFrameRow($df[1:50], 2);
  7.454 μs (34 allocations: 12.97 KiB)

julia> @btime $dfr[10];
  20.562 ns (1 allocation: 16 bytes)

julia> @btime $dfr[:x1];
  28.787 ns (1 allocation: 16 bytes)

julia> @btime $dfr[[10,11]];
  819.886 ns (25 allocations: 1.89 KiB)

julia> @btime $dfr[[:x10,:x11]];
  888.875 ns (28 allocations: 2.03 KiB)
```

bkamins:
The code is now good to look at. It passess tests. The failing PRs are due to https://github.com/queryverse/TableTraits.jl/commit/a590a1f74f033d72d73b9ee37f18394a0f5d8c9c#diff-6e4ca1ab98f0066326533b45809deb75.

bkamins:
OK - I have fixed `axes` implementation and exposed inner `DataFrameRow` constructor (in cases we would want to reuse precomputed values). Now the tests above work with `@inbounds` and the timings are:
```
julia> @btime DataFrameRow($df, 20, 1:10);
  22.618 ns (1 allocation: 64 bytes)

julia> @btime DataFrameRow($df, 20, :);
  22.619 ns (1 allocation: 48 bytes)

julia> @btime @inbounds DataFrameRow($df, 20, :);
  5.654 ns (1 allocation: 48 bytes)

julia> @btime @inbounds DataFrameRow($df, 20, 1:10);
  5.654 ns (1 allocation: 64 bytes)
```

(I have temporarily disabled testing the lines offended by TableTraits.jl)

bkamins:
If we agree on the design I will write comprehensive tests for this functionality as currently test coverage of `SubDataFrame` and `DataFrameRow` is not so good.

bkamins:
Also a question if we want to have for `SubDataFrame` constructor a similar approach as we have for `DataFrameRow` now (i.e. exposing the inner constructor).

bkamins:
Also note that this PR actually makes views of `DataFrame` useful, i.e. it is worth to use them for performance reasons as with views in Base, e.g.:
```
julia> @benchmark $df[1:10]
BenchmarkTools.Trial:
  memory estimate:  1.98 KiB
  allocs estimate:  21
  --------------
  minimum time:     1.593 μs (0.00% GC)
  median time:      1.748 μs (0.00% GC)
  mean time:        2.687 μs (22.19% GC)
  maximum time:     4.020 ms (99.89% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> @benchmark $view(df, :, 1:10)
BenchmarkTools.Trial:
  memory estimate:  96 bytes
  allocs estimate:  2
  --------------
  minimum time:     136.225 ns (0.00% GC)
  median time:      138.795 ns (0.00% GC)
  mean time:        163.452 ns (6.23% GC)
  maximum time:     48.924 μs (99.59% GC)
  --------------
  samples:          10000
  evals/sample:     800

julia> @benchmark $df[1:10, 1:10]
BenchmarkTools.Trial:
  memory estimate:  4.02 KiB
  allocs estimate:  42
  --------------
  minimum time:     2.228 μs (0.00% GC)
  median time:      2.742 μs (0.00% GC)
  mean time:        4.109 μs (20.74% GC)
  maximum time:     5.341 ms (99.87% GC)
  --------------
  samples:          10000
  evals/sample:     9

julia> @benchmark $view(df, 1:10, 1:10)
BenchmarkTools.Trial:
  memory estimate:  128 bytes
  allocs estimate:  3
  --------------
  minimum time:     46.266 ns (0.00% GC)
  median time:      48.321 ns (0.00% GC)
  mean time:        66.242 ns (17.00% GC)
  maximum time:     45.217 μs (99.78% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

bkamins:
and a second note is that `df[1, cols]` returns a `DataFrameRow` so we should make it fast as it is a fairly common operation.

bkamins:
OK - I have also exposed `SubDataFrame` inner constructor and added some additional tests.

src/dataframerow/dataframerow.jl
```diff
@@ -133,57 +233,28 @@ function rowhash(cols::Tuple{Vararg{AbstractVector}}, r::Int, h::UInt = zero(UIn
...
-# only the rows of the same DataFrame could be compared
-# rows are equal if they have the same values (while the row indices could differ)
-# if all non-missing values are equal, but there are missings, returns missing
-Base.:(==)(r1::DataFrameRow, r2::DataFrameRow) = isequal(r1, r2)
+    rowhash(ntuple(col -> parent(r)[getparentcols(r, col)], length(r)), row(r), h)
 
-function Base.isequal(r1::DataFrameRow, r2::DataFrameRow)
-    isequal_row(parent(r1), row(r1), parent(r2), row(r2))
+function Base.:(==)(r1::DataFrameRow, r2::DataFrameRow)
+    names(r1) == names(r2) || return false
```
nalimilan: Better check whether the `cols` fields are equal when `parent(r1) === parent(r2) && row(r1) == row(r2)` as a fast path, and only if parents differ check whether names are equal.

src/dataframerow/dataframerow.jl
```diff
@@ -133,57 +233,28 @@ function rowhash(cols::Tuple{Vararg{AbstractVector}}, r::Int, h::UInt = zero(UIn
...
-    (ncol(parent(r1)) == ncol(parent(r2))) ||
-        throw(ArgumentError("Rows of the data tables that have different number of columns cannot be compared ($(ncol(df1)) and $(ncol(df2)))"))
-    @inbounds for i in 1:ncol(parent(r1))
-        if !isequal(parent(r1)[i][row(r1)], parent(r2)[i][row(r2)])
-            return isless(parent(r1)[i][row(r1)], parent(r2)[i][row(r2)])
-        end
+    length(r1) == length(r2) ||
+        throw(ArgumentError("compared DataFrameRows must have the same number " *
+                            "of columns (got $(length(r1)) and $(length(r2)))"))
+    all(((a, b),) -> isless(a, b), zip(r1, r2))
```
nalimilan: ?

src/subdataframe/subdataframe.jl
```diff
@@ -1,154 +1,208 @@
...
+Base.parentindices(sdf::SubDataFrame) = (rows(sdf), getfield(sdf, :cols))
 
 Base.view(adf::AbstractDataFrame, colinds) = view(adf, :, colinds)
 Base.view(adf::AbstractDataFrame, rowinds, colind::ColumnIndex) =
     view(adf[colind], rowinds)
 Base.view(adf::AbstractDataFrame, rowinds, colind::Bool) =
     throw(ArgumentError("invalid column index $colind of type `Bool`"))
 Base.view(adf::AbstractDataFrame, rowinds, colinds) =
-    SubDataFrame(adf[colinds], rowinds)
-Base.view(adf::AbstractDataFrame, rowinds, ::Colon) = SubDataFrame(adf, rowinds)
```
nalimilan: Where is this method gone?

src/subdataframe/subdataframe.jl
```diff
@@ -1,154 +1,208 @@
...
 
 ##############################################################################
 ##
 ## AbstractDataFrame interface
 ##
 ##############################################################################
 
-index(sdf::SubDataFrame) = index(parent(sdf))
+struct SubIndex{T,S} <: AbstractIndex
+    sdf::SubDataFrame{T,S}
```
nalimilan: Why not have a single parameter for the `SubDataFrame`?

Or probably more logical: store the parent `Index` and the `cols`. Then you can define this in index.jl. That's more logical since `DataFrame` is based on `Index`, not the other way around, so `SubIndex` shouldn't be based on `SubDataFrame`.

src/subdataframe/subdataframe.jl
```diff
@@ -1,154 +1,208 @@
...
+end
+
+index(sdf::SubDataFrame{T,S}) where {T,S} = SubIndex{T,S}(sdf)
+
+Base.length(x::SubIndex) = length(getfield(x.sdf, :cols))
```
nalimilan:
```suggestion
Base.length(x::SubIndex) = ncol(x.sdf)
```

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,185 @@
...
-    DataFrameRow(parent(sdf)[colinds], rows(sdf)[rowind])
+    DataFrameRow(parent(sdf), rows(sdf)[rowind], index(sdf)[colinds])
 Base.getindex(df::DataFrame, rowind::Integer, ::Colon) =
-    DataFrameRow(df, rowind)
+    DataFrameRow(df, rowind, :)
 Base.getindex(sdf::SubDataFrame, rowind::Integer, ::Colon) =
-    DataFrameRow(parent(sdf), rows(sdf)[rowind])
+    DataFrameRow(parent(sdf), rows(sdf)[rowind], parentindices(sdf)[2])
+
+getparentcols(r::DataFrameRow, idx::Union{Integer, AbstractVector{<:Integer}}) =
```
nalimilan: I'd call this `parentcols` (we generally avoid "get" since it doesn't add anything).

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,185 @@
 """
-    DataFrameRow{<:AbstractDataFrame}
+    DataFrameRow{<:AbstractDataFrame, <:AbstractVector{Int}}
+
+A view of one row of an `AbstractDataFrame`.
+
+A `DataFrameRow` is meant to be constructed with `view` when one row
+and a selection of columns is requested.
+A `DataFrameRow` is typically appears when `getindex` is called on an
```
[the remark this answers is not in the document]

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,185 @@
...
+`AbstractDataFrame` and one row is selected and is returned when iterating
+the result of the call to the `eachrow` function.
+
+```julia
```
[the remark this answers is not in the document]
bkamins: I just removed it. I also remove the same section in `SubDataFrame` then. I have added **examples** section for `eachrow` as it was missing.

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,185 @@
...
+@inline function DataFrameRow(df::AbstractDataFrame, row::Integer, ::Colon)
+    @boundscheck if !checkindex(Bool, axes(df, 1), row)
+        throw(BoundsError("attempt to access a data frame with $(nrow(df)) " *
+                          "rows at index $row"))
+    end
+    cols = axes(df, 2)
+    DataFrameRow(df, row, cols, cols)
+end
+
+@inline DataFrameRow(df::AbstractDataFrame, row::Integer, cols::AbstractVector{Int}) =
```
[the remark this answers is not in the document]
bkamins: We may disallow it, but e.g. `cols` could be a view or a sparse vector (in this case conversion will be slow) or a `Base.Slice` (here conversion will be fast).
Probably these are not very common cases but I do not see why it hurts to keep them.

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,185 @@
...
+    DataFrameRow(df, row, convert(Vector{Int}, cols))
+@inline DataFrameRow(df::AbstractDataFrame, row::Integer, cols::AbstractUnitRange{Int}) =
+    DataFrameRow(df, row, convert(UnitRange{Int}, cols))
+@inline DataFrameRow(df::AbstractDataFrame, row::Integer, cols) =
```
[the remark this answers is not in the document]
bkamins: A good point - fixed.

src/abstractdataframe/iteration.jl
```diff
@@ -31,7 +31,7 @@ Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
 @inline function Base.getindex(itr::DataFrameRows, i::Int)
     @boundscheck checkbounds(itr, i)
-    return DataFrameRow(itr.df, i)
+    return DataFrameRow(itr.df, i, :)
```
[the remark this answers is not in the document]
bkamins: I have it now in deprecated.jl, but I can move it back.

The reason why I did not allow it is that I do the same for `SubDataFrame` (i.e. deprecate two-argument version) because there is a confusion if the index means a row or a column (as for an `AbstractDataFrame` normally single index means a column).

I assume that this constructor will not be used directly in user's code, as it is shorter to write `df[i, :]`, but I wanted to have an explicit call to the constructor here so that it is clearer in the code what we do.

In summary - I would leave the deprecation, because directly calling `DataFrameRow` constructor is not something that is needed in user's code.

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,185 @@
...
 Base.getindex(sdf::SubDataFrame, rowind::Integer, ::Colon) =
-    DataFrameRow(parent(sdf), rows(sdf)[rowind])
+    DataFrameRow(parent(sdf), rows(sdf)[rowind], parentindices(sdf)[2])
+
+getparentcols(r::DataFrameRow, idx::Union{Integer, AbstractVector{<:Integer}}) =
+    getfield(r, :cols)[idx]
+
+function getparentcols(r::DataFrameRow, idx::Symbol)
+    parentcols = index(parent(r))[idx]
+    getfield(r, :remap)[parentcols] == 0 && throw(KeyError("$idx not found"))
```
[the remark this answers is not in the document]
bkamins: OK - I will make it a lazy operation that is performed only when needed and cached.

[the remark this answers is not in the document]
bkamins: But in `haskey` I would leave this check unconditionally as it is the whole point of `haskey` to check it.

src/dataframerow/dataframerow.jl
bkamins: Agreed.

src/dataframerow/dataframerow.jl
bkamins: A shameful of copy-paste.

src/subdataframe/subdataframe.jl
bkamins: It is now handled by `Base.view(adf::AbstractDataFrame, rowinds, colinds)` catch all.

src/subdataframe/subdataframe.jl
bkamins: OK - moved

src/subdataframe/subdataframe.jl
bkamins: This would create a cyclic reference to `ncol(sdf::SubDataFrame) = length(index(sdf))`. Anyway - it is different now as I moved `SubIndex` to index.jl.

src/dataframerow/dataframerow.jl
bkamins: OK

bkamins:
I have adjusted the code to your comments. There are two minor things unresolved. When it goes through CI (the changes were significant so something might have slipped I will report the benchmark results against master and we can decide if we go forward with this PR).

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,185 @@
...
 """
-struct DataFrameRow{T<:AbstractDataFrame}
+struct DataFrameRow{T<:AbstractDataFrame, S<:AbstractVector{Int}}
     df::T
     row::Int
+    cols::S
+    remap::S # inverse of cols, it is of type S for efficiency in most common cases
+end
+
+DataFrameRow(df::AbstractDataFrame, row::Bool, cols::S, remap::S) where {S<:AbstractVector{Int}} =
```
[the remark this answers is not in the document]
bkamins: I had to revert it as otherwise much more code is needed to avoid method ambiguities.

bkamins:
All works, but please hold on with the review as I will do a bit of refactoring of the code today. I will let you know when I am finished.

bkamins:
Here is a list of changes (I had cascading things that needed to be fixed to form a common design):
* `axes(adf, i)` is more efficient
* `eachrow(df)` assumes that number of columns in the `df` does not change after the call (this improves iteration performance because we can remember the `SubIndex` and ruse it and should not be problematic I think)
* `summary` function now prints only type name (without its parameters as they got longer and the output was not nice)
* `ColumnIndex` is moved to `index.jl` and now does not allow `Bool` (in this way we can simplify the design and stop having to think all the time if we handled this corner case)
* parent of `DataFrameRow` and `SubDataFrame` is always source `DataFrame` it references to
* `DataFrameRow` and `SubDataFrame` allow for duplicate columns (they are views of the same column in parent so they are effectively indistinguishable and have the same name; of course materializing such a view will throw an error)
* `DataFrameRow` and `SubDataFrame` constructors and `getindex` all support disabling checking bounds and make sure they are inlined (without inlining there were some performance regressions)
* fixed a bug in `getindex(sdf::SubDataFrame, rowind::Integer, colinds::AbstractVector)` and `view(sdf::SubDataFrame, rowind::Integer, colinds::AbstractVector)`
* `DataFrameRow` and `SubDataFrame` use a new `SubIndex` type that is a view of `Index`
* `SubIndex` remembers column numbers in the parent (this means that column renaming is reflected in the view)
* `==`, `isequal` and `isless` fro `DataFrameRow` now work in a standard way for Base (i.e. after converting two `DataFrameRow`s to two `NamedTuple`s using `copy` the results of the operations should be the same)
* moved row equality tests based on tuples to utils.jl
* deprecated `DataFrameRow(df, row)` and  `SubDataFrame(df, rows)` as they were inconsistent with standard indexing in DataFrames.jl (single index indexing normally refers to columns which could be confusing)
* improved performance of comparison of `AbstractIndex`
* improved `merge`, `merge!` and `add_names` for indexes
* added an option to `getindex` from `AbstractIndex` using `Colon`
* documentation updates and fixes
* test updates and significantly increased coverage
* minor code layout corrections

src/subdataframe/subdataframe.jl
```diff
@@ -1,154 +1,160 @@
...
 
-A view of row subsets of an AbstractDataFrame
+A view of row subsets of an `AbstractDataFrame`.
+Currently supports `DataFrame` and `SubDataFrame`.
 
-A `SubDataFrame` is meant to be constructed with `view`.  A
-SubDataFrame is used frequently in split/apply sorts of operations.
+A `SubDataFrame` is meant to be constructed with `view` when a collection of
+rows and columns is selected.
+A `SubDataFrame` is used frequently in split/apply sorts of operations.
```
nalimilan: While you're at it:
```suggestion
 It is also created by some methods of [`by`](@ref) and [`combine`](@ref).
```

src/subdataframe/subdataframe.jl
```diff
@@ -158,11 +164,11 @@ end
...
 ##############################################################################
 
-Base.copy(sdf::SubDataFrame) = parent(sdf)[rows(sdf), :]
-
-without(sdf::SubDataFrame, c) = view(without(parent(sdf), c), rows(sdf), :)
-# Resolve a method ambiguity
-without(sdf::SubDataFrame, c::Vector{<:Integer}) = view(without(parent(sdf), c), rows(sdf), :)
+Base.copy(sdf::SubDataFrame) = parent(sdf)[rows(sdf), parentcols(sdf, :)]
 
+function without(df::SubDataFrame, icols::Vector{<:Integer})
```
nalimilan:
```suggestion
function without(df::SubDataFrame, icols::AbstractVector{<:Integer})
```
Couldn't this method be common to all `AbstractDataFrame` types?

test/dataframe.jl
```diff
@@ -784,10 +784,12 @@ module TestDataFrame
...
         df = DataFrame()
-        @test_throws ArgumentError df[true] = 1
-        @test_throws ArgumentError df[true] = [1,2,3]
-        @test_throws ArgumentError df[1:2, true] = [1,2]
-        @test_throws ArgumentError df[1, true] = 1
+        @test_throws MethodError df[true] = 1
+        @test_throws MethodError df[true] = [1,2,3]
+        @test_throws MethodError df[1:2, true] = [1,2]
+        @test_throws MethodError df[1, true] = 1
+        @test_throws ErrorException df[1, 100] = 1
```
nalimilan: `BoundsError` would be more appropriate.

test/dataframe.jl
```diff
@@ -969,9 +971,18 @@ module TestDataFrame
...
-    @testset "parent" begin
+    @testset "parent, size and axes" begin
         x = DataFrame(a = [1, 2, 3], b = [4, 5, 6])
         @test parent(x) === x
         @test parentindices(x) === (Base.OneTo(3), Base.OneTo(2))
+        @test size(x) == (3,2)
+        @test size(x, 1) == 3
+        @test size(x, 2) == 2
+        @test_throws ArgumentError size(x, 3)
+        @test axes(x) == (Base.OneTo(3), Base.OneTo(2))
```
nalimilan: Use `===`, or just `(1:3, 1:2)`.

test/dataframerow.jl
```diff
@@ -7,50 +7,109 @@ module TestDataFrameRow
...
                    d=CategoricalArray([:A, missing, :C, :A, missing, :C]))
     df2 = DataFrame(a = [1, 2, 3])
+    sdf = view(df, [5, 3], [3, 1, 2])
+
+    @test names(DataFrameRow(df, 1, :)) == [:a, :b, :c, :d]
+    @test names(DataFrameRow(df, 3, [3, 2])) == [:c, :b]
+    @test copy(DataFrameRow(df, 3, [3, 2])) == (c = "C", b = 1.2)
+    @test copy(DataFrameRow(sdf, 2, [3, 2])) == (b = 1.2, a = 3)
+    @test copy(DataFrameRow(sdf, 2, :)) == (c = "C", a = 3, b = 1.2)
+    @test DataFrameRow(df, 3, [3, 2]) == df[3, [3, 2]] == view(df, 3, [3, 2])
```
nalimilan: Use `===`?

test/dataframerow.jl
```diff
@@ -7,50 +7,109 @@ module TestDataFrameRow
...
+    @test_throws MethodError DataFrameRow(df, true, 1:2)
+    if VERSION ≥ v"1.0.0"
+        # this test throws a warning on Julia 0.7
+        @test_throws ArgumentError DataFrameRow(sdf, true, 1:2)
+    end
 
-    @test names(DataFrameRow(df, 1)) == [:a, :b, :c, :d]
+    # getindex
+    r = DataFrameRow(df, 2, :)
+    @test r[:] == r
```
nalimilan: Would `===` apply here too?

nalimilan:
Thanks. Do you think something can be done to fix the few (slight) regressions?

bkamins:
The slight regressions are AFAICT due to the fact that we create a separete `SubIndex` object following our discussion (and not keeping all information in `SubDataFrame` or `DataFrameRow` directly). I am convinced that your suggestion is a better design, but it has ~5ns penalty. I have no idea how to avoid it except moving back to the old design with one object created.

This penalty means that on 10^9 operations creating some object we loose ~5 seconds. I would say this should be acceptable as normally if you have to create that many objects you probably use `eachrow` and I avoid this penalty there - the new approach is faster than the old one in this case (and in general the inner constructor is exposed to allow users to handle this case if it is performance critical).

Also note that the old design did not allow for `@inbounds`. I will fix `@inline` with `@propagate_inbounds` in a few places in the new design and we will have a fully working `@inbounds`, which should be significantly faster (the most expensive check we do is `nrow(df)` which is not type stable and takes ~12ns).

src/subdataframe/subdataframe.jl
bkamins: It can (in fact - I initially removed it, but then mistakengly thought that `sdf[cols]` loses a view - but it returns `SubDataFrame` now so all should be correct).

src/dataframerow/dataframerow.jl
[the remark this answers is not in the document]
bkamins: OK - now I get what you wanted. Removed here and in `SubDataFrame`.

src/dataframerow/dataframerow.jl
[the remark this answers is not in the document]
bkamins: OK to keep it?

src/abstractdataframe/iteration.jl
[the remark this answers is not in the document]
bkamins: Do you agree to deprecate two-argument versions?

src/abstractdataframe/iteration.jl
```diff
@@ -8,31 +8,86 @@
 
 # Iteration by rows
 """
-    DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow{T}}
+    DataFrameRows <: AbstractVector{DataFrameRow}
```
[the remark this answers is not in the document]

src/abstractdataframe/iteration.jl
```diff
@@ -8,31 +8,86 @@
 
 # Iteration by rows
 """
-    DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow{T}}
```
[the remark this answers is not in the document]

src/abstractdataframe/iteration.jl
```diff
@@ -8,31 +8,86 @@
...
+│ Row │ x     │ y     │
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 11    │
+│ 2   │ 2     │ 12    │
+│ 3   │ 3     │ 13    │
+│ 4   │ 4     │ 14    │
+
+julia> eachrow(df)
+4-element DataFrames.DataFrameRows{DataFrame,DataFrames.SubIndex{Base.OneTo{Int64},Base.OneTo{Int64}}}:
```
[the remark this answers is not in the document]
bkamins: fixed - I print "DataFrameRows" only.

src/abstractdataframe/iteration.jl
```diff
@@ -8,31 +8,86 @@
...
+
+julia> copy.(eachrow(df))
+4-element Array{NamedTuple{(:x, :y),Tuple{Int64,Int64}},1}:
+ (x = 1, y = 11)
+ (x = 2, y = 12)
+ (x = 3, y = 13)
+ (x = 4, y = 14)
+
+julia> eachrow(view(df, [4,3], [2,1]))
+2-element DataFrames.DataFrameRows{Array{Int64,1}}:
```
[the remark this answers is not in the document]

src/abstractdataframe/iteration.jl
```diff
@@ -8,31 +8,86 @@
...
 
+Currently supports `DataFrame` and `SubDataFrame`.
+
 A value of this type is returned by the [`eachrow`](@ref) function.
 """
-struct DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow{T}}
-    df::T
+struct DataFrameRows{S<:AbstractDataFrame,T<:SubIndex} <: AbstractVector{DataFrameRow{T}}
+    df::S
+    subindex::T
```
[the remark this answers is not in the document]
bkamins: actually this is a very good point. It will improve the performance.

bkamins:
I am half-way through your comments, but the update on my answer above is that actually it is possible to avoid the performance regression in cases when `:` is used. The reason is that in this case we can avoid allocation of `SubIndex` and we can simply use `Index` instead and it is already allocated. Also the consequence is that `eachrow` as previously will reflect changes of the structure in the underlying data frame.

src/abstractdataframe/iteration.jl
```diff
@@ -8,31 +8,86 @@
...
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
-@inline function Base.getindex(itr::DataFrameRows, i::Int)
-    @boundscheck checkbounds(itr, i)
-    return DataFrameRow(itr.df, i)
-end
+Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
+
+@inline Base.getindex(itr::DataFrameRows{DataFrame}, i::Int) =
+    DataFrameRow(itr.df, itr.subindex, i)
+@inline Base.getindex(itr::DataFrameRows{<:SubDataFrame}, i::Int) =
```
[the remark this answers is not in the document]
bkamins: Yes - assuming the additional new `AbstractDataFrame` defines an appropriate `DataFrameRow` constructor.

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,135 @@
 """
-    DataFrameRow{<:AbstractDataFrame}
+    DataFrameRow{<:SubIndex,<:AbstractDataFrame}
```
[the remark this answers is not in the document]
bkamins: Good point - I allowed it and it speeds up things with colon selection.

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,135 @@
...
+that expect a collection as an argument.
+
+Indexing is one-dimensional like specifying a column of a `DataFrame`.
+
+It is possible to create a `DataFrameRow` with duplicate columns.
+All such columns will have a reference to the same entry in the parent `DataFrame`.
 """
-struct DataFrameRow{T<:AbstractDataFrame}
-    df::T
+# We allow D to be AbstractDataFrame, to allow for extensions
```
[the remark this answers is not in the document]
bkamins: OK - removed the comment and fixed the order.

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,135 @@
...
+* `row` : an `Integer` other than `Bool` indicating the requested row number
+* `cols` : any indexing type for columns, typically
+  a vector of `Int`, `Bool` or `Symbol`, or a colon
+
+### Notes
+
+A `DataFrameRow` supports iteration interface so you can pass it to functions
+that expect a collection as an argument.
+
+Indexing is one-dimensional like specifying a column of a `DataFrame`.
```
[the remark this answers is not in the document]
bkamins: I have added some comment, but in general we do not document it now even for `AbstractDataFrame` properly so I would leave a larger review of the documentation for later. Note, that it is not easy to write precisely what we allow as `NamedTuple` has actually different (I would say worse) indexing behavior now (see https://github.com/JuliaLang/julia/issues/29417).

src/other/index.jl
```diff
@@ -3,6 +3,8 @@
 # through cleanly.
 abstract type AbstractIndex end
 
+const ColumnIndex = Union{Signed, Unsigned, Symbol}
```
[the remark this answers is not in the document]
bkamins: The point is that `Integer` has three subtypes `Bool`, `Signed` and `Unsigned` so we exclude only `Bool`. Except `Bool` all other types (or possibly new user-defined types) are either `Signed` or `Unsigned`. So unless someone in the future introduces some weird design (i.e. incorrectly subtyping from `Integer`) we are safe. In particular:
```
julia> supertype(BigInt)
Signed
```

We cannot allow for `Integer` and not handle `Bool` as this is one of the transitions Julia made from 0.6 to 1.0 to disallow `Bool` indexing.

In summary - I would leave it as is. OK?

src/other/index.jl
```diff
@@ -208,3 +211,79 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
...
+# We allow I to be AbstractIndex, to allow for extensions
+# In DataFrames.jl I is always Index
+
+struct SubIndex{S<:AbstractVector{Int}, T<:AbstractVector{Int}, I<:AbstractIndex} <: AbstractIndex
+    parent::I
+    cols::S # columns from idx selected in SubIndex
+    remap::T # reverse of cols
+end
+
+@inline function SubIndex(parent::Index, ::Colon)
```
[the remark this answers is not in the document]
bkamins: We can use a broader signature.

src/other/index.jl
```diff
@@ -208,3 +211,79 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
...
+        fill!(remap, 0)
+        for (i, col) in enumerate(x.cols)
+            remap[col] = i
+        end
+    end
+    remap
+end
+
+Base.length(x::SubIndex) = length(x.cols)
+Base.names(x::SubIndex) = copy(_names(x))
```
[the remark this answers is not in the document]
bkamins: I have opened an issue for this https://github.com/JuliaData/DataFrames.jl/issues/1655.

src/other/index.jl
```diff
@@ -208,3 +211,79 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
...
+_names(x::SubIndex) = view(_names(x.parent), x.cols)
+
+function Base.haskey(x::SubIndex, key::Symbol)
+    haskey(x.parent, key) || return false
+    pos = x.parent[key]
+    remap = lazyremap!(x)
+    checkbounds(Bool, remap, pos) || return false
```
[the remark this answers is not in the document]
bkamins: I have thought about it, but the whole point of this function is to check if something exists (we do not throw an error here but return `false`) so I would leave it as is. OK?

src/other/index.jl
```diff
@@ -208,3 +211,79 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
...
+    remap = (1:l) .- f .+ 1
+    SubIndex(parent, cols, remap)
+end
+
+@inline function SubIndex(parent::Index, cols::AbstractVector{Int})
+    ncols = length(parent)
+    @boundscheck if !all(x -> 0 < x ≤ ncols, cols)
+        throw(BoundsError("invalid columns $cols selected"))
+    end
+    remap = Int[]
```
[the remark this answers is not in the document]
bkamins: I have to keep it `Int[]`. Now I remember why. The reason it that `SubIndex` is immutable so I have to keep `remap` as the same binding.

src/other/index.jl
```diff
@@ -208,3 +211,79 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
...
+end
+
+@inline SubIndex(parent::Index, cols::ColumnIndex) =
+    throw(ArgumentError("invalid column argument $cols"))
+
+@inline SubIndex(parent::Index, cols) = SubIndex(parent, parent[cols])
+
+# a helper function that lazily creates remap when needed
+# it assumes that remap is Vector{Int} unless it is an AbstractUnitRange
+@inline function lazyremap!(x::SubIndex)
```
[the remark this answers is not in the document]
bkamins: I will add an appropriate check of `remap` length and avoid inlining.

src/other/index.jl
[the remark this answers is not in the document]
bkamins: see above (I would leave it).

src/subdataframe/subdataframe.jl
```diff
@@ -1,154 +1,160 @@
 """
-    SubDataFrame{<:AbstractVector{Int}} <: AbstractDataFrame
+    SubDataFrame{<:AbstractVector{Int},<:SubIndex,<:AbstractDataFrame} <: AbstractDataFrame
 
-A view of row subsets of an AbstractDataFrame
+A view of row subsets of an `AbstractDataFrame`.
+Currently supports `DataFrame` and `SubDataFrame`.
```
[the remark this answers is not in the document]
bkamins: I removed the restriction. I simply meant that new `AbstractDataFrame` has to implement a custom method to make a `SubDataFrame` (and the same for `DataFrameRow`).

test/dataframerow.jl
bkamins: It will fail as `[3,2]` is mutable. I will add a test for `===` that goes through.

test/dataframerow.jl
bkamins: yes

bkamins:
@nalimilan I have pushed a working version that has good performance. In order to reach it I aggressively try to reuse parent `Index`. This means that the code got a bit more complex to reason about. I will go back to it in a few days and check it again with a fresh head.

In the mean-time, if you would want and have time, please comment (or resolve the questions I have left unresolved to track decision points).

Here are current benchmarks (we are left with one minor regression when we access one column of a `SubDataFrame`/`DataFrameRow` but it is offset by massive time savings on creation of these objects):

operation | new | master
---|---:|---:
view($df, 2:3, :) | 25 | 24
@inbounds view($df, 2:3, :) | 5 | 24
view($df, 20:30, 20:30) | 28 | 2170
@inbounds view($df, 20:30, 20:30) | 11 | 2170
view($df, [2,3,4,5], [2,3,4,5]) | 100 | 1079
@inbounds view($df, [2,3,4,5], [2,3,4,5]) | 77 | 1079
view($df, 20, :) | 23 | 25
@inbounds view($df, 20, :) | 5 | 25
view($df, 20, 20:30) | 28 | 2113
@inbounds view($df, 20, 20:30) | 12 | 2170
view($df, 2, [2,3,4,5]) | 71 | 1028
@inbounds view($df, 2, [2,3,4,5]) | 53 | 1028
$dfr1[10] | 23 | 22
@inbounds $dfr1[10] | 22 | 22
$dfr1[:x1] | 34 | 30
@inbounds $dfr1[:x1] | 32 | 30
$dfr1[[10,11]] | 104 | 835
@inbounds $dfr1[[10,11]] | 78 | 830
$dfr1[[:x10,:x11]] | 130 | 893
@inbounds $dfr1[[:x10,:x11]] | 111 | 893
$dfr2[10] | 22 | 22
@inbounds $dfr2[10] | 22 | 22
$dfr2[:x1] | 30 | 30
@inbounds $dfr2[:x1] | 30 | 30
$dfr2[[10,11]] | 70 | 835
@inbounds $dfr2[[10,11]] | 52 | 837
$dfr2[[:x10,:x11]] | 127 | 897
@inbounds $dfr2[[:x10,:x11]] | 109 | 896
collect(eachrow($df)) | 4540 | 20562
getindex.(eachrow($df), :x50) | 48835 | 74024

PS. @nalimilan If you know how to use PkgBenchmark.jl can you please set it up (if not I will get to it in the future, but I read through the docs and could not find instructions on a workflow how to best set it up in CI infrastructure).

src/dataframerow/dataframerow.jl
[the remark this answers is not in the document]
nalimilan: These are gone now, right?

src/subdataframe/subdataframe.jl
```diff
@@ -1,154 +1,142 @@
...
-SubDataFrame is used frequently in split/apply sorts of operations.
-
-```julia
-view(d::AbstractDataFrame, rows)
-```
+A `SubDataFrame` is meant to be constructed with `view` when a collection of
+rows and columns is selected.
+It is also created by some methods of [`by`](@ref) and [`combine`](@ref).
 
 ### Arguments
```
nalimilan: This isn't relevant now that `view` isn't shown here.

src/other/index.jl
```diff
@@ -208,3 +211,105 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
 
     return u
 end
+
+@inline parentcols(ind::Index) = Base.OneTo(length(ind))
+@inline parentcols(ind::Index, cols) = cols
+
+### SubIndex of Index. Used by SubDataFrame, DataFrameRow, and DataFrameRows
+
+# We allow I to be AbstractIndex, to allow for extensions
```
nalimilan: Remove this.

src/other/index.jl
```diff
@@ -208,3 +211,105 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
...
+end
+
+@inline SubIndex(parent::AbstractIndex, cols::ColumnIndex) =
+    throw(ArgumentError("cols argument must be a vector (got $cols)"))
+
+Base.@propagate_inbounds SubIndex(parent::AbstractIndex, cols) =
+    SubIndex(parent, parent[cols])
+
+# a helper function that lazily creates remap when needed
+# it assumes that remap is Vector{Int} unless it is an AbstractUnitRange
```
nalimilan: Is this assumption really made? We just need `remap` to be mutable and be able to hold `Int` values. So I don't think this comment is needed.

src/other/index.jl
```diff
@@ -208,3 +211,105 @@ function add_names(ind::Index, add_ind::Index; makeunique::Bool=false)
...
+
+### SubIndex of Index. Used by SubDataFrame, DataFrameRow, and DataFrameRows
+
+# We allow I to be AbstractIndex, to allow for extensions
+# In DataFrames.jl I is always Index
+
+struct SubIndex{I<:AbstractIndex,S<:AbstractVector{Int},T<:AbstractVector{Int}} <: AbstractIndex
+    parent::I
+    cols::S # columns from idx selected in SubIndex
+    remap::T # reverse of cols
```
```suggestion
    remap::T # reverse mapping from cols to their position in the SubIndex
```

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,117 @@
...
+    DataFrameRow{<:AbstractDataFrame,<:AbstractIndex}
 
-A view of one row of an AbstractDataFrame.
+A view of one row of an `AbstractDataFrame`.
+
+A `DataFrameRow` is constructed with `view` or `getindex` when one row and a
+selection of columns are requested, or when iterating the result
+of the call to the [`eachrow`](@ref) function.
+
+### Arguments
```
nalimilan: This doesn't apply here either.

src/abstractdataframe/iteration.jl
```diff
@@ -8,31 +8,83 @@
 
 # Iteration by rows
 """
-    DataFrameRows{T<:AbstractDataFrame} <: AbstractVector{DataFrameRow{T}}
+    DataFrameRows{D<:AbstractDataFrame,S<:AbstractIndex} <: AbstractVector{DataFrameRow{D,S}}
 
 Iterator over rows of an `AbstractDataFrame`,
 with each row represented as a `DataFrameRow`.
 
+Currently supports `DataFrame` and `SubDataFrame`.
```
nalimilan: Remove this too.

src/other/index.jl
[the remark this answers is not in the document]
nalimilan: OK, I wish Base also used `Union{Signed, Unsigned}`, but let's see how it goes.

src/other/index.jl
```diff
@@ -225,11 +226,29 @@ struct SubIndex{I<:AbstractIndex,S<:AbstractVector{Int},T<:AbstractVector{Int}}
...
 end
 
-parentcols(ind::SubIndex) = ind.cols
-
 SubIndex(parent::AbstractIndex, ::Colon) = parent
 
-@inline function SubIndex(parent::AbstractIndex, cols::AbstractUnitRange{Int})
+@inline parentcols(ind::SubIndex) = ind.cols
+
+Base.@propagate_inbounds parentcols(ind::SubIndex, idx::Union{Integer, AbstractVector{<:Integer}}) =
```
nalimilan: Would it makes sense to return a view for `idx::AbstractVector`?

nalimilan:
> PS. @nalimilan If you know how to use PkgBenchmark.jl can you please set it up (if not I will get to it in the future, but I read through the docs and could not find instructions on a workflow how to best set it up in CI infrastructure).

I don't remember the details, but you can have a look at CategoricalArrays' benchmark folder.

src/dataframerow/dataframerow.jl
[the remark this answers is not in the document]
bkamins: They are gone here but are supported by `SubIndex` which defines the following outer constructors:
```
SubIndex(parent::AbstractIndex, ::Colon)
SubIndex(parent::AbstractIndex, cols::AbstractUnitRange{Int})
SubIndex(parent::AbstractIndex, cols::AbstractVector{Int})
SubIndex(parent::AbstractIndex, cols::ColumnIndex)
SubIndex(parent::AbstractIndex, cols)
```

src/subdataframe/subdataframe.jl
bkamins: I have rewritten docstrings for `SubDataFrame` and `DataFrameRow`. Please give comments on them or just push a fix.

src/other/index.jl
bkamins: Removed. Also in `SubDataFrame`.

src/other/index.jl
bkamins: OK. Removed

src/dataframerow/dataframerow.jl
bkamins: fixed - but please review - I left the comment above unresolved to keep track of this.

src/abstractdataframe/iteration.jl

src/other/index.jl
bkamins: Good point. It should not break anything and will be somewhat faster.

bkamins:
As a note: changing the `parentindices` to view is faster. Not dramatically, but faster.

Timings for view approach:
```
julia> @btime $dfv[1:2,1:2];
  1.028 μs (31 allocations: 2.23 KiB)

julia> @btime view($dfv,1:2,1:2);
  80.593 ns (6 allocations: 304 bytes)
```

Timings when we materialize parent indices in a vector:
```
julia> @btime $dfv[1:2,1:2];
  1.079 μs (31 allocations: 2.28 KiB)

julia> @btime view($dfv,1:2,1:2);
  100.126 ns (6 allocations: 352 bytes)
```

However, I would say that the important thing here is that after this PR (if we decide to merge it) actually it makes no sense performance-wise not to use views unless someone really needs a copy. So we are reaching a state similar to what we have in Base, as creating a new `DataFrame` is really expensive.

src/dataframerow/dataframerow.jl
```diff
@@ -1,85 +1,123 @@
...
+Indexing is one-dimensional like specifying a column of a `DataFrame`.
+You can also access the data in a `DataFrameRow` using the `getproperty` and
+`setproperty!` functions and convert it to a `NamedTuple` using the `copy` function.
+
+It is possible to create a `DataFrameRow` with duplicate columns.
+All such columns will have a reference to the same entry in the parent `DataFrame`.
+
+If the selection of columns in a parent data frame is passed as `:` (a colon)
+then `DataFrameRow` will always have all columns from the parent,
+even if it is mutated.
```
```suggestion
even if they are added or removed after its creation.
```
(Same for `DataFrameRow`.)

nalimilan:
OK. Is there any drawback to using `view` in `parentindices`? E.g. what happens if the index is modified afterwards? Maybe it should be copied before storing it in a field (as opposed to using it just once for indexing)? It could also make sense to use a view for `AbstractArray{Symbol}`, but technically I don't think that's possible (as generators cannot be used for indexing).

bkamins:
There are several issues here (I refer to `SubDataFrame` but `DataFrameRow` is the same).

1. In what we have now parent of a `SubDataFrame` is always a de-aliased `DataFrame` (this might not hold with user-defined `AbstractDataFrame`, depending on how user implements the new type) - that is why in the PR I was writing that we support *DataFrame and SubDataFrame* only (i.e. not that we do not support other subtypes of `AbstractDataFrame` but that we guarantee). The de-aliasing approach is exactly the same as in `view` in Base (there also parent of a view of a view is the true parent and only indices are appropriately recalculated).
2. given point `SubDataFrame` does not care how much nested its creation is - it is only "one layer" over a root parent `DataFrame`
3. `SubDataFrame` references its parent by row and column **numbers**. This is the only information that `SubDataFrame` holds. So if the parent is mutaded then `SubDataFrame` will point to the row and column **numbers** established on its creation in the mutated parent (which means e.g. that if `SubDataFrame` points at column 2 and it gets renamed then it still points at the same column and will show a new name).

> But which index and which view exactly?

This is pretty tricky (also in Base it is tricky). Therefore in general you should not mutate the selectors you use to create views. It is best explained by an example I think.

The first "simple" scenario:
```
julia> df = DataFrame(a=1:5, b=11:15, c=21:25)
5×3 DataFrame
│ Row │ a     │ b     │ c     │
│ 1   │ 1     │ 11    │ 21    │
│ 2   │ 2     │ 12    │ 22    │
│ 3   │ 3     │ 13    │ 23    │
│ 4   │ 4     │ 14    │ 24    │
│ 5   │ 5     │ 15    │ 25    │

julia> selector = [3,2]
2-element Array{Int64,1}:
 3
 2

julia> dfv = view(df, :, selector)
5×2 SubDataFrame
│ Row │ c     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 21    │ 11    │
│ 2   │ 22    │ 12    │
│ 3   │ 23    │ 13    │
│ 4   │ 24    │ 14    │
│ 5   │ 25    │ 15    │

julia> push!(selector, 1)
3-element Array{Int64,1}:
 3
 2
 1

julia> dfv
5×3 SubDataFrame
│ Row │ c     │ b     │ a     │
│ 1   │ 21    │ 11    │ 1     │
│ 2   │ 22    │ 12    │ 2     │
│ 3   │ 23    │ 13    │ 3     │
│ 4   │ 24    │ 14    │ 4     │
│ 5   │ 25    │ 15    │ 5     │

julia> selector = [:c,:b]
2-element Array{Symbol,1}:
 :c
 :b

julia> dfv = view(df, :, selector)
5×2 SubDataFrame
│ Row │ c     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 21    │ 11    │
│ 2   │ 22    │ 12    │
│ 3   │ 23    │ 13    │
│ 4   │ 24    │ 14    │
│ 5   │ 25    │ 15    │

julia> push!(selector, :a)
3-element Array{Symbol,1}:
 :c
 :b
 :a

julia> dfv
5×2 SubDataFrame
│ Row │ c     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 21    │ 11    │
│ 2   │ 22    │ 12    │
│ 3   │ 23    │ 13    │
│ 4   │ 24    │ 14    │
│ 5   │ 25    │ 15    │
```
and you will have the same inconsistency in all operations that access columns of a `DataFrame` or `SubDataFrame` and produce a view as a result. That is: any operation that takes an `AbstractDataFrame` and produces.

The second "complex" scenario:
```
julia> df = DataFrame(a=1:5, b=11:15, c=21:25)
5×3 DataFrame
│ Row │ a     │ b     │ c     │
│ 1   │ 1     │ 11    │ 21    │
│ 2   │ 2     │ 12    │ 22    │
│ 3   │ 3     │ 13    │ 23    │
│ 4   │ 4     │ 14    │ 24    │
│ 5   │ 5     │ 15    │ 25    │

julia> selector1 = [3,2]
2-element Array{Int64,1}:
 3
 2

julia> dfv1 = view(df, selector1)
5×2 SubDataFrame
│ Row │ c     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 21    │ 11    │
│ 2   │ 22    │ 12    │
│ 3   │ 23    │ 13    │
│ 4   │ 24    │ 14    │
│ 5   │ 25    │ 15    │

julia> selector2 = [1]
1-element Array{Int64,1}:
 1

julia> dfv2 = view(dfv1, selector2)
5×1 SubDataFrame
│ Row │ c     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 21    │
│ 2   │ 22    │
│ 3   │ 23    │
│ 4   │ 24    │
│ 5   │ 25    │

julia> selector1[1] = 1
1

julia> dfv1
5×2 SubDataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 11    │
│ 2   │ 2     │ 12    │
│ 3   │ 3     │ 13    │
│ 4   │ 4     │ 14    │
│ 5   │ 5     │ 15    │

julia> dfv2
5×1 SubDataFrame
│ Row │ a     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │
│ 3   │ 3     │
│ 4   │ 4     │
│ 5   │ 5     │
```
And you see that `dfv2` reflects the change in selector of `dfv1`.

Now what would change if we would make a copy in `parentcols`

Scenario one "simple"
```
julia> df = DataFrame(a=1:5, b=11:15, c=21:25)
5×3 DataFrame
│ Row │ a     │ b     │ c     │
│ 1   │ 1     │ 11    │ 21    │
│ 2   │ 2     │ 12    │ 22    │
│ 3   │ 3     │ 13    │ 23    │
│ 4   │ 4     │ 14    │ 24    │
│ 5   │ 5     │ 15    │ 25    │

julia> selector = [3,2]
2-element Array{Int64,1}:
 3
 2

julia> dfv = view(df, :, selector)
5×2 SubDataFrame
│ Row │ c     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 21    │ 11    │
│ 2   │ 22    │ 12    │
│ 3   │ 23    │ 13    │
│ 4   │ 24    │ 14    │
│ 5   │ 25    │ 15    │

julia> push!(selector, 1)
3-element Array{Int64,1}:
 3
 2
 1

julia> dfv
5×3 SubDataFrame
│ Row │ c     │ b     │ a     │
│ 1   │ 21    │ 11    │ 1     │
│ 2   │ 22    │ 12    │ 2     │
│ 3   │ 23    │ 13    │ 3     │
│ 4   │ 24    │ 14    │ 4     │
│ 5   │ 25    │ 15    │ 5     │
```
(so no change here as we store the selector without copying it when it is an `AbstractVector{<Integer}`).

Scenario two "complex"
```
julia> df = DataFrame(a=1:5, b=11:15, c=21:25)
5×3 DataFrame
│ Row │ a     │ b     │ c     │
│ 1   │ 1     │ 11    │ 21    │
│ 2   │ 2     │ 12    │ 22    │
│ 3   │ 3     │ 13    │ 23    │
│ 4   │ 4     │ 14    │ 24    │
│ 5   │ 5     │ 15    │ 25    │

julia> selector1 = [3,2]
2-element Array{Int64,1}:
 3
 2

julia> dfv1 = view(df, selector1)
5×2 SubDataFrame
│ Row │ c     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 21    │ 11    │
│ 2   │ 22    │ 12    │
│ 3   │ 23    │ 13    │
│ 4   │ 24    │ 14    │
│ 5   │ 25    │ 15    │

julia> selector2 = [1]
1-element Array{Int64,1}:
 1

julia> dfv2 = view(dfv1, selector2)
5×1 SubDataFrame
│ Row │ c     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 21    │
│ 2   │ 22    │
│ 3   │ 23    │
│ 4   │ 24    │
│ 5   │ 25    │

julia> selector1[1] = 1
1

julia> dfv1
5×2 SubDataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 11    │
│ 2   │ 2     │ 12    │
│ 3   │ 3     │ 13    │
│ 4   │ 4     │ 14    │
│ 5   │ 5     │ 15    │

julia> dfv2
5×1 SubDataFrame
│ Row │ c     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 21    │
│ 2   │ 22    │
│ 3   │ 23    │
│ 4   │ 24    │
│ 5   │ 25    │
```
(and we have a change here as `dfv1` is changed as in simple scenario but `dfv2` is not changed because its selector was materialized and was not a view as above)

In summary - we will have the inconsistency between `AbstractVector{Int}` and `AbstractVector{Symbol}` anyway unless we always copy also `AbstractVector{Int}` or store `AbstractVector{Symbol}` and later pay the cost of remapping symbols to columns each time a column is selected. In both cases I think we should not change it.

Regarding view vs copy on subselection from a view. Given that both options have pros and cons I would go with what Base does in this case and Base makes a copy:
```
julia> x = [1:5;]
5-element Array{Int64,1}:
 1
 2
 3
 4
 5

julia> selector1 = [3,2]
2-element Array{Int64,1}:
 3
 2

julia> v1 = view(x, selector1)
2-element view(::Array{Int64,1}, [3, 2]) with eltype Int64:
 3
 2

julia> selector2 = [1]
1-element Array{Int64,1}:
 1

julia> v2 = view(v1, selector2)
1-element view(::Array{Int64,1}, [3]) with eltype Int64:
 3

julia> selector1[1] = 1
1

julia> v1
2-element view(::Array{Int64,1}, [1, 2]) with eltype Int64:
 1
 2

julia> v2
1-element view(::Array{Int64,1}, [3]) with eltype Int64:
 3
```

So in summary - I would tend to make a copy (as in the design proposed originally)

nalimilan:
Thanks, makes sense. Mutating the selector doesn't sound like a very common thing to do anyway.

bkamins:
@nalimilan merged; thank you!

PR #1651, opened 2018-12-23
Add optimized methods to compute grouped reductions

nalimilan:
```julia
               b = repeat(20000:-1:1, outer=[40]),
               c = randn(800000))
gd = groupby(df, :a)
using BenchmarkTools, Statistics

# Master
julia> @btime combine(gd, :c => sum);
  10.473 ms (439181 allocations: 17.39 MiB)

julia> @btime combine(gd, :c => sum, :b => mean);
  19.515 ms (719227 allocations: 26.24 MiB)

# This PR
julia> @btime combine(gd, :c => sum);
  1.578 ms (154 allocations: 946.95 KiB)

julia> @btime combine(gd, :c => sum, :b => mean);
  3.046 ms (186 allocations: 1.54 MiB)
```

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -90,7 +90,7 @@ end
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
                          hash::Val = Val(true),
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool, Bool}
+                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool}
     @assert groups === nothing || length(groups) == length(cols[1])
     rhashes, missings = hashrows(cols, skipmissing)
     # inspired by Dict code from base cf. https://github.com/JuliaData/DataTables.jl/pull/17#discussion_r102481481
@@ -130,13 +130,13 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
             groups[i] = gix
         end
     end
-    return ngroups, rhashes, gslots, false, false
+    return ngroups, rhashes, gslots, false
 end
 
 function row_group_slots(cols::Tuple{CategoricalVector},
                          hash::Val{false},
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool, Bool}
+                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool}
     col = cols[1]
     @assert groups === nothing || length(groups) == length(col)
 
@@ -149,16 +149,36 @@ function row_group_slots(cols::Tuple{CategoricalVector},
         # we could simply copy refs to groups, but the performance gain is negligible,
         # so always sort groups in the order of levels
         refmap = [0; CategoricalArrays.order(col.pool)]
+        seen = fill(false, length(refmap))
         if skipmissing
             refmap .+= 1
+            seen[1] = true
         else
             refmap[1] = ngroups
         end
         @inbounds for i in eachindex(groups)
-            groups[i] = refmap[col.refs[i]+1]
+            j = refmap[col.refs[i]+1]
+            groups[i] = j
+            seen[j] |= true
+        end
+        if !all(seen)
+            if skipmissing # Always keep first group even if empty
+                ngroups = 1
+                start = 2
+            else
+                ngroups = 0
+                start = 1
+            end
+            @inbounds for i in start:length(refmap)
+                ngroups += seen[i]
+                refmap[i] = ngroups
+            end
+            @inbounds for i in eachindex(groups)
+                groups[i] = refmap[groups[i]]
+            end
         end
     end
-    return ngroups, UInt[], Int[], true, true
+    return ngroups, UInt[], Int[], true
 end
 
 # Builds RowGroupDict for a given DataFrame.
@@ -171,7 +191,7 @@ end
 function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false,
                     skipmissing::Bool = false)
     groups = Vector{Int}(undef, nrow(df))
-    ngroups, rhashes, gslots, sorted, empty =
+    ngroups, rhashes, gslots, sorted =
         row_group_slots(ntuple(i -> df[i], ncol(df)), Val(hash), groups, skipmissing)
 
     # count elements in each group
@@ -203,25 +223,6 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
         popfirst!(starts)
         popfirst!(stops)
         ngroups -= 1
-    end
-
-    # if row_group_slots can return empty groups, remove them
-    if empty
-        i = 0
-        j = 0
-        @inbounds while j < ngroups
-            j += 1
-            while stops[j] < starts[j]
-                j < ngroups || @goto done
-                j += 1
-            end
-            i += 1
-            starts[i] = starts[j]
-            stops[i] = stops[j]
-        end
-        @label done
-        resize!(starts, i)
-        resize!(stops, i)
     end
 
     # sort groups if row_group_slots hasn't already done that
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -85,7 +85,6 @@ end
 # 3) slot array for a hash map, non-zero values are
 #    the indices of the first row in a group
 # 4) whether groups are already sorted
-# 5) whether some groups might be empty (index isn't used in `groups`)
 # Optional `groups` vector is set to the group indices of each row
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
                          hash::Val = Val(true),
@@ -159,7 +158,7 @@ function row_group_slots(cols::Tuple{CategoricalVector},
         @inbounds for i in eachindex(groups)
             j = refmap[col.refs[i]+1]
             groups[i] = j
-            seen[j] |= true
+            seen[j] = true
         end
         if !all(seen)
             if skipmissing # Always keep first group even if empty
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -232,7 +232,7 @@ function Base.map(f::Any, gd::GroupedDataFrame)
         ends[end] = length(idx)
         return GroupedDataFrame(parent, gd.cols, idx, collect(1:length(idx)), starts, ends)
     else
-        return GroupedDataFrame(parent, gd.cols, idx, Int[], Int[], Int[])
+        return GroupedDataFrame(gd.parent[1:0, gd.cols], gd.cols, Int[], Int[], Int[], Int[])
     end
 end
 
@@ -346,7 +346,7 @@ function combine(f::Any, gd::GroupedDataFrame)
         idx, valscat = _combine(f, gd)
         return hcat!(gd.parent[idx, gd.cols], valscat, makeunique=true)
     else
-        return similar(gd.parent[gd.cols], 0)
+        return gd.parent[1:0, gd.cols]
     end
 end
 combine(gd::GroupedDataFrame, f::Any) = combine(f, gd)
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -478,17 +478,19 @@ end
 
 function groupreduce(op, adjust, incol::AbstractVector{T}, gd::GroupedDataFrame) where T
     n = length(gd)
-    outcol = groupreduce_init(op, incol, gd)
+    res = groupreduce_init(op, incol, gd)
     if adjust !== nothing
         counts = zeros(Int, n)
     end
     @inbounds @simd for i in eachindex(incol, gd.groups)
         gix = gd.groups[i]
         adjust !== nothing && (counts[gix] += 1)
-        outcol[gix] = op(outcol[gix], incol[i])
+        res[gix] = op(res[gix], incol[i])
     end
     if adjust !== nothing
-        map!(adjust, outcol, outcol, counts)
+        outcol = map(adjust, res, counts)
+    else
+        outcol = res
     end
     # Undo pool sharing done by groupreduce_init
     if outcol isa CategoricalVector
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -160,8 +160,9 @@ all groups, and (if a named tuple or data frame) with the same fields or columns
 Due to type instability, returning a single value or a named tuple is dramatically
 faster than returning a data frame.
 
-Optimized methods are used when standard reduction functions (`sum`, `prod`,
-`minimum`, `maximum` and `mean`) are specified using the pair syntax (e.g. `col => sum`).
+Optimized methods are used when standard summary functions (`sum`, `prod`,
+`minimum`, `maximum`, `mean`, `var`, `std`, `first`, `last` and `length)
+are specified using the pair syntax (e.g. `col => sum`).
 When computing the `sum` or `mean` over floating point columns, results will be less
 accurate than the standard [`sum`](@ref) function (which uses pairwise summation). Use
 `col => x -> sum(x)` to avoid the optimized method and use the slower, more accurate one.
@@ -285,8 +286,9 @@ all groups, and (if a named tuple or data frame) with the same fields or columns
 Due to type instability, returning a single value or a named tuple is dramatically
 faster than returning a data frame.
 
-Optimized methods are used when standard reduction functions (`sum`, `prod`,
-`minimum`, `maximum` and `mean`) are specified using the pair syntax (e.g. `col => sum`).
+Optimized methods are used when standard summary functions (`sum`, `prod`,
+`minimum`, `maximum`, `mean`, `var`, `std`, `first`, `last` and `length)
+are specified using the pair syntax (e.g. `col => sum`).
 When computing the `sum` or `mean` over floating point columns, results will be less
 accurate than the standard [`sum`](@ref) function (which uses pairwise summation). Use
 `col => x -> sum(x)` to avoid the optimized method and use the slower, more accurate one.
@@ -424,7 +426,6 @@ for f in (:sum, :prod, :maximum, :minimum, :mean, :var, :std, :first, :last)
         funname(::typeof(check_aggregate($f∘skipmissing))) = :function
     end
 end
-funname(::typeof(length)) = :length
 
 # Find first value matching condition for each group
 # Optimized for situations where a matching value is typically encountered
@@ -445,7 +446,7 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
         end
     end
     if nfilled < length(outcol)
-        throw(ArgumentError("cannot compute maximum or minimum for groups with only missing values"))
+        throw(ArgumentError("some groups contain only missing values"))
     end
     outcol
 end
@@ -920,8 +921,9 @@ all groups, and (if a named tuple or data frame) with the same fields or columns
 Due to type instability, returning a single value or a named tuple is dramatically
 faster than returning a data frame.
 
-Optimized methods are used when standard reduction functions (`sum`, `prod`,
-`minimum`, `maximum` and `mean`) are specified using the pair syntax (e.g. `col => sum`).
+Optimized methods are used when standard summary functions (`sum`, `prod`,
+`minimum`, `maximum`, `mean`, `var`, `std`, `first`, `last` and `length)
+are specified using the pair syntax (e.g. `col => sum`).
 When computing the `sum` or `mean` over floating point columns, results will be less
 accurate than the standard [`sum`](@ref) function (which uses pairwise summation). Use
 `col => x -> sum(x)` to avoid the optimized method and use the slower, more accurate one.
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -9,6 +9,7 @@ Not meant to be constructed directly, see `groupby`.
 struct GroupedDataFrame{T<:AbstractDataFrame}
     parent::T
     cols::Vector{Int}    # columns used for grouping
+    groups::Vector{Int}  # group indices for each row
     idx::Vector{Int}     # indexing vector when grouped by the given columns
     starts::Vector{Int}  # starts of groups
     ends::Vector{Int}    # ends of groups
@@ -74,7 +75,7 @@ function groupby(df::AbstractDataFrame, cols::AbstractVector;
     intcols = index(df)[cols]
     sdf = df[intcols]
     df_groups = group_rows(sdf, false, sort, skipmissing)
-    GroupedDataFrame(df, intcols, df_groups.rperm,
+    GroupedDataFrame(df, intcols, df_groups.groups, df_groups.rperm,
                      df_groups.starts, df_groups.stops)
 end
 groupby(d::AbstractDataFrame, cols;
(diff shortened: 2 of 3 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -387,6 +423,98 @@ function gen_fun(::Type{NT}, f2) where NT
...
+# Use reducedim_init to get a vector of the right type,
+# but trick it into using the expected length
+groupreduce_init(op, incol, gd) =
+    Base.reducedim_init(identity, op, view(incol, 1:length(gd)), 2)
+
+for (op, initf) in ((:max, :typemin), (:min, :typemax))
+    @eval begin
+        function groupreduce_init(::typeof($op), incol::AbstractVector{T}, gd) where T
+            outcol = similar(incol, length(gd))
+            # Comparison is possible only between CatValues from the same pool
```
nalimilan: It's unfortunate that the standard approach doesn't work for categorical arrays. I've also realized something like `maximum(categorical([1 2; 3 4]), dims=1)` currently doesn't work for the same reason. Not sure what to do about this in general.

src/groupeddataframe/grouping.jl
```diff
@@ -387,6 +423,98 @@ function gen_fun(::Type{NT}, f2) where NT
...
+    # Undo pool sharing done by groupreduce_init
+    if outcol isa CategoricalVector
+        U = Union{CategoricalArrays.leveltype(outcol),
+                  eltype(outcol) >: Missing ? Missing : Union{}}
+        outcol = CategoricalArray{U, 1}(outcol.refs, incol.pool)
+    end
+    if isconcretetype(T)
+        return outcol
+    else
+        copyto_widen!(Tables.allocatecolumn(typeof(first(outcol)), n), outcol)
```
nalimilan: This is also not optimal, but it's required if we want to return the same element type as the non-optimized method. Hopefully it should be quite rare, since reducing over `Union{T,Missing}` columns will return `missing` if there is any missing value, which isn't very useful (though one can have a  `Union{T,Missing}` column without any missing value).

bkamins:
A quick comment (I just glanced at the code). We could add methods for counting (this is pretty common), variance, standadrd deviation, covariance. Of course this could be added later, but maybe these cases will give raise to a more general API design (I am not saying it is not general enough - but it would be good to know it would support such cases).

src/dataframerow/utils.jl
```diff
@@ -90,7 +90,7 @@ end
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
                          hash::Val = Val(true),
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool, Bool}
+                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool}
```
bkamins: comment above should be updated.

src/dataframerow/utils.jl
```diff
@@ -149,16 +149,36 @@ function row_group_slots(cols::Tuple{CategoricalVector},
...
             refmap .+= 1
+            seen[1] = true
         else
             refmap[1] = ngroups
         end
         @inbounds for i in eachindex(groups)
-            groups[i] = refmap[col.refs[i]+1]
+            j = refmap[col.refs[i]+1]
+            groups[i] = j
+            seen[j] |= true
```
bkamins: why `|=` and not simply `=`? Is it faster?

src/dataframerow/utils.jl
```diff
@@ -149,16 +149,36 @@ function row_group_slots(cols::Tuple{CategoricalVector},
...
+                ngroups += seen[i]
+                refmap[i] = ngroups
+            end
+            @inbounds for i in eachindex(groups)
+                groups[i] = refmap[groups[i]]
+            end
         end
     end
-    return ngroups, UInt[], Int[], true, true
+    return ngroups, UInt[], Int[], true
```
bkamins: maybe after squashing non-present groups `groups` vector should be resized just to make sure its length is appropriate (this should not be very expensive I think).

src/dataframerow/utils.jl
```diff
@@ -130,13 +130,13 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
     end
-    return ngroups, rhashes, gslots, false, false
+    return ngroups, rhashes, gslots, false
 end
 
 function row_group_slots(cols::Tuple{CategoricalVector},
                          hash::Val{false},
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool, Bool}
+                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool}
```
bkamins: maybe rename this function to `row_group_slots!` to show that it mutates its argument?

src/dataframerow/utils.jl
bkamins: Actually it seems we must do it as below you do:
```
stops = zeros(Int, ngroups)
@inbounds for g_ix in groups
    stops[g_ix] += 1
end
```

Can you disable this `@inbounds` in tests and do we have test with categorical vector that has an empty level?

bkamins: Also in lower code it is assumed that `length(groups) == ngroups`
(this means that maybe we should run tests of DataFrames.jl with globaly disabled `@inbounds` with `--check-bounds=yes` option)

src/groupeddataframe/grouping.jl
```diff
@@ -223,9 +230,9 @@ function Base.map(f::Any, gd::GroupedDataFrame)
         resize!(starts, j-1)
         resize!(ends, j-1)
         ends[end] = length(idx)
-        return GroupedDataFrame(parent, gd.cols, collect(1:length(idx)), starts, ends)
+        return GroupedDataFrame(parent, gd.cols, idx, collect(1:length(idx)), starts, ends)
     else
-        return GroupedDataFrame(parent, gd.cols, Int[], Int[], Int[])
+        return GroupedDataFrame(parent, gd.cols, idx, Int[], Int[], Int[])
```
bkamins: `idx` is undefined in this branch. Can you please also add a test that catches it?

src/groupeddataframe/grouping.jl
```diff
@@ -366,9 +378,33 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
...
-function gen_fun(::Type{NT}, f2) where NT
-    get_val(::Val{T}) where {T} = T
+struct Reduce{F, A}
+    f::F
+    adjust::A
+end
+Reduce(f) = Reduce(f, nothing)
+function (r::Reduce)(x)
+    v = reduce(r.f, x)
+    r.adjust === nothing ? v : r.adjust(v, length(x))
```
bkamins: I would rather define it `r.adjust(v, x)` to allow any adjustment function. It is a bit more flexible then. Next, if we wanted even more generality (given my earlier comment what functions could be supported) - an even broader `Reduce` definition could be considered.

I would document and export `Reduce` so that users can write their custom reducers (e.g. geometric or harmonic mean).

src/groupeddataframe/grouping.jl
```diff
@@ -366,9 +378,33 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
...
+end
+
+check_reduction(f::Any) = f
+check_reduction(::typeof(sum)) = Reduce(Base.add_sum)
+check_reduction(::typeof(prod)) = Reduce(Base.mul_prod)
+check_reduction(::typeof(maximum)) = Reduce(max)
+check_reduction(::typeof(minimum)) = Reduce(min)
+check_reduction(::typeof(mean)) = Reduce(Base.add_sum, /)
```
bkamins: If you wanted to follow my recommendation above here we would write:
```
check_reduction(::typeof(mean)) = Reduce(Base.add_sum, (s,x) -> s / length(x))
```

src/groupeddataframe/grouping.jl
```diff
@@ -366,9 +378,33 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
...
 
+for f in (:sum, :prod, :maximum, :minimum, :mean)
+    @eval begin
+        funname(::typeof(check_reduction($f))) = Symbol($f)
```
bkamins: maybe a more specific method name than `funname` could be used?

bkamins:
I have left first comments (+ a comment on https://github.com/JuliaData/CategoricalArrays.jl/issues/178)

matthieugomez:
I have to say: it is amazing how one can implement all of these fast aggregate functions in one commit. data.table / dplyr are now a few years old but I don't think they optimize `std` / `var` yet.

src/groupeddataframe/grouping.jl
nalimilan: That's the name of the method we use elsewhere already. What's wrong with it?

src/dataframerow/utils.jl
nalimilan: I figured it could avoid a write when not needed, but apparently it's slightly slower:
```julia
function f(refmap, groups, refs)
    seen = fill(false, length(refmap))
    @inbounds for i in eachindex(groups)
        j = refmap[refs[i]+1]
        groups[i] = j
        seen[j] |= true
    end
    seen, groups
end

N = 100_000
ngroups = 10
groups = Vector{Int}(undef, N)
refs = rand(1:ngroups, N)
refmap = rand(1:ngroups+1, ngroups+1)

# With |=
julia> @btime f(refmap, groups, refs);
  163.944 μs (2 allocations: 128 bytes)

# With =
julia> @btime f(refmap, groups, refs);
  149.227 μs (2 allocations: 128 bytes)
```

src/dataframerow/utils.jl
nalimilan: The `groups` vector has one entry for each row, not one for each group. So AFAICT there's no problem here (and we already pass `--check-bounds=yes`).

src/groupeddataframe/grouping.jl
nalimilan: The problem is, the whole point of these optimizations is to avoid constructing the `x` vector. And even if we constructed it (using a view, it wouldn't be too costly, though probably significant), its length cannot be computed efficiently for `skipmissing` cases, while it's almost free to compute it on the fly. So for now I'd rather keep this internal, and discuss with e.g. SplitApplyCombine developers to see how this could be generalized. Special-casing `length` doesn't sound too bad since that's an essential information and it's very cheap to compute anyway.

src/groupeddataframe/grouping.jl
nalimilan: See reply above.

nalimilan:
I have added commits which improve the logic and implement `var`, `std`, `first`, `last` and `length`. We don't support `count` at the moment as that function does something different (which is a bit annoying, as you need to pick an arbitrary column to get its length). When multiple pairs are passed, we now compute each function in turn, instead of computing all functions at the same time for a given group. This allows using an optimized method even if a non-optimized one is specified in another pair, makes the code simpler, and should be easier to parallelize.

> I have to say: it is amazing how one can implement all of these fast aggregate functions in one commit. data.table / dplyr are now a few years old and I don't think they optimize `std` / `var` yet.

Good point. AFAICT data.table uses a slow approach, allocating vectors for each group to compute the variance on it: https://github.com/Rdatatable/data.table/blob/b186b715837a069e06d1a49026ed501ae8a1e524/src/gsumm.c#L1151
I've mentioned this for addition to the H2O.ai benchmarks: https://github.com/h2oai/db-benchmark/issues/60

src/groupeddataframe/grouping.jl
bkamins: Why not use `nameof` directly here?

src/dataframerow/utils.jl

src/groupeddataframe/grouping.jl

src/dataframerow/utils.jl
bkamins: Thanks. Now I understand.

bkamins:
> We don't support count at the moment as that function does something different

I see the problem, but do you see a workaround in the future?

src/groupeddataframe/grouping.jl
nalimilan: Because we want the name of the original function, which isn't stored anywhere in the struct. Also what `funname` currently does is return `function` for anonymous functions (which you recently improved unless I'm mistaken).

nalimilan:
I'm not sure, I haven't thought a lot about it. We could allow `() => length` or similar to avoid the need to pick an arbitrary column.

src/groupeddataframe/grouping.jl
```diff
@@ -366,62 +379,260 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
...
+check_aggregate(::typeof(length)) = Aggregate(length)
+# SkipMissing does not support length
+
+for f in (:sum, :prod, :maximum, :minimum, :mean, :var, :std, :first, :last)
+    @eval begin
+        funname(::typeof(check_aggregate($f))) = Symbol($f)
+        funname(::typeof(check_aggregate($f∘skipmissing))) = :function
+    end
+end
+funname(::typeof(length)) = :length
```
bkamins: is this needed? already `funname(length)` should return `:length`. right?

src/groupeddataframe/grouping.jl
```diff
@@ -366,62 +379,260 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
...
+                nfilled += 1
+                break
             end
-            if res isa Union{AbstractDataFrame, NamedTuple, DataFrameRow, AbstractMatrix}
-                throw(ArgumentError("a single value or vector result is required when passing " *
-                                    "a vector or tuple of functions (got $(typeof(res)))"))
+        end
+    end
+    if nfilled < length(outcol)
+        throw(ArgumentError("cannot compute maximum or minimum for groups with only missing values"))
```
bkamins: this can be also called by `first` and `last`. right?

bkamins:
Regarding `count` discussion:
* `count`, `any` and `all` functions could be supported for boolean vectors by `groupreduce` even given its current implementation. Right?
* then if we wanted to handle predicates with `count`, `any` and `all`, which would be useful I think then we could use signatures matching `Base.Fix1{typeof(count)}` etc. to catch and handle these cases (the current would be `by(df, :group_col, :col=>Base.Fix1(count, predicate)`, but maybe this could be shortened if `Fix1` stared being used in Base - or at least exported); this would be cleanest; but maybe it would be simpler for a user to pass a tuple `by(df, :group_col, :col=>(count, predicate)` as it should be also easy enough to match?

Of course we do not have to do this in this PR.

nalimilan:
OK, I thought by `count` you mean counting rows in a group (i.e. `length`, but more intuitive). `count(pred, x)`, `any` and `all` shouldn't be too hard to support using the framework from this PR. The main issue is indeed the public API. I think we should investigate whether DataFramesMeta could automatically detect these functions (and things like `sum(skipmissing(:col))`) and transform them into something that can be optimized. Ideally that would be a generic API which can also be used directly, maybe living in SplitApplyCombine.

@andyferris @davidanthoff @piever Do you think it would be interesting/possible to move the optimized operations from this PR to a generic package like SplitApplyCombine, so that they can be shared across DataFrames, JuliaDB and Query?

test/grouping.jl
```diff
@@ -606,6 +631,94 @@ module TestGrouping
         end
     end
 
+    struct TestType end
+    Base.isless(::TestType, ::Int) = true
+    Base.isless(::Int, ::TestType) = false
+    Base.isless(::TestType, ::TestType) = false
+
+    @testset "combine with aggregation functions" begin
+        Random.seed!(1)
```
bkamins: We could run scenarios:
* `df = DataFrame(a = rand(1:5, 20), x1 = rand(Int, 20), x2 = rand(Complex{Int}, 20))`
* `df = DataFrame(a = categorical(rand(1:5, 20)), x1 = rand(Int, 20), x2 = rand(Complex{Int}, 20))`
* `df = DataFrame(a = categorical(rand(1:5, 20)), x1 = rand(Int, 20), x2 = rand(Complex{Int}, 20)); levels!(df.a, shuffle([levels(df.a); 10; 11]))`
* `df = DataFrame(a = categorical(rand(1:5, 20), ordered=true), x1 = rand(Int, 20), x2 = rand(Complex{Int}, 20))`
* `df = DataFrame(a = categorical(rand(1:5, 20), ordered=true), x1 = rand(Int, 20), x2 = rand(Complex{Int}, 20)); levels!(df.a, shuffle([levels(df.a); 10; 11]))`
e.g. 10 times to automatically improve the coverage.

nalimilan: Unfortunately that's not so simple. Most operations don't work for categorical values. I could add some cases in the categorical-specific tests, but what do you want to test exactly? For example, unused levels shouldn't make any difference here, and AFAIK they are tested elsewhere already (where it's more relevant). But please point uncovered scenarios if they are missing.

bkamins: My point is that it is a complex design so randomizing tests might catch something we have not thought of. The question is do we fully cover with tests:
* grouping on one column that is categorical ordered and unordered with missing level and non-missing level (these are 4 scenarios);
* the same but when grouping on multiple columns all categorical (possibly with mixed characteristics);
* the same but when grouping on multiple columns some categorical and some not-categorical;

If we have all these scenarios covered I am OK to leave it as is (or "reasonably covered" - i.e. we do not have to test things that you are 100% sure that work the same as the things that are tested).

But in general - that is why I have approved earlier because I think that it is only an extra precaution.

nalimilan: OK. What I meant is that to avoid combinatorial explosion of tests, I've tried to test each feature only once. All variations regarding the *grouping* column are handled in tests for `groupby`, since they make no difference for the `combine` code -- and AFAICT empty and reordered levels are tested there already. I guess we could add more tests with variations on the *aggregated* column here, but it's hard to decide what's relevant and what's not...

bkamins: OK - let us leave it as is. I think that aggregated column is less risky.

src/groupeddataframe/grouping.jl
```diff
@@ -366,62 +381,259 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
...
+end
+Reduce(f, condf=nothing) = Reduce(f, condf, nothing)
+
+check_aggregate(f::Any) = f
+check_aggregate(::typeof(sum)) = Reduce(Base.add_sum)
+check_aggregate(::typeof(prod)) = Reduce(Base.mul_prod)
+check_aggregate(::typeof(maximum)) = Reduce(max)
+check_aggregate(::typeof(minimum)) = Reduce(min)
+check_aggregate(::typeof(mean)) = Reduce(Base.add_sum, nothing, /)
+check_aggregate(::typeof(sum∘skipmissing)) = Reduce(Base.add_sum, !ismissing)
```
andyferris: Cool... I hadn't thought to introspect the type of the composition in this way - but this is a really clear way of expressing this!

andyferris:
> Do you think it would be interesting/possible to move the optimized operations from this PR to a generic package like SplitApplyCombine, so that they can be shared across DataFrames, JuliaDB and Query?

Sorry Milan, I'm not terribly familiar with some of the implementation of operations on DataFrames, so I'm having a bit of trouble following along. A grouped data frame is internally sorted by grouping key, rather than nested via `Dict`s?

I guess I'm wondering precisely what you are suggesting - there is already a generic `SplitApplyCombine.groupreduce` for arbitrary iterables which is optimized not to allocate or rearrange large amounts of memory (and I would love to have `groupvar` and `groupstd` and so-on as well, and combinations of everything working well with `skipmissing`). However, in the case of `SplitApplyCombine` it uses nesting more than the traditional SQL approach so the output is a `Dict` rather than a table with two columns.

> OK, I thought by count you mean counting rows in a group

I was also torn by `groupcount` vs `grouplength`.

nalimilan:
> Sorry Milan, I'm not terribly familiar with some of the implementation of operations on DataFrames, so I'm having a bit of trouble following along. A grouped data frame is internally sorted by grouping key, rather than nested via `Dict`s?

Yes, it stores a reference to a parent `DataFrame`, and information about which row goes to which group (in different forms).

> I guess I'm wondering precisely what you are suggesting - there is already a generic `SplitApplyCombine.groupreduce` for arbitrary iterables which is optimized not to allocate or rearrange large amounts of memory (and I would love to have `groupvar` and `groupstd` and so-on as well, and combinations of everything working well with `skipmissing`). However, in the case of `SplitApplyCombine` it uses nesting more than the traditional SQL approach so the output is a `Dict` rather than a table with two columns.

OK, I thought there was a function to group values according to a vector of group indices in SplitApplyCombine, but apparently that's not the case. It turns out it's often faster to compute group memberships in a first step (since you can compute hashes using SIMD and efficient memory access patterns; and with `CategoricalArray` columns group indices are already available without hashing nor dicts), and then do the actual reduction in a second step (since you can again use a very efficient loop). At least that's the approach DataFrames uses.

If you're also interested in a method taking group indices (in addition to a function computing the group, as currently), some code could be moved from DataFrames to SplitApplyCombine. It's a small part of the code, but it can be tricky, in particular when determining the initial values (especially for `minimum` and `maximum` and when skipping missing values) and the output type by progressive widening. But the fact that SplitApplyCombine returns a dict and DataFrames a vector is problematic...

andyferris:
> It turns out it's often faster to compute group memberships in a first step

Right. In *SplitApplyCombine* there is `groupinds` (and `groupview`) function but it does construct a `Dict`. Also it all works with *AcceleratedArrays* where secondary indices may store precomputed groups, that can be taken advantage of for a variety of operations including `groupreduce`. (The generic `groupreduce` operation I wrote has no O(input size) allocations so supports out-of-core streaming, which is the logic behind that as the default choice, though I'm happy to revisit any implemetation details).

I imagine there are a few more operations for different permutations of splitting data. For example, just speculating, if you already have the spliting indices then we could have some `split` and `splitreduce` functions...

> But the fact that SplitApplyCombine returns a dict and DataFrames a vector is problematic...

I just see this as: we need multiple generic functions for different types of nested/flattened inputs and outputs. Some will make sense for generic iteratables, others (like the SQL flavoured group operations) depend on concepts like columns (which so far `SplitApplyCombine` isn't aware of).

davidanthoff:
Just on the Query integration question: I've been designing/thinking about for a while now an alternative backend that process things in a column fashion, rather than row wise. That had been the plan from the get go, but it is just slow going, and as a first step the row wise processing implementation right now is more generic (but less performant). The general idea is that this column based backend kicks in if a query is purely table based, and if the query is one that we can run in an efficient way in a columnar way. The design I'm playing with right now would also implement a bunch of whole query rewrites/optimizations like in a database. In any case, once we start to implement all of this, I would expect that we would want to use algorithms like the one in this PR here, so the work here is most likely going to be super helpful down the road for this Query project!

But, be warned, don't expect anything anytime soon. I'm both still tinkering with the design and assembling the team to work on this here at Berkeley.

nalimilan:
OK, we don't have to do something right now, but if at some point you need code similar to this, we can try to share code across packages.

nalimilan:
I've added tests for floating point corner cases. Good to go?

bkamins:
Go ahead and merge it when you feel ready.

src/groupeddataframe/grouping.jl
```diff
@@ -366,62 +381,259 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
...
+function groupreduce!(res, f, op, condf, adjust,
+                      incol::AbstractVector{T}, gd::GroupedDataFrame) where T
+    n = length(gd)
+    if adjust !== nothing
+        counts = zeros(Int, n)
+    end
+    @inbounds for i in eachindex(incol, gd.groups)
+        gix = gd.groups[i]
+        x = incol[i]
+        if condf === nothing || condf(x)
```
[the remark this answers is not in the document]
bkamins: I guess in the future LLVM should be able to generate them automatically, so I would leave this only as a reference.

bkamins:
@nalimilan - I think it is good to go ahead and merge it (but please decide).

PR #1656, opened 2018-12-26
Make any AbstractDataFrame an iterable table

### LINKED ISSUE #1597: SubDataFrame is not an iterable table
A SubDataFrame is not an iterabletable, making it incompatible with that framework.

cf https://github.com/queryverse/IterableTables.jl/issues/76
cf https://github.com/JuliaPlots/StatPlots.jl/issues/197

nalimilan:
Fixes #1597.

@davidanthoff Could you suggest a test to ensure this doesn't regress?

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -44,6 +44,6 @@ Base.append!(df::DataFrame, x) = append!(df, DataFrame(x))
 # This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
 DataFrame(x::Vector{T}) where {T <: NamedTuple} = fromcolumns(Tables.columns(Tables.DataValueUnwrapper(x)))
 
-IteratorInterfaceExtensions.getiterator(df::DataFrame) = Tables.datavaluerows(df)
-IteratorInterfaceExtensions.isiterable(x::DataFrame) = true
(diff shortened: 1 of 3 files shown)


davidanthoff:
Two tests are simple:
````julia
df = DataFrame()
@test IteratorInterfaceExtensions.isiterable(df) == true
@test TableTraits.isiterabletable(df) == true
````
I'm not familiar enough with the subtypes of ``AbstractDataFrame`` to know which exists, but essentially I think we should just add the same two tests for those as well?

Tests for ``getiterator`` are more tricky. ``Tables.datavaluerows`` only exists if [DataValues.jl](https://github.com/queryverse/DataValues.jl) is loaded, I believe it is defined behind a ``@require`` block. So I think you would have to add ``DataValues`` to the test REQUIRE file, and then make sure it is loaded, and then something like this might do the trick:
````julia
df = DataFrame(a=[1,2,3], b=[1.,missing,3.])
@test collect(DataFrames.getiterator(df)) == [(a=1,b=DataValue(1.)), (a=2,b=DataValue{Float64}()), (a=3,b=DataValue(3.))]
````
For other subtypes of ``AbstractDataFrame`` I'm not sure again...

nalimilan:
Thanks! Looks like that did the trick, I've added a commit.

PR #1659, opened 2018-12-26
Flexible vcat

pdeffebach:
This PR adds some functionality for a "flexible" vcat that has more feature parity with dplyr's `bind_rows` and Stata's `append`. 

It introduces three keyword arguments: 

1) `widen` specifies whether or not the new dataframe contains the *union* of all the columns in its argument list. Any gaps are filled with missing. 

2) `keep` specifies a vector of column names that the resulting dataframe will have. 

3) `fillvalue` indicates what value we will use to fill in all missing data. 

`widen` can still be necessary with `keep`: if `keep` includes a symbol that is not present in all dataframes in the argument list, it will throw an error without `widen`. 

This is very much a WIP. We need to do the following (also commented in the code)

- [ ] Preserve order of the first dataframe in the argument list, if possible
- [ ] Get all the errors correct
- [ ] Make the `fill` part very efficient. Currently it allocates new vectors of `missings` and `copyto!`s them later to fill in the columns.


nalimilan:
Thanks. Do we really need `keep` though? It's easy enough to get rid of some columns before calling `vcat` (which isn't possible in Stata). Also, do you have a use case in mind for `fillvalue`? Maybe we can hardcode `missing` for now and see whether people actually need more flexibility.

pdeffebach:
I think `keep` is important. Imagine the following scenario

```
df1 = DataFrame(rand(10, 10)) # manageable width
df2 = DataFrame(rand(10, 100)) # too wide to want to work with
df3 = DataFrame(x1 = rand(10), x45 = rand(10))
```

Now you want to append but you only care about the columns `:x1`, `:x2`, and `:x45`. With just `widen` you would get all 100 columns in `df2`. And we can't subset `df1` because `:x45` isn't a column there. Similarly we have to be careful with subsetting `df3`. 

In a practical example, in Stata, I often have 2 dataset from different endline surveys, and I only care about questions that are common to both surveys, but there are like 200 questions so I don't want to bother with which ones are and are not in the Endline 1 survey. Even if I could subset the second survey, I wouldn't want to have to figure out which questions are in both and which are not. 

I'm fine to remove `fillvalue` and make `missing` the mandatory.

nalimilan:
Then maybe what's needed is actually an argument which can be `:inner`, `:outer`, `:left` or `:right`, as in `join`?

pdeffebach:
Yeah that could be consistent. The issue I found is that everything in `vcat` is generalized for splatting a vector of `DataFrame`s, so I wasn't sure about making some sort of assumption about the "first" data frame in the argument list.

nalimilan:
I guess `:left` could refer to the first argument, and `:right` to the last one. Or we could throw an error when there are more than two arguments.

pdeffebach:
`:upper`, `:lower`, `:inner`, `:outer` is a good strategy. I'm sure that's what `dplyr` would have done if `rbind` wasn't so widely used already. 

With more than 2 arguments, we could encourage people to use `reduce` or a `for` loop when `vcat`ing multiple DataFrames. Throwing an error would be quite breaking though, and I don't know how much people are calling `vcat` on a vector of DataFrames.

ryapric:
Coming over from the reference PR #1672 above.

A couple of initial thoughts out loud regarding both workstreams:

- Upon further review, I think both of these approaches might be a bit bloated. There already exists a method extension for `append!` in `src/dataframe.jl`. Would it not be better to modify this extension to achieve the desired effect, instead of defining new functions/methods, like we've both done here?

    - Edit: sorry, just noticed as well that what's even weirder, is there are method extensions for *both* `append!` and `vcat` *already on the master branch*, in different `src` files. I'm not really sure about the logic there, i.e. why both exist.

- I much prefer the splatting implementation you have here, instead of the array-based parameterization I have.

- Maybe instead of having both `widen` and `keep`, there's just one arg that defaults to the effects of `widen = true`, but any non-bool passed arg value is a vector of columns to keep? I can understand the `join`-similar syntax of `:inner`, etc., but since these *aren't* joins, I'm not sure of how intuitive those terms would be to users.

- I'm having a hard time trying to implement a clean `fillvalue` implementation myself, which requires type conversions on all respective `DataFrame` columns to the `Union` of each that is appended. I think there's two possibly good options there:

    - Keep `missing` as the default fill value, and let users act as they please

    - Let the `fillvalue` arg be set to `missing` as default. Then, just call something like:

        `if !ismissing(fillvalue) mapcols(x -> replace!(x, missing => fillvalue)) end`

    We may still run into coltype issues, though, but I'll know more once I try and implement that.

Still digging through everything, but that's what comes to mind so far.

nalimilan:
`append!` is very different from `vcat`, as it resizes vectors in-place. So we need both.

>     * Maybe instead of having both `widen` and `keep`, there's just one arg that defaults to the effects of `widen = true`, but any non-bool passed arg value is a vector of columns to keep? I can understand the `join`-similar syntax of `:inner`, etc., but since these _aren't_ joins, I'm not sure of how intuitive those terms would be to users.

Note that `OUTER UNION` exists in at least some of the major SQL implementations. Though I guess we could have a single argument which could either be a Boolean, a vector of names, or `:inner`/`:outer`/`:left`/`:right`. We just need to find a good name for it.

> I'm having a hard time trying to implement a clean `fillvalue` implementation myself, which requires type conversions on all respective `DataFrame` columns to the `Union` of each that is appended. I think there's two possibly good options there:
> 
>     * Keep `missing` as the default fill value, and let users act as they please
> 
>     * Let the `fillvalue` arg be set to `missing` as default. Then, just call something like:
>       `if !ismissing(fillvalue) mapcols(x -> replace!(x, missing => fillvalue)) end`
> 
> 
> We may still run into coltype issues, though, but I'll know more once I try and implement that.

I don't think there's a problem. We just need to allocate new columns using `Tables.allocatecolumn(promote_type(eltype(col), eltype(fillvalue)), len)`. Better not special-case `missing`. And actually the existing code already does that, and the current state of this PR takes advantage of it.

> I'm also seeing a lot of hard-coded test code for non-intersecting columns on master when calling `vcat`, so if adjusting `vcat` is still the plan, then those tests would need to be replaced. Now I'm questioning again slightly whether or not to just make this a separate function, especially in the spirit of not breaking someone's existing code... `vcatall`? `appendall!`? `bindrows`? `collapse`?

We've discussed these things a lot before [in a Discourse thread](https://discourse.julialang.org/t/vcat-does-not-append-dataframe-rows-as-it-did-in-0-10/9823). I prefer a single function since it's a standard Julia name and it's easier to discover. There's no reason why making the function more flexible should break code.

ryapric:
So [here's what I've managed to get to a WIP stage](https://github.com/ryapric/DataFrames.jl/tree/flexible_vcat2), myself, based on all the information you've all provided. What I've changed (from master) is:

- Add a `keepcols` keyword to `vcat`, which defaults to `nothing`. This is entirely ignored if not supplied explicitly by the user, maintaining perfect backwards-compatibility with how `vcat` acts today. It is checked if non-`nothing` in a few places, and only serves to conditionally prepare the incoming data. Then the rest of the existing `_vcat` code is executed as normal.

- `keepcols` accepts an array of `Symbol`s, which are the column names to keep in the output `DataFrame`. The default (if invoking the new usage) is to pass `[:allcols]`, which returns a `DataFrame` containing *all* of the columns across all passed `DataFrame`s. Values are filled with `missing` if they don't exist.

- Add some basic tests in advance so that I would know when the desired functionality was achieved. More to come.

- Edit the `ArgumentError`s thrown due to set-difference columns and not passing `keepcols`. The errors now will suggest the user to pass `keepcols` if this is the desired behavior. I have also edited the `vcat errors` section of `test/cat.jl`, so that the error outputs match these suggestions.

Were this to be ideal, it still needs at least a few things in my mind:

1. A way to reset the allocations & type conversions done to the passed `DataFrames`. I'm open to hearing everyone's thoughts on that, though, since Julia is pass-by-reference anyway.

1. Check to make sure the column order is preserved according to the first `DataFrame`. I haven't written tests for that yet, so maybe it's already behaving this way. Or, maybe that's not as important anymore?

1. More tests (I wrote just enough in advance to see if the basic applications worked, but there's more I'd like to write).

1. Reconsider the best name for the default arg to invoking `keepcols`. It could just be a single `Symbol`, `:all`, but then the type signature would be `Union{Symbol, Array{Symbol}, Nothing}}`. Not a problem, but verbose & unsightly, and only serving to allow development like this. If the join-like syntax is desired, then I don't think that would be too many tweaks from what exists now, so that's also an option.

pdeffebach:
Could you explain how that differs from the current implementation in the current state of this PR? 

@nalimilan How we should we decide whether to use a strategy that mimics `join` and whether to use a `keepcols` based strategy? I'm partial to the `join` approach, but I want to know how to make sure there is some consensus.

ryapric:
It appears that I am quite dumb. I was getting weird merge conflicts when looking at this PR's branch vs. master, so I just tried to recreate it a different way. But I got it sorted out now, and the state of this PR seems nicer. To answer your question: the functionality is the same, but mine was more verbose, and performed sullying allocations on the incoming data.

I've started from what is here, now, and pushed up some small suggested changes to the body, docs, and tests. I've taken out `widen`, and let `keep` handle all the conditional resolution. I've left `fillvalue` as-is, but in some ad-hoc testing, the chance of anything other than `missing` as a value runs the risk of converting a whole column to `Any`. So, I've warned users in the docstring of such, and added a suggested use case for `Missings.relplace()`.

Regarding the column-selection syntax: to prevent clutter in the body, I think we could really just send in the `keep` value and some of the initial header objects to another internal function in the call to define `header`, to handle greater flexibility. For example, maybe `:all` and `:outer` are synonymous in terms of what columns to select. So, *both* could resolve to `unionunique`. Maybe the same idea for `:inner`, `:same`, `:intersect`, etc, in that they all resolve to the same selection. This function could (and should) still be able to take a vector of `Symbol`s for finer selection control.

nalimilan:
> @nalimilan How we should we decide whether to use a strategy that mimics `join` and whether to use a `keepcols` based strategy? I'm partial to the `join` approach, but I want to know how to make sure there is some consensus.

As I said I think we can support both using the same argument. Not sure whether `keepcols` is the best name, need to think about it.

Do you want to finish this PR @pdeffebach? Else, @ryapric would better open a new PR, reviewing PRs against PRs will only help confusing everybody.

bkamins:
@pdeffebach 
I wanted to write to you here about #1691 again, but it seems no test files were changed in this PR 😸.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +983,88 @@ julia> vcat(df1, df2)
...
+        # it should be removed ASAP after deprecation
+        data = map(dfs) do df
+            if df isa DataFrame 
+                if haskey(df, name)
+                    return df[name] 
+                else
+                    # TODO: make this more efficient by not creating a 
+                    # full array of missing values. Instead, implement
+                    # this in the copyto! stage
+                    return fill(missing, nrow(df))
```
nalimilan: Sorry for the delay, I had missed the notification. `RepeatedVector` can probably help here (if using a generator isn't enough).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +983,88 @@ julia> vcat(df1, df2)
...
+        end
+    else 
+        header = keep 
+    end
+
     length(header) == 0 && return DataFrame()
     cols = Vector{AbstractVector}(undef, length(header))
     for (i, name) in enumerate(header)
-        data = [df[name] for df in dfs]
+        # TODO: replace with commented out code after getindex deprecation
```
bkamins: I have not looked at the PR in detail - but it seems you re-introduce some old code that was already removed here (at least in the comments)

pdeffebach: deleted.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: ugh why did I not think of a generator. It's fixed now.

pdeffebach:
I added more to-dos. I think it's more efficient now but is getting unwieldy. I will take a fresh look at it to try and simplify things.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: It turns out that a generator doesn't infer eltype correclty. Columns are of type `Any` rather than `Union{T, Missing}`. 

`RepeatedVector` doesn't work because it's not defined at this point in the code, so we can't use it here. (Should we have a place for all struct defintions?)

Let me know what to do next. I think there is an issue for this generator issue but I can't find it.

nalimilan: Ah. Maybe with `Iterators.repeated`? (`eltype` won't be defined for generators since it needs inference.)

pdeffebach:
`Iterators.repeated` did the trick for having the right column type at the end. 

fwiw I like this API a lot, I've been doing a lot of `vcat`ing in R and Stata this week and this seems like the right strategy. 

Should I go ahead and start working on tests and docs?

bkamins:
If @nalimilan is OK with the current implementation 😄, please add tests and docs and I will check if your implementation is consistent with the contract specified in the docs.

nalimilan:
Regarding the API, I think I'd still prefer a single argument which can take several forms. How about calling it `columns` and allow it to be either `:equal` (the default)/`:union`/`:intersect` (can be added later), or a vector of column names?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +992,78 @@ julia> vcat(df1, df2)
...
+            end
+            throw(ArgumentError(join(estrings, ". ")))
         end
-        throw(ArgumentError(join(estrings, ", ", ", and ")))
+    # Only have columns that are in all dataframes. No error thrown. 
+    elseif columns === :intersect 
+        header = intersectunique
+    elseif columns === :union
+        header = unionunique
+        # no setdiff cause we take in everything
```
nalimilan: These comments look out of place. Same above. Avoid describing the behavior already described in the docstring.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +983,83 @@ julia> vcat(df1, df2)
...
+        data = map(dfs) do df
+            if df isa DataFrame 
+                if haskey(df, name)
+                    return df[name]
+                else
+                    Iterators.repeated(missing, nrow(df))
+                end
+            else 
+                if haskey(df, name)
+                    return view(parent(df)[name], rows(df))
```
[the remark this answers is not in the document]
nalimilan: Do you think this will work?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +992,78 @@ julia> vcat(df1, df2)
...
         end
-        throw(ArgumentError(join(estrings, ", ", ", and ")))
+    # Only have columns that are in all dataframes. No error thrown. 
+    elseif columns === :intersect 
+        header = intersectunique
+    elseif columns === :union
+        header = unionunique
+        # no setdiff cause we take in everything
+    # Use names specified by user. Add missings as needed.
+    elseif columns isa AbstractVector{Symbol}
```
```suggestion
    else
```
is enough.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -956,20 +956,29 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
...
-different order. In such cases the order of names in the first passed
-`DataFrame` is used.
+The `columns` keyword argument determines the columns of the returned data frame:
+
+* `:equal` (the default): require all data frames to have the same column names.
+  If they appear in different orders, the order of the first provided data frame is used.
+* `:intersect`: only the columns present in *all* provided data frames are kept.
+  If the intersection is empty, an empty data frame is returned.
+* `:union`: columns present in *at least one* of the provided data frames are kept.
+  Columns not present in some data frames are filled with `missing` where necessary.
```
bkamins: please add information about how the order of columns is determined

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -956,20 +956,29 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
...
+* A vector of `Symbol`s: only listed columns are kept.
+  Columns not present in some data frames are filled with `missing` where necessary.
```
bkamins: Is it acceptable that the column is not present in any of the data frames?

bkamins: Also it would be nice to add a passage how eltype of the resulting columns is determined.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +992,78 @@ julia> vcat(df1, df2)
 │ 6   │ 6     │ 6     │
 ```
 """
-Base.vcat(df::AbstractDataFrame) = df
-Base.vcat(dfs::AbstractDataFrame...) = _vcat(collect(dfs))
-function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
+Base.vcat(df::AbstractDataFrame; 
+          columns::Union{Symbol, AbstractVector{Symbol}}=:equal) = df
```
bkamins: This implementation is incorrect when `columns` is a vector.
Also in the future we will change this to always make a copy of columns (this is a separate PR).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +992,78 @@ julia> vcat(df1, df2)
...
-        for (i, u) in enumerate(uniqueheaders)
-            matching = findall(h -> u == h, allheaders)
-            headerdiff = setdiff(coldiff, u)
-            cols = join(headerdiff, ", ", " and ")
-            args = join(matching, ", ", " and ")
-            estrings[i] = "column(s) $cols are missing from argument(s) $args"
+
+    # Throw an informative error if names arent the same across all dataframes
+    if columns === :equal
+            header = unionunique
```
bkamins: wrong indentation.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +992,78 @@ julia> vcat(df1, df2)
...
+            coldiff = setdiff(unionunique, intersectunique)
```
bkamins: it is probably enough to compare length.

src/abstractdataframe/abstractdataframe.jl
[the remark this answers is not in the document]
bkamins: I think we do not need that branch. Anyway `data` will be type unstable I think (but probably it is best to check it).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -983,35 +983,83 @@ julia> vcat(df1, df2)
 │ 6   │ 6     │ 6     │
 ```
 """
-Base.vcat(df::AbstractDataFrame) = df
-Base.vcat(dfs::AbstractDataFrame...) = _vcat(collect(dfs))
-function _vcat(dfs::AbstractVector{<:AbstractDataFrame})
+Base.vcat(df::AbstractDataFrame; 
+        widen::Bool = false,
```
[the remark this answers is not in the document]
pdeffebach: no longer needed.

src/abstractdataframe/abstractdataframe.jl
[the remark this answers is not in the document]
pdeffebach: getting rid of the branch works fine.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -956,20 +956,29 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
...
+* A vector of `Symbol`s: only listed columns are kept.
+  Columns not present in some data frames are filled with `missing` where necessary.
 
 # Example
 ```jldoctest
 julia> df1 = DataFrame(A=1:3, B=1:3);
 
 julia> df2 = DataFrame(A=4:6, B=4:6);
 
+julia> df3 = DataFrame(A=7:9, C=7:9);
```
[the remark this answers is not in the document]
pdeffebach: added examples for both `:union` and `:intersect`

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Sounds good. I might keep a few in while I work. Do you mind deleting the unnecessary ones at the end?

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
pdeffebach: added.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Yes it is acceptable that the column is not present in any of the DataFrames. This is mentioned in the bullets I think. 

I also added text about the `eltype`s including a small example.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: I added the subsetting if `columns` is a vector. Let's hope I get this in first and then you can rebase the ownership PR.

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Yeah I agree this error message is overkill. Particularly because I can't imagine someone `vcat`ing a long enough array of vectors to get confused. But it was here when I started the PR.

pdeffebach:
I'm going to write tests now. But feel free to review the documentation while I work on these.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -956,20 +956,38 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
...
 
-Vertically concatenate `AbstractDataFrames`.
+Vertically concatenate `AbstractDataFrame`s.
 
-Column names in all passed data frames must be the same, but they can have
-different order. In such cases the order of names in the first passed
-`DataFrame` is used.
+The `columns` keyword argument determines the columns of the returned data frame:
+
+* `:equal` (the default): require all data frames to have the same column names.
```
bkamins: We had a discussion with @nalimilan about the best name for this option. Alternatives are `:identical` or `:same`. @pdeffebach - do you have any opinion? (all are OK, but it would be good to choose the most intuitive)

test/cat.jl
```diff
@@ -228,6 +228,47 @@ module TestCat
         @test [df1; df2] == df3
     end
 
+    @testset "vcat with :union" begin
+        df1 = DataFrame(A = 1:3, B = 4:6)
+        df2 = DataFrame(A = 7:9)
+        df3 = DataFrame(B = 4:6, A = 1:3)
+
+        @test vcat(df1, df2; columns = :union) ≅ DataFrame([[1, 2, 3, 7, 8, 9], 
+                                                             [4, 5, 6, missing, missing, missing]],
```
nalimilan: Can you also test the element type of resulting columns?

Also, please cut lines at 92 chars. The `A=[...]` syntax is sligthly more readable IMHO.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -981,43 +997,91 @@ julia> vcat(df1, df2)
...
+├─────┼───────┤
+│ 1   │ 1     │
+│ 2   │ 2     │
+│ 3   │ 3     │
+│ 4   │ 7     │
+│ 5   │ 8     │
+│ 6   │ 9     │
 ```
 """
-Base.vcat(df::AbstractDataFrame) = df
```
nalimilan: I've removed this method since it doesn't make a copy, which is inconsistent with the general case (and can be dangerous). We could write an optimized implementation but I guess it doesn't matter a lot.

test/cat.jl
```diff
@@ -282,7 +323,13 @@ module TestCat
         err = @test_throws ArgumentError vcat(df1, df2, df3, df4, df1, df2, df3, df4, df1, df2, df3, df4)
         @test err.value.msg == "column(s) E and F are missing from argument(s) 1, 5 and 9, column(s) B are missing from argument(s) 2, 6 and 10, and column(s) F are missing from argument(s) 3, 7 and 11"
     end
-    x = view(DataFrame(A = Vector{Union{Missing, Int}}(1:3)), 2:2, :)
-    y = DataFrame(A = 4:5)
-    @test vcat(x, y) == DataFrame(A = [2, 4, 5])
+
```
nalimilan: One line break is enough.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: let's go with `:same`

test/cat.jl
pdeffebach: Added element type, made columns correct, and changed constructors.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: Good point. I'm going to add a method for an empty dataframe, though.

test/cat.jl
pdeffebach: fixed.

pdeffebach:
This is getting close. I pus some effort into making ` vcat` work, which makes something like. 

```
d = DataFrame()
for df in [df1, df2]
   vcat(d, df)
end
```

Obviously  this could be done with `reduce` but we now allow `push!` on an emtpy data frame so this seems like a good idea. 

Unfortunately some tests have forbid this behavior [here](    @test_throws ArgumentError vcat(df, missing_df)). That section also has some weird constructors I have never seen before like `DataFrame(Int, 0, 0)`. Not sure what the logic is behind this section.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -956,20 +956,39 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
...
 
-Vertically concatenate `AbstractDataFrames`.
+Vertically concatenate `AbstractDataFrame`s.
 
-Column names in all passed data frames must be the same, but they can have
-different order. In such cases the order of names in the first passed
-`DataFrame` is used.
+The `columns` keyword argument determines the columns of the returned data frame:
+
+* `:same` (the default): require all data frames to have the same column names.
```
bkamins: Finally we have merged `:equal` in `push!` per the earlier recommendation.
@nalimilan - do we keep `:equal` or change to `:same` everywhere?

src/abstractdataframe/abstractdataframe.jl
bkamins: The ownership PR sets this to `Base.vcat(df::AbstractDataFrame) = DataFrame(df)` and I think this is how it could stay (but a general `vcat` should handle it correctly - just please make sure that we test this case).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -981,43 +1000,94 @@ julia> vcat(df1, df2)
...
-        estrings = Vector{String}(undef, length(uniqueheaders))
-        for (i, u) in enumerate(uniqueheaders)
-            matching = findall(h -> u == h, allheaders)
-            headerdiff = setdiff(coldiff, u)
-            cols = join(headerdiff, ", ", " and ")
-            args = join(matching, ", ", " and ")
-            estrings[i] = "column(s) $cols are missing from argument(s) $args"
-        end
+
+    if columns === :same
```
bkamins: If we have `:same` and we `vcat` some data frame and an empty data frame should it be OK or throw an error. (note that empty means no rows and no columns, as no-rows and some columns data frame should go a standard test)

pdeffebach: I don't think so. It would be nice to have some behavior like 

```
df = DataFrame()

for file in files
    d = CSV.read(file) |> DataFrame
    vcat(df, d)
end
```

We think the same behavior is helpful with `push!` and allow that. At some point I might want a PR that nails down consistency across `append!`, `vcat`, and `push!`.

bkamins: Can you please comment how is it related with `push!`? `push!` adds exactly one row. While the case we are discussing here assumes that we add 0 rows in a data frame with 0 columns.

Regarding `append!` - this is exactly the behavior we have now. I.e. you can call `append!(df, DataFrame())` and this is a no-op. So the only inconsistent function is `vcat` (where it should not be no-op, but a copy of the first `DataFrame`)

bkamins: Also you are not allowed to `push!` a `DataFrame` to a `DataFrame`.

pdeffebach: Oh sorry! I misread your comment! It's about `:same` rather than other methods.

I vote for `vcat(DataFrame(), df` to never throw an error. The user's behavior inside the above loop I laid out above should be dictated by the data in their files, not the rules of `vcat` with an empty data frame. 

Could you comment on the issue I raised in another comment here? The `vcat(DataFrame(), df)` methods are prohibited from tests, but its also testing behavior I don't know about.

bkamins: `DataFrame(Int, 0, 0)` is the same as `DataFrame()`.

Tests:
```
@test_throws ArgumentError vcat(missing_df, df)
@test_throws ArgumentError vcat(df, missing_df)
```
test the old behavior (disallowing `vcat`-ing non-empty `DataFrame` with empty `DataFrame`) that is exactly what I propose to change (and this is a good sign that we have tests for this in place 😄 - in general we should this level of detail in testing corner cases everywhere).

pdeffebach:
I added tests for an empty dataframe and ensure everything passes!

We can hash out `:equal` vs `:same` in another PR I think.

bkamins:
But in this case let us leave `:equal` here, as we need to have consistency (and later change to `:same` in both places if we want to; @nalimilan - or simply we now decide to stick to `:equal`?).

Also note that this PR needs to be rebased (it now has conflict with master), pass tests after rebase, and then we can move forward with a final review before merging.

pdeffebach:
I hate git :(

bkamins:
test/cat.jl still has a conflict (and unfortunately it is rather large). Tests will not run unless the conflicts are resolved (this is the most painful thing for me also; imagine rebasing https://github.com/JuliaData/DataFrames.jl/pull/1742, which I already had to do twice if I remember correctly).

pdeffebach:
I think i skipped the last `merge remote tracking branch` last time and that's where it went wrong. I think i successfully rebased now.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Let's keep `:equal` and if we're not happy about it change uses of it in `push!` at the same time.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1015,43 +1034,94 @@ julia> vcat(df1, df2)
...
+│     │ Int64 │
+├─────┼───────┤
+│ 1   │ 1     │
+│ 2   │ 2     │
+│ 3   │ 3     │
+│ 4   │ 7     │
+│ 5   │ 8     │
+│ 6   │ 9     │
+
+d = DataFrame()
```
nalimilan: Make this a proper REPL doctest with output or remove it.

src/abstractdataframe/abstractdataframe.jl
nalimilan: So what's the best solution? Which PR should we merge first?

test/cat.jl
```diff
@@ -121,9 +121,29 @@ end
     df[1:2, 1:2] = 3
     df[[true,false,false,true], 2:3] = 3
 
-    # vector broadcasting assignment of subtables
-    df[1:2, 1:2] = [3,2]
+    # vector broadcasting assignment of subtables
```
nalimilan: Remove trailing spaces here and below.

test/cat.jl
```diff
@@ -121,9 +121,29 @@ end
...
+    @test vcat(missing_df, df) == df
+    @test vcat(df, missing_df) == df
+    @test eltypes(vcat(df, df)) == Type[Float64, Float64, Int]
+    @test size(vcat(df, df)) == (size(df, 1) * 2, size(df, 2))
+    res = vcat(df, df)
+    @test res[1:size(df, 1), :] == df
+    @test res[1+size(df, 1):end, :] == df
+    @test eltypes(vcat(df, df, df)) == Type[Float64, Float64, Int]
+    @test size(vcat(df, df, df)) == (size(df, 1) * 3, size(df, 2))
+    res = vcat(df, df, df)
```
nalimilan: Move this line above to avoid calling `vcat(df, df, df)` three times.

test/cat.jl
```diff
@@ -121,9 +121,29 @@ end
...
+
+    @test vcat(missing_df) == DataFrame()
+    @test vcat(missing_df, missing_df) == DataFrame()
+    @test vcat(missing_df, df) == df
+    @test vcat(df, missing_df) == df
+    @test eltypes(vcat(df, df)) == Type[Float64, Float64, Int]
+    @test size(vcat(df, df)) == (size(df, 1) * 2, size(df, 2))
+    res = vcat(df, df)
+    @test res[1:size(df, 1), :] == df
+    @test res[1+size(df, 1):end, :] == df
```
nalimilan: This doesn't seem correct?

test/cat.jl
```diff
@@ -155,14 +175,169 @@ end
...
+        df3 = DataFrame(A = [1, 11], B = [2, 12])
+        @test [df1; df2] == df3
+    end
+
+    @testset "vcat with :union" begin
+        df1 = DataFrame(A = 1:3, B = 4:6)
+        df2 = DataFrame(A = 7:9)
+        df3 = DataFrame(B = 4:6, A = 1:3)
+
+        @test vcat(df1, df2; columns = :union) ≅ DataFrame(A =Int[1, 2, 3, 7, 8, 9],
```
nalimilan: No need to explicitly write types. Also add space before `=`, and a line break before `DataFrame` to avoid going beyond 92 chars.

test/cat.jl
```diff
@@ -155,14 +175,169 @@ end
...
+    empty_dfs = [DataFrame(), DataFrame(), DataFrame()]         
+    df1 = DataFrame(A = 1:3, B = 4:6, C = 7:9)
     @test vcat(empty_dfs...) == reduce(vcat, empty_dfs) == DataFrame()
 
     df = DataFrame(x = trues(1), y = falses(1))
     dfs = [df, df, df]
     @test vcat(dfs...) ==reduce(vcat, dfs) == DataFrame(x = trues(3), y = falses(3))
 end
 
+    @testset "vcat out of order" begin
```
nalimilan: This whole block is duplicated, right?

test/cat.jl
```diff
@@ -155,14 +175,169 @@ end
...
+        @test err.value.msg == "column(s) E and F are missing from argument(s) 1, column(s) B are missing from argument(s) 2, and column(s) F are missing from argument(s) 3"
+        err = @test_throws ArgumentError vcat(df1, df1, df2, df2, df3, df3, df4, df4)
+        @test err.value.msg == "column(s) E and F are missing from argument(s) 1 and 2, column(s) B are missing from argument(s) 3 and 4, and column(s) F are missing from argument(s) 5 and 6"
+        err = @test_throws ArgumentError vcat(df1, df1, df1, df2, df2, df2, df3, df3, df3, df4, df4, df4)
+        @test err.value.msg == "column(s) E and F are missing from argument(s) 1, 2 and 3, column(s) B are missing from argument(s) 4, 5 and 6, and column(s) F are missing from argument(s) 7, 8 and 9"
+        err = @test_throws ArgumentError vcat(df1, df2, df3, df4, df1, df2, df3, df4, df1, df2, df3, df4)
+        @test err.value.msg == "column(s) E and F are missing from argument(s) 1, 5 and 9, column(s) B are missing from argument(s) 2, 6 and 10, and column(s) F are missing from argument(s) 3, 7 and 11"
+    end
+
+    @testset "views" begin
```
nalimilan:
```suggestion
    @testset "vcat with views" begin
```

src/abstractdataframe/abstractdataframe.jl
pdeffebach: added.

test/cat.jl
pdeffebach: fixed.

test/cat.jl
pdeffebach: fixed.

test/cat.jl
pdeffebach: changed.

test/cat.jl
pdeffebach: The second block was better because it used the better constructor. So I used that one.

src/abstractdataframe/abstractdataframe.jl
nalimilan: `julia>` is still missing.

test/cat.jl
nalimilan: Still there.

test/cat.jl
nalimilan: I mean that the file now contains two `@testset "vcat out of order"` sections. Also fix indentation.

test/cat.jl
pdeffebach: I added parentheses for clarity.

test/cat.jl
pdeffebach: I diff-checked them and the only difference was the constructors. So I deleted the one what had the `DataFrame(cols= [a, b, c], names = [:a, :b, :c])` constructors.

test/cat.jl
```diff
@@ -249,8 +302,6 @@ end
     @test err.value.msg == "column(s) B are missing from argument(s) 1"
     # multiple missing 1 column
     err = @test_throws ArgumentError vcat(df1, df2, df2, df2, df2, df2)
-    err2 = @test_throws ArgumentError reduce(vcat, [df1, df2, df2, df2, df2, df2])
```
pdeffebach: @oxinabox I think my rebase may have collided with your optimization for `reduce(vcat, dfs)`. Should this throw an error or not?

pdeffebach:
I think this PR is ready to go, pending one last question to Lyndon about a test for `reduce(vcat, dfs)`.

Sorry i took so long to get back to this PR! I see the benefit of not letting these stagnate and have a painful rebase.

test/cat.jl
oxinabox: If `vcat(xs...)` throws an error, then `reduce(vcat, xs)` should throw the same error.
If not, then not.

bkamins:
@pdeffebach can you please resolve merge conflicts in this PR (otherwise it will not go through CI).

pdeffebach:
I think I successfully resolved the conflicts

test/cat.jl
pdeffebach: Okay then this error should remain deleted. It's good to go.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1017,20 +1017,39 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
           makeunique=makeunique, copycols=copycols)
 
 """
-    vcat(dfs::AbstractDataFrame...)
+    vcat(dfs::AbstractDataFrame...; columns::Union{Symbol, AbstractVector{Symbol}}=:equal)
```
nalimilan: Sorry, but since this PR was opened we realized we use `cols` rather than `columns` in general (e.g. in function names), so better call this `cols` I guess.

test/cat.jl
```diff
@@ -382,11 +387,59 @@ end
...
+                  B = [4, 5, 6, missing, missing, missing],
+                  C = [missing, missing, missing, missing, missing, missing])
+
+    @test vcat(df1, df2, df3; columns = [:A, :B, :C]) ≅
+        DataFrame(A = [1, 2, 3, 7, 8, 9, 10, 11, 12],
+                  B = [4, 5, 6, missing, missing, missing, missing, missing, missing],
+                  C = [missing, missing, missing, missing, missing, missing, 13, 14, 15])
+end
+
+@testset "vcat on empty dataframe in loop" begin
```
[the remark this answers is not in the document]
pdeffebach: I thought at one point you wanted a test that mimicked the desired use of this functionality. But I can delete it.

test/cat.jl
```diff
@@ -347,25 +351,26 @@ end
...
+    @test vcat(df1, df2, df2) == DataFrame(A = [1, 2, 3, 14, 16, 18, 14, 16, 18],
+                                           B = [4, 5, 6, 8, 10, 12, 8, 10, 12],
+                                           C = [7, 8, 9, 2, 4, 6, 2, 4, 6])
+    @test vcat(df2, df1, df2) == DataFrame(C = [2, 4, 6, 7, 8, 9, 2, 4, 6],
+                                           B = [8, 10, 12, 4, 5, 6, 8, 10, 12],
+                                           A = [14, 16, 18, 1, 2, 3, 14, 16, 18])
     @test size(vcat(df1, df1, df1, df2, df2, df2)) == (18, 3)
     df3 = df1[[1, 3, 2]]
     res = vcat(df1, df1, df1, df2, df2, df2, df3, df3, df3, df3)
-    @test res == reduce(vcat, [df1, df1, df1, df2, df2, df2, df3, df3, df3, df3])
```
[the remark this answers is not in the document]
pdeffebach: not sure why that got left out.

test/cat.jl
```diff
@@ -347,25 +351,26 @@ end
...
-                                            [8, 10, 12, 4, 5, 6, 8, 10, 12],
-                                            [14, 16, 18, 1, 2, 3, 14, 16, 18]] ,[:C, :B, :A])
-
+    df2 = DataFrame(colwise(x->2x, df1), reverse(names(df1)))
+    @test vcat(df1, df2) == DataFrame(A = [1, 2, 3, 14, 16, 18],
+                                      B = [4, 5, 6, 8, 10, 12],
+                                      C = [7, 8, 9, 2, 4, 6])
+    # test with keyword argument for `columns`
+    @test vcat(df1, df2, columns = :equal) == DataFrame(A = [1, 2, 3, 14, 16, 18],
+                                                       B = [4, 5, 6, 8, 10, 12],
```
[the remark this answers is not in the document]
pdeffebach: fixed.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: changed to `cols` throughout

test/cat.jl
pdeffebach: I missed what this error was about. I have added it back in.

pdeffebach:
Okay I have responded to your most recent review.

src/abstractdataframe/abstractdataframe.jl
nalimilan: I don't see it. Did you push the commit?

test/cat.jl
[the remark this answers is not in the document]
nalimilan: Well a loop isn't different from calling `vcat` on two data frames, so...

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1042,43 +1061,104 @@ julia> vcat(df1, df2)
...
-        end
+
+    if columns === :equal
+        header = unionunique
+        coldiff = setdiff(unionunique, intersectunique)
+        
+        if !isempty(coldiff) 
+            # if any DataFrames are a full superset of names, skip them
+            filter!(u -> Set(u) != Set(header), uniqueheaders)
+            estrings = Vector{String}(undef, length(uniqueheaders))
```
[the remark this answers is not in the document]
nalimilan: The point of my suggestion was that you can do `estrings = map...`, and not set `estrings[i] =` inside the anonymous function, but just return the string.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1042,47 +1061,108 @@ julia> vcat(df1, df2)
...
+│     │ Int64 │
+├─────┼───────┤
+│ 1   │ 1     │
+│ 2   │ 2     │
+│ 3   │ 3     │
+│ 4   │ 7     │
+│ 5   │ 8     │
+│ 6   │ 9     │
+
+julia> d = DataFrame();
```
nalimilan: Call this `d4` then.

test/cat.jl
```diff
@@ -382,11 +391,59 @@ end
...
+    df1 = DataFrame(A = 1:3, B = 4:6)
+    df2 = DataFrame(A = 7:9)
+    df3 = DataFrame(A = 10:12, C = 13:15)
+
+    @test vcat(df1, df2; cols = :intersect) ≅ DataFrame(A = [1, 2, 3, 7, 8, 9])
+    @test vcat(df1, df2, df3; cols = :intersect) ≅ DataFrame(A = [1, 2, 3, 7, 8, 9,
+                                                                     10, 11, 12])
+end
+
+@testset "vcat with cols=:Vector" begin
```
nalimilan:
```suggestion
@testset "vcat with cols::Vector" begin
```

test/cat.jl
```diff
@@ -397,8 +454,8 @@ end
     @test err.value.msg == "column(s) B are missing from argument(s) 1"
     # multiple missing 1 column
     err = @test_throws ArgumentError vcat(df1, df2, df2, df2, df2, df2)
-    err2 = @test_throws ArgumentError reduce(vcat, [df1, df2, df2, df2, df2, df2])
-    @test err == err2
+    @test err.value.msg == "column(s) B are missing from argument(s) 2, 3, 4, 5 and 6"
+    err = @test_throws ArgumentError reduce(vcat, [df1, df2, df2, df2, df2, df2])
     @test err.value.msg == "column(s) B are missing from argument(s) 2, 3, 4, 5 and 6"
```
nalimilan: Is this really the error that is thrown? I would expect something about only two arguments, and if it was identical to the previous error then `@test err = err2` would be enough.

test/cat.jl
```diff
@@ -347,25 +350,31 @@ end
 
 @testset "vcat out of order" begin
     df1 = DataFrame(A = 1:3, B = 4:6, C = 7:9)
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
pdeffebach: fixed. I thought i checked the tests.

src/abstractdataframe/abstractdataframe.jl

test/cat.jl
pdeffebach: It's the error that is thrown. it's to check that `reduce(vcat, dfs)` behaves exactly similar to `vcat(dfs...)` 

However the `reduce(vcat,...)` error probably shouldn't be that since it refers to `argument(s)` and the `reduce` version has only 2 arguments. But I don't really want special error handling for just a small optimization of `recuce(vcat, dfs)`. 

I changed the code to test that the error messages are equal for now.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: fixed in documentation. I don't think there is anywhere else.

test/cat.jl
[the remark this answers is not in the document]
pdeffebach: deleted that testset.

test/cat.jl
nalimilan: OK, sounds fine. It's not a problem if the error is smarter than what we expect. ;-)

bkamins:
@pdeffebach Please let me know when you would be available to add/discuss the additional tests I have proposed (I would like to have new `vcat` in the current release and I am planning to do it soon).

bkamins:
@pdeffebach - I have added the requested tests and done some layout improvements and merged the PR. Thank you for the contribution.

pdeffebach:
Thank you for putting this over the finish line! So I didn't respond earlier.

bkamins:
Thank you for the initiative of the improvement.

PR #1664, opened 2019-01-03
Add custom functions to `describe`

### LINKED ISSUE #1436: allow `describe` function to take arbitrary functions
I'm absolutely loving the new `describe` function, but I'm wondering if it should be allowed to take arbitrary functions as arguments.

For example,
```julia
describe(df, stats=[:mean, :min, :custom1=>f])
```

Any thoughts on this?  I can do the PR if we decide on what for this should take.

pdeffebach:
This resolves #1436 by allowing the user to specify a `Pair` in their vector of `stats` . 

One design choice that needs consensus is whether or not we do the `skipmissing` for the user. I vote yes because we do that for `:mean` and we dont make the user do `skipmissing` there either.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -388,10 +388,11 @@ describe(df, :min, :sum => sum)
 ```
 
 """
-StatsBase.describe(df::AbstractDataFrame) = _describe(df, [:mean, :min, :median, 
-                                                           :max, :nunique, :nmissing, 
-                                                            :eltype])
-StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) = _describe(df, collect(stats))
+StatsBase.describe(df::AbstractDataFrame) = 
+    _describe(df, [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
+
+StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) = 
+    _describe(df, collect(stats))
 
 
 function _describe(df::AbstractDataFrame, stats::AbstractVector)   
@@ -405,13 +406,11 @@ function _describe(df::AbstractDataFrame, stats::AbstractVector)
         i = findfirst(s -> s == :all, stats)
         splice!(stats, i, allowed_fields) # insert in the stats vector to get a good order
     elseif :all in predefined_funs 
-            throw(ArgumentError("`:all` must be the only `Symbol` argument.")) 
-    else 
-        if !issubset(predefined_funs, allowed_fields)
-            not_allowed = join(setdiff(predefined_funs, allowed_fields), ", :")
-            allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
-            throw(ArgumentError(":$not_allowed not allowed." * allowed_msg))
-        end
+        throw(ArgumentError("`:all` must be the only `Symbol` argument.")) 
+    elseif !issubset(predefined_funs, allowed_fields)
+        not_allowed = join(setdiff(predefined_funs, allowed_fields), ", :")
+        allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
+        throw(ArgumentError(":$not_allowed not allowed." * allowed_msg))
     end
 
     custom_funs = Pair[s for s in stats if s isa Pair]
@@ -419,9 +418,8 @@ function _describe(df::AbstractDataFrame, stats::AbstractVector)
     ordered_names = [s isa Symbol ? s : s[1] for s in stats]
     
     if !allunique(ordered_names)
-        d = StatsBase.countmap(ordered_names)
-        duplicate_names = join(unique([name for name in ordered_names if d[name] > 1]), ", :") 
-        throw(ArgumentError("Duplicate names not allowed. Duplicated value(s) are: :$(duplicate_names)"))
+        duplicate_names = unique(ordered_names[nonunique(DataFrame(ordered_names = ordered_names))])
+        throw(ArgumentError("Duplicate names not allowed. Duplicated value(s) are: :$(join(duplicate_names, ", "))"))
     end
 
     # Put the summary stats into the return data frame
@@ -446,7 +444,7 @@ function _describe(df::AbstractDataFrame, stats::AbstractVector)
         if :first in predefined_funs 
             d[:first] = isempty(col) ? nothing : first(col)
         end
-        
+
         if :last in predefined_funs
             d[:last] = isempty(col) ? nothing : last(col)
         end
@@ -508,7 +506,7 @@ function get_stats(col::AbstractVector, stats::AbstractVector{Symbol})
     return d
 end
 
-function get_stats!(d::Dict, col::AbstractVector, stats::AbstractVector{Pair})
+function get_stats!(d::Dict, col::AbstractVector, stats::AbstractVector{<:Pair})
     for stat in stats 
         d[stat[1]] = try stat[2](col) catch end
     end
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -375,31 +375,40 @@ describe(df, stats = [:min, :max])
 ```
 
 """
-StatsBase.describe(df::AbstractDataFrame) = describe(df, :mean, :min, :median, 
+StatsBase.describe(df::AbstractDataFrame) = _describe(df, [:mean, :min, :median, 
                                                      :max, :nunique, :nmissing, 
-                                                     :eltype)
-function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair}...)
-   
+                                                     :eltype])
+function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) 
+    _describe(df, collect(stats))
+end
+
+function _describe(df::AbstractDataFrame, stats::Vector{>:Union{Symbol, Pair{Symbol}}})   
     predefined_funs = Symbol[s for s in stats if s isa Symbol] 
-    if :all in predefined_funs
+    
+    allowed_fields = [:mean, :std, :min, :q25, :median, :q75,
+                      :max, :nunique, :nmissing, :first, :last, :eltype]
+    
+    if predefined_funs == [:all]
         predefined_funs = allowed_fields
+    else 
+        if !issubset(predefined_funs, allowed_fields)
+            not_allowed = setdiff(predefined_funs, allowed_fields)
+            allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
+            # todo: fix the printing of this
+            throw(ArgumentError("$not_allowed not allowed." * allowed_msg))
+        end
     end
 
     custom_funs = Pair[s for s in stats if s isa Pair]
 
     # Get the names in the order they appear
     ordered_names = [stat isa Symbol ? stat : stat[1] for stat in stats]
-    # Make the names unique ? maybe not
-
-    allowed_fields = [:mean, :std, :min, :q25, :median, :q75,
-                      :max, :nunique, :nmissing, :first, :last, :eltype]
-
-    if !issubset(predefined_funs, allowed_fields)
-        not_allowed = setdiff(predefined_funs, allowed_fields)
-        allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
-        # todo: fix the printing of this
-        throw(ArgumentError("$not_allowed not allowed." * allowed_msg))
+    if !allunique(ordered_names)
+        d = StatsBase.countmap(ordered_names)
+        duplicate_names = [name for name in ordered_names if d[name] > 1] 
+        throw(ArgumentError("Duplicate names not allowed: $(duplicate_names)"))
     end
+
 
     # Put the summary stats into the return data frame
     data = DataFrame()
@@ -437,7 +446,6 @@ function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair}...
         data[stat] = [column_stats_dict[stat] for column_stats_dict in column_stats_dicts]
     end
 
-    # re-order columns according to the names from above 
     return data
 end
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -375,7 +375,7 @@ describe(df, stats = [:min, :max])
 ```
 
 """
-function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector{Symbol}} =
+function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector} =
                             [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
     # Check that people don't specify the wrong fields.
     allowed_fields = [:mean, :std, :min, :q25, :median, :q75,
@@ -393,8 +393,11 @@ function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractV
         end
     end
 
-    if !issubset(stats, allowed_fields)
-        disallowed_fields = setdiff(stats, allowed_fields)
+    custom_funs = [stat for stat in stats if stat isa Pair ]
+    predefined_funs = Symbol[stat for stat in stats if stat isa Symbol]
+
+    if !issubset(predefined_funs, allowed_fields)
+        disallowed_fields = setdiff(predefined_funs, allowed_fields)
         allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
         not_allowed = "Field(s) not allowed: :" * join(disallowed_fields, ", :") * "."
         throw(ArgumentError(not_allowed * allowed_msg))
@@ -407,31 +410,47 @@ function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractV
     # An array of Dicts for summary statistics
     column_stats_dicts = map(columns(df)) do col
         if eltype(col) >: Missing
-            d = get_stats(collect(skipmissing(col)), stats)
+            d = get_stats(collect(skipmissing(col)), predefined_funs)
         else
-            d = get_stats(col, stats)
+            d = get_stats(col, predefined_funs)
         end
 
-        if :nmissing in stats
+        if :nmissing in predefined_funs 
             d[:nmissing] = eltype(col) >: Missing ? count(ismissing, col) : nothing
         end
 
-        if :first in stats
+        if :first in predefined_funs 
             d[:first] = isempty(col) ? nothing : first(col)
         end
-
-        if :last in stats
+        
+        if :last in predefined_funs
             d[:last] = isempty(col) ? nothing : last(col)
         end
 
         return d
     end
 
-    for stat in stats
+    for stat in predefined_funs
         # for each statistic, loop through the columns array to find values
         # letting the comprehension choose the appropriate type
         data[stat] = [column_stats_dict[stat] for column_stats_dict in column_stats_dicts]
     end
+
+    # handle custom functions
+    for custom_fun in custom_funs
+        data[custom_fun[1]] = map(columns(df)) do col 
+            if eltype(col) >: Missing 
+                return try custom_fun[2](skipmissing(col)) catch nothing end
+            else
+                return try custom_fun[2](col) catch nothing end
+            end
+        end
+    end
+
+    # Get the order of column names as they were inputted by the user 
+    ordered_names = [stat isa Symbol ? stat : stat[1] for stat in stats]
+    permutecols!(data, [:variable; ordered_names]) # :variable is the first column
+
     return data
 end
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -378,18 +378,20 @@ describe(df, stats = [:min, :max])
 StatsBase.describe(df::AbstractDataFrame) = _describe(df, [:mean, :min, :median, 
                                                      :max, :nunique, :nmissing, 
                                                      :eltype])
-function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) 
+function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...)
     _describe(df, collect(stats))
 end
 
-function _describe(df::AbstractDataFrame, stats::Vector{>:Union{Symbol, Pair{Symbol}}})   
+function _describe(df::AbstractDataFrame, stats::Vector)   
     predefined_funs = Symbol[s for s in stats if s isa Symbol] 
-    
+
     allowed_fields = [:mean, :std, :min, :q25, :median, :q75,
                       :max, :nunique, :nmissing, :first, :last, :eltype]
-    
+
     if predefined_funs == [:all]
         predefined_funs = allowed_fields
+        i = findfirst(s -> s == :all, stats)
+        splice!(stats, i, allowed_fields) # insert in the stats vector to get a good order
     else 
         if !issubset(predefined_funs, allowed_fields)
             not_allowed = setdiff(predefined_funs, allowed_fields)
@@ -401,14 +403,13 @@ function _describe(df::AbstractDataFrame, stats::Vector{>:Union{Symbol, Pair{Sym
 
     custom_funs = Pair[s for s in stats if s isa Pair]
 
-    # Get the names in the order they appear
-    ordered_names = [stat isa Symbol ? stat : stat[1] for stat in stats]
+    ordered_names = [s isa Symbol ? s : s[1] for s in stats]
+    
     if !allunique(ordered_names)
         d = StatsBase.countmap(ordered_names)
         duplicate_names = [name for name in ordered_names if d[name] > 1] 
-        throw(ArgumentError("Duplicate names not allowed: $(duplicate_names)"))
+        throw(ArgumentError("Duplicate names not allowed: $(duplicate_names) are duplicated"))
     end
-
 
     # Put the summary stats into the return data frame
     data = DataFrame()
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -370,8 +370,8 @@ Consequently, `nmissing = 0` indicates that the column allows
 missing values, but does not currently contain any.
 
 Custom functions perform call `skipmissing` on columns of eltype `Union{T, Missing}`, 
-users should be aware that because of this, users cannot specify specialy handling
-of missing values in any of the functions given via a `Pair` argument
+users should be aware that because of this, they cannot specify special handling
+of missing values in any of the functions given via a `Pair` argument.
 
 
 **Examples**
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -335,21 +335,21 @@ Report descriptive statistics for a data frame
 
 ```julia
 describe(df::AbstractDataFrame)
-describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...)
+describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{<:Symbol}}...)
 ```
 
 **Arguments**
 
 * `df` : the `AbstractDataFrame`
-* `stats::Union{Symbol, Pair{Symbol}}...` : the summary statistics to report. 
+* `stats::Union{Symbol, Pair{<:Symbol}}...` : the summary statistics to report. 
   Arguments can be: 
     *  A symbol from the list `:mean`, `:std`, `:min`, `:q25`, 
       `:median`, `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and 
       `:nmissing`. The default statistics used
       are `:mean`, `:min`, `:median`, `:max`, `:nunique`, `:nmissing`, and `:eltype`.
     * `:all` as the only `Symbol` argument to return all statistics. 
-    * Finally, users can provide their own functions in the form of a 
-      `name => function` pairs where `name` is a `Symbol`. 
+    * A `name => function` pair where `name` is a `Symbol`. This will create
+      a column of summary statistics with the provided name. 
 
 **Result**
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -342,12 +342,14 @@ describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...)
 
 * `df` : the `AbstractDataFrame`
 * `stats::Union{Symbol, Pair{Symbol}}...` : the summary statistics to report. 
-    * Arguments can be symbols from the following: `:mean`, `:std`, `:min`, `:q25`, 
+  Arguments can be: 
+    *  A symbol from the list `:mean`, `:std`, `:min`, `:q25`, 
       `:median`, `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and 
       `:nmissing`. The default statistics used
       are `:mean`, `:min`, `:median`, `:max`, `:nunique`, `:nmissing`, and `:eltype`.
-    * Alternatively, specify `:all` as the only `Symbol` argument to return all statistics. 
-    * Finally, users can provide their own functions in the form of a `Pair{Symbol, Any}`.
+    * `:all` as the only `Symbol` argument to return all statistics. 
+    * Finally, users can provide their own functions in the form of a 
+      `name => function` pairs where `name` is a `Symbol`. 
 
 **Result**
 
@@ -369,9 +371,10 @@ If the column does not allow missing values, `nothing` is returned.
 Consequently, `nmissing = 0` indicates that the column allows
 missing values, but does not currently contain any.
 
-Custom functions perform call `skipmissing` on columns of eltype `Union{T, Missing}`, 
-users should be aware that because of this, they cannot specify special handling
-of missing values in any of the functions given via a `Pair` argument.
+If custom functions are provided, they are called repeatedly with the vector corresponding
+to each column as the only argument. For columns allowing for missing values,
+the vector is wrapped in a call to [`skipmissing`](@ref): custom functions must therefore
+support such objects (and not only vectors), and cannot access missing values.
 
 
 **Examples**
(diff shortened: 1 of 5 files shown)


pdeffebach:
I posted on discourse about a thought related to this. 

One application I could see for including your own functions into `describe` is weighted means. for example one might do 

```
describe(df, stats = [:mean, :mean_w1 => x -> mean(x, weights(w1)), :mean_w2 => mean(x, weights(w2))])
```

In Stata you can do this exact operation quite easily by specifying `pw = w1` etc. in the `summarize` command. 

We have the opportunity to be more flexible here, I think. But above I mention how I opted to have an automatic `skipmissing` be done to the columns. This would make weighted means, and a lot of other operations, pretty worthless. 

Maybe it's not worth doing custom functions at all, if more of these roadblocks present themselves.

bkamins:
Before going into details of the review I would like to discuss the design. Given current behavior of `by` I think it would be more natural to have the following signature of `describe`:
```
describe(::AbstractDataFrame, args...; kwargs...)
```

Then:
1. any element of `args` can be:
    * a symbol from a predefined list of symbols except `:all` which is handled in point 3 (and here we use the current mechanics and this gives us some kind of backward compatibility)
    * a function, then column name is generated using `funname`
2. kwargs have a form `column_name=function`
3. we have a special call `describe(::AbstractDataFrame, :all)` for backward compatibility
4. we set `makeunique=true`

I do not have a clear view about `skipmissing` by default (@nalimilan - any opinion?) as I can see pros and cons of both approaches.

nalimilan:
Skipping missing values by default would be consistent with what we currently do for default statistics, which is justified since we also print the number of missing values. I think we'd better defer the issue of the weighted mean: we need to find a general solution even outside of `describe` anyway (https://github.com/JuliaLang/julia/issues/30596 and [on Discourse](https://discourse.julialang.org/t/how-to-calculate-a-weighted-mean-with-missing-observations/19281)).

pdeffebach:
> we have a special call describe(::AbstractDataFrame, :all) for backward compatibility

Just a note, currently everything is specified in the `stats` keyword argument. I've got no problem with getting rid of that, but the switch from a keyword argument to `args...` would be breaking. 

One thing I considered when writing this PR was the order of arguments. Going back to the example of a weighted mean, a user might want to have the weighted mean and the un-weighted mean side by side in the resulting DataFrame for comparison.

bkamins:
> Just a note, currently everything is specified in the `stats` keyword argument.

I agree this would be breaking, but it would be more consistent with the rest of the package. Using a keyword argument with a vector is not a standard approach in Julia in general AFAIK.

Of course this would mean that we would have to go through a deprecation period, in which a special case of only `stats` keyword argument passed should be handled as it is now and print a deprecation warning (I know it is a pain but it cannot be helped).

Order of arguments is a good point. But in this case I would go for no keyword arguments at all and using `Pairs` as you have proposed, e.g. like this `describe(df, :mean, fun, :custom=>fun2)`. I simply do not see a big value out of using `stats` and a vector. But I am not very strongly advocating it.

nalimilan:
Indeed, passing pairs would be consistent with the new `combine` API.  Like for `combine`, we could allow passing either one or more functions or `col=>fun` pairs, or a vector thereof ("functions" could also be a symbol among the predefined list). We can deprecate the `stats` keyword argument, simply making it positional.

One issue with varargs is that it would force the specialization of the function on the specified stats, which is counter-productive here since we iterate over columns. Though that can be avoided by passing `collect(args)` to the method taking a vector.

bkamins:
@pdeffebach 
Independent of @nalimilan comments I have recently merged #1691. Because of this the test files require rebasing as their major clean-up was performed.

The change was removal of indentation in module block. If you prefer that I do the rebasing then please let me know (in general you have to use the current upstream master and reapply only the changes you have introduced in the PR).

pdeffebach:
@bkamins Let me know if I did this right! I'm still new at this.

bkamins:
@pdeffebach Thank you for your work. You have performed *git merge* and I have suggested *git rebase*. The difference is that after merge we have thousands of lines of differences to compare and it is very hard to track down your changes.

As a consequence in your PR you have now a lot of commits that are unrelated and already commited to master.

For me it would be much simpler, if it is OK with you, that you revert the merge locally and force push the old version (before the merge) and I will resolve the merge conflicts for you.

If you want to learn how to do it yourself the simplest option is to use [GitHub GUI](https://help.github.com/articles/resolving-a-merge-conflict-on-github/) to resolve merge conflicts instead of performing full merge of master into your branch.

Sorry for the problems (that is why I offer to fix them myself 😄), but I hope you can see in "Files changed" tab that now this PR is very hard to analyze.

pdeffebach:
Ah sorry! I swear I did `rebase` in gitkraken. I think it asked me to do a pull first and that's where it got messed up. 

I have force pushed after undo-ing a few things.

bkamins:
Thanks - now we are clean with master 👍.

Going back to the PR there are two things API and the implementation:
* regarding the API I can accept what you propose as it is fully non-breaking and recently @nalimilan convinced me to slow down breaking things 😄 (although I would prefer an API more similar to `combine`)
* regarding the implementation - I would find it cleaner if we had `get_stat` function that would have two methods - one taking a `DataFrame` and a `Symbol` expecting a standard symbol and doing the predefined calculations, the second taking a `DataFrame` and a `Pair` and calculating a custom function; both these methods would return a pair: `(column_name, column_values)`; then `describe` in main loop would simply iterate through the vector of statistics and add more columns to the output data frame. Such design would be easier to reason about; also then you would could more easily handle the case when user requests adding several columns with the same name (BTW: what do we want to do then: throw error, overwrite or generate a surrogate column name as with `makeunique`?)

It would be also good to hear if @nalimilan approves the above before we move forward.

After we have agreed on the design then in implementation I would add:
* updated docstrings;
* an update in the documentation (manual);
* added tests.

Thank you for pushing this forward!

nalimilan:
> regarding the API I can accept what you propose as it is fully non-breaking and recently @nalimilan convinced me to slow down breaking things smile (although I would prefer an API more similar to `combine`)

I didn't say we should avoid breaking things. :-) On the contrary, that's essential if we want to stabilize the API soon. I just think we should keep deprecations for a long time when they don't hinder progress.

I'm fine with the implementation you suggest

pdeffebach:
@bkamins I should do this change in a separate PR, though, right? Or should I keep pushing ahead with this one?

bkamins:
I would use this PR. Now it does not have merge conflicts, so it should not be a problem to push a commit adding the features?
(if you as if the "houskeeping" things - like documentation - should be left for a later PR I would definitely recommend to do it in one shot - in the past we had some PRs that did not update such things and now we have holes in the package that need patching).

Thank you for your effort as `describe` becomes a swiss army knife with it 😄.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -375,29 +375,30 @@ describe(df, stats = [:min, :max])
...
-    if stats == :all
-        stats = allowed_fields
+StatsBase.describe(df::AbstractDataFrame) = describe(df, :mean, :min, :median, 
+                                                     :max, :nunique, :nmissing, 
+                                                     :eltype)
+function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair}...)
+   
+    predefined_funs = Symbol[s for s in stats if s isa Symbol] 
+    if :all in predefined_funs
+        predefined_funs = allowed_fields
```
bkamins: where do you define `allowed_fields`?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -375,29 +375,30 @@ describe(df, stats = [:min, :max])
...
-                      :max, :nunique, :nmissing, :first, :last, :eltype]
-    if stats == :all
-        stats = allowed_fields
+StatsBase.describe(df::AbstractDataFrame) = describe(df, :mean, :min, :median, 
+                                                     :max, :nunique, :nmissing, 
+                                                     :eltype)
+function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair}...)
+   
+    predefined_funs = Symbol[s for s in stats if s isa Symbol] 
+    if :all in predefined_funs
```
bkamins: I think that when passing `:all` we should assume that it is the only `Symbol`-type of argument.

src/abstractdataframe/abstractdataframe.jl
bkamins: ah - later, but it has to be moved up I think.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -375,29 +375,30 @@ describe(df, stats = [:min, :max])
...
-                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
-    # Check that people don't specify the wrong fields.
-    allowed_fields = [:mean, :std, :min, :q25, :median, :q75,
-                      :max, :nunique, :nmissing, :first, :last, :eltype]
-    if stats == :all
-        stats = allowed_fields
+StatsBase.describe(df::AbstractDataFrame) = describe(df, :mean, :min, :median, 
+                                                     :max, :nunique, :nmissing, 
+                                                     :eltype)
+function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair}...)
```
bkamins: Probably it would be better to write `Union{Symbol, Pair{Symbol}}`?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -375,29 +375,30 @@ describe(df, stats = [:min, :max])
...
-            allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
-            throw(ArgumentError(":$stats not allowed." * allowed_msg))
-        else
-            stats = [stats]
-        end
-    end
+    custom_funs = Pair[s for s in stats if s isa Pair]
+
+    # Get the names in the order they appear
+    ordered_names = [stat isa Symbol ? stat : stat[1] for stat in stats]
```
bkamins: this should be fixed to handle `:all` case.

bkamins: also it would be good to make sure that `allunique(ordered_names)` is `true` as it will simplify the thinking about the code later simpler (and I guess we can assume that they must be unique).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -407,31 +408,36 @@ function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractV
...
     end
 
-    for stat in stats
+    for stat in ordered_names
         # for each statistic, loop through the columns array to find values
         # letting the comprehension choose the appropriate type
         data[stat] = [column_stats_dict[stat] for column_stats_dict in column_stats_dicts]
     end
+
+    # re-order columns according to the names from above
```
bkamins: this comment is not needed here I think.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Also note my previous remark about the implementation to avoid mostly useless specialization (second paragraph): https://github.com/JuliaData/DataFrames.jl/pull/1664#issuecomment-452073893

pdeffebach: does this mean I should add another function barrier `_describe` that is `_describe(df::AbstractDataFrame, Vector{<:Union{Symbol, Pair{Symbol}})`?

src/abstractdataframe/abstractdataframe.jl
pdeffebach: I added a check and now throw an error.

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
nalimilan: Yes.

pdeffebach:
Thanks for the comments! Making a function wrapper also means I get to abandon overly-specific type signatures, which is good. 

I updated the tests (and simplified them a bit). Now I need to update the docs. 

One API question: What do we want for `describe(df, :newcol => fun)`. Should we add it to our default list of columns or should we just return a DataFrame with a `:fun` column?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -375,29 +375,40 @@ describe(df, stats = [:min, :max])
...
-    if stats == :all
-        stats = allowed_fields
-    end
 
-    if stats isa Symbol
-        if !(stats in allowed_fields)
+    if predefined_funs == [:all]
+        predefined_funs = allowed_fields
+        i = findfirst(s -> s == :all, stats)
+        splice!(stats, i, allowed_fields) # insert in the stats vector to get a good order
```
pdeffebach: I need to also add an informative error when the user asks for `describe(df, :mean, :all)`.

bkamins: But this will be handled below. I have analyzed this case when reading the code 😄. Of course more informative message is welcome as now one would learn that `:all` is not allowed (and key is to describe all the options precisely in the docs).

If we are at it adding some failing tests would not hurt either.

pdeffebach:
Edited docs! and added a `@test_throws` error for the case of `describe(df, :all, :mean)`.

Let me know how I can improve the docs some more!

bkamins:
We have to to move old `describe` to deprecated.jl and keep it there for some time (and make sure there is no clash of methods).

bkamins:
I mean the version with a keyword argument.

pdeffebach:
I added a deprecation, but it's an awkward one, checking if a keyword argument is equal to `nothing`. 

Unfortunately, this method of deprecating `describe(df, stats = ...)` means that any stacktrace will go back to `deprecated` file, and not where the behavior is actually defined.

bkamins:
Looking at the code this seems temporarily unavoidable.

pdeffebach:
I think this is ready to be merged!

bkamins:
Looks good. @nalimilan - OK?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -334,16 +334,20 @@ end
...
 
 **Arguments**
 
-* `df` : the AbstractDataFrame
-* `stats::Union{Symbol,AbstractVector{Symbol}}` : the summary statistics to report. If
-  a vector, allowed fields are `:mean`, `:std`, `:min`, `:q25`, `:median`,
-  `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and `:nmissing`. If set to
-  `:all`, all summary statistics are reported.
+* `df` : the `AbstractDataFrame`
+* stats::Union{Symbol, Pair{Symbol}}... : the summary statistics to report.
```
```suggestion
* `stats::Union{Symbol, Pair{Symbol}}...` : the summary statistics to report. 
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -334,16 +334,20 @@ end
...
-* `df` : the AbstractDataFrame
-* `stats::Union{Symbol,AbstractVector{Symbol}}` : the summary statistics to report. If
-  a vector, allowed fields are `:mean`, `:std`, `:min`, `:q25`, `:median`,
-  `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and `:nmissing`. If set to
-  `:all`, all summary statistics are reported.
+* `df` : the `AbstractDataFrame`
+* stats::Union{Symbol, Pair{Symbol}}... : the summary statistics to report. 
+    * Arguments can be symbols from the following: `:mean`, `:std`, `:min`, `:q25`, 
+      `:median`, `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and 
+      `:nmissing`. The default statistics used when no `Symbol`s or `Pair`s are provided
```
```suggestion
      `:nmissing`. The default statistics used
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -334,16 +334,20 @@ end
...
 **Arguments**
 
-* `df` : the AbstractDataFrame
-* `stats::Union{Symbol,AbstractVector{Symbol}}` : the summary statistics to report. If
-  a vector, allowed fields are `:mean`, `:std`, `:min`, `:q25`, `:median`,
-  `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and `:nmissing`. If set to
-  `:all`, all summary statistics are reported.
+* `df` : the `AbstractDataFrame`
+* stats::Union{Symbol, Pair{Symbol}}... : the summary statistics to report. 
+    * Arguments can be symbols from the following: `:mean`, `:std`, `:min`, `:q25`,
```
nalimilan: "Arguments can be" should be moved outside of the list as it applies to all bullet points. Then you can go straight to the point for each type.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -334,16 +334,20 @@ end
...
-  `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and `:nmissing`. If set to
-  `:all`, all summary statistics are reported.
+* `df` : the `AbstractDataFrame`
+* stats::Union{Symbol, Pair{Symbol}}... : the summary statistics to report. 
+    * Arguments can be symbols from the following: `:mean`, `:std`, `:min`, `:q25`, 
+      `:median`, `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and 
+      `:nmissing`. The default statistics used when no `Symbol`s or `Pair`s are provided 
+      are `:mean`, `:min`, `:median`, `:max`, `:nunique`, `:nmissing`, and `:eltype`.
+    * Alternatively, specify `:all` as the only `Symbol` argument to return all statistics. 
+    * Finally, users can provide their own functions in the form of a `Pair{Symbol, Any}`.
```
nalimilan: `Pair{Symbol, Any}`should be `Pair{Symbol, <:Any}`. But better use something more readable like ``"`name => function` pairs, with `name` a symbol"``.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
 Consequently, `nmissing = 0` indicates that the column allows
 missing values, but does not currently contain any.
 
+Custom functions perform call `skipmissing` on columns of eltype `Union{T, Missing}`,
```
nalimilan: I suggest something like this:
```julia
If custom functions are provided, they are called repeatedly with the vector corresponding
to each column as the only argument. For columns allowing for missing values,
the vector is wrapped in a call to [`skipmissing`](@ref): custom functions must therefore
support such objects (and not only vectors), and cannot access missing values.
```

The last sentence isn't true right now since we allocate a vector without missing values, but we could avoid doing that in the future when neither the median nor quantiles need to be computed. That would be much faster.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
...
+
 **Examples**
 
 ```julia
 df = DataFrame(i = 1:10, x = rand(10), y = rand(["a", "b", "c"], 10))
 describe(df)
 describe(df, stats = :all)
-describe(df, stats = [:min, :max])
+describe(df, :min, :max)
+describe(df, :mean, :variance => Statistics.var)
```
nalimilan: Need `using Statistics` above. But better use something else like `sum` for simplicity.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
...
 ```
 
 """
-function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector{Symbol}} =
-                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
-    # Check that people don't specify the wrong fields.
+StatsBase.describe(df::AbstractDataFrame) = _describe(df, [:mean, :min, :median, 
+                                                     :max, :nunique, :nmissing,
```
nalimilan: Align with array. Better add a line break after `=`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
...
 ```
 
 """
-function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector{Symbol}} =
-                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
-    # Check that people don't specify the wrong fields.
+StatsBase.describe(df::AbstractDataFrame) = _describe(df, [:mean, :min, :median, 
+                                                     :max, :nunique, :nmissing, 
+                                                     :eltype])
+function StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...)
```
nalimilan: Short function form seems appropriate here.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
...
+    _describe(df, collect(stats))
+end
+
+function _describe(df::AbstractDataFrame, stats::Vector)
```
```suggestion
function _describe(df::AbstractDataFrame, stats::AbstractVector)
```

src/abstractdataframe/abstractdataframe.jl
nalimilan: Is this fixed now?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
...
-    end
 
-    if stats isa Symbol
-        if !(stats in allowed_fields)
+    if predefined_funs == [:all]
+        predefined_funs = allowed_fields
+        i = findfirst(s -> s == :all, stats)
+        splice!(stats, i, allowed_fields) # insert in the stats vector to get a good order
+    elseif :all in predefined_funs 
+            throw(ArgumentError("If the user specifies `:all` it must be the only `Symbol` argument."))
```
nalimilan: Avoid talking about "the user" to the user. :-) Anyway, avoid any unnecessary text.
```suggestion
            throw(ArgumentError("`:all` must be the only `Symbol` argument.")) 
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
...
+    elseif :all in predefined_funs 
+            throw(ArgumentError("If the user specifies `:all` it must be the only `Symbol` argument.")) 
+    else 
+        if !issubset(predefined_funs, allowed_fields)
+            not_allowed = setdiff(predefined_funs, allowed_fields)
             allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
-            throw(ArgumentError(":$stats not allowed." * allowed_msg))
-        else
-            stats = [stats]
+            # todo: fix the printing of this
```
nalimilan: ?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +369,58 @@ If the column does not allow missing values, `nothing` is returned.
...
-        disallowed_fields = setdiff(stats, allowed_fields)
-        allowed_msg = "\nAllowed fields are: :" * join(allowed_fields, ", :")
-        not_allowed = "Field(s) not allowed: :" * join(disallowed_fields, ", :") * "."
-        throw(ArgumentError(not_allowed * allowed_msg))
+    custom_funs = Pair[s for s in stats if s isa Pair]
+
+    ordered_names = [s isa Symbol ? s : s[1] for s in stats]
+    
+    if !allunique(ordered_names)
+        d = StatsBase.countmap(ordered_names)
```
nalimilan: This is really a heavy solution. `setdiff(ordered_names, unique(ordered_names))` should be enough.

test/dataframe.jl
```diff
@@ -424,47 +424,45 @@ end
                    dates  = Date.([2000, 2001, 2003, 2004]),
                    catarray = CategoricalArray([1,2,1,2]))
 
-    describe_output = DataFrame(variable = [:number, :number_missing, :string,
-                                            :string_missing, :dates, :catarray],
+    describe_output = DataFrame(variable = [:number, :number_missing, :string,
```
```suggestion
    describe_output = DataFrame(variable = [:number, :number_missing, :string,
```

test/dataframe.jl
```diff
@@ -424,47 +424,45 @@ end
                    dates  = Date.([2000, 2001, 2003, 2004]),
                    catarray = CategoricalArray([1,2,1,2]))
 
-    describe_output = DataFrame(variable = [:number, :number_missing, :string,
-                                            :string_missing, :dates, :catarray],
+    describe_output = DataFrame(variable = [:number, :number_missing, :string, 
+                                :string_missing, :dates, :catarray],
```
nalimilan: Incorrect alignment.

src/abstractdataframe/abstractdataframe.jl
bkamins: I was thinking about this earlier - performance wise it is ~ the same 😄. But your solution is more readable.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: yes this is fixed.

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
pdeffebach: changed so there is no need for `using Statistics`

src/abstractdataframe/abstractdataframe.jl
pdeffebach: fixed (over 80 chars but its okay).

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
pdeffebach: better printing.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: does this work? I don't think it does. I improved the printing though.

test/dataframe.jl
pdeffebach: fixed

pdeffebach:
Responded to all these comments, and fixed last printing issues. 

I think this is ready to be merged.

src/abstractdataframe/abstractdataframe.jl
nalimilan: OK it doesn't work. What we do currently in `names!` is `unique(nms[nonunique(DataFrame(nms=nms))])`. The reason why I'd rather avoid using `countmap` is that StatsBase is supposed to be moved to Statistics or other packages, so it may be deprecated at some point.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -407,31 +431,35 @@ function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractV
...
             d[:nmissing] = eltype(col) >: Missing ? count(ismissing, col) : nothing
         end
 
-        if :first in stats
+        if :first in predefined_funs 
             d[:first] = isempty(col) ? nothing : first(col)
         end
-
-        if :last in stats
+
```
```suggestion

```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -480,6 +508,12 @@ function get_stats(col::AbstractVector, stats::AbstractVector{Symbol})
     return d
 end
 
+function get_stats!(d::Dict, col::AbstractVector, stats::AbstractVector{Pair})
```
```suggestion
function get_stats!(d::Dict, col::AbstractVector, stats::AbstractVector{<:Pair})
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +371,57 @@ If the column does not allow missing values, `nothing` is returned.
...
+describe(df, :all)
+describe(df, :min, :max)
+describe(df, :min, :sum => sum)
 ```
 
 """
-function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector{Symbol}} =
-                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
-    # Check that people don't specify the wrong fields.
+StatsBase.describe(df::AbstractDataFrame) = _describe(df, [:mean, :min, :median,
```
nalimilan: Better put a line break here so that the array is on a single line. That will also fix the indentation below.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +371,57 @@ If the column does not allow missing values, `nothing` is returned.
...
 ```
 
 """
-function StatsBase.describe(df::AbstractDataFrame; stats::Union{Symbol,AbstractVector{Symbol}} =
-                            [:mean, :min, :median, :max, :nunique, :nmissing, :eltype])
-    # Check that people don't specify the wrong fields.
+StatsBase.describe(df::AbstractDataFrame) = _describe(df, [:mean, :min, :median, 
+                                                           :max, :nunique, :nmissing, 
+                                                            :eltype])
+StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) = _describe(df, collect(stats))
```
```suggestion
StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) =
    _describe(df, collect(stats))
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,39 +371,57 @@ If the column does not allow missing values, `nothing` is returned.
...
-    if stats isa Symbol
-        if !(stats in allowed_fields)
+    if predefined_funs == [:all]
+        predefined_funs = allowed_fields
+        i = findfirst(s -> s == :all, stats)
+        splice!(stats, i, allowed_fields) # insert in the stats vector to get a good order
+    elseif :all in predefined_funs 
+            throw(ArgumentError("`:all` must be the only `Symbol` argument.")) 
+    else 
+        if !issubset(predefined_funs, allowed_fields)
```
nalimilan: Make this an `elseif`.

src/deprecated.jl
```diff
@@ -1325,8 +1325,22 @@ end
...
-@deprecate showcols(df::AbstractDataFrame, all::Bool=false, values::Bool=true) describe(df, stats = [:eltype, :nmissing, :first, :last])
-@deprecate showcols(io::IO, df::AbstractDataFrame, all::Bool=false, values::Bool=true) show(io, describe(df, stats = [:eltype, :nmissing, :first, :last]), all)
+@deprecate showcols(df::AbstractDataFrame, all::Bool=false, values::Bool=true) describe(df, :eltype, :nmissing, :first, :last)
+@deprecate showcols(io::IO, df::AbstractDataFrame, all::Bool=false, values::Bool=true) show(io, describe(df, :eltype, :nmissing, :first, :last), all)
+function StatsBase.describe(df::AbstractDataFrame; stats=nothing)
+    if stats === nothing 
+        _describe(df, [:mean, :min, :median, 
+                       :max, :nunique, :nmissing, 
+                       :eltype])
+    elseif stats === :all
```
nalimilan: This branch should also print the deprecation AFAICT.

bkamins:
In line https://github.com/JuliaData/DataFrames.jl/pull/1664/files#diff-6108c932ccdff77aea536b99a80d4a79R493
you have to write `d[:std] = try std(col) catch end`
reusing `m` leads to the problem in #1730. Could you please change it?

bkamins:
Probably a better fix after thinking about it is to move:
```
if :std in stats
    d[:std] = try std(col, mean = m) catch end
end
```
inside the earlier `if` where we calculate the mean. Then we can reuse precalculated `m`.

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
pdeffebach: used this and fixed printing.

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/deprecated.jl
pdeffebach: fixed.

pdeffebach:
Fixed all the above comments, and changed the `if :std` to prevent the segfault

src/deprecated.jl
```diff
@@ -1325,8 +1325,23 @@ end
 import Base: vcat
 @deprecate vcat(x::Vector{<:AbstractDataFrame}) vcat(x...)
 
-@deprecate showcols(df::AbstractDataFrame, all::Bool=false, values::Bool=true) describe(df, stats = [:eltype, :nmissing, :first, :last])
-@deprecate showcols(io::IO, df::AbstractDataFrame, all::Bool=false, values::Bool=true) show(io, describe(df, stats = [:eltype, :nmissing, :first, :last]), all)
+@deprecate showcols(df::AbstractDataFrame, all::Bool=false, values::Bool=true) describe(df, :eltype, :nmissing, :first, :last)
+@deprecate showcols(io::IO, df::AbstractDataFrame, all::Bool=false, values::Bool=true) show(io, describe(df, :eltype, :nmissing, :first, :last), all)
+function StatsBase.describe(df::AbstractDataFrame; stats=nothing)
```
nalimilan: Doesn't this function overwrite the one defined above, printing a warning? If so, the other one should be commented out for now.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -334,16 +334,22 @@ end
...
-  `:all`, all summary statistics are reported.
+* `df` : the `AbstractDataFrame`
+* `stats::Union{Symbol, Pair{Symbol}}...` : the summary statistics to report. 
+  Arguments can be: 
+    *  A symbol from the list `:mean`, `:std`, `:min`, `:q25`, 
+      `:median`, `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and 
+      `:nmissing`. The default statistics used
+      are `:mean`, `:min`, `:median`, `:max`, `:nunique`, `:nmissing`, and `:eltype`.
+    * `:all` as the only `Symbol` argument to return all statistics. 
+    * Finally, users can provide their own functions in the form of a
```
nalimilan: "a ... pairs"

I'd also remove "users can provide their own functions": it breaks the logic of the bullets and it's quite verbose. Just say that `function` is a custom function. Also for `name` it could be useful to say that it's the name of the resulting column.

src/deprecated.jl
bkamins: It does overwrite, but I thought that the warning was not printed. But I think that commenting it out for now with TODO note is best.

pdeffebach: Sorry im a bit confused. What do I comment out and which TODO do I write?

src/abstractdataframe/abstractdataframe.jl

src/deprecated.jl
bkamins: Comment out the overwritten function and above it write TODO to remember to uncomment it when the deprecation period is finished. Have a look at abstractdataframe/iteration.jl for an example  (nothing is commented out there but this is the idea). When we have such TODO notes it is easy to grep the whole repository to find things that need fixing.

bkamins: by "overwritten function" I mean "overwritten method definition" to be precise.

pdeffebach: Done. I think I did this right but let me know.

bkamins: Thank you 👍. This is what will allow to avoid the problem @nalimilan mentioned above.

pdeffebach:
This should be ready for merging. Then we can fix that odd `unreachable reached` error.

bkamins:
Thank you. I will wait for @nalimilan to merge, as he has looked more into this PR 😄.

PR #1667, opened 2019-01-05
improve show for DataFrameRow and GroupedDataFrame for all supported MIME

### LINKED ISSUE #1663: Improve printing of `DataFrameRow`
To-Do:
Now as `DataFrameRow` type will be a more common type we should improve its printing in REPL, HTML and LaTeX.

bkamins:
Fixes #1663.
I reuse `AbstractDataFrame` `show` to display `DataFrameRow`, but make sure that we show a correct index of the row in the parent in `:Row` column and print `DataFrameRow` header where applicable.


bkamins:
I also have added `GroupedDataFrame` support.

bkamins:
@juliohm & @pdeffebach:
you might want to have a look at this PR if the changes to how `DataFrameRow` and `GroupedDataFrame` are shown follow what you would expect. Thanks!

pdeffebach:
I think this looks good!

src/abstractdataframe/io.jl
```diff
@@ -92,7 +92,14 @@ function html_escape(cell::AbstractString)
     return cell
 end
 
-function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary::Bool=true)
+Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame; summary::Bool=true) =
+    _show(io, mime, df, summary=summary)
+
+function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary::Bool=true, rowid=nothing)
```
```suggestion
function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary::Bool=true, rowid::Union{Int,Nothing}=nothing)
```

src/abstractdataframe/io.jl
```diff
@@ -92,7 +92,14 @@ function html_escape(cell::AbstractString)
...
+    n = size(df, 1)
+    if !(rowid isa Nothing)
```
nalimilan: A bit nicer:
```suggestion
    if rowid !== nothing
```

src/abstractdataframe/io.jl
```diff
@@ -215,12 +273,62 @@ function Base.show(io::IO, ::MIME"text/latex", df::AbstractDataFrame)
     write(io, "\\end{tabular}\n")
 end
 
+function Base.show(io::IO, mime::MIME"text/latex", dfr::DataFrameRow)
+    r, c = parentindices(dfr)
+    _show(io, mime, view(parent(dfr), [r], c), rowid=r)
+end
+
+function Base.show(io::IO, mime::MIME"text/latex", gd::GroupedDataFrame)
```
nalimilan: Maybe it would be possible to share code with other methods instead of repeating it? A small helper function taking `mime` could do the trick (for HTML, and maybe also the default output).

src/abstractdataframe/show.jl
```diff
@@ -234,7 +238,8 @@ end
                    rowindices::AbstractVector{Int},
                    maxwidths::Vector{Int},
                    leftcol::Int,
-                   rightcol::Int)
+                   rightcol::Int,
+                   rowid)
```
```suggestion
                   rowid::Union{Int,Nothing})
```

src/abstractdataframe/show.jl
```diff
@@ -251,6 +256,7 @@ required for printing have been precomputed.
   required to render each column.
 - `leftcol::Int`: The index of the first column in a chunk to be rendered.
 - `rightcol::Int`: The index of the last column in a chunk to be rendered.
+- `rowid`: Used internally to handle showing `DataFrameRow`
```
nalimilan: Everything is used internally here anyway. :-)

Also missing ending period.

src/abstractdataframe/show.jl
```diff
@@ -486,6 +500,43 @@ function showrows(io::IO,
     return
 end
 
+function _show(io::IO,
+               df::AbstractDataFrame;
+               allrows::Bool = !get(io, :limit, false),
+               allcols::Bool = !get(io, :limit, false),
+               splitcols = get(io, :limit, false),
+               rowlabel::Symbol = :Row,
+               summary::Bool = true, rowid=nothing)
```
nalimilan: Put argument on a new line for consistency (same elsewhere).

src/dataframerow/show.jl
```diff
@@ -1,7 +1,15 @@
...
-    @printf(io, "DataFrameRow (row %d)", row(r))
-    for (label, value) in pairs(r)
-        print(io, '\n', rpad(label, labelwidth, ' '), something(value, ""))
-    end
+function Base.show(io::IO, dfr::DataFrameRow;
+                   allcols::Bool = !get(io, :limit, false),
+                   splitcols = get(io, :limit, false),
+                   rowlabel::Symbol = :Row)
+    r, c = parentindices(dfr)
+    @printf(io, "DataFrameRow")
```
nalimilan:
```suggestion
    print(io, "DataFrameRow")
```

src/abstractdataframe/show.jl
```diff
@@ -537,45 +588,22 @@ julia> show(df, allcols=true)
 │ 3   │ 3     │ z      │
 ```
 """
-function Base.show(io::IO,
+Base.show(io::IO,
                    df::AbstractDataFrame;
```
nalimilan: Align arguments.

src/dataframerow/show.jl
```diff
@@ -1,7 +1,15 @@
...
+    _show(io, view(parent(dfr), [r], c), allcols=allcols, splitcols=splitcols,
+          rowlabel=rowlabel, summary=false, rowid=r)
 end
+
+Base.show(dfr::DataFrameRow;
+          allcols::Bool = !get(io, :limit, false),
```
nalimilan: We use different defaults for the corresponding `AbstractDataFrame` method.

test/dataframerow.jl
```diff
@@ -248,4 +243,39 @@ module TestDataFrameRow
         @test df[1, 1] == 200
         @test_throws KeyError dfr2.x2
     end
+
+    @testset "show" begin
+        df = DataFrame(a=nothing, b=1)
+        io = IOBuffer()
+        show(io, DataFrameRow(df, 1, :))
+        @test String(take!(io)) == "DataFrameRow\n│ Row │ a       │ b     │\n│     │ Nothing │ $(Int) │\n" *
```
nalimilan: `sprint` would be enough here, right? Also, a multiline string would be more readable.

src/abstractdataframe/io.jl
bkamins: Unfortunately they do not have exactly the same signature (for `"text/html"` we additionally allow a keyword argument).

I shared `DataFrameRow` mime for `"text/csv"` and `"text/tab-separated-values"` but it does not save that much.

src/abstractdataframe/show.jl
bkamins: Fixed (initially I exposed it, and later decided to hide the internals altogether)

src/dataframerow/show.jl
bkamins: right - fixed

src/abstractdataframe/io.jl
bkamins: Actually I have to keep separate methods as I get method ambiguity with methods defined in DelimitedFiles.jl (who would have thought).

bkamins:
Thank you for the review. I have fixed everything.

src/abstractdataframe/io.jl
nalimilan: But you could still call an internal function with shared code, right?

bkamins: The code is 90% similar, but 10% different - and the differences are scattered. I could try this, but it would contain if-else in many places as html and LaTeX control sequences have different semantics (I guess this is also the reason why original `show` functions for `AbstractDataFrame` although very similar are separate).

If you really insist I can do it, but I am not convinced that it is worth it.

src/abstractdataframe/io.jl
```diff
@@ -92,7 +92,14 @@ function html_escape(cell::AbstractString)
     return cell
 end
 
-function Base.show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary::Bool=true)
+Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame; summary::Bool=true) =
+    _show(io, mime, df, summary=summary)
+
+function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame; summary::Bool=true, rowid::Union{Int,Nothing}=nothing)
```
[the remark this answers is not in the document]
bkamins: I guess I should not blindly merge suggestions 😄. Anyway - you have prompted me to fix overlong line in the whole file + minor corrections with comparisions to `nothing`. Thanks

bkamins:
@nalimilan Strange things happen on nightly on Win and Linux (but not Mac).
Can you please have a look as I have no idea what might have caused this (essentially `sprint` prints bool `false` as `0` in the outputs).

And probably it is good that we caught this on an insignificant PR.

bkamins:
I have located the issue https://github.com/JuliaLang/julia/pull/30575.

What do you think we should do - keep printing `true`/`false` or be consistent with Base and start printing `0`/`1`?

nalimilan:
I'd say we should be consistent with Base (as everywhere else in printing code AFAIK).

pdeffebach:
It's also nice that we print the type at the top of columns, so `1` isn't ambiguous if its `Bool`

bkamins:
Given the discussion in https://github.com/JuliaLang/julia/pull/30575 I guess things are still being discussed and might change, so I will change the tests in DataFrames.jl (and regarding `Bool` we will print whatever Base produces in a given version of Julia anyway).

bkamins:
Given the changes I have also added column type information to LaTeX output as it becomes relevant now.

src/abstractdataframe/show.jl
```diff
@@ -350,6 +363,7 @@ NOTE: The value of `maxwidths[end]` must be the string width of
 - `displaysummary::Bool`: Should a brief string summary of the
   AbstractDataFrame be rendered to the I/O stream before printing the
   contents of the renderable rows? Defaults to `true`.
+- `rowid = nothing`: Used internally to handle showing `DataFrameRow`
```
```suggestion
- `rowid = nothing`: Used to handle showing `DataFrameRow`
```

src/abstractdataframe/show.jl
```diff
@@ -486,6 +503,44 @@ function showrows(io::IO,
...
+function _show(io::IO,
+               df::AbstractDataFrame;
+               allrows::Bool = !get(io, :limit, false),
+               allcols::Bool = !get(io, :limit, false),
+               splitcols = get(io, :limit, false),
+               rowlabel::Symbol = :Row,
+               summary::Bool = true,
+               rowid=nothing)
+    nrows = size(df, 1)
+    if !(rowid isa Nothing)
```
```suggestion
    if rowid !== nothing
```

PR #1672, opened 2019-01-09
Add bindrows(), which appends DataFrames regardless of column order

ryapric:
Hello! I've added a `bind_rows()` function to `src/dataframe.jl`, which takes an array of `DataFrames` and appends them to the first in the array. This is accomplished by iterating through the array, reordering columns, casting the proper types to `Union{Missing,eltype(c)}`, and filling the missing data with `missing`. I have also added appropriate unit tests to `test/dataframe.jl`.

I have read the contributing guidelines, and did my best to follow them, but there are a few lines that push past 79 characters. I did notice other definitions that weren't my own that did the same, however.

You will also notice that this PR adds a v1.0-canonical `Project.toml` file. I am developing on Julia v1.0.3, and in order to get tests to run cleanly, I needed to use the v1.0+ `Pkg`, or Julia fails due to "missing dependencies". If this file is undesired for some reason, please let me know, but I would like to know why, for my own knowledge. I have left the `REQUIRE` files untouched.

Please let me know if you have any questions. Thank you!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,34 @@
+name = "DataFrames"
+uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+repo = "https://github.com/JuliaData/DataFrames.jl.git"
+
+[deps]
+CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
+CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+DataStreams = "9a8bc11e-79be-5b39-94d7-1ccc349a1a85"
+IteratorInterfaceExtensions = "82899510-4779-5014-852e-03e436cf321d"
+Missings = "e1d29d7a-bbdc-5cf2-9ac0-f12de2c33e28"
+Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+SortingAlgorithms = "a2af1166-a08f-5f64-846c-94a0d3cef48c"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+TableTraits = "3783bdb8-4a98-5b6b-af9a-565f29a5fe9c"
+Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
+TranscodingStreams = "3bb67fe8-82b1-5028-8e26-92a6c54297fa"
+Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
+WeakRefStrings = "ea10d353-3f73-51f8-a26c-33c1cb351aa5"
+
+[extras]
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
+DataValues = "e7dc6d0d-1eca-5fa6-8ad6-5aecde8b7ea5"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
+LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


davidanthoff:
I had been thinking about something like that for [Query.jl](https://github.com/queryverse/Query.jl) as well, over [here](https://github.com/queryverse/Query.jl/issues/228). My main question had been whether this is a good name, or whether R just uses it because it has been around for a long time. I don't have an opinion yet, but if we add something like that to both packages, it might be nice to use a common name.

You should be able to run tests without a ``Project.toml``. If you run your tests via the ``test`` command at the Pkg REPL it should sort all of this out automatically. In my mind adding a Project.toml is not a good idea at this point. Things like CI, the automatic METADATA builds etc. then will use that instead of the REQUIRE file, and any mistakes in the REQUIRE file won't show up in CI anymore. That is not good, IMO, because it essentially means that CI is testing a different set of code/artifacts than what will determine whether things work on user machines.

ryapric:
Hi @davidanthoff, thanks for the reply!

I agree, the naming convention I chose was for consistency with `dplyr`, but I don't care what name is decided on. I do think it it should be consistent with *something*, though, for familiarity reasons; but something like `unionall()` (SQL-esque) doesn't feel right. Happy to continue that discussion on [the Query.jl issue](https://github.com/queryverse/Query.jl/issues/228), but since DataFrames.jl extends `append!()` already, I think it might fit better in this library.

Regarding a `Project.toml` file, migrating to Julia v1.0+, and CI concerns, here's what I'm seeing (and this is certainly a bigger discussion for outside of this PR):

Running `test` from the `Pkg` REPL, *or* calling `using Pkg; Pkg.test()` does *not* seem to work in *any* version of Julia 0.7+. The latest version of `Pkg` requires that a `name` field be specified in the `Project.toml` file, and errors out with 'Trying to test an unnamed project'. `Pkg` also seems to have no means of interacting with a `REQUIRE` file, either, as calling `Pkg.resolve()` does nothing without a `Project.toml` file present. I am able to replicate each of these concerns in three different Docker container images for Julia, and can provide shell scripts to demonstrate if requested.

Additionally: according to [Travis' page for Julia builds](https://docs.travis-ci.com/user/languages/julia/), the default build scripts executed should be as follows:

- If there exists a `Project.toml` file, AND the Julia version is >=0.7, then the command is:

        using Pkg
        Pkg.build()
        Pkg.test(coverage=true)

- Otherwise, it uses the "older form" (their words, not mine):

        if VERSION >= v"0.7.0-DEV.5183"
            using Pkg
        end
        Pkg.clone(pwd())
        Pkg.build("$pkgname")
        Pkg.test("$pkgname", coverage=true)

*All* of the Travis builds for this commit have failed, and the culprit seems to be the Travis configuration itself; ***I believe the build steps for Travis are not aligned with the documentation***.

For example, [this build](https://travis-ci.org/JuliaData/DataFrames.jl/jobs/477580849) uses Julia 1.0+, and so should be recognizing the `Project.toml` file and using the newer build steps. However, this is the command that is executed (line 537):

        julia --check-bounds=yes -e 'using Pkg; Pkg.clone(pwd()); Pkg.build("DataFrames"); Pkg.test("DataFrames"; coverage=true)'

This fails because Julia won't allow you to run `Pkg.clone()` on a project with the same UUID as what it tries to pull from the Registry (I am able to confirm this same error when running locally, and in three different versions of Julia in Docker containers). Additionally, `Pkg.clone()` throws deprecation warnings when called in Julia v0.7+.

My point being two things:

1. The Travis config seems to be documented one way, but configured another

1. The presence of a `Project.toml` file *should have no impact* on running builds against older versions of Julia; the commands to run each are not the same.

Again, these are issues I'll bring up separately with the Travis community maintainers, but worth pointing out here as it came up. It's of no great importance to me to remove the `Project.toml` file from this PR, but I do think that there are more systemic issues or concerns regarding *not* having one in registry packages, and the slow move to update due to the 0.7/1.0 release and the changes to `Pkg` is a bit concerning regarding best practices in contributing.

pdeffebach:
I have a working PR here: https://github.com/JuliaData/DataFrames.jl/pull/1659 that I started a few weeks ago. You should check that out.

ryapric:
Well color me frustrated for not checking the other open PRs first!

@pdeffebach forgive my inexperience here, but to help with that PR, should I fork _your_ fork, and work with that branch? This is a bit more branchy than I'm used to.

I'll close this PR here for now, and see what I can help with over there. But regarding the possible Travis issues, I'm still hoping to hear more on that here and [here](https://travis-ci.community/t/julia-builds-do-not-seem-to-be-configured-correctly-according-to-the-travis-documentation/1785).

PR #1674, opened 2019-01-13
Generalize optimized categorical grouping method to multiple keys

nalimilan:
The compiler is still able to generate efficient code specialized on the number of keys. Also require `groups !== nothing` for this method since we never call it without groups and it cannot compute the actual number of nonempty groups without doing all the work anyway.

```julia
df1 = DataFrame(key1=categorical(string.(rand(1:10, 500_000))),
                key2=categorical(string.(rand(1:10, 500_000))));
df2 = DataFrame(key1=categorical(string.(rand(1:1000, 500_000))),
                key2=categorical(string.(rand(1:1000, 500_000))));

# Master
julia> @btime groupby(df1, :key1);
  2.807 ms (66 allocations: 7.63 MiB)

julia> @btime groupby(df2, :key1);
  4.727 ms (68 allocations: 7.66 MiB)

julia> @btime groupby(df1, [:key1, :key2]);
  21.069 ms (499949 allocations: 23.08 MiB)

julia> @btime groupby(df2, [:key1, :key2]);
  57.993 ms (106682 allocations: 23.08 MiB)

# This PR
julia> @btime groupby(df1, :key1);
  2.324 ms (50 allocations: 7.63 MiB)

julia> @btime groupby(df2, :key1);
  4.241 ms (50 allocations: 7.66 MiB)

julia> @btime groupby(df1, [:key1, :key2]);
  3.381 ms (52 allocations: 7.63 MiB)

julia> @btime groupby(df2, [:key1, :key2]);
  38.733 ms (58 allocations: 22.23 MiB)
```

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -260,14 +260,20 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
     if skipmissing
         popfirst!(starts)
         popfirst!(stops)
+        groups .-= 1
         ngroups -= 1
     end
 
     # sort groups if row_group_slots hasn't already done that
     if sort && !sorted
         group_perm = sortperm(view(df, rperm[starts], :))
+        group_invperm = invperm(group_perm)
         permute!(starts, group_perm)
         Base.permute!!(stops, group_perm)
+        for i in eachindex(groups)
+            gix = groups[i]
+            groups[i] = gix == 0 ? 0 : group_invperm[gix]
+        end
     end
 
     return RowGroupDict(df, rhashes, gslots, groups, rperm, starts, stops)
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -106,6 +106,7 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
     @assert groups === nothing || length(groups) == length(cols[1])
     rhashes, missings = hashrows(cols, skipmissing)
     # inspired by Dict code from base cf. https://github.com/JuliaData/DataTables.jl/pull/17#discussion_r102481481
+    # but using open addressing with a table with as many slots as rows
     sz = Base._tablesz(length(rhashes))
     @assert sz >= length(rhashes)
     szm1 = sz-1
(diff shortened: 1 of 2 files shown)


src/dataframerow/utils.jl
```diff
@@ -145,50 +146,75 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+    # and this method needs to allocate a groups vector anyway
+    @assert groups !== nothing && all(col -> length(col) == length(groups), cols)
 
     # If missings are to be skipped, they will all go to group 1,
     # which will be removed by group_rows
-    ngroups = length(levels(col)) + (eltype(col) >: Missing)
+    ngroupstup = map(cols) do c
+        length(levels(c)) + (!skipmissing && eltype(c) >: Missing)
+    end
+    ngroups = prod(ngroupstup) + skipmissing
```
bkamins: why do you calculate it here and a well in line 171 but for `Int128`. We could do it only one time - right?

src/dataframerow/utils.jl
```diff
@@ -145,50 +146,75 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+
+    # Fall back to hashing if there would be too many empty combinations.
+    # The first check ensures the computation of ngroups did not overflow.
+    # The rationale for the 2 threshold is that while the fallback method is always slower,
+    # it allocates a hash table of size length(groups) instead of the remap vector
+    # of size ngroups (i.e. the number of possible combinations) in this method:
+    # so it makes sense to allocate more memory for better performance,
+    # but it needs to remain reasonable compared with the size of the data frame.
```
bkamins: is `2` a guess or you did the benchmarks (I do not insist to do them - just asking)

src/dataframerow/utils.jl
```diff
@@ -145,50 +146,75 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
         # so always sort groups in the order of levels
-        refmap = [0; CategoricalArrays.order(col.pool)]
-        seen = fill(false, length(refmap))
+        nlevels = length(levels(col))
+        refmap = Vector{Int}(undef, nlevels + 1)
+        refmap[1] = nlevels
+        refmap[2:end] .= CategoricalArrays.order(col.pool) .- 1
+        refmap
+    end
+    strides = (cumprod(collect(reverse(ngroupstup)))[end-1:-1:1]..., 1)::NTuple{N,Int}
```
bkamins: why do you need `collect` here? I guess it is not performance critical though.

src/dataframerow/utils.jl
```diff
@@ -145,50 +146,75 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+    # but it needs to remain reasonable compared with the size of the data frame.
+    if prod(Int128.(ngroupstup)) > typemax(Int) || ngroups > 2 * length(groups)
+        return invoke(row_group_slots,
+                      Tuple{Tuple{Vararg{AbstractVector}}, Val,
+                            Union{Vector{Int}, Nothing}, Bool},
+                      cols, hash, groups, skipmissing)
+    end
 
-    if groups !== nothing
+    seen = fill(false, ngroups)
```
bkamins: Have you checked if it would it be better to use `BitSet` here?.

src/dataframerow/utils.jl
nalimilan: Because we need it as an `Int` for performance in the loop, and as an `Int128` for the overflow check. I guess I could store the `Int128` version in a variable and convert it, but the cost is really negligible (there's one value per *key*).

src/dataframerow/utils.jl
nalimilan: I used the benchmark in the description, varying the number of groups, and it appears the optimized method is always faster (which is why I noted that in the comment). But of course it's not reasonable to allocate 20 or 100 times more memory just to be a bit faster (anyway this case should be pretty rare).

src/dataframerow/utils.jl
nalimilan: That's because `cumprod` doesn't accept tuples, unfortunately.

src/dataframerow/utils.jl
nalimilan: Using a `BitArray` would indeed reduce allocations from `2 * ngroups` to `ngroups + ngroups/8`. But I'm not sure it's a good trade-off since CategoricalArrays are optimized for small number of groups: probably better be as fast as possible, in particular in the most common case where the number of combinations is well below twice the number of rows.

src/dataframerow/utils.jl
nalimilan: OK. I've also tried it, and it's much slower. That sounds logical since a boolean indicator is really the simplest information we need to store here.

src/dataframerow/utils.jl

src/dataframerow/utils.jl
bkamins: OK given the discussion above

test/grouping.jl
```diff
@@ -210,10 +210,13 @@ module TestGrouping
...
-        # testing pool overflow
-        df2 = DataFrame(v1 = categorical(collect(1:1000)), v2 = categorical(fill(1, 1000)))
-        @test groupby(df2, [:v1, :v2]).starts == collect(1:1000)
-        @test groupby(df2, [:v2, :v1]).starts == collect(1:1000)
+        # test number of potential combinations higher than typemax(Int32)
+        N = 2000
+        df2 = DataFrame(v1 = levels!(categorical(rand(1:N, 100)), collect(1:N)),
+                        v2 = levels!(categorical(rand(1:N, 100)), collect(1:N)),
+                        v3 = levels!(categorical(rand(1:N, 100)), collect(1:N)))
+        df2b = mapcols(x -> convert(Vector{Int}, x), df2)
```
bkamins:
```suggestion
        df2b = mapcols(Vector{Int}, df2)
```

test/grouping.jl
```diff
@@ -210,10 +210,13 @@ module TestGrouping
...
+        @test groupby(df2, [:v1, :v2, :v3]) == groupby(df2b, [:v1, :v2, :v3])
```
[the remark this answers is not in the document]
nalimilan: Very good point, that actually uncovered a bug when sorting and when skipping missing introduced by https://github.com/JuliaData/DataFrames.jl/pull/1651. It's actually even more tricky, because the internal fields can be different, yet the result be equivalent (due to permutations). But AFAICT it's enough to check for equality and for equality of `groups` fields (which are redundant, and stored only for performance).

src/dataframerow/utils.jl
```diff
@@ -234,14 +260,20 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
...
         popfirst!(starts)
         popfirst!(stops)
+        groups .-= 1
         ngroups -= 1
     end
 
     # sort groups if row_group_slots hasn't already done that
     if sort && !sorted
         group_perm = sortperm(view(df, rperm[starts], :))
+        group_invperm = invperm(group_perm)
```
nalimilan: I'm very bad at permutations, but AFAICT that's the most efficient solution.

test/grouping.jl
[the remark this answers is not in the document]
nalimilan: Yes, they should be covered now.

src/dataframerow/utils.jl
nalimilan: Yeah, it's almost negligible compared with the time spent in grouping (this is only used for the slower fallback method).

test/grouping.jl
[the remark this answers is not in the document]
nalimilan: Actually I've found an even stricter way of checking this: each time we call `groupby`, ensure fields are consistent. Let me know what you think.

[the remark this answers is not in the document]
bkamins: Looks good. Thank you.

What I also meant that comparison of two `GroupedDataFrame`s is OK for the user (we compare values), but for the tests we should have a method that checks if actually we select exactly the same rows in both cases. Otherwise if you would do something like `groupby(DataFrame(a=[1,2,1,2], b=1), :a)` then all tests would pass even inf the `groupby` method were wrong (as `b` column is constant so it will always compare as `true` even if we take rows in an inconsistent order to `SubDataFrames`). Is it clear what I mean? The inconsistency would become apparent if we later written e.g. `b=[1,2,3,4]` in parent.

An alternative is to always use some `id` column in every `DataFrame` in our tests which has values from `1` to `nrows(df)` sorted. Then what we have now is guaranteed to catch such problems.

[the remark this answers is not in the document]
nalimilan: Indeed. After checking, it appears that all tests use unique (or random) values for a non-grouping column, so that should be OK.

PR #1682, opened 2019-01-16
Fix grouping column after map on GroupedDataFrame

nalimilan:
Column indices referred to the original data frame but did not necessarily point to the right column in the new parent. Add tests to cover this situation and more.

(Hide whitespace changes to review.)

### DIFF
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -105,35 +105,39 @@ module TestGrouping
         Random.seed!(1)
         df = DataFrame(a = repeat(Union{Int, Missing}[1, 3, 2, 4], outer=[2]),
                        b = repeat(Union{Int, Missing}[2, 1], outer=[4]),
-                       c = Vector{Union{Float64, Missing}}(randn(8)))
+                       c = repeat([0, 1], outer=[4]),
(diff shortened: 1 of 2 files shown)


test/grouping.jl
```diff
@@ -118,112 +116,118 @@ module TestGrouping
         f7(df) = (c2 = df[:c].^2,)
         f8(df) = DataFrame(c2 = df[:c].^2)
 
-        res = unique(df[cols])
-        res.cmax = [maximum(df[(df.a .== a) .& (df.b .== b), :c])
+        for cols in ([:a, :b], [:b, :a], [1, 2], [2, 1], [true, true, false])
```
bkamins: it would be good to add a test also when grouping columns are e.g. column 5 and 3 or something similar (not two first columns). I have checked it manually and it is now OK, but I did not find this covered in tests.

nalimilan: Indeed. I've added a commit. These tests are becoming more complex than the actual code, but I guess there's no way around that. ;-)

bkamins: Thanks. Indeed we probably cannot avoid it, but it is excellent to have them.

PR #1684, opened 2019-01-17
improve show of GroupedDataFrame

bkamins:
Use `repr` to show grouping column values and improve escaping in HTML and LaTeX output.

### DIFF
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -168,7 +168,7 @@ function Base.show(io::IO, mime::MIME"text/html", gd::GroupedDataFrame)
     N = length(gd)
     keynames = names(gd.parent)[gd.cols]
     parent_names = names(gd.parent)
-    keys = join(':' .* string.(keynames), ", ")
+    keys = html_escape(join(':' .* string.(keynames), ", "))
     keystr = length(gd.cols) > 1 ? "keys" : "key"
     groupstr = N > 1 ? "groups" : "group"
     write(io, "<p><b>$(typeof(gd).name) with $N $groupstr based on $keystr: $keys</b></p>")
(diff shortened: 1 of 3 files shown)


src/abstractdataframe/io.jl
```diff
@@ -168,15 +168,15 @@ function Base.show(io::IO, mime::MIME"text/html", gd::GroupedDataFrame)
...
+    keys = html_escape(join(':' .* string.(keynames), ", "))
     keystr = length(gd.cols) > 1 ? "keys" : "key"
     groupstr = N > 1 ? "groups" : "group"
     write(io, "<p><b>$(typeof(gd).name) with $N $groupstr based on $keystr: $keys</b></p>")
     if N > 0
         nrows = size(gd[1], 1)
         rows = nrows > 1 ? "rows" : "row"
 
-        identified_groups = [':' * string(parent_names[col], " = ", first(gd[1][col]))
+        identified_groups = [html_escape(':' * string(parent_names[col], " = ", repr(first(gd[1][col]))))
```
nalimilan: While we're at it, I wonder whether we should drop the leading `:` (we don't show it in the column header). Cc: @pdeffebach

bkamins: I mildly lean towards dropping `:`, but I am OK with both.

bkamins:
A side question: do you know why AppVeyor shows that it has failed on nightly x64, while the log shows that the tests have passed?

src/abstractdataframe/io.jl
pdeffebach: Yeah I do, too. I think we should drop the `:` for consistency.

bkamins: OK - I will drop it in this PR.

Issue #1675, opened 2019-01-13
DataFrameRow conversion to DataFrame broken


alyst:
On Julia v1.0.3 and DataFrames 0.16
```
julia> using DataFrames

julia> df = DataFrame(a = [1, 2], b = [:a, :b])
2×2 DataFrame
│ Row │ a     │ b      │
│     │ Int64 │ Symbol │
├─────┼───────┼────────┤
│ 1   │ 1     │ a      │
│ 2   │ 2     │ b      │

julia> df[1, :]
DataFrameRow (row 1)
a  1
b  a

julia> DataFrame(df[1, :])
0×0 DataFrame
```

I think it was working (as naturally expected) in 0.15. One consequence of this is that `append!(df, df[1, :])` fails with "Column names do not match".

nalimilan:
Woops, looks like the fallback Tables constructor is used. We should define a special constructor for `DataFrameRow`.

But returning an empty data frame is really weird. I would expect either an error or something meaningful. What do you think @quinnj?

> One consequence of this is that `append!(df, df[1, :])` fails with "Column names do not match".

AFAIK that never worked (it should be `push!` anyway): see https://github.com/JuliaData/DataFrames.jl/pull/1439.

bkamins:
`DataFrameRow` should behave in exactly the same way as a `NamedTuple` of values. Therefore:
* one should use `push!` instead of `append!`;
* constructing a `DataFrame` from a `DataFrameRow` as of current should fail;

The issue is that `Base.IteratorEltype` returns a different value for `NamedTuple` and for `DataFrameRow`. As we probably should not change it now we should add a special handling for `DataFrameRow`.

We should first decide if we want `DataFrame((a=1, b=2))` to work (now it fails). Maybe @quinnj could comment here (the issue is that if it would work we would perform an implicit broadcasting which I think we should not do). Depending on the answer an appropriate implementation for `DataFrameRow` should be provided (so my "default" is that `DataFrame(::DataFrameRow)` should throw an error).

An idiomatic pattern that currently works is `DataFrame(pairs(dfr)...)` and the same with a `NamedTuple`: `DataFrame(pairs(nt)...)`. In this approach we perform an implicit broadcasting, but it does not introduce ambiguity because it calls a different constructor.

bkamins:
We could implement a more convenient replacement for `DataFrame(pairs(dfr)...)`, e.g. that `DataFrame` constructor accepts a generator/iterator of `Paris`.

[...]

PR #1685, opened 2019-01-17
Improve push! and DataFrame for DataFrameRow

bkamins:
Changes in this PR:
* add `push!` for `DataFrameRow`
* add `DataFrame` constructor for `DataFrameRow` (it is inconsistent with `NamedTuple`, but we cannot guarantee consistency here anyway so it does not hurt to allow it)
* finish deprecation period for `push!` with string keys; now we do not allow it
* small performance improvements in column lookup

This replaces https://github.com/JuliaData/DataFrames.jl/pull/1439 and fixes https://github.com/JuliaData/DataFrames.jl/issues/1675.

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -203,3 +203,39 @@ function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
     end
     return false
 end
+
+function DataFrame(dfr::DataFrameRow)
+    row, cols = parentindices(dfr)
+    parent(dfr)[row:row, cols]
+end
+
+pushhelper!(x, r) = push!(x, x[r])
+
+function Base.push!(df::DataFrame, dfr::DataFrameRow)
+    if parent(dfr) === df && index(dfr) isa Index
+        # in this case we are sure that all we do is safe
+        r = row(dfr)
+        for col in _columns(df)
+            pushhelper!(col, r)
+        end
+    else
+        # DataFrameRow can contain duplicate columns and we disallow this
+        # corner case when push!-ing
+        size(df, 2) == length(dfr) || throw(ArgumentError("Inconsistent number of columns"))
+        i = 1
+        for nm in _names(df)
+            try
+                push!(df[i], dfr[nm])
+            catch
+                #clean up partial dfr
+                for j in 1:(i - 1)
+                    pop!(df[j])
+                end
+                msg = "Error adding value to column :$nm."
+                throw(ArgumentError(msg))
+            end
+            i += 1
+        end
+    end
+    df
+end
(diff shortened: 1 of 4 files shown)


src/dataframerow/dataframerow.jl
```diff
@@ -203,3 +203,39 @@ function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
     end
     return false
 end
+
+function DataFrame(dfr::DataFrameRow)
+    row, cols = parentindices(dfr)
+    parent(dfr)[row:row, cols]
+end
+
+pushhelper!(x, r) = push!(x, x[r])
```
nalimilan: I don't understand the advantage of this helper. :-)

src/dataframerow/dataframerow.jl
```diff
@@ -203,3 +203,39 @@ function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
...
+    else
+        # DataFrameRow can contain duplicate columns and we disallow this
+        # corner case when push!-ing
+        size(df, 2) == length(dfr) || throw(ArgumentError("Inconsistent number of columns"))
+        i = 1
+        for nm in _names(df)
+            try
+                push!(df[i], dfr[nm])
+            catch
+                #clean up partial dfr
```
```suggestion
                #clean up partial row
```

src/dataframerow/dataframerow.jl
nalimilan: Have you checked it really makes a difference? Without `@noinline` I'd expect it to be inlined since it's short, which would defeat its purpose. Also a comment would be useful.

bkamins: I would say that the difference is relevant.

Here are the benchmarks:

my approach using a barrier function:
```
julia> df = DataFrame(rand(100, 100));

julia> dfr = df[10, :];

julia> @benchmark push!($df, $dfr)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.113 μs (0.00% GC)
  median time:      2.228 μs (0.00% GC)
  mean time:        2.676 μs (0.00% GC)
  maximum time:     48.664 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     9
```

the approach without a barrier function:
```
julia> df = DataFrame(rand(100, 100));

julia> dfr = df[10, :];

julia> @benchmark push!($df, $dfr)
BenchmarkTools.Trial:
  memory estimate:  1.56 KiB
  allocs estimate:  100
  --------------
  minimum time:     3.791 μs (0.00% GC)
  median time:      3.920 μs (0.00% GC)
  mean time:        4.704 μs (0.59% GC)
  maximum time:     61.044 μs (91.37% GC)
  --------------
  samples:          10000
  evals/sample:     8
```

bkamins: I have added a comment in the code.

nalimilan: OK. How about adding `@noinline` just to be sure, in case the compiler changes its strategy?

bkamins: Added. These compiler-level things with Julia are sometimes tricky 😄.

PR #1687, opened 2019-01-19
Remove duplicate tests for DataFrame equality

petershintech:
It seems there are residuals of the same code to test the equality of DataFrame.


petershintech:
Yes, I can see those lines are exactly same as the following lines.

bkamins:
Thank you!

PR #1688, opened 2019-01-20
adding `show()` option `missingstring`  to change the string that prints for `missing` values

ianshmean:
I've found that it's sometimes visually helpful to print a smaller string than "missing" for `missing` values. This provides a way to change the string during `show()`, if desired.

i.e. `show(df,missingstring="-")`

Further work is needed for non-default outputs (HTML, LaTeX, CSV and TSV)

```
> df = DataFrame(a=1:3,b=[1,missing,3])
3×2 DataFrame
│ Row │ a     │ b       │
│     │ Int64 │ Int64⍰  │
├─────┼───────┼─────────┤
│ 1   │ 1     │ 1       │
│ 2   │ 2     │ missing │
│ 3   │ 3     │ 3       │

> show(df,missingstring="-")
3×2 DataFrame
│ Row │ a     │ b       │
│     │ Int64 │ Int64⍰  │
├─────┼───────┼─────────┤
│ 1   │ 1     │ 1       │
│ 2   │ 2     │ -       │
│ 3   │ 3     │ 3       │
```

```
> df = DataFrame(a = repeat([1, 2, 3, 4], outer=[2]),b = repeat([2, 1], outer=[4]),c = rand([randn(),missing],8));
> gd = groupby(df, :a);
> show(gd,missingstring="-")
GroupedDataFrame with 4 groups based on key: a
First Group (2 rows): a = 1
│ Row │ a     │ b     │ c         │
│     │ Int64 │ Int64 │ Float64⍰  │
├─────┼───────┼───────┼───────────┤
│ 1   │ 1     │ 2     │ 0.0727342 │
│ 2   │ 1     │ 2     │ -         │
⋮
Last Group (2 rows): a = 4
│ Row │ a     │ b     │ c         │
│     │ Int64 │ Int64 │ Float64⍰  │
├─────┼───────┼───────┼───────────┤
│ 1   │ 4     │ 1     │ -         │
│ 2   │ 4     │ 1     │ 0.0727342 │
```


src/abstractdataframe/show.jl
```diff
@@ -53,7 +53,8 @@ end
     DataFrames.getmaxwidths(df::AbstractDataFrame,
                             rowindices1::AbstractVector{Int},
                             rowindices2::AbstractVector{Int},
-                            rowlabel::Symbol)
+                            rowlabel::Symbol,
+                            missingstring::String)
```
bkamins:
```suggestion
                            missingstring::AbstractString)
```

src/abstractdataframe/show.jl
```diff
@@ -78,6 +79,7 @@ implicit row ID column contained in every `AbstractDataFrame`.
 - `rowlabel::AbstractString`: The label that will be used when rendered the
   numeric ID's of each row. Typically, this will be set to "Row".
 
+- `missingstring::String = "missing"`: The string that is printed for `missing` values.
```
bkamins:
```suggestion
- `missingstring::AbstractString = "missing"`: The string that is printed for `missing` values.
```

src/abstractdataframe/show.jl
```diff
@@ -95,10 +97,12 @@ function getmaxwidths(df::AbstractDataFrame,
                       rowindices1::AbstractVector{Int},
                       rowindices2::AbstractVector{Int},
                       rowlabel::Symbol,
+                      missingstring::String,
```
bkamins:
```suggestion
                      missingstring::AbstractString,
```

src/abstractdataframe/show.jl
```diff
@@ -257,6 +265,7 @@ required for printing have been precomputed.
   required to render each column.
 - `leftcol::Int`: The index of the first column in a chunk to be rendered.
 - `rightcol::Int`: The index of the last column in a chunk to be rendered.
+- `missingstring::String = "missing"`: The string that is printed for `missing` values.
```
bkamins:
```suggestion
- `missingstring::AbstractString = "missing"`: The string that is printed for `missing` values.
```

src/abstractdataframe/show.jl
```diff
@@ -276,6 +285,7 @@ function showrowindices(io::IO,
                         maxwidths::Vector{Int},
                         leftcol::Int,
                         rightcol::Int,
+                        missingstring::String,
```
bkamins:
```suggestion
                        missingstring::AbstractString,
```

src/abstractdataframe/show.jl
```diff
@@ -363,6 +374,7 @@ NOTE: The value of `maxwidths[end]` must be the string width of
 - `displaysummary::Bool`: Should a brief string summary of the
   AbstractDataFrame be rendered to the I/O stream before printing the
   contents of the renderable rows? Defaults to `true`.
+- `missingstring::String = "missing"`: The string that is printed for `missing` values.
```
bkamins:
```suggestion
- `missingstring::AbstractString = "missing"`: The string that is printed for `missing` values.
```

src/abstractdataframe/show.jl
```diff
@@ -390,6 +402,7 @@ function showrows(io::IO,
                   allcols::Bool = false,
                   rowlabel::Symbol = :Row,
                   displaysummary::Bool = true,
+                  missingstring::String = "missing",
```
bkamins:
```suggestion
                  missingstring::AbstractString = "missing",
```

src/abstractdataframe/show.jl
```diff
@@ -510,6 +525,7 @@ function _show(io::IO,
                splitcols = get(io, :limit, false),
                rowlabel::Symbol = :Row,
                summary::Bool = true,
+               missingstring::String = "missing",
```
bkamins:
```suggestion
               missingstring::AbstractString = "missing",
```

src/abstractdataframe/show.jl
```diff
@@ -575,6 +592,7 @@ while `splitcols` defaults to `true`.
   By default this is the case only if `io` has the `IOContext` property `limit` set.
 - `rowlabel::Symbol = :Row`: The label to use for the column containing row numbers.
 - `summary::Bool = true`: Whether to print a brief string summary of the data frame.
+- `missingstring::String = "missing"`: The string that is printed for `missing` values.
```
bkamins:
```suggestion
- `missingstring::AbstractString = "missing"`: The string that is printed for `missing` values.
```

src/abstractdataframe/show.jl
```diff
@@ -598,16 +616,18 @@ Base.show(io::IO,
           allcols::Bool = !get(io, :limit, false),
           splitcols = get(io, :limit, false),
           rowlabel::Symbol = :Row,
-          summary::Bool = true) =
+          summary::Bool = true,
+          missingstring::String = "missing") =
```
bkamins:
```suggestion
          missingstring::AbstractString = "missing") =
```

src/abstractdataframe/show.jl
```diff
@@ -598,16 +616,18 @@ Base.show(io::IO,
...
+          rowlabel=rowlabel, summary=summary, missingstring=missingstring)
 
 Base.show(df::AbstractDataFrame;
           allrows::Bool = !get(stdout, :limit, true),
           allcols::Bool = !get(stdout, :limit, true),
           splitcols = get(stdout, :limit, true),
           rowlabel::Symbol = :Row,
-          summary::Bool = true) =
+          summary::Bool = true,
+          missingstring::String = "missing") =
```
bkamins:
```suggestion
          missingstring::AbstractString = "missing") =
```

src/groupeddataframe/show.jl
```diff
@@ -4,7 +4,8 @@ function Base.show(io::IO, gd::GroupedDataFrame;
                    allcols::Bool = !get(io, :limit, false),
                    splitcols::Bool = get(io, :limit, false),
                    rowlabel::Symbol = :Row,
-                   summary::Bool = true)
+                   summary::Bool = true,
+                   missingstring::String = "missing")
```
bkamins:
```suggestion
                   missingstring::AbstractString = "missing")
```

ianshmean:
Note that the default for latex is `missingstring=""` given that this was the default functionality beforehand

bkamins:
Thank you. Please let me know when the tests pass and I will review the changes.

Regarding LaTeX - please change the default to `"missing"` - this is one of these small inconsistencies we have that need to be cleaned up.

The reason to change it to `"missing"` is that now `nothing` and `missing` print the same in LaTeX which is not desirable and moreover HTML and LaTeX are inconsistent (and the most common use of LaTeX is PDF export from Jupyter notebook so they should be the same).

ianshmean:
@bkamins The latex default is now "missing" and the tests now pass, however I had to make changes to the MIME types that need review

src/abstractdataframe/io.jl
```diff
@@ -92,11 +92,11 @@ function html_escape(cell::AbstractString)
     return cell
 end
 
-Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame; summary::Bool=true) =
-    _show(io, mime, df, summary=summary)
+Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame; summary::Bool=true, missingstring::AbstractString="missing") =
```
bkamins: can you please correctly align the indentation of the code here and in all places below/

bkamins:
Looks good to me. I have three comments:

### MIME

> however I had to make changes to the MIME types that need review

I think we should provide methods performing auto-conversion like in Base where we hace (@nalimilan - do you have the same opinion here?):
```
show(io::IO, m::AbstractString, x) = show(io, MIME(m), x)
```

(of course with proper type restrictions; one has to be careful here to check that method ambiguities were not introduced)

In particular the PR should be checked on Jupyter notebook + HTML and LaTeX/PDF export to make sure that nothing fails (I can do it if would be a problem for you - please let me know) - the reason is that in the current state the PR is very likely to fail there (but I have not checked; however, it breaks backward compatibility without the fix I propose so there is a risk of failure)

### Doumentation

It would be excellent to add a small note in the documentation about the newly introduced features

### Rebasing

The test files require rebasing because today a major test code clean-up was performed. The major change was removal of indentation in `module` block. If you prefer that I do the rebasing then please let me know (in general you have to use the current upstream master and reapply only the changes you have introduced in the PR).

Thank you for your contribution (and sorry that the seemingly simple PR became more work than expected but we are trying hard to hit DataFrames.jl 1.0 release with a clean package).

ianshmean:
Thanks for the guidance @bkamins. I would prefer to not do the rebase, so I graciously accept the offer! :) thanks

test/io.jl
```diff
@@ -11,37 +11,55 @@ module TestIO
...
             \\end{tabular}
             """
-        @test repr(MIME("text/latex"), df) == str
     end
 
     @testset "Huge LaTeX export" begin
         df = DataFrame(a=1:1000)
         ioc = IOContext(IOBuffer(), :displaysize => (10, 10), :limit => false)
-        show(ioc, "text/latex", df)
+        show(ioc, MIME(Symbol("text/latex")), df)
```
nalimilan: `Symbol` isn't needed AFAICT.

nalimilan:
> > 
> 
> I think we should provide methods performing auto-conversion like in Base where we hace (@nalimilan - do you have the same opinion here?):
> 
> ```
> show(io::IO, m::AbstractString, x) = show(io, MIME(m), x)
> ```

Indeed the fallback method doesn't support keyword arguments. Something like this shouldn't introduce ambiguities:
```julia
show(io::IO, m::AbstractString, x::AbstractDataFrame; kwargs...) = show(io, MIME(m), x; kwargs...)
```

But we should file an issue or PR against Base to improve this.

test/io.jl
bkamins: My point was that it should be just `"text/latex"` if we fix the signatures.

bkamins:
> But we should file an issue or PR against Base to improve this.

I am not sure methods in base are assumed to accept keyword arguments. AFAICT the intended way to do it is to pass the arguments in `IOContext` object. I did not push on it here because we have this non-standard keyword arguments approach in `show` anyway. We might consider cleaning it up, but this is another PR and a big change.

My thinking was to define the method also you have proposed - at least short term and decide about `IOContext` issue later.

bkamins:
@ianshmean I start rebasing it so please do not push anything to this PR in the meantime 😄.

nalimilan:
It's not clear whether all of these options should go through `IOContext`. I think the idea is that it should only be used for settings which should apply recursively to sub-elements, which isn't the case here. Anyway the best way to find out is to discuss this in Base.

bkamins:
I am also not 100% clear 😄 - but I have checked that out of 270 methods for `show` in Base only one accepts one keyword argument - I do not remember now which, but it was not something typical.

bkamins:
@ianshmean + @nalimilan any opinion what we should do with this PR (note that it requires a serious rebasing since we did a heavy cleanup of `show` in the meantime)

nalimilan:
If @ianshmean can rebase it I think we should add the `show` method accepting keyword arguments.

bkamins:
I mark it non-breaking as I understand what is proposed here can be added at any time.

bkamins:
@ianshmean - do you think you would find time to work on this PR to try finalizing it?
@ronisbr - does PettyTables.jl have this option (if yes then probably we can close this PR and recommend PrettyTables.jl)

ianshmean:
@bkamins Sorry, I don't have time to complete this now

bkamins:
OK - thank you for the response and submitting the PR. If PrettyTables.jl has this possibility I will close this PR then.

ronisbr:
In PrettyTables, missing is shown as “missing”. However it is trivial to use a data formatted to change missing to “-“, for example.

bkamins:
Thank you for responding. This is what I have assumed, in this case I close it.

As a reference:
```
(v,i,j)->ismissing(v) ? "-" : v
```
can be written as:
```
(v,i,j) -> coalesce(v, "-")
```

PR #1689, opened 2019-01-20
DataFrame for GroupedDataFrame

bkamins:
This follows the discussion in https://github.com/JuliaData/DataFramesMeta.jl/pull/122.

@nalimilan In my opinion the current behavior of `combine(::GroupedDataFrame)` (duplicating grouping columns) is not very intuitive. Is it intentional or you plan to change it in the future?


nalimilan:
I agree that behavior doesn't make a lot of sense. Probably better change `combine` at the same time. Or deprecate it in favor of `DataFrame(gd)`, which will allow reintroducing it later with a new behavior.

nalimilan:
That's actually related to https://github.com/JuliaData/DataFrames.jl/issues/1460 and https://github.com/JuliaData/DataFrames.jl/pull/1555. I'd be inclined to overwrite the grouping keys with columns of the same name if the returned data frame contains such columns.

bkamins:
I moved the discussion on the combine behavior to https://github.com/JuliaData/DataFrames.jl/issues/1460. Here I think it is better to implement a more efficient constructor anyway which I will propose in the revision.

test/grouping.jl
```diff
@@ -1028,4 +1028,18 @@ module TestGrouping
...
             """
     end
+
+    @testset "DataFrame" begin
+        df = DataFrame(A = [missing, :A, :B, :A, :B, missing], B = 1:6)
+        gd = groupby_checked(df, :A)
+        @test sort(DataFrame(gd), :B) ≅ sort(df, :B)
+        @test eltypes(DataFrame(gd)) == eltypes(df)
+        gd = groupby_checked(df, :A, skipmissing=true)
+        @test sort(DataFrame(gd), :B) == sort(dropmissing(df, disallowmissing=false), :B)
```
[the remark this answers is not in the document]
bkamins: I want to test explicitly here if we retain exactly the same types. `Missings.T.(eltypes(df))` does not allow me to check if the type is exactly the same. But I can rewrite these tests if you prefer.

bkamins:
Apart from the better constructor I have patched some holes in internal API.

src/groupeddataframe/grouping.jl
```diff
@@ -1106,3 +1106,11 @@ function _aggregate(d::AbstractDataFrame, fs::AbstractVector,
     sort && sort!(res, headers)
     res
 end
+
+function DataFrames.DataFrame(gd::GroupedDataFrame)
+    length(gd) == 0 && return similar(parent(gd), 0)
+    # below we assume that gd.ends[end] == length(gd.idxs)
+    # and that gd.starts and gd.ends are increasing and cover a continuous range
+    gd.starts[1] == 1 && return parent(gd)[gd.idx, :]
```
nalimilan: Isn't it always the case that `gd.starts[1]`? AFAICT that's what we explicitly set in `group_rows`.

OTC, if I'm wrong, is it really useful to distinguish this case? It should be quite cheap to create a view, right?

bkamins: OK - I will make a `view` - I wanted to avoid it in the most common case.
`gd.starts[1]` is greater than `1` if we have missing values in grouping and we set `skipmissing` to `true`.

test/grouping.jl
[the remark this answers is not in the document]

src/groupeddataframe/grouping.jl
nalimilan: >  `gd.starts[1]` is greater than `1` if we have missing values in grouping and we set `skipmissing` to `true`.

Indeed. We should probably just do `starts .-= starts[2] .+ 1 and `ends .-= ends[2] .+ 1. We already do that for `groups`.

bkamins: But I understand you mean to do it in `groupby` related mechanics not here. Right? I think it would be OK, but then `gd.idx` should not include the leading indices that are pointing at `missing` so `length(gd.idx)` may be less that `nrow(parent(gd))` and the same for `gd.ends[end]` (but the relationsip `gd.ends[end] == length(gd.idxs)` would still hold) - I assume that we are not relying on the fact that `gd.idx` has the same number of elements as rows in the parent, but that you know this part of code better 😄.

If we took this approach then we could avoid making the view here and just index by `gd.idx`.

nalimilan: Right. I don't think we assume `length(gd.idx) == nrow(parent(gd))`, but only test will tell for sure.

bkamins: OK, but I guess it should be a separate PR. Are willing to work on it?

bkamins:
I have added the deprecation for `combine(::GroupedDataFrame)`. I leave it in the grouping.jl file, as later we should change it to throw an error.

src/groupeddataframe/grouping.jl
nalimilan: See https://github.com/JuliaData/DataFrames.jl/pull/1692. But actually when `sort=true` there's no guaranty that `gd.starts[1] == 1`, only that `minimum(gd.starts) == 1`. And using `gd.idx` directly would return groups in an unsorted order. AFAIK that's unavoidable since we only sort group indices (`gd.start` and `gd.stops`), which should be faster than sorting row indices (`gd.idx`) when there are few groups. Not sure whether sorting group indices would be significantly slower.

bkamins: OK - I have fixed it (that is why I have said that split-apply-combine is tricky 😄). We cannot avoid making a copy of `gd.idx` in this case unfortunately, but I have proposed the fastest method to get what we want I could think of.

The fix depends on #1692, so I will rebase this PR, when we merge #1692 (I could resize to make it independent, but I think it is better to follow the idea of code simplification that can be achieved with #1692).

src/groupeddataframe/grouping.jl
```diff
@@ -1106,3 +1108,16 @@ function _aggregate(d::AbstractDataFrame, fs::AbstractVector,
     sort && sort!(res, headers)
     res
 end
+
+function DataFrame(gd::GroupedDataFrame)
+    length(gd) == 0 && return similar(parent(gd), 0)
+    idx = similar(gd.idx)
+    doff = 1
+    for (s,e) in zip(gd.starts, gd.ends)
```
nalimilan: Would a combination of `Iterators.flatten` and `copyto!` work? Not a big deal, but I wonder whether there's a simpler way of doing this relatively common operation.

bkamins: It could be made simpler (and it was simpler in DataFramesMeta.jl) but the approach I propose should be faster. Also I do not see how `Iterators.flatten` would be faster. In the approach I proposed we can use the fact that we have a small number of continuous blocks of memory.

Did you have in mind some alternative efficient pattern?

src/groupeddataframe/grouping.jl
bkamins: In parallel I was thinking about the same and consider using `view` instead of `copyto!` or creating a correct `rperm` from the start - it would probably help but would complicate the design a bit. But in general I am OK to drop #1692.

If you decide to drop #1692  I can fix this PR not to rely on #1692 and be safe.

src/groupeddataframe/grouping.jl
nalimilan: I wouldn't expect it to be faster, just simpler. But I don't really care.

bkamins:
OK - I have pushed the version that does not depend on #1692 so it should be good for checking when you have time.

src/groupeddataframe/grouping.jl
```diff
@@ -1106,3 +1112,19 @@ function _aggregate(d::AbstractDataFrame, fs::AbstractVector,
...
+    doff = 1
+    min_start = gd.starts[1]
+    for (s,e) in zip(gd.starts, gd.ends)
+        n = e - s + 1
+        copyto!(idx, doff, gd.idx, s, n)
+        doff += n
+        min_start = min(min_start, s)
+    end
+    resize!(idx, doff - 1)
+    @assert doff == length(gd.idx) + 2 - min_start
```
nalimilan: It's kind of weird to have this assertion here. Why not put something equivalent in `groupby_checked` instead? That way we will check it much more systematically, and the code here will be simpler.

bkamins: An excellent point. I have included a combo of tests to `groupby_checked` that I think are also useful for the future readers to understand what we expect to have in `GroupedDataFrame` object.

test/grouping.jl
```diff
@@ -177,9 +202,7 @@ end
         # groupby() without groups sorting
         gd = groupby_checked(df, cols)
         @test names(parent(gd))[gd.cols] == colssym
-        @test sort(combine(identity, gd), colssym) ==
-            sort(combine(gd), colssym) ==
```
nalimilan: Why not test `DataFrame(gd)` here and below? Can't hurt.

test/grouping.jl
```diff
@@ -1030,4 +1053,29 @@ end
...
+              sort(dropmissing(df, disallowmissing=false), :B)
+        @test eltypes(DataFrame(gd)) == [Union{Missing, Symbol}, Int]
+    end
+
+     df = DataFrame(a=Int[], b=[], c=Union{Missing, String}[])
+    gd = groupby_checked(df, :a)
+    @test size(DataFrame(gd)) == size(df)
+    @test eltypes(DataFrame(gd)) == [Int, Any, Union{Missing, String}]
+
+     dfv = view(dfx, 1:0, :)
```
```suggestion
    dfv = view(dfx, 1:0, :)
```

test/grouping.jl
```diff
@@ -1030,4 +1053,29 @@ end
...
+        @test sort(DataFrame(gd), :B) ≅ sort(df, :B)
+        @test eltypes(DataFrame(gd)) == [Union{Missing, Symbol}, Int]
+
+         gd = groupby_checked(df, :A, skipmissing=true)
+        @test sort(DataFrame(gd), :B) ==
+              sort(dropmissing(df, disallowmissing=false), :B)
+        @test eltypes(DataFrame(gd)) == [Union{Missing, Symbol}, Int]
+    end
+
+     df = DataFrame(a=Int[], b=[], c=Union{Missing, String}[])
```
```suggestion
    df = DataFrame(a=Int[], b=[], c=Union{Missing, String}[])
```

test/grouping.jl
bkamins: I have added a test here (although it is not very clean unfortunately, because of the `combine` mechanics), but I agree it is good to have it here.

PR #1692, opened 2019-01-21
Ensure GroupedDataFrame starts at index 1

nalimilan:
Storing indices of a ghost group could be confusing and makes the code more complex later.

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -258,6 +258,14 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
 
     # drop group 1 which contains rows with missings in grouping columns
     if skipmissing
+        s = starts[2]
+        if s > 1
+            N = length(rperm) - s + 1
+            copyto!(rperm, 1, rperm, s, N)
+            resize!(rperm, N)
+            starts .-= s .- 1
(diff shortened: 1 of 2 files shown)


src/dataframerow/utils.jl
```diff
@@ -258,6 +258,14 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
 
     # drop group 1 which contains rows with missings in grouping columns
     if skipmissing
+        s = starts[2]
+        if s > 1
+            N = length(rperm) - s + 1
+            copyto!(rperm, 1, rperm, s, N)
```
bkamins: have you considered using a `view` instead of `copyto!`? I think it should be faster and use less memory.

bkamins:
We might also consider avoiding considering `missing` group starting from:
```
rperm = Vector{Int}(undef, length(groups))
```
- make this vector shorter from the start and drop first `starts` and `stops` if needed. This should be even a bit faster when there are `missing` values as we would do less copying.

bkamins:
As a second thought - it would probably slow down things a bit when we have a lot of small groups.

bkamins:
If you are happy with current #1689 I think you can close this PR (unless you want to clean it up to always start from 1, but as discussed it is probably not strictly necessary)

PR #1697, opened 2019-01-26
use _columns in similar

oxinabox:
Using `_columns` gets ride of 1 level of indirection.
My bencmarks show 10% speedup,
and 1 less allocation from doing this.

using `_columns` is safe, as they are them selves passed to `similar` so the originals will not be in danger od mutation

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -244,7 +244,7 @@ that is different than the number of rows present in `df`.
 """
 function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
     rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
-    DataFrame(AbstractVector[similar(x, rows) for x in columns(df)], copy(index(df)))
+    DataFrame(AbstractVector[similar(x, rows) for x in _columns(df)], copy(index(df)))
 end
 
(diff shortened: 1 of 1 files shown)


bkamins:
Thank you for the proposal. The problem is that `_columns` is not a part of `AbstractDataFrame` interface. It is an internal function guaranteed to be defined only for `DataFrame` (for example `SubDataFrame` does not define it).

What we should do is to define a separate constructor for `DataFrame` that will use `_columns`.

However, a deeper problem is that this indirection should be optimized out by the compiler (at least this is what I would have thought should happen - and this was the reason behind the current implementation). We only create a `struct` around a `DataFrame`.

bkamins:
OK - I think can see where the performance problem lies. `getindex` for `DataFrame` does not support `@inbounds` (something @nalimilan prompted to add some time ago).

So the ideal solution to this PR (if you would be willing to dig into this - otherwise I can work on it) is to properly annotate `getindex` related methods with `Base.@propagate_inbounds` (in abstractdataframe/iteration.jl and dataframe/dataframe.jl).

oxinabox:
Ok, fair. I am not going to have the time to look into it.

bkamins:
OK - you can close it and I will open a new PR with the proposed fix (if still `_columns` for `DataFrame` will be faster I will include this custom method).

PR #1698, opened 2019-01-27
Add an example to skip :variable and :value variable.

petershintech:
Add one more example to skip :variable and :value variables when using unstack().

### DIFF
--- a/docs/src/man/reshaping_and_pivoting.md
+++ b/docs/src/man/reshaping_and_pivoting.md
@@ -244,7 +244,7 @@ julia> first(widedf, 6)
 │ 6   │ setosa        │ 6     │ 1.7         │ 0.4        │ 5.4         │ 3.9        │
 ```
 
-You can even skip the `:variable` and `:value` variables and use:
+You can even skip passing the `:variable` and `:value` values as positional arguments, as they will be used by default, and write:
 ```jldoctest reshape
 julia> widedf = unstack(longdf);
 


docs/src/man/reshaping_and_pivoting.md
```diff
@@ -244,6 +244,23 @@ julia> first(widedf, 6)
 │ 6   │ setosa        │ 6     │ 1.7         │ 0.4        │ 5.4         │ 3.9        │
 ```
 
+You can even skip the `:variable` and `:value` variables and use:
```
bkamins: can you please add a comment what happens then (it is rewritten as `unstack(::DataFrame, :variable, :value)`)

petershintech: Not sure what you meant, but I showed the output in the following REPL result. Choosing :variable and :value as default variables is obvious to me but please feel free to add more description if you want.

bkamins: What I meant is that we are not skipping the variables but positional arguments in the call. Would this be better?
```suggestion
You can even skip passing the `:variable` and `:value` values as positional arguments, as they will be used by default, and write:
```

nalimilan:
Thanks. Just being curious: do you actually use that feature? I've always wondered whether people had data with these variable names...

petershintech:
> Thanks. Just being curious: do you actually use that feature? I've always wondered whether people had data with these variable names...

Sticking to the standard names, :variable and :value, and skipping the arguments is quite convenient to me.

bkamins:
In my experience standard names are used most often when you `stack` some data frame, then do some manipulations on the result and `unstack`.

docs/src/man/reshaping_and_pivoting.md
bkamins:
```suggestion
You can even skip passing the `:variable` and `:value` values as positional arguments, as they will be used by default, and write:
```

petershintech:
@bkamins I committed your change to this PR. Thanks for your review.

bkamins:
Merged. Thank you!

petershintech:
Thanks for the merging @bkamins

PR #1700, opened 2019-01-27
Properly propagate inbounds in getindex

bkamins:
A follow up to https://github.com/JuliaData/DataFrames.jl/pull/1697.

Apparently Julia had some problems with efficiency of comprehensions when code is not type stable (as we get columns of unknown types) so I had to use the loop.

Here are the benchmark results. This PR:
```

julia> df = DataFrame(rand(100, 100));

julia> function s2(df::AbstractDataFrame, rows::Integer = size(df, 1))
           rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
           DataFrame(AbstractVector[similar(x, rows) for x in DataFrames._columns(df)], copy(DataFrames.index(df)))
       end
s2 (generic function with 2 methods)

julia> f1(df) = [x for x in eachcol(df, false)]

julia> f2(df) = [x for x in DataFrames._columns(df)]

julia> @btime similar($df, 0);
  7.068 μs (114 allocations: 15.31 KiB)

julia> @btime s2($df, 0);
  7.025 μs (114 allocations: 15.31 KiB)

julia> @btime f1($df);
  322.147 ns (4 allocations: 960 bytes)

julia> @btime f2($df);
  304.626 ns (2 allocations: 912 bytes)
```

and before:

```

julia> df = DataFrame(rand(100, 100));

julia> function s2(df::AbstractDataFrame, rows::Integer = size(df, 1))
           rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
           DataFrame(AbstractVector[similar(x, rows) for x in DataFrames._columns(df)], copy(DataFrames.index(df)))
       end
s2 (generic function with 2 methods)

julia> f1(df) = [x for x in eachcol(df, false)]

julia> f2(df) = [x for x in DataFrames._columns(df)]

julia> @btime similar($df, 0);
  7.197 μs (115 allocations: 15.33 KiB)

julia> @btime s2($df, 0);
  6.940 μs (114 allocations: 15.31 KiB)

julia> @btime f1($df);
  379.847 ns (4 allocations: 960 bytes)

julia> @btime f2($df);
  303.377 ns (2 allocations: 912 bytes)
```

CC @oxinabox

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -243,7 +243,7 @@ as `df`. An optional second argument can be provided to request a number of rows
 that is different than the number of rows present in `df`.
 """
 function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
-    rows < 0 && throw(ArgumentError("the number of rows must be positive"))
+    rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
     DataFrame(AbstractVector[similar(x, rows) for x in columns(df)], copy(index(df)))
 end
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -244,7 +244,11 @@ that is different than the number of rows present in `df`.
 """
 function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
     rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
-    DataFrame(AbstractVector[similar(x, rows) for x in columns(df)], copy(index(df)))
+    cols = Vector{AbstractVector}(undef, size(df, 2))
+    for (i, col) in enumerate(columns(df))
+        cols[i] = similar(col, rows)
+    end
+    DataFrame(cols, copy(index(df)))
 end
 
 ##############################################################################
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -244,11 +244,7 @@ that is different than the number of rows present in `df`.
 """
 function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
     rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
-    cols = Vector{AbstractVector}(undef, size(df, 2))
-    for (i, col) in enumerate(columns(df))
-        cols[i] = similar(col, rows)
-    end
-    DataFrame(cols, copy(index(df)))
+    DataFrame(AbstractVector[similar(x, rows) for x in columns(df)], copy(index(df)))
 end
 
 ##############################################################################
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -165,12 +165,16 @@ columns(df::AbstractDataFrame) = eachcol(df, false)
 Base.size(itr::DataFrameColumns) = (size(itr.df, 2),)
 Base.IndexStyle(::Type{<:DataFrameColumns}) = Base.IndexLinear()
 
-Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame,
-                                                             Pair{Symbol, AbstractVector}},
-                                       j::Int) = _names(itr.df)[j] => itr.df[j]
+@inline function Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame,
+                                                     Pair{Symbol, AbstractVector}}, j::Int)
+    @boundscheck checkbounds(itr, j)
+    @inbounds _names(itr.df)[j] => itr.df[j]
+end
 
-Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame, AbstractVector},
-                                       j::Int) = itr.df[j]
+@inline function Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame, AbstractVector}, j::Int)
+    @boundscheck checkbounds(itr, j)
+    @inbounds itr.df[j]
+end
 
 """
     mapcols(f::Union{Function,Type}, df::AbstractDataFrame)
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -165,18 +165,12 @@ columns(df::AbstractDataFrame) = eachcol(df, false)
 Base.size(itr::DataFrameColumns) = (size(itr.df, 2),)
 Base.IndexStyle(::Type{<:DataFrameColumns}) = Base.IndexLinear()
 
-@inline function Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame,
-                                                     Pair{Symbol, AbstractVector}},
-                               j::Int)
-    @boundscheck checkbounds(itr, j)
-    _names(itr.df)[j] => itr.df[j]
-end
+Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame,
+                                                             Pair{Symbol, AbstractVector}},
+                                       j::Int) = _names(itr.df)[j] => itr.df[j]
 
-@inline function Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame, AbstractVector},
-                               j::Int)
-    @boundscheck checkbounds(itr, j)
-    itr.df[j]
-end
+Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame, AbstractVector},
+                                       j::Int) = itr.df[j]
 
 """
     mapcols(f::Union{Function,Type}, df::AbstractDataFrame)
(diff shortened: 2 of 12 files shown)


src/dataframe/dataframe.jl
```diff
@@ -239,7 +239,10 @@ ncol(df::DataFrame) = length(index(df))
 ##############################################################################
 
 # df[SingleColumnIndex] => AbstractVector, the same vector
-function Base.getindex(df::DataFrame, col_ind::ColumnIndex)
+Base.@propagate_inbounds Base.getindex(df::DataFrame, col_ind::Union{Signed, Unsigned}) =
```
bkamins: `Union{Signed, Unsigned}` is unfortunately needed instead of `Integer` because we want to keep compatibility with Julia 0.7 where handling of `Bool` indexing is different than on Julia 1.0.

src/abstractdataframe/iteration.jl
```diff
@@ -165,18 +165,12 @@ columns(df::AbstractDataFrame) = eachcol(df, false)
...
 
-@inline function Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame,
-                                                     Pair{Symbol, AbstractVector}},
-                               j::Int)
-    @boundscheck checkbounds(itr, j)
-    _names(itr.df)[j] => itr.df[j]
-end
+Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns{<:AbstractDataFrame,
+                                                             Pair{Symbol, AbstractVector}},
+                                       j::Int) = _names(itr.df)[j] => itr.df[j]
```
nalimilan: Maybe keep the `@boundscheck`? It should give a more explicit error than the one coming from `_names(itr.df)[j]`. Then just add `@inbounds` to the next line, and `@inlined` should be enough (instead of `@propagate_inbounds`).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -244,7 +244,11 @@ that is different than the number of rows present in `df`.
 """
 function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
     rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
-    DataFrame(AbstractVector[similar(x, rows) for x in columns(df)], copy(index(df)))
```
nalimilan: I'm not really convinced by the move from a comprehension to a longer loop. Am I right it only reduces the time taken by `similar` by 2%?

bkamins: I was getting various performance results that I did not understand (i.e. if the compiler worked correctly there should be no cost to this and I would expect that in the long run there should be 0% difference when Julia compiler matures).

The cost was up to around 10% that was originally indicated by @oxinabox, but most of the time it was several percent. I would be OK with leaving the comprehension, but I just tried to squeeze out something as this was the original reason for this PR (but I would keep the improved `getindex` methods even if we revert `similar` as they are touching the fundamental functionality of the package that is often used).

oxinabox: my 10% was for `similar(df, 0)`
(which was the particular case that was worth optimising to me at the time I started looking at the code).

src/abstractdataframe/iteration.jl
bkamins: OK - I will fix this this way, but let us decide what we do in `similar` and then I will push a fix.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Even 10% doesn't sound a lot. What kind of operation are you doing so that 10% of something as specific as `similar(df, 0)` matters?!

oxinabox: In short, database query stuff. 
`simiilar(df, 0)` is what is made to make distinct empty results when nothing is returned.

nalimilan: Sure, it's always worth investigating. But given that we don't save a lot, I don't think it's worth making the code more complex.

BTW, if  `similar(df, 0)` matters for your user case, you could probably create a special `AbstractDataFrame` object which would just store the column types, it should be dramatically faster.

bkamins:
@nalimilan I have added the changes for your reference, but they have significantly worse performance.
The reason is twofold:
* comprehensions are not iterating with `@inbounds`
* `checkbounds` on `AbstractDataFrame` is significantly slower than `checkbounds` on an array (and I think this cannot be helped in general)

Here are the benchmarks for comparison:
```
[ Info: Recompiling stale cache file C:\Users\bogum\.julia\compiled\v1.1\DataFrames\AR9oZ.ji for DataFrames [a93c6f00-e57d-5684-b7b6-d8193f3e46c0]

julia>  df = DataFrame(rand(100, 100));

julia> function s2(df::AbstractDataFrame, rows::Integer = size(df, 1))
                  rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
                  DataFrame(AbstractVector[similar(x, rows) for x in DataFrames._columns(df)], copy(DataFrames.index(df)))
              end
s2 (generic function with 2 methods)

julia> f1(df) = [x for x in eachcol(df, false)]

julia> f2(df) = [x for x in DataFrames._columns(df)]

julia> @btime similar($df, 0);
  6.940 μs (115 allocations: 15.33 KiB)

julia> @btime s2($df, 0);
  6.811 μs (114 allocations: 15.31 KiB)

julia>  @btime f1($df);
  350.825 ns (4 allocations: 960 bytes)

julia> @btime f2($df);
  297.825 ns (2 allocations: 912 bytes)
```

Therefore I would leave `similar` as is, but I would revert to `Base.@propagate_inbounds` without `@boundscheck`.

bkamins:
I agree - if you are then OK with what we have now then can you please approve 😄.

bkamins:
The error for `eachcol` is OK I think:
```
julia> checkbounds(eachcol(df, false), 100)
ERROR: BoundsError: attempt to access 10-element DataFrames.DataFrameColumns{DataFrame,AbstractArray{T,1} where T} at index [100]
```
and I have improved the error messages for `getindex` of `DataFrame`.

src/dataframe/dataframe.jl
```diff
@@ -258,19 +264,40 @@ end
...
+@inline function Base.getindex(df::DataFrame, row_ind::Integer,
+                               col_ind::Union{Signed, Unsigned})
+    cols = _columns(df)
+    @boundscheck begin
+        if !checkindex(Bool, cols, col_ind)
+            throw(BoundsError("attempt to access a data frame with $(ncol(df)) " *
+                              "columns at index $row_inds"))
+        end
+        if !checkindex(Bool, axes(df, 1), row_ind)
+        throw(BoundsError("attempt to access a data frame with $(nrow(df)) " *
```
nalimilan: Wrong indentation.

src/dataframe/dataframe.jl
```diff
@@ -255,15 +264,40 @@ end
...
+    end
+    @inbounds _columns(df)[selected_column][row_ind]
 end
 
 # df[MultiRowIndex, SingleColumnIndex] => AbstractVector, copy
-function Base.getindex(df::DataFrame, row_inds::AbstractVector, col_ind::ColumnIndex)
+@inline function Base.getindex(df::DataFrame, row_inds::AbstractVector, col_ind::ColumnIndex)
     selected_column = index(df)[col_ind]
-    return _columns(df)[selected_column][row_inds]
+    if !checkindex(Bool, axes(df, 1), row_ind)
```
```suggestion
    @boundscheck if !checkindex(Bool, axes(df, 1), row_ind)
```

src/other/index.jl
```diff
@@ -145,8 +145,19 @@ end
 @inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) = collect(Int, idx)
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
 
-@inline Base.getindex(x::Index, idx::Symbol) = x.lookup[idx]
-@inline Base.getindex(x::Index, idx::AbstractVector{Symbol}) = [x.lookup[i] for i in idx]
+@inline function lookupname(l::Dict{Symbol, Int}, idx::Symbol)
+    i = get(l, idx, nothing)
+    if i === nothing
+        throw(BoundsError("column name $idx not found in the data frame"))
```
nalimilan: Maybe add `:`? Not sure what convention we use in this kind of situation.

Also, not to be done now, but I've been thinking for a long time that it would be great to suggest appropriate column names if we detect a case mismatch, or even better if we can identify a similar name based on some distance metric. :-)

src/other/index.jl
```diff
@@ -145,8 +145,19 @@ end
...
+    end
+    i
+end
+
+@inline Base.getindex(x::Index, idx::Symbol) = lookupname(x.lookup, idx)
+@inline function Base.getindex(x::Index, idx::AbstractVector{Symbol})
+    l = x.lookup
```
nalimilan: `Index` is immutable, so no need to store a temporary variable (then you can keep the single-line form).

src/dataframe/dataframe.jl

src/other/index.jl

src/other/index.jl
bkamins: Added `:` as indeed this is what we use it when printing errors (but note that we have just dropped `:` when showing `GroupedDataFrame` per your suggestion 😄).

I have implemented fuzzy matching listing columns that differ in at most one position in edit distance and disregarding case.

bkamins:
I have changed `BoundsError` to `ArgumentError` in case a column name is missing as it seemed better for me in that case. @nalimilan - are you OK with this?

nalimilan:
I re-approve!

PR #1703, opened 2019-01-29
Use optimized grouping method for PooledArrays

nalimilan:
The PR currently adds a dependency on PooledArrays. I'm not sure what's the best approach. We could define a custom method behind `@require PooledArrays` to avoid the dependency (calling a separate function to avoid duplicating the code). But I figured we may want to add a `pool!` function to parallel `categorical!`, in which case we would have to take the dependency anyway. Though we can defer this issue for later, and use `@require` until then.

Most of the work was to adapt tests to make them more generic. I've added a helper for that. Note that it is better to ignore whitespace to see what actually changed.

Obligatory benchmark:
```julia
using DataFrames, BenchmarkTools, PooledArrays

df1 = DataFrame(key1=categorical(string.(rand(1:10, 500_000))),
                key2=categorical(string.(rand(1:10, 500_000))));
df1p = DataFrame(key1=PooledArray(string.(rand(1:10, 500_000))),
                 key2=PooledArray(string.(rand(1:10, 500_000))));

df2 = DataFrame(key1=categorical(string.(rand(1:1000, 500_000))),
                key2=categorical(string.(rand(1:1000, 500_000))));
df2p = DataFrame(key1=PooledArray(string.(rand(1:1000, 500_000))),
                 key2=PooledArray(string.(rand(1:1000, 500_000))));

# master
julia> @btime groupby(df1, :key1);

  2.304 ms (51 allocations: 7.63 MiB)

julia> @btime groupby(df2, :key1);
  4.240 ms (51 allocations: 7.66 MiB)

julia> @btime groupby(df1, [:key1, :key2]);
  3.364 ms (53 allocations: 7.63 MiB)

julia> @btime groupby(df2, [:key1, :key2]);
  37.651 ms (59 allocations: 22.23 MiB)

julia> @btime groupby(df1p, :key1);
  28.769 ms (48 allocations: 15.45 MiB)

julia> @btime groupby(df2p, :key1);
  34.061 ms (48 allocations: 15.46 MiB)

julia> @btime groupby(df1p, [:key1, :key2]);
  58.622 ms (499949 allocations: 23.08 MiB)

julia> @btime groupby(df2p, [:key1, :key2]);
  82.056 ms (106523 allocations: 23.08 MiB)


# This PR
julia> @btime groupby(df1, :key1);
  2.357 ms (51 allocations: 7.63 MiB)

julia> @btime groupby(df2, :key1);
  4.197 ms (51 allocations: 7.66 MiB)

julia> @btime groupby(df1, [:key1, :key2]);
  3.337 ms (53 allocations: 7.63 MiB)

julia> @btime groupby(df2, [:key1, :key2]);
  41.909 ms (59 allocations: 22.23 MiB)


julia> @btime groupby(df1p, :key1);
  2.308 ms (51 allocations: 7.63 MiB)

julia> @btime groupby(df2p, :key1);
  4.163 ms (51 allocations: 7.66 MiB)

julia> @btime groupby(df1p, [:key1, :key2]);
  3.249 ms (53 allocations: 7.63 MiB)

julia> @btime groupby(df2p, [:key1, :key2]);
  37.511 ms (59 allocations: 22.24 MiB)
```

### DIFF
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -1,7 +1,32 @@
 module TestGrouping
 
-using Test, DataFrames, Random, Statistics
+using Test, DataFrames, Random, Statistics, PooledArrays
 const ≅ = isequal
+
+"""Check that groups in gd are equal to provided data frames, ignoring order"""
+function isequal_unordered(gd::GroupedDataFrame,
+                            dfs::AbstractVector{<:AbstractDataFrame})
+    n = length(gd)
+    @assert n == length(dfs)
+    remaining = Set(1:n)
+    for i in 1:n
+        for j in remaining
+            if gd[i] ≅ dfs[j]
+                pop!(remaining, j)
+                break
+            end
+        end
+    end
+    isempty(remaining) || error("gd is not equal to provided groups")
+end
+
+"""Helper to set the order of values in the pool and add unused values"""
+function _levels!(x::PooledArray, levels::AbstractVector)
+    res = similar(x)
+    copyto!(res, levels)
+    copyto!(res, x)
+end
+_levels!(x::CategoricalArray, levels::AbstractVector) = levels!(x, levels)
 
 function groupby_checked(df::AbstractDataFrame, keys, args...; kwargs...)
     gd = groupby(df, keys, args...; kwargs...)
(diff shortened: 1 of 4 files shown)


test/grouping.jl
```diff
@@ -1,8 +1,33 @@
...
+"""Check that groups in gd are equal to provided data frames, ignoring order"""
+function isequal_unordered(gd::GroupedDataFrame,
+                            dfs::AbstractVector{<:AbstractDataFrame})
+    n = length(gd)
+    @assert n == length(dfs)
+    remaining = Set(1:n)
+    for i in 1:n
+        for j in remaining
+            if gd[i] ≅ dfs[j]
+                pop!(remaining, j)
```
bkamins: does not this `pop!` potentially break iteration (I think it is OK, but just asking to be sure if you checked - some 1 year ago I was fixing something around this in Julia Base, but cannot recall in detail)

bkamins:
Assuming PooledArrays.jl will be maintained I think it is OK to have it as a dependency as it is really a small package.
The PR looks good (I have to sleep over it though).

test/grouping.jl
bkamins: Actually I can see it is wrong. Run the following code several times
```
function test()
    a = rand(1:1000, 20)
    x = Set(a)
    y = Int[]
    i = 0
    for v in x
        i += 1
        push!(y, pop!(x, v))
    end

    return isempty(x), i, a, y
end
```

and you will see that sometimes `i<20` and then `a` and `y` are not the same.

bkamins: And now I think I remember - you can safely delete elements, but you cannot add elements while iterating.

nalimilan: Good point. I'm not sure what the rules are, but here I think the fact that we call `break` immediately should prevent any problems anyway.

bkamins: True. In general I have an Pavlovian conditioning from other languages when I see that a collection that is iterated over is mutated inside the iterator 😄.

nalimilan:
Let's wait for https://github.com/JuliaComputing/PooledArrays.jl/pull/15.

bkamins:
Looks good 😄.

Issue #1696, opened 2019-01-24
Undeprecate `DataFrameRow(df, i)`


rofinn:
I'm not really clear why `DataFrameRow(df, i)` was deprecated, but that seems like a bad idea as `DataFrameRow(df, i, :)` doesn't seem like an intuitive API.

ararslan:
To me "row" already implies that you're selecting all columns from the row, so the `:` (which I assume indicates selecting all columns) feels redundant.

ararslan:
For reference, the deprecation was added in https://github.com/JuliaData/DataFrames.jl/pull/1648.

nalimilan:
Fine with me (though note `df[i, :]` is shorter). @bkamins?

rofinn:
@bkamins Thanks for the explanation. I still think that the column vs row indexing distinction could have been minimized by always indexing rows by integers and columns by symbols. 

One of our use cases is building up hash indices of our "primary key". We can then cache these hashes for future queries.

```
for i in 1:n
     r = DataFrameRow(df, i)

     # Some checks to avoid computing a hash for a row that we aren't going to need to query

     pk_hashes[i] = hash(tuple((r[c] for c in latest_key)...))
end
```

I suppose we could replace that with an `enumerate(eachrow(df))` call, but that seems a bit unnecessary if `DataFrameRow` already exists (assuming that we aren't calling `DataFrameRow(df, :, i)`).

bkamins:
> `r = df[i, :]`

that was my point. In this way a new user does not have to learn about DataFrames.jl internals - it is enough that one knows Base.

> column vs row indexing distinction could have been minimized by always indexing rows by integers and columns by symbols.

Actually given we now treat `AbstractDataFrame` as a collection of rows by default I would prefer `df[idx]` to always perform row selection than column selection - no matter what `idx` is, but I guess it would be too breaking. This would be mostly non-problematic for users as we support `df.col` for column selection syntax via `getproperty`.

bkamins:
But I think if the majority wants to keep `DataFrameRow(::AbstractDataFrame, ::Integer)` I will not object. I will make a PR and document this exception to the general rules we have.

rofinn:
> For something even more efficient, see the internal hashrow_cols function.

Thanks, I'll take a look at it. I'll note that the checks for skipping the `hash` call require looking at various columns, so the `hashrows_col!` approach would require multiple passes over the dataframe (one pass for the mask and another for the colwise hashing).

[...]

PR #1705, opened 2019-01-31
Undeprecate DataFrameRow constructor, fix line length, reorganize tests

bkamins:
This fixes https://github.com/JuliaData/DataFrames.jl/issues/1696.
Additionally I have cleaned up some related line length issues in the code
and reorganized test code, because it was hard to reason about.

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -69,7 +69,6 @@ end
 Base.@propagate_inbounds DataFrameRow(df::AbstractDataFrame, row::Integer) =
     DataFrameRow(df, row, :)
 
-
 row(r::DataFrameRow) = getfield(r, :row)
 Base.parent(r::DataFrameRow) = getfield(r, :df)
 Base.parentindices(r::DataFrameRow) = (row(r), parentcols(index(r)))
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -3,7 +3,7 @@
 
 A view of one row of an `AbstractDataFrame`.
 
-A `DataFrameRow` is constructed with `view` or `getindex` when one row and a
+A `DataFrameRow` is returned by `getindex` or `view` functions when one row and a
 selection of columns are requested, or when iterating the result
 of the call to the [`eachrow`](@ref) function.
 
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -7,6 +7,12 @@ A `DataFrameRow` is constructed with `view` or `getindex` when one row and a
 selection of columns are requested, or when iterating the result
 of the call to the [`eachrow`](@ref) function.
 
+You can also call a `DataFrameRow` constructor directly using the following syntax:
+
+```
+DataFrameRow(parent::AbstractDataFrame, row::Integer, cols=:)
+```
+
 A `DataFrameRow` supports the iteration interface and can therefore be passed to
 functions that expect a collection as an argument.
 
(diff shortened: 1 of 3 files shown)


test/dataframerow.jl
```diff
@@ -3,226 +3,281 @@ module TestDataFrameRow
...
-@test hash(DataFrameRow(df, 1, :)) == hash(DataFrameRow(df, 4, :))
-@test hash(DataFrameRow(df, 2, :)) == hash(DataFrameRow(df, 5, :))
-@test hash(DataFrameRow(df, 2, :)) != hash(DataFrameRow(df, 6, :))
-
-# check that hashrows() function generates the same hashes as DataFrameRow
-df_rowhashes, _ = DataFrames.hashrows(Tuple(columns(df)), false)
-@test df_rowhashes == [hash(dr) for dr in eachrow(df)]
-
-# test incompatible frames
-@test_throws UndefVarError DataFrameRow(df, 1, :) == DataFrameRow(df3, 1, :)
```
bkamins: I have removed this test as it made no sense.

test/dataframerow.jl
```diff
@@ -3,226 +3,281 @@ module TestDataFrameRow
...
+    @test DataFrameRow(df, 1, [:a]) == DataFrameRow(df2, big(1), :)
+    @test DataFrameRow(df, 1, :) != DataFrameRow(df, 2, :)
+    @test DataFrameRow(df, 1, :) != DataFrameRow(df, 3, :)
+    @test DataFrameRow(df, 1, :) == DataFrameRow(df, 4, :)
+    @test ismissing(DataFrameRow(df, 2, :) == DataFrameRow(df, 5, :))
+    @test isequal(DataFrameRow(df, 2, :), DataFrameRow(df, 5, :))
+    @test ismissing(DataFrameRow(df, 2, :) != DataFrameRow(df, 6, :))
+    @test !isequal(DataFrameRow(df, 2, :), DataFrameRow(df, 6, :))
+
+    dc_df = deepcopy(df)
```
bkamins: and added these tests instead of this broken `df3` test that I have removed.

src/dataframerow/dataframerow.jl
```diff
@@ -7,6 +7,12 @@ A `DataFrameRow` is constructed with `view` or `getindex` when one row and a
 selection of columns are requested, or when iterating the result
 of the call to the [`eachrow`](@ref) function.
 
+You can also call a `DataFrameRow` constructor directly using the following syntax:
```
```suggestion
The `DataFrameRow` constructor can also be called directly:
```
Above, maybe say "is returned" instead of constructed?

test/dataframerow.jl
```diff
@@ -3,226 +3,281 @@ module TestDataFrameRow
...
-ref = ["a", "b", "c"]
-df = DataFrame(permutedims(ref))
-dfr = df[1, :]
-@test Base.IteratorEltype(dfr) == Base.EltypeUnknown()
-@test collect(dfr) == ref
-@test eltype(collect(dfr)) === String
-for (v1, v2) in zip(ref, dfr)
-    @test v1 == v2
+@testset "getindex and setindex!" begin
+    df = DataFrame(a=Union{Int, Missing}[1, 2, 3, 1, 2, 2],
```
nalimilan: Maybe we could define this only once at the top of the file, and copy it before modifying it?

src/dataframerow/dataframerow.jl

test/dataframerow.jl
bkamins: I will do this. But this was exactly where the mess was `df` was overwritten several times + modified in place and you could not tell what is the actual value of `df` in a given group of tests.

bkamins: actually `deepcopy` is needed

test/dataframerow.jl
```diff
@@ -1,228 +1,265 @@
...
-# check that hashrows() function generates the same hashes as DataFrameRow
-df_rowhashes, _ = DataFrames.hashrows(Tuple(columns(df)), false)
-@test df_rowhashes == [hash(dr) for dr in eachrow(df)]
-
-# test incompatible frames
-@test_throws UndefVarError DataFrameRow(df, 1, :) == DataFrameRow(df3, 1, :)
-
-# test RowGroupDict
-N = 20
-d1 = rand(map(Int64, 1:2), N)
```
bkamins: I have rewritten the definition of `df5` because here is was not reproducible and it was possible that `d1` actually would hold only `1` or only `2` when the tests would fail.

PR #1708, opened 2019-02-01
Use standard Travis script

nalimilan:
Bounds checking is already enabled by default.


bkamins:
Can you point me to the reference where I can check it?

bkamins:
OK - I see it is in `Pkg.test`

PR #1710, opened 2019-02-04
Fix constructor for Vector{<:NamedTuple}

### LINKED ISSUE #1670: Converting to DataFrame fails for NamedTuple[]
I am submitting this issue based on a discussion on [Discourse](https://discourse.julialang.org/t/converting-array-of-named-tuples-to-dataframes/19378/2). It is possible to convert an Array of NamedTuples to a DataFrame if it is initialized as Any[]. However, this is not possible if the array is initialized as NamedTuple[].  The code below demonstrates the behavior. 

```
using DataFrames
#Works
x1 = [(x=rand(),y=rand([:a,:b])) for i in 1:10]
d1 = DataFrame(x1)

#Works
x2 = Any[(x=rand(),y=rand([:a,:b])) for i in 1:10]
d2 = DataFrame(x2)

#Does not work
x3 = NamedTuple[(x=rand(),y=rand([:a,:b])) for i in 1:10]
d3 = DataFrame(x3)

ERROR: MethodError: no method matching Tables.Schema(::Type{NamedTuple})
Closest candidates are:
  Tables.Schema(::Any, ::Nothing) at /home/dfish/.julia/packages/Tables/zyTIO/src/Tables.jl:159
  Tables.Schema(::Any, ::Any) at /home/dfish/.julia/packages/Tables/zyTIO/src/Tables.jl:160
  Tables.Schema(::Tuple{Vararg{Symbol,N} where N}, ::Type{T<:Tuple}) where T<:Tuple at /home/dfish/.julia/packages/Tables/zyTIO/src/Tables.jl:157
  ...
Stacktrace:
 [1] schema(::Tables.DataValueUnwrapper{Array{NamedTuple,1}}) at /home/dfish/.julia/packages/Tables/zyTIO/src/query.jl:29
 [2] columns at /home/dfish/.julia/packages/Tables/zyTIO/src/fallbacks.jl:152 [inlined]
 [3] DataFrame(::Array{NamedTuple,1}) at /home/dfish/.julia/packages/DataFrames/z2XOB/src/other/tables.jl:45
 [4] top-level scope at none:0
```
As pointed out by the other poster, x3 is a table, but is not handled by the interface. 

`Tables.istable(typeof(x3))`

Nosferican:
Should fix #1670

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -26,7 +26,7 @@ end
 Base.append!(df::DataFrame, x) = append!(df, DataFrame(x))
 
 # This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
-DataFrame(x::Vector{T}) where {T <: NamedTuple} = fromcolumns(Tables.columns(Tables.DataValueUnwrapper(x)))
+DataFrame(x::Vector{<:NamedTuple}) = fromcolumns(Tables.columns(Tables.IteratorWrapper(x)))
 
 IteratorInterfaceExtensions.getiterator(df::AbstractDataFrame) = Tables.datavaluerows(df)
 IteratorInterfaceExtensions.isiterable(x::AbstractDataFrame) = true
--- a/REQUIRE
+++ b/REQUIRE
@@ -9,6 +9,6 @@ DataStreams 0.3.0
 CodecZlib 0.4
 TranscodingStreams
 Compat 0.59.0
-Tables 0.1.14
+Tables 0.1.15
 IteratorInterfaceExtensions 0.1.1
 TableTraits 0.4.0
--- a/test/tables.jl
+++ b/test/tables.jl
@@ -109,6 +109,8 @@ Base.propertynames(d::DuplicateNamesColumnTable) = (:a, :a, :b)
         @test_throws ArgumentError (dn |> DataFrame)
 
         # non-Tables.jl constructor fallbacks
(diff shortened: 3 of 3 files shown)


src/other/tables.jl
```diff
@@ -26,7 +26,7 @@ end
 Base.append!(df::DataFrame, x) = append!(df, DataFrame(x))
 
 # This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
-DataFrame(x::Vector{T}) where {T <: NamedTuple} = fromcolumns(Tables.columns(Tables.DataValueUnwrapper(x)))
+DataFrame(x::Vector{<:NamedTuple}) = fromcolumns(Tables.columns(Tables.IteratorWrapper(x)))
```
nalimilan: Should this be `AbstractVector`?

bkamins: It can be a custom `AbstractVector` of with, in particular, a most common case a `SubArray`.

Nosferican: Guess we can check whether these are supported by Tables and if such move to use `AbstractVector`. I assume this method to avoid ambiguities should work fine with `AbstractVector`.

bkamins: Yes, but my general thinking is that we should make it consistent across the ecosystem, so if something would need to be changed in Tables.jl we should also change it there in sync. `SubArray` is actually quite possible to appear if you would perform subseting of rows using a `view` (which with very wide tables is exactly what you want to do given `NamedTuple` is immutable).

PR #1712, opened 2019-02-04
Add columnindex function

bkamins:
I propose to expose a `columnindex` function that gives you an integer index of a column given by name. Working also for vectors of `Symbol`s. I am not very attached to the name, but I could not think of anything better.


bkamins:
@nalimilan: do you understand the error thrown on Julia 0.7?
+
a question if we want to keep supporting Julia 0.7 since it is not maintained now.

nalimilan:
How useful is this? `findfirst(==(n), names(df))` isn't too complex, and even if it's a bit slower in practice it shouldn't matter, should it?

bkamins:
`findfirst(==(n), names(df))` is what I recommend to my students (see the last cell in https://github.com/bkamins/Julia-DataFrames-Tutorial/blob/master/05_columns.ipynb 😄), but:
1. casual students find it difficult to grasp (many things happen in this single line)
2. it only handles one name (not a vector)

I would propose to leave this PR open as there is no rush to add it and see if people would comment if they find it useful.

nalimilan:
OK. So you mean students need this function in real use cases?

I wonder whether we should coordinate with AxisArrays and NamedArrays about this. That sounds like an operation which makes sense for any kind of structure with dimension names (though data frames only have names for columns so a dedicated function can be enough). Also, we could imagine giving access to the `Index`, perhaps making it `<: AbstractVector{Symbol}`. These are just ideas.

bkamins:
You know - it is one of the first questions you ask. Does column `:blah` exist in a `DataFrame` (my answer: you can check it with `haskey`) and what is its number (and here the answer is a complex expression). Later I would say it is not that super useful because you rarely need to access column by number. The only case is when you say that I want to get all columns from column name `:blah1` to column name `:blah2` (often columns are grouped together logically).

Now your proposal is interesting. We could actually give access to `lookup` dict (just as we already give access to `names` field). Then the question is what name such function should have.

bkamins:
Actually exposing a `lookup` dict is not a good idea. Maybe we simply should export `index` function (possibly renaming it to `columnindex` or `keyindex`)? And follow the same approach we are considering with `names` - i.e. assume that user does not modify it and we do not try wrapping it to avoid write.

nalimilan:
That's a possibility, but that would expose lots of things for which we don't really have a strong use case. Honestly, I'd rather leave this aside for now and concentrate on essential features that need to be stabilized before 1.0.

It would also be interesting to have a look at other implementations.

bkamins:
1. Yes - we can leave it post 1.0, so we just can keep it open.
2. Base R and Pandas do not have it, data.table has it
3. Actually I realized that we can simply and safely expose `AbstractIndex` to the user. Most of the functions are read-only. The functions that mutate index can be renamed with `_` and not exported. Effectively we would expose `AbstractIndex` that will be a read only structure from the perspective of the user - and thus safe.

bkamins:
@quinnj + @nalimilan 
My personal preference is to have a `columnindex` function that takes a table and a list of columns and returns their canonical representation using only `Int`.

I am not sure if the name `columnindex` is best. Also maybe this function should be added to Tables.jl as a general interface (as probably it is useful to have a common function for all tabular data formats returning a canonical representation of column indices). Of course then we would specialize it in DataFrames.jl.

This would go in hand with what @quinnj mentioned that we could have a common column indexing mechanics across Tables.jl.

It would be great, from my perspective, to have this decision before DataFrames.jl 1.0.

quinnj:
Tables.jl has https://github.com/JuliaData/Tables.jl/blob/master/src/utils.jl#L208; which allows you to do `Tables.columnindex(Tables.schema(table).names, colname)`. But yeah, we could have another definition that essentially did that for you on any table.

bkamins:
@quinnj - excellent. So in fact we already have it 😄. So please confirm that, given your general intentions, it is OK that:
* I overload it in DataFrames.jl to accept: a `DataFrame` as a first argument?
* I export `columnindex` from Tables.jl in DataFrames.jl?

quinnj:
Well, I think we can just have:
```julia
Tables.columnindex(any_table, colname) = Tables.columnindex(Tables.schema(any_table).names, colname)
```
defined in Tables.jl itself, then people can just use `Tables.columnindex(table)` where-ever they want, nothing needed in DataFrames.

bkamins:
OK. Will you add this in Tables.jl? 😄

I will update this PR for DataFrames.jl anyway because we can do `Dict` lookup instead of linear lookup (this will be consistent with Tables.jl so nothing has to be exported, just a specializing method for performance).

quinnj:
Well, the Tables.jl method is `@pure`, so if it's used in a compiled setting, it'll be even faster ;)

bkamins:
OK - closing. I will update https://github.com/bkamins/Julia-DataFrames-Tutorial when Tables.jl with the change is tagged.

Issue #1655, opened 2018-12-26
Decision: remove `_names` and make `names` not perform a copy


bkamins:
We could decide to rename `_names` to `names` and avoid making a copy of names held in index (see https://github.com/JuliaData/DataFrames.jl/pull/1648#discussion_r243924332).

pdeffebach:
`names(df)` = [:a, :b, :c]` always seems intuitive. Would be cool if it were just a `getindex` call with some error checking.

bkamins:
The idea is that names are stored in the index either as a `Vector{Symbol}` or as a `view` (in case of `SubDataFrame` or `DataFrameRow`) and we could give a direct access to the user to these objects (this is a bit risky, as they should really be read-only).

@nalimilan maybe `names` should return a read-only custom version of an `AbstractVector`? (similar to what we have with `codeunits` and `String`)

bkamins:
@nalimilan a general purpose package would be good (or a feature in some standard package e.g. from https://github.com/JuliaCollections).
Could you please ask as I will be out for several days and let us know (please mention me in the question so that I can track it)? Thank you.

bkamins:
@nalimilan Have you had a chance to discuss this with Base devs? If no I can ask a question on Discourse.

nalimilan:
See https://discourse.julialang.org/t/convention-for-read-only-array-getters/19356/1.

bkamins:
@nalimilan I have created a proposal of a read-only array type here https://github.com/bkamins/ReadOnlyArrays.jl.

bkamins:
Any opinion on https://github.com/bkamins/ReadOnlyArrays.jl?

1. is it OK?
2. should we use it (then probably I should register this package)?
3. or should we move this code somewhere else?

bkamins:
Given the response and discussion I would propose that we remove `_names` and expose internal vector `names` (like `levels` in CategoricalArrays.jl). We should also document it.

Probably we should throw a depwarn here then giving a hint to use `copy(names(df))` if one wants a copy.

Do we go ahead this way?

nalimilan:
Fine we me. The advantage with this solution is that we can change it later to return a `ReadOnlyArray` if we want, it won't break valid code. Adding a deprecation would be annoying since there's no way to avoid it (except for adding a temporary argument to `names`, which then will have to be deprecated later...).

[...]

Issue #1695, opened 2019-01-24
Policy regarding in-place operations


nalimilan:
Some in-place operations like `append!` operate in-place on column vectors, but others like `allowmissing!` replace vectors. We should check whether this difference is justified and whether it is the most useful behavior.

For example, it might not be very useful for `append!` to resize column vectors in-place, since that does not save any memory and prevents it from changing the column eltypes. OTOH that could be a useful difference from `vcat` if one wants to preserve the original eltypes.

bkamins:
Today I hit a problem with `dropmissing!` related to this. I created a dataframe with a subset of two columns to perform pairwise deletion of missings and it corrupted the original wide `DataFrame`. I knew what to do (at least I hope I did 😄), but probably newcomers to DataFrames.jl will be caught off-guard here unless we properly document it.

nalimilan:
A possible idea: use a `!` suffix for operations that mutate the data frame but not the column vectors (the majority), and a `!!` suffix for operations that also mutate the vectors (which are not that common, and often trap users who don't want to mutate column vectors). That would be both explicit, consistent are still relatively convenient.

bkamins:
In general silently mutating vector type except for `df.x = something` (in this case it is obvious it can be mutating) should be signaled to the user so `!!` seems reasonable. As I have said in #1716 an implicit type mutation without signaling can lead to nasty bugs IMO.

pdeffebach:
Continuing from #1716 

> So is your proposal to make `push!` to do autopromotion (and create a new vector) and `push!!` to mutate the vector in place?

This kind of confusion shows why this kind of syntax-y solution is less than ideal. Though I do think it's interesting to think of a syntax that shows clearly the nested structure of a DataFrame

nalimilan:
So the idea was that a single `!` only mutates the data frame, so it can do anything to the object (including replacing column vectors), but `!!` can also mutate the column vectors (so it can't change their type). That's not really in contradiction with Base, since the single `!` version still doesn't change the type of the data frame, just like `push(Int[], 1.0)` converts `1.0` to `Int` since the type cannot change.

That said, I'm not saying that's the best solution.

bkamins:
I agree that there are pros and cons so my idea was to list all the affected functions and then see how many  changes we would have to make.

bkamins:
@nalimilan Given the discussion we had yesterday with @oxinabox I think I can explain both arguments behind `!!` briefly:
[...]

PR #1715, opened 2019-02-06
Make names return internal names

bkamins:
This PR follows #1655. It should not be treated lightly I think as it can lead to tricky bugs (so let us wait to see the feedback).

I have checked that we did not mutate `names(df)` anywhere in the package.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -83,8 +83,13 @@ abstract type AbstractDataFrame end
 ##
 ##############################################################################
 
+"""
+    names(df::AbstractDataFrame)
+
+Return a vector of column names in the `df` data frame. The returned vector
+should not be modified. If you need to modify it use `copy(names(df))`.
+"""
 Base.names(df::AbstractDataFrame) = names(index(df))
-_names(df::AbstractDataFrame) = _names(index(df))
 
 """
 Set column names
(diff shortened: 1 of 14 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -83,8 +83,13 @@ abstract type AbstractDataFrame end
 ##
 ##############################################################################
 
+"""
+    names(df::AbstractDataFrame)
+
+Return a vector of column names in the `df` data frame. The returned vector
+should not be modified. If you need to modify it use `copy(names(df))`.
+"""
```
nalimilan: Also add docstring to the manual?

bkamins: added (I also proposed to switch Basics and Split-apply-combine sections as it is more natural to start with basics)

bkamins:
@nalimilan https://github.com/JuliaData/DataFrames.jl/issues/1695 is not directly relevant here, but given the general "safety first", maybe `names` should make a copy (as this is not extremely expensive). Then maybe also `levels` should make a copy?

bkamins:
@nalimilan - given the direction we took (safety first) I would close it. Let `names` make a copy. In general `names` probably will not appear in performance critical code.

If you are OK with this then please just close this PR.

nalimilan:
We can keep it around, it doesn't really hurt.

bkamins:
OK 😄. But it is just a reminder really now.

We have so many changes in master that anyway it is simpler to open a new PR than to rebase (I just had to do it for #1752).

bkamins:
(@nalimilan can you please revoke the accept on this as we should not really merge it as is)

nalimilan:
OK, let's close then.

PR #1717, opened 2019-02-11
Add DataFrame constructors allowing NTuple and collection of Pair-s

bkamins:
Following https://github.com/JuliaData/DataFrames.jl/issues/1599 I have added 3 new constructors (allowing a Tuple for columns and for names - for completeness).

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -10,6 +10,9 @@ particularly a Vector or CategoricalVector.
 
 ```julia
 DataFrame(columns::Vector, names::Vector{Symbol}; makeunique::Bool=false)
+DataFrame(columns::Vector, names::NTuple{N,Symbol}; makeunique::Bool=false)
+DataFrame(columns::NTuple{N,AbstractVector}, names::Vector{Symbol}; makeunique::Bool=false)
+DataFrame(columns::NTuple{N,AbstractVector}, names::NTuple{N,Symbol}; makeunique::Bool=false)
 DataFrame(columns::Matrix, names::Vector{Symbol}; makeunique::Bool=false)
 DataFrame(kwargs...)
 DataFrame(pairs::Pair{Symbol}...; makeunique::Bool=false)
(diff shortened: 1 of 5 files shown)


bkamins:
I have added a constructor with collection of pairs also (without a deprecation - but I can add it if we found we need to).

bkamins:
@nalimilan this should be good to have a look at. No rush.

pdeffebach:
Is it necessary to have DataFrames constructors like this? Why not just off-load everything to Tables.jl?

bkamins:
1. constructor with iterable of `Pairs` produces a weird result (see https://github.com/JuliaData/DataFrames.jl/issues/1599 for an example) because of Tables.jl and is inconsistent with splatting the iterator
2. The other three constructors with Tuples now throw an error and are completely incompatibile to Tables.jl; we probably can live without them, they are added to complement `DataFrame(Vector,Vector)` constructor as in `setindex!` discussions (https://github.com/JuliaData/DataFrames.jl/pull/1646) we came to the conclusions that in such cases we should treat `AbstractVector{<:AbstractVector}` and `NTuple{N, AbstractVector}` in the same way so it is also consistent to treat them in the same way here.

The `AbstractVector` - `Tuple` duality is similar to the case of e.g. `push!` where we already accept both in the same way.

pdeffebach:
Thanks for the explanation!

Wow there are a lot of edge cases. Would it be better to throw errors due to ambiguity and then point the user towards something more specific?

bkamins:
All the functions for tuples are one-liners, so it is less code to implement them than to produce a proper deprecation.
I think we do not have to support a tuple for names argument (this would remove 2 methods from the list and leave only 2 new methods). I just thought it would be consistent to support tuples also there (but I am OK to allow only vectors of names or e.g. support only vector-vector and tuple-tuple constructor and disallow vector-tuple and tuple-vector combinations).

nalimilan:
> (but I am OK to allow only vectors of names or e.g. support only vector-vector and tuple-tuple constructor and disallow vector-tuple and tuple-vector combinations).

I agree we should drop methods mixing vectors and tuples. It's already complex enough to cover all reasonable cases...

bkamins:
OK - removed

PR #1721, opened 2019-02-14
end deprecation in dropmissing and dropmissing!

bkamins:
This ends deprecation of `disallowmissing` keyword argument (as I think everyone wants it so we do not have to wait with it that long). Also it shows (mostly in tests) the kind of thing #1695 will imply we should test against (I have moved out relevant tests to a separate testset for this).


bkamins:
and the tests additionally caught a hole in `disallowmissing!` and `allowmissing!` implementations 😄.

bkamins:
Sure - I just wanted to have a working example of the consequences, as it is always easier to look at a concrete case. It can wait.

Note that after this PR `dropmissing!` **does not** make copies of columns - it mutates all of them in place (in particular it mutates the `cols` columns that are dropped later before running the conversion if it is requested).

nalimilan:
> Note that after this PR `dropmissing!` **does not** make copies of columns - it mutates all of them in place (in particular it mutates the `cols` columns that are dropped later before running the conversion if it is requested).

Indeed it's wasteful to resize columns in place before making copies with a different type, especially given that it will be the most frequent case. But this could easily be improved later.

bkamins:
Agreed, but I was more referring to:

> e.g. in case we would like `dropmissing!` to avoid making copies of column vectors by default

because it indicated you think `dropmissing!` is doing copies now (and it only copies `cols` that have to change type which is unavoidable, but we can avoid resizing them earlier as you have noted).

bkamins:
OK to merge?

bkamins:
This is rebased against recently merged PRs and should be good for a final check before merging. Thank you.

bkamins:
Thank you! I will review it and merge if I do not spot any problems in the PR.

PR #1722, opened 2019-02-16
Improve explanation how join works with duplicates

bkamins:
Following the discussion on Slack.

### DIFF
--- a/docs/src/man/joins.md
+++ b/docs/src/man/joins.md
@@ -217,3 +217,5 @@ julia> join(a, b, on = [:City, :Job])
 │ 9   │ New York  │ Doctor │ 5        │ e      │
 
 ```
+
+Additionally, notice that in the last join rows 2 and 3 had the same values on `on` variables in both joined `DataFrame`s. In such a situation `:inner`, `:outer`, `:left` and `:right` joins will produce all combinations of matching rows. In our example rows from 2 to 5 were created as a result. The same behavior can be observed for rows 4 and 5 in both joined `DataFrame`s.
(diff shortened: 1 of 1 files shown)


bkamins:
@oxinabox - I have passed @nalimilan biological CI 😄. If the description is clear for you we can merge it.

oxinabox:
I think the explaination should be cross-referenced with the `validate` kwarg.
Either referencing back to it, or it referencing forward to the example.

bkamins:
Thank you. Actually the documentation was more outdated than I thought. I have cleaned it up a bit in general.

docs/src/man/joins.md
```diff
@@ -219,3 +200,41 @@ julia> join(a, b, on = [:City, :Job])
...
+
+```jldoctest joins
+julia> join(a, b, on = [(:City, :Location), (:Job, :Work)], validate=(true, true))
+ERROR: ArgumentError: Merge key(s) are not unique in both df1 and df2. First duplicate in df1 at 3. First duplicate in df2 at 3
+```
+
+Finally, using `indicator` keyword argument you can add a column to the resulting data frame giving an information if the given row appeared only in the left, the right or both data frames. Here is an example:
+
+```jldoctest joins
+julia> a = DataFrame(ID = [20, 40], Name = ["John", "Jane"])
```
nalimilan: Can't we use the existing data frames here?

bkamins: No, because in the existing data frames all keys match so we would see `"both"` everywhere.

PR #1723, opened 2019-02-17
Fix two-argument show for DataFrameRow

nalimilan:
Spotted at https://discourse.julialang.org/t/how-to-show-all-columns-with-dataframerow/20883.

### DIFF
--- a/test/dataframerow.jl
+++ b/test/dataframerow.jl
@@ -304,6 +304,18 @@ end
 end
 
 @testset "show" begin
+    function capture_stdout(f::Function)
+        oldstdout = stdout
+        rd, wr = redirect_stdout()
+        f()
+        str = String(readavailable(rd))
+        redirect_stdout(oldstdout)
+        size = displaysize(rd)
+        close(rd)
+        close(wr)
+        str, size
+    end
+
     df = DataFrame(a=nothing, b=1)
 
     @test sprint(show, DataFrameRow(df, 1, :)) == """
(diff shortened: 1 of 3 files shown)


test/dataframerow.jl
```diff
@@ -304,6 +304,18 @@ end
 end
 
 @testset "show" begin
+    function capture_stdout(f::Function)
+        oldstdout = stdout
+        rd, wr = redirect_stdout()
+        f()
+        str = String(readavailable(rd))
```
bkamins: I would write it as:

```
        redirect_stdout(oldstdout)
        size = displaysize(rd)
        close(wr)
        str = read(rd, String)
        close(rd)
```
as it will be non-blocking when `f` produces no output to `stdout`. Probably a minor thing.

nalimilan: TBH I've just copied the code from show.jl.

bkamins: This is a minor thing, but I just wanted to make sure you know the problem with the current implementation (you can check that it is blocking if you e.g. run `capture_stdout(()->nothing)`).

So maybe let us fix it in both places (as users might copy the code from DataFrames.jl for different purposes thinking it is a well written package 😄).

nalimilan: OK, I'm quite bad at streaming so I've just adopted your code. :-)

bkamins: Actually streaming in Julia is rather tricky from my experience but what I have proposed seems to work better.

PR #1735, opened 2019-02-23
Improve append!

bkamins:
This follows the comment from @nalimilan here https://discourse.julialang.org/t/best-way-to-iteratively-add-to-a-dataframe/21084/8.

Changes:
* allow to `append!` a data frame with 0 columns to an existing data frame (and this is a no-op)
* allow to `append!` to a `DataFrame` with 0 columns; in this case new columns are added (this is an exception) with freshly allocated column data; in this case it is problematic what should be the exact type of `AbstractVector` to create (this is why this change is slightly problematic); I have decided to recommend creating a `Vector` in this case

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1076,7 +1076,7 @@ julia> df1
 function Base.append!(df1::DataFrame, df2::AbstractDataFrame)
     if ncol(df1) == 0
         for (n, v) in eachcol(df2, true)
-            df1[n] = collect(v) # make sure df1 contains Vector-s
+            df1[n] = copy(v) # make sure df1 does not reuse df2
         end
         return df1
     end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1034,7 +1034,7 @@ end
 Add the rows of `df2` to the end of `df1`.
 
 Column names must be equal (including order), with the following exceptions:
-* If `df1` has no columns then newly allocated `Vector`s representing data in all
+* If `df1` has no columns then copies of
   columns from `df2` are added to it.
 * If `df2` has no columns then calling `append!` leaves `df1` unchanged.
 
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1033,7 +1033,11 @@ end
 
 Add the rows of `df2` to the end of `df1`.
 
-Column names must be equal (including order).
+Column names must be equal (including order), with the following exceptions:
+* If `df1` has no columns then newly allocated `Vector`s representing data in all
+  columns from `df2` are added to it.
+* If `df2` has no columns then calling `append!` leaves `df1` unchanged.
+
 Values corresponding to new rows are appended in-place to the column vectors of `df1`.
 Column types are therefore preserved, and new values are converted if necessary.
 An error is thrown if conversion fails: this is the case in particular if a column
@@ -1070,6 +1074,14 @@ julia> df1
 ```
 """
 function Base.append!(df1::DataFrame, df2::AbstractDataFrame)
+    if ncol(df1) == 0
+        for (n, v) in eachcol(df2, true)
+            df1[n] = collect(v) # make sure df1 contains Vector-s
+        end
+        return df1
+    end
+    ncol(df2) == 0 && return df1
+
     _names(df1) == _names(df2) || error("Column names do not match")
     nrows, ncols = size(df1)
     try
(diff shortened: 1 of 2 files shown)


src/dataframe/dataframe.jl
```diff
@@ -1070,6 +1074,14 @@ julia> df1
 ```
 """
 function Base.append!(df1::DataFrame, df2::AbstractDataFrame)
+    if ncol(df1) == 0
+        for (n, v) in eachcol(df2, true)
+            df1[n] = collect(v) # make sure df1 contains Vector-s
```
nalimilan: This makes a copy even when not necessary.

bkamins: I thought the opposite. Given column "ownership" contract we always have to make a copy. When do you think copy could be dropped?

bkamins:
> we should probably have a look at what other implementations do in this situation

what implementations do you have in mind (the issue is that we are fundamentally different from other implementations because `DataFrame` has a mutable schema, while other implementations I know assume that table schema is immutable).

src/dataframe/dataframe.jl
nalimilan: Ah, sorry, of course here we create new column vectors so we need to make the copy here. But I don't really understand your comment in the OP about `AbstractVector`: why not just use the type of `v` (i.e. use `copy(v)`)?

nalimilan:
I mean the usual suspects: R, dplyr and Pandas.

src/dataframe/dataframe.jl
bkamins: I was considering `copy`. The only downside is that for non standard `AbstractVectors` you do not know what you will get. E.g. if the source data frame contains `SVector`s then `copy` returns an `SVector` and `SVector` cannot be resized (and usually if you run `append!` you probably want to do it several times). `collect` guarantees we get a `Vector`.

But I agree that the choice here is not clear cut and we could use `copy` (and actually this is the problem with allowing `appending!` to an empty `DataFrame`).

bkamins:
Pandas:
```
>>> df = pd.DataFrame([[1, 2], [3, 4]], columns=list('AB'))
>>> pd.DataFrame()
Empty DataFrame
Columns: []
Index: []
>>> pd.DataFrame().append(df)
   A  B
0  1  2
1  3  4
```

In R:
```
>>> df = pd.DataFrame([[1, 2], [3, 4]], columns=list('AB'))
>>> pd.DataFrame()
Empty DataFrame
Columns: []
Index: []
>>> pd.DataFrame().append(df)
   A  B
0  1  2
1  3  4
>>> pd.DataFrame().append(df).append(pd.DataFrame())
   A  B
0  1  2
1  3  4
```

R:
```
> rbind(data.frame(), data.frame(a=1,b=2), data.frame(), data.frame(a=3,b=4))
  a b
1 1 2
2 3 4
```

So they both work as we assume (i.e. an empty data frame is "ignored")

src/dataframe/dataframe.jl
nalimilan: `collect` wouldn't give the right result for `CategoricalArray`. We could use `copyto!(Tables.allocatecolumn(eltype(v), length(v)), v)`, but it could lose information contained by the array. Since AFAIK DataFrames hasn't really been tested with static arrays, I'd go for the simpler `copy` for now.

bkamins: OK - fixed.

src/dataframe/dataframe.jl
```diff
@@ -1033,7 +1033,11 @@ end
 
 Add the rows of `df2` to the end of `df1`.
 
-Column names must be equal (including order).
+Column names must be equal (including order), with the following exceptions:
+* If `df1` has no columns then newly allocated `Vector`s representing data in all
```
```suggestion
* If `df1` has no columns then copies of
```

test/dataframe.jl
```diff
@@ -546,13 +546,27 @@ df = DataFrame(A = 1:10, B = 'A':'J')
...
-df2 = DataFrame(A=1:4, B = 1:4)
-@test append!(df, DataFrame(A = 3:4, B = [3.0, 4.0])) == df2
-@test_throws InexactError append!(df, DataFrame(A = 3:4, B = [3.5, 4.5]))
-@test df == df2
-@test_throws MethodError append!(df, DataFrame(A = 3:4, B = ["a", "b"]))
-@test df == df2
+
+@testset "append!" begin
+    df = DataFrame(A = 1:2, B = 1:2)
+    df2 = DataFrame(A=1:4, B = 1:4)
```
```suggestion
    df2 = DataFrame(A = 1:4, B = 1:4)
```

bkamins:
This should be good to merge.

PR #1737, opened 2019-02-23
improve push! and document it

bkamins:
Following https://discourse.julialang.org/t/best-way-to-iteratively-add-to-a-dataframe/21084/9 by @nalimilan .
At first I was reluctant to add it, but in the end I think it does not hurt to have it (I do not see a scenario when this is risky).

I have also added documentation for `push!`. I have split two cases of `push!` into two comments as this is really a different functionality and describing it together might be confusing.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1090,9 +1090,9 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
 
 Base.convert(::Type{DataFrame}, d::AbstractDict) = DataFrame(d)
 
-function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns::Symbol=:exact)
-    if !(columns in (:exact, :intersect))
-        throw(ArgumentError("`columns` keyword argument value must be `:exact` or `:intersect`"))
+function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns::Symbol=:equal)
+    if !(columns in (:equal, :intersect))
+        throw(ArgumentError("`columns` keyword argument must be `:equal` or `:intersect`"))
     end
     if ncol(df) == 0 && row isa NamedTuple
         for (n, v) in pairs(row)
@@ -1101,7 +1101,8 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns
         return df
     end
     i = 1
-    if columns == :exact && length(row) != size(df, 2)
+    # Only check for equal lengths, as an error will be thrown below if some names don't match
+    if columns === :equal && length(row) != size(df, 2)
         # TODO: add tests for this case after the deprecation period
         Base.depwarn("In the future push! will require that `row` has the same number" *
                       "of elements as is the number of columns in `df`." *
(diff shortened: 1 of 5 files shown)


src/dataframe/dataframe.jl
```diff
@@ -1115,7 +1161,53 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple})
     df
 end
 
-# array and tuple like collections
+"""
+    push!(df1::DataFrame, iterable::Any)
```
nalimilan: Could you merge this docstring with the one above (or rather move the one above here since this method is more general)? It would be easier to read as it would avoid repetitions. Also, `DataFrameRow` is more similar to `NamedTuple` than a random iterable.

src/dataframe/dataframe.jl
```diff
@@ -1090,14 +1090,60 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
...
+julia> df = DataFrame(A=1:3, B=1:3)
+3×2 DataFrame
+│ Row │ A     │ B     │
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 1     │
+│ 2   │ 2     │ 2     │
+│ 3   │ 3     │ 3     │
 
+julia> push!(df, (C="something", A=true, B=false))
```
nalimilan: I feel like we should deprecate this behavior and require setting an argument. What do you think?

src/dataframe/dataframe.jl
```diff
@@ -1090,14 +1090,61 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
...
+│ 2   │ 2     │ 2     │
+│ 3   │ 3     │ 3     │
+│ 4   │ 1     │ 0     │
+│ 5   │ 1     │ 2     │
+```
+"""
 function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple})
+    if ncol(df) == 0 && row isa NamedTuple
+        for (n, v) in pairs(row)
+            setproperty!(df, n, [v])
```
nalimilan: Probably better use `Tables.allocatecolumn`, which matters if you get a `CategoricalValue`.

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl
bkamins: Fixed (but please check it 😄)

src/dataframe/dataframe.jl

bkamins:
This should be good to have a look at.

src/dataframe/dataframe.jl
```diff
@@ -1090,15 +1090,19 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
...
-function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple})
+function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; exact::Bool=true)
+    if ncol(df) == 0 && row isa NamedTuple
+        for (n, v) in pairs(row)
+            setproperty!(df, n, fill!(Tables.allocatecolumn(typeof(v), 1), v))
+        end
+        return df
+    end
     i = 1
+    if exact && length(row) != size(df, 2)
```
nalimilan: How about using `columns=:equal` and `columns=:intersect` as in https://github.com/JuliaData/DataFrames.jl/pull/1659? Also, it's not just the number of elements that has to match: names must be the same too.

src/dataframe/dataframe.jl
```diff
@@ -1115,7 +1119,88 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple})
...
 
-# array and tuple like collections
+"""
+    push!(df1::DataFrame, iterable::Any)
+    push!(df::DataFrame, row::Union{AbstractDict, NamedTuple, DataFrameRow}; exact::Bool=true)
+
+Add in-place one row at the end of `df1` taking the values from the iterable `iterable`
+or `row`.
+
+Adding values from `iterable` is based on column number.
```
nalimilan: Instead of using two different names to refer to the same argument, I'd say "if `row` is a ...; else ...".

src/dataframe/dataframe.jl
bkamins: OK - fixed

src/dataframe/dataframe.jl
bkamins: Fixed

src/dataframe/dataframe.jl
bkamins: > Also, it's not just the number of elements that has to match: names must be the same too.

We do not have to check for this. If names do not match later the code will fail. We could add an additional test here but it will:
* slow down the run time if no error happens;
* speed up the run time if error happens.

(I thought it is better to be fast in no-error case)

bkamins:
@nalimilan - thank you.

PR #1739, opened 2019-02-25
Add compress argument to categorical!()

### LINKED ISSUE #1733: Add compress argument to categorical!()
It would be really nice to be able to compress categorical columns e.g.:

```julia
julia> import DataFrames, CategoricalArrays

julia> function DataFrames.categorical!(df::DataFrame, cname::Union{Integer, Symbol}, compress::Bool)
           df[cname] = categorical(df[cname], compress)
           df
       end

julia> df = DataFrame(a = ["A", "A", "B", "B"]);

julia> categorical!(df, :a, true);

julia> typeof(df[:a])
CategoricalArray{String,1,UInt8,String,CategoricalString{UInt8},Union{}}
```

harryscholes:
I have made `categorical!` add compression by default. Under the hood `categorical!` calls `CategoricalArrays.categorical`, which selects the smallest `Unsigned` type to represent the pool of categorical elements.

```julia
julia> using DataFrames

julia> df = DataFrame(a = 1:typemax(UInt8));

julia> categorical!(df, :a);

julia> eltypes(df)
1-element Array{DataType,1}:
 CategoricalValue{Int64,UInt8}

julia> df = DataFrame(a = 1:typemax(UInt8)+1);

julia> categorical!(df, :a);

julia> eltypes(df)
1-element Array{DataType,1}:
 CategoricalValue{Int64,UInt16}

julia> df = DataFrame(a = 1:typemax(UInt16)+1);

julia> categorical!(df, :a);

julia> eltypes(df)
1-element Array{DataType,1}:
 CategoricalValue{Int64,UInt32}
```

I also updated the categorical docs and tests.

Closes #1733

### DIFF
--- a/docs/src/man/categorical.md
+++ b/docs/src/man/categorical.md
@@ -135,7 +135,7 @@ julia> eltypes(df)
  CategoricalString{UInt32}
  String
 
-julia> categorical!(df, compress=true) # change all columns to be categorical with compression
+julia> categorical!(df, compress=true) # change all columns with eltype <: AbstractString to be categorical with compression
 6×2 DataFrame
 │ Row │ A            │ B            │
 │     │ Categorical… │ Categorical… │
--- a/docs/src/man/categorical.md
+++ b/docs/src/man/categorical.md
@@ -77,7 +77,8 @@ julia> sort(cv)
 
 ```
 
-By default, a `CategoricalArray` is able to represent 2<sup>32</sup> different levels. You can use less memory by calling the `compress` function, or by calling the `categorical` function with the `compress` argument set to `true`:
+By default, a `CategoricalArray` is able to represent 2<sup>32</sup> different levels. You
+can use less memory by calling the `compress` function:
 
 ```jldoctest categorical
 julia> cv = compress(cv)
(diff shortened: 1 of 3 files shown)


docs/src/man/categorical.md
```diff
@@ -77,23 +77,33 @@ julia> sort(cv)
 
 ```
 
-By default, a `CategoricalArray` is able to represent 2<sup>32</sup>differents levels. You can use less memory by calling the `compress` function:
+By default, a `CategoricalArray` is able to represent 2<sup>32</sup> different levels. You can use less memory by calling the `compress` function, or by calling the `categorical` function with the `compress` argument set to `true`:
```
bkamins: Can you please break this line as it seems overlong?
Also I would add that `compess` is a positional argument for clarity.

nalimilan:
Thanks. I think it would be better to make `compress` a keyword argument though. It's positional in CategoricalArrays only because it avoids a type unstability in the standard case (and even that would maybe be changed thanks to constant propagation in recent Julia versions). Here we don't care about that since data frames are type unstable already. Making it a keyword argument is more explicit and it also works when not providing a column name.

bkamins:
@nalimilan: do you have opinion if `compress` should be `true` or `false` by default?

nalimilan:
`false` sounds safer.

harryscholes:
Thanks for your comments. I've now made `compress` a keyword argument set to `false` by default.

docs/src/man/categorical.md
```diff
@@ -129,11 +131,11 @@ julia> categorical!(df, :A) # change the column `:A` to be categorical
...
 
 julia> eltypes(df)
-2-element Array{Type,1}:
- CategoricalArrays.CategoricalString{UInt32}
+2-element Array{DataType,1}:
+ CategoricalString{UInt32}
  String
 
-julia> categorical!(df) # change all columns to be categorical
+julia> categorical!(df, compress=true) # change all columns to be categorical with compression
```
bkamins: not all columns but only columns with `eltype` that is `AbstractString`.

harryscholes:
I can't find `compress!` in DataFrames.jl or CategoricalArrays.jl. It might be best if you polish up the docs how you see fit!

bkamins:
> I can't find `compress!` in DataFrames.jl or CategoricalArrays.jl.

Because it was not present 😞. I will add it in a separate PR.

bkamins:
Merged. Thank you!

PR #1741, opened 2019-03-02
document nrow and ncol

bkamins:
Clean up and enable `nrow` and `ncol` documentation.
My current understanding is that in the end we tend to want to keep them in 1.0.


bkamins:
I was thinking about separate docstrings, but came to the conclusion that having them in one place and also with `size` reference is a more user friendly approach.

PR #1743, opened 2019-03-04
Remove deepcopy

nalimilan:
It does not appear to be needed and other packages generally don't define it.


nalimilan:
For historical reference, `deepcopy` was added at https://github.com/JuliaData/DataFrames.jl/commit/c2ec6dea3613a98a42c15f978ffac4e5111fb70c because of changes in the function names: `shallowcopy` became `copy`, so `copy` had to become `deepcopy`. But actually that method wasn't/isn't needed.

PR #1746, opened 2019-03-07
change sorting documentation

### LINKED ISSUE #1745: Documentation says that a test/data/iris.csv file should exist, it doesnt.
http://juliadata.github.io/DataFrames.jl/stable/man/sorting.html

```jl
julia> using DataFrames, CSV

julia> iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../test/data/iris.csv"));
ERROR: ArgumentError: "C:\Users\Kristoffer\.julia\packages\DataFrames\IKMvt\src\../test/data/iris.csv" is not a valid file
Stacktrace:
 [1] #File#1(::Int64, ::Bool, ::Int64, ::Nothing, ::Int64, ::Nothing, ::Bool, ::Nothing, ::Bool, ::Array{String,1}, ::String, ::String, ::Bool, ::Char, ::Nothing, ::Nothing, ::Char, ::Nothing, ::Nothing, ::Nothing, ::Nothing, ::Nothing, ::Dict{Type,Type}, ::Symbol, ::Bool, ::Bool, ::Bool, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Type, ::String) at C:\Users\Kristoffer\.julia\packages\CSV\eWuJV\src\CSV.jl:135
```

bkamins:
I have dropped the dependency on the CSV.jl package and made the example smaller to allow the user to concentrate on the key issues when doing sorting. + I have added some small comments.

Fixes #1745


KristofferC:
http://juliadata.github.io/DataFrames.jl/stable/man/split_apply_combine.html also uses CSV

KristofferC:
Personal opinion, but I feel the docs are a little bit more fun to read when you have a dataset that could be real, and not just columns of random numbers with made up names.

bkamins:
Thank you.

This is a fair point (I prefer minimal examples showing exactly what we want to show, but this is a personal taste). E.g. the old example actually did not show what it claimed it did unfortunately, because the data set did not fit the use case.

@nalimilan - any opinion?

If we decide to use a real data set I will update all docs accordingly and select a good data set.

nalimilan:
I agree, why not keep the Iris dataset? We could use `RDatasets` to load it, IIRC we moved to CSV when RDatasets didn't support DataTables or something.

bkamins:
I have reverted iris.csv (but I have placed it in /assets in documentation part of the code).

@nalimilan We need to fix split-apply-combine docs also, but this is a separate PR (see https://github.com/JuliaData/DataFrames.jl/issues/1747).

docs/src/man/sorting.md
```diff
@@ -3,177 +3,183 @@
 Sorting is a fundamental component of data analysis. Basic sorting is trivial: just calling `sort!` will sort all columns, in place:
 
 ```jldoctest sort
-julia> using DataFrames, CSV
-
-julia> iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../test/data/iris.csv"));
+julia> using DataFrames, CSV,
```
nalimilan:
```suggestion
julia> using DataFrames, CSV
```

docs/src/man/sorting.md
```diff
@@ -3,177 +3,183 @@
...
+│     │ Float64⍰    │ Float64⍰   │ Float64⍰    │ Float64⍰   │ String⍰         │
+├─────┼─────────────┼────────────┼─────────────┼────────────┼─────────────────┤
+│ 1   │ 5.1         │ 2.5        │ 3.0         │ 1.1        │ Iris-versicolor │
+│ 2   │ 5.0         │ 2.0        │ 3.5         │ 1.0        │ Iris-versicolor │
+│ 3   │ 5.0         │ 2.3        │ 3.3         │ 1.0        │ Iris-versicolor │
+│ 4   │ 4.9         │ 2.4        │ 3.3         │ 1.0        │ Iris-versicolor │
 ```
 
-Keywords used above include `rev` (to sort a column or the whole `DataFrame` in reverse), and `by` (to apply a function to a column/`DataFrame`). Each keyword can either be a single value, or can be a tuple or array, with values corresponding to individual columns.
+Keywords used above include `rev` (to sort a column or the whole `DataFrame` in reverse),
```
nalimilan: The existing text could be confusing: this doesn't sort a column in isolation.
```suggestion
Keywords used above include `rev` (to sort in reverse),
```

docs/src/man/sorting.md
```diff
@@ -3,177 +3,183 @@
...
+and `by` (to apply a function to rows of a column/`DataFrame`).
```
nalimilan:
```suggestion
and `by` (to apply a function to values before comparing them).
```

docs/src/man/split_apply_combine.md
```diff
@@ -20,14 +20,14 @@ The kind of return value and the number and names of columns must be the same fo
...
 
-The name for the resulting column can be chosen either by passing a named tuple of pairs, or by returning a named tuple or a data frame. If no name is provided, it is generated automatically. For functions taking a single column (first form), the input column name is concatenated with the function name: for standard functions like `mean` this will produce columns with names like `SepalLength_mean`; for anonymous functions like `x -> sqrt(x)^e`, the produced columns will be `SepalLength_function`. For functions taking multiple columns (second form), names are `x1`, `x2`, etc.
+The name for the resulting column can be chosen either by passing a named tuple of pairs, or by returning a named tuple or a data frame. If no name is provided, it is generated automatically. For functions taking a single column (first form), the input column name is concatenated with the function name: for standard functions like `mean` this will produce columns with names like `SepalLength_mean`; for anonymous functions like `x -> sqrt(x)`, the produced columns will be `SepalLength_function`. For functions taking multiple columns (second form), names are `x1`, `x2`, etc.
```
nalimilan: Something like `x -> x+1` could be better than `x -> sqrt(x)`, as you can just pass `sqrt` directly.

bkamins: right (the old code was currently failing)

Issue #1747, opened 2019-03-07
Inconsistency in column naming by by


bkamins:
See:
```
julia> by(iris, :Species, :PetalLength => mean)
3×2 DataFrame
│ Row │ Species         │ PetalLength_Statistics.mean │
│     │ String⍰         │ Float64                     │
├─────┼─────────────────┼─────────────────────────────┤
│ 1   │ Iris-setosa     │ 1.464                       │
│ 2   │ Iris-versicolor │ 4.26                        │
│ 3   │ Iris-virginica  │ 5.552                       │

julia> by(iris, :Species, :PetalLength => mean, :PetalLength => sum)
3×3 DataFrame
│ Row │ Species         │ PetalLength_mean │ PetalLength_sum │
│     │ String⍰         │ Float64          │ Float64         │
├─────┼─────────────────┼──────────────────┼─────────────────┤
│ 1   │ Iris-setosa     │ 1.464            │ 73.2            │
│ 2   │ Iris-versicolor │ 4.26             │ 213.0           │
│ 3   │ Iris-virginica  │ 5.552            │ 277.6           │
```

The generated column name depends on the path taken (if we produce one or more columns).

PR #1748, opened 2019-03-08
fix names of functions from Statistics in by and upgrade documentation

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1747 (we should use `nameof` instead of `Symbol` to get **name of** function 😄).

I have also updated docstrings where they were missing outputs.

After this PR is OK and gets merged the next step is to enable checking docstrings by Documenter (a separate PR).


bkamins:
How do we go about this PR?
Should I merge it and later you rebase #1750? Or I should wait till you merge #1750 and then rebase this one?

nalimilan:
I've merged #1750 and rebased this PR, that sounds more logical than fixing something just to remove it. That way this PR is focused on improving docs.

bkamins:
OK - so do we marge it after it passes the CI?

PR #1750, opened 2019-03-12
Remove unnecessary funname methods

nalimilan:
We don't actually need to call these functions since we always have access to the original name of reductions.

This is a more radical fix for #1747.


src/groupeddataframe/grouping.jl
```diff
@@ -666,7 +659,7 @@ end
 function _combine(f::Any, gd::GroupedDataFrame)
     if f isa Pair{<:Union{Symbol,Integer}}
         incols = gd.parent[first(f)]
-        fun = check_aggregate(last(f))
+        fun = last(f)
```
nalimilan: `check_aggregate` is called below.

PR #1752, opened 2019-03-26
End deprecation period of eachcol

bkamins:
I had to open a new PR for this (the old one was https://github.com/JuliaData/DataFrames.jl/pull/1613).
It is now rebased to the current master.

### DIFF
--- a/docs/src/lib/types.md
+++ b/docs/src/lib/types.md
@@ -42,11 +42,10 @@ Similarly, the `eachcol` function returns a value of the `DataFrameColumns` type
 serves as an iterator over columns of an `AbstractDataFrame`.
 The return value can have two concrete types:
 
-* If the `eachcol` function is called with the `names` argument set to `true` (currently the default,
-  but in the future the default will change to `false`) then it returns a value of the
+* If the `eachcol` function is called with the `names` argument set to `true` (then it returns a value of the
   `DataFrameColumns{<:AbstractDataFrame, Pair{Symbol, AbstractVector}}` type, which is an
   iterator returning a pair containing the column name and the column vector.
-* If the `eachcol` function is called with `names` argument set to `false` then it returns a value of the
+* If the `eachcol` function is called with `names` argument set to `false` (the default) then it returns a value of the
   `DataFrameColumns{<:AbstractDataFrame, AbstractVector}` type, which is an
   iterator returning the column vector only.
 
--- a/docs/src/lib/types.md
+++ b/docs/src/lib/types.md
@@ -42,7 +42,7 @@ Similarly, the `eachcol` function returns a value of the `DataFrameColumns` type
 serves as an iterator over columns of an `AbstractDataFrame`.
 The return value can have two concrete types:
 
-* If the `eachcol` function is called with the `names` argument set to `true` (then it returns a value of the
+* If the `eachcol` function is called with the `names` argument set to `true` then it returns a value of the
   `DataFrameColumns{<:AbstractDataFrame, Pair{Symbol, AbstractVector}}` type, which is an
   iterator returning a pair containing the column name and the column vector.
 * If the `eachcol` function is called with `names` argument set to `false` (the default) then it returns a value of the
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -204,7 +204,7 @@ eltypes(df)
 ```
 
 """
-eltypes(df::AbstractDataFrame) = eltype.(columns(df))
+eltypes(df::AbstractDataFrame) = eltype.(eachcol(df, false))
 
 Base.size(df::AbstractDataFrame) = (nrow(df), ncol(df))
 function Base.size(df::AbstractDataFrame, i::Integer)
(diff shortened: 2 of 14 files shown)


src/abstractdataframe/iteration.jl
```diff
@@ -125,7 +124,7 @@ julia> collect(eachcol(df, true))
  :x => [1, 2, 3, 4]
  :y => [11, 12, 13, 14]
 
-julia> collect(eachcol(df, false))
+julia> collect(eachcol(df))
```
nalimilan: Maybe put the example with `eachcol(df, true)` at the end, since it's not the default?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -204,7 +204,7 @@ eltypes(df)
 ```
 
 """
-eltypes(df::AbstractDataFrame) = eltype.(columns(df))
+eltypes(df::AbstractDataFrame) = eltype.(eachcol(df, false))
```
```suggestion
eltypes(df::AbstractDataFrame) = eltype.(eachcol(df))
```
Same elsewhere.

src/abstractdataframe/iteration.jl
```diff
@@ -99,13 +99,12 @@ struct DataFrameColumns{T<:AbstractDataFrame, V} <: AbstractVector{V}
...
-    eachcol(df::AbstractDataFrame, names::Bool=true)
+    eachcol(df::AbstractDataFrame, names::Bool=false)
 
 Return a `DataFrameColumns` that iterates an `AbstractDataFrame` column by column.
-If `names` is equal to `true` (currently the default, in the future the default
-will be set to `false`) iteration returns a pair consisting of column name
+If `names` is equal to `true` iteration returns a pair consisting of column name
 and column vector.
-If `names` is equal to `false` then column vectors are yielded.
+If `names` is equal to `false` (the default) then only column vectors are yielded.
```
nalimilan: Put this sentence first (mentioning iteration in this one rather than in the one for `true`).

src/abstractdataframe/abstractdataframe.jl
bkamins: I left `false` everywhere in the source codes although it is the default to make it more explicit.
Do you prefer to drop it everywhere?

src/abstractdataframe/iteration.jl

src/abstractdataframe/iteration.jl

bkamins:
@nalimilan I have dropped `, false` in `eachcol` everywhere expect where it was tested to work correctly. This should be good to merge now. Then I can rebase "ownership" PR with this change.

docs/src/lib/types.md
```diff
@@ -42,11 +42,10 @@ Similarly, the `eachcol` function returns a value of the `DataFrameColumns` type
 serves as an iterator over columns of an `AbstractDataFrame`.
 The return value can have two concrete types:
 
-* If the `eachcol` function is called with the `names` argument set to `true` (currently the default,
-  but in the future the default will change to `false`) then it returns a value of the
+* If the `eachcol` function is called with the `names` argument set to `true` (then it returns a value of the
```
```suggestion
* If the `eachcol` function is called with the `names` argument set to `true` then it returns a value of the
```
Also, better present the default first.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -244,7 +244,7 @@ that is different than the number of rows present in `df`.
 """
 function Base.similar(df::AbstractDataFrame, rows::Integer = size(df, 1))
     rows < 0 && throw(ArgumentError("the number of rows must be non-negative"))
-    DataFrame(AbstractVector[similar(x, rows) for x in columns(df)], copy(index(df)))
+    DataFrame(AbstractVector[similar(x, rows) for x in eachcol(df, false)], copy(index(df)))
```
```suggestion
    DataFrame(AbstractVector[similar(x, rows) for x in eachcol(df)], copy(index(df)))
```

src/abstractdataframe/iteration.jl
```diff
@@ -120,48 +119,37 @@ julia> df = DataFrame(x=1:4, y=11:14)
...
-
-julia> map(eachcol(df, false)) do col
+julia> map(eachcol(df)) do col
            maximum(col) - minimum(col)
        end
 2-element Array{Int64,1}:
  3
  3
+
+julia> sum.(eachcol(df, false))
```
```suggestion
julia> sum.(eachcol(df))
```

test/indexing.jl
```diff
@@ -6,12 +6,12 @@ using Test, DataFrames
     df = DataFrame(a=1:3, b=4:6, c=7:9)
 
     @test df[1] == [1, 2, 3]
-    @test df[1] === eachcol(df, false)[1]
+    @test df[1] === eachcol(df)[1]
     @test df[1:2] == DataFrame(a=1:3, b=4:6)
-    @test eachcol(df[1:2], false)[1] === eachcol(df, false)[1]
+    @test eachcol(df[1:2])[1] === eachcol(df)[1] === eachcol(df)[1]
```
nalimilan:
```suggestion
    @test eachcol(df[1:2])[1] === eachcol(df)[1] === eachcol(df, true)[1]
```

test/indexing.jl
```diff
@@ -6,12 +6,12 @@ using Test, DataFrames
...
     @test df[:] == df
     @test df[:] !== df
-    @test eachcol(df[:], false)[1] === eachcol(df, false)[1]
+    @test eachcol(df[:])[1] === eachcol(df)[1] === eachcol(df)[1]
```
nalimilan:
```suggestion
    @test eachcol(df[:])[1] === eachcol(df)[1] === eachcol(df, true)[1]
```

test/indexing.jl
bkamins: I will add it to the batch so that it is not pending, but this will fail (I will later fix it so that it does not fail)

bkamins:
Thank you for finding more, 😄  - this was exactly my problem. I have grep-ed them, but there were so many places to fix that I must have missed something (as `, false` could appear also in other contexts)

Issue #1754, opened 2019-03-27
Building DataFrames.jl on windows requires admin privileges


felipenoris:
This happens due to CodecZlib.jl .

See https://github.com/bicycle1885/CodecZlib.jl/issues/43 .

bkamins:
@nalimilan - I have wanted to remove `readtable` and `writetable` form the package sources for some time now. This is yet another argument for it (and it should make DataFrames.jl have lighter dependencies in general + load a bit faster on startup).

nalimilan:
We could drop `readtable`, but it sounds quite easy to drop the additional dependencies, so... See https://github.com/JuliaData/DataFrames.jl/pull/1755.

davidanthoff:
Shouldn't this just be fixed in BinaryProvider? There seems to be an issue already: https://github.com/JuliaPackaging/BinaryProvider.jl/issues/149.

bkamins:
Yes - this is a long known issue. I simply think that we do not need these dependencies in DataFrames.jl anyway (in general my view is that DataFrames.jl should be 100% Julia including all its dependencies).

bkamins:
Just to expand on for possible future readers of this thread:

> DataFrames.jl should be 100% Julia including all its dependencies

We want to be sure that when DataFrames.jl fails badly (e.g. core dumps) this must mean only one of the following two reasons:
* there is a bug in Base
* an unsafe operation in-built into the language (like `@inbounds`) was used

so that the user can exclude the possible cause that e.g. some third-party C code used by DataFrames.jl has a bug.
This is in practice very relevant as columns of `DataFrame` can have a myriad of sources. People often report that DataFrames.jl fails. As of now - when we exclude the two points above by doing code analysis of DataFrames.jl functions only we can safely say that the bug is not in DataFrames.jl.

PR #1755, opened 2019-03-27
Remove dependency on CodecZlib and TranscodingStreams

nalimilan:
This avoids requiring admin privileges on Windows. Without these, `readtable` no longer supports .csv.gz files, but it's deprecated.

Fixes https://github.com/JuliaData/DataFrames.jl/issues/1754.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,8 +6,6 @@ SortingAlgorithms
 Reexport
 WeakRefStrings 0.4.0
 DataStreams 0.3.0
-CodecZlib 0.4
-TranscodingStreams
 Compat 0.59.0
 Tables 0.1.15
 IteratorInterfaceExtensions 0.1.1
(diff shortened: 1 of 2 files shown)


bkamins:
You also need to remove line
https://github.com/JuliaData/DataFrames.jl/pull/1755/files#diff-f703c426be50b4bbe69472caef0aa2feR25

davidanthoff:
What should users that depend on this functionality use now? Might be a good idea to add a pointer to the docs? Because CSV.jl doesn’t support this out of the box, AFAIK?

bkamins:
Could you please open a PR with the suggested wording of the message?
(or write it here and I will make a PR)
Thank you.

davidanthoff:
Well, I don't know what the suggestion would be... I think I would not have removed the functionality at this point, in particular since the problem actually doesn't seem related to admin priv issues on windows.

nalimilan:
Keeping binary dependencies just for a function which has been deprecated ages ago sounds a bit too much. My PR just saved `readtable` for a few more months, the alternative was to remove it altogether.

A bit of googling pointed me to https://github.com/JuliaData/CSV.jl/issues/46, but I'm not sure that's still the recommended solution.

Issue #1757, opened 2019-03-29
Showing unlimited number of columns on Jupyter Notebook?


bicycle1885:
In Jupyter Notebook (or JupyterLab), it seems an unlimited number of columns are shown as an output of a cell while the number of rows is limited. My data table contains 28,065 rows × 21,613 columns, and the horizontal scroll bar is very long and it takes a very long time to reander the table. I think there should be some reasonable cutoff (maybe 30 columns or so?) by default.

bkamins:
This issue keeps popping up. I think we should make `show` methods for HTML and LaTeX in sync with `show` for text/plain (i.e. use the same keyword arguments and have the same behavior).

@nalimilan - any objections. If not I can implement it (also `show` docstrings need some care).

nalimilan:
But how can we decide how many columns will fit in the display? Does Jupyter set `displaysize` so that we can at least do something approximate?

The ideal approach would be to use an interactive table with scrolling and lazy rendering, like TableView.jl (I think there are other packages implementing this). I guess Jupyter could use the Tables.jl interface for that.

bkamins:
We already read number of rows to display from `displaysize` we could also use it to determine number of columns (which typically will be 80).

Do you know who knows enough about interactive tables to give us a good advice how to proceed? (note, that in Jupyter we will still have to decide how to handle LaTeX which is not interactive, as PDF export uses LaTeX)

nalimilan:
OK, go for it then. Anyway it can't hurt if we keep printing everything when `displaysize` isn't defined.

IIRC @davidanthoff was part of the conversation about interactive HTML tables, but I don't remember where it happened.

bkamins:
AFAIK `displaysize` is always defined and defaults to 24x80.

nalimilan:
Ah, right. Fortunately the `:limit` `IOContext` property isn't set by default, so if we fall back to `false` as we do now for text output it should be OK.

bkamins:
Exactly.

davidanthoff:
The story for Jupyter is not simple, at this point... What used to work is that one could embed JavaScript in the HTML MIME bundle and thereby get interactivity into Jupyter Notebooks, but that no longer works on JuptyerLab, so I think that is essentially no longer an option.

[...]

PR #1761, opened 2019-03-30
Initial implementation of limiting number of rows/columns printed in HTML/LaTeX

bkamins:
Follows https://github.com/JuliaData/DataFrames.jl/issues/1757.

Please comment if you like it (as this is pretty subjective) and I will write tests when we settle on the design we like.


bicycle1885:
Thank you for doing that!

I think limiting the size to 30 rows is a little bit too restrictive. The default value of pandas is 60 in my environment (you can see this by `pandas.options.display.max_rows`). Also, setting the same value to the number of columns and rows will often result in a wide display of a table like below (this is a 10x10 view of a table):

image

Thinking that web browsers are good for displaying a long content rather a wide one, I think the default maximum number of rows should be larger (say 60 as per pandas).

nalimilan:
I don't really like the idea of hardcoding a number of rows/columns. Why not use `displaysize` to compute the (possibly approximate) number of rows/columns that fit on the display? I guess we should get in touch with Jupyter and Juno people to know how they want to communicate the appropriate size.

bicycle1885:
I agree with you; gracefully fitted in a display is ideal. However, it is not trivial to achieve that goal because it would depend on many factors such as browsers, fonts, and so on. It may be possible but I think it will take at least weeks or months. However, setting a hard limit to the output is very easy compared to that. Showing an unlimited number of columns is a pressing problem that loses user experience.

bkamins:
> I agree with you; gracefully fitted in a display is ideal. However, it is not trivial to achieve that goal because it would depend on many factors such as browsers, fonts, and so on.

This is my thinking. I really do not know how e.g. you can get information when producing LaTeX output how many rows/columns you are allowed to produce so that they still fit the page.

nalimilan:
I'm not suggesting to try estimating the size the output will take when the HTML is rendered -- that's impossible. But `displaysize` is defined in terms of number of character "columns", so it could make sense to apply the same rules as for the text/REPL printing. Anyway I think we need to get input from the Jupyter or Juno devs.

bkamins:
The problem is that `displaysize` does not contain sensible values (at least given my limited tests).

@stevengj, @pfitzseb: could you comment on the issue please?

What we need is to determine the default output width and height for HTML and LaTeX output when we display a data frame (in Jupyter they are both captured and the latter is used to produce PDF).

pfitzseb:
At least when it comes to Juno `displaysize(stdout)` *does* return something sensbile: The size of the REPL inside of Juno. Limiting the size of the inline display is sensible, but I don't think we need to think too much about it -- once WebIO is stable I'd recommend using [TableView.jl](https://github.com/JuliaComputing/TableView.jl) for HTML output in Juno.

bkamins:
@nalimilan - what do you think we should do with this PR (maybe increase the number of printed rows and columns to 100 - this should be more than any standard display size anyway and still be reasonably fast). 

For LaTeX any default will be probably problematic anyway.

stevengj:
> The problem is that `displaysize` does not contain sensible values (at least given my limited tests).

Should be fixed in IJulia master, due to https://github.com/JuliaLang/IJulia.jl/commit/475dcbe615fd8f0ec50887fa1e56be6f57409403

src/abstractdataframe/io.jl
```diff
@@ -92,16 +99,74 @@ function html_escape(cell::AbstractString)
...
 
-Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame; summary::Bool=true) =
-    _show(io, mime, df, summary=summary)
+"""
+    show(io::IO, mime::MIME, df::AbstractDataFrame;
+         allrows::Bool = !get(io, :limit, true),
+         allcols::Bool = !get(io, :limit, true),
+         summary::Bool=true)
+
+Render a data frame to an I/O stream using `mime`.
```
```suggestion
Render a data frame to an I/O stream in MIME type `mime`.
```

src/abstractdataframe/io.jl
```diff
@@ -92,16 +99,74 @@ function html_escape(cell::AbstractString)
...
+
+# Arguments
+- `io::IO`: The I/O stream to which `df` will be printed.
+- `mime::MIME`: supoorted mime types are: `"text/html"`, `"text/latex"`, `"text/csv"`,
```
```suggestion
- `mime::MIME`: supported MIME types are: `"text/plain"`, `"text/html"`, `"text/latex"`, `"text/csv"`,
```

src/abstractdataframe/io.jl
```diff
@@ -92,16 +99,74 @@ function html_escape(cell::AbstractString)
...
+   `"text/tab-separated-values"`
+- `df::AbstractDataFrame`: The data frame to print.
+- `allrows::Bool `: supported only for HTML and LaTeX mime types;
+   whether to print all rows, rather than a default number defined in
+   `HTML_NROWS` and `LATEX_NROWS` constants. The default value can be overriden by
```
nalimilan: I'd rather not define our custom `IOContext` property, especially if it's a temporary solution. Better use `get(io, :displaysize, HTML_DEFAULT)` (which is different from `displaysize(io)`). That's quite simple to choose the number of rows. For columns it's a bit more tricky: how hard do you think it would be to compute the width in characters needed to print a given number of columns?

src/abstractdataframe/io.jl
```diff
@@ -92,16 +99,74 @@ function html_escape(cell::AbstractString)
...
+
+julia> show(stdout, MIME("text/csv"), DataFrame(A = 1:3, B = ["x", "y", "z"]))
+"A","B"
+1,"x"
+2,"y"
+3,"z"
+```
+"""
+Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame;
+          allrows::Bool = !get(io, :limit, true),
```
nalimilan: Shouldn't we default to `false`? Otherwise we're going to break the use case where you want to print all the data to a file. If Jupyter doesn't set `:limit=>true`, that should be fixed.

bkamins:
Thank you for the feedback and the update of IJulia.

Let me list the things I am not very clear about (note though that I am not an expert in this area):
* we render data frames as HTML (not plain text) in Jupyter Notebook - so the character width is not constant
* in Jupyter Notebook it is not a problem if we do not fit the window (as we can scroll horizontally and vertically) - I would even say that this is desirable, as long as it is not so large that it delays shipment of the output to the screen
* in LaTeX ideally we should have the same as in Jupyter Notebook, but in general of course it is not possible as we have limited page width and height (and here also I am not very clear how we can easily check if something would fit the page)

Therefore my initial proposal was to have fixed number of columns and rows to be printed (note that I am talking in particular about number of columns not a number of characters to print horizontally as we do in REPL), possibly with different defaults for HTML and LaTeX (that is why after some consideration I was thinking about 100 rows and columns for HTML).

I am not overly attached to my proposal (there are bigger and more urgent things to do in DataFrames.jl if we want to hit v1.0 soon), so if someone knows how to get HTML/LaTeX printing right then please either let me know what would be the right approach or just push a PR with this.

My current thinking is that if we do not have a better proposal now we could stick to what I have implemented in this PR as a temporary solution to #1757 (which is problematic, try running `DataFrame(rand(100, 10000))` in Jupyter Notebook to see the lag).

src/abstractdataframe/io.jl
bkamins: > how hard do you think it would be to compute the width in characters needed to print a given number of columns?

This is what I am not sure of (see the post I was writing in parallel). There are two issues:
* we use variable width characters when printing
* we actually do not have to fit the `:displaysize` in Jupyter Notebook: it is OK (and I think desirable) to accept a bit larger output as we support scrolling

src/abstractdataframe/io.jl
bkamins: I thought we should not print all rows by default (the same with columns). The idea is that if you want to print all rows/columns you use `show` and explicitly set `allrows=true` or `allcols=true`. I think it should be not a huge burden. I would assume that users will almost never need to print whole a `DataFrame` to a file in e.g. `text/html` format so if you want to do it it is not a big burden to wrap `df` in `show` with appropriate kwargs.

src/abstractdataframe/io.jl
bkamins: For these reasons I was hesitant to use `:displaysize` or `displaysize(io)` but rather to have something we can customize. But I admit that I am not sure what is the best design here.

Maybe we could simply disallow customization of these parameters if this is a temporary solution (so we use the defaults or print everything). I do not think people would complain about it as most of the users would not even know that they can change the defaults or care about it. You always can print the exact slice of your data frame by writing `show(io, df[my_rows, my_columns], allrows=true, allcols=true)` so this should not be a great problem.

nalimilan: Yes that's only supposed to be a rough rule to choose the size. The point is that I don't think we're supposed to use `IOContext` for custom properties like that, especially with so generic names. Also they are propagated to `show` methods for elements, which really isn't what we want here (even if in practice it's not a problem).

Maybe better just hardcode the number of rows and columns for now, and not support any `IOContext` property that isn't really supposed to be used in Jupyter anyway. We'll see later whether we can find a better approach.

src/abstractdataframe/io.jl
nalimilan: I'm not sure I agree:
- A data frame can be fairly small, notably if it contains summary statistics over groups.
- For CSV output we definitely want to print everything, so better be consistent for HTML. Also the text/plain method defaults to `false`.
- I don't see in what situations one would like to print a partial data frame as HTML or LaTeX to a file (as opposed to Jupyter, which sets `:limit=>true`).

src/abstractdataframe/io.jl
bkamins: Right. I will do this.

src/abstractdataframe/io.jl
bkamins: OK

bkamins:
OK - I have pushed a commit changing `:limit` to default to `false` and to have hardcoded number of rows/cols to print when `:limit` is `true`.

I have changed the defaults a bit both for HTML and LaTeX to something that seemed reasonable for me.

stevengj:
> Therefore my initial proposal was to have fixed number of columns and rows to be printed.

That's basically what `displaysize(io)` in Jupyter gives you — they are set heuristically to 30x80 as a reasonable rough guideline to show (it's fine if you don't adhere to them exactly).   The difference from hard-coding  them in DataFrames is that this is user-configurable by setting the LINES and COLS environment variables in the kernel configuration file.

bkamins:
> That's basically what `displaysize(io)` in Jupyter gives you

Thank you for the explanation.

My current thinking is the following:
* it is OK to print much more (e.g. like up to 100 rows and 100 variables) as it is not problematic because Jupyter Notebook automatically adds scroll bars if the output is too large. What we need is to have a reasonable limit so that we do not produce overly large output as it leads to lags. (so we do not care too much about `displaysize(io)` anyway)
* if someone wants to print exactly a given number of rows and columns from a data frame the pattern to use is `show(io, MIME("text/html"), view(df, 1:rows,1:cols), allrows=true, allcols=true)` (agreed this is ugly, but at least it is doable).

stevengj:
It's a matter of taste.   Of course, it's okay to print more, but 99% of the time I don't want to scroll in my notebook.

bkamins:
My point is that you can decide to enable/disable vertical scrolling in Jupyter Notebook by clicking in the left zone of the output, so even if we have 100 rows it looks OK (you can even collapse it by double-clicking).

Here is a scrolled view:

image


And here it is expanded:

image


Here is a link that lets you to check this out http://localhost:8888/notebooks/large_df_demo.ipynb.

src/abstractdataframe/io.jl
```diff
@@ -92,16 +99,72 @@ function html_escape(cell::AbstractString)
...
+14
+
+julia> show(stdout, MIME("text/csv"), DataFrame(A = 1:3, B = ["x", "y", "z"]))
+"A","B"
+1,"x"
+2,"y"
+3,"z"
+```
+"""
+Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame;
```
nalimilan: Can you use a broader signature so that only one method is needed for different MIME types? That will also allow attaching the docstring to the right method. Then also move it to the top of the file.

bkamins: Done. Now the default fallback will be to `"text/plain"` if some unknown mime type is passed.

bkamins: I have removed one unused method for `printtable` (probably it was exported in the past?)

src/abstractdataframe/io.jl
```diff
@@ -1,17 +1,79 @@
...
+- `mime::MIME`: supported MIME types are: `"text/plain"`, `"text/html"`, `"text/latex"`,
+  `"text/csv"`, `"text/tab-separated-values"`
+- `df::AbstractDataFrame`: The data frame to print.
+- `allrows::Bool `: supported only for plain text, HTML and LaTeX MIME types;
+   whether to print all rows, rather than a default number defined in
+   `HTML_NROWS` and `LATEX_NROWS` constants for these MIME types.
+- `allcols::Bool`: supported only for HTML and LaTeX mime types;
+   whether to print all columns, rather than a default number defined in
+   `HTML_NCOLS` and `LATEX_NCOLS` constants for these MIME types.
+- `summary::Bool = true`: supported only for HTML mime type;
```
```suggestion
- `summary::Bool = true`: supported only for the HTML MIME type;
```

src/abstractdataframe/io.jl
```diff
@@ -1,17 +1,79 @@
...
+
+# Arguments
+- `io::IO`: The I/O stream to which `df` will be printed.
+- `mime::MIME`: supported MIME types are: `"text/plain"`, `"text/html"`, `"text/latex"`,
+  `"text/csv"`, `"text/tab-separated-values"`
+- `df::AbstractDataFrame`: The data frame to print.
+- `allrows::Bool `: supported only for plain text, HTML and LaTeX MIME types;
+   whether to print all rows, rather than a default number defined in
+   `HTML_NROWS` and `LATEX_NROWS` constants for these MIME types.
+- `allcols::Bool`: supported only for HTML and LaTeX mime types;
```
```suggestion
- `allcols::Bool`: supported only for HTML and LaTeX MIME types;
```

src/abstractdataframe/io.jl
```diff
@@ -1,17 +1,79 @@
...
+Base.show(io::IO, mime::MIME, df::AbstractDataFrame;
+          allrows::Bool = !get(io, :limit, false),
+          allcols::Bool = !get(io, :limit, false),
+          summary::Bool=true) =
+    if mime isa MIME"text/html"
+        _show(io, mime, df, allrows=allrows, allcols=allcols, summary=summary)
+    elseif mime isa MIME"text/latex"
+        _show(io, mime, df, allrows=allrows, allcols=allcols)
+    elseif mime isa MIME"text/csv"
+        printtable(io, df, header = true, separator = ',')
```
nalimilan: `allrows` and `allcols` are ignored. Better throw an error if they are `false`.

src/abstractdataframe/io.jl
nalimilan: > Now the default fallback will be to `"text/plain"` if some unknown mime type is passed.

Shouldn't this be an error instead?

nalimilan:
@bkamins @stevengj I don't have a strong opinion regarding whether the number of rows and columns to print should be hardcoded or chosen using `displaysize` as I don't use Jupyter. I tend to prefer the second approach, but it would require more work to translate the number of characters to a number of columns.

src/abstractdataframe/io.jl
bkamins: Well I was thinking of `"text/csv"` and `"text/tab-separated-values"` as outputs where we assume `io` has "infinite dimensionality". But I will change it to throw an error.

src/abstractdataframe/io.jl
bkamins: OK - I will make it an error.

bkamins:
@stevengj - you are the probably the best expert in Jupyter Notebooks here so could you please cast your vote.

My preference for fixed and "large" dimensions are for two reasons:
* counting character columns (as opposed to showing a given number of variables) will complicate the code (it is doable, but the question is if we want to throw another ~100 LOC to maintain to support it while we expect to switch to another display style in the future anyway)
* my thinking was that Jupyter Notebook has the "collapse" and "scrolling" functionality exactly to allow the output to be too large while still manageable

(we could also limit the number of rows printed based on `displaysize` but decide how variables to show using fixed parameter)

bkamins:
@nalimilan after merging `show` into one method we get the following ambiguity:
```
show(io::IO, mime::MIME, df::AbstractDataFrame) in DataFrames at /home/travis/build/JuliaData/DataFrames.jl/src/abstractdataframe/io.jl:51
show(io::IO, ::MIME{Symbol("text/csv")}, a) in DelimitedFiles at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/DelimitedFiles/src/DelimitedFiles.jl:828
  Possible fix, define
    show(::IO, ::MIME{Symbol("text/csv")}, ::AbstractDataFrame)
```

I guess we have to revert the old approach, or you see some better approach?

nalimilan:
You could define the suggested method and have it call `invoke`. Not really cleaner than the previous approach, but...

src/abstractdataframe/io.jl
nalimilan: Yeah, but it's kind of weird to change the number of rows/columns to print based on the MIME type rather than based on `io`.

bkamins:
> You could define the suggested method and have it call `invoke`.

After thinking about I I would go for a `Union` of supported types in the signature approach. Are you OK with this? If yes I will implement it.

(actually what we have here is a good example of the consequences of type piracy 😄).

nalimilan:
Yes, that sounds appropriate.

bkamins:
I have vaguely remembered that `Union` is considered ambiguous also, and just checked that unfortunately this is the case:
```
ERROR: MethodError: f(::Int64, ::Int64) is ambiguous. Candidates:
  f(x::Union{Float64, Int64}, y::Int64) in Main at REPL[4]:1
  f(x::Int64, y) in Main at REPL[3]:1
Possible fix, define
  f(::Int64, ::Int64)
```
so we cannot use `Union` approach. I will define specific methods for `MIME"text/csv"` and `"text/tab-separated-values"` (as they are both problematic) and use `invoke`. We can revert to the old code if people report problems.

bkamins:
`"text/plain"` had the same problem so I decided to write it a bit differently (and additionally we have a better support for `"text/plain"` printing).

@nalimilan I had to do another significant rewrite of a docstring so please freely fix it as usual 😄.

src/abstractdataframe/io.jl
```diff
@@ -1,17 +1,90 @@
...
+          allrows::Bool = !get(io, :limit, false),
+          allcols::Bool = !get(io, :limit, false),
+          summary::Bool=true) =
+    if mime isa MIME"text/html"
+        _show(io, mime, df, allrows=allrows, allcols=allcols, summary=summary)
+    elseif mime isa MIME"text/latex"
+        _show(io, mime, df, allrows=allrows, allcols=allcols)
+    else
+        throw(ArgumentError("Unknown MIME type: $mime"))
+    end
```
stevengj: Why are you even defining `Base.show(io::IO, mime::MIME, df::AbstractDataFrame)`?   Since the only two supported types are text/html and text/latex, and you do different things for these, you should just define those two, no?
```jl
"""
docstring
"""
```
function Base.show(io::IO, mime::MIME, df::AbstractDataFrame); end


Base.show(io::IO, mime::MIME"text/html", df::AbstractDataFrame;
                    allrows::Bool = !get(io, :limit, false), allcols::Bool = !get(io, :limit, false), summary::Bool=true)
   = _show(io, mime, df, allrows=allrows, allcols=allcols, , summary=summary)
Base.show(io::IO, mime::MIME"text/latex", df::AbstractDataFrame;
                    allrows::Bool = !get(io, :limit, false), allcols::Bool = !get(io, :limit, false))
   = _show(io, mime, df, allrows=allrows, allcols=allcols)
```

File: src/abstractdataframe/io.jl [hunk 10]
```diff
@@ -1,17 +1,90 @@
...
+julia> show(stdout, MIME("text/csv"), DataFrame(A = 1:3, B = ["x", "y", "z"]))
+"A","B"
+1,"x"
+2,"y"
+3,"z"
+```
+"""
+Base.show(io::IO, mime::MIME, df::AbstractDataFrame;
+          allrows::Bool = !get(io, :limit, false),
+          allcols::Bool = !get(io, :limit, false),
```
stevengj: Are these keywords really necessary?   The user could always pass `IOContext(io, :limit=>false)`, or `IOContext(io, :displaysize=>(10^9,10^9))` if you honor `displaysize…

src/abstractdataframe/io.jl
bkamins: @nalimilan - how do you see the change now (when actually most of the cases have to be specialized anyway)? I can revert to the old design with all specialized methods (then the question is to what should we attach the docstring).

src/abstractdataframe/io.jl
bkamins: The idea was to allow to independently change this for rows and columns. But given the discussion in this PR it seems that the whole approach is flawed. Thank you for the comment.

Currently on master we have the following rule:
```
    haslimit = get(io, :limit, true)
    if haslimit
        tty_rows, tty_cols = displaysize(io)
        mxrow = min(nrows,tty_rows)
    else
        mxrow = nrows
    end
```
so maybe we leave it and just decide on number of columns to print based on the same rules as we use for REPL vs `tty_cols`?

The only question is if we should base these rules on `displaysize` function (this is currently on master) or `:displaysize` property from `IOContext` (and fallback to `displaysize` function when it is not defined)?

stevengj: You should base it on the displaysize property from IOContext.

Via displaysize you could independently remove the limit for the rows and columns just by setting the rows or columns to a large number.

nalimilan: We added the keywords for convenience in interactive use at the REPL. I'd say better support them for all methods, even if the same behavior can be achieved using `IOContext`.

I agree it sounds good to use the same rules as for the REPL. But note that calling `displaysize(io)` or getting `:displaysize` from `IOContext` is equivalent AFAICT, as the former uses the latter when the stream is an `IOContext`. Looking up the `IOContext` property directly just allows you to use a default different from `displaysize()` if you want.

src/abstractdataframe/io.jl
nalimilan: The solution proposed by @stevengj looks good to attach the docstring to the right method, but AFAICT it has the drawback that we define a function which doesn't do anything. Maybe have it throw an "unsupported MIME type" error?

bkamins: OK

src/abstractdataframe/io.jl
bkamins: I am OK with any solution (to be frank I would not expect the functionality to be used much in practice anyway and I understand we are going to rewrite what we have here in the future anyway; we can keep in mind that currently `allrows` and `allcols` kwargs are not supported, so adding them will add another deprecation period when we remove them), so please decide and I can implement whatever is the concensus.

bkamins:
In the end I recommend to follow what @stevengj proposed (and not change kwargs of `show` for HTML and LaTeX from what we have - we can change it when we implement target functionalities). I use `:displaysize` to get display size information.

One strange thing I could not find a workaround for was that when I defined the dummy `mime::MIME` method that only threw an error (and to which a docstring was attached) then Jupyter Notebook did not detect the `"text/html"` method and produced `"text/LaTeX"` output in the notebook. Therefore I have dropped this method (but maybe someone understands the reason and this can be fixed).

src/abstractdataframe/io.jl
```diff
@@ -21,87 +68,27 @@ function digitsep(value::Integer)
...
                summary::Bool=true, rowid::Union{Int,Nothing}=nothing)
-    n = size(df, 1)
-    if rowid !== nothing && n != 1
+    if rowid !== nothing && size(df, 1) != 1
         throw(ArgumentError("rowid may be passed only with a single row data frame"))
     end
-    cnames = _names(df)
+
+    mxrow, mxcol = size(df)
+    if get(io, :limit, true)
```
nalimilan: Again can we default to `false`?

src/abstractdataframe/io.jl
```diff
@@ -21,87 +68,27 @@ function digitsep(value::Integer)
...
+        tty_rows, tty_cols = get(io, :displaysize, displaysize(io))
+        mxrow = min(mxrow, tty_rows)
+        mxcol = min(mxcol, tty_cols)
```
nalimilan: This is going to print much more columns than at the REPL. Would there be any way to use the width computed by the text/plain `show`? Or maybe divide this by a reasonable factor (e.g. 6-10)?

src/abstractdataframe/io.jl
```diff
@@ -21,87 +68,27 @@ function digitsep(value::Integer)
...
-    n = size(df, 1)
-    if rowid !== nothing && n != 1
+    if rowid !== nothing && size(df, 1) != 1
         throw(ArgumentError("rowid may be passed only with a single row data frame"))
     end
-    cnames = _names(df)
+
+    mxrow, mxcol = size(df)
+    if get(io, :limit, true)
+        tty_rows, tty_cols = get(io, :displaysize, displaysize(io))
```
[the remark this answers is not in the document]
bkamins: Indeed

src/abstractdataframe/io.jl
bkamins: OK

src/abstractdataframe/io.jl
bkamins: Done, based on actual column widths + 2 characters for padding.

bkamins:
I have pushed a commit doing a dynamic calculation of number of columns to show.

@stevengj - it seems to me (but please confirm as I am not a Jupyter expert) that the `:displaysize` is not large enough. When I run my code only around 70% of the width is used. Also when I resize the browser window nothing changes in the produced output.

In particular note that the font used in HTML typesetting is a bit smaller than the font used for displaying code.

bkamins:
I have also opened https://github.com/JuliaLang/IJulia.jl/issues/835 with the case of defining this more general `show` method case.

src/abstractdataframe/io.jl
```diff
@@ -1,17 +1,64 @@
...
+\\end{tabular}
+14
+
+julia> show(stdout, MIME("text/csv"), DataFrame(A = 1:3, B = ["x", "y", "z"]))
+"A","B"
+1,"x"
+2,"y"
+3,"z"
+```
+"""
```
nalimilan: Actually since https://github.com/JuliaLang/IJulia.jl/issues/835 doesn't work it looks like you can attach the docstring to any signature, even if it doesn't correspond to a real method:
```suggestion
"""
Base.show(io::IO, mime::MIME, df::AbstractDataFrame)
```

src/abstractdataframe/io.jl
nalimilan: Great!

stevengj:
> it seems to me (but please confirm as I am not a Jupyter expert) that the :displaysize is not large enough. When I run my code only around 70% of the width is used. Also when I resize the browser window nothing changes in the produced output.

Realize that IJulia has no way of knowing the size of your browser window or your font size — Jupyter does not provide this information to a kernel.    Indeed, the same kernel could be connected to multiple front ends simultaneously.   Hence, by default, it just picks a heuristic size of 30x80, which the user can override with environment variables.

If the default heuristic is using 70% of your width, that doesn't seem to be a bad guess.

bkamins:
> Realize that IJulia has no way of knowing the size of your browser window or your font size

This is what I have thought. Can you then please hint us (so that we can include it in the documentation) how a user can manually change number of rows and columns in `:displaysize` in an interactive session? Will changing `ENV["COLUMNS"]` and `ENV["LINES"]` dynamically change it? Thank you.

stevengj:
It only looks at `ENV["COLUMNS"]` and `ENV["LINES"]` at startup (you could modify them in your kernel configuration file).

The current way to override it is to write to `IOContext(io, displaysize=>(rows,cols))`.    I suppose we could make [`IJuliaStdio`](https://github.com/JuliaLang/IJulia.jl/blob/f81332558fce4736fbd1b80841a6dfcc44198e6d/src/stdio.jl) mutable so that you could change the default dynamically, but it seems better in general to rely on things that would work in other environments too.

src/abstractdataframe/io.jl
bkamins: Can you? I would assume that such a statement would be treated by Julia compiler as a function call and say that `io`is an undefined variable?

bkamins: We could add this signature and define additional `showable` methods (this should not be overly problematic) if you prefer.

bkamins:
@nalimilan - given this is a temporary solution what do you think we should do?

1. Stay with what we have in this PR now (the drawback is that it will be hard for users to "widen" the output if they wish as the canonical way to do this is non-obvious IMO)
2. Should we add some specific functions in DataFrames.jl to allow to control it
3. Add a multiplier to number of columns (i.e. produce 2x more characters than column count in `:displaysizie` - this should be reasonable in most of cases)

src/abstractdataframe/io.jl
nalimilan: I don't know how, but it works in my tests.

nalimilan:
I think what we have is OK. Anyway we need a better, interactive solution.

src/abstractdataframe/io.jl
bkamins: OK - I will try committing your suggestion and see if it works.

bkamins:
@nalimilan I have dug into `@doc` a bit and it seems to be relatively flexible in terms what it accepts as a second argument (which is surprising that even dynamically invalid statements are allowed, as long as they are correctly parsed statically)

PR #1763, opened 2019-03-31
Add :compact handling in show and organize show tests

### LINKED ISSUE #1762: Floating point number formatting
How can we show more decimals when a DataFrame is displayed in REPL?   

Two ideas:

Using `IOContext`:
1. Respect `:compact => false` 
2. Accept `:renderer => f` where f can be a custom function to render cells e.g.`f(value, column_name, column_index, column_type)::AbstractString`.  

Reference
https://discourse.julialang.org/t/display-more-decimals-in-dataframe/22545/1

bkamins:
Fixes #1762 and additionally wrapped tests of show in `@testset`

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -2,32 +2,32 @@ Base.summary(df::AbstractDataFrame) = # -> String
     @sprintf("%d×%d %s", size(df)..., typeof(df).name)
 
 let
-    local io = IOBuffer(Vector{UInt8}(undef, 80), read=true, write=true)
+    local buffer = IOBuffer(Vector{UInt8}(undef, 80), read=true, write=true)
     global ourstrwidth
 
     """
-        DataFrames.ourstrwidth(x::Any)
+        DataFrames.ourstrwidth(io::IO, x::Any)
 
     Determine the number of characters that would be used to print a value.
     """
-    function ourstrwidth(x::Any) # -> Int
-        truncate(io, 0)
-        ourshowcompact(io, x)
-        textwidth(String(take!(io)))
+    function ourstrwidth(io::IO, x::Any) # -> Int
+        truncate(buffer, 0)
+        ourshow(IOContext(buffer, :compact=>get(io, :compact, true)), x)
+        textwidth(String(take!(buffer)))
     end
 end
 
 """
-    DataFrames.ourshowcompact(io::IO, x::Any)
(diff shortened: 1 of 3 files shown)


src/abstractdataframe/show.jl
```diff
@@ -6,28 +6,28 @@ let
     global ourstrwidth
 
     """
-        DataFrames.ourstrwidth(x::Any)
+        DataFrames.ourstrwidth(x::Any, trueio::IO)
 
     Determine the number of characters that would be used to print a value.
     """
-    function ourstrwidth(x::Any) # -> Int
+    function ourstrwidth(x::Any, trueio::IO) # -> Int
```
nalimilan: Better put `trueio` first as with other functions. Maybe also rename `io` to something like `buffer` so that `trueio` can be just `io`?

src/abstractdataframe/show.jl
```diff
@@ -51,6 +51,7 @@ end
 
 """
     DataFrames.getmaxwidths(df::AbstractDataFrame,
+                            io::IO
```
```suggestion
                            io::IO,
```

src/abstractdataframe/show.jl
bkamins: OK

src/abstractdataframe/show.jl
```diff
@@ -2,32 +2,32 @@ Base.summary(df::AbstractDataFrame) = # -> String
...
 
     Determine the number of characters that would be used to print a value.
     """
-    function ourstrwidth(x::Any) # -> Int
-        truncate(io, 0)
-        ourshowcompact(io, x)
-        textwidth(String(take!(io)))
+    function ourstrwidth(io::IO, x::Any) # -> Int
+        truncate(buffer, 0)
+        ourshow(IOContext(buffer, :compact=>get(io, :compact, true)), x)
```
nalimilan: Isn't this redundant with what `ourshow` does?

bkamins: This is the most delicate part of this PR (and that is why I had to add `io` argument to `ourstringwidth`).

`ourstrwidth` prints to `buffer` not `io` so we have to make sure that `buffer` inherits `:compact` from `io` (this would produce a wrong result if `io` had `:compact` set to `false` as `buffer` by default does not have it defined, so `ourshow` would print to `buffer` assuming `:compact` is `true`).

PR #1767, opened 2019-04-05
Better fuzzy match

bkamins:
@nalimilan & @oxinabox:

PR following the discussion on Slack. We can now be more flexible, but I also want to make sure we do not swamp the error message with too many columns. There are three parameters to tweak:
* if we want to disregard case (I think yes - this is a typical mistake)
* what should be the maximal distance (I have made it 2, as @nalimilan commented that 1 is too small)
* what is the maximum number of matches to print (I have made it 8 unless those are only case mismatches, in this case we always print all such mismatches - it is unlikely to have more than 8 of them)

We can of course change these rules


src/other/index.jl
```diff
@@ -144,25 +144,29 @@ end
...
-    min(l1, l2) == 0 && return true
-    ts1, ts2 = chop.((s1, s2), head=1, tail=0)
-    uppercase(s1[1]) == uppercase(s2[1]) && return fuzzymatch(ts1, ts2, l1 - 1, l2 - 1)
-    l1 > l2 && return ts1 == s2
-    l1 < l2 && return s1 == ts2
-    ts1 == ts2
+# Fuzzy matching rules:
+# 1. ignore case
+# 2. maximum Levenshtein distance is 2
+# 3. always show matches with 0 difference (wrog case)
```
oxinabox:
```suggestion
# 3. always show matches with 0 difference (wrong case)
```

src/other/index.jl
```diff
@@ -144,25 +144,29 @@ end
...
+# 3. always show matches with 0 difference (wrong case)
+# 4. on top of 3. do not show more than 8 matches in total
+# Returns candidates ordered by (distance, name) pair
+function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
+        idxs = uppercase(string(idx))
+        dist = [(REPL.levenshtein(uppercase(string(x)), idxs), x) for x in keys(l)]
+        sort!(dist)
+        c = [count(x -> x[1] <= i, dist) for i in 0:2]
+        maxd = max(0, searchsortedlast(c, 8) - 1)
+        candidates = [s for (d, s) in sort!(dist) if d <= maxd]
```
```suggestion
        [s for (d, s) in dist if d <= maxd]
```

bkamins:
Thank you. Let us wait a few days as people might comment on matching rules.

bkamins:
For some reason coverage is reported to drop, but I do not understand why

bkamins:
The only consideration is that the new algorithm is obviously slower than the old one. I did some benchmarking and for 10000 columns the performance was OK, for 1_000_000 the lag was noticeable, but sub-second. So I guess it is OK (but it would not hurt if someone else also did a similar check 😄)

nalimilan:
Performance in case of user error is really not a concern IMO. :-)

bkamins:
Well - it is an expected exception :) and matters in interactive work. But as I have said - I have benchmarked it and it seems acceptable.

bkamins:
OK to merge this?

PR #1770, opened 2019-04-10
Improve consistency of DataFrame constructors

### LINKED ISSUE #1766: Why do we have `DataFrame(Int, 5, 5)`
In the constructors of DataFrame we have the following line:

https://github.com/JuliaData/DataFrames.jl/blob/8d6cd4327222c5acb58abdc093511f8465b8aca2/src/dataframe/dataframe.jl#L18

Do we know the original purpose of this constructor? It seems useful for pre-allocating a DataFrame and filling it in in a hot loop. But I don't think it's used anywhere aside from tests. 

If there are concerns about LOC in DataFrames we can deprecate it.

bkamins:
Fixes #1766.

My question is do we see the need for:
```
DataFrame(column_eltypes::Vector, names::AbstractVector{Symbol}, nrows::Integer;
          makeunique::Bool=false)
DataFrame(column_eltypes::Vector, names::AbstractVector{Symbol},
          categorical::AbstractVector{Bool}, nrows::Integer;
          makeunique::Bool=false)
```

I would also deprecate them both (I never use them). But maybe they have some sensible use case.


bkamins:
This is rebased against recently merged PRs and should be good for a final check before merging. Thank you.

bkamins:
OK - I will drop them. The side benefit is that we drop dependency on CategoricalArrays.jl in the constructors section 😄.

bkamins:
The only thing is style - two of the additionally deprecated constructors called the third. So I left them as is (but only moved to src/deprecated.jl) and called `@deprecate` only on the third.

src/dataframe/dataframe.jl
```diff
@@ -17,7 +17,6 @@ DataFrame(columns::Matrix, names::Vector{Symbol}; makeunique::Bool=false)
 DataFrame(kwargs...)
 DataFrame(pairs::Pair{Symbol}...; makeunique::Bool=false, copycols::Bool=true)
 DataFrame() # an empty DataFrame
-DataFrame(t::Type, nrows::Integer, ncols::Integer) # an empty DataFrame of arbitrary size
 DataFrame(column_eltypes::Vector, names::AbstractVector{Symbol}, nrows::Integer;
```
nalimilan: Remove them here too.

nalimilan:
Then the third one wouldn't actually be deprecated... Also deprecations are really ugly. Maybe only deprecate the `categorical` constructor for now? The others are relatively harmless.

src/dataframe/dataframe.jl
bkamins: right (I wanted, to but forgotten to; good to have a second pair of eyes 😄)

bkamins:
> Then the third one wouldn't actually be deprecated...

The other two call
`DataFrame(column_eltypes::AbstractVector{<:Type}, cnames::AbstractVector{Symbol}, nrows::Integer; makeunique::Bool=false)`
at their end so they will (unless I read something badly). I have also improved the deprecation layout.

My thinking is that we want to discourage creation of empty data frame and then filling it using `setindex!` as this is going to be slow and is not a good design. Users should either preallocate and call `DataFrame` or use `push!` or `append!`. That is why I prefer not to have any of these constructors.

nalimilan:
> The other two call
> `DataFrame(column_eltypes::AbstractVector{<:Type}, cnames::AbstractVector{Symbol}, nrows::Integer; makeunique::Bool=false)`
> at their end so they will (unless I read something badly). I have also improved the deprecation layout.

Yes but that doesn't provide users with a syntax which they can actually copy/paste. They just see a warning from inside a DataFrames function, without any indication about how to silence it.

bkamins:
> Yes but that doesn't provide users with a syntax which they can actually copy/paste.

Should I then add separate `@deprecate` for each case  with the suggested syntax hint?

nalimilan:
> Should I then add separate `@deprecate` for each case with the suggested syntax hint?

If you can find something this isn't completetely ugly... :-)

src/deprecated.jl
```diff
@@ -1,5 +1,49 @@
 import Base: @deprecate
 
+@deprecate DataFrame(t::Type, nrows::Integer, ncols::Integer) DataFrame(fill(t, ncols), nrows)
+
+@deprecate DataFrame(column_eltypes::AbstractVector{<:Type}, cnames::AbstractVector{Symbol},
+                     nrows::Integer;
+                     makeunique::Bool=false) DataFrame(AbstractVector[elty >: Missing ?
+                                                                      fill!(Tables.allocatecolumn(elty, nrows), missing) :
```
nalimilan: `Tables.allocatecolumn` isn't available from user code. Maybe just use `Vector` and `missings`?

bkamins: I was thinking about this and the use of `Index` below. OK - I will change both.

bkamins: I have pushed something that is shorter and should be equivalent (hopefully).

nalimilan:
Looks OK overall, but tests fail. :-/

bkamins:
I have forgotten to push back the fix. We need to use `Tables.allocatecolumn` because otherwise categorical vectors are not handled correctly.

bkamins:
CI failed in one case due to timeout.

nalimilan:
But does that mean tests rely on deprecated functions? Then better move them to deprecated.jl.

bkamins:
> But does that mean tests rely on deprecated functions?

OK. I moved them to deprecated (but the CI failure was I guess unrelated to this).

bkamins:
@nalimilan given the discussion we had I have reintroduced:
```
DataFrame(column_eltypes::Vector, names::AbstractVector{Symbol}, nrows::Integer=0;
          makeunique::Bool=false)
```

Note that I think that:
1. this constructor should require passing column names (but I am OK, if we decide that we allow the variant without passing column names)
2. I have added `nrows` to default to `0`; the reason is that the most useful usecase of this constructor is to create an empty `DataFrame` ready for adding rows later with e.g. `push!`.

If this is OK (and CI goes through). We should merge this before we finish `DataFrame!` PR.

src/deprecated.jl
```diff
@@ -1,5 +1,34 @@
 import Base: @deprecate
 
+@deprecate DataFrame(t::Type, nrows::Integer, ncols::Integer) DataFrame(Matrix{t}(undef, nrows, ncols))
+
+@deprecate DataFrame(column_eltypes::AbstractVector{<:Type},
+                     nrows::Integer) DataFrame(column_eltypes, gennames(length(column_eltypes)), nrows)
```
nalimilan: `gennames` isn't available outside of `DataFrames`. I guess `'x' .* string.(1:nrows)` is OK?

test/constructors.jl
```diff
@@ -302,11 +292,36 @@ end
...
+                   [:A, :B, :C], 100)
+    @test size(df, 1) == 100
+    @test size(df, 2) == 3
+    @test typeof(df[1]) == Vector{Union{Int, Missing}}
+    @test typeof(df[2]) == Vector{Union{Float64, Missing}}
+    @test typeof(df[3]) == Vector{Union{String, Missing}}
+    @test all(ismissing, df[1])
+    @test all(ismissing, df[2])
+    @test all(ismissing, df[3])
+    @test typeof(df[:, 1]) == Vector{Union{Int, Missing}}
```
nalimilan: Is this really useful? `df[1]` should be enough, right (I know it was in the original code)?

bkamins: removed (it was kind of preparation for `df[1]` deprecation, but not I am not sure if we will deprecate it after all)

src/deprecated.jl
bkamins: fixed (the code needs to be different, but the approach is right 😄)

bkamins:
Following the discussion in #1771 the following `DataFrame` constructors currently do not take `copycols` kwarg:
* `DataFrame(x::Vector{<:NamedTuple})`
* `DataFrame(columns::AbstractMatrix, cnames::AbstractVector{Symbol} = gennames(size(columns, 2));
          makeunique::Bool=false)`
* `DataFrame(column_eltypes::AbstractVector{T}, cnames::AbstractVector{Symbol},
                   nrows::Integer=0; makeunique::Bool=false)::DataFrame where T<:Type`

The reason that they will have to create respective columns anyway internally. The choice is:
* we leave them as is (then in #1771 I add additional methods for `DataFrame!`)
* we add `copycols` kwarg to them, which will be ignored, but the benefit is that we consistently accept `copycols` kwarg in all constructors

I am not sure which is better. Initially I thought option 1 is better (i.e. do not add `copycols` kwarg in cases it will not be used), but please comment before we move forward.

nalimilan:
Hard to decide. Let's not add them for now, we'll see whether people complain.

PR #1771, opened 2019-04-10
add DataFrame! function

bkamins:
Sometimes (mostly when reading data from a file) we need a convenience `DataFrame` constructor that does not copy columns by default and takes no arguments (to allow using `|>`).

@nalimilan had come concerns if this is really needed, but let us discuss if we want to keep or drop it.

CC @oxinabox, @quinnj


oxinabox:
I am pro this functionality.
Not yet sold one way or the other on the name

bkamins:
@nalimilan - any opinion on this (I want to have a feeling if we wait with this PR for the release)?

oxinabox:
We should wait for this PR before release.

If I am going to commit to going through and checking a ton of code to 
get rid of defensive copies and such,
then I want to check all my constructors at the same time.
And if the conclusion is that `DataFrame!` is nicer than `DataFrame(...; copycols=false)`,
than I want to make sure all my code uses it.

bkamins:
I am aware of exactly this and that is why I am asking. Note though, that currently this is proposed only for `DataFrame!(x)` form, not all possible `DataFrame` constructors. We also could change this to have `DataFrame!` accept all types of possible inputs (this should not be extremely hard).

So there are two questions:
* do we want to add such a constructor.
* do we keep `DataFrame!` name (which is, in particular, convenient for the reasons @oxinabox notes)

bkamins:
@oxinabox based on your 👍 and my 🤔 I came up with the decision to extend `DataFrame!` to allow all types of constructors that `DataFrame` allows 😄. Let us see what will be the feedback.

oxinabox:
I was 👍ing your 🤔 
as I am also 🤔

docs/src/lib/types.md
```diff
@@ -60,8 +60,8 @@ but they are columns of a `DataFrame` returned by `stackdf` and `meltdf`.
 ## [The design of handling of columns of a `DataFrame`](@id man-columnhandling)
 
 When a `DataFrame` is constructed columns are copied by default. You can disable
-this behavior by setting `copycols` keyword argument to `false`. The exception is
-if an `AbstractRange` is passed as a column, then it is always collected to a `Vector`.
+this behavior by setting `copycols` keyword argument to `false` or by using [`DataFrame!`](@ref) function.
```
```suggestion
this behavior by setting `copycols` keyword argument to `false` or by using the
[`DataFrame!`](@ref) function.
```

src/dataframe/dataframe.jl
```diff
@@ -257,6 +257,32 @@ function DataFrame(column_eltypes::AbstractVector{T}, nrows::Integer) where T<:T
     return DataFrame(column_eltypes, gennames(length(column_eltypes)), nrows)
 end
 
+"""
+    DataFrame!(args...; kwargs...)
+
+Create a `DataFrame` from a `args` and `kwargs` with `copycols=false`.
```
nalimilan: Maybe just "Equivalent to `DataFrame(args...; copycols=false, kwargs...)`"?

bkamins: OK. fixed

bkamins:
CI failed in one case due to timeout.

bkamins:
I have improved the tests coverage for `DataFrame!`. In particular note that although now Julia does not allow passing duplicate kwargs if they are listed **it is allowed** with splatting. I have updated a docstring to inform a user about it.

bkamins:
There is one bug in the code revealed by the tests. Now it exposes that actually not all constructors of `DataFrame` support `copycols` kwarg. I think I will add this kwarg to every `DataFrame` constructor (even if it is not used). Any objections?

src/dataframe/dataframe.jl
```diff
@@ -262,6 +262,9 @@ end
 
 Equivalent to `DataFrame(args...; copycols=false, kwargs...)`.
 
+In particular note that you can still pass `copycols` keyword argument
+to `DataFrame!` to override its default value.
```
oxinabox: but please don't as that would be hella confusing?

nalimilan: Yeah, probably better throw an error if `copycols` is provided. Throwing is always safer since we can allow it later if needed without breaking the API.

bkamins:
>  I think I will add this kwarg to every DataFrame constructor (even if it is not used).

do you think this would be a useful change (or we want to have `copycols` only in the constructors for which it makes sense to pass this `copycols` keyword argument)?

(I have disallowed passing `copycols` kwarg to `DataFrame!` in the last commit)

nalimilan:
> do you think this would be a useful change (or we want to have `copycols` only in the constructors for which it makes sense to pass this `copycols` keyword argument)?

Which constructors are in this case? Anyway that sounds like a separate PR.

bkamins:
> Which constructors are in this case?

The one taking a `Matrix` and the ones that do not take any data.

I will add them to #1770 if we agree for this change (we have to wait for #1770 to be merged anyway with this). If we do not add `copycols` kwarg to these constructors (as this kwarg will be ignored in them) then I can add additional methods for `DataFrame!` to cover these special cases in this PR.

bkamins:
I have rebased this PR against master and updated the missing `DataFrame!` methods along with some docs fixes. It should be good for a final review now.

src/dataframe/dataframe.jl
```diff
@@ -218,6 +218,48 @@ function DataFrame(column_eltypes::AbstractVector{T}, cnames::AbstractVector{Sym
...
+true
+"""
+function DataFrame!(args...; kwargs...)
+    if :copycols in keys(kwargs)
+        throw(ArgumentError("`copycols` keyword argument is not allowed"))
+    end
+    DataFrame(args...; copycols=false, kwargs...)
+end
+
+DataFrame!(columns::AbstractMatrix,
```
nalimilan: I don't think we should support this. It's misleading since a copy is performed. A sign of this is that the docstring gets significantly more complex. Better not allow it for now, as always we can add them later if needed, but removing them is harder.

bkamins: Actually I had the same thought over the weekend. I will add errors in the cases we do not allow `copycols` kwarg (we could drop the methods and an error would be thrown anyway, but this way it will be more clear what is the reason of the problem).

There is one special case `DataFrame!()` which I still allowed (although technically it does not copy) as this is a special case calling `DataFrame(; kwargs...)` constructor.

bkamins:
OK to merge? I have made some changes - as noted above - since the PR was approved.

Issue #1753, opened 2019-03-26
Convience methods for `getproperty` and `setproperty` in DataFrames with new ownership rules


pdeffebach:
I have remarked about this on slack and in #1695. On @nalimilan's suggestion I am posting an issue to discuss how we can make `df.col` easier with non-literals. 

**Motivation:** With #1695 (mostly) concluded, we will likely have the following two ways to get a single column from a DataFrame

* `df[:, :col]`: a copy
* `df.col` a non-copy. 
* `df[:col]`:  deprecated. 

I really like `df.col` syntax. I find it intuitive and requires little typing. However it only works with the actual literal `col`. If you have a variable `x` representing the symbol `:col`, you cannot do `df.x` to get the column `:col`. You also can't just do `df[:, x]` because that has different behavior -- a copy rather than the exact same vector. 

I want to find a syntax for 

`df.col = f.(df.col)` 

where you use the variable `x` to represent `:col`. 

**Alternative solutions for `getproperty`:**

* We can use `getproperty(df, x)` to have the same behavior as `df.col`. 
* We can maybe use `select!(df, x)` to have the same behavior. 

**Alternative solutions for `setproperty!:** 

* `setcol!(df, x, v)` will work. However I don't like it because now you have to worry about an extra parentheses. 

```
df.col = map(v) do e 
    @match e begin
    ... pattern matching
    end
end
```

Is complicated enough as is, without putting it in `setcol!(df, x, ...)`. 

Finally I like the symmetry of `getproperty` and `setproperty` looking the same. `df.col = f.(df.col)` is elegant.

**Solutions:**

Ideally: 

* `df.$x = v` would be valid Julia syntax, where the `$` escapes whatever is following it. 
* you could define a `struct` and overload `getproperty(df, s::ColumnSelector)`.  This is not possible, however, because Julia doesn't evaluate the expression after the dot at all. This doesn't work. 

Pragmatically:

* A macro? I explored this a bit and didn't get very far. 

I'm at a bit of a dead end in terms of ideas for this. But I do think it's important. The average scientist codes in global scope, and we aren't going to get them to put things in functions if they have to re-write all their code in a less intuitive way in order to do that. 

Thanks for reading my rant.

bkamins:
> Thanks for reading my rant.

I am done reading it 😄. Thank you for the input. This is indeed a hard nut to crack.

For an alternative to `getproperty` I thought of `getcol(df, x)`. Note that `x` can be also a number.

[...]

Issue #1727, opened 2019-02-20
API for groupwise column transformation


bkamins:
This Follows what @nalimilan identified in a discussion on Discourse.

A common operation when transforming a column is groupwise standardization of data.

Assume you have a data frame with grouping column `:g` and data column `:x` you want to mutate `:x` by subtracting groupwise mean of `:x` and divide by groupwise standard deviation of `:x`.

There are several options for this API. One of the ideas is the following change to `by` and `combine` functions:

* add a `groupwise::Bool` keyword argument which defaults to `false` and then we have a current behavior;
* if `groupwise` is `true` the following changes to the behavior happen:
    * the output data frame contains all columns from the original data frame (here we should decide if the order of columns should be unchanged OR we put grouping columns first as is currently done)
    * if `sort=false` then the existing rows are returned in the order present in the original data frame; if `sort=true` then they are sorted by groups
    * all functions `f` are assumed to return a single column with the length equal to the group length; this can be simply checked because what is proposed here behaves as-if user manually added the following entries to transformation list `col = :col => identity` for all non-grouping columns `:col` in the parent data frame

Then the standardization operation described above would be written e.g. as:

```
by(df, :g, s_x = :x => x -> (x .- mean(x) ./ std(x)))
````

as opposed to what we have to do now:

```
julia> using Statistics

julia> df = DataFrame(g = rand(Bool, 10), a=1:10, b=1:10, x=1:10)
10×4 DataFrame
│ Row │ g     │ a     │ b     │ x     │
│     │ Bool  │ Int64 │ Int64 │ Int64 │
├─────┼───────┼───────┼───────┼───────┤
│ 1   │ false │ 1     │ 1     │ 1     │
│ 2   │ true  │ 2     │ 2     │ 2     │
│ 3   │ false │ 3     │ 3     │ 3     │
│ 4   │ false │ 4     │ 4     │ 4     │
│ 5   │ false │ 5     │ 5     │ 5     │
│ 6   │ true  │ 6     │ 6     │ 6     │
│ 7   │ true  │ 7     │ 7     │ 7     │
│ 8   │ false │ 8     │ 8     │ 8     │
│ 9   │ false │ 9     │ 9     │ 9     │
│ 10  │ true  │ 10    │ 10    │ 10    │

julia> by(df, :g, a=:a=>identity, b=:b=>identity, x=:x=>identity, s_x = :x => x -> (x .- mean(x)) ./ std(x))
10×5 DataFrame
[...]

PR #1772, opened 2019-04-10
Add select, select! and deletecols

bkamins:
A follow up of https://github.com/JuliaData/DataFrames.jl/issues/1753.

Adds `select`, `select!` and `deletecols`. In this way we have a flexible range of functions for column subsetting in a `DataFrame` - in-place and creating a new `DataFrame`, with column copying or without it.

The functions always return a `DataFrame` (as opposed to `df[col]` vs `df[cols]`).


bkamins:
I get a strange error on Julia 0.7 - for some reason DataFrames.jl does not recognize the `select!` function in the tests. Anyone has an idea what could be the reason?

bkamins:
OK - I see why this fails on Julia 0.7. It defined (although deprecated) `select` and `select!` functions that are currently not in base.

@nalimilan - so the decision is if we want to keep supporting 0.7 (I would drop it), then I simply change the REQUIRE and testset specification. If you think we should keep supporting 0.7 could you hint what is a proper way to handle such a case using Compat.jl (especially as old `select` and `select!` got renamed)?

nalimilan:
Thanks. I'm hesitant about whether `select(df, col)` should return a data frame or a vector. [In JuliaDB](https://juliacomputing.github.io/JuliaDB.jl/latest/api/#IndexedTables.select-Tuple{IndexedTables.AbstractIndexedTable,Any}) it returns a vector, so doing something different would make it harder to unify APIs.

The advantage of returning a data frame is that `select` would be clearly a table operation, very similar to `mutate`/`transform`. For example, if we extend it to allow things like `select(df, col => normalize)` and `select(df, newcol = col => normalize, newcol2 = col2 => -)`, both would return a data frame. JuliaDB doesn't support the latter syntax: it requires you to write `select(t, (newcol = col => normalize, newcol2 = col2 => -))`. Not the end of the world, but it makes things more complex. Also using a one-element named tuple looks weird when you want to create a single column and get a data frame.

> @nalimilan - so the decision is if we want to keep supporting 0.7 (I would drop it), then I simply change the REQUIRE and testset specification. If you think we should keep supporting 0.7 could you hint what is a proper way to handle such a case using Compat.jl (especially as old `select` and `select!` got renamed)?

Let's drop 0.7, it doesn't seem very useful at this point.

src/dataframe/dataframe.jl
```diff
@@ -865,6 +865,52 @@ end
 deletecols!(df::DataFrame, c::Int) = deletecols!(df, [c])
 deletecols!(df::DataFrame, c::Any) = deletecols!(df, index(df)[c])
 
+"""
+    deletecols(df::DataFrame, ind, copycols::Bool=true)
```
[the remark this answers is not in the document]
bkamins: Fixed in all the places (there was this inconsistency even before this PR)

src/dataframe/dataframe.jl
```diff
@@ -913,6 +959,100 @@ function deleterows!(df::DataFrame, ind::AbstractVector{Bool})
...
+│     │ Int64 │
+├─────┼───────┤
+│ 1   │ 4     │
+│ 2   │ 5     │
+│ 3   │ 6     │
+```
+
+"""
+function select!(df::DataFrame, ind::AbstractVector{Int})
+    targetnames = _names(df)[ind]
```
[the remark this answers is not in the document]
bkamins: Right. I have added it later and forgotten to move.

src/dataframe/dataframe.jl
```diff
@@ -913,6 +959,100 @@ function deleterows!(df::DataFrame, ind::AbstractVector{Bool})
...
+    indmin, indmax = extrema(ind)
+    if indmin < 1
+        throw(ArgumentError("indices passed to select! must be positive"))
```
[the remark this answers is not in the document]
bkamins: Fixed.

test/dataframe.jl
```diff
@@ -346,6 +346,175 @@ end
...
+    d = copy(df)
+    deletecols!(d, 2:3)
+    @test d == DataFrame(a=1, d=4, e=5)
+
+    d = copy(df)
+    deletecols!(d, 2:3)
+    @test d == DataFrame(a=1, d=4, e=5)
+end
+
+@testset "deletecols" begin
```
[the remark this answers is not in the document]
bkamins: added

bkamins:
> I'm hesitant about whether select(df, col) should return a data frame or a vector

That is why I like to make a PR as then things are on the table.

My position is that we should return a `DataFrame` because:
* `select` guarantees that you get a `DataFrame` which I think users will like (one less thing to learn)
* we have an identity `select(df, inds)` is the same as `select!(df, inds)`

@piever - could you please comment on the `select` issue?

I also envision that in the long run `select` should allow various column transformations like `combine`, but I wanted to keep the initial simple.

Regarding dropping 0.7 - I will make a separate PR (to keep things atomic) and then we can rebase this one (as probably we need some more discussion on this PR with the community).

bkamins:
Regarding `select`, `select!`, `deletecols` and the old `deletecols!`. The question is (it can be a separate PR) if we want to support them for `SubDataFrame` also. There are two levels of the question:
* do we want to allow mutating the view by `select!` and `deletecols!` (here I would say no - the silent assumption is that views are not mutable) - so this should be an error;
* do we want to allow `select` and `deletecols` - I would say yes, but the question is if we should return a `SubDataFrame` or allocate a new `DataFrame`

piever:
> The advantage of returning a data frame is that select would be clearly a table operation, very similar to mutate/transform. For example, if we extend it to allow things like `select(df, col => normalize)` and `select(df, newcol = col => normalize, newcol2 = col2 => -)`, both would return a data frame. JuliaDB doesn't support the latter syntax: it requires you to write `select(t, (newcol = col => normalize, newcol2 = col2 => -))`. Not the end of the world, but it makes things more complex. Also using a one-element named tuple looks weird when you want to create a single column and get a data frame.

In the IndexedTables case,  `select(df, col => normalize)` would return a vector, whereas `select(df, (col => normalize,))` would return a one column table (named col), whereas to also rename it's either with pairs of pairs or named tuple:  `select(df, (:newcol => col => normalize,))` or `select(df, (newcol = col => normalize,))` (of course the pair version is most useful when the symbol is generated programmatically).

I think design-wise IndexedTables generally prefers named tuples to many keyword arguments, for example `table((a = rand(10), b = rand(10)))`.

I'd tend to agree that the trade off is that if one wants the "many args or many kwargs" version, than `select(df, col)` needs to be a `DataFrame`. 

I'll list what I believe may be reasons for the current IndexedTables design:

- one can use a wide variety of things as "selections", listed [here](https://juliacomputing.github.io/JuliaDB.jl/latest/basics/#Selectors-1). The important rule is IMO "Tuple of Selection – returns a table containing a column for every selector in the tuple.". This requires things like `select(df, sym::Symbol)::AbstractVector` and `select(df, v::AbstractVector) = v` but then is quite powerful:

```julia
julia> t = table((a=1:10, b=rand(10)));

julia> select(t, (:a, :b => log, :c => rand(10)))
Table with 10 rows, 3 columns:
a   b          c
────────────────────────
1   -2.02031   0.854743
2   -0.67662   0.726542
3   -0.318742  0.793672
4   -1.0202    0.0611947
5   -0.54456   0.987932
6   -0.312883  0.766104
7   -0.627879  0.0454886
8   -1.12356   0.907651
9   -1.42856   0.864524
10  -1.23728   0.734384
```

- many table operations accept a keyword argument. For example `map(f, t; select = All()) = map(f, select(t, select))`, which is generally useful because we need a "thin table" to be sure the compiler doesn't collapse when iterating rows. In this case it's easier for `map(f, t, select = col)` to return a vector so that the user can write a `f` that expects a regular vector. For example `map(log, t, select = :b)` or `filter(in([2, 3]), t, select = :a)`

But I agree that these rules require some getting used to and maybe DataFrames user prefer a simpler: `select` returns a `DataFrame` kind of rule.

bkamins:
@piever - thank you for a detailed explanation (actually many of these things (or similar) in the long run should land in `select` for `DataFrame`). Also `df[:col]` currently returns a vector in DataFrames.jl, so having `select(df, :col)` do the same would not be an end of the world.

However, as you have commented at the end - I feel that having a guarantee that `select` always returns a data frame will make novice user's life simpler (and if someone really wants to select one column writing `df[:col]` is available).

nalimilan:
Thanks @piever. I see the logic behind the design of `select` in JuliaDB. OTOH I'm not completely convinced of the utility of `select` returning vectors, since the same result can be achieved by accessing the column vector using `columns` (in JuliaDB) or `getindex` (in DataFrame) and applying any transformation you want to it directly if needed.

At any rate, it would be nice if we could have a least one common method to access a column vector in JuliaDB and DataFrames. If that's not `select`, what can it be? Could JuliaDB support `getproperty(t, col)` for that? Otherwise DataFrames could support `coumns(df, col)` like JuliaDB, but `columns(df)` would be redundant with `eachcol`.

bkamins:
We chose `eachcol` over `columns` to be consistent with Base.

piever:
The simplest JuliaDB syntax for selecting just one column is `column(t, sym)`. There is a fair amount of redundancy in JuliaDB in that `columns` works in general (for symbol, int and tuple) whereas `column` is only to select one column as an `AbstractVector`. Maybe DataFrames could use `column`?

I'm not sure `getproperty(t, col)` will be implemented on IndexedTables, there was some negative feedback on it (see https://github.com/JuliaComputing/IndexedTables.jl/pull/203#issuecomment-447945900) and not many people have asked for it.

nalimilan:
We could support `column(df, col)`, especially if we end up deprecating `df[col]`. A possible name for that operation was `getcol`, but indeed "get" isn't very Julian. At least it wouldn't be terrible to add this at some point if that helps developing a common table API.

But `getproperty(::IndexedTable, ::Symbol)` also makes sense to me. The comment by Jeff you refer to may be a bit different AFAICT since it's about StructArrays, which are more expected to behave like arrays of named tuples than tables. Anyway that can be discussed independently from `column(df, col)`.

bkamins:
I am ok to add `column(df, col)` (even if in the end we decide we leave `df[col]`).

bkamins:
I will wait a bit for comments and add `column` to this PR if there are none (also please comment at some suitable time if the rest of the functionality is then what we want).

bkamins:
I have added it to show the changes (if we decide to drop it we can remove last 2 commits)

piever:
In terms of API, JuliaDB uses the singular (`pushcol`, `setcol`, `popcol`, see [here](https://juliacomputing.github.io/JuliaDB.jl/latest/operations/#[pushcol](@ref)-1)) even though it also allows to set many columns at the same time: I personally think the plural (`deletecols` as in this PR) makes more sense, so maybe there it could be JuliaDB to adapt.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -229,6 +229,13 @@ Base.setproperty!(df::AbstractDataFrame, col_ind::Symbol, x) = setindex!(df, x,
 # Private fields are never exposed since they can conflict with column names
 Base.propertynames(df::AbstractDataFrame, private::Bool=false) = names(df)
 
+"""
+    column(df::AbstractDataFrame, col_ind::Union{Symbol, Signed, Unsigned})
```
nalimilan: `Union{Symbol, Signed, Unsigned}` isn't used anywhere currently, and it's not user-friendly. Maybe we should cheat a bit and say `Integer`?

Can you remind me why we don't allow `Bool` here? It works for single indices with arrays, e.g. `(1:3)[true] == 1`.

nalimilan:
> In terms of "working with columns" API, JuliaDB uses the singular (`pushcol`, `setcol`, `popcol`, see [here](https://juliacomputing.github.io/JuliaDB.jl/latest/operations/#%5Bpushcol%5D(@ref)-1)) even though it also allows to set many columns at the same time: I personally think the plural (`deletecols` as in this PR) makes a bit more sense, but don't have a strong opinion either way.

Actually we chose the plural at https://github.com/JuliaData/DataFrames.jl/issues/1514#issuecomment-423211183 (and following comments), and we already use it for `insertcols`. The inconsistency with JuliaDB is a bit annoying.

src/abstractdataframe/abstractdataframe.jl
bkamins: OK. `(1:3)[true]` is a bug that probably should be reported. Here is the expected behavior AFAIK:
```
julia> x = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> x[true]
ERROR: ArgumentError: invalid index: true of type Bool
```

nalimilan: Are you sure? There's a problem when indexing with a vector containing `Bool` values since there's an ambiguity with logical indexing, but in the scalar case there's no problem and it makes the implementation simpler (it can just use `Integer`).

bkamins: I have opened an issue in Julia for this. I am pretty sure. I index `x` above using a scalar `Bool`.

nalimilan: OK I hadn't realized that what you showed was the current behavior. Then indeed the inconsistency is probably a mistake.

bkamins:
Given the discussion in https://github.com/JuliaData/DataFrames.jl/issues/1695 we might drop adding `column` for now (unless there are strong opinions to add it). The issue is related to whether we retain `df[col]` and `df[col] = v` syntax.

nalimilan:
Yeah, let's merge this PR without `columns` and discuss that later.

nalimilan:
Reading https://discourse.julialang.org/t/common-api-for-tabular-data-backends/21546 again, I recalled that `select` is basically equivalent to `map` if we consider data frames as collections of rows. I guess one difference which could justify having both is that `select` on a `GroupedDataFrame` will recycle scalars to ensure that the result has as many rows as the input (just like `transform`/`mutate` would).

Then these three functions would cover all interesting combinations of these two dimensions: 1) keep or drop remaining columns, 2) recycle scalars or not. The missing function is the one which keeps columns but doesn't recycle, which is the one which doesn't make sense since remaining columns have by definition one value per row. Does that make sense?

bkamins:
> Then these three functions would cover all interesting combinations of these two dimensions

@nalimilan - could you please expand on this, as I am not 100% clear which functions in which use cases you mean.

bkamins:
OK - I have reverted the `column` commits for a separate discussion. Also adding functionality to `select` (which I think we should do in the long run) should be probably a separate PR.

nalimilan:
> > 
> 
> @nalimilan - could you please expand on this, as I am not 100% clear which functions in which use cases you mean.

I was referring to https://github.com/JuliaData/DataFrames.jl/issues/1727#issuecomment-480592991.

bkamins:
So you mean the triplet `transform/select/map` in the case of `GroupedDataFrame`? Then yes I think that `select` for a `GroupedDataFrame` should preserve number of rows (thus implicitly recycle scalars in particular). But this is a separate PR I think

bkamins:
If there will be no more comments on this I am going to merge this PR.

Issue #1595, opened 2018-11-13
Future of colwise


bkamins:
Following https://github.com/JuliaData/DataFrames.jl/pull/1590#discussion_r232497822 I would like to discuss here what we should do with `colwise` function.

Now its basic form `colwise(f, df)` is the same as `map(f, columns(df))`, but we have also two additional methods:
* one producing a matrix when multiple functions are passed as `f` argument (works only for a data frame);
* one producing an array of arrays when a single function is applied to a grouped data frame.

How useful do you think those are and is is worth to keep `colwise` defined or should we deprecate it.

nalimilan:
I'm not sure these two methods are very useful. Since they return arrays, they don't include any information about the names of applied functions or about the groups. Anyway this behavior is easy to get using array comprehensions.

Typically, when applying a function on each group, you want the grouping column to be able to interpret the result, so you'd use `map(df -> map(f, columns(df)), gd)`, or `map(gd) do f.(columns(df)) end`. It could be useful to provide a convenience function to do that, maybe via `map(f, columns(gd))` and `f.(columns(gd))`? The similarity with the `DataFrame` syntax is appealing.

bkamins:
I would deprecate `colwise` then. OK?

Users can use `map` or `mapcols`. I guess in the long run when @nalimilan redesigns `GroupedDataFrame` API some additional convenience methods for this type can be implemented.

nalimilan:
Yes, let's deprecate it!

PR #1778, opened 2019-04-14
Deprecate colwise

nalimilan:
This function does not make a lot of sense since it doesn't preserve column nor function names.
The `GroupedDataFrame` method loses grouping information and returns a vector of vectors, which
isn't very useful. It also conflicts with `Distances.colwise`.

`aggregate` is a much more useful replacement, and the current behavior of `colwise` can easily be
obtained using a comprehension, which is even shorter now that we support `eachcol`.

Fixes https://github.com/JuliaData/DataFrames.jl/issues/1595.


nalimilan:
> Just please confirm that you have checked that the removed tests passed the new temporary definitions (so that we are sure that we really have an equivalent functionality).

Yes. You're right to check that, since the first time I ran the tests they failed even if I would have thought everything was OK. :-)

I've added a commit moving the tests to deprecated.jl: as long as we keep the code, better also ensure it works.

PR #1784, opened 2019-04-24
Move from REQUIRE to Project.toml

nalimilan:


Project.toml
```diff
@@ -0,0 +1,41 @@
...
+
+[targets]
+test = ["DataStructures", "DataValues", "Dates", "LaTeXStrings", "Test"]
+
+[compat]
+julia = "1"
+Missings = ">= 0.2.3"
+CategoricalArrays = ">= 0.5.2"
+StatsBase = ">= 0.11.0"
+WeakRefStrings = ">= 0.4.0"
```
quinnj: don't need WeakRefStrings

bkamins:
CI fails because of lack of LinearAlgebra import, but I think we do not use it (we import it in test/dataframe.jl, but I think this import is not used in this file)

bjarthur:
what was the reason to drop support for julia 0.7?  the API is the same with 1.0, so it should still work, no?

bkamins:
The reason was that 0.7 and 1.0 do not handle iterators in the same way in all cases so CI failed on 0.7 for some functions that work on 1.0.

bjarthur:
can you be specific?  i thought the only difference was that depwarns were removed.

bkamins:
You can see the list of inconsistent behaviors in the changes introduced [here](https://github.com/JuliaData/DataFrames.jl/pull/1774).

On top of this `select` and `select!` functions that now DataFrames.jl defines are undefined now in Base, but in 0.7 they were deprecated (so technically defined in Base). This was the trigger for the change.

bjarthur:
thanks for the info.  i understand that with every passing day 0.7 becomes less relevant, but in this case, if i understand things correctly, you could've simply imported `select` if 0.7, and exported it otherwise.  like done in [NIDAQ.jl](https://github.com/JaneliaSciComp/NIDAQ.jl/blob/master/src/NIDAQ.jl#L25-L29) for `start`.  DataFrames.jl is a popular library that many other packages depend on, and so you're forcing lots of things to drop support for 0.7 that wouldn't otherwise.

bkamins:
I understand this concern and I am aware that this is doable, but this would require a broader consensus.

DataFrames.jl depends on Tables.jl, TableTraits.jl, IteratorInterfaceExtensions.jl that all require Julia 1.0 anyway if I understand things correctly. I guess you can, as a temporary solution use, in the package that depends on DataFrame.jl, have latest release depend on latest DataFrames.jl release and long term support older release line depend on DataFrames.jl that supported 0.7 (I know it is not perfect).

@nalimilan - do you have any perspective on it?

nalimilan:
I guess if somebody cares enough about 0.7 to make PRs against all of these packages we could support it, but we have already enough to do without this. What's the use case for supporting 0.7?

bjarthur:
indeed, there is plenty to do, but the difference btw 0.7 and 1 is trivial to support.  and generally it's good to be as inclusive as possible when specifying version compatibilities.

i just tested master of Tables.jl, TableTraits.jl, and IteratorInterfaceExtensions.jl on 0.7 (in less that 5 minutes of my time) and they all still work as is.  the requisite PR is simply to modify the Project.toml and travis.yml.  i will happily submit PRs if they will be merged.

ararslan:
Julia 0.7 is officially unmaintained so personally I don't see a whole lot of reason for packages to support it, but it's mostly functionally equivalent to 1.0 anyway, so typically it will work.

davidanthoff:
I would merge PRs to IteratorInterfaceExtensions.jl and TableTraits.jl that add back julia 0.7 compat. Probably shouldn't have removed it in the first place.

I am wondering though, whether a better option would be to adjust the bounds in the registry, and not actually tag a new release...

quinnj:
No one should be using or supporting 0.7 anymore. 0.7 was meant as a ***temporary*** compatibility release to help when migrating to 1.0. There have been no backport or bugfix releases to 0.7. There were a number of deprecations and breaking syntax changes removed in 0.7. -> 1.0, so there are definitely cases that work on 1.0 that don't work in 0.7.

bjarthur:
Tables.jl still works on 0.7, so there is zero effort to support it.

for all new code bases you're right, one should develop with 1.1.  but i've still got lots of legacy code.  i've upgraded it all to 0.7, but since my unit tests don't have 100% coverage, i still get deprecation warnings occasionally.  were i to use 1.1 with this legacy code it would just crash without a message telling me how to easily fix it.

bkamins:
I understand your concern. First let me explain how I understand what @quinnj has written (maybe he will not confirm it, but let me try 😄). Then I can write what I can think that is a mid-way solution (still not a perfect one).

So I think what @quinnj says (and this was also my fear), is that even though the package might load correctly it is not guaranteed that it will work correctly as 0.7 is unmaintained and is not 100% consistent with 1.0 (and the inconsistencies are often subtle). This - in the unlikely extreme - might mean that using the package on 0.7 you will get a wrong result (I think this is very unlikely, a more likely case is that something throws an error in an unexpected place). We cannot guarantee that this does not happen as we are not developing the packages against 0.7 since its support was stopped.

I, however, understand the porting issues (and have done a lot of such work in fall 2018) - and here 0.7 comes very handy. However, 0.7 should be treated as dev environment only for transition to 1.0. After thinking about it, the solution I would use (I am not sure if it would suit your needs, but on dev environment it might be good enough) is the following:
* you can download locally the needed packages without installing them
* then you can manually modify Project.toml to push down the required Julia version number
* and you can use such locally downloaded packages for temporary development till you migrate all the legacy code to 1.0

I know this is not perfect, but maybe it will suit you (however, with the current Julia package manager this is doable).

However, if in general the community feels that it is safe to allow running the packages on 0.7 globally (without guaranteeing all will work correctly) then I will not object as long as we agree that this is a temporary solution.

bkamins:
And an even simpler solution (I hope now I will not get banned from JuliaData for proposing such things 😄) is:
* install on 0.7 an old version of the packages that still support it
* pin them (so that they do not get updated if you update something else)
* overwrite the packages with contents of src from a newer versions that you want

The only reason I am writing about it (because in general it is not a recommended practice), is that it is 5 minutes of work and we already spent much more time discussing it. Of course such things are not recommended on production, but you probably can get away with such an experiment on dev.

bkamins:
Just to give you an example of the problem of 0.7 vs 1.0 inconsistency I recently had.
When we started to design broadcasting for DataFrames.jl we still supported 0.7. At that time it seemed that we need to use some unexported method from Base to make it work (broadcasting a data frame is very non-standard). This method (because it was not exported) was inconsistent between 0.7 and 1.0 (user facing functionality was the same but the internal design changed) and I could not make tests pass CI on 0.7.

bkamins:
Sorry for spamming, but I think I have a clean solution (I have not tested it though on 0.7 as I do not have it):
* install a version of the package that is installable on 0.7
* `dev` this package
* you can do any fixes (if something does not work for you) you want on dev, as this is exactly the space to do such things

davidanthoff:
I opened a PR against the registry to relax the bounds for IteratorInterfaceExtensions and TableTraits: https://github.com/JuliaRegistries/General/pull/1011.

Issue #1785, opened 2019-04-25
Inconvenient use of iterate in fallback DataFrame constructor


quinnj:
On [this line](https://github.com/JuliaData/DataFrames.jl/blob/master/src/other/tables.jl#L24), we call `all` on any iterable input, which starts iterating the source. This is unfortunate when iterating is expensive for the source, as is currently the case for an SQLite.Query. I'm brainstorming ways to work-around this (and not make iterating an SQLite.Query so stateful), but the sqlite interface is very limited in this regard. I also ran into another case where the source was iterable, yet empty, and then got caught in this branch, but then errored on the core DataFrames constructor because the empty column names comprehension produced a `Union{}[]`.

bkamins:
Yes - this kind of thing occurred when trying to convert JuliaDB table without column names if I remember correctly.

In general in line https://github.com/JuliaData/DataFrames.jl/blob/master/src/other/tables.jl#L25 we iterate over `x` anyway so we might produce a vectors of column values and names first and then check what they are.

Also maybe (but I am not an expert in the design here) we could move https://github.com/JuliaData/DataFrames.jl/blob/master/src/other/tables.jl#L29 before https://github.com/JuliaData/DataFrames.jl/blob/master/src/other/tables.jl#L23 so that if something follows Tables.jl interface it gets handled by `Tables.columns` which can be made efficient wrt the source?

bkamins:
Is this also fixed by #1789?

nalimilan:
@quinnj Can you confirm?

quinnj:
Yep!

PR #1789, opened 2019-04-26
Check for Tables.istable on the *type* first, to catch for-sure table…

### LINKED ISSUE #1788: Line shift in MySQL queries
On version v0.18.0, the `MySQL.Query(D, "SELECT * FROM test.table") |> DataFrame`
returns a shifted table from the query, ie the first line is missing and an imaginary last line is added.
> 2×2 DataFrame
> │ Row │ key   │ value   │
> │     │ Int32 │ String⍰ │
> ├─────┼───────┼─────────┤
> │ 1   │ 2     │ Two     │
> │ 2   │ 0     │ #undef  │

This does not happen on v0.17.1
> 2×2 DataFrame
> │ Row │ key   │ value   │
> │     │ Int32 │ String⍰ │
> ├─────┼───────┼─────────┤
> │ 1   │ 1     │ One     │
> │ 2   │ 2     │ Two     │

quinnj:
…s before we try to iterate things. This fixes #1788 and is a workable solution for #1785

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -16,7 +16,7 @@ DataFrame(columns::NTuple{N,AbstractVector}, names::NTuple{N,Symbol};
 DataFrame(columns::Matrix, names::Vector{Symbol}; makeunique::Bool=false)
 DataFrame(kwargs...)
 DataFrame(pairs::Pair{Symbol}...; makeunique::Bool=false, copycols::Bool=true)
-DataFrame(pairs::AbstractVector{<:Pair{Symbol, AbstractVector}}; copycols::Bool=true)
+DataFrame(pairs::AbstractVector{Pair{Symbol, <:AbstractVector}}; copycols::Bool=true)
 DataFrame(pairs::NTuple{N, Pair{Symbol, AbstractVector}}; copycols::Bool=true)
 DataFrame() # an empty DataFrame
 DataFrame(column_eltypes::Vector, names::AbstractVector{Symbol}, nrows::Integer=0;
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -16,6 +16,8 @@ DataFrame(columns::NTuple{N,AbstractVector}, names::NTuple{N,Symbol};
 DataFrame(columns::Matrix, names::Vector{Symbol}; makeunique::Bool=false)
 DataFrame(kwargs...)
 DataFrame(pairs::Pair{Symbol}...; makeunique::Bool=false, copycols::Bool=true)
+DataFrame(pairs::AbstractVector{<:Pair{Symbol, AbstractVector}}; copycols::Bool=true)
+DataFrame(pairs::NTuple{N, Pair{Symbol, AbstractVector}}; copycols::Bool=true)
 DataFrame() # an empty DataFrame
 DataFrame(column_eltypes::Vector, names::AbstractVector{Symbol}, nrows::Integer=0;
           makeunique::Bool=false)
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -23,7 +23,7 @@ function DataFrame(x::T; copycols::Bool=true) where {T}
     if Tables.istable(T)
         return fromcolumns(Tables.columns(x), copycols=copycols)
     end
-    if x isa AbstractVector
+    if x isa AbstractVector || x isa Tuple
         if all(v -> v isa Pair{Symbol, <:AbstractVector}, x)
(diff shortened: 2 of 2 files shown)


bkamins:
If you feel it is safer (now I feel it is 😄) maybe also remove the `applicable` part from the code as I discussed in https://github.com/JuliaData/DataFrames.jl/issues/1788#issuecomment-487204798?

It is not really needed turned out to be problematic.

bkamins:
Or even better change `applicable(iterate, x)` to `T <: AbstractVector` and leave the rest as is?

src/other/tables.jl
```diff
@@ -16,10 +16,13 @@ fromcolumns(x; copycols::Bool=true) =
...
 
-function DataFrame(x; copycols::Bool=true)
+function DataFrame(x::T; copycols::Bool=true) where {T}
     if x isa AbstractVector && all(col -> isa(col, AbstractVector), x)
         return DataFrame(Vector{AbstractVector}(x), copycols=copycols)
     end
+    if Tables.istable(T)
+        return fromcolumns(Tables.columns(x), copycols=copycols)
+    end
     if applicable(iterate, x)
```
bkamins:
```suggestion
    if x isa AbstractVector
```

src/other/tables.jl
```diff
@@ -16,10 +16,13 @@ fromcolumns(x; copycols::Bool=true) =
               Index(collect(Symbol, propertynames(x))),
               copycols=copycols)
 
-function DataFrame(x; copycols::Bool=true)
+function DataFrame(x::T; copycols::Bool=true) where {T}
     if x isa AbstractVector && all(col -> isa(col, AbstractVector), x)
         return DataFrame(Vector{AbstractVector}(x), copycols=copycols)
     end
+    if Tables.istable(T)
```
bkamins: if we add this I understand we should remove the code starting in line 32 below (https://github.com/JuliaData/DataFrames.jl/pull/1789/files#diff-b84fca182aa841ee4af16b7d5d6e3399R32)

quinnj: No, they do 2 different things. It's subtle, but here we're checking if the ***type*** of `T` `istable`, and if so, dispatching directly to Tables.jl. Below, we check if the ***instance*** `istable`, which is currently much more liberal (though there is a [PR](https://github.com/JuliaData/Tables.jl/pull/84) to make it conservative again). The idea here is that most official Tables.jl-compatible sources implement `Tables.istable(::Type{T}) = true` directly, so we can rely on that to dispatch directly to Tables.jl before needing to do the `iterate` thing.

bkamins: OK - thank you.

bkamins:
I have made a small fix (following a request of @oxinabox we start to try supporting `Tuple` where we support `AbstractVector` - and this caused CI tests to fail without this change). I have also updated the documentation a bit.

Thank you for the fix.

bkamins:
@quinnj if this `x isa Tuple` is also safe (I think so, but things seem to be delicate here 😄) then we can merge this and mare a patch release today. So please let me know if this is OK.

CC @cadoubs

quinnj:
Yes, this all looks good to me now.

src/dataframe/dataframe.jl
```diff
@@ -16,6 +16,8 @@ DataFrame(columns::NTuple{N,AbstractVector}, names::NTuple{N,Symbol};
 DataFrame(columns::Matrix, names::Vector{Symbol}; makeunique::Bool=false)
 DataFrame(kwargs...)
 DataFrame(pairs::Pair{Symbol}...; makeunique::Bool=false, copycols::Bool=true)
+DataFrame(pairs::AbstractVector{Pair{Symbol, <:AbstractVector}}; copycols::Bool=true)
```
nalimilan: This isn't completely correct IIUC: we also accept `Vector{Any}` as long as the entries are of the right type.

quinnj: But that's still checked in the fallback constructor, right?

bkamins: > This isn't completely correct IIUC

You are right, but I did not know how to express this better (that is why in the commit message I have said this). Since this PR is merged if you can propose a better docstring please advise.

A precise specification would be that "we accept `AbstractVector` of any type as long as all its entries are subtypes of `Pair{Symbol, <:AbstractVector}`" but I was not clear how to express it in a concise way.

nalimilan: I guess the only correct way to document this is to use `::AbstractVector` here, and give more details in the text.

bkamins: OK - I will fix it in #1790

PR #1790, opened 2019-04-27
Additional fixes of DataFrame constructors

bkamins:
This PR fixes the following problem:
```
julia> gd = groupby(DataFrame(rand(3,4)), :x1)
GroupedDataFrame with 3 groups based on key: x1
First Group (1 row): x1 = 0.6359046443565228
│     │ Float64  │ Float64  │ Float64  │ Float64  │
│ 1   │ 0.635905 │ 0.310854 │ 0.695098 │ 0.579583 │
⋮
Last Group (1 row): x1 = 0.0519420235728989
│     │ Float64  │ Float64  │ Float64  │ Float64  │
│ 1   │ 0.051942 │ 0.666328 │ 0.896186 │ 0.775949 │

julia> DataFrame!(gd)
3×4 DataFrame
│ Row │ x1         │ x2         │ x3          │ x4          │
│     │ SubArray…  │ SubArray…  │ SubArray…   │ SubArray…   │
├─────┼────────────┼────────────┼─────────────┼─────────────┤
│ 1   │ [0.635905] │ [0.310854] │ [0.695098]  │ [0.579583]  │
│ 2   │ [0.116937] │ [0.6347]   │ [0.0623185] │ [0.0985515] │
│ 3   │ [0.051942] │ [0.666328] │ [0.896186]  │ [0.775949]  │
```

Now calling `DataFrame!` on `GroupedDataFrame` will error.

We should merge this before we release a patch for 0.18.0.

@quinnj - actually the reason of this problem is that Tables.jl is very aggressive in trying to convert to a `DataFrame` anything that is "iterable" even if the schema is `nothing` (maybe this is what we want, but I wanted to make sure that this is intended there).

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -1172,7 +1172,12 @@ function _aggregate(d::AbstractDataFrame, fs::AbstractVector,
     res
 end
 
-function DataFrame(gd::GroupedDataFrame)
+function DataFrame(gd::GroupedDataFrame; copycols::Bool=true)
+    if copycols
+        throw(ArgumentError("It is not possible to construct a `DataFrame`" *
+                            "from GroupedDataFrame with `copycols=false`"))
+
+    end
     length(gd) == 0 && return similar(parent(gd), 0)
     idx = similar(gd.idx)
     doff = 1
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -1173,7 +1173,7 @@ function _aggregate(d::AbstractDataFrame, fs::AbstractVector,
 end
 
 function DataFrame(gd::GroupedDataFrame; copycols::Bool=true)
-    if copycols
+    if !copycols
         throw(ArgumentError("It is not possible to construct a `DataFrame`" *
                             "from GroupedDataFrame with `copycols=false`"))
 
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -1176,7 +1176,6 @@ function DataFrame(gd::GroupedDataFrame; copycols::Bool=true)
     if !copycols
         throw(ArgumentError("It is not possible to construct a `DataFrame`" *
                             "from GroupedDataFrame with `copycols=false`"))
-
     end
     length(gd) == 0 && return similar(parent(gd), 0)
     idx = similar(gd.idx)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataFrames"
 uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
-version = "0.18.0"
+version = "0.18.1"
 
 [deps]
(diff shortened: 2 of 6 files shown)


nalimilan:
I don't understand how this PR affects `DataFrame!(::GroupedDataFrame)`. AFAICT it already throws an error since `copycols` isn't supported by `DataFrame(::GropuedDataFrame)`?

src/groupeddataframe/grouping.jl
```diff
@@ -1172,7 +1172,12 @@ function _aggregate(d::AbstractDataFrame, fs::AbstractVector,
     res
 end
 
-function DataFrame(gd::GroupedDataFrame)
+function DataFrame(gd::GroupedDataFrame; copycols::Bool=true)
+    if !copycols
+        throw(ArgumentError("It is not possible to construct a `DataFrame`" *
+                            "from GroupedDataFrame with `copycols=false`"))
```
```suggestion
                            "from a `GroupedDataFrame` with `copycols=false`"))
```

bkamins:
> I don't understand how this PR affects `DataFrame!(::GroupedDataFrame)`

This is a pretty subtle consequence of method dispatch in Julia. Because `DataFrame(::GroupedDataFrame)` does not support `copycols` kwarg Julia dispatches to `DataFrame(::Any; copycols=false)` in other/tables.jl.

In this PR I add `copycols` keyword argument to `DataFrame(::GroupedDataFrame)` to make sure we dispatch to a right method.

bkamins:
Or to put it more generally this is a peculiar consequence of having decided to:

* have `DataFrame` constructor accepting `::Any` as a positional argument
* not defining `copycols` keyword argument for all possible `DataFrame` constructors

I have sanitized all other constructors against this case, but forgotten about this one as it was defined in other file.

The only other affected constructor is `DataFrame(x::Vector{<:NamedTuple})` which "in theory" does not support `copycols` keyword argument, but you can write `DataFrame([(a=1,b=2)], copycols=false)` and get a properly constructed data frame without a warning (although copying does happen here). I can also disallow this - but please comment if we want to (as opposed do `GroupedDataFrame` the returned result is "correct" with the only problem that `copycols` kwarg is ignored).

nalimilan:
Wow, OK, I didn't know about that dispatch subtlety.

> The only other affected constructor is `DataFrame(x::Vector{<:NamedTuple})` which "in theory" does not support `copycols` keyword argument, but you can write `DataFrame([(a=1,b=2)], copycols=false)` and get a properly constructed data frame without a warning (although copying does happen here). I can also disallow this - but please comment if we want to (as opposed do `GroupedDataFrame` the returned result is "correct" with the only problem that `copycols` kwarg is ignored).

Probably better disallow this too, at least until we find a good reason to allow it.

bkamins:
Following the review now this PR has three changes:
* improved `DataFrame(::GroupedDataFrame)` constructor and added it to the documentation
* improved documentation of iterable of `Pairs` constructors
* safer `DataFrame` constructor taking a `Vector` of `NamedTuple`s

@nalimilan This should be good to merge now if you confirm. I will make a patch release after merging this.

PR #1793, opened 2019-04-29
minor change in vcat implementation

bkamins:
Changes in this PR:
* make `reduce(vcat, dfs; cols)` behave in the same way as `vcat(dfs...; cols)` (this means: disregard data frames with 0 columns and acccept `cols` keyword argument)
* make `reduce(vcat, dfs; cols)` also accept a tuple of `AbstractDataFrame`s and not only an `AbstractVector` of them

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1101,7 +1101,12 @@ julia> vcat(d4, df1)
 """
 Base.vcat(dfs::AbstractDataFrame...;
           cols::Union{Symbol, AbstractVector{Symbol}}=:equal) =
-    _vcat([df for df in collect(dfs) if ncol(df) != 0]; cols=cols)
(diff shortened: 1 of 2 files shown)


bkamins:
Thank you. This PR can wait till the next release, but if we need to do 0.18.2 I would include it, as in my opinion it is a patch making sure `vcat` and `reduce(vcat,...) are working consistently (as they should).

nalimilan:
As you like.

pdeffebach:
Yeah thanks for this! I didn't know it was allowed to add a keyword argument when dispatching `reduce` on a function.

bkamins:
Actually the issue of dispatch with kwargs is quite tricky (I recently had another issue in DataFrames.jl related to this; here is a status of a long unresolved issue that should improve dispatch rules of functions taking kwargs https://github.com/JuliaLang/julia/issues/9498) but in this case it should be safe.

PR #1795, opened 2019-04-30
add instructions for setting up Jupyter Notebook with DataFrames.jl

bkamins:
The way to change the way data frames are displayed in Jupyter Notebook is non-obvious so I propose to add the instructions to the manual.


bkamins:
CI fails due to an unrelated reason (update of Tables.jl).

bkamins:
@tk3369 - can you confirm that the proposed instructions are clear to you and work (so that I have a second pair of eyes that checks this before merging). Thank you

bkamins:
@pszufe - also could you help checking that both options (changing kernel `"env"` and dynamic setup of `ENV` variable) work.

bkamins:
I have made some more changes to the instructions to make them more precise.

bkamins:
OK. Thanks

PR #1796, opened 2019-04-30
update tests to follow Tables.jl 0.2 rules

bkamins:
New Tables.jl 0.2 rules made tests in test/tables.jl fail.


quinnj:
@bkamins , I would prefer a solution like https://github.com/JuliaData/DataFrames.jl/pull/1797 that allows construction of additional input table sources.

PR #1797, opened 2019-04-30
Update use of Tables.jl

quinnj:


bkamins:
Sure - if you know that it is safe I am OK with this.

bkamins:
@nalimilan - could you also please have a look at it (we need to merge this PR as without it tests fail so opening new PRs is stalled).

PR #1801, opened 2019-05-02
fix constructor with zero rows

### LINKED ISSUE #1800: Push! row to DataFrame initialized via variable
I have found strange behaviour when initializing an empty DataFrame and then adding a row.
Depending on how it was initialized the behaviour differs in a way that does not make sense to me, and I would like to ask whether this is intentional or not. 
Also if its the former, I would be curious what the reasoning is, as I am not sure I fully understand Types in Julia yet. 

Following the Getting Started section, typing
```   
tp = Int[]  
df = DataFrame(A = Int[], B = Int[])
push!(df, (1,2))
```
returns the expected 
```
2×2 DataFrame
│ Row │ A     │ B     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2    │
```
However, when trying to make it more modular (I want to construct several frames in a very similar way, but with different content (types)), I tried to following
```   
tp = Int[]  
df = DataFrame(A = tp, B = tp)
push!(df, (1,2))
```
which returns:
```
2×2 DataFrame
│ Row │ A     │ B     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │
│ 2   │ 2     │ 2     │
```
The push! command still requires a tuple/array with the same number of elements as existing columns, but appears to repeat them into each column, so adding a row with n elements to a column with n columns yields a n,n - DataFrame. 

It gets stranger when using different types for the columns. Typing
```
tp = Int[]  
df = DataFrame(A = String[], B = tp, C = tp)
push!(df, ("a",1,2))
```
results in
```
│ Row │ A      │ B     │ C     │
│     │ String │ Int64 │ Int64 │
├─────┼────────┼───────┼───────┤
│ 1   │ a      │ 1     │ 1     │
```

and, when adding the row twice we find 
```
tp = Int[]  
df = DataFrame(A = String[], B = tp, C = tp)
push!(df, ("a",1,2))
push!(df, ("a",1,2))
```
we find
```
│ Row │ A      │ B     │ C     │
│     │ String │ Int64 │ Int64 │
├─────┼────────┼───────┼───────┤
│ 1   │ a      │ 1     │ 1     │
│ 2   │ a      │ 2     │ 2     │
```

If this is intended behaviour, I would be interested to learn, and maybe it might be worth adding a note to the documentation.

bkamins:
Fixes #1800.

Additionally fixes a bug under which `0`-length ranges were not converted to a vector. This should be merged before 0.18.2 release.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1419,6 +1419,10 @@ Otherwise if `columns=:equal` then `row` must contain exactly the same columns a
 As a special case, if `df` has no columns and `row` is a `NamedTuple` or `DataFrameRow`,
 columns are created for all values in `row`, using their names and order.
 
+Please note that `push!` must not be used on a `DataFrame` that contains columns
+that are aliases (equal when compared with `===`) as it will silently produce
+a wrong result in such a situation.
+
 # Examples
 ```jldoctest
 julia> df = DataFrame(A=1:3, B=1:3)
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1302,6 +1302,10 @@ An error is thrown if conversion fails: this is the case in particular if a colu
 in `df2` contains `missing` values but the corresponding column in `df1` does not
 accept them.
 
+Please note that `append!` must not be used on a `DataFrame` that contains columns
+that are aliases (equal when compared with `===`) as it will silently produce
+a wrong result in such a situation.
+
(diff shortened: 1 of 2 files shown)


bkamins:
@nalimilan This PR passes tests. The problem is that the reported coverage on CI dropped, but I am sure I have increased the coverage.

PR #1802, opened 2019-05-03
increase DataFrames version to 0.18.2

bkamins:


bkamins:
@JuliaRegistrator register()

PR #1804, opened 2019-05-06
Broadcasting in DataFrames

bkamins:
This PR is a result of a discussion to take a staged approach at `setindex!`/broadcasting in DataFrames.jl.
We start with broadcasting.

This PR invalidates #1643 and #1646 but I will keep them open for a reference until nothing that was discussed there becomes irrelevant.

In general we treat `AbstractDataFrame` as a matrix in broadcasting and `DataFrameRow` as a vector.

The tricky part is that we want to allow adding columns using broadcasting. What I proposed works as expected, but is not entirely clean and uses Base internals a lot. Also if you review this PR please have a look at the tests as they expose corner cases (which are consistent with Base, although sometimes surprising).

@mbauman - we have discussed with @nalimilan that it would be great if you commented on this PR as the design of broadcasting in Base is complex and I am not sure if I have taken the best path. Thank you!


src/other/broadcasting.jl
```diff
@@ -0,0 +1,94 @@
...
+                end
+                df[idx] = Vector{DummyColumnValue}(undef, nrow(df))
+            end
+        end
+    end
+    view(df, idxs)
+end
+
+function Base.copyto!(df::SubDataFrame, bc)
+    m = Base.materialize(bc)
```
mbauman: I don't think you should need this `materialize` here.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,94 @@
...
+    if ncol(df) == 1 && eltype(df[1]) === DummyColumnValue
+        if maximum(ndims.(Base.Broadcast.flatten(bc).args)) > 1
+            throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
+        end
+    end
+    for col in axes(df, 2)
+        if eltype(df[col]) === DummyColumnValue
+            colname = names(df)[col]
+            T = reduce(promote_type, typeof.(view(m, :, col)))
```
mbauman: It feels like you should be able to fuse the `typeof.()` here into the function we pass to `reduce`.  Perhaps:

```suggestion
            T = reduce(((::Union{T,Type{T}}, ::Union{S,Type{S}}) where {T,S})->promote_type(T, S), view(m, :, col); init=Union{})
```

src/other/broadcasting.jl
```diff
@@ -0,0 +1,94 @@
...
+            parent(df)[colname] = Tables.allocatecolumn(T, nrow(df))
+        end
+        df[col] .= view(m, :, col)
```
mbauman: This looks wrong to me.  Are you sure the dimensions work out to broadcast between `m` and `df` if you do it columnwise like this?  It feels like you're missing an error condition or two, but I haven't completely worked it out.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,94 @@
...
+        for col in axes(df, 2)
+            if eltype(df[col]) === DummyColumnValue
+                colname = names(df)[col]
+                parent(df)[colname] = Tables.allocatecolumn(typeof(bc.args[1][]), nrow(df))
+            end
+            fill!(df[col], bc.args[1][])
+        end
+        df
+    else
+        copyto!(dfr, convert(Broadcasted{Nothing}, bc))
```
mbauman: What's `dfr`?

src/other/broadcasting.jl
```diff
@@ -0,0 +1,94 @@
...
+
+Base.copyto!(dfr::DataFrameRow, bc::Base.Broadcast.Broadcasted{<:Base.Broadcast.AbstractArrayStyle{0}}) =
+    if bc.f === identity && bc.args isa Tuple{Any} && Base.Broadcast.isflat(bc)
+        foreach(col -> dfr[col] = bc.args[1][], 1:length(dfr))
+        dfr
+    else
+        copyto!(dfr, convert(Broadcasted{Nothing}, bc))
+    end
+
+function Base.materialize!(dest::SubDataFrame, bc::Base.Broadcast.Broadcasted{Style}) where {Style}
```
mbauman: Oh, are these patching up over the error conditions I was missing above?  Seems like it'd be easier to just detect them within `copyto!` instead of depending upon the try/catch.

docs/src/lib/indexing.md
```diff
@@ -79,3 +79,19 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+`AbstractDataFrame` is converted to a matrix in broadcasting.
+`DataFrameRow`, `GroupedDataFrame`, `DataFrameRows` and `DataFrameColumns` are `collect`ed in broadcasting.
+
+Additionally it is possible to assign a value to `AbstractDataFrame` and `DataFrameRow` using the `.=` operator.
+In such an operation `AbstractDataFrame` is two dimensional and `DataFrameRow` as a single dimensional.
+
+If column indexing using symbols is performed the order of columns in the operation is specified by the order of symbols.
+
+It is allowed to perform `df[col] .= value` even if `col` is not present in the `DataFrame` and it is a `Symbol`.
+In such a case a new column will be created.
```
mbauman: This feels a little quirky to me.  I get that it could be handy, but it seems easier to introduce broadcasting first and then add this feature later.

docs/src/lib/indexing.md
```diff
@@ -79,3 +79,19 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
...
+A data frame with zero columns is always considered to have zero rows.
+The assignment broadcasting is not supported for `df.col` style.
```
mbauman: It isn't?  Or are you just stating that the special column-creation behavior doesn't work?

src/other/broadcasting.jl
mbauman: Oh, and remember that this is the general `copyto!` method — it's what will get called for arbitrary iterables and you're not guaranteed to have something `view`able.

bkamins:
@nalimilan - this should be good to look at although it fails CI. There is some problem with `materialize!` on Julia 1.0 only (I will look at what is the reason of this tomorrow) but otherwise the PR is good.

src/other/broadcasting.jl
bkamins: I materialized to avoid possible problems with aliasing (I might be modifying RHS which might appear later in processing in LHS).

Additionally in general if I do not `materialize` then what is the recommended way to access values stored in `bc`?

src/other/broadcasting.jl
bkamins: I do not understand this `Type{T}` part in the signature. I would assume that it is not needed as otherwise I will a wrong result if e.g. my vector I want to reduce over is `[Int64, Float64]` - I will get `T` to be `Float64` and it should be `DataType`.

src/other/broadcasting.jl
bkamins: This is a very good point. I think that the signature of `copyto!` should be restricted. We accept here only such `bc` that are two-dimensional. Of course maybe this should be changed in general if we can avoid `m = Base.materialize(bc)` (given your comment above). I used view only for performance. It can be dropped.

src/other/broadcasting.jl
bkamins: Thank you for spotting. It was a copy-paste problem from below (`dfr` is `DataFrameRow` in the definitions below)

src/other/broadcasting.jl
bkamins: This is what I wanted to do, but the issue is that sometimes `copyto!` does not get even called when an error is thrown (due to dimension mismatch) so I have to clean up in `materialize!` as it is the outermost function called.

docs/src/lib/indexing.md
bkamins: This was a major discussion point with @nalimilan. Without this feature all would be much simpler.

@nalimilan - do you think we have to stick with this feature? Of course the problem would be how to add a column of constant values to a data frame. You still could do `df.newcol = fill(value, nrow(df))` (or use the `Tables.allocatecolumn` thing), but initially you wanted to avoid having to do this.

docs/src/lib/indexing.md
bkamins: `df.col` does not work in the column-creation case. I will state it more clearly (and maybe even we will then drop it if we decide not to support column creation).

bkamins:
@mbauman thank you for your time and having a look at these issues.

The crucial part for me is if we need to `materialize` (especially in aliasing case) and how to best handle the problem with `view` you have highlighted. So in general the learning question on my side is how do you should safely use lazy `bc`.

docs/src/lib/indexing.md
nalimilan: I guess we should summarize the discussion we had yesterday to give the rationale. The problem is that we currently allow `df[col] = value` to create a new column, so since we want to deprecate that to use `.=` for consistency with Base, we'd better provide a working replacement (or at least decide upfront whether we want it or not).

bkamins: Thank you @nalimilan for expanding on it. And we knew that having this would significantly add to the complexity of the design (as normally broadcasting assignment should not allocate new things (resize the target) and here we do this).

src/other/broadcasting.jl
mbauman: Yeah, I was trying to be a bit too clever.  The idea stands, though — it would be nice to do the typeof operation in the reduction function itself instead of broadcasting (and allocating an array) for `typeof.()`.  It'll just take a bit smarter of a function definition.  The tricky part is simply that the array you're reducing over holds values but the intermediate results are types.  So the first time you call the reduction function you get two values, and then every subsequent time you get a type and a value.

docs/src/lib/indexing.md
mbauman: Ah, I see.  Ok, yeah, I agree that this is definitely better than `df[:x] = 1` or even `df.x = 1`.  It's a shame we can't hook into `df.x .= 1`.  I guess folks aren't too thrilled with a two-step process like `addcol!(df, :x); df.x .= 1`?  Or I suppose an alternative spelling could be `df.x = Iterators.repeated(1)`, but I'm not familiar enough with DataFrame's  `setproperty!` conventions to know if this could happily coexist with the other meanings.

In any case, I'm sure you've already done thorough brainstorming here and this is a reasonable extension to make.  It's probably good to do this at the same time because it does impact the design rather substantially.

mbauman:
With respect to `materialize`, it's not really something that I intended for folks to call directly and I don't think there's a need to overload it except in very special cases.

The key thing to remember about broadcast is that it's entirely based upon indexing and axes.  The typical way to use a `bc` object is to just iterate over `eachindex` and iteratively index into it.  Note that what comes out of a dot-expression might not be a `Broadcasted` — it could be an immediately re-computed range, for example.  With regards to alias-safety, we don't have a very friendly API to unalias broadcasted expression trees, but you could crib off of how Base preprocesses the broadcasted tree to do this.

src/other/broadcasting.jl
bkamins: Thanks - this is what I thought and rightly `typeof.` is too expensive here (I did not want to optimize out everything in the first pass). Probably `mapreduce` instead of `reduce` should work here.

docs/src/lib/indexing.md
bkamins: Actually `df.x = Iterators.repeated(1)` (or something similar) is a possible option and an interesting idea.

In general our plan is to first do broadcasting and then redesign `setindex!` and `setproperty!` (the reason is that the former is new and the latter will have to go through deprecation and will be breaking).

bkamins:
Thank you for the tips on use of `bc`. I will try to rewrite the code in this way.

Regarding aliasing - the additional difficulty in DataFrames.jl as `DataFrame` is also a nested structure itself that can contain aliases (e.g. if the user really wants one can store the same vector several times under different names). But thinking about it probably we can simply document that the behavior of broadcasting when aliasing is a problem is unspecified and this is not something that should be done (actually recently we had to do a similar comment for `push!` and `append!` functions that are also not aliasing safe).

bkamins:
@nalimilan In Base broadcasting does not provide a cleanup on error, e.g.:
```
julia> x = ones(2)
2-element Array{Float64,1}:
 1.0
 1.0

julia> y = [2, "2"]
2-element Array{Any,1}:
 2
  "2"

julia> x .= y
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Float64

julia> x
2-element Array{Float64,1}:
 2.0
 1.0
```

Do you think it is acceptable we do the same, or we should make sure that we do not leave a data frame half-modified after an error?

bkamins:
@mbauman I have two design questions to you:

* what are the guarantees on the value of elements produced by `eachindex` called on `bc` object. In particular (as this is a relevant possibly problematic case because data frames do not support linear indexing) - is it possible that `eachindex(bc)` will return a linear index if `bc` is two dimensional.
* I wanted to make sure that I understand it correctly. The method https://github.com/JuliaData/DataFrames.jl/pull/1804/files#diff-57cbbe702db7443003fc8c590f239d70R48 (similar to one in Base) is needed only for performance. We could remove it and all should work. Is this right?

Thank you.

bkamins:
@nalimilan I am waiting with doing the changes in this PR as I would welcome a comment, given the discussed options, what you think we should do with column creation when we do broadcasting:
1. create columns both on `df[cols] .= v` and `df[col] .= v`
2. create columns only on `df[col] .= v` (this will simplify the design)
3. never create columns and retain implicit broadcasting in `df[col] = v` and `df[cols] = v`
4. never create columns and add explicit broadcasting via `df[col] = Iterators.repeated(v)` and `df[cols] = Iterators.repeated(v)` (we might decide to use some shorter alias `Iterators.repeated` if we wanted)

Thank you!

nalimilan:
I'd say 2 is fine if implementation isn't too complex. We can always implement `df[cols] .= v` later if we want.

bkamins:
OK. I will go for 2 then (it should be easier as we only have to handle a single special case so the general mechanics in all other cases will not be affected).

bkamins:
Just one more question. What should happen in this case:
```
df = DataFrame()
df[:a] .= 1
```
currently I create `:a` but keep `df` having `0`-rows, but maybe we should create `1` row for convenience (and consistency with `DataFrame` constructor and `insertcols`; BTW: we should rethink when we want to retain automatic broadcasting in these two cases).

I understand that in this case:
```
df = DataFrame()
df[:a] .= [1,2,3]
```
we should copy the vector and create 3 rows.

nalimilan:
OK so we would special-case data frames with no columns and automatically resize them as needed? Then indeed it makes sense for `df[:a] .= 1` to create one row, as we do in other places.

bkamins:
We could also disallow broadcasting into a `DataFrame` having zero columns (this would be simpler implementation-wise). But I just want to make sure that we discuss all corner cases before I start implementing it.

nalimilan:
As you prefer. I guess it's safer from an API standpoint to throw an error when there are no columns, that way we can make a decision later based on the collected experience with `.=`.

bkamins:
@mbauman + @nalimilan I have cleaned up the code following all the comments. Given the simplifications we allow now it is much cleaner and uses only `broadcastable`, `maybeview` and `copyto!` which I hope is the expected way to do it.

I have three comments:
* We allow adding columns using broadcasting only using `df[col] .= v` syntax if `!isempty(df))`.
* In general we could consider in the future to add support for something like `copyto!(df[1:2, 1:2], rand(2,2))`, which will now throw an error as I expect that the source in `copyto!` supports `eachindex` and the produced values are `CartesianIndex` (which is not true for not-broadcasted matrices as they use linear indexing). @nalimilan - do you think we should add this support or leave it for later?
* in `copyto!` I do not check if the source and destination have the same dimensions (this is done by broadcasting infrastructure earlier); however, this means that if someone uses `copyto!` directly one can corrupt a destination (if source is too large - here Base errors) or silently produce an unexpected result (is source is too small - here the same happens in base). Also please comment if you have any perspective on this.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
+struct LazyNewColDataFrame
+    df::DataFrame
+    col::Symbol
+end
+
+# we allow LazyNewColDataFrame only for data frames with some columns
```
nalimilan:
```suggestion
# we allow LazyNewColDataFrame only for data frames with at least one column
```

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
...
+Base.axes(x::LazyNewColDataFrame) = axes(x.df[1])
+
+Base.Broadcast.broadcastable(df::AbstractDataFrame) = Matrix(df)
```
nalimilan: Do we really need to do this? That doesn't sound like the most useful interpretation to me: `f.(df)` could also return a data frame. That would preserve column names.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
...
+
+Base.maybeview(df::AbstractDataFrame, idxs...) = view(df, idxs...)
+
+function Base.maybeview(df::AbstractDataFrame, idxs)
```
nalimilan: Or just `idx` (singular)?

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
...
+    if ncol(df) == 0
+        throw(ArgumentError("Broadcasting into a data frame with no columns is not allowed"))
+    end
+    if idxs isa Symbol
+        if !haskey(df, idxs)
+            if !(df isa DataFrame)
+                throw(ArgumentError("column $idxs not found"))
```
nalimilan: Can we print suggestions of similar column names as in `getindex`?

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
...
+            end
+            return LazyNewColDataFrame(df, idxs)
+        end
+    end
+    view(df, idxs)
+end
+
+function Base.copyto!(lazydf::LazyNewColDataFrame, bc)
+    if isempty(lazydf.df)
+        throw(ArgumentError("Broadcasting creation of a column of an empty data frame is not allowed"))
```
nalimilan:
```suggestion
        throw(ArgumentError("creating a column via broadcasting is not allowed on empty data frames"))
```

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
...
+    end
+    T = mapreduce(i -> typeof(bc[i]), promote_type, eachindex(bc); init=Union{})
+    col = Tables.allocatecolumn(T, nrow(lazydf.df))
+    copyto!(col, bc)
+    lazydf.df[lazydf.col] = col
+end
+
+function Base.copyto!(df::AbstractDataFrame, bc)
+    for I in eachindex(bc)
```
nalimilan: I guess it would be more efficient to iterate column-wise. That way we wouldn't have to lookup the column for each entry, but only once.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
...
+        # a safeguard against linear index
+        row, col = Tuple(I)
+        df[row, col] = bc[I]
+    end
+    df
+end
+
+function Base.copyto!(df::AbstractDataFrame, bc::Base.Broadcast.Broadcasted{<:Base.Broadcast.AbstractArrayStyle{0}})
+    if bc.f === identity && bc.args isa Tuple{Any} && Base.Broadcast.isflat(bc)
```
nalimilan: This is equivalent, right?
```suggestion
    if bc.f === identity && bc.args isa Tuple && Base.Broadcast.isflat(bc)
```

Can you add a comment to explain what this condition means? Same for the other branch.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,63 @@
...
+    if isempty(lazydf.df)
+        throw(ArgumentError("Broadcasting creation of a column of an empty data frame is not allowed"))
+    end
+    T = mapreduce(i -> typeof(bc[i]), promote_type, eachindex(bc); init=Union{})
+    col = Tables.allocatecolumn(T, nrow(lazydf.df))
+    copyto!(col, bc)
+    lazydf.df[lazydf.col] = col
+end
+
+function Base.copyto!(df::AbstractDataFrame, bc)
```
nalimilan: Can we use a more specific type for `bc`? I'm not sure we want to officially support this definition of `copyto!` for any type.

nalimilan:
>     * We allow adding columns using broadcasting only using `df[col] .= v` syntax if `!isempty(df))`.

OK.

>     * In general we could consider in the future to add support for something like `copyto!(df[1:2, 1:2], rand(2,2))`, which will now throw an error as I expect that the source in `copyto!` supports `eachindex` and the produced values are `CartesianIndex` (which is not true for not-broadcasted matrices as they use linear indexing). @nalimilan - do you think we should add this support or leave it for later?

Why not use `CartesianIndices` instead of `eachindex`? That shouldn't be slower. Anyway (as I noted in a comment) I think for performance we should iterate column-wise.

>     * in `copyto!` I do not check if the source and destination have the same dimensions (this is done by broadcasting infrastructure earlier); however, this means that if someone uses `copyto!` directly one can corrupt a destination (if source is too large - here Base errors) or silently produce an unexpected result (is source is too small - here the same happens in base). Also please comment if you have any perspective on this.

If possible I think the signatures should be stricter to ensure they are only used for broadcasting.

src/other/broadcasting.jl
bkamins: We can only return an object that supports `eachindex`. We could add this support to `DataFrame` alternatively (but I thought it is not safe to provide `eachindex` for a data frame that would iterate all cell indices, but I am not 100% sure about it). Do you prefer this?

However, bear in mind that the column names will not be retained anyway as in broadcasting access is provided via `eachindex` framework.

In short - a data frame in broadcasting must be considered as a matrix-like object (because `axes` on a data frame returns two dimensions). We can either perform the conversion or add `eachindex`. A third option is not to add this support at all and disallow data frame in RHS of broadcasting. What would you prefer?

Finally note that currently `DataFrameRow` is collected to a `Vector` when used in LHS of broadcasting, so casting a data frame to a `Matrix` felt like a consistent approach.

src/other/broadcasting.jl
bkamins: `idxs` may be a collection of column indices so I thought about plural. But I am OK to switch to singular. If felt consistent with the method argument names above in line 11.

src/other/broadcasting.jl
bkamins: OK - I will add this

src/other/broadcasting.jl
bkamins: But we have no guarantee what `eachindex` returns and in what order, and `eachindex` is the only guaranteed iteration infrastructure for `bc`. In short: we cannot get a column from `bc` object.

src/other/broadcasting.jl
bkamins: Your suggestion is not equivalent I think. In general this is a verbatim copy from Base giving a condition when we can perform a faster operation as we can get a scalar we want to broadcast over directly.

src/other/broadcasting.jl
bkamins: This is the approach in Base. If I understand @mbauman comment correctly in some cases broadcasting infrastructure does not return `Broadcasted` object.

bkamins:
> Why not use `CartesianIndices` instead of `eachindex`?

Because only `eachindex` is guaranteed to be defined by broadcasting infrastructure (again - if I understand @mbauman correctly). `CartesianIndices` do not have to be defined.

src/other/broadcasting.jl
nalimilan: OK. What's the best way to ensure `f.(df)` returns a data frame? I guess we could implement `eachindex` if that's needed.

I see the case of `DataFrame` as similar to `NamedArray` or `AxisArray`: we should preserve the type and column names if possible.

src/other/broadcasting.jl
nalimilan: AFAICT according to https://docs.julialang.org/en/latest/manual/interfaces/#man-interfaces-broadcasting-1, `Broadcastable` objects have to implement `axes` and `getindex`, so isn't that all we need? It would be really too bad to support broadcasting but with a terrible performance...

src/other/broadcasting.jl
nalimilan: Ah, right, `Tuple{Any}` implies length 1.

src/other/broadcasting.jl
nalimilan: Are you sure? [The docs](https://docs.julialang.org/en/latest/manual/interfaces/#Working-with-Broadcasted-objects-1) seem to say that we can rely on that.

src/other/broadcasting.jl
bkamins: Good point. I think I will leave out LHS broadcasting of data frame and `DataFrameRow` from this PR to a separate PR as these functionalities are independent from the core of this PR. Thank you for digging into this.

src/other/broadcasting.jl
bkamins: I understand you refer to:

> Iterating over the `CartesianIndices` of the `axes(::Broadcasted)` and using indexing with the resulting `CartesianIndex` object to compute the result.

I was afraid that there might be some strange `eachindex` methods for custom return values of `axes`. But it makes sense to assume that we can iterate over what `axes` returns. I will change this.

src/other/broadcasting.jl
bkamins: I will add a comment in the code though.

src/other/broadcasting.jl
bkamins: The docs say we have to handle `Broadcasted` but do not explicitly say that this is the only option. Anyway - I have tested several natural options and `Broadcasted` is always passed, so I guess we can restrict the signature and if some special cases are found later we can always handle special cases.

src/other/broadcasting.jl
bkamins: I have opened #1806.

bkamins:
This should be good to have another look. Thank you.

bkamins:
> Then things like `df[:, col] .+= 1` don't work

It works. What does not work is if `col` is not present in `df`, which is expected (one can only write `df[col] .= 1` to add a column). I will add the tests.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,67 @@
...
+        throw(ArgumentError("creating a column via broadcasting is not allowed on empty data frames"))
+    end
+    T = mapreduce(i -> typeof(bc[i]), promote_type, eachindex(bc); init=Union{})
+    col = Tables.allocatecolumn(T, nrow(lazydf.df))
+    copyto!(col, bc)
+    lazydf.df[lazydf.col] = col
+end
+
+function _copyto_heper!(dfcol::AbstractVector, bc::Base.Broadcast.Broadcasted, col::Int)
+    for row in axes(dfcol, 1)
```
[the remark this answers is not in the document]
bkamins: we can use `eachindex`. It should as fast if RHS is fast to access (i.e. when the compiler does the job right, which I guess cannot be guaranteed in general).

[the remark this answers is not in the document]
bkamins: `eachindex` and `axes(dfcol,1)` will differ if someone puts some wild custom vector with non-standard indexing into a data frame, but I guess we cannot handle this anyway.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,67 @@
...
+
+function Base.copyto!(df::AbstractDataFrame, bc::Base.Broadcast.Broadcasted{<:Base.Broadcast.AbstractArrayStyle{0}})
+    # special case of fast approach when bc is providing an untransformed scalar
+    if bc.f === identity && bc.args isa Tuple{Any} && Base.Broadcast.isflat(bc)
+        for col in axes(df, 2)
+            fill!(df[col], bc.args[1][])
+        end
+        df
+    else
+        copyto!(df, convert(Broadcasted{Nothing}, bc))
```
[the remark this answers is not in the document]
bkamins: this is a pattern from Base - I it forces dispatch to the other function (i.e. if we cannot use a fast batched `fill!` we fallback to element by element assignment).

docs/src/lib/indexing.md
```diff
@@ -79,3 +79,17 @@ For performance reasons, accessing, via `getindex` or `view`, a single `row` and
 ## `setindex!`
 
 Under construction
+
+## Broadcasting
+
+`AbstractDataFrame` is converted to a matrix in broadcasting.
```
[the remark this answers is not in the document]
bkamins: right

test/broadcasting.jl
```diff
@@ -0,0 +1,221 @@
...
+    @test all(Matrix(df[2:end]) .== ones(size(df[2:end])...))
+
+    dfv = @view df[1:2, 2:end]
+    dfv[:x2] .+= [1, 1]
+    @test dfv.x2 == [2, 2]
+    @test all(Matrix(dfv[2:end]) .== ones(size(dfv[2:end])...))
+    @test all(Matrix(df[1:2, 1:2]) .== 2)
+
+    dfr = df[1, 3:end]
+    dfr[[:x4, :x5]] .= [10, 10]
```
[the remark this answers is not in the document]
bkamins: This is one of the crucial points of the design decision. If `dfr` is to be one-dimensional it has to be treated in assignment as column vector as this is what we have in Base. This is related to the fact that `axes(dfr)` returns one element tuple, not two-element tuple with first element being a `Base.OneTo(1)`.

If we want to change this we have to make `DataFrameRow` have `ndims` equal to 2. But then the question is how would it be different from a `DataFrame` with one row (it would not differ except 
 for indexing rules). If we went this way then actually I think we would have to drop having `DataFrameRow` altogether.

Put it otherwise. This is the same as with arrays in Base:

```
julia> x = ones(2,3)
2×3 Array{Float64,2}:
 1.0  1.0  1.0
 1.0  1.0  1.0

julia> x[1,:]
3-element Array{Float64,1}:
 1.0
 1.0
 1.0

julia> x[1, :] .= [10, 10, 10]
3-element view(::Array{Float64,2}, 1, :) with eltype Float64:
 10.0
 10.0
 10.0

julia> x[1, :] .= [10 10 10]
```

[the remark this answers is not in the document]
bkamins: I leave it open for a discussion.

test/broadcasting.jl
```diff
@@ -0,0 +1,221 @@
...
+
+    dfv = @view df[1:2, 2:end]
+    @test_throws BoundsError dfv[10] .= ones(3)
+    @test_throws ArgumentError dfv[:z] .= ones(3)
+    @test df == cdf
+    dfr = df[1, 3:end]
+    @test_throws BoundsError dfr[10] .= ones(3)
+    @test_throws ArgumentError dfr[:z] .= ones(3)
+    @test df == cdf
+end
```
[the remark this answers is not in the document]
bkamins: OK

test/broadcasting.jl
```diff
@@ -0,0 +1,221 @@
...
+    dfv = @view df[1:2, 2:end]
+    dfr = df[1, 3:end]
+    @test_throws DimensionMismatch df[:x1] .= rand(3, 1)
+    @test_throws DimensionMismatch dfv[:x2] .= rand(2, 1)
+    @test_throws DimensionMismatch dfr[[:x4, :x5]] .= rand(3, 1)
+end
+
+@testset "normal data frame and data frame row in broadcasted assignment - two columns" begin
+    df = copy(refdf)
+    df[[1,2]] .= Matrix(df[[1,2]]) .+ 1
```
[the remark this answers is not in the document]
bkamins: `df[[1,2]] .= df[[1,2]] .+ 1` does not work because we have just decided to move a decision what to do with broadcasting of a data frame to the other PR. Currently data frame is not supported in broadcasting (this PR makes it only to support broadcasted assignment).

Please comment in https://github.com/JuliaData/DataFrames.jl/issues/1806 what you think should be a proper design when you have time.

mbauman:
With respect to your questions:

> what are the guarantees on the value of elements produced by eachindex called on bc object. In particular (as this is a relevant possibly problematic case because data frames do not support linear indexing) - is it possible that eachindex(bc) will return a linear index if bc is two dimensional.

Yes, if `bc` is a `Broadcasted`, it will always be a `CartesianIndex` if `ndims(bc) != 1`.  I suppose it _might_ be possible for some other package to specialize `broadcasted` to return some IndexLinear non-broadcasted thing, but I don't think it'll be possible to hit this method without wrapping it in a `Broadcasted`, so I think you're safe.  It looks like we could add the `eachindex(::IndexCartesian, ::Broadcasted)` method to Base to assuredly plug this case, but I think you're pretty safe even without that.

> I wanted to make sure that I understand it correctly. The method https://github.com/JuliaData/DataFrames.jl/pull/1804/files#diff-57cbbe702db7443003fc8c590f239d70R48 (similar to one in Base) is needed only for performance. We could remove it and all should work. Is this right?

Yes, that's exactly right.  It's just a hoisting of the scalar out of the loop since — when I tested it over a year ago — Julia wasn't automatically doing that for us.  It might be worth seeing if it's still needed... there's been so many compiler improvements since then.

bkamins:
@mbauman - thank you for the comments. A small thing is that I was surprised that `axes(bc, 1)` is not supported and you have to do `axes(bc)[1]` - is this intened?

@nalimilan - I have implemented your comments. There are two conversations open (one is a design decision the other is about broadcating of a data frame on RHS).

mbauman:
> A small thing is that I was surprised that `axes(bc, 1)` is not supported and you have to do `axes(bc)[1]` - is this intened?

Somewhat. I intentionally kept the API surface of `Broadcasted`s as minimal as possible so we could continue to iterate on them throughout the 1.x era.  It'd make sense to add that method, though.

test/broadcasting.jl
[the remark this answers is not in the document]
bkamins: @nalimilan - how do you see it (I would like to make a decision and then decide on https://github.com/JuliaData/DataFrames.jl/issues/1806 after we merge this PR). Then I will move to `setindex!`. The ultimate objective is to have a release relatively soon so that I can present a "cleaned up" and released functionality during JuliaCon2019, as I run a tutorial on DataFrames.jl there.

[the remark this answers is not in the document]
nalimilan: Interesting. Both approaches have some merit. Making `DataFrameRow` act like a single-row matrix kind of makes sense, but  that wouldn't be consistent with array indexing in Base: `A[1, :]` returns a vector. That would also be inconsistent with `NamedTuple`, and as I noted at https://github.com/JuliaData/DataFrames.jl/issues/1806 I think we want to allow broadcasting these two types together.

Anyway I think I misread the line and interpreted it as `df[[:x4, :x5]] .= [10, 10]` recycling `[10, 10]` over rows, which would be much more problematic.

In the `dfr` case, what's the advantage of `.=` over just `=`? Maybe we could just disallow broadcasting vectors and matrices with `DataFrameRow` for now? Only scalar broadcasting sounds useful.

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+@testset "normal data frame and data frame row in broadcasted assignment - one column" begin
+    df = copy(refdf)
+    df[1] .+= 1
+    @test df.x1 == [2, 2, 2]
+    @test df[2:end] == refdf[2:end]
+
+    dfv = @view df[1:2, 2:end]
+    dfv[1] .+= 1
+    @test dfv.x2 == [2, 2]
+    @test all(Matrix(dfv[2:end]) .== ones(size(dfv[2:end])...))
```
nalimilan: There are still lots of `all` calls which I think can be avoided.

test/broadcasting.jl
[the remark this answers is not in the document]
bkamins: OK - I will restrict broadcasting of `DataFrameRow` to scalars for now.
Actually this is related to #1806. I will comment on it there (but what you propose is safe to do).

test/broadcasting.jl
bkamins: OK - I will look through them.

test/broadcasting.jl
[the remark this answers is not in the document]
bkamins: I have thought about it more and my conclusion is that I would rather allow vectors.

First regarding

> Only scalar broadcasting sounds useful.

```
dfr .= 1 .+ [1,2,3] .+ sin.([5,6,7])
```

will not allocate, while

```
dfr = 1 .+ [1,2,3] .+ sin.([5,6,7])
```

will allocate.

Secondly - if we think of `DataFrameRow` as a kind of mutable `NamedTuple` it is hard to imagine it has different dimensionality than `1`.

Thirdly (this is a minor issue) - in order to restrict ourselves to scalars `DataFrameRow` would have to pretend being a scalar (which would in some corner cases produce strange errors).

Let us think about it a bit more. I will shortly open another PR regarding to `DataFrameRow` design that I think should be discussed before we move forward.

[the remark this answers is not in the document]
bkamins: Also we define `Vector(dfr)`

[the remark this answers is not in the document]
nalimilan: Fine with me. It's a bit weird that a data frame row behaves like a column vector, but it's not our fault if vectors are considered as columns in Julia.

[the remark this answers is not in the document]
bkamins: OK - so I will leave it as is and emphasize in the documentation. Actually I want to make `DataFrameRow` behave like a mutable `NamedTuple` (but it is also considered 1-dimensional, of course it is not mutable). See https://github.com/JuliaData/DataFrames.jl/pull/1813 for additional changes that improve this consistency.

test/broadcasting.jl
bkamins: I have cleaned out all `all` that were possible to remove.

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+    @test df.x2 == [2, 2, 2]
+    @test df[3:end] == refdf[3:end]
+
+    dfv = @view df[1:2, 3:end]
+    dfv[:, [:x3,:x4]] .= Matrix(dfv[[:x3,:x4]]) .+ [1 1
+                                                    1 1]
+    @test dfv.x3 == [2, 2]
+    @test dfv.x4 == [2, 2]
+    @test dfv[3:end] == refdf[1:2, 5:end]
+    @test all(Matrix(df[1:2, 1:2]) .== 2)
```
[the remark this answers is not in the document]
bkamins: OK - this is possible :). It was longer. In what cases do you think your test would fail while mine would pass (I thought they are equivalent)?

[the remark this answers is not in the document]
bkamins: But I will update these tests as requested.

bkamins:
Given the decision in #1806 the only question here if we want to support `DataFrameRow` on LHS of assignment. The reason is that we disallow `DataFrameRow` on RHS (and this is implemented already in https://github.com/JuliaData/DataFrames.jl/pull/1813 - please comment on that PR if you have time).

I think that the only risk in allowing `DataFrameRow` on LHS is that in the future we might decide to treat it as 2-dimensional with 1 row, but I think this is not likely (I do not see how `NamedTuple` could get to be interpreted this way). We still can disallow it with little harm, simply `df[1, :] .= v` will not work and someone will have to write `df[1:1, :] .= v` (and possibly transpose `v` if it is a vector, for scalars it will be the same behavior of both syntaxes) till we decide what `df[1, :]` means in broadcasting (for a reference of future readers: `df[1, :]` produces a `DataFrameRow` while `df[1:1, :] .= v` produces a `DataFrame` with one row).

bkamins:
@nalimilan Any opinion on allowing the support of `DataFrameRow` on LHS? This is I think the last blocking issue in the "broadcasting" story.

nalimilan:
Can we allow `DataFrameRow` on the LHS, but only when the RHS is not an `AbstractArray`? Or is that too complex? I think we're mostly interested in the scalar case.

bkamins:
> Can we allow `DataFrameRow` on the LHS, but only when the RHS is not an `AbstractArray`?

We could (similarly to https://github.com/JuliaData/DataFrames.jl/pull/1804/files#diff-57cbbe702db7443003fc8c590f239d70R55), but 1 day later I will ask you a question what `dfr[:] = vector` should do when I open a `setindex!` PR, so it is probably better decide it here anyway (or decide that in `setindex!` we will not allow a vector on RHS).

This is what we have now:
```
julia> using DataFrames

julia> df = DataFrame(rand(2,3))
2×3 DataFrame
│ Row │ x1       │ x2       │ x3        │
│     │ Float64  │ Float64  │ Float64   │
├─────┼──────────┼──────────┼───────────┤
│ 1   │ 0.880536 │ 0.995971 │ 0.0321657 │
│ 2   │ 0.37657  │ 0.477013 │ 0.216326  │

julia> df[1, :] = [1,2,3]
ERROR: MethodError: Cannot `convert` an object of type Array{Int64,1} to an object of type Float64

julia> df[1, :] = DataFrame(a=1,b=2,c=3)
1×3 DataFrame
│ Row │ a     │ b     │ c     │
│ 1   │ 1     │ 2     │ 3     │

julia> df
2×3 DataFrame
│ Row │ x1      │ x2       │ x3       │
│     │ Float64 │ Float64  │ Float64  │
├─────┼─────────┼──────────┼──────────┤
│ 1   │ 1.0     │ 2.0      │ 3.0      │
│ 2   │ 0.37657 │ 0.477013 │ 0.216326 │
```

In other words, now `dfr[:] = vector` now implicitly broadcasts a vector and tries to assign it to every column.

---

In summary - what I want to say is that we should have a convenient way to replace values in some row of a data frame with new values. Currently it is only possible if we convert RHS to a data frame. This - frankly - makes little sense and is a performance problem. We can provide this possibility either via broadcasting or `setindex!`. If we provide it via. `setindex!` probably we can restrict broadcasting to scalars, but the question is what will be the value of such restriction then (the point is that if we decide `dfr[:] = vector` is allowed then why `dfr .= vector` should be disallowed).

The same point applies to `dfr[:] = tuple` and `dfr .= tuple` (as `tuple` is considered column oriented in broadcasting):

```
julia> x = rand(2,3)
2×3 Array{Float64,2}:
 0.100988  0.440462  0.165818
 0.755196  0.477914  0.947304

julia> x .= (1,2)
2×3 Array{Float64,2}:
 1.0  1.0  1.0
 2.0  2.0  2.0
```

Also in `push!` we already assume that we push column vectors and tuples to a `DataFrame` to add a row.

Sorry for the laundry list, but the API issue here is complex for me so I want to spell out all issues I have with it.

nalimilan:
Allowing vectors on the RHS sounds like the way to go then. But just to be sure, this means we have to consider `DataFrameRow` as one-dimensional everywhere (at least for broadcasting), right?

bkamins:
> one-dimensional everywhere (at least for broadcasting), right?

Right - I would say an even stronger **everywhere**. In particular I think that the only problem is that:
```
df .= dfr

```
makes no sense (and we currently disallow it anyway)

but
```
df .= Ref(dfr)
```
makes sense and does what is expected.

In general `dfr` will behave the same as `collect(dfr)` or `copy(dfr)` depending on the context (if the context expects values only or column names also).

nalimilan:
What does `df .= Ref(dfr)` mean? Fill each cell with a `DataFrameRow`?

> In general `dfr` will behave the same as `collect(dfr)` or `copy(dfr)` depending on the context (if the context expects values only or column names also).

This sounds like a good property, even if it's annoying that we can't treat `DataFrameRow` as row vectors in other situations.

bkamins:
You are right `df .= Ref(dfr)` makes no sense - or it makes sense exactly in the way you have written (sorry for being too fast - I remembered that broadcasting of `dfr` was problematic).

Maybe let us add `permutedims` to `dfr` which would be a lazy wrapper, then this will work

```
df .= permutedims(dfr)
```

In general in Julia it would make much more sense that DataFrame would have size ncols x nrows (actually some packages assume this layout, e.g. I think Flux.jl if I remember correctly). But probably we have to live with this.

### Another thought

An alternative is to define `DataFrameRow` as two dimensional 1 x ncols object but also allowing linear indexing - as opposed to data frame that does not allow linear indexing (then we would not have all these problems).

Then `collect(dfr)` would produce a `Matrix` so we would loose some consistency with `NamedTuple`, but this probably we could live with this.

nalimilan:
> In general in Julia it would make much more sense that DataFrame would have size ncols x nrows (actually some packages assume this layout, e.g. I think Flux.jl if I remember correctly). But probably we have to live with this.

But then it would be at odds with the fact that `Matrix` is column-major (i.e. iterating over columns is faster than iterating over rows)...

> An alternative is to define `DataFrameRow` as two dimensional 1 x ncols object but also allowing linear indexing - as opposed to data frame that does not allow linear indexing (then we would not have all these problems).
> 
> Then `collect(dfr)` would produce a `Matrix` so we would loose some consistency with `NamedTuple`, but this probably we could live with this.

That inconsistency wouldn't be great IMO. That would make `DataFrameRow` a quite complex hybrid.

bkamins:
OK - we investigated all options I could see 😄.

The conclusion is: `DataFrameRow` has `ndim` equal to `1` and behaves like a column vector with all the consequences.

Additionally - for the time being `DataFrameRow` is allowed to take part in broadcasting on LHS, but is not allowed to take part in broadcasting on RHS (for consistency with `NamedTuples` - this is not a part of this PR).

Actually this is the functionality that is implemented now in this PR, so it should be good for a final review. Thank you.

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+
+    df = copy(refdf)
+    dfv = @view df[1:2, 2:end]
+    dfr = df[1, 3:end]
+    @test_throws DimensionMismatch df[:x1] .= rand(3, 1)
+    @test_throws DimensionMismatch dfv[:x2] .= rand(2, 1)
+    @test_throws DimensionMismatch dfr[[:x4, :x5]] .= rand(3, 1)
+    @test_throws DimensionMismatch df[:, :x1] .= rand(3, 1)
+    @test_throws DimensionMismatch dfv[:, :x2] .= rand(2, 1)
```
nalimilan: This block tests single-column matrices, but the code above only tests vectors. Maybe change this block to tests vectors, but put the whole set in a loop, wrapping vectors in a function which is `identity` for the first run, and `reshape(x, :, 1)` for the second run?

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+    dfv = @view df[1:2, 2:end]
+    dfv[:, 1] .+= [1, 1]
+    @test dfv.x2 == [2, 2]
+    @test dfv[2:end] == refdf[1:2, 3:end]
+    @test all(Matrix(df[1:2, 1:2]) .== 2)
+
+    df = copy(refdf)
+    dfv = @view df[1:2, 2:end]
+    dfr = df[1, 3:end]
+    @test_throws DimensionMismatch df[1] .= rand(3, 1)
```
nalimilan: This block tests single-column matrices, but the code above only tests vectors. Maybe change this block to tests vectors, but put the whole set in a loop, wrapping vectors in a function which is `identity` for the first run, and `reshape(x, :, 1)` for the second run?

(Same below.)

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+    df = copy(refdf)
+    df[1] .+= 1
+    @test df.x1 == [2, 2, 2]
+    @test df[2:end] == refdf[2:end]
+
+    dfv = @view df[1:2, 2:end]
+    dfv[1] .+= 1
+    @test dfv.x2 == [2, 2]
+    @test dfv[2:end] == refdf[1:2, 3:end]
+    @test all(Matrix(df[1:2, 1:2]) .== 2)
```
nalimilan: Also test that `df` is correct? Same in other places when using views and `DataFrameRow`.

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+
+    dfr = df[1, 3:end]
+    dfr[end-1:end] .= 10
+    @test all(Vector(dfr) .== [1, 10, 10])
```
```suggestion
    @test Vector(dfr) == [1, 10, 10]
```

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
+module TestBroadcasting
+
+using Test, DataFrames
+
+refdf = DataFrame(ones(3, 5))
```
nalimilan: It would be more robust to use random values: when they are all equal, the code could be completely broken (e.g. recycle the first value) and the tests wouldn't catch it.

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+    @test_throws DimensionMismatch df[:x1] .= rand(3, 1)
+    @test_throws DimensionMismatch dfv[:x2] .= rand(2, 1)
+    @test_throws DimensionMismatch dfr[[:x4, :x5]] .= rand(3, 1)
+    @test_throws DimensionMismatch df[:, :x1] .= rand(3, 1)
+    @test_throws DimensionMismatch dfv[:, :x2] .= rand(2, 1)
+end
+
+@testset "normal data frame and data frame row in broadcasted assignment - two columns" begin
+    df = copy(refdf)
+    df[[1,2]] .= Matrix(df[[1,2]]) .+ 1
```
nalimilan: Maybe also test that this works without the `.+ 1`?

Also I don't see tests with multiple columns on the LHS but a vector or single-column matrix on the RHS.

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+    @test_throws DimensionMismatch dfv[:, [:x3,:x4]] .= rand(2, 10)
+end
+
+@testset "assignment to a whole data frame and data frame row" begin
+    df = copy(refdf)
+    df .= 10
+    @test all(Matrix(df) .== 10)
+    dfv = view(df, 1:2, 1:4)
+    dfv .= 100
+    @test (all(Matrix(df[1:2, 1:4]) .== 100))
```
nalimilan: No need for enclosing parentheses.

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+    dfv[:, :] .= 100
+    @test (all(Matrix(df[1:2, 1:4]) .== 100))
+    @test (all(Vector(df[3, 1:4]) .== 10))
+end
+
+@testset "extending data frame in broadcasted assignment - one column" begin
+    df = copy(refdf)
+    df[:a] .= 1
+    @test all(Matrix(df) .== 1)
+    @test names(df)[end] == :a
```
nalimilan: Also test that `df[1:end-1] == refdf` just in case (here and below).

test/broadcasting.jl
```diff
@@ -0,0 +1,364 @@
...
+    df[:c1] .= v
+    @test df.c1 == v
+    @test df.c1 !== v
+    @test df.c1 isa CategoricalVector
+    @test levels(df.c1) == levels(v)
+    @test levels(df.c1) !== levels(v)
+    df[:c2] .= v[1]
+    @test df.c2 == [1,1,1]
+    @test df.c2 isa CategoricalVector
+    @test levels(df.c2) != levels(v)
```
nalimilan: This is somewhat unfortunate. We should allow `allocatecolumn` to take a value rather than just a type so that levels can be preserved (not for this PR).

src/other/broadcasting.jl
```diff
@@ -0,0 +1,72 @@
...
+        end
+    end
+    view(df, idxs)
+end
+
+function Base.copyto!(lazydf::LazyNewColDataFrame, bc::Base.Broadcast.Broadcasted)
+    if isempty(lazydf.df)
+        throw(ArgumentError("creating a column via broadcasting is not allowed on empty data frames"))
+    end
+    T = mapreduce(i -> typeof(bc[i]), promote_type, eachindex(bc); init=Union{})
```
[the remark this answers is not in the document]
bkamins: This is tricky because `bc` can have `Any` eltype (or in general - some abstract eltype). I could use `[v for v in bc_column]` style, and it would do automatic type promotion. However, there is a problem with `CategoricalArrays`, as if a column is categorical a comprehension will not create a proper container type (as you know very well 😄).

So - in summary - if we accept that we do not produce a narrowest possible type for the column this part of code can be dropped.

@nalimilan Please let me know which approach you would prefer.

In an ideal world I think that compiler should optimize out this `mapreduce` if it knows that `bc` has a homogeneous concrete eltype, but I have not checked if it happens.

test/broadcasting.jl
bkamins: I have opened https://github.com/JuliaData/Tables.jl/issues/99 to keep track of it

src/other/broadcasting.jl
```diff
@@ -0,0 +1,72 @@
...
+            fill!(df[col], bc.args[1][])
+        end
+        df
+    else
+        copyto!(df, convert(Broadcasted{Nothing}, bc))
+    end
+end
+
+function Base.copyto!(dfr::DataFrameRow, bc::Base.Broadcast.Broadcasted)
+    for I in eachindex(bc)
```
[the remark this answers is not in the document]
bkamins: I can change it but it will not help as `setindex!` does not support `@inbounds` yet.

Also it is a bit delicate because it can fail because `dfr` might potentially be corrupted (i.e. point at non-existing row or column) - so although `@inbounds` is safe here it might be unsafe in a broader context (we could add manual checks as you propose though).

Given this, and the fact that the operation is not type stable anyway so boundschecking will not be a significant cost anyway I would leave it as is.

test/broadcasting.jl
bkamins: I have made explicit non-random tests now (that should be relatively safe to recycling while allowing "eye" checking of the correctness of the result)

test/broadcasting.jl
bkamins: done - now I always check the whole `df` against what is expected.

test/broadcasting.jl
bkamins: I am not fully clear what this would achieve, all operations on `reshape(x, :, 1)` if `x` is a vector should throw an error. I added the test cases with `reshape` (that throw an error) and with some broadcasted vector transformation (that work OK).

test/broadcasting.jl
bkamins: see the remark above

test/broadcasting.jl
bkamins: I added some tests for each relevant case, but not all combinations as this would essentially test broadcasting mechanisms in base (i.e. that lower dimensional objects get properly expanded in `bc` to a matrix; in DataFrames.jl the crucial thing to test is if we properly assign values taken from a ready `bc`) and would add ~500 LOC to the tests here.

test/broadcasting.jl
bkamins: fixed everywhere

test/broadcasting.jl
bkamins: fixed

bkamins:
Thank you for the comments. I have rewritten the tests to be more robust (it would not hurt to have second eyes look at the - although now they are really cumbersome to check unfortunately). I have left four your comments unresolved as you might have some further comments on these issues.

src/other/broadcasting.jl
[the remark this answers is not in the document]
nalimilan: Ah, I guess `bc` could have eltype `Any` when calling e.g. `f.(x)`. Then that approach is fine. Though maybe check that the compiler optimizes this out when it has a concrete eltype, and if not add a check to skip it manually.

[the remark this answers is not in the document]
bkamins: OK - I will have look at it.

test/broadcasting.jl
nalimilan: Woops, yes, that's not the right place for this comment. Actually I should have said that in the test set below, column vectors and single-column matrices should be tested when assigning to a data frame. Does that make sense?

test/broadcasting.jl
```diff
@@ -0,0 +1,501 @@
+module TestBroadcasting
+
+using Test, DataFrames
+
+refdf = DataFrame(DataFrame(reshape(1.5:15.5, (3,5))))
```
```suggestion
refdf = DataFrame(reshape(1.5:15.5, (3,5)))
```

test/broadcasting.jl
bkamins: Yes :) - I already added column vectors. I will add single column matrices for safety (but I test general matrices and this should be handled correctly by Base).

nalimilan: OK I see you've added a test for that.

test/broadcasting.jl
quinnj: Why is this unfortunate? Does having consistent levels really matter? I'm not against adjusting allocatecolumn, but maybe we should also figure out a more general solution in DataAPI.jl that allows other array types a similar kind of flexibility.

nalimilan: It's unfortunate because the whole point of `CategoricalValue` is to preserve custom ordering of levels. If we allowed `allocatecolumn` to take values and not just types, any array type could reuse that mechanism if needed.

bkamins: My thinking is similar to @nalimilan. The whole design idea of `CategoricalValue` is that it carries over the level information if I understand the original intention of @nalimilan correctly here.

If someone does not need this (ignoring other differences in functionalities) then probably `PooledArray` is what should be used.

Regarding the generality required. Currently there are no important types that carry over metadata AFAIK, but e.g. in the future we might have metadata attached to a `DataFrame` object.

bkamins:
I have added tests, fixed one type instability in `axes` definition and added a "fast" path for most common simple case of writing `df[:col] .= scalar`.

Here are some benchmarks:
```
julia> function f2()
           for i in 1:10
               df = DataFrame(x=1:10^6)
               df[:a] = 1
               df[:b] = 1
               df[:c] = 1
               df[:d] = 1
               df[:e] = 1
               df[:f] = 1
               df[:g] = 1
               df[:h] = 1
               df[:i] = 1
               df[:j] = 1
           end
           nothing
       end

julia> function f3()
           for i in 1:10
               df = DataFrame(x=1:10^6)
               df[:a] .= 1:10^6
               df[:b] .= 1:10^6
julia> function f1()
           for i in 1:10
               df = DataFrame(x=1:10^6)
               df[:a] .= 1
               df[:b] .= 1
               df[:c] .= 1
               df[:d] .= 1
               df[:e] .= 1
               df[:f] .= 1
               df[:g] .= 1
               df[:h] .= 1
               df[:i] .= 1
               df[:j] .= 1
           end
           nothing
       end

julia> function f2()
           for i in 1:10
               df = DataFrame(x=1:10^6)
               df[:a] = 1
               df[:b] = 1
               df[:c] = 1
               df[:d] = 1
               df[:e] = 1
               df[:f] = 1
               df[:g] = 1
               df[:h] = 1
               df[:i] = 1
               df[:j] = 1
           end
           nothing
       end

julia> function f3()
           for i in 1:10
               df = DataFrame(x=1:10^6)
               df[:a] .= 1:10^6
               df[:b] .= 1:10^6
               df[:c] .= 1:10^6
               df[:d] .= 1:10^6
               df[:e] .= 1:10^6
               df[:f] .= 1:10^6
               df[:g] .= 1:10^6
               df[:h] .= 1:10^6
               df[:i] .= 1:10^6
               df[:j] .= 1:10^6
           end
           nothing
       end
f3 (generic function with 1 method)

julia> function f4()
           for i in 1:10
               df = DataFrame(x=1:10^6)
               df[:a] = 1:10^6
               df[:b] = 1:10^6
               df[:c] = 1:10^6
               df[:d] = 1:10^6
               df[:e] = 1:10^6
               df[:f] = 1:10^6
               df[:g] = 1:10^6
               df[:h] = 1:10^6
               df[:i] = 1:10^6
               df[:j] = 1:10^6
           end
           nothing
       end
f4 (generic function with 1 method)

julia> @benchmark f1()
BenchmarkTools.Trial:
  memory estimate:  839.33 MiB
  allocs estimate:  1960
  --------------
  minimum time:     1.289 s (51.17% GC)
  median time:      1.342 s (52.85% GC)
  mean time:        1.348 s (52.60% GC)
  maximum time:     1.418 s (54.07% GC)
  --------------
  samples:          4
  evals/sample:     1

julia> @benchmark f2()
BenchmarkTools.Trial:
  memory estimate:  839.28 MiB
  allocs estimate:  860
  --------------
  minimum time:     1.300 s (50.47% GC)
  median time:      1.347 s (52.75% GC)
  mean time:        1.347 s (52.59% GC)
  maximum time:     1.394 s (54.26% GC)
  --------------
  samples:          4
  evals/sample:     1

julia> @benchmark f3()
BenchmarkTools.Trial:
  memory estimate:  839.33 MiB
  allocs estimate:  1960
  --------------
  minimum time:     1.424 s (46.23% GC)
  median time:      1.490 s (47.96% GC)
  mean time:        1.487 s (47.94% GC)
  maximum time:     1.543 s (49.47% GC)
  --------------
  samples:          4
  evals/sample:     1

julia> @benchmark f4()
BenchmarkTools.Trial:
  memory estimate:  839.28 MiB
  allocs estimate:  760
  --------------
  minimum time:     1.302 s (50.31% GC)
  median time:      1.358 s (52.73% GC)
  mean time:        1.357 s (52.53% GC)
  maximum time:     1.411 s (54.19% GC)
  --------------
  samples:          4
  evals/sample:     1
```

The conclusion is that in "fast common patch" we are on par to what we had. If there is a need of `mapreduce` there is less than 10% penalty, but I guess this will not be a common use case (in such case you can still use the `df[:col] = vector` assignment anyway).

bkamins:
@nalimilan I know it is approved, but the change is significant to please confirm to me that you are OK with this to be merged 😄.

bkamins:
I am going to merge it and move forward to further broadcasting/setindex! issues if there are no more comments on this.

src/other/broadcasting.jl
```diff
@@ -28,7 +28,12 @@ function Base.copyto!(lazydf::LazyNewColDataFrame, bc::Base.Broadcast.Broadcaste
     if isempty(lazydf.df)
         throw(ArgumentError("creating a column via broadcasting is not allowed on empty data frames"))
     end
-    T = mapreduce(i -> typeof(bc[i]), promote_type, eachindex(bc); init=Union{})
+    if bc isa Base.Broadcast.Broadcasted{<:Base.Broadcast.AbstractArrayStyle{0}} &&
+       bc.f === identity && bc.args isa Tuple{Any} && Base.Broadcast.isflat(bc)
+        T = typeof(bc.args[1][])
```
nalimilan: So does this cover all cases where the eltype is known, or would it be useful to have an `if isconcrete(eltype(bc))` branch?

bkamins: `eltype(bc)` is a complex type that does not give information about what will be the type of `bc[idx]`.
 We cover only a simple case when the transformation function is `identity` and we have a scalar that is broadcasted (but I guess this is exactly the important case).

Why in general it is hard to cover? Because you can have something like e.g.:

```
x .= (v -> rand() ? 1 : "a").(1)
```
(I give an extreme case, but in general the point is that `eltype(bc)` is type BEFORE the function applied to the collection is evaluated), the simple case is x.= scalar, as we know that the function is `identity` and we have only one value so we can take it.

nalimilan: OK. I wonder whether the compiler is able to compute the element type statically when applying a type-stable function, instead of doing the full `mapreduce` call (which will be quite slow).

Actually, a more standard approach would probably be to allocate a column with the type of the first element, and widen the type if needed like `collect` does. I wonder whether we could reuse some functions from Tables.jl to do that. Or maybe from the grouping code.

nalimilan: For example, this seems to work. Maybe there's a better way to avoid the `nothing` hack, though:
```julia
julia> itr = Base.Broadcast.broadcasted(x -> rand() > 0.5 ? 1 : 1.5, 1:100)
Base.Broadcast.Broadcasted(getfield(Main, Symbol("##31#32"))(), (1:100,))

julia> last(Tables.columntable(nothing=>i for i in itr))
100-element Array{Float64,1}:
```

nalimilan: Actually we already have `copyto_widen` in grouping.jl. Widening the type of the second argument should be enough, right?

bkamins: We can iteratively `push!` or widen column (this is what Tables.jl essentially does). But I would leave it for later as this does not change API, the performance gains shall not be big (if any - note that now we have ~10% penalty) and the case we are discussing here is rare and can be handled more simply by writing `df[new_col] = vector_value`.

bkamins: > Actually we already have `copyto_widen!` in grouping.jl

I was writing in parallel and did not read this comment. I will have a look at it and report back.

bkamins: This works (I have to minimally tweak `copyto_widen!` for it to work with `bc`):
```
function copyto_widen!(res::AbstractVector{T}, bc::Base.Broadcast.Broadcasted) where T
    @inbounds for i in eachindex(bc)
        val = bc[i]
        S = typeof(val)
        if S <: T || promote_type(S, T) <: T
            res[i] = val
        else
            newres = Tables.allocatecolumn(promote_type(S, T), length(x))
            return copyto_widen!(newres, bc)
        end
    end
    return res
end

function Base.copyto!(lazydf::LazyNewColDataFrame, bc::Base.Broadcast.Broadcasted)
    if isempty(lazydf.df)
        throw(ArgumentError("creating a column via broadcasting is not allowed on empty data frames"))
    end
    if bc isa Base.Broadcast.Broadcasted{<:Base.Broadcast.AbstractArrayStyle{0}} &&
       bc.f === identity && bc.args isa Tuple{Any} && Base.Broadcast.isflat(bc)
        T = typeof(bc.args[1][])
        col = Tables.allocatecolumn(T, nrow(lazydf.df))
        copyto!(col, bc)
    else
        T = typeof(bc[iterate(eachindex(bc))[1]])
        col = copyto_widen!(Tables.allocatecolumn(T, nrow(lazydf.df)), bc)
    end
    lazydf.df[lazydf.col] = col
end
```

It is a bit faster (5% at most) than the current implementation. Do you want to switch to it then?

bkamins: ah - and it will be slower if the "offending" observation with different element type is at the end of the vector (because then it has to "redo" the work twice).
Also - it just threw me `EXCEPTION_ACCESS_VIOLATION` error.
In summary - I would stay with what we have now as it is safe and fast enough.

nalimilan: My idea was that you would then be able to drop the branch, since `copyto_widen!` is efficient when the type doesn't change. So that would simplify things. I also don't think it would be slower for the type-unstable case, since the `mapreduce` call will probably be quite slow, in particular since iteration may require evaluating one or several functions for complex cases.

But the crash you mention is worrying. If you can't work around it, can you push a branch somewhere and file an issue in Julia with a reproducer?

bkamins: The `copyto_widen!` approach was very slow when some new type occurred at the end of the vector. Also in general it was a bit slower than `copyto!` in the fast branch.

However, I have just pushed a cleaner approach. What it does it:
* it does normal broadcasting (and has a performance of normal broadcasting) in all most all cases
* except if we have in the end eltype that is categorical in which case one additional allocation is performed (as unfortunately normal broadcasting does not handle it well - but this is a more general problem I am sure you are aware of - if we solve it in general then we also can simplify the code here)

So in short - for now - what I propose should be maximally fast unless we broadcast over a vector of categorical values, in which case one additional copy is performed. I think it is an acceptable compromise.

 And as I have said - if in CategoricalArrays.jl you decide to add a special `Base.Broadcast.materialize` methods for categorical arrays we will be fast in all cases and can remove the unnecessary copy here (if it is not clear what should be added to CategoricalArrays.jl I can expand on it).

nalimilan: OK, we need to fix `broadcast` on `CategoricalArray`. Maybe we can get things working by defining `Base.similar(bc::Broadcasted{DefaultArrayStyle{N}}, ::Type{CategoricalString})` and so on. But we should really define something like `allocatecolumn` in Base and have `broadcast` use that, or we'll always suffer from this kind of inconsistency.

Though maybe we'll also need an optimized `materialize` method to speed up special cases, in particular assigning a single `CategoricalValue` to a column (either an existing `CategoricalVector` or a new column). There's also again the issue of propagating levels, which cannot be addressed by just overloading `similar`. What do you think?

bkamins: My thinking is:
### Simple thing - what to do with this PR:
* I have made yet another improvement - now it is fast in all cases except a complex broadcast when we apply a function to a broadcasted `CategoricalVector` that is really unlikely (I have added such complex broadcast to the tests)
* therefore if you are OK with this I would merge this PR

### Difficult thing - what to do in the long run with `CategoricalArray`:

* a half-way solution that will fix this PR in general (even in the complex case) is to make this work correctly:
```
julia> identity.(categorical(["a"]))
1-element Array{CategoricalString{UInt32},1}:
 "a"
```
fixing this is hopefully relatively easy and can be done in CategoricalArrays.jl

* a long term solution is to have `allocatecolumn` or something similar in Base as a general solution so that things like:
```
julia> [v for v in categorical(["a"])]
1-element Array{CategoricalString{UInt32},1}:
 "a"
```
works OK.

Now - I was surprised by this:
```
julia> map(x->x, categorical(["a"]))
1-element CategoricalArray{String,1,UInt32}:
 "a"
```
I did not have time to investigate (probably you know why this works) but maybe the idea used there can be reused (if I replace `CategoricalVector` by broadcasted version of it it does not work though).

nalimilan: OK, I think I've found a solution similar to what we did for `map`: https://github.com/JuliaData/CategoricalArrays.jl/pull/193. If we merge and tag a release, do you think we can get rid of the special casing here?

bkamins: Yes - that should fix it.

bkamins:
CI will fail now. I will retrigger it after https://github.com/JuliaRegistries/General/pull/1254 is merged.

bkamins:
As noted on discourse it works by using `similar(Vector{T}, nrow(lazydf.df))` which is strange as actually sometimes we do not get `Vector` but some other container type as a result.

nalimilan:
Yes that's kind of strange, but OTOH you're not supposed to store `CategoricalValue` in an `Array`. `CategoricalArray{T}` behaves like `Array{CategoricalValue{T}}` for all purposes.

src/other/broadcasting.jl
```diff
@@ -0,0 +1,77 @@
...
+Base.axes(x::LazyNewColDataFrame) = (Base.OneTo(nrow(x.df)),)
+
+Base.maybeview(df::AbstractDataFrame, idxs...) = view(df, idxs...)
+
+function Base.maybeview(df::AbstractDataFrame, idxs)
+    if ncol(df) == 0
+        throw(ArgumentError("Broadcasting into a data frame with no columns is not allowed"))
+    end
+    if idxs isa Symbol
+        if !haskey(df, idxs)
```
nalimilan:
```suggestion
        if !haskey(index(df), idxs)
```

bkamins:
I have merged this PR with master to make sure all is in sync.

Issue #1809, opened 2019-05-14
Unnecessary copy for any Queryverse source


davidanthoff:
I think (if I understand the current state of constructors and Tables.jl correctly) that we have a pretty major regression for the entire Queryverse integration with the recent versions because essentially anything a la:
```julia
df = DataFrame(anything_queryverse)
```
makes an unnecessary temporary copy of all the columns during the construction of the `DataFrame` instance.

I think here is what is happening currently: 

a) if `anything_queryverse` implements the `get_columns_copy_using_missing` interface from [TableTraits.jl](https://github.com/queryverse/TableTraits.jl/blob/816661ce6c5e16e92a161a7fa35201c597a023ad/src/TableTraits.jl#L35), then `Tables.columns` will return the column vectors that `TableTraits.get_columns_copy_using_missing ` returned, and then the `DataFrame` constructor will make a copy of each of these columns and use those. That is not good, because the whole point of the `get_columns_copy_using_missing` interface is that the source allocates a set of column vectors that the sink then owns, i.e. they are by the definition of the interface already copies that the sink can use.

b) if `anything_queryverse` implements the iterable tables interface (i.e an iterator of named tuples), then `Tables.columns` will allocate new column vectors and populate them with the values from the source, and then return those. And then the `DataFrame` constructor makes copies of those vectors, which again are really unnecessary large temporary allocations. After all, those vectors that are being allocated are _only_ allocated for this specific sink, so really they are best interpreted as copies for the sink as well.

I'm not sure what a good solution to this is, but it would be good to find some solution to this, given that this used to work well for a long time before the recent set of changes.

It would of course be easy to solve this if DataFrames.jl implemented the TableTraits.jl interfaces directly: they are explicit about column ownership, so the `DataFrame` constructor would always know whether it is getting columns from a source that it (DataFrame) then owns. I think the difficulty here stems from the fact that `Tables.columns` does not distinguish between a case where a source is returning a copy of columns and one where it is returning something that is still owned by the source. So what happens in the end is that `Table.columns` when called on any TableTraits.jl source will return vectors that are already copies, but for other sources `Tables.columns` can return a view. As the consumer, the `DataFrame` constructor then doesn't really know what to do: in the first case it shouldn't make a copy of what is returned by `Tables.columns`, in the second it should. I think the core problem is that those two cases are now conflated in the `Tables.columns` function.

@quinnj any idea what one could do about this?

[...]

PR #1810, opened 2019-05-14
Avoid a copy when materializing some TableTraits sources

davidanthoff:
This is a temporary, partial solution for #1809, until there is a way to handle this situation with Tables.jl. For now it fixes a performance regression for some cases described in #1809 (but not all of them). Would be great if this could be put into a patch release soon!

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -17,6 +17,10 @@ fromcolumns(x; copycols::Bool=true) =
               copycols=copycols)
 
 function DataFrame(x::T; copycols::Bool=true) where {T}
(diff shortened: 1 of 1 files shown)


davidanthoff:
I believe the tests [here](https://github.com/JuliaData/DataFrames.jl/blob/master/test/tabletraits.jl) should already exercise this code path.

I think the only package implementing this interface right now is https://github.com/queryverse/CSVFiles.jl, but we plan to add it to the other file IO packages in Queryverse eventually.

quinnj:
Closing in favor of #1811

PR #1811, opened 2019-05-15
Fix #1809 by avoiding making unnecessary copies when Tables signals t…

### LINKED ISSUE #1809: Unnecessary copy for any Queryverse source

quinnj:
…hat the incoming source has already been copied

cc: @davidanthoff @bkamins @nalimilan 

Relies on https://github.com/JuliaData/Tables.jl/pull/94


bkamins:
Looks good. We should merge it after https://github.com/JuliaData/Tables.jl/pull/94 is released, unless the design there is changed following what @davidanthoff commented there.

davidanthoff:
I think I would prefer that we fix this first (so merge https://github.com/JuliaData/Tables.jl/pull/94 and this PR here and tag). The other comment I made over there is really a larger, meta issue, and I think it would be better if that didn't hold up a fix.

src/other/tables.jl
```diff
@@ -11,6 +11,10 @@ Tables.materializer(df::AbstractDataFrame) = DataFrame
 
 getvector(x::AbstractVector) = x
 getvector(x) = collect(x)
+fromcolumns(x::Tables.CopiedColumns; kw...) =
```
nalimilan: Better not accept any keyword argument but ignore them: 
```suggestion
fromcolumns(x::Tables.CopiedColumns; copycols::Bool=true) =
```

quinnj: Why? It's an internal function and I was just trying to emphasize that `copycols` is ignored in this case.

nalimilan: Well it's a matter of taste, but `kw...` isn't very explicit.

quinnj:
Ok, I think this is good to go; not sure why appveyor check got messed up here, but the results for the most recent commit are all green: https://ci.appveyor.com/project/nalimilan/dataframes-jl/builds/24597257

bkamins:
@davidanthoff if we merge this do we need https://github.com/JuliaData/DataFrames.jl/pull/1810?

davidanthoff:
>  if we merge this do we need #1810?

No, I think with this we probably don't need #1810.

src/other/tables.jl
quinnj: kk, updated the keyword arg and added a comment that `copycols` is ignored for `CopiedColumns`

davidanthoff:
Thanks for sorting this out!

PR #1813, opened 2019-05-17
Make DataFrameRow behave more like NamedTuple

bkamins:
This PR implements methods that make `DataFrameRow` behave like `NamedTuple` in standard operations.
This is breaking for:
* `keys` (as we return now a `Tuple` not a `Vector`),
* `pairs` as we return `Pairs` not a `Generator`
* `map` where we return a `NamedTuple` not a `Vector`
* `broadcastable` where we disallow `DataFrameRow` on RHS (in preparation for a future functionality to match `NamedTuple`)

I will write the tests after we decide:
* if we want it (or part of it)
* in case we want it do we go through a deprecation period or simply change it (these are small changes, the only code that would be broken is code that expects a vector and consequently mutates it, and now it will get a `Tuple` or `NamedTuple`)

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -161,6 +161,7 @@ Base.get(dfr::DataFrameRow, key::Union{Integer, Symbol}, default) =
     haskey(dfr, key) ? dfr[key] : default
 Base.get(f::Callable, dfr::DataFrameRow, key::Union{Integer, Symbol}) =
     haskey(dfr, key) ? getfield(dfr, key) : f()
+Base.broadcastable(::DataFrameRow) = throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
 """
     copy(dfr::DataFrameRow)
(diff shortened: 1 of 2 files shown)


bkamins:
We already define `Base.IteratorEltype(::DataFrameRow) = Base.EltypeUnknown()` but I will fix it to define it on type not on value in this PR.

Also `collect(dfr)` returns a vector of values like for a `NamedTuple` (I have thought about it earlier and tested this now again) - how did you get the `Vector{Pair{Symbol,Any}}` result? (maybe I make some mistake)

nalimilan:
> Also `collect(dfr)` returns a vector of values like for a `NamedTuple` (I have thought about it earlier and tested this now again) - how did you get the `Vector{Pair{Symbol,Any}}` result? (maybe I make some mistake)

Sorry, I meant `collect(pairs(dfr))`.

bkamins:
Yes - there is a trade-off:
* generator approach gives good eltype after collecting
* proposed approach allows for indexing into `pairs(dfr)` (generator only allows for iteration)

Yes - we have discussed it when `Base.IteratorEltype(::DataFrameRow)` was introduced, and the conclusion was that calculation of the exact type would be too expensive.

I am not sure which is better/more useful - if you have a strong preference please indicate it.

When recommending please note that `pairs(dfr)` and `pairs(nt)` has one significant difference. The values returned by `pairs(dfr)` can change their type after this object is created because `dfr` is mutable.

nalimilan:
I'd lean towards keeping the current approach using a generator, as having a concrete element type (notably for homogeneous data frames) sounds more useful than being able to index into the result of `pairs`.

bkamins:
OK - I dropped down to a generator, fixed `IteratorEltype` and added tests.

PR #1814, opened 2019-05-18
Add docstrings for names, size, and order

### LINKED ISSUE #1265: Functions that need docstrings

ajwheeler:
closes  #1265 unless I'm mistaken


src/abstractdataframe/sort.jl
```diff
@@ -20,6 +20,14 @@ struct UserColOrdering{T<:ColumnIndex}
 end
 
 # This is exported, and lets a user define orderings for a particular column
+"""
+User defined order for a particular column.  Used in conjuction with sort!(df::DataFrame, cols).
```
bkamins: Please quote code in the text. Also it would be better if example were runnable. Finally if a function accepts some arguments they should be documented.

ajwheeler: You are referring to the kwargs of `order`, correct?  Are there any besides `rev`?  I was a bit confused by the source.

ajwheeler: Any by "runnable", do you mean that it should avoid undefined variables?

ajwheeler:
Thanks for the notes.  I'm not sure what you're referring to RE `DataFrameRow`.  Could you spell that out for me?  I don't see the connection to any of these methods.

src/abstractdataframe/sort.jl
bkamins: In the body of `ordering(col_ord::UserColOrdering, lt::Function, by::Function, rev::Bool, order::Ordering)` you have a list of accepted kwargs.

By "runnable" I mean that when we enable docstring testing during CI (which we plan in the future) the code should run without error (as otherwise CI will fail). In this case it means that the use of undefined variable. The additional benefit of ensuring "runnability" is that the user can copy-paste the code and test what it does.

bkamins:
`DataFrameRow` is a type defined by DataFrames.jl and it is exported. It supports `size` and `names` functions so it would be good to add appropriate docstrings in src/dataframerow/dataframerow.jl. Thank you!

bkamins:
@ajwheeler - after this is merged I can do what @nalimilan proposes above, but maybe you would be interested in doing this?

ajwheeler:
Sorry, I no longer have time to work on this.  Feel free to discard what
I've done if it's easier.

bkamins:
@nalimilan - you are better with docstrings than me 😄 . How do you advise to proceed with this?

bkamins:
I have opened https://github.com/JuliaData/DataFrames.jl/pull/2077 instead of this PR.

PR #1818, opened 2019-05-24
Extend describe from DataAPI to allow removing StatsBase dependency

quinnj:


src/deprecated.jl
```diff
@@ -1323,24 +1323,6 @@ macro tsv_str(s, flags...)
     inlinetable(s, flags...; separator='\t')
 end
 
-@deprecate showcols(df::AbstractDataFrame, all::Bool=false, values::Bool=true) describe(df, :eltype, :nmissing, :first, :last)
-@deprecate showcols(io::IO, df::AbstractDataFrame, all::Bool=false, values::Bool=true) show(io, describe(df, :eltype, :nmissing, :first, :last), all)
-function StatsBase.describe(df::AbstractDataFrame; stats=nothing)
```
bkamins: @quinnj Thank you for the PR. It will make our life easier. The only thing I would change is leave this deprecation in line 1328 to stick for some longer time as it was introduced very recently, so some old user code might rely on it.

src/DataFrames.jl
```diff
@@ -7,7 +7,8 @@ module DataFrames
 ##############################################################################
 
 using Statistics, Printf, REPL
-using Reexport, StatsBase, SortingAlgorithms, Compat, Unicode, PooledArrays
+using Reexport, SortingAlgorithms, Compat, Unicode, PooledArrays
+using DataAPI
```
nalimilan: Add this to previous line?

src/DataFrames.jl
```diff
@@ -33,7 +34,7 @@ export AbstractDataFrame,
        deletecols!,
        deletecols,
        deleterows!,
-       describe,
+       DataAPI.describe,
```
nalimilan: Is the prefix actually needed?

bkamins: `describe` is not exported by DataAPI.jl so I think it is needed.

nalimilan: OK. It would be possible to do `using DataAPI: describe` above, but that's not clearly better.

quinnj:
Alright, rebased the PR here. Tests looked good locally, though I ran into a test/indexing crash on the Julia 1.2-rc2 (reported https://github.com/JuliaLang/julia/issues/32607)

bkamins:
Thank you - I will have a detailed look at it soon.

Additionally it seems that tests also still try to reference StatsBase.jl (not sure why - probably we use some function from it for testing purposes).

bkamins:
@quinnj in test/utils.jl you can remove `Statistics, StatsBase, Random` from `using` in line 3 (https://github.com/JuliaData/DataFrames.jl/blob/master/test/utils.jl#L3). They are not used (only `Test` and `DataFrames` need to be imported).

bkamins:
Also - if I remember things correctly `DataAPI` does not export `describe` so instead of re-exporting `DataAPI` we should export `describe` from it. Right?

quinnj:
Updated test/utils.jl. DataAPI has a policy to specifically *not* export any functions, relying on extending packages to do so. Here, we're just exporting `describe` (after importing it, so really, we're exporting the DataAPI.describe function), which should ensure no breakage.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -365,20 +365,20 @@ support such objects (and not only vectors), and cannot access missing values.
...
-describe(df, :min, :sum => sum)
+describe([io,] df)
+describe([io,] df, :all)
+describe([io,] df, :min, :max)
+describe([io,] df, :min, :sum => sum)
 ```
 
 """
-StatsBase.describe(df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) =
+DataAPI.describe(io::IO, df::AbstractDataFrame, stats::Union{Symbol, Pair{Symbol}}...) =
```
bkamins: Do we have to quality here and below with `DataAPI.` prefix? I think it can be omitted. Right?

quinnj: We can omit, but I personally like to always explicitly qualify, just to be explicit about where this method is being defined.

src/deprecated.jl
nalimilan: @quinnj Please add back these deprecations. I see no reason to remove these, especially `showcols` which is likely to appear in old blog posts here and there.

bkamins: I have accepted this because the deprecations were "deprecated" long enough (but @nalimilan - I know your policy so if you really think we should revert I am OK with this). Thank you for such a careful consideration of things.

nalimilan: These don't cost us anything so I don't see the point of removing them...

quinnj: I don't think it's a good policy to just keep deprecations around because they're "cheap". That leads to users disregarding deprecations because they might think the functionality will stick around forever anyway; or even worse, they *won't* know what is a "real" deprecation vs. something we're just trying to fade out. These deprecations have been in multiple minor releases at this point, so I see no reason to keep them around.

nalimilan: Deprecations are so verbose that I really doubt users will keep using them instead of the replacement. All our deprecations are "real" and I think we've been quite clear about that. But note that there are books around that use the old APIs and we haven't even tagged a real major release, so keeping deprecations is really the minimum we can do to respect our users.

quinnj: We're still pre-1.0, so each minor release is like a major release.

nalimilan: Technically yes, but they happen much more frequently than real major releases, and they aren't really advertized.

PR #1822, opened 2019-05-25
Remove dump method

nalimilan:
This is an abuse of `dump`, which is supposed to always print the internals.
`summary(df)` can be used instead.


nalimilan:
Yeah, and there's no exact replacement anyway.

PR #1823, opened 2019-05-25
make categorical! accept missing values when converting

bkamins:
A small change how `categorical!` without specified columns works. I recommend to allow automatic conversion of columns with eltype `Union{AbstractString,Missing}` (and not only `AbstractString` as currently).

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1205,8 +1205,8 @@ end
     categorical!(df::DataFrame; compress::Bool=false)
 
 Change columns selected by `cname` or `cnames` in data frame `df`
-to `CategoricalVector`. If no columns are indicated then all columns that have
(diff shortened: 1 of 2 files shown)


src/dataframe/dataframe.jl
```diff
@@ -1206,7 +1206,7 @@ end
 
 Change columns selected by `cname` or `cnames` in data frame `df`
 to `CategoricalVector`. If no columns are indicated then all columns that have
-an `AbstractString` element type type will be converted to categorical.
+an `Union{AbstractString, Missing}` element type type will be converted to categorical.
```
nalimilan: To avoid any ambiguity, I'd say "all columns whose element type is a subtype of `Union{AbstractString, Missing}` will be...".

bkamins:
CI passes (only coveralls show some problems).

PR #1824, opened 2019-05-25
Improve nonunique

bkamins:
A small PR with the following changes in `nonunique`:
* return `BitVector` instead of `Vector{Bool}`
* correctly handle passing `true` as `cols` (it should error)
* make the operation faster for `DataFrame` (as in the old implementation we were performing an unnecessary copy)
* improve code coverage

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -878,7 +878,7 @@ function nonunique(df::AbstractDataFrame)
     return res
 end
 
-nonunique(df::AbstractDataFrame, cols::Union{Integer, Symbol}) =
+nonunique(df::AbstractDataFrame, cols) =
     if cols isa Union{Integer, Symbol}
         if cols isa Bool
             throw(ArgumentError("invalid index: $cols of type Bool"))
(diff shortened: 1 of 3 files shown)


src/dataframe/dataframe.jl
```diff
@@ -1567,3 +1567,5 @@ end
 function permutecols!(df::DataFrame, p::AbstractVector{Symbol})
     permutecols!(df, index(df)[p])
 end
+
+nonunique(df::DataFrame, cols) = nonunique(select(df, cols, copycols=false))
```
nalimilan: Maybe better implement `select` for all existing `AbstractDataFrame` types? Then we wouldn't need this special case.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -871,15 +871,22 @@ function nonunique(df::AbstractDataFrame)
     gslots = row_group_slots(ntuple(i -> df[i], ncol(df)), Val(true))[3]
     # unique rows are the first encountered group representatives,
     # nonunique are everything else
-    res = fill(true, nrow(df))
+    res = trues(nrow(df))
```
nalimilan: Isn't that a bit slower? There may be a speed-memory tradeoff, and in general we tend to favor speed since we assume the data frame stores many columns so it doesn't really matter to allocate another one temporarily.

bkamins:
TODO note:
also `completecases` needs fixing as now it does an unnecessary copy internally.

This will be added after `Not` is added and `select` is implemented for all `AbstractDataFrame` types.

bkamins:
I will close this once I move all functionality to https://github.com/JuliaData/DataFrames.jl/pull/1847

bkamins:
This PR is now redundant. It is covered by https://github.com/JuliaData/DataFrames.jl/pull/1847.

PR #1826, opened 2019-05-25
Fix map(f, ::GroupedDataFrame) when there are no groups

nalimilan:
Also fix a bug when the user-provided function returns an empty `DataFrame` with no columns.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -291,6 +291,10 @@ function Base.map(f::Any, gd::GroupedDataFrame)
     if length(gd) > 0
         idx, valscat = _combine(f, gd)
         parent = hcat!(gd.parent[idx, gd.cols], valscat, makeunique=true)
+        if length(idx) == 0
+            return GroupedDataFrame(parent, collect(1:length(gd.cols)), idx,
+                                    Int[], Int[], Int[])
+        end
         starts = Vector{Int}(undef, length(gd))
         ends = Vector{Int}(undef, length(gd))
         starts[1] = 1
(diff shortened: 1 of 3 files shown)


bkamins:
Can you please also test:
```
map(d -> DataFrame(X=Int[]), gd)
```

nalimilan:
Test added.

bkamins:
Let us discuss the second (added) test. Do you think it is OK that the result does not contain an empty column `X` also? Consider e.g. a scenario:

```
map(d -> DataFrame(X=filter(ismissing, d.X)), gd)
```

i.e. for each group we select only rows that have `missing` in column `X` (assuming this column exists).

Now the list columns of the resulting object will depend on the fact if there ware any missing values in column `X`.

A similar - probably possible enough - case is when in `map` we select all rows of a given group only if they pass some test (e.g. the average of some column is above some value - e.g. you have a yearly data on temperatures and you want to keep only months for which the average temperature was above some value; if the temperature is 0 degrees Celsius you will get some rows, but if it is 100 then you get nothing - and again the list of columns changes). We will get a similar situation.

Sorry for not being specific earlier - but I felt that something was lurking there - but did not realize the problem before I saw the actual test code.

nalimilan:
Well, the resulting `GroupedDataFrame` somewhat contains the empty column `X`: at least calling `parent` on it allows accessing it. But it can't be observed in the `GroupedDataFrame` since there are no groups, so you can't extract a `SubDataFrame` from it. This will change if we make `GroupedDataFrame` an `AbstractDataFrame`.

bkamins:
OK - so we are "safe" then, but maybe then it would be good to add the tests of what `parent` contains so that we (or someone else) is reminded of this corner case.

nalimilan:
Sure: https://github.com/JuliaData/DataFrames.jl/pull/1828.

PR #1830, opened 2019-05-26
small documentation fixes of split-apply-combine

### LINKED ISSUE #1817: does groupby preserve order in SubDataFrames?
Are the rows in a `SubDataFrame` returned by iterating the result of a `groupby` guaranteed to be in the same order as in the original dataframe? Eg
```julia
julia> df = DataFrame(A = repeat(1:4, inner = 2), B = repeat(1:2, outer = 4))
8×2 DataFrame
│ Row │ A     │ B     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │
│ 2   │ 1     │ 2     │
│ 3   │ 2     │ 1     │
│ 4   │ 2     │ 2     │
│ 5   │ 3     │ 1     │
│ 6   │ 3     │ 2     │
│ 7   │ 4     │ 1     │
│ 8   │ 4     │ 2     │

julia> [issorted(d.B) for d in groupby(df, [:A])]
4-element Array{Bool,1}:
 1
 1
 1
 1
```
My reading of the code suggests that this is the case, but I am wondering if this is something one can rely on, or an accident of the implementation. Since docs don't specify this, there is no particular reason the rows could not be exchanged.

bkamins:
closes #1817 and some additional small documentation fixes to make it more consistent

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -49,7 +49,7 @@ A `GroupedDataFrame` : a grouped view into `df`
 
 An iterator over a `GroupedDataFrame` returns a `SubDataFrame` view
 for each grouping into `df`.
-Each grouping preserves the order of rows in `df`.
+Within each group, the order of rows in `df` is preserved.
 
 A `GroupedDataFrame` also supports
 indexing by groups, `map` (which applies a function to each group)
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -36,19 +36,22 @@ groupby(cols; sort = false, skipmissing = false)
 
 ### Arguments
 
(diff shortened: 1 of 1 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -36,19 +36,22 @@ groupby(cols; sort = false, skipmissing = false)
...
 
-A `GroupedDataFrame` : a grouped view into `d`
+A `GroupedDataFrame` : a grouped view into `df`
 
 ### Details
 
 An iterator over a `GroupedDataFrame` returns a `SubDataFrame` view
-for each grouping into `d`. A `GroupedDataFrame` also supports
+for each grouping into `df`.
+Each grouping preserves the order of rows in `df`.
```
```suggestion
Within each group, the order of rows in `df` is preserved.
```

PR #1832, opened 2019-05-27
support julia 0.7

bjarthur:
i know not everyone thinks that julia 0.7 is still useful, but i do, and since the effort involved is so trivial, i'm happy to take in on myself.  see https://github.com/JuliaData/DataFrames.jl/pull/1784#issuecomment-493694454.

worth noting that all i had to do was `import select` if 0.7, and `export` if 1.0

### DIFF
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -52,8 +52,6 @@ export AbstractDataFrame,
        order,
        rename!,
        rename,
-       select,
-       select!,
        showcols,
        stack,
        stackdf,
@@ -63,8 +61,10 @@ export AbstractDataFrame,
 
 if VERSION >= v"1.1.0-DEV.792"
     import Base.eachcol, Base.eachrow
+    export select!, select
 else
     export eachcol, eachrow
+    import Base.select!, Base.select
 end
 
 ##############################################################################
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -898,7 +898,11 @@ end
     @test gd[1] isa SubDataFrame
     @test gd[1] == view(df, [1, 5], :)
     @test_throws BoundsError gd[5]
-    @test_throws ArgumentError gd[true]
+    if VERSION < v"1.0.0-"
+        @test gd[true] == gd[1]
+    else
+        @test_throws ArgumentError gd[true]
+    end
     @test_throws ArgumentError gd[[1, 2, 1]]
     @test_throws MethodError gd["a"]
     gd2 = gd[[false, true, false, false]]
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -898,10 +898,10 @@ end
     @test gd[1] isa SubDataFrame
     @test gd[1] == view(df, [1, 5], :)
     @test_throws BoundsError gd[5]
-    if VERSION < v"1.0.0-"
-        @test gd[true] == gd[1]
-    else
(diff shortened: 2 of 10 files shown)


bjarthur:
tests should pass once https://github.com/queryverse/IteratorInterfaceExtensions.jl/pull/3, https://github.com/queryverse/TableTraits.jl/pull/10, and https://github.com/JuliaData/Tables.jl/pull/98 get merged and tagged.

bkamins:
I think tests of your changes will fail on 0.7 due to dependencies restrictions (so relevant packages downstream need to get resolved first - that is why (especially because they are simple) I have recommended to do these changes in `dev` locally rather than merging them into remote), but I appreciate your efforts.

davidanthoff:
I would have thought that things should work even before the IteratorInterfaceExtensions and TableTraits bounds are adjusted: the previous versions of both packages supported julia 0.7 (I think), and then 1.0 releases for both versions didn't introduce any changes. So I would have thought that the package resolver would just pick the previous version of the two packages and everything would work?

bjarthur:
@davidanthoff the latest version of DataFrames (master and this PR) requires Tables >=0.2.3.  the last version of Tables  that supported julia 0.7 was 0.1.18, though i believe they actually all worked.

relatedly, it seems risky to me to use such an inequality when specifying compatibility.  should any subsequent version of a dependency not be compatible, and DataFrames' Project.toml not be simultaneously changed to reflect that, it would break.  wouldn't it be safer to assume that subsequent version don't work unless they are known to?

src/DataFrames.jl
```diff
@@ -63,8 +63,10 @@ export AbstractDataFrame,
 
 if VERSION >= v"1.1.0-DEV.792"
     import Base.eachcol, Base.eachrow
+    export select!, select
```
nalimilan: This is redundant with the export above.

test/grouping.jl
```diff
@@ -898,7 +898,11 @@ end
     @test gd[1] isa SubDataFrame
     @test gd[1] == view(df, [1, 5], :)
     @test_throws BoundsError gd[5]
-    @test_throws ArgumentError gd[true]
+    if VERSION < v"1.0.0-"
```
nalimilan: Better turn around the test so that the first test behavior is the one for Julia 1.0 and later (as elsewhere).

nalimilan:
> relatedly, it seems risky to me to use such an inequality when specifying compatibility. should any subsequent version of a dependency not be compatible, and DataFrames' Project.toml not be simultaneously changed to reflect that, it would break. wouldn't it be safer to assume that subsequent version don't work unless they are known to?

In theory yes, but in practice it's a nightmare to update all dependencies when bumping the minor version of a package (see e.g. what happened for StatsBase and Distributions). So for packages that haven't reached 1.0 yet it's more practical to only specify lower bounds.

src/DataFrames.jl
bjarthur: doh, fixed, thanks.

quinnj:
I still don't think this is worth doing. It's fine that this is a PR and if @bjarthur has all these packages as 0.7 compat for himself locally, that should allow him to do any necessary testing/porting to 1.0, but unless there are others that also happen to be in this unlikely state, I don't think it's worth the release tag/bounds changing churn.

bkamins:
I would like to close it. Any objections?

PR #1834, opened 2019-05-30
fix push! error message

bkamins:
This fixes a problem with printing `push!` error message when `t` is `nothing`.


bkamins:
>  it's probably a good idea to print quotes around strings anyway.

Do you say that the additional benefit of `repr` is that we enclose strings in quotes and this is good (if yes I agree 😄). Please confirm before I merge.

nalimilan:
Yes, that's what I meant.

PR #1836, opened 2019-05-31
Deprecate get for AbstractDataFrame

bkamins:
I propose to deprecate `get` method for `AbstractDataFrame`. I think it is not used and in general it is confusing to have it since we moved from `AbstractDict` interpretation.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -217,6 +217,8 @@ function Base.size(df::AbstractDataFrame, i::Integer)
     end
 end
 
+Base.isempty(df::AbstractDataFrame) = size(df, 1) == 0 || size(df, 2) == 0
+
 Base.lastindex(df::AbstractDataFrame) = ncol(df)
 Base.lastindex(df::AbstractDataFrame, i::Integer) = last(axes(df, i))
 Base.axes(df::AbstractDataFrame, i::Integer) = Base.OneTo(size(df, i))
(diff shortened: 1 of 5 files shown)


nalimilan:
Shouldn't we also deprecate `haskey`? It's useful, but it's quite inconsistent since we don't define `keys`? Or maybe we should define `keys` too. That could make sense if we decide to keep allowing `df[col]`, but it could also be confusing since a data frame is two-dimensional in most situations.

bkamins:
This was exactly my thinking that `haskey` is useful (as opposed to `get`).

We could rename `haskey` to `hascol` to make it more explicit and restrict its signature to allow only `Symbol` or `Integer`. Then we could decide if we go back to https://github.com/JuliaData/DataFrames.jl/pull/1712 and decide if we want `columnindex` (or `colindex` if we prefer) as this would be a complementary function to `hascol`.

nalimilan:
Could we just deprecate `haskey(n, df)` in favor of `n in names(df)` for now? Adding `hascol` is an interesting possibility, but that would only make sense if we also add a function to access columns like `getcol` or `column`. Better not rush that decision.

bkamins:
I could do it, but the problem is that `n in names(df)` will be *much* slower for two reasons:
* `names` makes a copy
* it will be a linear search, while `haskey` uses a dict inbuilt into `Index`

In short - if we do it, then it should be only a temporary decision with the plan to provide a first class function that checks if some column exists in a data frame.

How do you see this?

nalimilan:
Yes, it's much slower, but do we expect it to be used in performance-critical paths?

bkamins:
Yes - I just wanted to make sure that we note down the consequences.

bkamins:
I have deprecated `haskey`.

Issue #1610, opened 2018-11-29
DataFrames should be indexable by `CartesianIndex{2}`


ExpandingMan:
It should be possible to do, for example
```julia
for idx ∈ CartesianIndices(size(df))
    println(df[idx])
end
```

nalimilan:
I guess that would be consistent, but in what situations would it be useful?

ExpandingMan:
It's just a standard way to iterate over things.  I for one use it more and more.  It's already column major so it's a good way to iterate.  If you ever are iterating over a matrix and a dataframe at the same time, it might be useful.

I'd be happy to make the PR, when I get to it.

nalimilan:
Iterating over a data frame and a matrix at the same time is unlikely to be a good idea given how slow it's going to be compared to iterating over columns...

ExpandingMan:
Well, they are both column major so if you have to iterate over the dataframe anyway I don't really see what harm it would do (other than typing issues of course).

Regardless, is there some reason you are opposed to making dataframes indexable by `CartesianIndex`?

nalimilan:
Well the type-instability issue is a big problem in terms of performance. I'm not really opposed to supporting cartesian indexing, I'm just not sure it's really useful.

bkamins:
I think it is not a problem to have it for `AbstractDataFrame` and possibly for `DataFrameRow` also if we want a complete coverage.

bkamins:
We now allow `CartesianIndex{2}` indexing so I am closing this.

bkamins:
(we do not provide `CartesianIndices` though. If you feel it should be added please reopen and it can be discussed.

PR #1840, opened 2019-06-08
Add broadcasting of AbstractDataFrame

### LINKED ISSUE #1806: Allow data frame and DataFrameRow to take part in broadcasting
Currently data frame is not allowed in broadcasting and `DataFrameRow` is collected to a vector.

Following the discussion in #1804 we should probably implement custom broadcasting rules for them that would preserve column names if possible.

It is not entirely clear for me what should be the output but I guess this is an initially reasonable approach:
* `f.(data_frame)` returns a `DataFrame` transforming each element using `f` but retaining column names; the question is what should happen if we combine data frames in broadcasting with other data frames or data frame rows or three or higher dimensional objects;
* I am not sure what `f.(::DataFrameRow)` should produce - probably a `NamedTuple`, but the problem is that `NamedTuples` are not allowed in broadcasting later

Please comment if you have any opinion on this.

bkamins:
This resolves #1806 and #1610.

@nalimilan - in particular have a look at how `CategoricalVector` is transformed in broadcasting using current `similar` rules. This is visible in the test starting with `df3 = coalesce.(df, nothing)` line

@mbauman - could you have a look at this PR if you have time to spare (and also possibly at the whole other/broadcasting.jl file which now lays out full rules of broadcasting of data frames)


bkamins:
@nalimilan - the tests pass on Julia 1.1, but fail on Julia 1.0 due to a check against `CategoricalArray` type - do you have an idea what is going on? (if not I can install Julia 1.0 locally and check, but maybe you know the reason). My intuition is that this is `similar` related again 😞.

bkamins:
Fortunately all works after all. It was another change in broadcasting between 1.0 and 1.1 that caused the problem but it is unrelated to this PR.

bkamins:
> Maybe we should discuss separately whether we should keep using promote_type instead of promote_typejoin

You have the most experience here (but as you say we can leave it for later). I think the difference is not that big in practice as this case should not happen often (if at all). If we change it then probably we should make sure that we make it consistent everywhere (in particular in original `copyto_widen!`)

src/other/broadcasting.jl
```diff
@@ -1,3 +1,63 @@
...
+    DataFrameStyle()
+
+Base.Broadcast.BroadcastStyle(::DataFrameStyle, ::Base.Broadcast.BroadcastStyle) = DataFrameStyle()
+Base.Broadcast.BroadcastStyle(::Base.Broadcast.BroadcastStyle, ::DataFrameStyle) = DataFrameStyle()
+Base.Broadcast.BroadcastStyle(::DataFrameStyle, ::DataFrameStyle) = DataFrameStyle()
+
+function copyto_widen!(res::AbstractVector{T},
+                       bc::Base.Broadcast.Broadcasted{DataFrameStyle},
+                       pos, col) where T
+    for i in pos:length(axes(bc)[1])
```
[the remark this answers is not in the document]
bkamins: Unfortunately this does not work as this method is not defined for `bc`. I have discussed with @mbauman that this should be added for, but for now we have to use indexing.

src/other/broadcasting.jl
```diff
@@ -1,3 +1,63 @@
...
+        val = bc[CartesianIndex(i, col)]
```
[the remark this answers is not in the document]
bkamins: I do not think it is possible easily, unless @mbauman has an idea how to do it safely (essentially we would have to break down a 2-dimenstional `bc` into a collection of 1-dimensional columnwise `bc`s that would select appropriate columns from data frames taking part in the operation).

docs/src/lib/indexing.md
```diff
@@ -82,6 +83,11 @@ Under construction
 
 ## Broadcasting
 
+The following broadcasting rules apply to `AbstractDataFrame` objects:
+* `AbstractDataFrame` behaves in broadcasting like a matrix.
```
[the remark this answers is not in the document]

src/other/broadcasting.jl
```diff
@@ -1,3 +1,63 @@
...
+        end
+    end
+    return res
+end
+
+function Base.copy(bc::Base.Broadcast.Broadcasted{DataFrameStyle})
+    bcf = Base.Broadcast.flatten(bc)
+    colnames = unique([_names(df) for df in bcf.args if df isa AbstractDataFrame])
+    if length(colnames) != 1
+        throw(ArgumentError("Column names in broadcasted data frames must match"))
```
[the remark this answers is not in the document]
bkamins: fixed

test/broadcasting.jl
```diff
@@ -1,11 +1,84 @@
...
-using Test, DataFrames
+using Test, DataFrames, PooledArrays
 
 const ≅ = isequal
 
 refdf = DataFrame(reshape(1.5:15.5, (3,5)))
 
+@testset "broadcasting of AbstractDataFrame objects" begin
+    for df in (copy(refdf), view(copy(refdf), :, :))
+        @test identity.(df) == refdf
```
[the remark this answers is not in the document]
bkamins: added

test/broadcasting.jl
```diff
@@ -1,11 +1,84 @@
...
+        @test (x->x).(df) == refdf
+        @test (df .+ df) ./ 2 == refdf
```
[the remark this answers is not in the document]
bkamins: added

test/broadcasting.jl
```diff
@@ -1,11 +1,84 @@
...
+@testset "broadcasting of AbstractDataFrame objects corner cases" begin
+    df = DataFrame(c11 = categorical(["a", "b"]), c12 = categorical([missing, "b"]), c13 = categorical(["a", missing]),
+                   c21 = categorical([1, 2]), c22 = categorical([missing, 2]), c23 = categorical([1, missing]),
+                   p11 = PooledArray(["a", "b"]), p12 = PooledArray([missing, "b"]), p13 = PooledArray(["a", missing]),
+                   p21 = PooledArray([1, 2]), p22 = PooledArray([missing, 2]), p23 = PooledArray([1, missing]),
+                   b1 = [true, false], b2 = [missing, false], b3 = [true, missing],
+                   f1 = [1.0, 2.0], f2 = [missing, 2.0], f3 = [1.0, missing],
+                   s1 = ["a", "b"], s2 = [missing, "b"], s3 = ["a", missing])
+
+    df2 = DataFrame(c11 = categorical(["a", "b"]), c12 = [nothing, "b"], c13 = ["a", nothing],
```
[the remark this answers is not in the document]
bkamins: It is needed because I compare `eltypes` and `typeof` of `df2` and `df3` below.

src/other/broadcasting.jl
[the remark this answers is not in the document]
nalimilan: Right. Maybe we can special-case data frames? Anyway that's the only type for which it matters.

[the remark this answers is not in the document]
bkamins: What I imagine you would have to do is the following:
1. flatten `bc`
2. manually check dimensionality of all `args` and which of them is `AbstractDataFrame` (I already do this for)
3. pass to flattened function arguments by "manual" broadcast of all `args`

This is doable but complex (effectively you have to rewrite broadcasting machinery), unless @mbauman could propose something better.

If what I write above is the best solution I would:
1. leave the implementation as is
2. open an issue/PR for performance improvements as to make the above fully efficient is not an easy task (then we can safely merge the faster approach when we are sure it is fully safe).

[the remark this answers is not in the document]
mbauman: Well, you could just re-use broadcast here.  Just write a `preprocess`-like recursive function that hoists the columns out.  In other words (pseudo-pseudo code, untested, etc):

```julia
import Broadcast: Broadcasted, extrude, broadcast_unalias
@inline preprocess(dest, bc::Broadcasted{Style}, col) where {Style} = Broadcasted{Style}(bc.f, map(x->preprocess(dest, x, col), bc.args), bc.axes)
preprocess(dest, x, col) = extrude(broadcast_unalias(dest, x))
preprocess(dest, x::AbstractDataFrame, col) = extrude(broadcast_unalias(dest, x[:,col]))
```

Then:
```julia
# ...
    for j in axes(bc)[2]
        bccol = preprocess(dest, bc, j)
        inner(dest, bccol, j)
    end
# ...
function inner(dest, bc, j)
    for i in axes(bc)[1]
        dest[i, j] = bc[i, j]
    end
end
```

Another interesting/crazy thought would be to manually unroll a number (maybe 16?) of loops — does constant propagation save you if you do

```julia
m, n = size(bc)
n == 0 && return dest
for i=1:m; dest[i, 1] = bc[i, 1]; end
n == 1 && return dest
for i=1:m; dest[i, 2] = bc[i, 2]; end
n == 2 && return dest
# and so on...
```

Of course, that's not OffsetArray safe, but do you support offset data frames?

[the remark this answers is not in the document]
bkamins: Thank you. Just to confirm I understand the idea correctly - for each column you replace an `AbstractDataFrame` with a single column extracted from this data frame and let broadcast mechanics do the rest - right?

bkamins:
@nalimilan In the process of analysis of @mbauman suggestions I have identified an aliasing bug we had in broadcasting assignment.
I add fixes to this PR as they are required here.
You have test cases [here](https://github.com/JuliaData/DataFrames.jl/pull/1840/files#diff-a14f3df2199d7ba7ff58ee5070b5364aR619).

src/other/broadcasting.jl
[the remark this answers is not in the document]
nalimilan: I don't really understand all the mechanics, but does `bc[i, j]` in the inner loop imply indexing a data frame under the hood, or just the column vector? Is there any way we could add an optimized path which calls `copyto!` (which is much faster for `CategoricalArray` and `PooledArray`) when the source is a column vector in a data frame?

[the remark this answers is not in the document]
bkamins: I have just pushed out a working implementation of what @mbauman proposal. It is still WIP as I have not tested it end-to-end. Hopefully this will clarify.

However, I have one general note about allowing data frames on right hand side of broadcasting. It is going to be very slow for data frames with a large number of columns. The reason is that we have to unalias all pairs of columns in source and destination. So for `10_000` columns you have to do `100_000_000` checks, which is an overkill. Maybe it would be better to just copy a data frame if it is in source and has a number of columns over some limit?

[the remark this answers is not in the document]
bkamins: @nalimilan I managed to implement some improvements and now the broadcasting speed is not that bad even for large data frames. But for very large data frames it is better to convert a `DataFrame` into e.g. a `Matrix` as it will be much faster because we are sure we do not have to do do alias analysis.

bkamins:
@nalimilan  - this PR should be good for another look at. I have implemented fast handling of columns and alias detection in all cases (I hope). Additional test case suggestions would be appreciated as the code, although not very long, was non-trivial for me.

bkamins:
I have rebased this PR against master and added even more tests.

src/other/broadcasting.jl
```diff
@@ -33,6 +33,13 @@ function copyto_widen!(res::AbstractVector{T},
     return res
 end
 
+function getcolbc(bcf::Base.Broadcast.Broadcasted{Style}, colind) where {Style}
+    # we assume that bcf is already flattened and unaliased
+    Base.Broadcast.Broadcasted{Style}(bcf.f,
+        map(x->Base.Broadcast.extrude(x isa AbstractDataFrame ? x[colind] : x),
```
nalimilan: Move this to a separate line (with an intermediate variable) for readability?

src/other/broadcasting.jl
```diff
@@ -42,23 +49,23 @@ function Base.copy(bc::Base.Broadcast.Broadcasted{DataFrameStyle})
         throw(ArgumentError("Column names in broadcasted data frames must match. " *
                             "Non matching column names are $msg"))
     end
-    nrows = length(axes(bc)[1])
+    nrows = length(axes(bcf)[1])
```
nalimilan: `bcf` doesn't exist?

src/other/broadcasting.jl
```diff
@@ -128,13 +186,17 @@ function Base.copyto!(df::AbstractDataFrame, bc::Base.Broadcast.Broadcasted{<:Ba
...
 end
 
+Base.Broadcast.broadcast_unalias(dest::DataFrameRow, src) =
+    Base.Broadcast.broadcast_unalias(parent(dest), src)
+
 function Base.copyto!(dfr::DataFrameRow, bc::Base.Broadcast.Broadcasted)
-    for I in eachindex(bc)
-        dfr[I] = bc[I]
+    bc′ = Base.Broadcast.preprocess(dfr, bc)
+    for I in eachindex(bc′)
```
nalimilan: Would `@inbounds` be safe here?

src/other/broadcasting.jl
```diff
@@ -113,9 +120,60 @@ function _copyto_helper!(dfcol::AbstractVector, bc::Base.Broadcast.Broadcasted,
...
+    end
+    # col2 can be checked from col1 point as we are writing broadcasting
+    # results from 1 to ncol
+    wascopied = false
+    for col1 in axes(dest, 2)
+        for col2 in col1:ncol(src)
+            dcol = dest[col1]
+            scol = src[col2]
+            if Base.mightalias(dcol, scol)
+                if src isa SubDataFrame
```
nalimilan: Maybe we could allow `copy(::SubDataFrame, copycols=false)`, and have it return a `DataFrame` with `view` columns? Then this special-case wouldn't be needed, right?

src/other/broadcasting.jl
bkamins: done

src/other/broadcasting.jl
bkamins: It is defined in the first line of this function. In general we use `bcf` which is a flattened version of `bc` passed later so I changed this for consistency.

The difference between `bc` and `bcf` is that `bcf` is non-nested (it is a function and a list of arguments), while `bc` can be nested.

src/other/broadcasting.jl
bkamins: It should, but the design is so complex, that I would leave it as is for 6 months to make sure it is well tested. The main reason is that we have to do dynamic dispatch here anyway so `@inbounds` will save very little anyway.

src/other/broadcasting.jl
bkamins: We have to special case it. The reason is that unaliasing should not change the type of broadcasted object. In theory someone might write a code that depends on the fact that `SubDataFrame` was passed on RHS (`isa` test for example). This is one of these "delicate things" that broadcasting does under the hood.

In short we if we take `SubDataFrame` we have to return `SubDataFrame` after unaliasing, but making sure that columns in the parent are unaliased.

nalimilan: Ah, too bad. Then maybe just add a comment mentioning this requirement.

src/other/broadcasting.jl
```diff
@@ -35,9 +35,11 @@ end
 
 function getcolbc(bcf::Base.Broadcast.Broadcasted{Style}, colind) where {Style}
     # we assume that bcf is already flattened and unaliased
-    Base.Broadcast.Broadcasted{Style}(bcf.f,
-        map(x->Base.Broadcast.extrude(x isa AbstractDataFrame ? x[colind] : x),
-            bcf.args), bcf.axes)
+    function dfextruder(x)
```
nalimilan: `map(bcf.args) do ...` is quite simpler. :-)

bkamins:
Benchmarks

Simple broadcasting
```
julia> m = rand(100000, 10);

julia> df = DataFrame(m);

julia> @benchmark $m .+ 1
BenchmarkTools.Trial:
  memory estimate:  7.63 MiB
  allocs estimate:  2
  --------------
  minimum time:     2.548 ms (0.00% GC)
  median time:      3.682 ms (0.00% GC)
  mean time:        4.252 ms (20.66% GC)
  maximum time:     65.567 ms (93.74% GC)
  --------------
  samples:          1174
  evals/sample:     1

julia> @benchmark $df .+ 1
BenchmarkTools.Trial:
  memory estimate:  7.64 MiB
  allocs estimate:  145
  --------------
  minimum time:     889.600 μs (0.00% GC)
  median time:      1.112 ms (0.00% GC)
  mean time:        1.391 ms (10.95% GC)
  maximum time:     58.203 ms (94.78% GC)
  --------------
  samples:          3582
  evals/sample:     1

julia> @benchmark $m .+ [1 2 3 4 5 6 7 8 9 10]
BenchmarkTools.Trial:
  memory estimate:  7.63 MiB
  allocs estimate:  3
  --------------
  minimum time:     2.530 ms (0.00% GC)
  median time:      3.665 ms (0.00% GC)
  mean time:        4.212 ms (20.24% GC)
  maximum time:     64.252 ms (93.99% GC)
  --------------
  samples:          1186
  evals/sample:     1

julia> @benchmark $df .+ [1 2 3 4 5 6 7 8 9 10]
BenchmarkTools.Trial:
  memory estimate:  7.64 MiB
  allocs estimate:  156
  --------------
  minimum time:     931.900 μs (0.00% GC)
  median time:      1.147 ms (0.00% GC)
  mean time:        1.350 ms (9.48% GC)
  maximum time:     60.199 ms (94.59% GC)
  --------------
  samples:          3687
  evals/sample:     1

julia> @benchmark $m .+ (1:100000)
BenchmarkTools.Trial:
  memory estimate:  7.63 MiB
  allocs estimate:  2
  --------------
  minimum time:     2.537 ms (0.00% GC)
  median time:      3.624 ms (0.00% GC)
  mean time:        4.167 ms (20.01% GC)
  maximum time:     72.307 ms (94.54% GC)
  --------------
  samples:          1202
  evals/sample:     1

julia> @benchmark $df .+ (1:100000)
BenchmarkTools.Trial:
  memory estimate:  7.64 MiB
  allocs estimate:  156
  --------------
  minimum time:     1.019 ms (0.00% GC)
  median time:      1.239 ms (0.00% GC)
  mean time:        1.446 ms (8.60% GC)
  maximum time:     58.719 ms (94.23% GC)
  --------------
  samples:          3456
  evals/sample:     1

julia> @benchmark $m .+ $m
BenchmarkTools.Trial:
  memory estimate:  7.63 MiB
  allocs estimate:  2
  --------------
  minimum time:     2.451 ms (0.00% GC)
  median time:      3.571 ms (0.00% GC)
  mean time:        4.144 ms (20.80% GC)
  maximum time:     68.536 ms (94.01% GC)
  --------------
  samples:          1205
  evals/sample:     1

julia> @benchmark $m .+ $df
BenchmarkTools.Trial:
  memory estimate:  7.64 MiB
  allocs estimate:  155
  --------------
  minimum time:     1.187 ms (0.00% GC)
  median time:      1.415 ms (0.00% GC)
  mean time:        1.623 ms (8.59% GC)
  maximum time:     63.595 ms (91.71% GC)
  --------------
  samples:          3070
  evals/sample:     1

julia> @benchmark $df .+ $df
BenchmarkTools.Trial:
  memory estimate:  7.64 MiB
  allocs estimate:  166
  --------------
  minimum time:     944.500 μs (0.00% GC)
  median time:      1.171 ms (0.00% GC)
  mean time:        1.331 ms (9.21% GC)
  maximum time:     63.837 ms (95.01% GC)
  --------------
  samples:          3752
  evals/sample:     1
```

Assignment broadcasting (here the performance of data frames drops):
```
julia> @benchmark $m .+= 1
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     441.401 μs (0.00% GC)
  median time:      500.201 μs (0.00% GC)
  mean time:        562.079 μs (0.00% GC)
  maximum time:     3.071 ms (0.00% GC)
  --------------
  samples:          8841
  evals/sample:     1

julia> @benchmark $df .+= 1
BenchmarkTools.Trial:
  memory estimate:  7.63 MiB
  allocs estimate:  133
  --------------
  minimum time:     1.508 ms (0.00% GC)
  median time:      1.635 ms (0.00% GC)
  mean time:        1.795 ms (7.40% GC)
  maximum time:     51.025 ms (93.99% GC)
  --------------
  samples:          2775
  evals/sample:     1

julia> @benchmark $m .+= $m
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     473.200 μs (0.00% GC)
  median time:      552.801 μs (0.00% GC)
  mean time:        619.016 μs (0.00% GC)
  maximum time:     2.216 ms (0.00% GC)
  --------------
  samples:          8034
  evals/sample:     1

julia> @benchmark $df .+= $df
BenchmarkTools.Trial:
  memory estimate:  15.27 MiB
  allocs estimate:  218
  --------------
  minimum time:     2.442 ms (0.00% GC)
  median time:      2.828 ms (9.58% GC)
  mean time:        2.993 ms (8.48% GC)
  maximum time:     50.711 ms (89.78% GC)
  --------------
  samples:          1670
  evals/sample:     1
```

Now the case where we are hit by anty-aliasing analysis most:
```
julia> @benchmark $m .+= $m
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.100 μs (0.00% GC)
  median time:      7.420 μs (0.00% GC)
  mean time:        7.909 μs (0.00% GC)
  maximum time:     48.540 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     5

julia> @benchmark $df .+= $df
BenchmarkTools.Trial:
  memory estimate:  903.03 KiB
  allocs estimate:  10582
  --------------
  minimum time:     850.199 μs (0.00% GC)
  median time:      948.401 μs (0.00% GC)
  mean time:        1.123 ms (5.74% GC)
  maximum time:     50.658 ms (97.44% GC)
  --------------
  samples:          4439
  evals/sample:     1
```

This one is slow:
```
julia> @benchmark $m .+= $df
BenchmarkTools.Trial:
  memory estimate:  625.00 KiB
  allocs estimate:  40000
  --------------
  minimum time:     1.249 ms (0.00% GC)
  median time:      1.401 ms (0.00% GC)
  mean time:        1.644 ms (1.89% GC)
  maximum time:     45.792 ms (96.29% GC)
  --------------
  samples:          3046
  evals/sample:     1
```
because it does not use custom optimizations (we have a data frame on RHS, but not on LHS).

OTOH this is faster than $df .= df$:
```
julia> @benchmark $df .+= $m
BenchmarkTools.Trial:
  memory estimate:  502.75 KiB
  allocs estimate:  7045
  --------------
  minimum time:     615.500 μs (0.00% GC)
  median time:      689.050 μs (0.00% GC)
  mean time:        818.695 μs (5.18% GC)
  maximum time:     50.480 ms (98.05% GC)
  --------------
  samples:          6094
  evals/sample:     1
```
as aliasing analysis in this case is cheap.

So in summary: if you have a very large number of columns it is better to work with matrices than data frames (which probably could be expected).

src/other/broadcasting.jl
bkamins: fixed

bkamins:
In data frame unaliasing code I can add a line:
```
col1 == col2 && dcol === scol && continue
```
which:
1. greatly improves the performance in the common case when the same data frame is on the left and right hand side of broadcasting
2. further penalizes the case of many columns

The general reason why unaliasing is slow is because in this code we have to do:
```
            dcol = dest[col1]
            scol = src[col2]
```
which is not type stable (so all subsequent checks require dynamic dispatch) and this is performed O(`ncol(df)`^2) times.

nalimilan:
Sounds like a good idea. It shouldn't be very costly compared to the unalias checks we already perform. That will make `df .+= 1` faster, right?

src/other/broadcasting.jl
```diff
@@ -128,36 +128,41 @@ function Base.Broadcast.broadcast_unalias(dest::AbstractDataFrame, src)
     src
 end
 
+function _broadcast_unalias_helper(dest, scol, src, col2, wascopied)
```
nalimilan: Make the signature more specific for clarity?

bkamins:
After discussion with @nalimilan I have added some performance improvements in unaliasing. Here are the updated benchmarks. I report the relevant cases.

Long data frame:

```
julia> m = rand(100_000, 10);

julia> df = DataFrame(m);

julia> @benchmark $m .+ $m
BenchmarkTools.Trial:
  memory estimate:  7.63 MiB
  allocs estimate:  2
  --------------
  minimum time:     2.662 ms (0.00% GC)
  median time:      3.253 ms (0.00% GC)
  mean time:        3.842 ms (20.50% GC)
  maximum time:     7.473 ms (53.80% GC)
  --------------
  samples:          1299
  evals/sample:     1

julia> @benchmark $df .+ $df
BenchmarkTools.Trial:
  memory estimate:  7.64 MiB
  allocs estimate:  166
  --------------
  minimum time:     1.088 ms (0.00% GC)
  median time:      3.016 ms (0.00% GC)
  mean time:        3.029 ms (16.34% GC)
  maximum time:     6.696 ms (37.92% GC)
  --------------
  samples:          1648
  evals/sample:     1

julia> @benchmark $m .+= 1
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     428.500 μs (0.00% GC)
  median time:      441.201 μs (0.00% GC)
  mean time:        484.818 μs (0.00% GC)
  maximum time:     1.839 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark $df .+= 1
BenchmarkTools.Trial:
  memory estimate:  7.64 MiB
  allocs estimate:  143
  --------------
  minimum time:     1.414 ms (0.00% GC)
  median time:      2.009 ms (0.00% GC)
  mean time:        2.357 ms (13.52% GC)
  maximum time:     5.696 ms (38.79% GC)
  --------------
  samples:          2120
  evals/sample:     1
```

Wide data frame:
```
julia> m = rand(10, 1000);

julia> df = DataFrame(m);

julia> @benchmark $m .+ $m
BenchmarkTools.Trial:
  memory estimate:  78.20 KiB
  allocs estimate:  2
  --------------
  minimum time:     7.125 μs (0.00% GC)
  median time:      8.025 μs (0.00% GC)
  mean time:        10.341 μs (13.86% GC)
  maximum time:     617.800 μs (98.54% GC)
  --------------
  samples:          10000
  evals/sample:     4

julia> @benchmark $df .+ $df
BenchmarkTools.Trial:
  memory estimate:  556.36 KiB
  allocs estimate:  10550
  --------------
  minimum time:     917.701 μs (0.00% GC)
  median time:      1.003 ms (0.00% GC)
  mean time:        1.096 ms (4.56% GC)
  maximum time:     5.177 ms (67.39% GC)
  --------------
  samples:          4555
  evals/sample:     1

julia> @benchmark $m .+= 1
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     4.357 μs (0.00% GC)
  median time:      4.714 μs (0.00% GC)
  mean time:        4.862 μs (0.00% GC)
  maximum time:     40.800 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     7

julia> @benchmark $df .+= 1
BenchmarkTools.Trial:
  memory estimate:  533.95 KiB
  allocs estimate:  8044
  --------------
  minimum time:     550.200 μs (0.00% GC)
  median time:      601.600 μs (0.00% GC)
  mean time:        675.285 μs (6.58% GC)
  maximum time:     4.716 ms (84.92% GC)
  --------------
  samples:          7382
  evals/sample:     1

julia> @benchmark $m .+= $m
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.080 μs (0.00% GC)
  median time:      6.560 μs (0.00% GC)
  mean time:        6.964 μs (0.00% GC)
  maximum time:     54.600 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     5

julia> @benchmark $df .+= $df
BenchmarkTools.Trial:
  memory estimate:  965.53 KiB
  allocs estimate:  12582
  --------------
  minimum time:     815.999 μs (0.00% GC)
  median time:      839.800 μs (0.00% GC)
  mean time:        981.483 μs (6.75% GC)
  maximum time:     3.718 ms (72.33% GC)
  --------------
  samples:          5093
  evals/sample:     1
```

We can see that in the case of wide data frame (1000 columns) unaliasing costs ~25% of total time. The biggest penalty for this range is the fact that data frame is not type stable (when we process a small number of long columns we do not have this problem as most work is done behind a barrier function).

src/other/broadcasting.jl
bkamins: ok. Regarding the `col1 == col2 && dcol === scol && continue` check it did not help that much, so I left it out.

bkamins:
@nalimilan Thank you for working on it.

bkamins:
and @mbauman for an excellent suggestion of the column extraction approach.

bkamins:
Design principles of data frame broadcasting:
* we add a possibility to index an `AbstractDataFrame` using `CartesianIndex`
* `AbstractDataFrame` is directly broadcastable (it is passed as is)
* If `AbstractDataFrame` is used anywhere in broadcasting it enforces that the broadcast has `DataFrameStyle` (this works OK with Base, but will lead to errors if other types that define custom broadcasting rules are mixed with `AbstractDataFrame`, but in such cases it is not clear what should be produced so it is OK to throw an error then)
* when broadcasted over internally each column of an `AbstractDataFrame` is processed separately (both on LHS and RHS); this is achieved by dynamically generating a new broadcasting object for each processed column that contains only this column instead of the original data frame; this happens to work because, as @mbauman observed, the standard broadcasting mechanics will broadcast this vector just as expected (the important fact that allows this is that data frame is exactly two dimensional and vectors are columnwise - exactly as in data frames); this approach allows us to process broadcasting fast when the data frame has many rows; it has a performance overhead though if we have many columns to process (the processing complexity is O(`ncol(df)`)); additionally a minor thing that is done to simplify and improve the performance of this process is that the original broadcasted object is flattened so that we are sure that we are working on non-nested object
* when data frame is used in non-assignment broadcasting we do not have to unalias columns of a data frame; however if data frame is on left or right hand side of assignment broadcasting we have to perform unaliasing; typically this is O(`ncol(df)`)) operation; the only special case is `AbstractDataFrame` present both on LHS and RHS in which case the operation has O(`ncol(df)`^2)) cost. This is the second reason why performance of broadcasting of `AbstractDataFrames` when it contains many columns might get expensive; practical tests show that for up to `1000` columns the timings are in reasonable range; if you have tens of columns and thousands of rows the performance of broadcasting of `AbstractDataFrame` should be good

bkamins:
@nalimilan I have done a final round of code check after your approval, which lead to some small changes (unfortunately this PR is hard for me to get it right in one run). They are collected in the last two commits (the first adds tests, the second changed documentation and implementation).

Changes:
* I have made a documentation more precise regarding what kinds of broadcasting operations are allowed (but your review here would be most welcome as always 😄)
* now we throw an explicit dimension mismatch error in case we try to broadcast a data frame into higher dimension than 2 (earlier we also errored, but the error was cryptic); however, we allow to do broadcast assignment into higher dimensional objects (then data frame is treated as normal 2-dimensional array-like object)
* there was a tricky case when an additional anti-aliasing method was needed (initially I thought this case is not needed, but I came up with a scenario when it is required) - namely we also have to unalias when `src` is a `DataFrame` and `dest` is not a data frame
* I have added tests for: broadcast assignment of a data frame into other type of object (including the case that required anti-aliasing) and tests of broadcasting of data frame in combination with 3-dimensional object

docs/src/lib/indexing.md
```diff
@@ -83,6 +84,15 @@ Under construction
 
 ## Broadcasting
 
+The following broadcasting rules apply to `AbstractDataFrame` objects:
+* `AbstractDataFrame` behaves in broadcasting like a two-dimensional collection compatible with matrices.
+* If an `AbstractDataFrame` takes part in broadcasting then a `DataFrame` is always produced as a result.
+  In this case the requested broadcasting operation must produce exactly two dimensional object.
```
nalimilan: "a two-dimensional object" or "an object with exactly two dimensions"

docs/src/lib/indexing.md
```diff
@@ -83,6 +84,15 @@ Under construction
...
+  An exception is when `AbstractDataFrame` is used only as a source of broadcast assignment into an object
```
```suggestion
  An exception is when an `AbstractDataFrame` is used only as a source of broadcast assignment into an object
```

docs/src/lib/indexing.md
```diff
@@ -83,6 +84,15 @@ Under construction
...
+  of dimensionality higher than two. In such a case this is allowed and a data frame is treated as two-dimensional
```
nalimilan: The last sentence seems redundant with what has been said above.

src/other/broadcasting.jl
```diff
@@ -128,6 +132,29 @@ function Base.Broadcast.broadcast_unalias(dest::AbstractDataFrame, src)
     src
 end
 
+function Base.Broadcast.broadcast_unalias(dest, src::AbstractDataFrame)
+    wascopied = false
+    for (i, col) in enumerate(eachcol(src))
+        if Base.mightalias(dest, col)
+            if src isa SubDataFrame
+                if !wascopied
+                    src =SubDataFrame(copy(parent(src), copycols=false),
```
nalimilan: Missing space after `=`.

test/broadcasting.jl
```diff
@@ -533,4 +659,276 @@ end
...
+end
+
+@testset "broadcasting with 3-dimensional object" begin
+    y = zeros(4,3,2)
+    df = DataFrame(ones(4,3))
+    @test_throws DimensionMismatch df .+ y
+    @test_throws DimensionMismatch y .+ df
+    @test_throws DimensionMismatch df .+= y
+    y .+= df
+    @test all(y .== 1)
```
```suggestion
    @test y == ones(4,3,2)
```

docs/src/lib/indexing.md

docs/src/lib/indexing.md
bkamins: ok. removed

src/other/broadcasting.jl
bkamins: fixed

PR #1843, opened 2019-06-10
Deprecate empty!

### LINKED ISSUE #1841: empty! should preserve columns
Currently `empty!` removes all columns, making its argument a `0×0 DataFrame`. This doesn't seem consistent with the view of data frames as collections of rows (https://github.com/JuliaData/DataFrames.jl/issues/1514): according to that definition, `empty!` should just resize all column vectors to have length 0.

With this change, `push!` and `append!` would require columns to match, contrary to what happens for a `0×0 DataFrame`. That can be both a drawback (inconvenient) and an advantage (safety). Another point is that I'm not sure it's very useful to be able to make a `DataFrame` be `0×0`, since one can as well use `DataFrame()`.

We could also just remove that function for now, as it doesn't sound terrible useful.

nalimilan:
So that we can change it to be equivalent to `empty!.(eachcol(df))`.

Fix #1841.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -694,11 +694,9 @@ Base.setindex!(df::DataFrame, v, ::Colon, col_inds) =
 
 ##############################################################################
 ##
-## Mutating AbstractDict methods
+## Mutating methods
 ##
 ##############################################################################
-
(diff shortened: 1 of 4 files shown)


nalimilan:
Yeah, the deprecation will have to be updating before tagging a release, so that users don't have to migrate twice.

PR #1847, opened 2019-06-10
Add not to indexing

bkamins:
This is not finished yet and I will add a list of changes it does.
In general I add `Not` for indexing + remove `deletecols!` as it is obsolete with `Not` + clean up several related methods.


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -587,13 +588,12 @@ function completecases(df::AbstractDataFrame, col::Union{Integer, Symbol})
     res
 end
 
-completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex}) =
-    completecases(df[cols])
+completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex, Not}) =
+    completecases(select(df, cols, copycols=false))
```
nalimilan: Shouldn't  `df[cols]` support the same features as `select`?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -654,8 +654,7 @@ julia> dropmissing(df, [:x, :y])
 ```
 
 """
-function dropmissing(df::AbstractDataFrame,
-                     cols::Union{Integer, Symbol, AbstractVector, Regex}=1:size(df, 2);
+function dropmissing(df::AbstractDataFrame, cols=1:size(df, 2);
```
nalimilan: Why not keep the types? That makes things clearer both for developers and for users (via early errors).

src/dataframe/dataframe.jl
```diff
@@ -1024,15 +930,23 @@ select!(df::DataFrame, c::Any) = select!(df, index(df)[c])
 
 """
     select(df::DataFrame, inds, copycols::Bool=true)
+    select(dfv::SubDataFrame, inds, copycols::Bool=true)
+    select(dfr::DataFrameRow, inds)
 
-Create a new `DataFrame` that contains columns from `df` specified by `inds` and return it.
```
nalimilan: Better keep a general description of what the function does, disregarding the particular `AbstractDataFrame` type. The last sentence about `inds` can probably go here too.

src/dataframe/dataframe.jl
```diff
@@ -1024,15 +930,23 @@ select!(df::DataFrame, c::Any) = select!(df, index(df)[c])
...
-
-Argument `inds` can be any index that is allowed for column indexing of a `DataFrame`.
-
+If `df` is a `DataFrame` create a new `DataFrame` that contains columns from `df`
```
```suggestion
If `df` is a `DataFrame` return a new `DataFrame` that contains columns from `df`
```
(and then remove "and return it")

src/dataframe/dataframe.jl
```diff
@@ -1024,15 +930,23 @@ select!(df::DataFrame, c::Any) = select!(df, index(df)[c])
...
+specified by `inds` and return it.
 If `copycols=true` (the default), then returned `DataFrame` holds
 copies of column vectors in `df`.
 If `copycols=false`, then returned `DataFrame` shares column vectors with `df`.
 
+If `select` is called on `SubDataFrame` then
```
```suggestion
If `df` is a `SubDataFrame` then
```
(also break next line at 92 chars)

src/dataframe/dataframe.jl
```diff
@@ -1024,15 +930,23 @@ select!(df::DataFrame, c::Any) = select!(df, index(df)[c])
...
+`SubDataFrame` is returned if `copycols=false` and `DataFrame` with freshly allocated columns otherwise.
+
+If `select` is called on `DataFrameRow` then
```
```suggestion
If `df` is a `DataFrameRow` then a `DataFrameRow` pointing to the same parent data frame is returned.
```
(and remove next line)

src/dataframe/dataframe.jl
```diff
@@ -1024,15 +930,23 @@ select!(df::DataFrame, c::Any) = select!(df, index(df)[c])
 
 """
     select(df::DataFrame, inds, copycols::Bool=true)
+    select(dfv::SubDataFrame, inds, copycols::Bool=true)
```
nalimilan: Better have a single signature for `AbstractDataFrame`? Then the text can mention that `copycols` isn't supported for `DataFrameRow` (in theory we could allocate a new parent though if we wanted).

src/dataframe/dataframe.jl
```diff
@@ -1281,18 +1205,18 @@ function categorical!(df::DataFrame, cname::Union{Integer, Symbol};
     df
 end
 
-function categorical!(df::DataFrame, cnames::Vector{<:Union{Integer, Symbol}};
+function categorical!(df::DataFrame, cnames::AbstratVector{<:Union{Integer, Symbol}};
```
```suggestion
function categorical!(df::DataFrame, cnames::AbstractVector{<:Union{Integer, Symbol}};
```

src/abstractdataframe/abstractdataframe.jl
bkamins: `df[cols]` and `select` have two differences:
* `select` allows us to avoid copying of columns and `df[cols]` always copies columns if `cols` is a collection (as in this case)
* `select` always returns an `AbstractDataFrame` and `df[cols]` can return a vector (this is not the case here as if a single column is passed we have a special faster method above)

In general these differences are the reasons why `select` exists at all, otherwise, we could use `getindex`.

src/abstractdataframe/abstractdataframe.jl
bkamins: I can reintroduce them. The problem was that you still have to add `Colon` to the list and it seemed overly long (this correct handling of `Colon` is one of the small things I clean in this PR).

bkamins: and also `Not` has to be added

src/dataframe/dataframe.jl
bkamins: OK. For `DataFrameRow` I thought it is completely not useful and the default should be `copycols=false` (as opposed to the other methods) anyway so I thought that it is better not to support it.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Yes, it's verbose, but that's not the end of the world. We can create an alias if needed (a much more useful one than the current `ColumnIndex`).

src/dataframe/dataframe.jl

src/abstractdataframe/abstractdataframe.jl
bkamins: I was thinking about an alias - I will create it.

src/dataframe/dataframe.jl
nalimilan: Maybe don't support `select` on `DataFrameRow` at all? It sounds problematic to use a different default, and anyway a `DataFrameRow` isn't a data frame.

src/abstractdataframe/abstractdataframe.jl
bkamins: Actually it is not that easy - in different places we split the `Union{AbstractVector, Regex, Not, Colon}` in different ways to handle the specific cases fast. Maybe I will leave it as is explicitly.

nalimilan: Though if there's a more specific `Union`, it should take precedence in dispatch.

src/dataframe/dataframe.jl

src/abstractdataframe/abstractdataframe.jl
bkamins: Yes, but then users might get confused. I propose to leave it explicit.

bkamins:
This PR should be good to have a look at.

Notes:
* it supports `Not` for row and column indexing
* it does not support `Not` in `setindex!` (this should be done in a separate PR that cleans up `setindex!` that should be merged before the next release)
* in the long run all codes related to indexing should be refactored to make the design simpler I think, but this can be done after 1.0 release as discussed earlier

Test on nightly fail because of InteractiveUtils.jl not-found error (I did not investigate into it but I think this is unrelated).

Project.toml
```diff
@@ -31,11 +32,12 @@ test = ["DataStructures", "DataValues", "Dates", "LaTeXStrings", "Random", "Test
 
 [compat]
 julia = "1"
-Missings = ">= 0.2.3"
 CategoricalArrays = ">= 0.5.4"
-StatsBase = ">= 0.11.0"
 Compat = ">= 0.59.0"
-Tables = ">= 0.2.3"
+InvertedIndices = ">= 1.0.0"
```
nalimilan: This should be OK since the package has reached 1.0:
```suggestion
InvertedIndices = "1"
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -573,27 +574,29 @@ julia> completecases(df, [:x, :y])
 ```
 
 """
-function completecases(df::AbstractDataFrame)
+function completecases(df::AbstractDataFrame, col::Colon=:)
+    if ncol(df) == 0
```
nalimilan: Maybe we should just return `DataFrame()`?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -573,27 +574,29 @@ julia> completecases(df, [:x, :y])
...
-completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex}) =
-    completecases(df[cols])
+completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex, Not}) =
+    completecases(select(df, cols, copycols=false))
 
 """
     dropmissing(df::AbstractDataFrame; disallowmissing::Bool=true)
-    dropmissing(df::AbstractDataFrame, cols::Union{AbstractVector, Regex}; disallowmissing::Bool=true)
-    dropmissing(df::AbstractDataFrame, cols::Union{Integer, Symbol}; disallowmissing::Bool=true)
+    dropmissing(df::AbstractDataFrame, cols; disallowmissing::Bool=true)
```
nalimilan: `cols` should be described a bit in the text. I think we have a standard expression for values that can be used with indexing.

Also, the types are listed for `completecases`, so we should be consistent.

Same for `dropmissing!`.

src/abstractdataframe/abstractdataframe.jl
nalimilan: OK. But better keep the `Union` in the signature. Same in other functions.

src/dataframe/dataframe.jl
```diff
@@ -1016,25 +932,35 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
...
 
 select!(df::DataFrame, c::Int) = select!(df, [c])
 select!(df::DataFrame, c::Any) = select!(df, index(df)[c])
 
 """
-    select(df::DataFrame, inds, copycols::Bool=true)
+    select(df::AbstractDataFrame, inds, copycols::Bool=true)
 
-Create a new `DataFrame` that contains columns from `df` specified by `inds` and return it.
+If `df` is a data frame create a new data frame that contains columns from `df`
```
nalimilan: So `df` is not necessarily a data frame? :-/

src/dataframe/dataframe.jl
```diff
@@ -1016,25 +932,35 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
...
-Argument `inds` can be any index that is allowed for column indexing of a `DataFrame`.
+Argument `inds` can be any index that is allowed for column indexing.
 
+If `df` is a `DataFrame` return a new `DataFrame` that contains columns from `df`
+specified by `inds`.
 If `copycols=true` (the default), then returned `DataFrame` holds
 copies of column vectors in `df`.
 If `copycols=false`, then returned `DataFrame` shares column vectors with `df`.
 
+If `df` is a `SubDataFrame` then `SubDataFrame` is returned if `copycols=false`
```
```suggestion
If `df` is a `SubDataFrame` then a `SubDataFrame` is returned if `copycols=false`
```

src/dataframe/dataframe.jl
```diff
@@ -1016,25 +932,35 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
...
+and `DataFrame` with freshly allocated columns otherwise.
```
```suggestion
and a `DataFrame` with freshly allocated columns otherwise.
```

src/deprecated.jl
```diff
@@ -1402,4 +1402,7 @@ import Base: haskey
 @deprecate haskey(df::AbstractDataFrame, key::Any) key in names(df)
 
 import Base: empty!
-@deprecate empty!(df::DataFrame) deletecols!(df, 1:ncol(df))
\ No newline at end of file
+@deprecate empty!(df::DataFrame) deletecols!(df, 1:ncol(df))
+
+@deprecate deletecols!(df::DataFrame, inds) select!(df, Not(inds))
+@deprecate deletecols(df::DataFrame, inds; copycols::Bool=true) select(df, Not(inds), copycols=copycols)
```
nalimilan: Maybe also add a version without keyword arguments to keep the most common case simple.

src/other/index.jl
```diff
@@ -138,66 +138,66 @@ function Base.insert!(x::Index, idx::Integer, nm::Symbol)
...
-function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
-        idxs = uppercase(string(idx))
-        dist = [(REPL.levenshtein(uppercase(string(x)), idxs), x) for x in keys(l)]
-        sort!(dist)
-        c = [count(x -> x[1] <= i, dist) for i in 0:2]
-        maxd = max(0, searchsortedlast(c, 8) - 1)
-        [s for (d, s) in dist if d <= maxd]
+@inline function Base.getindex(x::AbstractIndex, idx::Integer)
+    if !(1 <= idx <= length(x))
+        throw(BoundsError("attempt to access $(length(x))-element AbstractIndex at index $idx"))
```
nalimilan: Maybe this should say "data frame" rather than "AbstractIndex" and "columns" instead of "element"? This is supposed to be seen by end users, right?

Same below.

src/other/index.jl
```diff
@@ -138,66 +138,66 @@ function Base.insert!(x::Index, idx::Integer, nm::Symbol)
...
-@inline Base.getindex(x::Index, idx::Symbol) = lookupname(x.lookup, idx)
-@inline Base.getindex(x::Index, idx::AbstractVector{Symbol}) =
-    [lookupname(x.lookup, i) for i in idx]
+@inline function Base.getindex(x::AbstractIndex, idx::AbstractRange{Int})
+    isempty(idx) && return idx
+    minidx, maxidx = extrema(idx)
+    if minidx < 1
+        throw(BoundsError("attempt to access $(length(x))-element AbstractIndex at index $minidx"))
+    end
+    if maxidx > length(x)
```
nalimilan: Merge with previous branch?

test/duplicates.jl
```diff
@@ -20,6 +21,24 @@ const ≅ = isequal
...
+
+    df = view(DataFrame(a = [1, 2, 3, 3, 4]), :, :)
+    udf = DataFrame(a = [1, 2, 3, 4])
+    @test nonunique(df) == [false, false, false, true, false]
+    @test udf == unique(df)
+    @test_throws ArgumentError unique!(df)
+    @test_throws ArgumentError unique(df, true)
+
+     pdf = view(DataFrame(a = CategoricalArray(["a", "a", missing, missing, "b", missing, "a", missing]),
+                         b = CategoricalArray(["a", "b", missing, missing, "b", "a", "a", "a"])), :,  :)
```
nalimilan:
```suggestion
                          b = CategoricalArray(["a", "b", missing, missing, "b", "a", "a", "a"])), :,  :)
```

test/duplicates.jl
```diff
@@ -20,6 +21,24 @@ const ≅ = isequal
...
+    updf = DataFrame(a = CategoricalArray(["a", "a", missing, "b", missing]),
+                    b = CategoricalArray(["a", "b", missing, "b", "a"]))
```
nalimilan:
```suggestion
                     b = CategoricalArray(["a", "b", missing, "b", "a"]))
```

test/iteration.jl
```diff
@@ -112,7 +112,7 @@ end
     @test copy(erv[1]) == (y3=33, y1=53, y4=43)
     df.z = 1
     @test length(erd[1]) == 5 # the added column is reflected
-    deletecols!(df, [4,5])
+    select!(df, Not([4,5]))
```
nalimilan: Can you add a basic test for `deletecols!` in test/deprecated.jl?

bkamins: added

src/other/index.jl
bkamins: The error message is different

src/other/index.jl

src/deprecated.jl
bkamins: It is not possible unfortunately as `copycols` is a kwarg so then we would have two declarations with the same signature and one `@deprecate` would get overwritten by the other.

src/dataframe/dataframe.jl

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
bkamins: fixed. I have also patched some more docstrings.

src/deprecated.jl
nalimilan: Actually you can cheat by using different types for positional arguments (e.g. `AbstractDataFrame`), but maybe that's not worth it. See https://github.com/JuliaLang/julia/issues/9498.

nalimilan: Actually it works even when the types are the same:
```julia
julia> f(x::Int; y=1) = 2

julia> f(x::Int) = 1

julia> f(1)
1

julia> f(1, y=2)
2
```
(That's probably one of the most obscure "features" of dispatch.)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -596,7 +596,9 @@ completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex, Not}) =
 
 """
     dropmissing(df::AbstractDataFrame; disallowmissing::Bool=true)
-    dropmissing(df::AbstractDataFrame, cols; disallowmissing::Bool=true)
+    dropmissing(df::AbstractDataFrame, ::Colon; disallowmissing::Bool=true)
```
nalimilan: Why not merge this with the one below? Or even merge the three last signatures?

Also need to break before `;`.

src/dataframe/dataframe.jl
```diff
@@ -1127,19 +1053,14 @@ Base.hcat(df1::DataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame...;
 ##############################################################################
 """
     allowmissing!(df::DataFrame)
+    allowmissing!(df::DataFrame, cols::Colon)
+    allowmissing!(df::DataFrame, cols::Union{Integer, Symbol})
+    allowmissing!(df::DataFrame, cols::Union{AbstractVector, Regex, Not})
 
-Convert all columns of a `df` from element type `T` to
+Convert columns `cols` of a `df` from element type `T` to
```
```suggestion
Convert columns `cols` of data frame `df` from element type `T` to
```

src/dataframe/dataframe.jl
```diff
@@ -1127,19 +1053,14 @@ Base.hcat(df1::DataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame...;
...
-    allowmissing!(df::DataFrame, col::Union{Integer, Symbol})
-
-Convert a single column of a `df` from element type `T` to
-`Union{T, Missing}` to support missing values.
-
-    allowmissing!(df::DataFrame, cols::AbstractVector})
-
-Convert multiple columns of a `df` from element type `T` to
-`Union{T, Missing}` to support missing values.
+If `cols` is ommited all columns in the data frame are converted.
```
```suggestion
If `cols` is omitted all columns in the data frame are converted.
```

src/dataframe/dataframe.jl
```diff
@@ -1163,21 +1084,22 @@ function allowmissing!(df::DataFrame, cols::AbstractVector{Bool})
...
 
-    disallowmissing!(df::DataFrame, col::Union{Integer, Symbol})
+"""
+    disallowmissing!(df::DataFrame)
+    disallowmissing!(df::DataFrame, cols::Colon)
+    disallowmissing!(df::DataFrame, cols::Union{Integer, Symbol})
+    disallowmissing!(df::DataFrame, cols::Union{AbstractVector, Regex, Not})
 
-Convert a single column of a `df` from element type `Union{T, Missing}` to
+Convert columns `cols` of a `df` from element type `Union{T, Missing}` to
```
```suggestion
Convert columns `cols` of data frame `df` from element type `Union{T, Missing}` to
```

src/dataframe/dataframe.jl
```diff
@@ -1163,21 +1084,22 @@ function allowmissing!(df::DataFrame, cols::AbstractVector{Bool})
...
 `T` to drop support for missing values.
 
-    disallowmissing!(df::DataFrame, cols::Union{AbstractVector, Regex})
-
-Convert multiple columns of a `df` from element type `Union{T, Missing}` to
-`T` to drop support for missing values.
+If `cols` is ommited all columns in the data frame are converted.
```
```suggestion
If `cols` is omitted all columns in the data frame are converted.
```

src/deprecated.jl
bkamins: I have tested a reverse order I have checked when originally replying to you. I have commented on it in https://github.com/JuliaLang/julia/issues/9498. I will add the other definition then.

src/abstractdataframe/abstractdataframe.jl

PR #1852, opened 2019-06-17
Implement hasproperty

nalimilan:
This is consistent with our definition of `getproperty` and `setproperty!`.
It also provides an efficient deprecation for `haskey`, though it turns out the current deprecation didn't support non-`Symbol`s, contrary to the function it replaces.


quinnj:
I agree this is fine to have here, but I'd also suggest just putting this in Base.

nalimilan:
> I agree this is fine to have here, but I'd also suggest just putting this in Base.

You mean `hascol`? `hasproperty` has just been added in Base, and for now it's in Compat.

quinnj:
Ah, I forgot it got put in! Yes, that's right. Carry on.

PR #1855, opened 2019-06-21
remove InteractiveUtils load

bkamins:
Loading `InteractiveUtils` is not needed and it causes tests to fail on nightly


bkamins:
@nalimilan CI fails due to coverage problems, but I do not see the reason for them.

Issue #1844, opened 2019-06-10
Make `getproperty(df, col)` return a full length view of the column


oxinabox:
Similar for the single arg `getindex` (`getindex(df, col::Symbol)`), and `eachcol`.

We are progressively knocking out ways of ending up with columns of different sizes.

Right now here are ways that you can't end up with incosistant sizes:
 - You can't assign in a new column that doesn't have the right size, as `setproperty` checks the size.
 - You can't pass in a `Vector` then resize it using another reference, as the `DataFrame` constructor copies.
 - You can't use normal indexing indexing to get access to columns that you then resize, as normal indexing returns copies 
 - you can't use `@view` indexing returns `SubDataFrame`s which disallow size mutating operations, and who's column vectors are views anyway, so also disallow resizing operations (right?) 
 - and size `setproperty`  (and 1 arg setindex) check the size of the column being added, you can't just insert one with incorrect size.

Right now the only way I can think of getting access to a inner column and then mutating its size,
is the use of `getproperty` or single argument `getindex` or `eachcol`.
Which return the actually underlying `Vector`.

And I was thinking: It would be great if we could just wrap those in some kind of `view` like array wrapper that doesn't allow resizing in place, but allows all the other operations one would hope from a `Vector`, including in-place `setindex` of elements.

Turns out such a view likes wrapper does exist.
The `SubArray`.
We can just return a the equivalent  of `@view df.col[:]`.
The overhead of creating a view is tiny, as is the overhead of working with a view (at least when the indexing is simple which it is in this case).
You can do all the things to a `SubArray` as long as you don't call `resize!` on it -- that is a `MethodError`.

If someone needs to actually access the raw array, then they can use `parent(df.col)`.
Only reason I can think that would really be needed is for if a method has a overly strict set of type constraints, then accessing the `parent` would be an alternative to `collect` (with its own trade-offs).

I think this would knock off the last possible way to end up with a corrupt `DataFrame`,
via column related shenanigans.

bkamins:
There would be still one hole:
```
df.a = vector
```
and then resize `vector`.

Still - patching up things is not that bad :). Actually this would be consistent with `df[row,cols]` returning a view.

An alternative (and this is something we really should discuss here - as we will always have holes) is to create a list of methods in DataFrames.jl that actively check if all columns have the same length before performing their operation (essentially - all expensive methods).

oxinabox:
Those are 2 good points, lets open seperate issues to discuss them,
[...]

PR #1856, opened 2019-06-23
WIP: make indexing and eachcol return views of data frame columns

bkamins:
This fixes https://github.com/JuliaData/DataFrames.jl/issues/1844.

I open this in case there are any design comments. This is still WIP as I have to do two things:
* review documentation in detail to make sure it is updated (I will also make other changes to documentation as I read it if it is outdated in this PR to simplify my life)
* go though DataFrames.jl internals to make sure we do not rely on `df[:col]` returning the actual stored vector not its `view`

### DIFF
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -44,7 +44,7 @@ If it is performed a description explicitly mentions that the data is *copied*.
 For performance reasons, accessing, via `getindex` or `view`, a single `row` and multiple `cols` of a `DataFrame`, a `SubDataFrame` or a `DataFrameRow` always returns a `DataFrameRow` (which is a view-like type).
 
 `DataFrame`:
-* `df[col]` -> the vector contained in column `col`;
+* `df[col]` -> a view of the vector contained in column `col`;
 * `df[cols]` -> a freshly allocated `DataFrame` containing the copies of vectors contained in columns `cols`;
 * `df[row, col]` -> the value contained in row `row` of column `col`, the same as `df[col][row]`;
 * `df[CartesianIndex(row, col)]` -> the same as `df[row,col]`;
--- a/docs/src/lib/types.md
+++ b/docs/src/lib/types.md
@@ -44,10 +44,10 @@ The return value can have two concrete types:
 
 * If the `eachcol` function is called with the `names` argument set to `true` then it returns a value of the
   `DataFrameColumns{<:AbstractDataFrame, Pair{Symbol, AbstractVector}}` type, which is an
-  iterator returning a pair containing the column name and the column vector.
+  iterator returning a pair containing the column name and the veiw of a column vector.
 * If the `eachcol` function is called with `names` argument set to `false` (the default) then it returns a value of the
   `DataFrameColumns{<:AbstractDataFrame, AbstractVector}` type, which is an
-  iterator returning the column vector only.
+  iterator returning the view of a column vector only.
 
 The `DataFrameRows` and `DataFrameColumns` types are subtypes of `AbstractVector` and support its interface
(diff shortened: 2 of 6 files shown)


bkamins:
I already see one slight issue with this PR. Earlier when we written `df[col]` we could distinguish type of `df` by the return value (at least in "normal" cases). Now it is not possible as `df[col]` always returns a `SubArray` no matter if `df` is a `DataFrame` or `SubDataFrame`. (see the required change to `hcat!` in this PR for the consequences)

nalimilan:
> I already see one slight issue with this PR. Earlier when we written `df[col]` we could distinguish type of `df` by the return value (at least in "normal" cases). Now it is not possible as `df[col]` always returns a `SubArray` no matter if `df` is a `DataFrame` or `SubDataFrame`. (see the required change to `hcat!` in this PR for the consequences)

Indeed. Though AFAICT that's not a big problem, right? We could add an argument to `getcol` to allow extracting the original vector rather than a view (for types where it makes sense; I'm thinking of `DataFrame` but also `GroupedDataFrame` soon probably).

bkamins:
I was thinking over this and I think it is mostly problem for me (as I do internal development). For people having a `view` always is actually more consistent.

Now you have raised a `getcol` function issue. We discussed earlier that it is not really needed. What do you think would be its use case.

bkamins:
Regarding `getcol`. Sorry for my mistake.
What I wanted to ask is the following:

what kind of behaviors (so essentially keyword arguments) we want in `getcol`

assuming `getcol(df, col)` replaces `df[col]` and will return a view.

nalimilan:
It could make sense to have a keyword argument to `getcol` to request "the simplest type which can represent a column": i.e. `Vector` if possible (for `DataFrame`, or later `GroupedDataFrame`), `SubVector` otherwise. But I don't know how to call that argument. Anyway we can defer this.

bkamins:
@nalimilan We have revenge of the `CategoricalArray` issue again. If we always return views then the problem is that a `view` of `CategoricalArray` is not `AbstractCategoricalArray`. This might break a lot of code that relies on checking if some column in a data frame is categorical. What do you think about this case? A direct approach is to have a special type for `view` of `CategoricalArray` defined in Categorical.jl.

nalimilan:
I think people should just switch to `AbstractArray{<:Union{CategoricalString,CategoricalValue}}`, as they should support views anyway.

bkamins:
@oxinabox It turns out that returning a `view` by default from `df.x` does not play well with CategoricalArrays.jl. The problem is that such a `view` is not a subtype of `AbstractCategoricalArray`. Unfortunately we cannot make it to be such a subtype because it is already a subtype of `SubArray`.

The consequence is that introducing such a change would have to be synchronized with changes in all packages that use CategoricalArrays.jl and DataFrames.jl to make sure that they do not break (and they probably would as up till now the normal way to check if something is a categorical variable was checking its type against `AbstractCategoricalArray` or even more narrowly `CategoricalArray`)..

nalimilan:
So in other words it's not completely clear that the involved work and increased complexity (for us, but also for users) are worth the limited increase in safety.

quinnj:
Personally, I think we'll run into a decent amount of cases where users will be surprised by this, not unlike the switch in `CSV.read` to return read-only `Column` by default.

oxinabox:
I think it is worth it,
but I think the way to do it is the return a custom column type.
This means we could have out own `NoResizeCategoricalVector <: AbstractCategoricalArray`.

bkamins:
The conclusion from the discussions and the implementation in #1866 is that we will return vectors not their views so I am closing this PR.

PR #1864, opened 2019-06-28
[WIP] complete and expand df

versipellis:
First PR I'm ever making to a big project, so bear with my noviceness :)

This PR spawns from a conversation on the Slack #data channel. It takes a dataframe as an input, and some vector of columns from that dataframe, and turns implicit into explicit missing values.

`expanddf(df, indexcols)` returns a dataframe of every possible combination of index cols, the length of which would be the product of the length of every column's unique values.

`completedf(df, indexcols, fill=missing)` wraps `expanddf`, but joins the original dataframe back, and provides the utility to fill the newly-missing values. 

I haven't written the documentation in anticipation that feedback might cause a number of changes to the implementation. See below for examples in use.

The equivalent in...
R Tidyverse: https://tidyr.tidyverse.org/reference/complete.html
Python Pandas: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.reindex.html
Stata: https://www.stata.com/manuals13/dfillin.pdf

Example:
```julia
julia> df = DataFrame(A=["05","05","05","06","06","07","07",missing],B=["a",missing,"c","a",missing,"c","d","a"],C=[1,1,10,43,5,10,3,10])

julia> indexcols = [:A,:B]

julia> df
8×3 DataFrame
│ Row │ A       │ B       │ C     │
│     │ String⍰ │ String⍰ │ Int64 │
├─────┼─────────┼─────────┼───────┤
│ 1   │ 05      │ a       │ 1     │
│ 2   │ 05      │ missing │ 1     │
│ 3   │ 05      │ c       │ 10    │
│ 4   │ 06      │ a       │ 43    │
│ 5   │ 06      │ missing │ 5     │
│ 6   │ 07      │ c       │ 10    │
│ 7   │ 07      │ d       │ 3     │
│ 8   │ missing │ a       │ 10    │

julia> expanddf(df, indexcols)
16×2 DataFrame
│ Row │ A       │ B       │
│     │ String⍰ │ String⍰ │
├─────┼─────────┼─────────┤
│ 1   │ 05      │ a       │
│ 2   │ 06      │ a       │
│ 3   │ 07      │ a       │
│ 4   │ missing │ a       │
│ 5   │ 05      │ missing │
⋮
│ 11  │ 07      │ c       │
│ 12  │ missing │ c       │
│ 13  │ 05      │ d       │
│ 14  │ 06      │ d       │
│ 15  │ 07      │ d       │
│ 16  │ missing │ d       │

julia> completedf(df, indexcols)
16×3 DataFrame
│ Row │ A       │ B       │ C       │
│     │ String⍰ │ String⍰ │ Int64⍰  │
│ 1   │ 05      │ a       │ 1       │
│ 2   │ 06      │ a       │ 43      │
│ 3   │ 07      │ a       │ missing │
│ 4   │ missing │ a       │ 10      │
│ 5   │ 05      │ missing │ 1       │
⋮
│ 11  │ 07      │ c       │ 10      │
│ 12  │ missing │ c       │ missing │
│ 13  │ 05      │ d       │ missing │
│ 14  │ 06      │ d       │ missing │
│ 15  │ 07      │ d       │ 3       │
│ 16  │ missing │ d       │ missing │

julia> completedf(df, indexcols, fill=10000)
16×3 DataFrame
│ Row │ A       │ B       │ C     │
│     │ String⍰ │ String⍰ │ Int64 │
├─────┼─────────┼─────────┼───────┤
│ 1   │ 05      │ a       │ 1     │
│ 2   │ 06      │ a       │ 43    │
│ 3   │ 07      │ a       │ 10000 │
│ 4   │ missing │ a       │ 10    │
│ 5   │ 05      │ missing │ 1     │
⋮
│ 11  │ 07      │ c       │ 10    │
│ 12  │ missing │ c       │ 10000 │
│ 13  │ 05      │ d       │ 10000 │
│ 14  │ 06      │ d       │ 10000 │
│ 15  │ 07      │ d       │ 3     │
│ 16  │ missing │ d       │ 10000 │
```


src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+
+##############################################################################
+##
+## Expand & Complete dataframes
+##
+##############################################################################
+
+
+function expanddf(df::AbstractDataFrame, indexcols::Array{Symbol,1})
+    colnames = names(df)
```
bkamins: you can use `_names` to avoid a copy

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
 end
+
+##############################################################################
+##
+## Expand & Complete dataframes
+##
+##############################################################################
+
+
+function expanddf(df::AbstractDataFrame, indexcols::Array{Symbol,1})
```
bkamins: I would leave `indexcols` either undefined or a wide definition accepting all we currently accept: `ColumnIndex`, `AbstractVector{<:ColumnIndex`, `Regex`, `Not`, `Colon`.

src/dataframe/dataframe.jl
bkamins: Or even avoid it and call `index(df)[indexcols]` and later work on integers - this will be more flexible.

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+    colnames = names(df)
+    # Check to make sure the symbols in indexcols are in the df
+    for i in indexcols
```
bkamins: if you use the `index` approach then here and possibly later you can greatly simplify code and improve its performance.

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+        if i ∉ colnames
+            throw(ArgumentError(":$i is not in dataframe"))
+        end
+    end
+    dummydf = similar(df[indexcols],0)
```
bkamins: `df[indexcols]` will copy and in general is going to be deprecated. If you want to use this design probably it is better to write `select(df, indexcols, copycols=false)`. Also only now you will catch a problem if you have duplicates in `indexcols` maybe it is better to handle it earlier.

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+        uniqueVals[col] = unique(df[col])
+    end
+
+    # Get a long vector of every possible combination
+    collected = collect(flatten(product(values(uniqueVals)...)))
+    dfsize = div(length(collected),length(indexcols))
+
+    # Re-format to dataframe
+    for i in 1:length(indexcols):length(collected)
+        push!(dummydf, collected[i:(i+length(indexcols)-1)])
```
bkamins: better use `view` here for performance. But seeing this code it is probably better to first construct the columns and then make a data frame from them. Here is a sample code I have written for similar thing in the past: https://github.com/PacktPublishing/Julia-1.0-Programming-Cookbook/blob/dfbb5b6c482d74f844840712dc206dcc567631ed/Chapter04/03.%20Generating%20full%20factorial%20designs/expand.jl. It does not materialize a `DataFrame` but is pretty similar.

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+        if i ∉ colnames
+            throw(ArgumentError(":$i is not in dataframe"))
+        end
+    end
+
+    # Expand the input df and left join
+    expanded = expanddf(df, indexcols)
+    expanded = join(expanded, df; on = indexcols, kind = :left)
+
+    # Replace missing values with the fill
```
bkamins: There is a tricky issue here that the original columns of `df` might have had contained missing values. I understand that they would be overwritten by `fill`. `join` actually allows you to distinguish case of `missing` originating from the `df` vs `missing` produced by `join` because you can temporarily generate the column that tells you if the source contained such a combination of values.

But maybe you wanted the implementation as it is now (or maybe we should give an option for the user to choose).

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+
+    # Re-format to dataframe
+    for i in 1:length(indexcols):length(collected)
+        push!(dummydf, collected[i:(i+length(indexcols)-1)])
+    end
+
+    return dummydf
+end
+
+function completedf(df::AbstractDataFrame, indexcols::Array{Symbol,1}; fill=missing::Any)
```
bkamins: same comment as above

src/dataframe/dataframe.jl
bkamins: also we might add an option to skip rows with `missing` values in `indexcols` like in `groupby`

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+    # Check to make sure the symbols in indexcols are in the df
+    for i in indexcols
+        if i ∉ colnames
+            throw(ArgumentError(":$i is not in dataframe"))
+        end
+    end
+
+    # Expand the input df and left join
+    expanded = expanddf(df, indexcols)
+    expanded = join(expanded, df; on = indexcols, kind = :left)
```
bkamins: This will produce multiple rows on duplicate combinations of values in `indexcols` - I just want to make sure we want it (or put it otherwise `expanded = join(expanded, df; on = indexcols, kind = :left)` may add rows to `expanded` data frame - do we want this?). Maybe - again - we should add a keyword argument to let the user decide on it.

versipellis:
Thank you for the review! I'll go through them in the next few days and make the changes/see if there's a sane approach for some of them.

bkamins:
Thank you. Also one more thing. If a column is categorical you probably should not use `unique` but `levels` - is this what you have indented? Probably @nalimilan can comment what is the best strategy to check if a column is categorcal given the planned changes in CategoricalArrays.jl (@nalimilan - note that `df` in this PR can be `SubDataFrame` in particular).

src/dataframe/dataframe.jl
nalimilan: An interesting related question is how to support `nesting` as in dplyr, which allows specifying that a combination of some variables should not be expanded for values which don't appear in the data. This can be a future extension, but we need to make sure we will be able to add it with the current interface. Maybe we could just support that by passing tuples of columns inside the array?

src/dataframe/dataframe.jl
nalimilan: dplyr seems to do that too:
> If you supply fill, these values will also replace existing explicit missing values in the data set.

But that doesn't seem like a great idea to me (you can do that manually if you want). Better distinguish these by default as you suggest.

nalimilan:
Thanks. Regarding the API, I wonder whether we need both `expand` and `complete` (I would drop the "df" suffix anyway): wouldn't it make sense to have a single function, with an argument to determine whether the other columns should be kept (`complete`) or not (`expand`)?

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
...
+    # Check to make sure the symbols in indexcols are in the df
+    for i in indexcols
+        if i ∉ colnames
+            throw(ArgumentError(":$i is not in dataframe"))
+        end
+    end
+    dummydf = similar(df[indexcols],0)
+
+    # Create a dictionary of symbol=>vector of unique values
+    uniqueVals = Dict{Symbol, Array{Any,1}}()
```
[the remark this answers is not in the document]
bkamins: That is why I have recommended to switch to to variable index representation. Then actually probably using a `Tuple` is simplest. If we stick with using names then `NamedTuple` can be used.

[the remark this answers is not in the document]
nalimilan: Tuples aren't great when the number of entries isn't known at compile time. If `index(df)[indexcols]` is used at the top, we can work only with integer indices and a vector, right?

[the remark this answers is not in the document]
bkamins: Right. Then you have `Vector{Int}` for selection (you have to special case `ColumnIndex` to convert it to a single entry vector - but this is a standard thing we do in other methods) and `Vector{AbstractVector}` to store the columns.

versipellis:
> Thanks. Regarding the API, I wonder whether we need both `expand` and `complete` (I would drop the "df" suffix anyway): wouldn't it make sense to have a single function, with an argument to determine whether the other columns should be kept (`complete`) or not (`expand`)?

That would certainly be more elegant. I kept the different functions mostly to keep in mind users transitioning from R. 

> Thank you. Also one more thing. If a column is categorical you probably should not use `unique` but `levels` - is this what you have indented? Probably @nalimilan can comment what is the best strategy to check if a column is categorcal given the planned changes in CategoricalArrays.jl (@nalimilan - note that `df` in this PR can be `SubDataFrame` in particular).

Quite honestly, I forgot about categorical as a column type, and haven't used it much myself. I can go take a look at how to implement this though :)

src/dataframe/dataframe.jl
versipellis: Sorry, could you explain a little more how that makes it more flexible? And what do you mean "and later work on integers"?

src/dataframe/dataframe.jl
versipellis: @nalimilan I would probably add that as an extension in the future, but will definitely keep it in mind. That said, DataFrames doesn't have an equivalent concept of `nest()`, does it? I'm trying to think of what structure would be passed in to indicate nested columns, or whether this PR should support creating the nesting itself.

src/dataframe/dataframe.jl
versipellis: I like the idea of distinguishing the original missing values from ones created by this package.

src/dataframe/dataframe.jl
versipellis: Are you referring to duplicate columns that would create duplicate combinations of values?

src/dataframe/dataframe.jl
```diff
@@ -1508,3 +1508,63 @@ end
 function permutecols!(df::DataFrame, p::AbstractVector{Symbol})
     permutecols!(df, index(df)[p])
 end
+
```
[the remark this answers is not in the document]
versipellis: My original intent was have this be primarily for DataFrame, but I'm more than happy to move it out to its own file, too. I originally considered this as a PR for DataFramesMeta, which (to me, at least) seemed like the closest dplyr-esque package, but I wasn't too sure.

src/dataframe/dataframe.jl
bkamins: You can supply to `index(df)[indexcols]` a wide selection of `indexcols` vales as opposed to your proposal. E.g. `indexcols` can be a `Regex`, or `Not` expression. Then `index(df)[indexcols]` guarantees you that you will receive as a result either `Int` or `Vector{Int}`. This way later you do not have to handle the multiple different possible options but work only with integers.

src/dataframe/dataframe.jl
bkamins: I am referring to a data structure like:
```
julia> df = DataFrame(x=[1,1,1,2,2,2], y=[1,2,1,2,1,2], z=[1,2,3,4,5,6])
6×3 DataFrame
│ Row │ x     │ y     │ z     │
│ 1   │ 1     │ 1     │ 1     │
│ 2   │ 1     │ 2     │ 2     │
│ 3   │ 1     │ 1     │ 3     │
│ 4   │ 2     │ 2     │ 4     │
│ 5   │ 2     │ 1     │ 5     │
│ 6   │ 2     │ 2     │ 6     │
```

now when you do `expanddf(df, [:x, :y])` you get 4 rows, but `completedf` will produce 6 rows. And the question is if this is intended or you should throw an error in such case that duplicate were found, or you should have a keyword argument if duplicates should be allowed.

src/dataframe/dataframe.jl
nalimilan: dplyr's `nesting` in `expand` is completely different from tidyr's `nest` AFAICT.

src/dataframe/dataframe.jl
versipellis: I considered constructing the columns first, but since there could be any number of columns, I didn't want to have to reach for metaprogramming to implement. This seemed like the most elegant approach I could think of at the time.

bkamins: maybe do something like
```
function _helper(col, c, stride, offset)
    for i in offset:stride:length(c)
        push!(col, c[i])
    end
end
for colind in 1:length(indexcols)
    _helper(dummydf[!, colind], collected, length(indexcols), colind)
end
```
(I am writing it from my head so this is not tested)

versipellis: Gotcha, I see what you're trying to get at. I'm still new to writing performance Julia code, but am I understanding correctly that building each column as an array and joining the arrays into a DF is faster?

bkamins: Yes - I am using a function barrier technique (see https://docs.julialang.org/en/latest/manual/performance-tips/#kernel-functions-1).

src/dataframe/dataframe.jl
versipellis: I see what you mean. My feeling is to throw an error in the case of `completedf`, and maintain the current behavior of `expanddf`.

This also makes me consider the case provided by your example where there's already a fully expanded DF being fed in. I think `expanddf`'s behavior works as is, but as above, `completedf` should throw an error. Thoughts?

src/dataframe/dataframe.jl
versipellis: I added a kwarg to `complete` called `replaceallmissing=false`. By default (false) it will only replace the missing values created by the expansion using the `:source` column from the join. true will replace all missings.

src/dataframe/dataframe.jl
versipellis: I couldn't get `dummydf[!,colind]` working (the exclamation mark syntax doesn't seem to work), so I ended up with something like this:
```
function _expandhelper(vect, stride, offset)
    assembledcol = []
    for i in offset:stride:length(vect)
        push!(assembledcol, vect[i])
    end
    return assembledcol
end
```
and
```
...
dummydf = DataFrame()
    for i in 1:length(indexcols)
        hcat(dummydf,_expandhelper(collected, length(indexcols), i))
    end
...
```
I've got something like that working, but I'm not entirely sure if it's as performant as your sketch.

src/dataframe/dataframe.jl
bkamins: I think it is not the issue if the table is fully epanded or not, but if it contains duplicates.
But if I understand you correctly you want `expanddf` to produce a deduplicated full factorial table and `completedf` to throw an error on duplicates. I think it is OK, but please document it.

src/dataframe/dataframe.jl
bkamins: ok

src/dataframe/dataframe.jl
bkamins: You need to rebase your PR against master to make `dummydf[!, colind]` work.

versipellis: Oh! That's why - whoops. Didn't realize that there was a new release; thanks!

versipellis:
@bkamins @nalimilan I think I've addressed all the comments that were raised, except for the one about nesting. I'm not sure if that discussion is completely within the scope of this PR however - would you both be OK if I resolve that comment for the time being?

bkamins:
@nalimilan - I am OK to leave out nesting. The only question is if we can later introduce it without breaking things (I do not know this functionality of dplyr in detail).

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+function _expandhelper(col, vect, stride, offset)
+    for i in offset:stride:length(vect)
+        push!(col, vect[i])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols)
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    allunique(indexcols) || throw(ArgumentError("Elements of $indexcols must be unique"))
+    sum(DataFrames.nonunique(df[:,indexcols]))>0 && @warn "duplicate rows in input; expand will only return unique combinations"
```
```suggestion
    length(nonunique(df, indexcols))>0 && @warn "duplicate rows in input; expand will only return unique combinations"
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+    colind = index(df)[indexcols]
+
+    dummydf = similar(select(df, indexcols, copycols=false),0)
```
```suggestion
    dummydf = similar(select(df, indexcols, copycols=false), 0)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+
+function _expandhelper(col, vect, stride, offset)
+    for i in offset:stride:length(vect)
+        push!(col, vect[i])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols)
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    allunique(indexcols) || throw(ArgumentError("Elements of $indexcols must be unique"))
```
```suggestion
    if !(indexcols isa Symbol || allunique(indexcols))
        throw(ArgumentError("Elements of $indexcols must be unique"))
    end
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+    sum(DataFrames.nonunique(df[:,indexcols]))>0 && @warn "duplicate rows in input; expand will only return unique combinations"
+    colind = index(df)[indexcols]
+
+    dummydf = similar(select(df, indexcols, copycols=false),0)
+
+    # Create a vect of vectors of unique values in each column
+    uniqueVals = []
+    for col in colind
+        # levels drops missing, handle the case where missing values are present
+        if sum(ismissing.(df[col])) != 0
```
```suggestion
        if any(ismissing, df[!, col])
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+    allunique(indexcols) || throw(ArgumentError("Elements of $indexcols must be unique"))
+    sum(DataFrames.nonunique(df[:,indexcols]))>0 && @warn "duplicate rows in input; expand will only return unique combinations"
+    colind = index(df)[indexcols]
+
+    dummydf = similar(select(df, indexcols, copycols=false),0)
+
+    # Create a vect of vectors of unique values in each column
+    uniqueVals = []
+    for col in colind
+        # levels drops missing, handle the case where missing values are present
```
```suggestion
        # levels drops missing, handle the case where missing values are present
        tempcol = levels(df[!, col])
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+        if sum(ismissing.(df[col])) != 0
+            tempcol = vcat(levels(df[col]), missing)
+        else
+            tempcol = levels(df[col])
+        end
+        push!(uniqueVals, tempcol)
+    end
+
+    # Get a long vector of every possible combination
+    collected = collect(flatten(product(values(uniqueVals)...)))
```
```suggestion
    collected = collect(Iterators.flatten(Iterators.product(uniqueVals...)))
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+    dfsize = div(length(collected),length(indexcols))
```
bkamins: where do you use this value?

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+
+    for i in 1:length(indexcols)
+        _expandhelper(dummydf[!,i], collected, length(indexcols), i)
```
```suggestion
        _expandhelper(dummydf[!, i], collected, length(indexcols), i)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+##
+##############################################################################
+
+function _expandhelper(col, vect, stride, offset)
+    for i in offset:stride:length(vect)
+        push!(col, vect[i])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols)
```
bkamins: @nalimilan - I am not sure here if `expand` and `complete` are the best names. What do you think. Maybe better keep the `df` suffix?

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+    dfsize = div(length(collected),length(indexcols))
+
+    for i in 1:length(indexcols)
+        _expandhelper(dummydf[!,i], collected, length(indexcols), i)
+    end
+
+    return dummydf
+end
+
+function complete(df::AbstractDataFrame, indexcols::Array{Symbol,1}; fill=missing::Any, replaceallmissing=false)
```
bkamins: `indexcols` probably should not have type restriction at all or please give a standard restriction `Union{ColumnIndex, AbstractVector{Integer}, AbstractVector{Symbol}, Not, Regex}`.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    allunique(indexcols) || throw(ArgumentError("Elements of $indexcols must be unique"))
```
bkamins: similar comments as above

src/dataframe/dataframe.jl
bkamins: also `replacemissing` should be restricted to `Bool`, and `fill` does not need `::Any`

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+    colind = index(df)[indexcols]
+
+    # Expand the input df and left join
+    expanded = expand(df, indexcols)
+    expanded = join(expanded, df; on = indexcols, kind = :left, indicator = :source)
+
+    # Replace missing values with the fill
+    if !ismissing(fill)
+        for n in names(expanded)
+            if (n ∉ indexcols) && (n != :source)
```
bkamins: the `:source` test is incorrect as the name might be duplicated. Better use the last column test (and then you can iterate not names but indices of a `DataFrame`).

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+                if replaceallmissing == true
+                    expanded[n] = coalesce.(expanded[n],fill)
```
```suggestion
                    expanded[!, n] = coalesce.(expanded[!, n], fill)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+                else
+                    #expanded[n] .= ifelse.(expanded.source .== "left_only", fill, expanded[n])
```
bkamins: why this omment is needed?

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+
+    # Expand the input df and left join
+    expanded = expand(df, indexcols)
+    expanded = join(expanded, df; on = indexcols, kind = :left, indicator = :source)
+
+    # Replace missing values with the fill
+    if !ismissing(fill)
+        for n in names(expanded)
+            if (n ∉ indexcols) && (n != :source)
+                if replaceallmissing == true
```
```suggestion
                if replaceallmissing
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+                    expanded[n] = coalesce.(expanded[n],fill)
+                else
+                    #expanded[n] .= ifelse.(expanded.source .== "left_only", fill, expanded[n])
+                    let col = expanded[n]
+                        for (i,v) in enumerate(expanded.source)
```
bkamins: `:source` might be an incorrect column name

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+
+    # Replace missing values with the fill
+    if !ismissing(fill)
+        for n in names(expanded)
+            if (n ∉ indexcols) && (n != :source)
+                if replaceallmissing == true
+                    expanded[n] = coalesce.(expanded[n],fill)
+                else
+                    #expanded[n] .= ifelse.(expanded.source .== "left_only", fill, expanded[n])
+                    let col = expanded[n]
```
bkamins: inside a function `let` does not help - you need a helper function to benefit from barrier function approach as above. `let` is a trick for global scope.

bkamins: also `expanded[!, n]`

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,77 @@ end
...
+                        for (i,v) in enumerate(expanded.source)
+                            v == "left_only" && (col[i] = fill)
+                        end
+                    end
+                end
+            end
+        end
+    end
+    select!(expanded, Not(:source))
```
bkamins: `:source` might be an incorrect name

src/dataframe/dataframe.jl
versipellis: I think using `sum(nonunique(df, indexcols))>0` is the correct test here as `length(nonunique(df, indexcols))>0` always returns true - the returned value of `nonunique(df, indexcols)` is an array of Bools that will be >0.

bkamins: Ah right (I mistakenly thought we return the actual indices).

src/dataframe/dataframe.jl
versipellis: That's a good point! Would using `names(df)[end]` be an appropriate solution to this?

versipellis: Tested it with a df with a column `:source` and it appears to work.

src/dataframe/dataframe.jl
versipellis: Tried my hand at barrier-izing it after reviewing the julia performance tips docs you linked me before. Let me know if I did it right and can resolve this!

src/dataframe/dataframe.jl
bkamins: yes `end` is OK, as the indicator column is always added at the end.

src/dataframe/dataframe.jl
bkamins: OK - I will go through the whole PR and let you know if I notice something.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+        push!(col, vect[i])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols)
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    if !(indexcols isa Symbol || allunique(indexcols))
+        throw(ArgumentError("Elements of $indexcols must be unique"))
+    end
+    sum(nonunique(df, indexcols))>0 && @warn "duplicate rows in input; expand will only return unique combinations"
```
bkamins: Now that I look at it we could use `row_group_slots` instead of `nonunique` to squeeze out more performance (it is enough to check how many groups were created).

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+    colind = index(df)[indexcols]
+    dummydf = similar(select(df, colind, copycols=false), 0)
+
+    # Create a vect of vectors of unique values in each column
+    uniqueVals = []
+    for col in colind
+        # levels drops missing, handle the case where missing values are present
+        if any(ismissing, df[!, col])
+            tempcol = vcat(levels(df[col]), missing)
```
```suggestion
            tempcol = vcat(levels(df[!, col]), missing)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+        else
+            tempcol = levels(df[col])
```
```suggestion
            tempcol = levels(df[!, col])
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+        end
+        push!(uniqueVals, tempcol)
+    end
+
+    # Get a long vector of every possible combination
+    collected = collect(Iterators.flatten(Iterators.product(uniqueVals...)))
```
bkamins: Now I have run your code. It will be much faster not to `Iterators.flatten` this iterator but rather to something like:

```
p = Iterators.product(uniqueVals...)
for i in 1:length(indexcols)
    _expandhelper(dummydf[!, i], p, i)
end
```

and define `_expandhelper` as:

```
function _expandhelper(col, prod, idx)
    for v in prod
        push!(col, v[i])
    end
end
```
the reason is that we do not create intermediate container with abstract eltype.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+    col = expandeddf[!, colname]
+    for (i,v) in enumerate(expandeddf[!,names(expandeddf)[end]])
+        v == "left_only" && (col[i] = fill)
+    end
+end
+
+function complete(df::AbstractDataFrame, indexcols; fill=missing, replaceallmissing=false::Bool)
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    allunique(indexcols) || throw(ArgumentError("Elements of $indexcols must be unique"))
+    sum(nonunique(df, indexcols))>0 && throw(ArgumentError("duplicate rows in input"))
```
bkamins: we call `expand` anyway so maybe we can avoid making these checks and throw an error in `expand` if something is wrong? Then maybe in `expand` we should add a kwarg if we want to have an error on duplicates (set to `true` by default)

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+            if (n ∉ indexcols) && (n != names(expanded)[end])
+                if replaceallmissing
+                    expanded[!, n] = coalesce.(expanded[!, n], fill)
+                else
+                    _completehelper(expanded, n, fill)
+                end
+            end
+        end
+    end
+    select!(expanded, Not(names(expanded)[end]))
```
```suggestion
    select!(expanded, 1:ncol(expanded)-1)
```
should be a bit faster

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+
+function complete(df::AbstractDataFrame, indexcols; fill=missing, replaceallmissing=false::Bool)
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    allunique(indexcols) || throw(ArgumentError("Elements of $indexcols must be unique"))
+    sum(nonunique(df, indexcols))>0 && throw(ArgumentError("duplicate rows in input"))
+    colind = index(df)[indexcols]
+
+    # Expand the input df and left join
+    expanded = expand(df, indexcols)
+    expanded = join(expanded, df; on = indexcols, kind = :left, indicator = :source)
```
bkamins: I would create here another binding, not rebind `expanded`. The reason is that you can use this fact below.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+
+    # Replace missing values with the fill
+    if !ismissing(fill)
+        for n in names(expanded)
```
bkamins: join will first add columns from `expanded` and only next from `df`, so it is enough to iterate columns added, not all columns.

Also note that this will **not** replace `missing` values in the `expaned`-part of the data frame, but I thing this is what we want (i.e. we should replace `missing`s in the joined columns from `df` but not in `expaned` where `missing` is actually a key - if you replaced `missing` with something else - you could - in particular - introduce duplicate rows in the `expanded` subpart of the data frame as the replacement value could be the same as key value actually present in that column)

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+            if (n ∉ indexcols) && (n != names(expanded)[end])
```
bkamins: then you do not need to do this check, but simply start iterating over added columns skipping the last column (which we know is artificially added).

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+                if replaceallmissing
+                    expanded[!, n] = coalesce.(expanded[!, n], fill)
```
```suggestion
                    expanded[!, n] = coalesce.(expanded[!, n], Ref(fill))
```

you have to protect `fill` against being broadcasted over.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,79 @@ end
...
+                else
+                    _completehelper(expanded, n, fill)
```
bkamins: You do not need this helper function (your code unfortunately was incorrect as `fill` might have type that is not subtype of `eltype` of `expanded[!, n]` and then it would break), better write:
```suggestion
                    expanded[!, i] .= ifelse.(expanded[!, end] .== "left_only", Ref(fill), expanded[!, i])
```
which will be already fast, as broadcasting introduces function barrier automatically and broadcasting will only allocate once (as in the case above before `else`).

bkamins:
@versipellis - sorry for a large number of comments, but in library code we want to make sure that it is exemplary (users then later often read it to learn how DataFrames.jl should be used). This means that also I, when read the code, sometimes change my mind when I notice that something can be done in a better way.

versipellis:
No problems @bkamins! This is my first PR to an established package, so I'm learning a lot through your comments as I go along, especially with regards to performant code :)

src/dataframe/dataframe.jl
versipellis: Aaah, gotcha. Didn't realize broadcasting did that already, thank you!

src/dataframe/dataframe.jl
versipellis: Something like this?
`row_group_slots(ntuple(i -> df[!, i], ncol(df)))[1]` and comparing that to the number of rows of the input `df`?

src/dataframe/dataframe.jl
versipellis: Just tested this and this indeed ended up being faster. Was the intermediate container that you were referring to the flattened Iterator?

src/dataframe/dataframe.jl
versipellis: This makes a lot of sense. Decent amount of code optimization happened from this, thank you!

versipellis:
@bkamins Implemented the latest round of changes and ran tests locally, seems to be fine on my end.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+##
+##############################################################################
+
+function _expandhelper(col, iterprod, idx)
+    for v in iterprod
+        push!(col, v[idx])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols; error=true::Bool)
```
```suggestion
function expand(df::AbstractDataFrame, indexcols; error::Bool=true)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+        throw(ArgumentError("Elements of $indexcols must be unique"))
+    end
+    if error == true
+        row_group_slots(ntuple(i -> df[!, i], ncol(df)))[1]!=nrow(df) && throw(ArgumentError("duplicate rows in input"))
+    else
+        row_group_slots(ntuple(i -> df[!, i], ncol(df)))[1]!=nrow(df) && @warn "duplicate rows in input; expand will only return unique combinations"
+    end
+
+    colind = index(df)[indexcols]
+    dummydf = similar(select(df, colind, copycols=false), 0)
```
bkamins: I think you can move these two lines to the top, and then you do need to do the first check `!(indexcols isa Symbol || allunique(indexcols))`, as they will error anyway on duplicates.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+        push!(col, v[idx])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols; error=true::Bool)
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    if !(indexcols isa Symbol || allunique(indexcols))
+        throw(ArgumentError("Elements of $indexcols must be unique"))
+    end
+    if error == true
```
bkamins: I think we have a problem if we have duplicates in columns selected to `dummydf` only (not a whole `df`, as e.g. we can have duplicates in selected columns, but there are no duplicates in the whole data frame `df`, as irrelevant columns might lead to uniqueness).

So we should pass to `row_group_slots` only columns that are in `colind` variable not all columns of `df`.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+        else
+            tempcol = levels(df[!, col])
+        end
+        push!(uniqueVals, tempcol)
+    end
+
+    # Get a long vector of every possible combination
+    p = Iterators.product(uniqueVals...)
+
+    for i in 1:length(indexcols)
```
```suggestion
    for i in axes(colind, 1)
```

bkamins: or `axes(dummydf, 2)`

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+        _expandhelper(dummydf[!, i], p, i)
+    end
+
+    return dummydf
+end
+
+function complete(df::AbstractDataFrame, indexcols; fill=missing, replaceallmissing=false::Bool)
```
```suggestion
function complete(df::AbstractDataFrame, indexcols; fill=missing, replaceallmissing::Bool=false)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+    colind = index(df)[indexcols]
+
+    # Expand the input df and left join
+    expanded = expand(df, indexcols)
```
```suggestion
    expanded = expand(df, colind)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+    joined = join(expanded, df; on = indexcols, kind = :left, indicator = :source)
```
```suggestion
    joined = join(expanded, df; on = _names(df)[colind], kind = :left, indicator = :source)
```

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,73 @@ end
...
+
+function expand(df::AbstractDataFrame, indexcols; error=true::Bool)
+    # Check to make sure the symbols in indexcols are in the df, and check for duplicate rows in the input df
+    if !(indexcols isa Symbol || allunique(indexcols))
+        throw(ArgumentError("Elements of $indexcols must be unique"))
+    end
+    if error == true
+        row_group_slots(ntuple(i -> df[!, i], ncol(df)))[1]!=nrow(df) && throw(ArgumentError("duplicate rows in input"))
+    else
+        row_group_slots(ntuple(i -> df[!, i], ncol(df)))[1]!=nrow(df) && @warn "duplicate rows in input; expand will only return unique combinations"
```
bkamins: @nalimilan - can you please double check that `row_group_slots` is used correctly, as you know this part of code better.

src/dataframe/dataframe.jl
versipellis: That's a very good point I missed in my original intent to allow for greater usability in `complete`. It does seem like a bit more of a challenge to have it work in `complete`, but I might tackle this in a future PR.

versipellis:
Pushed fixes - just waiting on @nalimilan for the remainder of the open requested changes above.

src/dataframe/dataframe.jl
```diff
@@ -1317,3 +1317,69 @@ end
...
+    end
+
+    # Create a vect of vectors of unique values in each column
+    uniqueVals = []
+    for col in colind
+        # levels drops missing, handle the case where missing values are present
+        if any(ismissing, df[!, col])
```
bkamins: now I have noticed we have the following inconsistency: you use `levels` instead of `unique`, this means that you return all **potential** levels in a column (even if they are not present). On the other hand we add `missing` only if it is actually present (not if it is potentially present).

Is this your intention?

src/dataframe/dataframe.jl
bkamins: I think this is actually not resolved. Why do you think this is a good behavior?
```
julia> df = DataFrame(x=1:5, y=1,z=1)
5×3 DataFrame
│ Row │ x     │ y     │ z     │
│ 1   │ 1     │ 1     │ 1     │
│ 2   │ 2     │ 1     │ 1     │
│ 3   │ 3     │ 1     │ 1     │
│ 4   │ 4     │ 1     │ 1     │
│ 5   │ 5     │ 1     │ 1     │

julia> expand(df, Not(:x))
1×2 DataFrame
│ Row │ y     │ z     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │

julia> expand(select(df, Not(:x)), :)
ERROR: ArgumentError: duplicate rows in input
```
I find it inconsistent. Given this maybe we should ignore duplicates and simply document that we ignore them (user can always deduplicate a data frame if one wants). This would simplify the code. And - it seems (we would have to write tests for this - best randomized) - that it is not problematic in `complete` - if we have duplicates simply we get duplicates in the output (essentially what `complete would do is add some rows with `missing` where the combinations were missing and retain the rest of the table as it was).

src/dataframe/dataframe.jl
nalimilan: We don't use the "df" suffix anywhere (except in `stackdf` that we want to remove), and that's not really Julian since that hardcodes the type in the function name, so I prefer `expand` (or another name). We could define it in DataAPI at some point. Regarding `complete`, I suggested above that it could be replaced with a keyword argument to `expand` to limit the number of exported functions. What do you think?

src/dataframe/dataframe.jl
nalimilan: `row_group_slots(ntuple(i -> df[!, i], ncol(df)), Val(false))[1]` will be even more efficient for pooled column types.

src/dataframe/dataframe.jl
nalimilan: Sounds correct. As I noted above, passing `Val(false)` as the second argument can make thing faster in some cases, so better do that.

Also, I'd make the `row_group_slot` call the top-level branch, and put `if error` inside it: that would avoid repeating the call. And while I'm at it, some spacing around `!=` wouldn't hurt.

src/dataframe/dataframe.jl
bkamins: I am OK with your proposal (and we can move `expand` to DataAPI.jl later when it is needed).

src/dataframe/dataframe.jl
bkamins: Given my later comment I think we can even avoid this - we really do not need to check for duplicates (they do no harm and if they are present `join` will expand `dummydf` as needed). If someone wants to avoid duplicates one can do it before calling `expand`.

nalimilan: OK, I was just reacting to the question.

bkamins: (sure - I also wanted to learn the optimal pattern)

src/dataframe/dataframe.jl
versipellis: It appears this is an issue that cropped up as a result of switching the duplicate rows check from using `nonunique` to using `row_group_slots`. Using the above example:
```
julia> sum(nonunique(df,[:y,:z]))>0
true
julia> row_group_slots(ntuple(i -> df[!, i], ncol(df)))[1]!=nrow(df)
false
```

Let me take a look at the implementation of `row_group_slots` in `nonunique` again, I appear to have missed something.

src/dataframe/dataframe.jl
versipellis: It was not. Honestly speaking, I don't work with `levels` a ton, so this is just a hole in my experience/knowledge showing that I need to address :)

src/dataframe/dataframe.jl
versipellis: Oh, I found the issue. `df` wasn't subsetted by `indexcols`. Testing now.

versipellis:
```
    if row_group_slots(ntuple(i -> select(df,colind)[!, i], ncol(select(df,colind))), Val(false))[1] != nrow(select(df,colind))
        if error == true
            throw(ArgumentError("duplicate rows in input"))
        else
            @warn "duplicate rows in input; expand will only return unique combinations"
        end
    end
```
seems to fix it.

```
julia> expand(df, Not(:x))
ERROR: ArgumentError: duplicate rows in input
julia> expand(df, Not(:x), error=false)
┌ Warning: duplicate rows in input; expand will only return unique combinations
└ @ DataFrames ~/.julia/dev/DataFrames/src/dataframe/dataframe.jl:1341
1×2 DataFrames.DataFrame
│ Row │ y     │ z     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │
```

src/dataframe/dataframe.jl
versipellis: @nalimilan I'd love your thoughts on this. I'd like to still be able to keep the order of categorical variables by using `levels` instead of reverting back to `unique`. The way I'm thinking is to check if the column is categorical, and if it is, calling `droplevels!` on it to ensure only categories that are in the subsetted dataset are expanded on. Is there a better approach to this?

bkamins: @nalimilan - I would keep using `levels` to retain the order of categorical variables, as @versipellis rightly commented. We have three options:
1. current functionality (retain all levels and add `missing` level only if it is present);
2. wide functionality (retain all levels and add `missing` level if it is potentially present in a categorical variable - we can detect this by examining `eltype` of column if it is categorical);
3. narrow functionality (retain only levels actually present in the column remembering their order);

Actually thinking about it I am leaning towards option 1. (current functionality) - but we just should document that we work this way. The reason is that `missing` is not a new level but rather indicates absence of information about the level from the list of of possible levels (I guess this was also your intention in CategoricalArrays.jl - right?).

src/dataframe/dataframe.jl
bkamins: As commented above - maybe we should ignore duplicates and just document that we ignore them. This will simplify the code and deduplicating before calling these functions is simple enough.

bkamins:
@versipellis - I think that the crucial change (apart from what we discuss above) is what @nalimilan suggested - i.e. merging these two functions into one and just adding a kwarg that will indicate what output we actually want.

src/dataframe/dataframe.jl
nalimilan: 1 sounds fine. I could be convinced to take approach 3 if other implementations do that.

bkamins: OK - so let us go for 1 if @versipellis also agrees

versipellis: 1 works for me as well. I'll make a note to document it as such.

versipellis:
@bkamins @nalimilan my original intention in keeping both functions separate as that's how it's done in other languages. I'm not opposed to merging them into one function, but I wasn't sure how much similarity Queryverse seeks to keep to similar packages in other languages.

bkamins:
My major concern was that that both `complete` and `expand` are potentially common names, so I preferred to use-up only one. It is also cleaner to have one function that does two very similar things (especially that this operation is not that commonly needed).

But this is only a slight preference. I would be OK also with two functions if we ended up this way.

nalimilan:
AFAICT `complete` only exists in dplyr, right? Like @bkamins, I'd rather only add `expand` to avoid exporting too many common names.

bkamins:
@versipellis - I would love to see this PR polished and merged before release 1.0 (which we will have this year probably). How do you see this?

versipellis:
@bkamins I'm just getting back to things after a long period of being away, let me take a look at this again tomorrow and see how I can wrap things up :)

versipellis:
OK. Pushing now. I've combined both together.

```julia
julia> df = DataFrames.DataFrame(A=["05","05","05","06","06","07","07",missing],
                       B=["a",missing,"c","a",missing,"c","d","a"],
                       source=[1,1,10,missing,5,10,3,missing])
8×3 DataFrames.DataFrame
│ Row │ A       │ B       │ source  │
│     │ String⍰ │ String⍰ │ Int64⍰  │
│ 1   │ 05      │ a       │ 1       │
│ 2   │ 05      │ missing │ 1       │
│ 3   │ 05      │ c       │ 10      │
│ 4   │ 06      │ a       │ missing │
│ 5   │ 06      │ missing │ 5       │
│ 6   │ 07      │ c       │ 10      │
│ 7   │ 07      │ d       │ 3       │
│ 8   │ missing │ a       │ missing │
julia> indexcols = [:A,:B]
2-element Array{Symbol,1}:
 :A
 :B
julia> DataFrames.expand(df,indexcols)
16×2 DataFrames.DataFrame
│ Row │ A       │ B       │
│     │ String⍰ │ String⍰ │
├─────┼─────────┼─────────┤
│ 1   │ 05      │ a       │
│ 2   │ 06      │ a       │
│ 3   │ 07      │ a       │
│ 4   │ missing │ a       │
│ 5   │ 05      │ c       │
│ 6   │ 06      │ c       │
│ 7   │ 07      │ c       │
│ 8   │ missing │ c       │
│ 9   │ 05      │ d       │
│ 10  │ 06      │ d       │
│ 11  │ 07      │ d       │
│ 12  │ missing │ d       │
│ 13  │ 05      │ missing │
│ 14  │ 06      │ missing │
│ 15  │ 07      │ missing │
│ 16  │ missing │ missing │
julia> DataFrames.expand(df,indexcols,complete=true,fill=10000)
16×3 DataFrames.DataFrame
│ Row │ A       │ B       │ C     │
│     │ String⍰ │ String⍰ │ Int64 │
├─────┼─────────┼─────────┼───────┤
│ 1   │ 05      │ a       │ 1     │
│ 2   │ 06      │ a       │ 43    │
│ 3   │ 07      │ a       │ 10000 │
│ 4   │ missing │ a       │ 10    │
│ 5   │ 05      │ c       │ 10    │
│ 6   │ 06      │ c       │ 10000 │
│ 7   │ 07      │ c       │ 10    │
│ 8   │ missing │ c       │ 10000 │
│ 9   │ 05      │ d       │ 10000 │
│ 10  │ 06      │ d       │ 10000 │
│ 11  │ 07      │ d       │ 3     │
│ 12  │ missing │ d       │ 10000 │
│ 13  │ 05      │ missing │ 1     │
│ 14  │ 06      │ missing │ 5     │
│ 15  │ 07      │ missing │ 10000 │
│ 16  │ missing │ missing │ 10000 │
```

versipellis:
@bkamins If this is acceptable to you, should I start working on the documentation for it?

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+##
+##############################################################################
+
+function _expandhelper(col, iterprod, idx)
+    for v in iterprod
+        push!(col, v[idx])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols; error::Bool=true, complete::Bool=false, fill=missing, replaceallmissing::Bool=false)
```
bkamins: can you please break this line into two as it is now overlong?

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+        push!(uniqueVals, tempcol)
+    end
+
+    # Get a long vector of every possible combination
+    p = Iterators.product(uniqueVals...)
+
+    for i in axes(colind, 1)
+        _expandhelper(dummydf[!, i], p, i)
+    end
+    if complete == false
```
bkamins: maybe just `if !complete`

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+        push!(col, v[idx])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols; error::Bool=true, complete::Bool=false, fill=missing, replaceallmissing::Bool=false)
+    colind = index(df)[indexcols]
+    dummydf = similar(select(df, colind, copycols=false), 0)
+
+    if row_group_slots(ntuple(i -> df[!,colind][!, i], ncol(df[!,colind])), Val(false))[1] != nrow(df[!,colind])
+        if error == true
```
bkamins: I think it should be:
```
if error && !complete
```
because if `complete == true` then duplicate rows are not a problem (they will not be dropped as `join` will reintroduce them).

bkamins:
Looks good (unless @nalimilan has some more comments). Thank you! I left some minor comments.

Yes - now it would be great to add documentation and tests (please try to make sure to cover all possible combinations of usage - in particular concerning: duplicates, different ways of handling `missings`, and categorical vectors).

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+    for v in iterprod
+        push!(col, v[idx])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols; error::Bool=true, complete::Bool=false, fill=missing, replaceallmissing::Bool=false)
+    colind = index(df)[indexcols]
+    dummydf = similar(select(df, colind, copycols=false), 0)
+
+    if row_group_slots(ntuple(i -> df[!,colind][!, i], ncol(df[!,colind])), Val(false))[1] != nrow(df[!,colind])
```
```suggestion
    if row_group_slots(ntuple(i -> df[!, colind][!, i], ncol(df[!, colind])), Val(false))[1] != nrow(df[!, colind])
```

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
 end
+
+##############################################################################
+##
+## Expand & Complete dataframes
+##
+##############################################################################
+
+function _expandhelper(col, iterprod, idx)
+    for v in iterprod
```
```suggestion
    @inbounds for v in iterprod
```

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+    end
+    if complete == false
+        return dummydf
+    else
+        joined = join(dummydf, df; on = _names(df)[colind], kind = :left, indicator = :source)
+
+        # Replace missing values with the fill
+        if !ismissing(fill)
+            for n in ncol(dummydf)+1:ncol(joined)-1
+                if replaceallmissing
```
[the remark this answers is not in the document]
bkamins: I think `replacemissing` would be better.

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
 function permutecols!(df::DataFrame, p::AbstractVector{Symbol})
     permutecols!(df, index(df)[p])
 end
+
+##############################################################################
+##
+## Expand & Complete dataframes
```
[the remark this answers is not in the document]
versipellis: Any suggestions on where it should live instead?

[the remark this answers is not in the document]
bkamins: I would put it in abstractdataframe/abstractdataframe.jl.
Also please update **common methods** section of `AbstractDataFrame` docstring.

When testing please make sure to make tests against `SubDataFrame` also.

versipellis:
In the process of writing the docstring.

I've never written tests before - I'm going through the Julia documentation for them now, is there any JuliaData-specific docs for writing tests I should also be looking at?

bkamins:
Just look at tests that we have in DataFrames.jl.
Mostly it should be enough if you use `@testset`, `@test` and `@test_throws`. Try to wrap everything within `@testset` so that the tests you write do not affect the global state of the test file.

bkamins:
@versipellis - what is your status with this PR? (I am planning to release 0.20.0 soon)

versipellis:
I've been absolutely slammed by real life in the last couple of months - sorry for falling off the grid. Let me take a look at this over the weekend and see how fast I can roll this out, I've done most of the docs work but haven't done too much with the tests yet.

bkamins:
Sure - no rush. In the worst case it will go int in 0.21 release. Please note that because the PR was stalled for some time it requires rebasing against master.

bkamins:
@versipellis - I wanted to check if you would want to work on this PR to try finalize it. If you do not have time for it I can take over your proposal and try to finalize it as a new PR (as I think we need the functionality you proposed 😄). Thank you!

bkamins:
Adding 2.0 milestone as for sure we want this functionality, but since it is not breaking it can be implemented later.

bkamins:
@versipellis - can you please comment if you would be still willing to work on this PR? If I do not get response from you I will open another PR adding the functionality to which I will write rights to finalize it (in such a case I will add a comment in the code that the initial idea for the functionality was from this PR).

versipellis:
Hi @bkamins, my apologies, real life has been absolutely a nightmare. I'll be working on this PR in the coming week or two.

I'm not very good with writing tests, however, which is where I've mostly been stuck with this up till now. I had been planning to write tests to make sure each of the args works, but that's all I got to. Do you have anything else you'd like to see covered in the testing?

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+        push!(col, v[idx])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols; error::Bool=true, complete::Bool=false, fill=missing, replaceallmissing::Bool=false)
+    colind = index(df)[indexcols]
+    dummydf = similar(select(df, colind, copycols=false), 0)
+
+    if row_group_slots(ntuple(i -> df[!, colind][!, i], ncol(df[!, colind])), Val(false))[1] != nrow(df[!, colind])
+        if error == true
```
```suggestion
        if error
```

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+        push!(uniqueVals, tempcol)
+    end
+
+    # Get a long vector of every possible combination
+    p = Iterators.product(uniqueVals...)
+
+    for i in axes(colind, 1)
+        _expandhelper(dummydf[!, i], p, i)
+    end
+    if complete == false
```
```suggestion
    if !complete
```

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+        return dummydf
+    else
+        joined = join(dummydf, df; on=_names(df)[colind], kind=:left, indicator=:source)
```
bkamins: now it should be `leftjoin`

src/dataframe/dataframe.jl
```diff
@@ -1410,3 +1410,67 @@ end
...
+##
+##############################################################################
+
+function _expandhelper(col, iterprod, idx)
+    @inbounds for v in iterprod
+        push!(col, v[idx])
+    end
+end
+
+function expand(df::AbstractDataFrame, indexcols; error::Bool=true, complete::Bool=false, fill=missing, replaceallmissing::Bool=false)
```
bkamins: Why do you think we need `replaceallmissing` kwarg? I think that `fill` is enough. If later someone wants to replace the remaining missing values that were originally in the data frame in non-indexcols then it is easily done.

bkamins:
Excellent - thank you. This time frame is OK (as you probably know we are pushing to finalize DataFrames.jl to 1.0 release and that is why I was asking about this feature).

So the steps I would recommend you to take are:
1. do not do anything until you have time to work on PR (we are actively adding new things now, so this way you will avoid rebasing the PR several times)
2. rebase the PR (Resolve conflicts button above)
3. address all the unresolved comments
4. add docstring
4. Write tests that will cover a Cartesian product of the following things:
    * data frame type: `DataFrame`, `SubDataFrame`
    * number of rows: `0`, `1`, many
    * `idxcols`: many columns, one column, zero columns (also testing for the case when what `idxcols` slects is not present in the data frame)
    * all combinations of kwarg values (for `fill` use two values: the default and some other value to check it is probably used to fill the columns)

bkamins:
> rebase the PR

Hi @versipellis - I recommended to rebase the PR, you have merged branches. Now it is impossible to make a review (you have added 8,500 lines and removed 5,000). I think that in the current state of this PR it is best if you just take the changes you have made, rebase it to master, add only one commit with changes and force push.

versipellis:
> > rebase the PR
> 
> Hi @versipellis - I recommended to rebase the PR, you have merged branches. Now it is impossible to make a review (you have added 8,500 lines and removed 5,000). I think that in the current state of this PR it is best if you just take the changes you have made, rebase it to master, add only one commit with changes and force push.

Sorry about that - I had switched laptops in the middle of the covid-19 stuff and working on this, and my own local repos were a giant mess; rebase was doing something funky. I'll do that.

bkamins:
That is what I assumed. That is why I recommend to use "force" in git - to prune old stuff and just make one commit on top of current master. Ideally - with the changes recommended in the comments (but they can be added later if that would be simpler for you.

Fortunately this should be simple as you just add a function (and not change code in many places).

bkamins:
I will implement this functionality in a separate PR for 1.4 release.

bkamins:
Closing in favor of https://github.com/JuliaData/DataFrames.jl/pull/3012

PR #1865, opened 2019-06-30
Fix combine with aggregation when skipmissing=true

nalimilan:
`gix == 0` for missing values when skipmissing=true, but the code was not designed to handle this. Due to `@inbounds`, it is likely to go unnoticed.

I've checked using benchmark at https://github.com/JuliaData/DataFrames.jl/pull/1651 that this doesn't affect performance (since unfortunately SIMD isn't enabled anyway in this loop).

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -591,7 +591,7 @@ function groupreduce!(res, f, op, condf, adjust,
     @inbounds for i in eachindex(incol, gd.groups)
         gix = gd.groups[i]
         x = incol[i]
-        if condf === nothing || condf(x)
+        if gix > 0 && (condf === nothing || condf(x))
             res[gix] = op(res[gix], f(x, gix))
             adjust !== nothing && (counts[gix] += 1)
         end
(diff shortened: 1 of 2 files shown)


nalimilan:
Actually I think results were generally correct: we just added values to index 0 in the results array, which probably corresponds to a meta-data area which isn't used (in my tests this doesn't corrupt the array). And since these values were supposed to be ignored anyway...

bkamins:
I have run your tests - and both ways of doing them produced different results. Anyhow - thank you for the fix.

bkamins:
Also - I thought we turned off `@inbounds` during tests?

nalimilan:
> I have run your tests - and both ways of doing them produced different results. Anyhow - thank you for the fix.

OK, then let's just hope nobody has used aggregations with `skipmissing=true`. :-)


> Also - I thought we turned off `@inbounds` during tests?

Yeah, but apparently tests didn't cover this particular combination of features... I guess coverage didn't catch that because there was no specific code path in the optimized methods I added -- so no specific test needed from its POV.

PR #1866, opened 2019-07-03
Deprecate df[col] and df[cols]` add `!` support

bkamins:
@nalimilan - please now only comment on specification in docs/src/lib/indexing.md (it is complex enough to check already unfortunately).

Other things are in flux (I fix deprecations where needed as I spot them, but this is not tested or finished).


docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+`getindex` on `DataFrameRow`:
+* `dfr[col]` -> the value contained in column `col` of `dfr`; the same as `dfr.col` is `col` is a valid identifier;
 * `dfr[cols]` -> a `DataFrameRow` with parent `parent(dfr)` if `cols` is a colon and `parent(dfr)[cols]` otherwise;
+
+`view` on `DataFrameRow`:
 * `@view dfr[col]` -> a `0`-dimensional view into `parent(dfr)[DataFrames.row(dfr), col]`;
 * `@view dfr[cols]` -> a `DataFrameRow` with parent `parent(dfr)` if `cols` is a colon and `parent(dfr)[cols]` otherwise;
 
+Note that when views are created with columns selector set to `:` then the view produced dynamically changes its columns'
+names and count if columns of parents are added/removed/renamed;
```
```suggestion
names and count if columns are added/removed/renamed in the parent;
```

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
 
-## `getindex`
+Note that `df.col` works like `df[!, col]` when getting column from an `AbstractDataFrame` or broadcasting into `df.col`,
+but it works like `df[:, col]` when used in setting column context. This behavior was chosen
+as this kind of behavior is typically expected as a reasonable default.
+
+## `getindex` and `view`
 
-The following list specifies return types of `getindex` operations depending on argument types.
+The following list specifies return types of `getindex` and `view` operations depending on argument types.
```
nalimilan: Not just return types: the behavior is fully specified. Same for `setindex!`.

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+Note that when views are created with columns selector set to `:` then the view produced dynamically changes its columns'
+names and count if columns of parents are added/removed/renamed;
+
 ## `setindex!`
 
-Under construction
+The following list specifies return types of `setindex!` operations depending on argument types.
+
+In particular a description explicitly mentions if the assignment is *in-place*,
+*replaces old vectors without copying source* or *replaces old vectors with copying source*.
```
```suggestion
*replaces old vectors without copying source* or *replaces old vectors copying source*.
```

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+
+`setindex!` on `DataFrame`:
+* `df[CartesianIndex(row, col)] = v` -> the same as `df[row, col] = v`;
+* `df[row, col] = v` -> set value of `col` in row `row` to `v` in-place;
+* `df[row, cols] = v` -> the same as `dfr = df[row, cols]; dfr[:] = v` in-place;
```
```suggestion
* `df[row, cols] = v` -> set row `row` of columns `cols` in-place; the same as `dfr = df[row, cols]; dfr[:] = v`;
```

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+* `df[rows, col] = v` -> set rows `rows` of column `col` in-place; `v` can be an abstract vector;
```
```suggestion
* `df[rows, col] = v` -> set rows `rows` of column `col` in-place; `v` must be an `AbstractVector`;
```

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+* `df[rows, cols]` -> set rows `rows` of columns `cols` in-place; `v` can be an `AbstractMatrix` or `v` can be `AbstractDataFrame`
```
```suggestion
* `df[rows, cols]` -> set rows `rows` of columns `cols` in-place; `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
```

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+                      (in this case column names must match);
+* `df[:, col] = v` -> the same as `df[1:nrow(df), col] = v` but replaces old vectors with a copy of source;
```
```suggestion
* `df[:, col] = v` -> similar to `df[1:nrow(df), col] = v` but replaces old vector with a copy of `v`;
```
Same below.

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+                      the same as `df.col = v` if `col` is a valid identifier;
+                      also if `col` is a `Symbol` that is not present in `df` then a new column in `df` is created;
```
```suggestion
                      if `col` is a `Symbol` that is not present in `df` then a new column is added to `df`;
```
Same below.

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+* `df[:, cols] = v` -> the same as `df[1:nrow(df), cols] = v` but replaces old vectors with a copy of source;
+* `df[!, col] = v` -> the same as `df[1:nrow(df), col] = v` but replaces old vectors without copying source;
+                      also if `col` is a `Symbol` that is not present in `df` then a new column in `df` is created without copying of source;
+* `df[!, cols] = v` -> the same as `df[1:nrow(df), cols] = v` but replaces old vectors without copying source if `v` is an `AbstractDataFrame`.
```
nalimilan: So this means no copy is made for `SubDataFrame`?

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
+
+`setindex!` on `SubDataFrame`:
+* `sdf[CartesianIndex(row, col)] = v` -> the same as `sdf[row, col] = v`;
+* `sdf[row, col]` -> set value of `col` in row `row` to `v` in-place;
+* `sdf[row, cols]` -> the same as `dfr = df[row, cols]; dfr[:] = v` in-place;
+* `sdf[rows, col]` -> set rows `rows` of column `col` in-place; `v` can be an abstract vector;
+* `sdf[rows, cols]` -> set rows `rows` of columns `cols` in-place;
+                       `v` can be an `AbstractMatrix` or `v` can be `AbstractDataFrame` when column names must match;
+* `sdf[:, col]` -> the same as `sdf[1:nrow(df), col] = v`; the same as `sdf.col = v` if `col` is a valid identifier;
```
nalimilan: These are somewhat inconsistent with what happens with `DataFrame`, where columns are replaced. Maybe we should require using `1:end` or `.=`?

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,146 @@ The rules for a valid type of index into a column are the following:
...
 The following broadcasting rules apply to `AbstractDataFrame` objects:
 * `AbstractDataFrame` behaves in broadcasting like a two-dimensional collection compatible with matrices.
 * If an `AbstractDataFrame` takes part in broadcasting then a `DataFrame` is always produced as a result.
-  In this case the requested broadcasting operation produce an object with exactly two dimensions.
+  In this case the requested broadcasting operation produces an object with exactly two dimensions.
   An exception is when an `AbstractDataFrame` is used only as a source of broadcast assignment into an object
   of dimensionality higher than two.
 * If multiple `AbstractDataFrame` objects take part in broadcasting then they have to have identical column names.
 
+Broadcasting `DataFrameRow` is currently not allowed.
```
```suggestion
Broadcasting `DataFrameRow` is currently not allowed (which is consistent with `NamedTuple`).
```

docs/src/lib/indexing.md
```diff
@@ -100,9 +169,21 @@ In such an operation `AbstractDataFrame` is considered as two-dimensional and `D
     The rule above means that, similar to single-dimensional objects in Base (e.g. vectors),
     `DataFrameRow` is considered to be column-oriented.
 
-If column indexing using `Symbol` names is performed the order of columns in the operation is specified
+Additional rules:
+* in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` and `df[row, cols] .= v` syntaxes the assignment to `df` is performed in-place;
+* in the `df[rows, col] .= v` and `df[rows, cols] .= v` syntaxes the assignment to `df` is performed in-place;
```
nalimilan: Why not merge this with the previous line?

docs/src/lib/indexing.md
```diff
@@ -100,9 +169,21 @@ In such an operation `AbstractDataFrame` is considered as two-dimensional and `D
     The rule above means that, similar to single-dimensional objects in Base (e.g. vectors),
     `DataFrameRow` is considered to be column-oriented.
 
-If column indexing using `Symbol` names is performed the order of columns in the operation is specified
+Additional rules:
```
nalimilan: This list is quite repetitive. How about listing the syntaxes which are in-place and those which aren't instead?

docs/src/lib/indexing.md
```diff
@@ -21,66 +21,111 @@ The rules for a valid type of index into a column are the following:
...
-* `df[rows, col]` -> a copy of the vector `df[col]` with only the entries corresponding to `rows` selected, the same as `df[col][rows]`;
-* `df[rows, cols]` -> a `DataFrame` containing copies of columns `cols` with only the entries corresponding to `rows` selected.
-* `@view df[col]` -> the vector contained in column `col` (this is equivalent to `df[col]`);
-* `@view df[cols]` -> a `SubDataFrame` with parent `df` if `cols` is a colon and `df[cols]` otherwise;
-* `@view df[row, col]` -> a `0`-dimensional view into `df[col]`, the same as `view(df[col], row)`;
-* `@view df[row, cols]` -> a `DataFrameRow` with parent `df` if `cols` is a colon and `df[cols]` otherwise;
-* `@view df[rows, col]` -> a view into `df[col]` with `rows` selected, the same as `view(df[col], rows)`;
-* `@view df[rows, cols]` -> a `SubDataFrame` with `rows` selected with parent `df` if `cols` is a colon and `df[cols]` otherwise.
+* `df[row, col]` -> the value contained in row `row` of column `col`, the same as `df[!, col][row]`;
+* `df[row, cols]` -> a `DataFrameRow` with parent `df` if `cols` is a colon and parent `df[!, cols]` otherwise;
```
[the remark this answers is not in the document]
bkamins: This is a delicate point. `df[!, :]` returns a copy of `df` and this is different because:
* it would be more expensive computationally (and we want to create `dfr` very cheaply)
* it would mean that when we mutate `df` (add/remove/rename columns) then `dfr` would not be mutated and it is mutated

docs/src/lib/indexing.md
bkamins: This rule is for `df`. For `sdf` there is a separate rule below `sdf[!, cos] = v` is an in-place operation.

docs/src/lib/indexing.md
bkamins: I had a problem what to do with these cases. The rules are inconsistent, but it is not possible to make them consistent (we can do only in-place operation).
I have added these rules because you can have code and then add `@view` in front of it later. Without these rules such an operation would break the code.

In order to be consistent we would have to drop special handling of `:` and `!` for `DataFrame`, but I understand we do not want this.

docs/src/lib/indexing.md
bkamins: I agree, but please leave it as is for now. I need it to write proper tests (and someone later checking the tests will need it to make sure that we have proper test coverage).

I propose to shorten the list after we are done with implementation and review of tests.

docs/src/lib/indexing.md
bkamins: The same reason as above - I need it to make sure I do not miss anything when writing tests. Later we can merge them

docs/src/lib/indexing.md
```diff
@@ -21,66 +21,111 @@ The rules for a valid type of index into a column are the following:
...
+* `@view df[rows, cols]` -> a `SubDataFrame` with `rows` selected with parent `df` if `cols` is a colon and `df[!, cols]` otherwise;
+* `@view df[:, col]` -> a view into a whole `df[!, col]`, the same as `view(df[!, col], :)`;
+* `@view df[:, cols]` -> a `SubDataFrame` with `rows` selected with parent `df` if `cols` is a colon and `df[!, cols]` otherwise;
+* `@view df[!, col]` -> the same as `view(df, :, col)`;
+* `@view df[!, cols]` -> the same as `view(df, :, cols)`.
 
 `SubDataFrame`:
-* `sdf[col]` -> a view of the vector contained in column `col` of `parent(sdf)` with `DataFrames.rows(sdf)` as a selector;
-* `sdf[cols]` -> a `SubDataFrame`, with parent `parent(sdf)` if `cols` is a colon and `parent(sdf)[cols]` otherwise;
+* `sdf[CartesianIndex(row, col)]` -> the same as `sdf[row,col]`;
```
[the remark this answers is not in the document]

docs/src/lib/indexing.md
```diff
@@ -21,66 +21,111 @@ The rules for a valid type of index into a column are the following:
...
 * `sdf[row, cols]` -> a `DataFrameRow` with parent `parent(sdf)` if `cols` is a colon and `parent(sdf)[cols]` otherwise;
-* `sdf[rows, col]` -> a copy of a vector `sdf[col]` with only rows `rows` selected;
-* `sdf[rows, cols]` -> a `DataFrame` containing columns `cols` and `df[rows, col]` as a vector in each `col` in `cols`.
-* `@view sdf[col]` -> a view of vector contained in column `col` of `parent(sdf)` with `DataFrames.rows(sdf)` as selector;
-* `@view sdf[cols]` -> a `SubDataFrame` with parent `parent(sdf)` if `cols` is a colon and `parent(sdf)[cols]` otherwise;
-* `@view sdf[row, col]` -> translates to `view(sdf[col], row)` (a `0`-dimensional view into `df[col]`);
-* `@view sdf[row, cols]` -> a `DataFrameRow` with parent `parent(sdf)` if `cols` is a colon and `parent(sdf)[cols]` otherwise;
-* `@view sdf[rows, col]` -> translates to `view(sdf[col], rows)` (a standard view into `sdf[col]` vector);
-* `@view sdf[rows, cols]` -> a `SubDataFrame` with parent `parent(sdf)` if `cols` is a colon and `sdf[cols]` otherwise.
+* `sdf[rows, col]` -> a subset of a vector `sdf[!, col]` with only rows `rows` selected;
```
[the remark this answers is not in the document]

docs/src/lib/indexing.md

docs/src/lib/indexing.md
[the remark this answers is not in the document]
nalimilan: Ah right `df[!, :]` doesn't copy vectors but it copies the data frame.

docs/src/lib/indexing.md
nalimilan: I'm referring to
> if `v` is an `AbstractDataFrame`.

It would be weird when `v` is a `SubDataFrame` to set columns of `df` to be `SubArray`s. OTOH maybe people shouldn't use `!` in that case. Not sure.

docs/src/lib/indexing.md
nalimilan: How about making these throw an error that points to `1:end` and `.=` for now? It sounds like a good idea to force users to use a different syntax from `DataFrame` here since the result is completely different. They should use a syntax that does in-place column setting both on `DataFrame` and `SubDataFrame`.

docs/src/lib/indexing.md
bkamins: I agree it is weird. But the same will happen with
```
foreach(col -> df[!, col] = sdf[!, col], axes(df, 2))
```
which is essentially the same operation (but without checking that the column names match).

The reason for this problem is that we allow to store in `DataFrame` any `AbstractVector` and not restrict it to `Vector`.

docs/src/lib/indexing.md
bkamins: OK - we can disallow it. We can always introduce it later if needed (I will update the docs for this case). In general setting whole columns of `SubDataFrame` is I guess a corner case that rarely happens in practice.

The only problem is that it covers `sdf.col = v` case which we should allow and it will be inconsistent (`sdf.col = v` is in place while `df.col = v` replaces with copy).

Also we still have an inconsistency that `df[:, col] .= v` allocates fresh columns while `sdf[:, col] .= v` is in place.

The problem is that in order to make this consistent we would have to make `:` in-place always. This is doable, but then we should have this `setcol!` function for column replacement and drop the whole `!` idea as it will be redundant then.

So the choice is:
1. use current design with `:` and `!` as it does what users expect most of the time (but either gives some inconsistency or is disallowed in some cases)
2. be strict with consistency and add `setcol!`

I think option 1. is better (we will still have inconsistency between `df.col = v` and `sdf.col = v` anyway if we go for option 2.). Then the question is if in option 1. we minimize inconsistency (by disallowing certain operations - your current proposal) or simply accept the inconsistency (this is the current state in the docs).

bkamins:
@nalimilan Following the discussion I have proposed a "minimal" version of the new syntax. It is much simpler, but requires introducing `setcol!` function (which is not that bad I think) and `fill` should be handled by CategoricalArrays.jl correctly (which anyway is desired).

In the proposal `df[!, col]` and `df.col` usage is reduced to the places where it makes a safe sense:
* `getindex` for `df` and `sdf`
* `setindex!` for `df`
* `df.col .= v` and `sdf.col .= v` in broadcasting (as this is handled by a separate mechanism outside control of DataFrames.jl)

In particular `!` syntax is disallowed with `@view` and disallowed in `setindex!` for `SubDataFrame` and in broadcasting.

This is the other extreme in comparison to the previous commit. Actually I like this version better as it is much simpler and consistent. The corner cases we have to handle with `setcol!` and `fill` should not be that problematic IMO.

bkamins:
@nalimilan - please check if all is described as we discussed. Thank you!

bkamins:
OK - it seems the proposed rules did not produce serious concerns. I start implementing them then.

bkamins:
This is good for a first look.

What is done:
* broadcasting (implementation and tests)
* `getindex` (implementation and tests)
* `setindex!` (deprecation)

What is not done:
* tests for `setindex!`
* rewrite of DataFrames.jl codebase to use new rules

Why it is good to have a look at it now:
* broadcasting and `getindex` required very low number of changes - this should be clean to check already
* old `setindex!` were massively inconsistent in corner cases; this means that mostly all that we have now goes to deprecated.jl; very little new functionality could be implemented as it would conflict with old functionality; I am not sure how to best approach it (maybe like in 0.7-1.0 transition in Julia - with the next release of DataFrames.jl closely following the previous one) - comments are welcome

bkamins:
The PR should be good for a review.

@nalimilan:
1. I am not sure how to approach the review as the changes required to the whole codebase both src and test are massive
2. I have one design question: if function `f` calls function `g` and `g` is deprecated I am not sure how to make `f` also throw a deprecation warning; we need this for `setindex!` of `SubDataFrame` (as it essentially passes down the work to deprecated `setindex!` methods of `DataFrame`)

src/dataframe/dataframe.jl
```diff
@@ -472,249 +463,67 @@ function upgrade_scalar(df::DataFrame, v::Any)
...
-                        v::Any,
-                        row_inds::AbstractVector{<:Real},
+                        row_inds,
                         col_ind::ColumnIndex)
-    insert_multiple_entries!(df, v, row_inds, col_ind)
+    if length(v) == length(df[row_inds, col_ind])
+        x = df[!, col_ind]
+        x[row_inds] = v
+    else
+        Base.depwarn("implicit vector broadcasting in in setindex! is deprecated", :setindex!)
```
quinnj:
```suggestion
        Base.depwarn("implicit vector broadcasting in setindex! is deprecated", :setindex!)
```

src/dataframe/dataframe.jl
```diff
@@ -472,249 +463,67 @@ function upgrade_scalar(df::DataFrame, v::Any)
...
-    (df[1:size(df, 1), 1:size(df, 2)] = v; df)
-
-# df[Any, :] = ...
-Base.setindex!(df::DataFrame, v, row_inds, ::Colon) =
-    (df[row_inds, 1:size(df, 2)] = v; df)
-
-# df[:, Any] = ...
-Base.setindex!(df::DataFrame, v, ::Colon, col_inds) =
-    (df[col_inds] = v; df)
-
```
quinnj: The ability to remove so many setindex! methods here is awesome!

src/dataframerow/dataframerow.jl
```diff
@@ -62,7 +62,11 @@ Base.@propagate_inbounds function DataFrameRow(sdf::SubDataFrame, row::Integer,
         throw(BoundsError("attempt to access a data frame with $(nrow(sdf)) " *
                           "rows at index $row"))
     end
-    colindex = SubIndex(index(parent(sdf)), parentcols(index(sdf), cols))
+    colindex = if index(sdf) isa Index
```
quinnj: I don't quite understand the change here.

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+
+import Base: setindex!
+@deprecate setindex!(sdf::SubDataFrame, val::Any, colinds::Any) (sdf[:, colinds] = val; sdf)
+@deprecate setindex!(df::DataFrame, v::AbstractVector, col_ind::ColumnIndex) (df[!, col_ind] = v; df)
+
+# df[SingleColumnIndex] = Single Item (EXPANDS TO NROW(df) if NCOL(df) > 0)
+function Base.setindex!(df::DataFrame, v, col_ind::ColumnIndex)
+    if haskey(index(df), col_ind)
+        Base.depwarn("Implicit broadcasting to an existing column in DataFrame assignment is deprecated." *
+                     "Use an explicit broadcast with df[:, col_ind] .= v.", :setindex!)
```
quinnj: Including the trailing period here might be confusing `df[:, col_ind] .= v.`. In general, I personally like putting syntax expressions in backticks, so this would be: `Use an explicit broadcast with [backtick]df[:, col_ind] .= v[backtick]`

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+    # a single replacement syntax
+    for j in 1:length(col_inds)
+        insert_single_column!(df, new_df[j], col_inds[j])
+    end
+    return df
+end
+
+# df[MultiColumnIndex] = AbstractVector (REPEATED FOR EACH COLUMN)
+function Base.setindex!(df::DataFrame, v::AbstractVector, col_inds::AbstractVector{Bool})
+    Base.depwarn("implicit vector broadcasting in in setindex! is deprecated", :setindex!)
```
quinnj:
```suggestion
    Base.depwarn("implicit vector broadcasting in setindex! is deprecated", :setindex!)
```

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+                     col_inds::AbstractVector{<:ColumnIndex}) (foreach(c -> (df[!, c] = copy(v)), col_inds); df)
+
+# df[MultiColumnIndex] = Single Item (REPEATED FOR EACH COLUMN; EXPANDS TO NROW(df) if NCOL(df) > 0)
+function Base.setindex!(df::DataFrame,
+                        val::Any,
+                        col_inds::AbstractVector{Bool})
+    setindex!(df, val, findall(col_inds))
+end
+function Base.setindex!(df::DataFrame, val::Any, col_inds::AbstractVector{<:ColumnIndex})
+    Base.depwarn("implicit broadcasting in in setindex! is deprecated", :setindex!)
```
quinnj:
```suggestion
    Base.depwarn("implicit broadcasting in setindex! is deprecated", :setindex!)
```

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+# df[:, SingleColumnIndex] = AbstractVector
+@deprecate setindex!(df::DataFrame, v::AbstractVector, ::Colon,
+                     col_ind::ColumnIndex) (df[!, col_ind] = v; df)
+
+# df[MultiRowIndex, SingleColumnIndex] = Single Item
+function Base.setindex!(df::DataFrame,
+                        v::Any,
+                        row_inds::AbstractVector{Bool},
+                        col_ind::ColumnIndex)
+    Base.depwarn("implicit broadcasting in in setindex! is deprecated", :setindex!)
```
quinnj:
```suggestion
    Base.depwarn("implicit broadcasting in setindex! is deprecated", :setindex!)
```

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+    setindex!(df, v, findall(row_inds), col_ind)
+end
+function Base.setindex!(df::DataFrame,
+                        v::Any,
+                        row_inds::AbstractVector{<:Integer},
+                        col_ind::ColumnIndex)
+    Base.depwarn("implicit broadcasting in in setindex! is deprecated", :setindex!)
```
quinnj:
```suggestion
    Base.depwarn("implicit broadcasting in setindex! is deprecated", :setindex!)
```

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+    insert_multiple_entries!(df, v, row_inds, col_ind)
+    return df
+end
+
+# df[MultiRowIndex, MultiColumnIndex] = AbstractVector
+function Base.setindex!(df::DataFrame,
+                        v::AbstractVector,
+                        row_inds::AbstractVector{Bool},
+                        col_inds::AbstractVector{Bool})
+    Base.depwarn("implicit broadcasting in in setindex! is deprecated", :setindex!)
```
quinnj:
```suggestion
    Base.depwarn("implicit broadcasting in setindex! is deprecated", :setindex!)
```

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+    setindex!(df, v, findall(row_inds), findall(col_inds))
+end
+function Base.setindex!(df::DataFrame,
+                        v::AbstractVector,
+                        row_inds::AbstractVector{Bool},
+                        col_inds::AbstractVector{<:ColumnIndex})
+    Base.depwarn("implicit broadcasting in in setindex! is deprecated", :setindex!)
```
quinnj:
```suggestion
    Base.depwarn("implicit broadcasting in setindex! is deprecated", :setindex!)
```

src/deprecated.jl
```diff
@@ -1409,3 +1409,310 @@ import Base: empty!
...
+    end
+    return df
+end
+
+# df[:] = DataFrame, df[:, :] = DataFrame
+function Base.setindex!(df::DataFrame,
+                        new_df::DataFrame,
+                        row_inds::Colon,
+                        col_inds::Colon=Colon())
+    Base.depwarn("replcating all df columns with copies of new_df is deprecated", :setindex!)
```
quinnj:
```suggestion
    Base.depwarn("replicating all df columns with copies of new_df is deprecated", :setindex!)
```

test/broadcasting.jl
```diff
@@ -223,309 +231,299 @@ end
...
-                         3.5  6.5  8.5  11.5  14.5
-                         4.5  6.5  9.5  12.5  15.5]
+    @test_throws ArgumentError dfv[!, :x2] .+= 1
 
     dfr = df[1, 3:end]
     dfr[[:x4, :x5]] .= 10
     @test Vector(dfr) == [7.5, 10.0, 10.0]
-    @test Matrix(df) == [2.5  5.5  7.5  10.0  10.0
-                         3.5  6.5  8.5  11.5  14.5
+    @test Matrix(df) == [2.5  4.5  7.5  10.0  10.0
```
quinnj: ??

src/dataframerow/dataframerow.jl
bkamins: The consequence of the change: if you do:
```
sdf = @view df[rows, :]
dfr = sdf[row, :]
```
then this has the same result as:
```
dfr = df[row, :]
```
In other words if you keep using `:` in `view` creation then the `:`-related property propagates.

The major consequence is that we have an improved performance of `DataFrameRow` creation from `SubDataFrame` (as we avoid creation of `SubIndex`). Additionally this makes the behavior consistent with the documentation (where we say that if you use `:` for column selection then view` follows changes of columns in the parent).

The practical way to think about it is the following: if you use `:` as column selector then view inherits `Index` from parent; if you use something else then `:` then in view we create a `SubIndex`. The original reason for this distinction was ensuring performance of operations like `eachrow` (where you know you select all columns always).

test/broadcasting.jl
bkamins: `dfv[!, :x2] .+= 1` above throws an error now in line 264 (`!` is disallowed in `SubDataFrames` as it indicates column replacement, while for `SubDataFrame` we can only do in-place assignment) therefore two entries in column `:x2` are not incremented by `1` in the test

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,131 @@ The rules for a valid type of index into a column are the following:
...
+* `df[rows, col] = v` -> set rows `rows` of column `col` in-place; `v` must be an `AbstractVector`;
+* `df[rows, cols] = v` -> set rows `rows` of columns `cols` in-place; `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
+                      (in this case column names must match);
+* `df[!, col] = v` -> replaces `col` with `v` without copying
+                      (with the exception that if `v` is an `AbstractRange` it gets converted to a `Vector`);
+                      also if `col` is a `Symbol` that is not present in `df` then a new column in `df` is created and holds `v`;
+                      equivalent to `df.col = v` if `col` is a valid identifier;
+
+Note in particular that only `df[!, col] = v` and `df.col = v` can be used to add a new column to a `DataFrame`.
+In particular `df[:, col] = v` does not add a column `v` to a `DataFrame` as it is in-place operation.
```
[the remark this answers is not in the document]
bkamins: This is not allowed. An error will be thrown. Literal `:` in row selection is treated the same as `1:nrow(df)` so it is in-place only. Clarification added.

docs/src/lib/indexing.md
```diff
@@ -21,77 +21,131 @@ The rules for a valid type of index into a column are the following:
...
+
+`setindex!` on `SubDataFrame`:
+* `sdf[row, col] = v` -> set value of `col` in row `row` to `v` in-place;
+* `sdf[CartesianIndex(row, col)] = v` -> the same as `sdf[row, col] = v`;
+* `sdf[row, cols] = v` -> the same as `dfr = df[row, cols]; dfr[:] = v` in-place;
+* `sdf[rows, col] = v` -> set rows `rows` of column `col`, in-place; `v` can be an abstract vector;
```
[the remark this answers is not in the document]
bkamins: must be. Fixed

src/dataframe/dataframe.jl
bkamins: There will be some new methods after the deprecation period, but for sure much less.

The problem with the old rules was that `:` was treated specially in different situations. For example `df[:, :] = df2` removed all columns from `df` and inserted into it columns from `df2`.

Now we have a clear rule `:` for row selection is not special. It is the same as `1:nrow(df)` so things get simpler (and are consistent with Base). For special operations we have `!` as row selector (so when one uses it this means that one knows that it is non-standard).

src/dataframerow/dataframerow.jl
bkamins: Actually your comment made me realize that we should have the same optimization in nested `SubDataFrame` (which I now added). Thank you.

src/deprecated.jl
bkamins: Agreed. Fixed

bkamins:
@quinnj - thank you for looking into this. I also replaced all remaining `in in` duplicates.

bkamins:
@nalimilan - this should be good to have a look at:
* I have minimized deprecation warning in all test files (there are few I thought we should test anyway even they throw a deprecation warning); a (massive) majority of deprecation warnings is thrown in test/deprecated.jl to make sure that we still pass all old tests in the deprecation period
* I have added a new section on `setindex!` rules in test/indexing.jl which goes one-to one with the documentation (for easier checking and for making sure what we should additionally test after the deprecation period ends)

src/groupeddataframe/grouping.jl
```diff
@@ -872,7 +872,7 @@ function append_rows!(rows, outcols::NTuple{N, AbstractVector},
         cn = colnames[j]
         local vals
         try
-            vals = rows[cn]
+            vals = rows isa AbstractDataFrame ? rows[!, cn] : rows[cn]
```
```suggestion
            vals = getproperty(rows, cn)
```

src/groupeddataframe/grouping.jl
```diff
@@ -906,7 +906,7 @@ function _combine_with_first!(first::Union{AbstractDataFrame,
             local newcols
             let i = i, j = j, outcols=outcols, rows=rows # Workaround for julia#15276
                 newcols = ntuple(length(outcols)) do k
-                    S = eltype(rows[k])
+                    S = eltype(rows isa AbstractDataFrame ? rows[!, k] : rows[k])
```
```suggestion
                    S = eltype(getproperty(rows, k))
```

src/subdataframe/subdataframe.jl
```diff
@@ -125,15 +134,22 @@ Base.@propagate_inbounds Base.getindex(sdf::SubDataFrame, rowinds::Union{Abstrac
...
+Base.@propagate_inbounds function Base.setindex!(sdf::SubDataFrame, val::Any, idx::CartesianIndex{2})
+    setindex!(sdf, val, idx[1], idx[2])
+end
+Base.@propagate_inbounds function Base.setindex!(sdf::SubDataFrame, val::Any, ::Colon, colinds::Any)
     parent(sdf)[rows(sdf), parentcols(index(sdf), colinds)] = val
     return sdf
 end
-
+Base.@propagate_inbounds function Base.setindex!(sdf::SubDataFrame, val::Any, ::typeof(!), colinds::Any)
+    throw(ArgumentError("setting index of SubDataFrame using ! as row selectior is not allowed"))
```
nalimilan:
```suggestion
    throw(ArgumentError("setting index of SubDataFrame using ! as row selector is not allowed"))
```

test/cat.jl
```diff
@@ -224,62 +224,24 @@ end
...
+                    3.0  3.0  3.0])
 
     @test vcat(missing_df) == DataFrame()
     @test vcat(missing_df, missing_df) == DataFrame()
     @test vcat(missing_df) == DataFrame()
     @test vcat(missing_df, missing_df) == DataFrame()
     @test vcat(missing_df, df) == df
     @test vcat(df, missing_df) == df
-    @test eltypes(vcat(df, df)) == Type[Float64, Float64, Int]
+    @test eltypes(vcat(df, df)) == Type[Float64, Float64, Float64]
```
nalimilan: Maybe keep the last column as `Int` just in case the code didn't handle the heterogeneity correctly.

test/indexing.jl
```diff
@@ -71,64 +59,55 @@ using Test, DataFrames
...
+    @test eachcol(df[Not(Int[]), Not([])])[1] !== df[!, 1]
 end
 
-@testset "getindex df[col] and df[cols]" begin
+@testset "getindex df[!, col]" begin
     x = [1, 2, 3]
     df = DataFrame(x=x, copycols=false)
     @test df.x === x
-    @test df[:x] === x
-    @test df[[:x]].x !== x
```
nalimilan: Restore equivalent of these tests if you add `df[, cols]`. Same for `view` below.

test/join.jl
```diff
@@ -343,47 +343,47 @@ end
...
     A = DataFrame(a = [1, 2, 3], b = ["a", "b", "c"])
     B = DataFrame(b = ["a", "b", "c"], c = CategoricalVector(["a", "b", "b"]))
-    levels!(B[:c], ["b", "a"])
-    @test levels(join(A, B, on=:b, kind=:inner)[:c]) == ["b", "a"]
-    @test levels(join(B, A, on=:b, kind=:inner)[:c]) == ["b", "a"]
-    @test levels(join(A, B, on=:b, kind =:left)[:c]) == ["b", "a"]
-    @test levels(join(A, B, on=:b, kind=:right)[:c]) == ["b", "a"]
-    @test levels(join(A, B, on=:b, kind=:outer)[:c]) == ["b", "a"]
-    @test levels(join(B, A, on=:b, kind =:semi)[:c]) == ["b", "a"]
+    levels!(B[!, :c], ["b", "a"])
```
nalimilan:
```suggestion
    levels!(B.c, ["b", "a"])
```
Same below.

src/dataframe/dataframe.jl
```diff
@@ -453,7 +444,7 @@ end
 
 function insert_multiple_entries!(df::DataFrame,
                                   v::Any,
-                                  row_inds::AbstractVector{<:Integer},
+                                  row_inds,
```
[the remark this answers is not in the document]
bkamins: we can add it. I removed it because in general we have to also handle `Not` and `Colon`, but I have just checked that we will remove this method after deprecation anyway (new methods do not use it), so your proposal is OK.

src/groupeddataframe/grouping.jl
bkamins: `k` is not a `Symbol` so we cannot use `getproperty` here

src/abstractdataframe/iteration.jl
```diff
@@ -199,10 +199,24 @@ julia> mapcols(x -> x.^2, df)
 """
 function mapcols(f::Union{Function,Type}, df::AbstractDataFrame)
     # note: `f` must return a consistent length
-    res = DataFrame()
-    for (n, v) in eachcol(df, true)
+    vs = AbstractVector[]
+    wasscalar = false
```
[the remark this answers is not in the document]
bkamins: A nice suggestion. Done

src/dataframe/dataframe.jl
```diff
@@ -73,9 +73,8 @@ df2 = DataFrame(A = 1:10, B = v, C = rand(10))
 summary(df1)
 describe(df2)
 first(df1, 10)
-df1[:A] + df2[:C]
+df1[!, :A] + df2[!, :C]
```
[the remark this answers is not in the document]
bkamins: added (+ added some more other examples)

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: I prefer using explicit typing where possible, especially since this code is particularly hard to follow.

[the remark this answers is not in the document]
bkamins: Sure - I have changed this. The only problem is (but not in this case) that it is a big mental burden to write proper sets of explicit types as often using `Union` leads to method ambiguities due to Julia's dispatch rules (sorry for being lazy sometimes 😄).

docs/src/lib/indexing.md
```diff
@@ -62,6 +62,7 @@ a `SubDataFrame` or a `DataFrameRow` always returns a `DataFrameRow` (which is a
                      the same as `df[!, col][rows]`;
 * `df[rows, cols]` -> a `DataFrame` containing copies of columns `cols` with only the entries corresponding to `rows` selected;
 * `df[!, col]` -> the vector contained in column `col` returned without copying; the same as `df.col` if `col` is a valid identifier.
+* `df[!, cols]` -> the same as `select(df, cols, copycols=false)`.
```
nalimilan: Also describe what this does? Indexing is more fundamental than `select`.

docs/src/lib/indexing.md
```diff
@@ -122,6 +127,7 @@ In particular a description explicitly mentions if the assignment is *in-place*.
                       (with the exception that if `v` is an `AbstractRange` it gets converted to a `Vector`);
                       also if `col` is a `Symbol` that is not present in `df` then a new column in `df` is created and holds `v`;
                       equivalent to `df.col = v` if `col` is a valid identifier;
+* `df[!, cols] = v` -> currently is disallowed, but its support is planned do be added in the future;
```
```suggestion
* `df[!, cols] = v` -> is currently disallowed, but is planned to be supported in the future;
```

docs/src/lib/indexing.md
```diff
@@ -100,9 +169,18 @@ In such an operation `AbstractDataFrame` is considered as two-dimensional and `D
     The rule above means that, similar to single-dimensional objects in Base (e.g. vectors),
     `DataFrameRow` is considered to be column-oriented.
 
-If column indexing using `Symbol` names is performed the order of columns in the operation is specified
-by the order of names.
+Additional rules:
+* in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` and `df[row, cols] .= v` syntaxes the assignment to `df` is performed in-place;
+* in the `df[rows, col] .= v` and `df[rows, cols] .= v` syntaxes the assignment to `df` is performed in-place;
+* in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector; if `col` is `Symbol` and it is missing from `df` then a new column is added;
+* `df[!, cols] = v` syntax currently is disallowed, but its support is planned do be added in the future;
```
```suggestion
* `df[!, cols] = v` syntax is currently disallowed, but is planned to be supported in the future;
```

docs/src/lib/indexing.md
bkamins: Added here and below. Thank you.

src/dataframe/dataframe.jl
```diff
@@ -432,6 +421,8 @@ function insert_single_column!(df::DataFrame,
             push!(_columns(df), dv)
         else
             if ncol(df) + 1 == Int(col_ind)
+                Base.depwarn("In the future setindex! will disallow adding columns" *
```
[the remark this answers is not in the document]
bkamins: I added an additional explanation.

As a separate thought - do you think we should export `hcat!`?

src/abstractdataframe/join.jl
```diff
@@ -28,7 +28,8 @@ struct DataFrameJoiner{DF1<:AbstractDataFrame, DF2<:AbstractDataFrame}
             left_on = [first(x) for x in on_cols]
             right_on = [last(x) for x in on_cols]
         end
-        new(dfl, dfr, dfl[left_on], dfr[right_on], left_on, right_on)
+        new(dfl, dfr, select(dfl, left_on, copycols=false),
```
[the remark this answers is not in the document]
bkamins: done

test/cat.jl
bkamins: OK. Fixed.

test/indexing.jl
bkamins: done

src/dataframe/dataframe.jl
```diff
@@ -472,249 +463,67 @@ function upgrade_scalar(df::DataFrame, v::Any)
...
-                        col_ind::ColumnIndex)
-    setindex!(df, v, findall(row_inds), col_ind)
-end
-function Base.setindex!(df::DataFrame,
-                        v::Any,
-                        row_inds::AbstractVector{<:Real},
+                        row_inds::Union{AbstractVector, Not}, # add Colon after deprecation
                         col_ind::ColumnIndex)
-    insert_multiple_entries!(df, v, row_inds, col_ind)
+    if length(v) == length(df[row_inds, col_ind])
```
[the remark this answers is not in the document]
bkamins: good point. I have replaced it with try-catch. This check will go away after the deprecation period.

src/dataframe/dataframe.jl
```diff
@@ -472,249 +463,67 @@ function upgrade_scalar(df::DataFrame, v::Any)
...
+        x = df[!, col_ind]
+        x[row_inds] = v
+    else
+        Base.depwarn("implicit vector broadcasting in setindex! is deprecated", :setindex!)
```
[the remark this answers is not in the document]
bkamins: Added, but this is probably a really rare case when someone does this.

src/dataframe/dataframe.jl
```diff
@@ -472,249 +463,67 @@ function upgrade_scalar(df::DataFrame, v::Any)
...
-                        new_df::DataFrame,
-                        row_inds::AbstractVector{<:Real},
-                        col_inds::AbstractVector{<:ColumnIndex})
-    for j in 1:length(col_inds)
-        insert_multiple_entries!(df, new_df[j], row_inds, col_inds[j])
+                        new_df::AbstractDataFrame,
+                        row_inds::Union{AbstractVector, Not}, # add Colon after deprecation
+                        col_inds::Union{AbstractVector, Regex, Not}) # add Colon after deprecation
+    idxs = index(df)[col_inds]
+    if names(df)[idxs] != names(new_df)
```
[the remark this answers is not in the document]
bkamins: added

src/dataframerow/dataframerow.jl
```diff
@@ -57,15 +57,25 @@ Base.@propagate_inbounds function DataFrameRow(df::DataFrame, row::Integer, cols
...
+Base.@propagate_inbounds DataFrameRow(df::DataFrame, row::Bool, cols) =
+    throw(ArgumentError("invalid row index of type Bool"))
+
 Base.@propagate_inbounds function DataFrameRow(sdf::SubDataFrame, row::Integer, cols)
     @boundscheck if !checkindex(Bool, axes(sdf, 1), row)
         throw(BoundsError("attempt to access a data frame with $(nrow(sdf)) " *
                           "rows at index $row"))
     end
-    colindex = SubIndex(index(parent(sdf)), parentcols(index(sdf), cols))
+    colindex = if index(sdf) isa Index # sdf was created using : as row selector
```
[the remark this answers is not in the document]
bkamins: done

src/deprecated.jl
```diff
@@ -1409,3 +1409,299 @@ import Base: empty!
...
+# df[MultiColumnIndex] = DataFrame
+function Base.setindex!(df::DataFrame, new_df::DataFrame, col_inds::AbstractVector{Bool})
+    setindex!(df, new_df, findall(col_inds))
+end
+function Base.setindex!(df::DataFrame,
+                        new_df::DataFrame,
+                        col_inds::AbstractVector{<:ColumnIndex})
+    Base.depwarn("Syntax `df[col_inds] = new_df` is deprecated." *
+                 "Use an explicit assignment of individual columns", :setindex!)
+    # the function has a different behavior when passed a vector of Symbols than
```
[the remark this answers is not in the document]
bkamins: I would leave this out. This is a super rare usage case and the behavior is very difficult to describe even if we split the definitions. E.g. in the old code you could add columns to `df` even using integers as `col_inds` but only if they form a continuous sequence of numbers beyond `ncol(df)` starting from `ncol(df+1)`, and only if they are sorted. So `[2,3,4]` is OK if `ncol(df) == 1` but e.g. `[3,2,4]` is not.

src/deprecated.jl
```diff
@@ -1409,3 +1409,299 @@ import Base: empty!
...
+    for j in 1:length(col_inds)
+        insert_single_column!(df, new_df[j], col_inds[j])
+    end
+    return df
+end
+
+# df[MultiColumnIndex] = AbstractVector (REPEATED FOR EACH COLUMN)
+function Base.setindex!(df::DataFrame, v::AbstractVector, col_inds::AbstractVector{Bool})
+    Base.depwarn("implicit vector broadcasting in setindex! is deprecated", :setindex!)
+    setindex!(df, v, findall(col_inds))
```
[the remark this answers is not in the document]
bkamins: OK - I changed it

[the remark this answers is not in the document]
bkamins: Below - in some places I was unable to reason what is the exact replacement, especially when `v` is `Any` as there are sometimes strange corner cases (of which we will hopefully get rid of after deprecation).

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: Probably better have Base define and export `hcat!` for matrices first? Though I guess we could export it, and just adapt when/if Julia exports it at some point.

src/deprecated.jl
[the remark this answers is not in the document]
nalimilan: My point is that it would allow suggesting a replacement for syntax at least when `col_inds` is a vector of symbols. As you like.

src/deprecated.jl
```diff
@@ -1409,3 +1409,299 @@ import Base: empty!
...
+                        col_inds::AbstractVector{Bool})
+    Base.depwarn("assigning a DataFrame to a single row of a DataFrame is deprecated", :setindex!)
+    setindex!(df, new_df, row_ind, findall(col_inds))
+end
+@deprecate setindex!(df::DataFrame, new_df::DataFrame, row_ind::Integer,
+                     col_inds::AbstractVector{<:ColumnIndex}) (foreach(c -> (df[row_ind, c] = new_df[1, c]), col_inds); df)
+
+# df[SingleRowIndex, MultiColumnIndex] = Single Item
+@deprecate setindex!(df::DataFrame, v::Any, row_ind::Integer,
+                     col_inds::AbstractVector{<:ColumnIndex}) (df[row_ind, col_inds] .= Ref(v); df)
```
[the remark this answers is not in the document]
bkamins: Here is the reason by example:
```
julia> using DataFrames

julia> df = DataFrame(a = [[1,2]], b=[[3,4]])
1×2 DataFrame
│ Row │ a      │ b      │
│     │ Array… │ Array… │
├─────┼────────┼────────┤
│ 1   │ [1, 2] │ [3, 4] │

julia> df[1, :] = [10, 10]
2-element Array{Int64,1}:
 10
 10

julia> df
1×2 DataFrame
│ Row │ a        │ b        │
│     │ Array…   │ Array…   │
├─────┼──────────┼──────────┤
│ 1   │ [10, 10] │ [10, 10] │
```
In short - the old rules implicitly broadcasted whatever they received. Therefore I have to protect it using `Ref` under new rules.

I did such proposals in places I felt I can safely make them. But in many places old rules have very non-obvious behavior to translate it to new rules in all cases correctly. In such cases I leave old rules "as is" and print a generic deprecation warning.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
bkamins: We can leave that decision for later. It is impossible currently to `hcat!` matrices as matrices are not resizable in Julia (as opposed to vectors).

src/dataframe/dataframe.jl
```diff
@@ -472,249 +463,67 @@ function upgrade_scalar(df::DataFrame, v::Any)
...
-function Base.setindex!(df::DataFrame,
-                        v::Any,
-                        row_inds::AbstractVector{Bool},
-                        col_ind::ColumnIndex)
-    setindex!(df, v, findall(row_inds), col_ind)
-end
-function Base.setindex!(df::DataFrame,
-                        v::Any,
-                        row_inds::AbstractVector{<:Real},
+                        row_inds::Union{AbstractVector, Not}, # add Colon after deprecation
```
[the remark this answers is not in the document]
bkamins: > knowing that a more specific definition will take precedence?

This is a problem as we have definitions in deprecated.jl and here that lead to method ambiguities (and the biggest problem is that, even if in theory possible, in practice it is really hard to make sure that we have covered all cases correctly and users will not be caught by "method ambiguity" error later - especially with so many methods as we have here).

It was simpler for me to leave it as proposed, as I am sure (or hope to be sure) it works.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: Ah yes of course -- only `vcat!` would make sense (for vectors). Then I guess we could add `hcat!` to DataAPI and reexport it.

src/deprecated.jl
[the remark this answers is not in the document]
bkamins: I know, but can you please propose one? I could not come up with a proposal that works in all cases.

src/deprecated.jl
```diff
@@ -1410,3 +1410,295 @@ import Base: empty!
...
+# df[:] = AbstractVector or Single Item
+function Base.setindex!(df::DataFrame, v, ::Colon)
+    Base.depwarn("`df[:] = v` syntax is deprecated; use `.=` broadcasting assignment", :setindex!)
+    df[1:size(df, 2)] = v
+    df
+end
+
+# df[SingleRowIndex, MultiColumnIndex] = 1-Row DataFrame
+@deprecate setindex!(df::DataFrame, new_df::DataFrame, row_ind::Integer,
+                        col_inds::AbstractVector{Bool}) (foreach(c -> (df[row_ind, c] = new_df[1, c]), findall(col_inds)); df)
```
nalimilan:
```suggestion
                   col_inds::AbstractVector{Bool}) (foreach(c -> (df[row_ind, c] = new_df[1, c]), findall(col_inds)); df)
```

test/indexing.jl
```diff
@@ -695,7 +759,7 @@ end
     @test sdf[!, :x2] isa SubArray
     @test parent(sdf[!, :x2]) === df.x2
     @test_throws ArgumentError @view sdf[!, :x]
-    @test_throws ArgumentError @view sdf[!, 1:2]
+    @test_throws select(sdf, 1:2, copycols=false) == @view sdf[!, 1:2]
```
nalimilan: ?

src/deprecated.jl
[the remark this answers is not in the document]
nalimilan: You know better than I do. :-p
Something with `foreach`?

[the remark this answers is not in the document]
bkamins: OK. I will do:

`foreach((j, colind) -> df[!, colind] = new_df[!, j] , enumerate(colinds))`

but it is (and was) a really unsafe operation to do.

test/indexing.jl
bkamins: This is a remainder of changed rules. This is now allowed (since yesterday 😄) and I will fix it. Thank you!

bkamins:
I am going to merge this in ~1 hour and then tag a release.

bkamins:
Thank you all who contributed into ideas, development and review of this PR.

Issue #1867, opened 2019-07-04
How about raise ArgumentError rather than just calling error() in insertcols!()?


petershintech:
I found the following case is the perfect case to raise ArgumentError(). What do you think of it?

```
function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:AbstractVector};
                     makeunique::Bool=false)
    name, item = name_col
    0 < col_ind <= ncol(df) + 1 || throw(BoundsError())
    size(df, 1) == length(item) || size(df, 2) == 0 || error("number of rows does not match")

    if hasproperty(df, name)
```

bkamins:
Sure - feel free to open a PR to improve both `throw(BoundsError())` and `error`.

bkamins:
fixed

PR #1870, opened 2019-07-04
Throw more specific error, ArgumentError, rather than a general Error()

petershintech:
#1867


bkamins:
Thank you. Could you please also:
* improve the `throw(BoundsError())` error message
* add tests to make sure that we correctly throw these errors?

PR #1872, opened 2019-07-04
Throw more specific error, ArgumentError, rather than a general Error()

petershintech:
Also changed the unit test to check ArgumentError, not ErrorException.
#1867

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -787,8 +787,10 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
     name, item = name_col
     if !(0 < col_ind <= ncol(df) + 1)
         throw(BoundsError("attempt to insert a column to a data frame with $(ncol(df)) columns at index $col_ind"))
+    end  
+    if !(size(df, 1) == length(item) || size(df, 2) == 0)
+        throw(ArgumentError("length of new column ($(length(item))) must match the number of rows in data frame ($(nrow(df)))"))
     end
-    size(df, 1) == length(item) || size(df, 2) == 0 || error("number of rows does not match")
 
     if hasproperty(df, name)
         if makeunique
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -789,7 +789,7 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
         throw(BoundsError("attempt to insert a column to a data frame with $(ncol(df)) columns at index $col_ind"))
     end  
     if !(size(df, 1) == length(item) || size(df, 2) == 0)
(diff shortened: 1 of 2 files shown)


src/dataframe/dataframe.jl
```diff
@@ -786,7 +786,7 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
                      makeunique::Bool=false)
     name, item = name_col
     0 < col_ind <= ncol(df) + 1 || throw(BoundsError())
-    size(df, 1) == length(item) || size(df, 2) == 0 || error("number of rows does not match")
+    size(df, 1) == length(item) || size(df, 2) == 0 || throw(ArgumentError("number of rows does not match"))
```
```suggestion
    if !(size(df, 1) == length(item) || size(df, 2) == 0)
        throw(ArgumentError("number of rows does not match"))
    end
```

bkamins: this fixes the overlong line

petershintech:
@bkamins Thanks. I agree your code is easy to read. In fact, I can see several long lines exploiting short circuits. Do you think we need to rewrite the lines to more readable ones? If so, I can help you.

src/dataframe/dataframe.jl
```diff
@@ -786,7 +786,9 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
                      makeunique::Bool=false)
     name, item = name_col
     0 < col_ind <= ncol(df) + 1 || throw(BoundsError())
-    size(df, 1) == length(item) || size(df, 2) == 0 || error("number of rows does not match")
+    if !(size(df, 1) == length(item) || size(df, 2) == 0)
+        throw(ArgumentError("number of rows does not match"))
```
```suggestion
        throw(ArgumentError("length of new column ($(length(item))) must match the number of rows in data frame ($(nrow(df)))"))
```

bkamins:
>  I can see several long lines exploiting short circuits.

Sure - thank you. The "soft" rule is that lines should not be longer than 92 characters.

petershintech:
I like the new message about more specific about the cause. Commit to the branch and ready to be merged. Thanks.

petershintech:
One quick question before merging. In this case,
if !(size(df, 1) == length(item) || size(df, 2) == 0)
should I throw ArgumentError or DimensionMismatch? I think the latter one is more specific about error cause so it would be better than the first one. What do you think of it?

nalimilan:
Right, `DimensionMismatch` sounds better.

petershintech:
Ok. I have replaced ArgumentError with DimensionMismatch and also touched a unit test for it. I can see the same change would be desirable for some other code lines but it can be captured in a separate PR. I think it is ready to merge. Thanks for all your inputs.

bkamins:
Yes, we used `ArgumentError` or `BoundsError` only everywhere up till now. We can be more specific of course 😄. Thank you.

Issue #1868, opened 2019-07-04
Why calling error() rather than just throw() in insert!()?


petershintech:
I found the following code in index.jl. Not sure why calling error() rather than just throwing the BoundsError()? Also would be better to provide some message to help users.
```
function Base.insert!(x::Index, idx::Integer, nm::Symbol)
    1 <= idx <= length(x.names)+1 || error(BoundsError())
    for i = idx:length(x.names)
```

bkamins:
Also agreed. There are also several more places where `error` is used (mostly because this is part of code that was not touched for a longer time and was not cleaned up). Feel free to open a PR if you wish.

petershintech:
Ok I will dig up a little bit more. I prefer small and isolated commits so you might receive multiple PRs for each small change.

bkamins:
OK - no problem.

PR #1873, opened 2019-07-04
Throw an exception with msg rather than error() with it.

petershintech:
Try to throw a similar message with other BoundsErrors but not sure the given message is the best one for this case. Please feel free to revise the message. #1868

### DIFF
--- a/src/other/index.jl
+++ b/src/other/index.jl
@@ -128,7 +128,9 @@ function Base.empty!(x::Index)
 end
 
 function Base.insert!(x::Index, idx::Integer, nm::Symbol)
-    1 <= idx <= length(x.names)+1 || throw(BoundsError("attempt to insert to a data frame with $(length(x.names)) columns at index $idx"))
+    if !(1 <= idx <= length(x.names)+1)
+        throw(BoundsError("attempt to insert a column to a data frame with $(length(x.names)) columns at index $idx"))
+     end
     for i = idx:length(x.names)
         x.lookup[x.names[i]] = i + 1
     end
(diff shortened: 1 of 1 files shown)


src/other/index.jl
```diff
@@ -128,7 +128,7 @@ function Base.empty!(x::Index)
 end
 
 function Base.insert!(x::Index, idx::Integer, nm::Symbol)
-    1 <= idx <= length(x.names)+1 || error(BoundsError())
+    1 <= idx <= length(x.names)+1 || throw(BoundsError("attempt to insert to a data frame with $(length(x.names)) columns at index $idx"))
```
bkamins:
```suggestion
    if !(1 <= idx <= length(x.names)+1)
    throw(BoundsError("attempt to insert to a data frame with $(length(x.names)) columns at index $idx"))
     end
```

bkamins: Looks good to me, but @nalimilan has a better feel for language issues 😄. Thank you.

nalimilan: Maybe this?
```suggestion
    if !(1 <= idx <= length(x.names)+1)
        throw(BoundsError("attempt to insert a column to a data frame with $(length(x.names)) columns at index $idx"))
     end
```

petershintech:
Agree. I committed @nalimilan's code to the branch. Would you merge the branch?

petershintech:
In test/dataframe.jl, I can't see any unit test checking insert!(). Where should I insert a unit test for this new code? Any suggestion?

bkamins:
Thank you

bkamins:
Adding test can be a separate PR. You can add test of `Index` `insert!` in test/index.jl.

PR #1874, opened 2019-07-04
Add an error message to give more details

petershintech:
As requested, add an error message to BoundsError(). Not sure if the given message is the best one in this case. Please feel free to change the message. #1867

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -786,7 +786,7 @@ function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:Abstr
                      makeunique::Bool=false)
     name, item = name_col
     if !(0 < col_ind <= ncol(df) + 1)
-        throw(BoundsError("attempt to insert to a data frame with $(ncol(df)) columns at index $col_ind"))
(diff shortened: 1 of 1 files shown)


src/dataframe/dataframe.jl
```diff
@@ -785,7 +785,7 @@ julia> insertcols!(d, 1, :c => [2, 3, 4])
 function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol, <:AbstractVector};
                      makeunique::Bool=false)
     name, item = name_col
-    0 < col_ind <= ncol(df) + 1 || throw(BoundsError())
+    0 < col_ind <= ncol(df) + 1 || throw(BoundsError("attempt to insert to a data frame with $(ncol(df)) columns at index $col_ind"))
```
```suggestion
    if !(0 < col_ind <= ncol(df) + 1)
        throw(BoundsError("attempt to insert to a data frame with $(ncol(df)) columns at index $col_ind"))
    end
```

bkamins: Again - looks good to me.

CC @nalimilan

petershintech:
I added "a column" in the error message to be aligned with #1873.

PR #1875, opened 2019-07-05
Throw a more specific exception when dimensions are mismatched

petershintech:
I have replaced two obvious cases where it is better to throw DimensionMismatch error rather than a more general ArgumentError. Not sure how to test this feature. Appreciate if I can have any help for unit test.


bkamins:
These two errors should never happen in practice. They can only happen due to a bug or serious user error and both methods are not part of public API. I think it is OK not to test for these errors but if you wish to add the tests you can do it by calling both methods directly in test/broadcast.jl.

PR #1876, opened 2019-07-07
Minor cleanup of abstractdataframe.jl

bkamins:
Small things I noticed when I review all code for `!` indexing change.


bkamins:
Thank you! (today I think I will be able to finish the monster `!` PR)

PR #1877, opened 2019-07-07
Reformat long error-throwing code lines

petershintech:
Make sure to limit error-throwing code lines within the width of 95 characters.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -240,7 +240,7 @@ Base.propertynames(df::AbstractDataFrame, private::Bool=false) = names(df)
 ##############################################################################
 
 """
-    similar(df::DataFrame[, rows::Integer])
+    similar(df::AbstractDataFrame, rows::Integer=nrow(df))
 
 Create a new `DataFrame` with the same column names and column element types
 as `df`. An optional second argument can be provided to request a number of rows
@@ -265,7 +265,7 @@ function Base.:(==)(df1::AbstractDataFrame, df2::AbstractDataFrame)
     for idx in 1:size(df1, 2)
         coleq = df1[idx] == df2[idx]
         # coleq could be missing
-        !isequal(coleq, false) || return false
+        isequal(coleq, false) && return false
         eq &= coleq
     end
     return eq
@@ -477,7 +477,6 @@ function get_stats(col::AbstractVector, stats::AbstractVector{Symbol})
             d[:std] = try std(col, mean = m) catch end
         end
     end
-
 
     if :nunique in stats
         if eltype(col) <: Real
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -817,7 +817,7 @@ function Base.convert(::Type{Matrix{T}}, df::AbstractDataFrame) where T
             if err isa MethodError && err.f == convert &&
                !(T >: Missing) && any(ismissing, col)
                 throw(ArgumentError("cannot convert a DataFrame containing missing values to Matrix{$T} " *
-                      "(found for column $name)"))
+                                     "(found for column $name)"))
             else
                 rethrow(err)
             end
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -405,7 +405,8 @@ function _describe(df::AbstractDataFrame, stats::AbstractVector)
 
     if !allunique(ordered_names)
         duplicate_names = unique(ordered_names[nonunique(DataFrame(ordered_names = ordered_names))])
-        throw(ArgumentError("Duplicate names not allowed. Duplicated value(s) are: :$(join(duplicate_names, ", "))"))
+        throw(ArgumentError("Duplicate names not allowed. Duplicated value(s) are: " *
+                            ":$(join(duplicate_names, ", "))"))
     end
 
     # Put the summary stats into the return data frame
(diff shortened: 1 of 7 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -816,7 +817,8 @@ function Base.convert(::Type{Matrix{T}}, df::AbstractDataFrame) where T
         catch err
             if err isa MethodError && err.f == convert &&
                !(T >: Missing) && any(ismissing, col)
-                error("cannot convert a DataFrame containing missing values to Matrix{$T} (found for column $name)")
+                error("cannot convert a DataFrame containing missing values to Matrix{$T} " *
```
bkamins: maybe change this to `ArgumentError`?

src/abstractdataframe/io.jl
```diff
@@ -82,7 +82,8 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
         if size(df, 2) == 0
             rowid = nothing
         else
-            size(df, 1) == 1 || throw(ArgumentError("rowid may be passed only with a single row data frame"))
+            size(df, 1) == 1 ||
```
bkamins: I personally would prefer to use `if` in situations when we do a line break after `||` as it might be confusing when reading the code. Or maybe just split the error message like in other places?

src/abstractdataframe/abstractdataframe.jl
petershintech: @bkamins I agree with you. With this PR, I just focused on line lengths. As you advised, I will commit the change.

src/abstractdataframe/io.jl
petershintech: Due to an indentation in the next line, I cannot see any problem in reading the broken lines. Is your preferred style a widely accepted practice in Julia community? If so, I have no problem with the change.

quinnj: There isn't a generally accepted style here in the Julia community (that I'm aware of), so it's more up to discussion and package maintainer preference. In this case, I personally don't have a strong preference, so I'm fine with whatever others want to go with.

bkamins: I have just checked in Base that this style is used there, so I think we can leave it as is.

petershintech:
As instructed, replaced with if statements. Also fix a bug in testing a conversion to matrix. Hopefully, it is good enough to merge.

src/abstractdataframe/io.jl
```diff
@@ -81,8 +81,8 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
     if rowid !== nothing
         if size(df, 2) == 0
             rowid = nothing
-        else
-            size(df, 1) == 1 || throw(ArgumentError("rowid may be passed only with a single row data frame"))
+        elseif size(df, 1) != 1 
+                throw(ArgumentError("rowid may be passed only with a single row data frame"))
```
bkamins: is the indentation here correct (same below)?

bkamins:
Thank you. I have left one comment (I am unsure as maybe GitHub is displaying it incorrectly).

Also I have two minor comments:
* in deprecated.jl we do not have to be super clean; most of this code will go away in a few months
* have you cherry-picked https://github.com/JuliaData/DataFrames.jl/pull/1877/commits/a411ee4384a244f3478bbf36363d7ae38e40d01f? Usually it is better to either merge master into your PR or rebase the PR against master, as then GitHub shows changes against master (so it is easier to see what is changed in the PR; now the diff in GitHub shows changes from two PRs)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -816,7 +816,8 @@ function Base.convert(::Type{Matrix{T}}, df::AbstractDataFrame) where T
         catch err
             if err isa MethodError && err.f == convert &&
                !(T >: Missing) && any(ismissing, col)
-                error("cannot convert a DataFrame containing missing values to Matrix{$T} (found for column $name)")
+                throw(ArgumentError("cannot convert a DataFrame containing missing values to Matrix{$T} " *
+                      "(found for column $name)"))
```
```suggestion
                                     "(found for column $name)"))
```

src/abstractdataframe/io.jl
```diff
@@ -214,8 +214,8 @@ function _show(io::IO, ::MIME"text/latex", df::AbstractDataFrame; rowid=nothing)
...
```
```suggestion
            throw(ArgumentError("rowid may be passed only with a single row data frame"))
```

src/abstractdataframe/show.jl
```diff
@@ -518,8 +518,8 @@ function _show(io::IO,
     if rowid !== nothing
         if size(df, 2) == 0
             rowid = nothing
-        else
-            nrows == 1 || throw(ArgumentError("rowid may be passed only with a single row data frame"))
+        elseif nrows != 1 
+                throw(ArgumentError("rowid may be passed only with a single row data frame"))
```
```suggestion
            throw(ArgumentError("rowid may be passed only with a single row data frame"))
```

petershintech:
Thanks @nalimilan for correcting my mistake in indentation.

src/abstractdataframe/io.jl
ararslan:
```suggestion
            throw(ArgumentError("rowid may be passed only with a single row data frame"))
```

petershintech:
Ok. I merged all proposed commits. Please let me know if the branch needs more touches.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -815,7 +816,8 @@ function Base.convert(::Type{Matrix{T}}, df::AbstractDataFrame) where T
         catch err
             if err isa MethodError && err.f == convert &&
                !(T >: Missing) && any(ismissing, col)
-                error("cannot convert a DataFrame containing missing values to Matrix{$T} (found for column $name)")
+                throw(ArgumentError("cannot convert a DataFrame containing missing values to Matrix{$T} " *
+                                     "(found for column $name)"))
```
```suggestion
                                    "(found for column $name)"))
```

bkamins:
Thank you!

I have spotted one misaligned line (I think). If in the future you allow force-pushing to your PR when creating it such small fixes can be without requiring an additional cycle (but it is of course up to you).

petershintech:
Ok. I will follow your advice.

bkamins:
Thank you for the PR!

Issue #1885, opened 2019-07-17
Preventing problems with aliased columns


jablauvelt:
A note in the documentation for `append!()` says:

> Please note that append! must not be used on a DataFrame that contains columns that are aliases (equal when compared with ===) as it will silently produce a wrong result in such a situation.

Unfortunately, I only found this note, and understood it, after many hours of confusion and debugging (I have only been using Julia for a couple of months). I was getting strange errors that didn't seem to be related to `append!()`, and therefore took a while to track down. It turns out that appending a DataFrame to another DataFrame with aliased columns caused the problem. Here's a reproducible example:

```
x1 = DataFrame(a = [1, 2, 3], b= ["x", "y", "z"])
x1.c = x1.b # here is the aliasing that causes the problem
x2 = DataFrame(a = [4, 5], b= ["x2", "y2"], c = ["z", "z"])
append!(x1, x2) # succeeds with no warning

# The dataframe looks normal when displayed:
show(x1)
# 5×3 DataFrame
# │ Row │ a     │ b      │ c      │
# │     │ Int64 │ String │ String │
# ├─────┼───────┼────────┼────────┤
# │ 1   │ 1     │ x      │ x      │
# │ 2   │ 2     │ y      │ y      │
# │ 3   │ 3     │ z      │ z      │
# │ 4   │ 4     │ x2     │ x2     │
# │ 5   │ 5     │ y2     │ y2     │

# But columns b and c are longer than column a:
[length(col) for col in eachcol(x1)]
# 3-element Array{Int64,1}:
# 5
# 7
# 7

# Here is what b (and c) actually look like:
x1.b
# 7-element Array{String,1}:
# "x"
# "y"
# "z"
# "x2"
# "y2"
# "z"
# "z"

# This leads to all sorts of strange errors, such as:
x1[x1.a .== 2, :]
# ERROR: BoundsError: attempt to access 7-element Array{String,1} at index [Base.LogicalIndex(Bool[false, true, false, false, false])]

# This is initially where I started seeing problems, but my append!() logic was
# many steps away from this filtering line, so it took a while to track down the culprit.
```

I know the note in the documentation technically warns users of this behavior, but many users who are less familiar with Julia, and aren't thinking of object references/copying and so forth, will also be very confused. I think it would help immensely to add some basic checks on DataFrames to ensure that their columns are consistent sizes or that there aren't problems created by aliased columns. These checks would go a long way towards protecting Julia's reputation of being safe and transparent.

[...]

Issue #1845, opened 2019-06-10
Add column length checks to expensive operations


oxinabox:
From: https://github.com/JuliaData/DataFrames.jl/issues/1844#issuecomment-500577183

> ... we will always have holes) is to create a list of methods in DataFrames.jl that actively check if all columns have the same length before performing their operation (essentially - all expensive methods).

In short: checking column lengths is cheap,
We should do that a bunch of operations,
even of we do think we have patched all holes that let a user resize columns out of sync,
we can still catch bugs in internal methods via this (defensive programming)

bkamins:
This is my point 😄. Thank you!
(I will get to it after I am finished with broadcasting and `setindex!` cleanups) - unless someone else will by then (which I am happy to review then)

bkamins:
Here is a benchmark of checking cost for 20,000 columns (which I think is a typically reasonable maximum in practice):
```
julia> df = DataFrame(rand(10, 20_000));

julia> function samelength(df)
           rows = nrow(df)
           for col in _columns(df)
               length(col) == rows || return false
           end
           return true
       end
samelength (generic function with 1 method)

julia> @benchmark samelength($df)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     492.100 μs (0.00% GC)
  median time:      497.400 μs (0.00% GC)
  mean time:        627.434 μs (0.00% GC)
  maximum time:     2.256 ms (0.00% GC)
  --------------
  samples:          7955
  evals/sample:     1
```

So given this the question is for what functions we should add this check.

oxinabox:
20,000 columns is definately an upper bound.
E.g. basically all spreedsheet programs crash out at over 16,384 columns.

I think it should be called if you do 
 - [ ] show / print
 - [ ] `first` (when called with the `nrows` argument)

Which will catch things during in interactive use.

bkamins:
It is tempting to check it also with every call of `nrows` but I have to check if we do not call it somewhere in hot loops.

I would also add:

- [ ] reshape operations
- [ ] filtering
- [ ] removing duplicates
- [ ] removing missings
- [ ] eachcol/eachrow upon creation of the object

oxinabox:
> It is tempting to check it also with every call of nrows but I have to check if we do not call it somewhere in hot loops.

If we do, then we should maybe define: `nrows(check_column_constancy=true)`
And then use `nrows(false)` in the hotloops.

[...]

PR #1887, opened 2019-07-18
First proposal of consistency checks

bkamins:
This fixes https://github.com/JuliaData/DataFrames.jl/issues/1885 and https://github.com/JuliaData/DataFrames.jl/issues/1845.

What it does:
* sanitizes `push!` and `append!`, as I concluded that it is best to do more strict checks there (as we do dynamic dispatch there anyway)
* adds consistency checks to `join`, `groupby` and `_show`

(this also lead to a bit different approach to error throwing)

Consistency check for 10000 column costs ~200 microseconds so I think it is acceptable. But please comment on this.

Also please comment where else you think it is worth to add consistency checks.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -291,9 +291,10 @@ function _check_consistency(df::DataFrame)
     ncols == 0 && return nothing
     nrows = length(cols[1])
     for i in 2:length(cols)
-        @assert length(cols[i]) == nrows "Length of column $(names(df)[i]) ($(length(df[!, i])))" *
+        @assert length(cols[i]) == nrows "Data frame is corrupt: length of column :$(names(df)[i]) ($(length(df[!, i])))" *
                                          " does not match length of column 1 ($(length(df[!, 1]))). " *
-                                         "It is likely caused by unintended resizing of the column."
+                                         "The column vector has likely been resized unintentionally " *
+                                         "(either directly or because it is shared with another data frame)."
     end
     nothing
 end
@@ -1177,7 +1178,7 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns
         for col in _columns(df)
             resize!(col, nrows)
         end
-        @error "Error adding value to column $(names(df)[current_col])."
+        @error "Error adding value to column :$(names(df)[current_col])."
         rethrow(err)
     end
     df
@@ -1291,7 +1292,7 @@ function Base.push!(df::DataFrame, row::Any)
         for col in _columns(df)
             resize!(col, nrows)
         end
-        @error "Error adding value to column $(names(df)[current_col])."
+        @error "Error adding value to column :$(names(df)[current_col])."
         rethrow(err)
     end
     df
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -287,11 +287,13 @@ ncol(df::DataFrame) = length(index(df))
 function _check_consistency(df::DataFrame)
     cols, idx = _columns(df), index(df)
     ncols = length(cols)
-    @assert length(idx.names) == length(idx.lookup) == ncols "Index and columns of a data frame do not match"
+    @assert length(idx.names) == length(idx.lookup) == ncols
     ncols == 0 && return nothing
     nrows = length(cols[1])
     for i in 2:length(cols)
-        @assert length(cols[i]) == nrows "Length of column $i does not match length of column 1"
+        @assert length(cols[i]) == nrows "Length of column $(names(df)[i]) ($(length(df[!, i])))" *
+                                         " does not match length of column 1 ($(length(df[!, 1]))). " *
+                                         "It is likely caused by unintended resizing of the column."
     end
     nothing
 end
@@ -1116,11 +1118,15 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame)
     _names(df1) == _names(df2) || error("Column names do not match")
     nrows, ncols = size(df1)
     targetrows = nrows + nrow(df2)
+    current_col = 0
     try
         for j in 1:ncols
+            current_col += 1
             append!(df1[!, j], df2[!, j])
         end
+        current_col = 0
         for col in _columns(df1)
+            current_col += 1
             @assert length(col) == targetrows
         end
     catch err
@@ -1128,6 +1134,7 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame)
         for col in _columns(df1)
             resize!(col, nrows)
         end
+        @error "Error adding value to column $(names(df)[current_col])."
         rethrow(err)
     end
     return df1
@@ -1142,7 +1149,6 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns
         throw(ArgumentError("`columns` keyword argument must be `:equal` or `:intersect`"))
     end
     nrows, ncols = size(df)
-    targetrows = nrows + 1
     if ncols == 0 && row isa NamedTuple
         for (n, v) in pairs(row)
             setproperty!(df, n, fill!(Tables.allocatecolumn(typeof(v), 1), v))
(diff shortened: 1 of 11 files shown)


src/dataframe/dataframe.jl
```diff
@@ -278,6 +278,26 @@ _columns(df::DataFrame) = getfield(df, :columns)
...
+##############################################################################
+##
+## DataFrame consistency check
+##
+##############################################################################
+
+function _check_consistency(df::DataFrame)
+    cols, idx = _columns(df), index(df)
+    ncols = length(cols)
+    @assert length(idx.names) == length(idx.lookup) == ncols "Index and columns of a data frame do not match"
```
nalimilan: More precisely, lengths don't match. This should never happen, right? If so, maybe just keep the default printing.

src/dataframe/dataframe.jl
```diff
@@ -278,6 +278,26 @@ _columns(df::DataFrame) = getfield(df, :columns)
...
+    ncols == 0 && return nothing
+    nrows = length(cols[1])
+    for i in 2:length(cols)
+        @assert length(cols[i]) == nrows "Length of column $i does not match length of column 1"
```
nalimilan: Print column name, its length, and mention that it's likely due to resizing the column vector? Since we expect this kind of error to happen due to user errors rather than bugs in our code, maybe this should be an `ArgumentError`? Not sure.

src/dataframe/dataframe.jl
```diff
@@ -1096,14 +1115,18 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame)
 
     _names(df1) == _names(df2) || error("Column names do not match")
     nrows, ncols = size(df1)
+    targetrows = nrows + nrow(df2)
     try
         for j in 1:ncols
             append!(df1[!, j], df2[!, j])
         end
+        for col in _columns(df1)
+            @assert length(col) == targetrows
```
nalimilan: Same here.

src/dataframe/dataframe.jl
```diff
@@ -1118,32 +1141,33 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns
...
     end
-    for nm in _names(df)
-        try
-            push!(df[!, i], row[nm])
-        catch
-            #clean up partial row
-            for j in 1:(i - 1)
-                pop!(df[!, j])
-            end
-            msg = "Error adding value to column :$nm."
```
nalimilan: This was probably useful to understand what is the problematic column. Not sure how we can convey both this information, and the full stacktrace for the actual error.

src/dataframe/dataframe.jl
```diff
@@ -1118,32 +1141,33 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns
     if !(columns in (:equal, :intersect))
         throw(ArgumentError("`columns` keyword argument must be `:equal` or `:intersect`"))
     end
-    if ncol(df) == 0 && row isa NamedTuple
+    nrows, ncols = size(df)
+    targetrows = nrows + 1
```
nalimilan: Just use `nrows + 1` below instead of defining a variable? Same in the next function.

src/dataframe/dataframe.jl
```diff
@@ -278,6 +278,26 @@ _columns(df::DataFrame) = getfield(df, :columns)
 nrow(df::DataFrame) = ncol(df) > 0 ? length(_columns(df)[1])::Int : 0
 ncol(df::DataFrame) = length(index(df))
 
+##############################################################################
+##
+## DataFrame consistency check
+##
+##############################################################################
+
+function _check_consistency(df::DataFrame)
```
nalimilan: I don't think we generally use underscores for private functions, do we?

src/dataframe/dataframe.jl
```diff
@@ -278,6 +278,26 @@ _columns(df::DataFrame) = getfield(df, :columns)
...
+    @assert length(idx.names) == length(idx.lookup) == ncols "Index and columns of a data frame do not match"
+    ncols == 0 && return nothing
+    nrows = length(cols[1])
+    for i in 2:length(cols)
+        @assert length(cols[i]) == nrows "Length of column $i does not match length of column 1"
+    end
+    nothing
+end
+
+_check_consistency(df::AbstractDataFrame) = _check_consistency(parent(df))
```
nalimilan: This doesn't sound fully generic, as we don't require `AbstractDataFrame` to have a parent (even if that's the case in practice). Is there a way to check consistency only using `AbstractDataFrame` API?

src/dataframe/dataframe.jl
bkamins: removed

src/dataframe/dataframe.jl
bkamins: message fixed. I would print `@assert` as this indicates "broken contract". I was considering `ArgumentError`, but I would reserve it for cases when user makes an error when calling a function (not - passes "broken" data).

src/dataframe/dataframe.jl
bkamins: Like above I personally would prefer to see `@assertion` error here to know that something really bad happened (opposed to simply passing a wrong argument).

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl

src/dataframe/dataframe.jl
bkamins: Actually I think that `parent` is part of `AbstractDataFrame` API (like it is part of arrays API in Base).
But if you think it is not then I can just special case for `SubDataFrame`. This is not a big deal whatever we choose I think 😄.

src/dataframe/dataframe.jl
bkamins: Not always, but I think we should. This is a more clear signal between a function that is not exported and really is internal (like this one) and a function that could be potentially exported or used by the end user (like `hcat!`).

src/dataframe/dataframe.jl
```diff
@@ -1118,32 +1148,37 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns
...
+        current_col = 0
+        for col in _columns(df)
+            current_col += 1
+            @assert length(col) == nrows + 1
+        end
+    catch err
+        for col in _columns(df)
+            resize!(col, nrows)
+        end
+        @error "Error adding value to column $(names(df)[current_col])."
```
```suggestion
        @error "Error adding value to column :$(names(df)[current_col])."
```

src/dataframe/dataframe.jl
```diff
@@ -1236,23 +1270,29 @@ julia> push!(df, Dict(:A=>1.0, :B=>2.0))
...
+        for col in _columns(df)
+            current_col += 1
+            @assert length(col) == nrows + 1
+        end
+    catch err
+        #clean up partial row
+        for col in _columns(df)
+            resize!(col, nrows)
+        end
+        @error "Error adding value to column $(names(df)[current_col])."
```
```suggestion
        @error "Error adding value to column :$(names(df)[current_col])."
```

src/dataframe/dataframe.jl
nalimilan: Clever.

src/dataframe/dataframe.jl
```diff
@@ -278,6 +278,28 @@ _columns(df::DataFrame) = getfield(df, :columns)
...
+##############################################################################
+
+function _check_consistency(df::DataFrame)
+    cols, idx = _columns(df), index(df)
+    ncols = length(cols)
+    @assert length(idx.names) == length(idx.lookup) == ncols
+    ncols == 0 && return nothing
+    nrows = length(cols[1])
+    for i in 2:length(cols)
+        @assert length(cols[i]) == nrows "Length of column $(names(df)[i]) ($(length(df[!, i])))" *
```
```suggestion
        @assert length(cols[i]) == nrows "Data frame is corrupt: length of column :$(names(df)[i]) ($(length(df[!, i])))" *
```

src/dataframe/dataframe.jl
```diff
@@ -278,6 +278,28 @@ _columns(df::DataFrame) = getfield(df, :columns)
...
+                                         " does not match length of column 1 ($(length(df[!, 1]))). " *
+                                         "It is likely caused by unintended resizing of the column."
```
```suggestion
                                         "The column vector has likely been resized unintentionally " *
                                         "(either directly or because it is shared with another data frame)."
```

src/dataframe/dataframe.jl
bkamins: So are you OK with `@assert` or we switch it to `ArgumentError` (I prefer `@assert`, but this is not a crucial thing for me to keep)

nalimilan: Yes that's fine.

Issue #1886, opened 2019-07-17
Allow column name transformations in unstack


bkamins:
It would be useful to allow `unstack` to have a kwarg that allows to apply a function to transform generated column names.

CC @oxinabox

oxinabox:
For sake of readers I am going to post more from the the whole conversation:

> 
> I wish unstack took a column name prefix
> I tend to do things like
> ```unstack(load_table, :time, :region, :load);```
> which gives me 1 row per time, and then 1 column per region,
> and the cells contain the load.
> The columns are thus named `North`, `South` etc. per the entriesi nthe Region column
> but I wish I could say to name them `load_North`, `load_South` etc.
> 
> Depending o nwhat other data I am joining it on to, I might want to say `load_North_region`
> So maybe even should just be a function?
> 
> You can join symbols with
> ```oldname->Symbol(:load_, oldname, :_region)```
> Which just seems more compact than a prefix and a suffix option.
> 
> Main usecase beyond suffixing and prefixing, I have, would be normalizing cases.
> Often things in the values are in “ALL CAPS” (with spaces even)

and I would like to transform that to `lower_case` before I make it a column name

bkamins:
And to add to this - normalizing **before** making it a column name could lead to a warning and data loss. So we should decide if we prefer this or throwing an error if duplicate column name is produced.

PR #1888, opened 2019-07-18
add transform keyword argument to unstack

bkamins:
Fix https://github.com/JuliaData/DataFrames.jl/issues/1886

CC @oxinabox


src/abstractdataframe/reshape.jl
```diff
@@ -171,12 +173,18 @@ unstack(df::AbstractDataFrame)
...
+  them into `Symbol`s that are target column names
+
 ### Result
 
 * `::DataFrame` : the wide-format DataFrame
 
 If `colkey` contains `missing` values then they will be skipped and a warning will be printed.
 
+If transforming `colkey` using the `transform` function produces duplicates
+then an error is thrown.
```
oxinabox:
```suggestion
in a way such that different `colkey` values map to that same transformed column name
then an error is thrown.
```

nalimilan:
How does this solution compare with tidyverse's new `pivot_wider` API? See https://rdrr.io/github/tidyverse/tidyr/man/pivot_wider.html

bkamins:
In terms of column naming: we are more flexible. What they have is a special case.

In terms of the "core funcitionality": `pivot_wider` allows multiple variable and value columns and we accept only one for each.

nalimilan:
OK. Maybe at some point we'll want to add keyword arguments for prefix and suffix, as it's more convenient than `x -> Symbol("prefix", x)`. But that's orthogonal to this PR.

Regarding the argument name, `transform` sounds too broad (it could refer to values). Maybe `rename`, since we have a method with that name with takes a function?

bkamins:
`rename` is OK for me. @oxinabox - any opinion?

oxinabox:
sounds good to me.
Maybe even: `rename_columns`?

bkamins:
I renamed it to `renamecols` as we use this `cols` suffix in other places.

bkamins:
@nalimilan - OK to merge this?

src/abstractdataframe/reshape.jl
```diff
@@ -171,12 +173,19 @@ unstack(df::AbstractDataFrame)
 
 * `value` : the value column, defaults to `:value`
 
+* `renamecols` : the function that takes column names from `colkey` and transforms
```
nalimilan: I'd say:
"a function that takes a `Vector{Symbol}` of column names from `colkey` and returns target column names as a vector of `Symbol`s or `String`s"

bkamins: actually now `colkey` can be anything as long as `renamecols` returns a vector of `Symbol`, `AbstractString` or `Vector{UInt8}`. I will update the docs

bkamins: And it works elementwise (I will make this clear also)

bkamins: done. Actually I renamed the kwarg to `renamecol` to make it clear that it operates on a single column name.

src/abstractdataframe/reshape.jl
```diff
@@ -171,12 +173,19 @@ unstack(df::AbstractDataFrame)
 
 * `value` : the value column, defaults to `:value`
 
+* `renamecol` : the function that takes a column name from `colkey` and returns
```
nalimilan: Maybe something like this? (It would also replace the note below.)
```
a function called on each unique value in `colkey` which must return the name of the column to be created (typically as a string or a `Symbol`). Duplicate names are not allowed.
```

Also, I like `renamecols` better since the function is applied to all columns, and that's the standard suffix we use.

bkamins: OK - fixed. I understand that by

> I would also replace the note below

you wanted me to remove it. Right?

PR #1890, opened 2019-07-18
allow scalar broadcasting into an empty data frame

### LINKED ISSUE #1883: Release 0.19.0
@JuliaRegistrator register()

### LINKED ISSUE #1889: Creating a column via broadcasting in an empty data frame
In DataFrames 0.18.4, I can for example do the following:
`df[cols] = 0.0`
where df is a data frame, cols is a vector of Symbol naming columns not in df.

As a result, df gets new columns of type Float64 containing 0.0 entries in all rows. If df is an empty data frame, i.e. with 0 rows, it still gets the new columns with type Float64, but of course having 0 rows.

This seems logical: creating a column through broadcasting to an empty data frame creates an empty column. In particular it means I don't have to test separately for the special case of an empty data frame.

In DataFrames 0.19.0 the same code gives me an error when df is an empty data frame:
`ArgumentError: creating a column via broadcasting is not allowed on empty data frames`
from broadcasting.jl in the Base.copyto! method.

`df[!,col] .= 0.0` works in 0.19.0, where df is non-empty and col identifies a single new column: a new column col is created with 0.0 entries in all rows. Can this also be made to work for empty df, similar to the behaviour of 0.18.4?

**NOTE** by "empty" here I mean a data frame with columns but with no rows. The behaviour for a data frame that is completely empty, no columns or rows, is not consistent with what I've described, as discussed [a few comments further down](https://github.com/JuliaData/DataFrames.jl/issues/1889#issuecomment-513314241).

bkamins:
Fix #1889
@grahamgill - in tests you can see what will be allowed and what will be still disallowed.

### DIFF
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -186,3 +186,10 @@ Note that `sdf[!, col] .= v` and `sdf[!, cols] .= v` syntaxes are not allowed as
 
 If column indexing using `Symbol` names in `cols` is performed, the order of columns in the operation is specified
 by the order of names.
+
+The `df[!, col] .= v` syntax follows several convinence special rules:
+* if `ncol(df) == 0` then it is allowed to add a column `v` as a freshly allocated column `col`;
+  the length of this column is equal to `length(v)` if `v` is a vector and `1` if it is a scalar or
+  a scalar broadcasting operation;
+* if `ncol(df) > 0` and `nrow(df) == 0` then it is allowed to add a column only if `v` is a scalar;
+  the length of this column is `0` and type the same as `typeof(v)`;
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -129,7 +129,7 @@ In particular a description explicitly mentions if the assignment is *in-place*.
                       (with the exception that if `v` is an `AbstractRange` it gets converted to a `Vector`);
                       also if `col` is a `Symbol` that is not present in `df` then a new column in `df` is created and holds `v`;
                       equivalent to `df.col = v` if `col` is a valid identifier;
-                      this operation is allowed if the following condition is met `ncol(df) == 0 || length(v) == nrow(df)`;
+                      this is allowed if `ncol(df) == 0 || length(v) == nrow(df)`;
 * `df[!, cols] = v` -> is currently disallowed, but is planned to be supported in the future;
 
 Note that only `df[!, col] = v` and `df.col = v` can be used to add a new column to a `DataFrame`.
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -129,6 +129,7 @@ In particular a description explicitly mentions if the assignment is *in-place*.
                       (with the exception that if `v` is an `AbstractRange` it gets converted to a `Vector`);
                       also if `col` is a `Symbol` that is not present in `df` then a new column in `df` is created and holds `v`;
                       equivalent to `df.col = v` if `col` is a valid identifier;
+                      this operation is allowed if the following condition is met `ncol(df) == 0 || length(v) == nrow(df)`;
 * `df[!, cols] = v` -> is currently disallowed, but is planned to be supported in the future;
 
 Note that only `df[!, col] = v` and `df.col = v` can be used to add a new column to a `DataFrame`.
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -172,7 +172,7 @@ In such an operation `AbstractDataFrame` is considered as two-dimensional and `D
     `DataFrameRow` is considered to be column-oriented.
 
 Additional rules:
-* in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` syntaxes we broadcast `v` into the contents of `df[row, col]` (this is consistent with Base behavior);
+* in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` syntaxes `v` is broadcasted into the contents of `df[row, col]` (this is consistent with Julia Base);
 * in the `df[row, cols] .= v` syntaxes the assignment to `df` is performed in-place;
 * in the `df[rows, col] .= v` and `df[rows, cols] .= v` syntaxes the assignment to `df` is performed in-place;
 * in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector; if `col` is `Symbol` and it is missing from `df` then a new column is added; the length of the column is always the value of `nrow(df)` before the assignment takes place;
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -172,7 +172,8 @@ In such an operation `AbstractDataFrame` is considered as two-dimensional and `D
     `DataFrameRow` is considered to be column-oriented.
 
 Additional rules:
-* in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` and `df[row, cols] .= v` syntaxes the assignment to `df` is performed in-place;
+* in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` syntaxes we broadcast `v` into the contents of `df[row, col]` (this is consistent with Base behavior);
+* in the `df[row, cols] .= v` syntaxes the assignment to `df` is performed in-place;
 * in the `df[rows, col] .= v` and `df[rows, cols] .= v` syntaxes the assignment to `df` is performed in-place;
 * in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector; if `col` is `Symbol` and it is missing from `df` then a new column is added;
 * `df[!, cols] = v` syntax is currently disallowed, but is planned to be supported in the future;
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -175,7 +175,7 @@ Additional rules:
 * in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` syntaxes we broadcast `v` into the contents of `df[row, col]` (this is consistent with Base behavior);
 * in the `df[row, cols] .= v` syntaxes the assignment to `df` is performed in-place;
 * in the `df[rows, col] .= v` and `df[rows, cols] .= v` syntaxes the assignment to `df` is performed in-place;
-* in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector; if `col` is `Symbol` and it is missing from `df` then a new column is added;
+* in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector; if `col` is `Symbol` and it is missing from `df` then a new column is added; the length of the column is always the value of `nrow(df)` before the assignment takes place;
 * `df[!, cols] = v` syntax is currently disallowed, but is planned to be supported in the future;
 * `df.col .= v` syntax is allowed and performs in-place assignment to an existing vector `df.col`.
 * in the `sdf[CartesianIndex(row, col)] .= v`, `sdf[row, col] .= v` and `sdf[row, cols] .= v` syntaxes the assignment to `sdf` is performed in-place;
@@ -187,10 +187,3 @@ Note that `sdf[!, col] .= v` and `sdf[!, cols] .= v` syntaxes are not allowed as
 
 If column indexing using `Symbol` names in `cols` is performed, the order of columns in the operation is specified
 by the order of names.
-
-The `df[!, col] .= v` syntax follows several convinence special rules:
-* if `ncol(df) == 0` then it is allowed to add a column `v` as a freshly allocated column `col`;
-  the length of this column is equal to `length(v)` if `v` is a vector and `1` if it is a scalar or
-  a scalar broadcasting operation;
-* if `ncol(df) > 0` and `nrow(df) == 0` then it is allowed to add a column only if `v` is a scalar;
-  the length of this column is `0` and type the same as `typeof(v)`;
--- a/src/other/broadcasting.jl
+++ b/src/other/broadcasting.jl
@@ -84,7 +84,7 @@ end
 # in this case we inherit it from the right hand side
 Base.axes(x::LazyNewColDataFrame) = (Base.OneTo(nrow(x.df)),)
 
-function Base.materialize!(dest::LazyNewColDataFrame,
+function Base.Broadcast.materialize!(dest::LazyNewColDataFrame,
                            bc::Base.Broadcast.Broadcasted{Style}) where {Style}
     ibc = Base.Broadcast.instantiate(bc)
     if length(axes(ibc)) == 1 && length(axes(ibc)[1]) == 0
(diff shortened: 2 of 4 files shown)


bkamins:
I have added new broadcasting rules for an empty `DataFrame` now. Please comment if this is what was expected. Tests are appreciated as the requested functionality is tricky (also I have updated the docs so you can check if what we have is what you expected).

itsdfish:
As far as I can tell, everything looks great. Thanks for taking the time to add those changes! I think this makes DataFrames more flexible.

bkamins:
@nalimilan + @quinnj : you might find the last commit interesting, see https://github.com/JuliaData/DataFrames.jl/pull/1890/commits/e6931245ae939ff3c5738ec7650efa8975287796

I also checked that `col_tmp` when returned from this function passes `col_tmp isa AbstractVector` test, so something bad happens when the function is compiled probably (most likely type inference). The problem arises when `col_tmp` is a `categorical([1,2,missing])` vector. It is fully reproducible by checking out this repo without the last commit.

The problem is present in Julia 1.0 only. Do you have any idea what could cause it?

CC @vtjnash - it seems to me like a bug in Julia 1.0.

bkamins:
Just as a reference this is a MWE of what was not working under Julia 1.0:
```
using DataFrames
v = categorical([1,2, missing])
df = DataFrame(reshape(1.5:15.5, (3,5)))
df[!, :c1] .= v
```

bkamins:
Fixes #1893.
An additional fix for `df[1,1] .=v`, `df[1,:col] .=v` and `df[CartesianIndex(1,1)] .=v` as it was not consistent with Base.

bkamins:
@nalimilan This should be good for a review. I have gotten rid of "speculative code (the possible bug in Base was worked around by a cleaner code). These are the only problems reported with 0.19 that needed fixing and are:
* treat `DataFrame()` as an object with undefined number of rows in broadcasting and create as many rows as the right hand side specifies; in particular one row if it is a scalar (this is consistent with the old behavior of DataFrames.jl and current behavior in non-broadcasting `df[!, :col] = v` which does not copy as opposed to broadcasting which copies).
* allow 0-row and multi-column `DataFrame` to be broadcasted into using a scalar like `df[!, :col] .= scalar` by creating a new column with 0 rows
* treat `df[1,1] .= v` as broadcasting into the contents of the cell (fixed, this is consistent with Base)

nalimilan:
>     * treat `DataFrame()` as an object with undefined number of rows in broadcasting and create as many rows as the right hand side specifies; in particular one row if it is a scalar (this is consistent with the old behavior of DataFrames.jl and current behavior in non-broadcasting `df[!, :col] = v` which does not copy as opposed to broadcasting which copies).

I'm not sure that's a great idea. That would introduce an inconsistency between 0-rows data frames depending on whether they have 0 columns or more. I think it's fine to allow for some flexibility for 0-column data frames in cases that would otherwise throw errors; but introducing inconsistencies is problematic. And indeed it makes the description of the rules more complex.

>     * allow 0-row and multi-column `DataFrame` to be broadcasted into using a scalar like `df[!, :col] .= scalar` by creating a new column with 0 rows

This sounds OK as long as 0-column and multi-column data frames behave consistently.

>     * treat `df[1,1] .= v` as broadcasting into the contents of the cell (fixed, this is consistent with Base)

+1

bkamins:
@nalimilan - your comments are along my initial thoughts. Let me give an extended explanation of the reason for the change and I would appreciate your feedback.

The problem is that: 1) users want what I have implemented here; 2) currently in `setindex!` behaves this way. What I mean is that you can do:
```
df = DataFrame()
df[!, :a] = v
```
and `v` can have any number of rows you like. And this first column then defines the `nrow` for a `DataFrame` afterwards. So - in `setindex!` actually we treat `0`-column `DataFrame` differently than a `DataFrame` with some columns. And what users want is to keep this behavior also for broadcasted assignment (in other words - the rules are longer, but are consistent between `setindex!` and broadcasted assignment).

Also in the past `df.col = 1` worked for `0` column data frames and produced one row. Now we disallow this and this is breaking (this is not end of the world but still ...).

nalimilan:
> 2) currently in `setindex!` behaves this way. What I mean is that you can do:
> 
> ```
> df = DataFrame()
> df[!, :a] = v
> ```
> 
> and `v` can have any number of rows you like. And this first column then defines the `nrow` for a `DataFrame` afterwards. So - in `setindex!` actually we treat `0`-column `DataFrame` differently than a `DataFrame` with some columns. And what users want is to keep this behavior also for broadcasted assignment (in other words - the rules are longer, but are consistent between `setindex!` and broadcasted assignment).

AFAICT that's a very different thing, since it doesn't conflict with the general behavior when there are columns: we just allow something that would otherwise fail (because the length of the new vector isn't `0`). But this PR proposes to introduce an inconsistency: broadcasting into a non-existent column in a zero-row data frame would give a zero-row data frame if it has columns, but a data frame with `length(v)` rows otherwise. That sounds really problematic to me.

> Also in the past `df.col = 1` worked for `0` column data frames and produced one row. Now we disallow this and this is breaking (this is not end of the world but still ...).

Yes, that's a bit annoying. But `df.col = [1]` is quite short and efficient. Maybe we can point to that as a replacement (and avoid throwing an error)?

bkamins:
>  But df.col = [1] is quite short and efficient

I agree (and that is why initially I wanted to create `0` rows always even for a `DataFrame` with `0` columns).

@itsdfish + @grahamgill : can you please comment how much you want 0-column data frame to be special and why? We have a tension here between consistency (which was our original design intention) and usability (this is what you wanted).

itsdfish:
@bkamins, from time to time, I find myself adding scalars to empty DataFrames when restructuring data. Wrapping scalars with `[]` isn't too bad. The main downside is that it cannot handle cases  such as df[!,:col] .=v`, where v could be a vector or scalar. Nonetheless, I would recommend doing whatever provides a simple and consistent interface.

bkamins:
@itsdfish thank you for the feedback.

I think we can wait one or two days for other feedback and then decide. If there are no new voices I will implement the following rule:

> when broadcasting into a data frame `df` the target number of rows is always `nrow(df)`

@nalimilan I guess this is the invariant you wanted - right?. This will mean that we cannot broadcast vectors of positive length into `0`-row data frames (no matter if they have columns or not) and we allow to broadcast scalars taking their type to create a column.
Actually we could allow broadcasting 0-length arrays also (Base allows this).

For the `setindex!` the rule is simpler:

> when using `df.col = v` you always get the vector of `length(v)` as a result; this has to be equal to `nrow(df)` if `ncol(df)>0`

bkamins:
@nalimilan - I have slept over this and I am still unsure what is best 😢 (I include both `setindex!` and broadcasting cases as I think it is relevant what we do in both). Here are the options. Can you please comment on them? Thank you

operation                                     | current | this PR                    | @nalimilan
----------------------------------------------|---------|----------------------------|------------
`df = DataFrame(); df[!, :a] .= 1`;           | fails   | new `[1]` col              | 
`df = DataFrame(); df[!, :a] .= sin.(1)`;     | fails   | new `[sin(1)]` col         | 
`df = DataFrame(); df[!, :a] .= [1,2]`;       | fails   | new `[1,2]` col            | 
`df = DataFrame(); df[!, :a] .= []`;          | fails   | fails - I can make it work | 
`df = DataFrame(); df[!, :a] .= sin.([])`;    | fails   | fails - it must fail       | 
`df = DataFrame(x=[]); df[!, :a] .= 1`;       | fails   | new `Int[]` col            | 
`df = DataFrame(x=[]); df[!, :a] .= sin.(1)`; | fails   | fails - I can make it work | 
`df = DataFrame(x=[]); df[!, :a] .= [1,2]`;   | fails   | fails                      | 
`df = DataFrame(x=[]); df[!, :a] .= []`;      | fails   | fails - I can make it work | 
`df = DataFrame(); df[!, :a] .= sin.([])`;    | fails   | fails - and must fail      | 
`df = DataFrame(); df[!, :a] = 1`;            | fails   | fails                      | 
`df = DataFrame(); df[!, :a] = [1,2]`;        | works   | works                      | 
`df = DataFrame(); df[!, :a] = []`;           | works   | works                      | 
`df = DataFrame(x=[]); df[!, :a] = 1`;        | fails   | fails                      | 
`df = DataFrame(x=[]); df[!, :a] = [1,2]`;    | fails   | fails                      | 
`df = DataFrame(x=[]); df[!, :a] = []`;       | works   | works                      |

nalimilan:
I'd suggest `df = DataFrame(); df[!, :a] .= v` and `df = DataFrame(x=[]); df[!, :a] .= v` to be consistent with `x = []; x .= v`, i.e. create an empty column. There wouldn't be any difference between `DataFrame()` and `DataFrame(x=[])` -- the idea being that the former should support all cases that the latter supports (and possibly a few more for convenience, but never less). I've filled the table.

I'd also be OK with throwing errors when attempting to use `.=` to broadcast to 0 rows (i.e. the current status), as it's probably not very useful and possibly confusing, and as it can be added later if needed.

bkamins:
@nalimilan Thank you for a prompt response. You are a robot 😄. Here are my comments to your suggestions (retaining your proposal, but showing the problematic cases):

```
df = DataFrame(x=[]); df[!, :a] .= [1,2];
```
must fail (it is a `DimensionMismatch`), which means that also under your proposal:
```
df = DataFrame(); df[!, :a] .= [1,2];
```
also should fail.

So essentially we have a choice:
* disallow broadcasting into `0`-rows data frame (current behavior)
* allow broadcasting into `0`-rows data frame (no matter if it has or does not have columns) and create a `0`-row column as long as RHS has length `1` and up to 1 dimension.

The last condition (length and dimensionality) is relevant as this is what Base currently allows you to do:
```
julia> x = []
0-element Array{Any,1}

julia> x .= rand()
0-element Array{Any,1}

julia> x
0-element Array{Any,1}

julia> x .= rand(1)
0-element Array{Any,1}

julia> x
0-element Array{Any,1}

julia> x .= rand(1, 1)
```

bkamins:
@itsdfish + @grahamgill:
we have distilled down the choice to two options as noted above. In short: `0`-rows is not allowed or it always creates `0`-rows vector if this is what Base would allow.

I am leaning towards the second option (this is consistent with Base).

You should get all you wanted initially except. For `0`-column `DataFrame`:

* `df[!, :col] .= [1,2]` will fail; you should write `df[!, :col] = [1,2]` to be non-copying and `df[!, :col] = copy([1,2])` to copy;
* `df[!, :col] .= [1]` will create a `0`-length vector (not `1`-length vector); the recommended syntax to add the vector you wanted is as above;

(also @nalimilan - just please confirm that this is exactly what we want 😄 as this is tricky, because it will lead to `setindex!` and broadcasted assignment inconsistency)

bkamins:
@nalimilan - just to expand on my last comment. `DataFrame()` object is IMO more common than `0`-row `DataFrame` with some columns. Actually a behavior currently implemented (allowing you do to `df[!, :col] .= [1]` creating a copy of `[1]` is more convenient than creating a `0`-row new column) so we could consider having 0-column `DataFrame` to behave specially.

itsdfish:
@bkamins, I'll defer to your recommendation. That looks good.

grahamgill:
> > But df.col = [1] is quite short and efficient
> 
> I agree (and that is why initially I wanted to create `0` rows always even for a `DataFrame` with `0` columns).
> 
> @itsdfish + @grahamgill : can you please comment how much you want 0-column data frame to be special and why? We have a tension here between consistency (which was our original design intention) and usability (this is what you wanted).

If a new column is created by broadcasting a scalar into a data frame with existing columns, then the new column necessarily has as many rows as the others. If that can also work when there are 0 rows, that's what interests me, because it eliminates a lot of edge case checking. If the data frame has both 0 columns and 0 rows, then I'd prefer consistency, with the resulting data frame also having 0 rows.

grahamgill:
Thanks for the summary. I prefer the second option. I didn't realise I was stirring the pot so much with my initial question in #1889 

> @itsdfish + @grahamgill:
> we have distilled down the choice to two options as noted above. In short: `0`-rows is not allowed or it always creates `0`-rows vector if this is what Base would allow.
> 
> I am leaning towards the second option (this is consistent with Base).
> 
> You should get all you wanted initially except. For `0`-column `DataFrame`:
> 
> * `df[!, :col] .= [1,2]` will fail; you should write `df[!, :col] = [1,2]` to be non-copying and `df[!, :col] = copy([1,2])` to copy;
> * `df[!, :col] .= [1]` will create a `0`-length vector (not `1`-length vector); the recommended syntax to add the vector you wanted is as above;
> 
> (also @nalimilan - just please confirm that this is exactly what we want 😄 as this is tricky, because it will lead to `setindex!` and broadcasted assignment inconsistency)

bkamins:
Thank you both for the feedback. Really appreciated.

@nalimilan - so we go your way I think broadcasting to `isempty(df)` always produces `0` rows. It will actually make this PR simpler (I do not have to mess with tricky corner cases). Just please confirm the corner cases I highlighted before I change this PR.

nalimilan:
Sounds good!

bkamins:
I was talking with @mbauman about this and what I will implement is:
1. assume that if `isempty(df) == true` then we always treat it as a 0 number of rows;
2. I will develop it in a way that it is 1-to-1 with Base (actually the behavior in Base might change here in the future; I will make the code in a way that will automatically keep track of Base behavior);

Thanks to all who contributed to the discussion.

bkamins:
Broadcasting rules: Round 4. (and I hope the last round).

What we do:
1. broadcasting always takes `nrow(df)` before the operation to determine the number of rows to create
2. we use exactly the same rules as Base what we accept or what we do not accept (and if something changes there we will synchronize automatically)

(and as a side effect - we need less code to express this)

@nalimilan - this should be good for a final review.

src/other/broadcasting.jl
```diff
@@ -78,7 +78,8 @@ struct LazyNewColDataFrame{T}
     col::T
 end
 
-# we allow LazyNewColDataFrame only for data frames with at least one column
+# we allow LazyNewColDataFrame also for data frames
+# that are empty, ie. `nrow(df) == 0`; in this case we create a 0 length column
```
nalimilan: This comment looks weird when you don't know the history: why wouldn't it be allowed for 0-row data frames? I'd just remove it, or put it elsewhere.

test/broadcasting.jl
```diff
@@ -615,11 +668,22 @@ end
...
+    @test eltype(df.b) == Char
+    @test names(df) == [:a, :b]
+
+    c = categorical(["a", "b", "c"])
+    df = DataFrame()
+    @test_throws DimensionMismatch df[!, :a] .= c
+
+    df[!, :b] .= c[1]
+    @test nrow(df) == 0
+    @test df.b isa CategoricalVector
```
```suggestion
    @test df.b isa CategoricalVector{String}
```

src/other/broadcasting.jl
bkamins: fixed

bkamins:
CI passes (we have only Coverage decrease as usual 😞)

docs/src/lib/indexing.md
```diff
@@ -129,6 +129,7 @@ In particular a description explicitly mentions if the assignment is *in-place*.
                       (with the exception that if `v` is an `AbstractRange` it gets converted to a `Vector`);
                       also if `col` is a `Symbol` that is not present in `df` then a new column in `df` is created and holds `v`;
                       equivalent to `df.col = v` if `col` is a valid identifier;
+                      this operation is allowed if the following condition is met `ncol(df) == 0 || length(v) == nrow(df)`;
```
```suggestion
                      this is allowed if `ncol(df) == 0 || length(v) == nrow(df)`;
```

bkamins:
Thank you. I am going to merge it today and make a release (unless you see that we should wait - e.g. for https://github.com/JuliaData/DataFrames.jl/pull/1887).

nalimilan:
Fine with me.

PR #1899, opened 2019-07-26
target setindex implementation

bkamins:
This is currently WIP (I have initially implemented new rules for things that we need to change to finish deprecations). It is not finalized yet, so have a look at it only to check what is added 😄.

Before I write tests and clean this up I would like to clarify what we do with `df[!, cols] = v` and `df[!, cols] .= v` syntax.

The particular, crucial, question is if we want to allow them to create columns (like `df[!, col]` allows) or not. If yes then the question is if we allow adding columns only at the end of a data frame or also reorder the columns already present.

The relevant case is for example:
```
df = DataFrame(a = [1,2], b=[3,4])
df[!, [:a,:c,:b]] .= 10
```
Do we want column `:c` to be created, and if yes should it be a last column of `df` (probably yes).

My problem is that the corner-cases will be surprising for users, so, as opposed to single column assignment, maybe it should be disallowed to add columns (but maybe you feel it is OK).


nalimilan:
I'd say it's fine to be restrictive for now. In particular, creating columns while providing names in an order different from the actual order sounds really weird. Let's start with a simple system, and possibly extend it later if we get requests.

src/dataframerow/dataframerow.jl
```diff
@@ -98,6 +98,25 @@ Base.@propagate_inbounds Base.getindex(r::DataFrameRow, idxs::Union{AbstractVect
...
+                        row_ind::Integer,
+                        col_inds::Union{AbstractVector, Regex, Not, Colon})
+    idxs = index(df)[col_inds]
+    if length(v) != length(idxs)
+        throw(DimensionMismatch("$(length(idxs)) columns were selected and the assigned" *
+                                " value contains $(length(v)) elements")
+    end
+
+    if !(v isa AbstractDict || all(((a, b),) -> a == b, zip(view(_names(df), idxs), keys(v))))
+        throw(ArgumentError("Selected column names do not match the names in assigned value"))
```
nalimilan: Print the names?

docs/src/lib/indexing.md
```diff
@@ -109,15 +109,12 @@ then view points to selected columns by their number at the moment of creation o
...
-
-In the current release of DataFrames.jl we are in the transition period when an old, undocumented, behavior
-of `setindex!` is still supported, but throws deprecation warnings.
-
-The behavior described below will be fully implemented in the next major release of DataFrames.jl.
+The following list specifies the behavior of `setindex!` operations depending on argument types.
 
 In particular a description explicitly mentions if the assignment is *in-place*.
 
+Note that `setindex!` operations are not atomic.
```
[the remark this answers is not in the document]
bkamins: this means that if `setindex!` operates and fails "halfway" it will throw an error and possibly leave garbage in target `DataFrame` (it does not clean-up as `push!` or `append!` now do)

docs/src/lib/indexing.md
```diff
@@ -150,7 +147,9 @@ Note that `sdf[!, col] = v`, `sdf[!, cols] = v` and `sdf.col = v` are not allowe
 * `dfr[col] = v` -> set value of `col` in row `row` to `v` in-place;
                     equivalent to `dfr.col = v` if `col` is a valid identifier;
 * `dfr[cols] = v` -> set values of entries in columns `cols` in `dfr` by elements of `v` in place;
-                     `v` can be an `AbstractVector` or `v` can be a `NamedTuple` or `DataFrameRow` when column names must match;
+                     `v` can be an iterable in which case it must have a number of elements equal to `length(dfr)`,
+                     or `v` can be a `NamedTuple`, `AbstractDict` or `DataFrameRow` when column names must match
```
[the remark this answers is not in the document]
bkamins: OK

bkamins:
So the "restrictive" approach is to disallow adding columns using `df[!, cols]` on LHS - right?

src/dataframerow/dataframerow.jl
bkamins: OK

docs/src/lib/indexing.md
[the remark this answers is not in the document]
bkamins: I will change the description

src/dataframerow/dataframerow.jl
bkamins: I will print column numbers in which there is no match (as otherwise it will get very verbose).

bkamins:
@nalimilan - this should be good to have a look at when you have some free time

test/deprecated.jl
```diff
@@ -1,10 +1,12 @@
 module TestDeprecated
 
-using Test, DataFrames, Random
+using Test, DataFrames, Random, Logging
 import DataFrames: identifier
 
 const ≅ = isequal
 
+old_logger = global_logger(NullLogger())
```
nalimilan: Is this really needed in this file (it's not modified AFAICT)?

test/indexing.jl
```diff
@@ -1056,7 +1179,163 @@ end
...
+    end
+
+    for rows in [:, 1:3], cols in [:, r"", Not(r"xx"), 1:4]
+        df = DataFrame(ones(3,4))
+        df[rows, cols] = reshape(1:12, 3, :)
+        @test df == DataFrame(reshape(1:12, 3, :))
+    end
+end
+
+@testset "old setindex! tests adjusted to new rules" begin
```
nalimilan: This description is a bit weird. Just keep what you think it useful, and drop the old ones.

test/reshape.jl
```diff
@@ -145,9 +145,10 @@ end
                    variable=["a", "b", "a", "b"], value=[3, 4, 5, 6])
     @test_logs (:warn, "Duplicate entries in unstack at row 3 for key 1 and variable a.") unstack(df, :id, :variable, :value)
     @test_logs (:warn, "Duplicate entries in unstack at row 3 for key (1, 1) and variable a.") unstack(df, :variable, :value)
-    a = unstack(df, :id, :variable, :value)
+    a, b = with_logger(NullLogger()) do
```
nalimilan: Why is this needed?

bkamins:
> I guess there are ambiguities if you don't use a loop? That sounds too bad.

Yes - unfortunately if we want to keep the deprecations it is much easier to do it this way. The problem is, as discussed some time ago that using `Union` is less specific than generating methods for given types.

After the deprecations are removed from code base, we can revert to the standard `Union` approach.

src/dataframe/dataframe.jl
```diff
@@ -498,51 +498,114 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Integer, col_ind::Column
...
+                                  mx::AbstractMatrix,
+                                  row_inds::typeof(!),
+                                  col_inds::$(T))
+        idxs = index(df)[col_inds]
+        if size(mx, 2) != length(idxs)
+            throw(DimensionMismatch("number of selected columns ($(length(idxs))) and number of columns in" *
+                                    " matrix ($(size(mx, 2))) do not match"))
+        end
+        for (j, col) in enumerate(idxs)
+            df[!, col] = mx[:, j]
```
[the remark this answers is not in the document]

src/dataframerow/dataframerow.jl
```diff
@@ -98,14 +98,39 @@ Base.@propagate_inbounds Base.getindex(r::DataFrameRow, idxs::Union{AbstractVect
...
+                                  v::Union{DataFrameRow, NamedTuple, AbstractDict},
+                                  row_ind::Integer,
+                                  col_inds::$(T))
+        idxs = index(df)[col_inds]
+        if length(v) != length(idxs)
+            throw(DimensionMismatch("$(length(idxs)) columns were selected but the assigned" *
+                                    " value contains $(length(v)) elements"))
+        end
+
+        if !(v isa AbstractDict || all(((a, b),) -> a == b, zip(view(_names(df), idxs), keys(v))))
```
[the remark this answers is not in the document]

src/deprecated.jl
```diff
@@ -1351,16 +1351,10 @@ import Base: delete!, insert!, merge!
...
 @deprecate merge!(df1::DataFrame, df2::AbstractDataFrame) (foreach(col -> df1[!, col] = df2[!, col], names(df2)); df1)
 
-import Base: setindex!
-@deprecate setindex!(df::DataFrame, x::Nothing, col_ind::Int) select!(df, Not(col_ind))
-
 import Base: map
 @deprecate map(f::Function, sdf::SubDataFrame) f(sdf)
 @deprecate map(f::Union{Function,Type}, dfc::DataFrameColumns{<:AbstractDataFrame, Pair{Symbol, AbstractVector}}) mapcols(f, dfc.df)
 
-import Base: length
```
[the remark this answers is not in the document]
bkamins: It has been deprecated for a long time. The problem is that some methods (setting a `DataFrameRow` and `push!`) in DataFrames.jl find that `DataFrame` has `length` and then try to iterate over it producing not very clear error messages.

I would remove this if you agree.

src/other/broadcasting.jl
```diff
@@ -229,28 +225,22 @@ function Base.copyto!(df::AbstractDataFrame, bc::Base.Broadcast.Broadcasted{<:Ba
...
-#         wrongnames = setdiff(union(colnames...), intersect(colnames...))
-#         msg = join(wrongnames, ", ", " and ")
-#         throw(ArgumentError("Column names in broadcasted data frames must match. " *
-#                             "Non matching column names are $msg"))
-#     end
+function Base.copyto!(crdf::ColReplaceDataFrame, bc::Base.Broadcast.Broadcasted)
+    bcf = Base.Broadcast.flatten(bc)
+    colnames = unique([_names(x) for x in bcf.args if x isa AbstractDataFrame])
+    if length(colnames) > 1 || (length(colnames) == 1 && view(_names(crdf.df), crdf.cols) != colnames[1])
+        wrongnames = setdiff(union(colnames...), intersect(colnames...))
```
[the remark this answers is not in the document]
bkamins: Fixed

src/other/broadcasting.jl
```diff
@@ -229,28 +225,22 @@ function Base.copyto!(df::AbstractDataFrame, bc::Base.Broadcast.Broadcasted{<:Ba
...
-#         startcol = similar(Vector{typeof(v1)}, nrows)
-#         startcol[1] = v1
-#         col = copyto_widen!(startcol, bcf′, 2, i)
-#         df.df[!, colidx] = col
-#     end
-#     df.df
-# end
+    bcf′ = Base.Broadcast.preprocess(crdf, bcf)
+    for (i, col) in enumerate(crdf.cols)
+        crdf.df[!, col] = [bcf′[CartesianIndex(j, i)] for j in axes(crdf.df, 1)]
```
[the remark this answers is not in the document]
bkamins: I have improved it for the `DataFrame` special case. Also I have fixed a bug in line 30 of this file. Thank you for asking for this.

test/broadcasting.jl
```diff
@@ -1345,4 +1347,69 @@ end
     @test df == DataFrame(a=[[1000,1000,1000]])
 end
 
+@testset "broadcasting into df[!, cols]" begin
+    for selector in [1:2, Between(:x1, :x2), Not(r"xx"), [:x1, :x2]]
+        df = DataFrame(x1=1:3, x2=4:6)
+        df[!, selector] .= "a"
+        @test df == DataFrame(fill("a", 3, 2))
```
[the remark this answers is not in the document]
bkamins: I have added some more tests.

test/dataframe.jl
```diff
@@ -144,78 +144,105 @@ end
 end
 
 @testset "push!(df, row)" begin
-    df=DataFrame( first=[1,2,3], second=["apple","orange","pear"] )
+    buf = IOBuffer()
+    sl = SimpleLogger(buf)
 
-    dfb= DataFrame( first=[1,2], second=["apple","orange"] )
-    dfc= DataFrame( first=[1,2], second=["apple","orange"] )
+    df=DataFrame(first=[1,2,3], second=["apple","orange","pear"])
```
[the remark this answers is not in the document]

test/deprecated.jl
bkamins: What do you mean by "it's not modified"?

This line disables outputting deprecation warnings from deprecated.jl.
There are hundreds of them, and they are expected.
I do not think anyone is able to digest them so I disable printing them (if something errors we will catch this as then CI will fail).

The reason of these changes is that after them it is much easier to find real problems in the actual testing code (as if you get an error or a warning it is easy to locate it by searching the logs). Does it make sense (it for sure makes my life easier)?

test/indexing.jl
bkamins: >  Just keep what you think it useful, and drop the old ones.

This is exactly what I did 😄. I renamed it to "additional tests".

test/reshape.jl
bkamins: I want to suppress outputting a warning to the log (we test above that actually the warning is produced) so that later it is easier to find unexpected warnings if they are printed.

bkamins:
This should be good for a second round of a review (probably there will be some comments - I tried to clean up everything, but the PR grew big). Thank you!

src/deprecated.jl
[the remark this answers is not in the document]
nalimilan: OK, that's fine if that creates problems elsewhere.

test/reshape.jl
nalimilan: Ah, cool! It's always annoying to print warnings during normal operation.

src/dataframerow/dataframerow.jl
```diff
@@ -98,14 +98,39 @@ Base.@propagate_inbounds Base.getindex(r::DataFrameRow, idxs::Union{AbstractVect
...
-        Base.depwarn("implicit broadcasting in DataFrameRow assignment is deprecated", :setindex!)
+for T in (:AbstractVector, :Regex, :Not, :Between, :All, :Colon)
+    @eval function Base.setindex!(df::DataFrame,
+                                  v::Union{DataFrameRow, NamedTuple, AbstractDict},
+                                  row_ind::Integer,
+                                  col_inds::$(T))
+        idxs = index(df)[col_inds]
+        if length(v) != length(idxs)
+            throw(DimensionMismatch("$(length(idxs)) columns were selected but the assigned" *
+                                    " value contains $(length(v)) elements"))
```
```suggestion
                                    " collection contains $(length(v)) elements"))
```

src/dataframe/dataframe.jl
```diff
@@ -498,51 +498,97 @@ function Base.setindex!(df::DataFrame, v::Any, row_ind::Integer, col_ind::Column
...
+                                  mx::AbstractMatrix,
+                                  row_inds::$T1,
+                                  col_inds::$T2)
+        idxs = index(df)[col_inds]
+        if size(mx, 2) != length(idxs)
+            throw(DimensionMismatch("number of selected columns ($(length(idxs))) and number of columns in" *
+                                    " matrix ($(size(mx, 2))) do not match"))
+        end
+        for (j, col) in enumerate(idxs)
+            df[row_inds, col] = row_inds isa typeof(!) ? mx[:, j] : view(mx, :, j)
```
nalimilan: Are you sure this is better than this? AFAIK the compiler is quite smart with `===`.
```suggestion
            df[row_inds, col] = (row_inds === !) ? mx[:, j] : view(mx, :, j)
```

src/other/broadcasting.jl
```diff
@@ -50,9 +50,14 @@ function Base.copy(bc::Base.Broadcast.Broadcasted{DataFrameStyle})
     colnames = unique([_names(df) for df in bcf.args if df isa AbstractDataFrame])
```
nalimilan:
```suggestion
    colnames = unique!([_names(df) for df in bcf.args if df isa AbstractDataFrame])
```

src/other/broadcasting.jl
```diff
@@ -204,10 +205,16 @@ function Base.copyto!(df::AbstractDataFrame, bc::Base.Broadcast.Broadcasted)
     bcf = Base.Broadcast.flatten(bc)
     colnames = unique([_names(x) for x in bcf.args if x isa AbstractDataFrame])
```
nalimilan:
```suggestion
    colnames = unique!([_names(x) for x in bcf.args if x isa AbstractDataFrame])
```

src/dataframe/dataframe.jl
bkamins: It is I just did want to avoid this as `(! === row_inds)` is a syntax error. But I am OK with the change

bkamins:
Regarding `setindex!` to a `DataFrameRow` and also `push!`. We have two options:
* leave things as is
* make LHS to be required to be: `Base.Generator`, `Tuple`, `NamedTuple`, `AbstractDict`, `AbstractVector` (or `AbstractArray` for the last, but I prefer `AbstarctVector`)

The `Base.Generator` would cover all generators. Then if we allow `AbstractArray` we would not check dimensionality of a generator and if we only accept `AbstractVector` we would check if `Base.Generator` is unidimensional.

What do you think would be better?

nalimilan:
I'd be fine with restricting as you suggest. But better make a new PR?

bkamins:
But for `setindex!` I have to define it in this PR (as we add this functionality in this PR).

The only question is if we should "math dimensionality" (so accept only `AbstractVector` and 1D generator) or accept all `AbstractArrays` and all generators?

For `push!` it will be a separate PR for sure (as we need to go through deprecation cycle).

nalimilan:
I don't have a strong opinion really.

bkamins:
OK - I will go with what Base allows for Arrays (any dimensionality) then.

bkamins:
@nalimilan - merged. Thank you!

PR #1900, opened 2019-07-26
Change eltype to true eltype in describe

### LINKED ISSUE #1765: `eltype` for `describe` is confusing users
Referencing [this thread](https://discourse.julialang.org/t/pca-not-working/22645/3) on discourse. The user used `describe` and thought their columns were not missing because they weren't listed as such in `eltype`. 

* Would it be possible to return `Union{T, Missing}` with nice printing? 
* We could use a more informative name.

bkamins:
Fixes #1765.

Actually I think that it is better to print true `eltype` not a stripped one. It is a bit more verbose, but is explicit and in this case it is not a problem (someone wants `eltype` in `describe` so one gets it verbatim).

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -435,6 +435,10 @@ function _describe(df::AbstractDataFrame, stats::AbstractVector)
             d[:last] = isempty(col) ? nothing : last(col)
         end
 
+        if :eltype in predefined_funs
+            d[:eltype] = eltype(col)
+        end
+
         return d
     end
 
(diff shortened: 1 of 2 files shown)


pdeffebach:
I had this on my radar a bit and I thought about overriding `show` for `Union{T, Missing}` in all data frames printing to `T⍰` so `describe` would look better. Is that a reasonable change?

bkamins:
We could consider it but this is an orthogonal change (i.e. after this PR we could implement it independently). My only consideration was that this is a special case that would happen very rarely.

Ideally Base could make `Union{T, Missing}` be expressed as `T?` on syntactic level (an open debate for a long time). Then we could consider making:
* printing `missing` as `?` (this actually would be a big space saver) - this would be similar to `Bool` printing as `1`/`0` that we now have in Base
* your proposal of printing types being an union with `Missing`

nalimilan:
Yes, that would be more consistent. But it would be really nice to get a short syntax in Base...

PR #1901, opened 2019-07-26
Add custom show to DataFrameRows and DataFrameColumns

### LINKED ISSUE #1894: show for DataFrameRows should display a table
It would be very nice (a lot more readable) if `show` on `DataFrameRows` would display a table.

Just discussed this with @bkamins at JuliaCon.

bkamins:
Fixes #1894

Please have a look if we like it before I add tests (which is cumbersome).


src/abstractdataframe/io.jl
```diff
@@ -156,6 +156,17 @@ function Base.show(io::IO, mime::MIME"text/html", dfr::DataFrameRow; summary::Bo
...
 end
 
+function Base.show(io::IO, mime::MIME"text/html", dfrs::DataFrameRows; summary::Bool=true)
+    write(io, "<p>DataFrameRows</p>")
+    _show(io, mime, parent(dfrs), summary=summary, rowid=r)
+end
+
+function Base.show(io::IO, mime::MIME"text/html", dfcs::DataFrameColumns{T,V};
+                   summary::Bool=true) where {T,V}
+    write(io, "<p>DataFrameColumns (with names = $(V <: Pair))</p>")
```
nalimilan: Isn't this going to explode when there are many columns? Maybe just print `m×n DataFrameRows`, just like we do for `DataFrame`? Also, pass `summary=false`?

nalimilan: Scratch that, I hadn't realized this is just a Boolean. Though my point still holds about printing the dimensions and passing `summary=false`.

bkamins: Agreed. There are many more things to fix but I just wanted to make sure that we agree on the general approach.

bkamins:
Now it should be good to have a look at.

oschulz:
Oh, nice, thanks!!

bkamins:
CI problems are a typical marginal coverage drop problem.

src/abstractdataframe/io.jl
```diff
@@ -152,8 +152,23 @@ end
...
+    df = parent(dfrs)
+    summary && write(io, "<p>$(nrow(df))×$(ncol(df)) DataFrameRows</p>")
+    _show(io, mime, df, summary=false)
+end
+
+function Base.show(io::IO, mime::MIME"text/html", dfcs::DataFrameColumns{T,V};
+                   summary::Bool=true) where {T,V}
+    df = parent(dfcs)
+    if summary
+        write(io, "<p>$(nrow(df))×$(ncol(df)) DataFrameColumns (with names = $(V <: Pair))</p>")
```
```suggestion
        write(io, "<p>$(nrow(df))×$(ncol(df)) DataFrameColumns (with names=$(V <: Pair))</p>")
```

src/abstractdataframe/io.jl
```diff
@@ -387,6 +410,16 @@ function Base.show(io::IO, mime::MIME"text/tab-separated-values", dfr::DataFrame
     show(io, mime, view(parent(dfr), [r], c))
 end
 
+function Base.show(io::IO, mime::MIME"text/csv",
```
nalimilan: Use short function syntax?

test/show.jl
```diff
@@ -359,6 +359,32 @@ end
         │ 1   │ 3.141592653589793 │"""
 end
 
+@testset "Test using :compact parameter of IOContext" begin
```
nalimilan: I don't see `:compact`?

src/abstractdataframe/io.jl
bkamins: fixed (also in other places in the file)

test/show.jl
bkamins: fixed (copy-paste error)

PR #1903, opened 2019-07-27
add disallowmissing, allowmissing and categorical; fix a bug in categorical!

### LINKED ISSUE #1720: Support disallowmissing
One of the things out of the confusion in #1718 
was that we should have a nonmutating `disallowmissing`  that applies to all abstract dataframes,
including `SubDataFrames`

bkamins:
Fixes #1720

To make things consistent I add `disallowmissing`, `allowmissing` and `categorical` support.

I also fix a bug in `categorical!` which incorrectly handled `:` passed as column selector.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -31,6 +31,9 @@ The following are normally implemented for AbstractDataFrames:
 * [`dropmissing!`](@ref) : remove rows with missing values in-place
 * [`nonunique`](@ref) : indexes of duplicate rows
 * [`unique!`](@ref) : remove duplicate rows
+* [`disallowmissing`](@ref) : drop support for missing values in columns
+* [`allowmissing`](@ref) : add support for missing values in columns
+* [`categorical`](@ref) : change column types to categorical
 * `similar` : a DataFrame with similar columns as `d`
 * `filter` : remove rows
 * `filter!` : remove rows in-place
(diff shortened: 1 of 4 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,3 +1283,168 @@ julia> ncol(df)
 
 """
 (nrow, ncol)
+
+"""
+    disallowmissing(df::AbstractDataFrame,
+        cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:)
+
+Create a new `DataFrame` in which columns `cols` of data frame `df` are converted
```
```suggestion
Return a copy of data frame `df` with columns `cols` converted
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,3 +1283,168 @@ julia> ncol(df)
...
+2×1 DataFrame
+│ Row │ a     │
+│     │ Int64 │
+├─────┼───────┤
+│ 1   │ 1     │
+│ 2   │ 2     │
+```
+"""
+function Missings.disallowmissing(df::AbstractDataFrame,
+                         cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:)
```
nalimilan: Wrong indentation. Same below.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,3 +1283,168 @@ julia> ncol(df)
...
+    idxcols = Set(index(df)[cols])
```
nalimilan: Why not this?
```suggestion
    idxcols = unique!(index(df)[cols])
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,3 +1283,168 @@ julia> ncol(df)
...
+        end
+    end
+    DataFrame(newcols, _names(df), copycols=false)
+end
+
+"""
+    allowmissing(df::AbstractDataFrame,
+        cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:)
+
+Create a new `DataFrame` in which columns `cols` of data frame `df` are converted
```
```suggestion
Return a copy of data frame `df` with columns `cols` converted
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,3 +1283,168 @@ julia> ncol(df)
...
+    end
+    DataFrame(newcols, _names(df), copycols=false)
+end
+
+"""
+    categorical(df::AbstractDataFrame,
+                cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:;
+                compress::Bool=false)
+
+Create a new `DataFrame` and change columns selected by `cols` from `df`
```
```suggestion
Return a copy of data frame `df` with columns `cols` converted
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,3 +1283,168 @@ julia> ncol(df)
...
+        else
+            push!(newcols, copy(x))
+        end
+    end
+    DataFrame(newcols, _names(df), copycols=false)
+end
+
+"""
+    categorical(df::AbstractDataFrame,
+                cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:;
```
nalimilan: This isn't completely correct, since passing `:` explicitly gives a different result. Maybe (in another PR) we should change that, and require people to pass `categorical(df, String)` or `categorical(df, Union{Missing,String})` explicitly to get the current behavior?

src/abstractdataframe/abstractdataframe.jl
bkamins: `Set` will be faster later (we repeatedly do `i in idxcols` check) and `unique!` will fail if `cols` is a `ColumnIndex` (a single column selector not wrapped in a vector)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,3 +1283,168 @@ julia> ncol(df)
 
 """
 (nrow, ncol)
+
+"""
+    disallowmissing(df::AbstractDataFrame,
+        cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:)
```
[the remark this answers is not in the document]
bkamins: fixed (I thought to use 4 spaces, but you are right)

src/abstractdataframe/abstractdataframe.jl

src/abstractdataframe/abstractdataframe.jl
bkamins: I have fixed the docs to make it clearer (and I intentionally put the difference between `:` and no argument in the Examples section earlier).

Now the question is if we find the behavior of `categorical(df)` useful. I find it OK (i.e. when doing e.g. GLM later this is what you want to do) so I think we can leave it as is. But I guess it was you who created the initial implementation - right - so I do not want to override your intentions here 😄.

If we found `categorical(df)` unwanted (as you suggest) then I would add a deprecation warning in this PR and add a keyword argument that would allow the user to specify which types should be converted (but I think it is too complex - almost no one would use it, and the default behavior that we currently have seems to fit 95% of cases and if someone wants something else one simply will pass a list of columns t convert).

bkamins:
@nalimilan - thank you for having a look at this PR.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Fine with me. But better not show `cols=:`, and list two signatures instead, otherwise it's really misleading (the real value would be `[eltype(col) <: Union{AbstractString, Missing} for col in eachcol(df)]`).

src/abstractdataframe/abstractdataframe.jl
nalimilan: Still not AFAICT?

src/abstractdataframe/abstractdataframe.jl
bkamins: This is exactly what I did (see the PR after I pushed).

bkamins:
Now it should be good - but please kindly have a look (sorry for the mess).

bkamins:
OK - I will merge it when CI passes and I go through it without noticing anything.

PR #1906, opened 2019-08-02
Fix setindex! deprecation message

bkamins:
Make it more specific for 0-column data frames.


bkamins:
When I made this deprecation in 0.19 this was not possible (we allowed broadcasting into 0-row data frames 0.19.1 and I forgot to review depwarn messages).

PR #1911, opened 2019-08-05
Update documentation to reflect ! indexing

### LINKED ISSUE #1910: Column of dataframe syntax change?
When referring to columns of a dataframe, the getting started guide of the documentation says
"Columns can be accessed via df.col or df[:col]. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name"

However, I'm getting a depreciation warning for that the `df[:col]` syntax. Is this an internal depreciation warning or is the syntax changing but the docs just haven't been updated yet?

Example
```julia 
julia> df = DataFrame(A = 1:3, B = ['a', 'b', 'c'])
3×2 DataFrame
│ Row │ A     │ B    │
│     │ Int64 │ Char │
├─────┼───────┼──────┤
│ 1   │ 1     │ 'a'  │
│ 2   │ 2     │ 'b'  │
│ 3   │ 3     │ 'c'  │

julia> df[:B]
┌ Warning: `getindex(df::DataFrame, col_ind::ColumnIndex)` is deprecated, use `df[!, col_ind]` instead.
│   caller = top-level scope at none:0
└ @ Core none:0
3-element Array{Char,1}:
 'a'
 'b'
 'c
```

bkamins:
Fixes #1910

(I hope I have not missed anything)

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -45,7 +45,7 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
-Columns can be accessed via `df.col` or `df[:col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position.
+Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position.
 
 ```jldoctest dataframe
 julia> df.A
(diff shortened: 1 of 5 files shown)


docs/src/man/getting_started.md
```diff
@@ -45,7 +45,7 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
-Columns can be accessed via `df.col` or `df[:col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position.
+Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position.
```
ararslan:
```suggestion
Columns can be accessed directly (i.e. without copying) via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position.
```
This just reads a little nicer to me.

bkamins:
Thank you. I will merge it tomorrow if there no additional comments.

PR #1914, opened 2019-08-07
Add Between and All

bkamins:
Adds `Between` and `All` column indexing

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -34,7 +34,7 @@ test = ["DataStructures", "DataValues", "Dates", "LaTeXStrings", "Random", "Test
 julia = "1"
 CategoricalArrays = ">= 0.5.4"
 Compat = "2.0.0"
-DataAPI = ">= 1.0.1"
+DataAPI = "1.0.1"
 InvertedIndices = "1"
 IteratorInterfaceExtensions = "0.1.1, 1"
 Missings = ">= 0.2.3"
--- a/Project.toml
+++ b/Project.toml
@@ -34,6 +34,7 @@ test = ["DataStructures", "DataValues", "Dates", "LaTeXStrings", "Random", "Test
 julia = "1"
 CategoricalArrays = ">= 0.5.4"
 Compat = "2.0.0"
+DataAPI = ">= 1.0.1"
 InvertedIndices = "1"
 IteratorInterfaceExtensions = "0.1.1, 1"
 Missings = ">= 0.2.3"
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -525,7 +525,7 @@ end
 
 """
     completecases(df::AbstractDataFrame, cols::Colon=:)
-    completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex, Not})
+    completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex, Not, Between, All})
     completecases(df::AbstractDataFrame, cols::Union{Integer, Symbol})
 
 Return a Boolean vector with `true` entries indicating rows without missing values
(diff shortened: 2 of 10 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1286,7 +1286,7 @@ julia> ncol(df)
 
 """
     disallowmissing(df::AbstractDataFrame,
-                    cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:)
+                    cols::Union{ColumnIndex, AbstractVector, Regex, Between, All, Not, Colon}=:)
```
```suggestion
                    cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon}=:)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1359,7 +1359,7 @@ julia> allowmissing(df)
 ```
 """
 function Missings.allowmissing(df::AbstractDataFrame,
-                               cols::Union{ColumnIndex, AbstractVector, Regex, Not, Colon}=:)
+                               cols::Union{ColumnIndex, AbstractVector, Regex, Between, All, Not, Colon}=:)
```
```suggestion
                               cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon}=:)
```

src/other/index.jl
```diff
@@ -174,9 +174,14 @@ end
...
 end
 
-@inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) = getindex(x, collect(Int, idx))
+@inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) =
+    getindex(x, collect(Int, idx))
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
-@inline Base.getindex(x::AbstractIndex, notidx::Not) = setdiff(1:length(x), getindex(x, notidx.skip))
+@inline Base.getindex(x::AbstractIndex, notidx::Not) =
+    setdiff(1:length(x), getindex(x, notidx.skip))
+@inline Base.getindex(x::AbstractIndex, idx::Between) = collect(x[idx.first]:x[idx.last])
```
nalimilan: I know the existing code does this, but do we really need to call `collect`? Ideally we would just return an `Int` range here above and below.

src/other/index.jl
```diff
@@ -174,9 +174,14 @@ end
...
+@inline Base.getindex(x::AbstractIndex, idx::All) =
+    idx.cols == () ? collect(1:length(x)) : union(getindex.(Ref(x), idx.cols)...)
```
```suggestion
    isempty(idx.cols) ? collect(1:length(x)) : union(getindex.(Ref(x), idx.cols)...)
```
(To avoid recompiling `union` for each number of arguments.)

nalimilan:
Cool. Don't forget to bump the required DataAPI version. X-ref https://github.com/JuliaData/DataAPI.jl/pull/7

src/other/index.jl
bkamins: Good point. We do not need this. I have changed it in places where it is safe (there are two cases in `getindex` where still it is better to `collect`).

bkamins:
@nalimilan OK to merge this?

Project.toml
```diff
@@ -34,6 +34,7 @@ test = ["DataStructures", "DataValues", "Dates", "LaTeXStrings", "Random", "Test
 julia = "1"
 CategoricalArrays = ">= 0.5.4"
 Compat = "2.0.0"
+DataAPI = ">= 1.0.1"
```
nalimilan:
```suggestion
DataAPI = "1.0.1"
```

bkamins:
@nalimilan - I understand that what you proposed now (a strict reference to DataAPI.jl version) is what we should use - right? If you confirm I will merge this. Thank you.

PR #1918, opened 2019-08-11
add type argument to categorical and categorical!

### LINKED ISSUE #1282: Categorical! to include Char
The current "default" behavior is to make `AbstractVector{T} where T <: AbstractString` a `CategoricalArrays.CategoricalVector` and more specifically `CategoricalArrays.CategoricalVector{T} where {T <: CategoricalArrays.CategoricalString, ...}`

However, I believe that a better default would be to include columns that are `eltype(col) <: Char` as well.

bkamins:
Fixes #1282

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -966,13 +966,13 @@ disallowmissing!(df::DataFrame, cols::Colon=:) =
 ##############################################################################
 
 """
+    categorical!(df::DataFrame, cols::Type=Union{AbstractString, Missing};
+                 compress::Bool=false)
     categorical!(df::DataFrame, cname::Union{Integer, Symbol};
                  compress::Bool=false)
     categorical!(df::DataFrame, cnames::Vector{<:Union{Integer, Symbol}};
                  compress::Bool=false)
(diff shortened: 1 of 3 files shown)


bkamins:
CI fails only do to coverage on deprecated.jl which is not touched by this PR.

nalimilan:
I wonder whether it wouldn't be better to use a positional argument for this. Not sure why, maybe because types are often used that way. What do you think?

bkamins:
Makes sense, and I would say I know why 😄. You either pass a list of columns as a second argument **or** you pass the type as column selector. I have just pushed this fix.

src/dataframe/dataframe.jl
```diff
@@ -972,11 +972,15 @@ disallowmissing!(df::DataFrame, cols::Colon=:) =
                  compress::Bool=false)
     categorical!(df::DataFrame, cnames::Union{Regex, Not};
                  compress::Bool=false)
-    categorical!(df::DataFrame; compress::Bool=false)
+    categorical!(df::DataFrame, cols::Type=Union{AbstractString, Missing};
```
nalimilan: Maybe put this one first since it's the simplest method?

PR #1923, opened 2019-08-14
Update deprecated Travis build commands

bkamins:
The old path-getting syntax is currently deprecated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ notifications:
 
 after_success:
   - julia -e 'import DataFrams; cd(dirname(pathof(DataFrames)));
-              Pkg.add("Coverage"); using Coverage;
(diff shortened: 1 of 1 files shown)


ararslan:
Changing to the package directory is actually unnecessary. You can just make it
```yml
after_success:
  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
```

bkamins:
Fixed. Do you know why it was changed in the original code?

ararslan:
Holdover from the 0.6/0.7 transition, if I'm not mistaken. That directory change is the rough equivalent to the old `cd(Pkg.dir("DataFrames"))`. Changing to that directory was necessary in the Pkg2 days.

PR #1924, opened 2019-08-19
export Tables and Tables.columnindex

bkamins:
@quinnj and @nalimilan I have proposed to export `Tables` and `Tables.columnindex`. I am OK not to export `Tables.columnindex` if you feel it should not be exported and accessed with `Tables.` prefix.


nalimilan:
I'm fine with exporting `columindex`, but should we reexport Tables?

bkamins:
I was also thinking about it. The good thing is that Tables.jl exports nothing, so it does not pollute the namespace. But the benefit is minor (most users will not need it). Similarly here - I am also OK not to export it.

nalimilan:
Probably better not export it then.

bkamins:
OK - fixed

Issue #1895, opened 2019-07-21
Support getproperty on DataFrameRows


oschulz:
In some applications that require a vector of something struct-like, `DataFrameRows` is better choice than `DataFrame`. It would be very helpful if `.field` via getproperty would also work on `DataFrameRows` since column access is still often required. This would also make `DataFrameRows` a mutable, API-compatible alternative to `TypedTables.Table`.

Just discussed this with @bkamins at JuliaCon.

ararslan:
This already works.
```
julia> df = DataFrame(a=1:3, b=2:4)
3×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │
│ 2   │ 2     │ 3     │
│ 3   │ 3     │ 4     │

julia> row = first(eachrow(df))
DataFrameRow
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │

julia> row.a += 1
2

julia> df
3×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 2     │ 2     │
│ 2   │ 2     │ 3     │
│ 3   │ 3     │ 4     │
```

bkamins:
@oschulz idea is to add `getproperty` to `DataFrameRows` that would essentially be defined as:
```
getproperty(dfrs::DataFrameRows, s::Symbol) = getproperty(parent(dfrs), s)
```
I think it is acceptable to add it. We could also add it to `DataFrameCols`.

So you could write `eachrow(df).a` and it would the same as `df.a`.

ararslan:
If I'm understanding correctly, that seems very breaking. What I showed above is something I personally have used several times just in the past week...

bkamins:
@ararslan - your proposal is exactly like `eachrow` was designed to be used.

@oschulz proposal is breaking as now we expose properties `df` and `index` in `DataFrameRows` object, but I would assume that people do not rely on this.

The idea was to make `DataFrameRows` object behave like [StructArrays.jl](https://github.com/piever/StructArrays.jl) currently do. Do you see any serious shortcomings of such approach (apart from the fact - noted above - that this was not the original design intention).

ararslan:
Okay, I think I'm not properly understanding the proposal then.

[...]

PR #1925, opened 2019-08-19
add getproperty and propertynames for data frame iterators

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1895.

CC @oschulz

### DIFF
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -82,6 +82,10 @@ Base.@propagate_inbounds Base.getindex(itr::DataFrameRows, i::Int) =
 Base.@propagate_inbounds Base.getindex(itr::DataFrameRows{<:SubDataFrame}, i::Int) =
     DataFrameRow(parent(itr.df), itr.index, rows(itr.df)[i])
 
+Base.getproperty(df::DataFrameRows, col_ind::Symbol) = getproperty(itr.df, col_ind)
+# Private fields are never exposed since they can conflict with column names
+Base.propertynames(df::DataFrameRows, private::Bool=false) = names(itr.df)
+
 # Iteration by columns
 """
     DataFrameColumns{<:AbstractDataFrame, V} <: AbstractVector{V}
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/iteration.jl
```diff
@@ -75,12 +75,16 @@ x  3
 eachrow(df::AbstractDataFrame) = DataFrameRows(df, index(df))
 
 Base.IndexStyle(::Type{<:DataFrameRows}) = Base.IndexLinear()
-Base.size(itr::DataFrameRows) = (size(itr.df, 1), )
+Base.size(itr::DataFrameRows) = (size(parent(itr), 1), )
 
 Base.@propagate_inbounds Base.getindex(itr::DataFrameRows, i::Int) =
-    DataFrameRow(itr.df, itr.index, i)
+    DataFrameRow(parent(itr), getfield(itr, :index), i)
```
nalimilan: Define an `index` method for `DataFrameRows`?

test/iteration.jl
```diff
@@ -115,4 +115,17 @@ end
     @test copy(erd[1]) == (y1 = 51, y2 = 21, y3 = 31) # the removed columns are reflected
 end
 
+@testset "getproperty and propertynames" begin
+    df_base = DataFrame([11:16 21:26 31:36 41:46])
+    for df in [df_base, view(df_base, 1:3, 1:3)]
```
nalimilan:
```suggestion
    for df in (df_base, view(df_base, 1:3, 1:3))
```

test/iteration.jl
```diff
@@ -115,4 +115,17 @@ end
...
+        for x in [eachcol(df), eachcol(df, true), eachrow(df)]
```
nalimilan:
```suggestion
        for x in (eachcol(df), eachcol(df, true), eachrow(df))
```

src/abstractdataframe/iteration.jl
bkamins: Actually maybe we should remove the `:index` field. We currently do not use it. It is always set to `index(df)` and we do not export `index` anyway.

nalimilan: OK, simplification is always nice.

bkamins: I have changed this - let us see how it goes.

bkamins:
This should be good to have a look at.

PR #1928, opened 2019-08-25
fix deprecation for setindex! using 1-row DataFrame

bkamins:
This fixes https://discourse.julialang.org/t/overwrite-partial-dataframe-row-with-new-array/27953.

@nalimilan After this is merged we should tag a release. Ideally I would also merge https://github.com/JuliaData/DataFrames.jl/pull/1925 and https://github.com/JuliaData/DataFrames.jl/pull/1899 if they could be ready relatively soon (I have time to work on it) and then tag 0.20.0 release (this is my preferred option if you have time to have a look, especially at #1899 which is large). Otherwise we should just pick this PR over 0.19.3 and tag 0.19.4 with only this fix.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataFrames"
 uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
-version = "0.19.2"
+version = "0.19.3"
 
 [deps]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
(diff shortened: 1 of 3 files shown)


bkamins:
What was mildly non-breaking, but on the border is:
* deprecation of `Tuple` for `on` in `join`
* different printing of `DataFrameRows` and `DataFrameColumns`

And what is breaking is https://github.com/JuliaData/DataFrames.jl/pull/1899 (as I disallow some old functionality).

So if it is OK with you I would:
* bump a version here to 0.19.3
* merge https://github.com/JuliaData/DataFrames.jl/pull/1925
* tag a release 0.19.3 now

and leave https://github.com/JuliaData/DataFrames.jl/pull/1899 for later (this is a complex PR so it is better to sleep over it). Please confirm if we can go this way.

PR #1931, opened 2019-08-26
Cleanup tests

bkamins:
What I have changed:
* added tests for `DataFrame` constructors from `eachrow` and `eachcol` as they were missing
* removed `"error"` word from testset names, so that they do not produce false positives when parsing log files.


bkamins:
@quinnj - you might want to have a look at tests of `eachrow` and `eachcol` constructors - I guess they behave exactly as you have designed Tables.jl, but please confirm. Thank you.

PR #1932, opened 2019-08-26
Why not let `first(df)` have the same behaviour as `head(df)`?

xiaodaigh:


xiaodaigh:
Oh. Is that what base is doing? Then forget

bkamins:
@nalimilan we could allow `head` and `tail` though (I know they have been long deprecated). The `first(collection, n)` is defined only for `AbstractString` in Base. Having `firs` alias to `head` is not that bad I think. Same with `tail` and `last` (but this is not something crucial).

This is a similar discussion as with `size` and `nrow` and `ncol`.

xiaodaigh:
OK. I like head tail ncol and nrow. They are concise and works well.

bkamins:
And many people like them - that is why I would discuss to leave `head` and `tail`, although polluting name space with additional function names is not a very good practice in general.

mkborregaard:
I agree - head, tail, ncol and nrow are useful aliases, well-known from other languages.

bkamins:
Moving the discussion to https://github.com/JuliaData/DataFrames.jl/issues/1933 as this PR cannot be reopened.

PR #1934, opened 2019-08-26
Make sort type-stable to improve performance

### LINKED ISSUE #1927: sort! performance
Maybe I am wrong, but it seems that the sort function is slow:

```
using Random
using DataFrames

N = convert(Int, 1e7) 

A = rand(0:100, N, 4)
A = DataFrame(A, [:a, :b, :c, :d])

t_begin = time()
sort!(A, [:b, :d])
println("Sort time was $(time() - t_begin)")                                                                                                                              
```
the dataframe is sorted in ~30s 

The equivalent in pandas instead:

```
import pandas as pd 
from time import time
import numpy as np

N = int(1e7)
A = np.random.randint(0, 1000, size=(N, 4))
A = pd.DataFrame(A, columns=['a', 'b', 'c', 'd'])

begin_t = time() 
A.sort_values(['b', 'd'], inplace=True)
print("Sort time was {}".format(time()-begin_t))
```

takes ~3 seconds

Is this similar to the problem of GroupBy performance?

nalimilan:
Specialize on the type of sorting columns to ensure type-stability, like we do for grouping.

Fixes #1927.

### DIFF
--- a/src/abstractdataframe/sort.jl
+++ b/src/abstractdataframe/sort.jl
@@ -44,7 +44,7 @@ function ordering(col_ord::UserColOrdering, lt::Function, by::Function, rev::Boo
 end
 
 ordering(col::ColumnIndex, lt::Function, by::Function, rev::Bool, order::Ordering) =
-             Order.ord(lt,by,rev,order)
+    Order.ord(lt,by,rev,order)
 
 # DFPerm: defines a permutation on a particular DataFrame, using
 #         a single ordering (O<:Ordering) or a list of column orderings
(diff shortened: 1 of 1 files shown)


nalimilan:
> What is the precompilation overhead for the first run? (just of curiosity)

The overhead for the first run is large, but not worse than the existing performance (about 10 seconds on my machine).

> * Have you tested this on significantly heterogeneous columns? I am asking, because AFAICT the line https://github.com/JuliaData/DataFrames.jl/pull/1934/files#diff-4311d0d4a7500122af7804eea4238aa5R80 will not be stable in general. Normally in Base I saw a recursion that handled this kind of comparison (peeling the head of `Tuple` to get the type stability). Another approach would be a `@generated` function approach here that would unroll the loop

Good point. However I just tested it and it's as fast with two columns of different types. Looking at the code for `Sort.lt`, the compiler is able to figure out the types.

bkamins:
>  However I just tested it and it's as fast with two columns of different types. Looking at the code for Sort.lt, the compiler is able to figure out the types.

Yes, that is why I have said "multiple heterogeneous columns". For 2-element `Tuple` I think the compiler is able to handle the loop without problem. So we can treat it as a border case for the future (as I guess sorting on 1 or 2 columns covers all typical cases, also if `Tuples` are homogeneous the compiler should handle this without problem).

In short: I am OK to leave it as is (remembering that there is some slight improvement possible for the very complex `sort!` case).

bkamins:
As a side comment: maybe consider adding more tests with standard types that can be expected in usage for sorting on 1 and 2 arguments (this is what I have started trying to do - and that is why the tests grow). This way we can make sure that in the future they will get compiled ahead of time (if ideas from PackageCompiler.jl go mainstream).

bkamins:
@nalimilan Feel free to merge it if you are not planning to add anything to this PR for now.

nalimilan:
I just wanted to check that the compilation time wasn't too high compared to the current version, given that the gain is only clear for about 1M rows. But that sounds OK, adding about 2s for the first time you sort on a given column type isn't unreasonable (in practice most people should sort on `Float64`, maybe `Int`).

I'm not a fan of testing multiple types just in the perspective of precompilation, as that makes tests more complex and slower without a direct gain at the moment. I think we will need separate code if we want to allow precompilation. It can be be simpler and easier to manage since it doesn't have to check correctness: we could simply create a big data frame with columns of all common types, and run all common operations on all columns.

nalimilan:
Ah, should have noted: sorting with three or four columns of different types is still quite fast.

bkamins:
Excellent - thank you!
So just to be clear - In the simple scenario of sorting 1 column did the compilation time increase or stayed the same (I would expect it should stay the same and the only problem is that you need to recompile for different types - right)?

nalimilan:
It seems to have increased a little bit (from 2.3s to 2.7s). But yeah, the main drawback is that recompilation will happen for each new type.

bkamins:
2.3 to 2.7s is not significant, so I think it is OK.

PR #1937, opened 2019-09-02
Make push! more strict

bkamins:
This is a deprecation PR for now.
Note that we still allow `AbstractArray`s (not only `AbstractVector`s) as this is what base allows:
```
julia> x = ones(2,2)
2×2 Array{Float64,2}:
 1.0  1.0
 1.0  1.0

julia> x[:, 1] = zeros(1,1,1,2)
1×1×1×2 Array{Float64,4}:
[:, :, 1, 1] =
 0.0

[:, :, 1, 2] =
 0.0

julia> x
2×2 Array{Float64,2}:
 0.0  1.0
 0.0  1.0
```

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1191,7 +1191,7 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; columns
 end
 
 """
-    push!(df::DataFrame, row)
+    push!(df::DataFrame, row::Union{Tuple, AbstractArray, Base.Generator})
     push!(df::DataFrame, row::Union{DataFrameRow, NamedTuple, AbstractDict};
           columns::Symbol=:intersect)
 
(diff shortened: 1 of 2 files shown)


bkamins:
However, when assigning to columns we are strict (and require an `AbstractVector` not any `AbstractArray`).

bkamins:
CI fails only because of coverage.

bkamins:
We have three options for LHS for `push!` (and then we will sync `setindex!` for `DataFrameRow`):
1. allow `Any` and then `push!` behaves like `map` (i.e. tries to iterate and fails if it is not possible), this will allow types from `Iterators` to work
2. allow `AbstractArray` and `Tuple` - this is a strict approach
3. allow `AbstractArray`, `Tuple` and `Base.Generator` - this is an intermediate approach that will allow generator expressions on LHS
4. allow `Any` except `AbstractString` (as this is the only possibly problematic iterable in Base that users can a mistake when using it)

I am a bit afraid of 1. as it can be surprising, but I am OK to leave it (then I will change `setindex!` implementation).

I prefer 3 over 2 as I do not see a scenario when the user makes a mistake by passing a generator on LHS (and such a mistake can happen in scenario 1).

Option 4 is a refinement of option 1 to disallow a possibly only problematic case.

From my perspective I am OK with 1, I introduced other options to safeguard people against silly mistakes in code producing unexpected results.

In general I do not have very strong preference over any of these options. Option 2 should cover most of use cases anyway. Other options are just a matter how flexible we want to be in the "rare cases".

Can you please give your opinion?

bkamins:
@nalimilan - do you have any opinion on this? As noted above I am pretty flexible here (i.e. I care more about newcomers than myself when rising this PR), so if you feel 1. should stay I will close it and just make `setindex!` in sync with 1. in another PR.

nalimilan:
I'd favor 1 or 2: either be strict, or accept any iterator. Intermediate solutions sound too ad-hoc and don't have a precedent in Base AFAIK.

bkamins:
OK - I have decided to disallow Base.Generator. The main reason is that we can always be more flexible later. If we merge this then I will also make a small PR to `setindex!` that will follow the same rule (now it allows `Base.Generator`).

PR #1938, opened 2019-09-03
Stop using makeunique=true for grouping keys in combine

### LINKED ISSUE #1460: check for preserved index column during combine

nalimilan:
Instead, check whether columns with the same names as grouping keys have been returned, and throw an error if they are not equal.

This is breaking since columns named key_1, etc. won't be created, but in practice it should not be frequently used. Introducing a deprecation would be more annoying than helpful since it would require adding a keyword argument to disable adding grouping keys and force users to set it to silence
the warning, which happens in most common cases.

Fixes #1460.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -435,7 +435,16 @@ of `combine(map(f, groupby(df, cols)))`.
 function combine(f::Any, gd::GroupedDataFrame)
     if length(gd) > 0
         idx, valscat = _combine(f, gd)
-        return hcat!(gd.parent[idx, gd.cols], valscat, makeunique=true)
+        keys = _names(gd.parent)[gd.cols]
+        for key in keys
+            if hasproperty(valscat, key) &&
+               !isequal(valscat[!, key], view(gd.parent[!, key], idx))
+               throw(ArgumentError("column :$key in returned data frame " *
+                                   "is not equal to grouping key :$key"))
+            end
+        end
+        return hcat!(gd.parent[idx, gd.cols],
+                     without(valscat, intersect(keys, _names(valscat))))
     else
         return gd.parent[1:0, gd.cols]
     end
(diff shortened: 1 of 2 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -435,7 +435,16 @@ of `combine(map(f, groupby(df, cols)))`.
 function combine(f::Any, gd::GroupedDataFrame)
     if length(gd) > 0
         idx, valscat = _combine(f, gd)
-        return hcat!(gd.parent[idx, gd.cols], valscat, makeunique=true)
+        keys = _names(gd.parent)[gd.cols]
+        for key in keys
```
bkamins: a small gripe `keys` is a standard function name from Base. It is not a big problem though - please decide if you want to rename it or not.

bkamins:
I am OK to skip deprecation period in this case.

nalimilan:
> One small thing that could be added is a paragraph in the documentation of `by` and `combine` explaining this behavior.

Good idea. I've also noticed that the same change needs to be applied to `map`.

src/groupeddataframe/grouping.jl
nalimilan: I'd say that's not a problem, especially given that the function is quite short.

bkamins:
Thank you. I am OK to merge it.

PR #1940, opened 2019-09-03
Deprecate eltypes

### LINKED ISSUE #1596: Future of eltypes
Since now we can write `eltypes(df)` as `eltype.(columns(df))` there is a question do we want to keep `eltypes` function?

nalimilan:
Here's how it looks.

Fixes #1596.


nalimilan:
Merging since we only got positive replies.

PR #1941, opened 2019-09-04
Use nonmissingtype instead of Missings.T

nalimilan:
The latter is deprecated since Julia now exports `nonmissingtype`.
Missings defines `nonmissingtype` on previous Julia versions.
Also simplify the code a bit: there's no need to check `T >: Missing` before calling `nonmissingtype`, as it is the identity for other types.

Waiting for https://github.com/JuliaRegistries/General/pull/3224.


nalimilan:
Failure is due to a deprecation warning printed by CategoricalArrays in a log test. https://github.com/JuliaData/CategoricalArrays.jl/pull/205 should fix this.

bkamins:
OK, but then we should first release CategoricalArrays.jl and then change the Project.toml in this PR to require its proper version also. Right?

nalimilan:
I don't think we need to require that version: the test failure doesn't reflect a real issue, it's just due to how we test that warnings are printed.

bkamins:
OK

nalimilan:
CategoricalArrays tagged: https://github.com/JuliaRegistries/General/pull/3292

bkamins:
Looks good to be merged.

bkamins:
So - are you going to cherry-pick this PR and make 0.19.4 or we wait for https://github.com/JuliaData/DataFrames.jl/pull/1934, https://github.com/JuliaData/DataFrames.jl/pull/1937 and https://github.com/JuliaData/DataFrames.jl/pull/1946 and make 0.20.0 release?

nalimilan:
Done: https://github.com/JuliaRegistries/General/pull/3328

PR #1946, opened 2019-09-05
fix documentation examples

### LINKED ISSUE #1945: Sorting error using examples from docs
For both `v0.19.3` and `master` on Julia v1.2:

```julia
using DataFrames, CSV
iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv"))
sort!(iris)
```
```
ERROR: LoadError: ArgumentError: CSV.Column is read-only; to get a mutable vector, do `copy(col)` or to make all columns mutable do `CSV.read(file; copycols=true)` or `CSV.File(file) |> DataFrame`
Stacktrace:
 [1] setindex!(::CSV.Column{Float64,Float64}, ::Float64, ::Int64) at /home/ian/.julia/packages/CSV/KgT4M/src/tables.jl:26
 [2] permute!!(::CSV.Column{Float64,Float64}, ::Array{Int64,1}) at ./combinatorics.jl:107
```

bkamins:
Fixes #1945

### DIFF
--- a/docs/src/man/sorting.md
+++ b/docs/src/man/sorting.md
@@ -5,7 +5,8 @@ Sorting is a fundamental component of data analysis. Basic sorting is trivial: j
 ```jldoctest sort
 julia> using DataFrames, CSV
 
-julia> iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv"));
+julia> iris = CSV.File(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv")) |>
+              DataFrame;
 
 julia> first(iris, 4)
 4×5 DataFrame
--- a/docs/src/man/sorting.md
+++ b/docs/src/man/sorting.md
@@ -5,8 +5,7 @@ Sorting is a fundamental component of data analysis. Basic sorting is trivial: j
 ```jldoctest sort
 julia> using DataFrames, CSV
 
-julia> iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv"),
-                       copycols=true);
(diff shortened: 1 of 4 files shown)


docs/src/man/sorting.md
```diff
@@ -5,7 +5,8 @@ Sorting is a fundamental component of data analysis. Basic sorting is trivial: j
 ```jldoctest sort
 julia> using DataFrames, CSV
 
-julia> iris = CSV.read(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv"));
+julia> iris = CSV.File(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv")) |>
```
nalimilan: Better use the more common syntax `DataFrame(...)`? That will avoid confusing newcomers.

bkamins: fixed (although in my codes I use the `File` pattern 😄)

nalimilan: No, I meant using `DataFrame(CSV.File(...))` instead of `|>`.

bkamins: OK. Let us stick with `CSV.read` with `copycols=true` (what I pushed now) as it is also a pattern that is good to show I think.

nalimilan: I prefer `DataFrame(CSV.File(...))`, as it makes it less likely that people will call `CSV.read` and forget to pass `copycols=true`. Actually I'd rather get rid of `CSV.read`.

BTW, there are other occurrences in the docs IIRC, better change all of them at the same time.

bkamins: OK - done. At least we agree what should be used 😄.

Issue #1942, opened 2019-09-04
Allow partial re-ordering for `permutecols!`


nilshg:
`permutecols!` currently requires passing a full list of column names in the desired order. It would be nice to have the option of only reordering the start of the DataFrame, a common operation that mirrors Stata's `order` command.

Desired outcome would be something like:
```
using DataFrames
df = DataFrame(rand(100,20))
leading_cols = [:x5, :x9, :x13]
df = df[!, vcat(lead_cols, setdiff(names(df), leading_cols)]
```

bkamins:
What about reordering only columns that appear in the second argument (i.e. that we allow any subset of columns - not only the starting columns). This will not be ambiguous.

nalimilan:
Why not. Though that doesn't allow the symmetric operations of moving some columns to the end. Another (alternative or complementary) solution would be to take pairs giving the name/index and its new position.

Anyway, does that really need to be in 1.0?

bkamins:
I removed the milestone (but I will implement it pretty fast as this is a simple change, so it does not matter). The problem with `name/index => newindex` is that it can be sometimes hard to disambiguate what was the intention (is the operation sequential or one shot).

To move a column `5` to the end you will pass `[6:ncol(df); 5]`.

The permutation notation is unambiguous as it specifies a new order of columns within a subset of columns (so columns not included in the permutation do not move).

bkamins:
A current approach is in https://github.com/JuliaData/DataFrames.jl/pull/1964

bkamins:
Given our current implementation the reordering you asked for is achieved by:
```
select!(df, :x1, :x5, :x2, :)`
```
(i.e. give the first three columns and then indicate by `:` to include the rest of them)

so I am closing this issue (please reopen if you feel there is something more to be added here)

PR #1955, opened 2019-09-11
add permutecols and extend permutecols!

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1942

@nilshg - can you please test this PR for bugs and if it follows what you would expect?

@nalimilan - as usual apart from general comments a review of documentation would be appreciated.

Note that earlier `permutecols` made no sense as it was the same as `df[:, perm]` but now it makes sense, as we accept subsets of columns.

### DIFF
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -62,6 +62,7 @@ show
 sort
 sort!
 unique!
+permutecols
 permutecols!
 vcat
 append!
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1361,8 +1361,10 @@ end
     permutecols!(df::DataFrame, p::AbstractVector)
 
 Permute the columns of `df` in-place, according to permutation `p`. Elements of `p` may be
-either column indices (`Int`) or names (`Symbol`), but cannot be a combination of both. All
-columns must be listed.
+either column indices (`Int`) or names (`Symbol`), but cannot be a combination of both.
+Any subset of columns may be listed, but it may not contain duplicates.
+If the subset does not contain all columns, then the permutation is performed
+within the subset passed.
 
 ### Examples
 
@@ -1378,9 +1380,7 @@ julia> df = DataFrame(a=1:5, b=2:6, c=3:7)
 │ 4   │ 4     │ 5     │ 6     │
 │ 5   │ 5     │ 6     │ 7     │
 
-julia> permutecols!(df, [2, 1, 3]);
-
-julia> df
+julia> permutecols!(df, [2, 1, 3])
 5×3 DataFrame
 │ Row │ b     │ a     │ c     │
 │     │ Int64 │ Int64 │ Int64 │
@@ -1391,9 +1391,7 @@ julia> df
 │ 4   │ 5     │ 4     │ 6     │
 │ 5   │ 6     │ 5     │ 7     │
 
-julia> permutecols!(df, [:c, :a, :b]);
-
-julia> df
+julia> permutecols!(df, [:c, :a, :b])
 5×3 DataFrame
 │ Row │ c     │ a     │ b     │
 │     │ Int64 │ Int64 │ Int64 │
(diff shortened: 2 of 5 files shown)


bkamins:
That is exactly the point of difference of `permutecols!` and `permutecols` - this example is intended to show the difference between them.

Call `permutecols!(df, [:b, :a])` takes the data frame with columns `[:c, :a, :b]`, because earlier `permutecols!` changed the column order in `df` data frame.

And `permutecols(df, [:b, :a])` in the second takes the data frame with columns `[:a, :b, :c]`, as earlier `permutecols` did not alter the column order of `df`.

bkamins:
Now I have had a look at Stata documentation of `order`. We might add a kwarg `location` that takes a default value of `nothing` which does what I currently implemented, then it can take following values:
* `:first` - to put them in front of the data frame
* `:last` - to put them at the end of the data frame
* `:before=>column_name`
* `:after=>column_name`

where `column_name` is a `Symbol` giving the name of the column. @nalimilan - what do you think about it? If we added all this we would get "more or less" all that is potentially needed. The only question is if this should be still called `permutecols!` 😄.

bkamins:
actually it is even simpler to implement `location` as a last positional argument (so I would opt for this). Any preference here.

bkamins:
OK. I propose to rename the functions to `reordercols!` and `reordercols`.
I have implemented all the functionality I have proposed.

I would appreciate if you did some testing if you like it and commented. If we agree that this is what we want then I will write tests of the functionality and update docstrings.

nilshg:
What's the syntax in Pkg to install a specific fork?

bkamins:
use `PackageSpec` to point to url and branch. But as this change is very simple it is easier to `dev` normal package and just copy the definitions into the appropriate file (this is a bit hacky but in this case it is simplest as the code-block to copy is a simple copy-paste).

src/dataframe/dataframe.jl
```diff
@@ -1358,11 +1358,13 @@ function Base.push!(df::DataFrame, row::Any)
...
 
 """
-    permutecols!(df::DataFrame, p::AbstractVector)
+    reordercols!(df::DataFrame, cols::AbstractVector)
 
-Permute the columns of `df` in-place, according to permutation `p`. Elements of `p` may be
-either column indices (`Int`) or names (`Symbol`), but cannot be a combination of both. All
-columns must be listed.
+Permute the columns of `df` in-place, according to permutation `cols`. Elements of `cols` may be
+any valid column selection, but cannot be a combination of both.
```
nalimilan: What is "both"? Also, not in sync with `reordercols`.

docs/src/lib/functions.md
```diff
@@ -62,6 +62,7 @@ show
 sort
 sort!
 unique!
+permutecols
```
nalimilan: If functions are renamed, this needs to be updated. Same below for exports.

nalimilan:
Having a `location` argument similar to Stata sounds like a good idea. But a keyword argument sounds more logical for that. We may also want to add a keyword argument to sort columns in alphabetic order at some point (which could be combined with `All`, `Between` or a regex).

nalimilan:
Regarding the name, I'm not sure. "permute" sounds OK.

Also I've just realized that dplyr [uses `select`](http://www.datasciencemadesimple.com/re-arrange-re-order-column-dataframe-r-using-dplyr/) for this. In Julia the equivalent would be `select(df, cols, All())` or `select(df, Not(cols), All())`. Not sure that's really better, but probably worth considering as it simplifies the API further.

bkamins:
@nalimilan This is an excellent observation. Actually this means that what we need to do is just to `permutecols!` to `select!`. The only thing we would loose is that `permutecols!` throws an error when an invalid permutation is passed, but I do not think it is crucial to keep it.

The reason is that to put columns in front currently you can just write:
```
select(df, All(col1, col2, All()))`
```
to put column in the back you write:
```
select(df, All(Not([col1, col2]), col1, col2))
```

the `:before` and `:after` cases would not be that simple to implement, but I think they are not super useful anyway.

The only change we could consider is to have `select(df, cols...)` be transformed to `select(df, All(cols...))`.

@nilshg and @nalimilan If you are OK with this I will put up another PR for this (i.e. deprecation of `permutecols!` and adding slurping to `select`/`select!`).

bkamins:
See #1955. I finally decided to leave `permutecols!` as it is a more fundamental operation than `select!` (we would have to leave it internally anyway). But if you prefer we can deprecate it.

bkamins:
I meant https://github.com/JuliaData/DataFrames.jl/pull/1964.

nalimilan:
Why is it more fundamental?

bkamins:
Because it has to modify internals of a `DataFrame` in-place (reordering the columns and the index). Of course I can move the logic of `permutecols!` into `select!`, but maybe it is OK to leave `permutecols!` as it additionally checks if you actually passed a permutation (which `select!` does not do and I think it should not).

nalimilan:
So that's not really more fundamental than `select!` I'd say. Not sure whether checking that a permutation has been passed justifies having a separate function. Do you think that's really common? Julia doesn't offer a similar function for matrices, and neither dplyr not Pandas provide separate functions AFAICT.

bkamins:
OK - I will remove it.

Issue #1922, opened 2019-08-14
Add kwarg do disallowmissing that skips conversion of columns with missing values


bkamins:
As in the title and per discussion on Slack.
This should be done for `disallowmissing` and `disallowmissing!`.

I will implement it (unless someone else is willing to do this - then please let me know).

CC @nalimilan

Issue #1953, opened 2019-09-09
API for aggregate


bkamins:
I think, since we make `aggregate` stay, we could allow the similar API in `aggregate` that we allow in `by`. Namely:
1. functions could be passed as several positional arguments, e.g. `aggregate(df, sum, std)`
2. we should add `skipmissing` kwarg for consistency
3. we could add a kwarg that provides a suffix to the column name, namely `aggregate(df, _fun=x->x)` would add `_fun` suffix to the generated columns

If we are OK with all these changes I can implement them. This probably does not have to be done before 1.0, so I do not set a milestone.

Comments are welcome.

bkamins:
@nalimilan Actually we should decide before 1.0 if we keep `aggregate`.

bkamins:
I am closing it. I would keep `aggregate` as is (i.e. not deprecate it), but not develop it further because now we have functionalities in `select` and `combine` that are much more powerful.

The only reason to keep `aggregate` is that it uses `makeunique=true` option by default, which is important when you pass multiple anonymous functions to (I think a rare case but still possible). Plus keeping it is less breaking and it does not hurt that much to have it.

Please reopen if you disagree.

PR #1956, opened 2019-09-11
add error to disallowmissing and disallowmissing!

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1922

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1253,8 +1253,7 @@ from element type `Union{T, Missing}` to `T` to drop support for missing values.
 
 If `cols` is omitted all columns in the data frame are converted.
 
-If `error=false` then columns containing `missing` as value will be skipped in the conversion,
-otherwise an error is thrown if such columns exist.
+
 
 **Examples**
 
@@ -1283,7 +1282,7 @@ julia> df = DataFrame(a=[1,missing])
 ├─────┼─────────┼────────┤
 │ 1   │ 1       │ 1      │
 │ 2   │ missing │ 2      │
-julia> disallowmissing(df, error=false)
+
 2×2 DataFrame
 │ Row │ a       │ b     │
 │     │ Int64⍰  │ Int64 │
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1245,12 +1245,15 @@ julia> ncol(df)
 
 """
     disallowmissing(df::AbstractDataFrame,
-                    cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon}=:)
+                    cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon}=:;
+                    skip::Bool=false)
 
 Return a copy of data frame `df` with columns `cols` converted
 from element type `Union{T, Missing}` to `T` to drop support for missing values.
 
 If `cols` is omitted all columns in the data frame are converted.
+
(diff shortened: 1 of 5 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1271,15 +1274,35 @@ julia> disallowmissing(df)
...
 ```
+
+julia> df = DataFrame(a=[1,missing])
+2×2 DataFrame
+│ Row │ a       │ b      │
+│     │ Int64⍰  │ Int64⍰ │
+├─────┼─────────┼────────┤
+│ 1   │ 1       │ 1      │
+│ 2   │ missing │ 2      │
+julia> disallowmissing(df, skip=true)
```
```suggestion

julia> disallowmissing(df, skip=true)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1245,13 +1245,16 @@ julia> ncol(df)
...
-                    cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon}=:)
+                    cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon}=:;
+                    skip::Bool=false)
 
 Return a copy of data frame `df` with columns `cols` converted
 from element type `Union{T, Missing}` to `T` to drop support for missing values.
 
 If `cols` is omitted all columns in the data frame are converted.
 
+If `skip=true` then columns containing `missing` as value will be skipped in the conversion.
```
```suggestion
If `skip=true` then columns containing a `missing` value will be skipped instead of throwing an error.
```

src/dataframe/dataframe.jl
```diff
@@ -970,42 +970,49 @@ allowmissing!(df::DataFrame, cols::Colon=:) =
...
+    disallowmissing!(df::DataFrame, cols::Union{Integer, Symbol}; skip::Bool=false)
+    disallowmissing!(df::DataFrame, cols::Union{AbstractVector, Regex, Not, Between, All};
+                     skip::Bool=false)
 
 Convert columns `cols` of data frame `df` from element type `Union{T, Missing}` to
 `T` to drop support for missing values.
 
 If `cols` is omitted all columns in the data frame are converted.
+
+If `skip=true` then columns containing `missing` as value will be skipped in the conversion.
```
```suggestion
If `skip=true` then columns containing a `missing` value will be skipped instead of throwing an error.
```

bkamins:
Thank you for the corrections. I thought `skip` is enough (and I still think it, what else could `skip` mean in the context of this function?) but I am ok with either of the names you proposed if you feel they are better.

bkamins:
In https://github.com/JuliaData/DataFrames.jl/issues/1953#issue-491172598 I plan to add `skipmissing` to `aggregate` but working rowwise (as in `groupby`). I would keep `skipmissing` for row-selection.

Here `skip` is for "column selection". Do you think I should add to https://github.com/JuliaData/DataFrames.jl/issues/1953#issue-491172598 also `skip` kwarg that skips columns with missing? Then I would add to `aggegate` a general column selection scheme that is the same as in `disallowmissing` (so that you can select on which columns you want to perform aggregation). Can you please comment here (or in https://github.com/JuliaData/DataFrames.jl/issues/1953#issue-491172598) before I merge this PR so that we are consistent here? Thank you!

nalimilan:
Hmm... I guess `ignoreerror` would be more explicit, but it's not very nice. Maybe `error=false`? That would potentially allow for extensions like `error=nothing` or `error=0` in `aggregate`, which I guess could be useful to avoid hardcoding the value that is returned. (That said, I don't want to base the design on `aggregate` too much since I kind of hoped we could get rid of it.)

bkamins:
Considering all the options `error` seems OK. So - do we go with it?

nalimilan:
Fine with me if you like it.

Issue #1959, opened 2019-09-18
Issues using the `df[!, col]` syntax during broadcasts


eperim:
When updating my code to comply with v0.19, I was changing the `df[col]` calls to `df[!, col]` as instructed. However, that led to 
```julia 
MethodError: no method matching length(::DataFrames.LazyNewColDataFrame{Symbol})
```
while it works fine if I use `df[:, col]` instead.

I don't have a MWE for this, because it happens somewhere deep in a stack containing proprietary packages. However, it seems to me that the `!` syntax is creating some intermediate types under the hood for which not all properties are defined, and that can lead to issues when doing broadcasts over columns.

I'd suggest that, at least, the deprecation warning be fixed to suggest the safe `:` syntax instead of the unsafe `!` one.

eperim:
@oxinabox

oxinabox:
>  I'd suggest that, at least, the deprecation warning be fixed to suggest the safe : syntax instead of the unsafe ! one.

Problem is they do do different things. It might be nice if it suggested both.
Mostly people should use the `:` one, but the `!` ione is supposed to be identical (including unsafety) to `df[col]`


But also there must be some bug with `!`

eperim:
It doesn't seem to be exactly identical for me. Calling `df[col]` worked fine before, but calling `df[!, col]` leads to that error from the original message.

bkamins:
What you must be writing is something like:
```
df[!, col] .= some_operations
```
which replaces column `col` with a new column.
```
df[:, col] .= some_operations
```
is almost the same, but it is an in-place operation, so in some cases it will have a different effect (the most important is when you want to create a new column). The most important are the 1-2 functions up the stack. Chiefly - what function wanted to call `length` on `LazyNewColDataFrame` (in DataFrames.jl we never make such a call).

Use one or the other depending on what your objective is.

Now - to solve the problem - can you please share with me the stacktrace (you can cut off all proprietary methods and leave only the ones that are from Base or DataFrames.jl). The reason is that broadcasting API, as documented [here](https://docs.julialang.org/en/latest/manual/interfaces/#man-interfaces-broadcasting-1) does not require objects that are assigned to using broadcasted assignment to define `length`, they only need to define `axes`. This means that the operation you are trying to do calls `length` on an object that is not required to have `length` defined.

bkamins:
[...]

PR #1960, opened 2019-09-18
correctly use dotview instead of maybeview in broadcasting contexts

bkamins:
Improve warning message for `df[col]` in broadcasted assignment.

### DIFF
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -152,7 +152,7 @@ Note that `sdf[!, col] = v`, `sdf[!, cols] = v` and `sdf.col = v` are not allowe
                     equivalent to `dfr.col = v` if `col` is a valid identifier;
 * `dfr[cols] = v` -> set values of entries in columns `cols` in `dfr` by elements of `v` in place;
                      `v` can be:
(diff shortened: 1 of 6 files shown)


bkamins:
I judge it is appropriate, as it works and I understand why 😄 (so this is not a hacky patch).

However, I will wait a few days as maybe @mbauman might have some comments here.

bkamins:
Having slept over this PR:
my previous solution was not good and actually what was required and sufficient was to redefine `maybeview` to `dotview`.

What is going on:
* `maybeview` is not reserved for broadcasting purposes only; it is a general mechanism to make a view (or leave things as is); so as @oxinabox indicated - we had a bad design in DataFrames.jl
* fortunately `dotview` is used for broadcasting assignment only (these are small details that make design of broadcasting in Base awesome)

This change fixes #1959 in broadcasting contexts but also fixes usage of views in non-broadcasting contexts, like `@views typeof(df[!, :])` which were broken (in general I did not realize that with 0.19 we have fixed this kind of problems in general and only `!` bug was left - I think this was not caught earlier, because people did not use data frames much in chaining operations when working with views).

PR #1961, opened 2019-09-22
allow to create columns using : as row selector

bkamins:
Allow `df[:, col] = vector` and `df[:, col] .= v` to create a column if `col` is not present in `df`. A fresh vector is always allocated.

Given the comments on Slack I am convinced we can add this without harm, but there is no rush (though personally I would prefer to settle this PR before 1.0, as I would prefer not to mess with `getindex`/`setindex!`/broadcasting assignment after 1.0 if possible).

CC @oxinabox @nalimilan @pdeffebach

### DIFF
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -122,6 +122,8 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
 * `df[CartesianIndex(row, col)] = v` -> the same as `df[row, col] = v`;
 * `df[row, cols] = v` -> set row `row` of columns `cols` in-place; the same as `dfr = df[row, cols]; dfr[:] = v`;
 * `df[rows, col] = v` -> set rows `rows` of column `col` in-place; `v` must be an `AbstractVector`;
+                         if `rows` is `:` and `col` is a `Symbol` that is not present in `df` then a new column
+                         in `df` is created and holds a `copy` of `v`; equivalent to `df.col = copy(v)` if `col` is a valid identifier;
 * `df[rows, cols] = v` -> set rows `rows` of columns `cols` in-place; `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
                       (in this case column names must match);
 * `df[!, col] = v` -> replaces `col` with `v` without copying
@@ -185,7 +187,11 @@ Additional rules:
 * in the `df[CartesianIndex(row, col)] .= v`, `df[row, col] .= v` syntaxes `v` is broadcasted into the contents of `df[row, col]` (this is consistent with Julia Base);
 * in the `df[row, cols] .= v` syntaxes the assignment to `df` is performed in-place;
 * in the `df[rows, col] .= v` and `df[rows, cols] .= v` syntaxes the assignment to `df` is performed in-place;
-* in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector; if `col` is `Symbol` and it is missing from `df` then a new column is added; the length of the column is always the value of `nrow(df)` before the assignment takes place;
+  if `rows` is `:` and `col` is `Symbol` and it is missing from `df` then a new column is allocated and added;
+  the length of the column is always the value of `nrow(df)` before the assignment takes place;
+* in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector;
+  if `col` is `Symbol` and it is missing from `df` then a new column is allocated added;
+  the length of the column is always the value of `nrow(df)` before the assignment takes place;
 * the `df[!, cols] .= v` syntax replaces existing columns `cols` in data frame `df` with freshly allocated vectors;
 * `df.col .= v` syntax is allowed and performs in-place assignment to an existing vector `df.col`.
 * in the `sdf[CartesianIndex(row, col)] .= v`, `sdf[row, col] .= v` and `sdf[row, cols] .= v` syntaxes the assignment to `sdf` is performed in-place;
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -135,10 +135,6 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
                        `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
                        (in the latter case column names must match);
 
-Note that `df[!, col] = v`, `df.col = v` and `df[:, col] = v` can be used to add a new column to a `DataFrame`.
-The difference is that `df[:, col] = v` copies `v` before adding it to `df` and `df[!, col] = v` and `df.col = v`
-add `v` without copying it.
-
 `setindex!` on `SubDataFrame`:
 * `sdf[row, col] = v` -> set value of `col` in row `row` to `v` in-place;
 * `sdf[CartesianIndex(row, col)] = v` -> the same as `sdf[row, col] = v`;
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -135,9 +135,9 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
                        `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
                        (in the latter case column names must match);
 
-Note that only `df[!, col] = v` and `df.col = v` can be used to add a new column to a `DataFrame`.
-In particular as `df[:, col] = v` is an in-place operation it does not add a column `v` to a `DataFrame` if `col` is missing
-(an error is thrown if such operation is attempted).
+Note that `df[!, col] = v`, `df.col = v` and `df[:, col] = v` can be used to add a new column to a `DataFrame`.
+The difference is that `df[:, col] = v` copies `v` before adding it to `df` and `df[!, col] = v` and `df.col = v`
+add `v` without copying it.
 
 `setindex!` on `SubDataFrame`:
 * `sdf[row, col] = v` -> set value of `col` in row `row` to `v` in-place;
--- a/src/other/broadcasting.jl
+++ b/src/other/broadcasting.jl
@@ -96,6 +96,14 @@ Base.maybeview(df::AbstractDataFrame, idx::CartesianIndex{2}) = df[idx]
 Base.maybeview(df::AbstractDataFrame, row::Integer, col::ColumnIndex) = df[row, col]
 Base.maybeview(df::AbstractDataFrame, rows, cols) = view(df, rows, cols)
 
+function Base.dotview(df::DataFrame, ::Colon, cols::ColumnIndex)
+    haskey(index(df), cols) && return view(df, :, cols)
+    if !(cols isa Symbol)
+        throw(ArgumentError("creating new columns using an integer index by broadcasting is disallowed"))
+    end
+    LazyNewColDataFrame(df, cols)
+end
+
 function Base.dotview(df::DataFrame, ::typeof(!), cols)
     if !(cols isa ColumnIndex)
         return ColReplaceDataFrame(df, index(df)[cols])
--- a/src/other/broadcasting.jl
+++ b/src/other/broadcasting.jl
@@ -109,7 +109,7 @@ function Base.dotview(df::DataFrame, ::typeof(!), cols)
         return ColReplaceDataFrame(df, index(df)[cols])
     end
     if !(cols isa Symbol) && cols > ncol(df)
-        throw(ArgumentError("creating new columns using an integer index by broadcasting is disallowed"))
+        throw(ArgumentError("creating new columns using an integer index is disallowed"))
     end
     LazyNewColDataFrame(df, cols)
 end
--- a/test/broadcasting.jl
+++ b/test/broadcasting.jl
@@ -1500,6 +1500,13 @@ end
     @test_throws MethodError df[:, 1] .= z
 
     df = DataFrame(ones(3, 4))
+    z = "abc"
+    df[:, :z] .= z
+    @test df.z == fill("abc", 3)
+    @test_throws ArgumentError df[:, 6] .= z
+    @test_throws MethodError df[:, 1] .= z
+
+    df = DataFrame(ones(3, 4))
     z = fill("abc", 1, 1, 1)
     @test_throws DimensionMismatch df[:, :z] .= z
 end
--- a/test/broadcasting.jl
+++ b/test/broadcasting.jl
@@ -1490,4 +1490,14 @@ end
     @test @views typeof(df[!, 1:2]) <: SubDataFrame
 end
 
+@testset "broadcasting of df[:, col] = vector" begin
+    df = DataFrame(ones(3, 4))
+    z = ["a", "b", "c"]
+    df[:, :z] .= z
+    @test df.z == z
+    @test df.z !== z
+    @test_throws ArgumentError df[:, 6] .= z
+    @test_throws MethodError df[:, 1] .= z
+end
+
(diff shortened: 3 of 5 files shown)


pdeffebach:
I confirmed that this behavior works in R. So this change helps users coming from R (at least those who do things like `df[, "col"] = c(1, 2, 3)`. 

I do see the appeal of, during a code review or debugging, clearly knowing which operations are creating variables and which are modifying existing variables. Stata has this obviously with the nice distinction between `gen` and `replace`. 

In the docs the current behavior is: 

```
Note that only `df[!, col] = v` and `df.col = v` can be used to add a new column to a `DataFrame`.
In particular as `df[:, col] = v` is an in-place operation it does not add a column `v` to a `DataFrame` if `col` is missing
(an error is thrown if such operation is attempted).
```

This is hardly as clear as `gen` and `replace` in Stata so I don't code readability is a casualty of this proposed change.

bkamins:
I have to fix
```
Note that only `df[!, col] = v` and `df.col = v` can be used to add a new column to a `DataFrame`.
In particular as `df[:, col] = v` is an in-place operation it does not add a column `v` to a `DataFrame` if `col` is missing (an error is thrown if such operation is attempted).
```
this message - thank you for catching this (this is what current master does, and what will be changed by this PR). If this PR is accepted then it should state that also `df[:, col] =v ` can create a column.

But I am not fully clear how to read your comment:
> This is hardly as clear as `gen` and `replace` in Stata so I don't code readability is a casualty of this proposed change.

Do you want to say that you want `df[:, col] = v` to add a new column or you do not want this?

bkamins:
I have corrected the documentation so you can review now what is the target behavior after this PR.

nalimilan:
As I said, I'd rather wait and see whether we get more requests for this. Adding this after 1.0 wouldn't be terrible since it wouldn't break anything.

bkamins:
Fortunately the PR is relatively simple, so it can wait. I will ask on Slack for people to comment here if they want it.

nickrobinson251:
We would use this in code an Invenia, I think. Ping @eperim

pdeffebach:
> Do you want to say that you want df[:, col] = v to add a new column or you do not want this?

I guess I'm asking what the motivations for the original design were. Which principles would this change be violating?

nalimilan:
> I guess I'm asking what the motivations for the original design were. Which principles would this change be violating?

The principle that `df[:, col] = v` always mutates the column vector in-place. Among other things, this means that values in `v` are converted to the type of the existing column vector, which wouldn't be the case if we created a column when it doesn't exist yet. Errors could happen due to conversion failure when the column exists, which could be surprising as the error won't happen if the column doesn't exist.

bkamins:
@nalimilan - to the point 😄. This was my initial thinking and that is why I have hesitated.

If we decide to incorporate this proposal then:

> Errors could happen due to conversion failure when the column exists, which could be surprising as the error won't happen if the column doesn't exist.

Would have to be extended in an interpretation as something like

> if `col` does not exist we assume that a lazy "virtual" column actually existed and had `eltype` the same as `v` has and the operation just serves to materialize it

of course this is only a conceptual discussion as in practice we just say that `df[:, col] = v` is the same as `df[!, col] = copy(v)` and this is how it is implemented.

bkamins:
Just to be clear where the problem can happen.

Assume that you have two lines of code:

```
try
    df[:, col] = [1,2,3]
    df[:, col] = ["a", "b", "c"]
catch
end
```

You ask yourself what is the contents of `col` after this code.

The answer now is that: if `col` is not present nothing happens, if `col` is present then it depends on its `eltype`.

After the change it would be .

bkamins:
sorry for the mess with the last two commits - I pushed changes to a wrong branch

bkamins:
@nalimilan - I guess there is a consensus here.

@quinnj - maybe you could drop in your opinion?

bkamins:
OK - as @oxinabox converted me to support this PR 😄 here is another reason to merge it.

The operations `df[:, col] = v` and `df[!, col] .= v` are not equivalent. The difference is that `df[:, col] = v` will preserve more of `v` metadata than `df[!, col] .= v` will. Here is an example:
```
julia> df = DataFrame(a=[1,2,3])
3×1 DataFrame
│ Row │ a     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │
│ 3   │ 3     │

julia> df[!, :b] .= Any[1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> df.b
3-element Array{Int64,1}:
 1
 2
 3
```
and we see that due to broadcasting `df.b` has eltype `Int64`. With `df[:, :b] = Any[1,2,3]` the element type would not change and remain `Any`.

The other difference is that `df[:, col] = [1,2,3]` will work on an empty data frame, while `df[!, col] .= [1,2,3]` is an error.

Any opinions on these two points?

oxinabox:
> The operations `df[:, col] = v` and `df[!, col] .= v` are not equivalent. The difference is that `df[:, col] = v` will preserve more of `v` metadata than `df[!, col] .= v` will. Here is an example:

Note, however that `df[:, col] = v` *is* equivelent `df[!, col] = v` (for nonexistant `col`).
The key difference is `.=` vs `=`, not `:` vs `!`.
because it is a statement above writing every element into a column,
vs assigning the column

```
julia> df[!, :b] .= Any[1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> df[!, :c] = Any[1,2,3]
3-element Array{Any,1}:
 1
 2
 3

julia> df
3×3 DataFrame
│ Row │ a     │ b     │ c   │
│     │ Int64 │ Int64 │ Any │
├─────┼───────┼───────┼─────┤
│ 1   │ 1     │ 1     │ 1   │
│ 2   │ 2     │ 2     │ 2   │
│ 3   │ 3     │ 3     │ 3   │
```

So all seems well to me.

bkamins:
> Note, however that df[:, col] = v is equivelent df[!, col] = v (for nonexistant col).

It is not equivalent, as `df[:, col] = v` performs a copy while `df[!, col] = v` does not and `copy` can change the type of RHS.

Here is an example:

```
julia> df = DataFrame()
0×0 DataFrame


julia> x = view([1,2,3], 1:2)
2-element view(::Array{Int64,1}, 1:2) with eltype Int64:
 1
 2

julia> df[!, :x] = x
2-element view(::Array{Int64,1}, 1:2) with eltype Int64:
 1
 2

julia> df.x
2-element view(::Array{Int64,1}, 1:2) with eltype Int64:
 1
 2

julia> df = DataFrame()
0×0 DataFrame

julia> df[:, :x] = x
2-element Array{Int64,1}:
 1
 2

julia> df.x
2-element Array{Int64,1}:
 1
 2
```

oxinabox:
> It is not equivalent, as df[:, col] = v performs a copy while df[!, col] = v does not and copy can change the type of RHS.

Sorry I mispoke. Multiple uses of the word equivelent going around.
I Imean top say the matching operation is `df[:, col] = v` and `df[!, col] = v`,
rather than `df[:, col] = v` and  `df[!, col] .= v`.
And `df[:, col] = v` and `df[!, col] = v` do act generally similarly,
and differ in the way we want them to:
In particular one copies and the other doesn't.
(and if copying your vector has stide effects then indeed that happens)

bkamins:
Agreed :). In general thank you for discussing this as it shows (at least for me) that we should add what you originally proposed if we consider all these examples.

What is strictly equivalent after this PR is `df[:, col] .= v` and `df[!, col] .= v` if `col` is not present in `df`.

nilshg:
On balance I'm probably in favour, I think the usability aspect of this wins out. 

What I think we should have though is a sort of table in the docs that basically looks like

```
   |   !    |   :  |
--------------------
 = |   x    |   y  |
.= |   z    |   a  |
```
where `x, y, z, a`  show the results of using the respective combination of operators. In my opinion this aspect of DataFrames is actually a strength compared to R/pandas, where things mostly "just work" but are often a bit opaque, but we need to convince users of this by making it as easy as possible to grasp what's happening.

bkamins:
I have made probably 3 to 5 such tables in the PRs that lead to the current design. Unfortunately they got very complex very quickly (to the point of being not workable) and https://juliadata.github.io/DataFrames.jl/latest/lib/indexing/ was the best I could come up with.

Simply there are too many dimensions:
1. `=` or `.=`
2. view or not-view
3. LHS:
    * columns: one, many
    * rows: `!`, many, `:`, one
4. RHS:
    * columns: one, many
    * rows: `!`, many, `:`, one

so you get 2 * 2 * 2 * 4 * 2 * 4 = 256 combinations if you wanted to put them all in a table, which would have to have 6 dimensions.

Still - I would be more than happy to discuss how to make such a table (maybe simplifying/dropping some cases - so please comment if you have ideas on it).

What I took care about (subject to my bugs) is that https://juliadata.github.io/DataFrames.jl/latest/lib/indexing/ should be definitive and precise (i.e. if something does not work as explained there it is a bug).

bkamins:
Ah - and another dimension is that RHS does not have to be from DataFrames.jl but can be largely anything that supports broadcasting if you use `.=`.

pdeffebach:
If we merge this, then we are going to have to be prepared for lots of questions along the lines of 

"Why is my code slow when working with DataFrames? here is my code"

```
df[:, :x] = df[:, :y]
```

In which case there is an extra copy happening. We are copying both to retrieve `df.y` and to insert `df.x`.

oxinabox:
>"Why is my code slow when working with DataFrames? here is my code"
> ...
> In which case there is an extra copy happening. We are copying both to retrieve df.y and to insert df.x.

And then we answer those questions and it is fine.

Or maybe (and I suspect this will be the case), the cost of a extra copy of sequential memory is so low most of the time, that most people neither notice nor care.

In an ideal world something (be it Julia, LLVM, or elsewhere) would detect that one of those  copy's is never used and avoid making it.
(One way is to build copy on write tooling).

Better to have the friendly and safe API than anything else in my opinion.

bkamins:
I agree with @oxinabox. DataFrames.jl should be safe in the first place, as it is an "intro" package for people coming to Julia. When one learns more about the language one is prepared for more advanced concepts.

I teach DataFrames.jl a lot and I would prefer not to have to teach the meaning of `!` in my first class, but rather say - use `:` you already know and understand. When we get into the realm of `!` one should be able to understand the nuances of https://juliadata.github.io/DataFrames.jl/latest/lib/indexing/ (and if - hopefully one day - we get Julia and DataFrames.jl to be used on mass scale - many users will never reach this stage). Still we want to provide `!` as we have advanced users who want to benefit from using it when it is appropriate.

bkamins:
@nalimilan - so what do we do with this PR - merge or leave it open for now? 😄

bkamins:
After the discussion it will not go into 0.20 release. It is not decided yet if it will make into 1.0 (I would prefer it to be merged in 1.0 release given the comments above and the fact that for many years prior to the indexing redesign it used to work).

oxinabox:
Please yes can we have this.

docs/src/lib/indexing.md
```diff
@@ -133,9 +135,9 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
                        `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
                        (in the latter case column names must match);
 
-Note that only `df[!, col] = v` and `df.col = v` can be used to add a new column to a `DataFrame`.
-In particular as `df[:, col] = v` is an in-place operation it does not add a column `v` to a `DataFrame` if `col` is missing
-(an error is thrown if such operation is attempted).
+Note that `df[!, col] = v`, `df.col = v` and `df[:, col] = v` can be used to add a new column to a `DataFrame`.
```
nalimilan: Maybe this paragraph isn't needed anymore? This just repeats the general behavior applied to a special case. If you want to stress what is special about non-existing columns, better mention the fact that conversion may happen when the column exists, but (obviously) not when it doesn't.

src/other/broadcasting.jl
```diff
@@ -96,6 +96,14 @@ Base.maybeview(df::AbstractDataFrame, idx::CartesianIndex{2}) = df[idx]
 Base.maybeview(df::AbstractDataFrame, row::Integer, col::ColumnIndex) = df[row, col]
 Base.maybeview(df::AbstractDataFrame, rows, cols) = view(df, rows, cols)
 
+function Base.dotview(df::DataFrame, ::Colon, cols::ColumnIndex)
+    haskey(index(df), cols) && return view(df, :, cols)
+    if !(cols isa Symbol)
+        throw(ArgumentError("creating new columns using an integer index by broadcasting is disallowed"))
```
nalimilan: Is  "broadcasting" really relevant here? The same problem would happen with plain assignment, right?
```suggestion
        throw(ArgumentError("creating new columns using an integer index is disallowed"))
```

test/broadcasting.jl
```diff
@@ -1490,4 +1490,14 @@ end
     @test @views typeof(df[!, 1:2]) <: SubDataFrame
 end
 
+@testset "broadcasting of df[:, col] = vector" begin
```
nalimilan: Also test `= scalar`?

docs/src/lib/indexing.md
bkamins: I removed it. It was just a duplication indeed.

src/other/broadcasting.jl
bkamins: right

test/broadcasting.jl
bkamins: Fixed (also added one more other test)

bkamins:
@nalimilan - thank you for looking into this. I guess users want it 😄. The major benefit - if I understand the intention correctly - is that with this you do not have to teach users that `!` exists at all at the beginning. And you can only introduce it later as a performance optimization.

bkamins:
Something bad starts to happen on nightly, but probably this is a CI issue. Otherwise the PR passes after the updates.

test/broadcasting.jl
```diff
@@ -1490,14 +1490,18 @@ end
...
     z = ["a", "b", "c"]
     df[:, :z] .= z
     @test df.z == z
     @test df.z !== z
     @test_throws ArgumentError df[:, 6] .= z
     @test_throws MethodError df[:, 1] .= z
+
+    df = DataFrame(ones(3, 4))
+    z = fill("abc", 1, 1, 1)
+    @test_throws DimensionMismatch df[:, :z] .= z
```
nalimilan: This still doesn't test scalars, right? :-)

nalimilan:
> The major benefit - if I understand the intention correctly - is that with this you do not have to teach users that `!` exists at all at the beginning. And you can only introduce it later as a performance optimization.

Though note that new users who don't know about `!` won't be able to replace and existing column with another one, which matters if you want to change its type. So that only saves you a few hours until they complain. :-)

test/broadcasting.jl
bkamins: right - I have accidentally removed the right test (this was an additional test I wanted to add, but it seems I have overwritten the correct one with this)

bkamins:
CI passes except nightly - but I have looked at the problems and they seem unrelated

bkamins:
Thanks to all for contributing into the discussion and the review of this PR. It is going to be included in 0.20 release.

I hope the users will appreciate this addition.

pdeffebach:
@Bogumil thanks for the hard work on this. This will reduce the mental overhead of working with small data cleaning as we can now use `:` wherever we want. Thanks a ton!

Issue #1962, opened 2019-09-25
can't join more than two dataframes?


felluksch:
Hi, 
I would expect that you could join more than two DataFrames with `join(df1, df2, df3, on = :id)` since the documention doesn't explicitly say otherwise (and even states "We often need to combine **two or more** data sets").

```julia
using StatsKit

df1 = DataFrame(id = 1:10, df1_col2 = 1:10, df1_col3 = 1:10)
df2 = DataFrame(id = 1:10, df2_col2 = 1:10, df2_col3 = 1:10)
df3 = DataFrame(id = 1:10, df3_col2 = 1:10, df3_col3 = 1:10)
df = join(df1, df2, on = :id) # works
df = join(df1, df2, df3, on = :id) # doesn't work
arr = [df1, df2, df3]
df = join(arr..., on = :id) # doesn't work

df = join(df1, df2, on = :id) # workaround
df = join(df, df3, on = :id)
```
I find it just a bit weird that you have to split it up. Or am I missing something?
Thanks for the work you put into this package btw!

bkamins:
I have not implemented this, but I guess this can be added. The challenge is to make it efficient (in the worst case we can do it recursively for the time being). I guess you would want joining and the `kind` kwarg to be applied left-to-right order - right (like in `foldl`)?

felluksch:
If you want to invest into this, would be great of course. But maybe it'll suffice to just add a disclaimer to the doc and tell the people they have to write their own function. In my case I wrote something like this: 
```julia
function joinDataFrames()
    arr = [df1, df2, df3, df4]
    old_e = arr[1]
    df = arr[1]
    for e in arr
        new_e = e
        if old_e != new_e
            df = join(old_e, new_e, on = :id)
        end
        old_e = df
    end
    return df
end
println(joinDataFrames())
```
But of course it would be more convenient to have it as core functionality as I feel it's a common use case (think several timecoded files you want to join on their timestamps).

bkamins:
See #1963

felluksch:
Thanks a lot for the immediate implementation! I guess we're done here.

PR #1963, opened 2019-09-25
allow joins for more than two data frames

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1962

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -258,7 +258,7 @@ Join two or more `DataFrame` objects
    column named `Symbol` for whether a row appeared in only `df1` (`"left_only"`),
    only `df2` (`"right_only"`) or in both (`"both"`). If `Symbol` is already in use,
    the column name will be modified if `makeunique=true`.
-   This argument is supported when joining exactly two data frames.
+   This argument is only supported when joining exactly two data frames.
 
 * `validate` : whether to check that columns passed as the `on` argument
    define unique keys in each input data frame (according to [`isequal`](@ref)).
@@ -272,9 +272,9 @@ and `:right`), all columns of the returned data table will support missing value
 When merging `on` categorical columns that differ in the ordering of their levels, the
 ordering of the left `DataFrame` takes precedence over the ordering of the right `DataFrame`.
 
-If there are more than two data frames passed to `join` the joining is performed
+If more than two data frames are passed, the join is performed
 recursively with left associativity.
-Also in this case `indicator` keyword argument is not supported.
+In this case the `indicator` keyword argument is not supported.
 
 ### Result
 
@@ -432,8 +432,8 @@ function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame,
                    on::Union{Symbol, AbstractVector{Symbol}} = Symbol[],
                    kind::Symbol = :inner, makeunique::Bool=false,
                    validate::Union{Pair{Bool, Bool}, Tuple{Bool, Bool}}=(false, false))
-    if !(kind in [:inner, :outer, :cross])
-        throw(ArgumentError("Only inner, outer, and cross join is supported when " *
+    if !(kind in (:inner, :outer, :cross))
+        throw(ArgumentError("Only inner, outer, and cross joins are supported when " *
                             "joining more than two data frames"))
     end
     join(join(df1, df2, on=on, kind=kind, makeunique=makeunique, validate=validate),
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -216,19 +216,22 @@ end
 """
     join(df1, df2; on = Symbol[], kind = :inner, makeunique = false,
          indicator = nothing, validate = (false, false))
+    join(df1, df2, dfs...; on = Symbol[], kind = :inner, makeunique = false,
+         validate = (false, false))
 
-Join two `DataFrame` objects
+Join two or more `DataFrame` objects
 
 ### Arguments
 
-* `df1`, `df2` : the two AbstractDataFrames to be joined
+* `df1`, `df2`, `dfs...` : the `AbstractDataFrames` to be joined
 
 ### Keyword Arguments
 
 * `on` : A column, or vector of columns to join df1 and df2 on. If the column(s)
     that df1 and df2 will be joined on have different names, then the columns
     should be `left => right` pairs, or a vector of such pairs.
-    `on` is a required argument for all joins except for `kind = :cross`
+    `on` is a required argument for all joins except for `kind = :cross`.
+    If more than two data frames are joined then only a vector of column names is allowed.
 
 * `kind` : the type of join, options include:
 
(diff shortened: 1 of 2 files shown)


nalimilan:
Why not. Though can you check that other software allows that? If not, there may be a reason.

I guess it would be more efficient to join data frames by pairs? Not sure how hard it would be to implement.

src/abstractdataframe/join.jl
```diff
@@ -417,6 +426,19 @@ function Base.join(df1::AbstractDataFrame,
     return joined
 end
 
+function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame,
+                   dfs::AbstractDataFrame...;
+                   on::Union{Symbol, AbstractVector{Symbol}} = Symbol[],
```
nalimilan: I guess we should disallow (vectors of) pairs? Also the empty default may give weird results: should it operate on columns that are common to all inputs (which isn't what it does in this PR)?

bkamins: I disallow `Pairs`/vectors of `Pairs` as you propose. Have I missed something?

An empty default to `on` leads to an error unless the join is `kind=:cross` in which case it is not allowed to pass a non-empty `on`.

In general - the behavior of `on` is the same for many as for two data frames - you have to be explicit what you want to join on.

src/abstractdataframe/join.jl
```diff
@@ -216,19 +216,22 @@ end
...
+* `df1`, `df2`, `dfs...` : the `AbstractDataFrames` to be joined
 
 ### Keyword Arguments
 
 * `on` : A column, or vector of columns to join df1 and df2 on. If the column(s)
     that df1 and df2 will be joined on have different names, then the columns
     should be `left => right` pairs, or a vector of such pairs.
-    `on` is a required argument for all joins except for `kind = :cross`
+    `on` is a required argument for all joins except for `kind = :cross`.
+    If more than two data frames are joined then only a vector of column names is allowed.
```
nalimilan: Only a vector, or a single name?

bkamins: also a single name. Will fix.

bkamins:
This PR should be good for a final review. Thank you!

src/abstractdataframe/join.jl
```diff
@@ -252,6 +258,7 @@ Join two `DataFrame` objects
    column named `Symbol` for whether a row appeared in only `df1` (`"left_only"`),
    only `df2` (`"right_only"`) or in both (`"both"`). If `Symbol` is already in use,
    the column name will be modified if `makeunique=true`.
+   This argument is supported when joining exactly two data frames.
```
nalimilan:
```suggestion
   This argument is only supported when joining exactly two data frames.
```

src/abstractdataframe/join.jl
```diff
@@ -263,7 +270,11 @@ For the three join operations that may introduce missing values (`:outer`, `:lef
 and `:right`), all columns of the returned data table will support missing values.
 
 When merging `on` categorical columns that differ in the ordering of their levels, the
-ordering of the left `DataFrame` takes precedence over the ordering of the right `DataFrame`
+ordering of the left `DataFrame` takes precedence over the ordering of the right `DataFrame`.
+
+If there are more than two data frames passed to `join` the joining is performed
```
nalimilan:
```suggestion
If more than two data frames are passed, the join is performed
```

src/abstractdataframe/join.jl
```diff
@@ -263,7 +270,11 @@ For the three join operations that may introduce missing values (`:outer`, `:lef
...
+recursively with left associativity.
+Also in this case `indicator` keyword argument is not supported.
```
nalimilan:
```suggestion
In this case the `indicator` keyword argument is not supported.
```

src/abstractdataframe/join.jl
```diff
@@ -417,6 +427,19 @@ function Base.join(df1::AbstractDataFrame,
     return joined
 end
 
+function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame,
+                   dfs::AbstractDataFrame...;
+                   on::Union{Symbol, AbstractVector{Symbol}} = Symbol[],
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   validate::Union{Pair{Bool, Bool}, Tuple{Bool, Bool}}=(false, false))
+    if !(kind in [:inner, :outer, :cross])
```
nalimilan:
```suggestion
    if !(kind in (:inner, :outer, :cross))
```

src/abstractdataframe/join.jl
```diff
@@ -417,6 +427,19 @@ function Base.join(df1::AbstractDataFrame,
...
+        throw(ArgumentError("Only inner, outer, and cross join is supported when " *
```
nalimilan:
```suggestion
        throw(ArgumentError("Only inner, outer, and cross joins are supported when " *
```

src/abstractdataframe/join.jl
```diff
@@ -263,7 +270,11 @@ For the three join operations that may introduce missing values (`:outer`, `:lef
 and `:right`), all columns of the returned data table will support missing values.
 
 When merging `on` categorical columns that differ in the ordering of their levels, the
-ordering of the left `DataFrame` takes precedence over the ordering of the right `DataFrame`
+ordering of the left `DataFrame` takes precedence over the ordering of the right `DataFrame`.
+
+If there are more than two data frames passed to `join` the joining is performed
+recursively with left associativity.
```
nalimilan: What does this imply in practice? I haven't thought about this too deeply, but I guess it could be more efficient for future optimizations to perform joins in a different order in some cases. I guess that wouldn't change the result as long as we use the same order for columns?

bkamins: I have deliberately restricted this feature to `:inner`, `:outer` and `:cross` to make sure that the only difference would be:
1. efficiency (as you note - but I have left it for later)
2. column names in case `makeunique=true` (different columns might get renamed) - this can be handled in future more efficient implementation, as we have to anyway guarantee predictable column naming result (that is why I specify the contract - this allows the user to know how the columns might get renamed)
3. possibly ordering of rows in `:outer` and `:cross` joins (but again - this can be handled in more efficient implementation, and again - we have to make sure we have a contract here)

On the other hand other kinds of joins either make little sense or would produce different results (e.g. `:right` join would produce different values of rows in the output depending on the order of joins) - that is why I excluded these options (and I think no one really would need them in practice).

bkamins:
Thank you for the fixes!

src/abstractdataframe/join.jl
nalimilan: OK. Any data/thoughts about whether other implementations allow passing several data sets and why?

bkamins: Pandas `join` allows multiple data frames: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.join.html.

In R there is `merge_all` and `merge_recurse` in reshape.

In general I think it is easy to apply `foldl` on a vector of data frames to achieve what is wanted so it was not a top priority to have.
However, I do not see a problem with supporting it, and that is why I thought it is OK to have this PR.

nalimilan: OK, cool.

PR #1968, opened 2019-09-27
add groupindices for AbstractVectors

matthieugomez:
This pull request adds `groupindices` for multiple vectors. It corresponds loosely to functions such as [interaction](https://www.rdocumentation.org/packages/base/versions/3.6.1/topics/interaction) in R and [group](https://www.stata.com/support/faqs/data-management/creating-group-identifiers/) in Stata.

On the one hand, it's a bit weird to have something like this `DataFrames` since it does not act or return Dataframes. On the other hand, defining this function in this package allows one to benefit from the code performance of `row_group_slots` (and its future improvement).

Another point is that maybe the function should return a PooledArray. This is because functions that use the result of `groupindices` typically do a grouping operation, and therefore are interested in the group structure of the array. In this case, the function could be written
```julia
function groupindices(cols...)
    if !isa(cols, Tuple{Vararg{AbstractVector}})
        throw(ArgumentError("Arguments of groupindices should be AbstractVectors"))
    end
    if isempty(cols)
        throw(ArgumentError("groupindices was called without argument")) 
    end
    n = length(first(cols))
    if any(length(col) != n for col in cols)
        throw(DimensionMismatch("Arguments of groupindices should have the same length"))
    end
    groups = Vector{Int}(undef, n)
    ngroups, rhashes, gslots, sorted = row_group_slots(cols, Val(false), groups, true)
    pool = zeros(Union{Int, Missing}, ngroups)
    for x in groups
        p = pool[x]
        if !ismissing(p) && p == 0
            pool[x] = x == 1 ? missing : x - 1
        end
    end
    if !ismissing(pool[1])
        # no missing value was found
        deleteat!(pool, 1)
        groups .-= 1
    end
    invpool = Dict{Union{Int, Missing}, Int}(pool[i] => i for i in eachindex(pool))
    PooledArray(PooledArrays.RefArray(groups), invpool, pool)
end
```

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -1235,6 +1235,14 @@ function DataFrame(gd::GroupedDataFrame; copycols::Bool=true)
 end
 
 """
+    groupvars(gd::GroupedDataFrame)
+
+Return a vector of column names in `parent(gd)` used for grouping.
+"""
+groupvars(gd::GroupedDataFrame) = _names(gd)[gd.cols]
+
+
+"""
     groupindices(gd::GroupedDataFrame)
 
 Return a vector of group indices for each row of `parent(gd)`.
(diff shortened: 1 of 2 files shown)


nalimilan:
I'd find it a bit weird (as you note) to ship this function in DataFrames. You can already get the same result with `groupindices(groupby(DataFrame(cols...), All()))`. If you want a function which doesn't rely on `DataFrame` at all, it should probably live in a lower-level package like TableOperations or SplitApplyCombine. We could even consider moving grouping code from DataFrames to such a package, and use it from there.

bkamins:
Good point - I agree that this can live in a more lower-level package (I felt the same but I could not express this as precisely).

bkamins:
@matthieugomez - are you OK to close this PR and move it to some more general package?

PR #1969, opened 2019-10-01
update documentation of column selection

bkamins:
The old docs were broken (the order of examples was wrong and corrupted the value of `df`). Additionally explain how to move columns to front/end of a data frame (a recently asked question that I think is common enough to include it in the examples).


docs/src/man/getting_started.md
```diff
@@ -424,6 +407,49 @@ julia> df[(df.A .> 500) .& (300 .< df.C .< 400), :]
...
+1×2 DataFrame
+│ Row │ x1    │ x2    │
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 2     │
+```
+
+Finally, you can use `Not` and `All` selectors in more complex cases. The following examples move all columns containing `"x"` respectively to the front and to the end of a data frame:
+```
+julia> df = DataFrame(r="r", x1="x1", x2="x2", y="y")
```
nalimilan: Maybe use number values instead of repeating column names?

bkamins: OK. I will fix this. Anyway, this PR should be updated after https://github.com/JuliaData/DataFrames.jl/pull/1971 is merged.

It is in general non blocking for 0.20 release anyway. Thank you!

bkamins:
After #1971 what made sense to leave in this PR are examples of moving columns to the front/back (this is a pretty often needed functionality) and references to `select` and `select!`.

bkamins:
Thank you!

PR #1972, opened 2019-10-02
clean up permutecols! and tests

bkamins:
This PR:
1. removes `permutecols!` from dataframe.jl (I could swear I deleted it, but somehow it stayed)
2. fixes tests that fail on Julia nightly due to float printing changes
3. fixes a tests that fails due to recent CategoricalArrays.jl change of levels handling

@nalimilan could you have a look at it. The third point (CategoricalArrays.jl) makes all new PRs fail CI (points 1 and 2 do not lead to CI errors but clearly should be fixed)


nalimilan:
Thanks. AFAICT we have to require CategoricalArrays 0.7.0 for tests to pass. I've added a commit.

bkamins:
Thank you for the fix. This requires conflicted with CSV.jl as it is still not registered with CategoricalArrays 0.7 (hopefully https://github.com/JuliaRegistries/General/pull/4098 will be merged soon).

PR #1974, opened 2019-10-05
Allow permutaion of names in rename!

innerlee:
Before:
```julia
julia> df = DataFrame(A = 1:3, B = 'A':'C')
3×2 DataFrame
│ Row │ A     │ B    │
│     │ Int64 │ Char │
├─────┼───────┼──────┤
│ 1   │ 1     │ 'A'  │
│ 2   │ 2     │ 'B'  │
│ 3   │ 3     │ 'C'  │

julia> rename!(df, :A=>:B, :B=>:A)
ERROR: Tried renaming A to B, when B already exists in the Index.
Stacktrace:
...
```

This pr:
```julia
julia> df = DataFrame(A = 1:3, B = 'A':'C')
3×2 DataFrame
│ Row │ A     │ B    │
│     │ Int64 │ Char │
├─────┼───────┼──────┤
│ 1   │ 1     │ 'A'  │
│ 2   │ 2     │ 'B'  │
│ 3   │ 3     │ 'C'  │

julia> rename!(df, :A=>:B, :B=>:A)
3×2 DataFrame
│ Row │ B     │ A    │
│     │ Int64 │ Char │
├─────┼───────┼──────┤
│ 1   │ 1     │ 'A'  │
│ 2   │ 2     │ 'B'  │
│ 3   │ 3     │ 'C'  │
```


bkamins:
This is a nice proposal.
Can you please rebase it against current master (otherwise it will keep failing CI - I have just merged a PR that fixes this)?
After it is rebased I will review it today.
Thank you!

bkamins:
I have given a quick glance at the PR. Is your proposal is [atomic](https://en.wikipedia.org/wiki/Atomicity_(database_systems))?

When we make this change it would probably good to ensure this. What I mean is either `rename!` succeeds or if it fails it should not render the original `DataFrame` partially changed.

innerlee:
The current implementation will cause the df not able to print when wrong:

```julia
julia> df = DataFrame(A = 1:3, B = 'A':'C')
3×2 DataFrame
│ Row │ A     │ B    │
│     │ Int64 │ Char │
├─────┼───────┼──────┤
│ 1   │ 1     │ 'A'  │
│ 2   │ 2     │ 'B'  │
│ 3   │ 3     │ 'C'  │

julia> rename!(df, :A => :B)
ERROR: Tried renaming to B, when it already exists in the Index.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] rename!(::DataFrames.Index, ::Array{Pair{Symbol,Symbol},1}) at /mnt/hdd/dev/DataFrames.jl/src/other/index.jl:62
 [3] rename!(::DataFrames.Index, ::Pair{Symbol,Symbol}) at /mnt/hdd/dev/DataFrames.jl/src/other/index.jl:67
 [4] rename!(::DataFrame, ::Pair{Symbol,Symbol}) at /mnt/hdd/dev/DataFrames.jl/src/abstractdataframe/abstractdataframe.jl:116
 [5] top-level scope at none:0

julia> df
Error showing value of type DataFrame:
ERROR: AssertionError: length(idx.names) == length(idx.lookup) == ncols
Stacktrace:
 [1] _check_consistency(::DataFrame) at /mnt/hdd/dev/DataFrames.jl/src/dataframe/dataframe.jl:290
 [2] #_show#343(::Bool, ::Bool, ::Bool, ::Symbol, ::Bool, ::Nothing, ::Function, ::IOContext{REPL.Terminals.TTYTerminal}, ::DataFrame) at /mnt/hdd/dev/DataFrames.jl/src/abstractdataframe/show.jl:517
 [3] #_show at ./none:0 [inlined]
 [4] #show#344 at /mnt/hdd/dev/DataFrames.jl/src/abstractdataframe/show.jl:609 [inlined]
 [5] #show at ./none:0 [inlined]
 [6] #show#357(::Bool, ::Bool, ::Bool, ::Symbol, ::Bool, ::Function, ::IOContext{REPL.Terminals.TTYTerminal}, ::MIME{Symbol("text/plain")}, ::DataFrame) at /mnt/hdd/dev/DataFrames.jl/src/abstractdataframe/io.jl:53
 [7] show(::IOContext{REPL.Terminals.TTYTerminal}, ::MIME{Symbol("text/plain")}, ::DataFrame) at /mnt/hdd/dev/DataFrames.jl/src/abstractdataframe/io.jl:53
 [8] display(::REPL.REPLDisplay, ::MIME{Symbol("text/plain")}, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/REPL.jl:131
 [9] display(::REPL.REPLDisplay, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/REPL.jl:135
 [10] display(::Any) at ./multimedia.jl:287
 [11] #invokelatest#1 at ./essentials.jl:742 [inlined]
 [12] invokelatest at ./essentials.jl:741 [inlined]
 [13] print_response(::IO, ::Any, ::Any, ::Bool, ::Bool, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/REPL.jl:155
 [14] print_response(::REPL.AbstractREPL, ::Any, ::Any, ::Bool, ::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/REPL.jl:140
 [15] (::getfield(REPL, Symbol("#do_respond#38")){Bool,getfield(REPL, Symbol("##48#57")){REPL.LineEditREPL},REPL.LineEditREPL,REPL.LineEdit.Prompt})(::Any, ::Any, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/REPL.jl:714
 [16] #invokelatest#1 at ./essentials.jl:742 [inlined]
 [17] invokelatest at ./essentials.jl:741 [inlined]
 [18] run_interface(::REPL.Terminals.TextTerminal, ::REPL.LineEdit.ModalInterface, ::REPL.LineEdit.MIState) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/LineEdit.jl:2273
 [19] run_frontend(::REPL.LineEditREPL, ::REPL.REPLBackendRef) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/REPL.jl:1035
 [20] run_repl(::REPL.AbstractREPL, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/REPL/src/REPL.jl:192
 [21] (::getfield(Base, Symbol("##734#736")){Bool,Bool,Bool,Bool})(::Module) at ./client.jl:362
 [22] #invokelatest#1 at ./essentials.jl:742 [inlined]
 [23] invokelatest at ./essentials.jl:741 [inlined]
 [24] run_main_repl(::Bool, ::Bool, ::Bool, ::Bool, ::Bool) at ./client.jl:346
 [25] exec_options(::Base.JLOptions) at ./client.jl:284
 [26] _start() at ./client.jl:436
```

Do we need to restore it to original when it fails?

bkamins:
Yes - I think `rename!` should be atomic - this is what I have just commented in parallel 😄.

innerlee:
yea we thought the same issue. Note the original implementation is not atomic also:

```julia
julia> df = DataFrame(A = 1:3, B = 'A':'C', C = [:x, :y, :z])
3×3 DataFrame
│ Row │ A     │ B    │ C      │
│     │ Int64 │ Char │ Symbol │
├─────┼───────┼──────┼────────┤
│ 1   │ 1     │ 'A'  │ x      │
│ 2   │ 2     │ 'B'  │ y      │
│ 3   │ 3     │ 'C'  │ z      │

julia> rename!(df, :C=>:D, :B=>:A, :A=>:B)
ERROR: Tried renaming B to A, when A already exists in the Index.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] rename!(::DataFrames.Index, ::Array{Pair{Symbol,Symbol},1}) at /home/lizz/.julia/packages/DataFrames/yH0f6/src/other/index.jl:52
 [3] rename!(::DataFrames.Index, ::Pair{Symbol,Symbol}, ::Pair{Symbol,Symbol}, ::Vararg{Pair{Symbol,Symbol},N} where N) at /home/lizz/.julia/packages/DataFrames/yH0f6/src/other/index.jl:60
 [4] rename!(::DataFrame, ::Pair{Symbol,Symbol}, ::Pair{Symbol,Symbol}, ::Vararg{Pair{Symbol,Symbol},N} where N) at /home/lizz/.julia/packages/DataFrames/yH0f6/src/abstractdataframe/abstractdataframe.jl:117
 [5] top-level scope at none:0

julia> df
3×3 DataFrame
│ Row │ A     │ B    │ D      │
│     │ Int64 │ Char │ Symbol │
├─────┼───────┼──────┼────────┤
│ 1   │ 1     │ 'A'  │ x      │
│ 2   │ 2     │ 'B'  │ y      │
│ 3   │ 3     │ 'C'  │ z      │
```

How about simply backup the original index and restore if goes wrong?

src/other/index.jl
```diff
@@ -46,14 +46,36 @@ function names!(x::Index, nms::Vector{Symbol}; makeunique::Bool=false)
...
 function rename!(x::Index, nms)
+    xbackup = copy(x)
+    toholder = Dict()
     for (from, to) in nms
         from == to && continue # No change, nothing to do
-        if haskey(x, to)
+        if !haskey(xbackup, from)
+            copy!(x.lookup, xbackup.lookup)
+            x.names .= xbackup.names
+            error("Tried renaming $from to $to, when $from does not exist in the Index.")
```
bkamins: Can you please switch here and below to `throw(ArgumentError(` to raise a more specific exception?

src/other/index.jl
```diff
@@ -46,14 +46,36 @@ function names!(x::Index, nms::Vector{Symbol}; makeunique::Bool=false)
 end
 
 function rename!(x::Index, nms)
+    xbackup = copy(x)
+    toholder = Dict()
```
bkamins: Why not `Dict{Symbol,Int}()?

src/other/index.jl
```diff
@@ -46,14 +46,36 @@ function names!(x::Index, nms::Vector{Symbol}; makeunique::Bool=false)
...
-        x.lookup[to] = col = pop!(x.lookup, from)
+        if haskey(x, to)
+            toholder[to] = x.lookup[to]
+        end
+        if !haskey(toholder, from) && !haskey(x.lookup, from)
+            copy!(x.lookup, xbackup.lookup)
+            x.names .= xbackup.names
+            error("Tried renaming $from to multiple values.")
+        end
+        x.lookup[to] = col = haskey(toholder, from) ? pop!(toholder, from) : pop!(x.lookup, from)
```
bkamins:
```suggestion
        col = pop!(toholder, from, pop!(x.lookup, from))
        x.lookup[to] = col
```

We try to avoid doing multiple assignments in one line. Also I have shortened the expression a bit (but I am OK to leave it as you implemented it).

test/dataframe.jl
```diff
@@ -1127,6 +1127,16 @@ end
...
+
+    df = DataFrame(A = 1:3, B = 'A':'C', C = [:x, :y, :z])
+    @test rename!(df, :A => :B, :B => :A) === df
+    @test names(df) == [:B, :A, :C]
+    @test rename!(df, :A => :B, :B => :A, :C => :D) === df
+    @test names(df) == [:A, :B, :D]
+    @test rename!(df, :A => :B, :B => :C, :D => :A) === df
+    @test names(df) == [:B, :C, :A]
+    @test rename!(df, :A => :B, :B => :C, :C => :A) === df
+    @test names(df) == [:C, :A, :B]
```
bkamins: Can you please also add tests for each case of the error.
Ideally also randomized tests could be added here to make sure we cover all corner cases as the logic is not simple.

src/other/index.jl
innerlee: Interesting, the default argument is not lazy:

```julia
julia> a = Dict(:B=>1,:C=>3)
Dict{Symbol,Int64} with 2 entries:
  :B => 1
  :C => 3

julia> b = Dict(:B=>2)
Dict{Symbol,Int64} with 1 entry:
  :B => 2

julia> pop!(a, :B, pop!(b, :B))
1

julia> a
Dict{Symbol,Int64} with 1 entry:
  :C => 3

julia> b
Dict{Symbol,Int64} with 0 entries
```

innerlee:
The new implementation is a breaking change, because it changed the old behavior ``New names are processed sequentially. A new name must not already exist in the `DataFrame` at the moment an attempt to rename a column is performed.``

src/other/index.jl
bkamins: You are right. This is my mistake.

bkamins:
It is breaking, but I would say that it is "mildly breaking" - i.e. we allow something that was an error earlier, therefore we do not have to go through deprecation.

src/other/index.jl
```diff
@@ -46,14 +46,41 @@ function names!(x::Index, nms::Vector{Symbol}; makeunique::Bool=false)
 end
 
 function rename!(x::Index, nms)
+    xbackup = copy(x)
+    processedfrom = Set(Symbol[])
```
bkamins:
```suggestion
    processedfrom = Set{Symbol}()
```

src/other/index.jl
```diff
@@ -46,14 +46,41 @@ function names!(x::Index, nms::Vector{Symbol}; makeunique::Bool=false)
...
+    processedto = Set(Symbol[])
```
bkamins:
```suggestion
    processedto = Set{Symbol}()
```

bkamins:
Thinking more about it - it is breaking even if something worked (I guess this is what you meant, like `:A=>:B, :B=>:C` with a data frame having only column `:A`, but I would assume that this was a rare/non-existent use case, and writing a correct deprecation code would be hard).

innerlee:
yes the mental model of the two approaches are different. The original one is some kind of online processing, the current version is not

innerlee:
Well, actually the new model is a preparation for https://github.com/JuliaData/DataFrames.jl/pull/1975

innerlee:
the test passed locally.

innerlee:
The 1.0 CI error is due to
```julia
julia> rename!(df, :X => :Y)
ERROR: MethodError: no method matching copy!(::Dict{Symbol,Int64}, ::Dict{Symbol,Int64})
Stacktrace:
 [1] rename!(::DataFrames.Index, ::Array{Pair{Symbol,Symbol},1}) at /mnt/hdd/dev/DataFrames.jl/src/other/index.jl:68
 [2] rename!(::DataFrames.Index, ::Pair{Symbol,Symbol}) at /mnt/hdd/dev/DataFrames.jl/src/other/index.jl:87
 [3] rename!(::DataFrame, ::Pair{Symbol,Symbol}) at /mnt/hdd/dev/DataFrames.jl/src/abstractdataframe/abstractdataframe.jl:116
 [4] top-level scope at none:0
```

src/other/index.jl
```diff
@@ -46,14 +46,41 @@ function names!(x::Index, nms::Vector{Symbol}; makeunique::Bool=false)
 end
 
 function rename!(x::Index, nms)
+    xbackup = copy(x)
+    processedfrom = Set{Symbol}()
+    processedto = Set{Symbol}()
+    toholder = Dict{Symbol,Int}()
     for (from, to) in nms
+        if from ∈ processedfrom
+            merge!(empty!(x.lookup), xbackup.lookup)
```
nalimilan: Isn't this just `copy!(x.lookup, xbackup.lookup)`? (Use `Future.copy!` on Julia 1.0.)

Or maybe even simpler: outside of the loop, do
```julia
   if !allunique(keys(nms))
...
   if !allunique(values(nms))
```

innerlee: Yes it is exactly `copy!`. 

```julia
julia> nms = [:a=>:b, :a=>:c, :b=>:c]
3-element Array{Pair{Symbol,Symbol},1}:
 :a => :b
 :a => :c
 :b => :c

julia> allunique(keys(nms))
true
```

nalimilan: Yeah replace `keys(nms)` with `first.(nms)` (or `(first(x) for x in nms)` to avoid an allocation).

innerlee: Implementation-wise, [`allunique`](https://github.com/JuliaLang/julia/blob/c6da87ff4bc7a855e217856757ad3413cf6d1f79/base/set.jl#L365-L373) is the same as here. 
However, we can get exact items that make things error by doing the check manually.

bkamins: Only `(first(x) for x in nms)` as broadcasting over dictionaries is not suported

nalimilan: Ah, indeed. Though please use `copy!` at least, that will make the code simpler in the long run.

test/dataframe.jl
```diff
@@ -1127,6 +1127,33 @@ end
...
+    @test names(df) == [:A, :B, :D]
+    @test rename!(df, :A => :B, :B => :C, :D => :A) === df
+    @test names(df) == [:B, :C, :A]
+    @test rename!(df, :A => :C, :B => :A, :C => :B) === df
+    @test names(df) == [:A, :B, :C]
+    @test rename!(df, :A => :A, :B => :B, :C => :C) === df
+    @test names(df) == [:A, :B, :C]
+
+    @test_throws ArgumentError rename!(df, :X => :Y)
+    @test names(df) == [:A, :B, :C]
```
nalimilan: Maybe also test that `df` is equal to what we expect. That way you'll also check that the index's `lookup` field is correct.

bkamins: If you add tests can you also add the randomized test I have proposed so that in the future it is kept being checked also?

src/other/index.jl
innerlee: I added a line of note on the `merge! ∘ empty!` to avoid adding a new dependency of `Future` for this package.
Since the package is not specially designed for 1.0, the `Future.blah` thing looks strange for people of 1.1 or after.

bkamins:
Thank you for the fixes. Regarding the `copy!` issue I leave it to @nalimilan to decide, as he has a better feel for such things.

In general please note (this is a general comment to make sure we are on the same page in the future) that DataFrames.jl intends to support Julia 1.0 at least as long as it is supported by JuliaLang team.

src/other/index.jl
nalimilan: `Future` is in the stdlib, so that's OK to have it as a dependency. And `Future.copy!` is much easier to understand to people used to Julia >= 1.1 than `merge!(empty(...), ...)`.

bkamins:
@nalimilan - OK to merge this?

Then #1985 would be rebased against the changes in this PR as it also imports `copy!` from `Future` (CC @Ellipse0934).

nalimilan:
Thanks @innerlee!

PR #1977, opened 2019-10-08
allow more flexible cols in groupby and in particular handle empty cols

bkamins:
Allow more flexible `cols` in `groupby` and `by` (e.g. `Regex`, `Not`, `All`).
In particular also correctly handle the case when cols produces an empty selection of columns. In this case we create a single group having all columns.

### DIFF
--- a/src/groupeddataframe/grouping.jl
+++ b/src/groupeddataframe/grouping.jl
@@ -39,7 +39,7 @@ groupby(d::AbstractDataFrame, cols; sort=false, skipmissing=false)
 ### Arguments
 
 * `df` : an `AbstractDataFrame` to split
-* `cols` : data table columns to group by
+* `cols` : data frame columns to group by
 * `sort` : whether to sort rows according to the values of the grouping columns `cols`
 * `skipmissing` : whether to skip rows with `missing` values in one of the grouping columns `cols`
 
@@ -52,6 +52,10 @@ A `GroupedDataFrame` : a grouped view into `df`
 An iterator over a `GroupedDataFrame` returns a `SubDataFrame` view
 for each grouping into `df`.
 Within each group, the order of rows in `df` is preserved.
+
+`cols` can be any valid data frame indexing expression.
+In particular if it is an empty vector then a single group `GroupedDataFrame`
+is created.
 
 A `GroupedDataFrame` also supports
 indexing by groups, `map` (which applies a function to each group)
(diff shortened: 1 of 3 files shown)


test/grouping.jl
```diff
@@ -1200,4 +1200,25 @@ end
...
+    gdf = groupby(df, r"x")
+    @test groupvars(gdf) == [:x1, :x2]
+    @test groupindices(gdf) == [1,2,3]
+
+    gdf = groupby(df, Not(r"x"))
+    @test groupvars(gdf) == [:y]
+    @test groupindices(gdf) == [1,2,3]
+
+    gdf = groupby(df, [])
+    @test groupvars(gdf) == Symbol[]
```
```suggestion
    @test groupvars(gdf) == []
```
(`Symbol` doesn't make any difference here.)

test/grouping.jl
```diff
@@ -1200,4 +1200,25 @@ end
...
+    @test groupindices(gdf) == [1,1,1]
+
+    gdf = groupby(df, r"z")
+    @test groupvars(gdf) == Symbol[]
```
```suggestion
    @test groupvars(gdf) == []
```

src/groupeddataframe/grouping.jl
```diff
@@ -39,7 +39,7 @@ groupby(d::AbstractDataFrame, cols; sort=false, skipmissing=false)
 ### Arguments
 
 * `df` : an `AbstractDataFrame` to split
-* `cols` : data table columns to group by
+* `cols` : data frame columns to group by
```
nalimilan: Really funny... I wonder how this one survived.

test/grouping.jl
```diff
@@ -1200,4 +1200,25 @@ end
     end
 end
 
+@testset "non standard cols arguments" begin
+    df = DataFrame(x1=[1,2,2], x2=[1,1,2], y=[1,2,3])
+    gdf = groupby(df, r"x")
```
```suggestion
    gdf = groupby_checked(df, r"x")
```

test/grouping.jl
```diff
@@ -1200,4 +1200,25 @@ end
...
+    @test groupvars(gdf) == [:x1, :x2]
+    @test groupindices(gdf) == [1,2,3]
+
+    gdf = groupby(df, Not(r"x"))
```
```suggestion
    gdf = groupby_checked(df, Not(r"x"))
```

test/grouping.jl
```diff
@@ -1200,4 +1200,25 @@ end
...
+    @test groupvars(gdf) == [:y]
+    @test groupindices(gdf) == [1,2,3]
+
+    gdf = groupby(df, [])
+    @test groupvars(gdf) == Symbol[]
+    @test groupindices(gdf) == [1,1,1]
+
+    gdf = groupby(df, r"z")
```
```suggestion
    gdf = groupby_checked(df, r"z")
```

bkamins:
@nalimilan - thank you for looking into it! (and as always keen 👁)

bkamins:
> I would expect some things to break with the current code, e.g. printing.

This is what I expected also but I have tested it locally and all worked. I will perform some more tests then (and add them to tests) and report back here what is the result (I will remove WIP in the PR name then).

bkamins:
I have added all tests I could think of + I have went through code to make sure all works correctly. It seems we can safely change this.

As an addition I have cleaned up `aggregate` a bit to be more consistent with the rest of the functions (as it seems it is going to stay in 1.0).

bkamins:
This should be good to have a look at now.

bkamins:
bump

src/groupeddataframe/grouping.jl
```diff
@@ -53,6 +53,10 @@ An iterator over a `GroupedDataFrame` returns a `SubDataFrame` view
 for each grouping into `df`.
 Within each group, the order of rows in `df` is preserved.
 
+`cols` can be any valid data frame indexing expression.
+In particular if it is an empty vector then a single group `GroupedDataFrame`
```
```suggestion
In particular if it is an empty vector then a single-group `GroupedDataFrame`
```

test/grouping.jl
```diff
@@ -1200,4 +1200,42 @@ end
...
+    @test gdf[:] == gdf
+    @test gdf[1:1] == gdf
+
+    @test map(nrow, gdf) == groupby_checked(DataFrame(x1=3), [])
+    @test map(:x2 => identity, gdf) == groupby_checked(DataFrame(x2_identity=[1,1,2]), [])
+    @test aggregate(df, sum) == aggregate(df, [], sum) == aggregate(df, 1:0, sum)
+    @test aggregate(df, sum) == aggregate(df, [], sum, sort=true, skipmissing=true)
+    @test DataFrame(gdf) == df
+
+    @test sprint(show, groupby(df, [])) == "GroupedDataFrame with 1 group based on key: \n" *
```
```suggestion
    @test sprint(show, groupby_checked(df, [])) == "GroupedDataFrame with 1 group based on key: \n" *
```

Also use triple quotes?

bkamins: OK - I wanted to check this without a wrapper as it tests printing only, but probably this should also pass.

bkamins:
@nalimilan - thank you for working on it.

PR #1979, opened 2019-10-09
Fix handling of hash collisions in row_group_slots

### LINKED ISSUE #1978: 32-bit BoundsError
```
Julia Version 1.2.0
Commit c6da87ff4b (2019-08-20 00:03 UTC)
Platform Info:
  OS: Linux (i686-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i9-9980H CPU @ 2.30GHz
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

I've run into an issue when trying to do a `join` with two large DataFrames. The CI runners for 32-bit versions for Julia `1.0, 1.1, 1.2` all fail. The 64-bit versions work perfectly fine:

Stacktrace from the CI:
```
BoundsError: attempt to access 1365559-element Array{Int32,1} at index [0]
  Stacktrace:
   [1] getindex at ./array.jl:731 [inlined]
   [2] group_rows(::DataFrame, ::Bool, ::Bool, ::Bool) at /mnt/builds/DcGs9yxw/0/{REDACTED}/depot/packages/DataFrames/0Em9Q/src/dataframerow/utils.jl:255
   [3] group_rows at /mnt/builds/DcGs9yxw/0/{REDACTED}/depot/packages/DataFrames/0Em9Q/src/dataframerow/utils.jl:248 [inlined]
   [4] #join#237(::Array{Symbol,1}, ::Symbol, ::Bool, ::Nothing, ::Tuple{Bool,Bool}, ::Function, ::DataFrame, ::DataFrame) at /mnt/builds/DcGs9yxw/0/{REDACTED}/depot/packages/DataFrames/0Em9Q/src/abstractdataframe/join.jl:344
   [5] (::getfield(Base, Symbol("#kw##join")))(::NamedTuple{(:on, :makeunique),Tuple{Array{Symbol,1},Bool}}, ::typeof(join), ::DataFrame, ::DataFrame) at ./none:0
...
```

I've been playing around in `VirtualBox` with an `Ubuntu 32-bit` instance. Below is an example of how I am using `DataFrames`. Note this example on my VirtualBox instance causes a `SIGABRT`.
```
using DataFrames
using DataFramesMeta

df_1 = DataFrame(1:2000000)
df_1 = @transform(df_1, time=first.(:A))
df_2 = DataFrame(1:2000000)
df_2 = @transform(df_2, time=first.(:A))

join(df_1, df_2, on=[:time, :A], makeunique=true)
```

After spending sometime and looking at [`group_rows`](https://github.com/JuliaData/DataFrames.jl/blob/v0.18.4/src/dataframerow/utils.jl#L246) I was able to create another example which forces the above stacktrace. It will always crash after `g_ix=57979`.
```
using DataFrames

df = DataFrame(A=1:2000000)
groups = Vector{Int}(undef, nrow(df))
ngroups, rhashes, gslots, sorted = DataFrames.row_group_slots(ntuple(i -> df[i], ncol(df)), Val(true), groups, false)
stops = zeros(Int, ngroups)

for g_ix in groups
    stops[g_ix] += 1
end
```

nalimilan:
Original code was correct but JuliaData/DataTables.jl#79 incorrectly moved `break` outside of the `if`.

Fixes #1978.

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -131,8 +131,8 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
                 elseif rhashes[i] == rhashes[g_row] # occupied slot, check if miss or hit
                     if isequal_row(cols, i, g_row) # hit
                         gix = groups !== nothing ? groups[g_row] : 0
+                        break
                     end
-                    break
                 end
                 slotix = slotix & szm1 + 1 # check the next slot
                 probe += 1
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -51,12 +51,23 @@ function groupby_checked(df::AbstractDataFrame, keys, args...; kwargs...)
 
         # correct start-end relations
         for i in eachindex(se)
+            firstkeys = gd.parent[gd.idx[se[i][1]], gd.cols]
+            # all grouping keys must be equal within a group
+            @assert all(j -> gd.parent[gd.idx[j], gd.cols] ≅ firstkeys, se[i][1]:se[i][2])
             @assert se[i][1] <= se[i][2]
             if i > 1
                 # the blocks returned by groupby must be continuous
                 @assert se[i-1][2] + 1 == se[i][1]
             end
         end
+
+        # all grouping keys must be equal within a group
+        for (s, e) in zip(gd.starts, gd.ends)
+            firstkeys = gd.parent[gd.idx[s], gd.cols]
+            @assert all(j -> gd.parent[gd.idx[j], gd.cols] ≅ firstkeys, s:e)
+        end
+        # all groups have different grouping keys
+        @test allunique(eachrow(gd.parent[gd.idx[gd.starts], gd.cols]))
     end
 
     gd
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -608,6 +608,13 @@ end
     end
 end
 
+@testset "grouping with hash collisions" begin
+    # Hash collisions are almost certain on 32-bit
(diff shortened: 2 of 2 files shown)


mattBrzezinski:
Good morning.

I can verify this fixes the issue later today, and write some unit tests for it as well!

@nalimilan would you like me to push commits to this branch? Or just create a new one for the unit-tests?

nalimilan:
I've added a test: can you just check that it reproduces the problem on 32-bit without the patch? Thanks!

bkamins:
Now I have four Julia versions installed in parallel to be able to develop DataFrames.jl 😄.

The test correctly reproduces the error prior to the change in this PR:
```
julia> gd = groupby(df, :A)
ERROR: ReadOnlyMemoryError()
Stacktrace:
 [1] getindex at .\array.jl:728 [inlined]
 [2] group_rows(::DataFrame, ::Bool, ::Bool, ::Bool)
 [3] #groupby#157(::Bool, ::Bool, ::typeof(groupby), ::DataFrame, ::Array{Symbol,1})
 [4] #groupby at .\none:0 [inlined]
 [5] #groupby#158 [inlined]
 [6] groupby(::DataFrame, ::Symbol)
 [7] top-level scope at REPL[6]:1
```

nalimilan:
I've added some more checks to make sure that even if no error happens for some reason, we don't mix two different values that hash equal in the same group.

mattBrzezinski:
Just tested this and the issue is resolved! :)

Issue #1861, opened 2019-06-25
Improve select and select! performance with `Not`


bkamins:
`select` and `select!` internally removes columns. Therefore it is relatively easy to add special methods for `Not` column selector that will have a much better performance than current implementation that does `setdiff` twice.

Ellipse0934:
Hi,
I tried really hard to improve performance using the `Not` selector but could not do it. I could achieve a little worse than the current performance. Unfortunately `Not` is having too many allocations internally which gets worse with the DataFrame's size. (There is an issue at InvInd/issues which mentions this)
I did manage to improve the performance significantly by rewriting `select!`. Currently we generate a list of columns to be removed then remove them sequentially and each removal is  O(df.columns). After which there is the cost of using `permute!`.
Instead how about swapping in-place in df.columns as each swap is O(1) and we need to make |inds| swaps at max. Then just resizing df.columns to |inds| to purge unwanted cols. (And same swapping for colindices)
```julia
function mselect!(df::DataFrame, inds::AbstractVector{Int})
    if isempty(inds)
        empty!(_columns(df))
        empty!(index(df))
        return df
    end
    indmin, indmax = extrema(inds)
    if indmin < 1
        throw(ArgumentError("indices must be positive"))
    end
    if indmax > ncol(df)
        throw(ArgumentError("indices must not be greater than number of columns"))
    end
    if !allunique(inds)
        throw(ArgumentError("indices must not contain duplicates"))
    end
    
    itr = 1
    ord = collect(1:length(_columns(df)))
    pos = copy(ord)
    for i = inds
            ord[itr], ord[pos[i]] = ord[pos[i]], ord[itr]
            _columns(df)[itr], _columns(df)[pos[i]] = _columns(df)[pos[i]], _columns(df)[itr]
            index(df).names[itr], index(df).names[pos[i]] = index(df).names[pos[i]], index(df).names[itr]
            pos[ord[pos[i]]], pos[ord[itr]] = pos[ord[itr]], pos[ord[pos[i]]]
            itr += 1
    end
    l = length(index(df).names)
    for i = 1:length(inds)
          index(df).lookup[index(df).names[i]] = i
    end
    for i = (length(inds) + 1):l
        delete!(index(df).lookup,index(df).names[i])
    end
   resize!(_columns(df), length(inds))
   resize!(index(df).names, length(inds)) 
  df
end
```
Performance Improvement:

```julia
cols = 90000
[...]

PR #1985, opened 2019-10-13
Rewrite `select!` to improve performance

Ellipse0934:
As discussed in Issue #1861 . This will significantly improve performance for the `select!` function(over 1000x for 100,000 columns).
Passes the tests locally and I couldn't find any bugs upon testing.

### DIFF
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -74,6 +74,10 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
+if VERSION < v"1.1"
+    import Future.copy!
+end
+
 ##############################################################################
 ##
 ## Load files
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -74,10 +74,6 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
-if VERSION < v"1.1"
-    import Future.copy!
-end
-
 ##############################################################################
 ##
 ## Load files
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -818,11 +818,11 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
     if !allunique(inds)
         throw(ArgumentError("indices must not contain duplicates"))
     end
-    copy!(_names(df), _names(df)[inds])
     copy!(_columns(df), _columns(df)[inds])
-    empty!(index(df).lookup)
-    for (i, n) in enumerate(_names(df))
-        index(df).lookup[n] = i
+    x = index(df)
+    copy!(_names(x), _names(df)[inds])
+    for (i, n) in enumerate(x.names)
+        x.lookup[n] = i
     end
     df
 end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -818,15 +818,11 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
     if !allunique(inds)
         throw(ArgumentError("indices must not contain duplicates"))
     end
-    targetnames = _names(df)[inds]
-    targetcolumns = _columns(df)[inds]
-    empty!(_columns(df))
-    append!(_columns(df), targetcolumns)
-    x = index(df)
-    empty!(x)
-    append!(_names(x), targetnames)
-    for (i, n) in enumerate(x.names)
-        x.lookup[n] = i
+    copy!(_names(df), _names(df)[inds])
+    copy!(_columns(df), _columns(df)[inds])
+    empty!(index(df).lookup)
+    for (i, n) in enumerate(_names(df))
+        index(df).lookup[n] = i
     end
     df
 end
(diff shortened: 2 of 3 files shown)


Ellipse0934:
Need to import Future.copy! for backwards compatibility with ver 1.0. Will do it tomorrow.

bkamins:
Yes - exactly like in #1974 (so whichever gets merged first the other will need to rebase later).

Ellipse0934:
Can you tell me why something like `VERSION < v"1.1" && import Future.copy!` is a bad idea ? 
Future won't be unnecessarily imported. Behavior should be the same. Let me test it out now.

bkamins:
Ah - and I think we need to add `Future` to `[deps]` in Project.toml like in #1974 (again @ararslan - can you please have look and confirm?).

src/DataFrames.jl
```diff
@@ -74,6 +74,10 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
+if VERSION < v"1.1"
+    import Future.copy!
```
nalimilan: I think this should be enough since you don't overload the function:
```suggestion
    using Future: copy!
```

src/dataframe/dataframe.jl
```diff
@@ -818,15 +818,12 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
...
-
-    targetnames = _names(df)[inds]
-    for i in setdiff(ncol(df):-1:1, inds)
-        splice!(_columns(df), i)
-        delete!(index(df), i)
+    copy!(_names(df), _names(df)[inds])
+    copy!(_columns(df), _columns(df)[inds])
+    empty!(index(df).lookup)
+    for (i, n) in enumerate(_names(df))
+        index(df).lookup[n] = i
```
nalimilan: Store `index(df)` in a variable to avoid calling the function over and over.

src/dataframe/dataframe.jl
```diff
@@ -818,15 +818,12 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
     if !allunique(inds)
         throw(ArgumentError("indices must not contain duplicates"))
     end
-
-    targetnames = _names(df)[inds]
-    for i in setdiff(ncol(df):-1:1, inds)
-        splice!(_columns(df), i)
-        delete!(index(df), i)
+    copy!(_names(df), _names(df)[inds])
```
```suggestion
    copy!(_names(df), view(_names(df), inds))
```

src/dataframe/dataframe.jl
```diff
@@ -818,15 +818,12 @@ function select!(df::DataFrame, inds::AbstractVector{Int})
...
+    copy!(_columns(df), _columns(df)[inds])
```
```suggestion
    copy!(_columns(df), view(_columns(df), inds))
```

bkamins: This code will not work correctly because of aliasing. We need to copy:
```
julia> x = [1,2,3,4,5]
5-element Array{Int64,1}:
 1
 2
 3
 4
 5

julia> y = view(x,[2,1])
2-element view(::Array{Int64,1}, [2, 1]) with eltype Int64:
 2
 1

julia> copy!(x, y)
2-element Array{Int64,1}:
 2
 2

julia> x
2-element Array{Int64,1}:
 2
 2
```
fortunately here it is not expensive in comparison to other operations.

src/dataframe/dataframe.jl
bkamins: here same as below - we need to copy.

src/dataframe/dataframe.jl
bkamins: This is what I initially proposed, but I think the compiler should be able to optimize out this so I thought it was OK. But for clarity maybe even `index(df).lookup` could be stored in a variable to make sure all is optimal here.

src/dataframe/dataframe.jl
nalimilan: Ah right. Copying is fine.

Ideally we could simply write `deleteat!(x, Not(inds))` or maybe `filter!(x, inds), but that's not a big deal.

src/dataframe/dataframe.jl
Ellipse0934: I verified that the compiler does make this optimization by looking at `@code_typed`. Is it still required ?

nalimilan: Maybe yes, just to make sure it won't regress.

src/DataFrames.jl
ararslan: This will also require a dependency on Future declared in Project.toml deps.

nalimilan: Actually https://github.com/JuliaData/DataFrames.jl/pull/1974 just did that, so you can just drop this block AFAICT.

bkamins: and you need to rebase this PR then.

Ellipse0934:
I think we are ready to merge. 
Ping: @bkamins

bkamins:
I am OK to merge this. Let us wait for @nalimilan or @ararslan to confirm so that we have a second pair of eyes :), as this is a fundamental function in the DataFrames.jl ecosystem 😄.

bkamins:
@Ellipse0934 and @nalimilan - thank you for contributing to this. The change is really of order of magnitude improvement as @Ellipse0934 noted!

PR #1986, opened 2019-10-14
deprecate names!

### LINKED ISSUE #1943: merge `names! into `rename!`
`names!` has non-conflicting signature with `rename!` and serves the same purpose.
Additionally add an appropriate method to `rename` that renames columns as `names!` but creates a new `DataFrame`.

bkamins:
Fixes #1943.

@nalimilan after this is the last PR (apart from some already open ones) that I think is needed before 0.20 release.


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -75,40 +74,8 @@ _names(df::AbstractDataFrame) = _names(index(df))
...
-df = DataFrame(i = 1:10, x = rand(10), y = rand(["a", "b", "c"], 10))
-names!(df, [:a, :b, :c])
-names!(df, [:a, :b, :a])  # throws ArgumentError
-names!(df, [:a, :b, :a], makeunique=true)  # renames second :a to :a_1
-```
-
-"""
-function names!(df::AbstractDataFrame, vals; makeunique::Bool=false)
-    names!(index(df), vals, makeunique=makeunique)
+function rename!(df::AbstractDataFrame, vals::Vector{Symbol}; makeunique::Bool=false)
```
nalimilan: How about allowing an `AbstractVector{Symbol}` everywhere?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -145,6 +116,11 @@ rename(f::Function, df::AbstractDataFrame)
   the original names to new names
 * `f` : a function which for each column takes the old name (a Symbol)
   and returns the new name (a Symbol)
+* `vals` : column names, normally a Vector{Symbol} the same length as
```
```suggestion
* `vals` : new column names as a vector of `Symbol`s of the same length as
```

test/dataframe.jl
```diff
@@ -38,6 +38,36 @@ const ≇ = !isequal
     end
 end
 
+@testset "additional rename! tests" begin
+    Random.seed!(123)
+    for i in 1:1000
```
nalimilan: Why do we need this brute-force approach? Can't you just choose a set of values that cover all possible cases?

bkamins: It takes 0.1 second to run these 1000 tests. I add this because I was not sure if we cover all possible cases manually (the code for `rename!` in https://github.com/JuliaData/DataFrames.jl/pull/1974 is tricky, so I prefer to be on a safe side).

src/abstractdataframe/abstractdataframe.jl
bkamins: I will change this. I left `Vector{Symbol}` as this is what was originally there and I thought having another collection of `Symbol`s is unlikely so I left it as is. But it is OK to change this.

test/dataframe.jl
nalimilan: Hmm, but are you sure that with 1000 iterations we cover all cases? :-)

Anyway, please add a comment explaining a bit why we do this.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -121,17 +88,21 @@ function rename!(f::Function, df::AbstractDataFrame)
...
+rename(df::AbstractDataFrame, vals::AbstractVector{Symbol}; makeunique::Bool=false) =
+    rename!(copy(df), vals, makeunique=makeunique)
 rename(df::AbstractDataFrame, args...) = rename!(copy(df), args...)
 rename(f::Function, df::AbstractDataFrame) = rename!(f, copy(df))
 
 """
 Rename columns
 
 ```julia
+rename!(df::AbstractDataFrame, vals::Vector{Symbol}; makeunique::Bool=false)
```
```suggestion
rename!(df::AbstractDataFrame, vals::AbstractVector{Symbol}; makeunique::Bool=false)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -121,17 +88,21 @@ function rename!(f::Function, df::AbstractDataFrame)
...
 rename!(df::AbstractDataFrame, (from => to)::Pair{Symbol, Symbol}...)
 rename!(df::AbstractDataFrame, d::AbstractDict{Symbol,Symbol})
 rename!(df::AbstractDataFrame, d::AbstractArray{Pair{Symbol,Symbol}})
 rename!(f::Function, df::AbstractDataFrame)
+rename(df::AbstractDataFrame, vals::Vector{Symbol}; makeunique::Bool=false)
```
```suggestion
rename(df::AbstractDataFrame, vals::AbstractVector{Symbol}; makeunique::Bool=false)
```

test/dataframe.jl
bkamins: Thank you for the docs fixes - a keen eye as always 😄.

I have added a note in the test. In general I am convinced that 1000 should cover all cases (I have chosen column count - 8 and names set so that with high probability all cases are covered). But I am not sure 😢, if I were sure I would not write a randomized test.

In general we want to be sure that cyclical renames like `:X=>:Y, :Y=>:Z, :Z=>:X` for various cycle lengths are processed correctly and also that non cyclical renames like `:X=>:Y, :Z=>:X` are processed correctly also when they happen in combination and in the presence of other columns that are not renamed and in two subcategories: a) the proposed renaming scheme is valid and b) the proposed renaming scheme is invalid (so as you see there are really many possible combinations of cases and it is really hard to make sure that you cover all of them if you do not generate the tests randomly; for sure the existing tests have not covered all the possibilities).

PR #1989, opened 2019-10-20
Make `map(f, df)` and `foreach(f, df)` work

hymd-research:
by adding `Base.length(df)` and `Base.iterate(df)` (both `Base.map` and `Base.foreach` call `Base.iterate`, and `Base.iterate` calls `Base.length`) .

`map(f, df)` and `foreach(f, df)` work just like `filter(f, df)`.


bkamins:
You can use `eachrow(df)` instead of `df` to get what you need.

PR #1994, opened 2019-10-27
Add reference to DataFrames Tutorial in index.md

dmolina:
I think that the website [https://github.com/bkamins/Julia-DataFrames-Tutorial/](https://github.com/bkamins/Julia-DataFrames-Tutorial/) is a very useful resource, and it could be good to mention it in the documentation.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -6,6 +6,8 @@ This resource aims to teach you everything you need
 to know to get up and running with tabular data manipulation using the DataFrames.jl package
 and the Julia language.
 
+This documentation is always updated. There is available also a [DataFrames Tutorial in NoteBooks](https://github.com/bkamins/Julia-DataFrames-Tutorial/). That tutorial is not guarantee that be completely updated, but it is a good complementary resource to learn about DataFrames usage. 
+
 If there is something you expect DataFrames to be capable of, but
 cannot figure out how to do, please reach out with questions in Domains/Data on
 [Discourse](https://discourse.julialang.org/new-topic?title=[DataFrames%20Question]:%20&body=%23%20Question:%0A%0A%23%20Dataset%20(if%20applicable):%0A%0A%23%20Minimal%20Working%20Example%20(if%20applicable):%0A&category=Domains/Data&tags=question).
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -6,7 +6,7 @@ This resource aims to teach you everything you need
 to know to get up and running with tabular data manipulation using the DataFrames.jl package
 and the Julia language.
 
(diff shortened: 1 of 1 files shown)


nalimilan:
Well since the author is the same as this package, we should probably move content to here when applicable. @bkamins?

bkamins:
@nalimilan - do you mean to migrate the whole of https://github.com/bkamins/Julia-DataFrames-Tutorial to DataFrames.jl manual?
It is doable without a problem (and I am OK with this and could do the changes).
The reason I was hesitant to do this (and kept them separate) is that https://github.com/bkamins/Julia-DataFrames-Tutorial is dependence-heavy (and I think it should as its objective is to show the ecosystem around DataFrames.jl). What do you think would be the best approach?

nalimilan:
I don't know, IIRC we discussed moving some parts to the DataFrames manual. I agree that the manual shouldn't require installing too many dependencies, so maybe it makes sense to point to the broader tutorial from the manual.

docs/src/index.md
```diff
@@ -6,6 +6,8 @@ This resource aims to teach you everything you need
 to know to get up and running with tabular data manipulation using the DataFrames.jl package
 and the Julia language.
 
+This documentation is always updated. There is available also a [DataFrames Tutorial in NoteBooks](https://github.com/bkamins/Julia-DataFrames-Tutorial/). That tutorial is not guarantee that be completely updated, but it is a good complementary resource to learn about DataFrames usage.
```
nalimilan: Maybe don't insist on the risk of the tutorial being out of date, given that it's maintained by @bkamins. But mentioning that the tutorial illustrates other packages would make sense.

dmolina: Yes, you are right. I have removed the mention to the risk of being out of date, and add that "it illustrates its usage in conjunction with other packages.".

dmolina: Also, @bkamins  I was referring only to make more accesible the Tutorial, I understand the aim of avoid the additional dependences. Only I think that the Tutorial is a great learning material, and it has a lot of sense mention it in the Documentation.

bkamins:
@dmolina - thank you! I think this is good to merge one @nalimilan checks it :). I will try to update the core documentation after v0.20 is released.

docs/src/index.md
```diff
@@ -6,6 +6,8 @@ This resource aims to teach you everything you need
 to know to get up and running with tabular data manipulation using the DataFrames.jl package
 and the Julia language.
 
+This documentation is always updated. There is available also a [DataFrames Tutorial in NoteBooks](https://github.com/bkamins/Julia-DataFrames-Tutorial/). That tutorial is a good complementary resource to learn about DataFrames usage, and it illustrates its usage in conjunction with other packages.
```
bkamins: some small comments:
```suggestion
This documentation is always updated. There is available also a [DataFrames Tutorial using Jupyter Notebooks](https://github.com/bkamins/Julia-DataFrames-Tutorial/). That tutorial is a good complementary resource to learn about DataFrames.jl, and it illustrates its usage in conjunction with other packages.
```

PR #1995, opened 2019-10-27
fix a bug in select!

bkamins:
`lookup` of `Index` was not cleaned by `select!` before reassignment. This PR fixes this.


bkamins:
Thank you!

Issue #1999, opened 2019-11-05
Documenting the difference betwen df[!, :col] and df[:, :col]


aminya:
In the documentation, I see `!` sign being used which is acting similar to `:`, but it isn't documented anywhere what are the differences.
https://github.com/JuliaData/DataFrames.jl/blob/master/docs/src/man/getting_started.md

nilshg:
Have you looked at https://juliadata.github.io/DataFrames.jl/stable/lib/indexing/ which described all types of indexing strategies in great detail?

bkamins:
Ideally I would like also to have a decision on https://github.com/JuliaData/DataFrames.jl/pull/1961 to make the target API of `:` vs `!` final (please comment in that PR if you have an opinion on this - as the decision is pending as @nalimilan and I want to collect more feedback where it should go). Thank you!

aminya:
> 
> 
> Have you looked at https://juliadata.github.io/DataFrames.jl/stable/lib/indexing/ which described all types of indexing strategies in great detail?

I think it should be also described brefiely in Getting started. I didn't find this page until you sent it.

PR #2000, opened 2019-11-05
Documenting the difference betwen df[!, :col] and df[:, :col]

aminya:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1999

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -47,7 +47,9 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position. 
 
-Since `df[!, :col]` does not make a copy, changing the elements of the column returned by this syntax will affect the values stored in the original `df`. Use the `df[:, :col]` syntax to get a copy of the column, which can be modified without changing `df`.
+Since `df[!, :col]` does not make a copy, changing the elements of the column returned by this syntax will affect the values stored in the original `df`. 
+
+To get a copy of the column use `df[:, :col]`. Changing the column returned by this syntax does not change `df`.
 
 ```jldoctest dataframe
 julia> df.A
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -47,7 +47,7 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position. 
 
-Notice that, changing the elements of the column returned by `df[!, :col]` will affect the values stored in the original `df`. To get a copy of the column use `df[:, :col]`. Changing the column returned by `df[:, :col]` does not change `df`.
+Since `df[!, :col]` does not make a copy, changing the elements of the column returned by this syntax will affect the values stored in the original `df`. Use the `df[:, :col]` syntax to get a copy of the column, which can be modified without changing `df`.
 
 ```jldoctest dataframe
 julia> df.A
@@ -60,13 +60,31 @@ julia> df.A
 julia> df.A === df[!, :A]
 true
 
+julia> df.A === df[:, :A]
+false
+
+julia> df.A == df[:, :A]
+true
+
 julia> df.A === df[!, 1]
+true
+
+julia> df.A === df[:, 1]
+false
+
+julia> df.A == df[:, 1]
 true
 
 julia> firstcolumn = :A
 :A
 
 julia> df[!, firstcolumn] === df.A
+true
+
+julia> df[:, firstcolumn] === df.A
+false
+
+julia> df[:, firstcolumn] == df.A
 true
 ```
 
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -45,7 +45,9 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/man/getting_started.md
```diff
@@ -45,7 +45,9 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
-Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position.
+Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position. 
+
+`Notice that, changing the elements of the array returned by `df[!, :col]` will affect the original `df`, while `df[:, :col]` returns a copy of the array, and changing it doesn't change `df` values.
```
bkamins:
```suggestion
Notice that, changing the elements of the column returned by `df[!, :col]` will affect the values stored in the original `df`, while `df[:, :col]` returns a copy of the column, and changing it does not change `df`.
```

docs/src/man/getting_started.md
```diff
@@ -45,7 +45,9 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
-Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position.
+Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position. 
+
+Notice that, changing the elements of the column returned by `df[!, :col]` will affect the values stored in the original `df`, while `df[:, :col]` returns a copy of the column, and changing it does not change `df`.
```
nalimilan: AFAICT `df[:, :col]` hasn't been mentioned yet at this point. Can you rephrase this so that it introduces `:`, rather than making it sound like users already know it?

aminya: Fixed in https://github.com/JuliaData/DataFrames.jl/pull/2000/commits/c1de00de14e23f0fc19aa208fbbf5c5b130068eb

nalimilan:
Thanks. I've made some changes, let me know what you think.

docs/src/man/getting_started.md
```diff
@@ -47,7 +47,7 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 Columns can be directly (i.e. without copying) accessed via `df.col` or `df[!, :col]`. The latter syntax is more flexible as it allows passing a variable holding the name of the column, and not only a literal name. Note that column names are symbols (`:col` or `Symbol("col")`) rather than strings (`"col"`). Columns can also be accessed using an integer index specifying their position. 
 
-Notice that, changing the elements of the column returned by `df[!, :col]` will affect the values stored in the original `df`. To get a copy of the column use `df[:, :col]`. Changing the column returned by `df[:, :col]` does not change `df`.
+Since `df[!, :col]` does not make a copy, changing the elements of the column returned by this syntax will affect the values stored in the original `df`. Use the `df[:, :col]` syntax to get a copy of the column, which can be modified without changing `df`.
```
aminya: This is too picky, but regarding the last part, I like my sentence structure more :smile: Combining the two sentences and using `which` makes it a little vague (like: `what can be modified?`). 
Otherwise, it looks good.

nalimilan: Can you suggest a wording?

aminya:
```suggestion
Since `df[!, :col]` does not make a copy, changing the elements of the column returned by this syntax will affect the values stored in the original `df`. 

To get a copy of the column use `df[:, :col]`. Changing the column returned by this syntax does not change `df`.
```

scls19fr:
Maybe a good place to warn about the difference could be also in the warning

    Warning: `getindex(df::DataFrame, col_ind::ColumnIndex)` is deprecated, use `df[!, col_ind]` 

which could be changed for

    Warning: `getindex(df::DataFrame, col_ind::ColumnIndex)` is deprecated, use `df[!, col_ind]` (reference) or `df[:, col_ind]` to get a copy

when user are doing `df[:column_name]`

aminya:
> Maybe a good place to warn about the difference could be also in the warning
> 
> ```
> Warning: `getindex(df::DataFrame, col_ind::ColumnIndex)` is deprecated, use `df[!, col_ind]` 
> ```
> 
> which could be changed for
> 
> ```
> Warning: `getindex(df::DataFrame, col_ind::ColumnIndex)` is deprecated, use `df[!, col_ind]` (reference) or `df[:, col_ind]` to get a copy
> ```
> 
> when user are doing `df[:column_name]`

`:` is not deprecated. They are two different methods.

nalimilan:
I've slightly edited the wording.

> `:` is not deprecated.

I don't think that's what @scls19fr suggested. @scls19fr Can you file a new PR? You can edit the file directly on GitHub if you want.

scls19fr:
Absolutely @nalimilan what I'm saying is that `df[:col]` is deprecated and `df[!, :col]` or `df[:, :col]` should be used instead (and `!` or `:` should be used depending if user want a reference or a copy of column data)

I've been searching for this depwarn using
https://github.com/JuliaData/DataFrames.jl/search?q=use+%60df%5B%21%2C+col_ind%5D%60&unscoped_q=use+%60df%5B%21%2C+col_ind%5D%60

but couldn't find where it's located.

@nalimilan could you point me to code where it's located?

I will submit a PR also

nalimilan:
It's here: https://github.com/JuliaData/DataFrames.jl/blob/9b30a99303414839579510c4c3138752333fe446/src/deprecated.jl#L1390:L1396

scls19fr:
Hum... not so easy to add information there I thought it could be done using a simple string... but that's not the case (or at least I don't know how to provide deprecated macro an informative string)

nalimilan:
You have to replace this with a normal function which will call `Base.depwarn`. See examples in the file.

Issue #1982, opened 2019-10-11
vcat! or push!(..., columns=:union)


rapus95:
Is there a reason why I can't find a good way to add rows, where each comes from a dict with only a subset of the columns set, in-place?

bkamins:
I was thinking about allowing it in `push!` (as this is I guess what you want). I can add it https://github.com/JuliaData/DataFrames.jl/pull/1958 as this is not merged yet.

However, if you are willing please discuss what the target functionality should be. The particular issues are:
* what if the column is missing in added row and the column in a target data frame does not allow `missing` (do we throw an error then - I guess yes)
* what if the target does not have a column that is present in the added row (do we error then? - I guess yes).

The current way to do it is to use `vcat` with `:union` option, and you have to materialize each row as 1-row `DataFrame` (I know it is inefficient and that is why I am OK to consider making `push!` more flexible). The difference is that `push!` will not change target metadata, while `vcat` with `:union` will be still more flexible.

Any opinion on this?

rapus95:
For now I implemented the following as some helper function for my usecase (supporting additional values and assuming that missing is a valid element either way).

```julia
function fillingpush!(df::DataFrame, dict::AbstractDict; additionalvalues...)
    temp = similar(df, 1)
    for col in names(temp)
        temp[col] = get(dict, col, get(additionalvalues, col, missing))
    end
    push!(df, temp)
end
```

> However, if you are willing please discuss what the target functionality should be. The particular issues are:
> 
> * what if the column is missing in added row and the column in a target data frame does not allow `missing` (do we throw an error then - I guess yes)

I'd say yes here aswell.

> * what if the target does not have a column that is present in the added row (do we error then? - I guess yes).

When `:union` is selected I'd act differently depending upon whether the requested column supports Missing:
 * Supports Missing: Add column that has missing for every row of the first DataFrame
 * Does not support Missing: Throw an error as we couldn't keep the type.
 
Thus, as a general rule, I'd say `push!` never changes column types (not even widening them to support missing). Everyone can do that explicitly anyway.
Thus:
```julia
canhold_or_convert(X,Y) #tests, if we can put x::X into a cell of type Y
A,B #be DataFrames
easy = intersect(names(A), names(B))
@assert canhold_or_convert.(types(B[easy]),types(A[easy]))
onlyA = setdiff(names(A),names(B))
@assert canhold_or_convert.(Missing, types(A[onlyA]))
[...]

PR #2003, opened 2019-11-06
Support `append!`ing row-oriented tables

tkf:
This PR makes code such as `append!(DataFrame(), [(a=1,)])` work.

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -32,7 +32,13 @@ function DataFrame(x::T; copycols::Bool=true) where {T}
     return fromcolumns(Tables.columns(x), copycols=copycols)
 end
 
-Base.append!(df::DataFrame, x) = append!(df, DataFrame(x, copycols=false))
+function Base.append!(df::DataFrame, x)
+    if Tables.columnaccess(x)
+        return append!(df, DataFrame(x, copycols=false))
+    else
+        return foldl(push!, Tables.rows(x); init=df)
+    end
+end
 
 # This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
 function DataFrame(x::Vector{<:NamedTuple}; copycols::Bool=true)
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -32,13 +32,7 @@ function DataFrame(x::T; copycols::Bool=true) where {T}
     return fromcolumns(Tables.columns(x), copycols=copycols)
 end
 
-function Base.append!(df::DataFrame, x)
-    if Tables.columnaccess(x)
-        return append!(df, DataFrame(x, copycols=false))
-    else
-        return foldl(push!, Tables.rows(x); init=df)
-    end
-end
+Base.append!(df::DataFrame, x) = append!(df, DataFrame(Tables.columns(x), copycols=false))
 
 # This supports the Tables.RowTable type; needed to avoid ambiguities w/ another constructor
 function DataFrame(x::Vector{<:NamedTuple}; copycols::Bool=true)
--- a/test/tables.jl
+++ b/test/tables.jl
@@ -129,6 +129,9 @@ Base.propertynames(d::DuplicateNamesColumnTable) = (:a, :a, :b)
         append!(df, etu)
         @test size(df) == (6, 3)
 
+        append!(df, [nt])
+        @test size(df) == (7, 3)
+
         # categorical values
         cat = CategoricalVector(["hey", "there", "sailor"])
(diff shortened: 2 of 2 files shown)


bkamins:
The tests currently fail.

But even if they would pass why do we need the second branch?

tkf:
> why do we need the second branch?

It looks like `append!` does not support row-oriented tables (checked with 1788d8a786868d41f19c8583ec7ae297f1172c07). Presumably, this is to make sure that it is copy-free?

```julia
julia> using DataFrames
       append!(DataFrame(), [(a=1,)])
ERROR: ArgumentError: It is not possible to construct a `DataFrame`from a `Vector{<:NamedTuple}` with `copycols=false`
Stacktrace:
 [1] DataFrame(::Array{NamedTuple{(:a,),Tuple{Int64}},1}; copycols::Bool) at /home/takafumi/.julia/dev/DataFrames/src/other/tables.jl:40
 [2] append!(::DataFrame, ::Array{NamedTuple{(:a,),Tuple{Int64}},1}) at /home/takafumi/.julia/dev/DataFrames/src/other/tables.jl:35
 [3] top-level scope at REPL[1]:2
```

I'll look into the test failure.

tkf:
So the problem was that `push!(df, row)` does not support arbitrary row object specified by Tables.jl.  I see that there is a depwarn so maybe this is going to be fixed in the future?  For now, it seems to be using `Tables.columns` outside `DataFrame` constructor is the simplest solution.

For copy-free `append!` of row-oriented tables, I think you would need widening-based code like. [I implemented this already](https://github.com/tkf/BangBang.jl/blob/9cac31a559402fb8eee3ebd3c8d7879147d1d30f/src/dataframes_impl.jl#L49-L72) so I can turn this into a PR.  But it would be a somewhat non-trivial patch so I'm not sure what would be the best approach here.  (Unless there is some internal function that I can just call.)

bkamins:
I think that your current proposal is a good temporary solution (maybe @quinnj might want to have a look at it).

I think there are the following things to be discussed additionally:
* do we want to allow `copycols` kwarg for the `DataFrame` and `DataFrame!` constructors in case they have to make a copy (i.e. it is impossible not to perform a copy) - currently we disallow it
* extension of `push!` to support arbitrary row object specified by Tables.jl. This I think can be discussed further in https://github.com/JuliaData/DataFrames.jl/issues/1982, as it adds to the wider design issue of `push!` that we currently discuss.

tkf:
> * do we want to allow `copycols` kwarg for the `DataFrame` and `DataFrame!` constructors in case they have to make a copy (i.e. it is impossible not to perform a copy) - currently we disallow it

I think somewhat related question is: "does `copycols=false` mean to avoid _any_ kind of data copy?" (or maybe that's what you mean)  See `Tables.columns` at the end of this constructor:

https://github.com/JuliaData/DataFrames.jl/blob/1788d8a786868d41f19c8583ec7ae297f1172c07/src/other/tables.jl#L22-L33

So, `copycols=false` does not stop `Tables.columns` materializing ("copying") the columns.  However, this case is manually disallowed for:

https://github.com/JuliaData/DataFrames.jl/blob/1788d8a786868d41f19c8583ec7ae297f1172c07/src/other/tables.jl#L38-L42

This is the reason why `append!(DataFrame(a=[1]), (x for x in [(a=1,)]))` works but `append!(DataFrame(a=[1]), [(a=1,)])` does not.

bkamins:
Agreed - I think we should make this consistent. My current feeling is that we should allow `copycols` always and say that if it is `false` we **try** not to copy, but we might copy it is not possible not to copy.

@nalimilan - any opinion on this (the other way would be to be always strict and guarantee that `copycols=false` fails if non-copying is not possible).

nalimilan:
Makes sense.

bkamins:
I have opened https://github.com/JuliaData/DataFrames.jl/pull/2004 for the `copycols` part of the discussion.

quinnj:
To fix `append!(DataFrame(), [(a=1,)])`, I think the right correction is to define:
```julia
Base.append!(df::DataFrame, x::Vector{<:NamedTuple}) = append!(df, DataFrame(x; copycols=true))
```
This corresponds to the special-case `DataFrame` constructor we have for `Vector{<:NamedTuple}`.

tkf:
@quinnj I guess #2004 is equivalent to that (with less code)?

quinnj:
Yes, that looks right.

nalimilan:
Closing since https://github.com/JuliaData/DataFrames.jl/pull/2004 fixed this in a different way. Thanks @tkf for catching this.

PR #2004, opened 2019-11-07
change how copycols in Vector{<:NamedTuple} constructor

bkamins:
Additionally I have cleaned a minor bug in the documentation of `DataFrame` constructor.

See https://github.com/JuliaData/DataFrames.jl/pull/2003#issuecomment-550217542 for a rationale of this change.


tkf:
Thanks. I guess this PR would fix this test case I added #2003?

https://github.com/JuliaData/DataFrames.jl/blob/16904d9df87fbde98c9fbd6409c677d95b7954d6/test/tables.jl#L132-L133

So, I guess you don't need #2003 if you are going to use this `copycols` semantics?

bkamins:
Yes, I think it makes #2003 not needed.

PR #2006, opened 2019-11-08
allow mixing Symbol and Pair{Symbol,Symbol} in on for joining

### LINKED ISSUE #2001: Allow mix of `Symbol` and `Pair` in `join`
Currently we can do:
```
df1 = DataFrame(a = [1,2], b = ["a","b"], c = rand(2))
df2 = DataFrame(a = [1,2], B = ["a","b"], d = rand(2))

join(df1, df2, on = [:a => :a, :b => :B]) 
join(df1, df2, on = [:a]) 
join(df1, df2, on = [:b => :B], makeunique = true)
```
but it's not possible to do:
```
join(df1, df2, on = [:a, :b => :B])
```
This would be very convenient for cases where one has a large number of columns to match on, with only one or two that don't match and require a `Pair`.

bkamins:
Fixes #2001

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -240,7 +240,7 @@ Join two or more `DataFrame` objects
 ### Keyword Arguments
 
 * `on` : A column name to join `df1` and `df2` on. If the columns on which `df1` and `df2`
-  will be joined have different names, then a`left=>right` pair can be passed.
+  will be joined have different names, then a `left=>right` pair can be passed.
   It is also allowed to perform a join on multiple columns, in which case
   a vector of column names or column name pairs can be passed
   (mixing names and pairs is allowed).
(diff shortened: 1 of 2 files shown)


bkamins:
Yes - the parsing of `on` is very cheap compared to joining later so I preferred a clean design over squeezing out performance from the preprocessing part.

nalimilan:
Looks good, but can you update the docstring(s)?

bkamins:
When I read it before I thought it was good. But I agree that it needs an update. I have pushed a fix of docstring.

src/abstractdataframe/join.jl
```diff
@@ -227,12 +239,14 @@ Join two or more `DataFrame` objects
...
 ### Keyword Arguments
 
-* `on` : A column, or vector of columns to join df1 and df2 on. If the column(s)
-    that df1 and df2 will be joined on have different names, then the columns
-    should be `left => right` pairs, or a vector of such pairs.
-    `on` is a required argument for all joins except for `kind = :cross`.
-    If more than two data frames are joined then only a column name
-    or a vector of column names are allowed.
+* `on` : A column name to join `df1` and `df2` on. If the columns on which `df1` and `df2`
+  will be joined have different names, then a`left=>right` pair can be passed.
```
bkamins:
```suggestion
  will be joined have different names, then a `left=>right` pair can be passed.
```

Issue #2010, opened 2019-11-11
describe should also apply to Vector


scls19fr:
Hello,

`describe` function is very convenient to have a quick look at dataframe value.
But it will be great if it could apply to Vector also

I was expecting to be able to do:

    describe(df[!, :column])

Kind regards

bkamins:
Use `describe(select(df, :column))`.

`describe` for an `AbstractVector` is defined in StatsBase.jl:

```
julia> using StatsBase

julia> describe([1,2,3,4])
Summary Stats:
Length:         4
Missing Count:  0
Mean:           2.500000
Minimum:        1.000000
1st Quartile:   1.750000
Median:         2.500000
3rd Quartile:   3.250000
Maximum:        4.000000
Type:           Int64
```

scls19fr:
Thanks @bkamins maybe [doc](http://juliadata.github.io/DataFrames.jl/stable/lib/functions/#DataAPI.describe) should be improved to show such a usage (`describe(select(df, :column))`)

bkamins:
OK - I will update the docs anyway today so I can add it.

bkamins:
See https://github.com/JuliaData/DataFrames.jl/pull/2011 (please leave a comment if something is not clear)

scls19fr:
Thanks @bkamins

PR #2011, opened 2019-11-11
update documentation of select

### LINKED ISSUE #1997: select!(df, Not(tuple)) does not work
The issue is pretty simple:

```julia
unwanted = (:col1, :col2, :col3)
select!(df, Not(unwanted))
```

will fail (getindex error)

```julia
unwanted = [:col1, :col2, :col3]
select!(df, Not(unwanted))
```

will work fine. 

Maybe the behaviour from (2) could be generalised to iterables?

bkamins:
Fixes #1997 and #2010

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -476,7 +476,7 @@ julia> df[in.(df.A, Ref([1, 5, 601])), :]
 
 Equivalently, the `in` function can be called with a single argument to create a function object that tests whether each value belongs to the subset (partial application of `in`): `df[in([1, 5, 601]).(df.A), :]`.
 
-#### Column selection syntax
+#### Column selection using `select` and `select!`
 
 You can also use the [`select`](@ref) and [`select!`](@ref) functions to select columns in a data frame.
 The `select` function creates a new data frame:
@@ -502,6 +502,7 @@ julia> select(df, r"x") # select columns containing 'x' character
 ├─────┼───────┼───────┤
 │ 1   │ 1     │ 2     │
 ```
+
 To perform the operations in-place use `select!`:
 ```jldoctest dataframe
 julia> select!(df, Not(:x1))
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -479,6 +479,7 @@ Equivalently, the `in` function can be called with a single argument to create a
 #### Column selection using `select` and `select!`
 
 You can also use the [`select`](@ref) and [`select!`](@ref) functions to select columns in a data frame.
+
 The `select` function creates a new data frame:
 ```jldoctest dataframe
 julia> df = DataFrame(x1=1, x2=2, y=3)
@@ -503,7 +504,46 @@ julia> select(df, r"x") # select columns containing 'x' character
 │ 1   │ 1     │ 2     │
 ```
 
-To perform the operations in-place use `select!`:
+It is important to note that `select` always returns a data frame,
+even if a single column is selected (as opposed to indexing syntax).
+```jldoctest dataframe
+julia> select(df, :x1)
+1×1 DataFrame
+│ Row │ x1    │
+│     │ Int64 │
+├─────┼───────┤
+│ 1   │ 1     │
+
+julia> df[:, :x1]
+1-element Array{Int64,1}:
+ 1
+```
+
+By default `select` copies columns of a passed source data frame.
+In order to avoid copying set its `copycols` keyword argument to `false`:
+```
+julia> df2 = select(df, :x1)
+1×1 DataFrame
+│ Row │ x1    │
+│     │ Int64 │
+├─────┼───────┤
+│ 1   │ 1     │
+
+julia> df2.x1 === df.x1
+false
+
+julia> df2 = select(df, :x1, copycols=false)
+1×1 DataFrame
+│ Row │ x1    │
+│     │ Int64 │
+├─────┼───────┤
+│ 1   │ 1     │
+
+julia> df2.x1 === df.x1
+true
+```
+
+To perform the selection operation in-place use `select!`:
 ```jldoctest dataframe
 julia> select!(df, Not(:x1))
 1×2 DataFrame
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -371,7 +371,7 @@ julia> df[!, :A] == df[:, :A]
 true
 ```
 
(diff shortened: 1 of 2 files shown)


src/dataframe/dataframe.jl
```diff
@@ -772,6 +772,8 @@ Mutate `df` in place to retain only columns specified by `inds...` and return it
 
 Arguments passed as `inds...` can be any index that is allowed for column indexing
 provided that the columns requested in each of them are unique and present in `df`.
+In particular, you are allowed to use regular expressions,
```
nalimilan: I'd avoid using "you" in docstrings. Better say "are supported" or "`inds` can be". Also use `@ref`?

docs/src/man/getting_started.md
```diff
@@ -371,6 +371,9 @@ true
 
 In the first cases, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`, since a vector can contain one or more column names. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted.
 
+Note, that similarly to Base Julia you are not allowed to select rows nor columns using `Tuple`s, use `AbstractVector`s instead.
```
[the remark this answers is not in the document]
bkamins: I have shortened it (I decided to add it as people ask about it - so it seems it creates confusion for newcomers).

For me the general answer to all the questions like this would be "it works like Base Julia" period (this was the design goal we had ~1 year ago and now we have it) 😄.

docs/src/man/getting_started.md
```diff
@@ -425,6 +428,24 @@ julia> df[:, All(Not(r"x"), :)]
 ```
 
 You can also use the [`select`](@ref) and [`select!`](@ref) functions to select columns in a data frame.
```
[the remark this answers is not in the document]
bkamins: OK - I have moved it down to a separate subheading (so now we have two subheadings - one for indexing and one for selecting)

docs/src/man/getting_started.md
```diff
@@ -499,7 +520,27 @@ julia> describe(df)
 
 ```
 
-Of course, one can also compute descrptive statistics directly on individual columns:
+If you are interested in describing only a subset of columns then the easiest way to do it is to
```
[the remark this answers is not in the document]
bkamins: Right - fixed.

Regarding an additional arg do `describe` I have opened https://github.com/JuliaData/DataFrames.jl/issues/2014 to keep it in mind (in general I agree we could add it).

src/dataframe/dataframe.jl
bkamins: changed to "are supported".

AFAIK `@ref` will not work as `Not` etc. are not defined in DataFrames.jl and cross-package `@ref` does not work (at least as of present).

docs/src/man/getting_started.md
```diff
@@ -476,6 +476,49 @@ julia> df[in.(df.A, Ref([1, 5, 601])), :]
...
+│ 1   │ 2     │ 3     │
+
+julia> select(df, r"x") # select columns containing 'x' character
+1×2 DataFrame
+│ Row │ x1    │ x2    │
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 2     │
+```
+To perform the operations in-place use `select!`:
```
nalimilan:
```suggestion

To perform the operations in-place use `select!`:
```

docs/src/man/getting_started.md
```diff
@@ -476,6 +476,49 @@ julia> df[in.(df.A, Ref([1, 5, 601])), :]
 
 Equivalently, the `in` function can be called with a single argument to create a function object that tests whether each value belongs to the subset (partial application of `in`): `df[in([1, 5, 601]).(df.A), :]`.
 
+#### Column selection syntax
```
nalimilan:
```suggestion
#### Column selection using `select` and `select!`
```

docs/src/man/getting_started.md
```diff
@@ -476,6 +476,49 @@ julia> df[in.(df.A, Ref([1, 5, 601])), :]
...
+
+You can also use the [`select`](@ref) and [`select!`](@ref) functions to select columns in a data frame.
```
nalimilan: How about mentioning the main differences between `select` and indexing, namely: 1) that `select(df, :x1)` returns a data frame rather than a vector? Also, `select` makes copies like `df[:, cols]`, which isn't obvious.

bkamins: added

docs/src/man/getting_started.md
```diff
@@ -476,6 +476,90 @@ julia> df[in.(df.A, Ref([1, 5, 601])), :]
...
+├─────┼───────┤
+│ 1   │ 1     │
+
+julia> df[:, :x1]
+1-element Array{Int64,1}:
+ 1
+```
+
+By default `select` copies columns of a passed source data frame.
+In order to avoid copying set its `copycols` keyword argument to `false`:
```
nalimilan:
```suggestion
In order to avoid copying, pass `copycols=false`:
```

docs/src/man/getting_started.md
```diff
@@ -476,6 +476,90 @@ julia> df[in.(df.A, Ref([1, 5, 601])), :]
...
+├─────┼───────┤
+│ 1   │ 1     │
+
+julia> df2.x1 === df.x1
+true
+```
+
+To perform the selection operation in-place use `select!`:
+```jldoctest dataframe
+julia> select!(df, Not(:x1))
```
nalimilan: I'd show `df` right after this, instead of the result of `select!`. Then you can drop the second example.

docs/src/man/getting_started.md
```diff
@@ -369,7 +371,7 @@ julia> df[!, :A] == df[:, :A]
 true
 ```
 
-In the first cases, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`, since a vector can contain one or more column names. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted.
+In the first cases, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`, since a vector can contain one or more column names. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted. Note, that in the first case a vector is required to be passed (not just any iterable), so e.g. `df[:, (:x1, :x2)]` is not allowed, but `df[:, [:x1, :x2]]` is valid.
```
nalimilan:
```suggestion
In the first cases, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`, since a vector can contain one or more column names. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted. Note that in the first case a vector is required to be passed (not just any iterable), so e.g. `df[:, (:x1, :x2)]` is not allowed, but `df[:, [:x1, :x2]]` is valid.
```

docs/src/man/getting_started.md
bkamins: fixed - is this what you had in mind?

Issue #2012, opened 2019-11-13
Add `flatten` function


pdeffebach:
[This post](https://discourse.julialang.org/t/juliadb-flatten-equivalent-in-dataframes/31048/2) on Discourse requested a `flatten` function. 

I wrote what I think is a pretty efficient implementation of a `flatten` function while answering the question. It could probably be added to DataFrames.

Or even better, it could be added to `TablesApi` with a few tweaks.

bkamins:
I think it is OK to add it to DataFrames.jl. Would you care to make a PR?

A small note though. The implementation should not use indexing into data frames in the triple nested loop (it will be a performance killer); either a barier function or delayed data frame creation (both discussed in https://github.com/bkamins/Julia-DataFrames-Tutorial/blob/master/11_performance.ipynb) should be used here for performance.

nalimilan:
FWIW JuliaDB already has that function: https://juliacomputing.github.io/JuliaDB.jl/latest/api/#IndexedTables.flatten
It would make sense to add it to DataAPI.

bkamins:
closing ad #2013 added the functionality

PR #2013, opened 2019-11-13
Add flatten function

pdeffebach:
Addresses #2012 . I think I made the performance improvements requested by @bkamins.


src/other/utils.jl
```diff
@@ -58,3 +58,40 @@ function gennames(n::Integer)
     return res
 end
 
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+
+When column `veccol` of `df` has non-zero length (i.e. a vector), returns a data frame where each element
```
bkamins: returns "data frame" or `DataFrame`?

src/other/utils.jl
```diff
@@ -58,3 +58,40 @@ function gennames(n::Integer)
     return res
 end
 
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+
+When column `veccol` of `df` has non-zero length (i.e. a vector), returns a data frame where each element 
+of `veccol` is flattened, row `i` of `df` will be duplicated according to the length of `df[i, veccol]`
```
bkamins: except for `veccol` which will be flattened - right?

src/other/utils.jl
```diff
@@ -58,3 +58,40 @@ function gennames(n::Integer)
...
+"""
+function Base.Iterators.flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+    N = nrow(df)
+    lengths = Vector{Int64}(undef, nrow(df))
```
bkamins: why `Int64` and not `Int`?

src/other/utils.jl
```diff
@@ -58,3 +58,40 @@ function gennames(n::Integer)
...
+
+    for i in 1:N
+        lengths[i] = length(df[i, veccol])
```
bkamins: maybe simply `length.(df[!, veccol])`?

src/other/utils.jl
```diff
@@ -58,3 +58,40 @@ function gennames(n::Integer)
...
+                counter += 1
+            end
+        end
+    end
+
+    for name in names(new_df)
+        copy_length!(new_df[!, name], df[!, name], lengths)  
+    end
+
+    insertcols!(new_df, columnindex(df, veccol), veccol => reduce(vcat, df[!, veccol]))
```
bkamins: is `reduce` or ` collect(Iterators.flatten(df[!, veccol]))` faster?

src/other/utils.jl
bkamins: also what is the contract on `eltype` of `veccol` - simply that it has `length`, or it should be iterable, or it should be an `AbstractVector`?

src/DataFrames.jl
```diff
@@ -43,6 +43,7 @@ export AbstractDataFrame,
        disallowmissing!,
        dropmissing,
        dropmissing!,
+       flatten,
```
bkamins: also add it to the manual please

src/other/utils.jl
```diff
@@ -58,3 +58,40 @@ function gennames(n::Integer)
...
+    end
+
+    new_N = sum(lengths)
+    
+    new_df = similar(df[!, Not(veccol)], new_N)
+
+    function copy_length!(longnew, shortold, lengths)
+        counter = 1
+        @inbounds for i in 1:length(shortold)
+            @inbounds for j in 1:lengths[i]
```
bkamins: one `@inbounds` should be enough. Also (but here I am not sure, so please check) - is it possible to use @simd to speed up this filling (but maybe not as `fill!` does not use SIMD).

bkamins:
Thank you for a quick PR.

pdeffebach:
I have made the requested changes. I explored `@simd` and it makes no different. 

`collect(Iterators.flatten(df[!, veccol]))` is actually *way* slower than `reduce(vcat, df[!, veccol])`. However not all things that support iteration support `vcat`. For instance `generators` don't play well with `vcat` while the Iterators method does. 

My initial guess would be to make a small function barrier that allows the fast version for `Vector{<:AbstractArray}` and Iterators as a fallback.

src/other/utils.jl
```diff
@@ -58,3 +58,42 @@ function gennames(n::Integer)
...
 
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+
+When column `veccol` of `df` has elements non-zero length, for example a `Vector` 
+of `Vector`s. Returns a DataFrame where each element of `veccol` is flattened. 
+Elements of row `i` of `df` other than `veccol` will be duplicated according to 
+the length of `df[i, veccol]`
+
+**Arguments**
```
bkamins:
```suggestion
# Arguments
```

src/other/utils.jl
```diff
@@ -58,3 +58,42 @@ function gennames(n::Integer)
...
+
+* `df`: An `AbstractDataFrame`
```
bkamins:
```suggestion
- `df`: an `AbstractDataFrame` to flatten
```

src/other/utils.jl
```diff
@@ -58,3 +58,42 @@ function gennames(n::Integer)
...
+* `veccol`: A `Symbol` or `Integer` where `df[:, veccol]` is a column whose
```
bkamins:
```suggestion
* `veccol`: a `Symbol` or `Integer` where `veccol` is a column of `df` whose 
```

src/other/utils.jl
```diff
@@ -58,3 +58,42 @@ function gennames(n::Integer)
     return res
 end
 
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+
+When column `veccol` of `df` has elements non-zero length, for example a `Vector` 
+of `Vector`s. Returns a DataFrame where each element of `veccol` is flattened.
```
bkamins:
```suggestion
of `Vector`s. Returns a `DataFrame` where each element of `veccol` is flattened. 
```

src/other/utils.jl
```diff
@@ -58,3 +58,42 @@ function gennames(n::Integer)
     return res
 end
 
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+
+When column `veccol` of `df` has elements non-zero length, for example a `Vector`
```
bkamins:
```suggestion
When column `veccol` of `df` has iterable elements that define `length`, for example a `Vector` 
```
I understand we allow zero-length iterables and we only require that they support `length` (as in general some iterables do not support `length`)

bkamins:
I have left some comments (sorry for your confusion, but we are just in a process of making DataFrames.jl consistent with Base recommendations for docstrings).

Regarding the key issue of `reduce` vs `flatten` given the contract you provided (i.e. any iterable that has `length` is accepted) I think we should write:
```
flatcol = df[!, veccol] # this can be extracted once on the top of the function 
newflatcol = if flatcol isa AbstractVector{<:AbstractVector}
    reduce(vcat, flatcol)
else
    collect(Iterators.flatten(flatcol))
end
```
(code is tentative to show what I mean)

src/other/utils.jl
```diff
@@ -58,3 +58,42 @@ function gennames(n::Integer)
...
+the length of `df[i, veccol]`
+
+**Arguments**
+
+* `df`: An `AbstractDataFrame`
+* `veccol`: A `Symbol` or `Integer` where `df[:, veccol]` is a column whose 
+elements support iteration.
+"""
+
+function Base.Iterators.flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
```
non-Jedi: I'm not sure that it's correct for this to be a method of `Iterators.flatten`. To me, this is a fundamentally different operation (un-nesting an iterable versus un-nesting a single iterable field). Since `flatten` is in `Base.Iterators`, you can even use the `flatten` function-name without collision.

pdeffebach: I think you are correct. JuliaDB does not extend Base.Iterators.flatten. I have deleted the extension.

bkamins: But then we should introduce it to DataAPI.jl and mutually extend in DataFrames.jl and JuliaDB.jl (CC @nalimilan).

pdeffebach:
I have responded to these changes, including new code for fast `flatten`ing.

I no longer extend `Base.Iterators` because JuliaDB does not. If it is best practice to extend, please let me know.

bkamins:
Ah - I have just noted. This function should be in abstractdataframe.jl, in utils.jl we keep utility functions for the DataFrames.jl package itself.

pdeffebach:
I have moved the contents of the PR to abstractdataframe.jl. Can't get the diff to go away for utils.jl unfortunately. 

I can add the Issue and PR to Tables.jl if that is what we decide to do.

test/utils.jl
```diff
@@ -29,5 +29,16 @@ end
     @test repeat(df, inner = 2, outer = 3) == ref
     @test repeat(view(df, 1:2, :), inner = 2, outer = 3) == ref
 end
+#
+# flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+#
+@testset "flatten" begin
```
bkamins: can you please add tests having different containers in the flattened column (in particular things like: tuples, scalars, ranges, and mixtures of them will probably be typical use cases).

bkamins:
I guess it should be DataAPI.jl not Tables.jl as noted earlier, but here probably @nalimilan is a good person to judge.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,45 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
     DataFrame(newcols, _names(df), copycols=false)
 end
+
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+
+When column `veccol` of `df` has iterable elements that define `length`, for example a `Vector` 
+of `Vector`s. Returns a DataFrame where each element of `veccol` is flattened. 
+Elements of row `i` of `df` other than `veccol` will be duplicated according to 
+the length of `df[i, veccol]`
```
```suggestion
the length of `df[i, veccol]`.
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,45 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
     end
     DataFrame(newcols, _names(df), copycols=false)
 end
+
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+
+When column `veccol` of `df` has iterable elements that define `length`, for example a `Vector` 
+of `Vector`s. Returns a DataFrame where each element of `veccol` is flattened.
```
```suggestion
of `Vector`s. Returns a `DataFrame` where each element of `veccol` is flattened. 
```

Can you also explain what "flattened" means?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,45 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
     end
     DataFrame(newcols, _names(df), copycols=false)
 end
+
+"""
+    flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
```
nalimilan: The general convention is to call the second argument `cols`, and allow it to point either to a single or to multiple columns. Better use the same naming convention, even if you don't want to support multiple columns for now (would that be hard?).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,45 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+
+When column `veccol` of `df` has iterable elements that define `length`, for example a `Vector` 
+of `Vector`s. Returns a DataFrame where each element of `veccol` is flattened. 
+Elements of row `i` of `df` other than `veccol` will be duplicated according to 
+the length of `df[i, veccol]`
+
+# Arguments
```
nalimilan: No need for this section as it repeats what has been said above. But an example would be useful.

src/other/utils.jl
```diff
@@ -56,5 +56,4 @@ function gennames(n::Integer)
         res[i] = Symbol(@sprintf "x%d" i)
     end
     return res
-end
```
nalimilan: Unrelated.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,45 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+- `df`: an `AbstractDataFrame` to flatten
+- `veccol`: a `Symbol` or `Integer` where `veccol` is a column of `df` whose 
+   elements support iteration.
+"""
+function flatten(df::AbstractDataFrame, veccol::Union{Integer, Symbol})
+    col_to_flatten = df[!, veccol]
+    lengths = length.(col_to_flatten)
+    new_df = similar(df[!, Not(veccol)], sum(lengths))
+
+    function copy_length!(longnew, shortold, lengths)
```
nalimilan: Move the definition of `flatten` and add type annotations for clarity. Also maybe call this `repeat_lengths!` or something like that?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,45 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+        counter = 1
+        @inbounds for i in 1:length(shortold)
+            for j in 1:lengths[i]
+                longnew[counter] = shortold[i]
```
nalimilan: Just in case the compiler isn't smart enough, do `x = shortold[i]` outside of `for j` and then:
```suggestion
                longnew[counter] = x
```

Or even do this:
```julia
l = lengths[i]
longnew[counter:counter+l] .= shortold[i]
counter += l
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,45 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+                counter += 1
+            end
+        end
+    end
+
+    for name in names(new_df)
```
nalimilan: Instead of names, use `axes(new_df, 2)` to avoid a dict lookup. Or maybe `mapcols!` is even shorter.

test/utils.jl
```diff
@@ -29,5 +29,16 @@ end
     @test repeat(df, inner = 2, outer = 3) == ref
     @test repeat(view(df, 1:2, :), inner = 2, outer = 3) == ref
 end
+#
```
nalimilan: Remove this comment.

src/abstractdataframe/abstractdataframe.jl
bkamins: I considered making support for multiple `cols` but in this case it did not make sense for me (would we do a cross-product of these cols)?

bkamins: Still - I agree that having the names consistent is better.

src/abstractdataframe/abstractdataframe.jl
bkamins: You have to use dict lookup as `new_df` does not have all columns that the source had (here we drop the flattened column which is added only at the end).

pdeffebach:
Thank you for your feedback. I have applied the suggestions. 

Regarding multiple columns. We could implement this. It would require each element of each column having the same length. Then we would have to check that they have the same length. That's all. It seems easy to implement.

pdeffebach:
If I were to allow multiple columns to be `flatten`ed, whats the argument that it takes? I forget where we stand on this. A `Vector` of `Symbols` and `Ints`? A tuple? Just a splatted `args...`?

bkamins:
> It would require each element of each column having the same length. Then we would have to check that they have the same length.

Why do you think this is more natural/better than cross-join as I initially thought?

> whats the argument that it takes?

Any argument that `AbstractIndex` can digest for `getindex`. If you want to be even more general and allow splatting have a look how `select` is implemented.

pdeffebach:
Could you clarify what you mean by cross-product? Are you imagining going from `df` to `df2` below?

```
julia> df = DataFrame(a = [1, 2], b = ["p", "q"], c = ["r", "s"]) 
2×3 DataFrame
│ Row │ a     │ b      │ c      │
│     │ Int64 │ String │ String │
├─────┼───────┼────────┼────────┤
│ 1   │ 1     │ p      │ r      │
│ 2   │ 2     │ q      │ s      │

julia> df2 = DataFrame(a = [1, 1, 2, 2], b_c = ["p", "r", "q", "s"])
4×2 DataFrame
│ Row │ a     │ b_c    │
│     │ Int64 │ String │
├─────┼───────┼────────┤
│ 1   │ 1     │ p      │
│ 2   │ 1     │ r      │
│ 3   │ 2     │ q      │
│ 4   │ 2     │ s      │
```

That seems close to reshape. 

I will check out `select` and add some code.

bkamins:
I am not sure the result of what operation your example is intended to show.
By cross-product I imagine the result of cross-join of the respective entries of columns so:
```
julia> DataFrame(a = ["a", "b"], b = [[1, 2], [3,4]], c=[[11,12],[13,14]])
2×3 DataFrame
│ Row │ a      │ b      │ c        │
│     │ String │ Array… │ Array…   │
├─────┼────────┼────────┼──────────┤
│ 1   │ a      │ [1, 2] │ [11, 12] │
│ 2   │ b      │ [3, 4] │ [13, 14] │
```
after flattening by columns `:b` and `:c` would become:
```
julia> DataFrame(a = ["a", "a", "a", "a", "b", "b", "b" ,"b"],
                      b = [1,1,2,2,3,3,4,4], c=[11,12,11,12,13,14,13,14])
8×3 DataFrame
│ Row │ a      │ b     │ c     │
│     │ String │ Int64 │ Int64 │
├─────┼────────┼───────┼───────┤
│ 1   │ a      │ 1     │ 11    │
│ 2   │ a      │ 1     │ 12    │
│ 3   │ a      │ 2     │ 11    │
│ 4   │ a      │ 2     │ 12    │
│ 5   │ b      │ 3     │ 13    │
│ 6   │ b      │ 3     │ 14    │
│ 7   │ b      │ 4     │ 13    │
│ 8   │ b      │ 4     │ 14    │
```
This seems more natural to me as essentially what we say then is that we treat columns that are not flattened as scalars always (even if they contained iterables).
Or conversly - then combining flattening over a column containing vectors and a column containing scalars is the same as flattening by column containing vectors only.

Is this clear what I mean.

Of course I am open to your opinion - I simply find this "cross"- approach potentially more useful than requiring equal lengths of iterables in all columns (but I might have a wrong judgement here).

pdeffebach:
I see now. That does seem like a more useful function, especially since nothing is stopping the user from adding a new column for the second vector column in my earlier proposal. 

Though I guess we should step back and ask, who wants this? The original `flatten` discourse issue seemed intuitive enough. It's not clear there is a demand for a featureful `flatten` function that takes in multiple input vectors.

bkamins:
This was my first response above - that I think allowing a single column is enough (when I reviewed the PR the first time I have considered it and decided that it would create more confusion than help no matter what definition we applied).

pdeffebach:
Sounds good. 

I don't understand why tests aren't passing. They don't pass locally when I run `] test DataFrames` and they don't run on Travis. But when I do `include("test/reshape.jl")` it runs just fine.

bkamins:
This `longnew[counter:counter + l]` seems to be incorrect (have not tested it though)

pdeffebach:
But why would it work when I run the `include(test/reshape.jl")` myself but not in `] test DataFrames`? I'm quite confused by it.

bkamins:
Because you used `@inbounds` (which you really should not do unless you are 100% sure that your code is correct). In this case your code is not correct. The problem (or maybe a blessing is that AFAICT CI always disables `@inbounds` and catches the bug).

You should write `longnew[counter:counter + l - 1]` (note `- 1` part). What you have written seems to be a typical off-by-1 bug.

To make sure it is the case remove `@inbounds` in your tests locally or start Julia with `--check-bounds=yes` switch.

pdeffebach:
Thank you for your help. Sorry for doing `@inbounds` too early. Thank you for teaching me about how the test tests work!

I have fixed the code, checked that it works without the `@inbounds`, then added back the `@inbounds`. Test failures are now exclusively due to `CategoricalArrays` failues.

bkamins:
> Sorry for doing `@inbounds` too early.

I hope it is OK that I was so direct 😄. It is just a fixed pattern in my head how these things have to be done to be on a safe side (I painfully learned many times not to trust myself when using `@inbounds`).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,80 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
     end
     DataFrame(newcols, _names(df), copycols=false)
 end
+
+"""
+    flatten(df::AbstractDataFrame, cols::Union{Integer, Symbol})
+
+When column `cols` of `df` has iterable elements that define `length`, for example a `Vector`
```
nalimilan: This sentence has no verb. Better start with "Return" from the second sentence and include this part in it?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,80 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+    flattened_col = col_to_flatten isa AbstractVector{<:AbstractVector} ?
+        reduce(vcat, col_to_flatten) :
+        collect(Iterators.flatten(col_to_flatten))
+
+    insertcols!(new_df, columnindex(df, cols), cols => flattened_col)
+
+    return new_df
+end
+
+function repeat_lengths!(longnew::V, shortold::V, lengths::Vector{Int}) where V <: AbstractVector
```
nalimilan: You don't need to be that strict I think... :-)
```suggestion
function repeat_lengths!(longnew::AbstractVector, shortold::AbstractVector, lengths::AbstractVector{Int})
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,80 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+    counter = 1
+    @inbounds for i in 1:length(shortold)
```
```suggestion
    @inbounds for i in eachindex(longnew, shortold, lengths)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,80 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+        l = lengths[i]
+        longnew[counter:counter + l - 1] .= shortold[i]
```
nalimilan: Just for clarity:
```suggestion
        longnew[counter:(counter + l - 1)] .= shortold[i]
```

test/reshape.jl
```diff
@@ -347,5 +347,20 @@ end
                     renamecols=x->string("_", x, "_"))
     @test wide3 == w2[!, Not(2)]
 end
+#
```
nalimilan: Remove this comment.

test/reshape.jl
```diff
@@ -347,5 +347,20 @@ end
...
+    df_tup = DataFrame(a = [1, 2], b = [(1, 2), (3, 4)])
+    ref = DataFrame(a = [1, 1, 2, 2], b = [1, 2, 3, 4])
+    @test flatten(df_vec, :b) == flatten(df_tup, :b) == ref
+    df = DataFrame(a = [1, 2], b = [[1, 2], ["x", "y"]])
+    ref = DataFrame(a = [1, 1, 2, 2], b = [1, 2, "x", "y"])
+    @test flatten(df, :b) == ref
+    df = DataFrame(a = [1, 2], b = [(i for i in 1:5), (i for i in 6:10)])
+    ref = DataFrame(a = [fill(1, 5); fill(2, 5)], b = collect(1:10))
+    @test flatten(df, :b) == ref
+end
```
nalimilan: You mentioned a failure with CategoricalArrays. Can you add a test for that? That applies both to columns to flatten and to other columns.

It would also be good to have a test where vectors to flatten have different element types. One particularly tricky case to check is e.g. `Union{Missing, Int}[1, 2]}` the `Missing` type should be preserved.

nalimilan:
I agree multi-column support isn't needed right now. I just wanted to make sure we keep that possibility open.

pdeffebach:
I have made the changes requested. 

Regarding test failures, they are happening in `test/broadcasting.jl` and appear to be related to `CategoricalArrays`. Originally I thought it was an out-standing issue that was being fixed in separate PRs, but when I switch branches to `master` the test failures are no longer there. 

I thought it implausible that my code changes would have affected this behavior, so I am a bit confused  as to why these failures are showing up now.

src/abstractdataframe/abstractdataframe.jl
pdeffebach: this isn't feasible because they need to have the same indices.

test/reshape.jl
```diff
@@ -348,4 +348,23 @@ end
...
+    df_mixed_types = DataFrame(a = [1, 2], b = [[1, 2], ["x", "y"]])
+    ref_mixed_types = DataFrame(a = [1, 1, 2, 2], b = [1, 2, "x", "y"])
+    @test flatten(df_mixed_types, :b) == ref_mixed_types
+    df_three = DataFrame(a = [1, 2, 3], b = [[1, 2], [10, 20], [100, 200, 300]])
+    ref_three = DataFrame(a = [1, 1, 2, 2, 3, 3, 3], b = [1, 2, 10, 20, 100, 200, 300])
+    @test flatten(df_three, :b) == ref_three
+    df = DataFrame(a = [1, 2], b = [(i for i in 1:5), (i for i in 6:10)])
+    ref = DataFrame(a = [fill(1, 5); fill(2, 5)], b = collect(1:10))
+    @test flatten(df, :b) == ref
+    df_miss = DataFrame(a = [1, 2], b = d = [Union{Missing, Int}[1, 2], Union{Missing, Int}[3, 4]])
```
nalimilan: Even stricter as it also tests promotion:
```suggestion
    df_miss = DataFrame(a = [1, 2], b = [Union{Missing, Int}[1, 2], [3, 4]])
```

test/reshape.jl
```diff
@@ -348,4 +348,23 @@ end
...
+    ref = DataFrame(a = [1, 1, 2, 2], b = Union{Missing, Int}[1, 2, 3, 4])
```
nalimilan: Not needed:
```suggestion
    ref = DataFrame(a = [1, 1, 2, 2], b = [1, 2, 3, 4])
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,80 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+    insertcols!(new_df, columnindex(df, col), col => flattened_col)
+
+    return new_df
+end
+
+function repeat_lengths!(longnew::AbstractVector, shortold::AbstractVector, lengths::AbstractVector{Int})
+    counter = 1
+    @inbounds for i in eachindex(shortold)
+        l = lengths[i]
+        longnew[counter:(counter + l - 1)] .= shortold[i]
```
nalimilan: Actually I realized that this is needed to properly handle the case where `shortold[i]` would be an array. Might be worth a test.
```suggestion
        longnew[counter:(counter + l - 1)] .= Ref(shortold[i])
```

nalimilan:
Thanks. I don't see any failures on Travis.

Can you also add a test for CategoricalArrays?

pdeffebach:
I have addressed all comments. Thank you for catching the potential bug with `Ref()`.

test/reshape.jl
```diff
@@ -348,4 +348,32 @@ end
...
+    ref = DataFrame(a = [1, 1, 2, 2], b = [1, 2, 3, 4])
+    @test flatten(df_miss, :b) == ref
+    v1 = [[1, 2], [3, 4]]
+    v2 = [[5, 6], [7, 8]]
+    v = [v1, v2]
+    df_vec_vec = DataFrame(a = [1, 2], b = v)
+    ref_vec_vec = DataFrame(a = [1, 1, 2, 2], b = [v1 ; v2])
+    @test flatten(df_vec_vec, :b) == ref_vec_vec
+    df_cat = DataFrame(a = [1, 2], b = [CategoricalArray([1, 2]), CategoricalArray([1, 2])])
+    ref_cat = DataFrame(a = [1, 1, 2, 2], b = CategoricalArray([1, 2, 1, 2]))
```
nalimilan: Did you mean this to be `ref`? Anyway, there's no need to make `b` categorical as it doesn't make any difference for `==`. But you have to test that the resulting column is a categorical array.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,80 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
     DataFrame(newcols, _names(df), copycols=false)
 end
+
+"""
+    flatten(df::AbstractDataFrame, col::Union{Integer, Symbol})
+
+When column `col` of data frame `df` has iterable elements that define `length` (for example
+a `Vector` of `Vector`s), return a `DataFrame` where each element of `col` is flattened, meaning 
+the column corresponding to `col` becomes a longer `Vector` where the original entries 
+are concatenated. Elements of row `i` of `df` other than `col` will be duplicated according to
```
bkamins: There is one crucial thing to clarify in the docs before we merge this.

Here "duplicated" does not mean "copied". I think we should clarify this. Essentially this means that if we duplicate mutable objects the column will hold their aliases (which saves memory as a plus, but can be surprising if we do not explicitly mention this).

Do you agree that it is something that should be mentioned?

pdeffebach: You are right. We should either mention it or perform a copy. `flatten`ing and then modifying the original data frame could be a pain point for users.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1435,3 +1435,80 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+│ 4   │ 2     │ s      │
+
+```
+"""
+function flatten(df::AbstractDataFrame, col::ColumnIndex)
+    col_to_flatten = df[!, col]
+    lengths = length.(col_to_flatten)
+    new_df = similar(df[!, Not(col)], sum(lengths))
+
+    for name in names(new_df)
```
```suggestion
    for name in _names(new_df)
```

src/abstractdataframe/abstractdataframe.jl
bkamins: Yes - and I was not sure which approach is better. Let us wait for @nalimilan to comment on this (and we can merge the PR soon). Apart from this comment and a small comment below I think we are done. Thank you!

nalimilan: Good catch. I don't think we can make copies preventively as many objects don't support `copy` and there's no trait to detect that. Just mentioning this should be enough. Maybe say "repeated" rather than "duplicated" for clarity.

bkamins: Yes - copy would have to be a `deepcopy` (sorry for not mentioning this). But let us go with what @nalimilan recommends - and just clarify in the documentation that we do not copy.

pdeffebach:
I made the requested changes, perhaps my second detailed sentence isn't 100% correct, feel free to suggest a change via the github interface that I can easily merge.

bkamins:
OK - I have merged this.

@pdeffebach for your work - thank you for an excellent contribution (and by the way I hope you enjoyed the process - this is something we go through every time a significant PR is made 😄).

@nalimilan - thank you for detailed comments on the PR.

Issue #2017, opened 2019-11-16
Allow `rename!(df, pair::Pair{String, String})` as a signature


xiaodaigh:
```julia
rename!(gmsc, (Symbol("NumberOfTime60-89DaysPastDueNotWorse") => Symbol("NumberOfTime60_89DaysPastDueNotWorse"), Symbol("NumberOfTime30-59DaysPastDueNotWorse") => Symbol("NumberOfTime30_59DaysPastDueNotWorse")))
```
is my rename code. I think it's better if we allow `String` in the rename signature

```julia
rename!(gmsc, ("NumberOfTime60-89DaysPastDueNotWorse" => "NumberOfTime60_89DaysPastDueNotWorse", "NumberOfTime30-59DaysPastDueNotWorse" => "NumberOfTime30_59DaysPastDueNotWorse"))
```
because it's more amendable to creating column names using algorithms for manipulating strings.

bkamins:
This is a discussion we had some time ago with @nalimilan.
The conclusion was to allow `Symbol` only (at least for now).
@nalimilan - do you remember the reason for this?

@xiaodaigh you can use broadcasting to save having to write one `Symbol`.

nalimilan:
I think one of the reasons to use symbols everywhere is that they are more efficient for column lookup. But even that may have changed due to https://github.com/JuliaLang/julia/pull/32437, so we should benchmark it.

Regarding `rename` in particular, I would say it's a case where accepting strings would make sense since 1) performance isn't critical, 2) string processing functions do not accept nor return symbols, and 3) there's little risk of confusion with other functions like `getindex`. But unfortunately, even with that change, when one passes a function to `rename` it will have to take a symbol, nor a string, unless we radically change the API.

bkamins:
Then it is OK to make a PR adding this.

The "function" form has to accept `Symbol` but we could allow it to return anything (a string in particular) and cast it to `Symbol`, so it should be just one `string` call extra, which is not a huge deal.

pdeffebach:
This probably hurts consistency of API, as now people will expect `String`s to work in other contexts. I think it's best to let users call `Symbol` themselves.

nalimilan:
I'm not too worried about that since `rename` is quite particular.

PR #2018, opened 2019-11-17
allow strings in rename!

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2017

What I change here:
* remove `rename` for `Index` as it is not used and could create confusion; also clean up `rename!` to be strictly typed as it is easier to reason about
* in `rename!(f, df)` form allow value returned by `f` to be anything and cast it to `Symbol`
* add `rename!` accepting a vector of `AbstractString` (typed), do not allow mixing `Symbol` and `AbstractString` in this form (use `Symbol` broadcasting if needed).
* in collection of `Pair`s explicitly accept only three forms: `AbstractVector`, `AbstractDict`, list of `Pair`s as kwargs BUT allow mixing of `Symbol` and `AbstractString` (this might be useful eg. when old names are `Symbols` but new names are `AbstractString`) - if you feel mixing should be not allowed please comment

What is to be done:
* additional tests
* update documentation (I am waiting for https://github.com/JuliaData/DataFrames.jl/pull/2009 to be merged before I change this)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -67,22 +67,47 @@ _names(df::AbstractDataFrame) = _names(index(df))
...
-function rename!(df::AbstractDataFrame, args...)
-    rename!(index(df), args...)
+function rename!(df::AbstractDataFrame, vals::AbstractVector{<:AbstractString};
+                 makeunique::Bool=false)
+    rename!(index(df), Symbol.(vals), makeunique=makeunique)
     return df
 end
+
+function rename!(df::AbstractDataFrame,
+                 args::Union{AbstractVector{<:Pair}, AbstractDict})
```
nalimilan: Why not do this to avoid having to check manually?
```suggestion
                 args::Union{AbstractVector{<:Pair{Union{Symbol,String}, Union{Symbol,String}}, AbstractDict})
```

bkamins: The reason is type promotion rules:
```
julia> ["a"=>"b", :c=>:d]
2-element Array{Pair{Any,Any},1}:
 "a" => "b"
  :c => :d
```
and you would have to write manually:
```
julia> Pair{Union{Symbol,String}, Union{Symbol,String}}["a"=>"b", :c=>:d]
2-element Array{Pair{Union{String, Symbol},Union{String, Symbol}},1}:
 "a" => "b"
  :c => :d
```
which I do not think anyone would do in practice.

nalimilan: I see. But do we really want to allow mixing these?

bkamins: I can disallow them, but `Pair{Union{Symbol,String}, Union{Symbol,String}}` actually allows mixing.

So do we want to be stricter and actually only allow `Pair{Union{T}, Union{U}} where {T<:Union{Symbol, AbstractString}, U<:Union{Symbol, AbstractString}}` (so essentially that `first` and `last` are homogeneous).

Or we go for `Pair{Union{Symbol,String}, Union{Symbol,String}}` signature and technically allow mixing but in practice disallow it due to the problem I have described above (this would be simplest to change in the code).

nalimilan: I don't have a strong opinion, but being strict would probably be OK.

bkamins:
```suggestion
                 args::Union{AbstractVector{<:Pair{Union{Symbol,AbstractString}, Union{Symbol,AbstractString}}, AbstractDict})
```

bkamins:
Well - I decided to go with your original suggestion (so we are not strict but we are strict with the signature).

I will not add tests to this functionality and update the documentation (unless we have some more things to discuss regarding the functionality provided).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -67,22 +67,47 @@ _names(df::AbstractDataFrame) = _names(index(df))
...
+                 makeunique::Bool=false)
+    rename!(index(df), Symbol.(vals), makeunique=makeunique)
     return df
 end
+
+function rename!(df::AbstractDataFrame,
+                 args::Union{AbstractVector{<:Pair{Union{Symbol,AbstractString}, Union{Symbol,AbstractString}}, AbstractDict})
+    args_vec = Vector{Pair{Symbol, Symbol}}(undef, length(args))
+    for (i, (from, to)) in args
+        if !(from isa Union{Symbol, AbstractString})
```
nalimilan: This isn't needed with the strict signature, right?

src/abstractdataframe/abstractdataframe.jl
bkamins: OK - I will change it to be strict then.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -67,22 +67,48 @@ _names(df::AbstractDataFrame) = _names(index(df))
...
+                 makeunique::Bool=false)
+    rename!(index(df), Symbol.(vals), makeunique=makeunique)
     return df
 end
+
+rename!(df::AbstractDataFrame, args::AbstractVector{Pair{Symbol,Symbol}}) =
+    rename!(index(df), args)
+
+function rename!(df::AbstractDataFrame,
+                 args::Union{AbstractVector{<:Pair{Symbol,<:AbstractString}},
```
nalimilan: Is this needed because a `Union` triggers an ambiguity? I wouldn't have though so...

bkamins: What do you mean by "is this needed", as it can have two meanings - I will comment on both of them then:
1. why double `<:` is needed (in front of `Pair` and in front of `AbstrctString`) - this is needed because this is how parametric subtyping works
2. why I make a union of 7 different specifications? The reason is that writing `Pair{<:Union{Symbol,AbstractString}, <:Union{Symbol,AbstractString}}` would allow mixing `Symbols` and strings and we agreed that we want to disallow mixing. I could have written a macro that would iterate through all the options to generate appropriate methods but I decided that it would be too much magic and it is simpler just to make such a large union.

nalimilan: Ah, right. That's really too bad we don't have a more concise way of doing this...

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -67,22 +67,48 @@ _names(df::AbstractDataFrame) = _names(index(df))
...
+                             AbstractVector{<:Pair{<:AbstractString,Symbol}},
+                             AbstractVector{<:Pair{<:AbstractString,<:AbstractString}},
+                             AbstractDict{Symbol,Symbol},
+                             AbstractDict{Symbol,<:AbstractString},
+                             AbstractDict{<:AbstractString,Symbol},
+                             AbstractDict{<:AbstractString,<:AbstractString}})
+    args_vec = Vector{Pair{Symbol, Symbol}}(undef, length(args))
```
nalimilan: I realize this could just be `[Symbol(from) => Symbol(to) for (from, to) in args]`, right?

bkamins: sure - I do not know why I did not use this pattern 😄.

bkamins:
Thank you!

PR #2021, opened 2019-11-21
Key in warning doesn't reflect the the right entry

laborg:
```julia
using DataFrames
a = DataFrame( i=["a","a","b","b"], j=["a1","a1","b1","b1"],
    var1=[1,2,3,3], val=[11,22,33,44])

unstack(a,:i,:var1,:val)
#┌ Warning: Duplicate entries in unstack at row 4 for key b and variable 3.
# > correct key

unstack(a,[:i,:j],:var1,:val)
#┌ Warning: Duplicate entries in unstack at row 4 for key ("a", "a1") and variable 3.
# > key ("a", "a1") isn't the duplicate

# this PR:
unstack(a,[:i,:j],:var1,:val)
#┌ Warning: Duplicate entries in unstack at row 4 for key ("b", "b1") and variable 3.
# > corrected key
```

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -287,7 +287,7 @@ function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Symbol},
         i = rowkey[k]
         if !warned_dup && mask_filled[i, j]
             @warn("Duplicate entries in unstack at row $k for key "*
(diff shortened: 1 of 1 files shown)


bkamins:
I think it is OK to leave it as is (it is only a change in the warning printed). Thank you for the contribution.

PR #1958, opened 2019-09-13
Add :setequal to cols kwarg in vcat, push! and append!

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1904

I think we should make `vcat` and `push!` consistent with the name of kwarg. Should it be `cols` (now used in `vcat`) or `columns` (now used in `push!`). Any opinions?

PR #2022, opened 2019-11-22
make SubIndex strict about duplicates

bkamins:
see the discussion in https://github.com/JuliaData/DataFrames.jl/pull/1958.

Rationale for this change:
* the case when the cost of creation of `DataFrameRow` or `SubDataFrame` will be affected is only when indexing by `AbstractVector{Int}` is performed (abut not ranges - as they will be fast) - so I guess the case when it is needed is pretty rare
* if someone really wants to avoid checking for duplicates it is possible to call `SubIndex` default constructor with appropriate values
* the additional benefit is that functions that previously called `lazyremap!` have a bit less work to do later

### DIFF
--- a/src/other/index.jl
+++ b/src/other/index.jl
@@ -329,7 +329,6 @@ Base.@propagate_inbounds function parentcols(ind::SubIndex, idx::Symbol)
     parentcol = ind.parent[idx]
     @boundscheck begin
         remap = ind.remap
-        length(remap) == 0 && lazyremap!(ind)
         remap[parentcol] == 0 && throw(ArgumentError("$idx not found"))
     end
     return parentcol
@@ -345,22 +344,29 @@ Base.@propagate_inbounds parentcols(ind::SubIndex, ::Colon) = ind.cols
 
 Base.@propagate_inbounds parentcols(ind::SubIndex, idx::Not) = parentcols(ind, ind[idx])
 
-Base.@propagate_inbounds function SubIndex(parent::AbstractIndex, cols::AbstractUnitRange{Int})
+function SubIndex(parent::AbstractIndex, cols::AbstractUnitRange{Int})
     l = last(cols)
     f = first(cols)
-    @boundscheck if !checkindex(Bool, Base.OneTo(length(parent)), cols)
+    if !checkindex(Bool, Base.OneTo(length(parent)), cols)
         throw(BoundsError("invalid columns $cols selected"))
     end
     remap = (1:l) .- f .+ 1
     SubIndex(parent, cols, remap)
 end
 
-Base.@propagate_inbounds function SubIndex(parent::AbstractIndex, cols::AbstractVector{Int})
+function SubIndex(parent::AbstractIndex, cols::AbstractVector{Int})
     ncols = length(parent)
-    @boundscheck if !all(x -> 0 < x ≤ ncols, cols)
-        throw(BoundsError("invalid columns $cols selected"))
+    remap = zeros(Int, ncols)
+    for (i, col) in enumerate(cols)
+        if !(1 <= col <= ncols)
+            throw(BoundsError("column index must be greater than zero " *
+                                "and not larger than number columns in the parent"))
+        end
+        if remap[col] != 0
+            throw(ArgumentError("duplicate selected column detected"))
+        end
+        remap[col] = i
     end
-    remap = Int[]
     SubIndex(parent, cols, remap)
 end
 
@@ -370,21 +376,6 @@ end
 Base.@propagate_inbounds SubIndex(parent::AbstractIndex, cols) =
     SubIndex(parent, parent[cols])
 
-# a helper function that lazily creates remap when needed
-function lazyremap!(x::SubIndex)
-    remap = x.remap
-    remap isa AbstractUnitRange{Int} && return remap
-    if length(remap) == 0
-        resize!(remap, length(x.parent))
-        # we set non-existing mappings to 0
-        fill!(remap, 0)
-        for (i, col) in enumerate(x.cols)
-            remap[col] = i
-        end
-    end
-    remap
-end
-
 Base.length(x::SubIndex) = length(x.cols)
 Base.names(x::SubIndex) = copy(_names(x))
 _names(x::SubIndex) = view(_names(x.parent), x.cols)
@@ -393,7 +384,6 @@ function Base.haskey(x::SubIndex, key::Symbol)
     haskey(x.parent, key) || return false
     pos = x.parent[key]
     remap = x.remap
-    length(remap) == 0 && lazyremap!(x)
     checkbounds(Bool, remap, pos) || return false
     remap[pos] > 0
 end
@@ -405,7 +395,6 @@ Base.keys(x::SubIndex) = names(x)
 
 function Base.getindex(x::SubIndex, idx::Symbol)
     remap = x.remap
-    length(remap) == 0 && lazyremap!(x)
     remap[x.parent[idx]]
 end
 
--- a/test/cat.jl
+++ b/test/cat.jl
@@ -267,10 +267,8 @@ end
 
     df = DataFrame(A=1:5, B=11:15, C=21:25)
     @test vcat(view(df, 1:2, :), view(df, 3:5, [3,2,1])) == df
-    @test_throws ArgumentError vcat(view(df, 1:2, [1,2,3,1,2,3]),
-                                    view(df, 3:5, [3,2,1,1,2,3])) == df
-    @test_throws ArgumentError all(==(df[1, :]), eachrow(vcat(view(df, [1,1,1], [1,2,3,1,2,3]),
-                                                              view(df, [1,1,1], [3,2,1,1,2,3]))))
+    @test_throws ArgumentError view(df, 1:2, [1,2,3,1,2,3])
+    @test_throws ArgumentError view(df, 3:5, [3,2,1,1,2,3])
 end
 
 @testset "vcat copy" begin
--- a/test/cat.jl
+++ b/test/cat.jl
@@ -267,9 +267,10 @@ end
 
     df = DataFrame(A=1:5, B=11:15, C=21:25)
     @test vcat(view(df, 1:2, :), view(df, 3:5, [3,2,1])) == df
-    @test vcat(view(df, 1:2, [1,2,3,1,2,3]), view(df, 3:5, [3,2,1,1,2,3])) == df
-    @test all(==(df[1, :]), eachrow(vcat(view(df, [1,1,1], [1,2,3,1,2,3]),
-                                         view(df, [1,1,1], [3,2,1,1,2,3]))))
+    @test_throws ArgumentError vcat(view(df, 1:2, [1,2,3,1,2,3]),
+                                    view(df, 3:5, [3,2,1,1,2,3])) == df
+    @test_throws ArgumentError all(==(df[1, :]), eachrow(vcat(view(df, [1,1,1], [1,2,3,1,2,3]),
+                                                              view(df, [1,1,1], [3,2,1,1,2,3]))))
 end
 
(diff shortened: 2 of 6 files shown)


src/other/index.jl
```diff
@@ -345,22 +344,29 @@ Base.@propagate_inbounds parentcols(ind::SubIndex, ::Colon) = ind.cols
 
 Base.@propagate_inbounds parentcols(ind::SubIndex, idx::Not) = parentcols(ind, ind[idx])
 
-Base.@propagate_inbounds function SubIndex(parent::AbstractIndex, cols::AbstractUnitRange{Int})
+function SubIndex(parent::AbstractIndex, cols::AbstractUnitRange{Int})
     l = last(cols)
     f = first(cols)
-    @boundscheck if !checkindex(Bool, Base.OneTo(length(parent)), cols)
```
nalimilan: Why not still allow using `@inbounds` to skip the checks? Doesn't seem it would harm.

bkamins: I have checked that the benefit is so small and in only a particular case of range that I thought it is not worth allowing it to make sure we are always consistent (as noted - if someone really wants speed one can call `SubIndex` directly - probably this is not advisable, but possible). This check is really cheap - most of the time compiler will optimize it out due to constant propagation.

But I can reintroduce it if you prefer.

nalimilan: OK. So the `@boundscheck` is still used in places where it does make a difference?

bkamins: I have removed all `@boundscheck` in constructors (there were two places of this - this one and the one that now replaces `lazyremap!` functionality). But we do `@boundscheck` in other functions (that get called on the created objects).

src/other/index.jl
```diff
@@ -345,22 +344,29 @@ Base.@propagate_inbounds parentcols(ind::SubIndex, ::Colon) = ind.cols
...
-Base.@propagate_inbounds function SubIndex(parent::AbstractIndex, cols::AbstractVector{Int})
+function SubIndex(parent::AbstractIndex, cols::AbstractVector{Int})
     ncols = length(parent)
-    @boundscheck if !all(x -> 0 < x ≤ ncols, cols)
-        throw(BoundsError("invalid columns $cols selected"))
+    remap = zeros(Int, ncols)
+    for (i, col) in enumerate(cols)
+        if !(1 <= col <= ncols)
+            throw(BoundsError("column index must be greater than zero " *
+                                "and not larger than number columns in the parent"))
```
```suggestion
                              "and not larger than number columns in the parent"))
```

src/other/index.jl
nalimilan: I thought at https://github.com/JuliaData/DataFrames.jl/pull/1958#discussion_r348938463 you said the performance impact was large for `DataFrameRow`?

bkamins:
> I thought at #1958 (comment) you said the performance impact was large for `DataFrameRow`?

I have said that the performance hit is large for selection of kind `df[1, [1,2,3]]`.
Here are the benchmarks for `DataFrameRow` (as this is the most critical kind):

Current implementation
```
julia> df = DataFrame(rand(50, 50));

julia> x = collect(1:50);

julia> y = names(df);

julia> @benchmark df[1, :]
BenchmarkTools.Trial:
  memory estimate:  32 bytes
  allocs estimate:  1
  --------------
  minimum time:     35.749 ns (0.00% GC)
  median time:      36.557 ns (0.00% GC)
  mean time:        52.476 ns (11.99% GC)
  maximum time:     39.307 μs (99.87% GC)
  --------------
  samples:          10000
  evals/sample:     993

julia> @benchmark df[1, 1:50]
BenchmarkTools.Trial:
  memory estimate:  80 bytes
  allocs estimate:  2
  --------------
  minimum time:     46.107 ns (0.00% GC)
  median time:      47.422 ns (0.00% GC)
  mean time:        67.729 ns (15.78% GC)
  maximum time:     41.870 μs (99.86% GC)
  --------------
  samples:          10000
  evals/sample:     989

julia> @benchmark df[1, $x]
BenchmarkTools.Trial:
  memory estimate:  144 bytes
  allocs estimate:  3
  --------------
  minimum time:     88.554 ns (0.00% GC)
  median time:      93.652 ns (0.00% GC)
  mean time:        122.653 ns (11.30% GC)
  maximum time:     41.732 μs (99.51% GC)
  --------------
  samples:          10000
  evals/sample:     961

julia> @benchmark df[1, $y]
BenchmarkTools.Trial:
  memory estimate:  4.55 KiB
  allocs estimate:  16
  --------------
  minimum time:     2.722 μs (0.00% GC)
  median time:      2.911 μs (0.00% GC)
  mean time:        4.634 μs (19.63% GC)
  maximum time:     4.658 ms (99.86% GC)
  --------------
  samples:          10000
  evals/sample:     9
```

Proposed in this PR:
```
julia> df = DataFrame(rand(50, 50));

julia> x = collect(1:50);

julia> y = names(df);

julia> @benchmark df[1, :]
BenchmarkTools.Trial:
  memory estimate:  32 bytes
  allocs estimate:  1
  --------------
  minimum time:     37.261 ns (0.00% GC)
  median time:      38.570 ns (0.00% GC)
  mean time:        53.947 ns (12.11% GC)
  maximum time:     40.133 μs (99.87% GC)
  --------------
  samples:          10000
  evals/sample:     993

julia> @benchmark df[1, 1:50]
BenchmarkTools.Trial:
  memory estimate:  80 bytes
  allocs estimate:  2
  --------------
  minimum time:     45.804 ns (0.00% GC)
  median time:      50.152 ns (0.00% GC)
  mean time:        81.071 ns (13.05% GC)
  maximum time:     41.938 μs (99.87% GC)
  --------------
  samples:          10000
  evals/sample:     989

julia> @benchmark df[1, $x]
BenchmarkTools.Trial:
  memory estimate:  560 bytes
  allocs estimate:  3
  --------------
  minimum time:     126.487 ns (0.00% GC)
  median time:      138.047 ns (0.00% GC)
  mean time:        171.026 ns (10.79% GC)
  maximum time:     46.290 μs (99.41% GC)
  --------------
  samples:          10000
  evals/sample:     891

julia> @benchmark df[1, $y]
BenchmarkTools.Trial:
  memory estimate:  4.95 KiB
  allocs estimate:  16
  --------------
  minimum time:     2.744 μs (0.00% GC)
  median time:      2.944 μs (0.00% GC)
  mean time:        4.555 μs (20.23% GC)
  maximum time:     4.988 ms (99.90% GC)
  --------------
  samples:          10000
  evals/sample:     9
```

And what I say is that these nanosecond differences will be negligible anyway as if you try doing anything later with such `DataFrameRow` it will be more expensive anyway (as I assume that if you create it you actually want to do something with it that is more complex). Here is an example of cost of casting it to a `NamedTuple`:
```
julia> dfr =  df[1, y];

julia> @benchmark copy($dfr)
BenchmarkTools.Trial:
  memory estimate:  4.20 KiB
  allocs estimate:  115
  --------------
  minimum time:     5.483 μs (0.00% GC)
  median time:      5.833 μs (0.00% GC)
  mean time:        7.541 μs (10.77% GC)
  maximum time:     7.461 ms (99.89% GC)
  --------------
  samples:          10000
  evals/sample:     6
```

nalimilan:
That's fine. But why not keep these `@boundscheck` annotations as they don't hurt?

bkamins:
We can discuss it. It is a matter of taste. I prefer as little magic as possible in the source code, so the average DataFrames.jl user can understand what is going on (the codebase for indexing and views is already quite complex I think).

If we revert `@boundscheck` I have to add `lazyremap!` function back to the code base and then also `parentindices`, `getindex` and `haskey` will have do one additional check (as we will dynamically have to check after creation of `SubIndex` if it was created with `@inbounds` or without it).

But - if you prefer I will revert `@boundscheck` - so feel free to decide if it is OK with you.

bkamins:
So in short having or not having `@boundscheck` is a tradeoff: for a small performance gain in a rare case, we add code complexity and small performance hit in several other places.

test/dataframerow.jl
```diff
@@ -298,7 +298,7 @@ end
     @test parentindices(sdf[2, r"a"]) == (3, [1])
     @test parentindices(sdf[2, r"x"]) == (3, Int[])
     @test parent(sdf[1, 1:2]) === df
-    @test parentindices(sdf[1, [2, 2]]) == (4, [1, 1])
+    @test_throws ArgumentError parentindices(sdf[1, [2, 2]]) == (4, [1, 1])
```
nalimilan:
```suggestion
    @test_throws ArgumentError parentindices(sdf[1, [2, 2]])
```

test/dataframerow.jl
```diff
@@ -318,7 +318,7 @@ end
...
```
nalimilan:
```suggestion
    @test_throws ArgumentError parentindices(sdf[1, [2, 2]])
```

test/dataframerow.jl
```diff
@@ -341,22 +341,8 @@ end
...
-    dfr1.x2 = 100
-    @test values(dfr1) == (100, 100, 100)
-    @test df[2, 2] == 100
-    @test_throws ArgumentError dfr1.x1
-    @test dfr2.x1 == 11
-    dfr2.x1 = 200
-    @test values(dfr2) == (200, 200, 200)
-    @test df[1, 1] == 200
-    @test_throws ArgumentError dfr2.x2
+    @test_throws ArgumentError dfr1 = df[2, [2,2,2]]
```
nalimilan:
```suggestion
    @test_throws ArgumentError df[2, [2,2,2]]
```

test/dataframerow.jl
```diff
@@ -341,22 +341,8 @@ end
...
+    @test_throws ArgumentError dfr2 = sdf[2, [2,2,2]]
```
nalimilan:
```suggestion
    @test_throws ArgumentError sdf[2, [2,2,2]]
```

test/subdataframe.jl
```diff
@@ -241,14 +241,7 @@ end
...
     df = DataFrame([11:16 21:26 31:36 41:46])
-    sdf = view(df, [3,1,4], [3,3,3])
-    @test names(sdf) == fill(:x3, 3)
-    @test sdf[!, 1] == [33, 31, 34]
-    @test sdf[!, 1] === sdf[!, 2] === sdf[!, 3]
-    @test sdf.x3[1] == 33
-    sdf.x3[1] = 333
-    @test df.x3[3] == 333
-    @test_throws ArgumentError sdf.x1
+    @test_throws ArgumentError sdf = view(df, [3,1,4], [3,3,3])
```
nalimilan:
```suggestion
    @test_throws ArgumentError view(df, [3,1,4], [3,3,3])
```

test/cat.jl
```diff
@@ -267,9 +267,10 @@ end
 
     df = DataFrame(A=1:5, B=11:15, C=21:25)
     @test vcat(view(df, 1:2, :), view(df, 3:5, [3,2,1])) == df
-    @test vcat(view(df, 1:2, [1,2,3,1,2,3]), view(df, 3:5, [3,2,1,1,2,3])) == df
-    @test all(==(df[1, :]), eachrow(vcat(view(df, [1,1,1], [1,2,3,1,2,3]),
-                                         view(df, [1,1,1], [3,2,1,1,2,3]))))
+    @test_throws ArgumentError vcat(view(df, 1:2, [1,2,3,1,2,3]),
```
nalimilan: `vcat` isn't needed anymore, right? Same below for `all`.

nalimilan:
OK, fine with me.

bkamins:
Re: `vcat` not needed - fixed. Thank you (I wanted to minimize changes in the tests but here it does not make sense to retain them - so also thank you for the other fixes).

I have added another tests for duplicate columns when trying to make a `view` of a `SubDataFrame`.

bkamins:
@nalimilan - thank you!

PR #2025, opened 2019-11-26
add Array convert for data frame and data frame row

bkamins:
Following the discussion on Slack

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -189,6 +189,11 @@ Base.convert(::Type{Vector{T}}, dfr::DataFrameRow) where T =
 Base.Vector(dfr::DataFrameRow) = convert(Vector, dfr)
 Base.Vector{T}(dfr::DataFrameRow) where T = convert(Vector{T}, dfr)
 
+Base.convert(::Type{Array}, dfr::DataFrameRow) = Vector(dfr)
+Base.convert(::Type{Array{T}}, dfr::DataFrameRow) where {T} = Vector{T}(dfr)
+Base.Array(dfr::DataFrameRow) = Vector(dfr)
+Base.Array{T}(dfr::DataFrameRow) where {T} = Vector{T}(dfr)
+
 Base.keys(r::DataFrameRow) = Tuple(_names(r))
 Base.values(r::DataFrameRow) =
     ntuple(col -> parent(r)[row(r), parentcols(index(r), col)], length(r))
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Looks good to me

ChrisRackauckas:
Thanks for this!

src/dataframerow/dataframerow.jl
```diff
@@ -189,6 +189,11 @@ Base.convert(::Type{Vector{T}}, dfr::DataFrameRow) where T =
 Base.Vector(dfr::DataFrameRow) = convert(Vector, dfr)
 Base.Vector{T}(dfr::DataFrameRow) where T = convert(Vector{T}, dfr)
 
+Base.convert(::Type{Array}, dfr::DataFrameRow) = Vector(dfr)
```
nalimilan: This is a bit more controversial as `DataFrameRow` could also be viewed as a single-row matrix. Maybe let's leave it out for now?

BTW, we define `Base.size(r::DataFrameRow) = (length(index(r)),)`, but isn't this at odds with how broadcasting works for `DataFrameRow`? I.e. it could also be `(1,length(index(r)))`.

bkamins: We allow `Vector` to be called on  `DataFrameRow`. Since we do this then all other interpretations are ruled out I think. Doing `df[row, cols]` drops one level of dimensionality from a data frame to make it 1-dimensional object just like in arrays.

Note that we do not allow:
```
julia> using DataFrames

julia> df = DataFrame(ones(2,3))
2×3 DataFrame
│ Row │ x1      │ x2      │ x3      │
│     │ Float64 │ Float64 │ Float64 │
│ 1   │ 1.0     │ 1.0     │ 1.0     │
│ 2   │ 1.0     │ 1.0     │ 1.0     │

julia> df[1, :] .= [1, 2, 3]
DataFrameRow
│ Row │ x1      │ x2      │ x3      │
│     │ Float64 │ Float64 │ Float64 │
│ 1   │ 1.0     │ 2.0     │ 3.0     │
```
but we do not allow:
```
julia> df[1, :] .= [1 2 3]
```
but we allow (as `1:1` does not drop a dimension):
```
julia> df[1:1, :] .= [1 2 3]
1×3 SubDataFrame
│ Row │ x1      │ x2      │ x3      │
│     │ Float64 │ Float64 │ Float64 │
│ 1   │ 1.0     │ 2.0     │ 3.0     │
```

Currently we do not allow:
```
julia> df[1, :] .= df[1, :]
ERROR: ArgumentError: broadcasting over `DataFrameRow`s is reserved
```
(i.e. for `DataFrameRow` to be on RHS of broadcasted assignment, as we are waiting for Base to decide how to treat `NamedTuple` in the same place).

In summary: all the infrastructure of DataFrames.jl for `DataFrameRow` assumes it is 1-dimensional.

nalimilan: OK, I guess that's fine then.

bkamins:
Thank you!

ChrisRackauckas:
Awesome! Any idea on when this will get released?

bkamins:
I think in a few days at most. See https://github.com/JuliaData/DataFrames.jl/pull/2028 for the list of pending PRs (essentially 1 PR that is finished but requires a final check + possibly 2 PRs that we will exclude from that release if we cannot get consensus on them)

PR #2029, opened 2019-12-01
Stop reexporting packages

nalimilan:
Doing this automatically includes all functions exported by CategoricalArrays, Missings and InvertedIndices to our API. Instead, export selected functions that we really use and/or overload. This has the most significant consequences for CategoricalArrays (`levels`, `isordered`, `CategoricalArray`...). The advantage is that this will allow us to drop this dependency if we want (e.g. by defining `categorical`, `allowmissing!`, etc. in a lightweight interface package).


bkamins:
Here are my thoughts:
* InvertedIndices.jl: it is ok just to export `Not`
* Missings.jl: there are three functions that are useful, and would not get exported: `levels`, `passmissing`, `nonmissingtype` (the last one for compatibility between Julia 1.0 and 1.3) - I thought it was intentional that we wanted to have them automatically exported when DataFrames.jl is loaded
* CategoricalArrays.jl: I think here we it does not make much sense to export `categorical` and `categorical!` and not export the rest

So in summary we should decide:
* if we want DataFrames.jl to be "low-level" and recommend users to use StatsKit.jl by default (and only solely load DataFrames.jl if there is really a reason) then we should drop CategoricalArrays.jl and exports from Missings.jl
* if we want to keep DataFrames.jl relatively stand-alone we need some more exports as otherwise users will anyway often need to load Missings.jl and CategoricalArrays.jl for compatibility/ability to work smoothly

nalimilan:
> Missings.jl: there are three functions that are useful, and would not get exported: `levels`, `passmissing`, `nonmissingtype` (the last one for compatibility between Julia 1.0 and 1.3) - I thought it was intentional that we wanted to have them automatically exported when DataFrames.jl is loaded

`levels` now lives in DataAPI. It sounds OK to reexport it from there. Reexporting `nonmissingtype` on older Julia versions also sounds OK, as anyway it's exported by Base in recent versions. I guess we could also do that with `passmissing`, though at that point it's almost as if we reexported all of Missings.jl. But at least the API would be fixed.

> CategoricalArrays.jl: I think here we it does not make much sense to export `categorical` and `categorical!` and not export the rest

I exported them because `categorical!` doesn't (and cannot) exist in CategoricalArrays, so we have to export it ourselves, and then exporting `categorical` sounds logical. But I see your point, it's a bit weird to export these and not e.g. `levels!` and `isordered`.

That discussion extends to potential `pool` and `pool!` methods that we could add for `PooledArray`.

> So in summary we should decide:
> 
>     * if we want DataFrames.jl to be "low-level" and recommend users to use StatsKit.jl by default (and only solely load DataFrames.jl if there is really a reason) then we should drop CategoricalArrays.jl and exports from Missings.jl
> 
>     * if we want to keep DataFrames.jl relatively stand-alone we need some more exports as otherwise users will anyway often need to load Missings.jl and CategoricalArrays.jl for compatibility/ability to work smoothly

"Low-level" is a bit strong, but generally speaking it would be nice to have a relatively self-contained package that allows using `DataFrame` without dependencies which aren't strictly necessary. Some had proposed to have a DataFramesBase package with only type definitions, but that would be so limited that it wouldn't be much more useful than what Tables.jl provides IMHO. Though we could decide to put more things in DataFramesBase, e.g. everything except heavy dependencies.

Alternatively, recommending people to use StatsKit sounds like a good approach in general. If we want people to be able to do data management without the stats methods, we could also introduce a new broader package, just like tidyverse vs. tibble in R.

bkamins:
Since we will import Missings.jl anyway I see no problem with exporting `nonmissingtype` (if needed) and `passmissing` + `levels` (no matter what source package is used). This will not have a performance impact, and these functions are just useful.

Now regarding CategoricalArrays.jl the question is: is re-exporting CategoricalArrays.jl expensive or is even using it internally expensive? Do you have timings comparing this?
If only re-exporting is expensive then we could do what you propose now (i.e. I do not see a large problem to tell people a simple sentence "use StatsKit.jl if you want to do general stats and use categorical arrays in particular"). However, if even loading CategoricalArrays.jl internally is expensive then we have a problem as it is deeply buried in split-apply-combine code and I have just checked (and you probably know better) that it will be probably hard do decouple it.

bkamins:
I have checked and:
* this PR does not decrease load time of DataFrames.jl (but can you please confirm with your tests)
* we have settled to have CategoricalArrays.jl as an internal dependency for now, which means that we can as well re-export it

If we decide not to reexport then I would export:
`categorical`, `levels`, `cut`, `recode`, `recode!`, `nonmissingtype`, `passimissing`, and `Not`.

bkamins:
@nalimilan - given the discussion we had - do you think it is worth to stop reexporting the packages?

nalimilan:
OK, let's forget this. I've opened #2074 for the documentation fixes.

PR #2031, opened 2019-12-02
deprecate stackdf and meltdf

### LINKED ISSUE #1736: Find a better API for stackdf and meltdf
These names really make no sense. We could use `@view` if possible or add an argument to the corresponding functions.

bkamins:
Fixes #1736. It does not have to go into v0.20 if some problems are found.


src/abstractdataframe/reshape.jl
```diff
@@ -12,9 +12,10 @@ column `variable_name` (`:variable` by default) a vector of `Symbol`s holding
 the name of the corresponding `measure_vars` variable,
 and with columns for each of the `id_vars`.
 
-See also [`stackdf`](@ref)) and [`meltdf`](@ref) for stacking methods that return a
-view into the original data frame. See [`unstack`](@ref) for converting from
-long to wide format.
+If `view=true` then return a stacked view of a data frame (long format).
+The result is a view because the columns are special `AbstractVectors`
+that return indexed views into the original data frame.
```
nalimilan: What does "indexed" mean?

src/abstractdataframe/reshape.jl
```diff
@@ -31,6 +32,8 @@ long to wide format.
   of each of `measure_vars`
 - `value_name` : the name of the new stacked column containing the values from
   each of `measure_vars`
+- `view` : if the stacked data frame should be a view or it should contain
```
nalimilan:
```suggestion
- `view` : whether the stacked data frame should be a view rather than contain
```

src/abstractdataframe/reshape.jl
```diff
@@ -50,7 +53,9 @@ d1s_name = melt(d1, [:a, :b, :e], variable_name=:somemeasure)
 
 function stack(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
                id_vars::AbstractVector{<:Integer}; variable_name::Symbol=:variable,
-               value_name::Symbol=:value)
+               value_name::Symbol=:value, view::Bool=false)
+    view && return _stackdf(df, measure_vars, id_vars, variable_name=variable_name,
```
nalimilan: `if view` would look cleaner IMHO.

src/abstractdataframe/reshape.jl
```diff
@@ -413,99 +412,61 @@ function CategoricalArrays.CategoricalArray(v::RepeatedVector)
...
-d1s = stackdf(d1, [:c, :d])
-d1s2 = stackdf(d1, [:c, :d], [:a])
-d1m = meltdf(d1, [:a, :b, :e])
-```
-"""
-(stackdf, meltdf)
-
-function stackdf(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
+function _stackdf(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
                  id_vars::AbstractVector{<:Integer}; variable_name::Symbol=:variable,
```
nalimilan: Wrong indentation here and below.

But maybe better rename the internal function while you're at it? E.g. `stackview` or `viewstack`?

src/abstractdataframe/reshape.jl
bkamins: I just copy-pasted this from the original description. I guess indexed is intended to mean that it is a custom view. But probably we can just remove "indexed". OK?

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl
bkamins: ok. fixed

src/abstractdataframe/reshape.jl
bkamins: I have removed "indexed"

bkamins:
OK - all should be fixed. I just used `_` in front of internal functions `_stackview` and `_meltview`.

bkamins:
Only coverage fails here.

bkamins:
I will add deprecation of `melt` and `meltdf` in this PR as it duplicates `stack` and `stackdf` and we do not need both since we have `Not` for indexing. I will let you know when this is implemented.

nalimilan:
Maybe merge this now and file another PR as this one seems to be ready.

bkamins:
Thank you for having a look at this.

I had to clean up this PR anyway (I have left some references to `stackdf` in the docs - sorry for this) so I also deprecated `melt`. This is a rather simple fix, as it is either adding `Not` to an argument or changing argument order (you can see the changes in deprecated.jl).

bkamins:
This should be good to have a look at. I also wanted to have it in one shot as previously we changed methods that are removed entirely now.

docs/src/lib/types.md
```diff
@@ -74,8 +74,8 @@ On the contrary, functions that create a view of a `DataFrame` *do not* by defin
 the columns, and therefore require particular caution. This includes `view`, which returns
 a `SubDataFrame` or a `DataFrameRow`, and `groupby`, which returns a `GroupedDataFrame`.
 
-A partial exception to this rule are the [`stackdf`](@ref) and [`meltdf`](@ref) functions which
-create a `DataFrame` that contains views of the columns from the source `DataFrame`.
+A partial exception to this rule are the [`stack`](@ref) function with `view=true` which
```
```suggestion
A partial exception to this rule is the [`stack`](@ref) function with `view=true` which
```

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -123,10 +123,10 @@ julia> last(d, 6)
 │ 6   │ SepalWidth │ 3.0      │ virginica     │
 ```
 
-`melt` is an alternative function to reshape from wide to long format. It is based on `stack`, but it prefers specification of the id columns as:
+If you prefer prefers to specify the id columns then use `Not` with `stack` like this:
```
nalimilan:
```suggestion
If you prefer to specify the id columns then use `Not` with `stack` like this:
```

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -261,10 +261,10 @@ julia> first(widedf, 6)
 │ 6   │ setosa        │ 6     │ 1.7         │ 0.4        │ 5.4         │ 3.9        │
 ```
 
-`stackdf` and `meltdf` are two additional functions that work like `stack` and `melt`, but they provide a view into the original wide DataFrame. Here is an example:
+If you pass `view=true` to `stack` then you get a data frame whose columns are views into the original wide data frame. Here is an example:
```
nalimilan: Maybe use a more impersonal formulation?
```suggestion
Passing `view=true` to `stack` returns a data frame whose columns are views into the original wide data frame. Here is an example:
```

bkamins:
Incorporated all changes. Thank you for reviewing this.

bkamins:
Thank you!

Issue #2032, opened 2019-12-02
Unify `push!`, `append!` and `vcat` implementation.


rapus95:
I intend to rewrite the internal code that does the merging of a DataFrame with another DataFrame, DataFrameRow, NamedTuple or AbstractDict.
Idea is as follows:
* add the capabillity to specify some sort of maximum column-type widening -> `promotetype`
* add a few `cols` column merging algorithms
* `push!`, `append!` and `vcat` are just front end redirectors which call into the following functions
* work consists of two functions:
  * `infersignature(signatureA, signatureB; cols, promotetype)`
  * `__merge!(df, others..., newsignature)`

Details:
----------
   `infersignature(signatureA::Signature, signatureB::Signature; cols::Symbol, promotetype::Type)`

  this infers the signature of the resulting DataFrame from the provided signatures, the `cols` keyword 
  and the `promotetype` according to the following rules:
 1. combine the column names according to `cols` keyword, including:
    * `:orderequal`: names and order must be the same, error otherwise
    * `:setequal`: names can appear in any order but must be present in both signatures, error otherwise
    * `:subset`: the names of A must be a strict subset of (or equal to) the names of B, error otherwise
     * `:superset`: the names of A must be a strict superset of (or equal to) the names of B, error otherwise
    * `:left`: take the names of A
    * `:right`: take the names of B
    * `:intersect`: take the names that are present in both signatures
    * `:union`: take the names that are present in at least one signature
 2. for all column names of the new signature, promote the column types together with the `promotetype` keyword. It'll roughly works like this:
```julia
function promote(tA::Type, tB::Type, promotetype::Type)
  prom = promote_type(tA,tB)
  prom <: Union{promotetype,tA} && return prom
  return tA
end
```
3. return the resulting signature
-----
`__merge!(df::DataFrame, others..., newsignature::Signature)`

this merges all arguments into the first argument, resulting in the signature provided.
Rules:
1. 0-column dataframes act as a true neutral element
2. 0-row dataframes act as a type signature provider (implemented in infersignature by manipulating the signature inference step)
3. Each column of `df` whose type fits the corresponding requested type is reused and resized (hence in-place). Otherwise a new column is allocated and filled.
4. In case of any conversion errors, we reverse resizing (thus removing newly added entries) for all columns.
5. Once we have all columns that will be needed, we replace and reorder the columns in the provided dataframe to mirror the requested signature.

-----

Ref: #1982, #1991

[...]

PR #2034, opened 2019-12-04
Unify `push!`, `append!` and `vcat` implementation.

rapus95:
See https://github.com/JuliaData/DataFrames.jl/issues/2032

There's still a lot of work to be done. But for easier tracking and feedback I open it now.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -51,6 +51,9 @@ the ["Indexing" section](https://juliadata.github.io/DataFrames.jl/stable/lib/in
 """
 abstract type AbstractDataFrame end
 
+const ColumnNameMergingType = Union{Symbol, AbstractVector{Symbol}, Function}
+const ColumnTypeMergingType = Union{Type, AbstractVector{Type}, Function}
+
 ##############################################################################
 ##
 ## Basic properties of a DataFrame
@@ -1063,16 +1066,16 @@ julia> vcat(d4, df1)
 
 """
 Base.vcat(dfs::AbstractDataFrame...;
-          cols::Union{Symbol, AbstractVector{Symbol}, Function}=:setequal) =
+          cols::ColumnNameMergingType=:setequal) =
     reduce(vcat, dfs; cols=cols)
 
 Base.reduce(::typeof(vcat),
             dfs::Union{AbstractVector{<:AbstractDataFrame}, Tuple{Vararg{AbstractDataFrame}}};
-            cols::Union{Symbol, AbstractVector{Symbol}, Function}=:setequal) =
+            cols::ColumnNameMergingType=:setequal) =
     _vcat([df for df in dfs if ncol(df) != 0]; cols=cols)
 
 function _inferheadernames(dfs::AbstractVector{<:AbstractDataFrame};
-    cols::Union{Symbol, AbstractVector{Symbol}, Function}=:setequal)
+    cols::ColumnNameMergingType=:setequal)
     # Array of all headers
     allheaders = map(names, dfs)
     # Array of unique headers across all data frames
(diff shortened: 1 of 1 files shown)


bkamins:
Thank you for working on it. As noted earlier I would recommend to write down the target API specification for `vcat`, `push!` and `append!` (similarly to what you did in #2032 but there you were commenting about internal design and here we should a specification of how functions are intended to work). Actually we might add a separate section for `vcat`, `push!` and `append!` to the documentation like we did for indexing.

rapus95:
I don't intend to change the API currently, though, I want to make the internals more capable (as stated in #2032). If you suggest some API change to "unleash" the full capabilities to the public, I'm open for suggestions. Though, I might first get the implementation done (better than later finding it unsolvable to be implemented performant) and already having a new API.

bkamins:
I have quickly went through the PR and it seems you are touching `vcat` only (which is OK), but you are changing the API of `vcat` in it (which is also fine in general, but you say that you do not do it so I am not fully sure what is the intention).

If you intend to change only the internals and do not touch the API then it is fine - I guess then the starting point for a review is that CI should pass the old tests.

rapus95:
Ah indeed, though, for now, much will change. When `vcat` is done I will try to generalize the helper functions in order to expand their usage onto `push!` and `append!`. At the end API I'll return to previous API if wanted (for now they just show, what the function would be capable of)

bkamins:
OK.
Please note that from my side it is easiest to review PRs that are either documented in detail (so we know what we want in the end - this is most often the case for large changes that cannot be accomplished in a single PR) or self-contained (i.e. providing some functionality end-to-end that can be checked - this is most often the case for small PRs).

rapus95:
I'll try to split up the whole PR into small logically separated parts where each should be verifiable in a simple fashion and the whole verification comes by modus ponens :D

rapus95:
I've found a single change (that might be breaking but plausible in my opinion):
0-row but non-0-width dataframes are taken into account when designing the new column to the extent that the actual vcat/append! etc fail if the type of a 0-length column cannot be converted into the inferred type.
Tl;Dr those functions can fail even when appending a 0-length dataframe. I find that plausible because supplying a 0-length dataframe wouldn't be useful at all if you didn't want to influence the new table. Most useful case would be to "prepare" a typestable (in regards to not changing column types) appending of more data. You could then prepare that by merging with a 0-row signature ahead of time. Then we can guarantuee the following append to succeed if the supplied types had been supplied before.

bkamins:
I cannot reproduce what you are describing:
```
julia> using DataFrames

julia> df1 = DataFrame(a=["1", "2"])
2×1 DataFrame
│ Row │ a      │
│     │ String │
├─────┼────────┤
│ 1   │ 1      │
│ 2   │ 2      │

julia> df2 = DataFrame(a=Int[])
0×1 DataFrame


julia> vcat(df1, df2)
2×1 DataFrame
│ Row │ a   │
│     │ Any │
├─────┼─────┤
│ 1   │ 1   │
│ 2   │ 2   │

julia> append!(df1, df2)
2×1 DataFrame
│ Row │ a      │
│     │ String │
├─────┼────────┤
│ 1   │ 1      │
│ 2   │ 2      │
```
(and there is no error in either of the cases - the behavior is different as `vcat` does type promotion and `append!` does not, but this is a separate issue).

Can you please supply example codes with what you had in mind (as supposedly this is something different). Showing: current behavior and the proposed behavior? Thank you.

rapus95:
I've resolved it. There shouldn't be any difference. 😀

bkamins:
This should be closed when #2152 is merged. We can open a new PR for post 1.0 functionality.

bkamins:
OK we have merged #2152. If something more would to be added please open a new PR.

PR #2037, opened 2019-12-07
Add upper bounds to versions in Project.toml

nalimilan:


bkamins:
@nalimilan - I have pushed a fix to your PR that caused the CI to fail.

It was related to LaTeXStrings.jl. I removed it from tests. The package is not actively maintained and it also caused recently problems on nightly. I think we do not need to test DataFrames.jl against this package.

PR #2045, opened 2019-12-08
Remove the dependency on `Compat.jl`

DilumAluthge:
It turns out that the dependency on `Compat.jl` is not necessary.

This pull request removes the dependency on `Compat.jl`.

cc: @nalimilan


nalimilan:
Did you close this because we actually need Compat?

DilumAluthge:
Yeah :( I guess hasproperty is not available in 1.0. So unless you drop support for 1.0, you’ll need to keep Compat

PR #2046, opened 2019-12-08
Index GroupedDataFrame with array of keys

jlumpe:
Use array of keys (`GroupKey`/`Tuple`/`NamedTuple`) in `getindex(::GroupedDataFrame, ...)`.

I also fixed a couple of issues remaining in my last PR: a test for `IndexStyle(::GroupKeys)` and added some missing bindings to the documentation.

### DIFF
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -212,8 +212,6 @@ performance on par with integer indexing.
 The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
 
 * `gd[i::Integer]` -> Get the `i`th group.
-* `gd[a::AbstractArray{<:Integer}]` -> Get a reduced `GroupedDataFrame` containing only the groups
-  with indices in `a`.
 * `gd[key::NamedTuple]` -> Get the group corresponding to the given values of the
   grouping columns. The fields of the `NamedTuple` must match the grouping columns
   columns passed to [`groupby`](@ref) (including order).
@@ -223,3 +221,12 @@ The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
 * `gd[key::GroupKey]` -> Get the group corresponding to the [`GroupKey`](@ref)
   `key` (one of the elements of the vector returned by [`keys(::GroupedDataFrame)`](@ref)).
   This should be nearly as fast as integer indexing.
+* `gd[a::AbstractVector]` -> Select multiple groups and return them in a new
+  `GroupedDataFrame` object. Groups may be selected by integer position using an
+  array of `Integer`s or `Bool`s, similar to a standard array. Alternatively the
+  array may contain keys of any of the types supported for dictionary-like
+  indexing (`GroupKey`, `Tuple`, or `NamedTuple`). Selected groups must be
+  unique, and different types of indices cannot be mixed.
+* `gd[n::Not]` -> Any of the above types wrapped in `Not`. The result
+   will be a new `GroupedDataFrame` containing all groups in `gd` *not* selected
+   by the wrapped index.
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -255,6 +255,12 @@ end
 # Non-standard indexing
 #
 
+# Non-standard indexing relies on converting to integer indices first
+# The full version (to_indices) is required rather than to_index even though
+# GroupedDataFrame behaves as a 1D array due to the behavior of Colon and Not.
+# Note that this behavior would be the default if it was <:AbstractArray
+Base.getindex(gd::GroupedDataFrame, idx...) = getindex(gd, Base.to_indices(gd, idx)...)
+
 # The allowed key types for dictionary-like indexing
 const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple}
 # All allowed scalar index types
@@ -280,11 +286,39 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
-# Convert array of indices to integer array
-Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{<:GroupKeyTypes}) = [Base.to_index(gd, k) for k in idxs]
+# Array of (possibly non-standard) indices
+function Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T}
+    # A concrete eltype which is <: GroupIndexTypes, don't need to check
+    isconcretetype(T) && T <: GroupIndexTypes && return [Base.to_index(gd, i) for i in idxs]
 
-# Non-standard indexing relies on converting to integer indices first
-Base.getindex(gd::GroupedDataFrame, idx) = gd[Base.to_index(gd, idx)]
+    # Edge case - array is empty
+    isempty(idxs) && return Int[]
+
+    # Infer eltype based on type of first index
+    idx1 = idxs[1]
+    E1 = typeof(idx1)
+
+    E = if E1 <: Integer && E1 !== Bool
+        Integer
+    elseif E1 <: GroupKey
+        GroupKey
+    elseif E1 <: Tuple
+        Tuple
+    elseif E1 <: NamedTuple
+        NamedTuple
+    else
+        throw(ArgumentError("Invalid index: $idx1 of type $E1"))
+    end
+
+    ints = zeros(Int, length(idxs))
+    for (i, idx) in enumerate(idxs)
+        idx isa GroupIndexTypes || throw(ArgumentError("Invalid index: $idx of type $(typeof(idx))"))
+        idx isa E || throw(ArgumentError("Mixed index types in array not allowed"))
+        ints[i] = Base.to_index(gd, idx)
+    end
+
+    return ints
+end
 
 
 #
@@ -300,14 +334,13 @@ function Base.to_indices(gd::GroupedDataFrame, (idx,)::Tuple{<:Not})
     return (idxs,)
 end
 
-# Nested Not
-# Base.to_index(gd::GroupedDataFrame, idx::Not{Not}) = Base.to_index(gd, idx.skip.skip)
-
 # InvertedIndex wrapping a boolean array
-# Base.getindex(gd::GroupedDataFrame, idx::Not{<:AbstractVector{Bool}}) = gd[.!idx.skip]
-
-# InvertedIndex wrapping a colon (gives empty GroupedDataFrame)
-# Base.getindex(gd::GroupedDataFrame, idx::Not{Colon}) = gd[Int[]]
+# The definition above works but we need to define a specialized method to avoid
+# ambiguity in dispatch
+function Base.to_indices(gd::GroupedDataFrame,
+                         (idx,)::Tuple{Not{<:Union{BitArray{1}, Vector{Bool}}}})
+    (Base.LogicalIndex(.!idx.skip),)
+end
 
 
 #
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -249,27 +249,34 @@ end
 # Non-standard indexing
 #
 
-# Index with GroupKey
-function Base.getindex(gd::GroupedDataFrame, key::GroupKey)
-    gd === parent(key) && return gd[getfield(key, :idx)]
+# The allowed key types for dictionary-like indexing
+const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple}
+
+# Find integer index for dictionary keys:
+function Base.to_index(gd::GroupedDataFrame, key::GroupKey)
+    gd === parent(key) && return getfield(key, :idx)
     throw(ErrorException("Cannot use a GroupKey to index a GroupedDataFrame other than the one it was derived from."))
 end
 
-# Index with tuple
-function Base.getindex(gd::GroupedDataFrame, key::Tuple)
+function Base.to_index(gd::GroupedDataFrame, key::Tuple)
     for i in 1:length(gd)
-        isequal(Tuple(_groupvalues(gd, i)), key) && return gd[i]
+        isequal(Tuple(_groupvalues(gd, i)), key) && return i
     end
     throw(KeyError(key))
 end
 
-# Index with named tuple
-function Base.getindex(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
+function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(N, gd.cols))
         throw(KeyError(key))
     end
-    return gd[Tuple(key)]
+    return Base.to_index(gd, Tuple(key))
 end
+
+# Dictionary-like indexing with single key
+Base.getindex(gd::GroupedDataFrame, key::GroupKeyTypes) = gd[Base.to_index(gd, key)]
+
+# Dictionary-like indexing with multiple keys
+Base.getindex(gd::GroupedDataFrame, idxs::AbstractVector{<:GroupKeyTypes}) = gd[[Base.to_index(gd, k) for k in idxs]]
 
 
 #
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -288,8 +288,8 @@ end
 
 # Array of (possibly non-standard) indices
 function Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T}
-    # A concrete eltype which is <: GroupIndexTypes, don't need to check
-    if isconcretetype(T) && T <: GroupIndexTypes
+    # A concrete eltype which is <: GroupKeyTypes, don't need to check
+    if isconcretetype(T) && T <: GroupKeyTypes
         return [Base.to_index(gd, i) for i in idxs]
     end
 
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -312,7 +312,7 @@ function Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T}
 
     ints = zeros(Int, length(idxs))
     for (i, idx) in enumerate(idxs)
-        idx isa GroupIndexTypes || throw(ArgumentError("Invalid index: $idx of type $(typeof(idx))"))
+        (idx isa GroupIndexTypes && !(idx isa Bool)) || throw(ArgumentError("Invalid index: $idx of type $(typeof(idx))"))
         idx isa E || throw(ArgumentError("Mixed index types in array not allowed"))
         ints[i] = Base.to_index(gd, idx)
     end
@@ -335,10 +335,14 @@ function Base.to_indices(gd::GroupedDataFrame, (idx,)::Tuple{<:Not})
 end
 
 # InvertedIndex wrapping a boolean array
-# The definition above works but we need to define a specialized method to avoid
+# The definition above works but we need to define specialized methods to avoid
 # ambiguity in dispatch
 function Base.to_indices(gd::GroupedDataFrame,
                          (idx,)::Tuple{Not{<:Union{BitArray{1}, Vector{Bool}}}})
+    (Base.LogicalIndex(.!idx.skip),)
+end
+function Base.to_indices(gd::GroupedDataFrame,
+                         (idx,)::Tuple{Not{<:AbstractVector{Bool}}})
     (Base.LogicalIndex(.!idx.skip),)
 end
 
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -133,6 +133,12 @@ Compat.lastindex(gd::GroupedDataFrame) = gd.ngroups
 Base.first(gd::GroupedDataFrame) = gd[1]
 Base.last(gd::GroupedDataFrame) = gd[end]
 
+# These seem to have to be defined for some to_indices() logic to work, as long
+# as GroupedDataFrame is not <: AbstractArray
+Base.IndexStyle(::Type{<:GroupedDataFrame}) = IndexLinear()
+Base.IndexStyle(::GroupedDataFrame) = IndexLinear()
+Base.keys(::IndexLinear, gd::GroupedDataFrame) = Base.OneTo(length(gd))
+
 # Single integer indexing
 Base.getindex(gd::GroupedDataFrame, idx::Integer) =
     view(gd.parent, gd.idx[gd.starts[idx]:gd.ends[idx]], :)
@@ -251,6 +257,8 @@ end
 
 # The allowed key types for dictionary-like indexing
 const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple}
+# All allowed scalar index types
+const GroupIndexTypes = Union{Integer, GroupKeyTypes}
 
 # Find integer index for dictionary keys:
 function Base.to_index(gd::GroupedDataFrame, key::GroupKey)
@@ -272,11 +280,34 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
-# Dictionary-like indexing with single key
-Base.getindex(gd::GroupedDataFrame, key::GroupKeyTypes) = gd[Base.to_index(gd, key)]
+# Convert array of indices to integer array
+Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{<:GroupKeyTypes}) = [Base.to_index(gd, k) for k in idxs]
 
-# Dictionary-like indexing with multiple keys
-Base.getindex(gd::GroupedDataFrame, idxs::AbstractVector{<:GroupKeyTypes}) = gd[[Base.to_index(gd, k) for k in idxs]]
+# Non-standard indexing relies on converting to integer indices first
+Base.getindex(gd::GroupedDataFrame, idx) = gd[Base.to_index(gd, idx)]
+
+
+#
+# Indexing with Not/InvertedIndex
+#
+
+# InvertedIndex wrapping any other valid index type
+# to_indices() is needed here rather than to_index() for some reason, but we
+# only need to accept one-tuples.
+function Base.to_indices(gd::GroupedDataFrame, (idx,)::Tuple{<:Not})
+    (skip_idx,) = Base.to_indices(gd, (idx.skip,))
+    idxs = Base.OneTo(length(gd))[Not(skip_idx)]
+    return (idxs,)
+end
+
+# Nested Not
+# Base.to_index(gd::GroupedDataFrame, idx::Not{Not}) = Base.to_index(gd, idx.skip.skip)
+
+# InvertedIndex wrapping a boolean array
+# Base.getindex(gd::GroupedDataFrame, idx::Not{<:AbstractVector{Bool}}) = gd[.!idx.skip]
+
+# InvertedIndex wrapping a colon (gives empty GroupedDataFrame)
+# Base.getindex(gd::GroupedDataFrame, idx::Not{Colon}) = gd[Int[]]
 
 
 #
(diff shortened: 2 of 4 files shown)


docs/src/lib/indexing.md
```diff
@@ -212,8 +212,9 @@ performance on par with integer indexing.
 The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
 
 * `gd[i::Integer]` -> Get the `i`th group.
-* `gd[a::AbstractArray{<:Integer}]` -> Get a reduced `GroupedDataFrame` containing only the groups
-  with indices in `a`.
+* `gd[a::AbstractArray{<:Union{Integer, Boolean}}]` -> Use array-style indexing
```
bkamins: `Boolean` type does not exist. If you meant `Bool` it is a subtype of `Integer` so no need to mention it. However, we discussed of adding `Not` support in this PR.

docs/src/lib/indexing.md
```diff
@@ -223,3 +224,6 @@ The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
 * `gd[key::GroupKey]` -> Get the group corresponding to the [`GroupKey`](@ref)
   `key` (one of the elements of the vector returned by [`keys(::GroupedDataFrame)`](@ref)).
   This should be nearly as fast as integer indexing.
+*  `gd[keys::AbstractArray{<:Union{GroupKey, Tuple, NamedTuple}}]` ->
```
bkamins: I think it can be merged with the `AbstractArray` selector above.

src/groupeddataframe/grouping.jl
```diff
@@ -412,28 +412,38 @@ true
...
-function Base.getindex(gd::GroupedDataFrame, key::GroupKey)
-    gd === parent(key) && return gd[getfield(key, :idx)]
-    throw(ErrorException("Cannot use a GroupKey to index a GroupedDataFrame other than the one it was derived from."))
+
+# The allowed key types for dictionary-like indexing
+const AbstractGroupKey = Union{GroupKey, Tuple, NamedTuple}
+
+
+# Find the integer index of a group given its key
+function _findgroup(gd::GroupedDataFrame, key::GroupKey, strict::Val{S}=Val{true}()) where {S}
```
bkamins: Why do you use `strict` parameter here? It is not used later. I would just drop it and assume that we are always strict.
Also with constant propagation the `Val` trick in practice is not needed now (it was needed in early versions of Julia that had much less efficient compiler).

test/grouping.jl
```diff
@@ -1332,6 +1363,7 @@ end
     @test length(gdkeys) == length(expected)
     @test size(gdkeys) == size(expected)
     @test eltype(gdkeys) == DataFrames.GroupKey{typeof(gd)}
+    @test IndexStyle(gdkeys) == IndexLinear()
```
bkamins: this is already tested four lines above

jlumpe: Not sure why I thought I hadn't put that in yet.

bkamins: @nalimilan did 😄

jlumpe:
Resolved issues and added support for `InvertedIndex`.

src/groupeddataframe/grouping.jl
```diff
@@ -412,28 +412,53 @@ true
 """
 Base.keys(gd::GroupedDataFrame) = GroupKeys(gd)
 
-# Index with GroupKey
-function Base.getindex(gd::GroupedDataFrame, key::GroupKey)
-    gd === parent(key) && return gd[getfield(key, :idx)]
+
+# The allowed key types for dictionary-like indexing
+const AbstractGroupKey = Union{GroupKey, Tuple, NamedTuple}
```
bkamins: maybe better call it `GroupKeyUnion`, as "abstract" suggests abstract type.

@nalimilan - any opinion?

docs/src/lib/indexing.md
```diff
@@ -223,3 +221,12 @@ The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
 * `gd[key::GroupKey]` -> Get the group corresponding to the [`GroupKey`](@ref)
   `key` (one of the elements of the vector returned by [`keys(::GroupedDataFrame)`](@ref)).
   This should be nearly as fast as integer indexing.
+* `gd[a::AbstractArray]` -> Select multiple groups and return them in a new
+  `GroupedDataFrame` object. Groups may be selected by integer position using an
+  array of `Integer`s or `Bool`s, similar to a standard array. Alternatively the
+  array may contain keys of any of the types supported for dictionary-like
+  indexing (`GroupKey`, `Tuple`, or `NamedTuple`). Selected groups must be
+  unique.
+* `gd[n::InvertedIndex]` -> Any of the above types wrapped in
```
bkamins: maybe use `Not` here and everywhere in the code instead of `InvertedIndex` as this is what users use (so it will be easier to read the code in case they need to).

src/groupeddataframe/grouping.jl
```diff
@@ -412,28 +412,53 @@ true
...
+Base.getindex(gd::GroupedDataFrame, key::AbstractGroupKey) = gd[_findgroup(gd, key)]
+
+# Array indexing with dictionary-like keys
+Base.getindex(gd::GroupedDataFrame, idxs::AbstractArray{T}) where {T<:AbstractGroupKey} = gd[[_findgroup(gd, k) for k in idxs]]
+
+# InvertedIndex
+function Base.getindex(gd::GroupedDataFrame, idx::InvertedIndex{T}) where T
+    selected = trues(length(gd))
+    if T <: AbstractArray
+        selected[[_findgroup(gd, k) for k in idx.skip]] .= false
```
bkamins: not a big thing, but maybe we can do it in a loop to avoid the allocation of the inner array?

bkamins:
Looks good apart from the comment I left (and maybe @nalimilan will want to add something 😄)

bkamins:
Also please make sure you test:
* `Not(:)` (which I think will not work correctly now)
* `end` and `Not(end)` (they should work I think, but have not tested it)
* `:` (it is tested as far as I recall but can you please check it)

Thank you.

bkamins:
I have thought over your proposal and I think we need to be more strict in a way we introduce allowed indexing. I will leave the comments in relevant methods.

src/groupeddataframe/grouping.jl
```diff
@@ -412,28 +412,53 @@ true
...
     end
-    return gd[Tuple(key)]
+    return _findgroup(gd, values(key))
 end
 
+# Dictionary-like indexing
+Base.getindex(gd::GroupedDataFrame, key::AbstractGroupKey) = gd[_findgroup(gd, key)]
+
+# Array indexing with dictionary-like keys
+Base.getindex(gd::GroupedDataFrame, idxs::AbstractArray{T}) where {T<:AbstractGroupKey} = gd[[_findgroup(gd, k) for k in idxs]]
```
bkamins: 1. This should be fixed as it allows mixing different styles of grouping keys (i.e. mixing named tuples, tuples, and `GroupKey` in a single vector).
2. Also here - and in the whole of this file (also in places you do not touch ) for indexing `AbstractArray` should be changed to `AbstractVector` - as we disallow changing the dimensionality of objects we use here.
3. All `getindex` methods for `GroupedDataFrame` should be moved into the same place in the file as otherwise it will be hard to maintain.

src/groupeddataframe/grouping.jl
bkamins: Also the design here is not optimal because:
1. it allows mixing integers, tuples, named tuples and `GroupKey`s in `idx.skip` that I think we should not allow
2. having `Bool` value of `true` in `idx.skip` (either bare or in a vector) will lead to wrong results (first index will be selected while it should error)

Like in the case of standard `getindex` above we should make sure that the values passed have homogeneous type, an if this type is integer, we should check if no values is `Bool` there.

A typical pattern to handle this (the `Not` case and the standard case above) would be:
1. have "fast" methods in case `eltype` of passed index is well specified (like the case you handle below with eltype being `Bool`)
2. have "catch all" slower method (that would be called only in obscure cases like `Any[1, 2, 3]`) that would check homogeneity of what was passed to it (and make sure no `Bool` slipped in) and then perform the selection similarly to what you do now.

docs/src/lib/indexing.md
jlumpe: Is `Not` supported with a `@ref` link?

src/groupeddataframe/grouping.jl
jlumpe: That makes sense, it's only used internally anyway.

docs/src/lib/indexing.md
bkamins: No, but neither `InvertedIndex` is - it is a limitation of Documenter.jl AFAICT - you cannot `@ref` to another package.

jlumpe:
Before we go any further with this - I went a little further than your request to group all the indexing code together and just split `grouping.jl`  into multiple files which I think are much more manageable. If I did it in this PR it would be hard to tell what else I changed so I [opened a separate one](#2050). If you want to merge that I can rebase this PR on top of it.

bkamins:
Let us wait for @nalimilan to comment on #2050 then.

docs/src/lib/indexing.md
```diff
@@ -223,3 +221,12 @@ The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
 * `gd[key::GroupKey]` -> Get the group corresponding to the [`GroupKey`](@ref)
   `key` (one of the elements of the vector returned by [`keys(::GroupedDataFrame)`](@ref)).
   This should be nearly as fast as integer indexing.
+* `gd[a::AbstractVector]` -> Select multiple groups and return them in a new
+  `GroupedDataFrame` object. Groups may be selected by integer position using an
+  array of `Integer`s or `Bool`s, similar to a standard array. Alternatively the
+  array may contain keys of any of the types supported for dictionary-like
+  indexing (`GroupKey`, `Tuple`, or `NamedTuple`). Selected groups must be
+  unique, and different types of indices cannot be mixed.
+* `gd[n::Not]` -> Any of the above types wrapped in [`Not`](@ref). The result
```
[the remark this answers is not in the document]
bkamins: I think it will not.

@jlumpe - have you tried building the documentation locally to check?

src/groupeddataframe/grouping.jl
```diff
@@ -412,28 +412,60 @@ true
...
+
+# InvertedIndex wrapping scalar index
+function Base.getindex(gd::GroupedDataFrame, idx::InvertedIndex{T}) where {T<:GroupIndexTypes}
+    selected = trues(length(gd))
+    selected[_findgroup(gd, idx.skip)] = false
+    return gd[selected]
+end
+
+# InvertedIndex wrapping array
+function Base.getindex(gd::GroupedDataFrame, idx::InvertedIndex{T}) where {E<:GroupIndexTypes,T<:AbstractVector{E}}
```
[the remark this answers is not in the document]
jlumpe: Was this syntax not valid in a previous version of Julia? I'm trying to figure out why I've been using the longer version.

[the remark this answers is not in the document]
bkamins: It was valid AFAICT.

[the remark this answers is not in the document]
nalimilan: IIRC it's been introduced in 0.7 or 0.6.

src/groupeddataframe/grouping.jl
```diff
@@ -412,28 +412,60 @@ true
...
+# Dictionary-like indexing
+Base.getindex(gd::GroupedDataFrame, key::GroupKeyTypes) = gd[_findgroup(gd, key)]
+
+# Array indexing with dictionary-like keys
+Base.getindex(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T<:GroupKeyTypes} = gd[[_findgroup(gd, k) for k in idxs]]
+
+# InvertedIndex wrapping scalar index
+function Base.getindex(gd::GroupedDataFrame, idx::InvertedIndex{T}) where {T<:GroupIndexTypes}
+    selected = trues(length(gd))
+    selected[_findgroup(gd, idx.skip)] = false
```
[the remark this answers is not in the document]
jlumpe: I would think this would be faster - the existing `getindex(gd, ::AbstractArray)` works directly with either ints or bools because the argument is just used as an index for `.starts` and `.ends`.  This way is a bit array allocation and an `O(1)` write, `setdiff` allocates an integer array but also has to do an `O(n)` scan through all the indices.

jlumpe:
Updates:
* Rebased onto master with new organization of `src/groupeddataframe/` files.
* `getindex(gd, a::AbstractArray)` narrowed to `getindex(gd, ::AbstractVector)`
* Removed unnecessary `where` statements
* Significantly cleaned up code
* Implemented `gd[Not(:)]` and test
* Tests for GroupedDataFrame `first()`, `last()`, `lastindex()`, `gd[end]`
* Correctly infer element type of arrays with `eltype() == Any` or anything else nonspecific, throw an exception for non-homogeneous arrays (even when all elements are individually valid scalar indices)
  * Applies to `Not(::AbstractArray)` as well

Issues not addressed:
* `Not(end)` isn't syntactically valid, `x[end]` is syntactic sugar for `x[lastindex(x)]` but this isn't available outside of a `:ref` expression.
* `AbstractVector{Any}` full of Bools does not work as an index. I left it out because `AbstractArray` doesn't check for it either.

docs/src/lib/indexing.md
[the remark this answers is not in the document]
jlumpe: You're right, it looks like `Not`/`InvertedIndex` is in a different package. I'll remove the `@ref`s.

bkamins:
Thank you for the update.

> `Not(end)` isn't syntactically valid

It is valid in DataFrames.jl:
```
julia> df = DataFrame(rand(3,4))
3×4 DataFrame
│ Row │ x1        │ x2       │ x3       │ x4        │
│     │ Float64   │ Float64  │ Float64  │ Float64   │
├─────┼───────────┼──────────┼──────────┼───────────┤
│ 1   │ 0.0193821 │ 0.489792 │ 0.649841 │ 0.887786  │
│ 2   │ 0.35907   │ 0.727754 │ 0.392126 │ 0.0994079 │
│ 3   │ 0.315079  │ 0.969064 │ 0.956188 │ 0.9801    │

julia> df[:, Not(end)]
3×3 DataFrame
│ Row │ x1        │ x2       │ x3       │
│     │ Float64   │ Float64  │ Float64  │
├─────┼───────────┼──────────┼──────────┤
│ 1   │ 0.0193821 │ 0.489792 │ 0.649841 │
│ 2   │ 0.35907   │ 0.727754 │ 0.392126 │
│ 3   │ 0.315079  │ 0.969064 │ 0.956188 │

julia> df[Not(end), :]
2×4 DataFrame
│ Row │ x1        │ x2       │ x3       │ x4        │
│     │ Float64   │ Float64  │ Float64  │ Float64   │
├─────┼───────────┼──────────┼──────────┼───────────┤
│ 1   │ 0.0193821 │ 0.489792 │ 0.649841 │ 0.887786  │
│ 2   │ 0.35907   │ 0.727754 │ 0.392126 │ 0.0994079 │
```
and also for normal vectors:
```
julia> x = [1,2,3,4]
4-element Array{Int64,1}:
 1
 2
 3
 4

julia> x[Not(end)]
3-element Array{Int64,1}:
 1
 2
 3
```
I will have a look at your code to check.

> `AbstractVector{Any}` full of Bools does not work as an index.

Yes - this is how it should work

bkamins:
It also seems from your implementation that you do not support `gd[Not(1)]` nor things like `gd[Not(Not(1))]`. Please correct me if I am wrong.

I would rather delegate `Not` (except for non-standard indexing and optimizations you have) to be handled by the mechanics provided by InvertedIndices.jl and then all should work as expected.

Also - as I have commended earlier - I would recommend you to use `Not` not `InvertedIndex` in the code, as this is the style we use everywhere else in the code base (interestingly I have checked that InvertedIndices.jl mixes both of them in different methods, but I think it is better to stick to one form everywhere).

src/groupeddataframe/grouping.jl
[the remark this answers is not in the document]
nalimilan: My point is that in the end `getindex` on vectors will actually do an `O(n)` pass anyway to count the number of `true` entries (and probably twice, for `starts` and `ends`). Since you know that value, it should be better to directly allocate an array of integer indices. Maybe try something like `n = _findgroup(gd, idx.skip); vcat(1:n, n+2:length(gd))`?

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -331,28 +300,48 @@ true
...
+Base.first(gd::GroupedDataFrame) = gd[1]
+Base.last(gd::GroupedDataFrame) = gd[end]
+
+# Single integer indexing (basis for all other scalar methods)
+Base.getindex(gd::GroupedDataFrame, idx::Integer) =
+    view(gd.parent, gd.idx[gd.starts[idx]:gd.ends[idx]], :)
+
+# Find the integer index of a group given any valid scalar key type
+# (Integer case already defined by default, does not accept Bool)
+function Base.to_index(gd::GroupedDataFrame, key::GroupKey)
```
nalimilan: Is it really useful to override `Base.to_index` for this? Otherwise, better use a different name to avoid relying on Base internals.

src/groupeddataframe/grouping.jl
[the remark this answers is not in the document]
nalimilan: Ah, but that will only work for single indices. I guess it's OK to keep a logical index then. (Ideally we would have an optimized `setdiff` method for ranges, but...)

src/groupeddataframe/groupeddataframe.jl
jlumpe: It's definitely not necessary, but this is the intended function of `Base.to_index`. I don't think it's actually used by any of the default indexing machinery if the first argument isn't an `AbstractArray`.

bkamins: As a side note: InvertedIndices.jl uses `Base.to_index`.

src/groupeddataframe/grouping.jl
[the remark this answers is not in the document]
jlumpe: I see what you mean. Yes I think that's probably the best way of doing it.

jlumpe:
You're right about `end`, I'm not sure why I thought you couldn't use it in non-trivial expressions but it looks like you can do pretty much whatever you want with it in between the brackets. I have added this test.

jlumpe:
For everything involving `InvertedIndex.jl`'s built-in behavior and `Base.to_index` and related, it all seems to only really work on `AbstractArray`s. Is there a reason `GroupedDataFrame` doesn't subclass it (I think this discussion already exists but I can't find it)? If it did we'd just need definitions for `getindex` with integers and arrays of integers and then `Base.to_index` for scalar keys or arrays of them, and we'd get everything else.

bkamins:
> I think this discussion already exists but I can't find it

I think the main reason is that this decision is very committing so as long as it can be postponed we try not to do it. The major issue is that Julia does not allow multiple inheritance currently.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -410,3 +399,95 @@ function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple}, default)
...
+    elseif T1 <: NamedTuple
+        NamedTuple
+    else
+        throw(ArgumentError("Invalid index: $idx1 of type $T1"))
+    end
+
+    ints = zeros(Int, length(idxs))
+    for (i, idx) in enumerate(idxs)
+        idx isa GroupIndexTypes || throw(ArgumentError("Invalid index: $idx1 of type $T1"))
+        idx isa T || throw(ArgumentError("Mixed index types in array not allowed"))
```
bkamins: mixing integer types except `Bool` is allowed in Base, so we should also allow it here:
```
julia> x = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> x[Any[1,0x2]]
2-element Array{Int64,1}:
 1
 2
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -410,3 +399,95 @@ function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple}, default)
...
+    selected = trues(length(gd))
+    selected[_to_index_array(gd, idx.skip)] .= false
+    return gd[selected]
+end
+
+# InvertedIndex wrapping a boolean array
+Base.getindex(gd::GroupedDataFrame, idx::InvertedIndex{<:AbstractVector{Bool}}) = gd[.!idx.skip]
+
+# InvertedIndex wrapping a colon (gives empty GroupedDataFrame)
+Base.getindex(gd::GroupedDataFrame, idx::InvertedIndex{Colon}) = gd[Int[]]
```
bkamins: the fallback method for `Not` seems to be still missing.

src/groupeddataframe/groupeddataframe.jl
jlumpe: Using `Base.to_indices` in the default `getindex` method, this calls `to_index` when it needs to.

jlumpe:
* Handle arrays of non-concrete type, check for homogeneity
* Indexing with `Not` now uses the `InvertedIndices.jl` machinery. Implements `to_indices(gd, (idx,)::Tuple{Not})` instead of `getindex(gd, ::Not)`. Recurses on `ints = Base.to_indices(gd, (idx.skip,))` and then gets inverted integer indices with `to_indices(Base.OneTo(length(gd)), ints)`. This handles all cases, but one extra definition is needed to avoid a method ambiguity.
* Tests for `Not{Not}`
* Test for mixed non-`Bool` `Integer` arrays

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +251,102 @@ Base.IndexStyle(::Type{<:GroupKeys}) = IndexLinear()
...
+    (skip_idx,) = Base.to_indices(gd, (idx.skip,))
+    idxs = Base.OneTo(length(gd))[Not(skip_idx)]
+    return (idxs,)
+end
+
+# InvertedIndex wrapping a boolean array
+# The definition above works but we need to define a specialized method to avoid
+# ambiguity in dispatch
+function Base.to_indices(gd::GroupedDataFrame,
+                         (idx,)::Tuple{Not{<:Union{BitArray{1}, Vector{Bool}}}})
```
bkamins: Why not `Tuple{Not{<:AbstractVector{Bool}}`? Just making sure.

To be sure that all works OK, can you add a test for a `GroupedDataFrame` with two groups and indexed by `false:true`, `Not(false:true)`?

bkamins:
Looks good. Thank you. I just left one comment (with a request for an additional test).

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +251,102 @@ Base.IndexStyle(::Type{<:GroupKeys}) = IndexLinear()
...
+    elseif E1 <: NamedTuple
+        NamedTuple
+    else
+        throw(ArgumentError("Invalid index: $idx1 of type $E1"))
+    end
+
+    ints = zeros(Int, length(idxs))
+    for (i, idx) in enumerate(idxs)
+        idx isa GroupIndexTypes || throw(ArgumentError("Invalid index: $idx of type $(typeof(idx))"))
+        idx isa E || throw(ArgumentError("Mixed index types in array not allowed"))
```
bkamins: one more thing. There should be an additional test here `idx isa Bool && throw(ArgumentError`...
as it seems that currently indexing by `Any[2, true]` would go through your function and work the same as `[2,1]` (can you please check it and add to the tests)?

src/groupeddataframe/groupeddataframe.jl
jlumpe: This is covered by the new method.

src/groupeddataframe/groupeddataframe.jl
jlumpe: Turned out I needed them both, `Invertedindices.jl` defines both of these so they both cause ambiguities. Added the test.

jlumpe:
Remaining issues should be resolved.

bkamins:
@nalimilan - please have a look at it when you have time so that we can merge it if it is OK.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +251,106 @@ Base.IndexStyle(::Type{<:GroupKeys}) = IndexLinear()
...
+    idxs = Base.OneTo(length(gd))[Not(skip_idx)]
+    return (idxs,)
+end
+
+# InvertedIndex wrapping a boolean array
+# The definition above works but we need to define specialized methods to avoid
+# ambiguity in dispatch
+function Base.to_indices(gd::GroupedDataFrame,
+                         (idx,)::Tuple{Not{<:Union{BitArray{1}, Vector{Bool}}}})
+    (Base.LogicalIndex(.!idx.skip),)
```
nalimilan: It would probably be more efficient to compute the integer indices directly, using `findall(!, idx.skip)`. Could also use the one-liner syntax.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +251,106 @@ Base.IndexStyle(::Type{<:GroupKeys}) = IndexLinear()
...
+    # Edge case - array is empty
+    isempty(idxs) && return Int[]
+
+    # Infer eltype based on type of first index
+    idx1 = idxs[1]
+    E1 = typeof(idx1)
+
+    E = if E1 <: Integer && E1 !== Bool
+        Integer
+    elseif E1 <: GroupKey
```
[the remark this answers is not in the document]
bkamins: In Base integer indices are allowed to have mixed types as long as they are not `Bool` if mixed types are present.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +251,106 @@ Base.IndexStyle(::Type{<:GroupKeys}) = IndexLinear()
...
+    return ints
+end
+
+
+#
+# Indexing with Not/InvertedIndex
+#
+
+# InvertedIndex wrapping any other valid index type
+# to_indices() is needed here rather than to_index() for some reason, but we
```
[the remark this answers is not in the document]
bkamins: Agreed, good point. We have not used `to_index` nor `to_indices` previously in DataFrames.jl. I am OK to start using it, but @nalimilan is right, that then we must understand exactly the logic how they should be used properly.

src/groupeddataframe/groupeddataframe.jl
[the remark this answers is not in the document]
nalimilan: OK. So there's no simpler way to achieve this?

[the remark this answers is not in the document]
bkamins: We do not have to be super elegant here (and roughly it is a minimal way to implement it - I have not thought much if it can be improved).
This method is just for completeness, as it will be hardly ever called (it will be used only if the users passes a vector that has an abstract eltype).

src/groupeddataframe/groupeddataframe.jl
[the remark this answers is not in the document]
jlumpe: Figured it out, `to_indices(a, t)` is what is used by `getindex(a::AbstractArray, t...)` but the default implementation just calls `to_index(a, i)` for each `i` in `t`. This is needed to override the `to_indices` methods defined in `InvertedIndices.jl`.

jlumpe:
Resolved rest of @nalimilan's suggestions.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +251,111 @@ Base.IndexStyle(::Type{<:GroupKeys}) = IndexLinear()
...
+    if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(N, gd.cols))
+        throw(KeyError(key))
+    end
+    return Base.to_index(gd, Tuple(key))
+end
+
+# Array of (possibly non-standard) indices
+function Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T}
+    # A concrete eltype which is <: GroupIndexTypes, don't need to check
+    if isconcretetype(T) && T <: GroupIndexTypes
```
bkamins: Just to clarify - it should be enough to use `GroupKeyTypes` here as if `T` is concrete and is `Integer` this function should never be called - right?

I am asking because below you are iterating `idxs`, and if `idxs` would happen to be `AbstractVector{Bool}` the line https://github.com/JuliaData/DataFrames.jl/pull/2046/files#diff-6349421a054bb7e74b7f27ae86304cbaR293 would fail.

bkamins:
Looks good. I just left one minimal comment (we should either narrow the condition there or special case when `T` is `Bool`).

@nalimilan - apart from this I would merge this. Thank you!

src/groupeddataframe/groupeddataframe.jl
jlumpe: Yes, correct.

nalimilan:
Thanks @jlumpe!

PR #2047, opened 2019-12-08
Avoid computing group indices when possible

nalimilan:
This can make simple aggregation functions more than twice faster.

Some benchmarking shows no regression from current state even when a custom function (which doesn't benefit from optimization) is used:
```julia
using DataFrames, PooledArrays, BenchmarkTools
n = 100_000;
for k in (10, 10_000)
    global df = DataFrame(x=PooledArray(rand(1:k, n)), y=rand(n));
    println("groupby with $k groups")
    @btime groupby(df, :x);
    global gd = groupby(df, :x);

    println("sum with $k groups")
    @btime by(df, :x, :y => sum);
    @btime combine(gd2, :y => sum) setup=(gd2=deepcopy(gd));

    println("sum+length with $k groups")
    @btime by(df, :x, :y => sum, :y => length);
    @btime combine(gd2, :y => sum, :y => length) setup=(gd2=deepcopy(gd));

    println("x -> x[1] with $k groups")
    @btime by(df, :x, :y => (x -> x[1]));
    @btime combine(gd2, :y => (x -> x[1])) setup=(gd2=deepcopy(gd));
end

# Before
groupby with 10 groups
  463.561 μs (49 allocations: 1.53 MiB)
sum with 10 groups
  612.750 μs (188 allocations: 1.54 MiB)
  97.593 μs (139 allocations: 10.28 KiB)
sum+length with 10 groups
  599.455 μs (220 allocations: 1.54 MiB)
  106.351 μs (171 allocations: 12.14 KiB)
x -> x[1] with 10 groups
  702.722 μs (269 allocations: 2.30 MiB)
  153.622 μs (220 allocations: 793.52 KiB) # Indices have already been computed by first run
groupby with 10000 groups
  1.505 ms (52 allocations: 1.77 MiB)
sum with 10000 groups
  1.881 ms (210 allocations: 2.35 MiB)
  382.246 μs (158 allocations: 596.09 KiB)
sum+length with 10000 groups
  2.099 ms (253 allocations: 2.66 MiB)
  475.889 μs (201 allocations: 910.25 KiB)
x -> x[1] with 10000 groups
  4.099 ms (89172 allocations: 5.12 MiB)
  2.436 ms (89120 allocations: 3.35 MiB) # Indices have already been computed by first run

# After
groupby with 10 groups
  152.112 μs (44 allocations: 784.13 KiB)
sum with 10 groups
  271.348 μs (173 allocations: 794.02 KiB)
  92.858 μs (129 allocations: 9.89 KiB)
sum+length with 10 groups
  362.215 μs (205 allocations: 795.77 KiB)
  180.631 μs (161 allocations: 11.64 KiB) # Counts groups again every time
x -> x[1] with 10 groups
  689.666 μs (264 allocations: 2.30 MiB)
  152.599 μs (215 allocations: 793.34 KiB) # Indices have already been computed by first run
groupby with 10000 groups
  235.042 μs (45 allocations: 872.06 KiB)
sum with 10000 groups
  780.547 μs (193 allocations: 1.44 MiB)
  524.955 μs (148 allocations: 605.55 KiB)
sum+length with 10000 groups
  997.213 μs (235 allocations: 1.67 MiB)
  710.533 μs (190 allocations: 841.55 KiB) # Counts groups again every time
x -> x[1] with 10000 groups
  4.060 ms (89186 allocations: 5.12 MiB)
  2.540 ms (89133 allocations: 3.35 MiB) # Indices have already been computed by first run
```

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -244,6 +244,7 @@ end
 
 function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Symbol},
                   colkey::Int, value::Int, keycol, valuecol, g, renamecols)
+    g.idx === nothing && compute_indices!(g)
     groupidxs = [g.idx[g.starts[i]:g.ends[i]] for i in 1:length(g.starts)]
     rowkey = zeros(Int, size(df, 1))
     for i in 1:length(groupidxs)
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -244,13 +244,13 @@ end
 
 function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Symbol},
                   colkey::Int, value::Int, keycol, valuecol, g, renamecols)
-    g.idx === nothing && compute_indices!(g)
-    groupidxs = [g.idx[g.starts[i]:g.ends[i]] for i in 1:length(g.starts)]
+    idx, starts, ends = g.idx, g.starts, g.ends
+    groupidxs = [idx[starts[i]:ends[i]] for i in 1:length(starts)]
     rowkey = zeros(Int, size(df, 1))
     for i in 1:length(groupidxs)
         rowkey[groupidxs[i]] .= i
     end
-    df1 = df[g.idx[g.starts], g.cols]
+    df1 = df[idx[starts], g.cols]
     Nrow = length(g)
     Ncol = length(levels(keycol))
     unstacked_val = [similar_missing(valuecol, Nrow) for i in 1:Ncol]
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -263,7 +263,7 @@ function compute_indices(groups::AbstractVector{<:Integer}, ngroups::Integer)
     end
     stops .-= 1
 
-    # drop group 1 which contains rows with missings in grouping columns (if any)
+    # group 1 corresponds to missings to drop (if any)
     popfirst!(starts)
     popfirst!(stops)
 
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -180,6 +180,8 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
     end
 
     seen = fill(false, ngroups)
+    # Compute vector mapping missing to -1 if skipmissing=true,
+    # and sorting CategoricalArray levels (since it's cheap)
     refmaps = map(cols) do col
         nlevs = nlevels(col)
         if col isa CategoricalVector
@@ -191,10 +193,11 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
             refmap[2:end] .= CategoricalArrays.order(col.pool) .- 1
         else # PooledVector
             # First value in refmap is never used
+            # (corresponds to ref 0 which is only used by CategoricalArray)
             refmap = collect(-1:nlevs-1)
-            if eltype(col) >: Missing
+            if eltype(col) >: Missing && skipmissing
                 missingind = get(col.invpool, missing, 0)
-                if skipmissing && missingind > 0
+                if missingind > 0
                     refmap[missingind+1] = -1
                     refmap[missingind+2:end] .-= 1
                 end
@@ -210,6 +213,7 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
         end
         vals = map((m, r, s) -> m[r+1] * s, refmaps, refs, strides)
         j = sum(vals) + 1
+        # x < 0 happens with -1 in refmap, which corresponds to missing
         if skipmissing && any(x -> x < 0, vals)
             j = 0
         else
@@ -226,8 +230,8 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
             remap[i] = ngroups
         end
         @inbounds for i in eachindex(groups)
-            g_ix = groups[i]
-            groups[i] = g_ix > 0 ? remap[g_ix] : 0
+            gix = groups[i]
+            groups[i] = gix > 0 ? remap[gix] : 0
         end
         # To catch potential bugs inducing unnecessary computations
         @assert oldngroups != ngroups
@@ -245,8 +249,8 @@ end
 function compute_indices(groups::AbstractVector{<:Integer}, ngroups::Integer)
     # count elements in each group
     stops = zeros(Int, ngroups+1)
-    @inbounds for g_ix in groups
-        stops[g_ix+1] += 1
+    @inbounds for gix in groups
+        stops[gix+1] += 1
     end
 
     # group start positions in a sorted table
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -236,30 +236,20 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
     return ngroups, UInt[], Int[], sorted
 end
 
-# Builds RowGroupDict for a given DataFrame.
-# Partly uses the code of Wes McKinney's groupsort_indexer in pandas (file: src/groupby.pyx).
-# - hash: whether row hashes should be computed (if false, the rhashes and gslots fields
-#   hold empty vectors)
-# - sort: whether groups should be sorted
-# - skipmissing: whether rows with missing values should be skipped
-#   rather than put into a separate group
-function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false,
-                    skipmissing::Bool = false)
-    groups = Vector{Int}(undef, nrow(df))
-    ngroups, rhashes, gslots, sorted =
-        row_group_slots(ntuple(i -> df[!, i], ncol(df)), Val(hash), groups, skipmissing)
 
+# Partly uses the code of Wes McKinney's groupsort_indexer in pandas (file: src/groupby.pyx).
+function compute_indices(groups::AbstractVector{<:Integer}, ngroups::Integer)
     # count elements in each group
-    stops = zeros(Int, ngroups)
+    stops = zeros(Int, ngroups+1)
     @inbounds for g_ix in groups
-        stops[g_ix] += 1
+        stops[g_ix+1] += 1
     end
 
     # group start positions in a sorted table
-    starts = Vector{Int}(undef, ngroups)
-    if !isempty(starts)
+    starts = Vector{Int}(undef, ngroups+1)
+    if length(starts) > 1
         starts[1] = 1
-        @inbounds for i in 1:(ngroups-1)
+        @inbounds for i in 1:ngroups
             starts[i+1] = starts[i] + stops[i]
         end
     end
@@ -268,31 +258,24 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
     rperm = Vector{Int}(undef, length(groups))
     copyto!(stops, starts)
     @inbounds for (i, gix) in enumerate(groups)
-        rperm[stops[gix]] = i
-        stops[gix] += 1
+        rperm[stops[gix+1]] = i
+        stops[gix+1] += 1
     end
     stops .-= 1
 
-    # drop group 1 which contains rows with missings in grouping columns
-    if skipmissing
-        popfirst!(starts)
-        popfirst!(stops)
-        groups .-= 1
-        ngroups -= 1
-    end
+    # drop group 1 which contains rows with missings in grouping columns (if any)
+    popfirst!(starts)
+    popfirst!(stops)
 
-    # sort groups if row_group_slots hasn't already done that
-    if sort && !sorted
-        group_perm = sortperm(view(df, rperm[starts], :))
-        group_invperm = invperm(group_perm)
-        permute!(starts, group_perm)
-        Base.permute!!(stops, group_perm)
-        for i in eachindex(groups)
-            gix = groups[i]
-            groups[i] = gix == 0 ? 0 : group_invperm[gix]
-        end
-    end
+    return rperm, starts, stops
+end
 
+# Build RowGroupDict for a given DataFrame
+function group_rows(df::AbstractDataFrame)
+    groups = Vector{Int}(undef, nrow(df))
+    ngroups, rhashes, gslots, sorted =
+        row_group_slots(ntuple(i -> df[!, i], ncol(df)), Val(true), groups, false)
+    rperm, starts, stops = compute_indices(groups, ngroups)
     return RowGroupDict(df, rhashes, gslots, groups, rperm, starts, stops)
 end
 
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -121,7 +121,6 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
         # Use -1 for non-missing values to catch bugs if group is not found
         gix = skipmissing && missings[i] ? 0 : -1
         probe = 0
-        # If skipmissing=true, assign rows containing at least one missing to group 0
         if !skipmissing || !missings[i]
             while true
                 g_row = gslots[slotix]
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -98,7 +98,8 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
 # 3) slot array for a hash map, non-zero values are
 #    the indices of the first row in a group
 # 4) whether groups are already sorted
-# Optional `groups` vector is set to the group indices of each row
+# Optional `groups` vector is set to the group indices of each row (starting at 1)
+# With skipmissing=true, rows with missing values are attributed index 0.
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
                          hash::Val = Val(true),
                          groups::Union{Vector{Int}, Nothing} = nothing,
@@ -111,16 +112,16 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
     @assert sz >= length(rhashes)
     szm1 = sz-1
     gslots = zeros(Int, sz)
-    # If missings are to be skipped, they will all go to group 1,
-    # which will be removed by group_rows
-    ngroups = skipmissing ? 1 : 0
+    # If missings are to be skipped, they will all go to group 0,
+    # which will be removed by functions down the stream
+    ngroups = 0
     @inbounds for i in eachindex(rhashes)
         # find the slot and group index for a row
         slotix = rhashes[i] & szm1 + 1
-        # Use 0 for non-missing values to catch bugs if group is not found
-        gix = skipmissing && missings[i] ? 1 : 0
+        # Use -1 for non-missing values to catch bugs if group is not found
+        gix = skipmissing && missings[i] ? 0 : -1
         probe = 0
-        # If skipmissing=true, assign rows containing at least one missing to group 1
+        # If skipmissing=true, assign rows containing at least one missing to group 0
         if !skipmissing || !missings[i]
             while true
                 g_row = gslots[slotix]
@@ -157,12 +158,12 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
     # and this method needs to allocate a groups vector anyway
     @assert groups !== nothing && all(col -> length(col) == length(groups), cols)
 
-    # If missings are to be skipped, they will all go to group 1,
-    # which will be removed by group_rows
+    # If skipmissing=true, rows with missings all go to group 0,
+    # which will be removed by functions down the stream
     ngroupstup = map(cols) do c
         nlevels(c) + (!skipmissing && eltype(c) >: Missing)
     end
-    ngroups = prod(ngroupstup) + skipmissing
+    ngroups = prod(ngroupstup)
 
     # Fall back to hashing if there would be too many empty combinations.
     # The first check ensures the computation of ngroups did not overflow.
@@ -179,9 +180,6 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
     end
 
     seen = fill(false, ngroups)
-    # If skipmissing=true, missings will all go to group 1,
-    # which will be removed by group_rows
-    seen[1] = skipmissing
     refmaps = map(cols) do col
         nlevs = nlevels(col)
         if col isa CategoricalVector
@@ -212,11 +210,12 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
         end
         vals = map((m, r, s) -> m[r+1] * s, refmaps, refs, strides)
         j = sum(vals) + 1
-        if skipmissing
-            j = any(x -> x < 0, vals) ? 1 : j + 1
+        if skipmissing && any(x -> x < 0, vals)
+            j = 0
+        else
+            seen[j] = true
         end
         groups[i] = j
-        seen[j] = true
     end
     if !all(seen) # Compress group indices to remove unused ones
         oldngroups = ngroups
@@ -227,7 +226,8 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
             remap[i] = ngroups
         end
         @inbounds for i in eachindex(groups)
-            groups[i] = remap[groups[i]]
+            g_ix = groups[i]
+            groups[i] = g_ix > 0 ? remap[g_ix] : 0
         end
         # To catch potential bugs inducing unnecessary computations
         @assert oldngroups != ngroups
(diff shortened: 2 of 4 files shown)


bkamins:
Wy do we have to mage `GroupedDataFrame` mutable for this? Because of `ngroups`?

nalimilan:
No (it doesn't change), that's because of the nullable fields which need to be populated on the first call to `compute_indices!` to avoid recomputing them several times.

bkamins:
Not sure what would be better, but we could make them empty like `Int[]`, and then check if they are empty by `isempty`.

The only corner case to check against would be if it is possible to have them empty in practice (e.g. a grouping of an empty data frame, or an empty `GroupedDataFrame` as a result of selection of zero entries - in the latter case it seems that `starts` and `ends` are empty but not `idx`)..

I have just checked - you can `groupby` an empty `DataFrame`:
```
julia> groupby(DataFrame(a=Int[], b=Int[]), :a)
GroupedDataFrame with 0 groups based on key: a

julia> by(DataFrame(a=Int[], b=Int[]), :a, nrow)
0×1 DataFrame
```

So I am not sure what would be best (maybe your design is cleaner and I guess there is no much harm in having `GroupedDataFrame` mutable - right?)

nalimilan:
Yeah, we could check whether `gd.ngroups == length(gd.idx)`,  but using `nothing` is so much cleaner for this. Actually, it allowed me to catch very easily all places where `compute_indices!(gd)` was needed. Is there any clear drawback to using a mutable struct? Maybe we need to store the fields in variables to avoid slowing down some loops, but for users it shouldn't matter, right?

bkamins:
> Maybe we need to store the fields in variables to avoid slowing down some loops, but for users it shouldn't matter, right?

This is what I would assume. It is just a reflex to avoid `mutable struct`, but in this case it is a heavy object anyway so it should be OK.

I will go through in detail through this PR once the CI start passing as it now fails (or you prefer me to check it now?).

nalimilan:
Test pass now, so feel free to have a look. I still need to check everything is OK and clean, though.

src/dataframerow/utils.jl
```diff
@@ -236,30 +236,20 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
     return ngroups, UInt[], Int[], sorted
 end
 
-# Builds RowGroupDict for a given DataFrame.
```
bkamins: A general question: does whole this code belong to `DataFrameRow`? It seems not - maybe we should move it to other part of the codebase?

src/groupeddataframe/grouping.jl
```diff
@@ -169,33 +170,72 @@ function groupby(df::AbstractDataFrame, cols;
...
 # Single integer indexing
-Base.getindex(gd::GroupedDataFrame, idx::Integer) =
+function Base.getindex(gd::GroupedDataFrame, idx::Integer)
+    gd.idx === nothing && compute_indices!(gd)
     view(gd.parent, gd.idx[gd.starts[idx]:gd.ends[idx]], :)
+end
 
 # Array of integers
-function Base.getindex(gd::GroupedDataFrame, idxs::AbstractArray{T}) where {T<:Integer}
+function Base.getindex(gd::GroupedDataFrame, idxs::AbstractArray{<:Integer})
```
bkamins: `idxs` should be `AbstractVector`

src/groupeddataframe/grouping.jl
```diff
@@ -1064,10 +1146,16 @@ function _combine(f::Any, gd::GroupedDataFrame)
     end
     agg = check_aggregate(fun)
     if agg isa AbstractAggregate && f isa Pair{<:ColumnIndex}
-        idx = gd.idx[gd.starts]
+        idx = Vector{Int}(undef, length(gd))
+        # Find a representative row for each group
+        # TODO: see whether short-circuiting would be possible without slowing down the loop
+        @inbounds for (i, g_ix) in enumerate(gd.groups)
+            g_ix > 0 && (idx[g_ix] = i)
```
bkamins: maybe you can write `idx[g_ix] = ifelse(g_ix > 0, i , 0)`. This should allow avoiding branching in this loop and enable SIMD (I am not sure how performance critical this part of code is).

bkamins:
This is a complex PR. I understand you do two changes in it:
1. code that makes sure we do lazy evaluation of group indices (only when they are needed)
2. code that performs computations that do not require doing point 1. above if this is possible

The point 2. is rather clear (I assume that we handle `sort` and `skipmissing` in this case efficiently here also - right?).

However, I see that you did some changes to the logic of point 1. (e.g. regarding `sort` and `skipmissing` options) - can you please add some comment how we avoid this? (the code is "compressed", so I would prefer to make sure I correctly read its logic). Thank you

src/dataframerow/utils.jl
nalimilan: Yes, it's here for historical reasons I think. I always feel it's not the right place. Let's fix this as part of https://github.com/JuliaData/DataFrames.jl/pull/2050.

src/groupeddataframe/grouping.jl
nalimilan: Unfortunately no, since the index would be `0` in that case. One approach I have tried is doing `g_ix+1`, allocating a vector with one more entry, and resizing it afterwards. But anyway there are no SIMD instructions for this operation on most CPUs (I think some AVX512 machines have them, but I haven't been able to find one to test), so it doesn't make a difference for performance.

src/abstractdataframe/reshape.jl
```diff
@@ -244,6 +244,7 @@ end
 
 function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Symbol},
                   colkey::Int, value::Int, keycol, valuecol, g, renamecols)
+    g.idx === nothing && compute_indices!(g)
```
bkamins: maybe it would be cleaner to define `indices` (or some more appropriate name) unexported function that we would use instead of `g.idx` everywhere and it would do this check internally. This should be a safer design in the long therm I think.

nalimilan:
> However, I see that you did some changes to the logic of point 1. (e.g. regarding `sort` and `skipmissing` options) - can you please add some comment how we avoid this? (the code is "compressed", so I would prefer to make sure I correctly read its logic). Thank you

So the PR was indeed a bit messy in that regard. I've tried to make this cleaner by computing directly indices in `0:ngroups`, with 0 always referring to rows which are supposed to be skipped. That way the first group corresponds to index 1 everywhere (instead of subtracting 1 when `skipmissing=true`). I've also added comments which should make this more explicit, let me know what you think.

src/dataframerow/utils.jl
```diff
@@ -111,16 +112,16 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
     @inbounds for i in eachindex(rhashes)
         # find the slot and group index for a row
         slotix = rhashes[i] & szm1 + 1
-        # Use 0 for non-missing values to catch bugs if group is not found
-        gix = skipmissing && missings[i] ? 1 : 0
+        # Use -1 for non-missing values to catch bugs if group is not found
+        gix = skipmissing && missings[i] ? 0 : -1
         probe = 0
-        # If skipmissing=true, assign rows containing at least one missing to group 1
+        # If skipmissing=true, assign rows containing at least one missing to group 0
```
bkamins: I do not understand this comment (especially given the earlier comment).
The code below seems to be executed if `skipmissing==false` or if `missings[i]==false` so this.
Is this related to the fact that we initialized `gslots` to `0`. But if this is the case then how did this work before as the condition and the initialization has not changed.

src/dataframerow/utils.jl
```diff
@@ -212,11 +210,12 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
         end
         vals = map((m, r, s) -> m[r+1] * s, refmaps, refs, strides)
         j = sum(vals) + 1
-        if skipmissing
-            j = any(x -> x < 0, vals) ? 1 : j + 1
+        if skipmissing && any(x -> x < 0, vals)
```
bkamins: Can you please comment when `vals` can take a negative value?

src/dataframerow/utils.jl
```diff
@@ -268,31 +262,25 @@ function group_rows(df::AbstractDataFrame, hash::Bool = true, sort::Bool = false
...
 
-    # drop group 1 which contains rows with missings in grouping columns
-    if skipmissing
-        popfirst!(starts)
-        popfirst!(stops)
-        groups .-= 1
-        ngroups -= 1
-    end
+    # When skipmissing=true was used, group 0 corresponds to missings to drop
+    # Otherwise it's empty
```
bkamins: maybe we can put an `@assert` for this statement - it should not be expensive I think.

bkamins:
Reading through this PR I have found the following problem:
```
julia> df1 = DataFrame(x=["a", "b", "c"], y=["1", "2", "3"])
3×2 DataFrame
│ Row │ x      │ y      │
│     │ String │ String │
├─────┼────────┼────────┤
│ 1   │ a      │ 1      │
│ 2   │ b      │ 2      │
│ 3   │ c      │ 3      │

julia> df2 = categorical(df1, :x)
3×2 DataFrame
│ Row │ x            │ y      │
│     │ Categorical… │ String │
├─────┼──────────────┼────────┤
│ 1   │ a            │ 1      │
│ 2   │ b            │ 2      │
│ 3   │ c            │ 3      │

julia> levels!(df2.x, ["b", "c", "a"])
 "a"
 "b"
 "c"

julia> by(df1, :x, identity)
3×2 DataFrame
│ Row │ x      │ y      │
│     │ String │ String │
├─────┼────────┼────────┤
│ 1   │ a      │ 1      │
│ 2   │ b      │ 2      │
│ 3   │ c      │ 3      │

julia> by(df2, :x, identity)
3×2 DataFrame
│ Row │ x            │ y      │
│     │ Categorical… │ String │
├─────┼──────────────┼────────┤
│ 1   │ b            │ 2      │
│ 2   │ c            │ 3      │
│ 3   │ a            │ 1      │
```
In other words - we always sort in `by`/`groupby` if we pass categorical grouping columns, which is inconsistent with what we promise (this should only happen if `sort=true` is passed).

bkamins:
I have prepared a test suite that should pass:
```
@testset "by additional tests" begin
    ref_df = DataFrame(x=["c", "a", "b", "c", "c", "a", "a", "a", "b", "c"],
                       y=["2", "2", "2", "3", "1", "1", "1", "1", "1", "1"],
                       z=1:10)
    ref1 = DataFrame(x=["c", "a", "b"], z=[[1, 4, 5, 10], [2, 6, 7, 8], [3, 9]])
    ref1_s = sort(ref1, :x)
    ref1_m = ref1[[1,3], :]
    ref1_mk = DataFrame(x=["c", missing, "b"], z=[[1, 4, 5, 10], [2, 6, 7, 8], [3, 9]])
    ref1_sm = sort(ref1_m, :x)
    ref1_smk = sort(ref1_mk, :x)
    ref2 = DataFrame(x=["c", "a", "b", "c", "c", "a", "b"],
                     y=["2", "2", "2", "3", "1", "1", "1"],
                     z=[[1], [2], [3], [4], [5, 10], [6, 7, 8], [9]])
    ref2_s = sort(ref2, [:x, :y])
    ref2_m = ref2[[1,3,4,5,7], :]
    ref2_mk = DataFrame(x=["c", missing, "b", "c", "c", missing, "b"],
                        y=["2", "2", "2", "3", "1", "1", "1"],
                        z=[[1], [2], [3], [4], [5, 10], [6, 7, 8], [9]])
    ref2_sm = sort(ref2_m, [:x, :y])
    ref2_smk = sort(ref2_mk, [:x, :y])

    for trans1 = (:no, :pool, :nom, :ord), trans2 = (:no, :pool, :nom, :ord), rev=(true, false)
        @testset "trans1=$trans1, trans2=$trans2, rev=$rev" begin
            df = copy(ref_df)
            if trans1 == :pool
                df.x = PooledArray(df.x)
            elseif trans1 == :nom
                df.x = categorical(df.x)
                rev && levels!(df.x, reverse(levels(df.x)))
            elseif trans1 == :ord
                df.x = categorical(df.x)
                ordered!(df.x, true)
                rev && levels!(df.x, reverse(levels(df.x)))
            end
            if trans2 == :pool
                df.y = PooledArray(df.y)
            elseif trans2 == :nom
                df.y = categorical(df.y)
                rev && levels!(df.y, reverse(levels(df.y)))
            elseif trans2 == :ord
                df.y = categorical(df.y)
                ordered!(df.y, true)
                rev && levels!(df.y, reverse(levels(df.y)))
            end
            @test by(df, :x, z=:z=>Base.vect) == ref1
            @test by(df, :x, z=:z=>Base.vect, sort=true) == ref1_s
            @test by(df, [:x, :y], z=:z=>Base.vect) == ref2
            @test by(df, [:x, :y], z=:z=>Base.vect, sort=true) == ref2_s
            df.x = replace(df.x, "a"=>missing)
            @test by(df, :x, z=:z=>Base.vect, skipmissing=true) == ref1_m
            @test by(df, :x, z=:z=>Base.vect, sort=true, skipmissing=true) == ref1_sm
            @test by(df, [:x, :y], z=:z=>Base.vect, skipmissing=true) == ref2_m
            @test by(df, [:x, :y], z=:z=>Base.vect, sort=true, skipmissing=true) == ref2_sm
            @test by(df, :x, z=:z=>Base.vect) ≅ ref1_mk
            @test by(df, :x, z=:z=>Base.vect, sort=true) ≅ ref1_smk
            @test by(df, [:x, :y], z=:z=>Base.vect) ≅ ref2_mk
            @test by(df, [:x, :y], z=:z=>Base.vect, sort=true) ≅ ref2_smk
        end
    end
end
```
but currently (also on master, but it is really needed to run these tests in this PR I think) it produces:
```
Test Summary:                         | Pass  Fail  Total
by additional tests                   |  274   110    384
  trans1=no, trans2=no, rev=true      |   12           12
  trans1=no, trans2=no, rev=false     |   12           12
  trans1=no, trans2=pool, rev=true    |   12           12
  trans1=no, trans2=pool, rev=false   |   12           12
  trans1=no, trans2=nom, rev=true     |    9     3     12
  trans1=no, trans2=nom, rev=false    |   12           12
  trans1=no, trans2=ord, rev=true     |    9     3     12
  trans1=no, trans2=ord, rev=false    |   12           12
  trans1=pool, trans2=no, rev=true    |   12           12
  trans1=pool, trans2=no, rev=false   |   12           12
  trans1=pool, trans2=pool, rev=true  |   11     1     12
  trans1=pool, trans2=pool, rev=false |   11     1     12
  trans1=pool, trans2=nom, rev=true   |    9     3     12
  trans1=pool, trans2=nom, rev=false  |   12           12
  trans1=pool, trans2=ord, rev=true   |    9     3     12
  trans1=pool, trans2=ord, rev=false  |   12           12
  trans1=nom, trans2=no, rev=true     |    4     8     12
  trans1=nom, trans2=no, rev=false    |    9     3     12
  trans1=nom, trans2=pool, rev=true   |    4     8     12
  trans1=nom, trans2=pool, rev=false  |    9     3     12
  trans1=nom, trans2=nom, rev=true    |    3     9     12
  trans1=nom, trans2=nom, rev=false   |    8     4     12
  trans1=nom, trans2=ord, rev=true    |    3     9     12
  trans1=nom, trans2=ord, rev=false   |    8     4     12
  trans1=ord, trans2=no, rev=true     |    4     8     12
  trans1=ord, trans2=no, rev=false    |    9     3     12
  trans1=ord, trans2=pool, rev=true   |    4     8     12
  trans1=ord, trans2=pool, rev=false  |    9     3     12
  trans1=ord, trans2=nom, rev=true    |    3     9     12
  trans1=ord, trans2=nom, rev=false   |    8     4     12
  trans1=ord, trans2=ord, rev=true    |    3     9     12
  trans1=ord, trans2=ord, rev=false   |    8     4     12
ERROR: Some tests did not pass: 274 passed, 110 failed, 0 errored, 0 broken.
```
If you are OK with it I think it should be added to this PR and the methods should be corrected to pass this test. What do you think?

nalimilan:
Yes we always sort categorical columns because that's not more expensive than not sorting. Do you think that's a problem? I don't think we promise not to sort... When `sort=false` the order is undefined (due to how the `Dict` works in the general case).

src/dataframerow/utils.jl
nalimilan: The problem is, we don't know whether `skipmissing=true` was used in this function. We could store it in the `GroupedDataFrame` just for checking, but normally tests would catch bugs in that regard anyway.

src/abstractdataframe/reshape.jl
nalimilan: You mean move the check inside `compute_indices!`?

src/dataframerow/utils.jl
nalimilan: Yeah, that indicates a missing value in the corresponding column. I'll add a comment.

src/dataframerow/utils.jl
nalimilan: The group index is always set to `gix`, which previously was `1` for missing values and is now `0`. `gslots` is something different: there a 0 indicates an unfilled slot in the hash table (it's not used for missing values).

Can you suggest a better wording?

src/abstractdataframe/reshape.jl
bkamins: even more. I would define the following for example:
```
function Base.getproperty(g::GroupedDataFrame, f::Symbol)
    f === :idx && getfield(g, :idx) === nothing && compute_indices!(g)
    getfield(g, f)
end
```
in this way future developers will not have to think about checking if `:idx` was computed or not.

bkamins: of course we do not have to do it.

src/dataframerow/utils.jl

src/dataframerow/utils.jl
bkamins: This is what I assumed, but my question is how does this happen (i.e. what is the source of this negative value). For this to happen either `m[r+1]` or `s` should be negative. I understand that `s` is always non-negative, and `m` is `refmaps` where I thought we are non-negative. For sure I miss something, but I am asking so that in the future I understand this part of code better.

src/dataframerow/utils.jl
bkamins: Again - I need to understand this code first 😭. If `skipmissing==true` and find missing then we set `gix=0` and later set `groups[i] = gix` to skip this group - right? If this is so I would just add something that group `0` is later skipped?

bkamins: Also maybe in https://github.com/JuliaData/DataFrames.jl/pull/2047/files#diff-223a9905c9dcda6f61927d162475af85L129 it would be better to write:
```
ngroups += 1
gix = ngroups
```
to make the code simpler? (this is a minor thing, but I prefer to have more readable code especially where it is complex)

bkamins:
Summing up the discussion I would recommend:
* set `sort` to `true` everywhere by default
* add a comment that if `sort=false` we heuristically choose the fastest possible grouping method and in this case the order of groups is undefined.

src/groupeddataframe/grouping.jl
```diff
@@ -846,21 +883,53 @@ check_aggregate(::typeof(length)) = Aggregate(length)
...
+            throw(ArgumentError("some groups contain only missing values"))
+        end
+    else # Finding first row is faster than computing all group indices
+        groups = gd.groups
+        if rev
+            r = length(groups):-1:1
+        else
+            r = 1:length(groups)
+        end
+        filled = fill(false, gd.ngroups)
```
bkamins: Why in this file in general do we use `fill(false, some_size` instead of `falses(some_size)`. I know there is a performance trade-off here, but I thought that recently `BitArray` as consistently faster than `Vector{Bool}` have you maybe benchmarked in recently?

src/abstractdataframe/reshape.jl
nalimilan: Yeah I had quickly considered this option. It's appealing. If we do that, it should also apply to `starts` and `ends` fields.

src/dataframerow/utils.jl
nalimilan: Yes, that's it. Feel free to suggest a wording. (Actually the comment is a bit redundant with the one at the top of the loop...)

I'm not sure `gix = ngroups += 1` is the most complex thing in this code :-). But why not divide it in two steps if you prefer...

src/dataframerow/utils.jl
nalimilan: Well the code above does all kinds of weird computations so that `refmaps` has a `-1` entry for `missing` (the first entry for `CategoricalArray`, an arbitrary entry for `PooledArray`). Do you think a comment is needed above?

src/groupeddataframe/grouping.jl
nalimilan: Probably worth checking, yes.

nalimilan:
> Summing up the discussion I would recommend:
> 
>     * set `sort` to `true` everywhere by default
> 
>     * add a comment that if `sort=false` we heuristically choose the fastest possible grouping method and in this case the order of groups is undefined.

I'll do that in another PR, as it's really a separate issue.

src/abstractdataframe/reshape.jl
bkamins: Well - please decide what you prefer. It is an internal issue, and you are the lead developer of this part of code 😄.

src/dataframerow/utils.jl
bkamins: > (Actually the comment is a bit redundant with the one at the top of the loop...)

This is exactly what misguided me initially.

bkamins:
```suggestion
        # If skipmissing=true, rows containing at least one missing get a group number 0 which is skipped later
```

src/dataframerow/utils.jl
bkamins: Yes, I think ideally a comment how the contents of `remaps` should be interpreted would be ideal. Thank you.

bkamins:
Agreed for a separate issue.

And thank you for being willing to comment the code more. I think it is really important as it is really complex. Also it might help if in signatures of methods, where it is known, we added the expected type of passed value (in some places it is left unspecified and then it is hard to decode what certain variable means). Of course feel free to skip this request.

src/groupeddataframe/grouping.jl
nalimilan: So `BitArray` is still significantly slower on Julia 1.3, though the difference is irregular:
```julia


function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
                    gd::GroupedDataFrame; rev::Bool=false)
    ngroups = gd.ngroups
    # Use group indices if they have already been computed
    groups = gd.groups
    if rev
        r = length(groups):-1:1
    else
        r = 1:length(groups)
    end
    filled = fill(false, gd.ngroups)
    nfilled = 0
    @inbounds for i in r
        gix = groups[i]
        x = incol[i]
        if gix > 0 && (condf === nothing || condf(x)) && !filled[gix]
            filled[gix] = true
            outcol[gix] = x
            nfilled += 1
            nfilled == ngroups && break
        end
    end
    if nfilled < length(outcol)
        throw(ArgumentError("some groups contain only missing values"))
    end
    outcol
end


function fillfirst2!(condf, outcol::AbstractVector, incol::AbstractVector,
                     gd::GroupedDataFrame; rev::Bool=false)
    ngroups = gd.ngroups
    # Use group indices if they have already been computed
    groups = gd.groups
    if rev
        r = length(groups):-1:1
    else
        r = 1:length(groups)
    end
    filled = falses(gd.ngroups)
    nfilled = 0
    @inbounds for i in r
        gix = groups[i]
        x = incol[i]
        if gix > 0 && (condf === nothing || condf(x)) && !filled[gix]
            filled[gix] = true
            outcol[gix] = x
            nfilled += 1
            nfilled == ngroups && break
        end
    end
    if nfilled < length(outcol)
        throw(ArgumentError("some groups contain only missing values"))
    end
    outcol
end


using BenchmarkTools
n = 100_000
for k in (10, 100, 1000, 10_000)
    println("$k groups")
    df = DataFrame(x=rand(1:k, n), y=rand(n));
    gd = groupby(df, :x);
    out = Vector{Float64}(undef, k)
    in = Vector{Float64}(undef, n)
    @btime fillfirst!(nothing, $out, $in, $gd);
    @btime fillfirst2!(nothing, $out, $in, $gd);
end

10 groups
  70.659 ns (1 allocation: 96 bytes)
  94.355 ns (2 allocations: 128 bytes)
100 groups
  657.304 ns (1 allocation: 192 bytes)
  1.175 μs (2 allocations: 128 bytes)
1000 groups
  8.644 μs (1 allocation: 1.06 KiB)
  14.491 μs (2 allocations: 240 bytes)
10000 groups
  140.208 μs (1 allocation: 9.94 KiB)
  198.912 μs (2 allocations: 1.36 KiB)
```

bkamins: Interesting. Thank you for checking this!

nalimilan:
Thanks. I had to hoist a few field accesses manually to avoid performance regressions with the `getproperty` approach (I'll update the benchmarks shortly). That's annoying but it's not the end of the world either, and with a mutable struct it's a good idea anyway.

Let me know whether I've addressed all of your remarks about comments.

bkamins:
Regarding comments - I think what we have now is OK. Thank you.

bkamins:
Thank you!

PR #2049, opened 2019-12-09
Allow stack to make column variable contain a categorical vector of strings

### LINKED ISSUE #1947: Make stack create CategoricalVector{String} vector from column names
Currently it is `Vector{Symbol}`. See https://github.com/JuliaInterop/RCall.jl/issues/331 for a discussion (but I think this would be also useful in general).

bkamins:
Fixes #1947

I have left an option for the column to contain `Symbol`s (via a keyword argument), as it simplified the deprecation period and maybe some old code uses this.

After the deprecation the only thing to change is the default value of the kwarg (if we decide to use the design I have proposed).

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -385,7 +385,7 @@ struct RepeatedVector{T} <: AbstractVector{T}
 end
 
 Base.parent(v::RepeatedVector) = v.parent
-Base.levels(v::RepeatedVector) = levels(parent(v))
+DataAPI.levels(v::RepeatedVector) = levels(parent(v))
 CategoricalArrays.isordered(v::RepeatedVector) = isordered(parent(v))
 
 function CategoricalArrays._isordered(v::RepeatedVector)
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -1,7 +1,7 @@
 """
     stack(df::AbstractDataFrame, [measure_vars], [id_vars];
           variable_name::Symbol=:variable, value_name::Symbol=:value,
-          view::Bool=false, stringvar::Bool=false)
+          view::Bool=false, vareltype::Type=CategoricalString)
 
 Stack a data frame `df`, i.e. convert it from wide to long format.
 
@@ -32,8 +32,9 @@ that return views into the original data frame.
   each of `measure_vars`
 - `view` : whether the stacked data frame should be a view rather than contain
    freshly allocated vectors.
-- `stringvar` : if `true` column `variable_name` is a categorical vector
-   of strings; if `false` (currently the deprecated default) it is a vector of `Symbol`.
+- `vareltype` : determines the element type of column `variable_name`. By default
+   a categorical vector is created. If `vareltype=Symbol` it is a vector of `Symbol`,
+   and if `vareltype=String` a vector of `String` is produced.
 
 # Examples
 ```julia
@@ -51,23 +52,27 @@ d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
 """
 function stack(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
                id_vars::AbstractVector{<:Integer}; variable_name::Symbol=:variable,
-               value_name::Symbol=:value, view::Bool=false, stringvar::Bool=false)
+               value_name::Symbol=:value, view::Bool=false,
+               vareltype::Type=CategoricalString)
     if view
         return _stackview(df, measure_vars, id_vars, variable_name=variable_name,
-                          value_name=value_name, stringvar=stringvar)
+                          value_name=value_name, vareltype=vareltype)
     end
     N = length(measure_vars)
     cnames = _names(df)[id_vars]
     pushfirst!(cnames, value_name)
     pushfirst!(cnames, variable_name)
-    if stringvar
+    if vareltype <: CategoricalString
         nms = String.(_names(df)[measure_vars])
         catnms = categorical(nms)
         levels!(catnms, nms)
-    else
-        Base.depwarn("`stringvar=false` as a default is deprecated." *
-                         "In the future it will default to `true`", :vcat)
+    elseif vareltype <: Symbol
         catnms = _names(df)[measure_vars]
+    elseif vareltype <: String
+        catnms = PooledArrays.PooledArray(String.(_names(df)[measure_vars]))
+    else
+        throw(ArgumentError("`vareltype` keyword argument accepts only `CategoricalString`, " *
+                            "`String` or `Symbol` as a value."))
     end
     DataFrame(AbstractVector[repeat(catnms, inner=nrow(df)), # variable
                              vcat([df[!, c] for c in measure_vars]...),           # value
@@ -77,30 +82,30 @@ end
 
 function stack(df::AbstractDataFrame, measure_var::Int, id_var::Int;
                variable_name::Symbol=:variable, value_name::Symbol=:value,
-               view::Bool=false, stringvar::Bool=false)
+               view::Bool=false, vareltype::Type=CategoricalString)
     stack(df, [measure_var], [id_var],
-          variable_name=variable_name, value_name=value_name, view=view, stringvar=stringvar)
+          variable_name=variable_name, value_name=value_name, view=view, vareltype=vareltype)
 end
 
 function stack(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer}, id_var::Int;
                variable_name::Symbol=:variable, value_name::Symbol=:value,
-               view::Bool=false, stringvar::Bool=false)
+               view::Bool=false, vareltype::Type=CategoricalString)
     stack(df, measure_vars, [id_var],
-          variable_name=variable_name, value_name=value_name, view=view, stringvar=stringvar)
+          variable_name=variable_name, value_name=value_name, view=view, vareltype=vareltype)
 end
 
 function stack(df::AbstractDataFrame, measure_var::Int, id_vars::AbstractVector{<:Integer};
                variable_name::Symbol=:variable, value_name::Symbol=:value,
-               view::Bool=false, stringvar::Bool=false)
+               view::Bool=false, vareltype::Type=CategoricalString)
     stack(df, [measure_var], id_vars;
-          variable_name=variable_name, value_name=value_name, view=view, stringvar=stringvar)
+          variable_name=variable_name, value_name=value_name, view=view, vareltype=vareltype)
 end
 
 function stack(df::AbstractDataFrame, measure_vars, id_vars;
                variable_name::Symbol=:variable, value_name::Symbol=:value,
-               view::Bool=false, stringvar::Bool=false)
+               view::Bool=false, vareltype::Type=CategoricalString)
     stack(df, index(df)[measure_vars], index(df)[id_vars];
-          variable_name=variable_name, value_name=value_name, view=view, stringvar=stringvar)
+          variable_name=variable_name, value_name=value_name, view=view, vareltype=vareltype)
 end
 
 # no vars specified, by default select only numeric columns
@@ -109,10 +114,10 @@ numeric_vars(df::AbstractDataFrame) =
 
 function stack(df::AbstractDataFrame, measure_vars = numeric_vars(df);
                variable_name::Symbol=:variable, value_name::Symbol=:value,
-               view::Bool=false, stringvar::Bool=false)
+               view::Bool=false, vareltype::Type=CategoricalString)
     mv_inds = index(df)[measure_vars]
     stack(df, mv_inds, setdiff(1:ncol(df), mv_inds);
-          variable_name=variable_name, value_name=value_name, view=view, stringvar=stringvar)
+          variable_name=variable_name, value_name=value_name, view=view, vareltype=vareltype)
 end
 
 """
@@ -388,11 +393,6 @@ Base.parent(v::RepeatedVector) = v.parent
 DataAPI.levels(v::RepeatedVector) = levels(parent(v))
 CategoricalArrays.isordered(v::RepeatedVector) = isordered(parent(v))
 
-function CategoricalArrays._isordered(v::RepeatedVector)
-    p = parent(v)
-    p isa AbstractCategoricalArray ? isordered(p) : false
-end
-
 function Base.getindex(v::RepeatedVector, i::Int)
     N = length(v.parent)
     idx = Base.fld1(mod1(i,v.inner*N),v.inner)
@@ -415,17 +415,22 @@ end
 
 function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
                     id_vars::AbstractVector{<:Integer}; variable_name::Symbol=:variable,
-                    value_name::Symbol=:value, stringvar::Bool=false)
+                    value_name::Symbol=:value, vareltype::Type=CategoricalString)
     N = length(measure_vars)
     cnames = _names(df)[id_vars]
     pushfirst!(cnames, value_name)
     pushfirst!(cnames, variable_name)
-    if stringvar
+    if vareltype <: CategoricalString
         nms = String.(_names(df)[measure_vars])
         catnms = categorical(nms)
         levels!(catnms, nms)
-    else
+    elseif vareltype <: Symbol
         catnms = _names(df)[measure_vars]
+    elseif vareltype <: String
+        catnms = String.(_names(df)[measure_vars])
+    else
+        throw(ArgumentError("`vareltype` keyword argument accepts only `CategoricalString`, " *
+                            "`String` or `Symbol` as a value."))
     end
     DataFrame(AbstractVector[RepeatedVector(catnms, nrow(df), 1), # variable
                              StackedVector(Any[df[!, c] for c in measure_vars]),    # value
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/reshape.jl
```diff
@@ -1,12 +1,13 @@
 """
     stack(df::AbstractDataFrame, [measure_vars], [id_vars];
-          variable_name::Symbol=:variable, value_name::Symbol=:value, view::Bool=false)
+          variable_name::Symbol=:variable, value_name::Symbol=:value,
+          view::Bool=false, stringvar::Bool=false)
```
nalimilan: Maybe call this `categorical` instead? That sounds more explicit.

src/abstractdataframe/reshape.jl
```diff
@@ -48,54 +51,68 @@ d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
...
-    DataFrame(AbstractVector[repeat(_names(df)[measure_vars], inner=nrow(df)), # variable
+    cnames = _names(df)[id_vars]
+    pushfirst!(cnames, value_name)
+    pushfirst!(cnames, variable_name)
+    if stringvar
+        nms = String.(_names(df)[measure_vars])
+        catnms = categorical(nms)
+        levels!(catnms, nms)
+    else
+        Base.depwarn("`stringvar=false` as a default is deprecated." *
```
nalimilan: Usually we use `nothing` as a default so that people can get the old behavior without a deprecation. Or if we want to remove that possibility in the future, better say it.

BTW, there are three possibly useful types: categorical, symbol and string. Maybe the argument could take the type instead of being a Boolean?

src/abstractdataframe/reshape.jl
```diff
@@ -48,54 +51,68 @@ d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
...
+                         "In the future it will default to `true`", :vcat)
+        catnms = _names(df)[measure_vars]
+    end
+    DataFrame(AbstractVector[repeat(catnms, inner=nrow(df)), # variable
```
nalimilan: Align the comment.

src/abstractdataframe/reshape.jl
bkamins: I will reply here to both comments as they are related. I think it is OK to allow either `Symbol`, `String` or categorical, with the default to categorical.

Maybe we should call kwarg `vartype` and allow it to take the following values:
* `:symbol`
* `:string`
* `:categorical` (this will be the default after the deprecation)

Now the ability to handle the deprecation of kwargs is not clean in Julia unfortunately. We could go for `nothing` as a temporary default as you propose (and not document it) and not document it (so that we can later remove it without deprecation).

Is what I propose OK?

nalimilan: Why not take types rather than symbols? That sounds more Julian.

Regarding the deprecation, if we want to keep the current behavior in the next version and only change it in the following one, better use `nothing`. But we could also decide that it's not too breaking and just change the default now so that we don't have to wait for another release.

nalimilan:
While you're at it, it would also make sense to return a `PooledArray` when we don't return a `CategoricalArray`.

bkamins:
OK, so I will make:
1) a breaking change (it is mildly breaking fortunately) - so the default will be categorical I think
2) I know you wanted types, but what type would you pass for categorical? (that is why I did not propose them initially)?

Now - when making a proposal of what type to pass please take note that if `view=true` we return `RepeatedVector` not categorical or pooled one anyway (it will just differ if it contains `String` or `CategoricalString`).

This also means that we should merge this later (not immediately), as we might have some patch releases before 0.21.

nalimilan:
>     1. a breaking change (it is mildly breaking fortunately) - so the default will be categorical I think

OK.

>     2. I know you wanted types, but what type would you pass for categorical? (that is why I did not propose them initially)?

You can use `CategoricalString` (soon to be `CategoricalValue{String}` in the next breaking CategoricalArrays release), right?

> Now - when making a proposal of what type to pass please take note that if `view=true` we return `RepeatedVector` not categorical or pooled one anyway (it will just differ if it contains `String` or `CategoricalString`).

Hmm, yeah, we should check how `RepeatedVector` behaves with `CategoricalString`. For example, `levels` might give weird results. Maybe we should override that function.

> This also means that we should merge this later (not immediately), as we might have some patch releases before 0.21.

That, or create a release-0.21 branch for backports when we need it.

bkamins:
> Hmm, yeah, we should check how RepeatedVector behaves with CategoricalString. For example, levels might give weird results.

I have checked already when I opened a PR - it works OK (that I why I have left this implementation).

> you can use `CategoricalString` (soon to be `CategoricalValue{String}` in the next breaking CategoricalArrays release)

this is what I did not want to use 😄, exactly because it would change soon and because it is long. But now I think it is OK, as this will be the default so users will not have to write it.

PS. When do you plan to break CategoricalArrays.jl with a new release?

nalimilan:
> > 
> 
> I have checked already when I opened a PR - it works OK (that I why I have left this implementation).

Well it doesn't preserve the order of levels (which is the whole point...):
```julia
julia> levels(DataFrames.RepeatedVector(levels!(categorical(["a", "b", "c"]), ["c", "b", "a"]), 3, 2))
 "a"
 "b"
 "c"
```

> this is what I did not want to use smile, exactly because it would change soon and because it is long. But now I think it is OK, as this will be the default so users will not have to write it.

Yes that's one of the advantages of using it by default...

> PS. When do you plan to break CategoricalArrays.jl with a new release?

Not sure. I guess we can do it as soon as we want. It could depend on whether we want to wait for the next breaking DataFrames release or not.

bkamins:
Well - it is clearly impossible to preserve, but `copy` does:
```
julia> levels(copy(DataFrames.RepeatedVector(levels!(categorical(["a", "b", "c"]), ["c", "b", "a"]), 3, 2)))
3-element Array{String,1}:
 "c"
 "b"
 "a"
```
and this is what I have meant earlier.

The design choices I have contemplated earlier were:
1. always materialize varaibles column (this goes against a `view` approach, but would get us what we want)
2. disallow `categorical` in combination with `view=true` (but this goes against the idea that categorical is the default)
3. do what I did now (which looses order, but you can recover it when you materialize the column)

I am not sure what is best. Maybe we should go for yet another solution?

4. for both view-types make them check if the underlying vector is a `CategoricalVector` and if yes make `levels` return the levels of the parent.

Thinking of it now - I think option 4. should hit the spot of all requirements. Are you OK with this? (this would also partially solve the problem that CategoricalArrays.jl does not provide opt-in trait system).

Regarding breaking release - OK, let us sync it with 0.21 for DataFrames.jl. I hope we can make it faster than in 6 months like 0.20 took (but time will show - maybe some complex issues would arise in the process).

nalimilan:
Yes, option 4 sounds good if it's not too hard to implement. Could also implement `isordered`.

> Regarding breaking release - OK, let us sync it with 0.21 for DataFrames.jl. I hope we can make it faster than in 6 months like 0.20 took (but time will show - maybe some complex issues would arise in the process).

OK. We can make a CategoricalArrays release some time before that, but it won't be used by many people until DataFrames is released.

bkamins:
@nalimilan - can you please have a look if the list of methods I added in the last commit is complete or something more needs to be added to make `RepeatedVector` pretend to be a `CategoricalArray`?

src/abstractdataframe/reshape.jl
```diff
@@ -367,6 +384,15 @@ struct RepeatedVector{T} <: AbstractVector{T}
     outer::Int
 end
 
+Base.parent(v::RepeatedVector) = v.parent
+Base.levels(v::RepeatedVector) = levels(parent(v))
```
nalimilan:
```suggestion
DataAPI.levels(v::RepeatedVector) = levels(parent(v))
```

src/abstractdataframe/reshape.jl
```diff
@@ -367,6 +384,15 @@ struct RepeatedVector{T} <: AbstractVector{T}
...
+CategoricalArrays.isordered(v::RepeatedVector) = isordered(parent(v))
+
+function CategoricalArrays._isordered(v::RepeatedVector)
```
nalimilan: I don't really like depending on internal functions like this. We should be able to fix this in CategoricalArrays by dispatching on `AbstractArray{<:CategoricalValue}` instead. Anyway that really concerns a corner case (calling `CategoricalArray(df.variable)`), so that's not urgent.

src/abstractdataframe/reshape.jl
bkamins: Ah - right :)

src/abstractdataframe/reshape.jl
bkamins: OK - so I remove it here and assume that CategoricalArrays.jl will fix this.
(it also touches `categorical(df.variable)` which is pretty common though).

src/abstractdataframe/reshape.jl
```diff
@@ -48,54 +52,72 @@ d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
...
-    insert!(cnames, 1, variable_name)
-    DataFrame(AbstractVector[repeat(_names(df)[measure_vars], inner=nrow(df)), # variable
+    cnames = _names(df)[id_vars]
+    pushfirst!(cnames, value_name)
+    pushfirst!(cnames, variable_name)
+    if vareltype <: CategoricalString
+        nms = String.(_names(df)[measure_vars])
+        catnms = categorical(nms)
+        levels!(catnms, nms)
+    elseif vareltype <: Symbol
```
nalimilan:
```suggestion
    elseif vareltype === Symbol
```
(Same below.)

src/abstractdataframe/reshape.jl
```diff
@@ -48,54 +52,72 @@ d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
...
+        catnms = _names(df)[measure_vars]
+    elseif vareltype <: String
```
nalimilan:
```suggestion
    elseif vareltype === String
```

src/abstractdataframe/reshape.jl
```diff
@@ -48,54 +52,72 @@ d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
...
+        catnms = PooledArrays.PooledArray(String.(_names(df)[measure_vars]))
```
nalimilan: Is the module prefix needed?

src/abstractdataframe/reshape.jl
```diff
@@ -367,6 +389,10 @@ struct RepeatedVector{T} <: AbstractVector{T}
     outer::Int
 end
 
+Base.parent(v::RepeatedVector) = v.parent
+DataAPI.levels(v::RepeatedVector) = levels(parent(v))
+CategoricalArrays.isordered(v::RepeatedVector) = isordered(parent(v))
```
nalimilan:
```suggestion
CategoricalArrays.isordered(v::RepeatedVector{<:Union{CategoricalValue,CategoricalString}) = isordered(parent(v))
```

test/reshape.jl
```diff
@@ -279,6 +279,13 @@ end
     @test_throws ArgumentError d1s[!, 1][1.0]
     @test_throws ArgumentError d1s[!, 2][1.0]
 
+    d1ss = stack(d1, [:a, :b], view=true)
+    @test d1ss[!, 1][[1,24]] == ["a", "b"]
+    d1ss = stack(d1, [:a, :b], view=true, vareltype=String)
+    @test d1ss[!, 1][[1,24]] == ["a", "b"]
```
nalimilan: Also check that the result is a `PooledArray`?

src/abstractdataframe/reshape.jl
```diff
@@ -31,6 +32,9 @@ that return views into the original data frame.
   each of `measure_vars`
 - `view` : whether the stacked data frame should be a view rather than contain
    freshly allocated vectors.
+- `vareltype` : determines the element type of column `variable_name`. By default
+   a categorical vector is created. If `vareltype=Symbol` it is a vector of `Symbol`,
```
nalimilan:
```suggestion
   a `CategoricalVector{String}` is created. If `vareltype=Symbol` it is a vector of `Symbol`,
```

src/abstractdataframe/reshape.jl
```diff
@@ -1,12 +1,13 @@
 """
     stack(df::AbstractDataFrame, [measure_vars], [id_vars];
-          variable_name::Symbol=:variable, value_name::Symbol=:value, view::Bool=false)
+          variable_name::Symbol=:variable, value_name::Symbol=:value,
+          view::Bool=false, vareltype::Type=CategoricalString)
```
nalimilan: Maybe `variable_eltype` for consistency with `variable_name`?

test/reshape.jl
```diff
@@ -386,4 +393,58 @@ end
...
+    @test d1s.variable isa CategoricalVector{String}
+    @test levels(d1s.variable) == ["d", "c"]
+    d1s = stack(d1, [:d, :c], view=true)
+    @test d1s.variable isa DataFrames.RepeatedVector{<:CategoricalString}
+    @test levels(d1s.variable) == ["d", "c"]
+    @test d1s[:, 1] isa CategoricalVector{String}
+    @test levels(d1s[:, 1]) == ["d", "c"]
+
+    d1s = stack(d1, [:d, :c], vareltype=String)
+    @test d1s.variable isa PooledArrays.PooledArray{String}
```
nalimilan:
```suggestion
    @test d1s.variable isa PooledVector{String}
```

src/abstractdataframe/reshape.jl
bkamins: it is not true for `view=true` as then it is `RepeatedVector`. So I was not sure what to write.
Given this comment do you still think we should apply your change?

src/abstractdataframe/reshape.jl
bkamins: OK, but I have always thought `<:` is the recommended style in such cases :).

src/abstractdataframe/reshape.jl
nalimilan: Ah, right. Then maybe just "a categorical vector of strings" or something like that.

src/abstractdataframe/reshape.jl
nalimilan: When I see `<:` I always wonder whether the RHS is an abstract type. But that's really a minor point, YMMV.

src/abstractdataframe/reshape.jl
bkamins: Strictly speaking it is possible that some type defines `isordered` and does not expose the eltype.

But if you think we should add it it should be `AbstractVector{<:Union{CategoricalValue, CategoricalString, Missing}}` I think.

Apart from the simple case of `misssing` in the vector also something like this is true:
```
julia> eltype(categorical([1,2,"a"]))
Union{Missing, CategoricalValue{Any,UInt32}}
```
(have you intended this? - I guess yes, but want to double check)

nalimilan: Ah yes. But I don't think we need to care about other types, for now at least `isordered` is only for CategoricalArrays.

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl

src/abstractdataframe/reshape.jl
bkamins: ok - restricted.

src/abstractdataframe/reshape.jl

test/reshape.jl
bkamins: This is a `view=true` test, so it will be `RepeatedArray` in all cases.

src/abstractdataframe/reshape.jl
bkamins: good point. Fixed

test/reshape.jl
```diff
@@ -386,4 +396,64 @@ end
...
+    d1s = stack(d1, [:d, :c], variable_eltype=Symbol)
+    @test d1s.variable isa Vector{Symbol}
+    @test levels(d1s.variable) == [:c, :d]
+    d1s = stack(d1, [:d, :c], view=true, variable_eltype=Symbol)
+    @test d1s.variable isa DataFrames.RepeatedVector{Symbol}
+    @test levels(d1s.variable) == [:c, :d]
+    @test d1s[:, 1] isa Vector{Symbol}
+    @test levels(d1s[:, 1]) == [:c, :d]
+
+    d2 = categorical(d1, :)
```
bkamins: I have applied all your comments - thank you. In particular here I have added the test we were talking about (I hope I understood you correctly what you wanted us to test here?)

nalimilan: Yes, that's it. Though it would be even better if you tested that when columns have the same levels their order is preserved (as well as `isordered`).

bkamins: added. Note though that `:value` column is in this case merging the original levels (as it combines two vectors in our case), but I think it is OK (normally `:value` will not be categorical, but I just wanted to make sure we are OK in this case also).

nalimilan: Hmm, but if you keep levels of `:c` and `:d` in their default order and without marking them as ordered, you can't check that order is preserved, right? My idea was that you can make both variables use the same levels (including unused ones).

bkamins: I am not fully sure what you mean by "you can't check that order is preserved" and by "both variables use the same levels (including unused ones)" (in the output we have a single column.).

what `stack` does essentially is that it calls `vcat(d2.c, d2.d)` and all that happens is exactly what `vcat` provides. I think that if you feel that we do not get what we like here then `vcat` implementation for categorical arrays should be changed - not the code of `stack` (but maybe I misunderstand something).

In particular note that if you `vcat` two ordered categorical arrays they become unordered and their levels are just added sequentially:
```
julia> x = categorical(1:3, ordered=true)
3-element CategoricalArray{Int64,1,UInt32}:
 1
 2
 3

julia> y = categorical(4:6, ordered=true)
3-element CategoricalArray{Int64,1,UInt32}:
 4
 5
 6

julia> isordered(vcat(x, y))
false

julia> levels(vcat(x, y))
6-element Array{Int64,1}:
 1
 2
 3
 4
 5
 6

julia> levels(vcat(y, x))
6-element Array{Int64,1}:
 4
 5
 6
 1
 2
 3
```
What other behavior would you expect here?

nalimilan: I was thinking about something like this:
```julia
julia> x = levels!(categorical(1:3, ordered=true), collect(6:-1:1))
3-element CategoricalArray{Int64,1,UInt32}:
 1
 2
 3

julia> y = levels!(categorical(4:6, ordered=true), collect(6:-1:1))
3-element CategoricalArray{Int64,1,UInt32}:
 4
 5
 6

julia> isordered(vcat(x, y))
true

julia> levels(vcat(x, y))
6-element Array{Int64,1}:
 6
 5
 4
 3
 2
 1
```

That's just the behavior of `vcat`, but it may still be worth checking since in other places we use other tricks to allocate an array and fill it manually, which may break levels.

bkamins: sure - done! Thank you.

bkamins:
Thank you. Once CI passes how do you think we should move forward, since it is breaking (so it should go into 0.21 but not 0.20.x releases). What is the best practice?

Shall we merge this to master and just later if we have bug fixes then we will merge them to master + create a patch branch? I think it would be good to have a clear policy (possibly written down somewhere in the documentation), as after 1.0 release we will have a much slower (probably ~1 year) turnover with introduction of breaking changes.

nalimilan:
Yeah, let's just merge and see later whether we need to backport some things in a separate branch later. After 1.0 we'll probably have to avoid merging breaking things into master (just like Julia does).

bkamins:
OK - I have merged this as we already have some other breaking PRs over 0.20 release + it seems we got no signals from the users about problems with 0.20.

PR #2050, opened 2019-12-09
Split groupeddataframe/grouping.jl into several files

jlumpe:
This file is pretty massive at 1500 lines and it was getting difficult to navigate through. I've split it logically into several files, similar to the organization of `src/abstractdataframe/`:

* `groupeddataframe.jl` - Type definition and basic methods like `length`, `names`, and `==`.
* `groupkeys.jl` - Definitions for `GroupKey`, `GroupKeys`, and their methods.
* `getindex.jl` - All methods of `getindex(::GroupedDataFrame, ...)` and other functions related to indices/keys.
* `splitapplycombine.jl` - `groupby`, `map`, `combine`, `by`, and `aggregate`.


bkamins:
@nalimilan is maintaining split-apply-combine part of the codebase, so probably it is best if he comments if this is feasible (I guess he has some projects changing this part of code in paralel and such a change is seriously breaking).

In the worst case we can leave this proposal as a post 1.0 release change (as it is planned in general that we would refactor the codebase after 1.0 release to make it more consistent and manageable).

jlumpe:
`splitapplycombine.jl` actually still contains over 1000 lines and should probably be broken up further, but I'm not familiar enough with that part of the code. The upside is that git treats it a renaming of the original file so hopefully any changes to it should be able to be merged automatically.

nalimilan:
Why not do this in theory, but I'd rather wait until https://github.com/JuliaData/DataFrames.jl/pull/2047 is merged as rebasing over moving code is really a nightmare (I don't think git is that smart). Regarding the details, I'd say getindex.jl is too short to deserve its own file: better put its contents in groupeddataframe.jl.

Also as @bkamins noted at https://github.com/JuliaData/DataFrames.jl/pull/2047#pullrequestreview-328952646 we should also move some code out of dataframerow/utils.jl, which really isn't the best place to have it.

jlumpe:
Saw #2047 got merged, rebased on top of it. @nalimilan I've kept all the indexing stuff in the main file for now, but #2406 is going to add a significant amount to it. It also depends on the definitions of `GroupKey` and `GroupKeys` so those can't be put into a different file if the indexing stuff isn't.

bkamins:
Thank you!

PR #2052, opened 2019-12-14
A small map documentation cleanup

bkamins:
I am not sure I proposed the best wording, but the old documentation was using "data frame" in several places that I felt was confusing.


src/groupeddataframe/grouping.jl
```diff
@@ -520,24 +520,27 @@ a vector or tuple thereof, and `f` must be a callable. If `cols` is a single col
...
 If the first argument is a vector, tuple or named tuple of such pairs, each pair is
-handled as described above. If a named tuple, field names are used to name
+handled as described above. If it is a named tuple then field names are used to name
 each generated column.
 
 If the first argument is a callable `f`, it is passed a [`SubDataFrame`](@ref) view for each group,
-and the returned `DataFrame` then consists of the returned rows plus the grouping columns.
-If the returned data frame contains columns with the same names as the grouping columns,
+and the returned `GroupedDataFrame` then consists of the returned rows plus the grouping columns.
+If the value returned  by `f` contains columns with the same names as the grouping columns,
```
nalimilan: I think this behavior is only supported when an `AbstractDataFrame` is returned, right? So I'd say:
```suggestion
If the data frame returned  by `f` contains columns with the same names as the grouping columns,
```

src/groupeddataframe/grouping.jl
```diff
@@ -520,24 +520,27 @@ a vector or tuple thereof, and `f` must be a callable. If `cols` is a single col
...
 
-`f` can return a single value, a row or multiple rows. The type of the returned value
-determines the shape of the resulting data frame:
-- A single value gives a data frame with a single column and one row per group.
-- A named tuple of single values or a [`DataFrameRow`](@ref) gives a data frame with one column
+`f` can return a single value, a row or multiple rows.
+In all cases, the resulting `GroupedDataFrame` contains all the grouping columns in addition
+to those generated by the application of `f`.
+The type of the returned value determines the shape of the resulting groups in `GroupedDataFrame`.
+Additional columns in groups are created using the following rules:
```
nalimilan: I"m not sure repeating "in groups" really helps.
```suggestion
Additional columns are created using the following rules:
```

src/groupeddataframe/grouping.jl
```diff
@@ -548,8 +551,6 @@ faster than returning a data frame.
 As a special case, if a tuple or vector of pairs is passed as the first argument, each function
 is required to return a single value or vector, which will produce each a separate column.
 
-In all cases, the resulting `GroupedDataFrame` contains all the grouping columns in addition
```
nalimilan: I think this sentence is also used for `combine`: maybe apply the same move there?

src/groupeddataframe/grouping.jl
bkamins: I have said "value" because it does not have to be a data frame. It also can be a named tuple of vectors.

src/groupeddataframe/grouping.jl
bkamins: OK - I will sync them when we settle the comment I have left above regarding the return value.

src/groupeddataframe/grouping.jl
nalimilan: OK. Maybe "table" or "table object"? I tend to think of "value" as more or less implying "scalar".

bkamins: fixed

bkamins:
I have updated `by` and `combine` as discussed.

src/groupeddataframe/grouping.jl
```diff
@@ -520,24 +520,27 @@ a vector or tuple thereof, and `f` must be a callable. If `cols` is a single col
...
 
-`f` can return a single value, a row or multiple rows. The type of the returned value
-determines the shape of the resulting data frame:
-- A single value gives a data frame with a single column and one row per group.
-- A named tuple of single values or a [`DataFrameRow`](@ref) gives a data frame with one column
+`f` can return a single value, a row or multiple rows.
+In all cases, the resulting `GroupedDataFrame` contains all the grouping columns in addition
+to those generated by the application of `f`.
+The type of the returned value determines the shape of the resulting groups in `GroupedDataFrame`.
+Additional columns are created using the following rules:
```
nalimilan: Actually it's not just columns: the number of rows of the result also depends on what `f` returns. That's why the old text referred to "shape".

bkamins: OK. Fixed.

bkamins:
Looks good. Thank you.

PR #2055, opened 2019-12-15
Add Tables.jl interface for DataFrame(Rows|Columns)

tkf:
This PR adds Tables.jl interface for `DataFrameRows` and `DataFrameColumns`.  It is useful for defining data manipulation functions expecting iterators.  Example:

```julia
julia> tablemap(f, xs) = Tables.materializer(xs)(map(f, xs))
tablemap (generic function with 1 method)

julia> tablemap(x -> (A=x.a, B=x.b), eachrow(DataFrame(a=[3], b=[4])))
1×2 DataFrame
│ Row │ A     │ B     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 3     │ 4     │
```

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -48,17 +48,14 @@ DataFrame!(x::Vector{<:NamedTuple}) =
                         "`$(typeof(x))` without allocating new columns: use " *
                         "`DataFrame(x)` instead"))
 
-for T in [DataFrameRows, DataFrameColumns]
-    @eval begin
-        Tables.istable(::Type{<:$T}) = true
-        Tables.columnaccess(::Type{<:$T}) = true
-        Tables.rowaccess(::Type{<:$T}) = true
-        Tables.columns(itr::$T) = Tables.columns(parent(itr))
-        Tables.rows(itr::$T) = Tables.rows(parent(itr))
-        Tables.schema(itr::$T) = Tables.schema(parent(itr))
-        Tables.materializer(itr::$T) = Tables.materializer(parent(itr))
-    end
-end
+Tables.istable(::Type{<:Union{DataFrameRows,DataFrameColumns}}) = true
+Tables.columnaccess(::Type{<:Union{DataFrameRows,DataFrameColumns}}) = true
+Tables.rowaccess(::Type{<:Union{DataFrameRows,DataFrameColumns}}) = true
+Tables.columns(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.columns(parent(itr))
+Tables.rows(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.rows(parent(itr))
+Tables.schema(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.schema(parent(itr))
+Tables.materializer(itr::Union{DataFrameRows,DataFrameColumns}) =
+    Tables.materializer(parent(itr))
 
 IteratorInterfaceExtensions.getiterator(df::AbstractDataFrame) = Tables.datavaluerows(df)
 IteratorInterfaceExtensions.isiterable(x::AbstractDataFrame) = true
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -48,6 +48,15 @@ DataFrame!(x::Vector{<:NamedTuple}) =
                         "`$(typeof(x))` without allocating new columns: use " *
                         "`DataFrame(x)` instead"))
 
+const _DataFrameIter = Union{DataFrameRows, DataFrameColumns}
+Tables.istable(::Type{<:_DataFrameIter}) = true
+Tables.columnaccess(::Type{<:_DataFrameIter}) = true
+Tables.rowaccess(::Type{<:_DataFrameIter}) = true
+Tables.columns(itr::_DataFrameIter) = Tables.columns(getfield(itr, :df))
+Tables.rows(itr::_DataFrameIter) = Tables.rows(getfield(itr, :df))
+Tables.schema(itr::_DataFrameIter) = Tables.schema(getfield(itr, :df))
+Tables.materializer(itr::_DataFrameIter) = Tables.materializer(getfield(itr, :df))
+
 IteratorInterfaceExtensions.getiterator(df::AbstractDataFrame) = Tables.datavaluerows(df)
 IteratorInterfaceExtensions.isiterable(x::AbstractDataFrame) = true
 TableTraits.isiterabletable(x::AbstractDataFrame) = true
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -54,8 +54,16 @@ Tables.rowaccess(::Type{<:Union{DataFrameRows,DataFrameColumns}}) = true
 Tables.columns(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.columns(parent(itr))
 Tables.rows(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.rows(parent(itr))
 Tables.schema(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.schema(parent(itr))
-Tables.materializer(itr::Union{DataFrameRows,DataFrameColumns}) =
-    Tables.materializer(parent(itr))
+Tables.materializer(itr::DataFrameRows) =
+    eachrow ∘ prefer_singleton_callable(Tables.materializer(parent(itr)))
+Tables.materializer(itr::DataFrameColumns) =
+    eachcol ∘ prefer_singleton_callable(Tables.materializer(parent(itr)))
+
+# A hack to workaround the type-instability of `∘`:
+prefer_singleton_callable(::Type{T}) where T = SingletonCallable{T}()
+prefer_singleton_callable(f) = f
+struct SingletonCallable{T} end
+(::SingletonCallable{T})(x) where T = T(x)
 
 IteratorInterfaceExtensions.getiterator(df::AbstractDataFrame) = Tables.datavaluerows(df)
 IteratorInterfaceExtensions.isiterable(x::AbstractDataFrame) = true
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -53,10 +53,10 @@ for T in [DataFrameRows, DataFrameColumns]
         Tables.istable(::Type{<:$T}) = true
         Tables.columnaccess(::Type{<:$T}) = true
         Tables.rowaccess(::Type{<:$T}) = true
-        Tables.columns(itr::$T) = Tables.columns(getfield(itr, :df))
-        Tables.rows(itr::$T) = Tables.rows(getfield(itr, :df))
-        Tables.schema(itr::$T) = Tables.schema(getfield(itr, :df))
-        Tables.materializer(itr::$T) = Tables.materializer(getfield(itr, :df))
+        Tables.columns(itr::$T) = Tables.columns(parent(itr))
+        Tables.rows(itr::$T) = Tables.rows(parent(itr))
+        Tables.schema(itr::$T) = Tables.schema(parent(itr))
+        Tables.materializer(itr::$T) = Tables.materializer(parent(itr))
     end
 end
 
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -56,8 +56,14 @@ Tables.rows(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.rows(parent(itr
 Tables.schema(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.schema(parent(itr))
 Tables.materializer(itr::DataFrameRows) =
     eachrow ∘ prefer_singleton_callable(Tables.materializer(parent(itr)))
-Tables.materializer(itr::DataFrameColumns) =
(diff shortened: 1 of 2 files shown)


src/other/tables.jl
```diff
@@ -48,6 +48,15 @@ DataFrame!(x::Vector{<:NamedTuple}) =
                         "`$(typeof(x))` without allocating new columns: use " *
                         "`DataFrame(x)` instead"))
 
+const _DataFrameIter = Union{DataFrameRows, DataFrameColumns}
```
bkamins: Intuitively I would prefer to use `@eval` here on two types, as then we do not have to introduce an unexported union that users might not understand when using `methods` function.

bkamins:
In general - this should not be problematic to add these definitions. Can you please elaborate in what cases you would find this useful (the issue is that you can always call `parent` on an iterator and get the data frame to use with Tables.jl).

src/other/tables.jl
```diff
@@ -48,6 +48,15 @@ DataFrame!(x::Vector{<:NamedTuple}) =
...
+Tables.istable(::Type{<:_DataFrameIter}) = true
+Tables.columnaccess(::Type{<:_DataFrameIter}) = true
+Tables.rowaccess(::Type{<:_DataFrameIter}) = true
+Tables.columns(itr::_DataFrameIter) = Tables.columns(getfield(itr, :df))
```
bkamins: It is better use `parent` than `getfield` here and below I think.

tkf:
Thanks for the review!

> Can you please elaborate in what cases you would find this useful (the issue is that you can always call `parent` on an iterator and get the data frame to use with Tables.jl).

I need this interface to make this test for [`Transducers.copy`](https://tkf.github.io/Transducers.jl/dev/manual/#Base.copy) pass: https://github.com/tkf/Transducers.jl/pull/104/files#diff-da78dddbbd001af539780c4952b0e9fcR23-R27

I can't use `parent` in general as I want to process wrapper types as-is, e.g., `SubArray`-of-`StructVector`.

test/tables.jl
```diff
@@ -182,7 +182,7 @@ end
 
      df2 = DataFrame!(eachrow(df))
      @test df == df2
-     @test !any(((a,b),) -> a === b, zip(eachcol(df), eachcol(df2)))
+     @test all(((a,b),) -> a === b, zip(eachcol(df), eachcol(df2)))
```
tkf: With this PR, `df2 = DataFrame!(eachrow(df))` does not copy columns any more. Is it an OK change?

bkamins: I would say that it is even better. @nalimilan - are you OK with this?
If someone writes `DataFrame!` an explicit opt-out from copying, if possible, is assumed.

src/other/tables.jl
```diff
@@ -48,6 +48,18 @@ DataFrame!(x::Vector{<:NamedTuple}) =
                         "`$(typeof(x))` without allocating new columns: use " *
                         "`DataFrame(x)` instead"))
 
+for T in [DataFrameRows, DataFrameColumns]
+    @eval begin
```
nalimilan: Do you really need a loop? Isn't `::Union{DataFrameRows, DataFrameColumns}` enough?

test/tables.jl
nalimilan: Yes, makes sense.

src/other/tables.jl
nalimilan: Ah, I hadn't seen @bkamins's comment above. I'd just repeat the `Union` without defining a custom type alias.

tkf: I'm OK with both approaches. I wrote this as @bkamins preferred this approach. Ref: https://github.com/JuliaData/DataFrames.jl/pull/2055#discussion_r358008369

nalimilan:
Shouldn't `materializer` return a `DataFramesRows`/`DataFramesColumns` object rather than a `DataFrame`?

tkf:
My understanding is that `materializer(x)(input) isa typeof(x)` does not have to hold.  For example, there is a [fallback definition `materializer(x) = columntable`](https://github.com/JuliaData/Tables.jl/blob/de50b30f7e393e41d56d427511c2a5ca2831d22b/src/Tables.jl#L102) which creates a `NamedTuple`-of-`Vector`s.

src/other/tables.jl
bkamins: I am OK with both - the @eval approach is used in Base often. But using `Union` without defining the alias is also OK (I just prefer not to introduce the alias here).

nalimilan:
Maybe it doesn't have to, but should it? Or why shouldn't it? :-)

tkf:
Actually `materializer(::DataFramesRows) = DataFramesRows ∘ DataFrame` makes sense.

bkamins:
Well `materializer` is @quinnj's idea, so probably he can comment 😄. When I analyze it (it does not have a documentation) I get the understanding that it should **materialize** `x`, and `DataFrameRows` and `DataFrameColums` are views, while `DataFrame` is a materialized object (but maybe this is a wrong way to look at it - @quinnj can you please comment here what is the design purpose of `materialize`).

bkamins:
> Actually `materializer(::DataFramesRows) = DataFramesRows ∘ DataFrame` makes sense.

As I have commented - whether it makes sense depends on the contract @quinnj wants `materialize` to support I think (I was writing in parallel).

tkf:
@bkamins I just noticed your comment after writing a patch to do (something like) `materializer(::DataFramesRows) = eachrow ∘ DataFrame`.  But I agree waiting for @quinnj's comment is a good idea.  If returning a `DataFrame` is a better idea, I'll just remove that commit.

The reason why I thought returning a `DataFramesRows` make sense was that we can then have `tablemap(f ∘ g, xs) == tablemap(f, tablemap(g, xs)` which is a nice property.

src/other/tables.jl
tkf: Does https://github.com/JuliaData/DataFrames.jl/pull/2055/commits/22e32dbe445bf3ce1e565cee0f56e9d2a5275197 look good?

src/other/tables.jl
```diff
@@ -48,6 +48,23 @@ DataFrame!(x::Vector{<:NamedTuple}) =
...
+Tables.istable(::Type{<:Union{DataFrameRows,DataFrameColumns}}) = true
+Tables.columnaccess(::Type{<:Union{DataFrameRows,DataFrameColumns}}) = true
+Tables.rowaccess(::Type{<:Union{DataFrameRows,DataFrameColumns}}) = true
+Tables.columns(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.columns(parent(itr))
+Tables.rows(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.rows(parent(itr))
+Tables.schema(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.schema(parent(itr))
+Tables.materializer(itr::DataFrameRows) =
+    eachrow ∘ prefer_singleton_callable(Tables.materializer(parent(itr)))
+Tables.materializer(itr::DataFrameColumns) =
+    eachcol ∘ prefer_singleton_callable(Tables.materializer(parent(itr)))
```
bkamins: If @quinnj comments that we should return here `DataFrameColumns` and not `DataFrame` then we should inherit from `itr` if it was created with `names` positional argument set to `true` or `false`.

quinnj:
We don't currently have strict requirements on `materializer`; I think, as mentioned, it's preferable if it returns the same type as the input, but I don't think that's always feasible (immutable or view-like inputs, certain DB or file-based format tables, etc.).

These changes seem fine by me, though I will note that I recently tried to use `DataFrameRows` and wanted to open an issue up about it's useability: namely the fact that it doesn't inherit from `<: AbstractDataFrame` means things like `names(df_rows)` don't work, but would be nice if they did.

Anyway, I'm good w/ this.

bkamins:
> namely the fact that it doesn't inherit from <: AbstractDataFrame means things like names(df_rows) don't work, but would be nice if they did.

It cannot because it inherits from `AbstractVector` and we do not have multiple inheritance. I will open a PR adding common `AbstractDataFrame` methods for them.

src/other/tables.jl
```diff
@@ -48,6 +48,29 @@ DataFrame!(x::Vector{<:NamedTuple}) =
...
+function Tables.materializer(itr::DataFrameColumns)
+    f = prefer_singleton_callable(Tables.materializer(parent(itr)))
+    if eltype(itr) <: Pair
+        return x -> eachcol(f(x), true)
+    else
+        return x -> eachcol(f(x), false)
+    end
+end
+
+# A hack to workaround the type-instability of `∘`:
```
nalimilan: Can you explain why this is needed?

tkf: I illustrate this in https://github.com/JuliaLang/julia/issues/23618#issuecomment-566130606

nalimilan: OK. Though a simpler solution would have been to hardcode `DataFrame` as the materializer (since that's the definition we have above).

It would be nice to at least add a comment pointing to the issue, otherwise it's quite hard to guess without context (especially if one day this gets fixed).

tkf: > Though a simpler solution would have been to hardcode `DataFrame` as the materializer (since that's the definition we have above).

I thought somebody may have their AbstractDataFrame subtype but I guess I didn't need to worry about it.

nalimilan: In theory yes, in practice it's relatively unlikely, and anyway they could just provide their custom method.

tkf: Please see https://github.com/JuliaData/DataFrames.jl/pull/2057

tkf: https://github.com/JuliaData/DataFrames.jl/pull/2057

tkf:
Actually, can we go back to `materializer(::DataFramesRows) = DataFrame`? :sweat_smile:

> When I analyze it (it does not have a documentation) I get the understanding that it should **materialize** `x`, and `DataFrameRows` and `DataFrameColums` are views, while `DataFrame` is a materialized object

@bkamins I just realized that your point totally makes sense.  I needed something I can `push!`/`append!` to.  At the moment, `DataFramesRows` does not support this and I'm guessing DataFrames devs don't want to expand the API surface.  If that's the case, can we go back to the original `materializer` definition?

Sorry, I should've tried to re-implement https://github.com/tkf/Transducers.jl/pull/107 after the API was changed...

bkamins:
> I'm guessing DataFrames devs don't want to expand the API surface

It is a simple rule: both objects are `view`s so they should not be allowed to mutate the schema of the `parent` (i.e. changing numer of rows or columns or renaming them).

So I understand the change should be made to make `materialize` produce a `DataFrame` - right.?

Also just to confirm. This new `DataFrame` should be a copy of the source (not reuse the columns) - right?

nalimilan:
I guess returning a `DataFrame` would be OK if `materializer` is supposed to return a mutable table (like `similar` for arrays), but we would have to make this clear. Otherwise it sounds weird not to return `DataFrameRows`.

tkf:
> It is a simple rule: both objects are `view`s so they should not be allowed to mutate the schema of the `parent` (i.e. changing numer of rows or columns or renaming them).

The "rule" sounds arbitrary to me, in the sense that you made it so you can change it.  I can't find any explicit API contracts defined for `eachrows` and `eachcols`.  I agree that the function name sounds like that the caller should not expect more than iterator API.  However, if `DataFrameRows` is also used for `Tables.rows` #2051, it may makes sense to adding mutation support.  For example, Tables.jl mentions that `Tables.rows(table) === table` is a reasonable choice if `table` is already a row iterator.  This includes a `table` type that support `push!`.

bkamins:
I see your point, so let us wait for other to comment about their preference.

My thinking was that `DtaFrameRows` for `AbstractDataFrame` is like `(v for x in x)` in Base.

Also note that `DataFrameRows` might wrap a `SubDataFrame` in which case it will not be resizable anyway, as `SubDataFrame` is not resizable.

nalimilan:
@quinnj Do you think `materializer` is supposed to return a mutable table, i.e. to which rows or columns can be added?

quinnj:
I think `materializer` should return the most well-supported, "standard" table type for a family of table types. The point is that some higher-order table processor somewhere wants to manipulate a table and spit back out something like what you put in; so while it definitely makes sense to return the _exact_ same type as input, I also think for general useability, it makes sense to return the most "standard" type, in this case a `DataFrame`.

nalimilan:
Yeah it's a tough choice. It's not ideal that if you passed a table which iterates rows (`DataFrameRows`) you get a table that cannot be iterated over (`DataFrame`). Though I guess you should call `Tables.rows` on the result anyway if you want to iterate over rows, so returning a `DataFrame` should be OK.

tkf:
@bkamins @nalimilan @quinnj Thanks for the discussion! I opened #2058

bkamins:
I would prefer in the future to squash-merge PRs, as otherwise it is hard for me to write appropriate release notes. Thank you!

quinnj:
Huh......I think I did a rebase-merge on another repo and then it must remember your latest preference regardless of repo? I almost always squash, but there was a case on another repo where I wanted to rebase-merge and then I must have just hit the default here. Sorry about that.

PR #2056, opened 2019-12-16
Add names to data frame iterators

bkamins:
Following https://github.com/JuliaData/DataFrames.jl/pull/2055#issuecomment-566225609. Actually it seems that only `names` was missing for `eachcol` and `eachrow` return values.


bkamins:
I have come to the conclusion that `propertynames` should always return a `Tuple` in DataFrames.jl. This is mildly breaking but I think we do not need deprecation for this (but because this is a decision to be made for now I have not updated the tests of this functionality).

Adding 1.0 tag, as this changes the API.

bkamins:
The point is that if you want a vector you use `names` and if you want a `Tuple` you use `propertynames` so this is consistent and both options are always available.

PR #2057, opened 2019-12-16
Simplify materializer for DataFrameRows and DataFrameColumns

tkf:
Address https://github.com/JuliaData/DataFrames.jl/pull/2055#discussion_r358445036

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -55,21 +55,18 @@ Tables.columns(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.columns(pare
 Tables.rows(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.rows(parent(itr))
 Tables.schema(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.schema(parent(itr))
 Tables.materializer(itr::DataFrameRows) =
-    eachrow ∘ prefer_singleton_callable(Tables.materializer(parent(itr)))
+    x -> eachrow(Tables.materializer(parent(itr))(x))
 function Tables.materializer(itr::DataFrameColumns)
-    f = prefer_singleton_callable(Tables.materializer(parent(itr)))
     if eltype(itr) <: Pair
-        return x -> eachcol(f(x), true)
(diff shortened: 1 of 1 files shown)


src/other/tables.jl
```diff
@@ -55,21 +55,18 @@ Tables.columns(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.columns(pare
 Tables.rows(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.rows(parent(itr))
 Tables.schema(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.schema(parent(itr))
 Tables.materializer(itr::DataFrameRows) =
-    eachrow ∘ prefer_singleton_callable(Tables.materializer(parent(itr)))
+    x -> eachrow(Tables.materializer(parent(itr))(x))
 function Tables.materializer(itr::DataFrameColumns)
-    f = prefer_singleton_callable(Tables.materializer(parent(itr)))
```
bkamins: Why not just have two methods for `DataFrameColumns` instead of one method with `eltype` check? But this is a mild preference (maybe wait for @nalimilan to comment before changing this 😄)

nalimilan: Yes that would sound more logical.

src/other/tables.jl
```diff
@@ -55,21 +55,18 @@ Tables.columns(itr::Union{DataFrameRows,DataFrameColumns}) = Tables.columns(pare
...
+        return x -> eachcol(Tables.materializer(parent(itr))(x), false)
     end
 end
-
-# A hack to workaround the type-instability of `∘`:
-prefer_singleton_callable(::Type{T}) where T = SingletonCallable{T}()
-prefer_singleton_callable(f) = f
-struct SingletonCallable{T} end
-(::SingletonCallable{T})(x) where T = T(x)
+# Note: Don't try to factor `Tables.materializer(...)` out from the
```
nalimilan: I'd rather keep this out: if we added a comment everytime somebody could possibly do something wrong... in some parts of the code there would be more comments than code. :-)

PR #2058, opened 2019-12-19
Materialize DataFrame(Rows|Columns) as DataFrame

tkf:
This PR reverts last two commits in #2055 so that

```julia
materializer(::Union{DataFrameRows,DataFrameColumns}) = DataFrame
```

Ref: https://github.com/JuliaData/DataFrames.jl/pull/2055#issuecomment-566421559

### DIFF
--- a/test/tables.jl
+++ b/test/tables.jl
@@ -62,7 +62,7 @@ Base.propertynames(d::DuplicateNamesColumnTable) = (:a, :a, :b)
         @test Tables.columnaccess(table)
         @test Tables.schema(table) === Tables.Schema((:a, :b), Tuple{Int64, Symbol})
         @test Tables.schema(table) == Tables.schema(Tables.rows(table)) == Tables.schema(Tables.columns(table))
-        @test @inferred(Tables.materializer(table)(Tables.columns(table))) isa typeof(df)
+        @test @inferred(Tables.materializer(table)(Tables.columns(table))) isa DataFrame
 
         row = first(Tables.rows(table))
         @test propertynames(row) == (:a, :b)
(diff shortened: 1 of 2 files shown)


bkamins:
@tkf - does this PR mean that #2057 should be closed? (I guess yes, but I want to be sure)

tkf:
@bkamins Yes, I forgot to close it.

test/tables.jl
```diff
@@ -56,18 +56,13 @@ Base.propertynames(d::DuplicateNamesColumnTable) = (:a, :a, :b)
...
-        eachcol(df, true),
-    ]
+    @testset "basics $(nameof(typeof(table)))" for table in [df, eachrow(df), eachcol(df)]
         @test Tables.istable(table)
         @test Tables.rowaccess(table)
         @test Tables.columnaccess(table)
         @test Tables.schema(table) === Tables.Schema((:a, :b), Tuple{Int64, Symbol})
         @test Tables.schema(table) == Tables.schema(Tables.rows(table)) == Tables.schema(Tables.columns(table))
-        @test @inferred(Tables.materializer(table)(Tables.columns(table))) isa typeof(table)
+        @test @inferred(Tables.materializer(table)(Tables.columns(table))) isa typeof(df)
```
bkamins:
```suggestion
        @test @inferred(Tables.materializer(table)(Tables.columns(table))) isa DataFrame
```

bkamins: A small fix. I would pass `DataFrame` explicitly. Also I propose to make an outer loop passing through `[df, view(df, :, :)]` vector. The point is that `Tables.materializer` produces `DataFrame` also for `SubDataFrame`. I think we should test for this here (actually this is relevant wrt. our earlier discussion, as `SubDataFrame` is not resizable).

bkamins:
Thank you!

PR #2060, opened 2019-12-20
add conversions and merge for DataFrameRow

### LINKED ISSUE #2059: DataFrameRow should be more NamedTuple like
DataFrameRow is basically a mutable `NamedTuple` it should support:
 - splatting into kwargs positions `foo(x; row...)`
 - splatting into named tuple literal `(; row..., a=1, b=2)`
 - Converting into `NamedTuple`
 - `merge(row, nt)`, `merge(row, row2)`, `merge(nt, row)`
 - `pairs(row)`

cc @willtebbutt

bkamins:
Fixes #2059.

Notes:
1) I did not add docstrings for these functionalities following the logic that we treat `DataFrameRow` as if it is a `NamedTuple`
2) I skipped defining `merge` for more than 2 passed arguments as this is not supported in 1.0 (it is new in 1.1) and additionally it is problematic (there is no clean way to define `merge` with `Union` approach as it overwrites the original `merge(a::NamedTuple, b::NamedTuple, cs::NamedTuple...)` definition which I think we should avoid); I think thought that in practice this is not needed
3) in tests I check to make sure splatting as kwargs works as expected for `DataFrameRow`

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -205,7 +205,7 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
-Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(dfr))
+Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(dfr))}(values(dfr))
 
 """
     copy(dfr::DataFrameRow)
@@ -215,7 +215,7 @@ This method returns a `NamedTuple` so that the returned object
 is not affected by changes to the parent data frame of which `dfr` is a view.
 
 """
-Base.copy(r::DataFrameRow) = NamedTuple(r)
+Base.copy(dfr::DataFrameRow) = NamedTuple(dfr)
 
 Base.convert(::Type{NamedTuple}, dfr::DataFrameRow) = NamedTuple(dfr)
 Base.convert(::Type{Tuple}, dfr::DataFrameRow) = Tuple(dfr)
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -205,24 +205,24 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
+Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(r))
+
 """
     copy(dfr::DataFrameRow)
 
 Convert a [`DataFrameRow`](@ref) to a `NamedTuple`.
 """
-Base.copy(r::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(r))
+Base.copy(r::DataFrameRow) = NamedTuple(r)
 
-Base.NamedTuple(dfr::DataFrameRow) = copy(dfr)
-
-Base.convert(::Type{NamedTuple}, dfr::DataFrameRow) = copy(dfr)
+Base.convert(::Type{NamedTuple}, dfr::DataFrameRow) = NamedTuple(dfr)
 Base.convert(::Type{Tuple}, dfr::DataFrameRow) = Tuple(dfr)
 
-Base.merge(a::DataFrameRow) = copy(a)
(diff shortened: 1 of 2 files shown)


src/dataframerow/dataframerow.jl
```diff
@@ -212,6 +212,18 @@ Convert a [`DataFrameRow`](@ref) to a `NamedTuple`.
 """
 Base.copy(r::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(r))
 
+Base.NamedTuple(dfr::DataFrameRow) = copy(dfr)
```
oxinabox: I am tempted to say that this should be considered the primary function,
and `copy` should call this.

While I find the _its a view, `copy` materializes a view_ not completely unreasonable (but maybe it shoukd be in the docstring),
this feels like the main operation.

bkamins: `copy` has a docstring. I will change the rest.

oxinabox: Yeah, but copy's doc string should say why it is returning a named tuple.

bkamins: OK - can you please propose something that would be appropriate (I have learned that native speakers are usually better at writing doc stings 😄; thank you)

src/dataframerow/dataframerow.jl
```diff
@@ -205,12 +205,24 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
+Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(r))
+
 """
     copy(dfr::DataFrameRow)
 
 Convert a [`DataFrameRow`](@ref) to a `NamedTuple`.
```
oxinabox:
```suggestion
Constucts a `NamedTuple` with the same content as the [`DataFrameRow`](@ref).
A  [`DataFrameRow`](@ref) is a view into a row of a  [`DataFrame`](@ref), following the convention that `copy` materializes the view into a independent concrete object, this thus creates a `NamedTuple`.

```

src/dataframerow/dataframerow.jl
```diff
@@ -205,12 +205,26 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
...
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
+Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(r))
+
 """
     copy(dfr::DataFrameRow)
 
-Convert a [`DataFrameRow`](@ref) to a `NamedTuple`.
+Constucts a `NamedTuple` with the same content as the [`DataFrameRow`](@ref).
+A  [`DataFrameRow`](@ref) is a view into a row of a  [`DataFrame`](@ref), following the convention that `copy` materializes the view into a independent concrete object, this thus creates a `NamedTuple`.
```
bkamins:
```suggestion
A  [`DataFrameRow`](@ref) is a view into a row of a  [`DataFrame`](@ref), following
the convention that `copy` materializes the view into an independent concrete object,
this thus creates a `NamedTuple`.
```

src/dataframerow/dataframerow.jl
```diff
@@ -205,12 +205,28 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
+Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(r))
+
 """
     copy(dfr::DataFrameRow)
 
-Convert a [`DataFrameRow`](@ref) to a `NamedTuple`.
+Constucts a `NamedTuple` with the same content as the [`DataFrameRow`](@ref).
```
```suggestion
Construct a `NamedTuple` with the same contents as the [`DataFrameRow`](@ref).
```

src/dataframerow/dataframerow.jl
```diff
@@ -205,12 +205,28 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
...
+A  [`DataFrameRow`](@ref) is a view into a row of a  [`DataFrame`](@ref), following
```
nalimilan: It seems weird to say what is a `DataFrameRow` here. I'm not convinced we need to justify the behavior choice, but if you really want to, maybe this?
```
This method returns a `NamedTuple` so that the returned object
is not affected by changes to the parent data frame of which `dfr` is a view.
```

bkamins: @oxinabox felt that it is worth to justify it. I will update the wording to use your proposal.

bkamins: done

PR #2063, opened 2019-12-23
Fix grouped maximum, minimum, var and std with only missing values

### LINKED ISSUE #2061: Error in `by` when composing with `skipmissing`
On 0.20 I observe the following:

```
julia> using DataFrames, Dates

julia> df = DataFrame(id = [1,1,2,2], date = [Date(2010, 10, 10), missing, missing, missing]);

julia> grouped = by(df, :id, :date => maximum ∘ skipmissing)
2×2 DataFrame
│ Row │ id    │ date_function          │
│     │ Int64 │ Date                   │
├─────┼───────┼────────────────────────┤
│ 1   │ 1     │ 2010-10-10             │
│ 2   │ 2     │ -252522163911150-01-01 │
```

This seems like a bug - as @bkamins pointed out on Slack
```
julia> grouped = by(df, :id, :date => x -> (maximum ∘ skipmissing)(x))
```
gives the expected error with `missing`-only groups

nalimilan:
When a group only contains missing values, grouped `maximum` and `minimum` have to throw an error like the corresponding functions. Otherwise they incorrectly return `typemin`/`typemax`. Also fix `var` and `std` to return `NaN` in that case instead of `-0.0`.

This problem cannot happen without `skipmissing`, as empty groups are not allowed.

Fixes #2061.

### DIFF
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -488,9 +488,14 @@ check_aggregate(::typeof(minimum)) = Reduce(min)
 check_aggregate(::typeof(mean)) = Reduce(Base.add_sum, nothing, /)
 check_aggregate(::typeof(sum∘skipmissing)) = Reduce(Base.add_sum, !ismissing)
 check_aggregate(::typeof(prod∘skipmissing)) = Reduce(Base.mul_prod, !ismissing)
-check_aggregate(::typeof(maximum∘skipmissing)) = Reduce(max, !ismissing)
-check_aggregate(::typeof(minimum∘skipmissing)) = Reduce(min, !ismissing)
 check_aggregate(::typeof(mean∘skipmissing)) = Reduce(Base.add_sum, !ismissing, /)
+# Empty groups can only happen with skipmissing and are only a problem for min and max
+function check_notempty(res, count::Integer)
+    count > 0 || throw(ArgumentError("some groups contain only missing values"))
+    res
+end
+check_aggregate(::typeof(maximum∘skipmissing)) = Reduce(max, !ismissing, check_notempty)
+check_aggregate(::typeof(minimum∘skipmissing)) = Reduce(min, !ismissing, check_notempty)
 
 # Other aggregate functions which are not strictly reductions
 struct Aggregate{F, C} <: AbstractAggregate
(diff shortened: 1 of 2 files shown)


src/groupeddataframe/grouping.jl
```diff
@@ -864,9 +864,14 @@ check_aggregate(::typeof(minimum)) = Reduce(min)
 check_aggregate(::typeof(mean)) = Reduce(Base.add_sum, nothing, /)
 check_aggregate(::typeof(sum∘skipmissing)) = Reduce(Base.add_sum, !ismissing)
 check_aggregate(::typeof(prod∘skipmissing)) = Reduce(Base.mul_prod, !ismissing)
-check_aggregate(::typeof(maximum∘skipmissing)) = Reduce(max, !ismissing)
-check_aggregate(::typeof(minimum∘skipmissing)) = Reduce(min, !ismissing)
 check_aggregate(::typeof(mean∘skipmissing)) = Reduce(Base.add_sum, !ismissing, /)
+# Empty groups can only happen with skipmissing and are only a problem for min and max
+function check_notempty(res, count::Integer)
```
nalimilan: This approach has the side effect of making an unnecessary copy of the result. Maybe I can find something better...

src/groupeddataframe/grouping.jl
```diff
@@ -1002,10 +1007,10 @@ function groupreduce!(res, f, op, condf, adjust,
     end
     outcol = adjust === nothing ? res : map(adjust, res, counts)
     # Undo pool sharing done by groupreduce_init
-    if outcol isa CategoricalVector
+    if outcol isa CategoricalVector && outcol.pool === incol.pool
```
bkamins: I do not see where this "unnecessary" copy is made. It seems that the check you do here is needed. Also it should not be very expensive - right?

nalimilan: This change is actually just a consequence of the fact that a copy is made: the pool can be in a different order after `map`. The unnecessary copy is due to passing `adjust !== nothing`, which induces a call to `map`.

src/groupeddataframe/grouping.jl
nalimilan: I've pushed a new commit which avoids the copy.

src/groupeddataframe/grouping.jl
bkamins: OK. The changes also look good.

bkamins:
Sorry for merging https://github.com/JuliaData/DataFrames.jl/pull/2050. I have forgotten that it will introduce the conflict with this PR. I hope rebasing will not be that hard.

nalimilan:
No worries, I had anticipated that. Actually the diff is so simple that git figured it out automatically!

Issue #1981, opened 2019-10-09
Allow returning DataFrame() or NamedTuple() in combine


bkamins:
Re-posted from Slack.

> I have a function inside a `DataFrame` by that I want to have return rows only in some cases.  When it doesn't return rows I can't just return `nothing` or `DataFrame()` -- I have to explicitly construct a zero-row `DataFrame` with the same columns.  Is there any way to avoid this?

My opinion is that `nothing` is not a good option (it can be a valid value), but returning `DataFrame()` if other groups return `DataFrame`s or `NamedTuple()` if other groups return `NamedTuple`s should be allowed. Essentially `DataFrame()` should be treated as zero-row `DataFrame` with appropriate columns. The same for `NamedTuples`.

@nalimilan - probably you have most to say here. If we decide to do it it should be before 1.0 release I think.

nalimilan:
Makes sense. It shouldn't be too hard, maybe all what's needed is return without doing anything instead of throwing an error when the result has no columns here:
https://github.com/JuliaData/DataFrames.jl/blob/57aec338a4f270c1fb95bdbd3f845b0031fb617b/src/groupeddataframe/grouping.jl#L891-L893
and here:
https://github.com/JuliaData/DataFrames.jl/blob/57aec338a4f270c1fb95bdbd3f845b0031fb617b/src/groupeddataframe/grouping.jl#L810-L812

bkamins:
Thank you for the analysis. I think we can disallow `NamedTuple()` if this is problematic, but it seems it should not (of course this would have to be tested):
```
julia> NamedTuple() isa NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}
true
```

We could simply append nothing if we get an object with no columns.

The thing we would need to change is probably:
https://github.com/JuliaData/DataFrames.jl/blob/57aec338a4f270c1fb95bdbd3f845b0031fb617b/src/groupeddataframe/grouping.jl#L842
and
https://github.com/JuliaData/DataFrames.jl/blob/57aec338a4f270c1fb95bdbd3f845b0031fb617b/src/groupeddataframe/grouping.jl#L922
as it is possible that the first group might happen to be empty (we would have to have a special logic that waits till we hit a value with at least one column).

nalimilan:
[...]

PR #2066, opened 2019-12-26
Allow containers with no columns in combine

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1981

Allow `NamedTuple()`, `DataFrame` and zero columns matrix as return value in `combine` and `map`. They are ignored and drop columns.

Additionally fixes some small issues in tests.

@nalimilan - one potential performance consideration is that now I use `colnames` as `Vector{Symbol}` - do you think it will be problematic?

Also I had to separate `_combine_with_first_row!` and `_combine_with_first!` as separate functions as for some strange reason alternatively in some cases Julia had problems with type inference.

Importantly  - we allow these 0-column containers only for vector-oriented containers. They are disallowed in scalar-oriented containers.

Finally - this PR touches parts of code that I do not know very well, so feel free to comment/fix/add test proposals, as some strange interactions might kick in.

### DIFF
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -201,6 +201,7 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives the same additional columns
   and as many rows for each group as the rows returned for that group.
+  In this case returning a table with no columns drops group.
 
 `f` must always return the same kind of object (as defined in the above list) for
 all groups, and if a named tuple or data frame, with the same fields or columns.
@@ -340,6 +341,7 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives a `DataFrame` with the same
   additional columns and as many rows for each group as the rows returned for that group.
+  In this case returning a table with no columns drops group.
 
 `f` must always return the same kind of object (as defined in the above list) for
 all groups, and if a named tuple or data frame, with the same fields or columns.
@@ -790,7 +792,7 @@ function _combine(f::Any, gd::GroupedDataFrame)
          !isa(firstres, Union{AbstractDataFrame, NamedTuple, DataFrameRow, AbstractMatrix}))
          nms = [Symbol(names(gd.parent)[index(gd.parent)[first(f)]], '_', funname(fun))]
     end
-    valscat = DataFrame(collect(AbstractVector, outcols), collect(Symbol, nms))
+    valscat = DataFrame(collect(AbstractVector, outcols), nms)
     return idx, valscat
 end
 
@@ -818,14 +820,17 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
     let eltys=eltys, n=n # Workaround for julia#15276
         initialcols = ntuple(i -> Tables.allocatecolumn(eltys[i], n), _ncol(first))
     end
-    outcols = _combine_with_first!(first, initialcols, idx, 1, 1, f, gd, incols,
-                                   tuple(propertynames(first)...))
-    idx, outcols, propertynames(first)
+    targetcolnames = collect(Symbol, propertynames(first))
+    outcols = first isa Union{AbstractDataFrame,
+                              NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}} ?
+              _combine_with_first!(first, initialcols, idx, 1, 1, f, gd, incols, targetcolnames) :
+              _combine_with_first_row!(first, initialcols, idx, 1, 1, f, gd, incols, targetcolnames)
+    idx, outcols, targetcolnames
 end
 
 function fill_row!(row, outcols::NTuple{N, AbstractVector},
                    i::Integer, colstart::Integer,
-                   colnames::NTuple{N, Symbol}) where N
+                   colnames::Vector{Symbol}) where N
     if !isa(row, Union{NamedTuple, DataFrameRow})
         throw(ArgumentError("return value must not change its kind " *
                             "(single row or variable number of rows) across groups"))
@@ -856,12 +861,12 @@ function fill_row!(row, outcols::NTuple{N, AbstractVector},
     return nothing
 end
 
-function _combine_with_first!(first::Union{NamedTuple, DataFrameRow},
+function _combine_with_first_row!(first::Union{NamedTuple, DataFrameRow},
                               outcols::NTuple{N, AbstractVector},
                               idx::Vector{Int}, rowstart::Integer, colstart::Integer,
                               f::Any, gd::GroupedDataFrame,
                               incols::Union{Nothing, AbstractVector, NamedTuple},
-                              colnames::NTuple{N, Symbol}) where N
+                              colnames::Vector{Symbol}) where N
     len = length(gd)
     gdidx = gd.idx
     starts = gd.starts
@@ -889,7 +894,7 @@ function _combine_with_first!(first::Union{NamedTuple, DataFrameRow},
                     end
                 end
             end
-            return _combine_with_first!(row, newcols, idx, i, j, f, gd, incols, colnames)
+            return _combine_with_first_row!(row, newcols, idx, i, j, f, gd, incols, colnames)
         end
         idx[i] = gdidx[starts[i]]
     end
@@ -911,7 +916,7 @@ else
 end
 
 function append_rows!(rows, outcols::NTuple{N, AbstractVector},
-                      colstart::Integer, colnames::NTuple{N, Symbol}) where N
+                      colstart::Integer, colnames::Vector{Symbol}) where N
     if !isa(rows, Union{AbstractDataFrame, NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}})
         throw(ArgumentError("return value must not change its kind " *
                             "(single row or variable number of rows) across groups"))
@@ -944,18 +949,37 @@ function _combine_with_first!(first::Union{AbstractDataFrame,
                               idx::Vector{Int}, rowstart::Integer, colstart::Integer,
                               f::Any, gd::GroupedDataFrame,
                               incols::Union{Nothing, AbstractVector, NamedTuple},
-                              colnames::NTuple{N, Symbol}) where N
+                              colnames::Vector{Symbol}) where N
     len = length(gd)
     gdidx = gd.idx
     starts = gd.starts
     ends = gd.ends
     # Handle first group
-    j = append_rows!(first, outcols, colstart, colnames)
-    @assert j === nothing # eltype is guaranteed to match
-    append!(idx, Iterators.repeated(gdidx[starts[rowstart]], _nrow(first)))
+
+    @assert isempty(colnames) == (_ncol(first) == 0) == isempty(outcols)
+    if !isempty(colnames)
+        j = append_rows!(first, outcols, colstart, colnames)
+        @assert j === nothing # eltype is guaranteed to match
+        append!(idx, Iterators.repeated(gdidx[starts[rowstart]], _nrow(first)))
+    end
     # Handle remaining groups
     @inbounds for i in rowstart+1:len
         rows = wrap(do_call(f, gdidx, starts, ends, gd, incols, i))
+        if !(rows isa Union{AbstractDataFrame, NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}})
+            throw(ArgumentError("return value must not change its kind " *
+                                "(single row or variable number of rows) across groups"))
+        end
+        _ncol(rows) == 0 && continue
+        if isempty(colnames)
+            append!(colnames, collect(Symbol, propertynames(rows)))
+            if rows isa AbstractDataFrame
+                eltys = eltype.(eachcol(rows))
+            else
+                eltys = map(eltype, rows)
+            end
+            initialcols = ntuple(i -> Tables.allocatecolumn(eltys[i], 0), _ncol(rows))
+            return _combine_with_first!(rows, initialcols, idx, i, 1, f, gd, incols, colnames)
+        end
         j = append_rows!(rows, outcols, 1, colnames)
         if j !== nothing # Need to widen column type
             local newcols
@@ -1021,6 +1045,7 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives a `DataFrame` with the same
   additional columns and as many rows for each group as the rows returned for that group.
+  In this case returning a table with no columns drops group.
 
 `f` must always return the same kind of object (as defined in the above list) for
 all groups, and if a named tuple or data frame, with the same fields or columns.
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -201,7 +201,8 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives the same additional columns
   and as many rows for each group as the rows returned for that group.
-  (a table with zero columns drops the group)
+  As a special case, returning an empty table with zero columns is allowed,
+  whatever the number of columns returned for other groups.
 
 `f` must always return the same kind of object (as defined in the above list) for
 all groups, and if a named tuple or data frame, with the same fields or columns.
@@ -341,7 +342,8 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives a `DataFrame` with the same
   additional columns and as many rows for each group as the rows returned for that group.
-  (a table with zero columns drops the group)
+  As a special case, returning an empty table with zero columns is allowed,
+  whatever the number of columns returned for other groups.
 
 `f` must always return the same kind of object (as defined in the above list) for
 all groups, and if a named tuple or data frame, with the same fields or columns.
@@ -1055,7 +1057,8 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives a `DataFrame` with the same
   additional columns and as many rows for each group as the rows returned for that group.
-  (a table with zero columns drops the group)
+  As a special case, returning an empty table with zero columns is allowed,
+  whatever the number of columns returned for other groups.
 
 `f` must always return the same kind of object (as defined in the above list) for
 all groups, and if a named tuple or data frame, with the same fields or columns.
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -820,17 +820,17 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
     let eltys=eltys, n=n # Workaround for julia#15276
         initialcols = ntuple(i -> Tables.allocatecolumn(eltys[i], n), _ncol(first))
     end
-    targetcolnames = collect(Symbol, propertynames(first))
-    outcols = first isa Union{AbstractDataFrame,
+    targetcolnames = tuple(propertynames(first)...)
+    outcols, finalcolnames = first isa Union{AbstractDataFrame,
                               NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}} ?
               _combine_with_first!(first, initialcols, idx, 1, 1, f, gd, incols, targetcolnames) :
               _combine_with_first_row!(first, initialcols, idx, 1, 1, f, gd, incols, targetcolnames)
-    idx, outcols, targetcolnames
+    idx, outcols, collect(Symbol, finalcolnames)
 end
 
 function fill_row!(row, outcols::NTuple{N, AbstractVector},
                    i::Integer, colstart::Integer,
-                   colnames::Vector{Symbol}) where N
+                   colnames::NTuple{N, Symbol}) where N
     if !isa(row, Union{NamedTuple, DataFrameRow})
         throw(ArgumentError("return value must not change its kind " *
                             "(single row or variable number of rows) across groups"))
@@ -866,7 +866,7 @@ function _combine_with_first_row!(first::Union{NamedTuple, DataFrameRow},
                               idx::Vector{Int}, rowstart::Integer, colstart::Integer,
                               f::Any, gd::GroupedDataFrame,
                               incols::Union{Nothing, AbstractVector, NamedTuple},
-                              colnames::Vector{Symbol}) where N
+                              colnames::NTuple{N, Symbol}) where N
     len = length(gd)
     gdidx = gd.idx
     starts = gd.starts
@@ -898,7 +898,7 @@ function _combine_with_first_row!(first::Union{NamedTuple, DataFrameRow},
         end
         idx[i] = gdidx[starts[i]]
     end
-    outcols
+    return outcols, colnames
 end
 
 # This needs to be in a separate function
@@ -916,7 +916,7 @@ else
 end
 
 function append_rows!(rows, outcols::NTuple{N, AbstractVector},
-                      colstart::Integer, colnames::Vector{Symbol}) where N
+                      colstart::Integer, colnames::NTuple{N, Symbol}) where N
     if !isa(rows, Union{AbstractDataFrame, NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}})
         throw(ArgumentError("return value must not change its kind " *
                             "(single row or variable number of rows) across groups"))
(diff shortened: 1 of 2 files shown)


nalimilan:
> @nalimilan - one potential performance consideration is that now I use `colnames` as `Vector{Symbol}` - do you think it will be problematic?

Only benchmarking will tell, but I'm afraid this will make a difference. Would it be possible to have a loop in `_combine_with_first` to call the function on each group until a non-empty group is returned, and only then call `combine_with_first!`?

bkamins:
> Would it be possible to have a loop in `_combine_with_first` to call the function on each group until a non-empty group is returned, and only then call `combine_with_first!`?

I was thinking about it, but it required more code changes. I will try to do it the way you propose to keep `colnames` a `Tuple`.

bkamins:
OK - I have reverted to use `Tuple` for `colnames`.

nalimilan:
> I was thinking about it, but it required more code changes.

Really? I would have though `_combine_with_first` could just iterate over groups and check whether the result is empty. Then `_combine_with_first!` would just have to skip empty groups, without having to care about calling itself recursively. Am I missing something?

bkamins:
There are following points:
* I would have to change both `_combine(f::AbstractVector{<:Pair}, gd::GroupedDataFrame)` and `_combine(f::Any, gd::GroupedDataFrame)`
* we would be to remember that if we skipped rows then we should not allow single-row return values
* we would have to special case for situation when all groups return a no-column table.

In short - I felt that if I went this way I should change both `_combine` functions more significantly. But if you prefer I can rewrite it this way. This will save us one level of recursion (and probably one recompilation). What do you think?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -856,7 +861,7 @@ function fill_row!(row, outcols::NTuple{N, AbstractVector},
     return nothing
 end
 
-function _combine_with_first!(first::Union{NamedTuple, DataFrameRow},
+function _combine_with_first_row!(first::Union{NamedTuple, DataFrameRow},
                               outcols::NTuple{N, AbstractVector},
```
nalimilan: Fix indentation.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -818,9 +820,12 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
     let eltys=eltys, n=n # Workaround for julia#15276
         initialcols = ntuple(i -> Tables.allocatecolumn(eltys[i], n), _ncol(first))
     end
-    outcols = _combine_with_first!(first, initialcols, idx, 1, 1, f, gd, incols,
-                                   tuple(propertynames(first)...))
-    idx, outcols, propertynames(first)
+    targetcolnames = tuple(propertynames(first)...)
+    outcols, finalcolnames = first isa Union{AbstractDataFrame,
+                              NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}} ?
```
nalimilan: Weird indentation. Turn this into an `if`?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -818,9 +820,12 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
...
+              _combine_with_first!(first, initialcols, idx, 1, 1, f, gd, incols, targetcolnames) :
```
nalimilan: Maybe call this `_combine_with_first_table!`? Or even `_combine_tables_with_first!` and `_combine_rows_with_first!`?

BTW, it would be nice to file an issue in Julia with a branch to reproduce the bug (you can see in the file that I had similar issues in older Julia versions).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -950,12 +955,31 @@ function _combine_with_first!(first::Union{AbstractDataFrame,
     starts = gd.starts
     ends = gd.ends
     # Handle first group
-    j = append_rows!(first, outcols, colstart, colnames)
-    @assert j === nothing # eltype is guaranteed to match
-    append!(idx, Iterators.repeated(gdidx[starts[rowstart]], _nrow(first)))
+
+    @assert isempty(colnames) == (_ncol(first) == 0) == isempty(outcols)
```
nalimilan: `colnames` and `outcols` are guaranteed to have the same length thanks to the signature, so you can omit the last check. Actually you could just check `_ncol(first) == N`?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -950,12 +955,31 @@ function _combine_with_first!(first::Union{AbstractDataFrame,
...
+    if !isempty(colnames)
+        j = append_rows!(first, outcols, colstart, colnames)
+        @assert j === nothing # eltype is guaranteed to match
+        append!(idx, Iterators.repeated(gdidx[starts[rowstart]], _nrow(first)))
+    end
     # Handle remaining groups
     @inbounds for i in rowstart+1:len
         rows = wrap(do_call(f, gdidx, starts, ends, gd, incols, i))
+        if !(rows isa Union{AbstractDataFrame, NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}})
```
nalimilan: This check was missing in the existing code?

test/grouping.jl
```diff
@@ -643,29 +641,29 @@ end
         by(df, :a, (:c => sum,)) ==
         by(df, :a, [:c => sum]) ==
         by(df, :a, c_sum = :c => sum) ==
-        by(d -> (c_sum=sum(d.c),), df, :a)
+        by(d -> (c_sum=sum(d.c),), df, :a) ==
```
nalimilan: Wow that's embarrassing, good catch!

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -201,6 +201,7 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives the same additional columns
   and as many rows for each group as the rows returned for that group.
+  In this case returning a table with no columns drops group.
```
nalimilan: Maybe just say `"(a table with zero rows drops the group)"`? That's just a natural extension of the general rule given by the previous sentence, and the presence of columns or not doesn't change the result as long as there are no rows.

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
     @test_throws KeyError gd[(a=:A, b=:X)]
 end
 
+@testset "Allow returning DataFrame() or NamedTuple() to drop group" begin
+    for i in typemin(UInt8):typemax(UInt8),
+        er in (DataFrame(), view(DataFrame(ones(2,2)), 2:1, 2:1),
+               view(DataFrame(ones(2,2)), 1:2, 2:1),
+               NamedTuple(), rand(0,0), rand(5,0))
+        df = DataFrame(a = 1:8, x1 = Bool.(collect(bitstring(UInt8(i))) .- '0'))
```
nalimilan: The definition of `x1` is really tricky. Can you add a comment explaining what it does? IIUC the idea is to cover all possible sequences?

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+        res = by(sdf -> sdf.x1[1] ? DataFrame(x1=[true]) : er, df, :a)
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? DataFrame(x1=[true]) : er, groupby(df, :a)))
+        if nrow(res) == 0
+            @test res == DataFrame(a=Int[])
```
```suggestion
            @test res == DataFrame(a=[])
            @test typeof(res.a) == Vector{Int}
```

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+            @test res == df[df.x1, :]
+        end
+
+        res = by(sdf -> sdf.x1[1] ? (x1=[true],) : er, df, :a)
+        if typemin(UInt8) < i < typemax(UInt8)
+            @test_throws ArgumentError by(sdf -> sdf.x1[1] ? (x1=true,) : er, df, :a)
+        end
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? (x1=[true],) : er, groupby(df, :a)))
+        if nrow(res) == 0
+            @test res == DataFrame(a=Int[])
```
```suggestion
            @test res == DataFrame(a=[])
            @test typeof(res.a) == Vector{Int}
```

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+            @test res == df[df.x1, :]
+        end
+
+        res = by(sdf -> sdf.x1[1] ? hcat(true) : er, df, :a)
+        if typemin(UInt8) < i < typemax(UInt8)
+            @test_throws ArgumentError by(sdf -> sdf.x1[1] ? true : er, df, :a)
+        end
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? hcat(true) : er, groupby(df, :a)))
+        if nrow(res) == 0
+            @test res == DataFrame(a=Int[])
```
```suggestion
            @test res == DataFrame(a=[])
            @test typeof(res.a) == Vector{Int}
```

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+        else
+            @test res == df[df.x1, :]
+        end
+
+        res = by(sdf -> sdf.x1[1] ? hcat(true) : er, df, :a)
+        if typemin(UInt8) < i < typemax(UInt8)
+            @test_throws ArgumentError by(sdf -> sdf.x1[1] ? true : er, df, :a)
+        end
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? hcat(true) : er, groupby(df, :a)))
```
```suggestion
        @test res == DataFrame(map(sdf -> sdf.x1[1] ? hcat(true) : er, groupby_checked(df, :a)))
```

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+            @test res == DataFrame(a=Int[])
+        else
+            @test res == df[df.x1, :]
+        end
+
+        res = by(sdf -> sdf.x1[1] ? (x1=[true],) : er, df, :a)
+        if typemin(UInt8) < i < typemax(UInt8)
+            @test_throws ArgumentError by(sdf -> sdf.x1[1] ? (x1=true,) : er, df, :a)
+        end
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? (x1=[true],) : er, groupby(df, :a)))
```
```suggestion
        @test res == DataFrame(map(sdf -> sdf.x1[1] ? (x1=[true],) : er, groupby_checked(df, :a)))
```

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
 end
 
+@testset "Allow returning DataFrame() or NamedTuple() to drop group" begin
+    for i in typemin(UInt8):typemax(UInt8),
+        er in (DataFrame(), view(DataFrame(ones(2,2)), 2:1, 2:1),
+               view(DataFrame(ones(2,2)), 1:2, 2:1),
+               NamedTuple(), rand(0,0), rand(5,0))
+        df = DataFrame(a = 1:8, x1 = Bool.(collect(bitstring(UInt8(i))) .- '0'))
+        res = by(sdf -> sdf.x1[1] ? DataFrame(x1=[true]) : er, df, :a)
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? DataFrame(x1=[true]) : er, groupby(df, :a)))
```
```suggestion
        @test res == DataFrame(map(sdf -> sdf.x1[1] ? DataFrame(x1=[true]) : er, groupby_checked(df, :a)))
```

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
     @test_throws KeyError gd[(a=:A, b=:X)]
 end
 
+@testset "Allow returning DataFrame() or NamedTuple() to drop group" begin
+    for i in typemin(UInt8):typemax(UInt8),
```
nalimilan: Isn't this a bit too much? :-)

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+        er in (DataFrame(), view(DataFrame(ones(2,2)), 2:1, 2:1),
+               view(DataFrame(ones(2,2)), 1:2, 2:1),
+               NamedTuple(), rand(0,0), rand(5,0))
```
nalimilan: Maybe also test `DataFrame(a=Int[])`, `DataFrame(a=Any[])`, `NamedTuple(a=Int[])` and `NamedTuple(a=Any[])`?

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+        df = DataFrame(a = 1:8, x1 = Bool.(collect(bitstring(UInt8(i))) .- '0'))
+        res = by(sdf -> sdf.x1[1] ? DataFrame(x1=[true]) : er, df, :a)
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? DataFrame(x1=[true]) : er, groupby(df, :a)))
+        if nrow(res) == 0
+            @test res == DataFrame(a=Int[])
+        else
+            @test res == df[df.x1, :]
+        end
+
+        res = by(sdf -> sdf.x1[1] ? (x1=[true],) : er, df, :a)
```
nalimilan: Maybe loop over `(x1=[true],)`, `DataFrame(x1=[true])` and `hcat(true)` to avoid repeting this?

test/grouping.jl
```diff
@@ -1458,4 +1456,42 @@ end
...
+        if typemin(UInt8) < i < typemax(UInt8)
+            @test_throws ArgumentError by(sdf -> sdf.x1[1] ? (x1=true,) : er, df, :a)
```
nalimilan: Can you also test that returning e.g. `(x2=[true],)` fails?

nalimilan:
OK, makes sense. This strategy will have to be revised if we want to allow returning an empty tuple for `_combine_with_first_row!`, but that's not a priority anyway.

nalimilan:
Though can you check that performance doesn't regress e.g. on benchmarks like the one at https://github.com/JuliaData/DataFrames.jl/pull/1601 (also test a function taking several columns and returning a named tuple)?

bkamins:
> OK, makes sense. This strategy will have to be revised if we want to allow returning an empty tuple for _combine_with_first_row!, but that's not a priority anyway.

I came to the conclusion that `_combine_with_first_row!` should not allow empty return values. It is a nice invariant to have that guarantees that we know how many rows we are going to get.
(but maybe in the future I will change my mind 😄)

src/groupeddataframe/splitapplycombine.jl
bkamins: OK. Fixed (but "columns" not "rows" 😄 )

src/groupeddataframe/splitapplycombine.jl
bkamins: I was not sure what would be best. Changed to `if`

src/groupeddataframe/splitapplycombine.jl
bkamins: Fixed. I will try to recreate the dispatch problem and file an issue.

src/groupeddataframe/splitapplycombine.jl
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
bkamins: right - this check is left from vector version.

src/groupeddataframe/splitapplycombine.jl
bkamins: It would produce an error later (most likely inside `append_rows!`). I think it was better to have it here because:
1. it gives a cleaner error message early
2. in a `_ncol(rows)` test `rows` could be a `DataFrameRow` and if it had 0 columns we would allow it by `continue` and we should not.
3. a few lines below I do `if rows isa AbstractDataFrame` and in `else` I am sure it is a `NamedTuple` of vectors in `else` branch (otherwise I would have to add a third branch throwing an error).

test/grouping.jl

test/grouping.jl
bkamins: Yes. Unfortunately (probably) this was a most natural way to define it for me. I have changed it to something more normal now.

test/grouping.jl
bkamins: OK - I will reduce it from 256 to 16

test/grouping.jl
bkamins: added

test/grouping.jl
bkamins: good point. Initially I had more heterogeneous tests, but now it can be merged.

test/grouping.jl
bkamins: I understand you meant `x1` instead of `a` here?

bkamins: added also `rand(0, 1)` as it is a similar special case.

test/grouping.jl
bkamins: interestingly it uncovered another bug where we called `names` on `NamedTuple` in the code in corner cases.

bkamins:
> (also test a function taking several columns and returning a named tuple)

I have added a tests taking a single column and taking two columns

bkamins:
Here are some benchmarks (but feel free to ask for more, as I might have missed some important case):
* Data preparation:
```
using DataFrames, BenchmarkTools, Random
Random.seed!(1234);
df = DataFrame(a = categorical(rand(1:10000, 10^6)), b=rand(10^6));
gdf = groupby(df, :a);
```
* running on 0.20 (after precompilation):
```
julia> @time combine(gdf, first);
  1.249979 seconds (166.29 k allocations: 15.258 MiB)

julia> @time combine(gdf, (:a, :b) => x -> (a=x.a[1], b=x.b[1]));
  1.436761 seconds (480.75 k allocations: 29.987 MiB)

julia> @time combine(gdf, sdf -> sdf[1:2, :]);
 35.002942 seconds (101.96 M allocations: 22.739 GiB, 5.82% gc time)

julia> @time combine(gdf, (:a, :b) => x -> (a=x.a[1:2], b=x.b[1:2]));
 51.706808 seconds (203.13 M allocations: 37.077 GiB, 5.64% gc time)
```
* running on this PR
```
julia> @time combine(gdf, first);
  1.269710 seconds (166.28 k allocations: 15.258 MiB)

julia> @time combine(gdf, (:a, :b) => x -> (a=x.a[1], b=x.b[1]));
  1.507670 seconds (641.45 k allocations: 37.083 MiB, 0.49% gc time)

julia> @time combine(gdf, sdf -> sdf[1:2, :]);
 37.266022 seconds (102.03 M allocations: 22.742 GiB, 6.47% gc time)

julia> @time combine(gdf, (:a, :b) => x -> (a=x.a[1:2], b=x.b[1:2]));
 54.060982 seconds (203.44 M allocations: 37.091 GiB, 6.62% gc time)
```

So it seems we have about 5% regression in the "slow" case.

BTW. Interestingly `DataFrame`/`DataFrameRow` is processed faster than `NamedTuple` in output. Do you know why?

test/grouping.jl
```diff
@@ -1480,4 +1478,40 @@ end
...
+        er in (DataFrame(), view(DataFrame(ones(2,2)), 2:1, 2:1),
+               view(DataFrame(ones(2,2)), 1:2, 2:1),
+               NamedTuple(), rand(0,0), rand(5,0),
+               DataFrame(x1=Int[]), DataFrame(x1=Any[]),
+               (x1=Int[],), (x1=Any[],), rand(0,1)),
+        fr in (DataFrame(x1=[true]), (x1=[true],), hcat(true), [true])
+        df = DataFrame(a = 1:N, x1 = x1)
+        res = by(sdf -> sdf.x1[1] ? fr : er, df, :a)
+        @test res == DataFrame(map(sdf -> sdf.x1[1] ? fr : er, groupby_checked(df, :a)))
+        if fr == [true]
```
bkamins: we have a column naming issue here. It is unrelated to this PR:
```
julia> by(:a => x -> x[1] != 1 ? (x1=[2],) : [1], DataFrame(a=1:2), :a)
2×2 DataFrame
│ Row │ a     │ a_function │
│     │ Int64 │ Int64      │
├─────┼───────┼────────────┤
│ 1   │ 1     │ 1          │
│ 2   │ 2     │ 2          │

julia> by(:a => x -> x[1] == 1 ? (x1=[1],) : [2], DataFrame(a=1:2), :a)
2×2 DataFrame
│ Row │ a     │ x1    │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 1     │
│ 2   │ 2     │ 2     │
```
What do you think we should do in such cases (now we inherit column name from the first row in a different way than we check when combining for name consistency).

The problematic line is https://github.com/JuliaData/DataFrames.jl/pull/2066/files#diff-23657e51a9cc9e627fc153ba1e6e04c1L799, as it "manually" overrides column naming mechanics using in combine code.

As you can see it can lead to significant inconsistency (column name passed in `NamedTuple` got silently overriden by `a_function`).

An alternative solution would be to disallow mixing vectors and named tuple of a single vector (but current design allows this).

bkamins:
Thinking of it the regressions are in the table-oriented combine, which is slow anyway - we should expect some regression as we do additional checks in the loop for emptiness.

I think this is an additional argument to avoid allowing dropping rows in row-oriented combine (as it should be as fast as possible).

bkamins:
@nalimilan Steps to reproduce the problem with method dispatch at current commit:
1. Everywhere in `src/groupeddataframe/splitapplycombine.jl` replace function names `_combine_tables_with_first!` and `_combine_rows_with_first!` with some common name (e.g. `_combine_with_first!` as it was in the past)
2. run `test/grouping.jl` file
3. The test will get stuck at some operation run in line https://github.com/JuliaData/DataFrames.jl/blob/master/test/grouping.jl#L824

The problem with writing a short code reproducing it is that it "hanging" happens randomly (i.e. at different iteration of this multiply-nested loop). Also if you e.g. extract a "failing" case to an outer expression and run it stand-alone it goes through without hanging most of the time (sometimes it hangs).

Can you please let me know if you are able to reproduce the problem?

src/groupeddataframe/splitapplycombine.jl
nalimilan: I really meant "rows". A column with zero columns also has zero rows, and all tables with zero rows give the same result, right?

test/grouping.jl
nalimilan: Yeah that's not great, but I'm not sure what better solution we could find. Ideally I guess we could find whether at least one named tuple has been returned, and take its names (that would be doable if we pass a Boolean to `_combine_to_with_first!`). Maybe file an issue about that?

nalimilan:
OK, a 5% slowdown is acceptable. A type inference issue would have had dramatic effects. I'm surprised that returning a named tuple of vectors is slower, there must be a type instability somewhere.

I've followed the steps you wrote, but unfortunately I can't reproduce the problem. :-/ Was that with Julia 1.3?

src/groupeddataframe/splitapplycombine.jl
bkamins: There is a subtle thing:

a table with zero rows must have zero columns or have column names matching `colnames`. If table has zero rows but non-zero columns that have wrong names an error will be thrown (and this is intended).

What would you then suggest to write to convey this message?

test/grouping.jl
bkamins: I have opened https://github.com/JuliaData/DataFrames.jl/issues/2071. Actually I think this should be an error (unless the automatic column name for a vector and the actual column name in named tuple match)

bkamins:
Yes - it is Version 1.3.0 (2019-11-26) on Windows (maybe here is the difference).
Anyway - I think that for our purposes it is better to have two separate names anyway (the code is easier to understand this way).

src/groupeddataframe/splitapplycombine.jl
nalimilan: OK. Maybe "As a special case, returning an empty table with zero rows and columns is allowed, whatever the number of columns returned for other groups."?

nalimilan:
Have you tried pushing a branch to see whether the problem happens on Travis? I agree renaming makes sense here anyway, but we'd better track this bug down while we have a reproducer (or kind of...).

src/groupeddataframe/splitapplycombine.jl
bkamins: Actually we allow non-zero rows as long as it is zero columns, e.g. in `rand(10, 0)`, at least now. We could disallow it, but it would introduce another check that would only slow things down a bit, for no practical benefit.

I will write "As a special case, returning an empty table with zero columns is allowed, whatever the number of columns returned for other groups." as `isempty` and "zero columns" are correct conditions (they are duplicate but I think it is OK to give both as it will be easier to understand it by users).

bkamins:
@nalimilan - can you please have a final look at this PR if it requires something more. Then we could merge it and I would move to making a PR implementing the proposal in https://github.com/JuliaData/DataFrames.jl/pull/1975 (probably in the limited version so that we can slowly move forward and simplify checking if all is correct)

bkamins:
Thank you!

bkamins:
@nalimilan - thank you!

PR #2067, opened 2019-12-26
Overload Compat.eachrow and Compat.eachcol

tkf:
As `eachrow` and `eachcol` are in Compat.jl https://github.com/JuliaLang/Compat.jl/pull/658, I think it makes sense to overload the functions in Compat.jl rather than creating new functions.  What do you think?

I also added a test to make sure that similar overlooks do not happen in the future.


Project.toml
```diff
@@ -34,7 +34,7 @@ test = ["DataStructures", "DataValues", "Dates", "Logging", "Random", "Test"]
 [compat]
 julia = "1"
 CategoricalArrays = "0.7"
-Compat = "2, 3"
+Compat = "2.2, 3"
```
tkf: `eachrow` and `eachcol` need Compat 2.2: https://github.com/JuliaLang/Compat.jl/releases/tag/v2.2.0

test/runtests.jl
```diff
@@ -29,7 +29,8 @@ my_tests = ["utils.jl",
             "tabletraits.jl",
             "indexing.jl",
             "broadcasting.jl",
-            "deprecated.jl"]
+            "deprecated.jl",
+            "compat.jl"]
```
bkamins: deprecated.jl should be a last file tested.

src/DataFrames.jl
```diff
@@ -50,9 +52,7 @@ export AbstractDataFrame,
        unique!,
        unstack
 
-if VERSION >= v"1.1.0-DEV.792"
-    import Base.eachcol, Base.eachrow
-else
+if VERSION < v"1.1.0-DEV.792"
```
bkamins: Two questions as I do not know Compat.jl well enough:
* do we have to `export` something we import via `Compat.`? (probably yes, but I just wanted to make sure)
* shoud `import Compat.` be used even on current versions of Julia or it should be in the `if` block here.

bkamins:
Thank you for the PR. I have left two questions to learn for the future as I do not know Compat.jl well enough.

src/DataFrames.jl
tkf: > * do we have to `export` something we import via `Compat.`?

My understanding is that `Compat` is "just" a Julia package.  So, if you want `eachrow` to work without explicitly importing `Compat`, I think you need to re-export `eachrow` from DataFrames.jl.

> * shoud `import Compat.` be used even on current versions of Julia or it should be in the `if` block here.

I don't think it matters since `Base.eachrow === Compat.eachrow` for `julia` >= 1.1 and `julia` does not care how `eachrow` is imported.  But my guess is that it is rather an implementation detail of `julia` so

```julia
if VERSION < v"1.1.0-DEV.792"
    import Base: eachcol, eachrow
else
    import Compat: eachcol, eachrow
end
```

makes sense if you want to play on the very safe side.

test/compat.jl
```diff
@@ -0,0 +1,12 @@
+module TestCompat
```
nalimilan: I wouldn't say this deserves a separate file. Can you just test this at the same time as `eachcol` and `eachrow`?

src/DataFrames.jl
tkf: I pushed the change I just suggested 9f1d862285b9399a42f775efda0fd5e4739b2f07.  Let me know if you want to revert it.

nalimilan:
Thanks!

PR #2068, opened 2019-12-26
Make grouping.jl test pass when using include

nalimilan:
I don't really understand why we get `"DataFrames.GroupedDataFrame"` when doing `include("test/grouping.jl")`... Any idea?

### DIFF
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -1050,7 +1050,7 @@ end
     show(io, gd)
     str = String(take!(io.io))
     @test str == """
-    GroupedDataFrame with 4 groups based on key: A
+    $GroupedDataFrame with 4 groups based on key: A
     First Group (1 row): A = 1
     │ Row │ A     │ B      │ C       │
     │     │ Int64 │ String │ Float32 │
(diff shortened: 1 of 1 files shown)


bkamins:
Thank you for fixing this. It was bugging me for some time also (we could fix it everywhere in the show-related tests).

The reason is that we have tests wrapped in modules. Here is a result on a clean Julia session:
```
julia> module M
       using DataFrames
       println(DataFrame(rand(2,3)))
       end
2×3 DataFrames.DataFrame
│ Row │ x1        │ x2       │ x3       │
│     │ Float64   │ Float64  │ Float64  │
├─────┼───────────┼──────────┼──────────┤
│ 1   │ 0.451307  │ 0.176419 │ 0.991432 │
│ 2   │ 0.0617158 │ 0.884206 │ 0.292654 │
Main.M

julia>
```

if you load DataFrames.jl global scope before running tests all is OK:
```
julia> using DataFrames

julia> module M
       using DataFrames
       println(DataFrame(rand(2,3)))
       end
2×3 DataFrame
│ Row │ x1       │ x2       │ x3       │
│     │ Float64  │ Float64  │ Float64  │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 0.219251 │ 0.410601 │ 0.976862 │
│ 2   │ 0.275108 │ 0.694158 │ 0.481246 │
Main.M
```

PR #2072, opened 2019-12-28
[DO NOT MERGE] Method dispatch problem in combine

bkamins:
Try to reproduce problems discussed in https://github.com/JuliaData/DataFrames.jl/pull/2066#issuecomment-569403575 on Travis.

### DIFF
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -2,6 +2,7 @@
 # Type definition and basic methods
 #
 
+
 """
     GroupedDataFrame
 
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -201,6 +201,7 @@ which is determined using the following rules:
   for each group as the length of the returned vector for that group.
 - A data frame, a named tuple of vectors or a matrix gives the same additional columns
   and as many rows for each group as the rows returned for that group.
+  In this case returning a table with no columns drops group.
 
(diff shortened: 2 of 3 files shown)


bkamins:
@nalimilan as discussed this PR shows a problem, most likely with method dispatch. In comparison to https://github.com/JuliaData/DataFrames.jl/pull/2066 it has the methods `_combine_tables_with_first!` and `_combine_rows_with_first!` merged into one method ``_combine_with_first!` (this was an earlier approach in DataFrames.jl and in theory should be equivalent).

What does this branch show. In comparison to
https://travis-ci.org/JuliaData/DataFrames.jl/builds/630430168
that are results of Travis for #2066 here CI
https://travis-ci.org/JuliaData/DataFrames.jl/builds/630430967
has two problems:
1. build times are longer, especially under Julia 1.3 and nightly
2. in 4 cases out of 15 CI stalled (this was the original reason why this PR is opened, as I experience the same problems on my local Julia)

CC @JeffBezanson (as this probably is related to method dispatch in Julia 1.3 in general)

bkamins:
re-run tests, as maybe the issue is fixed on nightly.

bkamins:
@nalimilan - do you know if we can force CI to be triggered. I tried all standard methods, but they do not seem to work.

nalimilan:
Maybe mark it as ready?

bkamins:
I am closing it as Julia Base has diverged so much that the tests fail in so many places that it is not useful to keep it any more (I do not get the error but it might be we have other errors)

PR #2074, opened 2020-01-02
List missing functions in AbstractDataFrame API

nalimilan:


bkamins:
While we are at it, here is the list of functions an `AbstractDataFrame` should support on top of what is listed there:
`aggregate`
`by`
`flatten`
`getinex`
`groupby`
`isempty`
`lastindex`
`mapcols`
`stack`
`rename`
`show`
`unstack`
`view`
`axes`
`eachrow`
`eachcol`
`getpropery`
`hasproperty`
`isequal`
`issorted`
`ndims`
`parent`
`propertynames`
`size`
`sort`
`sortperm`

Not sure if they all should be listed. Also maybe we should sort what we select to show?

nalimilan:
Yes I wonder whether it makes sense to list everything. Maybe we'd better remove the list altogether and have people look at the index of functions in the manual? Ideally it could list the first paragraph of each docstring, and we would put a short summary of the function there...

bkamins:
> Maybe we'd better remove the list altogether

I always felt that it is strange to list only some of the functions.

In general - I am OK to whatever proposal you would fit best (and be willing to do in this PR 😄). We can also do some "half-measure" if doing it fully correctly is too much work.

A list of all exported functions defined for `AbstractDataFrame` can be generated by:
```
unique([m.name for m in methodswith(AbstractDataFrame) if !occursin("deprecated.jl", String(m.file))])
```

bkamins:
I think we can merge this and do further improvements later.

PR #2076, opened 2020-01-02
remove lastindex without a dimension

bkamins:
The removed method should not be used. It was needed when we supported single-index indexing.


bkamins:
Thank you!

PR #2078, opened 2020-01-04
update insertcols!

### LINKED ISSUE #1851: Redesign `insertcols!`
It would be good to add `copycols` kwarg to `insertcols!` and make it insert columns at the end of the data frame by default.

bkamins:
Fixes #1851 

Changes:
* allow multiple columns to be inserted
* support `copycols` keyword argument
* correctly handle `AbstractRange` columns (it was a bug in the existing code)
* deprecate passing columns as keyword arguments (in general we move away in DataFrames.jl from passing columns in kwargs except for the constructor; this reserves kwargs for passing "options" to the functions)
* deprecate implicit broadcasting (it was incorrectly implemented anyway + I guess we want to be consistent here - I can change this if we prefer to allow for implicit broadcasting, then I will fix it to use a correct method to create a vector)

As a general comment. If we go for these changes implicit broadcasting will be removed from DataFrames.jl except for constructor in the forms `DataFrame(:a=>1)` and `DataFrame(a=1)` but I guess we should keep these for user convenience.


src/dataframe/dataframe.jl
```diff
@@ -589,10 +580,8 @@ end
 ##############################################################################
 
 """
-    insertcols!(df::DataFrame, ind::Int; name=col,
-                makeunique::Bool=false)
-    insertcols!(df::DataFrame, ind::Int, (:name => col)::Pair{Symbol,<:AbstractVector};
-                makeunique::Bool=false)
+    insertcols!(df::DataFrame, ind::Int, (name=>col)::Pair{Symbol,<:AbstractVector}...;
+                makeunique::Bool=false, copycols::Bool=false)
```
nalimilan: Maybe we should use `copycols=true` by default? Not sure.

src/dataframe/dataframe.jl
```diff
@@ -604,6 +593,9 @@ Insert a column into a data frame in place. Return the updated `DataFrame`.
 - `makeunique` : Defines what to do if `name` already exists in `df`;
   if it is `false` an error will be thrown; if it is `true` a new unique name will
   be generated by adding a suffix
+- `copycols` : whether vectors passed as columns should be copied
+
+If `col` is an `AbstractRange` then insert a `collect(col)` into a `DataFrame`.
```
```suggestion
If `col` is an `AbstractRange` then the result of `collect(col)` is inserted.
```

src/dataframe/dataframe.jl
```diff
@@ -616,7 +608,7 @@ julia> d = DataFrame(a=1:3)
 │ 2   │ 2     │
 │ 3   │ 3     │
 
-julia> insertcols!(d, 1, b=['a', 'b', 'c'])
+julia> insertcols!(d, 1, :b=>'a':'c')
```
```suggestion
julia> insertcols!(d, 1, :b => 'a':'c')
```

src/dataframe/dataframe.jl
```diff
@@ -625,54 +617,78 @@ julia> insertcols!(d, 1, b=['a', 'b', 'c'])
...
 
-julia> insertcols!(d, 1, :c => [2, 3, 4])
-3×3 DataFrame
-│ Row │ c     │ b    │ a     │
-│     │ Int64 │ Char │ Int64 │
-├─────┼───────┼──────┼───────┤
-│ 1   │ 2     │ 'a'  │ 1     │
-│ 2   │ 3     │ 'b'  │ 2     │
-│ 3   │ 4     │ 'c'  │ 3     │
+julia> insertcols!(d, 2, :c=>2:4, :c=>3:5, makeunique=true)
```
```suggestion
julia> insertcols!(d, 2, :c => 2:4, :c => 3:5, makeunique=true)
```

src/dataframe/dataframe.jl
```diff
@@ -625,54 +617,78 @@ julia> insertcols!(d, 1, b=['a', 'b', 'c'])
...
     end
-    insert!(index(df), col_ind, name)
-    insert!(_columns(df), col_ind, item)
-    df
+    return df
 end
 
-insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol}; makeunique::Bool=false) =
-    insertcols!(df, col_ind, name_col[1] => upgrade_scalar(df, name_col[2]), makeunique=makeunique)
+function insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol}; makeunique::Bool=false)
```
nalimilan: Move these to deprecated.jl?

src/dataframe/dataframe.jl
```diff
@@ -625,54 +617,78 @@ julia> insertcols!(d, 1, b=['a', 'b', 'c'])
...
+    Base.depwarn("Implicit broadcasting when inserting a column is deprecated. Pass an `AbstractVector` instead. ", :insertcols!)
+    # this is not a fully correct way to upgrade a scalar but this is how upgrade_scalar
+    # worked so I keep it not to be breaking the method will be removed anyways
+    insertcols!(df, col_ind, name_col[1] => fill(name_col[2], (ncol(df) == 0) ? 1 : nrow(df)), makeunique=makeunique)
+end
 
 function insertcols!(df::DataFrame, col_ind::Int; makeunique::Bool=false, name_col...)
+    Base.depwarn("inserting colums using a keyword argument is deprecated. Use `Pair` syntax instead. ", :insertcols!)
```
nalimilan: Maybe the message could show what the new syntax looks like, e.g. `"insertcols!(df, col_ind, :$(keys(name_col)[1]) => v)`?

test/dataframe.jl
```diff
@@ -165,43 +165,95 @@ end
...
-    @test df[!,:a] == [1, 2]
-    @test df[!, :b] == [3.0, 4.0]
-    @test df[!, :newcol] == ["a", "b"]
-    @test df[!, :newcol_1] == ["a1", "b1"]
+    @test df.a == [1, 2]
+    @test df.b == [3.0, 4.0]
+    @test df.newcol == ["a", "b"]
+    @test df.newcol_1 == ["a1", "b1"]
+
+    @test insertcols!(df, 1, :c1 => 1:2) == df
```
```suggestion
    @test insertcols!(df, 1, :c1 => 1:2) === df
```

test/dataframe.jl
```diff
@@ -165,43 +165,95 @@ end
...
+    @test df.c1 isa Vector{Int}
+    x = [1, 2]
+    @test insertcols!(df, 1, :c2 => x, copycols=true) == df
```
```suggestion
    @test insertcols!(df, 1, :c2 => x, copycols=true) === df
```

test/dataframe.jl
```diff
@@ -165,43 +165,95 @@ end
...
+    @test df.c2 == x
+    @test df.c2 !== x
+    @test insertcols!(df, 1, :c3 => x) == df
```
```suggestion
    @test insertcols!(df, 1, :c3 => x) === df
```

src/dataframe/dataframe.jl
bkamins: I have thought about it. I rejected it because it would be breaking. But given that `insertcols!` is not used frequently maybe let us do it. It is similar case as `DataFrame` constructor with keyword arguments. We copy there by default. I will make the change.

src/dataframe/dataframe.jl
bkamins: ok

src/dataframe/dataframe.jl
bkamins: ok

bkamins:
OK, I have decided to just make `copycols=true` by default, without deprecation, so it is mildly breaking, but I think it should not be very problematic, and going through deprecation period is a pain with changing the defaults for keyword arguments.

bkamins:
@pdeffebach - since we leave implicit broadcasting in `DataFrame` constructor maybe we should leave it also here (I was more convinced to remove it in this function, but maybe we should leave it for the same reasons).

An alternative would be to add a type called e.g. `Repeat` that would explicitly say that we want repeat the value passed in it as many times as needed (and `1`) by default.

pdeffebach:
As long as we are using the `Pair` syntax, we aren't "breaking" any rules per-se. As in, it's not obvious that `=>` has the same behavior as `=`. And obviously we can't do `.=>` because that means something different. 

So I propose keeping it for the time being.

bkamins:
> So I propose keeping it for the time being.

You mean to keep allowing implicit broadcasting?

pdeffebach:
Yes I believe so. I would vote to keep the implicit broadcasting. 

However if there are performance considerations (checking if something is Broadcastable for example) then I would recommend removing it.

pdeffebach:
We could also use multiple dispatch to have a `DataFrameFill` type that does the broadcasting. But that seems overkill for such a simple issue.

bkamins:
There are no performance considerations fortunately. I will just copy the code from `DataFrame` constructor (as old code using `fill` that was earlier here was not correct)

@nalimilan - please let me know if you also think we should keep implicit broadcasting here. If we agree I will add it back again.

src/dataframe/dataframe.jl
```diff
@@ -589,12 +580,14 @@ end
 ##############################################################################
 
 """
-    insertcols!(df::DataFrame, ind::Int; name=col,
-                makeunique::Bool=false)
-    insertcols!(df::DataFrame, ind::Int, (:name => col)::Pair{Symbol,<:AbstractVector};
-                makeunique::Bool=false)
+    insertcols!(df::DataFrame, ind::Int, (name=>col)::Pair{Symbol,<:AbstractVector}...;
```
nalimilan: With this you could avoid repeating the signature. Maybe that's clearer?
```suggestion
    insertcols!(df::DataFrame, [ind::Int], (name=>col)::Pair{Symbol,<:AbstractVector}...;
```

src/dataframe/dataframe.jl
```diff
@@ -589,12 +580,14 @@ end
...
+                makeunique::Bool=false, copycols::Bool=true)
+    insertcols!(df::DataFrame, (name=>col)::Pair{Symbol,<:AbstractVector}...;
+                makeunique::Bool=false, copycols::Bool=true)
 
 Insert a column into a data frame in place. Return the updated `DataFrame`.
+If `ind` argument is omitted it is by default set to `ncol(df)+1`
```
```suggestion
If `ind` is omitted it is set to `ncol(df)+1`
```

src/dataframe/dataframe.jl
```diff
@@ -589,12 +580,14 @@ end
...
+(the column is inserted as a last column of a `DataFrame`).
```
```suggestion
(the column is inserted as the last column).
```

src/dataframe/dataframe.jl
```diff
@@ -589,12 +580,14 @@ end
...
 
 # Arguments
 - `df` : the DataFrame to which we want to add a column
```
```suggestion
- `df` : the DataFrame to which we want to add columns
```
Same below for other arguments.

nalimilan:
Damn, this broadcasting issue keeps coming up in different places. It wouldn't be too bad to allow it here too, but I wonder how far this will lead us in the end...

> An alternative would be to add a type called e.g. `Repeat` that would explicitly say that we want repeat the value passed in it as many times as needed (and `1`) by default.

There's already `Iterators.Repeated(1)`. I guess we could abuse it slightly.

bkamins:
> There's already `Iterators.Repeated(1)`. I guess we could abuse it slightly.

I have thought about it, but it seemed to verbose.

OK - I will add just standard implicit broadcasting. Only `DataFrame` and `insertcols!` need it, all other places allowing it are fortunately purged 😄.

src/dataframe/dataframe.jl
```diff
@@ -47,7 +47,10 @@ DataFrame(::GroupedDataFrame)
   to `false` then the constructor will still copy the passed columns
   if it is not possible to construct a `DataFrame` without materializing new columns.
 
-All columns in `columns` should have the same length.
+All columns in `columns` should be `AbstractVector`s and have the same length.
```
```suggestion
All columns in `columns` must be `AbstractVector`s and have the same length.
```

src/dataframe/dataframe.jl
```diff
@@ -47,7 +47,10 @@ DataFrame(::GroupedDataFrame)
...
+An exception are `DataFrame(kwargs...)` and `DataFrame(pairs::Pair{Symbol,<:Any}...)`
+form constructors which additionally allow a column to be of any other type that is not
+an `AbstractArray`, in which case the passed value is implicitly broadcasted.
```
nalimilan: Maybe just say "repeated"? Broadcast isn't actually used, even for structure which may support it.
```suggestion
an `AbstractArray`, in which case the passed value is automatically repeated to fill
a new vector of the appropriate length.
```

src/dataframe/dataframe.jl
```diff
@@ -589,21 +583,25 @@ end
...
+(the column is inserted as the last column).
 
 # Arguments
-- `df` : the DataFrame to which we want to add a column
+- `df` : the DataFrame to which we want to add columns
 - `ind` : a position at which we want to insert a column
 - `name` : the name of the new column
-- `col` : an `AbstractVector` giving the contents of the new column
+- `col` : an `AbstractVector` giving the contents of the new column or a value of any
+  other type that is not `AbstractArray` which will be implicitly broadcasted
```
```suggestion
  type other than `AbstractArray` which will be repeated to form a new vector
```

src/dataframe/dataframe.jl
```diff
@@ -625,58 +623,93 @@ julia> insertcols!(d, 1, b=['a', 'b', 'c'])
...
-                nn = Symbol("$(name)_$k")
-                if !hasproperty(df, nn)
-                    name = nn
-                    break
+    for (name, item) in name_cols
+        if !(item isa AbstractArray)
+            # perform implicit broadcasting of an item that is not an array
+            if ncol(df) == 0
+                target_row_count = 1
+                for (_, v) in name_cols
```
nalimilan: Maybe do this loop once at the beginning, and perform the dimension check at the same time? Then this loop will be simpler as you will already know the target row count and error handling won't be necessary anymore.

src/dataframe/dataframe.jl
```diff
@@ -625,58 +623,93 @@ julia> insertcols!(d, 1, b=['a', 'b', 'c'])
...
                 end
-                k += 1
+            else
+                target_row_count = nrow(df)
             end
-        else
-            msg = "Duplicate variable name $name. Pass makeunique=true" *
-                  " to make it unique using a suffix automatically."
-            throw(ArgumentError(msg))
+            target_row_count
```
nalimilan: ?

src/dataframe/dataframe.jl
```diff
@@ -625,58 +623,93 @@ julia> insertcols!(d, 1, b=['a', 'b', 'c'])
...
+            item = fill!(Tables.allocatecolumn(typeof(item), target_row_count),
+                                item)
```
```suggestion
                         item)
```

bkamins:
> If we wanted to follow that terminology, `pushcols!` would be the function to add columns at the end.

Actually I had exactly the same thought, when implementing this change following the request from Slack. However, we diverge from `insert!` already when allowing multiple columns to be inserted  (and this was a primary use case people also wanted).

In summary - I would be OK to add `pushcols!` instead of the additional method for `insertcols!`. The reason why we can handle this with one method (which is not possible in Base) is that we know that we only allow adding `Pair` which cannot be an index into a `DataFrame`.

So do you have any opinion which would be better?

src/dataframe/dataframe.jl
bkamins: removed

bkamins:
I have extracted-out all error checking to be done at the beginning of the function. This way if we start adding columns we are sure that we can do it (so we will not throw an error 'half-way" adding only some of columns but not all).

src/dataframe/dataframe.jl
```diff
@@ -640,9 +641,48 @@ function insertcols!(df::DataFrame, col_ind::Int, name_cols::Pair{Symbol,<:Any}.
...
 
+    if !makeunique
+        if !allunique(first.(name_cols))
+            throw(ArgumentError("Names of columns to be inserted into a data frame must be " *
+                                "unique when `makeunique=true`"))
+        end
+        for (n, _) in name_cols
+            if hasproperty(df, n)
+                throw(ArgumentError("Column $n is already present in the data frame " *
+                                "which is not allowed when `makeunique=true`"))
```
nalimilan: Wrong indentation.

src/dataframe/dataframe.jl
```diff
@@ -625,58 +624,122 @@ julia> insertcols!(d, 1, b=['a', 'b', 'c'])
...
-end
+    if target_row_count == -1
+        target_row_count = 1
+    end
 
-insertcols!(df::DataFrame, col_ind::Int, name_col::Pair{Symbol}; makeunique::Bool=false) =
-    insertcols!(df, col_ind, name_col[1] => upgrade_scalar(df, name_col[2]), makeunique=makeunique)
+    for (name, item) in name_cols
+        if !(item isa AbstractVector)
+            if iterm isa AbstractArray
```
```suggestion
            if item isa AbstractArray
```
But maybe turn this into `@assert`

src/dataframe/dataframe.jl
```diff
@@ -640,9 +641,48 @@ function insertcols!(df::DataFrame, col_ind::Int, name_cols::Pair{Symbol,<:Any}.
...
+        target_row_count = nrow(df)
+    end
+
+    for (n, v) in name_cols
+        if v isa AbstractVector
+            if target_row_count == -1
+                target_row_count = length(v)
+            elseif length(v) != target_row_count
+                throw(DimensionMismatch("length of new column $n which is $(length(v)) must match" *
+                                        " the target number of rows in data frame ($target_row_count)"))
```
nalimilan: This error may be a bit weird if you try to add two columns with different lengths to a zero-column data frame. Maybe check above the loop that all columns have the same size? Then you can just say here "number of rows" without the notion of target.

nalimilan:
That's OK, I just wanted to be sure we are aware of this.

src/dataframe/dataframe.jl
bkamins: OK - I will make an @assert

src/dataframe/dataframe.jl
bkamins: ok - will fix

src/dataframe/dataframe.jl
bkamins: OK - I will change it (but at least you seem to have gotten the meaning of `"target"` correctly here 😄)

pdeffebach:
Thanks for doing this! Two questions. 

Is the change from `:b => v` to `v => :b`, i.e. new column name coming at the end, not happening before 1.0? I thought it would be in this PR.

bkamins:
`:b => v` is not a part of `select` API. The reason is that `v` is not a function.

If we introduce it to `select` it will likely get expanded to `Col(:b) => (x -> v) => :b` if `v` is an `AbstractArray` and `:b => (x -> v) => :b` if `v` is some other type so this is going to be consistent.

The problem is that you cannot both have `:b => v` and `v => :b` as `v` could be a symbol or an `Int` so we have to decide on one or the other. If you feel `v => :b` would be better than `:b => v` please comment.

The reason I felt `:b => v` is preferred is that it avoids introducing breaking changes especially in `DataFrame` constructor.

pdeffebach:
Thanks for the insight!

> The problem is that you cannot both have :b => v and v => :b as v could be a symbol or an Int so we have to decide on one or the other. If you feel v => :b would be better than :b => v please comment.

I think its important to be consistent by having the destination column be the last thing written. If that's what we've decided to do in `select` (which I think we have), then we should do it here too.

bkamins:
OK - so you would also change it in `DataFrame` constructor?

src/dataframe/dataframe.jl

pdeffebach:
I will post an issue outlining the divergent syntaxes and we can discuss it over there, I think, since this would be a big change.

bkamins:
OK. We should not merge this nor the `select` PR before this issue is settled.

PR #2079, opened 2020-01-04
remove readtable, writetable and related macros

### LINKED ISSUE #1602: Remove deprecated readtable/writetable
In order to clean up the code base I would remove as much as possible form deprecated.jl in the after the next release.

This is related to my thinking that when we release DataFrames.jl version 1.0 it should not contain deprecated functionality.

I am raising this as an issue, because I think that if we go this way it would be also good to indicate in release notes of 0.X what deprecated functionality will be removed in 0.(X+1) - not as a guarantee, but in cases we know it.

In particular - what would you say to remove after the next release `readtable`/`writetable` and related functions?

CC @nalimilan

bkamins:
Fixes #1602

I propose to finally remove readtable and writetable and related macros from deprecated.
I have checked that this reduces load time of DataFrames.jl about 0.15 seconds.
The functions are long deprecated and not maintained. I think CSV.jl is now mature-enough to be considered as a full replacement (after the recent fix of `Tabler.rows` all performance issues with CSV.jl are resolved AFAICT).


bkamins:
You have been managing CSV.jl for years I would say 😄. Actually my crucial point is that no one knows now if `readtable` and `writetable` actually work correctly in all cases as they have not been maintained since https://github.com/JuliaData/DataFrames.jl/pull/1243, which is ages for DataFrames.jl.

bkamins:
OK - thank you. Let us get rolling with this.

@quinnj - from this moment CSV.jl is the king (you can count on my support there, as likely we are going to keep getting issues there, like the one I have just opened in https://github.com/JuliaData/CSV.jl/issues/558).

PR #2081, opened 2020-01-07
bugfix: io context no longer ignored

vonDonnerstein:
Without this bugfix IOContext specifications (e.g. to which precision to print Float64) will be ignored.


nalimilan:
Thanks! Do you feel like adding a test for this to ensure it doesn't regress in the future?

bkamins:
I have checked that this is the only place where passing context to `sprint` should matter, but yes - it would be good to add a test 😄.

bkamins:
@nalimilan - OK to merge?

Issue #1779, opened 2019-04-17
Corner case for show of a data frame in REPL


bkamins:
@nalimilan - I understand that you are working on making `show` of a data frame in REPL more slick. Here is a corner case to test against (currently it is not rendered very nicely):

```
julia> df = DataFrame(x="a"^1000)
1×1 DataFrame. Omitted printing of 1 columns
│ Row │ │     │ ├─────┼
│ 1   │

julia> show(stdout, df, splitcols=true)
1×1 DataFrame
│ Row │ │     │ ├─────┼
│ 1   │

│ Row │ x




              │
│     │ String




              │
├─────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
──────────────┤
[...]

PR #2087, opened 2020-01-20
Clean up column omission when columns overflow IOContext width

dgkf:
Addresses #1779

Specifically,
- For `MIME"text/plain"`
   - When `allcols=true` and any column would cause the display to overflow the IOContext, prints "mxn DataFrame. Display too narrow: Omitted printing of $ncols columns" 
   - When `allcols=false` and no columns are displayed, only print the header (avoiding situations where all the row numbers continue to print despite all columns being omitted)
- For `MIME"text/html"`
   - This was just an off-by-one error using `searchsortedfirst`
   - Also short-circuit the printing if no columns get printed as to not print a table of only row indices
- For `MIME"text/latex"`
   - Added same off-by-one error fix for `searchsortedfirst`
   - I **think** this is all that's needed. It definitely prevents the inclusion of the column, but I think that it still prints out columnless rows of a table. I didn't render any documents to fully test this. I'm not sure what the expected behavior would be.

#### Added Tests:
- Added tests for all MIME types and parameterized variants of behavior.
- Latex was tested against a generated table that lacked the column text, but I haven't tried to render it and would appreciate guidance on expected result.

### DIFF
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -122,7 +122,7 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
         write(io, "<p>$header</p>")
         if mxcol == 0
             write(io, "</tbody></table>")
-            mxcol == 0 && return
+            return
         end
     end
     for row in 1:mxrow
@@ -249,6 +249,14 @@ function _show(io::IO, ::MIME"text/latex", df::AbstractDataFrame; rowid=nothing)
         mxcol = min(mxcol, ttymxcol)
     end
 
+    nomit = size(df, 2) - mxcol
+    caption = nomit > 0 ? "omitted printing of $nomit columns" : ""
+
+    if mxcol == 0 && size(df, 2) > 0
+        write(io, "\\emph{$caption}\n")
+        return
+    end
+
     cnames = _names(df)[1:mxcol]
     alignment = repeat("c", mxcol)
     write(io, "\\begin{tabular}{r|")
@@ -293,6 +301,8 @@ function _show(io::IO, ::MIME"text/latex", df::AbstractDataFrame; rowid=nothing)
         write(io, " \\\\\n")
     end
     write(io, "\\end{tabular}\n")
+    nomit > 0 && write(io, "\\emph{$caption}\n")
+    return
 end
 
 function Base.show(io::IO, mime::MIME"text/latex", dfr::DataFrameRow)
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -120,7 +120,10 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
             header *= " (omitted printing of $(size(df, 2)-mxcol) columns)"
         end
         write(io, "<p>$header</p>")
-        mxcol == 0 && return
+        if mxcol == 0
+            write(io, "</tbody></table>")
+            mxcol == 0 && return
+        end
     end
     for row in 1:mxrow
         write(io, "<tr>")
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -92,7 +92,8 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
         tty_rows, tty_cols = displaysize(io)
         mxrow = min(mxrow, tty_rows)
         maxwidths = getmaxwidths(df, io, 1:mxrow, 0:-1, :X) .+ 2
-        mxcol = min(mxcol, searchsortedfirst(cumsum(maxwidths), tty_cols))
+        ttymxcol = searchsortedfirst(cumsum(maxwidths), tty_cols) - 1
+        mxcol = min(mxcol, ttymxcol)
     end
 
     cnames = _names(df)[1:mxcol]
@@ -114,12 +115,12 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
     write(io, "</thead>")
     write(io, "<tbody>")
     if summary
-        omitmsg = if mxcol < size(df, 2)
-                      " (omitted printing of $(size(df, 2)-mxcol) columns)"
-                  else
-                      ""
-                  end
-        write(io, "<p>$(digitsep(size(df, 1))) rows × $(digitsep(ncol(df))) columns$omitmsg</p>")
+        header = "$(digitsep(size(df, 1))) rows × $(digitsep(ncol(df))) columns"
+        if mxcol < size(df, 2)
+            header *= " (omitted printing of $(size(df, 2)-mxcol) columns)"
+        end
+        write(io, "<p>$header</p>")
+        mxcol == 0 && return
     end
     for row in 1:mxrow
         write(io, "<tr>")
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -242,7 +242,8 @@ function _show(io::IO, ::MIME"text/latex", df::AbstractDataFrame; rowid=nothing)
         tty_rows, tty_cols = get(io, :displaysize, displaysize(io))
         mxrow = min(mxrow, tty_rows)
         maxwidths = getmaxwidths(df, io, 1:mxrow, 0:-1, :X) .+ 2
-        mxcol = min(mxcol, searchsortedfirst(cumsum(maxwidths), tty_cols))
+        ttymxcol = searchsortedfirst(cumsum(maxwidths), tty_cols) - 1
+        mxcol = min(mxcol, ttymxcol)
     end
 
     cnames = _names(df)[1:mxcol]
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -405,28 +405,20 @@ function showrows(io::IO,
     end
 
     rowmaxwidth = maxwidths[ncols + 1]
-    colwidths = maxwidths .+ length(" ") .+ length(" |")  # add border + padding
-    colwidths[end] += length("|")  # add row column initial border
-    chunkbounds = getchunkbounds(maxwidths, splitcols, displaysize(io)[2])
-    nchunks = allcols ? length(chunkbounds) - 1 : min(length(chunkbounds) - 1, 1)
+    colwidths = maxwidths .+ textwidth(" ") .+ textwidth(" |")  # padding+border
+    colwidths[end] += textwidth("|")  # add row column initial border
+    coltextwidths = colwidths[end] .+ colwidths[1:end-1]
+    coloverflows = coltextwidths .> displaysize(io)[2]
+    chunkbounds = getchunkbounds(maxwidths, splitcols || !allcols, displaysize(io)[2])
 
-    overflowsdisplay = false
+    nchunks = length(chunkbounds) - 1
+    nchunks = allcols ? nchunks : (chunkbounds[2] < 1 ? 0 : min(nchunks, 1))
+    nchunks = !allcols && splitcols && any(coloverflows) ? 0 : nchunks
+
     header = displaysummary ? summary(df) : ""
-
-    # when allcols=true and any column overflows display, print omission message
-    if allcols && any(colwidths[end] .+ colwidths[1:end-1] .> displaysize(io)[2])
-        header *= ". Display too narrow: Omitted printing of $ncols columns"
-        overflowsdisplay = true
-    end
-
-    # when allcols=false and all cols won't fit, print omission message
-    if !allcols && length(chunkbounds) > 2
-        header *= ". Omitted printing of $(chunkbounds[end] - chunkbounds[2]) columns"
-        overflowsdisplay = chunkbounds[end] - chunkbounds[2] == ncols
-    end
-
+    nomit = ncols - sum(chunkbounds[1:nchunks+1])
+    header *= nomit > 0 ? ". Omitted printing of $(nomit) columns" : ""
     println(io, header)
-    overflowsdisplay && return
 
     for chunkindex in 1:nchunks
         leftcol = chunkbounds[chunkindex] + 1
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -405,14 +405,28 @@ function showrows(io::IO,
     end
 
     rowmaxwidth = maxwidths[ncols + 1]
+    colwidths = maxwidths .+ length(" ") .+ length(" |")  # add border + padding
+    colwidths[end] += length("|")  # add row column initial border
     chunkbounds = getchunkbounds(maxwidths, splitcols, displaysize(io)[2])
     nchunks = allcols ? length(chunkbounds) - 1 : min(length(chunkbounds) - 1, 1)
 
+    overflowsdisplay = false
     header = displaysummary ? summary(df) : ""
+
+    # when allcols=true and any column overflows display, print omission message
+    if allcols && any(colwidths[end] .+ colwidths[1:end-1] .> displaysize(io)[2])
+        header *= ". Display too narrow: Omitted printing of $ncols columns"
+        overflowsdisplay = true
+    end
+
+    # when allcols=false and all cols won't fit, print omission message
     if !allcols && length(chunkbounds) > 2
         header *= ". Omitted printing of $(chunkbounds[end] - chunkbounds[2]) columns"
+        overflowsdisplay = chunkbounds[end] - chunkbounds[2] == ncols
     end
+
     println(io, header)
+    overflowsdisplay && return
 
     for chunkindex in 1:nchunks
         leftcol = chunkbounds[chunkindex] + 1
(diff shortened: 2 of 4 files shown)


src/abstractdataframe/io.jl
```diff
@@ -114,12 +115,15 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
...
-                  end
-        write(io, "<p>$(digitsep(size(df, 1))) rows × $(digitsep(ncol(df))) columns$omitmsg</p>")
+        header = "$(digitsep(size(df, 1))) rows × $(digitsep(ncol(df))) columns"
+        if mxcol < size(df, 2)
+            header *= " (omitted printing of $(size(df, 2)-mxcol) columns)"
+        end
+        write(io, "<p>$header</p>")
+        if mxcol == 0
+            write(io, "</tbody></table>")
+            mxcol == 0 && return
```
bkamins: we did this test above so this should be enough I think:
```suggestion
            return
```

src/abstractdataframe/show.jl
```diff
@@ -405,14 +405,28 @@ function showrows(io::IO,
     end
 
     rowmaxwidth = maxwidths[ncols + 1]
+    colwidths = maxwidths .+ length(" ") .+ length(" |")  # add border + padding
```
bkamins: probably it is cleaner to use `textwidth` not `length` (but in this case it is the same - it is just a matter of consistency in the codes).

src/abstractdataframe/show.jl
```diff
@@ -405,14 +405,28 @@ function showrows(io::IO,
...
+    colwidths[end] += length("|")  # add row column initial border
     chunkbounds = getchunkbounds(maxwidths, splitcols, displaysize(io)[2])
     nchunks = allcols ? length(chunkbounds) - 1 : min(length(chunkbounds) - 1, 1)
 
+    overflowsdisplay = false
     header = displaysummary ? summary(df) : ""
+
+    # when allcols=true and any column overflows display, print omission message
+    if allcols && any(colwidths[end] .+ colwidths[1:end-1] .> displaysize(io)[2])
```
bkamins: when `allcols=true` we should disregard displaysize and just print everything if `splitcols=false`. What you propose should be done only if `splitcols=true` (`splitcols=false` means that the user wants to force printing everything as it is).

bkamins:
Thank you for the PR. I have left some minor comments.

> I'm not sure what the expected behavior would be.

For LaTeX I think when we do not display any column we should not print any rows. Do you think the same?

dgkf:
> For LaTeX I think when we do not display any column we should not print any rows. Do you think the same?

I agree. I should have added more description about why I think this is ambiguous. Unlike the "text/plain" and "text/html" outputs, the "text/latex" output doesn't currently print any sort of message about omitted columns. Should I add that message, or should we just output nothing?

src/abstractdataframe/show.jl
dgkf: Cool, I didn't know about `textwidth`. I'll switch over to using it.

bkamins:
> Should I add that message, or should we just output nothing?

I would add a message (like in HTML).

Now I think you see that even simplest PRs are usually lead to many small decisions, as we have a lot of legacy code written several years ago in the package that no one touched since 😄 (and for big PRs it is even more complex as you have to also consider the whole JuliaData ecosystem).

src/abstractdataframe/io.jl
dgkf: Good catch - a sloppy refactoring of some earlier code. I'll remove it.

dgkf:
I addressed all of your notes. Specifically:

- now uses `textlength`
- removed superfluous column condition (and redid a lot of the logic for `splitcols` anyways)
- columns now get omitted after the first column that overflows the display when `splitcols=false`
- all columns get omitted if any column individually overflows the display when `splitcols=true`
- latex output now considers io display limits, printing an omission message below the table when columns aren't displayed. 

A note, the existing latex overflow is compared is against io display width, which probably isn't indicative of when an outputted table will overflow the render target. I also couldn't find a way of reasonably generating a latex table in a limited io display, so I'm not sure what situations this arises in. Regardless, I don't think this is worth trying to fix in this pull request, and probably would be more appropriately solved in PrettyTables.jl.

bkamins:
When `allcols==false` then `splitcols` value should be ignored (see the docstring of `show`).

So I understand the logic should be:
1. `allcols==true && splitcols==true` => stop printing at first over-wide column (this is how I understand it works in your PR)
2. `allcols==true && splitcols==false` => always print everything disregarding display width (this is how I understand it works in your PR) also docstring should be updated to cover this case more precisely
3. `allcols==false` => ignore `splitcols` and print only as much as fits in a single page (this should be fixed as currently you vary the behavior depending on `splitcols` value, and we do not split cols at all if `allcols==false`)

dgkf:
Thanks @bkamins. I based this on the docstring as is, so let me know if the desired functionality should change. This is how it works currently, and to my understanding is the desired functionality.

- `allcols == true`: (**always** prints all columns, regardless of whether it would overflow display width)  
    - `splitcols == true` _or_ `splitcols == false`: when any chunk overflows, display all output, even if it overflows the display width.
- `allcols == false`: (omit column as needed)
    - `splitcols == true`: when any chunk overflows, omit all output.
    - `splitcols == false`: when any chunk (in this case only the first, which would only overflow when the first column is too wide) overflows display, omit all output.

The issue this stemmed from raised the `allcols == false && splitcols == false` case. Should we also omit columns when `allcols == true`? My thinking is that we shouldn't.

bkamins:
Well - we have an inconsistency here in the docstrings, as https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/show.jl#L582 says something different 😢.
We should make them consistent (especially that only https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/show.jl#L582 is shown to the user as the other docstring is added to an internal function).

So this is my understanding what is a right functionality:

- `allcols == true`: (**always** prints all columns, regardless of whether it would overflow display width)  
    - `splitcols == true` _or_ `splitcols == false`: when any chunk overflows, display all output, even if it overflows the display width. (**bkamins: AGREED**)
- `allcols == false`: (omit column as needed) (**bkamins: Here I would ignore `splitcols` following https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/show.jl#L582 description and just print as many columns as fit the display width - I do not think there is a case when the user would prefer not to have any output if some output would be possible to be displayed**)
    - `splitcols == true`: when any chunk overflows, omit all output.
    - `splitcols == false`: when any chunk (in this case only the first, which would only overflow when the first column is too wide) overflows display, omit all output.

src/abstractdataframe/show.jl
```diff
@@ -203,38 +198,38 @@ julia> df = DataFrame(A = 1:3, B = ["x", "yy", "z"]);
...
+julia> DataFrames.getchunkedcols(maxwidths, true, displaysize()[2])
+1-element Array{Array{Any,1},1}:
+ [1, 2]
 ```
 """
-function getchunkbounds(maxwidths::Vector{Int},
-                        splitcols::Bool,
-                        availablewidth::Int) # -> Vector{Int}
+function getchunkedcols(maxwidths::Vector{Int},
+                         splitcols::Bool,
```
bkamins: wrong indent

src/abstractdataframe/show.jl
```diff
@@ -203,38 +198,38 @@ julia> df = DataFrame(A = 1:3, B = ["x", "yy", "z"]);
...
+                         availablewidth::Int) # -> Vector{Int}
```
bkamins: wrong return value specification. I think you can just drop it, as you document it in the docstring.

src/abstractdataframe/show.jl
```diff
@@ -203,38 +198,38 @@ julia> df = DataFrame(A = 1:3, B = ["x", "yy", "z"]);
...
     ncols = length(maxwidths) - 1
     rowmaxwidth = maxwidths[ncols + 1]
     if splitcols
-        chunkbounds = [0]
+        chunkedcols = [[]] # begin with a single chunk with no columns
```
bkamins: This probably should be `[Int[]]` right as now you create `Array{Array{Any,1},1}`?

src/abstractdataframe/show.jl
```diff
@@ -203,38 +198,38 @@ julia> df = DataFrame(A = 1:3, B = ["x", "yy", "z"]);
...
+                push!(chunkedcols, [j])
                 totalwidth = rowmaxwidth + 4 + maxwidths[j] + 3
+            else
+                push!(chunkedcols[end], j)
             end
         end
-        push!(chunkbounds, ncols)
     else
-        chunkbounds = [0, ncols]
+        chunkedcols = [[1:ncols...]]
```
bkamins:
```suggestion
        chunkedcols = [1:ncols;]
```
is more efficient

src/abstractdataframe/show.jl
```diff
@@ -361,9 +356,8 @@ NOTE: The value of `maxwidths[end]` must be the string width of
 - `allcols::Bool = false`: Whether to print all columns, rather than
   a subset that fits the device width.
 - `splitcols::Bool`: Whether to split printing in chunks of columns fitting the
-  screen width rather than printing all columns in the same block. Unless
-  `allcols==true`, all columns will be omitted if any single chunk overflows
-  the screen width. 
+  screen width rather than printing all columns in the same block. By default
+  this is the case only if `io` has the `IOContext` property `limit` set.
```
bkamins: this is not true for this method, it takes `false` as default (as opposed to what `show` does)

src/abstractdataframe/show.jl
```diff
@@ -410,21 +404,27 @@ function showrows(io::IO,
...
     colwidths[end] += textwidth("|")  # add row column initial border
     coltextwidths = colwidths[end] .+ colwidths[1:end-1]
-    coloverflows = coltextwidths .> displaysize(io)[2]
-    chunkbounds = getchunkbounds(maxwidths, splitcols || !allcols, displaysize(io)[2])
 
-    nchunks = length(chunkbounds) - 1
-    nchunks = allcols ? nchunks : (chunkbounds[2] < 1 ? 0 : min(nchunks, 1))
-    nchunks = !allcols && splitcols && any(coloverflows) ? 0 : nchunks
+    chunkedcols = getchunkedcols(maxwidths, splitcols || !allcols, displaysize(io)[2])
+    chunkwidths = [colwidths[end] + sum(colwidths[cols]) for cols=chunkedcols]
```
bkamins:
```suggestion
    chunkwidths = [colwidths[end] + sum(colwidths[cols]) for cols in chunkedcols]
```
seems more readable

bkamins: also why do we add `colwidths[end]` to every value?

src/abstractdataframe/show.jl
```diff
@@ -410,21 +404,27 @@ function showrows(io::IO,
...
+    chunkoverflows = chunkwidths .> displaysize(io)[2]
+
+    if !allcols && any(chunkoverflows)
+        chunkedcols = chunkedcols[1:(findfirst(chunkoverflows) - 1)]
```
bkamins: why not this and omit the next condition?
```suggestion
        chunkedcols = chunkedcols[1:max(1, findfirst(chunkoverflows) - 1)]
```

src/abstractdataframe/show.jl
```diff
@@ -410,21 +404,27 @@ function showrows(io::IO,
...
+    end
+
+    if !allcols && length(chunkedcols) > 0
+        chunkedcols = chunkedcols[[1]]
+    end
 
     header = displaysummary ? summary(df) : ""
-    nomit = ncols - sum(chunkbounds[1:nchunks+1])
+    nomit = length(chunkedcols) > 0 ? ncols - sum(length.(chunkedcols)) : ncols
```
bkamins: maybe:
```suggestion
    nomit = isempty(chunkedcols) ? ncols : ncols - sum(length, chunkedcols)
```
it seems cleaner and should allocate less

src/abstractdataframe/show.jl
```diff
@@ -173,20 +173,15 @@ function getprintedwidth(maxwidths::Vector{Int}) # -> Int
...
-i-th chunk bound is the index MINUS 1 of the first column in the
-i-th chunk. The (i + 1)-th chunk bound is the EXACT index of the
-last column in the i-th chunk. For example, the bounds [0, 3, 5]
-imply that the first chunk contains columns 1-3 and the second chunk
-contains columns 4-5.
+When rendering an `AbstractDataFrame` to a REPL window in chunks, each of which
+will fit within the width of the REPL window, this function will return an
+array of chunk column indices that should be included in each chunk. For
+example, if printing is divided into two chunks containing columns 1 through 3
+and 4 through 5 respectively, `getchunkedcols` will return [[1, 2, 3], [4, 5]].
```
bkamins: after reviewing the code it seems that it is better to return a vector of `UnitRange` - this is not a big deal as this method does not have to be super efficient, but it seems cleaner to return `[1:3, 4:5]` in your example - for sure it will allocate less (so it would be a better style).

test/show.jl
```diff
@@ -199,18 +199,25 @@ end
     io = IOContext(IOBuffer(), :displaysize=>(10,20), :limit=>true)
     show(io, df_wide, splitcols=false, allcols=false)
     str = String(take!(io.io))
-    @test str == """1×3 DataFrame. Omitted printing of 2 columns
-                 │ Row │ x1    │
-                 │     │ Int64 │
-                 ├─────┼───────┤
-                 │ 1   │ 1     │"""
+    @test str == """
```
bkamins: It would be ideal if we had tests for all cases of combinations of `allcols` and `splitcols` and overwide column in different placements you implement (maybe we have it already, but I am just asking to be sure).

src/abstractdataframe/show.jl
dgkf: Good catch - I'm still getting used to this notation. I'm curious why this is used instead of a return type assertion. Is it just that this code predates that language feature?

```julia
function getchunkedcols(x, y, z)::Vector{Vector{Int}}
    ...
end
```

src/abstractdataframe/show.jl
dgkf: `colwidths[end]` represents the `Row` column width (see `getmaxwidths`, which eventually populates `maxwidths`). I updated the docstring in `showrows` to reflect that this is what is expected.

src/abstractdataframe/show.jl
bkamins: `::Vector{Vector{Int}}` is not type assertion but type conversion clause, so you can just omit it.
To see that this is not assertion but conversion consider this code:
```
julia> f(x)::Float64 = x

julia> f(1)
1.0
```

src/abstractdataframe/show.jl
bkamins: Ah - OK, and in `chunkedcols` this last column is never included - right?

src/abstractdataframe/show.jl
bkamins: and yes - this is just very old code. Current style is not to add return type in the comment, but rather specify it in the docstring.

src/abstractdataframe/show.jl
dgkf: That's right, in `getchunkedcols` (or `getchunkbounds` before I modified it), `maxwidths` doesn't influence the number of columns that get chunked.

src/abstractdataframe/show.jl
dgkf: This won't cover the case where no chunk overflows the display, but because `allcols == false`, only the first chunk is to be displayed anyways. I've reworked the code a bit to make it a bit clearer and will include changes in an upcoming commit.

bkamins: sorry I have meant `min` not `max` - fixed above.

dgkf: No worries, I understood what you were getting at. This still misses the case that I mentioned above.

If there's an alternative to `findfirst(...)` with a default value if nothing is found then we can clean this up quite a bit more, but as is we need to avoid `findfirst` returning `nothing` and throwing an error before we collapse the two conditionals.

bkamins: no matter if you changed it and it works. To avoid producing `nothing` use `something(findfirst(...), the_value_you_want_on_nothing)`

bkamins:
For the future - a standard thing to do for CI is to define column as e.g. `DataFrame(a=Int64[1,2,3])` so that it is consistent across architectures.

dgkf:
Thanks, keep the tips coming. I got a lot to learn in the Julia space. I did take a look at your other tests using data from rand where Int64 was specified, but figured String was a pretty safe bet and the exact type isn't important for the test.

bkamins:
@dgkf - do you have time to work on it to mark it for 1.0 release? If not it can wait

bkamins:
I have resolved the comments I think are resolved. Three unresolved are left - could you please have a look at them. Thank you!

bkamins:
Closing as we ended up with https://github.com/JuliaData/DataFrames.jl/pull/2403 approach. Thank you for working on this.

Issue #2088, opened 2020-01-22
Additional functions supported for DataFrame.jl


bkamins:
We could consider adding support for `count` for `AbstractDataFrame` like we now support `filter` (also https://github.com/JuliaData/DataFrames.jl/issues/2048 is related).

Additionally now `filter` is very inefficeint, we could introduce additional form:
```
filter(col => fun, df)
```
that will be much faster as it would be type stable. This is a style planned for `select` (the only difference is that `fun` is applied to each element, not a whole column).

bkamins:
https://github.com/JuliaData/DataFrames.jl/pull/2091 adds `filter`/`filter!` proposal. I leave `count` for later when we decide what to do with #2048 and similar.

bkamins:
I would close it. I have not seen any requests for `count`.

Issue #1935, opened 2019-08-28
How to perform `by` on two variables? Should we auto-splat?


xiaodaigh:
Suppose I want to compute the correlation between two variable within each group, how do I do that e.g. I want to compute the correlation between `x` and `y` within each group of `grp`?

See MWE

```julia
using DataFrames
df = DataFrame(grp = rand(["a","b"], 100), x= rand(100), y = rand(100))
by(df, :grp, (:x, :y)=>cor)
```
I though the above would be ok but only the below works.  The below also doesn't work.

```julia
by(df, :grp, (:x, :y)=>((x,y))->cor(x,y))
by(df, :grp, (:x, :y)=>(x,y)->cor(x,y))
by(df, :grp, (:x, :y)=>(xy)->cor(xy[1],xy[2))
```

xiaodaigh:
```julia
y(df,:grp,(:x,:y)=>xy->cor(xy...))
```

the above works but auto-splatting is a nicer interface IMO

```julia
by(df, :grp, (:x, :y)=>cor)
```

xiaodaigh:
I think the auto-splat version is more intuitive because

```julia
f = (x,y)->x+y
f(1,2)
```
It's not call using `f((1,2))`, so `(:x,:y) = > cor` invokes the feeling that `cor` is a two argument function. So auto-splatting makes more sense.

bkamins:
@nalimilan was designing this API so probably he can comment best.

The problem with auto-splatting is that it loses some information (you are not able to pass column names then to the called function).
However, I agree that auto-splatting is something I typically expect and the pattern I use very often is `x -> f(x...)` as you have noted. Actually I almost never need to know column names.

Also if we went for this change this would be breaking.

nalimilan:
Yes, the problem is that with the current system you can both apply splatting manually, or access columns by name (or do anything with the named tuple). As @bkamins noted, if we splat arguments automatically, there's no way to retrieve the column names. Also, in some cases you don't want splatting, e.g. to call `sum` on several columns.

Something we could do is use `All(:x, :y)` to enable automatic splatting. Or conversely deprecate the current behavior and require people to use `All(:x, :y)` to get a named tuple. Not sure which one is best, but note that our behavior was inspired by [JuliaDB](https://juliacomputing.github.io/JuliaDB.jl/latest/basics/#map-a-function-to-the-table-of-:x-and-:y-1), and consistency is somewhat nice.

[...]

PR #2091, opened 2020-01-26
add Pair to filter and filter!

bkamins:
Related to #2088. Easy filters will be faster and easier to write with it.


bkamins:
@nalimilan - this should be good to have a look at.

As discussed on Slack we have a decision if we want the function to be applied for each row (which is natural here). I think that in the end after all these `Pair` related PRs are merged I will add a section to the manual explaining the `Pair` usage scenarios.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -863,9 +863,15 @@ end
 
 """
     filter(function, df::AbstractDataFrame)
+    filter(cols => function, df::AbstractDataFrame)
 
 Return a copy of data frame `df` containing only rows for which `function`
-returns `true`. The function is passed a `DataFrameRow` as its only argument.
+returns `true`. The function is passed a `DataFrameRow` as its only argument
```
nalimilan: I'd turn this into three sentences (no cols, single col, multiple cols), each starting with ``If `cols` ...``

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -886,15 +892,40 @@ julia> filter(row -> row[:x] > 1, df)
 ├─────┼───────┼────────┤
 │ 1   │ 3     │ b      │
 │ 2   │ 2     │ a      │
+
+julia> filter(:x => >(1), df)
```
nalimilan: As noted in another PR, I'd rather use something which doesn't involve the fancy syntax:
```suggestion
julia> filter(:x => x -> x > 1, df)
```

Also show an example with several columns?

Same below.

src/abstractdataframe/abstractdataframe.jl

bkamins:
all comments should be fixed now.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -866,12 +866,16 @@ end
...
 Return a copy of data frame `df` containing only rows for which `function`
-returns `true`. The function is passed a `DataFrameRow` as its only argument
-if `cols` is not specified.
-If `cols` is passed then elements of column `cols` are passed to the function,
-if `cols` is a single column or as a `NamedTuple` if multiple columns are passed
-(any valid column selector is accepted). Passing `cols` leads to a more efficient
-execution of the operation for large data frames.
+returns `true`.
+
+If `cols` is not specified then he function is passed `DataFrameRow`s as its only argument.
```
nalimilan: Mixing plural and singular is weird:
```suggestion
If `cols` is not specified then the function is passed `DataFrameRow`s.
```

bkamins:
Thank you. I have applied the same fix to `filter!` docstring.

bkamins:
This should be good for a final reveiw. Thank you.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -867,9 +867,19 @@ end
...
     filter(function, df::AbstractDataFrame)
+    filter(cols => function, df::AbstractDataFrame)
 
 Return a copy of data frame `df` containing only rows for which `function`
-returns `true`. The function is passed a `DataFrameRow` as its only argument.
+returns `true`.
+
+If `cols` is not specified then the function is passed `DataFrameRow`s.
+If `cols` is a single column then then elements of column `cols` are passed to the function.
+If `cols` represents multiple columns (any valid column selector is accepted) then
```
nalimilan: This should be kept consistent with `by`/`combine`/`select` (https://github.com/JuliaData/DataFrames.jl/issues/1935): if we decide to splat, we should also do it here. But I guess we can merge the PR as-is, and change the code before the next breaking release if we decide to change other functions too.

bkamins: Agreed. We can wait a few days with merging this I think (when we get the response on #data).

bkamins: I will rewrite it. We will auto-splat

bkamins:
I have implemented auto-splatting for `filter` and `filter!` so that we can have a look how it works.

Here are some benchmarks:
```
julia> df = DataFrame(x = rand(1:10, 10^6), y = rand(1:10, 10^6));

julia> @benchmark filter(row -> row.x == row.y, df)
BenchmarkTools.Trial:
  memory estimate:  94.79 MiB
  allocs estimate:  4999012
  --------------
  minimum time:     140.924 ms (2.87% GC)
  median time:      147.238 ms (2.94% GC)
  mean time:        158.191 ms (2.99% GC)
  maximum time:     230.591 ms (1.84% GC)
  --------------
  samples:          32
  evals/sample:     1

julia> @benchmark filter([:x, :y] => ==, df)
BenchmarkTools.Trial:
  memory estimate:  3.25 MiB
  allocs estimate:  41
  --------------
  minimum time:     3.825 ms (0.00% GC)
  median time:      4.902 ms (0.00% GC)
  mean time:        5.356 ms (6.40% GC)
  maximum time:     11.774 ms (36.95% GC)
  --------------
  samples:          933
  evals/sample:     1

julia> @benchmark df[df.x .== df.y, :]
BenchmarkTools.Trial:
  memory estimate:  2.42 MiB
  allocs estimate:  32
  --------------
  minimum time:     2.098 ms (0.00% GC)
  median time:      2.957 ms (0.00% GC)
  mean time:        3.203 ms (5.94% GC)
  maximum time:     9.557 ms (50.90% GC)
  --------------
  samples:          1553
  evals/sample:     1
```

so we are way faster than old `filter` and only somewhat slower than direct indexing (part of the reason why we are slower is that we have to do the `::Bool` check when applying the function).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -873,9 +873,9 @@ Return a copy of data frame `df` containing only rows for which `function`
 returns `true`.
 
 If `cols` is not specified then the function is passed `DataFrameRow`s.
-If `cols` is a single column then then elements of column `cols` are passed to the function.
-If `cols` represents multiple columns (any valid column selector is accepted) then
-`NamedTuple`s formed with elements of the selected columns are passed to the function.
+If `cols` argument is passed then it should be valid column selector (additionally
```
```suggestion
If `cols` is specified then it should be a valid column selector
```
(for consistency with the previous sentence)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -873,9 +873,9 @@ Return a copy of data frame `df` containing only rows for which `function`
...
+allowing column duplicates are allowed if a vector of `Int` or `Symbol is passed),
```
```suggestion
(column duplicates are allowed if a vector of `Int` or `Symbol` is passed),
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -873,9 +873,9 @@ Return a copy of data frame `df` containing only rows for which `function`
...
+the the function is passed elements of the selected columns as consecutive positional arguments.
```
```suggestion
the function is passed elements of the selected columns as separate positional arguments.
```
(I'm not sure "consecutive" adds anything here, we can't pass non-consecutive arguments. :-)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -909,38 +909,37 @@ julia> filter(:x => x -> x > 1, df)
...
-end
-
-_filter_helper(f, itr) = collect(f(r)::Bool for r in itr)
+Base.filter(f, df::AbstractDataFrame) = _filter_helper(df, f, eachrow(df))
+Base.filter((col, f)::Pair{<:ColumnIndex}, df::AbstractDataFrame) =
+    _filter_helper(df, f, df[!, col])
+Base.filter((cols, f)::Pair{<:AbstractVector{Int}}, df::AbstractDataFrame) =
+    (cdf = _columns(df); _filter_helper(df, f, (cdf[i] for i in cols)...))
+Base.filter((cols, f)::Pair{<:AbstractVector{Symbol}}, df::AbstractDataFrame) =
+  filter([index(df)[col] for col in cols] => f, df)
```
```suggestion
    filter([index(df)[col] for col in cols] => f, df)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -909,38 +909,37 @@ julia> filter(:x => x -> x > 1, df)
...
+Base.filter((cols, f)::Pair, df::AbstractDataFrame) =
+  filter(index(df)[cols] => f, df)
```
```suggestion
    filter(index(df)[cols] => f, df)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -909,38 +909,37 @@ julia> filter(:x => x -> x > 1, df)
...
 """
     filter!(function, df::AbstractDataFrame)
 
 Remove rows from data frame `df` for which `function` returns `false`.
 
 If `cols` is not specified then the function is passed `DataFrameRow`s.
-If `cols` is a single column then then elements of column `cols` are passed to the function.
-If `cols` represents multiple columns (any valid column selector is accepted) then
-`NamedTuple`s formed with elements of the selected columns are passed to the function.
+If `cols` argument is passed then it should be valid column selector (additionally
```
nalimilan: Same suggestions.

nalimilan:
> (part of the reason why we are slower is that we have to do the `::Bool` check when applying the function).

Shouldn't the compiler be able to optimize this check away?

bkamins:
> Shouldn't the compiler be able to optimize this check away?

It should in theory but it seems this check adds an overhead (and we have to keep it). Anyway - at least there is hope that this will "just improve" in the future versions of Julia.

src/abstractdataframe/abstractdataframe.jl
bkamins: All suggestions applied. Thank you!

bkamins:
Before we merge. If I use broadcasting here like:
```
_filter_helper(df, f, cols...) = df[((x...) -> f(x...)::Bool).(cols...), :]
```
then we are much faster and the `Bool` check seems to be properly optimized out (again the only downside is additional compilation cost as broadcasting is a bit more expensive to compile, but I think it should be OK). So do we change it here and in `select` to broadcasting for `ByRow` operations?

nalimilan:
Let's take the same approach as for `ByRow` and have the discussion there.

PR #2098, opened 2020-02-04
improve performance of Tables.rowtable

bkamins:
This is needed for performance because `Tables.rows(df)` is type unstable (to avoid huge compilation costs in corner cases). However, if someone wants `Tables.rowtable` the compilation cannot be avoided (very wide data frames are still very slow to generate), so we go through `Tables.columntable` to improve performance.


bkamins:
@quinnj

Without this PR the following:
```
julia> df = DataFrame(rand(10, 1000));

julia> Tables.rowtable(df)
Internal error: encountered unexpected error in runtime:
StackOverflowError()
```
hangs Julia session on my laptop, while

```
Tables.rowtable(Tables.columntable(df))
```
has an unavoidable compilation hicckup but goes through smoothly.

nalimilan:
Isn't that a bug in Tables.jl? It shouldn't crash like that.

bkamins:
Maybe not a bug but some deep recursion. I am waiting for a new release of Tables.jl to check if it gets resolved then as @quinnj has made improvements to this code.

Still we will have to add a custom `Tables.rowtable` method in DataFrames.jl for performance reasons.

nalimilan:
But shouldn't Tables.jl have a fallback that does what this PR does? If even DataFrame (which is type unstable) wants this, wouldn't other type want the same definition?

bkamins:
The issue is that Tables.jl calls `Tables.rows` internally, which we manually override in DataFrames.jl to be `eachrow(df)` (otherwise `CSV.write` for very wide data frames could be prohibitively slow). So it cannot be fixed in Tables.jl (as the package does not know that we override `Tables.rows`).

In theory you are right, but I can imagine that some other packages redefine `Tables.rows` in some special way that is more efficient (as opposed to DataFrames.jl which is less efficient).

PR #2101, opened 2020-02-06
add inner/left/right/outer/semi/anti/cross-join

### LINKED ISSUE #2092: Reconsider overloading Base.join?
`Base.join` has a clear docstring which is to join an array of strings with delimiters. This seems like a different operation than what the join on DataFrames provides. Since the operation is different, it shouldn't share the namespace (`DataFrames.join` shouldn't extend `Base.join`)

bkamins:
Fixes #2092


bkamins:
CI failed due to an unrelated error (failure to clone Git repo), so this should be good to review. Thank you!

bkamins:
@nalimilan - if you will not have any comments on this I will review it again and want to merge it in a few days. Thank you!

src/abstractdataframe/join.jl
```diff
@@ -430,23 +345,622 @@ function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
-    join(join(df1, df2, on=on, kind=kind, makeunique=makeunique, validate=validate),
-         dfs..., on=on, kind=kind, makeunique=makeunique, validate=validate)
-end
+"""
+    innerjoin(df1, df2; on = Symbol[], makeunique = false,
+              indicator = nothing, validate = (false, false))
+    innerjoin(df1, df2, dfs...; on = Symbol[], makeunique = false,
+              validate = (false, false))
+
+Inner join two or more data frame objects and return a `DataFrame` containing
```
nalimilan: Can you explain what an inner join is? The `join` docstring defined each kind. Same for other kinds.

Also I'm not sure whether "to inner join" is a real verb. :-) Maybe "Perform an inner join"?

src/abstractdataframe/join.jl
```diff
@@ -430,23 +345,622 @@ function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
+  pair can be passed. It is also allowed to perform a join on multiple columns,
+  in which case a vector of column names or column name pairs can be passed
+  (mixing names and pairs is allowed). If more than two data frames are joined
+  then only a column name or a vector of column names are allowed.
+  `on` is a required argument.
+- `makeunique` : if `false` (the default), an error will be raised
+  if duplicate names are found in columns not joined on;
+  if `true`, duplicate names will be suffixed with `_i`
+  (`i` starting at 1 for the first duplicate).
+- `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
```
nalimilan: This doesn't make a lot of sense here, right (since rows must appear in both inputs)? Maybe better remove it. For other functions, it might be worth updating the description to mention that result will never be `"left_only"`/`"right_only"`/`"both"`.

src/abstractdataframe/join.jl
```diff
@@ -430,23 +345,622 @@ function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
+julia> job2 = DataFrame(identifier = [1, 2, 4], Job = ["Lawyer", "Doctor", "Farmer"])
+3×2 DataFrame
+│ Row │ identifier │ Job    │
+│     │ Int64      │ String │
+├─────┼────────────┼────────┤
+│ 1   │ 1          │ Lawyer │
+│ 2   │ 2          │ Doctor │
+│ 3   │ 4          │ Farmer │
+
+julia> innerjoin(name, job2, on = [:ID => :identifier])
```
nalimilan: Using a vector with a single value could mislead users into thinking that a pair can only be used with a vector. Same below.
```suggestion
julia> innerjoin(name, job2, on = :ID => :identifier)
```

src/deprecated.jl
```diff
@@ -282,3 +282,70 @@ function insertcols!(df::DataFrame, col_ind::Int; makeunique::Bool=false, name_c
...
     return insertcols!(df, col_ind, makeunique=makeunique, keys(name_col)[1] => name_col[1])
 end
+
+function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame;
+                   on::Union{<:OnType, AbstractVector} = Symbol[],
+                   kind::Symbol = :inner, makeunique::Bool=false,
+                   indicator::Union{Nothing, Symbol} = nothing,
+                   validate::Union{Pair{Bool, Bool}, Tuple{Bool, Bool}}=(false, false))
+    if kind == :inner
+        Base.depwarn("inner joining data frames using join is deprecated, use " *
```
nalimilan: Why not show the exact syntax (even if it's verbose)?

src/abstractdataframe/join.jl
nalimilan: Actually, it could make sense to have a list of all possible joins pointing to other functions (like the one we currently have in `?join`). For example that's what dplyr does (since it follows the classic R tradition of documenting multiple methods in a single docstring).

docs/src/man/joins.md
```diff
@@ -36,21 +36,25 @@ julia> join(people, jobs, on = :ID)
...
--   Inner: The output contains rows for values of the key that exist in both the first (left) and second (right) arguments to `join`.
--   Left: The output contains rows for values of the key that exist in the first (left) argument to `join`, whether or not that value exists in the second (right) argument.
--   Right: The output contains rows for values of the key that exist in the second (right) argument to `join`, whether or not that value exists in the first (left) argument.
--   Outer: The output contains rows for values of the key that exist in the first (left) or second (right) argument to `join`.
--   Semi: Like an inner join, but output is restricted to columns from the first (left) argument to `join`.
--   Anti: The output contains rows for values of the key that exist in the first (left) but not the second (right) argument to `join`. As with semi joins, output is restricted to columns from the first (left) argument.
--   Cross: The output is the cartesian product of rows from the first (left) and second (right) arguments to `join`.
+-   Inner: The output contains rows for values of the key that exist in both the first (left) and second (right) arguments to `innerjoin`.
```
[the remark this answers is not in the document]
bkamins: Updated. I have also updated the descriptions to reflect the fact that some joins accept more than two tables.

src/abstractdataframe/join.jl
bkamins: 1. For each join I have added a one sentence description what it does (a bit more precise than we had before).
2. Changed to `"Perform ..."` style.
3. Added `See also` section for each docstring

src/abstractdataframe/join.jl
bkamins: Indeed. I have cleaned such issues in other places. Actually I will remove it here and also for `antijoin` and `semijoin` where it currently even errors.

It seems that the idea to split the functions was good, as it simplifies thinking about their contract (and things that slipped through can be caught).

src/abstractdataframe/join.jl
```diff
@@ -430,23 +345,622 @@ function Base.join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
+julia> job2 = DataFrame(identifier = [1, 2, 4], Job = ["Lawyer", "Doctor", "Farmer"])
+3×2 DataFrame
+│ Row │ identifier │ Job    │
+│     │ Int64      │ String │
+├─────┼────────────┼────────┤
+│ 1   │ 1          │ Lawyer │
+│ 2   │ 2          │ Doctor │
+│ 3   │ 4          │ Farmer │
+
+julia> innerjoin(name, job2, on = :ID => :identifier)
```
bkamins: Yes - I was thinking what would be the best approach. I will add a comment that the vector can be used.

src/deprecated.jl
bkamins: OK - added

bkamins:
Thank you for the comments. Apart from the suggested changes I have improved the description of `validate` keyword argument for the case when more than two data frames are joined.

bkamins:
CI went down due to large deprecation of `join` in deprecated.jl.

src/abstractdataframe/join.jl
```diff
@@ -631,8 +660,8 @@ rightjoin(df1::AbstractDataFrame, df2::AbstractDataFrame;
     outerjoin(df1, df2, dfs...; on = Symbol[], kind = :inner, makeunique = false,
               validate = (false, false))
 
-Outer join two or more data frame objects and return a `DataFrame` containing
-the result.
+Perfrom an outer join two or more data frame objects and return a `DataFrame` containing
```
nalimilan:
```suggestion
Perform an outer join two or more data frame objects and return a `DataFrame` containing
```

src/abstractdataframe/join.jl
```diff
@@ -347,12 +347,12 @@ end
...
 """
     innerjoin(df1, df2; on = Symbol[], makeunique = false,
-              indicator = nothing, validate = (false, false))
+              validate = (false, false))
     innerjoin(df1, df2, dfs...; on = Symbol[], makeunique = false,
               validate = (false, false))
 
-Inner join two or more data frame objects and return a `DataFrame` containing
-the result.
+Perform an inner join two or more data frame objects and return a `DataFrame` containing
```
nalimilan:
```suggestion
Perform an inner join of two or more data frame objects and return a `DataFrame` containing
```
Or maybe "on"? Same elsewhere.

src/abstractdataframe/join.jl
```diff
@@ -347,12 +347,12 @@ end
...
+the result. Only include rows with keys that match in all passed data frames.
```
nalimilan: Written like this, it's not obvious that it's the definition an inner join. Maybe say "An inner join includes..." Same for other kinds.

docs/src/man/joins.md
```diff
@@ -39,18 +39,18 @@ julia> innerjoin(people, jobs, on = :ID)
 In relational database theory, this operation is generally referred to as a join.
 The columns used to determine which rows should be combined during a join are called keys.
 
-There are seven kinds of joins supported by the DataFrames package:
+There are seven kinds of joins supported by the following functions in DataFrames.jl package:
```
nalimilan: While we're at it, this sentence was weird:
```suggestion
The following functions are provided to perform seven kinds of joins:
```

src/abstractdataframe/join.jl
bkamins: ok - all suggestions applied

src/abstractdataframe/join.jl
```diff
@@ -556,8 +556,8 @@ leftjoin(df1::AbstractDataFrame, df2::AbstractDataFrame;
     rightjoin(df1, df2; on = Symbol[], makeunique = false,
               indicator = nothing, validate = (false, false))
 
-Perform a right join two or more data frame objects and return a `DataFrame` containing
-the result. Include all rows from `df2`.
+Perform a right join on twodata frame objects and return a `DataFrame` containing
```
nalimilan:
```suggestion
Perform a right join on two data frame objects and return a `DataFrame` containing
```

src/abstractdataframe/join.jl
```diff
@@ -777,8 +777,8 @@ outerjoin(df1::AbstractDataFrame, df2::AbstractDataFrame, dfs::AbstractDataFrame
 """
     semijoin(df1, df2; on = Symbol[], makeunique = false, validate = (false, false))
 
-Perform a semi join two or more data frame objects and return a `DataFrame` containing
-the result. Return a subset of rows of `df1` that match with the keys in `df2`.
+Perform a semi join of two data frame objects and return a `DataFrame` containing the result.
+A semi join returns a subset of rows of `df1` that match with the keys in `df2`.
```
nalimilan:
```suggestion
A semi join returns the subset of rows of `df1` that match with the keys in `df2`.
```

src/abstractdataframe/join.jl
```diff
@@ -872,8 +872,8 @@ semijoin(df1::AbstractDataFrame, df2::AbstractDataFrame;
 """
     antijoin(df1, df2; on = Symbol[], makeunique = false, validate = (false, false))
 
-Perform an anti join two or more data frame objects and return a `DataFrame` containing
-the result. Return a subset of rows of `df1` that do not match with the keys in `df2`.
+Perform an anti join of two data frame objects and return a `DataFrame` containing the result.
+An anti join returns a subset of rows of `df1` that do not match with the keys in `df2`.
```
nalimilan:
```suggestion
An anti join returns the subset of rows of `df1` that do not match with the keys in `df2`.
```

bkamins:
Thank you for working on this. I applied the changes.

PR #2107, opened 2020-02-11
Consistent period printing on Julia 1.5-DEV

omus:
Keeps printing behaviour the same on Julia 1.5-DEV as it currently exists. The change https://github.com/JuliaLang/julia/pull/34387 changes the 2-arg show to display the parser-friendly version.

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -24,11 +24,23 @@ end
 Render a value to an IO object. Unlike
 `show`, render strings without surrounding quote marks.
 """
-ourshow(io::IO, x::Any) =
-    show(IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x)), x)
+function ourshow(io::IO, x::Any)
+    ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
+    show(ctxt, "text/plain", x)
+end
+
 ourshow(io::IO, x::AbstractString) = escape_string(io, x, "")
 ourshow(io::IO, x::Symbol) = ourshow(io, string(x))
 ourshow(io::IO, x::Nothing) = nothing
+
+# AbstractChar: https://github.com/JuliaLang/julia/pull/34730 (1.5.0-DEV.261)
+# Irrational: https://github.com/JuliaLang/julia/pull/34741 (1.5.0-DEV.266)
+if VERSION < v"1.5.0-DEV.261" || VERSION < v"1.5.0-DEV.266"
+    function ourshow(io::IO, x::T) where T <: Union{AbstractChar, Irrational}
+        ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
+        show(ctxt, x)
+    end
+end
 
 """Return compact string representation of type T"""
 function compacttype(T::Type, maxwidth::Int=8)
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -25,8 +25,17 @@ Render a value to an IO object. Unlike
 `show`, render strings without surrounding quote marks.
 """
 function ourshow(io::IO, x::Any)
-    ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
-    show(ctxt, "text/plain", x)
+    io = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
+
+    # First try 3-arg show
+    sx = sprint(show, "text/plain", x, context=io)
+
+    # If the output contains line breaks, try 2-arg show instead.
+    if occursin('\n', sx)
+        sx = sprint(show, x, context=io)
+    end
+
+    print(io, sx)
 end
 
 ourshow(io::IO, x::AbstractString) = escape_string(io, x, "")
@@ -37,8 +46,8 @@ ourshow(io::IO, x::Nothing) = nothing
 # Irrational: https://github.com/JuliaLang/julia/pull/34741 (1.5.0-DEV.266)
 if VERSION < v"1.5.0-DEV.261" || VERSION < v"1.5.0-DEV.266"
     function ourshow(io::IO, x::T) where T <: Union{AbstractChar, Irrational}
-        ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
-        show(ctxt, x)
+        io = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
+        show(io, x)
     end
 end
 
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -21,12 +21,14 @@ end
 """
     DataFrames.ourshow(io::IO, x::Any)
 
-Render a value to an IO object. Unlike
-`show`, render strings without surrounding quote marks.
+Render a value to an `IO` object compactly and omitting type information, by
+calling 3-argument `show`, or 2-argument `show` if the former contains line breaks.
+Unlike `show`, render strings without surrounding quote marks.
 """
 function ourshow(io::IO, x::Any)
     io = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
 
+    # This mirrors the behavior of Base.print_matrix_row
     # First try 3-arg show
     sx = sprint(show, "text/plain", x, context=io)
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -6,7 +6,7 @@ fatalerrors = length(ARGS) > 0 && ARGS[1] == "-f"
 quiet = length(ARGS) > 0 && ARGS[1] == "-q"
 anyerrors = false
 
-using Test, Random, DataFrames
+using DataFrames, Dates, Test, Random
 
 my_tests = ["utils.jl",
             "cat.jl",
--- a/test/show.jl
+++ b/test/show.jl
@@ -1,6 +1,6 @@
 module TestShow
 
-using DataFrames, Random, Test
+using DataFrames, Dates, Random, Test
 
 function capture_stdout(f::Function)
     oldstdout = stdout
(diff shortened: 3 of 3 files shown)


src/abstractdataframe/show.jl
```diff
@@ -24,9 +24,13 @@ end
...
 """
-ourshow(io::IO, x::Any) =
-    show(IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x)), x)
+function ourshow(io::IO, x::Any)
+    ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
+    show(ctxt, "text/plain", x)
+end
+
 ourshow(io::IO, x::AbstractString) = escape_string(io, x, "")
+ourshow(io::IO, x::Char) = ourshow(io, repr(x))
```
omus: We can conditionally create this method based upon the Julia version once https://github.com/JuliaLang/julia/pull/34730 is merged.

bkamins: Maybe use `AbstractChar` as in relevant code in the PR for Base?

omus: Good thinking

omus:
Julia 1.5.0-DEV.242 without this PR (d2726d1ba):
```julia
julia> using DataFrames, Dates

julia> DataFrame(a = [Date(2020, 2, 11)], b = [DateTime(2020, 2, 11, 15)], c = [Day(1)])
1×3 DataFrame
│ Row │ a                  │ b                               │ c      │
│     │ Date               │ DateTime                        │ Day    │
├─────┼────────────────────┼─────────────────────────────────┼────────┤
│ 1   │ Date("2020-02-11") │ DateTime("2020-02-11T15:00:00") │ Day(1) │
```

Julia 1.5.0-DEV.242 with this PR:
```julia
julia> using DataFrames, Dates

julia> DataFrame(a = [Date(2020, 2, 11)], b = [DateTime(2020, 2, 11, 15)], c = [Day(1)])
1×3 DataFrame
│ Row │ a          │ b                   │ c     │
│     │ Date       │ DateTime            │ Day   │
├─────┼────────────┼─────────────────────┼───────┤
│ 1   │ 2020-02-11 │ 2020-02-11T15:00:00 │ 1 day │
```

src/abstractdataframe/show.jl
```diff
@@ -24,12 +24,20 @@ end
...
+    ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
+    show(ctxt, "text/plain", x)
+end
+
 ourshow(io::IO, x::AbstractString) = escape_string(io, x, "")
 ourshow(io::IO, x::Symbol) = ourshow(io, string(x))
 ourshow(io::IO, x::Nothing) = nothing
 
+# https://github.com/JuliaLang/julia/pull/34730
+if VERSION < v"1.5.0-DEV.261"
```
bkamins: Just to be sure - I understand that the workaround is needed only for `AbstractChar`?

omus: I just double checked what was affected from the change in https://github.com/JuliaLang/julia/pull/34387 and `AbstractChar` and `Irrational` need work arounds. It's possible that other external types will also need changes.

We may want to add compact support for `Irrational` as well:

Currently with PR
```julia
julia> DataFrame(x = π)
1×1 DataFrame
│ Row │ x                      │
│     │ Irrational{:π}         │
├─────┼────────────────────────┤
│ 1   │ π = 3.1415926535897... │
```

Without PR:
```julia
julia> DataFrame(x = π)
1×1 DataFrame
│ Row │ x         │
│     │ Irration… │
├─────┼───────────┤
│ 1   │ π         │
```

omus: I'm fine with either option here. What are other's preferences?

bkamins: Maybe let us then also handle `Irrational` just to be less breaking (but probably this is not crucial). Other types - this is OK, let us just try to be consistent with Base.

Could you please add a test for `Char` and `Irrational` then (just as additional columns to the "dates" test)? Thank you.

omus: I separated the tests. If you still want me to add them to the dates test I can change it.

src/abstractdataframe/show.jl
```diff
@@ -24,12 +24,23 @@ end
...
+    ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
+    show(ctxt, "text/plain", x)
+end
+
 ourshow(io::IO, x::AbstractString) = escape_string(io, x, "")
 ourshow(io::IO, x::Symbol) = ourshow(io, string(x))
 ourshow(io::IO, x::Nothing) = nothing
 
+# https://github.com/JuliaLang/julia/pull/34730
+# if VERSION < v"1.5.0-DEV.261"
```
omus: I'll update this to:
```julia
if VERSION < v"1.5.0-DEV.261" || VERSION < v"1.5.0-DEV.X"
```
Once https://github.com/JuliaLang/julia/pull/34741 is merged.

omus: Done

src/abstractdataframe/show.jl
```diff
@@ -24,12 +24,24 @@ end
...
+    show(ctxt, "text/plain", x)
+end
+
 ourshow(io::IO, x::AbstractString) = escape_string(io, x, "")
 ourshow(io::IO, x::Symbol) = ourshow(io, string(x))
 ourshow(io::IO, x::Nothing) = nothing
 
+# https://github.com/JuliaLang/julia/pull/34730
+# https://github.com/JuliaLang/julia/pull/34741
+if VERSION < v"1.5.0-DEV.261" || VERSION < v"1.5.0-DEV.266"
```
bkamins: just to understand - why this is not the same as simply `if VERSION < v"1.5.0-DEV.266"`?

omus: I just did it this way to record the changes to `AbstractChar` and `Irrational` occurred in different revisions. I can add a comment to make that more explicit or just drop that if it it seems unnecessary.

I may need to split this up further due to how `Irrational` works in Julia 1.0. Will post details

omus:
Something I missed is that Julia 1.0 - 1.1 before this PR displays π as:
```julia
1×1 DataFrame
│ Row │ a                      │
│     │ Irrational{:π}         │
├─────┼────────────────────────┤
│ 1   │ π = 3.1415926535897... │
```
where as Julia 1.2+ display:
```julia
1×1 DataFrame
│ Row │ a         │
│     │ Irration… │
├─────┼───────────┤
│ 1   │ π         │
```
I'll could either update the tests to reflect this variation or and a special case to `ourshow`. I guess I'll update `ourshow` to be consistent.

omus:
Looks like pre-1.2 that form of compact printing was unavailable. To avoid type piracy I'll just update the tests.

bkamins:
Thank you for working on this. I think what we have now is OK. As usual - even small PRs do not turn out to be as simple as expected.

src/abstractdataframe/show.jl
omus: I've added the comment. I can drop the `||` if you prefer still but I figured it doesn't really matter as this check only happens at compile time.

bkamins: Thank you. I have approved the PR. I am just waiting for @nalimilan to have a look at it and comment or merge.

src/abstractdataframe/show.jl
```diff
@@ -24,12 +24,24 @@ end
 Render a value to an IO object. Unlike
 `show`, render strings without surrounding quote marks.
 """
-ourshow(io::IO, x::Any) =
-    show(IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x)), x)
+function ourshow(io::IO, x::Any)
+    ctxt = IOContext(io, :compact=>get(io, :compact, true), :typeinfo=>typeof(x))
```
nalimilan: Thanks. Matrix printing in Base does something a bit more tricky (and which I find a bit ugly, but well...):
https://github.com/JuliaLang/julia/pull/34387/files#diff-6f3c3cea343e12203f3635ea6504d6b3R105
We should probably do the same.

test/show.jl
```diff
@@ -311,6 +314,33 @@ end
...
+    @test sprint(show, df) == """
+    1×3 DataFrame
+    │ Row │ a          │ b                   │ c     │
+    │     │ Date       │ DateTime            │ Day   │
+    ├─────┼────────────┼─────────────────────┼───────┤
+    │ 1   │ 2020-02-11 │ 2020-02-11T15:00:00 │ 1 day │"""
+
+    # Irrational
+    df = DataFrame(a = π)
+    if VERSION < v"1.2.0-DEV.276"
```
nalimilan: This doesn't looks like an acceptable result given how verbose is, and we still claim to support 1.0. It might be OK if that only affects `Rational` as it's unlikely to be stored in a data frame. But is there a risk that it affects other custom types (notably from packages, like Unitful.jl)?

test/runtests.jl
```diff
@@ -6,7 +6,7 @@ fatalerrors = length(ARGS) > 0 && ARGS[1] == "-f"
 quiet = length(ARGS) > 0 && ARGS[1] == "-q"
 anyerrors = false
 
-using Test, Random, DataFrames
+using DataFrames, Dates, Test, Random
```
nalimilan: Isn't this redundant with the next change?

omus: This is needed since `show.jl` runs the test inside a submodule `TestShow`. Without this the Date tests will fail with:
```
Test showing special types: Test Failed at /Users/omus/.julia/dev/DataFrames/test/show.jl:320
  Expression: sprint(show, df) == "1×3 DataFrame\n│ Row │ a          │ b                   │ c     │\n│     │ Date       │ DateTime            │ Day   │\n├─────┼────────────┼─────────────────────┼───────┤\n│ 1   │ 2020-02-11 │ 2020-02-11T15:00:00 │ 1 day │"
   Evaluated: "1×3 DataFrame\n│ Row │ a          │ b                   │ c      │\n│     │ Dates.Date │ Dates.DateTime      │ Dates… │\n├─────┼────────────┼─────────────────────┼────────┤\n│ 1   │ 2020-02-11 │ 2020-02-11T15:00:00 │ 1 day  │" == "1×3 DataFrame\n│ Row │ a          │ b                   │ c     │\n│     │ Date       │ DateTime            │ Day   │\n├─────┼────────────┼─────────────────────┼───────┤\n│ 1   │ 2020-02-11 │ 2020-02-11T15:00:00 │ 1 day │"
Stacktrace:
 [1] top-level scope at /Users/omus/.julia/dev/DataFrames/test/show.jl:320
 [2] top-level scope at /Users/omus/Development/Julia/1.3/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
 [3] top-level scope at /Users/omus/.julia/dev/DataFrames/test/show.jl:284
```

test/show.jl
omus: It should be noted that this is just mirroring how DataFrames and Julia already work. For example here is the behaviour of DataFrames v0.20.0 with various versions of Julia:

Julia 1.0.5
```julia
julia> using DataFrames; DataFrame(a = π)
1×1 DataFrame
│ Row │ a                      │
│     │ Irrational{:π}         │
├─────┼────────────────────────┤
│ 1   │ π = 3.1415926535897... │
```

Julia 1.1.1
```
julia> using DataFrames; DataFrame(a = π)
1×1 DataFrame
│ Row │ a                      │
│     │ Irrational{:π}         │
├─────┼────────────────────────┤
│ 1   │ π = 3.1415926535897... │
```

Julia 1.2.0
```julia
julia> using DataFrames; DataFrame(a = π)
1×1 DataFrame
│ Row │ a         │
│     │ Irration… │
├─────┼───────────┤
│ 1   │ π         │
```

Julia 1.3.1
```julia
julia> using DataFrames; DataFrame(a = π)
1×1 DataFrame
│ Row │ a         │
│     │ Irration… │
├─────┼───────────┤
│ 1   │ π         │
```

src/abstractdataframe/show.jl
omus: I've updated the code to use the same behaviour

test/show.jl
nalimilan: Ah OK. Clearly nobody cares or we would have got reports about it. :-)

omus:
Thank you. Feel free to merge/squash and merge when you're ready

bkamins:
Thank you!

PR #2111, opened 2020-02-12
Support Tables.jl 1.0

quinnj:


src/other/tables.jl
```diff
@@ -7,16 +7,24 @@ Tables.rows(df::AbstractDataFrame) = eachrow(df)
...
+Tables.getcolumn(df::AbstractDataFrame, nm::Symbol) = df[!, nm]
+Tables.columnnames(df::AbstractDataFrame) = names(df)
+
+Tables.getcolumn(dfr::DataFrameRow, i::Int) = dfr[i]
+Tables.getcolumn(dfr::DataFrameRow, nm::Symbol) = dfr[nm]
+Tables.columnnames(dfr::DataFrameRow) = names(dfr)
+
 getvector(x::AbstractVector) = x
-getvector(x) = collect(x)
+getvector(x) = [x[i] for i = 1:length(x)]
```
bkamins: what could the type of `x` in practice? (I guess it is an iterable that is not an `AbstractVector` but has length and is indexable; then what is left here except `Tuple` and `NamedTuple`). I understand `collect` allowed iterables that were not indexable - right?

quinnj:
Yes, that's correct. Do we already have a branch I could cherry-pick this to? Or should I make one?

bkamins:
There is no branch for patch releases yet (astonishingly no bugs in v0.20 were reported).

After merging this PR to master a new branch with this commit cherry-picked would be needed branching from tag v0.20 (similarly to what @nalimilan did for 0.19.4).

src/other/tables.jl
bkamins: @quinnj - can you please comment on this - so that I have a better understanding of the intention of this method. Thank you!

test/tables.jl
```diff
@@ -177,8 +181,8 @@ end
     df = DataFrame(rand(3,4))
     @test columnindex.(Ref(df), names(df)) == 1:4
     @test columnindex(df, :a) == 0
-    @test_throws ErrorException columnindex(df, 1)
-    @test_throws ErrorException columnindex(df, "x1")
+    # @test_throws ErrorException columnindex(df, 1)
```
bkamins: Also @quinnj - could you please comment why are these two tests disabled now?

bkamins: Here is a fix https://github.com/JuliaData/Tables.jl/pull/151.
It assumes that @quinnj wants to allow only `Symbol` as `name`.

src/other/tables.jl
quinnj: For Tables.jl 1.0, a column must be an indexable collection with known length. So `[x[i] for i = 1:length(x)]` is the most correct in terms of what the interface requires.

test/tables.jl
quinnj: Yes, sorry for not following up on this.

src/other/tables.jl
nalimilan: Though `collect` could be faster for some types, right? I just saw you mentioned on Slack that it was faster to iterate over JSON arrays than to call `getindex` on each entry. :-p

PR #2115, opened 2020-02-13
Show friendly error when trying to iterate DataFrames

non-Jedi:
Apologies if this has already been proposed before and rejected. I couldn't find it with cursory searching. I semi-frequently see people confused by the `MethodError` stacktrace thrown when trying to iterate over a DataFrame (e.g. https://discourse.julialang.org/t/i-do-not-understand-this-error-methoderror-no-method-matching-iterate-dataframe/34530 although that person has more confusion than just the exception). This PR adds a method for `iterate(::AbstractDataFrame)` which simply throws an error with more explanation of what went wrong.

### DIFF
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -21,7 +21,8 @@ Base.summary(dfrs::DataFrameRows) = "$(length(dfrs))-element DataFrameRows"
 Base.summary(io::IO, dfrs::DataFrameRows) = print(io, summary(dfrs))
 
 Base.iterate(::AbstractDataFrame) =
-    error("AbstractDataFrame is not iterable. Use eachrow(adf) to iterate rows.")
+    error("AbstractDataFrame is not iterable. Use eachrow(df) to get a row iterator" *
+          " or eachcol(df) to get a column iterator")
 
 """
     eachrow(df::AbstractDataFrame)
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -20,6 +20,9 @@ end
 Base.summary(dfrs::DataFrameRows) = "$(length(dfrs))-element DataFrameRows"
 Base.summary(io::IO, dfrs::DataFrameRows) = print(io, summary(dfrs))
 
+Base.iterate(::AbstractDataFrame) =
+    error("AbstractDataFrame is not iterable. Use eachrow(adf) to iterate rows.")
+
 """
     eachrow(df::AbstractDataFrame)
 
--- a/test/iteration.jl
+++ b/test/iteration.jl
@@ -43,7 +43,7 @@ using Test, DataFrames
     @test eltype.(eachcol(mapcols(Vector{Float64}, df))) == [Float64, Float64]
     @test eltype(map(Vector{Float64}, eachcol(df))) == Vector{Float64}
 
-    @test_throws MethodError for x in df; end
+    @test_throws ErrorException for x in df; end
 end
 
 @testset "mapcols" begin


src/abstractdataframe/iteration.jl
```diff
@@ -20,6 +20,13 @@ end
 Base.summary(dfrs::DataFrameRows) = "$(length(dfrs))-element DataFrameRows"
 Base.summary(io::IO, dfrs::DataFrameRows) = print(io, summary(dfrs))
 
+# Trying to iterate over a DataFrame directly (for row in df) tends to
+# be one of the first things a beginner does. Allowing a MethodError
+# to be raised is more "correct", but an explanation of the
+# appropriate function for iterating rows is more friendly.
+Base.iterate(::AbstractDataFrame) =
+    error("AbstractDataFrame is not iterable. Use eachrow(adf) to iterate rows.")
```
bkamins: maybe also add or `eachcol(df) to iterate columns`?
(depending on the background people might expect the iteration to be per-row or per-column)

bkamins:
I think it makes sense to add it. Thank you!

src/abstractdataframe/iteration.jl
non-Jedi: Very good point. The internet connection I'm currently on blocks port 22, so I can't push a fix right now, but I'll try to remember to do so later. Or you can just fix it when merging; whichever works.

bkamins:
```suggestion
    error("AbstractDataFrame is not iterable. Use eachrow(df) to get a row iterator" *
          " or eachcol(df) to get a column iterator")
```

src/abstractdataframe/iteration.jl
```diff
@@ -20,6 +20,14 @@ end
 Base.summary(dfrs::DataFrameRows) = "$(length(dfrs))-element DataFrameRows"
 Base.summary(io::IO, dfrs::DataFrameRows) = print(io, summary(dfrs))
 
+# Trying to iterate over a DataFrame directly (for row in df) tends to
+# be one of the first things a beginner does. Allowing a MethodError
+# to be raised is more "correct", but an explanation of the
+# appropriate function for iterating rows is more friendly.
+Base.iterate(::AbstractDataFrame) =
+    error("AbstractDataFrame is not iterable. Use eachrow(df) to get a row iterator" *
+          " or eachcol(df) to get a column iterator")
```
bkamins: @nalimilan - are you OK with this?

non-Jedi: Looks fine to me. :)

bkamins: I use `f() = dostuff()` if `dostuff()` is a single expression even if it spans more than one line (like in the case here). But this is a personal preference, and maybe @nalimilan would disagree 😄.

We use the style:
```
long_function_name_with_many_parameters(x) =
    single_expression_that_does_stuff()
```
a lot in DataFrames.jl (this is especially useful when you have many parameters with long signature).

src/abstractdataframe/iteration.jl
```diff
@@ -20,6 +20,14 @@ end
 Base.summary(dfrs::DataFrameRows) = "$(length(dfrs))-element DataFrameRows"
 Base.summary(io::IO, dfrs::DataFrameRows) = print(io, summary(dfrs))
 
+# Trying to iterate over a DataFrame directly (for row in df) tends to
```
nalimilan: Can you remove this comment? People can easily understand why this method has been defined, and if not they can look at the git history.

src/abstractdataframe/iteration.jl
nalimilan: That's fine with me, as long as there's a single expression.

bkamins:
@non-Jedi - do you have time to have a look at comments of @nalimilan so that we can merge this PR?

src/abstractdataframe/iteration.jl
non-Jedi: Done.

bkamins:
Thank you!

PR #2120, opened 2020-02-16
[BREAKING] deprecate names=true in eachcol

### LINKED ISSUE #2090: redesign of eachcol
Following the discussion on Slack we can consider to redesign `eachcol` in the following way:
* deprecate two argument form, allow only `eachcol(df)` that would be equivalent to `eachcol(df, false)`
* define `getindex(::DataFrameColumns, ::Symbol)`
* define `keys(::DataFrameColumns)` to produce symbols (this is breaking)
* define `pairs(::DataFrameColumns)` to produce symbol=>value pairs (this is breaking)
* after the deprecation period drop the second parameter in type definition of `DataFrameColumns`

Overall this change would simplify the design. We would have to make sure though that such changes do not break some API assumptions that are made for arrays (it seems not, but this might be delicate). Other packages, like AxisArrays.jl do not override default `keys` and `pairs` definitions.

bkamins:
Fixes #2090.

In the next move we will drop V parameter of `DataFrameColumns`.

Making this PR revealed that `names=true` is in practice almost never used (at least internally), so this change - while big - should not be that problematic.

### DIFF
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -94,8 +94,9 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
 """
     DataFrameColumns{<:AbstractDataFrame, V} <: AbstractVector{V}
 
-An `AbstractVector` that allows for an iteration over columns of an `AbstractDataFrame`.
-Additionally it is allowed to index `DataFrameColumns` using column names.
+An `AbstractVector` that allows iteration over columns of an `AbstractDataFrame`.
+Indexing into `DataFrameColumns` objects using integer or symbol indices
+returns the corresponding column (without copying).
 """
 struct DataFrameColumns{T<:AbstractDataFrame} <: AbstractVector{AbstractVector}
     df::T
@@ -108,7 +109,7 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
     eachcol(df::AbstractDataFrame)
 
 Return a `DataFrameColumns` that is an `AbstractVector`
-that alows to iterate an `AbstractDataFrame` column by column.
+that allows iterating an `AbstractDataFrame` column by column.
 Additionally it is allowed to index `DataFrameColumns` using column names.
 
 # Examples
@@ -158,16 +159,17 @@ Base.getproperty(itr::DataFrameColumns, col_ind::Symbol) = getproperty(parent(it
 Base.propertynames(itr::DataFrameColumns, private::Bool=false) = propertynames(parent(itr))
 
 """
-    keys(::DataFrameColumns)
+    keys(dfc::DataFrameColumns)
 
-Get a vector of column names of a parent data frame.
+Get a vector of column names of `dfc`.
 """
 Base.keys(itr::DataFrameColumns) = names(parent(itr))
 
 """
     pairs(dfc::DataFrameColumns)
 
-An iterator that accesses each column of a parent data frame, returning `name => col`,
+Return an iterator of pairs associating the name of each column of `dfc`
+with the corresponding column vector, i.e. `name => col`
 where `name` is the column name of the column `col`.
 """
 Base.pairs(itr::DataFrameColumns) = Base.Iterators.Pairs(itr, keys(itr))
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -90,30 +90,26 @@ Base.getproperty(itr::DataFrameRows, col_ind::Symbol) = getproperty(parent(itr),
 Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(parent(itr))
 
 # Iteration by columns
+
 """
     DataFrameColumns{<:AbstractDataFrame, V} <: AbstractVector{V}
 
-Iterator over columns of an `AbstractDataFrame` constructed using
-[`eachcol(df, true)`](@ref) if `V` is a `Pair{Symbol,AbstractVector}`. Then each
-returned value is a pair consisting of column name and column vector.
-If `V` is an `AbstractVector` (a value returned by [`eachcol(df, false)`](@ref))
-then each returned value is a column vector.
+An `AbstractVector` that allows for an iteration over columns of an `AbstractDataFrame`.
+Additionally it is allowed to index `DataFrameColumns` using column names.
 """
-struct DataFrameColumns{T<:AbstractDataFrame, V} <: AbstractVector{V}
+struct DataFrameColumns{T<:AbstractDataFrame} <: AbstractVector{AbstractVector}
     df::T
 end
 
-Base.summary(dfcs::DataFrameColumns{T, V}) where {T,V} =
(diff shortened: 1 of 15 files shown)


bkamins:
I decided to redesign `eachcol` in one shot. What will be breaking:
* `eachcol(df, true)` will print differently (it will be a plain array)
* `eachcol(df, true)` will not have `getproperty` and `propertynames` defined

Given that probably `eachcol(df, true)` is not used that much and we introduced `getproperty` only recently I think it should be OK to go this way.

quinnj:
There was a discussion the other day on slack that mentioned a convention of adding `[BREAKING]` to PRs that introduce breaking changes. This can help later when you review tagbot release notes w/ PRs that start w/ [BREAKING].

bkamins:
We now use a label `breaking` for this. But I can also add [BREAKING] in the name

bkamins:
@nalimilan the PR is good to have a look at. the ❌ is due to coveralls only.

src/abstractdataframe/iteration.jl
```diff
@@ -90,30 +90,26 @@ Base.getproperty(itr::DataFrameRows, col_ind::Symbol) = getproperty(parent(itr),
...
+
 """
     DataFrameColumns{<:AbstractDataFrame, V} <: AbstractVector{V}
 
-Iterator over columns of an `AbstractDataFrame` constructed using
-[`eachcol(df, true)`](@ref) if `V` is a `Pair{Symbol,AbstractVector}`. Then each
-returned value is a pair consisting of column name and column vector.
-If `V` is an `AbstractVector` (a value returned by [`eachcol(df, false)`](@ref))
-then each returned value is a column vector.
+An `AbstractVector` that allows for an iteration over columns of an `AbstractDataFrame`.
```
```suggestion
An `AbstractVector` that allows iteration over columns of an `AbstractDataFrame`.
```

src/abstractdataframe/iteration.jl
```diff
@@ -90,30 +90,26 @@ Base.getproperty(itr::DataFrameRows, col_ind::Symbol) = getproperty(parent(itr),
...
+Additionally it is allowed to index `DataFrameColumns` using column names.
```
```suggestion
Indexing into `DataFrameColumns` objects using integer or symbol indices
returns the corresponding column (without copying).
```

src/abstractdataframe/iteration.jl
```diff
@@ -143,39 +139,39 @@ julia> sum.(eachcol(df))
...
+Base.getindex(itr::DataFrameColumns, j::Symbol) = parent(itr)[!, j]
+
 Base.getproperty(itr::DataFrameColumns, col_ind::Symbol) = getproperty(parent(itr), col_ind)
 # Private fields are never exposed since they can conflict with column names
 Base.propertynames(itr::DataFrameColumns, private::Bool=false) = propertynames(parent(itr))
 
+"""
+    keys(::DataFrameColumns)
+
+Get a vector of column names of a parent data frame.
```
nalimilan: Do we really need to mention the parent?
```suggestion
Get a vector of column names of `dfc`.
```

src/abstractdataframe/iteration.jl
```diff
@@ -90,30 +90,26 @@ Base.getproperty(itr::DataFrameRows, col_ind::Symbol) = getproperty(parent(itr),
...
 """
-    eachcol(df::AbstractDataFrame, names::Bool=false)
+    eachcol(df::AbstractDataFrame)
 
-Return a `DataFrameColumns` that iterates an `AbstractDataFrame` column by column.
-If `names` is equal to `false` (the default) iteration returns column vectors.
-If `names` is equal to `true` pairs consisting of column name and column vector
-are yielded.
+Return a `DataFrameColumns` that is an `AbstractVector`
+that alows to iterate an `AbstractDataFrame` column by column.
```
```suggestion
that allows iterating an `AbstractDataFrame` column by column.
```

src/abstractdataframe/iteration.jl
```diff
@@ -143,39 +139,39 @@ julia> sum.(eachcol(df))
...
 end
 
+Base.getindex(itr::DataFrameColumns, j::Symbol) = parent(itr)[!, j]
+
 Base.getproperty(itr::DataFrameColumns, col_ind::Symbol) = getproperty(parent(itr), col_ind)
 # Private fields are never exposed since they can conflict with column names
 Base.propertynames(itr::DataFrameColumns, private::Bool=false) = propertynames(parent(itr))
 
+"""
+    keys(::DataFrameColumns)
```
```suggestion
    keys(dfc::DataFrameColumns)
```

src/abstractdataframe/iteration.jl
```diff
@@ -143,39 +139,39 @@ julia> sum.(eachcol(df))
...
+
+Get a vector of column names of a parent data frame.
+"""
+Base.keys(itr::DataFrameColumns) = names(parent(itr))
+
+"""
+    pairs(dfc::DataFrameColumns)
+
+An iterator that accesses each column of a parent data frame, returning `name => col`,
```
```suggestion
Return an iterator of pairs associating the name of each column of `dfc`
with the corresponding column vector, i.e. `name => col`
```

test/deprecated.jl
```diff
@@ -539,6 +539,108 @@ end
     @test_throws ArgumentError join(df1, df2, df3, on=:id, kind=:xxx)
 end
 
+@testset "eachcol(df, true)" begin
+    df = DataFrame(a=1:3, b=4:6, c=7:9)
+    @test eachcol(df)[1] === last(eachcol(df, true)[1])
+    @test eachcol(df)[1] === last(eachcol(df, true)[1])
+
+
+    # show for eachcol(df, true) is breaking but this is a minor thing
```
nalimilan: Just remove this?

test/deprecated.jl
```diff
@@ -539,6 +539,108 @@ end
...
+
+    #     │ Row │ x       │
+    #     │     │ Float64 │
+    #     ├─────┼─────────┤
+    #     │ 1   │ 3.14159 │"""
+
+    df = DataFrame(rand(3,4), [:a, :b, :c, :d])
+    df2 = DataFrame(eachcol(df, true))
+    @test df == df2
+    @test !any(((a,b),) -> a === b, zip(eachcol(df), eachcol(df2)))
```
nalimilan: This block doesn't contain any relevant tests for `eachcol(df)`?

test/deprecated.jl
bkamins: ok - removed the whole portion of commented-out tests.

test/deprecated.jl
bkamins: Line 585 was irrelevant - removed. But below it is relevant as we test that `eachcol(df, true)` does not copy columns.

bkamins:
❌ is due to coverage

test/deprecated.jl
nalimilan: I just wanted to ensure none of these tests should be kept for `eachcol(df)` too. But I guess you have checked this.

bkamins: a good point - I have rechecked this and added two tests. Thank you.

bkamins:
The failing build on nightly is due to https://github.com/JuliaLang/julia/issues/34889 so I will merge this.

pablosanjose:
I just made a PR with a fix for JuliaLang/#34889, sorry about that.

bkamins:
@pablosanjose - thank you for such a quick fix.

PR #2122, opened 2020-02-17
export Tables and imoprt DataAPI and Tables internally

### LINKED ISSUE #2114: Should we export Tables
I was thinking if we want to export `Tables` name from `DataFrames`.
What we would get is that I constantly type `using DataFrames, Tables` and the change would simply allow to write `using DataFrames` and get access to `Tables`.

This is just a loose thought. We do not have to do this. Does anyone have some opinion on this?

bkamins:
Fixes #2114.

Given @nalimilan comment I just export Tables.jl.
Also internally I switched to import Tables.jl and DataAPI.jl (not using).

I think it is cleaner if all functionality from these packages is explicitly qualified by package name when used (except we decide to export some names explicitly, as we currently do for: `columnindex`, `All`, `Between` and `describe`)

### DIFF
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -1,14 +1,16 @@
 module DataFrames
 
 using Statistics, Printf, REPL
-using Reexport, SortingAlgorithms, Compat, Unicode, PooledArrays, DataAPI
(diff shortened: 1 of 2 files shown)


bkamins:
Thank you!

PR #2131, opened 2020-02-25
Flatten multiple columns at once

pearlzli:
This is a feature I wanted for `flatten`, to be able to use it on multiple vector-valued columns at a time. I couldn't quite figure out how to word the error message for when the lengths don't match up between the columns (see the `df3` example I've added to the docstring). I definitely welcome your suggestions on that and anything else!

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1682,7 +1682,8 @@ function flatten(df::AbstractDataFrame,
     for col in idxcols
         v = df[!, col]
         if any(x -> length(x[1]) != x[2], zip(v, lengths))
-            throw(ArgumentError("Vector lengths across columns in col, within the same row, must be the same"))
+            r = findfirst(!=(0), length.(v) .- lengths)
+            throw(ArgumentError("Lengths of iterables stored in columns $col1 and $col are not the the same in row $r"))
         end
     end
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1677,10 +1677,11 @@ ERROR: ArgumentError: Vector lengths across columns in col, within the same row,
 function flatten(df::AbstractDataFrame,
                  cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
     idxcols = index(df)[cols]
-    col1 = idxcols isa AbstractVector ? idxcols[1] : idxcols
+    col1 = first(idxcols)
     lengths = length.(df[!, col1])
     for col in idxcols
-        if length.(df[!, col]) != length.(df[!, col1])
+        v = df[!, col]
+        if any(x -> length(x[1]) != x[2], zip(v, lengths))
             throw(ArgumentError("Vector lengths across columns in col, within the same row, must be the same"))
         end
     end
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1602,7 +1602,8 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
 end
 
 """
-    flatten(df::AbstractDataFrame, cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
+    flatten(df::AbstractDataFrame, cols::Union{ColumnIndex, AbstractVector, Regex, Not,
+            Between, All, Colon})
 
 When columns `cols` of data frame `df` have iterable elements that define `length` (for
 example a `Vector` of `Vector`s), return a `DataFrame` where each element of each `col` in
@@ -1671,7 +1672,8 @@ julia> df3 = DataFrame(a = [1, 2], b = [[1, 2], [3, 4]], c = [[5, 6], [7]])
 │ 2   │ 2     │ [3, 4] │ [7]    │
 
 julia> flatten(df3, [:b, :c])
-ERROR: ArgumentError: Vector lengths across columns in col, within the same row, must be the same
+ERROR: ArgumentError: Lengths of iterables stored in columns 2 and 3 are not the the same in
+row 2
 ```
 """
 function flatten(df::AbstractDataFrame,
@@ -1685,7 +1687,8 @@ function flatten(df::AbstractDataFrame,
         v = df[!, col]
         if any(x -> length(x[1]) != x[2], zip(v, lengths))
             r = findfirst(!=(0), length.(v) .- lengths)
-            throw(ArgumentError("Lengths of iterables stored in columns $col1 and $col are not the the same in row $r"))
+            throw(ArgumentError("Lengths of iterables stored in columns $col1 and $col are"
+                                * " not the the same in row $r"))
         end
     end
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1676,6 +1676,8 @@ ERROR: ArgumentError: Vector lengths across columns in col, within the same row,
 """
 function flatten(df::AbstractDataFrame,
                  cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
+    _check_consistency(df)
+
     idxcols = index(df)[cols]
     col1 = first(idxcols)
     lengths = length.(df[!, col1])
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1691,6 +1691,7 @@ function flatten(df::AbstractDataFrame,
     for name in _names(new_df)
         repeat_lengths!(new_df[!, name], df[!, name], lengths)
     end
+    length(idxcols) > 1 && sort!(idxcols)
     for col in idxcols
         col_to_flatten = df[!, col]
         flattened_col = col_to_flatten isa AbstractVector{<:AbstractVector} ?
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1689,8 +1689,8 @@ function flatten(df::AbstractDataFrame,
         if any(x -> length(x[1]) != x[2], zip(v, lengths))
             r = findfirst(x -> x != 0, length.(v) .- lengths)
             colnames = names(df)
-            throw(ArgumentError("Lengths of iterables stored in columns $(colnames[col1])"
-                                 * " and $(colnames[col]) are not the the same in row $r"))
+            throw(ArgumentError("Lengths of iterables stored in columns $(colnames[col1])" *
+                                " and $(colnames[col]) are not the the same in row $r"))
         end
     end
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1602,8 +1602,8 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
 end
 
 """
-    flatten(df::AbstractDataFrame, cols::Union{ColumnIndex, AbstractVector, Regex, Not,
-            Between, All, Colon})
+    flatten(df::AbstractDataFrame,
+            cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
 
 When columns `cols` of data frame `df` have iterable elements that define `length` (for
 example a `Vector` of `Vector`s), return a `DataFrame` where each element of each `col` in
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1688,8 +1688,9 @@ function flatten(df::AbstractDataFrame,
         v = df[!, col]
         if any(x -> length(x[1]) != x[2], zip(v, lengths))
             r = findfirst(x -> x != 0, length.(v) .- lengths)
-            throw(ArgumentError("Lengths of iterables stored in columns $col1 and $col are"
-                                * " not the the same in row $r"))
+            colnames = names(df)
+            throw(ArgumentError("Lengths of iterables stored in columns $(colnames[col1])"
+                                 * " and $(colnames[col]) are not the the same in row $r"))
         end
     end
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1607,7 +1607,7 @@ end
 
 When columns `cols` of data frame `df` have iterable elements that define `length` (for
 example a `Vector` of `Vector`s), return a `DataFrame` where each element of each `col` in
-`cols` is flattened, meaning the column corresponding to `col` becomes a longer `Vector`
+`cols` is flattened, meaning the column corresponding to `col` becomes a longer vector
 where the original entries are concatenated. Elements of row `i` of `df` in columns other
 than `cols` will be repeated according to the length of `df[i, col]`. These lengths must
 therefore be the same for each `col` in `cols`, or else an error is raised. Note that these
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1672,8 +1672,8 @@ julia> df3 = DataFrame(a = [1, 2], b = [[1, 2], [3, 4]], c = [[5, 6], [7]])
 │ 2   │ 2     │ [3, 4] │ [7]    │
 
 julia> flatten(df3, [:b, :c])
-ERROR: ArgumentError: Lengths of iterables stored in columns :b and :c are not the the same in
-row 2
+ERROR: ArgumentError: Lengths of iterables stored in columns :b and :c
+are not the the same in row 2
 ```
 """
 function flatten(df::AbstractDataFrame,
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1602,35 +1602,47 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
 end
 
 """
-    flatten(df::AbstractDataFrame, col::Union{Integer, Symbol})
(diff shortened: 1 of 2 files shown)


bkamins:
Thank you for contributing.

If I remember correctly we did not add the multiple-columns option because it was not clear what should be the the behaviour then:
* do what you propose (each flattened column should have the same length and we kind of "broadcast" the operation)
* do not require equal lengths and perform a Cartesian product of the columns indicated to be flattened

Before we move forward with this PR I would wait for people to comment which functionality would be more useful (in general we could have both via a keyword argument, but then the question would be what should be the default).

pearlzli:
Sure, that sounds good! The second behavior you mentioned also sounds reasonable. My small case for the first one (to be the default): unlike the second, it can't be replicated by multiple calls to `flatten`.

bkamins:
Makes sense.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1602,35 +1602,47 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+    flatten(df::AbstractDataFrame, cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
 
-When column `col` of data frame `df` has iterable elements that define `length` (for example
-a `Vector` of `Vector`s), return a `DataFrame` where each element of `col` is flattened, meaning
-the column corresponding to `col` becomes a longer `Vector` where the original entries
-are concatenated. Elements of row `i` of `df` in columns other than `col` will be repeated
-according to the length of `df[i, col]`. Note that these elements are not copied,
-and thus if they are mutable changing them in the returned `DataFrame` will affect `df`.
+When columns `cols` of data frame `df` have iterable elements that define `length` (for
+example a `Vector` of `Vector`s), return a `DataFrame` where each element of each `col` in
```
```suggestion
example a `Vector` of `Vector`s), return a `DataFrame` where each element of each column `col` in
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1602,35 +1602,47 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+ `cols` is flattened, meaning the column corresponding to `col` becomes a longer `Vector`
+where the original entries are concatenated. Elements of row `i` of `df` in columns other
+than `cols` will be repeated according to the length of `df[i, col]`. These lengths must
+therefore be the same for each `col` in `cols`, or else an error is raised. Note that these
+ elements are not copied, and thus if they are mutable changing them in the returned
```
```suggestion
elements are not copied, and thus if they are mutable changing them in the returned
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1602,35 +1602,47 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
+ `DataFrame` will affect `df`.
```
```suggestion
`DataFrame` will affect `df`.
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1602,35 +1602,47 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
 
-When column `col` of data frame `df` has iterable elements that define `length` (for example
-a `Vector` of `Vector`s), return a `DataFrame` where each element of `col` is flattened, meaning
-the column corresponding to `col` becomes a longer `Vector` where the original entries
-are concatenated. Elements of row `i` of `df` in columns other than `col` will be repeated
-according to the length of `df[i, col]`. Note that these elements are not copied,
-and thus if they are mutable changing them in the returned `DataFrame` will affect `df`.
+When columns `cols` of data frame `df` have iterable elements that define `length` (for
+example a `Vector` of `Vector`s), return a `DataFrame` where each element of each `col` in
+ `cols` is flattened, meaning the column corresponding to `col` becomes a longer `Vector`
```
```suggestion
`cols` is flattened, meaning the column corresponding to `col` becomes a longer `Vector`
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,41 @@ julia> flatten(df2, :b)
...
 ```
 """
-function flatten(df::AbstractDataFrame, col::ColumnIndex)
-    col_to_flatten = df[!, col]
-    lengths = length.(col_to_flatten)
-    new_df = similar(df[!, Not(col)], sum(lengths))
+function flatten(df::AbstractDataFrame,
+                 cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
+    idxcols = index(df)[cols]
+    col1 = idxcols isa AbstractVector ? idxcols[1] : idxcols
```
bkamins: you have to add
`isempty(idxcols) && return copy(df)`
before this line.
also can you please add an appropriate test for empty column selection?

```suggestion
    col1 = first(idxcols)
```
is enough

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,41 @@ julia> flatten(df2, :b)
...
+    lengths = length.(df[!, col1])
+    for col in idxcols
+        if length.(df[!, col]) != length.(df[!, col1])
```
```suggestion
        v = df[!, col]
        if any(x -> length(x[1]) != x[2], zip(v, lengths))
```
will allocate less and be faster

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,41 @@ julia> flatten(df2, :b)
...
+            throw(ArgumentError("Vector lengths across columns in col, within the same row, must be the same"))
```
```suggestion
            r = findfirst(!=(0), length.(v) .- lengths)
            throw(ArgumentError("Lengths of iterables stored in columns $col1 and $col are not the the same in row $r"))
```
(here we do not have to be super efficient as this will be executed only on error and we prefer a more informative error message.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,41 @@ julia> flatten(df2, :b)
...
+        end
+    end
 
+    new_df = similar(df[!, Not(cols)], sum(lengths))
     for name in _names(new_df)
         repeat_lengths!(new_df[!, name], df[!, name], lengths)
     end
+    for col in idxcols
```
bkamins: before this line add:
```
length(idxcols) > 1 && sort!(idxcols)
```
otherwise the result will be incorrect if the user passed columns in the order that is not increasing as in the original data frame.

Also could you please add a test checking this specific case? Thank you.

test/reshape.jl
```diff
@@ -396,6 +396,22 @@ end
     @test df_flat_cat.b isa CategoricalArray
 end
 
+@testset "flatten multiple columns" begin
+    df = DataFrame(a = [1, 2], b = [[1, 2], [3, 4]], c = [[5, 6], [7, 8]])
+    ref = DataFrame(a = [1, 1, 2, 2], b = [1, 2, 3, 4], c = [5, 6, 7, 8])
+    @test flatten(df, [:b, :c])        == ref
```
bkamins: e.g. here add the test `flatten(df, [:c, :b])`.

Also do not align `==` just use a single space

bkamins: e.g. here add the test `flatten(df, [:c, :b])`.

Also do not align `==` just use a single space

bkamins: and a test for `flatten(df, [])`

bkamins:
Thank you for the PR. I have left some comments that I think would be good to consider before we move forward.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,44 @@ julia> flatten(df2, :b)
...
+julia> df3 = DataFrame(a = [1, 2], b = [[1, 2], [3, 4]], c = [[5, 6], [7]])
+2×3 DataFrame
+│ Row │ a     │ b      │ c      │
+│     │ Int64 │ Array… │ Array… │
+├─────┼───────┼────────┼────────┤
+│ 1   │ 1     │ [1, 2] │ [5, 6] │
+│ 2   │ 2     │ [3, 4] │ [7]    │
+
+julia> flatten(df3, [:b, :c])
+ERROR: ArgumentError: Vector lengths across columns in col, within the same row, must be the same
```
bkamins: outdated error message

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,44 @@ julia> flatten(df2, :b)
...
+function flatten(df::AbstractDataFrame,
+                 cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
+    idxcols = index(df)[cols]
+    col1 = first(idxcols)
+    lengths = length.(df[!, col1])
+    for col in idxcols
+        v = df[!, col]
+        if any(x -> length(x[1]) != x[2], zip(v, lengths))
+            r = findfirst(!=(0), length.(v) .- lengths)
+            throw(ArgumentError("Lengths of iterables stored in columns $col1 and $col are not the the same in row $r"))
```
bkamins: can you please break this line (and I think also above one line also) so that it is not longer that 92 chars, as we try to stick to this limit. Thank you.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,44 @@ julia> flatten(df2, :b)
...
-    new_df = similar(df[!, Not(col)], sum(lengths))
+function flatten(df::AbstractDataFrame,
+                 cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
+    idxcols = index(df)[cols]
+    col1 = first(idxcols)
+    lengths = length.(df[!, col1])
+    for col in idxcols
+        v = df[!, col]
+        if any(x -> length(x[1]) != x[2], zip(v, lengths))
+            r = findfirst(!=(0), length.(v) .- lengths)
```
bkamins: as an afterthought, I do not remember if `!=(0)` is defined on Julia 1.0. If CI fails on 1.0 we need to change it to `x -> x != 0`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,44 @@ julia> flatten(df2, :b)
...
+ERROR: ArgumentError: Vector lengths across columns in col, within the same row, must be the same
 ```
 """
-function flatten(df::AbstractDataFrame, col::ColumnIndex)
-    col_to_flatten = df[!, col]
-    lengths = length.(col_to_flatten)
-    new_df = similar(df[!, Not(col)], sum(lengths))
+function flatten(df::AbstractDataFrame,
+                 cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
+    idxcols = index(df)[cols]
```
bkamins: still the `isempty(idxcols)` special case (and relevant test in tests) is missing.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1662,44 @@ julia> flatten(df2, :b)
...
+julia> flatten(df3, [:b, :c])
+ERROR: ArgumentError: Vector lengths across columns in col, within the same row, must be the same
 ```
 """
-function flatten(df::AbstractDataFrame, col::ColumnIndex)
-    col_to_flatten = df[!, col]
-    lengths = length.(col_to_flatten)
-    new_df = similar(df[!, Not(col)], sum(lengths))
+function flatten(df::AbstractDataFrame,
+                 cols::Union{ColumnIndex, AbstractVector, Regex, Not, Between, All, Colon})
```
bkamins: since `flatten` is an expensive operation on all columns I think it is safer to add `_check_consistency(df)` as a first operation in this function as it will add only a little overhead, and we later heavily rely on the fact that the data frame is conistent.

pearlzli:
Have to step away for a few hours, but will finish incorporating your suggestions later today. Thanks again!

src/abstractdataframe/abstractdataframe.jl
pearlzli: Tested it on my computer with Julia 1.0, and you're right that the newer syntax doesn't work.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1602,35 +1602,48 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
-the column corresponding to `col` becomes a longer `Vector` where the original entries
-are concatenated. Elements of row `i` of `df` in columns other than `col` will be repeated
-according to the length of `df[i, col]`. Note that these elements are not copied,
-and thus if they are mutable changing them in the returned `DataFrame` will affect `df`.
+    flatten(df::AbstractDataFrame, cols::Union{ColumnIndex, AbstractVector, Regex, Not,
+            Between, All, Colon})
+
+When columns `cols` of data frame `df` have iterable elements that define `length` (for
+example a `Vector` of `Vector`s), return a `DataFrame` where each element of each `col` in
+`cols` is flattened, meaning the column corresponding to `col` becomes a longer `Vector`
```
```suggestion
`cols` is flattened, meaning the column corresponding to `col` becomes a longer vector
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1602,35 +1602,48 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
...
-    flatten(df::AbstractDataFrame, col::Union{Integer, Symbol})
-
-When column `col` of data frame `df` has iterable elements that define `length` (for example
-a `Vector` of `Vector`s), return a `DataFrame` where each element of `col` is flattened, meaning
-the column corresponding to `col` becomes a longer `Vector` where the original entries
-are concatenated. Elements of row `i` of `df` in columns other than `col` will be repeated
-according to the length of `df[i, col]`. Note that these elements are not copied,
-and thus if they are mutable changing them in the returned `DataFrame` will affect `df`.
+    flatten(df::AbstractDataFrame, cols::Union{ColumnIndex, AbstractVector, Regex, Not,
+            Between, All, Colon})
```
nalimilan: Either keep this on the previous line, align them with the character after `Union{`, or add a line break before `cols` instead.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1663,49 @@ julia> flatten(df2, :b)
...
+
+    idxcols = index(df)[cols]
+    isempty(idxcols) && return copy(df)
+    col1 = first(idxcols)
+    lengths = length.(df[!, col1])
+    for col in idxcols
+        v = df[!, col]
+        if any(x -> length(x[1]) != x[2], zip(v, lengths))
+            r = findfirst(x -> x != 0, length.(v) .- lengths)
+            throw(ArgumentError("Lengths of iterables stored in columns $col1 and $col are"
```
nalimilan: It would be nice to print the name or indices of columns, as specified in `cols`, rather than always the indices.

bkamins: OK - now I see the ambiguity. However `cols` can be anything, so we should use column names then.

src/abstractdataframe/abstractdataframe.jl
bkamins: I would break before `cols` - just as in the method definition.

pearlzli:
Thanks for your comments, @nalimilan! I think I've addressed them now.

bkamins:
Looks good. Thank you!

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1663,50 @@ julia> flatten(df2, :b)
...
+julia> df3 = DataFrame(a = [1, 2], b = [[1, 2], [3, 4]], c = [[5, 6], [7]])
+2×3 DataFrame
+│ Row │ a     │ b      │ c      │
+│     │ Int64 │ Array… │ Array… │
+├─────┼───────┼────────┼────────┤
+│ 1   │ 1     │ [1, 2] │ [5, 6] │
+│ 2   │ 2     │ [3, 4] │ [7]    │
+
+julia> flatten(df3, [:b, :c])
+ERROR: ArgumentError: Lengths of iterables stored in columns 2 and 3 are not the the same in
```
```suggestion
ERROR: ArgumentError: Lengths of iterables stored in columns :b and :c are not the the same in
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1663,50 @@ julia> flatten(df2, :b)
...
+    idxcols = index(df)[cols]
+    isempty(idxcols) && return copy(df)
+    col1 = first(idxcols)
+    lengths = length.(df[!, col1])
+    for col in idxcols
+        v = df[!, col]
+        if any(x -> length(x[1]) != x[2], zip(v, lengths))
+            r = findfirst(x -> x != 0, length.(v) .- lengths)
+            colnames = names(df)
+            throw(ArgumentError("Lengths of iterables stored in columns $(colnames[col1])" *
```
```suggestion
            throw(ArgumentError("Lengths of iterables stored in columns :$(colnames[col1])" *
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1663,50 @@ julia> flatten(df2, :b)
...
+                                " and $(colnames[col]) are not the the same in row $r"))
```
```suggestion
                                " and :$(colnames[col]) are not the the same in row $r"))
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1663,50 @@ julia> flatten(df2, :b)
...
+julia> df3 = DataFrame(a = [1, 2], b = [[1, 2], [3, 4]], c = [[5, 6], [7]])
+2×3 DataFrame
+│ Row │ a     │ b      │ c      │
+│     │ Int64 │ Array… │ Array… │
+├─────┼───────┼────────┼────────┤
+│ 1   │ 1     │ [1, 2] │ [5, 6] │
+│ 2   │ 2     │ [3, 4] │ [7]    │
+
+julia> flatten(df3, [:b, :c])
+ERROR: ArgumentError: Lengths of iterables stored in columns :b and :c are not the the same in
```
```suggestion
ERROR: ArgumentError: Lengths of iterables stored in columns :b and :c
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1650,22 +1663,50 @@ julia> flatten(df2, :b)
...
+row 2
```
```suggestion
are not the the same in row 2
```

bkamins:
Thank you!

pearlzli:
Thanks to both of you as well!

Issue #2095, opened 2020-01-31
Speed up key lookup in GroupedDataFrame


bkamins:
When https://github.com/JuliaData/DataFrames.jl/pull/2046 is merged I think we should redesign https://github.com/JuliaData/DataFrames.jl/pull/2046/files#diff-6349421a054bb7e74b7f27ae86304cbaR275 (but I did not want to postpone merging of that PR).

What I mean is that on the first call to this function (or maybe even earlier when we call `compute_indices`) we should build a dictionary from `Tuples` to integer indices in a `GroupedDataFrame`. Then the lookup of Tuple would be very fast if done repeatedly.

If we did this we could actually say that the way to add an index to a `DataFrame`, which people very often ask for, is to simply run `groupby` on it. Then you would have a very fast lookup of the index which is a common operation.

@nalimilan - do you have any thoughts about it (i.e. if we want it - as it would eat up some memory + what would be the best way to implement it - that is: at what moment it would be best to populate such a dictionary)?

nalimilan:
Yes that linear search is really bad for performance.

Actually we already have a hash table structure: the `gslots` vector returned by `row_group_slots` allows mapping a hash to the index of the first row in the group, and the `groups` vector allows getting the index of the group corresponding to that row.

But currently we don't store it in `GroupedDataFrame` as it's not used. Keeping that information would only use a significant amount of memory when the number of groups is large, so maybe that's OK. We could add an option to discard it.

bkamins:
> Actually we already have a hash table structure

This is what I thought 😄 (judging by the logic of what we do), but I just do not have enough experience with this part of code base to be sure how exactly this can be achieved.

> We could add an option to discard it.

This is exactly what I thought. If someone uses `by` we do not need to store them as `GroupedDataFrame` gets discarded anyway immediately.

Summing up: do you have time to make a PR for this? (if not - I can have a look). Thank you!

nalimilan:
One subtlety I hadn't spotted is that we have specialized methods for categorical and pooled arrays, which doesn't create the hash table. And creating it would be quite wasteful when it's not used. So we should probably do the same as for group indices: set it to `nothing`, and only create it on first access.

I probably won't have the time soon, so go ahead.

bkamins:
Agreed, but only in case hash table is not created (if hash table is created then I understand it is best to create the mapping immediately). Right?

nalimilan:
[...]

PR #2132, opened 2020-02-26
use Dict{Any,Int} to speed up GroupDataFrame lookup

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2095.

I used `Dict{Any, Int}` interface which is fast enough I think. In the future this can be improved as @nalimilan noted, but already we have huge performance gains here and the end-result makes `GroupedDataFrame` practically usable as an indexed `DataFrame`.

### DIFF
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -172,13 +172,14 @@ function Base.getindex(gd::GroupedDataFrame, idxs::AbstractVector{<:Integer})
         end
     end
     GroupedDataFrame(gd.parent, gd.cols, new_groups, gd.idx,
-                     new_starts, new_ends, length(new_starts))
+                     new_starts, new_ends, length(new_starts), nothing)
 end
 
 # Index with colon (creates copy)
 Base.getindex(gd::GroupedDataFrame, idxs::Colon) =
-    GroupedDataFrame(gd.parent, gd.cols, gd.groups, gd.idx,
-                     gd.starts, gd.ends, gd.ngroups)
+    GroupedDataFrame(gd.parent, gd.cols, gd.groups, getfield(gd, :idx),
+                     getfield(gd, :starts), getfield(gd, :ends), gd.ngroups,
+                     getfield(gd, :keymap))
 
 
 #
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -22,10 +22,10 @@ mutable struct GroupedDataFrame{T<:AbstractDataFrame}
 end
 
 function genkeymap(gd, cols)
-# currently we use Dict{Any,Int} because then filed :keymap in GroupedDataFrame has a concrete
-# type which makes the access to it faster as we do not have a dynamic dispatch when indexing
-# into it. In the future an optimization of this approach can be investigated
-# (also taking compilation time into account).
+    # currently we use Dict{Any,Int} because then filed :keymap in GroupedDataFrame
+    # has a concrete type which makes the access to it faster as we do not have a dynamic
+    # dispatch when indexing into it. In the future an optimization of this approach
+    # can be investigated (also taking compilation time into account).
     d = Dict{Any,Int}()
     gdidx = gd.idx
     sizehint!(d, length(gd.starts))
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -22,6 +22,10 @@ mutable struct GroupedDataFrame{T<:AbstractDataFrame}
 end
 
 function genkeymap(gd, cols)
+# currently we use Dict{Any,Int} because then filed :keymap in GroupedDataFrame has a concrete
+# type which makes the access to it faster as we do not have a dynamic dispatch when indexing
+# into it. In the future an optimization of this approach can be investigated
+# (also taking compilation time into account).
     d = Dict{Any,Int}()
     gdidx = gd.idx
     sizehint!(d, length(gd.starts))
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -41,6 +41,7 @@ function Base.getproperty(gd::GroupedDataFrame, f::Symbol)
         if getfield(gd, f) === nothing
             gd.keymap = genkeymap(gd, ntuple(i -> parent(gd)[!, gd.cols[i]], length(gd.cols)))
         end
+        return getfield(gd, f)::Dict{Any,Int}
     else
         return getfield(gd, f)
     end
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -22,7 +22,7 @@ mutable struct GroupedDataFrame{T<:AbstractDataFrame}
 end
 
 function genkeymap(gd, cols)
-    # currently we use Dict{Any,Int} because then filed :keymap in GroupedDataFrame
+    # currently we use Dict{Any,Int} because then field :keymap in GroupedDataFrame
     # has a concrete type which makes the access to it faster as we do not have a dynamic
     # dispatch when indexing into it. In the future an optimization of this approach
     # can be investigated (also taking compilation time into account).
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -12,12 +12,22 @@ Not meant to be constructed directly, see `groupby`.
 """
 mutable struct GroupedDataFrame{T<:AbstractDataFrame}
     parent::T
-    cols::Vector{Int}                  # columns used for grouping
-    groups::Vector{Int}                # group indices for each row in 0:ngroups, 0 skipped
-    idx::Union{Vector{Int},Nothing}    # indexing vector sorting rows into groups
-    starts::Union{Vector{Int},Nothing} # starts of groups after permutation by idx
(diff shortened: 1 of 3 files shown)


src/groupeddataframe/groupeddataframe.jl
```diff
@@ -158,13 +173,14 @@ function Base.getindex(gd::GroupedDataFrame, idxs::AbstractVector{<:Integer})
...
     GroupedDataFrame(gd.parent, gd.cols, new_groups, gd.idx,
-                     new_starts, new_ends, length(new_starts))
+                     new_starts, new_ends, length(new_starts), nothing)
 end
 
 # Index with colon (creates copy)
 Base.getindex(gd::GroupedDataFrame, idxs::Colon) =
-    GroupedDataFrame(gd.parent, gd.cols, gd.groups, gd.idx,
-                     gd.starts, gd.ends, gd.ngroups)
+    GroupedDataFrame(gd.parent, gd.cols, gd.groups, getfield(gd, :idx),
```
bkamins: @nalimilan - this is partially not related to this PR, but I introduce `getfield` instead of `getpropery` to avoid materializing indices when not needed. I assume this is what we want, right?.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -12,12 +12,22 @@ Not meant to be constructed directly, see `groupby`.
...
+    starts::Union{Vector{Int},Nothing}   # starts of groups after permutation by idx
+    ends::Union{Vector{Int},Nothing}     # ends of groups after permutation by idx
+    ngroups::Int                         # number of groups
+    keymap::Union{Dict{Any,Int},Nothing} # mapping of key tuples to group indices
+end
+
+function genkeymap(gd, cols)
+    d = Dict{Any,Int}()
+    sizehint!(d, length(gd.starts))
+    for (i, s) in enumerate(gd.starts)
```
nalimilan: I guess this wouldn't hurt?
```suggestion
    @inbounds for (i, s) in enumerate(gd.starts)
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -27,6 +37,11 @@ function Base.getproperty(gd::GroupedDataFrame, f::Symbol)
             gd.idx, gd.starts, gd.ends = compute_indices(gd.groups, gd.ngroups)
         end
         return getfield(gd, f)::Vector{Int}
+    elseif f == :keymap
```
nalimilan:
```suggestion
    elseif f === :keymap
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -27,6 +37,11 @@ function Base.getproperty(gd::GroupedDataFrame, f::Symbol)
...
+        if getfield(gd, f) === nothing
```
nalimilan: Just for consistency with other fields:
```suggestion
        # Keymap is computed lazily the first time it is accessed
        if getfield(gd, f) === nothing
```

nalimilan: Just for consistency with other fields:
```suggestion
        # Keymap is computed lazily the first time it is accessed
        if getfield(gd, f) === nothing
```

bkamins: duplicate

bkamins:
In your proposed change `@inbounds` should be OK so I have applied it. The only downside is that we will silently populate garbage in `d` if `parent(gd)` was mutated after `gd.starts` was created. The performance gain will be low (as `keymap` stores keys in `Any`).

So: I have applied your suggestion, but if you find that what I have written above is convincing I will remove `@inbounds` - so please let me know what you think. Thank you!

Regarding the size of the change - this is exactly why I wanted `Dict{Any, Int}` - the PR is small and rather simple, and probably in 90% of cases is near optimal, and we have many different things to finish. Further optimizations can be done post 1.0.

bkamins:
I have made a bad bug with indexing using `starts` and not using `idx` in this PR 😢. I will fix it in the evening (and it is not good that current tests did not catch this - which means that they have to be significantly improved, as they make assumptions that the data frames that are grouped have some special structure - as usual - I will go for my favorite randomized tests to make sure we do not make any such assumptions in the testcases 😄).

nalimilan:
> In your proposed change `@inbounds` should be OK so I have applied it. The only downside is that we will silently populate garbage in `d` if `parent(gd)` was mutated after `gd.starts` was created. The performance gain will be low (as `keymap` stores keys in `Any`).
> 
> So: I have applied your suggestion, but if you find that what I have written above is convincing I will remove `@inbounds` - so please let me know what you think. Thank you!

If the performance gain is small then yes, maybe better not use `@inbounds`.

> I have made a bad bug with indexing using `starts` and not using `idx` in this PR cry. I will fix it in the evening (and it is not good that current tests did not catch this - which means that they have to be significantly improved, as they make assumptions that the data frames that are grouped have some special structure - as usual - I will go for my favorite randomized tests to make sure we do not make any such assumptions in the testcases smile).

Woops. Does that mean the tests use only one row per group? That's indeed quite restrictive.

bkamins:
Regarding timing of `@inbounds` the difference is minimal, so I will remove it.
For 1,000,000 groups this is timing with `@inbounds`:
```
BenchmarkTools.Trial: 
  memory estimate:  132.68 MiB
  allocs estimate:  3375484
  --------------
  minimum time:     481.781 ms (0.00% GC)
  median time:      590.817 ms (21.87% GC)
  mean time:        601.083 ms (21.64% GC)
  maximum time:     693.690 ms (31.86% GC)
  --------------
  samples:          9
  evals/sample:     1
```
and this is without it:
```
BenchmarkTools.Trial: 
  memory estimate:  132.68 MiB
  allocs estimate:  3375484
  --------------
  minimum time:     477.255 ms (0.00% GC)
  median time:      584.295 ms (21.89% GC)
  mean time:        593.354 ms (21.69% GC)
  maximum time:     684.765 ms (31.66% GC)
  --------------
  samples:          9
  evals/sample:     1
```
so it is around 1%.

test/grouping.jl
```diff
@@ -1478,29 +1478,35 @@ end
...
-        @test gd[Array{Any}(a)] ≅ gd2
-
-        # Duplicate keys
-        a2 = converter.(keys(gd)[[1, 2, 1]])
-        @test_throws ArgumentError gd[a2]
+    df_ref = DataFrame(a = repeat([:A, :B, missing], outer=4),
+                       b = repeat(1:2, inner=6), c = 1:12)
+    Random.seed!(1234)
+    for df in [df_ref, df_ref[randperm(nrow(df_ref)), :]], grpcols = [[:a, :b], :a, :b],
+              dosort in [true, false], doskipmissing in [true, false]
```
```suggestion
               dosort in [true, false], doskipmissing in [true, false]
```

test/grouping.jl
```diff
@@ -1478,29 +1478,35 @@ end
...
-        @test gd[Array{Any}(a)] ≅ gd2
-
-        # Duplicate keys
-        a2 = converter.(keys(gd)[[1, 2, 1]])
-        @test_throws ArgumentError gd[a2]
+    df_ref = DataFrame(a = repeat([:A, :B, missing], outer=4),
+                       b = repeat(1:2, inner=6), c = 1:12)
+    Random.seed!(1234)
+    for df in [df_ref, df_ref[randperm(nrow(df_ref)), :]], grpcols = [[:a, :b], :a, :b],
+               dosort in [true, false], doskipmissing in [true, false]
```
bkamins:
```suggestion
        dosort in [true, false], doskipmissing in [true, false]
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -12,12 +12,23 @@ Not meant to be constructed directly, see `groupby`.
...
+    cols::Vector{Int}                    # columns used for grouping
+    groups::Vector{Int}                  # group indices for each row in 0:ngroups, 0 skipped
+    idx::Union{Vector{Int},Nothing}      # indexing vector sorting rows into groups
+    starts::Union{Vector{Int},Nothing}   # starts of groups after permutation by idx
+    ends::Union{Vector{Int},Nothing}     # ends of groups after permutation by idx
+    ngroups::Int                         # number of groups
+    keymap::Union{Dict{Any,Int},Nothing} # mapping of key tuples to group indices
+end
+
+function genkeymap(gd, cols)
```
bkamins:
```suggestion
function genkeymap(gd, cols)
# currently we use Dict{Any,Int} because then filed :keymap in GroupedDataFrame has a concrete
# type which makes the access to it faster as we do not have a dynamic dispatch when indexing
# into it. In the future an optimization of this approach can be investigated
# (also taking compilation time into account).
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -12,12 +12,22 @@ Not meant to be constructed directly, see `groupby`.
...
+    d = Dict{Any,Int}()
```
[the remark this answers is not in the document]
bkamins: added

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -22,10 +22,10 @@ mutable struct GroupedDataFrame{T<:AbstractDataFrame}
 end
 
 function genkeymap(gd, cols)
-# currently we use Dict{Any,Int} because then filed :keymap in GroupedDataFrame has a concrete
-# type which makes the access to it faster as we do not have a dynamic dispatch when indexing
-# into it. In the future an optimization of this approach can be investigated
-# (also taking compilation time into account).
+    # currently we use Dict{Any,Int} because then filed :keymap in GroupedDataFrame
```
nalimilan:
```suggestion
    # currently we use Dict{Any,Int} because then field :keymap in GroupedDataFrame
```

PR #2140, opened 2020-03-04
change ⍰ to ? when showing a DataFrame with missing eltype

### LINKED ISSUE #2139: ⍰ character in header output
I was wondering what the odd little character in the header output was:

Screenshot 2020-03-04 at 11 31 40

I thought at first there was an error with the font - since it's more or less the same as the `.notdef` character which is displayed when a font can't display a suitable glyph. (See [here](https://docs.microsoft.com/en-us/typography/opentype/spec/recom#glyph-0-the-notdef-glyph) for more info.)

I found out that it's an APL SYMBOL QUAD QUESTION, '⍰': Unicode U+2370 (category So: Symbol, other). 

So given that it's not entirely clear whether there's a "problem" with the dataframe or the font (and assuming I'm not the only person daft enough to wonder), is this the wrong choice of character and would some other symbol be more suitable?

An obvious suggestion is an ordinary question mark or exclamation mark. Unicode gives many  possibilities - ⁈ (2048), ⁉︎ (2049). Or even the interrobang ‽ (203D). (A mixture of "where? and watch out!). 

```
5×3 DataFrame
│ Row │ i     │ x       │ y       │
│     │ Int64 │ Int64⁈  │ String⁈ │
├─────┼───────┼─────────┼─────────┤
│ 1   │ 1     │ missing │ missing │
│ 2   │ 2     │ 4       │ missing │
│ 3   │ 3     │ missing │ c       │
│ 4   │ 4     │ 2       │ d       │
│ 5   │ 5     │ 1       │ e       │
```

These would presumably fall back to `?` anyway... Just don't use emojis..🤣

bkamins:
Fixes #2139

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -306,7 +306,7 @@ julia> DataFrame(a = 1:2, b = [1.0, missing],
                  c = categorical('a':'b'), d = [1//2, missing])
 2×4 DataFrame
 │ Row │ a     │ b        │ c            │ d         │
-│     │ Int64 │ Float64⍰ │ Categorical… │ Rationa…⍰ │
+│     │ Int64 │ Float64? │ Categorical… │ Rationa…? │
 ├─────┼───────┼──────────┼──────────────┼───────────┤
 │ 1   │ 1     │ 1.0      │ 'a'          │ 1//2      │
 │ 2   │ 2     │ missing  │ 'b'          │ missing   │
@@ -315,7 +315,7 @@ julia> DataFrame(a = 1:2, b = [1.0, missing],
 we can observe that:
 
 * the first column `:a` can hold elements of type `Int64`;
-* the second column `:b` can hold `Float64` or `Missing`, which is indicated by `⍰` printed after the name of type;
+* the second column `:b` can hold `Float64` or `Missing`, which is indicated by `?` printed after the name of type;
 * the third column `:c` can hold categorical data; here we notice `…`, which indicates that the actual name of the type was long and got truncated;
 * the type information in fourth column `:d` presents a situation where the name is both truncated and the type allows `Missing`.
 
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -305,11 +305,11 @@ Also notice that when `DataFrame` is printed to the console or rendered in HTML 
 julia> DataFrame(a = 1:2, b = [1.0, missing],
                  c = categorical('a':'b'), d = [1//2, missing])
 2×4 DataFrame
-│ Row │ a     │ b        │ c    │ d         │
-│     │ Int64 │ Float64? │ Cat… │ Rationa…? │
-├─────┼───────┼──────────┼──────┼───────────┤
-│ 1   │ 1     │ 1.0      │ 'a'  │ 1//2      │
-│ 2   │ 2     │ missing  │ 'b'  │ missing   │
+│ Row │ a     │ b        │ c    │ d        │
+│     │ Int64 │ Float64? │ Cat… │ Ration…? │
+├─────┼───────┼──────────┼──────┼──────────┤
+│ 1   │ 1     │ 1.0      │ 'a'  │ 1//2     │
+│ 2   │ 2     │ missing  │ 'b'  │ missing  │
 ```
 
 we can observe that:
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -305,11 +305,11 @@ Also notice that when `DataFrame` is printed to the console or rendered in HTML 
 julia> DataFrame(a = 1:2, b = [1.0, missing],
                  c = categorical('a':'b'), d = [1//2, missing])
 2×4 DataFrame
-│ Row │ a     │ b        │ c    │ d        │
-│     │ Int64 │ Float64? │ Cat… │ Ration…? │
-├─────┼───────┼──────────┼──────┼──────────┤
-│ 1   │ 1     │ 1.0      │ 'a'  │ 1//2     │
-│ 2   │ 2     │ missing  │ 'b'  │ missing  │
+│ Row │ a     │ b        │ c    │ d         │
+│     │ Int64 │ Float64? │ Cat… │ Rationa…? │
+├─────┼───────┼──────────┼──────┼───────────┤
+│ 1   │ 1     │ 1.0      │ 'a'  │ 1//2      │
+│ 2   │ 2     │ missing  │ 'b'  │ missing   │
 ```
 
 we can observe that:
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -305,11 +305,11 @@ Also notice that when `DataFrame` is printed to the console or rendered in HTML 
 julia> DataFrame(a = 1:2, b = [1.0, missing],
                  c = categorical('a':'b'), d = [1//2, missing])
 2×4 DataFrame
-│ Row │ a     │ b        │ c            │ d         │
-│     │ Int64 │ Float64? │ Categorical… │ Rationa…? │
-├─────┼───────┼──────────┼──────────────┼───────────┤
-│ 1   │ 1     │ 1.0      │ 'a'          │ 1//2      │
-│ 2   │ 2     │ missing  │ 'b'          │ missing   │
+│ Row │ a     │ b        │ c    │ d         │
+│     │ Int64 │ Float64? │ Cat… │ Rationa…? │
+├─────┼───────┼──────────┼──────┼───────────┤
+│ 1   │ 1     │ 1.0      │ 'a'  │ 1//2      │
+│ 2   │ 2     │ missing  │ 'b'  │ missing   │
 ```
 
 we can observe that:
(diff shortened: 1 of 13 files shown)


docs/src/man/getting_started.md
```diff
@@ -306,7 +306,7 @@ julia> DataFrame(a = 1:2, b = [1.0, missing],
                  c = categorical('a':'b'), d = [1//2, missing])
 2×4 DataFrame
 │ Row │ a     │ b        │ c            │ d         │
-│     │ Int64 │ Float64⍰ │ Categorical… │ Rationa…⍰ │
+│     │ Int64 │ Float64? │ Categorical… │ Rationa…? │
```
nalimilan: While we're breaking things, maybe worth also revisiting the printing of `Categorical` so that users don't have to adjust doctests in the next release again? We discussed printing just `Cat` to save some space.

bkamins: OK - changed. I will hold off with merging until you have a look and approve the change from `Categorical` to `Cat`.

(there are extensive changes to the manual because of this change - and you will see there the visual consequences of the change to assess it)

docs/src/man/joins.md
```diff
@@ -241,12 +241,12 @@ julia> b = DataFrame(ID = [20, 60], Job = ["Lawyer", "Doctor"])
...
 julia> outerjoin(a, b, on=:ID, validate=(true, true), indicator=:source)
 3×4 DataFrame
-│ Row │ ID    │ Name    │ Job     │ source       │
-│     │ Int64 │ String⍰ │ String⍰ │ Categorical… │
-├─────┼───────┼─────────┼─────────┼──────────────┤
-│ 1   │ 20    │ John    │ Lawyer  │ both         │
-│ 2   │ 40    │ Jane    │ missing │ left_only    │
-│ 3   │ 60    │ missing │ Doctor  │ right_only   │
+│ Row │ ID    │ Name    │ Job     │ source     │
+│     │ Int64 │ String? │ String? │ Cat…       │
```
nalimilan: It would be nice to use the available space here, as for other types, and print `Categoric…`.

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -268,27 +303,27 @@ julia> d = stack(iris, view=true);
...
-│     │ Symbol      │ Float64⍰ │ Categorical…⍰ │ Int64 │
-├─────┼─────────────┼──────────┼───────────────┼───────┤
-│ 1   │ SepalLength │ 5.1      │ setosa        │ 1     │
-│ 2   │ SepalLength │ 4.9      │ setosa        │ 2     │
-│ 3   │ SepalLength │ 4.7      │ setosa        │ 3     │
-│ 4   │ SepalLength │ 4.6      │ setosa        │ 4     │
-│ 5   │ SepalLength │ 5.0      │ setosa        │ 5     │
-│ 6   │ SepalLength │ 5.4      │ setosa        │ 6     │
+│ Row │ variable    │ value │ Species     │ id    │
+│     │ Cat…        │ Any   │ String      │ Int64 │
```
nalimilan: `Any` sounds like a bug in `stack`?

bkamins: Yes - there was a bud. I fix it in https://github.com/JuliaData/DataFrames.jl/pull/2145. We should merge this PR first, and then I will update #2145 to reflect the changes here.

docs/src/man/joins.md
bkamins: I can fix it but let us decide on the rules of abbreviation of `Categorical`, because this was the problem at the time when this was discussed originally.

Do we print `Cat`/`Categoric`/`Categorical` (or some other list) or allow to shorten `Categorical` at any letter?

nalimilan: Sorry, my suggestion was clear. I meant that we should print `Cat…` when the column isn't wider than that, but use all the available space when there is (just like for other types).

bkamins: > but use all the available space when there is (just like for other types).

The issue is that we do not do this for other types now (that is why I asked). We either show the whole type or trim what is printed to up to 8 characters.

I have pushed a change that follows what you wanted if I understand your intention correctly (please comment if this is correct and then I will update the documentation).

nalimilan: Woops, I realize I didn't even know the current rules, yet I was the one who proposed that one at https://github.com/JuliaData/DataFrames.jl/pull/1499#issuecomment-417861696! It's interesting to note that there are very few types whose names are above 8 characters, so in practice only `CategoricalValue` was in this case and we special-cased it.

Maybe showing `Cat` unless there's enough space to print `CategoricalValue` in full is OK then. Otherwise, I feel like we should find a more general rule that would apply to other types. Do we have other examples of types with long names? :-/

bkamins: Probably there are other types (e.g. if we store arrays of arrays/tuples/named tuples):
```
julia> DataFrame(a=(1,2),b=[[1]],c=(a=1,b=2))
1×3 DataFrame
│ Row │ a      │ b      │ c              │
│     │ Tuple… │ Array… │ NamedTup…      │
├─────┼────────┼────────┼────────────────┤
│ 1   │ (1, 2) │ [1]    │ (a = 1, b = 2) │
```
but 8 was chosen so that "typical types", even these above, were printed relatively OK.

So in practice I think only `CategoricalValue` is problematic. Given what you commented I think my current proposal to:
* print full type if it is possible (very rare)
* print "Categorical..." for wider columns
* print "Cat..." for columns not wide enough for "Categorical..."

should be OK. If you approve I will rewrite the docstrings to follow the current implementation (which does what I have described here).

nalimilan: Fine with me. Though maybe a smarter algorithm would be to break before the last character that is not a lowercase letter and for which the content fits. That would give the behavior you describe for `CategoricalValue`, but also improve the printing of `NamedTuple` above so that you get `NamedTuple` if there's enough space.

bkamins: I would leave as is for now - as I prefer to close this PR and move forward with other that are waiting for it to be merged (what you describe is doable and better, but would require careful testing, the reason is that there are surprisingly many "corner cases" in names of types in Julia: `Union`s, parameters, numbers, curly braces etc. - I remember struggling with them last time we changed this). We can of course improve it in the future.

So I now will update the docstrings.

nalimilan: Sure.

bkamins: Well - as usual :smile: - you have prompted me to fix things in the end. I have updated the `compacttype` implementation to do what I think is the right thing. That is:
* always try to print at least 8 text width **wide** type name (and correctly handle wide characters which we did not do earlier), also correctly always print 8 text width wide compacttype if it has to be truncated to 8 characters (this was also incorrect before in some cases)
* Handle `Any`, `Missing`, `Union` and `Categorical...` as special cases
* if type has to be compacted (i.e. it does not fit the `maxwidth`) then only try printing its name (so then we always drop the parameter part for `UnionAll` and concrete parametric types)

I did not do "truncate at first capital letter" as I came to the conclusion that it is better just to try printing as much as possible within `maxwidth` given the other rules above (with the exception of categorical which is abbreviated to `Cat` or `Categorical` as described in the rules above).

bkamins:
OK - this should be good for a final check (if you agree with the description I have given for the reimplementation of `compacttype`).

src/abstractdataframe/show.jl
```diff
@@ -55,22 +55,48 @@ end
...
+    textwidth(sT) ≤ maxwidth && return sT
+
     if T >: Missing
         T = nonmissingtype(T)
         sT = string(T)
-        suffix = "⍰"
-        # handle the case when after removing Missing union type name is short
-        length(sT) ≤ 8 && return sT * suffix
+        suffix = "?"
+        maxwidth -= 1 # we will add "?" at the end
```
nalimilan: Previously we didn't include `⍰` in the count, right? I think we'd better not include it, as it means that narrow columns allowing and not allowing for missing values will print a different abbreviation for the type, e.g. `DataFrame(x=BigFloat[1, 2], y=Union{BigFloat, Missing}[1, 2])` will use "BigFloa" and "BigFlo" (IIUC).

Also, by taking into account `?` and `…` the number of characters used for the type is now 6 instead of 8 when it's too long and allows for missing values. Previously `BigFloat` would always be printed in full, and now it isn't. Is this intentional?

src/abstractdataframe/show.jl
```diff
@@ -55,22 +55,48 @@ end
...
         suffix = ""
     end
-    T <: Union{CategoricalString, CategoricalValue} && return "Categorical…"*suffix
-    # we abbreviate consistently to total of 8 characters
-    match(Regex("^.\\w{0,$(7-length(suffix))}"), sT).match * "…"*suffix
+
+    maxwidth -= 1 # we will add "…" at the end
+
+    if T <: Union{CategoricalString, CategoricalValue}
+        return (maxwidth ≥ 11 ? "Categorical…" : "Cat…") * suffix
```
nalimilan: If the column is wider than 17 characters, can we print `CategoricalValue`/`CategoricalString` in full? For simplicity, we can use the same threshold for both, anyway the latter is going away soon.

src/abstractdataframe/show.jl
bkamins: Well - it was intentional (to normalize everything to 8), but I can change it. I will switch not to count "?" when determining width (which means sometimes it will be 9 like for `BigFloat?`).

src/abstractdataframe/show.jl
bkamins: Currently there is one nuance that `T` can also be `Union{CategoricalString, CategoricalValue}` (probably a very improbable case). I will fix it then and we can do what you propose.

bkamins:
@nalimilan - your requests are fixed. As usual (and now I remember why I had problems last time) - things were more problematic than expected. I had to distinguish the phase of initial width counting and display phase to ignore "?" (the problem was when actually the type name is the widest entry in the column and it contains missing). Things should be good now.

nalimilan:
Yeah, sorry for derailing the PR. These things are quite tricky.

I hope the choice to special-case `CategoricalValue` will turn out to be the right one. The alternative would be to use `Cat` by default, but print as many characters as possible beyond that, like for other types.

bkamins:
Thank you!

PR #2141, opened 2020-03-04
add haskey to GroupedDataFrame and GroupKey

bkamins:
When we defined `keys` for `GroupedDataFrame` and `GroupKey` we have not defined `haskey` for them. I think it is natural to define both.

When we have it defined then actually `GroupedDataFrame` can be used as a dictionary.

A design decision is if we want to define all the methods I proposed, but I think it should be OK.

### DIFF
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -219,6 +219,8 @@ end
 Base.parent(key::GroupKey) = getfield(key, :parent)
 Base.length(key::GroupKey) = length(parent(key).cols)
 Base.keys(key::GroupKey) = Tuple(groupvars(parent(key)))
+Base.haskey(key::GroupKey, idx::Symbol) = idx in groupvars(parent(key))
+Base.haskey(key::GroupKey, idx::Union{Signed,Unsigned}) = 1 <= idx <= length(key)
 Base.names(key::GroupKey) = groupvars(parent(key))
 # Private fields are never exposed since they can conflict with column names
 Base.propertynames(key::GroupKey, private::Bool=false) = keys(key)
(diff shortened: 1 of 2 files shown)


src/groupeddataframe/groupeddataframe.jl
```diff
@@ -460,6 +462,19 @@ true
 """
 Base.keys(gd::GroupedDataFrame) = GroupKeys(gd)
 
+# assume that GroupKey was generated so that we do not have to check the index
```
nalimilan: What does "was generated" mean exactly?

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -460,6 +462,19 @@ true
...
+Base.haskey(gd::GroupedDataFrame, key::GroupKey) = gd === parent(key)
+Base.haskey(gd::GroupedDataFrame, key::Tuple) = haskey(gd.keymap, key)
+
+function Base.haskey(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
+    if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(N, gd.cols))
```
nalimilan: Design question: is it more useful to return `false` than throwing an error when the number or names of columns doesn't match? Throwing could help catch errors, and I'm not sure I can imagine cases where it would be useful to return `false` instead.

src/groupeddataframe/groupeddataframe.jl
bkamins: "was generated" means that it was produced by DataFrames.jl functions (most likely `keys`).

It is possible to create a `GroupKey` object manually (though it is not exported). Then you can make it point to any index (e.g. very large one that is invalid).

I used this approach because it is also used in `to_index`. But if you prefer (and maybe I prefer it to when you ask 😄) we could also check here if `:idx` field of `GroupKey` is valid and if it is not throw an error (I think we should not return `false` then).

src/groupeddataframe/groupeddataframe.jl
bkamins: Good point. I will change it.

src/groupeddataframe/groupeddataframe.jl
bkamins: I will just  change it also.

nalimilan: I guess it wouldn't be a problem for performance, so maybe better also check the index indeed.

bkamins:
OK - I have made `haskey` error in clearly error cases.

A small note here is that `haskey` in Base does not have this property - it always produces `true` or `false` even in "nonsense" cases for which we throw now.

nalimilan:
Yeah, we'll see whether the inconsistency is justified or not. As usual, we can turn errors into something else later, but not the reverse.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -460,6 +462,36 @@ true
...
+    if length(key) != length(gd.cols)
+        return throw(ArgumentError("The length of key does not match the " *
+                                   "number of grouping columns"))
+    end
+    return haskey(gd.keymap, key)
+end
+
+function Base.haskey(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
+    if any(n != _names(gd)[c] for (n, c) in zip(N, gd.cols))
+        return throw(ArgumentError("The column names of keys do not match " *
```
bkamins:
```suggestion
        return throw(ArgumentError("The column names of key do not match " *
```

PR #2145, opened 2020-03-08
[BREAKING] fix eltype in stack with view=true

bkamins:
Fixes a wrong definition of `eltype` when `stack` with `view=true` is used. We switch from dynamic to static calculation of eltype (which is needed as Base uses static eltype information for vectors to manipulate them).

Do not merge before #2140 is merged as docs (the manual) need to be updated here which is not done yet.

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -335,7 +335,7 @@ end
 StackedVector(d::AbstractVector) =
     StackedVector{promote_type(map(eltype, d)...)}(d)
 
-function Base.getindex(v::StackedVector, i::Int)
+function Base.getindex(v::StackedVector{T}, i::Int)::T where T
     lengths = [length(x)::Int for x in v.components]
     cumlengths = [0; cumsum(lengths)]
     j = searchsortedlast(cumlengths .+ 1, i)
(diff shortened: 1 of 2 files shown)


bkamins:
It is mildly breaking as it changes `eltype` of `StackedVector`.

src/abstractdataframe/reshape.jl
```diff
@@ -328,11 +328,14 @@ StackedVector(d::AbstractVector...)
...
-struct StackedVector <: AbstractVector{Any}
+struct StackedVector{T} <: AbstractVector{T}
     components::Vector{Any}
 end
 
-function Base.getindex(v::StackedVector,i::Int)
+StackedVector(d::AbstractVector) =
+    StackedVector{promote_type(map(eltype, d)...)}(d)
+
+function Base.getindex(v::StackedVector, i::Int)
```
nalimilan: Doesn't this need a `convert(T, ...)` for the returned value?

bkamins: Good point. This is consistent with `stack` without view and how `unstack` later works. I will add this change.
 
(the alternative would be to use `typejoin` not `promote_type` in line 336. Then we do not need to convert, but it would not be consistent with `stack` that materializes.)

bkamins:
Thank you!

PR #2147, opened 2020-03-09
[BREAKING] make id_vars go first in stack

### LINKED ISSUE #645: Output format of reshape functions
@tshort I got familiar with the reshape functions in #644, and had a couple questions about the output format:
1) Placing the id vars to the right of the measure vars -- seemed the opposite of what I'd used elsewhere -- what's the rationale?
2) Trivial: lowercase `variable` and `value` -- are they just from before the uppercasing idiom took hold, or is it a reducing clashes / purposely stand out thing?

```
julia> d = dataset("reshape2", "smiths")
julia> melt(d, [:Subject, :Time], [:Age, :Weight, :Height])
6x4 DataFrame
|-------|----------|-------|--------------|------|
| Row # | variable | value | Subject      | Time |
| 1     | Age      | 33.0  | "John Smith" | 1    |
| 2     | Age      | NA    | "Mary Smith" | 1    |
| 3     | Weight   | 90.0  | "John Smith" | 1    |
| 4     | Weight   | NA    | "Mary Smith" | 1    |
| 5     | Height   | 1.87  | "John Smith" | 1    |
| 6     | Height   | 1.54  | "Mary Smith" | 1    |
```

bkamins:
Fixes #645.

Now we put `id_vars` first in the data frame returned from `stack`


src/abstractdataframe/reshape.jl
```diff
@@ -55,14 +57,19 @@ function stack(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
                id_vars::AbstractVector{<:Integer}; variable_name::Symbol=:variable,
                value_name::Symbol=:value, view::Bool=false,
                variable_eltype::Type=CategoricalString)
+    if stack_didnotwarn
```
nalimilan: I'm not sure it's a good idea to print this. There's no way to turn this off so it will annoying everybody. And if you relied on the columns order, you'll find out the problem soon enough.

The only way to deprecate this cleanly would be to introduce a keyword argument, but that would be quite annoying.

bkamins: The point is that I did not do a normal deprecation exactly because I do not want the message to be annoying. It will be printed only once per Julia session (as opposed a normal depwarn that would be printed every time).

The deprecation would only be in 0.21, in 1.0 release it will be gone.

@oxinabox, @pdeffebach  - can you please comment on it from a "user's voice" perspective. Thank you!

oxinabox: > as opposed a normal depwarn that would be printed every time).

is that still not fixed?

Rather than using a global  you can do `@warn msg maxlog=1`
to prevent warning showing up more than once

---

I never use `stack`, only `unstack`, so i have no stake in this game.

Warnings are, in general are incredibly annoying.
I would rather a well-documented and breaking change than most warnings.

bkamins: > `@warn msg maxlog=1`

Thank you - one learns something all the time (I knew there should be a solution for this :smile:). If we end up printing a deprecation warning I will change the implementation to use your tip.

pdeffebach:
I want to add that I think this is a good idea and I wonder if, despite allowing indexing by numbers, the order of columns shouldn't be considered a part of the API that we can "break". i.e., tell users not to rely on the order to columns in their code.

bkamins:
> tell users not to rely on the order to columns in their code.

This probably should be a soft recommendation (as it is in other tabular data systems where you are typically advised to index column by name).

But in general sometimes indexing by a position is useful (and that is why we provide it) so I would be hesitant just to say that "the order of columns is undefined".

Anyway - the decision for this PR is to either
1. print the warning once per session in 0.21 release, and drop the warning in 1.0 release.

or

2. do not print the warning at all and just note in the release notes

(I would prefer option 1. as printing a warning only once is not that bad I think)

nalimilan:
I vote for 2. IIUC @oxinabox has the same preference, but he said he has no stake in this game, so...  :-)

bkamins:
Well it seems only I want this warning to be printed. I have dropped it in the commit I have just pushed.

bkamins:
Thank you!

Issue #2150, opened 2020-03-11
add missing columns when push! ing?


kleinschmidt:
I'm trying to build up a data frame one row at a time, except that not every row has all the columns.  I'd like to be able to `push!` the rows on one at a time, filling in the missing columns with `missing`.  using `cols=:subset` gets me halfway there:

```julia
julia> df = DataFrame()
0×0 DataFrame

julia> push!(df, (a=1, b=2))
1×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │

julia> allowmissing!(df)
1×2 DataFrame
│ Row │ a      │ b      │
│     │ Int64⍰ │ Int64⍰ │
├─────┼────────┼────────┤
│ 1   │ 1      │ 2      │
julia> push!(df, (a=1, c=3), cols=:subset)
2×2 DataFrame
│ Row │ a      │ b       │
│     │ Int64⍰ │ Int64⍰  │
├─────┼────────┼─────────┤
│ 1   │ 1      │ 2       │
│ 2   │ 1      │ missing │
```

But notice that no column for `c` is added.  Having to call `allowmissing!` is also potentially annoying since I'll have to call that for every row or have to do the bookkeeping of keeping track of which columns have already missings in them in the loop over rows.  But I think that's more related to #1716 .

I'd like either an additional argument like `fill`, or another value for `cols` which opts into adding columns as they are encountered.

bkamins:
I think it is a duplicate (actually a subset) of https://github.com/JuliaData/DataFrames.jl/issues/2032. Could you please have a look there and if yes this can be closed.

Just as a comment - things are not easy here as you have to decide about widening rule. If you have a column of `Int`s and want to push a float to it - should the column be converted to floats then?

In particular - maybe you want something simpler, which could be implemented before https://github.com/JuliaData/DataFrames.jl/issues/2032. Something like `cols=:union` where, as you proposed missing columns would be added filled with missing in existing rows and the new entry given type it has.

Now having to call `allowmissing!` is another issue, and potentially another kwarg could be added, like `widen`, again it is tricky, do we want to allow promotion:
* only to `Missing` union
* using `promote_type` (could lead to conversion of existing data)
* using `typeunion` (will produce abstract types)
* using `Union` of encountered types (this is not what normally is done)

[...]

PR #2152, opened 2020-03-13
allow :union as cols kwarg in push! and append!

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2150

@kleinschmidt and @nalimilan - I have implemented what I think is a useful minimal functionality. As usual things were more complex than I have initially expected.

(also comments helping to refine a docstring are welcome, as I usually do not get it right)


src/dataframe/dataframe.jl
```diff
@@ -1309,25 +1309,74 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; cols::S
...
+            else
+                val = missing
+            end
+            S = typeof(val)
+            T = eltype(col)
+            V = promote_type(S, T)
+            if S <: T || V <: T
+                push!(col, val)
+            else
+                newcol = Tables.allocatecolumn(V, targetrows)
```
kleinschmidt: is this widening happening only for `col=:union`?  while that's useful it seems odd to have the widening behavior be different from `:subset`.  is this meant to only apply in the case where `S === missing`?

Ah I see now that the docstring does mention that this behavior is different.  I wonder whether it makes more sense to *only* do the widening in the case of missing?

bkamins:
> Only issue is that the widening seems aggressive, but you know the tradeoffs here better than I do. I'd be happy to just have it widen in the case that missings are introduced.

Such considerations were exactly the reason why last time we have decided not to add this functionality, because there were too many options.

My thinking was  that if we want something simple then it should be flexible, i.e. if the user passes `:union` then it is safe to assume that the operation will not fail due to promotion problems (it might fail due to aliasing, but this is a corner case that probably normal users do not have to consider). Therefore you can safely write something like:
```
df = DataFrame()
for row in some_source_of_rows
    push!(df, row, cols=:union)
end
```
and expect it will not fail.
So essentially this is like reading a CSV or JSON file row by row and doing automatic detection of column type on the go.

Of course we can be more defensive - as you propose. The question is in what cases you would benefit from `push!` throwing an error when type promotion would be required (a typical problem is that you have a column of `Int` (as e.g. source data were rouned to integers) and later there is some `Float64` in your data and without automatic promotion `push!`-ing would fail). Another example is that you initially have a `missing` in some column that you `push!`, so you create `Vector{Missing}` and later - if you have actual data in the column without allowing promotion you will have an error.

A midway solution would be to allow promotion only if `promote_type` produces a concrete type or `Union{concrete_type, Missing}` or `Union{previous_eltype_of_column,Missing}`, and fail if it produces an abstract type.

So in short - I am open to change what I have proposed, and all options have pros and cons unfortunately.

> Edit: also needs tests but you probably knew that already ;)

What do you mean by this (I thought I have provided an adequate test coverage). The tests I have written pass, the :x: is only due to unrelated CI failures.

kleinschmidt:
I don't see any changes to test files here, just the source...

pdeffebach:
I would have expected something like 

```
push!(df, x; cols = :union, promote_type = true)
```

Where `cols = :union` automatically promotes to `missing` but errors otherwise, except when `promote_type` is used.

bkamins:
@kleinschmidt - thank you (I have forgotten to stage the test file - pushed now :smile:)

bkamins:
@pdeffebach - intuitively I would prefer to keep a single kwarg not to complicate the design and pick the behavior of `:union` that is most useful.

Given this consideration I preferred to make `:union` do promotion as this seemed to me "more useful".

In general I think you shoulud use `:union` if you are unsure what you will get. If you know what you expect to get I think one should construct a `DataFrame` with all needed columns empty and of correct types and then use `cols=:subset` in `push!`.

pdeffebach:
> In general I think you shoulud use :union if you are unsure what you will get.

Sounds good! `:union` is an appropriate term for this, with its double meaning and all.

bkamins:
So this was my thinking. But just to make sure we are on the right track. Let us try to think of a case when this kind of promotion in `:union` would be a bad thing. I.e. in what scenario one would prefer to get an error rather than auto-promotion. I could not come up with such a convincing scenario, but maybe there is some that would change our minds. Thank you for the time you spend on it!

pdeffebach:
I agree that no one who is too "lazy" to make sure all their named-tuples have the same number of fields is going to care that much about unexpected autopromotion. 

The only scenario I can think of is someone who wants auto-promotion of types but doesn't realize their named tuples have different elements. Then they unexpectedly create new columns. 

*But* this seems very unlikely. 

1. The vast majority of use-cases is someone applying a function that returns a named tuple and then pushes to the data frame in a loop. Why would the function return an inconsistent set of fields?

2. Perhaps they are parsing a string. Then they should use CSV.jl or JSONTables.jl, if they aren't doing that they are probably an advanced user who can write their own function which normalizes fields. 

The only other concern would be discoverability. This is solved by good documentation and perhaps a few discourse answers.

kleinschmidt:
I think the scenario I had in mind was more that someone would use `cols=:union` once and then expect that auto-promotion is the default for the other options as well.

kleinschmidt:
But I agree that's unlikely given that CSV.jl/JSON3.jl handle the promotion for you....

bkamins:
> I think the scenario I had in mind was more that someone would use `cols=:union` once

Yeah, I was afraid of it too. I even considered adding `push!!` instead of `cols=:union` to avoid this duality. But I came to the conclusion that:

* if someone uses `push!` one has to read the documentation in order to use `cols=:union` (as otherwise how can you learn about it)
* if you do not do `cols=:union` in `push!` and get an error - well - then you read the documentation to learn that `cols=:union` is a special case

Because of this in the initial message in this PR I have made the comment

> (also comments helping to refine a docstring are welcome, as I usually do not get it right)

as proper documentation is important here.

The key concern was that `push!` in Base clearly does not auto-promote and will never do (because it is not able to mutate the type of the container as opposed to the case of a `DataFrame`). And again - exactly for these reasons we did not implement `:union` for `push!` in the past.

So in summary we have the situation:
1. have a small inconsistency in `push!` as proposed in current PR which I simply think is a convenient functionality for its use case (reading data from an unknown source) at the cost of a small inconsistency
2. not have this functionality at all (and here the point of @kleinschmidt initial comment was that he wanted it :smile:), or have a separate function for it (like `push!!` which has a problem that I feared that would be less discoverable by users than an additional option to `push!`), or further complicate the API of `push!` with another kwarg (I wanted to avoid this as `push! is already complex as witnessed by its docstring)

I understand that we converge to accepting what is currently implemented in the PR - right?

Thank you for discussing this.

nalimilan:
Sorry for not commenting earlier. I think I'd rather have a separate `promote` keyword argument. That makes things simpler and more powerful, as various combinations with `cols` could make sense. Apart from the ones with `cols=:union`, I think it can be not too uncommon with the default `cols=:setequal` too. As @bkamins mentioned, if you start with one row (e.g. when running computations in a loop and aggregating results) and later get missing values (or the reverse), you need promotion, yet it would be weird (and hard to find) to use `cols=:union` since you don't want to add new columns at all. The same thing can happen if you want to append a new dataset loaded from CSV (for `append!`).

If we want a flexible mode, we could have another argument for convenience which will set `cols=:union, promote=true`.

bkamins:
What you propose is essentially the initial design that was planned.

It is more complex, but we can go for it. I would simply set `promote` column by default to different values depending on `cols` column.

I think it is natural that for `:union` and `:subset` it is set to `true` (to allow population with `missing` if needed as a common case) and for all other it can be set to `false`.

If we go for it in a separate PR I can update `append!` to be synced with `push!` as it would go along the same logic as you note.

@nalimilan - please confirm that you are OK with this plan and then I can go forward with implementing the functionality.

bkamins:
@nalimilan - I have implemented what you proposed. In the end it seems it is not that problematic as I feared. The only issue is that if we have column aliases then with `promote=true` we will not be able to revert the `DataFrame` to its original state, but I think it is acceptable.

bkamins:
I have added the same arguments to `append!` to support https://github.com/JuliaData/DataFrames.jl/issues/2173. It is left to update the tests now (I will do it soon).

bkamins:
@nalimilan - all tests added.

src/dataframe/dataframe.jl
```diff
@@ -1108,8 +1112,9 @@ julia> df1
 │ 6   │ 6     │ 6     │
 ```
 """
-function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:setequal)
-    if !(cols in (:orderequal, :setequal))
+function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:setequal,
+                      promote::Bool=(cols in [:union, :subset]))
+    if !(cols in (:orderequal, :setequal, :intersect, :subset, :union))
         throw(ArgumentError("`cols` keyword argument must be any of :setequal, :orderequal"))
```
nalimilan, line 1118:
```suggestion
        throw(ArgumentError("`cols` keyword argument must be " *
                            ":orderequal, :setequal, :intersect, :subset or :union)"))
```

src/dataframe/dataframe.jl
```diff
@@ -1129,21 +1134,29 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
             throw(ArgumentError("Columns number " *
                                 join(mismatches, ", ", " and ") *
                                 " do not have the same names in both passed data frames" *
-                                "and `cols=:orderequal`"))
+                                "and `cols==:orderequal`"))
```
nalimilan, line 1137: Add spaces around `==` ?

src/dataframe/dataframe.jl
```diff
@@ -1129,21 +1134,29 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
+                                " were found in only one of the passed data frames " *
+                                "and passed `cols==:setequal`"))
+        end
+    elseif cols == :intersect
+        wrongnames = setdiff(_names(df1), _names(df2))
+        if !isempty(wrongnames)
+            throw(ArgumentError("Column names :" *
+                                join(wrongnames, ", ", :" and :") *
+                                " were found in only in destination data frame " *
+                                "and passed `cols==:intersect`"))
```
nalimilan, line 1159:
```suggestion
                                join(wrongnames, ", ", :" and :") *
                                " were found in only in destination data frame " *
                                " and `cols==:intersect`"))
```

src/dataframe/dataframe.jl
```diff
@@ -1129,21 +1134,29 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
         end
     elseif cols == :setequal
         wrongnames = symdiff(_names(df1), _names(df2))
         if !isempty(wrongnames)
             throw(ArgumentError("Column names :" *
                                 join(wrongnames, ", ", :" and :") *
-                                "were found in only one of the passed data frames " *
-                                "and passed `cols=:setequal`"))
+                                " were found in only one of the passed data frames " *
+                                "and passed `cols==:setequal`"))
```
nalimilan, line 1151:
```suggestion
                                "and `cols==:setequal`"))
```

src/dataframe/dataframe.jl
```diff
@@ -1129,21 +1134,29 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
             throw(ArgumentError("Columns number " *
                                 join(mismatches, ", ", " and ") *
                                 " do not have the same names in both passed data frames" *
-                                "and `cols=:orderequal`"))
+                                "and `cols==:orderequal`"))
         else
             mismatchmsg = " Column names :" *
             throw(ArgumentError("Column names :" *
                                 join(wrongnames, ", ", :" and :") *
```
nalimilan, line 1141:
```suggestion
                                join(wrongnames, ", :", " and :") *
```

src/dataframe/dataframe.jl
```diff
@@ -1129,21 +1134,29 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
-                                "were found in only one of the passed data frames " *
-                                "and `cols=:orderequal`"))
+                                " were found in only one of the passed data frames " *
+                                "and `cols==:orderequal`"))
         end
     elseif cols == :setequal
         wrongnames = symdiff(_names(df1), _names(df2))
         if !isempty(wrongnames)
             throw(ArgumentError("Column names :" *
                                 join(wrongnames, ", ", :" and :") *
```
nalimilan, line 1149:
```suggestion
                                join(wrongnames, ", :", " and :") *
```

src/dataframe/dataframe.jl
```diff
@@ -1129,21 +1134,29 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
-                                "were found in only one of the passed data frames " *
-                                "and passed `cols=:setequal`"))
+                                " were found in only one of the passed data frames " *
+                                "and passed `cols==:setequal`"))
+        end
+    elseif cols == :intersect
+        wrongnames = setdiff(_names(df1), _names(df2))
+        if !isempty(wrongnames)
+            throw(ArgumentError("Column names :" *
+                                join(wrongnames, ", ", :" and :") *
```
nalimilan, line 1157:
```suggestion
                                join(wrongnames, ", :", " and :") *
```

src/dataframe/dataframe.jl
```diff
@@ -1153,13 +1166,49 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
+            if n in _names(df2)
+                df1_c = df1[!, j]
+                df2_c = df2[!, n]
+                S = eltype(df2_c)
+                T = eltype(df1_c)
+                V = promote_type(S, T)
+                if S <: T || V <: T || !promote
+                    append!(df1_c, df2_c)
+                else
+                    newcol = Tables.allocatecolumn(V, targetrows)
```
nalimilan, line 1178: This might preserve more properties of the input column:
```suggestion
                    newcol = similar(df1_c, V, targetrows)
```

src/dataframe/dataframe.jl
```diff
@@ -1153,13 +1166,49 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
     try
         for (j, n) in enumerate(_names(df1))
             current_col += 1
-            append!(df1[!, j], df2[!, n])
+            if n in _names(df2)
```
nalimilan, line 1169:
```suggestion
            if hasproperty(df2, n)
```

src/dataframe/dataframe.jl
```diff
@@ -1153,13 +1166,49 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
+                    copyto!(newcol, 1, df1_c, 1, nrows)
+                    copyto!(newcol, nrows+1, df2_c, 1, targetrows - nrows)
+                    _columns(df1)[j] = newcol
+                end
+            else
+                # we can safely do it - if we got here this means that this is allowed
+                if Missing <: eltype(df1[!, j]) || !promote
+                    append!(df1[!, j], df2[!, n])
+                else
+                    newcol = Tables.allocatecolumn(Union{Missing,eltype(df1[!, j])} , targetrows)
```
nalimilan, line 1188:
```suggestion
                    newcol = similar(df1[!, j], Union{Missing, T}, targetrows)
```

src/dataframe/dataframe.jl
```diff
@@ -1153,13 +1166,49 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
+                    copyto!(newcol, 1, df1[!, j], 1, nrows)
+                    for i in nrows+1:targetrows
+                        @inbounds newcol[i] = missing
+                    end
```
nalimilan, line 1192:
```suggestion
                    newcol[nrows+1:targetrows] .= missing
```

src/dataframe/dataframe.jl
```diff
@@ -1153,13 +1166,49 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
     try
         for (j, n) in enumerate(_names(df1))
             current_col += 1
-            append!(df1[!, j], df2[!, n])
+            if n in _names(df2)
+                df1_c = df1[!, j]
+                df2_c = df2[!, n]
+                S = eltype(df2_c)
```
nalimilan, line 1172: Maybe call this `U` and define it after `T` to use the same order as arguments?

src/dataframe/dataframe.jl
```diff
@@ -1153,13 +1166,49 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
         for col in _columns(df1)
             current_col += 1
             @assert length(col) == targetrows
         end
+        if cols == :union
+            for n in setdiff(_names(df2), _names(df1))
+                newcol = Tables.allocatecolumn(Union{Missing,eltype(df2[!, n])} , targetrows)
+                for i in 1:nrows
+                    @inbounds newcol[i] = missing
+                end
```
nalimilan, line 1207:
```suggestion
                newcol = similar(df2[!, n], Union{Missing, S}, targetrows)
                @inbounds newcol[1:nrows] .= missing
```

src/dataframe/dataframe.jl
```diff
@@ -1153,13 +1166,49 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
+                copyto!(newcol, nrows+1, df2[!, n], 1, targetrows - nrows)
+                df1[!, n] = newcol
```
nalimilan, line 1209: Why not `_columns(df1)[j] = newcol` here? Or rather, why not `df1[!, n] = newcol` above since it's more standard?

src/dataframe/dataframe.jl
```diff
@@ -1176,30 +1225,79 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
 Base.convert(::Type{DataFrame}, d::AbstractDict) = DataFrame(d, copycols=false)
 
 function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; cols::Symbol=:setequal,
-                    columns::Union{Nothing,Symbol}=nothing)
+                    columns::Union{Nothing,Symbol}=nothing, promote::Bool=(cols in [:union, :subset]))
```
nalimilan, line 1228:
```suggestion
                    columns::Union{Nothing,Symbol}=nothing,
                    promote::Bool=(cols in [:union, :subset]))
```

src/dataframe/dataframe.jl
```diff
@@ -1176,30 +1225,79 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
...
     if ncols == 0 && row isa NamedTuple
         for (n, v) in pairs(row)
             setproperty!(df, n, fill!(Tables.allocatecolumn(typeof(v), 1), v))
         end
         return df
     end
+
+    if cols == :union
+        if !all(x -> x isa Symbol, keys(row))
+            throw(ArgumentError("when `cols=:union all keys of row must be Symbol"))
```
nalimilan, line 1250: Couldn't we just call `Symbol` on keys and see what happens?

Also, we should stick to a single convention regarding `cols=:something` vs. `cols == :something`.

src/dataframe/dataframe.jl
```diff
@@ -1176,30 +1225,79 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
...
+            else
+                val = missing
+            end
+            S = typeof(val)
+            T = eltype(col)
+            V = promote_type(S, T)
+            if S <: T || V <: T || !promote
+                push!(col, val)
+            else
+                newcol = Tables.allocatecolumn(V, targetrows)
```
nalimilan, line 1265:
```suggestion
                newcol = similar(col, V, targetrows)
```

src/dataframe/dataframe.jl
nalimilan: This comment no longer applies I think.

src/dataframe/dataframe.jl
```diff
@@ -1176,30 +1225,79 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
...
+            col = _columns(df)[i]
+            if haskey(row, colname)
+                val = row[colname]
+            else
+                val = missing
+            end
+            S = typeof(val)
+            T = eltype(col)
+            V = promote_type(S, T)
+            if S <: T || V <: T || !promote
```
nalimilan, line 1262: I wonder whether it wouldn't be better for performance to avoid calling `promote_type` if `S <: T || !promote` (for type-unstable cases). Same for `append!`, though it should be less of a concern there since you add multiple rows.

Also in some cases maybe having something like `push_helper` could make sense?

src/dataframe/dataframe.jl
```diff
@@ -1176,30 +1225,79 @@ Base.convert(::Type{DataFrame}, A::AbstractMatrix) = DataFrame(A)
...
+                newcol[end] = val
+                _columns(df)[i] = newcol
+            end
+        end
+        for (colname, col) in zip(_names(df), _columns(df))
+            if length(col) != targetrows
+                for col2 in _columns(df)
+                    resize!(col2, nrows)
+                end
+                throw(AssertionError("Error adding value to column :$colname"))
```
nalimilan, line 1276: When can this happen?

src/dataframe/dataframe.jl
```diff
@@ -1223,7 +1321,18 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; cols::S
             else
                 val = row[nm]
             end
-            push!(col, val)
+            S = typeof(val)
+            T = eltype(col)
+            V = promote_type(S, T)
+            if S <: T || V <: T || !promote
+                # if S <: T || V <: T this should never throw an exception
```
nalimilan, line 1328: This comment applies in other similar places, right?

src/dataframe/dataframe.jl
```diff
@@ -1223,7 +1321,18 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple}; cols::S
...
+                push!(col, val)
+            else
+                newcol = Tables.allocatecolumn(V, targetrows)
```
nalimilan, line 1331:
```suggestion
                newcol = similar(col, V, targetrows)
```

src/dataframe/dataframe.jl
```diff
@@ -1269,6 +1378,12 @@ depends on the `cols` argument value in the following way:
   are used to populate a new row in `df`.
 * If `cols=:subset` then `push!` behaves like for `:intersect` but if some column
   is missing in `row` then a `missing` value is pushed to `df`.
+* If `cols=:union` then `push!` adds columns missing in `df` that are present in `row`,
+  for columns present in `df` but missing in `row` a `missing` value is pushed.
```
nalimilan, line 1382:
```suggestion
* If `cols=:union` then columns missing in `df` that are present in `row` are added to `df`
  (using `missing` for existing rows) and a `missing` value is pushed to columns
  missing in `row` that are present in `df`.
```

src/dataframe/dataframe.jl
```diff
@@ -1269,6 +1378,12 @@ depends on the `cols` argument value in the following way:
...
+
+If `promote=true` and eltype of a column present in `df` does not allow the type
+of a pushed argument then a new column with a promoted eltype allowing it is freshly
+allocated and stored in `df`. If `promote=false` an error is thrown.
```
nalimilan, line 1386:
```suggestion
If `promote=true` and element type of a column present in `df` does not allow the type
of a pushed argument then a new column with a promoted element type allowing it is freshly
allocated and stored in `df`. If `promote=false` an error is thrown.
```

src/dataframe/dataframe.jl
```diff
@@ -1297,30 +1405,7 @@ julia> push!(df, (true, false))
 │ 3   │ 3     │ 3     │
 │ 4   │ 1     │ 0     │
 
-julia> push!(df, df[1, :])
```
nalimilan, line 1300: Why remove this?

src/dataframerow/dataframerow.jl
```diff
@@ -337,62 +337,124 @@ end
...
-        if parent(dfr) === df && index(dfr) isa Index
-            # in this case we are sure that all we do is safe
-            r = row(dfr)
-            for col in _columns(df)
-                # use a barrier function to improve performance
-                pushhelper!(col, r)
+    if cols == :union
+        for (i, colname) in enumerate(_names(df))
+            col = _columns(df)[i]
+            if haskey(dfr, colname)
```
nalimilan, line 383:
```suggestion
            if hasproperty(dfr, colname)
```

src/dataframerow/dataframerow.jl
```diff
@@ -337,62 +337,124 @@ end
...
+        end
+        for (colname, col) in zip(_names(df), _columns(df))
+            if length(col) != targetrows
+                for col2 in _columns(df)
+                    resize!(col2, nrows)
                 end
-                msg = "Number of columns of `DataFrameRow` does not match that of " *
-                      "target data frame (got $(length(dfr)) and $ncols)."
-                ncols == length(dfr) || throw(ArgumentError(msg))
+                throw(AssertionError("Error adding value to column :$colname"))
```
nalimilan, line 405: This is already checked in the `catch` block, isn't it?

src/dataframerow/dataframerow.jl
```diff
@@ -337,62 +337,124 @@ end
...
+                val = dfr[nm]
+            end
+            S = typeof(val)
+            T = eltype(col)
+            V = promote_type(S, T)
+            if S <: T || V <: T || !promote
+                # if S <: T || V <: T this should never throw an exception
                 push!(col, val)
+            else
+                newcol = Tables.allocatecolumn(V, targetrows)
```
nalimilan, line 454:
```suggestion
                newcol = similar(col, V, targetrows)
```

src/dataframe/dataframe.jl
bkamins: ok

src/dataframe/dataframe.jl
bkamins: Yeah - this is always tricky.

src/dataframe/dataframe.jl
bkamins: This is tricky. Here I use `df1[!, n] = newcol` as it is more standard :smile:, but above `df1[!, n] = newcol` will error when we assign the first column, because it will not have the length equal to number of rows before the assignment (i.e. we are making a data frame taller earlier, so we have to do a workaround).

src/dataframe/dataframe.jl
bkamins: yes - should I add it?

src/dataframerow/dataframerow.jl
bkamins: We do not reach `catch` block if `cols == :union`. This error is abnormal, and a most likely cause is that `DataFrame` contained column aliases.

src/dataframe/dataframe.jl
bkamins: I actually had it in most places. Added in two more.

src/dataframe/dataframe.jl
bkamins: The most likely cause is column aliases (or error in Base Julia).

src/dataframe/dataframe.jl
bkamins: OK - fixed this by reordering conditions.

I do not see where `push_helper` could help as we are processing each column exactly once.

src/dataframe/dataframe.jl
bkamins: OK. This is what we do in the constructor.

src/dataframe/dataframe.jl
bkamins: I reordered the lines. I would leave `S` and `T` as this is consistent with the rest of the code.

src/dataframe/dataframe.jl
bkamins: OK - I have reverted it.

src/dataframe/dataframe.jl
bkamins: We have to call `promote_type` twice now but I think it is OK, as we call it for the second time only if we allocate a column. Also maybe compiler will be smart enough to optimize this out.

bkamins:
Thank you for a huge effort to review this. I hope I have fixed everything.

bkamins:
CI passes here.

test/dataframe.jl
```diff
@@ -718,10 +718,142 @@ end
...
+            df1.x = 1:3
+            df1.y = df1.x
+            df2 = DataFrame(x=1:3, y=4:6)
+            with_logger(sl) do
+                @test_throws AssertionError append!(df1, df2, cols=cols, promote=promote)
+            end
+            @test df1 == DataFrame(x=1:3, y=1:3)
+            df2 = DataFrame(y=4:6, x=1:3)
+            with_logger(sl) do
+                @test_throws (cols==:orderequal ? ArgumentError :
```
nalimilan, line 795:
```suggestion
                @test_throws (cols == :orderequal ? ArgumentError :
```

src/dataframe/dataframe.jl
nalimilan: Maybe add a comment when using `_columns` then?

src/dataframe/dataframe.jl
nalimilan: OK.

What about the `cols=` vs. `cols ==` convention?

src/dataframe/dataframe.jl
bkamins: added

src/dataframe/dataframe.jl
bkamins: this is fixed. GitHub just shows an outdated version of the code here.

src/dataframe/dataframe.jl
```diff
@@ -1164,6 +1164,9 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
     nrows, ncols = size(df1)
     targetrows = nrows + nrow(df2)
     current_col = 0
+    # in the code below we use a direct access to _columns because
```
nalimilan, line 1167: There's the same pattern in `push!`, right?

bkamins: right. added

bkamins:
Thank you!

PR #2154, opened 2020-03-18
Fix typo

prosoitos:
I think that DataFrame and not vector was meant in this sentence.

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -419,7 +419,7 @@ julia> df[!, :A] == df[:, :A]
 true
 ```
 
-In the first cases, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`, since a vector can contain one or more column names. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted. Note that in the first case a vector is required to be passed (not just any iterable), so e.g. `df[:, (:x1, :x2)]` is not allowed, but `df[:, [:x1, :x2]]` is valid.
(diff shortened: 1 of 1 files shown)


bkamins:
The previous wording was correct. What is meant that vector `[:A]` in this case contains one column but in general it could contain more.

However, if you find the current wording unclear please go ahead and correct the sentence (I am not a native so ...).

docs/src/man/getting_started.md
```diff
@@ -419,7 +419,7 @@ julia> df[!, :A] == df[:, :A]
 true
 ```
 
-In the first cases, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`, since a vector can contain one or more column names. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted. Note that in the first case a vector is required to be passed (not just any iterable), so e.g. `df[:, (:x1, :x2)]` is not allowed, but `df[:, [:x1, :x2]]` is valid.
+In the first cases, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`, since a DataFrame can contain one or more column names. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted. Note that in the first case a vector is required to be passed (not just any iterable), so e.g. `df[:, (:x1, :x2)]` is not allowed, but `df[:, [:x1, :x2]]` is valid.
```
bkamins, line 422:
```suggestion
In the first case, `[:A]` is a vector, indicating that the resulting object should be a `DataFrame`. On the other hand, `:A` is a single symbol, indicating that a single column vector should be extracted. Note that in the first case a vector is required to be passed (not just any iterable), so e.g. `df[:, (:x1, :x2)]` is not allowed, but `df[:, [:x1, :x2]]` is valid.
```

bkamins:
Thank you for your suggestion. I have removed the confusing part in the end as it was not needed (and only caused a distraction).

prosoitos:
Great. Thanks!

Issue #2110, opened 2020-02-12
Add transform function


bkamins:
The PR adding `transform` function should follow after merging #2080.

PR #2155, opened 2020-03-19
add transform and transform! and select cleanup

bkamins:
Adds transform and transform! functions.

Additionally small bug fixes in `select` (`Any()` corner case when `copycols=false`) and in making a view of a `SubDataFrame` and some small rearrangements of documentation.

### DIFF
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -136,7 +136,7 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
     if col_idx isa Int
         res = fun(df[!, col_idx])
     else
-        res = fun(i -> ntuple(cdf[col_idx[i]], length(col_idx))...)
+        res = fun(ntuple(i -> cdf[col_idx[i]], length(col_idx))...)
     end
     if res isa Union{AbstractDataFrame, NamedTuple, DataFrameRow, AbstractMatrix}
         throw(ArgumentError("return value from function $fun " *
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -121,7 +121,9 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
                            df::AbstractDataFrame, newdf::DataFrame,
                            transformed_cols::Dict{Symbol, Any}, copycols::Bool)
     col_idx, (fun, newname) = nc
-    # it is allowed to request column tranformation only once
+    # It is allowed to request a tranformation operation into a newname column
+    # only once. This is ensured by the logic related to transformed_cols dictionaly
+    # in _select, therefore in select_transform! such a duplicate should not happen
     @assert !hasproperty(newdf, newname)
     cdf = eachcol(df)
     if col_idx isa Int
@@ -150,12 +152,15 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
 end
 
 """
-    select!(df::DataFrame, inds...)
+    select!(df::DataFrame, args...)
 
-Mutate `df` in place to retain only columns specified by `inds...` and return it.
+Mutate `df` in place to retain only columns specified by `args...` and return it.
 
-Arguments passed as `inds...` can be any index that is allowed for column indexing.
-In particular, regular expressions, `All`, `Between`, and `Not` selectors are supported.
+Arguments passed as `args...` can be:
+
+* Any index that is allowed for column indexing. In particular, regular expressions,
+  `All`, `Between`, and `Not` selectors are supported.
+* Column transformation operations using the `Pair` notation that is described below.
 
 Columns can be renamed using the `old_column => new_column_name` syntax,
 and transformed using the `old_column => fun => new_column_name` syntax.
@@ -239,36 +244,36 @@ julia> select!(df, :, [:c, :b] => (c,b) -> c .+ b .- sum(b)/length(b))
 
 julia> df = DataFrame(a=1:3, b=4:6);
 
-julia> select!(df, names(df) .=> sum .=> [:A, :B]);
+julia> select!(df, names(df) .=> sum);
 
 julia> df
 1×2 DataFrame
-│ Row │ A     │ B     │
+│ Row │ a_sum │ b_sum │
 │     │ Int64 │ Int64 │
 ├─────┼───────┼───────┤
 │ 1   │ 6     │ 15    │
 ```
 
 """
-function select!(df::DataFrame, inds::AbstractVector{Int})
-    if isempty(inds)
+function select!(df::DataFrame, args::AbstractVector{Int})
+    if isempty(args)
         empty!(_columns(df))
         empty!(index(df))
         return df
     end
-    indmin, indmax = extrema(inds)
+    indmin, indmax = extrema(args)
     if indmin < 1
         throw(ArgumentError("indices must be positive"))
     end
     if indmax > ncol(df)
         throw(ArgumentError("indices must not be greater than number of columns"))
     end
-    if !allunique(inds)
+    if !allunique(args)
         throw(ArgumentError("indices must not contain duplicates"))
     end
-    copy!(_columns(df), _columns(df)[inds])
+    copy!(_columns(df), _columns(df)[args])
     x = index(df)
-    copy!(_names(x), _names(df)[inds])
+    copy!(_names(x), _names(df)[args])
     empty!(x.lookup)
     for (i, n) in enumerate(x.names)
         x.lookup[n] = i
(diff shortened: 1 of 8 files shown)


bkamins:
Fixes #2110

bkamins:
I have just realized that with the current column copying rules we do not need `ColRename` type. It is just an `identity` transformation of a column.

bkamins:
@nalimilan - could you have a look at this - it should be simple to check and contains several fixes that I would prefer to rebase other related PRs against. Thank you!

bkamins:
CI fails only due to coverage.

bkamins:
Following the comments I have also added an option to pass vectors of pairs without splatting (fortunately the change needed is minor), so now you can write:
```
julia> df = DataFrame(rand(4,3))
4×3 DataFrame
│ Row │ x1       │ x2        │ x3       │
│     │ Float64  │ Float64   │ Float64  │
├─────┼──────────┼───────────┼──────────┤
│ 1   │ 0.292811 │ 0.608872  │ 0.72134  │
│ 2   │ 0.457833 │ 0.216678  │ 0.768774 │
│ 3   │ 0.351017 │ 0.839151  │ 0.941342 │
│ 4   │ 0.626779 │ 0.0706048 │ 0.718542 │

julia> select(df, names(df) .=> mean, names(df) .=> std)
1×6 DataFrame
│ Row │ x1_mean │ x2_mean  │ x3_mean  │ x1_std   │ x2_std   │ x3_std   │
│     │ Float64 │ Float64  │ Float64  │ Float64  │ Float64  │ Float64  │
├─────┼─────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
│ 1   │ 0.43211 │ 0.433827 │ 0.787499 │ 0.146672 │ 0.353087 │ 0.105119 │
```

src/abstractdataframe/selection.jl
```diff
@@ -49,12 +42,12 @@ normalize_selection(idx::AbstractIndex, sel) =
 
 function normalize_selection(idx::AbstractIndex, sel::ColumnIndex)
     c = idx[sel]
-    return c => ColRename() => _names(idx)[c]
+    return c => identity => _names(idx)[c]
```
nalimilan: It's nice that this equivalence holds for simplicity!

src/abstractdataframe/selection.jl
```diff
@@ -123,23 +116,12 @@ function normalize_selection(idx::AbstractIndex,
...
-    # nothing is not possible otherwise as a value in this dict
-    transformed_cols[newname] = nothing
-end
-
 function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
                                     <:Pair{<:Union{Base.Callable, ByRow}, Symbol}},
                            df::AbstractDataFrame, newdf::DataFrame,
                            transformed_cols::Dict{Symbol, Any}, copycols::Bool)
     col_idx, (fun, newname) = nc
+    # it is allowed to request column tranformation only once
```
nalimilan: I don't understand this comment. You mean that it's not allowed to specify multiple times the same transformation on the same column(s)?
```suggestion
    # a given transformation can only be requested once for each set of input(s)
```

src/abstractdataframe/selection.jl
```diff
@@ -256,6 +236,17 @@ julia> select!(df, :, [:c, :b] => (c,b) -> c .+ b .- sum(b)/length(b))
 │ 1   │ 0.841471 │ 4     │ -0.158529    │
 │ 2   │ 0.909297 │ 5     │ 0.909297     │
 │ 3   │ 0.14112  │ 6     │ 1.14112      │
+
+julia> df = DataFrame(a=1:3, b=4:6);
+
+julia> select!(df, names(df) .=> sum .=> [:A, :B]);
```
nalimilan: Maybe better show the syntax without specifying output names (or both)? The most common case is probably the one where you want automatic names.

src/abstractdataframe/selection.jl
```diff
@@ -302,6 +293,16 @@ function select!(df::DataFrame, cs...)
     return df
 end
 
+"""
+    transform!(df::DataFrame, inds...)
+
+Mutate `df` in place to add columns specified by `inds...` and return it.
+Equivalent to `select!(df, :, inds...)`.
+
+See [`select!`](@ref) for a detailed rules how the transformation is applied
```
nalimilan:
```suggestion
See [`select!`](@ref) for detailed rules regarding accepted values for `inds`.
```
BTW, also in `select`, `inds` isn't a good name for this now, and the `select` docstring should mention immediately that it can be either an index *or a transformation*.

src/abstractdataframe/selection.jl
```diff
@@ -434,8 +445,17 @@ select(df::DataFrame, c::Union{AbstractVector{<:Integer}, AbstractVector{Symbol}
 select(df::DataFrame, c::ColumnIndex; copycols::Bool=true) =
     select(df, [c], copycols=copycols)
 
-select(df::DataFrame, cs...; copycols::Bool=true) =
-    _select(df, [normalize_selection(index(df), c) for c in cs], copycols)
+function select(df::DataFrame, cs...; copycols::Bool=true)
+    cs_vec = []
+    for v in cs
```
nalimilan: Actually I realize that `reduce(vcat, cs)` seems to have the same effect since it doesn't treat pairs as collections. Am I right? Also valid for the `combine` PR.

src/abstractdataframe/selection.jl
```diff
@@ -555,6 +583,18 @@ function select(dfv::SubDataFrame, inds...; copycols::Bool=true)
...
 end
+
+"""
+    transform(df::AbstractDataFrame, inds...; copycols::Bool=true)
+
+Create a new data frame that contains columns from `df` and adds columns
+specified by `inds` and return it.
+Equivalent to `select(df, :, inds..., copycols=copycols)`.
+
+See [`select`](@ref) for a detailed rules how the transformation is applied
```
nalimilan:
```suggestion
See [`select`](@ref) for detailed rules regarding accepted values for `inds`.
```

src/abstractdataframe/selection.jl
```diff
@@ -31,6 +22,8 @@ struct ByRow{T}
     fun::T
 end
 
+Base.broadcastable(x::ByRow) = Ref(x)
```
nalimilan: Is this tested?

src/abstractdataframe/selection.jl
bkamins: I will add both then. I wanted to show that you can also specify target column names this way.

src/abstractdataframe/selection.jl
bkamins: yes, this is tested in https://github.com/JuliaData/DataFrames.jl/pull/2155/files#diff-002e65995072b19e4150f93385cb5f61R897

src/abstractdataframe/selection.jl
bkamins: Unfortunately not as we do not want to unwrap vectors other than `AbstractVector{<:Pair}`:
```
julia> vcat([:a, :b], [1,2])
4-element Array{Any,1}:
  :a
  :b
 1
 2
```
and we want:
```
julia> [[:a, :b], [1,2]]
2-element Array{Array{T,1} where T,1}:
 [:a, :b]
 [1, 2]
```
Also `vcat` would lead to strange uncontrolled things like:
```
julia> vcat(:a, [:a=>:b :c=>:d])
2×2 Array{Any,2}:
    :a      :a
 :a=>:b  :c=>:d
```
that would get through and produce strange errors later (we could check for this, but for me the code I proposed was simpler, also I make `cs_vec` of eltype `Any` to save a bit of compilation cost).

src/abstractdataframe/selection.jl
bkamins: I mean that it is not allowed to make a transformation operation into the same "new column name" more than once.
This rule is handled elsewhere in the code and if such operation is passed to `select_transform!` then `@assert !hasproperty(newdf, newname)` should hold (so this line makes sure we do not have a bug in the code earlier).

By transformation we understand:
* an ordinary transformation
* column renaming
* single column selection (as we decided then we also do not want to allow for duplicates)

I will make a better comment.

src/abstractdataframe/selection.jl
bkamins: OK. I will change this.

src/abstractdataframe/selection.jl
```diff
@@ -150,12 +152,15 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
...
 
-Mutate `df` in place to retain only columns specified by `inds...` and return it.
+Mutate `df` in place to retain only columns specified by `args...` and return it.
 
-Arguments passed as `inds...` can be any index that is allowed for column indexing.
-In particular, regular expressions, `All`, `Between`, and `Not` selectors are supported.
+Arguments passed as `args...` can be:
+
+* Any index that is allowed for column indexing. In particular, regular expressions,
+  `All`, `Between`, and `Not` selectors are supported.
```
nalimilan:
```suggestion
* Any index that is allowed for column indexing. In particular, symbols, integers,
  regular expressions, `All`, `Between`, and `Not` selectors are supported.
```
Same below.

src/abstractdataframe/selection.jl
```diff
@@ -150,12 +152,15 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
...
+* Column transformation operations using the `Pair` notation that is described below.
```
nalimilan: Also vectors of indices or pairs are allowed.

bkamins: I will add it

bkamins:
TODO: sync with nrow

src/abstractdataframe/selection.jl
```diff
@@ -129,7 +129,7 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
     if col_idx isa Int
         res = fun(df[!, col_idx])
     else
-        res = fun((cdf[i] for i in col_idx)...)
+        res = fun(i -> ntuple(cdf[col_idx[i]], length(col_idx))...)
```
nalimilan:
```suggestion
        res = fun(ntuple(i -> cdf[col_idx[i]], length(col_idx))...)
```

bkamins: Yes - I have changed this in parallel before I saw your post - it is fixed now. Thank you for catching it.

bkamins:
nrow synced, now only coverage fails.

src/abstractdataframe/selection.jl
nalimilan: According to Coverall it isn't. Is it wrong?
https://coveralls.io/builds/29767300/source?filename=src%2Fabstractdataframe%2Fselection.jl#L25

bkamins: Strange - I added an explicit test already (I now double checked and the test is `transform(df, names(df) .=> ByRow(-))`). I have added 2 more tests of this.

bkamins:
OK - thank you. I will merge this and again update the "combine" PR now to reflect these changes.

Yes coverage is not 100% reliable (but it is better to have it than not to have it probably :smile:).

Issue #2089, opened 2020-01-24
Add "begin" tests for Julia 1.4


bkamins:
Add tests to make sure we properly handle `df[begin, begin]` etc. on Julia 1.4

PR #2160, opened 2020-03-22
update to Juia 1.4

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2089

+ change Julia version against which we run tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: julia
 
 julia:
   - 1.0
-  - 1.3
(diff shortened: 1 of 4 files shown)


bkamins:
We have an interesting case here. In order to test `begin` in indexing I had to have a separate file for them to be able to keep testing on 1.0 (I do not see a cleaner way to do this).

ararslan:
Yeah, I think that's the only viable solution.

PR #2163, opened 2020-03-22
Add .editorconfig and comment on GitHub Actions

jonas-schulze:
The `.editorconfig` is very close to the one I use personally. The lines about `end_of_line` and `charset` are best guesses. What do you think about it? If you want one but not the other, I'd be happy to split this PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+
+[*.jl]
+indent_style = space
+indent_size = 4
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,6 +13,7 @@ Thanks for taking the plunge!
 * Aim for atomic commits, if possible, e.g. `change 'foo' behavior like so` & `'bar' handles such and such corner case`, rather than `update 'foo' and 'bar'` & `fix typo` & `fix 'bar' better`
 * Pull requests are tested against release and development branches of Julia, so using `Pkg.test("DataFrames")` as you develop can be helpful
 * The style guidelines outlined below are not the personal style of most contributors, but for consistency throughout the project, we've adopted them
+* It is recommended to disable GitHub Actions on your fork; check Settings > Actions
 
 ## Style Guidelines
 


bkamins:
We use all the settings you have proposed, I am not sure if this is a standard to include this file in repositories.

@ararslan - can you please comment what you think?

ararslan:
I agree, it does not seem to be standard for Julia package repositories to include this. Also why do actions need to be disabled on forks?

jonas-schulze:
> I agree, it does not seem to be standard for Julia package repositories to include this.

To be honest, I haven't seen any Julia package to include an `.editorconfig`, but I think it would be really nice addition (and that there's no downside as it complies with the Style Guidelines).

> Also why do actions need to be disabled on forks?

Before opening #2157 I pushed all the changes of `JuliaData:master` to `jonas-schulze:master`, which a.o. added some GitHub Actions workflows. Then, today I got an email that a `Run failed: TagBot - master (d98b9be)`. It turned out that my fork had GitHub Actions enabled (maybe it's the default) and all the runs had failed with a time out (after 45min each).

I'd like to make other contributors aware, so that they are not as confused as I was. 🙃

I disabled Actions alltogether, but maybe someone only wants to disable third party ones (like TagBot). That's why the proposed change reads so general. An abstract of the above is in the commit message.

bkamins:
Just to be clear I think both additions are not a problem, but @ararslan can you please decide as this PR is mostly related community and workflow standards (and here I am not an expert). Thank you!

ararslan:
+1 to the comment about GitHub Actions, if that's something that was a source of confusion. I'm less enthused about the editor config file but I'm not opposed.

bkamins:
So let us wait if some more JuliaData members will have an opinion here. Thank you both!

nalimilan:
Which editors take this file into account by default?

jonas-schulze:
bump 🙂

jonas-schulze:
@bkamins bump 🙂

bkamins:
This PR is pending the approval by @nalimilan.

bkamins:
@nalimilan - I would merge this. It does not hurt to have it and it will prompt people what is the expected style, even if they do not use .editorconfig.

bkamins:
Thank you!

Issue #2162, opened 2020-03-22
Automatically fill in `select` and `combine` for scalars?


pdeffebach:
More thoughts from [this](https://discourse.julialang.org/t/dataframesmeta-jl-and-the-state-of-the-dataframes-ecosystem/36221/30) discourse thread. 

```
julia> df = DataFrame(rand(10, 5))
julia> select(df, :, [:x1] => mean => :y)
ERROR: ArgumentError: New columns must have the same length as old columns
```

In Stata, 

```
gen mean_x = mean(x)
```

would spread the argument for you, creating a full column of just `mean(x)`

I also seem to remember that DataFramesMeta used to have this behavior, but now it throws an error.

bkamins:
This was intentional because `combine` does the same:

```
julia> df.y = [ones(5); zeros(5)];

julia> by(df, :y, :x3 => x -> 1, :x4 => identity)
ERROR: ArgumentError: all functions must return values of the same length
```

However, I have also been considering if we should not add this as a third case of automatic broadcasting that is allowed (other two are in `DataFrame` and in `insertcols!`). Since this is non-breaking I have opted to disallow it and we can later allow it in both `select` and `combine` if we want. @nalimilan probably has also some thoughts here (as he designed the `combine` part).

There are special cases of several types that would be excluded for `select` in the process even if we allowed it (`AbstractDataFrame`, `NamedTuple`, `DataFrameRow`, `AbstractMatrix`) due to compatibility with `combine` and future extensions, but this should not be problematic I think.

pdeffebach:
Thanks for this. 

I understand more the reasoning. If you have one function returning a scalar and another returns a vector, there is ambiguiity. 

I agree that `AbstractDataFrame` and friends should be added eventually so they should throw errors for now. But they are already special-cased to throw errors. 

I'm not sure about SemVer but if this is a post-1.0 thing we can wait.

bkamins:
It is not hard to add, the question is if we want to add it.

> If you have one function returning a scalar and another returns a vector, there is ambiguiity.

There is a kind of "defensive" ambiguity now (i.e. we do not implicitly broadcast). But we could without a problem, i.e. all functions would have to return either one element or vectors of the same length (just like in broadcasting). We would just have to check if it would not have significant consequences for the performance of `combine` (@nalimilan can tell best, but I feel that this would be doable, although current design of `combine` separates `row` and `table` oriented return values from passed functions).

[...]

PR #2165, opened 2020-03-23
[BREAKING] allow pseudo-broadcasting in select

bkamins:
Resolves the `select` part of https://github.com/JuliaData/DataFrames.jl/issues/2162.

I have disallowed "squashing" scalars to 0-rows if vectors have 0-rows. So this is a decision to make what we should do (I disallowed it as later allowing it would be non-breaking).

BTW `DataFrame` and `insertcols!` are currently inconsistent here (`DataFrame` does not allow squashing and `insertcols!` allows it). We should decide what should be the target functionality and make it consistent across these three use cases (and `combine` in the future)

### DIFF
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -170,61 +170,66 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
     transformed_cols[newname] = nothing
 end
 
+SELECT_ARG_RULES =
+    """
+    Arguments passed as `args...` can be:
+
+    * Any index that is allowed for column indexing. In particular, symbols, integers,
+      vectors of symbols, vectors of integers, vectors of bools, regular expressions,
+      `All`, `Between`, and `Not` selectors are supported.
+    * Column transformation operations using the `Pair` notation that is described below
+      and vectors of such pairs.
+
+    Columns can be renamed using the `old_column => new_column_name` syntax,
+    and transformed using the `old_column => fun => new_column_name` syntax.
+    `new_column_name` must be a `Symbol`, and `fun` a function or a type. If `old_column`
+    is a `Symbol` or an integer then `fun` is applied to the corresponding column vector.
+    Otherwise `old_column` can be any column indexing syntax, in which case `fun`
+    will be passed the column vectors specified by `old_column` as separate arguments.
+    If `fun` returns a value of type other than `AbstractVector` then it will be broadcasted
+    into a vector matching the target number of rows in the data frame,
+    unless its type is one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
+    `AbstractMatrix`, in which case an error is thrown as currently these
+    return types are not allowed.
+    As a particular rule, values wrapped in a `Ref` or a `0`-dimensional `AbstractArray`
+    are unwrapped and then broadcasted.
+
+    To apply `fun` to each row instead of whole columns, it can be wrapped in a `ByRow`
+    struct. In this case if `old_column` is a `Symbol` or an integer then `fun` is applied
+    to each element (row) of `old_column` using broadcasting. Otherwise `old_column` can be
+    any column indexing syntax, in which case `fun` will be passed one argument for each of
+    the columns specified by `old_column`. If `ByRow` is used it is not allowed for
+    `old_column` to select an empty set of columns.
+
+    Column transformation can also be specified using the short `old_column => fun` form.
+    In this case, `new_column_name` is automatically generated as `\$(old_column)_\$(fun)`.
+    Up to three column names are used for multiple input columns and they are joined
+    using `_`; if more than three columns are passed then the name consists of the
+    first two names and `etc` suffix then, e.g. `[:a,:b,:c,:d] => fun` produces
+    the new column name `:a_b_etc_fun`.
+
+    Column renaming and transformation operations can be passed wrapped in vectors
+    (this is useful when combined with broadcasting).
+
+    As a special rule passing `nrow` without specifying `old_column` creates a column named `:nrow`
+    containing a number of rows in a source data frame, and passing `nrow => new_column_name`
+    stores the number of rows in source data frame in `new_column_name` column.
+
+    If a collection of column names is passed to `select!` or `select` then requesting
+    duplicate column names in target data frame are accepted (e.g. `select!(df, [:a], :, r"a")`
+    is allowed) and only the first occurrence is used. In particular a syntax to move
+    column `:col` to the first position in the data frame is `select!(df, :col, :)`.
+    On the contrary, output column names of renaming, transformation and single column
+    selection operations must be unique, so e.g. `select!(df, :a, :a => :a)` or
+    `select!(df, :a, :a => ByRow(sin) => :a)` are not allowed.
+    """
+
 """
     select!(df::DataFrame, args...)
 
 Mutate `df` in place to retain only columns specified by `args...` and return it.
 
-Arguments passed as `args...` can be:
-
-* Any index that is allowed for column indexing. In particular, symbols, integers,
-  vectors of symbols, vectors of integers, vectors of bools, regular expressions,
-  `All`, `Between`, and `Not` selectors are supported.
-* Column transformation operations using the `Pair` notation that is described below
-  and vectors of such pairs.
-
-Columns can be renamed using the `old_column => new_column_name` syntax,
-and transformed using the `old_column => fun => new_column_name` syntax.
-`new_column_name` must be a `Symbol`, and `fun` a function or a type. If `old_column`
-is a `Symbol` or an integer then `fun` is applied to the corresponding column vector.
-Otherwise `old_column` can be any column indexing syntax, in which case `fun`
-will be passed the column vectors specified by `old_column` as separate arguments.
-If `fun` returns a value of type other than `AbstractVector` then it will be broadcasted
-into a vector matching the target number of rows in the data frame,
-unless its type is one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
-`AbstractMatrix`, in which case an error is thrown as currently these
-return types are not allowed.
-As a particular rule, values wrapped in a `Ref` or a `0`-dimensional `AbstractArray`
-are unwrapped and then broadcasted.
-
-To apply `fun` to each row instead of whole columns, it can be wrapped in a `ByRow`
-struct. In this case if `old_column` is a `Symbol` or an integer then `fun` is applied
-to each element (row) of `old_column` using broadcasting. Otherwise `old_column` can be
-any column indexing syntax, in which case `fun` will be passed one argument for each of
-the columns specified by `old_column`. If `ByRow` is used it is not allowed for
-`old_column` to select an empty set of columns.
-
-Column transformation can also be specified using the short `old_column => fun` form.
-In this case, `new_column_name` is automatically generated as `\$(old_column)_\$(fun)`.
-Up to three column names are used for multiple input columns and they are joined
-using `_`; if more than three columns are passed then the name consists of the
-first two names and `etc` suffix then, e.g. `[:a,:b,:c,:d] => fun` produces
-the new column name `:a_b_etc_fun`.
-
-Column renaming and transformation operations can be passed wrapped in vectors
-(this is useful when combined with broadcasting).
-
-As a special rule passing `nrow` without specifying `old_column` creates a column named `:nrow`
-containing a number of rows in a source data frame, and passing `nrow => new_column_name`
-stores the number of rows in source data frame in `new_column_name` column.
-
-If a collection of column names is passed to `select!` then requesting duplicate column
-names in target data frame are accepted (e.g. `select!(df, [:a], :, r"a")` is allowed)
-and only the first occurrence is used. In particular a syntax to move column `:col`
-to the first position in the data frame is `select!(df, :col, :)`.
-On the contrary, output column names of renaming, transformation and single column
-selection operations must be unique, so e.g. `select!(df, :a, :a => :a)` or
-`select!(df, :a, :a => ByRow(sin) => :a)` are not allowed.
+$SELECT_ARG_RULES
 
 Note that including the same column several times in the data frame via renaming
 or transformations that return the same object without copying will create column aliases.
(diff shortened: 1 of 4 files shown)


src/abstractdataframe/selection.jl
```diff
@@ -483,6 +509,9 @@ function _select(df::AbstractDataFrame, normalized_cs, copycols::Bool)
             transformed_cols[newname] = nc
         end
     end
+    # we allow resizing newdf only if up to some point only scalars were put
```
pdeffebach: I think that this logic would actually supercede all of `combine`'s features. 

Eventually, we want `select` to work on a grouped data frame, performing each operation by group in order to allow `select(gdf, :, :x1 => (x -> x .- mean(x)))`. But with this automatic collapsing with the following call

```
select(gdf, :x1 => mean, :x1 => var)
```

we would get the exact same behavior as `combine`. 

This would be an interesting direction. I was expecting always `spread`ing to have the same number of rows as `nrow` df. 

note that the above ambiguity only applies when you aren't keeping any of the old columns.

bkamins: > Eventually, we want `select` to work on a grouped data frame

Yes - and we want then it to behave exactly like `combine` except that it will return a `GroupedDataFrame` (and not a `DataFrame` like combine)

> I was expecting always spreading to have the same number of rows as `nrow(df)`.

This would conflict with what `combine` does now. We can discuss if `1` row or `nrow(df)` is better in this case, but at least in `combine` we produce `1` row (which is intuitive at least for me).

> note that the above ambiguity only applies when you aren't keeping any of the old columns.

Exactly - this is what I handle in the code.

src/abstractdataframe/selection.jl
```diff
@@ -319,9 +342,12 @@ and transformed using the `old_column => fun => new_column_name` syntax.
 is a `Symbol` or an integer then `fun` is applied to the corresponding column vector.
 Otherwise `old_column` can be any column indexing syntax, in which case `fun`
 will be passed the column vectors specified by `old_column` as separate arguments.
-If `fun` returns a value of type other than `AbstractVector` then it will be wrapped
-into a 1-element vector, unless its type is one of `AbstractDataFrame`, `NamedTuple`,
-`DataFrameRow`, `AbstractMatrix`, in which case an error is thrown as currently these
+If `fun` returns a value of type other than `AbstractVector` then: it will be wrapped
```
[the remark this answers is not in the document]
bkamins: right - fixed (wrapped was a leftover from the previous version when I did not do pseudo broadcasting).

src/abstractdataframe/selection.jl
bkamins: And what is your opinion about spreading to 0 rows - should we allow such a spread or throw an error (current behavior)?

pdeffebach: > And what is your opinion about spreading to 0 rows - should we allow such a spread or throw an error (current behavior)?

So the current behavior is 

```
julia> select(df, :x1 => (x -> []))
0×1 DataFrame

```

and 

```
julia> select(df,:, :x1 => (x -> []))
ERROR: ArgumentError: New columns must have the same length as old columns
```

Versus `Ref`

julia> select(df,:x1 => (x -> Ref([])))
1×1 DataFrame
│ Row │ x1_function            │
│     │ Array{Any,1}           │
├─────┼────────────────────────┤
│ 1   │ 0-element Array{Any,1} │

julia> select(df,:, :x1 => (x -> Ref([])))
5×4 DataFrame
│ Row │ x1       │ x2         │ a     │ x1_function            │
│     │ Float64  │ Float64    │ Int64 │ Array{Any,1}           │
├─────┼──────────┼────────────┼───────┼────────────────────────┤
│ 1   │ 0.435143 │ 0.849017   │ 1     │ 0-element Array{Any,1} │
│ 2   │ 0.943903 │ 0.00494995 │ 1     │ 0-element Array{Any,1} │
│ 3   │ 0.351229 │ 0.898012   │ 2     │ 0-element Array{Any,1} │
│ 4   │ 0.689727 │ 0.65101    │ 2     │ 0-element Array{Any,1} │
│ 5   │ 0.866909 │ 0.343819   │ 2     │ 0-element Array{Any,1} │
```

If this is what you mean, I think this is good behavior. 

Sorry about not realizing `transform` would preserve grouping. This is good, as long as `combine(gdf, :, :x1 => mean)`  works and "spreads" everything.

Comment (bkamins) [reply src/abstractdataframe/selection.jl]: What I mean with 0-length broadcasting is that:
```
julia> select(df, :, [] => (() -> 9) => :x)
ERROR: ArgumentError: New columns must have the same length as old columns
```
now errors, while:
```
julia> df = DataFrame(a=[])
0×1 DataFrame


julia> insertcols!(df, 1, :x => 9)
0×2 DataFrame
```
works just like:
```
julia> v = []
0-element Array{Any,1}

julia> v .= 9
0-element Array{Any,1}
```
works, but this fails:
```
julia> DataFrame(a=[], b=1)
ERROR: DimensionMismatch("column length 0 for column(s) a, and is incompatible with column length 1 for column(s) b")
```
So the question is if when we pseudo-broadcast a scalar into a 0-row container should that be allowed or no (I see pros and cons of both approaches and we are currently inconsistent already so we should pick one).

> `combine(gdf, :, :x1 => mean)`  works and "spreads" everything.

yes, it would spread (it is just not implemented yet)

File: src/abstractdataframe/selection.jl [hunk 3]
```diff
@@ -125,11 +125,20 @@ end
 
 function select_transform!(nc::Pair{Int, Pair{ColRename, Symbol}},
                            df::AbstractDataFrame, newdf::DataFrame,
-                           transformed_cols::Dict{Symbol, Any}, copycols::Bool)
+                           transformed_cols::Dict{Symbol, Any}, copycols::Bool,
+                           allow_resizing_newdf::Base.RefValue{Bool})
```
nalimilan, line 129: `Ref` is the supertype so it works:
```suggestion
                           allow_resizing_newdf::RefBool})
```

src/abstractdataframe/selection.jl
```diff
@@ -138,7 +147,8 @@ end
 function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
                                     <:Pair{<:Union{Base.Callable, ByRow}, Symbol}},
                            df::AbstractDataFrame, newdf::DataFrame,
-                           transformed_cols::Dict{Symbol, Any}, copycols::Bool)
+                           transformed_cols::Dict{Symbol, Any}, copycols::Bool,
+                           allow_resizing_newdf::Base.RefValue{Bool})
```
nalimilan, line 151:
```suggestion
                           allow_resizing_newdf::Ref{Bool})
```

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
nalimilan: Why not `broadcasted`? That's the standard term in Julia. "recycle" would also make sense (in the R tradition).

src/abstractdataframe/selection.jl
```diff
@@ -179,9 +199,12 @@ and transformed using the `old_column => fun => new_column_name` syntax.
...
-If `fun` returns a value of type other than `AbstractVector` then it will be wrapped
-into a 1-element vector, unless its type is one of `AbstractDataFrame`, `NamedTuple`,
-`DataFrameRow`, `AbstractMatrix`, in which case an error is thrown as currently these
+If `fun` returns a value of type other than `AbstractVector` then: it will be spread
+into a vector matching the target number of rows in the data frame
+(as a particular rule a values stored in a `Ref` or a `0`-dimensional `AbstractArray`
+ are unwrapped and treated in the same way),
+unless its type is one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
+`AbstractMatrix`, in which case an error is thrown as currently these
 return types are not allowed.
```
nalimilan, line 208:
```suggestion
If `fun` returns a value of type other than `AbstractVector` then it will be broadcasted
into a vector matching the target number of rows in the data frame,
unless its type is one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
`AbstractMatrix`, in which case an error is thrown as currently these
return types are not allowed.
As a particular rule, values wrapped in a `Ref` or a `0`-dimensional `AbstractArray`
are unwrapped and then broadcasted.
```

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
nalimilan: Another term is "repeated".

src/abstractdataframe/selection.jl
```diff
@@ -125,11 +125,20 @@ end
...
+                           transformed_cols::Dict{Symbol, Any}, copycols::Bool,
+                           allow_resizing_newdf::Base.RefValue{Bool})
     col_idx, (_, newname) = nc
     # it is allowed to request column tranformation only once
     @assert !hasproperty(newdf, newname)
+    # disallow shortening to 0 rows
+    if allow_resizing_newdf[] && nrow(newdf) == 1 && nrow(df) > 1
+        newdfcols = _columns(newdf)
+        for (i, col) in enumerate(newdfcols)
+            newdfcols[i] = fill(first(col), nrow(df))
```
nalimilan, line 137: This is needed for `CategoricalArray` I think:
```suggestion
            newdfcols[i] = fill!(similar(col, nrow(df)), first(col))
```

src/abstractdataframe/selection.jl
```diff
@@ -160,7 +178,9 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
             newdf[!, newname] = res
         end
     else
-        newdf[!, newname] = [res]
+        res_unwrap = res isa Union{AbstractArray{<:Any, 0}, Ref} ? res[] : res
+        # disallow squashing a scalar to 0 rows
+        newdf[!, newname] = fill(res_unwrap, max(1, nrow(newdf)))
```
nalimilan, line 183:
```suggestion
        newdf[!, newname] = fill!(Tables.allocatecolumn(typeof(res_unwrap), max(1, nrow(newdf))), res_unwrap)
```

test/select.jl
```diff
@@ -819,6 +819,49 @@ end
     @test_throws ArgumentError select(sdf, :x1 => identity => :r1, copycols=false)
 end
 
+@testset "pseudo-broadcasting" begin
+    df = DataFrame([1 2 3
+                    4 5 6])
+    df2 = DataFrame([1 2 3])
+    df3 = DataFrame(x1=Int[], x2=Int[], x3=Int[])
+    for v in [9, Ref(9), view([9], 1)]
+        @test select(df, [] => (() -> 9) => :a, :, (:) => (+) => :d) ==
```
nalimilan, line 828: I think you wanted to use `v` instead of `9`, right?

src/abstractdataframe/selection.jl
bkamins: Indeed - it is always tricky.

src/abstractdataframe/selection.jl
bkamins: maybe `fill` should be overloaded in CategoricalArrays.jl. Its contract in Base does not promise a standard array?

test/select.jl
bkamins: sure. now you see how I developed these tests :). Corrected

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
bkamins: changed to your "broadcasted" suggestion

bkamins:
Changes applied. I had a `fill` bug in two other places.

bkamins:
@nalimilan - maybe we should allow pseudo-broadcasting into 0-rows (now we disallow it). This would allow dropping groups in `combine` if in some group has 0 rows in "vectorized" part, something like:
```
combine(gdf, :col => x -> filter(>(0), x) => pos_col, nrow)
```
would now fail in pseudo-broadcasting if some group had `:col` with only negative entries.

bkamins:
OK - I have allowed broadcasting into 0-row data frame, to make it consistent with the work I do for https://github.com/JuliaData/DataFrames.jl/issues/2166 now.

src/abstractdataframe/selection.jl
nalimilan: Yes, probably. I've filed https://github.com/JuliaData/CategoricalArrays.jl/issues/257.

nalimilan: Can you add tests with `CategoricalArray` BTW?

bkamins: Tests added - all was good, but still I discovered that there are many possible cases we should test against.

test/select.jl
```diff
@@ -889,6 +891,93 @@ end
...
           DataFrame(ones(0, 4))
+
+    df = DataFrame(x1=1:2, x2=categorical(1:2),
+                   x3=[missing,2], x4=categorical([missing, 2]))
+
+    df2 = select(df, names(df) .=> first)
+    @test df2 ≅ DataFrame(x1_first=1, x2_first=1, x3_first=missing,
+                          x4_first=missing)
+    @test df2.x1_first isa Vector{Int}
+    @test df2.x2_first isa CategoricalVector
```
nalimilan, line 902: While you're at it, test the element type? That's especially useful for `last` as `Missing` could be dropped.

bkamins: ok - I made tests more specific

bkamins:
I have also synchronized `DataFrame` constructor with our current broadcasting rules. This made me fix one bug in the constructor.

test/select.jl
```diff
@@ -899,24 +899,24 @@ end
     @test df2 ≅ DataFrame(x1_first=1, x2_first=1, x3_first=missing,
                           x4_first=missing)
     @test df2.x1_first isa Vector{Int}
-    @test df2.x2_first isa CategoricalVector
+    @test df2.x2_first isa CategoricalVector{Int}
     @test df2.x3_first isa Vector{Missing}
     @test df2.x4_first isa Vector{Missing}
```
nalimilan, line 904: This result isn't ideal, as levels are lost. But I guess there's no general solution to this.

src/dataframe/dataframe.jl
```diff
@@ -104,37 +105,44 @@ struct DataFrame <: AbstractDataFrame
...
-        for (i, c) in enumerate(columns)
-            if isa(c, AbstractRange)
-                columns[i] = collect(c)
-            elseif !isa(c, AbstractVector)
-                throw(DimensionMismatch("columns must be 1-dimensional"))
-            elseif copycols
-                columns[i] = copy(c)
+        len == -1 && (len = 1) # we got no vectors so make one row of scalars
+
+        # we write into columns as we know that it is guarantted
```
nalimilan, line 126: It doesn't sound right for an inner to constructor to rely on the behavior of outer constructors, since it can be called directly by user code. Is there another solution?

```suggestion
        # we write into columns as we know that it is guaranteed
```

test/select.jl
bkamins: I think the decision how this should be handled should be in CategoricalArrays.jl not here - here we should use the mechanics exposed there. Do you agree?

src/dataframe/dataframe.jl
bkamins: > since it can be called directly by user code

it cannot, as we do not expose `Index`.

> It doesn't sound right for an inner to constructor to rely on the behavior of outer constructors

It already did - I simply corrected a bug in an outer constructor.

> Is there another solution?

Yes, but it would involve making an additional copy of `columns` vector inside an inner constructor, which I understand we wanted to avoid (outer constructors now already alsways make copies and these copies cannot be avoided there).

nalimilan: OK, let's say it's an internal outer constructor then. :-D

test/select.jl
nalimilan: I'm not sure it can be fixed, even there. Basically it's due to the fact that we allocate the column based on the type of the returned value, not based on the possible types that could have been returned. That's the same problem as with `[v for v in x]` when `x` happens to contain only missing values. We'll have to live with it, or move to `DataValue{Union{Missing, CategoricalValue{Int}}`.

bkamins: Ah - right

bkamins:
Thank you!

Issue #1704, opened 2019-01-30
Group Indices function


matthieugomez:
It would be great to have a function that returns a vector of unique identifier for each group of a GroupedDataframe (as a `PooledDataArray` or as a `CategoricalVector`). Similar functions exist in other packages:  `group_indices` in `dplyr`, `.GRP` in `data.table`, `group` in Stata. In particular, it would make it easier for external packages to benefit from the all recent work done in `group_rows`.

bkamins:
You can get it like this:
```
getindex.(gd, 1, (gd.cols,))
```

or like this
```
copy.(getindex.(gd, 1, (gd.cols,)))
```

But I agree that it would be good to define a function that does this in the package (recently we have discussed this with @nalimilan).

What name for such a function would you see and should it return `DataFrameRow`s or rather `NamedTuple`s?

nalimilan:
@matthieugomez Can you detail what kind of output you would expect? A single vector giving groups? Then it's present as integer indices in `gd.groups`. Or do you need labels too?

bkamins:
Yes - now I see that `dplyr` function `group_indices` is something we have as `groups` field.
Maybe it would be good to expose it also (probably wrapping in `ReadOnlyArray`).

matthieugomez:
Oh yes you're right, an integer vector would be enough. The only thing is that groups with missing values should return a certain integer (0 or some other value, as long as it is documented).

nalimilan:
OK. We already use `0` for missing values when `skipmissing=true`. I guess we could add a `group_indices` or `groupindices` function (unless we can override an already existing and more standard function for that).

matthieugomez:
Thinking about it, maybe it would be safer to return a missing value for groups where one variable is missing when `skipmissing = true`.

nalimilan:
We could do that, but that would require allocating a new vector. What do other implementations do?

matthieugomez:
By default, Stata returns missing values in case one variable has a missing value. With the option missing, missing values are considered like distinct values.
Dplyr and datatable always consider missing values as distinct values.

matthieugomez:
By the way, is not it a bit weird that groups with a missing value are included by default in `groupby` `by` etc...? Would not it be less error-prone to remove them? (I think Stata and Panda remove them, but R dplyr and datatable keep them).

bkamins:
It think it is very good that you raise these issues before DataFrames.jl 1.0, as now we have time to change the API if we want.

[...]

PR #2167, opened 2020-03-25
[BREAKING] add groupcols and valuecols functions

bkamins:
`valuevars` function will be a useful complement to `groupvars` function.

If we have it we can actually deprecate `aggregate` already now, as then:
```
aggregate(df, :a, sum)
```
becomes
```
combine(gdf, valuevars(gdf) .=> sum)
```

@nalimilan In particular that adding broadcasting of `All` will not allow to deprecate `aggregate` as `All()` would include grouping columns which are skipped by `aggregate`.

If you think that some other name would be better please comment.

### DIFF
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -23,8 +23,8 @@ by
 combine
 groupby
 groupindices
-groupvars
-valuevars
+groupcols
+valuecols
 keys
 get
 map
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -24,6 +24,7 @@ combine
 groupby
 groupindices
 groupvars
+valuevars
 keys
 get
 map
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -41,7 +41,7 @@ export AbstractDataFrame,
        flatten,
        groupby,
        groupindices,
-       groupvars,
+       groupcols,
        innerjoin,
        insertcols!,
        leftjoin,
@@ -62,7 +62,7 @@ export AbstractDataFrame,
        transform!,
        unique!,
        unstack,
-       valuevars
+       valuecols
 
 if VERSION >= v"1.1.0-DEV.792"
     import Base.eachcol, Base.eachrow
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -59,7 +59,8 @@ export AbstractDataFrame,
        semijoin,
        stack,
        unique!,
-       unstack
+       unstack,
+       valuevars
 
 if VERSION >= v"1.1.0-DEV.792"
     import Base.eachcol, Base.eachrow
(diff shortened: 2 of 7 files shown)


nalimilan:
Good catch. It probably makes sense to add this only if we do deprecate `aggregate`.

About the name, I don't remember whether we discussed the choice of "vars" rather than "cols" or "keys" when `groupvars` was introduced.

bkamins:
I thought you wanted to deprecate `aggregate` :) and this PR is exactly aimed to allow for this (but maybe users will find it useful also in other contexts?).

*Question 1* So - do we go for it?

Regarding the name - I do not think we discussed it.
Indeed probably `cols` would be better than `vars`, i.e. `groupcols` instead of `groupvars`. I can add such a deprecation in this PR.

*Question 2* Should we change `groupvars` to `groupcols`?

*Question 3* Would then `valuecols` be OK (i.e. do we think that non-grouping columns in `GroupedDataFrame` can be called value columns)?

nalimilan:
> I thought you wanted to deprecate `aggregate` :) and this PR is exactly aimed to allow for this (but maybe users will find it useful also in other contexts?).

Yes but I thought *you* were hesitant about that. :-p

Actually I found the issue where we discussed `groupvars`: https://github.com/JuliaData/DataFrames.jl/issues/1704#issuecomment-459432627. So the name comes from dplyr, so the question is whether it's worth diverging from it to be consistent with our convention. I'd say yes.

"value columns" sounds goo, though I haven't thought too much about it. It's a bit similar to `stack`/`unstack`.

bkamins:
I was hesitant about `aggregate`, but given the functionality we add I think we can drop it :smile:.

Yes `value` is similar to what `stack`/`unstack` produce, but I think it is OK (though not identical, as in the docs we call them "measure" variables).

But then maybe new names should be: `keycols` and `valuecols`. The reason is that now `GroupedDataFrame` supports key-value interface and in particular we support `keys` etc.?

nalimilan:
I'm hesitant between `groupcols` and `keycols`. "keys" might not be very intuitive.

bkamins:
The intuition is that now `groupby` essentially adds a key to a data frame. But I am OK with `groupcols` also. So - do you have a preference, or we should run a quick pool on #data on Slack?

nalimilan:
Let's make a poll, it's always fun...

JeffreySarnoff:
I have become disenamoured with method names that use "key" and do not open locks.

bkamins:
I have reviewed `aggregate` and cleaned it up a bit. The conclusion is that I would keep it for corner-case uses (i.e. passing multiple anonymous functions), but not develop it further - just keep the functionality we have now.

nalimilan:
> The conclusion is that I would keep it for corner-case uses (i.e. passing multiple anonymous functions), but not develop it further - just keep the functionality we have now.

Can you show examples of things that cannot be done with `combine`? It's going to be so powerful now that I wonder what's the point of having another limited function like `aggregate`.

bkamins:
I will shortly open a PR deprecating `aggregate` per discussion and comment there.

bkamins:
Thank you!

PR #2158, opened 2020-03-21
[BREAKING] sync combine with select

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2156.

I have initially implemented the functionality discussed in #2156 and added `keepkeys` kwarg. Here is an example:
```
julia> df = DataFrame(rand(3, 4))
3×4 DataFrame
│ Row │ x1       │ x2       │ x3       │ x4       │
│     │ Float64  │ Float64  │ Float64  │ Float64  │
├─────┼──────────┼──────────┼──────────┼──────────┤
│ 1   │ 0.181501 │ 0.794075 │ 0.86532  │ 0.417367 │
│ 2   │ 0.743854 │ 0.257419 │ 0.350772 │ 0.293578 │
│ 3   │ 0.431316 │ 0.922369 │ 0.862359 │ 0.362968 │

julia> @time by(df, [:x1, :x2], :x3 => ByRow(cos))
  6.172474 seconds (20.32 M allocations: 1.041 GiB, 6.61% gc time)
3×3 DataFrame
│ Row │ x1       │ x2       │ x3_cos   │
│     │ Float64  │ Float64  │ Float64  │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 0.181501 │ 0.794075 │ 0.648397 │
│ 2   │ 0.743854 │ 0.257419 │ 0.939108 │
│ 3   │ 0.431316 │ 0.922369 │ 0.650648 │

julia> @time by(df, [:x1, :x2], :x3 => ByRow(cos), [:x1, :x2] => (x1, x2) -> x1 ./ x2)
  0.907711 seconds (2.40 M allocations: 122.147 MiB, 2.84% gc time)
3×4 DataFrame
│ Row │ x1       │ x2       │ x3_cos   │ x1_x2_function │
│     │ Float64  │ Float64  │ Float64  │ Float64        │
├─────┼──────────┼──────────┼──────────┼────────────────┤
│ 1   │ 0.181501 │ 0.794075 │ 0.648397 │ 0.228569       │
│ 2   │ 0.743854 │ 0.257419 │ 0.939108 │ 2.88967        │
│ 3   │ 0.431316 │ 0.922369 │ 0.650648 │ 0.467618       │

julia> @time by(df, [:x1, :x2], :x3 => ByRow(cos), [:x1, :x2] => (x1, x2) -> x1 ./ x2, keepkeys=false)
[...]

PR #2168, opened 2020-03-26
add nrow to select

bkamins:
Allow an easy way to get a number of rows in `select`. It was such a clean idea that I could not resist it (but maybe other people will think otherwise).

I think having this as a special case is justified.

When this is merged the full benefit will be in https://github.com/JuliaData/DataFrames.jl/pull/2158, as `nrow` is routinely required in split-apply-combine operations and now we did not have a clean and quick way to do it.

### DIFF
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -47,6 +47,13 @@ normalize_selection(idx::AbstractIndex, sel) =
         end
     end
 
+
+normalize_selection(idx::AbstractIndex, sel::Pair{typeof(nrow), Symbol}) =
+    length(idx) == 0 ? (Int[] => (() -> 0) => last(sel)) : (1 => length => last(sel))
+
+normalize_selection(idx::AbstractIndex, sel::typeof(nrow)) =
+    normalize_selection(idx, nrow => :nrow)
+
 function normalize_selection(idx::AbstractIndex, sel::ColumnIndex)
     c = idx[sel]
     return c => ColRename() => _names(idx)[c]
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/selection.jl
```diff
@@ -198,6 +205,10 @@ using `_`; if more than three columns are passed then the name consists of the
 first two names and `etc` suffix then, e.g. `[:a,:b,:c,:d] => fun` produces
 the new column name `:a_b_etc_fun`.
 
+As a special rule passing `nrow` as an argument creates a column named `:nrow`
+containing a number of rows in a source data frame, and passing `nrow => new_column_name`
+stores a number of rows in a source data frame in `new_column_name` column.
```
nalimilan:
```suggestion
stores the number of rows in source data frame in `new_column_name` column.
```

src/abstractdataframe/selection.jl
```diff
@@ -338,6 +349,10 @@ using `_`; if more than three columns are passed then the name consists of the
...
```
nalimilan:
```suggestion
stores the number of rows in source data frame in `new_column_name` column.
```

src/abstractdataframe/selection.jl
```diff
@@ -198,6 +205,10 @@ using `_`; if more than three columns are passed then the name consists of the
 first two names and `etc` suffix then, e.g. `[:a,:b,:c,:d] => fun` produces
 the new column name `:a_b_etc_fun`.
 
+As a special rule passing `nrow` as an argument creates a column named `:nrow`
```
nalimilan: Maybe that's more explicit?
```suggestion
As a special rule passing `nrow` without specifying `old_column` creates a column named `:nrow`
```

bkamins:
Thank you. I will integrate it to other PRs now.

Issue #2166, opened 2020-03-24
[BREAKING] Making combine more flexible


bkamins:
I am splitting a discussion of additional functionalities of combine to a separate issue (to keep size of changes manageable). The two related things that should be added are:
* allow column selection and renaming in combine (just like in `select`)
* allow pseudo-broadcasting of scalars if they are mixed with vectors (just like in `select` very soon)

This is non breaking, so can go for 1.x release (but probably we will have it before this but this is not blocking 0.21 release)

pdeffebach:
Added to this is a possible `combine(gd, :, cols => fun)` so a user can transform a column by group and keep all the others, returning a data frame. We can do the same with `by`.

bkamins:
Agreed (if I understand your point correctly this is a special case of "allow column selection and renaming in `combine` - just writing it to make sure we have the same understanding what I meant above).

bkamins:
(first note - column renaming is already here - we got it for free)
 
@nalimilan - we will have to work on this issue for 1.0. The reason is that we currently have a very significant bug in `_combine` (I am showing results from 0.20.2, the same is on master).

When solving it I can try implement implicit broadcasting as a side effect I think (I will have to double check to make sure):

```
julia> df = DataFrame(g=[1,2]);

julia> f1(i) = i[1] == 1 ? ["a", "b"] : ["c"];

julia> f2(i) = i[1] == 1 ? ["d"] : ["e", "f"];

julia> by(df, :g, :g => f1, :g => f2)
res = [([1, 1, 2], ["a", "b", "c"]), ([1, 2, 2], ["d", "e", "f"])]
3×3 DataFrame
│ Row │ g     │ g_f1   │ g_f2   │
│     │ Int64 │ String │ String │
├─────┼───────┼────────┼────────┤
│ 1   │ 1     │ a      │ d      │
│ 2   │ 1     │ b      │ e      │
│ 3   │ 2     │ c      │ f      │
```

bkamins:
Also, for a reference, we kind of already allow mixing single values and vectors in special cases (which is also a bug):
```
julia> f1(i) = i[1] == 1 ? ["a"] : ["c"];

julia> f2(i) = i[1] == 1 ? "d" : "e";

julia> df = DataFrame(g=[1,2]);

julia> by(df, :g, :g => f1, :g => f2)
2×3 DataFrame
│ Row │ g     │ g_f1   │ g_f2   │
│     │ Int64 │ String │ String │
[...]

PR #2170, opened 2020-04-05
[BREAKING] add to combine and by: column selection, pseudo broadcasting, fix bug with unequal column lengths

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2166

I still need to update the documentation and tests, but all from #2166 is implemented here.

One comment is that we treat `NamedTuple` as-is (i.e. we do not unpack `NamedTuples` when pseudo-broadcasting)

### DIFF
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -47,9 +47,9 @@ The kind of return value and the number and names of columns must be the same fo
 
 It is allowed to mix single values and vectors if multiple transformations
 are requested. In this case single value will be broadcasted to match the length
-of columns specified by returned vectors. In this case as a convinience values stored
-in `Ref` and 0-dimensional arrays are extracted from them.
-
+of columns specified by returned vectors. 
+As a particular rule, values wrapped in a `Ref` or a `0`-dimensional `AbstractArray`
+are unwrapped and then broadcasted.
 
 If a single value or a vector is returned by the `function` and `target_col` is not
 provided, it is generated automatically, by concatenating source column name and
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -11,22 +11,22 @@ function, which is a shorthand for `groupby` followed by `map` and/or `combine`.
 `by` takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split
 the `DataFrame` on, and (3) a specification of one or more functions to apply to
 each subset of the `DataFrame`. This specification can be of the following forms:
-1. a `cols => function` pair indicating that `function` should be called with
-   positional arguments holding columns `cols`, which can be a any valid column selector
-2. a `cols => function => target_col` form additionally
-   specifying the name of the target column (this assumes that `function` returns a single value or a vector)
-3. a `col => target_col` pair, which renames the column `col` to `target_col`
-4. a `nrow` or `nrow => target_col` form which efficiently computes the number of rows in a group
-   (without `target_col` the new column is called `:nrow`)
-5. standard column selectors (integers, symbols, vectors of integers, vectors of symbols,
+1. standard column selectors (integers, symbols, vectors of integers, vectors of symbols,
    `All`, `:`, `Between`, `Not` and regular expressions)
-5. several arguments of the forms given above, or vectors thereof
-6. a function which will be called with a `SubDataFrame` corresponding to each group;
+2. a `cols => function` pair indicating that `function` should be called with
+   positional arguments holding columns `cols`, which can be a any valid column selector
+3. a `cols => function => target_col` form additionally
+   specifying the name of the target column (this assumes that `function` returns a single value or a vector)
+4. a `col => target_col` pair, which renames the column `col` to `target_col`
+5. a `nrow` or `nrow => target_col` form which efficiently computes the number of rows in a group
+   (without `target_col` the new column is called `:nrow`)
+6. several arguments of the forms given above, or vectors thereof
+7. a function which will be called with a `SubDataFrame` corresponding to each group;
    this form should be avoided due to its poor performance unless a very large
    number of columns are processed (in which case `SubDataFrame` avoids excessive
    compilation)
 
-All forms except 6 can be also passed as the first argument to `map`.
+All forms except 1 and 6 can be also passed as the first argument to `map`.
 
 In all of these cases, `function` can return either a single row or multiple rows.
 `function` can always generate a single column by returning a single value or a vector.
@@ -47,7 +47,7 @@ The kind of return value and the number and names of columns must be the same fo
 
 It is allowed to mix single values and vectors if multiple transformations
 are requested. In this case single value will be broadcasted to match the length
-of columns specified by returned vectors. 
+of columns specified by returned vectors.
 As a particular rule, values wrapped in a `Ref` or a `0`-dimensional `AbstractArray`
 are unwrapped and then broadcasted.
 
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -18,13 +18,15 @@ each subset of the `DataFrame`. This specification can be of the following forms
 3. a `col => target_col` pair, which renames the column `col` to `target_col`
 4. a `nrow` or `nrow => target_col` form which efficiently computes the number of rows in a group
    (without `target_col` the new column is called `:nrow`)
+5. standard column selectors (integers, symbols, vectors of integers, vectors of symbols,
+   `All`, `:`, `Between`, `Not` and regular expressions)
 5. several arguments of the forms given above, or vectors thereof
 6. a function which will be called with a `SubDataFrame` corresponding to each group;
    this form should be avoided due to its poor performance unless a very large
    number of columns are processed (in which case `SubDataFrame` avoids excessive
    compilation)
 
-All forms except 6 can be also passed as the first argument to `by`.
+All forms except 6 can be also passed as the first argument to `map`.
 
 In all of these cases, `function` can return either a single row or multiple rows.
 `function` can always generate a single column by returning a single value or a vector.
@@ -42,6 +44,12 @@ Here are the rules specifying the shape of the resulting `DataFrame`:
   of groups is large
 
 The kind of return value and the number and names of columns must be the same for all groups.
+
+It is allowed to mix single values and vectors if multiple transformations
+are requested. In this case single value will be broadcasted to match the length
+of columns specified by returned vectors. In this case as a convinience values stored
+in `Ref` and 0-dimensional arrays are extracted from them.
+
 
 If a single value or a vector is returned by the `function` and `target_col` is not
 provided, it is generated automatically, by concatenating source column name and
@@ -115,6 +123,25 @@ julia> by(iris, :Species,
 │ 1   │ Iris-setosa     │ 0.292449 │ 73.2    │
 │ 2   │ Iris-versicolor │ 0.717655 │ 213.0   │
 │ 3   │ Iris-virginica  │ 0.842744 │ 277.6   │
+
+julia> by(iris, :Species, 1:2, 1:2 .=> mean, nrow)
+150×6 DataFrame
+│ Row │ Species        │ SepalLength │ SepalWidth │ SepalLength_mean │ SepalWidth_mean │ nrow  │
+│     │ String         │ Float64     │ Float64    │ Float64          │ Float64         │ Int64 │
+├─────┼────────────────┼─────────────┼────────────┼──────────────────┼─────────────────┼───────┤
+│ 1   │ Iris-setosa    │ 5.1         │ 3.5        │ 5.006            │ 3.418           │ 50    │
+│ 2   │ Iris-setosa    │ 4.9         │ 3.0        │ 5.006            │ 3.418           │ 50    │
+│ 3   │ Iris-setosa    │ 4.7         │ 3.2        │ 5.006            │ 3.418           │ 50    │
+│ 4   │ Iris-setosa    │ 4.6         │ 3.1        │ 5.006            │ 3.418           │ 50    │
+│ 5   │ Iris-setosa    │ 5.0         │ 3.6        │ 5.006            │ 3.418           │ 50    │
+⋮
+│ 145 │ Iris-virginica │ 6.7         │ 3.3        │ 6.588            │ 2.974           │ 50    │
+│ 146 │ Iris-virginica │ 6.7         │ 3.0        │ 6.588            │ 2.974           │ 50    │
+│ 147 │ Iris-virginica │ 6.3         │ 2.5        │ 6.588            │ 2.974           │ 50    │
+│ 148 │ Iris-virginica │ 6.5         │ 3.0        │ 6.588            │ 2.974           │ 50    │
+│ 149 │ Iris-virginica │ 6.2         │ 3.4        │ 6.588            │ 2.974           │ 50    │
+│ 150 │ Iris-virginica │ 5.9         │ 3.0        │ 6.588            │ 2.974           │ 50    │
+
 ```
 
 The `by` function also supports the `do` block form. However, as noted above,
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -350,9 +350,21 @@ const F_ARGUMENT_RULES =
     `source_cols => fun => target_col`, or `source_col => target_col`.
     Function `fun` is passed `SubArray` views as positional arguments for each column
     specified to be selected and can return a vector or a single value
-    (defined precisely below). As a special case `nrow` or `nrow => target_col` can
-    be passed without specifying input columns to efficiently calculate number of
-    rows in each group. If `nrow` is passed the resulting column name is `:nrow`.
+    (defined precisely below).
+
+    As a special case the following are also allowed:
+    * `nrow` or `nrow => target_col` can be passed without specifying input
+      columns to efficiently calculate number of rows in each group.
+      If `nrow` is passed the resulting column name is `:nrow`.
+    * passing any valid column selector; in such a case an `identity` transformation
+      is applied to a set of passed columns. All rules specified in `select` for
+      this case apply.
+
+    Also if `pairs` are passed then it is allowed to mix single values and vectors
+    as return values of different `fun`s. In this case single value will be
+    broadcasted to match the length of columns specified by returned vectors.
+    In this case as a convinience values stored in `Ref` and 0-dimensional
+    arrays are extracted from them.
 
     If the first or last argument is `pair` then it must be a `Pair` following the
     rules for pairs described above, except that in this case function defined
@@ -484,6 +496,30 @@ julia> combine(gd, :b => :b1, :c => :c1,
 │ 6   │ 2     │ 7     │ 9     │
 │ 7   │ 1     │ 4     │ 5     │
 │ 8   │ 1     │ 8     │ 9     │
+
+julia> combine(gd, :b, :c => sum) # passing columns and broadcasting
+8×3 DataFrame
+│ Row │ a     │ b     │ c_sum │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 2     │ 6     │
+│ 2   │ 1     │ 2     │ 6     │
+│ 3   │ 2     │ 1     │ 8     │
+│ 4   │ 2     │ 1     │ 8     │
+│ 5   │ 3     │ 2     │ 10    │
+│ 6   │ 3     │ 2     │ 10    │
+│ 7   │ 4     │ 1     │ 12    │
+│ 8   │ 4     │ 1     │ 12    │
+
+julia> combine(gd, [:b,:c] .=> Ref)
+4×3 DataFrame
+│ Row │ a     │ b_Ref    │ c_Ref    │
+│     │ Int64 │ SubArra… │ SubArra… │
+├─────┼───────┼──────────┼──────────┤
+│ 1   │ 1     │ [2, 2]   │ [1, 5]   │
+│ 2   │ 2     │ [1, 1]   │ [2, 6]   │
+│ 3   │ 3     │ [2, 2]   │ [3, 7]   │
+│ 4   │ 4     │ [1, 1]   │ [4, 8]   │
 ```
 
 See [`by`](@ref) for more examples.
@@ -517,7 +553,7 @@ end
 
 function combine(gd::GroupedDataFrame,
                  @nospecialize(cs::Union{Pair, AbstractVector{<:Pair}, typeof(nrow),
-                                         AbstractVector{Integer}, AbstractVector{Symbol},
+                                         AbstractVector{<:Integer}, AbstractVector{Symbol},
                                          ColumnIndex, Colon, Regex, Not, All, Between}...);
                  keepkeys::Bool=true)
     @assert !isempty(cs)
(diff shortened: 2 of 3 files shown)


bkamins:
With this `select` and `transform` for `GroupedDataFrame` should be easy (as we have all the functionality - just need to do `groupby` on the result in an efficient way like in `map`).

bkamins:
PR should be good to review.

I have added tests, improved performance a bit and written documentation (I have tried hard, but this is probably the weakest point of the PR :cry:).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
+                                aggj += 1
+                            end
+                            agg2idx_map[j] = aggj
+                        end
+                    end
+                    res[i] = idx, res[i][2][agg2idx_map]
+                else
+                    if idx != res[i][1]
+                        throw(ArgumentError("all functions must return values of the same length"))
+                    end
```
nalimilan, line 1096:
```suggestion
                elseif idx != res[i][1]
                    throw(ArgumentError("all functions must return vectors of the same length"))
                end
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
-    idx = res[1][1]
+    idx_loc = findfirst(x -> x[1] !== idx_agg, res)
+    if isnothing(idx_loc)
+        @assert !isnothing(idx_agg)
+        idx = idx_agg
+    else
+        idx = res[idx_loc][1]
+        agg2idx_map = nothing
+        for i in 1:length(res)
+            if res[i][1] != idx
```
nalimilan, line 1078: Maybe this would we worth it? Surprisingly, `==` doesn't use that optimization.
```suggestion
            if res[i][1] !== idx && res[i][1] != idx
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
+                if res[i][1] === idx_agg
```
nalimilan, line 1079: Why not do this out of the `for` loop instead of doing `agg2idx_map = nothing`? AFAICT you're certain that `res[i][1] === idx_agg` will be true for at least one `i`, right?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
+                    # we perform pseudo broadcasting here
+                    # keep -1 as a sentinel for errors
+                    if isnothing(agg2idx_map)
+                        agg2idx_map = fill(-1, length(idx))
+                        aggj = 1
+                        for j in 1:length(idx)
```
nalimilan, line 1085: Should be worth it here. Also make sure that all types are inferred.
```suggestion
                        @inbounds for j in 1:length(idx)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
+                end
+            end
+        end
+    end
     outcols = map(x -> x[2], res)
+    # this check is redundant given we check idx above now
+    # but it is safer to double check and it is cheap
     if !all(x -> length(x) == length(outcols[1]), outcols)
         throw(ArgumentError("all functions must return values of the same length"))
     end
```
nalimilan, line 1106: Then it's no longer a user error:
```suggestion
    @assert all(x -> length(x) == length(outcols[1]), outcols)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
+                        agg2idx_map = fill(-1, length(idx))
+                        aggj = 1
+                        for j in 1:length(idx)
+                            while idx_agg[aggj] != idx[j]
+                                aggj += 1
+                            end
+                            agg2idx_map[j] = aggj
+                        end
+                    end
+                    res[i] = idx, res[i][2][agg2idx_map]
```
nalimilan, line 1092: Is there a way to `@assert` than `aggj` has the right ending value?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -979,8 +1110,15 @@ end
 function _combine(fun::Base.Callable, gd::GroupedDataFrame, ::Nothing)
     firstres = fun(gd[1])
     firstmulticol = firstres isa MULTI_COLS_TYPE
+    if !(firstres isa Union{AbstractVecOrMat, AbstractDataFrame,
```
nalimilan, line 1113: Would there be a way to avoid repeating this in each method?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1017,7 +1162,9 @@ end
 function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractDataFrame},
                              f::Any, gd::GroupedDataFrame,
                              incols::Union{Nothing, AbstractVector, Tuple},
-                             firstmulticol::Val)
+                             firstmulticol::Val, idx_agg)
+    extrude_case = false
```
nalimilan, line 1166: Why "case"?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1340,6 +1493,30 @@ julia> by(df, :a, :b => :b1, :c => :c1,
...
+│ 1   │ 1     │ 2     │ 6     │
+│ 2   │ 1     │ 2     │ 6     │
+│ 3   │ 2     │ 1     │ 8     │
+│ 4   │ 2     │ 1     │ 8     │
+│ 5   │ 3     │ 2     │ 10    │
+│ 6   │ 3     │ 2     │ 10    │
+│ 7   │ 4     │ 1     │ 12    │
+│ 8   │ 4     │ 1     │ 12    │
+
+julia> by(df, :a, [:b,:c] .=> Ref)
```
nalimilan, line 1511:
```suggestion
julia> by(df, :a, [:b, :c] .=> Ref)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1027,26 +1174,35 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
     elseif first isa DataFrameRow
         n = length(gd)
         eltys = [eltype(parent(first)[!, i]) for i in parentcols(index(first))]
-    else # NamedTuple giving a single row
+    elseif firstmulticol == Val(false) && first.x1 isa Union{AbstractArray{<:Any, 0}, Ref}
+        extrude_case = true
+        first = wrap_row(first.x1, firstmulticol)
+        n = length(gd)
+        eltys = (typeof(first.x1),)
```
nalimilan, line 1181: I'd use `first[1]` instead of `first.x1`, as the name is quite arbitrary.
```suggestion
    elseif firstmulticol == Val(false) && first[1] isa Union{AbstractArray{<:Any, 0}, Ref}
        extrude_case = true
        first = wrap_row(first[1], firstmulticol)
        n = length(gd)
        eltys = (typeof(first[1]),)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
+            end
             idx, outcols, _ = _combine_with_first(wrap(firstres), fun, gd, incols,
-                                                  Val(firstmulticol))
+                                                  Val(firstmulticol),
+                                                  firstres isa AbstractVector ? nothing : idx_agg)
+            @assert length(outcols) == 1
             res[i] = idx, outcols[1]
         end
     end
-    # TODO: avoid recomputing idx for each pair when !all(isagg, f)
```
nalimilan, line 970: This one is still valid I think (the redundant computation happens in `_combine_with_first`, so you could move it there).

docs/src/man/split_apply_combine.md
```diff
@@ -18,13 +18,15 @@ each subset of the `DataFrame`. This specification can be of the following forms
...
+5. standard column selectors (integers, symbols, vectors of integers, vectors of symbols,
+   `All`, `:`, `Between`, `Not` and regular expressions)
 5. several arguments of the forms given above, or vectors thereof
 6. a function which will be called with a `SubDataFrame` corresponding to each group;
    this form should be avoided due to its poor performance unless a very large
    number of columns are processed (in which case `SubDataFrame` avoids excessive
    compilation)
 
-All forms except 6 can be also passed as the first argument to `by`.
+All forms except 6 can be also passed as the first argument to `map`.
```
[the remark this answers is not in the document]
bkamins: `map` does not support vectors of `Pair`s. I will correct the numbering.

docs/src/man/split_apply_combine.md
```diff
@@ -18,13 +18,15 @@ each subset of the `DataFrame`. This specification can be of the following forms
 3. a `col => target_col` pair, which renames the column `col` to `target_col`
 4. a `nrow` or `nrow => target_col` form which efficiently computes the number of rows in a group
    (without `target_col` the new column is called `:nrow`)
+5. standard column selectors (integers, symbols, vectors of integers, vectors of symbols,
+   `All`, `:`, `Between`, `Not` and regular expressions)
 5. several arguments of the forms given above, or vectors thereof
```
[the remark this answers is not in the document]
bkamins: OK

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -350,9 +350,21 @@ const F_ARGUMENT_RULES =
...
-    (defined precisely below). As a special case `nrow` or `nrow => target_col` can
-    be passed without specifying input columns to efficiently calculate number of
-    rows in each group. If `nrow` is passed the resulting column name is `:nrow`.
+    (defined precisely below).
+
+    As a special case the following are also allowed:
+    * `nrow` or `nrow => target_col` can be passed without specifying input
+      columns to efficiently calculate number of rows in each group.
+      If `nrow` is passed the resulting column name is `:nrow`.
+    * passing any valid column selector; in such a case an `identity` transformation
```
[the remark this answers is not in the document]
bkamins: OK - I leave the comment unresolved as probably it would be good if you have a look how I corrected it. Thank you.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -540,7 +577,55 @@ function combine(gd::GroupedDataFrame,
...
+            @assert v isa AbstractVector{Int}
+            process_vectors = true
+        end
+    end
+    processed_cols = Set{Symbol}()
+    if process_vectors
+        cs_norm = Pair[]
+        i = 0
+        while i < length(cs_norm_pre)
+            i += 1
```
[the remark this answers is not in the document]
bkamins: OK. I felt that with `while` I have more control here, but as we delete only in "unvisited" part this is the same.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -585,50 +670,58 @@ wrap(x::AbstractMatrix) =
...
+    return wrap(x)
+end
+
+function wrap_row(x::Union{AbstractArray{<:Any, 0}, Ref},
+                  ::Val{firstmulticol}) where firstmulticol
+    if firstmulticol
+        throw(ArgumentError(ERROR_COL_COUNT))
     end
-    wrap(x)
+    return (x1 = x[],)
```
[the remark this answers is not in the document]
bkamins: This is the point that it cannot (this was my initial implementation), but then `Ref([1,2,3])` gets unwrapped to a vector and in consequence we think it is a multi-row case of vectors holding `Int`s and not a single-row case where each row holds a vector (which is a correct way to handle this).

Actually this is a key functionality we provide here - it was requested on discourse - to allow for `Ref([1,2,3])` instead of `[[1,2,3]]` because the latter:
1. allocates
2. is slower
3. does not check if only one row per group is produced

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
                 throw(ArgumentError("a single value or vector result is required when passing " *
                                     "multiple functions (got $(typeof(res)))"))
             end
+            if !(firstres isa AbstractVector) && isnothing(idx_agg)
```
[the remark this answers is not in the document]
bkamins: No - exactly the opposite. This is to make sure we compute `idx_agg` exactly once.

In short we say if `firstres` is a vector do not compute `idx_agg` or do not compute it if it was already computed (as `idx_agg` is identical for each function returning single-row-per-group).

Later we use `=== idx_agg` to identify columns that need to be pseudo-broadcasted (because they were produced by single-row-per-group return value type).

I leave this unresolved as this is a crucial part of the change I have implemented so it is crucial that we are on the same page here.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1027,26 +1174,35 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
...
         if any(x -> x <: AbstractVector, eltys)
             throw(ArgumentError("mixing single values and vectors in a named tuple is not allowed"))
         end
     end
-    idx = Vector{Int}(undef, n)
+    if isnothing(idx_agg)
+        idx = Vector{Int}(undef, n)
+    else
+        idx = idx_agg
+    end
```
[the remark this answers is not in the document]
bkamins: This function is going to be recompiled every time anyway, because of changes in `first` type and `f` type (i.e. different type of return value and different type of function will cause recompilation). We decided to leave `@nospecialize` for later, so I would leave it as is for now here.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,14 +1056,51 @@ function _combine(f::AbstractVector{<:Pair},
...
-                                                  Val(firstmulticol))
+                                                  Val(firstmulticol),
+                                                  firstres isa AbstractVector ? nothing : idx_agg)
+            @assert length(outcols) == 1
             res[i] = idx, outcols[1]
         end
     end
-    # TODO: avoid recomputing idx for each pair when !all(isagg, f)
-    idx = res[1][1]
+    idx_loc = findfirst(x -> x[1] !== idx_agg, res)
```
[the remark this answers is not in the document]
bkamins: Well - it is OK. If `idx_agg === nothing` then `idx_loc=1` (`res` contains elements whose first entry must be a vector and it is non-empty).

I will add a comment.

src/groupeddataframe/splitapplycombine.jl
bkamins: Yes - I think that I even used this trick somewhere in the code, but forgotten to add it here. Probably a PR to Base would be in place.

src/groupeddataframe/splitapplycombine.jl
bkamins: No, `idx_agg` will be `nothing` if no columns were of one-row type (i.e. all transformation functions returned vectors). I use `nothing` for `agg2idx_map` as a sentinel that no column of this type was seen yet. If it is encountered for the first time I populate it.

I will leave this open as this is a second crucial part of the PR (i.e. that I recompute `idx_agg` and `agg2idx_map` at most once and only if at least one function producing single rows were passed by the user).

src/groupeddataframe/splitapplycombine.jl
bkamins: Good point. I will also move it to a barrier function - just to be sure.

src/groupeddataframe/splitapplycombine.jl
bkamins: Right :smile:

src/groupeddataframe/splitapplycombine.jl
bkamins: Now I will add such a check if you want to do `@inbounds`. In general `aggj` can end up smaller than `length(idx_agg)` because `idx` might have dropped groups at the end (and then we quash these values in `idx_agg` from 1 to 0 rows).

src/groupeddataframe/splitapplycombine.jl
bkamins: No - this is a part of optimization you wanted (in the earlier comment as TODO for performance): we do not want to calculate `idx_agg` each time as single-row function is encountered, so we have to compute it before we call `_combine_with_first` exactly once if it is needed to be computed.

src/groupeddataframe/splitapplycombine.jl
bkamins: And what would you suggest. We have to keep track of two cases:
1. Case when we extrude value from `Ref` or 0-dimensional array
2. Case when we do not extrude it

Later we have to remember it because `first` can be a vector and still this is a single-row case (because we extruded this vector from e.g. `Ref`).

I leave it open as maybe you have a better name for it.

src/groupeddataframe/splitapplycombine.jl
bkamins: 1. if functions return a single row we compute `idx_agg` only once
2. for all functions returning vectors each of them potentially can return a different `idx` and only later we can find if they have the same lengths or not.

What could be optimized is that for functions returning vectors we could work like this:
1. for a first function like this compute `idx` as is now in `_combine_tables_with_first`
2. for any other function have a different `_combine_tables_with_first` implementation that takes the `idx` vector computed in step 1 and instead of creating it only check if it would be consistent with what `idx` holds.

I will add such a comment but leave it as a future optimization.

bkamins: Just to be clear - we currently avoid multiple allocation of `idx` if single-row columns are given (and this is then `idx_agg`). I kept `idx_agg` name as single-row columns are essentially aggregations of groups.

I leave it unresolved as again - this is a crucial concept of this PR.

bkamins:
Comments applied. Just to show that we avoid recomputing `idx_agg` when not needed.

Preparation:
```
df = DataFrame(g = 10^6:-1:1)
```

In this PR:
```
julia> @btime by(df, :g, :g => first => :x1, :g => first => :x2, :g => first => :x3, :g => first => :x4, :g => first => :x5, :g => first => :x6);
  163.177 ms (365 allocations: 182.54 MiB)
```
on 0.20.2:
```
julia> @btime by(df, :g, x1 = :g => first, x2 = :g => first, x3 = :g => first, x4 = :g => first, x5 = :g => first, x6 = :g => first);
  186.147 ms (417 allocations: 236.90 MiB)
```
(note the differences in allocations due to the fact that we allocate `idx_agg` only once now)

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -926,11 +1024,25 @@ isagg(p::Pair) = check_aggregate(last(p)) isa AbstractAggregate && first(p) isa
 
 const MULTI_COLS_TYPE = Union{AbstractDataFrame, NamedTuple, DataFrameRow, AbstractMatrix}
 
+function _agg2idx_map_helper(idx, idx_agg)
+    agg2idx_map = fill(-1, length(idx))
+    aggj = 1
+    @inbounds for (j, idxj) in enumerate(idx)
+        while idx_agg[aggj] != idxj
+            aggj += 1
+            @assert aggj <= length(idx_agg)
```
nalimilan, line 1033: I was thinking about a single check at the end of the loops. This one is going to be quite costly, right?

src/groupeddataframe/splitapplycombine.jl
nalimilan: Sorry, my question was about whether it would be possible to reorganize the code to reduce duplication (not about performance).

src/groupeddataframe/splitapplycombine.jl
nalimilan: I think I'd simply have said "extrude".

nalimilan:
> Well - we could "pseudo-broadcast" the contents of `NamedTuple`, that is if someone passes `(a=1, b=[1,2,3])` to expand it to `(a=[1,1,1], b=[1,2,3])`, note that this happens in `DataFrame` if you write `DataFrame(a=1, b=[1,2,3])`. But I came to the conclusion that it is cleaner to treat `NamedTuple` "as is" and directly use what the user has passed (as `DataFrame` is only constructed using `DataFrame(a=1, b=[1,2,3])` but it gets expanded inside a function not within `combine`).

Actually I don't see in what cases this would make a difference. Currently things like `by(df, :x1, :x2 => (x -> (a=[1, 2], b=3)))` throw an error, right? With `ByRow` it's allowed, but it would make no sense to broadcast as the named tuple isn't destructured into columns. Is that what you're referring to?

bkamins:
> Currently things like `by(df, :x1, :x2 => (x -> (a=[1, 2], b=3)))` throw an error, right?

Right. And it would keep to be an error.

However also `by(df, :x2 => (x -> (a=[1, 2], b=3)))` throws an error, but `by(df, :x2 => (x -> DataFrame(a=[1, 2], b=3)))` does not (and conceptually it is almost the same).

The issue is that `DataFrame` constructor does pseudo-broadcasting while, of course, `NamedTuple` constructor does not do it. My question is if we should do it after `NamedTuple` is constructed (currently we do not and throw an error, and I think this is a right approach, but maybe you would disagree so I prefer to ask).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -372,7 +389,7 @@ const KWARG_PROCESSING_RULES =
     """
 
 """
-    combine(gd::GroupedDataFrame, pairs...; keepkeys::Bool=true)
+    combine(gd::GroupedDataFrame, args...; keepkeys::Bool=true)
     combine(fun::Union{Function, Type}, gd::GroupedDataFrame; keepkeys::Bool=true)
     combine(pair::Pair, gd::GroupedDataFrame; keepkeys::Bool=true)
```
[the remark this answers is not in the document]
bkamins: we allow this:
1. to be less breaking (we allowed this in the past)
2. to be consistent with `map` (where we allow it as otherwise there would be no way to do "fast path" in `map` as it requires a `Pair` argument)

src/groupeddataframe/splitapplycombine.jl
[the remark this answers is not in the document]
bkamins: added

src/groupeddataframe/splitapplycombine.jl
bkamins: I would not do it. This loop is cheap relative to all other work we do in combine - even for very large tables it is just a single pass in a loop. Note that it is only executed if vectors and single values are mixed (so we do a lot of more expensive operations anyway).

Now what is a consequence of avoiding this check? the `while idx_agg[aggj] != idxj` loop might go to infinity and/or produce a segfault. I think the benefit is not worth it. I would either leave this assertion or drop `@inbounds`.

src/groupeddataframe/splitapplycombine.jl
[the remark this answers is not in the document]
nalimilan: OK, consistency with `map` is convincing. We could allow passing multiple pairs too, but that can be done later if needed.

nalimilan:
OK, let's keep throwing an error for now, that way we're safe to change it later if we want. If we allowed named tuples mixing scalars and vectors, it would indeed make sense to also broadcast in that case.

src/groupeddataframe/splitapplycombine.jl
nalimilan: OK, as you like.

bkamins: I have benchmarked it. The difference in speed with `@assert` is not noticeable:
```
julia> x = 1:10^6;

julia> function f1(x, l)
           s = 0.0
           i = 1
           while i < l
               s += sin(i) # do something that will not be optimized out
               @assert i <= length(x)
               i += 1
           end
           s
       end;

julia> function f2(x, l)
           s = 0.0
           i = 1
           while i < l
               s += sin(i) # do something that will not be optimized out
               i += 1
           end
           s
       end;

julia> @benchmark f1($x, 10^6)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     16.509 ms (0.00% GC)
  median time:      16.773 ms (0.00% GC)
  mean time:        16.828 ms (0.00% GC)
  maximum time:     24.072 ms (0.00% GC)
  --------------
  samples:          298
  evals/sample:     1

julia> @benchmark f2($x, 10^6)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     16.465 ms (0.00% GC)
  median time:      16.782 ms (0.00% GC)
  mean time:        16.839 ms (0.00% GC)
  maximum time:     23.997 ms (0.00% GC)
  --------------
  samples:          297
  evals/sample:     1
```

src/groupeddataframe/splitapplycombine.jl
bkamins: OK - changed.

src/groupeddataframe/splitapplycombine.jl
bkamins: OK - I have added an intermediate function that captures duplicate code (essentially it handles combining when returning multiple columns is allowed).

src/groupeddataframe/splitapplycombine.jl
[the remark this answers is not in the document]
bkamins: I disallowed this as conceptually `map` maps a single function.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1171,11 +1159,25 @@ function _combine(p::Pair, gd::GroupedDataFrame, ::Nothing)
     return idx, valscat
 end
 
+function _combine_multicol(firstres, fun::Any, gd::GroupedDataFrame,
+                        incols::Union{Nothing, AbstractVector, Tuple})
```
nalimilan, line 1163:
```suggestion
                           incols::Union{Nothing, AbstractVector, Tuple})
```

bkamins:
Thank you!

PR #2174, opened 2020-04-06
[BREAKING] Deprecate aggregate

bkamins:
in the future use of `select` and `combine` is to be preferred (also as `aggregate` was currently inconsistent).

@nalimilan - but please have a look at deprecation code to see what is required to make it exactly the same (most of the time much simpler code would be enough, but we want 100% consistency).


bkamins:
@nalimilan - only coverage fails.

docs/src/man/split_apply_combine.md
```diff
@@ -133,28 +133,6 @@ julia> by(iris, :Species) do df
 │ 3   │ Iris-virginica  │ 5.552   │ 0.304588  │
 ```
 
-A second approach to the Split-Apply-Combine strategy is implemented in the `aggregate` function, which also takes three arguments: (1) a DataFrame, (2) one or more columns to split the DataFrame on, and (3) one or more functions that are used to compute a summary of each subset of the DataFrame. Each function is applied to each column that was not used to split the DataFrame, creating new columns of the form `$name_$function` like with `by` (see above). We show several examples of the `aggregate` function applied to the `iris` dataset below:
```
nalimilan, line 136: Maybe show how to apply a function to each non-grouping column using `by`?

docs/src/man/getting_started.md
```diff
@@ -699,36 +699,6 @@ julia> mean(df.A)
 2.5
 ```
 
-### Column-Wise Operations
```
nalimilan, line 702: Why not show how to get the same result with `select`? Though this probably doesn't need a separate section heading as it can be seen as a complement to `describe`.

src/deprecated.jl
```diff
@@ -357,3 +357,32 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
 end
 
 @deprecate groupvars(gd::GroupedDataFrame) groupcols(gd)
+
+@deprecate aggregate(d::AbstractDataFrame, fs::Any; sort::Bool=false) begin
+    df = select(d, names(d) .=> [fs])
```
nalimilan, line 362: This seems to be enough:
```suggestion
    df = select(d, names(d) .=> fs)
```

src/deprecated.jl
```diff
@@ -357,3 +357,32 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
 end
 
 @deprecate groupvars(gd::GroupedDataFrame) groupcols(gd)
+
+@deprecate aggregate(d::AbstractDataFrame, fs::Any; sort::Bool=false) begin
```
nalimilan, line 361: To make the deprecation a bit simpler, it would be nice to branch on `sort` and use `depwarn` to print a deprecation warning which doesn't use `sort` when it's not provided (using `sort=nothing` by default).

src/deprecated.jl
```diff
@@ -357,3 +357,32 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+end
+
+@deprecate aggregate(d::AbstractDataFrame, fs::AbstractVector; sort::Bool=false) begin
+    df = hcat([select(d, names(d) .=> [f]) for f in fs]..., makeunique=true)
+    sort && sort!(df)
+    return df
+end
+
+@deprecate aggregate(gd::GroupedDataFrame, f::Any; sort::Bool=false) begin
+    df = combine(gd, valuecols(gd) .=> [f])
```
nalimilan, line 374:
```suggestion
    df = combine(gd, valuecols(gd) .=> f)
```

src/deprecated.jl
```diff
@@ -357,3 +357,32 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
 @deprecate groupvars(gd::GroupedDataFrame) groupcols(gd)
+
+@deprecate aggregate(d::AbstractDataFrame, fs::Any; sort::Bool=false) begin
+    df = select(d, names(d) .=> [fs])
+    sort && sort!(df)
+    return df
+end
+
+@deprecate aggregate(d::AbstractDataFrame, fs::AbstractVector; sort::Bool=false) begin
+    df = hcat([select(d, names(d) .=> [f]) for f in fs]..., makeunique=true)
```
nalimilan, line 368: Could we check whether there are functions with duplicate names in `fs`? If not, then `select(d, names(d) .=> fs)` is enough right? Same for the GDF method.

src/deprecated.jl
```diff
@@ -357,3 +357,32 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+
+@deprecate aggregate(gd::GroupedDataFrame, fs::AbstractVector; sort::Bool=false) begin
+    df = hcat([combine(gd, valuecols(gd) .=> [f], keepkeys=i==1) for (i, f) in enumerate(fs)]...,
+                       makeunique=true)
+    sort && sort!(df, setdiff(names(df), groupcols(gd)))
+    return df
+end
+
+@deprecate aggregate(d::AbstractDataFrame, cols, fs::Any;
+          sort::Bool=false, skipmissing::Bool=false) aggregate(groupby(d, cols, sort=sort,
```
nalimilan, line 387: Better show the final syntax, otherwise people will get two deprecation warnings, and one is already scary enough. :-)

src/deprecated.jl
bkamins: It would not be equivalent to what `aggregate` did in the past (but maybe it is not a problem - the differences will be apparent only in very obscure cases when some custom broadcasting is defined for `fs`, or it s). So - given this - what do you think - leave what I have written (which is safe) or what you propose (which is cleaner but sometimes might fail).

bkamins: Actually I rewrite it anyway. So I resolve it.

src/deprecated.jl
bkamins: OK - I made two branches.

src/deprecated.jl
bkamins: The same - I have rewritten it.

src/deprecated.jl
bkamins: right - I will rewrite it.

src/deprecated.jl
bkamins: OK

src/deprecated.jl
bkamins: ah - it is not right, but I will write what is right.

docs/src/man/split_apply_combine.md
bkamins: added

docs/src/man/getting_started.md
bkamins: added

bkamins:
I have fixed all. The way I did it is to print a "nice" warning that is not 100% correct, but it is incorrect only when duplicate column names are produced (which is noted in the deprecation comment).

docs/src/man/split_apply_combine.md
```diff
@@ -269,3 +247,40 @@ Last Group (5 rows): g = 501
 │ 4   │ 501   │ 2504  │
 │ 5   │ 501   │ 2505  │
 ```
+
+In order to apply a function to each non-grouping column ofr a `GroupedDataFrame` you can write:
+```jldoctest sac
+julia> gd = groupby(iris, :Species);
+
+julia> combine(gd, valuecols(gd) .=> mean)
```
nalimilan, line 255: Why not show this using `by`, and put it above, at the same place where `aggregate` was shown?

src/deprecated.jl
```diff
@@ -357,3 +357,85 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+                     "in `fs` have unique names.", :aggregate)
+    end
+    return df
+end
+
+function aggregate(gd::GroupedDataFrame, f::Any; sort::Bool=false)
+    df = combine(gd, valuecols(gd) .=> [f])
+    if sort
+        Base.depwarn("`aggregate` is deprecated. Instead use" *
+                     "sort!(combine(gd, valuecols(gd) .=> f), setdiff(names(df), groupcols(gd)))`.",
```
nalimilan, line 395: `df` doesn't exist for the user. Maybe break this into two steps separated by `;`.

Also:
```suggestion
                     " `sort!(combine(gd, valuecols(gd) .=> f), setdiff(names(df), groupcols(gd)))`.",
```

src/deprecated.jl
```diff
@@ -357,3 +357,85 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+                     :aggregate)
+        sort!(df, setdiff(names(df), groupcols(gd)))
+    else
+        Base.depwarn("`aggregate` is deprecated. Instead use" *
+                     "combine(gd, [valuecols(gd) .=> f for f in fs]...)`",
```
nalimilan, line 400:
```suggestion
                     " `combine(gd, [valuecols(gd) .=> f for f in fs]...)`",
```

src/deprecated.jl
```diff
@@ -357,3 +357,85 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+    return df
+end
+
+function aggregate(gd::GroupedDataFrame, fs::AbstractVector; sort::Bool=false)
+    df = hcat([combine(gd, valuecols(gd) .=> [f], keepkeys=i==1) for (i, f) in enumerate(fs)]...,
+                       makeunique=true)
+    if sort
+        Base.depwarn("`aggregate` is deprecated. Instead" *
+                     " use `sort!(combine(gd, [names(gd) .=> f for f in fs]...), `" *
+                     "setdiff(names(df), groupcols(gd))) if functions in `fs` have unique names.", :aggregate)
```
nalimilan, line 412: Same here.

docs/src/man/split_apply_combine.md
bkamins: I will add it in https://github.com/JuliaData/DataFrames.jl/pull/2177, as without that PR it is not very easy to do.

bkamins: But I will keep it here as it shows how `valuecols` can be used.

src/deprecated.jl
bkamins: fixed

src/deprecated.jl
```diff
@@ -357,3 +357,87 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+    df = combine(gd, valuecols(gd) .=> [f])
+    if sort
+        Base.depwarn("`aggregate` is deprecated. Instead use" *
+                     " `df = combine(gd, valuecols(gd) .=> f); " *
+                     "sort!(df, setdiff(names(df), groupcols(gd)))`.",
+                     :aggregate)
+        sort!(df, setdiff(names(df), groupcols(gd)))
+    else
+        Base.depwarn("`aggregate` is deprecated. Instead use" *
+                     " `combine(gd, [valuecols(gd) .=> f for f in fs]...)`",
```
nalimilan, line 401: Maybe this would be more explicit? People might not understand what `fs` refers to. Same elsewhere.
```suggestion
        Base.depwarn("`aggregate(gd, fs, sort=$sort)` is deprecated. Instead use" *
                     " `combine(gd, [valuecols(gd) .=> f for f in fs]...)`",
```

src/deprecated.jl
```diff
@@ -391,13 +391,13 @@ end
...
-        Base.depwarn("`aggregate` is deprecated. Instead use" *
+        Base.depwarn("`aggregate(gd, f, sort=true)` is deprecated. Instead use" *
                      " `df = combine(gd, valuecols(gd) .=> f); " *
                      "sort!(df, setdiff(names(df), groupcols(gd)))`.",
                      :aggregate)
         sort!(df, setdiff(names(df), groupcols(gd)))
     else
-        Base.depwarn("`aggregate(gd, fs, sort=$sort)` is deprecated. Instead use" *
+        Base.depwarn("`aggregate(gd, f)` is deprecated. Instead use" *
                      " `combine(gd, [valuecols(gd) .=> f for f in fs]...)`",
```
nalimilan, line 401:
```suggestion
                     " `combine(gd, valuecols(gd) .=> f)`",
```

bkamins: I thought I have commented on this already - sorry for missing this.
What you propose is not correct. We have to apply all `valuecols(gd)` to all functions in `fn` (in other words - we need to do a Cartesian product). The alternative to what I have written would be e.g.:
```
(hcat(valuecols(gd)...) .=> fs)...
```
but I am not sure how many users would understand what is going on here.

Also we could write (this will be without `...`):
```
[n => f for f in fs for n in valuecols(df)]
```

So - given the options what would you prefer?

bkamins:
I have merged this with master and used the new `names` functionality.

src/deprecated.jl
nalimilan: But `fs` doesn't exist in that method?

src/deprecated.jl
```diff
@@ -357,3 +357,89 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+    end
+    return df
+end
+
+function aggregate(gd::GroupedDataFrame, f::Any; sort::Bool=false)
+    df = combine(gd, valuecols(gd) .=> [f])
+    if sort
+        Base.depwarn("`aggregate(gd, f, sort=true)` is deprecated. Instead use" *
+                     " `df = combine(gd, valuecols(gd) .=> f); " *
+                     "sort!(df, names(df, Not(groupcols(gd))))`.",
```
nalimilan, line 396: Doesn't this work thanks to #2181?
```suggestion
                     "sort!(df, Not(groupcols(gd)))`.",
```

src/deprecated.jl
```diff
@@ -357,3 +357,89 @@ function by(d::AbstractDataFrame, cols::Any; sort::Bool=false, skipmissing::Bool
...
+
+function aggregate(gd::GroupedDataFrame, fs::AbstractVector; sort::Bool=false)
+    df = hcat([combine(gd, valuecols(gd) .=> [f], keepkeys=i==1) for (i, f) in enumerate(fs)]...,
+                       makeunique=true)
+    if sort
+        Base.depwarn("`aggregate(gd, fs, sort=true)` is deprecated. Instead use " *
+                     "`df = combine(gd, [names(gd) .=> f for f in fs]...); " *
+                     "sort!(df, names(df, Not(groupcols(gd))))`" *
+                     " if functions in `fs` have unique names.", :aggregate)
+        sort!(df, names(df, Not(groupcols(gd))))
```
nalimilan, line 415:
```suggestion
        sort!(df, Not(groupcols(gd)))
```

src/deprecated.jl
bkamins: right - nice catch (actually I wanted to do this this way initially :smile: )

src/deprecated.jl
bkamins: wow - you are right. There are so many deprecated methods. Thank you for being so precise.

bkamins:
Thank you!

bkamins:
Thank you!

PR #2176, opened 2020-04-06
fix deprecation in append

bkamins:
correct use of `eachcol` in `append!`


bkamins:
@nalimilan I merged this PR with master as it is very simple so it is a good place to check if all works correctly everywhere after merging the "big" PRs.

bkamins:
Thank you!

PR #2177, opened 2020-04-06
add cols to names

bkamins:
allow `cols` argument to `names` so that it is easy to use any valid selector in `select` and `combine` when broadcasting.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -69,7 +69,7 @@ abstract type AbstractDataFrame end
 
     If `cols` column selector argument is passed then restrict returned
     column names to those matching the selector
-    (this is useful with regular expressions, `All`, `Not`, ad `Between`).
+    (this is useful with regular expressions, `All`, `Not`, and `Between`).
 """
 Base.names(df::AbstractDataFrame) = names(index(df))
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -67,9 +67,9 @@ abstract type AbstractDataFrame end
 
     Return a `Vector{Symbol}` of names of columns contained in `df`.
 
-    If `cols` column selector argument is passed then restrict returned
+    If a `cols` column selector is passed then restrict returned
     column names to those matching the selector
-    (this is useful with regular expressions, `All`, `Not`, and `Between`).
+    (this is useful in particular with regular expressions, `Not`, and `Between`).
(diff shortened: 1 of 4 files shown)


bkamins:
This removes the need (at least temporaty) for broadcasting of `All`, `Not` and `Between`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -63,10 +63,21 @@ abstract type AbstractDataFrame end
 
 """
     names(df::AbstractDataFrame)
+    names(df::AbstractDataFrame, cols)
 
     Return a `Vector{Symbol}` of names of columns contained in `df`.
+
+    If `cols` column selector argument is passed then restrict returned
+    column names to those matching the selector
+    (this is useful with regular expressions, `All`, `Not`, ad `Between`).
```
jonas-schulze, line 72:
```suggestion
    (this is useful with regular expressions, `All`, `Not`, and `Between`).
```

bkamins:
@jonas-schulze - thank you!

bkamins:
TODO: add https://github.com/JuliaData/DataFrames.jl/pull/2174/files#r406131188 (whatever is merged first the other should add the missing example)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -63,10 +63,21 @@ abstract type AbstractDataFrame end
 
 """
     names(df::AbstractDataFrame)
+    names(df::AbstractDataFrame, cols)
 
     Return a `Vector{Symbol}` of names of columns contained in `df`.
+
+    If `cols` column selector argument is passed then restrict returned
```
nalimilan, line 70:
```suggestion
    If a `cols` column selector is passed then restrict returned
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -63,10 +63,21 @@ abstract type AbstractDataFrame end
...
+    column names to those matching the selector
+    (this is useful with regular expressions, `All`, `Not`, and `Between`).
```
nalimilan, line 72:
```suggestion
    (this is useful in particular with regular expressions, `Not`, and `Between`).
```
(`All` doesn't sound too useful, and passing indices could be useful in some cases)

Also, don't we generally repeat the list of all possible selectors in all docstrings?

bkamins: OK - I will fix this

bkamins:
Thank you!

Issue #2121, opened 2020-02-17
Add a wrapper type for passing named tuples to functions when transforming


bkamins:
See:
https://github.com/JuliaData/DataFrames.jl/issues/1935#issuecomment-586967550
for a dicsussion.

This applies to: `combine`, `by`, `select`, `transform` and `filter`.

dgkf:
Continuing the discussion from #1935, it's still ambiguous to me how this interacts with broadcasted `selection .=> tranform` pairs.

Consider a `DataFrame` with an `:age` column and a large number of measurement columns (for example, let's say `[:x, :y, :z]`). Perhaps you want to see what the correlation is between age and each other column.

```julia
# knowing what they're called in advance, you can do
Splat.(:age, [:x, :y, :z]) .=> cor

# if instead, we don't know what the measurement columns are 
# named, we might use a `Regex` or `Not`
Splat.(:age, Not(:age)) .=> cor
```

This raises some ambiguity. Syntactically, you can't broadcast `Splat` (without a specialized method) because `Not(:age)` doesn't have a length and doesn't yet know what columns it's selecting. Similar ambiguity arises if you're trying to work with pairwise combinations of columns. 

I don't think all of these necessarily need to be accommodated, as such a niche use case might not warrant such syntactic shorthand, but raise it for consideration about where to draw this line.

bkamins:
We could add broadcasting to `Splat` in the future, but I do not find it a crucial functionality for now. In such complex cases I think it is cleaner to just write:
```
by(df, :col) do sdf
    whatever_you_need_to_do_with_your_sdf
end
```

bkamins:
We have decided to auto-splat by default (to lower compilation cost).
The actions are the following:
1. in this issue we should decide what wrapper we want to use for the old behavior (i.e. to make a function pass a `NamedTuple`), my initial idea is `NT`, but it does not seem very nice, maybe `Table` is a better name.
2. I will update https://github.com/JuliaData/DataFrames.jl/pull/2091 where I will propose the API for selecting columns for auto-splatting (the difference is that we will allow duplicates in auto-splatting, which was not allowed for `NamedTuple` API)
3. Then I will update https://github.com/JuliaData/DataFrames.jl/pull/2080
4. Then we have to clean up `by`/`combine` API (this is the hard part, as we have to go through deprecation period here)

bkamins:
In order to resolve this we need to:
[...]

PR #2183, opened 2020-04-11
add AsTable wrapper, disallow NamedTuple in ByRow

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2121

I still need to write tests and update manual and documentation


bkamins:
@nalimilan - the PR should be good to review. Thank you!

bkamins:
Traditionally only coveralls fail

docs/src/man/getting_started.md
```diff
@@ -634,7 +634,7 @@ julia> df
 ```
 
 `transform` and `transform!` functions work identically to `select` and `select!` with the only difference that
-they retain all columns that are present in the source data frame, for example:
+they retain all columns that are present in the source data frame. Here are some more advanced examples:
```
pdeffebach, line 637: Can we add a `rowmean`-esque example in these docs?

bkamins: added - it is showing the ability to infer collection type even if all elements are `missing` as you wanted.

bkamins: However, this functionality is not for Julia 1.0 users :(.

docs/src/man/split_apply_combine.md
```diff
@@ -28,6 +28,12 @@ each subset of the `DataFrame`. This specification can be of the following forms
 
 All forms except 1 and 6 can be also passed as the first argument to `map`.
 
+Three are two special rules that apply to `cols => function` syntax:
+1. if `cols` is wrapped in `AsTable` object then a `NamedTuple` containing columns
+   selected by `cols` is passed to `function`
+2. if `function` is wrapped in a `ByRow` object, then it will be passed values from single
```
nalimilan, line 34: Is it really worth mentioning `ByRow` for `by`? It's rarely useful as it completely ignores grouping. It's good to support it for consistency, but having it in the docstring sounds enough. Then `AsTable` can be mentioned in point 2, when the kind of arguments the function takes is mentioned.

docs/src/man/split_apply_combine.md
```diff
@@ -60,31 +66,27 @@ We show several examples of the `by` function applied to the `iris` dataset belo
...
-│ Row │ SepalLength │ SepalWidth │ PetalLength │ PetalWidth │ Species        │ id    │
-│     │ Float64     │ Float64    │ Float64     │ Float64    │ String         │ Int64 │
-├─────┼─────────────┼────────────┼─────────────┼────────────┼────────────────┼───────┤
-│ 1   │ 6.7         │ 3.3        │ 5.7         │ 2.5        │ Iris-virginica │ 145   │
-│ 2   │ 6.7         │ 3.0        │ 5.2         │ 2.3        │ Iris-virginica │ 146   │
-│ 3   │ 6.3         │ 2.5        │ 5.0         │ 1.9        │ Iris-virginica │ 147   │
-│ 4   │ 6.5         │ 3.0        │ 5.2         │ 2.0        │ Iris-virginica │ 148   │
-│ 5   │ 6.2         │ 3.4        │ 5.4         │ 2.3        │ Iris-virginica │ 149   │
-│ 6   │ 5.9         │ 3.0        │ 5.1         │ 1.8        │ Iris-virginica │ 150   │
+julia> iris = DataFrame(CSV.File(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv")))
```
nalimilan, line 69: I think we used `first` and `last` because we couldn't find a way to reproduce the default output in doctests. Unless you have an idea about how to fix this, probably better not make it even harder to enable doctests one day.

docs/src/man/split_apply_combine.md
```diff
@@ -124,6 +126,54 @@ julia> by(iris, :Species,
 │ 2   │ Iris-versicolor │ 0.717655 │ 213.0   │
 │ 3   │ Iris-virginica  │ 0.842744 │ 277.6   │
 
+julia> by(iris, :Species,
+          :PetalLength, :SepalLength,
+          AsTable([:PetalLength, :SepalLength]) => sum)
```
nalimilan, line 131: Here too, this example doesn't sound very useful as grouping has no effect. Not sure what kind of example could be useful, but if we can't find one maybe better not show anything (showing it in `select` can be enough). Same below.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -876,7 +876,8 @@ returns `true`.
 If `cols` is not specified then the function is passed `DataFrameRow`s.
 If `cols` is specified then it should be a valid column selector
 (column duplicates are allowed if a vector of `Int` or `Symbol` is passed),
-the function is passed elements of the selected columns as separate positional arguments.
+the function is passed elements of the selected columns as separate positional arguments,
+unless it is `AsTable` selector in which case `NamedTuple`s of these arguments are passed.
```
nalimilan, line 880:
```suggestion
unless it is an `AsTable` selector, in which case a `NamedTuple` of these arguments is passed.
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -930,7 +940,22 @@ Base.filter((cols, f)::Pair{<:AbstractVector{Symbol}}, df::AbstractDataFrame) =
...
+
+function Base.filter((cols, f)::Pair{<:AsTable}, df::AbstractDataFrame)
+    dff = select(df, cols.colselector, copycols=false)
+    if ncol(dff) == 0
+        throw(ArgumentError("At least one column must be passed to filter on"))
+    end
+    return _filter_helper_astable(df, Tables.namedtupleiterator(dff), f)
+end
+
+_filter_helper_astable(df, nti, f) = df[(x -> f(x)::Bool).(nti), :]
```
nalimilan, line 958: Define this above?
```suggestion
_filter_helper_astable(df::AbstractDataFrame, nti::NamedTuple, f) = df[(x -> f(x)::Bool).(nti), :]
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -930,7 +940,22 @@ Base.filter((cols, f)::Pair{<:AbstractVector{Symbol}}, df::AbstractDataFrame) =
 Base.filter((cols, f)::Pair, df::AbstractDataFrame) =
     filter(index(df)[cols] => f, df)
 
-_filter_helper(df, f, cols...) = df[((x...) -> f(x...)::Bool).(cols...), :]
+function _filter_helper(df, f, cols...)
```
nalimilan, line 943:
```suggestion
function _filter_helper(df::AbstractDataFrame, f, cols...)
```
(Would any type restriction apply to `cols`?)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1001,8 +1038,23 @@ Base.filter!((cols, f)::Pair{<:AbstractVector{Symbol}}, df::AbstractDataFrame) =
 Base.filter!((cols, f)::Pair, df::AbstractDataFrame) =
     filter!(index(df)[cols] => f, df)
 
-_filter!_helper(df, f, cols...) =
-    deleterows!(df, findall(((x...) -> !(f(x...)::Bool)).(cols...)))
+function _filter!_helper(df, f, cols...)
```
nalimilan, line 1041:
```suggestion
function _filter!_helper(df::AbstractDataFrame, f, cols...)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1001,8 +1038,23 @@ Base.filter!((cols, f)::Pair{<:AbstractVector{Symbol}}, df::AbstractDataFrame) =
...
+
+function Base.filter!((cols, f)::Pair{<:AsTable}, df::AbstractDataFrame)
+    dff = select(df, cols.colselector, copycols=false)
+    if ncol(dff) == 0
+        throw(ArgumentError("At least one column must be passed to filter on"))
+    end
+    return _filter!_helper_astable(df, Tables.namedtupleiterator(dff), f)
+end
+
+_filter!_helper_astable(df, nti, f) =
```
nalimilan, line 1056:
```suggestion
_filter!_helper_astable(df::AbstractDataFrame, nti::NamedTuple, f) =
```

src/abstractdataframe/selection.jl
```diff
@@ -17,6 +17,31 @@
 
 A type used for selection operations to signal that the wrapped function should
 be applied to each element (row) of the selection.
+
+Note that `ByRow` always wraps retun values of `fun` into a vector.
```
nalimilan, line 21: You mean that they are not destructured into columns? By the way, maybe we should reserve this in case we want to allow creating multiple columns by returning a named tuple in the future?

Also:
```suggestion
Note that `ByRow` always wraps return values of `fun` into a vector.
```

src/abstractdataframe/selection.jl
```diff
@@ -155,7 +196,8 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}},
         allow_resizing_newdf[] = false
         respar = parent(res)
         if copycols && !(fun isa ByRow) &&
-            (res isa SubArray || any(i -> respar === parent(cdf[i]), col_idx))
+            (res isa SubArray || any(i -> respar === parent(cdf[i]),
+                col_idx isa AsTable ? col_idx.colselector : col_idx))
```
nalimilan, line 200: It would be more readable to define an intermediary variable.

src/abstractdataframe/selection.jl
```diff
@@ -188,6 +230,8 @@ SELECT_ARG_RULES =
     is a `Symbol` or an integer then `fun` is applied to the corresponding column vector.
     Otherwise `old_column` can be any column indexing syntax, in which case `fun`
     will be passed the column vectors specified by `old_column` as separate arguments.
+    Additionally if `old_column` is an `AsTable` type wrapping a column selection argument,
+    then `fun` is passed a `NamedTuple` containing the selected columns.
```
nalimilan, line 234:
```suggestion
    The only exception is when `old_column` is an `AsTable` type wrapping a selector,
    in which case `fun` is passed a `NamedTuple` containing the selected columns.

```

src/abstractdataframe/selection.jl
```diff
@@ -287,6 +333,17 @@ julia> df
 │     │ Int64 │ Int64 │
 ├─────┼───────┼───────┤
 │ 1   │ 6     │ 15    │
+
+julia> df = DataFrame(a=1:3, b=4:6);
+
+julia> select!(df, AsTable(:) => ByRow(x -> map(x -> x^2, x)))
```
nalimilan, line 339: Maybe also show an example which is useful in practice, like the row means?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -219,7 +219,8 @@ If `pair` is passed then it must follow the rules specified for transformations
 [`select`](@ref) and have the form `source_cols => fun`, `source_cols => fun => target_col`,
 or `source_col => target_col`.
 Function defined by `fun` is passed `SubArray` views as positional arguments for
-each column specified to be selected and can return any return value defined below.
+each column specified to be selected and can return any return value defined below,
+or a `NamedTuple` containing these `SubArray`s if `source_cols` is a `AsTable` selector.
```
nalimilan, line 223:
```suggestion
or a `NamedTuple` containing these `SubArray`s if `source_cols` is an `AsTable` selector.
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -358,7 +375,8 @@ const F_ARGUMENT_RULES =
     for [`select`](@ref) and have the form `source_cols => fun`,
     `source_cols => fun => target_col`, or `source_col => target_col`.
     Function `fun` is passed `SubArray` views as positional arguments for each column
-    specified to be selected and can return a vector or a single value
+    specified to be selected, or a `NamedTuple` containing these `SubArray`s if
+    `source_cols` is a `AsTable` selector, and can return a vector or a single value
```
nalimilan, line 379:
```suggestion
    `source_cols` is an `AsTable` selector. It can return a vector or a single value
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1064,6 +1113,10 @@ function _combine(f::AbstractVector{<:Pair},
         else
             if source_cols isa Int
                 incols = (parentdf[!, source_cols],)
+            elseif source_cols isa AsTable
+                incols = Tables.columntable(select(parentdf,
```
nalimilan, line 1117: Why not use the indexing syntax (here and elsewhere), for consistency with other branches?

src/other/utils.jl
```diff
@@ -1,3 +1,15 @@
+"""
+    AsTable
+
+A type used for selection operations to signal that the columns selected by a wrapped
```
nalimilan, line 4:
```suggestion
A type used for selection operations to signal that the columns selected by the wrapped
```

src/other/utils.jl
```diff
@@ -1,3 +1,15 @@
+"""
+    AsTable
```
nalimilan, line 2: Maybe mirror `?All`:
```suggestion
    AsTable(cols)
```
The internal field could also use that name.

bkamins:
> You mean that they are not destructured into columns? By the way, maybe we should reserve this in case we want to allow creating multiple columns by returning a named tuple in the future?

Let me start with the example on v0.20:
```
julia> using DataFrames

julia> df = DataFrame(g=1:2);

julia> by(df, :g, :g => x -> [(a=x,),])
2×2 DataFrame
│ Row │ g     │ g_function │
│     │ Int64 │ NamedTup…  │
├─────┼───────┼────────────┤
│ 1   │ 1     │ (a = [1],) │
│ 2   │ 2     │ (a = [2],) │
```
in other words now we are reproducing the behavior we already have.

Also on master `select` will do the same without `ByRow`:
```
julia> df = DataFrame(g=1:2)
2×1 DataFrame
│ Row │ g     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │

julia> df = DataFrame(g=1:2);

julia> select(df, :g => x -> [(a=x,)])
1×1 DataFrame
│ Row │ g_function    │
│     │ NamedTuple…   │
├─────┼───────────────┤
│ 1   │ (a = [1, 2],) │
```

In summary `ByRow` returns a vector so it is not unwrapped as in other cases where we do not unwrap vectors.

If we disallowed this then there would be no possibility to return a produce a vector of `NamedTuple`s from `ByRow`.

Note that we disallow returning `NamedTuple` (and other types but let us focus on `NamedTuple`s) if it is not a `ByRow` operation. This is reserved for the case you describe in the future. However, note - that in this case this does not disallow producing a `DataFrame` holding `NamedTuple`s as you can wrap them in a vector (as in the example above) or in `Ref` and you have no problem.

In summary: we currently treat a `NamedTuple` of objects as a table (so we disallow it) but treat a vector of `NamedTuples` just as a vector - so we allow it. Therefore my point is that we are consistent now with the design.
Also I do not see too much use of returning multiple columns from a single row of data (it is easy enough to do with multiple `ByRow` without performance loss as we are operating on single values anyway)

But if you feel that `NamedTuple` destruction from `ByRow` would be a potentially useful case then we can add this as an exception.

bkamins:
> My main remark is about showing more useful examples.

I will try to think of something, but I really believe that the default API is more useful most of the time (except for row aggregations - I will try to come up with some better cases).

@pdeffebach - can you please suggest something as I know that you had many good ideas here.

docs/src/man/getting_started.md
bkamins: added

docs/src/man/split_apply_combine.md
bkamins: I would leave this as is (and maybe disable doctests for selected chunks if this is not fixable). My idea is that this is a manual, so we should show things that one would normally do. And normally you just show the data frame and rely on default behaviour I think.

@mortenpi - do you know if it is possible to set a custom width and height of `stdout` in doctests?

docs/src/man/split_apply_combine.md
bkamins: agreed (`ByRow` works because it exists). removed. `AsTable` is mentioned and another example is present.

docs/src/man/split_apply_combine.md
bkamins: ok - changed both.

src/abstractdataframe/selection.jl
bkamins: changed

src/abstractdataframe/selection.jl
bkamins: ok - changed

src/groupeddataframe/splitapplycombine.jl
bkamins: A good question :). The reason is that `source_cols.colselector` can be a `ColumnIndex`, in which case indexing syntax would not produce a table, but a vector, and then `Tables.columntable` would be broken. (I will make sure that we test for this)

src/abstractdataframe/abstractdataframe.jl
bkamins: `AbstractVector` - I will add this.

src/abstractdataframe/abstractdataframe.jl
bkamins: OK - I will move it and make the type restriction more strict.

src/abstractdataframe/selection.jl
bkamins: Yes - they are not destructed. I have expanded on why I did it below. But please comment there if you feel otherwise.

src/other/utils.jl
bkamins: OK - I will rename the internal field

bkamins:
Regarding `_filter_helper` family:
1. I fixed the signatures (but with `cols...` it can be either multiple vectors or a single `DataFrameRows` object so I decided to leave it undefined)
2. I could not find a good place to move them and keep consistency in the code layout and readability, so I left them below the functions that use them

pdeffebach:
> @pdeffebach - can you please suggest something as I know that you had many good ideas here.

Sum and mean by row are the two use-cases, I think. A cool thing about `AsTable` would be working with the names super easily. 

```
function findwinner(nt)
    m = maximum(nt)
    findall(x -> x == m, nt)
end
```

This returns a vector of the  symbols equal to the max. 

Also, I agree with Milan that unwrapping a named tuple would be convenient. A whole workflow around named-tuple to named-tuple functions seems very useful. I think I understand the logic of your objections though and they are valid points.

bkamins:
Thank you for the suggestion. I will think of something along these lines and add it.

> unwrapping a named tuple would be convenient

It will be allowed in the future. I just did not plan for it in `ByRow` (this can be changed as an exception if we want it). Do you have some example where `AsTable(cols) => ByRow(fun)` would usefully return a `NamedTuple`?

pdeffebach:
> Do you have some example where `AsTable(cols) => ByRow(fun)` would usefully return a `NamedTuple`?

With the kinds of surveys I worked with the past two years, maybe something like 

```
function add_mean(nt)
    m = mean(skipmissing(nt))
    countmissing = sum(ismissing, nt)
    return (rowmean = m, countmissing = countmissing)
end
```

I can't say with complete confidence how much I would use this (I am not analyzing survey data on a day-to-day basis any more). But it seems like a natural workflow that is consistent with using `AsTable` without `ByRow`.

bkamins:
Yeah - so it is essentially, when you want to conditionally on row filter out some columns, and based on these filtered columns produce multiple outputs. This is a valid use case. The question is how often it is needed (as a normal way to do such things would be to transform from wide to long format and then do aggregation).

As a side note a `Vector` of `NamedTuple`s is not that bad data structure - it adds a layer of grouping variables in a `DataFrame` and I tend to like it (this is something e.g. BigQuery provides).

bkamins:
> This returns a vector of the symbols equal to the max.

Added such an example to getting started section.

docs/src/man/getting_started.md
```diff
@@ -673,6 +673,46 @@ julia> Tables.rowtable(df)
...
+│ 3   │ 0.312707   │ 0.424718  │ 0.873544  │
+│ 4   │ 0.00790928 │ 0.773223  │ 0.278582  │
+│ 5   │ 0.488613   │ 0.28119   │ 0.751313  │
+│ 6   │ 0.210968   │ 0.209472  │ 0.644883  │
+│ 7   │ 0.951916   │ 0.251379  │ 0.0778264 │
+│ 8   │ 0.999905   │ 0.0203749 │ 0.848185  │
+│ 9   │ 0.251662   │ 0.287702  │ 0.0856352 │
+│ 10  │ 0.986666   │ 0.859512  │ 0.553206  │
+
+julia> transform(df, AsTable(:) => ByRow(argmax) => :prediction)
```
nalimilan, line 700: This example is really cool!

docs/src/man/getting_started.md
```diff
@@ -673,6 +673,46 @@ julia> Tables.rowtable(df)
  (x1 = 1, x2 = 3, y = 5)
  (x1 = 2, x2 = 4, y = 6)
 ```
+Now assume that a data frame `df` contains predictions from a model producing scores
+for three levels `a`, `b` and `c` of a nominar target variable.
+For each row we want to get the level with the highest score.
```
nalimilan, line 678: I'd avoid the context, finding the column with the highest value can happen in many situations, no need to justify.
```suggestion
With this approach, we can easily compute for each row the name of column with the highest score:
```

docs/src/man/getting_started.md
```diff
@@ -634,7 +634,9 @@ julia> df
 ```
 
 `transform` and `transform!` functions work identically to `select` and `select!` with the only difference that
-they retain all columns that are present in the source data frame, for example:
+they retain all columns that are present in the source data frame. Here are some more advanced examples.
+
+First we show how to generate a column that is a sum of all other columns in the data frame using `All()` selector:
```
nalimilan, line 639:
```suggestion
First we show how to generate a column that is a sum of all other columns in the data frame
using the `All()` selector:
```

docs/src/man/getting_started.md
```diff
@@ -653,8 +655,86 @@ julia> transform(df, All() => +)
 │ 1   │ 1     │ 3     │ 5     │ 9         │
 │ 2   │ 2     │ 4     │ 6     │ 12        │
 ```
+Here we wrap rows of the data frame into a `NamedTuple` while remembering
+source column names.
```
nalimilan, line 659:
```suggestion
source column names:
```

docs/src/man/split_apply_combine.md
nalimilan: It's still there. :-)

docs/src/man/split_apply_combine.md
nalimilan: As you like. Anyway enabling doctests will take some work.

src/abstractdataframe/selection.jl
```diff
@@ -17,6 +17,31 @@
 
 A type used for selection operations to signal that the wrapped function should
 be applied to each element (row) of the selection.
+
+Note that `ByRow` always wraps return values of `fun` into a vector.
```
nalimilan, line 21:
```suggestion
Note that `ByRow` always collects values returned by `fun` in a vector. Therefore,
returning a `NamedTuple` produces a column of `NamedTuple` values, rather than
multiple columns.
```

docs/src/man/split_apply_combine.md
mortenpi: If I understand the question correctly, then it is an open issue in Documenter, but there might a workaround via setting the appropriate environment variables (`COLUMNS` and [maybe `LINES`?](https://stackoverflow.com/questions/1780483/lines-and-columns-environmental-variables-lost-in-a-script)): https://github.com/JuliaDocs/Documenter.jl/issues/942#issuecomment-597905221

bkamins: Thank you! (you understand the question correctly :smile:)

@nalimilan - so we are safe, and when we switch to enable doctests the PR (or a standard practice) in Documenter.jl probably will be merged.

nalimilan:
Thanks for the explanation. I'm not sure the consistency is so visible for users, as they may also expect that returning a `NamedTuple` always creates multiple columns -- ignoring the fact that the result of `ByRow` is put into a vector.

I wonder what kind of solution we could use to allow both behaviors in the future if needed. Wrapping named tuples in `Ref`? That would mean that if you really want to return a `Ref` with `ByRow`, you would have to wrap it in another `Ref`. Maybe that's not too bad, and that would be consistent with the pseudo-broadcasting we apply when `ByRow` isn't used.

Anyway for now I'd err on the safe side and throw an error (as usual), returning named tuples from `ByRow` isn't the most urgent need.

bkamins:
> Anyway for now I'd err on the safe side and throw an error (as usual

OK. Later you will be able to pass `(a = (x=1,y=2),)` to get your `NamedTuple` unwrapped into column `:a`.

docs/src/man/split_apply_combine.md
bkamins: Ah - removed now (I have removed it in other place).

bkamins:
CI passes here (nightly fails due to https://github.com/JuliaData/DataFrames.jl/issues/2184).

docs/src/man/getting_started.md
```diff
@@ -656,24 +656,6 @@ julia> transform(df, All() => +)
...
-│ 2   │ 2     │ 4     │ 6     │ (x1 = 2, x2 = 4, y = 6) │
-```
-Note that the same column could be generated by using the `Tables.rowtable` function:
-```
-julia> Tables.rowtable(df)
-2-element Array{NamedTuple{(:x1, :x2, :y),Tuple{Int64,Int64,Int64}},1}:
- (x1 = 1, x2 = 3, y = 5)
- (x1 = 2, x2 = 4, y = 6)
-```
 With this approach, we can easily compute for each row the name of column with the highest score:
```
nalimilan, line 659:
```suggestion
Using the `ByRow` wrapper, we can easily compute for each row the name of column with the highest score:
```

src/abstractdataframe/selection.jl
```diff
@@ -19,40 +19,23 @@ A type used for selection operations to signal that the wrapped function should
...
-3×2 DataFrame
-│ Row │ a     │ b     │
-│     │ Int64 │ Int64 │
-├─────┼───────┼───────┤
-│ 1   │ 1     │ 4     │
-│ 2   │ 2     │ 5     │
-│ 3   │ 3     │ 6     │
-```
+to allow for future extensions `NamedTuple` or `DataFrameRow` from `fun` is
+currently disallowed.
```
nalimilan, line 23:
```suggestion
to allow for future extensions, returning `NamedTuple` or `DataFrameRow`
from `fun` is currently disallowed.
```

src/abstractdataframe/selection.jl
```diff
@@ -19,40 +19,23 @@ A type used for selection operations to signal that the wrapped function should
...
 """
 struct ByRow{T}
     fun::T
 end
 
+_by_row_helper(x::Any) = x
+_by_row_helper(x::Union{NamedTuple, DataFrameRow}) =
+    throw(ArgumentError("return value of type $(typeof(x)) " *
+                        "is currently not allowed in ByRow."))
```
nalimilan, line 32:
```suggestion
                        "is currently not allowed with ByRow."))
```

bkamins:
Thank you!

stillyslalom:
> @bkamins
> Yeah - so it is essentially, when you want to conditionally on row filter out some columns, and based on these filtered columns produce multiple outputs. This is a valid use case. The question is how often it is needed (as a normal way to do such things would be to transform from wide to long format and then do aggregation). 

I have a use case where I'm doing something like a coordinate transformation (it can't be treated as an aggregation). In v0.20, I was able to do something like the following (boiled down considerably):
```julia
using Statistics, DataFrames, Polynomials

function forcecalc(df, weight_offsets)
    α = mean(df.α)  # angle of attack
    AF = mean(df.AF) - weight_offsets[1](α)    # axial force
    NF = mean(df.NF) - weight_offsets[2](α)    # normal force
    (L = -NF*cos(α) - AF*sin(α),    # lift
     D = -NF*sin(α) + AF*cos(α))    # drag
end

α = vcat([fill(v, 30) for v in -2:2:16]...)
AF =  10sind.(α) .+ rand(length(α))
NF = 100cosd.(α) .+ rand(length(α))
df = DataFrame(α = α, AF = AF, NF = NF)

wt = [Polynomial(rand(3)), Polynomial(rand(3))] # weight offsets
forces = by(df, :α, [:α, :AF, :NF] => df -> forcecalc(df, wt))
```
I'm subtracting a weight offset polynomial from raw force measurements (:AF and :NF), then rotating the force vector. I haven't found an elegant way to do this in v0.21, but my desired syntax is something like the following:
```julia
function forcecalc(nt::NamedTuple, weight_offsets)
    α = nt.α  # angle of attack
    AF = nt.AF - weight_offsets[1](α)    # axial force
    NF = nt.NF - weight_offsets[2](α)    # normal force
    (L = -NF*cos(α) - AF*sin(α),    # lift
     D = -NF*sin(α) + AF*cos(α))    # drag
end

varnames = [:AF, :NF]
gdf = groupby(df, :α)
measurements = combine(gdf, varnames .=> mean .=> varnames)
forces = transform(measurements, [:α, :AF, :NF] => ByRow(v -> forcecalc(v, wt)))
```
Note that in the former case, I'm both averaging measurements and applying the transformation in the 'apply' step of split-apply-combine, whereas in the latter case, I'm applying the transformation after s-a-c.

bkamins:
I have not analyzed your code in detail, but given I know the design assumptions :) the code that should work without changing anything in your 0.20 code except the last line is:
```
forces = combine(AsTable([:α, :AF, :NF]) => nt -> forcecalc(nt, wt), groupby(nt, :α))
```
Can you please confirm that it works as expected?

(still maybe it is possible to write it in a cleaner way - I am just doing code transformation that should be equivalent between 0.20 and 0.21)

stillyslalom:
Save for a typo (should be `groupby(df, :α)` instead of `groupby(nt, :α)`), it works! I think I was confused by the documentation for `combine`, which seems to imply that a GroupedDataFrame can be passed as either the first or second argument.
```
combine(gd::GroupedDataFrame, args...; keepkeys::Bool=true, ungroup::Bool=true)
combine(fun::Union{Function, Type}, gd::GroupedDataFrame; keepkeys::Bool=true, ungroup::Bool=true)
```
> Arguments passed as args... can be:
> * Column transformation operations using the Pair notation that is described below and vectors of such pairs.

That said, I would prefer to perform reduction and transformation in two separate steps (as in my "v0.22" example), since the broader workflow involves applying the same reduction operation, followed by a variety of transformations.

bkamins:
>  which seems to imply that a GroupedDataFrame can be passed as either the first or second argument.

This is indeed allowed and has a different behavior:

1) `GroupedDataFrame` passed as a first argument requires all following arguments to be `Pair`s and return only one columns
2) `GroupedDataFrame` passed as a second argument allows the first argument to be a `Pair` or a function (to allow `do` block syntax) and in this case you are allow to return one column or a table (i.e. multiple columns in one shot)

stillyslalom:
Does the second behavior suffer the same performance impact for `Pair`s as it does for functions?

bkamins:
No for `Pair` it is type-stable, and if you pass a pair like `:a => sum` that uses aggregation functions it will use the fastest code path available.

PR #2188, opened 2020-04-15
Port to CategoricalArrays 0.8

nalimilan:
Replace `CategoricalString` with `CategoricalValue{String}` and stop using `CategoricalArrays.index`. This simplifies the code a bit. Also drop some specializations which are equivalent to optimized methods defined in CategoricalArrays.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -719,16 +719,13 @@ function completecases(df::AbstractDataFrame, col::Colon=:)
     end
     res = trues(size(df, 1))
     for i in 1:size(df, 2)
-        _nonmissing!(res, df[!, i])
+        res .&= .!ismissing.(df[!, i])
     end
     res
 end
 
-function completecases(df::AbstractDataFrame, col::ColumnIndex)
-    res = trues(size(df, 1))
-    _nonmissing!(res, df[!, col])
-    res
-end
+completecases(df::AbstractDataFrame, col::ColumnIndex) =
+    .!ismissing.(df[!, col])
 
 completecases(df::AbstractDataFrame, cols::Union{AbstractVector, Regex, Not, Between, All}) =
     completecases(df[!, cols])
@@ -1719,7 +1716,7 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
         x = df[!, i]
         if i in idxcols
             # categorical always copies
-            push!(newcols, categorical(x, compress))
+            push!(newcols, categorical(x, compress=compress))
         else
             push!(newcols, copy(x))
         end
@@ -1735,7 +1732,7 @@ function CategoricalArrays.categorical(df::AbstractDataFrame,
         x = df[!, i]
         if eltype(x) <: cols
             # categorical always copies
-            push!(newcols, categorical(x, compress))
+            push!(newcols, categorical(x, compress=compress))
         else
             push!(newcols, copy(x))
         end
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -645,21 +645,6 @@ end
 ##
 ##############################################################################
 
-function _nonmissing!(res, col)
-    @inbounds for (i, el) in enumerate(col)
-        res[i] &= !ismissing(el)
-    end
-    return nothing
-end
-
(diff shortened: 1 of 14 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -719,16 +719,13 @@ function completecases(df::AbstractDataFrame, col::Colon=:)
     end
     res = trues(size(df, 1))
     for i in 1:size(df, 2)
-        _nonmissing!(res, df[!, i])
+        res .&= .!ismissing.(df[!, i])
```
bkamins, line 722: you can remove `_nonmissing` definition above.

src/abstractdataframe/reshape.jl
```diff
@@ -209,15 +206,15 @@ function _unstack(df::AbstractDataFrame, rowkey::Int, colkey::Int, value::Int,
             end
             i = length(unstacked_val[1])
         else
-            i = refkeycol_order[refkref]
+            i = refkref
```
bkamins, line 209: maybe add type restrictions for `_unstack` parameters?

src/dataframerow/dataframerow.jl
```diff
@@ -278,26 +278,13 @@ Base.merge(a::DataFrameRow, b::DataFrameRow) = merge(NamedTuple(a), NamedTuple(b
...
-        hash(CategoricalArrays.index(v.pool)[ref], h)
-    end
-end
-
 # hash of DataFrame rows based on its values
 # so that duplicate rows would have the same hash
 # table columns are passed as a tuple of vectors to ensure type specialization
 rowhash(cols::Tuple{AbstractVector}, r::Int, h::UInt = zero(UInt))::UInt =
-    hash_colel(cols[1], r, h)
+    hash(cols[1][r], h)
```
bkamins, line 285: I understand that now it is not needed and proper hashing is ensured by CategoricalArrays.jl?

src/deprecated.jl
```diff
@@ -18,8 +18,8 @@ function DataFrame(column_eltypes::AbstractVector{T}, cnames::AbstractVector{Sym
     end
     for i in eachindex(categorical)
         categorical[i] || continue
-        elty = CategoricalArrays.catvaluetype(nonmissingtype(updated_types[i]),
-                                              CategoricalArrays.DefaultRefType)
+        elty = CategoricalValue{nonmissingtype(updated_types[i]),
```
bkamins, line 21: this deprecation is 1 year old - maybe just remove this method from deprecated.jl?

test/broadcasting.jl
```diff
@@ -699,9 +699,9 @@ end
         @test levels(df.c1) == levels(v)
         @test levels(df.c1) !== levels(v)
         df[!, :c2] .= v[2]
-        @test df.c2 == get.([v[2], v[2], v[2]])
+        @test df.c2 == fill(v[2], 3)
         @test df.c2 isa CategoricalVector
-        @test levels(df.c2) != levels(v)
+        @test levels(df.c2) == levels(v)
```
bkamins, line 704: I understand this is a consequence of copying of all levels by `setindex!`?

bkamins:
Looks good apart from what I commented on. Thank you!

src/deprecated.jl
nalimilan: You know I love keeping old deprecated methods. :-) I'd say let's remove them all at once later if we want.

test/broadcasting.jl
nalimilan: Yes, that's a more logical behavior than before.

src/dataframerow/dataframerow.jl
nalimilan: Yes. Actually I'm not even sure it was required before. It's slightly faster, but not that much, and working with `DataFrameRow` is inefficient anyway (we have much more efficient code when hashing a full column, based on hashing each level only once). What do you think?

```julia
julia> df = categorical(DataFrame(rand(["a", "b", "c"], 10, 10)));

julia> r = view(df, 1, :);

# master
julia> @btime hash($r);
  471.533 ns (7 allocations: 352 bytes)

# this PR
julia> @btime hash($r);
  552.241 ns (7 allocations: 352 bytes)

bkamins: It is OK given it is hardly used anyway.

PR #2189, opened 2020-04-15
[BREAKING] Rename deleterows! to delete!

nalimilan:
`delete!` was renamed to `deletecols!` a few releases ago, now it's time to finish the plan (https://github.com/JuliaData/DataFrames.jl/issues/1514) and rename `deleterows!` to `delete!` for consistency with the rest of the API that treats data frames as collections of rows.


nalimilan:
I'd rather leave that for later. ;-)

PR #2190, opened 2020-04-15
clean-up reshaping code

bkamins:
@nalimilan It would be good to merge this before #2188 and before the PR allowing string indexing.

The old design was overly complex and was hard to reason about for me (both mentioned PRs will be simpler after this change).

The functionality is untouched.

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -228,7 +228,7 @@ function unstack(df::AbstractDataFrame, rowkeys, colkey::ColumnIndex,
     @assert rowkey_ints isa AbstractVector{Int}
     length(rowkey_ints) == 0 && throw(ArgumentError("No key column found"))
     length(rowkey_ints) == 1 && return unstack(df, rowkey_ints[1], colkey, value,
-                                           renamecols=renamecols)
+                                               renamecols=renamecols)
     g = groupby(df, rowkey_ints, sort=true)
     keycol = categorical(df[!, colkey])
     droplevels!(keycol)
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -51,14 +51,47 @@ d1m = stack(d1, Not([:a, :b, :e]))
 d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
 ```
 """
-function stack(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
-               id_vars::AbstractVector{<:Integer}; variable_name::Symbol=:variable,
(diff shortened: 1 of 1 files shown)


src/abstractdataframe/reshape.jl
```diff
@@ -225,48 +219,36 @@ function _unstack(df::AbstractDataFrame, rowkey::Int, colkey::Int, value::Int,
...
-    length(rowkeys) == 0 && throw(ArgumentError("No key column found"))
-    length(rowkeys) == 1 && return unstack(df, rowkeys[1], colkey, value, renamecols=renamecols)
-    g = groupby(df, rowkeys, sort=true)
+function unstack(df::AbstractDataFrame, rowkeys, colkey::ColumnIndex,
+                 value::ColumnIndex; renamecols::Function=identity)
+    rowkey_ints = index(df)[rowkeys]
+    @assert rowkey_ints isa AbstractVector{Int}
+    length(rowkey_ints) == 0 && throw(ArgumentError("No key column found"))
+    length(rowkey_ints) == 1 && return unstack(df, rowkey_ints[1], colkey, value,
+                                           renamecols=renamecols)
```
nalimilan, line 231:
```suggestion
                                               renamecols=renamecols)
```

src/abstractdataframe/reshape.jl
```diff
@@ -51,74 +51,67 @@ d1m = stack(d1, Not([:a, :b, :e]))
 d1s_name = stack(d1, Not([:a, :b, :e]), variable_name=:somemeasure)
 ```
 """
-function stack(df::AbstractDataFrame, measure_vars::AbstractVector{<:Integer},
-               id_vars::AbstractVector{<:Integer}; variable_name::Symbol=:variable,
+function stack(df::AbstractDataFrame,
+               measure_vars = findall(col -> eltype(col) <: Union{AbstractFloat, Missing},
```
nalimilan, line 55: While we're at it, shouldn't this use `Real`?

bkamins: I have not designed it. I think the intention was to exclude integers (only take floats), as integers are often used as ids not as values.

bkamins:
thank you for having a look at it.

src/abstractdataframe/reshape.jl
nalimilan: OK. Not sure I love that behavior but...

bkamins: Well - I would prefer not to have the default at all (we do not use such a pattern anywhere else in the package), but let us leave it as it does not hurt I think and removing it would be breaking.

bkamins:
Thank you!

PR #2191, opened 2020-04-15
add Not to delete!

bkamins:
This is not urgent, but simple and will improve consistency.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -762,8 +762,8 @@ end
 
 Delete rows specified by `inds` from a `DataFrame` `df` in place and return it.
 
-Internally `deleteat!` is called for all columns so `inds` must
-be: a vector of sorted and unique integers, a boolean vector or an integer.
+Internally `deleteat!` is called for all columns so `inds` must be:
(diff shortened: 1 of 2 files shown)


src/dataframe/dataframe.jl
```diff
@@ -791,19 +791,21 @@ function Base.delete!(df::DataFrame, inds)
...
     foreach(col -> deleteat!(col, inds), _columns(df))
-    df
+    return df
 end
 
 function Base.delete!(df::DataFrame, inds::AbstractVector{Bool})
     if length(inds) != size(df, 1)
         throw(BoundsError(df, (inds, :)))
     end
-    drop = findall(inds)
```
nalimilan, line 802: I think this was intentional: computing integer indices once is faster than using logical indexing for each column, as it requires computing the sum again and again. Though a comment wouldn't hurt.

bkamins: indeed - now I remember thinking about it some tome ago. I will revert. Anyway this PR should not be merged before #2199 as it is conflicting. Thank you!

bkamins:
Thank you. I will rebase it and merge after #2199 is done

bkamins:
Thank you!

Issue #2192, opened 2020-04-16
Why does categorical!(df::DataFrame, ...) exist?


matthieugomez:
I don't understand why `categorical(df, cols)` exists. 

If it is about applying a function (`categorical`) to multiple columns of a dataframes, there should be a way to do it for any function, not just `categorical` (I guess that's what `transform` is about?).

I'm not following the package closely, so sorry if this was discussed somewhere.

bkamins:
Thank you for raising this point (we have a similar case with `allowmissing` and `disallowmissing`). Simply in the past our API was much less flexible.

It has one minor functionality that:
1. by default it converts string column
2. it can take type as a column selector

But apart from this it is true that roughly (I am dropping some corner cases):
```
@deprecate categorical(df, cols) transform(df, cols .=> categorical .=> cols)
```
and
```
@deprecate categorical!(df, cols) transform!(df, cols .=> categorical .=> cols)
```

Having said this, I do not see a huge problem to keep `categorical` and `categorical!` as convenience functions. We define such functions for most common transformations, so that users have an easier life. @nalimilan - what do you think?

nalimilan:
Yes I think it makes sense to have some convenience functions. That's the clearest for `allowmissing` as it's a very common need, but `categorical` is already exported since it exists in CategoricalArrays, so adding a new method isn't very costly.

For other functions, `mapcols` can also be used instead of `transform`. I guess we could add `mapcols!` for in-place operation.

matthieugomez:
Thanks for the explanation! I think it'd be better to have only one way to do things, but that's your choice! I appreciate all the improvement happening these days, by the way. Thanks for your work!

PR #2195, opened 2020-04-16
add mapcols! and repeat!, fix corner cases of repeat

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2192

Note that `repeat` in Base has a bug if `inner<0` or `outer<0` so I catch it now properly before we dispatch there


src/dataframe/dataframe.jl
```diff
@@ -1368,3 +1368,79 @@ function Base.push!(df::DataFrame, row::Any)
     end
     df
 end
+
+"""
+    repeat!(df::DataFrame; inner::Integer = 1, outer::Integer = 1)
+
+Update a data frame `df` in place by repeating rows. `inner` specifies how many
```
nalimilan, line 1375:
```suggestion
Update a data frame `df` in place by repeating its rows. `inner` specifies how many
```

src/dataframe/dataframe.jl
```diff
@@ -1368,3 +1368,79 @@ function Base.push!(df::DataFrame, row::Any)
...
+function repeat!(df::DataFrame; inner::Integer = 1, outer::Integer = 1)
+    inner < 0 && throw(ArgumentError("inner keyword argument must be non-negative"))
+    outer < 0 && throw(ArgumentError("outer keyword argument must be non-negative"))
+    return mapcols!(x -> repeat(x, inner = inner, outer = outer), df)
+end
+
+"""
+    repeat!(df::DataFrame, count::Integer)
+
+Update a data frame `df` in place by repeating rows the number of times
```
nalimilan, line 1419:
```suggestion
Update a data frame `df` in place by repeating its rows the number of times
```

test/utils.jl
```diff
@@ -1,33 +1,84 @@
...
+
+    @test_throws MethodError repeat!(view(df, 1:2, :), 2)
+end
+
+@testset "repeat! inner_outer" begin
+    df = DataFrame(a = 1:2, b = 3:4)
+    ref = DataFrame(a = repeat(1:2, inner = 2, outer = 3),
+                    b = repeat(3:4, inner = 2, outer = 3))
+    repeat!(df, inner = 2, outer = 3)
+    @test df == ref
```
nalimilan, line 64: Also test that column vectors are `===` or `!==` from original column vectors (depending on our decision)?

bkamins:
> OTOH it doesn't make much sense for performance.

Yes - that is why I have made it make a full copy. It is safer this way in case we have column aliases (we should not have typically, but in some cases we might have). And doing alias detection in this function would be an over-complication.

test/utils.jl
bkamins: OK - added.

bkamins:
@nalimilan this is not crucial to merge now, but IMHO is finished and rather simple to decide.

bkamins:
I have added a note that `repeat!` freshly allocates columns.

bkamins:
Thank you!

Issue #2164, opened 2020-03-22
Help `skipmissing` with a DataFrameRow full of `Missing`s


pdeffebach:
In reference to #2161, we have the following behavior. 

```
julia> df = DataFrame(a = [1, missing], b = [2, missing])
2×2 DataFrame
│ Row │ a       │ b       │
│     │ Int64?  │ Int64?  │
├─────┼─────────┼─────────┤
│ 1   │ 1       │ 2       │
│ 2   │ missing │ missing │

julia> sum(skipmissing(df[2, :]))
ERROR: ArgumentError: reducing over an empty collection is not allowed
```

But with NamedTuples we can do 

```
julia> nt = NamedTuple{(:a,:b), Tuple{Union{Int,Missing}, Union{Int,Missing}}}((missing, missing));

julia> sum(skipmissing(nt))
0
```

Given that a DataFrameRow is *kind of* like a NamedTuple we should be able to support this behavior. I don't know off the top of my head what we would need to make this work. Probably a customized `Base.mapreduce_empty` implementation.

bkamins:
The problem is that `DataFrameRow` does not have column types encoded in its type (and this is a typical assumption of `eltype`). The reason is that column types can change at any time (e.g. after you have taken the column type information and before using this information).

Therefore `DataFrameRow` now supports the default and promises `Any` eltype only.

I am not sure what would be a solution here. The problem is that functions in Base mostly assume `eltype` can be determined from type not from value.

pdeffebach:
>  (e.g. after you have taken the column type information and before using this information).

I'm not sure I fully understand. Could you give an example of how this might break? I think `eltype` is all that would be needed here so maybe there is a way to make it work. I will check tomorrow.

bkamins:
```
julia> df = DataFrame(ones(2,3))
2×3 DataFrame
│ Row │ x1      │ x2      │ x3      │
│     │ Float64 │ Float64 │ Float64 │
├─────┼─────────┼─────────┼─────────┤
│ 1   │ 1.0     │ 1.0     │ 1.0     │
│ 2   │ 1.0     │ 1.0     │ 1.0     │

julia> for x in skipmissing(df[1, :])
           df[!, :] .= "a"
           @show x
       end
x = 1.0
x = "a"
x = "a"
```

nalimilan:
[...]

PR #2196, opened 2020-04-16
dataframerow related docstrings

pdeffebach:
In reference to #2164 . Just some short text to make it a bit clearer (hopefully) how to work with `eachrow` a bit better. 

Long term solutions might include a `eachrow(df, namedtuples = true)`. But this seems unnecessary given `select` with `ByRow`.

### DIFF
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -32,7 +32,7 @@ with each row represented as a `DataFrameRow`.
 
 Because `DataFrameRow`s have an `eltype` of `Any`, use `copy(dfr::DataFrameRow)` to obtain 
 a named tuple, which supports iteration and property access like a `DataFrameRow`, 
-but also passes information of the `eltypes` of the columns of `df`. 
+but also passes information on the `eltypes` of the columns of `df`. 
 
 # Examples
 ```jldoctest
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -30,6 +30,10 @@ Base.iterate(::AbstractDataFrame) =
 Return a `DataFrameRows` that iterates a data frame row by row,
 with each row represented as a `DataFrameRow`.
 
+Because `DataFrameRow`s have an `eltype` of `Any`, use `copy(dfr::DataFrameRow)` to obtain 
+a named tuple, which supports iteration and property access like a `DataFrameRow`, 
+but also passes information of the `eltypes` of the columns of `df`. 
+
 # Examples
 ```jldoctest
 julia> df = DataFrame(x=1:4, y=11:14)
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -276,7 +276,7 @@ function Base.NamedTuple(dfr::DataFrameRow)
     pc = parentcols(index(dfr))
     cols = _columns(parent(dfr))
     s = ntuple(i -> eltype(cols[pc[i]]), length(dfr))
-    NamedTuple{k, s...}(v)
+    NamedTuple{k, Tuple{s...}}(v)
 end
 
 """
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -14,8 +14,7 @@ DataFrameRow(parent::AbstractDataFrame, row::Integer, cols=:)
 ```
 
 A `DataFrameRow` supports the iteration interface and can therefore be passed to
-functions that expect a collection as an argument. In such situations `eltype`
-of `DataFrameRow` is always `Any`.
+functions that expect a collection as an argument. Its element type is always `Any`.
 
 Indexing is one-dimensional like specifying a column of a `DataFrame`.
 You can also access the data in a `DataFrameRow` using the `getproperty` and
@@ -272,11 +271,12 @@ Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
 function Base.NamedTuple(dfr::DataFrameRow)
-    k = keys(dfr)
-    v = values(dfr)
-    n = names(dfr)
-    s = eltype.(parent(dfr)[!, ni] for ni in n)
-    NamedTuple{Tuple(k), Tuple{s...}}(Tuple(v))
+    k = Tuple(_names(dfr))
+    v = ntuple(i -> dfr[i], length(dfr))
+    pc = parentcols(index(dfr))
+    cols = _columns(parent(dfr))
+    s = ntuple(i -> eltype(cols[pc[i]]), length(dfr))
+    NamedTuple{k, s...}(v)
 end
 
 """
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -14,7 +14,8 @@ DataFrameRow(parent::AbstractDataFrame, row::Integer, cols=:)
 ```
 
 A `DataFrameRow` supports the iteration interface and can therefore be passed to
-functions that expect a collection as an argument.
+functions that expect a collection as an argument. In such situations `eltype`
+of `DataFrameRow` is always `Any`.
 
 Indexing is one-dimensional like specifying a column of a `DataFrame`.
 You can also access the data in a `DataFrameRow` using the `getproperty` and
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -271,7 +271,13 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
-Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(dfr))}(values(dfr))
+function Base.NamedTuple(dfr::DataFrameRow)
+    k = keys(dfr)
+    v = values(dfr)
+    n = names(dfr)
+    s = eltype.(parent(dfr)[!, ni] for ni in n)
+    NamedTuple{Tuple(k), Tuple{s...}}(Tuple(v))
+end
 
 """
     copy(dfr::DataFrameRow)
--- a/test/dataframerow.jl
+++ b/test/dataframerow.jl
@@ -270,7 +270,7 @@ end
     df = DataFrame(a=[1, missing, missing], b=[2.0, 3.0, 0.0])
     dfr = DataFrameRow(df, 1, :)
     nt = first(Tables.namedtupleiterator(df))
-    @test copy(dfr) === nt === NamedTuple{(:a, :b),Tuple{Union{Missing, Int64},Float64}}((1, 2.0))
+    @test copy(dfr) === nt === NamedTuple{(:a, :b),Tuple{Union{Missing, Int},Float64}}((1, 2.0))
     @test sum(skipmissing(copy(df[3, :]))) == 0
     @test convert(Vector, dfr)::Vector{Union{Float64, Missing}} == [1.0, 2.0]
     @test convert(Vector{Int}, dfr)::Vector{Int} == [1, 2]
--- a/test/dataframerow.jl
+++ b/test/dataframerow.jl
@@ -209,7 +209,7 @@ end
     @test propertynames(r) == keys(r) == Symbol.(names(df))
     @test r.a === 1
     @test r.b === 2.0
-    @test copy(r[[:a,:b]]) === (a=1, b=2.0)
+    @test copy(r[[:a,:b]]) == (a=1, b=2.0)
 
     r.a = 2
     @test r.a === 2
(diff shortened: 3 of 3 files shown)


Project.toml
```diff
@@ -9,30 +9,19 @@ DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 Future = "9fa8497b-333b-5362-9e8d-4d0656e87820"
 InvertedIndices = "41ab1584-1d38-5bbf-9106-f11c6c58b48f"
 IteratorInterfaceExtensions = "82899510-4779-5014-852e-03e436cf321d"
+Lazy = "50d2b5c4-7a5e-59d5-8109-a42b560f39c0"
```
bkamins, line 12: I do not think we depend on Lazy.jl - right?

src/abstractdataframe/iteration.jl
```diff
@@ -30,6 +30,10 @@ Base.iterate(::AbstractDataFrame) =
 Return a `DataFrameRows` that iterates a data frame row by row,
 with each row represented as a `DataFrameRow`.
 
+Because `DataFrameRow`s have an `eltype` of `Any`, use `copy(dfr::DataFrameRow)` to obtain 
+a named tuple, which supports iteration and property access like a `DataFrameRow`, 
+but also passes information of the `eltypes` of the columns of `df`.
```
bkamins, line 35: this is a good idea but then the definition of:
```
Base.copy(r::DataFrameRow) = NamedTuple{Tuple(keys(r))}(values(r))
```
should be updated to reflect it (we should also pass `eltypes` of columns when constructing a `NamedTuple`).

Now it does not work as you describe.

When you add it please also add a test of this behavior.

bkamins:
Thank you for the PR. I have left two small comments (and probably @nalimilan will want to have a look at it also).

pdeffebach:
I am working on this, and there is a problem. Check this out

```
julia> nt1 = NamedTuple{(:a, :b),Tuple{Union{Missing, Int64},Float64}}((1, 2.0))
NamedTuple{(:a, :b),Tuple{Union{Missing, Int64},Float64}}((1, 2.0))

julia> nt2 = NamedTuple{(:a, :b),Tuple{Union{Missing, Int64},Int64}}((1, 2))
NamedTuple{(:a, :b),Tuple{Union{Missing, Int64},Int64}}((1, 2))

julia> eltype(nt1)
Any

julia> eltype(nt2)
Union{Missing, Int64}
```

This means that, for example, with `sum`, you get an error saying `no method matching `zero(::Any)`. 

This error pops up with `Tables.namedtupleiterator` as well, so it fails with `select` as well. 

```
julia> df2
3×3 DataFrame
│ Row │ a       │ c       │ d        │
│     │ Int64?  │ Int64?  │ Float64? │
├─────┼─────────┼─────────┼──────────┤
│ 1   │ 1       │ 4       │ 1.0      │
│ 2   │ missing │ 5       │ 2.0      │
│ 3   │ missing │ missing │ missing  │

julia> transform(df2, AsTable([:a, :d]) => ByRow(sum ∘ skipmissing))
ERROR: MethodError: no method matching zero(::Type{Any})
```

I suppose the solution here means fixing `skipmissing`, wouldn't it? cc @nalimilan because it is a `missing` and `skipmissing` problem, ultimately.

bkamins:
The core reason is how `typejoin` in Base works:
```
julia> typejoin(Int64, Union{Missing, Float64})
Any
```
and I do not think it would be changed as messing with this would be breaking.

Though we have:
```
julia> typejoin(Int64, Float64)
Real
```
so it would make sense just to propagate `Missing` in the above case also to produce `Union{Missing, Real}`.

@StefanKarpinski  - do you have an opinion on this?

Anyway - this will not alter what we should implement in DataFrames.jl so you can go ahead with updating the PR and count that the problem you mention gets resolved elsewhere.

nalimilan:
Good catch. I've filed https://github.com/JuliaLang/julia/issues/35504 to see whether we can improve `promote_typejoin`. See also https://github.com/JuliaData/SplitApplyCombine.jl/issues/22.

A workaround would be to add an `init` argument to `sum`, `mean`, etc. to allow specifying the kind of zero you want. But that would be cumbersome for users (and require changing many functions).

StefanKarpinski:
Propagating missing support for unions like this seems like a good idea. However, have you considered allowing the broadening of columns when people assign `missing`? I.e. you could allow `df.col[i] = missing` to work even if `df.col` has type `Vector{Int}` for example, by transparently replacing `df.col` with a widened copy of type `Vector{Union{Int,Missing}}`. That seems like it would be a big usability boost since people wouldn't have to worry so much about the distinction. You could safely load data that doesn't actually have missing values without support for them and then if someone needs to write a `missing` later, they still can.

pdeffebach:
> I.e. you could allow `df.col[i] = missing`

This wouldn't work because of our guarantee that `df.col` is *just* a concrete, normal, vector type. We need people to have 

```
function foo(v::Vector{Float64})
###
end

v = df[!, :col]
foo(f)
```

If we allow `v[i] = missing` then we are being inconsistent with the promotion rules of of Vectors. 

We could, however, make `allowmissing` the defaut. But then when adding a `Vector{Float64}` to the data frame, we would have to promote it's type, copying it.

nalimilan:
Actually we can't allow that since `df.col[i] = v` is parsed as `setindex!(getproperty(df, col), v, i)`. I guess it could be changed to parse to a function which would default to doing that. What we could allow is `df[col, i] = v` to automatically promote. But as @pdeffebach says I'm not sure it would be a good idea.

Anyway that's completely orthogonal to this issue, right? :-)

pdeffebach:
The motivation for this PR is very niche, but nonetheless ready for review.

src/dataframerow/dataframerow.jl
```diff
@@ -256,7 +257,13 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
-Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(dfr))}(values(dfr))
+function Base.NamedTuple(dfr::DataFrameRow)
+    k = keys(dfr)
+    v = values(dfr)
+    n = names(dfr)
+    s = eltype.(parent(dfr)[!, ni] for ni in n)
```
bkamins, line 264: this will be inefficient - better use integer indexing or `parentcols`

src/dataframerow/dataframerow.jl
```diff
@@ -256,7 +257,13 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
-Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(dfr))}(values(dfr))
+function Base.NamedTuple(dfr::DataFrameRow)
+    k = keys(dfr)
```
bkamins, line 261: `keys` will return a `Tuple` of `Symbol`s after #2199. So in general this PR should wait till we merge the other one as that one is much bigger

src/abstractdataframe/iteration.jl
nalimilan, line 35:
```suggestion
but also passes information on the `eltypes` of the columns of `df`. 
```

src/dataframerow/dataframerow.jl
```diff
@@ -14,7 +14,8 @@ DataFrameRow(parent::AbstractDataFrame, row::Integer, cols=:)
 ```
 
 A `DataFrameRow` supports the iteration interface and can therefore be passed to
-functions that expect a collection as an argument.
+functions that expect a collection as an argument. In such situations `eltype`
+of `DataFrameRow` is always `Any`.
```
nalimilan, line 18:
```suggestion
functions that expect a collection as an argument. Its element type is always `Any`.
```
That's always the case, so "in such situations" doesn't seem appropriate to me.)

bkamins:
@pdeffebach - can you please rebase this PR against master (as #2199 was merged) and integrate the comments, so that we can move forward with this? Thank you!

src/dataframerow/dataframerow.jl
bkamins: In the end #2199 `keys` returns a `Vector`, but it makes a copy, so probably it is better to write:
```suggestion
    k = Tuple(_names(dfr))
```

src/dataframerow/dataframerow.jl
```diff
@@ -256,7 +257,13 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
 Base.broadcastable(::DataFrameRow) =
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
-Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(dfr))}(values(dfr))
+function Base.NamedTuple(dfr::DataFrameRow)
+    k = keys(dfr)
+    v = values(dfr)
```
bkamins, line 262:
```suggestion
    v = ntuple(i -> dfr[i], length(dfr))
```

src/dataframerow/dataframerow.jl
bkamins, line 264:
```suggestion
    pc = parentcols(index(dfr))
    cols = _columns(parent(dfr))
    s = ntuple(i -> eltype(cols[pc[i]]), length(dfr))
```

src/dataframerow/dataframerow.jl
```diff
@@ -256,7 +257,13 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
...
+    n = names(dfr)
+    s = eltype.(parent(dfr)[!, ni] for ni in n)
+    NamedTuple{Tuple(k), Tuple{s...}}(Tuple(v))
```
bkamins, line 265:
```suggestion
    NamedTuple{k, s...}(v)
```

bkamins:
@pdeffebach - I have made some suggestions how to improve the performance of your proposal (of course I might be wrong so it would be good to see a benchmark of one vs the other) + tests (in particular an important test is if `DataFrameRow` has a subset of columns from a `SubDataFrame` that itself had a subset of columns from a parent `DataFrame`).

src/dataframerow/dataframerow.jl
```diff
@@ -270,7 +270,14 @@ Base.get(f::Base.Callable, dfr::DataFrameRow, key::ColumnIndex) =
...
     throw(ArgumentError("broadcasting over `DataFrameRow`s is reserved"))
 
-Base.NamedTuple(dfr::DataFrameRow) = NamedTuple{Tuple(keys(dfr))}(values(dfr))
+function Base.NamedTuple(dfr::DataFrameRow)
+    k = Tuple(_names(dfr))
+    v = ntuple(i -> dfr[i], length(dfr))
+    pc = parentcols(index(dfr))
+    cols = _columns(parent(dfr))
+    s = ntuple(i -> eltype(cols[pc[i]]), length(dfr))
+    NamedTuple{k, s...}(v)
```
bkamins, line 279:
```suggestion
    NamedTuple{k, Tuple{s...}}(v)
```
It was my mistake. Sorry for this. I have benchmarked both and it is only minimally slower than the previous version, so this seems OK:
```
julia> dfr
DataFrameRow
│ Row │ x      │ y      │
│     │ Int64? │ Int64? │
├─────┼────────┼────────┤
│ 1   │ 1      │ 2      │

julia> @benchmark NamedTuple($dfr)
BenchmarkTools.Trial: 
  memory estimate:  480 bytes
  allocs estimate:  9
  --------------
  minimum time:     814.301 ns (0.00% GC)
  median time:      824.548 ns (0.00% GC)
  mean time:        884.224 ns (3.93% GC)
  maximum time:     54.187 μs (98.25% GC)
  --------------
  samples:          10000
  evals/sample:     83

julia> @benchmark fun($dfr)
BenchmarkTools.Trial: 
  memory estimate:  496 bytes
  allocs estimate:  11
  --------------
  minimum time:     904.462 ns (0.00% GC)
  median time:      924.692 ns (0.00% GC)
  mean time:        1.032 μs (4.61% GC)
  maximum time:     126.462 μs (99.00% GC)
  --------------
  samples:          10000
  evals/sample:     39
```
(`fun` is the proposed implementation, `NamedTuple` is the old implementation)

test/dataframerow.jl
```diff
@@ -267,9 +267,11 @@ end
 end
 
 @testset "convert, copy and merge" begin
-    df = DataFrame(a=[1, missing], b=[2.0, 3.0])
+    df = DataFrame(a=[1, missing, missing], b=[2.0, 3.0, 0.0])
     dfr = DataFrameRow(df, 1, :)
-
+    nt = first(Tables.namedtupleiterator(df))
+    @test copy(dfr) === nt === NamedTuple{(:a, :b),Tuple{Union{Missing, Int64},Float64}}((1, 2.0))
```
bkamins, line 273:
```suggestion
    @test copy(dfr) === nt === NamedTuple{(:a, :b),Tuple{Union{Missing, Int},Float64}}((1, 2.0))
```
correcting as I guess without it the PR will fail on 32-bit Julia

bkamins:
@nalimilan - good to merge?

bkamins:
Thank you!

Issue #2135, opened 2020-03-02
Circular reference in DataFrame bug


bkamins:
This should be fixed
```
julia> df = DataFrame(x=Any[1])
1×1 DataFrame
│ Row │ x   │
│     │ Any │
├─────┼─────┤
│ 1   │ 1   │

julia> df[1, :x] = df
Error showing value of type DataFrame:
ERROR: StackOverflowError:
```

(though it is probably a minor thing)

bkamins:
Also:
```
julia> df[1,1] = df[1, :]
DataFrameRowError showing value of type DataFrameRow{DataFrame,DataFrames.Index}:
ERROR: StackOverflowError:
```

PR #2200, opened 2020-04-20
[BREAKING] Circular ref bug show

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2135

This is not a full fix, as Base is much better at detecting circular references, but doing it right would require a much bigger rewrite of the code. Now we correctly handle 1-level of indirection (but do not correctly handle a case when e.g. a data frame would contain a vector containing a dict containing the original data frame - but I think this case is so obscure that it can be skipped).

Additionally now we:
1. always print `missing`, `nothing` and `#undef` consistently everywhere (text, LaTeX, HTML, CSV, TSV) - the only restriction I left is CSV and TSV throwing an error when trying to write #undef
2. consistently highlight `missing`, `nothing` and `#undef` everywhere (text, LaTeX, HTML) - so they are never mixed up with strings
2. if a `AbstractDataFrame`, `DataFrameRow`, `DataFrameRows`, `DataFrameColumns` or `GroupedDataFrame` is stored in a data frame then only its summary is printed (this makes printing much nicer (otherwise the output were not readable anyway) - and as a side effect - avoids the original circular reference problem)

Finally this PR fixes a problem that old printing mechanism in DataFrames.jl was not thread safe. After fixing it is thread safe (we do not reuse one global `buffer` but allocate it anew for each top-level call of `show`).

This PR is largely orthogonal to #2199 and can be merged post 0.21 release, but before 1.0 release as it is mildly breaking.


src/abstractdataframe/show.jl
```diff
@@ -112,8 +122,9 @@ end
                             rowindices1::AbstractVector{Int},
                             rowindices2::AbstractVector{Int},
                             rowlabel::Symbol,
-                            rowid=nothing,
-                            show_eltype::Bool=true)
+                            rowid,
+                            show_eltype::Bool,
+                            buffer)
```
nalimilan, line 127:
```suggestion
                            rowid::Union{Integer, Nothing},
                            show_eltype::Bool,
                            buffer::IOBuffer)
```

src/abstractdataframe/show.jl
```diff
@@ -140,47 +151,37 @@ implicit row ID column contained in every `AbstractDataFrame`.
...
 function getmaxwidths(df::AbstractDataFrame,
                       io::IO,
                       rowindices1::AbstractVector{Int},
                       rowindices2::AbstractVector{Int},
                       rowlabel::Symbol,
-                      rowid=nothing,
-                      show_eltype::Bool=true)
+                      rowid,
+                      show_eltype::Bool,
+                      buffer)
```
nalimilan, line 165:
```suggestion
                      rowid::Union{Integer, Nothing},
                      show_eltype::Bool,
                      buffer::IOBuffer)
```

src/abstractdataframe/show.jl
```diff
@@ -328,6 +297,7 @@ required for printing have been precomputed.
 - `leftcol::Int`: The index of the first column in a chunk to be rendered.
 - `rightcol::Int`: The index of the last column in a chunk to be rendered.
 - `rowid`: Used to handle showing `DataFrameRow`.
+- `buffer`: buffer passed around to avoid reallocations in `ourstrwidth`
```
nalimilan, line 300: Also add it to signature.

src/abstractdataframe/show.jl
```diff
@@ -346,7 +316,8 @@ function showrowindices(io::IO,
                         maxwidths::Vector{Int},
                         leftcol::Int,
                         rightcol::Int,
-                        rowid)
+                        rowid,
+                        buffer)
```
nalimilan, line 320:
```suggestion
                        rowid::Union{Integer, Nothing},
                        buffer::IOBuffer)
```

src/abstractdataframe/show.jl
```diff
@@ -402,12 +371,13 @@ end
...
-             displaysummary::Bool = true,
-             eltypes::Bool = true,
-             rowid=nothing)
+             splitcols::Bool,
+             allcols::Bool,
+             rowlabel::Symbol,
+             displaysummary::Bool,
+             eltypes::Bool,
+             rowid,
+             buffer)
```
nalimilan, line 380:
```suggestion
             rowid::Union{Integer, Nothing},
             buffer::IOBuffer)
```

src/abstractdataframe/show.jl
```diff
@@ -437,6 +407,7 @@ NOTE: The value of `maxwidths[end]` must be the string width of
 - `eltypes::Bool = true`: Whether to print the column type
    under the column name in the heading. Defaults to `true`.
 - `rowid = nothing`: Used to handle showing `DataFrameRow`
+- `buffer`: buffer passed around to avoid reallocations in `ourstrwidth`
```
nalimilan, line 410:
```suggestion
- `rowid::Union{Integer, Nothing} = nothing`: Used to handle showing `DataFrameRow`
- `buffer::IOBuffer`: buffer passed around to avoid reallocations in `ourstrwidth`
```

src/abstractdataframe/show.jl
```diff
@@ -459,12 +430,14 @@ function showrows(io::IO,
...
-                  displaysummary::Bool = true,
-                  eltypes::Bool = true,
-                  rowid=nothing)
+                  splitcols::Bool,
+                  allcols::Bool,
+                  rowlabel::Symbol,
+                  displaysummary::Bool,
+                  eltypes::Bool,
+                  rowid,
+                  buffer)
```
nalimilan, line 439:
```suggestion
                  rowid::Union{Integer, Nothing},
                  buffer::IOBuffer)
```

src/abstractdataframe/io.jl
```diff
@@ -134,11 +139,23 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
         end
         for column_name in cnames
             if isassigned(df[!, column_name], row)
-                cell = sprint(ourshow, df[row, column_name])
+                cell_val = df[row, column_name]
+                if ismissing(cell_val)
+                    write(io, "<td><em>missing</em></td>")
+                elseif isnothing(cell_val)
+                    write(io, "<td><em>nothing</em></td>")
```
nalimilan, line 146: It was nice to print `nothing` as the empty string to make the output of `decribe` more readable (and similar use cases outside of DataFrames). That's also how `nothing` is printed at the REPL.

bkamins:
> I wonder whether we couldn't just adapt the `show` method for data frames so that they check `:compact` or `:limited` and show only the summary in that case.

I would leave it for a separate PR (as it probably will lead to a cascade of other things to check - I am just not sure that I can get it 100% right then - these `show`-related PRs are hairy if we want to consistently support all MIME types).

src/abstractdataframe/io.jl
```diff
@@ -134,11 +139,23 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
...
+                elseif cell_val isa SHOW_TABULAR_TYPES
+                    write(io, "<td><em>")
```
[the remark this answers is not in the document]
bkamins: I use italics everywhere where a printed value can be confused with a string being printed and I think this is a case here (though admittedly very unlikely). I would leave it as is.

src/abstractdataframe/io.jl
```diff
@@ -134,11 +139,23 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
         end
         for column_name in cnames
             if isassigned(df[!, column_name], row)
-                cell = sprint(ourshow, df[row, column_name])
+                cell_val = df[row, column_name]
+                if ismissing(cell_val)
+                    write(io, "<td><em>missing</em></td>")
```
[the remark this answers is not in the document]
bkamins: I Jupyter every second row is printed grey (see the picture below) so I think it would be problematic.

Screenshot from 2020-04-22 10-43-35

src/abstractdataframe/show.jl
```diff
@@ -25,25 +20,40 @@ Render a value to an `IO` object compactly and omitting type information, by
...
+    sx = sprint(show, "text/plain", x, context=io_ctx)
 
     # If the output contains line breaks, try 2-arg show instead.
     if occursin('\n', sx)
-        sx = sprint(show, x, context=io)
+        sx = sprint(show, x, context=io_ctx)
     end
 
-    print(io, sx)
+    #
```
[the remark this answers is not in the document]
bkamins: I have added a comment

src/abstractdataframe/show.jl
bkamins: added

src/abstractdataframe/io.jl
bkamins: Ah - describe is a good point:
```
julia> df = DataFrame(a=1:2, b=["1", "2"], c=[1,nothing], d=[missing, 1])
2×4 DataFrame
│ Row │ a     │ b      │ c       │ d       │
│     │ Int64 │ String │ Union…  │ Int64?  │
├─────┼───────┼────────┼─────────┼─────────┤
│ 1   │ 1     │ 1      │ 1       │ missing │
│ 2   │ 2     │ 2      │ nothing │ 1       │

julia> describe(df)
4×8 DataFrame
│ Row │ variable │ mean    │ min     │ median  │ max     │ nunique │ nmissing │ eltype                │
│     │ Symbol   │ Union…  │ Any     │ Union…  │ Any     │ Union…  │ Union…   │ Type                  │
├─────┼──────────┼─────────┼─────────┼─────────┼─────────┼─────────┼──────────┼───────────────────────┤
│ 1   │ a        │ 1.5     │ 1       │ 1.5     │ 2       │ nothing │ nothing  │ Int64                 │
│ 2   │ b        │ nothing │ 1       │ nothing │ 2       │ 2       │ nothing  │ String                │
│ 3   │ c        │ nothing │ nothing │ nothing │ nothing │ 2       │ nothing  │ Union{Nothing, Int64} │
│ 4   │ d        │ 1.0     │ 1       │ 1.0     │ 1       │ nothing │ 1        │ Union{Missing, Int64} │
```

But note that `nothing` is printed in REPL if it is stored in a container:
```
julia> [1, nothing]
2-element Array{Union{Nothing, Int64},1}:
 1
  nothing

julia> (1, nothing)
(1, nothing)
```
and that is why I have made this change.

Given this what do you think would be best? (I would be OK to leave the old behavior and print nothing)

bkamins:
All suggestions are applied. I think the major decision is what to do with `nothing`.

src/abstractdataframe/io.jl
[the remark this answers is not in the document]
nalimilan: Actually aren't tables the kind of object which is the less likely to be confused with a string? They always include the name of the type. Conversely, many custom types that we don't print in italics could be confused with strings.

There may be other reasons to print table differently from other values though, e.g. to highlight nesting.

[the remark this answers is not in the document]
bkamins: This and the fact that this is non-standard printing of the object (for other objects we use normal `show`). Anyway - I think this is a minor thing so I can change it if you prefer.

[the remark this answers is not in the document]
nalimilan: Ah but as I said I'd love to have this be the standard `show` with `:compact => true`. Exceptions are bad. ;-)

src/abstractdataframe/io.jl
nalimilan: I vote to continue printing `nothing` as the empty string. The inconsistency with other containers isn't great, but well...

bkamins: OK - I will change it.

src/abstractdataframe/io.jl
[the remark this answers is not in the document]
bkamins: So do we:
1. change to always use `:compact => true` when rendering cells
2. do not special case tabular data, but define a special printing path for them when `:compact => true`

?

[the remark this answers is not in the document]
bkamins: Actually such a change will also resolve the problem when data frame is deeply nested in another container that the current solution did not solve.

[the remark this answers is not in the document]
bkamins: As `:context` is not fully settled in Base, for the time being we keep what is proposed in this PR.

bkamins:
Thank you. I will merge it after #2911 is done.

PR #2209, opened 2020-04-24
Fix exponent in docs

nalimilan:
Documenter doesn't seem to support this.

### DIFF
--- a/docs/src/man/categorical.md
+++ b/docs/src/man/categorical.md
@@ -80,7 +80,7 @@ julia> sort(cv)
 
 ```
 
-By default, a `CategoricalArray` is able to represent 2<sup>32</sup> different levels.
+By default, a `CategoricalArray` is able to represent ``2^32`` different levels.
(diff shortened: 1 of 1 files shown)


docs/src/man/categorical.md
```diff
@@ -80,7 +80,7 @@ julia> sort(cv)
 
 ```
 
-By default, a `CategoricalArray` is able to represent 2<sup>32</sup> different levels.
+By default, a `CategoricalArray` is able to represent ``2^32`` different levels.
```
bkamins, line 83: what do double backquotes mean?

bkamins: @nalimilan - bump - I would like to understand the change before merging :)

nalimilan: That's the syntax for equations. It's too bad we need such an advanced feature for this, but I guess it's better than using Unicode superscripts?

Issue #2206, opened 2020-04-23
Cleaner syntax


matthieugomez:
I've followed a bit the recent updates of this package. This is very impressive — thanks @bkamins 
 and @nalimilan for all your work.

I have one comment about the syntax. An operation such as computing the mean of a variable in a dataframe can be classified along two dimensions (i) whether the new dataframe as the same size as the original dataframe (ii) whether it is a by operation or not.

dplyr and stata make it very easy to alternate across these two dimensions.

in dplyr 

| mutate(df, mean(x))                | mutate(group_by(df, id), mean(x))       |
|----------------------------- |---------------------------------------|
| summarize(d, mean(x))          | summarize(group_by(df, id), mean(x)) |

In stata:

| egen  x = mean(x)                   | egen x = mean(x), by(id)                    |
|------------------------------|-------------------------------------|
| collapse (mean) x                    | collapse (mean) x, by(id)                   |

This is very neat. There is a symmetry between top vs bottom, and left vs right. People can understand what these commands do just by reading them

The current syntax of Dataframes.jl is not as neat IMO. On the current master, we have:

| select(df, :, :x => mean)          | by(df, :id, :, :x => mean)                   |
|------------------------------|-----------------------------------|
| select(df, :x => mean)          | by(df, :id, :x => mean)                      |

I wish DataFrames.jl would follow the example of  dplyr and stata here. For instance, a nice syntax could be the following:

| transform(df, :x => mean)       | transform(df, :x => mean, by = :id)                |
|------------------------------|---------------------------------------------|
| aggregate(df, :x => mean)      | aggregate(df, :x => mean, by = :id)                |

In short, my suggestion would be to (i) remove the function `by` (ii) allow  `by` kwarg in `transform` (ii) define a new function that would be the Julia equivalent of summarize (dplyr)/ collapse (stata).

bkamins:
This is essentially https://github.com/JuliaData/DataFrames.jl/issues/2172, which will be added later (as it is non-breaking). We leave `by` for backward compatibility reasons.

As for `mutate` vs `summarize` - why do you see that this distinction is needed? If I understand the functionality there correctly we already provide it with a single function (currently `select`/`transform` and `combine`/`by`).

bkamins:
[...]

PR #2210, opened 2020-04-25
Proposal for `gen` and `collapse`

pdeffebach:
This is a very naive implementation of @bkamins 's plan for `collapse`, etc. described [here](https://github.com/JuliaData/DataFrames.jl/issues/2206#issuecomment-619390620). 

Please keep comments focused on #2206. This is just to demonstrate how I imagine this should work.

### DIFF
--- a/src/gen_proposal.jl
+++ b/src/gen_proposal.jl
@@ -1,0 +1,46 @@
+function gen(df::AbstractDataFrame, args...; kwargs...)
+	transform(df, args...; kwargs...)
+end
+
+function genby(df::AbstractDataFrame, id, args...; kwargs...)
+	by(df, id, :, args...; kwargs...)
+end
+
+function keep(df::AbstractDataFrame, args...; kwargs...)
+		t = gensym()
+	df[!, t] = df[!, 1]
+	ndf = select(df, t, args...; kwargs...)
+	return select(ndf, Not(t))
+end
+
+function keepby(df::AbstractDataFrame, id, args...; kwargs...)
+	t = gensym()
+	df[!, t] = df[!, 1]
+	ndf = by(df, id, t, args...; kwargs...)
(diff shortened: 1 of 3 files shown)


bkamins:
OK - I treat it as a reference to a discussion. This PR should not be merged and will be closed when #2206 is resolved.

pdeffebach:
I will update this PR as consensus develops. When we all agree we can implement it for real.

bkamins:
OK, but as @nalimilan commented in #2206 I do not expect that we will allow to define so many functions in DataFrames.jl. We will rather keep a minimal list of functions that cover all required functionality. The functions you propose are good for an extension package to DataFrames.jl so that users can opt-in to use them.

src/gen_proposal.jl
```diff
@@ -0,0 +1,46 @@
+function gen(df::AbstractDataFrame, args...; kwargs...)
+	transform(df, args...; kwargs...)
```
bkamins, line 2: first small note (just for the future discussions) - this is not a correct implementation as `args` could overwrite all source columns of `df`.

src/gen_proposal.jl
```diff
@@ -0,0 +1,46 @@
...
+end
+
+function genby(df::AbstractDataFrame, id, args...; kwargs...)
+	by(df, id, :, args...; kwargs...)
+end
+
+function keep(df::AbstractDataFrame, args...; kwargs...)
+		t = gensym()
```
bkamins, line 10: a second small note `gensym()` cannot be used here as it does not guarantee we will not have column name conflicts. But again - as above this is a minor issue than can be fixed.

bkamins:
I am closing this as 0.21 was released. If you decide to put these definitions in some accompanying package ping me and I can help with the review. Thank you for working on this!

PR #2212, opened 2020-04-27
Make ByRow subtype Function

oxinabox:
`ByRow` is a function wrapper.
Its callable, so it works like a function.
As such it is not unreasonable to for it to subtype `Function`
Which means it is also covered by `Base.Callable`.
Which means we can simplify some of the more complex type constraints.
(I noticed this because these get more complex in #2199 )

In general I think if something is callable and doesn't already have a supertype, it is a fairly good idea to make it subtype `Function`.

(i would be in general a fan of dropping the [constraints here entirely](https://white.ucc.asn.au/2020/04/19/Julia-Antipatterns.html#dispatching-on--function), but that would require closer looking at the code to see if it would cause any change in dispatch)

### DIFF
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -102,8 +102,7 @@ function normalize_selection(idx::AbstractIndex,
 end
 
 normalize_selection(idx::AbstractIndex,
-                    sel::Pair{<:Any,<:Pair{<:Union{Base.Callable, ByRow},
-                                           <:AbstractString}}) =
+                    sel::Pair{<:Any,<:Pair{<:Base.Callable,<:AbstractString}}) =
     normalize_selection(idx, first(sel) => first(last(sel)) => Symbol(last(last(sel))))
 
 function normalize_selection(idx::AbstractIndex,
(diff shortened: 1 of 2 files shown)


bkamins:
Thank you. Makes sense. Is it rebased against master (I have just merged #2199). If not can you please rebase?

bkamins:
If we go for it also remove line `Base.broadcastable(x::ByRow) = Ref(x)` as `Function` is broadcastable in Base already.

oxinabox:
done

src/abstractdataframe/selection.jl
```diff
@@ -102,28 +99,27 @@ function normalize_selection(idx::AbstractIndex,
...
+                             sel::Pair{<:Any, <:Base.Callable})
     if first(sel) isa AsTable
         rawc = first(sel).cols
         wanttable = true
     else
         rawc = first(sel)
         wanttable = false
     end
-    if rawc isa AbstractVector{Int}
+    if rawc isa AbstractVector{Int}
```
nalimilan, line 122:
```suggestion
    if rawc isa AbstractVector{Int}
```

bkamins:
Merged. Thank you!

PR #2213, opened 2020-04-27
cleanup after string PR

bkamins:
This is a PR implementing the suggestions of @oxinabox to #2199.

@nalimilan - the major comment was if we want to allow `AbstractVector{<:SymbolOrString}`
in place of `Union{AbstractVector{Symbol}, AbstractVector{<:AbstractString}}` (in technical terms, as in some places in the code this has a different manifestation, but essentially - I was defensive in #2199 and disallowed mixing `Symbol`s and strings, and @oxinabox suggests to allow this).

(as a note - my preference is to disallow mixing - as I do not see a use case where it is really useful)

@oxinabox - note that currently we also disallow mixing integers and `Symbol`s or strings (of course this is not the same, but ...).

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -45,7 +45,7 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
-Columns can be directly (i.e. without copying) accessed via `df.col`, `df."col"`, `df[!, :col]` or `df[!, "col"]`. The two latter syntaxes are more flexible as they allow passing a variable holding the name of the column, and not only a literal name. Note that column names can be either symbols (written as `:col`, `:var"col"` or `Symbol("col")`) or strings (written as `"col"`; in particular variable interpolation using `$` is not allowed).
+Columns can be directly (i.e. without copying) accessed via `df.col`, `df."col"`, `df[!, :col]` or `df[!, "col"]`. The two latter syntaxes are more flexible as they allow passing a variable holding the name of the column, and not only a literal name. Note that column names can be either symbols (written as `:col`, `:var"col"` or `Symbol("col")`) or strings (written as `"col"`). Note that in the forms `df."col"` and `:var"col"` variable interpolation into a string using `$` does not work.
 Columns can also be accessed using an integer index specifying their position.
 
 Since `df[!, :col]` does not make a copy, changing the elements of the column vector returned by this syntax will affect the values stored in the original `df`. To get a copy of the column use `df[:, :col]`: changing the vector returned by this syntax does not change `df`.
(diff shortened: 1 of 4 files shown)


docs/src/man/getting_started.md
```diff
@@ -45,7 +45,7 @@ julia> df = DataFrame(A = 1:4, B = ["M", "F", "F", "M"])
 
 ```
 
-Columns can be directly (i.e. without copying) accessed via `df.col`, `df."col"`, `df[!, :col]` or `df[!, "col"]`. The two latter syntaxes are more flexible as they allow passing a variable holding the name of the column, and not only a literal name. Note that column names can be either symbols (written as `:col`, `:var"col"` or `Symbol("col")`) or strings (written as `"col"`).
+Columns can be directly (i.e. without copying) accessed via `df.col`, `df."col"`, `df[!, :col]` or `df[!, "col"]`. The two latter syntaxes are more flexible as they allow passing a variable holding the name of the column, and not only a literal name. Note that column names can be either symbols (written as `:col`, `:var"col"` or `Symbol("col")`) or strings (written as `"col"`; in particular variable interpolation using `$` is not allowed).
```
nalimilan, line 48: This isn't correct for `df[!, col]`.

TBH I'm not sure this really needs to be mentioned here, this is the "getting started" page.

nalimilan:
> @nalimilan - the major comment was if we want to allow `AbstractVector{<:SymbolOrString}`
> in place of `Union{AbstractVector{Symbol}, AbstractVector{<:AbstractString}}` (in technical terms, as in some places in the code this has a different manifestation, but essentially - I was defensive in #2199 and disallowed mixing `Symbol`s and strings, and @oxinabox suggests to allow this).
> 
> (as a note - my preference is to disallow mixing - as I do not see a use case where it is really useful)
> 
> @oxinabox - note that currently we also disallow mixing integers and `Symbol`s or strings (of course this is not the same, but ...).

I agree that we'd better disallow mixing for now. We have more important features to focus on and this can be changed after 1.0.

docs/src/man/getting_started.md
bkamins: right - `$` is disallowed only in `getproperty`/`setproperty!`. @oxinabox - are you OK to skip commenting on this? (in general the behaviour is not DataFrames.jl related but just how Julia Base works)

bkamins:
We agreed with @oxinabox not to allow mixing now on Slack. So we are all on the same page here now.

docs/src/man/getting_started.md
oxinabox: I think it is important.
this is going to to be the first time most people encounter a string in this position.
And I would naturally assume that interpolation would work in this string unless told otherwise.

nalimilan: Well Julia will tell them otherwise quite soon if they think it works... :-)

bkamins: I think we can keep it. I have also added a comment that interpolation does not work for `:var"a$i"`.

bkamins:
@oxinabox - thank you for the earlier review. As you can see from the last commit it even prompted me to fix one bug that I have overlooked earlier.

PR #2214, opened 2020-04-27
[BREAKING] new design of select, transform and combine

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2206.

This is still a draft. I have only implemented the `AbstractDataFrame` part now (`GroupedDataFrame` is pending) and not updated the documentation and tests. But you can have a look at the code to see what changes we essentially agreed to do.


bkamins:
@matthieugomez, @pdeffebach, @nalimilan - the PR should be good to have a quick look (and tests - maybe some corner cases that are incorrect will be caught).

All functions should be implemented as we discussed.

I have not updated tests nor documentation so there might be some holes, but here is how it works:
```
julia> df = DataFrame(g=[2,3,1,1,2,2,3,1,2,1], x=1:10)
10×2 DataFrame
│ Row │ g     │ x     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 2     │ 1     │
│ 2   │ 3     │ 2     │
│ 3   │ 1     │ 3     │
│ 4   │ 1     │ 4     │
│ 5   │ 2     │ 5     │
│ 6   │ 2     │ 6     │
│ 7   │ 3     │ 7     │
│ 8   │ 1     │ 8     │
│ 9   │ 2     │ 9     │
│ 10  │ 1     │ 10    │

julia> gdf = groupby(df, :g);

julia> select(gdf, :g, :g => :g1, :g => (x->x) => :g2, :g => mean, :x, :x => :x1, :x => (x->x) => :x2, :x => mean)
10×8 DataFrame
│ Row │ g     │ g1    │ g2    │ g_mean  │ x     │ x1    │ x2    │ x_mean  │
│     │ Int64 │ Int64 │ Int64 │ Float64 │ Int64 │ Int64 │ Int64 │ Float64 │
├─────┼───────┼───────┼───────┼─────────┼───────┼───────┼───────┼─────────┤
│ 1   │ 2     │ 2     │ 2     │ 2.0     │ 1     │ 1     │ 1     │ 5.25    │
│ 2   │ 3     │ 3     │ 3     │ 3.0     │ 2     │ 2     │ 2     │ 4.5     │
│ 3   │ 1     │ 1     │ 1     │ 1.0     │ 3     │ 3     │ 3     │ 6.25    │
│ 4   │ 1     │ 1     │ 1     │ 1.0     │ 4     │ 4     │ 4     │ 6.25    │
│ 5   │ 2     │ 2     │ 2     │ 2.0     │ 5     │ 5     │ 5     │ 5.25    │
│ 6   │ 2     │ 2     │ 2     │ 2.0     │ 6     │ 6     │ 6     │ 5.25    │
│ 7   │ 3     │ 3     │ 3     │ 3.0     │ 7     │ 7     │ 7     │ 4.5     │
│ 8   │ 1     │ 1     │ 1     │ 1.0     │ 8     │ 8     │ 8     │ 6.25    │
│ 9   │ 2     │ 2     │ 2     │ 2.0     │ 9     │ 9     │ 9     │ 5.25    │
│ 10  │ 1     │ 1     │ 1     │ 1.0     │ 10    │ 10    │ 10    │ 6.25    │
```

matthieugomez:
I’ve tried it a little bit and it looks very good from a user perspective!

bkamins:
@nalimilan - there is one thing on top of what was my idea in #2206 (and something you said you wanted to allow anyway).

We should treat `GroupedDataFrame` as canonical irrespective of its ordering, as long as it contains all groups. I guess no one will oppose this :smile:, as it was only me who wanted to be strict. The reason for this is that `groupby` automatically sorts `GroupedDataFrame` - even if not asked for - if all grouping columns are `CategoricalVector`. I will try to make the changes needed to reflect this fact.

bkamins:
I have finished the development and tests of the new functionality.
In the process some optimizations of the old code to reduce copying of the data were done.

Documentation is half-finished (i.e. I tried to update it everywhere but for sure it is not perfect). Maybe even we should agree to make a separate PR for polishing it later (but if you would have good suggestions now of course I am open for them).

src/abstractdataframe/selection.jl
```diff
@@ -514,22 +532,71 @@ julia> select(df, AsTable(:) => ByRow(mean))
...
+select(df::AbstractDataFrame, args...; copycols::Bool=true) =
+    _manipulate(df, args..., copycols=copycols, keeprows=true)
+
+"""
+    transform(df::AbstractDataFrame, args...; copycols::Bool=true)
+
+Create a new data frame that contains columns from `df` and adds columns
+specified by `args` and return it.
+The result is guaranteed to have the same number of rows as `df`.
+Equivalent to `select(df, :, args..., copycols=copycols)`.
```
matthieugomez, line 544: I would remove it actually — I don’t think this is particularly helpful.

docs/src/man/split_apply_combine.md
```diff
@@ -6,10 +6,24 @@ framework for handling this sort of computation is described in the paper
...
 written by Hadley Wickham.
 
-The DataFrames package supports the split-apply-combine strategy through the `by`
-function, which is a shorthand for `groupby` followed by `map` and/or `combine`.
-`by` takes in three arguments: (1) a `DataFrame`, (2) one or more columns to split
-the `DataFrame` on, and (3) a specification of one or more functions to apply to
+The DataFrames package supports the split-apply-combine strategy through the
+`combine`, `select`/`select!` and `transform`/`transform!` functions.
+
+In order to perform operations by groups you first need to create a `GroupedDataFrame`
```
matthieugomez, line 12: I would put groupeddataframe at the end. The beauty of the current syntax is that one can define transform/select/combine and then say “and by the way, they are all defined on groupedDataFrame”

docs/src/man/split_apply_combine.md
```diff
@@ -6,10 +6,24 @@ framework for handling this sort of computation is described in the paper
...
+object from your data frame using `groupby` function that takes two arguments:
+(1) a data frame to be grouped, and (2) a set of columns to group by.
+
+The differences between the above functions are the following:
+* `select`: return a data frame with the number and order of rows exactly the same
+  as the source, preserve only columns that have been calculated;
+* `transform`: return a data frame with the number and order of rows exactly the same
+  as the source, preserve all columns from the source and columns that have been calculated;
+* `select!`: is an in-place version of `select`;
```
matthieugomez, line 21: I would put this in the same paragraph as select. Same thing for transform!

src/abstractdataframe/selection.jl
bkamins: But what do you propose to add instead? We should give some precise specification how the function works.

docs/src/man/split_apply_combine.md
bkamins: The problem with the current manual is that this is split_apply_combine.md section.

docs/src/man/split_apply_combine.md
bkamins: OK - changed.

bkamins:
> I will do a pull request with dplyr/stata comparison, following the tables we did in the issue

That would be great.

In general your comments show what I have felt when updating the documentation - we should probably totally reorganize the manual.

But I think - I am not sure if you agree - that probably it is best done as a separate PR - focusing solely on the documentation enchancements. The reason is two fold:
* this PR is already large (1300 LOC added) and reworking documentation is usually very heavy on LOC changes (which will make it even harder to review)
* in this way the work of other contributors can be reflected in the repo (I value very much the discussions have in #2206 and in general, so it would be great to have the contributors reflected in the commit history of the repository)
* we can make a release of 0.21 quicker and then work on the documentation updates which would go in as 0.21.1 release that would contain a better documentation

If we agreed to this then I would propose:
1. to make minimal (but still correct) changes to the docstrings and the manual.
2. have later a PR/PRs that will make the docstrings and the manual much better than it is now.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -407,31 +251,33 @@ const KWARG_PROCESSING_RULES =
...
+    combine(fun::Union{Function, Type}, gd::GroupedDataFrame;
+            keepkeys::Bool=true, regroup::Bool=false)
+    combine(pair::Pair, gd::GroupedDataFrame; keepkeys::Bool=true, regroup::Bool=false)
+    combine(fun::Union{Function, Type}, df::AbstractDataFrame)
+    combine(pair::Pair, df::AbstractDataFrame)
 
 Transform a [`GroupedDataFrame`](@ref) into a `DataFrame`.
 
+As a special case if `combine` is passed an `AbstractDataFrame` it applies `fun`
+or `pair` to the passed data frame as a whole.
```
nalimilan, line 270:
```suggestion
If an `AbstractDataFrame` is passed, `fun` or `pair` is applied to the passed data frame as a whole.
```
```suggestion
Apply operations to each group in a [`GroupedDataFrame`](@ref) and return
the combined result as a `DataFrame`.
If an `AbstractDataFrame` is passed, apply operations to the data frame as a whole.
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -407,31 +251,33 @@ const KWARG_PROCESSING_RULES =
...
-The resulting data frame will be sorted if `sort=true` was passed to the
+The resulting data frame will be sorted by `keycols(gdf)` if `sort=true` was passed to the
 [`groupby`](@ref) call from which `gd` was constructed. Otherwise, ordering of rows
-is undefined.
-
-See also:
-- [`by(f, df, cols)`](@ref) is a shorthand for `combine(f, groupby(df, cols))`.
-- [`map`](@ref): `combine(f, groupby(df, cols))` is a more efficient equivalent
-  of `combine(map(f, groupby(df, cols)))`.
+follows the order of groups in `gdf`.
```
nalimilan, line 280: So rows always appear in the order of groups in `gdf`, and when `sort=true` was passed this order reflect the sorting by `keycols(gdf)`?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -598,13 +460,20 @@ function combine(gd::GroupedDataFrame, p::Pair; keepkeys::Bool=true)
...
-                 @nospecialize(cs::Union{Pair, typeof(nrow),
-                                         ColumnIndex, MultiColumnIndex}...);
-                 keepkeys::Bool=true)
+combine(gd::GroupedDataFrame,
+        cs::Union{Pair, typeof(nrow), ColumnIndex, MultiColumnIndex}...;
+        keepkeys::Bool=true, regroup::Bool=false) =
+    _combine_executor(gd, cs..., keepkeys=keepkeys, regroup=regroup,
+                      copycols=true, keeprows=false)
+
+function _combine_executor(gd::GroupedDataFrame,
```
nalimilan, line 473: The main definition I can find of "executor" is "a person who is named to carry out the terms of a will." :-D
Maybe use a more joyful term?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -693,22 +563,89 @@ end
...
+                if keeprows
+                    isequal(valscat[!, key], parent(gd)[!, key]) ||
+                    throw(ArgumentError("column :$key in returned data frame " *
+                                        "is not equal to grouping key :$key"))
+
+                else
+                    isequal(valscat[!, key], view(parent(gd)[!, key], idx)) ||
+                    throw(ArgumentError("column :$key in returned data frame " *
+                                        "is not equal to grouping key :$key"))
+                end
```
nalimilan, line 586:
```suggestion
                if (keeprows && !isequal(valscat[!, key], parent(gd)[!, key])) ||
                    (!keeprows && !isequal(valscat[!, key], view(parent(gd)[!, key], idx)))
                    throw(ArgumentError("column :$key in returned data frame " *
                                        "is not equal to grouping key :$key"))
                end
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -693,22 +563,89 @@ end
...
-                     select(valscat, Not(intersect(keys, _names(valscat))), copycols=false))
+        if keeprows
+            newparent = hcat!(select(parent(gd), gd.cols, copycols=copycols),
+                              select(valscat, Not(intersect(keys, _names(valscat))),
+                                     copycols=false), copycols=false)
+        else
+            newparent = hcat!(parent(gd)[idx, gd.cols],
+                              select(valscat, Not(intersect(keys, _names(valscat))),
+                                     copycols=false), copycols=false)
+        end
```
nalimilan, line 597:
```suggestion
        if keeprows
            newparent = select(parent(gd), gd.cols, copycols=copycols)
        else
            newparent = parent(gd)[idx, gd.cols]
        end
        hcat!(newparent, select(valscat, Not(intersect(keys, _names(valscat))), copycols=false),
              copycols=false)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -693,22 +563,89 @@ end
...
+            ends[end] = length(idx)
+
+            groups = zeros(Int, length(idx))
+            for i in 1:j
+                @inbounds for k in starts[i]:ends[i]
+                    groups[k] = i
+                end
+            end
+            # all groups must be filled
+            @assert minimum(groups) == 1
```
nalimilan, line 637: I guess this is what this is supposed to test?
```suggestion
            @assert !any(iszero, groups)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1090,10 +1027,32 @@ function _agg2idx_map_helper(idx, idx_agg)
 end
 
 function _combine(f::AbstractVector{<:Pair},
-                  gd::GroupedDataFrame, nms::AbstractVector{Symbol})
+                  gd::GroupedDataFrame, nms::AbstractVector{Symbol},
+                  copycols::Bool, keeprows::Bool) # TODO: remove these defaults
```
nalimilan, line 1031: What defaults?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1090,10 +1027,32 @@ function _agg2idx_map_helper(idx, idx_agg)
...
     # here f should be normalized and in a form of source_cols => fun
     @assert all(x -> first(x) isa Union{Int, AbstractVector{Int}, AsTable}, f)
     @assert all(x -> last(x) isa Base.Callable, f)
+
+    if keeprows
+        if !_check_cannonical(gd)
```
nalimilan, line 1037: Should be "canonical".

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1090,10 +1027,32 @@ function _agg2idx_map_helper(idx, idx_agg)
...
+            throw(ArgumentError("select or transform functions require that " *
+                                "GroupedDataFrame is not subsetted"))
```
nalimilan, line 1039:
```suggestion
            throw(ArgumentError("select and transform do not support " *
                                "GroupedDataFrames from which some groups have been dropped "*
                                "(including skipmissing=true)")))
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1090,10 +1027,32 @@ function _agg2idx_map_helper(idx, idx_agg)
...
+        let i = 0
+            for (s, e) in zip(gd.starts, gd.ends)
+                v = gd.idx[s]
+                for k in s:e
+                    i += 1
+                    idx_keeprows[i] = v
+                end
+            end
+            @assert i == nrow(parent(gd))
+        end
```
nalimilan, line 1051:
```suggestion
        i = 0
        for (s, e) in zip(gd.starts, gd.ends)
            v = gd.idx[s]
            for k in s:e
                i += 1
                idx_keeprows[i] = v
            end
        end
        @assert i == nrow(parent(gd))
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1167,30 +1135,55 @@ function _combine(f::AbstractVector{<:Pair},
...
                         agg2idx_map = _agg2idx_map_helper(idx, idx_agg)
                     end
-                    res[i] = idx, res[i][2][agg2idx_map]
+                    res[i] = idx_agg, res[i][2][agg2idx_map]
                 elseif idx != res[i][1]
-                    throw(ArgumentError("all functions must return vectors of the same length"))
+                    if keeprows
+                        throw(ArgumentError("all functions must return vectors of " *
+                                            "the length equal to the group rows count " *
+                                            "in the source GroupedDataFrame"))
```
nalimilan, line 1143:
```suggestion
                        throw(ArgumentError("all functions must return vectors with " *
                                            "as many values as rows in each group"))
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1167,30 +1135,55 @@ function _combine(f::AbstractVector{<:Pair},
...
+    # to keep track how the column was generated
+    # a correct index is stored in idx variable
+
+    for (i, (col_idx, col)) in enumerate(res)
+        if keeprows && res[i][1] !== idx_keeprows # we need to reorder the column
+            newcol = similar(col)
+            # we can probably make it more efficient, but I leave it as an optimization for the future
+            for i in axes(col, 1)
+                newcol[gd.idx[i]] = col[i]
+            end
```
nalimilan, line 1162:
```suggestion
            gd_idx = gd.idx
            @inbounds for j in eachindex(gd.idx, col)
                newcol[gd_idx[j]] = col[j]
            end
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1153,11 +1117,15 @@ function _combine(f::AbstractVector{<:Pair},
...
+    if !keeprows && isnothing(idx_loc)
         @assert !isnothing(idx_agg)
         idx = idx_agg
     else
-        idx = res[idx_loc][1]
+        if keeprows
+            idx = idx_keeprows
+        else
+            idx = res[idx_loc][1]
+        end
```
nalimilan, line 1128:
```suggestion
        idx = keeprows? idx_keeprows : res[idx_loc][1]
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1167,30 +1135,55 @@ function _combine(f::AbstractVector{<:Pair},
...
+                        throw(ArgumentError("all functions must return vectors of the same length"))
+                    end
                 end
             end
         end
     end
+
+    # remember that here first field in res[i] is not useful - it is just needed
+    # to keep track how the column was generated
+    # a correct index is stored in idx variable
```
nalimilan, line 1154: Is that field never useful at all, or just for the loop below? It sounds weird to say we store a field that's actually not needed in that form.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1447,187 +1440,57 @@ function _combine_tables_with_first!(first::Union{AbstractDataFrame,
...
-4×2 DataFrame
-│ Row │ a     │ sum_log_c │
-│     │ Int64 │ Float64   │
-├─────┼───────┼───────────┤
-│ 1   │ 1     │ 1.60944   │
-│ 2   │ 2     │ 2.48491   │
-│ 3   │ 3     │ 3.04452   │
-│ 4   │ 4     │ 3.46574   │
+Apply `args` to `gd` following the rules described in [`combine`](@ref).
+Ensure that the return value has number of rows equal to `nrow(parent(gd))`.
```
nalimilan, line 1447: That doesn't sound correct, as broadcasting happens contrary to `combine`.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1447,187 +1440,57 @@ function _combine_tables_with_first!(first::Union{AbstractDataFrame,
...
-│ 1   │ 1     │ [2, 2]   │ [1, 5]   │
-│ 2   │ 2     │ [1, 1]   │ [2, 6]   │
-│ 3   │ 3     │ [2, 2]   │ [3, 7]   │
-│ 4   │ 4     │ [1, 1]   │ [4, 8]   │
+An equivalent of
+`select(gd, args..., copycols=false, keepkeys=true, regroup=regroup)`
+but updates the `parent(gd)` in place.
+"""
+function select!(gd::GroupedDataFrame{DataFrame}, args...; regroup::Bool=false)
+    newdf = select(gd, args..., copycols=false, regroup=false)
```
nalimilan, line 1476:
```suggestion
    newdf = select(gd, args..., copycols=false)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1447,187 +1440,57 @@ function _combine_tables_with_first!(first::Union{AbstractDataFrame,
...
+    df = parent(gd)
+    copy!(_columns(df), _columns(newdf))
+    x = index(df)
+    copy!(_names(x), _names(newdf))
+    empty!(x.lookup)
+    for (i, n) in enumerate(x.names)
+        x.lookup[n] = i
+    end
```
nalimilan, line 1484: Don't we have a function doing this somewhere? I think this pattern is used in several places.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1447,187 +1440,57 @@ function _combine_tables_with_first!(first::Union{AbstractDataFrame,
...
-│ Row │ a     │ b_Ref    │ c_Ref    │
-│     │ Int64 │ SubArra… │ SubArra… │
-├─────┼───────┼──────────┼──────────┤
-│ 1   │ 1     │ [2, 2]   │ [1, 5]   │
-│ 2   │ 2     │ [1, 1]   │ [2, 6]   │
-│ 3   │ 3     │ [2, 2]   │ [3, 7]   │
-│ 4   │ 4     │ [1, 1]   │ [4, 8]   │
+An equivalent of
+`select(gd, args..., copycols=false, keepkeys=true, regroup=regroup)`
+but updates the `parent(gd)` in place.
```
nalimilan, line 1473:
```suggestion
but updates `parent(gd)` in place.
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1447,187 +1440,57 @@ function _combine_tables_with_first!(first::Union{AbstractDataFrame,
...
-│ 3   │ 2     │ 1     │ 2     │ 3       │
-│ 4   │ 2     │ 1     │ 6     │ 7       │
-│ 5   │ 3     │ 2     │ 3     │ 5       │
-│ 6   │ 3     │ 2     │ 7     │ 9       │
-│ 7   │ 4     │ 1     │ 4     │ 5       │
-│ 8   │ 4     │ 1     │ 8     │ 9       │
-```
+An equivalent of
+`transform(gd, args..., copycols=false, keepkeys=true, regroup=regroup)`
+but updates the `parent(gd)` in place.
```
nalimilan, line 1493:
```suggestion
but updates `parent(gd)` in place.
```

bkamins:
Thank you for the comments.

> except for the `ungroup` argument about which I'm not totally convinced as it goes against the general description of `combine`.

I understand you mean `regroup`. In general I wanted to leave it out and always return `DataFrame` from `select`/`select!`/`transform`/`transform!`/`combine`. The reasons why I have opted to include it are:
* it is more efficient to regroup immediately (we know how the groups look like) - if we did not have this efficiency gain I would not add it; but if we add it then there are points 2 and 3 below
* now we have full symmetry between all five functions we define (all support this keyword argument)
* we drop reserving `map`, as it was not obvious what it was doing for the users on the first sight
* `combine` now means "combine rows" not "combine groups" so this is consistent (maybe I have missed correcting it in the docstring - I will have a look)

@pdeffebach + @matthieugomez : do you have an opinion on this?

matthieugomez:
Yes, I think the kwarg should be in either in all functions or none.
Combine means combine rows now so it does not sound wrong.

Take it with a grain of salt, but, because I don’t plan to use
regroup=false, I would find it cleaner if it simply did not exist (that is,
transform select and combine always regroup). It would make it clearer that
groupby is a short-lived clause.

As for the doc, yes I completely agree it’s better to improve it later.

bkamins:
> I would find it cleaner if it simply did not exist

The issue is that if `gdf` is a `GroupedDataFrame` and you write:
```
gdf2 = select(gdf, some_transformations, regroup=true)
```
you get a `gdf2` which is a `GroupedDataFrame` but now - how do you get a `DataFrame` that contains the rows in a correct order?

A natural answer would be `DataFrame(gdf2)`, but it will not be correct, as it produces a `DataFrame` that is sorted by group order in `gdf2`.

A correct answer is to use `parent(gdf2)`, but I think that it will not be obvious for the users. Also, if later `gdf2` is sub-setted, then using the `parent` function is incorrect (it is correct only if you apply it to a freshly produced `GroupedDataFrame` by `groupby`, `combine`, `select` or `transform`).

src/groupeddataframe/splitapplycombine.jl
bkamins: your proposal is a bit slower (I guess `minimum` is able to avoid branching in the loop). Additionally my proposal additionally checks for errors (which we had in the past) in case `groups` is invalid and does not start from `1`. Therefore I leave my implementation.

src/groupeddataframe/splitapplycombine.jl
bkamins: This function is long and complex. I would prefer `i` to be defined only locally (it has no global meaning) - so that it does not pollute the namespace of `_combine`. Otherwise it is harder to debug cases if by accident `i` is used later before being properly initiated. Therefore I would leave the `let` block (although it does not improve anything performance-wise).

src/groupeddataframe/splitapplycombine.jl
bkamins: I removed `@inbounds` for now. An `@inbounds` of this kind was a cause of `map` error not being caught for years (here we rely on the fact that `gd_idx` is a permutation - and it should be, but we make so many changes that I would prefer not to risk such bugs). This should be optimized in the future, that is why I have left a note above

bkamins:
@nalimilan - as a comment on the `regroup` kwarg (to be sure we have the same understanding). It works as follows:

>    If `regroup=false` (the default) a `DataFrame` is returned.
>    If `regroup=true` a `GroupedDataFrame` grouped using `keycols(gdf)` is returned.

If you find word `regroup` non intuitive we can change it.
It is meant to say that we apply grouping again to the produced `DataFrame` (so we regroup it).

docs/src/man/getting_started.md
```diff
@@ -792,7 +792,9 @@ julia> mean(df.A)
 2.5
 ```
 
-We can also apply a function to each column of a `DataFrame` using `select`. For example:
+We can also apply a function to each column of a `DataFrame` using `select`.
```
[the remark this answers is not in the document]
bkamins: changed

docs/src/man/split_apply_combine.md
```diff
@@ -6,10 +6,24 @@ framework for handling this sort of computation is described in the paper
...
+(1) a data frame to be grouped, and (2) a set of columns to group by.
+
+The differences between the above functions are the following:
+* `select`: return a data frame with the number and order of rows exactly the same
+  as the source, preserve only columns that have been calculated;
+  `select!`: is an in-place version of `select`;
+* `transform`: return a data frame with the number and order of rows exactly the same
+  as the source, preserve all columns from the source and columns that have been calculated;
+  `transform!`: is an in-place version of `transform`;
+* `combine`: does not put restrictions on number of rows returned, the order of rows
```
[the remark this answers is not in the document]
bkamins: done

docs/src/man/split_apply_combine.md
```diff
@@ -27,19 +41,20 @@ each subset of the `DataFrame`. This specification can be of the following forms
    number of columns are processed (in which case `SubDataFrame` avoids excessive
    compilation)
 
-All forms except 1 and 6 can be also passed as the first argument to `map`.
```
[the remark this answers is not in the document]
bkamins: we do not support `map` now. That is why I have removed it.

docs/src/man/split_apply_combine.md
```diff
@@ -147,11 +205,61 @@ julia> by(iris, :Species, 1:2 => cor, nrow)
 
 ```
 
-The `by` function also supports the `do` block form. However, as noted above,
+If we use `select` or `transform` instead of `combine` we always obtain the number
+and of order of rows in the result equal to the source. In the example below
+the return values in columns `:SepalLength_SepalWidth_cor` and `:nrow` are
+broadcasted to match the number of elements in each group:
+```
+julia> select(gdf, 1:2 => cor, nrow)
```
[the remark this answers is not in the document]
bkamins: OK - removed.

src/abstractdataframe/selection.jl
```diff
@@ -181,6 +179,13 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}, AsTable},
                 newdfcols[i] = fill!(similar(col, length(res)), first(col))
             end
         end
+
+        # this means that we use `select` or `transform` not `combine`
```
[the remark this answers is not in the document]
bkamins: I hope now I have improved the description. The logic is:

* `allow_resizing_newdf[]` tells us if we can resize the `newdf`
* if we are in `select` it is never allowed (and it is assumed that if `newdf` has 0 columns it virtually has `nrow(df)` rows although `nrow(newdf)==0` at this point)
* if we are in `combine` it is allowed until we see a first vector (if we have seen only scalars up to some point `allow_resizing_newdf[]` is `true`)

The case in this part of code is the following:
* we are processing a vector `res`
* if we already have some columns in `newdf` we do not care here what is its length (later we will make sure we process it correctly and check its shape when we add it to `newdf` using `setindex!`)
* but we can have a situation when `res` is a first value pushed to `newdf`, this is a `ncol(newdf) == 0` part of the condition; now here things differ between `select` and `combine`; in `combine` we have that `allow_resizing_newdf[]` will be `true` so the condition is not met; but in `select` we have that `allow_resizing_newdf[]` is `false` here and in this case we require this vector to have length equal to the number of rows of the source `df` data frame

src/abstractdataframe/selection.jl
```diff
@@ -538,10 +605,11 @@ function select(df::DataFrame, cs...; copycols::Bool=true)
             push!(cs_vec, v)
         end
     end
-    _select(df, [normalize_selection(index(df), c) for c in cs_vec], copycols)
+    return _process(df, [normalize_selection(index(df), c) for c in cs_vec],
+                    copycols, keeprows)
 end
 
-function _select(df::AbstractDataFrame, normalized_cs, copycols::Bool)
+function _process(df::AbstractDataFrame, normalized_cs, copycols::Bool, keeprows::Bool)
```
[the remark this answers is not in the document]
bkamins: OK - changed to `manipulate` and `_manipulate`.

src/deprecated.jl
```diff
@@ -444,3 +446,25 @@ function aggregate(d::AbstractDataFrame, cols, fs::AbstractVector;
 end
 
 @deprecate deleterows!(df::DataFrame, inds) delete!(df, inds)
+
+@deprecate by(f::Union{Base.Callable, Pair}, d::AbstractDataFrame, cols::Any;
+   sort::Bool=false, skipmissing::Bool=false,
+   keepkeys::Bool=true) combine(groupby(d, cols, sort=sort, skipmissing=skipmissing),
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
bkamins: The old description was wrong, but it was late when I was writing this. The rule is simply:

> Ordering of rows follows the order of groups in `gdf`.

I have fixed this

src/groupeddataframe/splitapplycombine.jl
bkamins: Right - changed to `_combine_prepare`. Not sure it is best though. The idea is that we need an "intermediate" function as:
* `select` calls `_combine_prepare` which later calls `_combine_helper`
* `combine` calls `_combine_prepare` (which later calls `_combine_helper`) if its first argument is `GroupedDataFrame`
* `combine` directly calls `_combine_helper` if its first argument is a `Pair` or a `Callable`

src/groupeddataframe/splitapplycombine.jl
bkamins: sorry - I have forgotten to remove the comment

src/groupeddataframe/splitapplycombine.jl
bkamins: I just removed this function and replaced it with `minimum(gd.groups) == 0` as it is easy enough to digest.

src/groupeddataframe/splitapplycombine.jl
bkamins: Well - it was in the original code (and it was not used). I have corrected the comment.

Now we use it - it tells us how the column was generated as we have three modes:
* from aggregate
* from a function
* by a direct reuse of the column from the parent (this happens in `select` and `transform` which keep the order of rows from the parent)

src/groupeddataframe/splitapplycombine.jl
bkamins: `combine` also broadcasts (just to any number of rows).

I have updated the docs + added examples + added `See also` sections everywhere.

src/groupeddataframe/splitapplycombine.jl
bkamins: A good point. I have introduced `_replace_columns!` function which also simplified `select!` for `DataFrame`

bkamins:
I have updated the PR following the first round of comments. I have marked the comments resolved except for places where it is non obvious or potentially disputable.

matthieugomez:
Sorry, I am confused. I meant:  I don’t plan to use regroup=true.
The keyword argument does not have a great name — I though regroup means put together all the groups, ie return a DataFrame. Now I understand that it means “group again”, but that’s not how I understood it.

bkamins:
Yes - this is what I have noticed - do you have a proposal of a better name, maybe just `group`?

matthieugomez:
I think ungroup would be the clearest (but maybe we could just remove it).

bkamins:
Do I understand it correctly that `ungroup=true` (the default) would return a `DataFrame` and `ungroup=false` would return a `GroupedDataFrame`?

We should not remove this kwarg for performance reasons (in many cases grouping is more expensive than the operation performed later on the data). In the example I am still using the old `regroup` kwarg:
```
julia> @time df = DataFrame(g=rand(10^8));
  0.620453 seconds (28 allocations: 1.490 GiB, 1.99% gc time)

julia> @time gdf = groupby(df, :g);
 14.075247 seconds (33 allocations: 2.490 GiB, 0.45% gc time)

julia> @time df2 = combine(gdf, :g => length);
  1.532653 seconds (179 allocations: 2.328 GiB, 18.65% gc time)

julia> @time gdf2 = groupby(df2, :g);
 14.380990 seconds (34 allocations: 2.490 GiB, 1.82% gc time)

julia> @time gdf3 = combine(gdf, :g => length, regroup=true);
 11.977388 seconds (500.00 M allocations: 12.014 GiB, 7.66% gc time)
```
So you can see that it is cheaper to agregate and `regroup=true` than just to to `groupby` later.

bkamins:
OK, I was surprised that the gain was so small. Now I have fixed the performance (barrier functions came in handy; @nalimilan - as a side note the old `map` code was not only buggy but also inefficient due to type instability); the benchmark after the update is:
```
julia> @time df = DataFrame(g=rand(10^8));
  0.790909 seconds (28 allocations: 1.490 GiB, 22.32% gc time)

julia> @time gdf = groupby(df, :g);
 14.090509 seconds (50 allocations: 2.490 GiB, 1.26% gc time)

julia> @time df2 = combine(gdf, :g => length);
  1.498985 seconds (179 allocations: 2.328 GiB, 16.99% gc time)

julia> @time gdf2 = groupby(df2, :g);
 14.069919 seconds (34 allocations: 2.490 GiB, 0.82% gc time)

julia> @time gdf3 = combine(gdf, :g => length, regroup=true);
  1.847234 seconds (186 allocations: 3.073 GiB, 12.57% gc time)
```
(and you see that `regroup` is almost 0 cost)

the same with `select`:
```
julia> gdf.idx;

julia> @time select(gdf, :g => length);
  6.707589 seconds (200.00 M allocations: 8.196 GiB, 6.05% gc time)

julia> @time select(gdf, :g => length, regroup=true);
  6.758230 seconds (200.00 M allocations: 8.196 GiB, 6.17% gc time)
```

docs/src/man/split_apply_combine.md
```diff
@@ -27,19 +42,21 @@ each subset of the `DataFrame`. This specification can be of the following forms
...
-All forms except 1 and 6 can be also passed as the first argument to `map`.
-
 As a special rule that applies to `cols => function` syntax, if `cols` is wrapped
 in an `AsTable` object then a `NamedTuple` containing columns selected by `cols` is
 passed to `function`.
 
 In all of these cases, `function` can return either a single row or multiple rows.
 `function` can always generate a single column by returning a single value or a vector.
-Additionally, if `by` is passed exactly one `function` and `target_col` is not specified,
+Additionally, if `combine` is passed exactly one `function` as a first argument
```
pdeffebach, line 51:
```suggestion
Additionally, if `combine` is passed exactly one `function`  or `cols => function` pair as a first argument
```

docs/src/man/split_apply_combine.md
```diff
@@ -114,9 +175,8 @@ julia> by(iris, :Species, nrow, :PetalLength => mean => :mean)
 │ 2   │ Iris-versicolor │ 50    │ 4.26    │
 │ 3   │ Iris-virginica  │ 50    │ 5.552   │
 
-julia> by(iris, :Species,
-          [:PetalLength, :SepalLength] =>
-          (p, s) -> (a=mean(p)/mean(s), b=sum(p))) # multiple columns are passed as arguments
+julia> combine([:PetalLength, :SepalLength] => (p, s) -> (a=mean(p)/mean(s), b=sum(p)),
```
pdeffebach, line 178: Are we deprecating `combine(fun, gdf)`? If not, we should add an example showing how this call is equivelent to a function syntax. 

I have no strong feelings about the deprecation, but wouldn't mind a `DF()` column selector to be added the future.

src/groupeddataframe/splitapplycombine.jl
bkamins: I had to rewrite it for performance reasons anyway, so I resolve this comment.

docs/src/man/split_apply_combine.md
bkamins: We are not deprecating it. The example you ask for is given in the docstring of combine. I will add it also here as you requested.

matthieugomez:
Ungroup= true (default) would mean that it returns a DataFrame.

bkamins:
OK - I have pushed a commit changing `regroup` to `ungroup`

bkamins:
Only coverage fails

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1441,56 +1418,194 @@ end
 
 """
     select(gd::GroupedDataFrame, args...;
-           copycols::Bool=true, keepkeys::Bool=true, regroup::Bool=false)
+           copycols::Bool=true, keepkeys::Bool=true, ungroup::Bool=true)
 
 Apply `args` to `gd` following the rules described in [`combine`](@ref).
-Ensure that the return value has number of rows equal to `nrow(parent(gd))`.
+The return value has number of rows equal to `nrow(parent(gd))`
+(if single value is returned it is always broadcasted to have this number of rows).
```
nalimilan, line 1425:
```suggestion
The returned object has as many rows as `parent(gd)`.
If an operation returns a single value it is always broadcasted to have this number of rows.
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1441,56 +1418,194 @@ end
...
+│ 2   │ 1     │ 8     │ 19    │
+│ 3   │ 1     │ 8     │ 19    │
+│ 4   │ 2     │ 4     │ 17    │
+│ 5   │ 2     │ 4     │ 17    │
+│ 6   │ 1     │ 8     │ 19    │
+│ 7   │ 1     │ 8     │ 19    │
+│ 8   │ 2     │ 4     │ 17    │
+
+julia> select(gd, :b => :b1, :c => :c1,
+              [:b, :c] => +, keepkeys=false) # auto-splatting, renaming and keepkeys
```
nalimilan, line 1516:
```suggestion
              [:b, :c] => +, keepkeys=false) # multiple arguments, renaming and keepkeys
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1441,56 +1418,194 @@ end
 
 """
     select(gd::GroupedDataFrame, args...;
-           copycols::Bool=true, keepkeys::Bool=true, regroup::Bool=false)
+           copycols::Bool=true, keepkeys::Bool=true, ungroup::Bool=true)
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -609,42 +611,17 @@ function combine_helper(f, gd::GroupedDataFrame,
...
-                end
-            end
-            # all groups must be filled
-            @assert minimum(groups) == 1
-
+            groups = gen_groups(idx)
+            @assert groups[end] <= length(gd)
             return GroupedDataFrame(newparent, collect(1:length(gd.cols)), groups,
-                                    collect(1:length(idx)), starts, ends, j, nothing)
+                                    nothing, nothing, nothing, groups[end], nothing)
```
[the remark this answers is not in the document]
bkamins: I was thinking about it. The reason to remove it is the following:
1. we would have to compute them anyway (they are not guaranteed to be given like in the branch above)
2. if we compute them now then we have to store them (so it takes memory), and it is likely that they might be even not needed
3. the cost of computing indices with `compute_indices` is higher than computing them here but not that much higher (it doest 3 loops instead of 2 loops that could be used in an optimal case so the saving is at most ~33%)

For `10^8` groups we have:
```
julia> x = [1:10^8;];

julia> @time DataFrames.compute_indices(x, 10^8);
  1.419073 seconds (8 allocations: 2.235 GiB, 11.12% gc time)
```
and with an optimal timing we would have it a bit over 1 second (at the cost of allocating these 3 vectors - and maybe we even do not have to pay this cost at all if later this `GroupedDataFrame` will be used to do some aggregation function like `sum` when we do not need to pay this cost at all).

In summary:
* if we have to pay this cost we will be slow anyway (as it means we had to use non-aggregating operations which will be more expensive than `compute_indices` anyway)
* if we do not have to pay it it means we are doing aggregations only which are much faster than `compute_indices` (and this is probably the most likely use case)
* additionally we save some memory (this "some" can be a lot if there are a lot of small groups)

bkamins:
Thank you for the comments. The PR is updated.

The decision to store or not to store `idx`, `starts` and `ends` is indeed double edged. I have opted to be maximally fast and space efficient in the "fast case" - i.e. when we use fast aggregation functions later on a `GroupedDataFrame` which is the case in which we fight for performance..

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -407,31 +245,34 @@ const KWARG_PROCESSING_RULES =
...
+    combine(gd::GroupedDataFrame, args...; keepkeys::Bool=true, ungroup::Bool=true)
+    combine(fun::Union{Function, Type}, gd::GroupedDataFrame;
+            keepkeys::Bool=true, ungroup::Bool=true)
+    combine(pair::Pair, gd::GroupedDataFrame; keepkeys::Bool=true, ungroup::Bool=true)
+    combine(fun::Union{Function, Type}, df::AbstractDataFrame, ungroup::Bool=true)
+    combine(pair::Pair, df::AbstractDataFrame, ungroup::Bool=true)
 
-Transform a [`GroupedDataFrame`](@ref) into a `DataFrame`.
+Apply operations to each group in a [`GroupedDataFrame`](@ref) and return
+the combined result as a `DataFrame`.
```
nalimilan, line 262: I guess `ungroup` should be mentioned here directly, or the type of the result shouldn't be mentioned here?

bkamins: Ah - OK. I have updated both `combine` and `select`.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1423,9 +1425,15 @@ end
...
-Apply `args` to `gd` following the rules described in [`combine`](@ref).
-The returned object has as many rows as `parent(gd)`.
-If an operation returns a single value it is always broadcasted to have this number of rows.
+Apply `args` to `gd` following the rules described in [`combine`](@ref) and return the
+result as a `DataFrame` if `ungroup=true` or `GroupedDataFrame` if `ungroup=false`.
+
+The `parent` of the returned value has as many rows as `parent(gd)`. If an operation
+in `args` returns a single value it is always broadcasted to have this number of rows.
+
+Apply operations to each group in a [`GroupedDataFrame`](@ref) and return the combined
```
nalimilan, line 1434: This part is repeated.

bkamins: fixed. sorry for this.

bkamins:
OK - so I am merging this as we know CI shall pass and I will open a new PR just bumping the version and there we will check that everything passes on a clean PR.

@nalimilan - will you now make an update to CSV.jl that fixes CategoricalArrays.jl compatibility so that we can synchronize the releases. Thank you!

matthieugomez:
That’s awesome! Thanks a lot for this!

Issue #2221, opened 2020-05-03
possible test failure in upcoming Julia version 1.5 


rfourquet:
A PkgEval run for a Julia [pull request](https://github.com/JuliaLang/julia/pull/29240) which changes the generated numbers for `rand(a:b)` indicates that the tests of this package might fail in Julia 1.5 (and on Julia current master branch).

Also, you might be interested in using the new [StableRNGs.jl](https://github.com/rfourquet/StableRNGs.jl) registered package, which provides guaranteed stable streams of random numbers across Julia releases.

Apologies if this is a false positive. Cf. https://github.com/JuliaCI/NanosoldierReports/blob/ab6676206b210325500b4f4619fa711f2d7429d2/pkgeval/by_hash/52c2272_vs_47c55db/logs/DataFrames/1.5.0-DEV-87d2a04de3.log

bkamins:
@rfourquet - thank you for caring so much to inform the package maintainers about the change!

bkamins:
@nalimilan - I will make this update after https://github.com/JuliaData/DataFrames.jl/pull/2214 is merged. It affects only tests so it can be done after 0.21 release

bkamins:
In #2223 I have made the tests independent of PRNG used (which should have been done from the start).

PR #2223, opened 2020-05-04
updates required for Julia 1.5

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2221.

Additionaly updates DataFrames.jl to changes in internal implementation of broadcasting in Base.


bkamins:
@nalimilan - fortunately the changes required are independent of everything else we change.

bkamins:
@nalimilan - I have made one more update that was failing on 32-bit.

We have also one problem here https://travis-ci.org/github/JuliaData/DataFrames.jl/jobs/682985140. It is on 64 bit and happens randomly when trying to evaluate a `Regex`. I could not locate the bug on our side, so it might be some that got broken in Base. We from time to time get a strange error:
```
LoadError: LoadError: InexactError: check_top_bit(Int32, 4294967248)

Stacktrace:

 [1] top-level scope at /Users/travis/build/JuliaData/DataFrames.jl/test/runtests.jl:48
 [2] include(::String) at ./client.jl:457
 [3] top-level scope at none:6
 [4] eval(::Module, ::Any) at ./boot.jl:331
 [5] exec_options(::Base.JLOptions) at ./client.jl:272
 [6] _start() at ./client.jl:506
```

@Keno - maybe you can quickly diagnose what is the reason of the problem here?

bkamins:
@nalimilan, @Keno - after my update the tests pass on 1.0, 1.4 everywhere.
Also on nightly pass everywhere (Windows, Linux, both 64 and 32 bit) except OSx 64 bit where we get the same `LoadError: LoadError: InexactError: check_top_bit(Int32, 4294967248)` problem (so this seems to be a problem with Julia Base on OSx, but maybe we have some bug in DataFrames.jl that only manifests itself under this combination?).

bkamins:
OK - the problem with pcre on Julia nightly seems to be known so we can safely merge this PR.

PR #2228, opened 2020-05-05
Ignore `Not` which column not in data frame

### LINKED ISSUE #2197: Return data frame unaltered when Not only includes columns that are not in data frame
Currently we have this behavior:

```julia
julia> df = DataFrame(a=1:2);

julia> select(df, Not(:b))
ERROR: ArgumentError: column name :b not found in the data frame; existing most similar names are: :a
Stacktrace:
 [1] lookupname at /Users/bieganek/.julia/packages/DataFrames/S3ZFo/src/other/index.jl:253 [inlined]
 [2] getindex at /Users/bieganek/.julia/packages/DataFrames/S3ZFo/src/other/index.jl:259 [inlined]
 [3] getindex at /Users/bieganek/.julia/packages/DataFrames/S3ZFo/src/other/index.jl:192 [inlined]
 [4] select(::DataFrame, ::InvertedIndex{Symbol}; copycols::Bool) at /Users/bieganek/.julia/packages/DataFrames/S3ZFo/src/dataframe/dataframe.jl:881
 [5] select(::DataFrame, ::InvertedIndex{Symbol}) at /Users/bieganek/.julia/packages/DataFrames/S3ZFo/src/dataframe/dataframe.jl:881
 [6] top-level scope at REPL[3]:1
```

However, it seems like the data frame should be returned unaltered if the `Not` only includes columns that don't exist in the data frame.

My motivation for this is that I want to apply `select!(df, Not(...))` to a bunch of data frames, some of which have the columns I'm dropping, and some of which do not have the columns I'm dropping.

pdeffebach:
Fixed #2197

```
df = DataFrame(a = [1], b = [2])
select(df, Not(:c)) # returns df
```

I still have to add tests.


bkamins:
The rule you propose is incorrect as what is in `Not` can be any selector, e.g. `Not(r"x")` or `Not(Between(:x1, :x2))`, so unfortunately this PR will not be that easy (I have not spent much time on designing a good set of rules - maybe it is enough to add a third branch).

pdeffebach:
Thanks for the feedback. I think the simplest thing to do in this context is add a bunch of `if` `else` statements. The tests show the behavior I want. Let me know if you have more feedback.

src/other/index.jl
```diff
@@ -218,9 +218,26 @@ end
     if notidx.skip isa AbstractVector 
         toskip = filter(t -> haskey(x, t), notidx.skip)
         return setdiff(1:length(x), toskip)
+    elseif notidx.skip isa Between
+        if haskey(x, notidx.skip.first) && haskey(x, notidx.skip.last)
+            toskip = x[notidx.first]:x[notidx.last]
+            return setdiff(1:length(x), toskip)
+        elseif !(haskey(x, notidx.skip.first)) && !(haskey(x, notidx.skip.last))
+            return 1:length(x)
```
bkamins, line 226: why do you think it should not error?

src/other/index.jl
```diff
@@ -218,9 +218,26 @@ end
...
+        else
+            # one of these will error
+            getindex(x, notidx.skip.first)
+            getindex(x, notidx.skip.last)
+        end
+    elseif notidx.skip isa Regex
```
bkamins, line 232: why this branch is needed - it seems it is the same as the last `else`?

test/index.jl
```diff
@@ -360,6 +360,16 @@ end
     @test DataFrames.parentcols(SubIndex(i3, Not(Not(r"")))) == 1:2
 end
 
+@testset "Not indexing with columns that don't exist" begin
+    df = DataFrame(a=1, b=2, c=3)
+    @test select(df, Not(:d)) == df
+    @test select(df, Not(Between(:x, :z))) == df
+    @test select(df, Not(r"zzz")) == df
+    @test select(df, Not(:x)) == df
+    @test select(df, Not(All(:d, r"zzz", Between(:q, :p)))) == df
```
bkamins, line 369: Can you please add a test for just `All()` also?

bkamins:
It looks almost good. Thank you!

> The tests show the behavior I want.

Right :smile:. We need to think where to put a precise description of the behaviour you implement. Probably https://juliadata.github.io/DataFrames.jl/latest/lib/indexing/ is the right place. The key thing to highlight there is that we will have different rules what errors if you select rows and what errors if you select columns. As now we are consistent:
```
julia> df = DataFrame(rand(2,2))
2×2 DataFrame
│ Row │ x1       │ x2       │
│     │ Float64  │ Float64  │
├─────┼──────────┼──────────┤
│ 1   │ 0.784868 │ 0.736577 │
│ 2   │ 0.230237 │ 0.628254 │

julia> df[Not(3), :]
ERROR: BoundsError: attempt to access 2-element Base.OneTo{Int64} at index [Not(3)]

julia> df[:, Not(3)]
ERROR: BoundsError: attempt to access data frame with 2 columns
  at index [3]
```
but this would not be the case any more after this PR.

Thinking about this - maybe (I am not 100% sure what would be best). An alternative of the only case we should accept for `Not` with an invalid column index should be `Not(::Symbol)`, and `Not(::AbstractString)`, as in practice these seem to be the real cases (or more flexibly - it should not be an error if column name is passed, but should be an error when column index is passed). I am not 100% sure here what is best here, as we have a tension between flexibility and consistency. I will have to think about it more.

@CameronBieganek and @nalimilan - maybe you have some thoughts.

src/other/index.jl
```diff
@@ -214,8 +214,35 @@ end
 @inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) =
     getindex(x, collect(Int, idx))
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
-@inline Base.getindex(x::AbstractIndex, notidx::Not) =
-    setdiff(1:length(x), getindex(x, notidx.skip))
+@inline function Base.getindex(x::AbstractIndex, notidx::Not)
+    if notidx.skip isa AbstractVector 
+        toskip = filter(t -> haskey(x, t), notidx.skip)
+        return setdiff(1:length(x), toskip)
```
CameronBieganek, line 220: Maybe I'm missing something, but can't you omit the `filter` here and just use `setdiff`? E.g., this works:

```julia
julia> setdiff(1:3, 3:4)
2-element Array{Int64,1}:
 1
 2
```

Also, what happens if you have a vector of symbols, e.g. `Not([:c, :d])`?

CameronBieganek:
> An alternative of the only case we should accept for `Not` with an invalid column index should be `Not(::Symbol)`, and `Not(::AbstractString)`, as in practice these seem to be the real cases (or more flexibly - it should not be an error if column name is passed, but should be an error when column index is passed). I am not 100% sure here what is best here, as we have a tension between flexibility and consistency.

If I understand correctly, you're asking whether `Not(3)` should be an error and `Not(:x3)` should be allowed in your above example? I think that might be a good idea. As you mentioned, it's inconsistent with `Not(:x3)`, but at least it's consistent with `df[Not(3), :]`.

By the way, it's not clear from the discussion on this PR so far, but I think `Not(::Vector{Symbol})` should also be supported. I think that was how I originally came across this. 🙂

src/other/index.jl
bkamins: right - `toskip` should be probably be a comprehension calling `getindex` with `haskey` after the `if` part.

bkamins:
My point is to consider supporting missing column in only:
```
Not(::Union{Symbol, AbstractString, AbstractVector{Symbol}, AbstractVector{<:AbstractString}, Between{Symbol, Symbol})
```
(i.e. cases where you pass a column name explicitly and it is not a nested selection)

nalimilan:
Having `Not(3)` and `Not(:x3)` be inconsistent sounds problematic to me. I wonder whether we could use something like `Not(Cols(:x3))` for that, since `Cols` is already more flexible by dropping duplicates silently.

src/other/index.jl
pdeffebach: I would be fine having this error. 

Here is my reasoning for the current behavior. Imagine a dataframe with infinite columns, and a contiguous set of columns comprises your data frame. When you do `Not(Between(:x, :z))`, either 

1. `:x` and `:z` are both to the left of your columns of interest. Then no error, do nothing
2. `:x` and `:z` are both to the right of your columns of interest. Then no error, do nothing. 
3. `:x` is on the left of your columns of interest, `:z` is on the right. In which case you should delete all columns. 

This PR, currently, only assumes options 2 and 3.

src/other/index.jl
pdeffebach: Yes. Though it's a bit less readable now, and the order of the shortcircuited OR matters.

pdeffebach:
I responded to all comments about the implementation. 

I think it's too much to require users to always to `Not(Cols(:x))`. Personally I am fine with a bit of inconsistency between row indexing and column indexing.

pdeffebach:
Finally, should I refactor this to just do dispatch?

bkamins:
Having reviewed the revision I still feel that we should only special case:
```
Not(::Union{Symbol, AbstractString, AbstractVector{Symbol}, AbstractVector{<:AbstractString}, Between{Symbol, Symbol})
```
and handle everything else like we do currently.

The reason is that now I see that there is a bug with handling `Bool` indexing. Then it is also clear that it will be hard to handle other corner cases of indexing.

Simply the proposed design (that is general) does not compose well with indexing of `DataFrames`.

If you feel we should special case some more cases except `Not(::Union{Symbol, AbstractString, AbstractVector{Symbol}, AbstractVector{<:AbstractString}, Between{Symbol, Symbol})` then please just add them to the list, but I would be hesitant to allow an implementation that ignores type of what is stored in `Not` as proposed now. Simply the indexing rules of `DataFrame` are too complex (of course they could be replicated but it will lead to a very messy code).

src/other/index.jl
bkamins: But if `!(haskey(x, notidx.skip.first)) && !(haskey(x, notidx.skip.last))` is true then you are not sure if

`:x` is on the left and `:z` is on the right (and do what you proposed)

or

`:z` is on the left and `:x` is on the right (and then exactly the opposite should happen)

That is why I felt it should also error.

If we agree then `Between` should be removed from the list in my main comment.

bkamins:
> Finally, should I refactor this to just do dispatch?

This is what I would do, and in dispatch I would only allow the types that we are sure we handle correctly.

pdeffebach:
Thanks! I refactored so it uses dispatch. 

I also changed the `Between` behavior to be more strict. It errors unless both elements are in the index. 

Final decision is how to type this up in docs and if we should update getindex for rows as well. I don't think we should, also because we want to maintain consistency with the behavior of `Not` with ordinary vectors.

src/other/index.jl
```diff
@@ -214,8 +214,28 @@ end
 @inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) =
     getindex(x, collect(Int, idx))
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
-@inline Base.getindex(x::AbstractIndex, notidx::Not) =
+
+@inline Base.getindex(x::AbstractIndex, notidx::Not{<:AbstractVector}) =
```
bkamins, line 218: As I have noted already this definition is incorrect in general (e.g. it does not correctly handle boolean indexing).
Therefore I would prefer to narrow it down to 
```suggestion
@inline Base.getindex(x::AbstractIndex, notidx::Union{Not{<:AbstractVector{Symbol}}, Not{<:AbstractVector{<:AbstractString}}}) =
```

src/other/index.jl
```diff
@@ -214,8 +214,28 @@ end
...
+@inline function Base.getindex(x::AbstractIndex, notidx::Not{<:All})
+    if isempty(notidx.skip.cols) 
+        return Int[]
+    else 
+        intersect(getindex.(Ref(x), Not.(notidx.skip.cols))...)
+    end
+end
+@inline Base.getindex(x::AbstractIndex, notidx::Not{<:Between}) =
+    return setdiff(1:length(x), getindex(x, notidx.skip))
+@inline Base.getindex(x::AbstractIndex, notidx::Not{<:Regex}) =
```
bkamins, line 229: This is enough I think:
```suggestion
@inline Base.getindex(x::AbstractIndex, notidx::Not) =
```

src/other/index.jl
```diff
@@ -214,8 +214,28 @@ end
...
     setdiff(1:length(x), getindex(x, notidx.skip))
+@inline function Base.getindex(x::AbstractIndex, notidx)
```
bkamins, line 231: I guess you meant this:
```suggestion
@inline function Base.getindex(x::AbstractIndex, notidx::Union{Not{Symbol},Not{<:AbstractString}})
```

test/index.jl
```diff
@@ -360,6 +360,17 @@ end
     @test DataFrames.parentcols(SubIndex(i3, Not(Not(r"")))) == 1:2
 end
 
+@testset "Not indexing with columns that don't exist" begin
```
bkamins, line 363: The tests should cover every possible combination types of inputs for `Not`. So these should be:
`Bool`, `Integer`, `AbstractVector{Int}`, `AbstractRange{Int}`, `AbstractVector{<:Integer}`, `AbstractRange{<:Integer}`, `AbstractRange{Bool}`, `AbstractVector{Bool}`, `Colon`, `Not` (nested), `Between`, `All`, `Regex`, `Symbol`, `AbstractString`, `AbstractVector{Symbol}`, `AbstractVector{<:AbstractString}`, `AbstractVector{Any}` (this one testing different combinations allowed in the implementation)

pdeffebach: Okay I've implemented the functionality that throws errors for integer values. 

`Not()` is way under-tested in general. So I will add tests normal `Not` behavior (columns in the data frame) soon.

bkamins: > Not() is way under-tested in general.

It was not tested before, because we had a very simple logic of processing it before. Since now we deviate from the "default" we should test all cases.

src/other/index.jl
```diff
@@ -214,8 +214,33 @@ end
 @inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) =
     getindex(x, collect(Int, idx))
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
-@inline Base.getindex(x::AbstractIndex, notidx::Not) =
+
+@inline Base.getindex(x::AbstractIndex, notidx::Union{Not{<:AbstractVector{Symbol}}, Not{<:AbstractVector{<:AbstractString}}}) =
```
bkamins, line 218:
```suggestion
@inline Base.getindex(x::AbstractIndex, notidx::Union{Not{<:AbstractVector{Symbol}},
                                                      Not{<:AbstractVector{<:AbstractString}}}) =
```

src/other/index.jl
```diff
@@ -214,8 +214,33 @@ end
...
+@inline Base.getindex(x::AbstractIndex, notidx::Not{<:Regex}) =
+    setdiff(1:length(x), getindex(x, notidx.skip))
+@inline function Base.getindex(x::AbstractIndex, notidx::Union{Not{Symbol},Not{<:AbstractString}})
+    if haskey(x, notidx.skip) 
+        return setdiff(1:length(x), getindex(x, notidx.skip))
+    else 
+        return 1:length(x)
+    end
+end
+@inline Base.getindex(x::AbstractIndex, notidx) =
```
bkamins, line 240: This is an incorrect definition and should be removed

src/other/index.jl
bkamins: Do you have a problem with accepting this proposed change or its intention is not clear to you?

bkamins:
Thank you - I have left some more comments.

src/other/index.jl
pdeffebach: No I just implemented it not realizing those cases were general. added.

test/index.jl
pdeffebach: let me know if there is anything you would like to change about the current block of tests because they are a bit obtuse and there is going to be a big test diff anyways.

src/other/index.jl
```diff
@@ -214,8 +214,28 @@ end
...
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
+
+@inline Base.getindex(x::AbstractIndex, notidx::Union{Not{<:AbstractVector{Symbol}},
+                                                      Not{<:AbstractVector{<:AbstractString}}}) =
+    setdiff(1:length(x), [getindex(x, idx) for idx in notidx.skip if haskey(x, idx)])
+@inline function Base.getindex(x::AbstractIndex, notidx::Not{<:All})
+    if isempty(notidx.skip.cols) 
+        return Int[]
+    else 
+        intersect(getindex.(Ref(x), Not.(notidx.skip.cols))...)
```
bkamins, line 225:
```suggestion
        return intersect(getindex.(Ref(x), Not.(notidx.skip.cols))...)
```

src/other/index.jl
```diff
@@ -214,8 +214,28 @@ end
 @inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) =
     getindex(x, collect(Int, idx))
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
+
+@inline Base.getindex(x::AbstractIndex, notidx::Union{Not{<:AbstractVector{Symbol}},
+                                                      Not{<:AbstractVector{<:AbstractString}}}) =
+    setdiff(1:length(x), [getindex(x, idx) for idx in notidx.skip if haskey(x, idx)])
```
bkamins, line 220:
```suggestion
    setdiff(1:length(x), [getindex(x, idx) for idx in notidx.skip if haskey(x, idx)])
    
```

src/other/index.jl
```diff
@@ -214,8 +214,28 @@ end
...
+    if haskey(x, notidx.skip) 
+        return setdiff(1:length(x), getindex(x, notidx.skip))
+    else 
+        return 1:length(x)
+    end
+end
 @inline Base.getindex(x::AbstractIndex, notidx::Not) =
     setdiff(1:length(x), getindex(x, notidx.skip))
+
+
```
bkamins, line 238:
```suggestion
```

pdeffebach:
I'm slowly working my way through tests over the next few days.

bkamins:
Thank you. No rush. We anyway should wait for @nalimilan to have a look at this PR.

pdeffebach:
Okay I added all tests. I think all cases are covered.

bkamins:
Thank you! I have restarted the CI as for some reason it did not spit up correctly after your last commit.

src/other/index.jl
```diff
@@ -214,6 +214,27 @@ end
 @inline Base.getindex(x::AbstractIndex, idx::AbstractRange{<:Integer}) =
     getindex(x, collect(Int, idx))
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
+
+@inline Base.getindex(x::AbstractIndex, notidx::Union{Not{<:AbstractVector{Symbol}},
+                                                      Not{<:AbstractVector{<:AbstractString}}}) =
+    setdiff(1:length(x), [getindex(x, idx) for idx in notidx.skip if haskey(x, idx)])
```
bkamins, line 220: One more thing I have noted. This approach allows duplicates in `notidx.skip`. Normally we error when duplicates are passed. Maybe we should have `allunique` test on it and throw an error if duplicates are found?

bkamins:
CI fails

test/index.jl
```diff
@@ -44,11 +44,6 @@ using DataFrames: Index, SubIndex, fuzzymatch
     @test_throws ArgumentError i[[1,1]]
     @test_throws ArgumentError i[[:A,:A]]
     @test_throws ArgumentError i[["A","A"]]
-    @test_throws BoundsError i[Not(0)]
```
bkamins, line 47: why have you removed these?

pdeffebach: These are tested below, now, in 

```
@testset "Not indexing with columns that don't exist" begin
```

bkamins: ok

src/other/index.jl
```diff
@@ -214,6 +214,38 @@ end
...
     getindex(x, collect(Int, idx))
 @inline Base.getindex(x::AbstractIndex, ::Colon) = Base.OneTo(length(x))
+
+@inline function Base.getindex(x::AbstractIndex, notidx::Union{Not{<:AbstractVector{Symbol}},
+                                                         Not{<:AbstractVector{<:AbstractString}}})
+    allunique(notidx.skip) || throw(ArgumentError("Elements of $(notidx.skip) must be unique"))
+    setdiff(1:length(x), [getindex(x, idx) for idx in notidx.skip if haskey(x, idx)])
+end
+
+@inline function Base.getindex(x::AbstractIndex, notidx::Not{<:AbstractVector{<:Integer}})
```
bkamins, line 224: why is this method needed? I would think it is covered by https://github.com/JuliaData/DataFrames.jl/pull/2228/files#diff-084f63e064c680ae660e2676e15d3d4eR249.

Then also https://github.com/JuliaData/DataFrames.jl/pull/2228/files#diff-084f63e064c680ae660e2676e15d3d4eR229 is not needed

bkamins:
Apart from comment https://github.com/JuliaData/DataFrames.jl/pull/2228/files#r432882413 it looks good. So the only thing left is documentation update.

src/other/index.jl
pdeffebach: After you suggested we throw an error for repeated elements in the array I had to add a check that isn't present in the `Any` method. 

But then since `Bool <: Integer` I had to add another method because obviously `Vector{Bool}` will contain repeated elements. 

Let me know if that makes sense.

bkamins: My question is because we make this check in line https://github.com/JuliaData/DataFrames.jl/pull/2228/files#diff-084f63e064c680ae660e2676e15d3d4eR250. So now you would be doing the same check twice in this branch.

Now the check is needed in https://github.com/JuliaData/DataFrames.jl/pull/2228/files#diff-084f63e064c680ae660e2676e15d3d4eR220 because in line https://github.com/JuliaData/DataFrames.jl/pull/2228/files#diff-084f63e064c680ae660e2676e15d3d4eR221 you do not use a default `getindex` (which does this test) but a custom one (which does not do it).

Is this clear now what I meant? Thank you for working on this!

pdeffebach:
Thanks for the catch. I have removed the methods. I now only check `isunique` for string and symbol vectors and the integers get the check in the default `getindex` method. 

This should be ready to go.

bkamins:
Looks good now. I approve it as it can be merged from the functionality perspective.

However, can you please add update to the documentation? The manual page describing indexing should cover special cases where we allow invalid indices.

docs/src/man/getting_started.md
```diff
@@ -481,6 +481,24 @@ julia> df[!, Not(:x1)]
 │ 1   │ 2     │ 3     │
 ```
 
+The `Not` selector will ignore `String` and `Symbol` inputs that are not in the data frame.
```
bkamins, line 484:
```suggestion
The `Not` selector will ignore string and `Symbol` inputs that are not in the data frame. 
```

bkamins:
Thank you for the update of the getting started example.

What I meant earlier though is that it is crucial to update docs/src/lib/indexing.md with a description of precise rules - in the line where `Not` is introduced, as an additional comment (as this is a documentation and it has to be definitive description of the behaviour, while getting started - as you did it now - should give an intuitive explanation of the rule).

docs/src/lib/indexing.md
```diff
@@ -29,6 +29,12 @@ The rules for a valid type of index into a column are the following:
     * an `All` or `Between` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
     * a colon literal `:`.
 
+Indexing with `Not` has special behavior, in particular:
+* `String`s and `Symbol`s that do not belong in the column names of the data frame are ignored. The same applies for elements of `AbstractVector{<:String}` and `AbstractVector{Symbol}` that do not belong in the column names of the data frame;
```
bkamins, line 33:
```suggestion
* strings and `Symbol`s that do not belong in the column names of the data frame are ignored. The same applies for elements of `AbstractVector{<:AbstractString}` and `AbstractVector{Symbol}` that do not belong in the column names of the data frame; 
```

bkamins: I would also add that the same applies to the case if they are present in `All` expressions.

On the other hand - the three rules you described below are not special - they follow the standard rules for `Not` from InvertedIndices.jl. So we can do two things - either just remove them, or if you think it is worth to mention it, change the line above to be more verbose and say something like:
```
Indexing with `Not` selects indices that are a complement of the selection specified by its argument, in particular:
```

nalimilan:
Sorry for not reacting again earlier. After discussing this deeper with @bkamins I'm still not happy with the inconsistency, and I've also realized that we need something more general to allow trying to select columns that may not exist.

Even dplyr, which is quite flexible, doesn't allow negating non-existent names with `!` as noted by @CameronBieganek at https://github.com/JuliaData/DataFrames.jl/issues/2197#issuecomment-614952934. What they do instead is to require you to do `!any_of(...)`, whose [documentation](https://tidyselect.r-lib.org/reference/all_of.html) says "It is especially useful with negative selections, when you would like to make sure a variable is removed." That `any_of` function has the advantage that it also works when positively selecting columns.

We agreed with @bkamins that it would be better to adopt a powerful design along these lines. We already have `All` and recently introduced `Cols` in DataAPI, so we could use these to parallel the dplyr distinction. But our preferred solution so far is to add an argument to `Cols` to specify that it's fine for columns not to exist. Then you could do something like `Not(Cols(...; strict=false))`. That would probably be easier to grasp than subtle differences between `All` and `Cols`, and it would avoid introducing yet another name like `AnyOf`. We could even print an error message pointing to that argument when a column doesn't exist. What do you think?

pdeffebach:
Conditional on not allowing `Not(:zzz)`, which I still think is a good idea and an acceptable inconsistency, I think the second best solution would be petition to add a keyword `strict` to `Not` types in `InvertedIndices.jl`.

nalimilan:
Let's try that and see what they think.

bkamins (changes requested):
@pdeffebach - do we want to resurrect this PR. I think we should assume that we cannot count on changes in external packages. What we do must be self contained to DataFrames.jl

pdeffebach:
Let's close this. R and Stata don't support this, and typos are common.

PR #2233, opened 2020-05-07
Fix docstring references in the manual

### LINKED ISSUE #2232: first/last/etc. documentation problem
http://juliadata.github.io/DataFrames.jl/stable/lib/types/

Below the "Common methods" header, some of the links are broken.
* first
* last
* unique
perhaps more.

mortenpi:
Try to fix #2232.

### DIFF
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -87,7 +87,6 @@ last
 unique
 propertynames
 similar
-ourshow
 ourstrwidth
 sortperm
 pairs
@@ -95,7 +94,6 @@ parent
 issorted
 getmaxwidths
 getprintedwidth
-gennames
 getchunkbounds
 showrowindices
 compacttype
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -79,14 +79,25 @@ unique!
 vcat
 ```
 
+## Unsorted
+
 ```@docs
-join
-summary
 first
 last
 unique
-isequal
-sum
-transform!
 propertynames
+similar
+ourshow
+ourstrwidth
+sortperm
+pairs
+parent
+issorted
+getmaxwidths
+getprintedwidth
+gennames
+getchunkbounds
+showrowindices
+compacttype
+showrows
 ```
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -18,7 +18,6 @@ outerjoin
 antijoin
 semijoin
 crossjoin
-by
 combine
 groupby
 groupindices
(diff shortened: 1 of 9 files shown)


mortenpi:
Ok, this should now actually resolve all the docstring referencing issues:

* Remove references to docstring that don't exist from the manual
* Remove at-refs to functions that don't have (DataFrames-specific) docstrings anymore
* Add in all the docstrings that exists, but were not included in the manual.

I also enabled `strict = true` for `makedocs` -- this should fail the documentation build if a PR introduces some new referencing problems.

docs/src/lib/functions.md
```diff
@@ -80,3 +78,26 @@ transform!
...
+parent
+issorted
+getmaxwidths
+getprintedwidth
+gennames
+getchunkbounds
+showrowindices
+compacttype
+showrows
+```
```
mortenpi, line 103: These are all the docstrings that were not included in the manual at the moment. These should be put into their appropriate sections in the manual before merging this, but I would prefer to leave this to someone who is more familiar with the overall structure of the package :slightly_smiling_face:

bkamins: Thank you for doing this. Are we allowed to exclude the unexported functions from this list? (e.g. `ourshow` or `gennames`)

mortenpi: They can't be completely removed from the manual, without Documenter complaining, but I would propose having a (hidden) internals page to get around this (3885ff3ad14ff7be3269a64b41ebd584c753f3db).

bkamins: OK - thx.

docs/src/lib/functions.md
```diff
@@ -80,3 +78,24 @@ transform!
...
+sortperm
+pairs
+parent
+issorted
+getmaxwidths
+getprintedwidth
+getchunkbounds
+showrowindices
+compacttype
+showrows
```
bkamins, line 100:
```suggestion
first
last
unique
propertynames
similar
sortperm
pairs
parent
issorted
```

docs/src/lib/internals.md
```diff
@@ -0,0 +1,15 @@
...
+# Internals
+
+!!! warning "Internal API"
+
+    The functions, methods and types listed on this page are internal to DataFrames and are
+    **not considered to be part of the public API**.
+
+```@docs
+ourshow
+gennames
```
bkamins, line 14:
```suggestion
compacttype
gennames
getchunkbounds
getmaxwidths
getprintedwidth
ourshow
ourstrwidth
showrowindices
showrows
```

bkamins:
Thank you!

PR #2249, opened 2020-05-12
better deprecations of by and map

bkamins:
Now the deprecations are more correct (still we generate other column names in auto-generation of column names mode but I would ignore it). The cost is that they are messy.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -441,7 +441,8 @@ end
 @deprecate deleterows!(df::DataFrame, inds) delete!(df, inds)
 
 @deprecate by(f::Base.Callable, d::AbstractDataFrame, cols::Any;
-    sort::Bool=false, skipmissing::Bool=false) combine(f, groupby(d, cols, sort=sort,
+    sort::Bool=false, skipmissing::Bool=false) combine(f,
+    groupby(d, cols, sort=sort, skipmissing=skipmissing))
 @deprecate by(f::Pair{<:ColumnIndex}, d::AbstractDataFrame, cols::Any;
     sort::Bool=false, skipmissing::Bool=false) combine(f,
     groupby(d, cols, sort=sort, skipmissing=skipmissing))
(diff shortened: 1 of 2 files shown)


bkamins:
@nalimilan - only Coverage fails here

src/deprecated.jl
```diff
@@ -447,20 +440,29 @@ end
...
 @deprecate deleterows!(df::DataFrame, inds) delete!(df, inds)
 
-@deprecate by(f::Union{Base.Callable, Pair}, d::AbstractDataFrame, cols::Any;
-   sort::Bool=false, skipmissing::Bool=false) combine(groupby(d, cols, sort=sort,
-                                                              skipmissing=skipmissing), f)
+@deprecate by(f::Base.Callable, d::AbstractDataFrame, cols::Any;
+    sort::Bool=false, skipmissing::Bool=false) combine(f, groupby(d, cols, sort=sort,
+                                                                 skipmissing=skipmissing))
+@deprecate by(f::Pair, d::AbstractDataFrame, cols::Any;
+    sort::Bool=false, skipmissing::Bool=false) combine(first(f) isa ColumnIndex ?
```
nalimilan, line 447: Can we print different messages depending on whether `first(f) isa ColumnIndex`? Otherwise the deprecation message is going to be totally scary.

bkamins: The problem is that if you have something like:
```
by(df, :a, :a => f1, [:a, :b] => f2)
```
there is a combination of both. But in the single argument case it can be improved

nalimilan: Yeah, it's hard to do something perfect in all cases, but if we can make simple cases simple it's much nicer for users.

bkamins: It seems that we will re-introduce `by` so I will wait with this PR till we make that decision :).

bkamins: OK - it seems we drop `by` so I have updated the deprecations

bkamins:
Thank you!

PR #2250, opened 2020-05-13
add support for extended help

cormullion:
A first pass at adding support for Extended help. It looks like this:

Screenshot 2020-05-13 at 09 31 07

It's not always easy to judge the right amount of help before inserting `Extended help`, but a rough guide was to make sure you can still see the function definition when the `julia>` prompt reappears...

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -6,6 +6,16 @@ An AbstractDataFrame that stores a set of named columns
 The columns are normally AbstractVectors stored in memory,
 particularly a Vector or CategoricalVector.
 
+# Examples
+```julia
+df = DataFrame()
+v = ["x","y","z"][rand(1:3, 10)]
+df1 = DataFrame(Any[collect(1:10), v, rand(10)], [:A, :B, :C])
+df2 = DataFrame(A = 1:10, B = v, C = rand(10))
+```
+
+# Extended help
+
 # Constructors
 ```julia
 DataFrame(columns::AbstractVector, names::AbstractVector{Symbol};
@@ -671,6 +681,8 @@ Insert a column into a data frame in place. Return the updated `DataFrame`.
 If `ind` is omitted it is set to `ncol(df)+1`
 (the column is inserted as the last column).
 
+# Extended help
+
 # Arguments
 - `df` : the DataFrame to which we want to add columns
 - `ind` : a position at which we want to insert a column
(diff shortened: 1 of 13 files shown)


src/dataframe/dataframe.jl
```diff
@@ -6,6 +6,16 @@ An AbstractDataFrame that stores a set of named columns
 The columns are normally AbstractVectors stored in memory,
 particularly a Vector or CategoricalVector.
 
+# Examples
```
bkamins, line 9: We have examples below. I think we should not duplicate it. @nalimilan - do you have an opinion?

nalimilan:
Thanks for this! It prompted me to [comment](https://github.com/JuliaLang/julia/pull/34226#issuecomment-630110183) on the PR which added extended help support to Julia, as it seems like we could avoid adding most of these if extended help automatically stopped at `# Examples`.

src/dataframe/dataframe.jl
cormullion: Is the duplication the problem, or is having examples immediately available t..he problem?

Screenshot 2020-05-20 at 10 24 16

It's frustrating for a user to ask for help then to have to ask again because the previous answer wasn't helpful.

nalimilan: Both I'd say. Examples shouldn't generally be presented in the short help IMHO.

cormullion: I think we'll agree to disagree :) I think almost the opposite!

cormullion: Let's say you've forgotten the syntax for `replace()`. I'd show this:

Screenshot 2020-05-20 at 10 36 11

But you would show this:

Screenshot 2020-05-20 at 10 37 15

I'd say my way was helpful.

nalimilan: Actually I think we need clearer guidance regarding what should be put in the extended help in general. Your proposal amounts to putting the full `replace` docstring in the short help. :-)

cormullion: Yes, that's a fair summary! :) My initial understanding of the purpose behind @tholy's "extended help" PR was to provide the user with a readable and useful amount of information in a single screen without forcing them to scroll back through 100s of lines of small print, LaTeX formulas, exhaustive analyses, and scholarly references just to find the basic syntax for the function. I didn't think the concept of "short help without examples" was being proposed...

nalimilan: OK. But then you probably should have kept `# Arguments` in the short help in this PR, right? Basically, most docstrings shouldn't be changed at all.

cormullion: "It's not always easy to judge the right amount of help before inserting Extended help, but a rough guide was to make sure you can still see the function definition when the julia> prompt reappears..." - sometimes I may have guessed too quickly that there was too much information for a screen. 

I don't think I'll be doing more of these...

PR #2237, opened 2020-05-08
group functions by their roles in the manual

bkamins:
Following @mortenpi PR.

PR #2262, opened 2020-05-17
add empty and empty!

### LINKED ISSUE #2251: Implement `empty` / `empty!` as the truncate operation
Proposal to have these definitions implemented and added to the API
```
empty(df::AbstractDataFrame) = similar(df, 0)
empty!(df::AbstractDataFrame) = filter!(row -> false, df)
```

### LINKED ISSUE #2261: Rename `tail` to avoid clash with `Base.tail`
I typically import `Base: tail` into scope. If I then do `using DataFrames`, I get a warning

```
WARNING: using DataFrames.tail in module Main conflicts with an existing identifier.
```

Although `Base.tail` is not exported, it is documented (https://docs.julialang.org/en/v1/base/collections/#Base.tail) and I think many people use it regularly.

To avoid this clash, maybe `DataFrames` can use a different name for this function? Or it could just extend `Base.tail`.

bkamins:
Fixes #2251, #2261

@nalimilan - I think what @Nosferican proposed in #2251 can be implemented before 1.0. At the same time I propose to remove rarely used and long-deprecated functions from src/deprecated.jl, this will resolve #2261 in particular.


bkamins:
This PR should not be merged before https://github.com/JuliaData/DataFrames.jl/pull/2237 (I would like to have documentation cleaned up, and then I will rebase it)

bkamins:
@nalimilan - I have reverted the things that are not problematic to keep as deprecated. Now I remove only definitions that are really bad IMO and should not exist even as deprecations.

docs/src/lib/functions.md
```diff
@@ -91,4 +91,6 @@ sortperm
 pairs
 parent
 issorted
+empty!
+empty
```
nalimilan, line 95:
```suggestion
empty
empty!
```

bkamins:
Thank you!

cossio:
Thanks for fixing this!

Issue #2241, opened 2020-05-08
Problems in groupreduce_init


bkamins:
@nalimilan. We have two issues with `Reduce` design related to the `eltype` identified by `groupreduce_init`.

The first is:
```
julia> df = DataFrame(g=[1,1,1,2,2,2], x=Any[1,1,1,1.5,1.5,1.5])
6×2 DataFrame
│ Row │ g     │ x   │
│     │ Int64 │ Any │
├─────┼───────┼─────┤
│ 1   │ 1     │ 1   │
│ 2   │ 1     │ 1   │
│ 3   │ 1     │ 1   │
│ 4   │ 2     │ 1.5 │
│ 5   │ 2     │ 1.5 │
│ 6   │ 2     │ 1.5 │

julia> combine(groupby(df, :g), :x => sum)
ERROR: InexactError: Int64(1.5)
```
or
```
julia> df = DataFrame(g=[1,1,1,2,2,2], x=Any[1,1,1,1,1,missing])
6×2 DataFrame
│ Row │ g     │ x       │
│     │ Int64 │ Any     │
├─────┼───────┼─────────┤
│ 1   │ 1     │ 1       │
│ 2   │ 1     │ 1       │
│ 3   │ 1     │ 1       │
│ 4   │ 2     │ 1       │
│ 5   │ 2     │ 1       │
│ 6   │ 2     │ missing │

julia> combine(groupby(df, :g), :x => sum)
ERROR: MethodError: Cannot `convert` an object of type Missing to an object of type Int64
```

The second is that after https://github.com/JuliaLang/Statistics.jl/pull/25 we do not consistently identify `eltype` of the vector that is to store aggregated values (in DataFrames.jl if we get ints we keep them as ints, which can overflow - Base does a conversion to florat).

Proposed approach to solve it:
1. in `groupreduce_init` check if a column that is aggregated has a concrete type or a `Union` of a concrete type and `missing`; if yes do a fast path; if not - we have to go through the whole vector  to check what we allow
2. pass `adjust` to `groupreduce_init` and if it is `/` then appropriately set `eltype` of `res` to take this fact into account

nalimilan:
[...]

PR #2263, opened 2020-05-19
fix bug in reduceinit

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2241.

I have bumped version in the PR. When we merge it I will cherry pick it (+ maybe some other non-breaking PRs we have merged) and make a 0.21.1 release.

### DIFF
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -876,24 +876,17 @@ function groupreduce_init(op, condf, adjust,
         else
             V = U >: Missing ? Union{typeof(x), Missing} : typeof(x)
         end
+        v = similar(incol, V, length(gd))
+        fill!(v, x)
+        return v
     else
-        idx = findfirst(!ismissing, incol)
-        # here the definition of x is simpler as we do V = Any anyway
-        if isnothing(idx)
-
-            end
-        else
-            x = initf(incol[idx])
-        end
-        # do not try to determine the narrowest possible type as this is not
-        # possible to do correctly in general without processing groups
-        # it will get fixed later in groupreduce!
-        V = Any
+        # do not try to determine the narrowest possible type nor statring value
+        # as this is not possible to do correctly in general without processing
+        # groups it will get fixed later in groupreduce!; later we
+        # will make use of the fact that this vector is filled with #undef
+        # while above the vector is filled with a concrete value
+        return Vector{Any}(undef, length(gd))
     end
-    v = similar(incol, V, length(gd))
-
-    fill!(v, x)
-    return v
 end
 
 for (op, initf) in ((:max, :typemin), (:min, :typemax))
@@ -901,8 +894,9 @@ for (op, initf) in ((:max, :typemin), (:min, :typemax))
         function groupreduce_init(::typeof($op), condf, adjust,
                                   incol::AbstractVector{T}, gd::GroupedDataFrame) where T
             @assert isnothing(adjust)
+            S = nonmissingtype(T)
             # !ismissing check is purely an optimization to avoid a copy later
-            outcol = similar(incol, condf === !ismissing ? nonmissingtype(T) : T, length(gd))
+            outcol = similar(incol, condf === !ismissing ? S : T, length(gd))
             # Comparison is possible only between CatValues from the same pool
             if incol isa CategoricalVector
                 U = Union{CategoricalArrays.leveltype(outcol),
@@ -911,7 +905,7 @@ for (op, initf) in ((:max, :typemin), (:min, :typemax))
             end
             # It is safe to use a non-missing init value
             # since missing will poison the result if present
-            S = nonmissingtype(T)
+            # we assume here that groups are non-empty (current design assures this)
             if isconcretetype(S) && hasmethod($initf, Tuple{S})
                 fill!(outcol, $initf(S))
             else
@@ -936,8 +930,8 @@ function copyto_widen!(res::AbstractVector{T}, x::AbstractVector) where T
     return res
 end
 
-function groupreduce!(res, f, op, condf, adjust, checkempty::Bool,
-                      incol::AbstractVector{T}, gd::GroupedDataFrame) where T
+function groupreduce!(res::AbstractVector{U}, f, op, condf, adjust, checkempty::Bool,
+                      incol::AbstractVector{T}, gd::GroupedDataFrame) where {U,T}
     n = length(gd)
     if adjust !== nothing || checkempty
         counts = zeros(Int, n)
@@ -947,26 +941,48 @@ function groupreduce!(res, f, op, condf, adjust, checkempty::Bool,
         gix = groups[i]
         x = incol[i]
         if gix > 0 && (condf === nothing || condf(x))
-            res[gix] = op(res[gix], f(x, gix))
+            # this check should be optimized out if T is not Any
+            if U === Any && !isassigned(res, gix)
+                res[gix] = f(x, gix)
+            else
+                res[gix] = op(res[gix], f(x, gix))
+            end
             if adjust !== nothing || checkempty
                 counts[gix] += 1
             end
         end
     end
-    outcol = adjust === nothing ? res : map(adjust, res, counts)
+    # handle the case of an unitialized reduction
+    if U === Any
+        if op isa typeof(Base.add_sum)
+            initf = zero
+        elseif op isa typeof(Base.mul_prod)
+            initf = one
+        else
+            initf = x -> throw(ErrorException("Unrecognized op $op"))
+        end
+        @inbounds for gix in eachindex(res)
+            if !isassigned(res, gix)
+                res[gix] = initf(nonmissingtype(T))
+            end
+        end
+    end
+    if adjust !== nothing
+        res .= adjust.(res, counts)
+    end
     if checkempty && any(iszero, counts)
         throw(ArgumentError("some groups contain only missing values"))
     end
     # Undo pool sharing done by groupreduce_init
-    if outcol isa CategoricalVector && outcol.pool === incol.pool
-        U = Union{CategoricalArrays.leveltype(outcol),
-                  eltype(outcol) >: Missing ? Missing : Union{}}
-        outcol = CategoricalArray{U, 1}(outcol.refs, copy(outcol.pool))
+    if res isa CategoricalVector && res.pool === incol.pool
+        V = Union{CategoricalArrays.leveltype(res),
+                  eltype(res) >: Missing ? Missing : Union{}}
+        res = CategoricalArray{V, 1}(res.refs, copy(res.pool))
     end
-    if isconcretetype(eltype(outcol))
-        return outcol
+    if isconcretetype(eltype(res))
+        return res
     else
-        copyto_widen!(Tables.allocatecolumn(typeof(first(outcol)), n), outcol)
+        return copyto_widen!(Tables.allocatecolumn(typeof(first(res)), n), res)
     end
 end
 
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -867,7 +867,7 @@ function groupreduce_init(op, condf, adjust,
     end
 
     Tnm = nonmissingtype(T)
-    if Tnm !== Any && isconcretetype(Tnm) && applicable(initf, Tnm)
+    if isconcretetype(Tnm) && applicable(initf, Tnm)
         tmpv = initf(Tnm)
         initv = op(tmpv, tmpv)
         x = adjust isa Nothing ? initv : adjust(initv, 1)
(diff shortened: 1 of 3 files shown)


src/groupeddataframe/splitapplycombine.jl
```diff
@@ -855,11 +855,45 @@ end
...
+        else
+            initv = f(incol[idx])
+            x = adjust isa typeof(/) ? (initv / 1) : initv
+        end
+        # do not try to determine the narrowest possible type as this is not
+        # possible to do correctly in general without processing groups
+        # it will get fixed later in groupreduce!
+        V = Any
+    end
+    v = Vector{V}(undef, length(gd))
```
bkamins, line 887: @nalimilan - just to be sure - it is impossible here that `v` should be a `CategoricalVector` here - right?
(I ask this because in code below I have seen that we check against `CategoricalVector` in reductions and I was not sure why).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -855,11 +855,45 @@ end
 
 # Use reducedim_init to get a vector of the right type,
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -855,11 +855,45 @@ end
...
 # but trick it into using the expected length
-groupreduce_init(op, condf, incol, gd) =
-    Base.reducedim_init(identity, op, view(incol, 1:length(gd)), 2)
+function groupreduce_init(op, condf, adjust, incol, gd)
+    U = eltype(incol)
+    T = Base.promote_union(U)
+
+    if op isa typeof(Base.add_sum)
+        f = zero
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -855,11 +855,45 @@ end
...
+    elseif op isa typeof(Base.mul_prod)
+        f = one
+    else
+        throw(ErrorException("Unrecognized op $op"))
+    end
+
+    if T !== Any && applicable(f, T)
+        initv = f(T)
+        x = adjust isa typeof(/) ? (initv / 1) : initv
```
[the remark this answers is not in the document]
bkamins: in the future we can potentially have different `adjust` types so I would prefer to be explicit here.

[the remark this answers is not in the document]
nalimilan: That's the point. `ajust` is guaranteed to be passed the final value and the counts, so `adjust(initv, 1)` is supposed to give you a correct init value, right?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -855,11 +855,45 @@ end
...
+        V = U >: Missing ? Union{typeof(x), Missing} : typeof(x)
```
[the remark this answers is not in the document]
bkamins: I thought your design idea was to check `condf` in https://github.com/JuliaData/DataFrames.jl/pull/2263/files#diff-23657e51a9cc9e627fc153ba1e6e04c1R974 below.

> this assumes addition/subtraction are type stable

A good point - it wrongly assumes:
```
julia> typeof(zero(Real))
Int64

julia> big(1) isa Real
true
```
which will lead to bugs. I will fix this.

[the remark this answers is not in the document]
bkamins: See dump from my tests :):
```
julia> combine(gdf, :x => sum∘skipmissing);
op = Base.add_sum
condf = Base.var"#66#67"{typeof(ismissing)}(ismissing)
adjust = nothing
incol = Real[1, 9223372036854775807, 1, 1, 1, 1]
v = [139634466330832, 139634465689520]
ERROR: InexactError: Int64(9223372036854775808)
```

src/groupeddataframe/splitapplycombine.jl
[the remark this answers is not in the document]
bkamins: OK

src/groupeddataframe/splitapplycombine.jl
bkamins: I have changed to `similar(incol, V, length(gd))` but it really I cannot see in practice the case where it will differ (a quick look at `methods` of `similar` shows that normally it will always produce `Vector`).

src/groupeddataframe/splitapplycombine.jl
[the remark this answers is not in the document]
bkamins: OK - I have also added handling for:

> This doesn't give the right eltype for `sum ∘ skipmissing`, does it?

Though this should not change much

bkamins:
I have rewritten the internals. A careful look would be welcome :).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -853,13 +853,58 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
...
+    else
+        throw(ErrorException("Unrecognized op $op"))
+    end
+
+    Tnm = nonmissingtype(T)
+    if Tnm !== Any && isconcretetype(Tnm) && applicable(initf, Tnm)
+        tmpv = initf(Tnm)
+        initv = op(tmpv, tmpv)
+        x = adjust isa Nothing ? initv : adjust(initv, 1)
+        if condf isa typeof(!ismissing)
```
nalimilan, line 874:
```suggestion
        if condf === !ismissing
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -853,13 +853,58 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
...
+            V = typeof(x)
+        else
+            V = U >: Missing ? Union{typeof(x), Missing} : typeof(x)
+        end
+    else
+        idx = findfirst(!ismissing, incol)
+        # here the definition of x is simpler as we do V = Any anyway
+        if isnothing(idx)
+            if condf isa typeof(!ismissing)
```
nalimilan, line 883:
```suggestion
            if condf === !ismissing
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -853,13 +853,58 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
...
+                throw(ArgumentError("Aggregating over a column containing " *
+                                    "only missing values that does not have a " *
+                                    "concrete union element type is not allowed"))
```
nalimilan, line 886: Shouldn't we check `applicable(initf, T)`? As you noted, `zero(Real)` is defined so we could fill a `Vector{Any}` or `Vector{Real}` with that. Maybe even do that in all cases before looking for the first non-missing value as it's faster.

test/grouping.jl
```diff
@@ -2217,4 +2218,72 @@ end
     end
 end
 
+@testset "corner cases of group_reduce" begin
+    df = DataFrame(g=[1,1,1,2,2,2], x=Any[1,1,1,1.5,1.5,1.5])
+    gdf = groupby_checked(df, :g)
+    @test combine(groupby(df, :g), :x => sum) == DataFrame(g=1:2, x_sum=[3.0, 4.5])
```
nalimilan, line 2224: Also need to check the eltype.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -853,13 +853,58 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
     outcol
 end
 
-# Use reducedim_init to get a vector of the right type,
-# but trick it into using the expected length
-groupreduce_init(op, condf, incol, gd) =
-    Base.reducedim_init(identity, op, view(incol, 1:length(gd)), 2)
+# Use a strategy similar to reducedim_init from Base to get the vector of the right type,
```
nalimilan, line 856:
```suggestion
# Use a strategy similar to reducedim_init from Base to get the vector of the right type
```

src/groupeddataframe/splitapplycombine.jl
bkamins: OK - we can fall back here as you propose. However, we have to do `findfirst`, as otherwise we have the following problem:
```
julia> x = Real[UInt8(250), UInt8(250)]
2-element Array{Real,1}:
 0xfa
 0xfa

julia> sum(x)
0x00000000000001f4

julia> Base.add_sum(Base.add_sum(zero(Real), x[1]), x[2])
500
```
and we would be inconsistent with Base.

test/grouping.jl
bkamins: I will add the check, but please note that this check does not tell us what `groupreduce_init` returned, as the final eltype of the column is determined elsewhere.

bkamins:
In the test I hopefully covered all relevant cases.

src/groupeddataframe/splitapplycombine.jl
bkamins: > Maybe even do that in all cases before looking for the first non-missing value as it's faster.

In the end you are probably right and I will change to that. But the reason is different. We are not sure which values we will encounter in subgroups. I will propose yet a different implemenation

bkamins: Pondering on this even more I came to the conclusion that we should use yet a different strategy because of:
```
julia> sum(["a"])
"a"
```
I will try to propose something efficient for this (in general using `fillfirst!` I think is what we should do).

If the behaviour of `sum` is not intended then we should use a different strategy so please comment how you see this. Thank you!.

bkamins: I just implemented what Base does

bkamins:
The design I now proposed seems to be more robust that the earlier approaches. There is a bit of risk of performance drop, but normally the compiler should optimize-out the branches. I will test performance when we are happy with the design.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -906,38 +941,60 @@ function groupreduce!(res, f, op, condf, adjust, checkempty::Bool,
...
+            end
             if adjust !== nothing || checkempty
                 counts[gix] += 1
             end
         end
     end
-    outcol = adjust === nothing ? res : map(adjust, res, counts)
+    # handle the case of an unitialized reduction
+    if U === Any
+        if op isa typeof(Base.add_sum)
```
nalimilan, line 957:
```suggestion
        if op === Base.add_sum
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -906,38 +941,60 @@ function groupreduce!(res, f, op, condf, adjust, checkempty::Bool,
...
+            initf = zero
+        elseif op isa typeof(Base.mul_prod)
```
nalimilan, line 959:
```suggestion
        elseif op === Base.mul_prod
```

src/groupeddataframe/splitapplycombine.jl
nalimilan: OK. It's too bad that we have to make things so complex to match what Base does. But in the end you didn't use `fillfirst!`, right? Wouldn't it be simpler for us?

bkamins: `fillfirst!` is used for `max` and `min` like operations as there it is safe. We could use `fillfirst!` for `add` and `prod` like operations but then the result of `fillfirst!` would have to be changed to `zero` or `one` in order to get a neutral element. This would mean that the "hard cases" at the end (which match Base behavior) would throw an error, so we would have an inconsistency between `combine(gdf, :x=>sum)` and `combine(gdf, :x=>(x->sum(x))` unfortunately. Also what I do now is faster than using `fillfirst!` (though the gain is probably non-noticeable).

test/grouping.jl
```diff
@@ -2217,4 +2218,102 @@ end
...
+    @test combine(gdf, :x => sum)[1, 2] isa Missing
+    @test eltype(combine(gdf, :x => sum)[!, 2]) === Missing
+    @test_throws MethodError combine(gdf, :x => sum∘skipmissing)
+
+    # these questions can go to a final exam in "mastering combine" class
+    df = DataFrame(g=[1,2,3], x=["a", "b", "c"])
+    gdf = groupby_checked(df, :g)
+    @test combine(gdf, :x => sum => :a, :x => prod => :b) ==
+          combine(gdf, :x => (x -> sum(x)) => :a, :x => (x -> prod(x)) => :b)
+    df = DataFrame(g=[1,2,3], x=Any["a", "b", "c"])
```
nalimilan, line 2305:
```suggestion
    df = DataFrame(g=[1,2,3], x=["a", "b", "c"])
```

bkamins:
I see you change `isa` tests to `===` tests. Out of curiosity - why do you prefer this?

src/groupeddataframe/splitapplycombine.jl
nalimilan: Ah yes this `sum(["a"])` thing is really absurd, but I don't see how Base could handle this differently.

One case where performance is going to be suboptimal is for small unions like `Union{Int8,Int32}` or `Union{Nothing,Int}`. It probably doesn't matter a lot, but in the future we could imagine adding a fast path using `fillfirst!` for `Number`, where we can assume `zero` and `one` are defined. Though even in these cases it would be faster only if the number of groups is relatively small I guess.

nalimilan:
> I see you change `isa` tests to `===` tests. Out of curiosity - why do you prefer this?

It's shorter and easier to read when you need to do `isa typeof(...)`. Otherwise I don't have a strong preference.

src/groupeddataframe/splitapplycombine.jl
bkamins: Right, but I would not think that `Union{Int8,Int32}` or `Union{Nothing,Int}` are type specifications we should put a priority on - agreed?

nalimilan: Right.

bkamins:
OK - when this is merged I will make a patch release in a separate branch also.

Issue #2264, opened 2020-05-20
Index a grouped data frame by a vector of pairs


pdeffebach:
With the new ability to index with symbols, it's hard to index a grouped data frame with a named tuple when you are working with column names as strings.

bkamins:
That is why it is allowed to index them by `Tuple`s also. Is there something you could propose alternatively (also as I have commented on discourse - it is easy to create a `NamedTuple` from a `DataFrameRow` - as you know :smile:).

Still - do you have some proposal what we could change? Maybe we can additionally allow iterables of `key => value` paris?

pdeffebach:
See #2265 for an implementation. I think it's readable and makes strings "first class" etc.

bkamins:
done on master

PR #2265, opened 2020-05-20
Add String => Any grouped indexing

pdeffebach:
In reference to #2264 

I think this is a bit more intuitive than the current `Tuple` interface. A bit more readable. 

I couldn't quite figure out the right way to test this. I added a helper function but need to figure out how to use it better.

### DIFF
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -316,6 +316,12 @@ end
 
 Base.to_index(gd::GroupedDataFrame, key::Tuple) = gd.keymap[key]
 
+Base.to_index(gd::GroupedDataFrame, key::Pair) = 
+    Base.to_index(gd, (; Symbol(first(key)) => last(key)))
+
+Base.to_index(gd::GroupedDataFrame, key::Vector{Pair}) = 
+    Base.to_index(gd, (; [Symbol(first(k)) => last(k) for k in keys]...))
+
 function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(N, gd.cols))
         throw(KeyError(key))
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -323,8 +323,8 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
-function Base.to_index(gd::GroupedDataFrame, key::Union{Tuple{Vararg{Pair{Symbol}}}), Tuple{Vararg{Pair{<:AbstractString}}})}
-    if length(key) != length(gd.cols) || any(Symbol(first(k)) != _names(gd)[c] for (n, c) in zip(key, gd.cols))
+function Base.to_index(gd::GroupedDataFrame, key::Union{Tuple{Vararg{Pair{Symbol}}}, Tuple{Vararg{Pair{<:AbstractString}}}})
+    if length(key) != length(gd.cols) || any(Symbol(first(n)) != _names(gd)[c] for (n, c) in zip(key, gd.cols))
         throw(KeyError(key))
     end
     return Base.to_index(gd, last.(key))
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -316,17 +316,19 @@ end
 
 Base.to_index(gd::GroupedDataFrame, key::Tuple) = gd.keymap[key]
 
-Base.to_index(gd::GroupedDataFrame, key::Pair) = 
-    Base.to_index(gd, (; Symbol(first(key)) => last(key)))
-
-Base.to_index(gd::GroupedDataFrame, key::Vector{Pair}) = 
-    Base.to_index(gd, (; [Symbol(first(k)) => last(k) for k in keys]...))
-
 function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(N, gd.cols))
         throw(KeyError(key))
     end
     return Base.to_index(gd, Tuple(key))
+end
+
+function Base.to_index(gd::GroupedDataFrame, key::Tuple{Vararg{<:Pair}})
+    println("hey there")
+    if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(Symbol.(first.(key)), gd.cols))
+        throw(KeyError(key))
+    end
+    return Base.to_index(gd, ntuple(i -> last(key[i]), length(key)))
 end
 
 # Array of (possibly non-standard) indices
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -323,12 +323,11 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
-function Base.to_index(gd::GroupedDataFrame, key::Tuple{Vararg{<:Pair}})
-    println("hey there")
-    if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(Symbol.(first.(key)), gd.cols))
+function Base.to_index(gd::GroupedDataFrame, key::Union{Tuple{Vararg{Pair{Symbol}}}), Tuple{Vararg{Pair{<:AbstractString}}})}
+    if length(key) != length(gd.cols) || any(Symbol(first(k)) != _names(gd)[c] for (n, c) in zip(key, gd.cols))
         throw(KeyError(key))
     end
-    return Base.to_index(gd, ntuple(i -> last(key[i]), length(key)))
+    return Base.to_index(gd, last.(key))
 end
 
 # Array of (possibly non-standard) indices
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -1326,6 +1326,17 @@ end
     df = DataFrame(a = repeat([:A, :B, missing], outer=4), b = repeat(1:2, inner=6), c = 1:12)
     gd = groupby_checked(df, [:a, :b])
 
+    function gkey_to_pair(key)
+        k = keys(key)
+        v = values(key)
+        if length(k) == 1 
+            return string(first(k)) => first(v)
+        else
+            return map(k, v) do ki, vi 
+                string(ki) => vi
+            end
+        end
+    end
     @test map(NamedTuple, keys(gd)) ≅
         [(a=:A, b=1), (a=:B, b=1), (a=missing, b=1), (a=:A, b=2), (a=:B, b=2), (a=missing, b=2)]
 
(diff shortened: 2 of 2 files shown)


src/groupeddataframe/groupeddataframe.jl
```diff
@@ -316,6 +316,12 @@ end
 
 Base.to_index(gd::GroupedDataFrame, key::Tuple) = gd.keymap[key]
 
+Base.to_index(gd::GroupedDataFrame, key::Pair) =
```
bkamins, line 319: We do not allow this form for other styles, but probably it is OK to allow this, as grouping by a single column is pretty common.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -316,6 +316,12 @@ end
 
 Base.to_index(gd::GroupedDataFrame, key::Tuple) = gd.keymap[key]
 
+Base.to_index(gd::GroupedDataFrame, key::Pair) = 
+    Base.to_index(gd, (; Symbol(first(key)) => last(key)))
+
+Base.to_index(gd::GroupedDataFrame, key::Vector{Pair}) =
```
bkamins, line 322: Current version will not work due to type invariance in Julia. Probably it is better to use:
```suggestion
Base.to_index(gd::GroupedDataFrame, key::Union{AbstractVector{<:Pair}, Base.Iterators.Pairs}) = 
```

bkamins: ah - maybe `Union` is not OK because of https://github.com/JuliaData/DataFrames.jl/pull/2265/files#diff-6349421a054bb7e74b7f27ae86304cbaR333 and current method dispatch rules.

Still then https://github.com/JuliaData/DataFrames.jl/pull/2265/files#diff-6349421a054bb7e74b7f27ae86304cbaR333 requires also an update to allow vector of vectors of pairs.

src/groupeddataframe/groupeddataframe.jl
bkamins: Ah - now I remember why we disallow it - it will be ambiguous if `AbstractVector{<:Pair}` should select one cell based on many pairs or it should select multiple cells based on a single pair each.

pdeffebach: I am fine if we only allow vectors of `Pairs`.

bkamins:
@pdeffebach - as usual I think it is best to write down the API specification, then write tests, then implement the functionality against the tests. Otherwise we will be not sure that what we design is consistent.

I think it is best if you start with https://github.com/JuliaData/DataFrames.jl/pull/2265/files#diff-6349421a054bb7e74b7f27ae86304cbaL306 definitions. Then there is a specification of vector indexing as I have noted.

When I slept over it the problem with vectors of pairs is that they conflict with API for selecting collections of groups. Maybe it will be better to allow tuples of pairs? It is not the end of the world - it can be worked around, but simply keeping `GroupKeyTypes` union unchanged would be simpler.

In general note that accepting pairs is only a convenience thing. Internally we use `Tuple` indexing anyway, and use names only to make sure all is consistent.

pdeffebach:
Thank you for your comments. I understand the way the code is structured much more now. I have reworked so that we have a Tuple of Pairs, which is as close as we can get to a NamedTuple. Basically when you construct a named tuple with `Symbol`s and variables you write 

```
nt = (; :a => :A, :b => 1)
```

Unfortunately you can't do this with strings. But now you can write 

```
p = ("a" => :A, "b" => 1)
```

The API is consequently that a grouped data frame

```
julia> gd[("a" => :A, "b" => 2)]
hey there
2×3 SubDataFrame
│ Row │ a       │ b     │ c     │
│     │ Symbol? │ Int64 │ Int64 │
├─────┼─────────┼───────┼───────┤
│ 1   │ A       │ 2     │ 7     │
│ 2   │ A       │ 2     │ 10    │

julia> gd[[("a" => :A, "b" => 2), ("a" => :B, "b" => 1)]]
hey there
hey there
GroupedDataFrame with 2 groups based on keys: a, b
First Group (2 rows): a = :A, b = 2
│ Row │ a       │ b     │ c     │
│     │ Symbol? │ Int64 │ Int64 │
├─────┼─────────┼───────┼───────┤
│ 1   │ A       │ 2     │ 7     │
│ 2   │ A       │ 2     │ 10    │
⋮
Last Group (2 rows): a = :B, b = 1
│ Row │ a       │ b     │ c     │
│     │ Symbol? │ Int64 │ Int64 │
├─────┼─────────┼───────┼───────┤
│ 1   │ B       │ 1     │ 2     │
│ 2   │ B       │ 1     │ 5     │
```

I have added tests for the desired behavior.

bkamins:
> Basically when you construct a named tuple with Symbols and variables you write
> ```
> nt = (; :a => :A, :b => 1)
> ```
> Unfortunately you can't do this with strings. But now you can write
> ```
> p = ("a" => :A, "b" => 1)
> ```

This is an excellent example and I think it is a good design. Thank you for working on this. I will have a look at the implementation and comment.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,14 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function Base.to_index(gd::GroupedDataFrame, key::Tuple{Vararg{<:Pair}})
+    println("hey there")
```
bkamins, line 327:
```suggestion
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,14 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
+    if length(key) != length(gd.cols) || any(n != _names(gd)[c] for (n, c) in zip(Symbol.(first.(key)), gd.cols))
```
bkamins, line 328: This seems to be minimally faster (but your implementation is also OK, so please leave what you prefer)
```suggestion
    if length(key) != length(gd.cols) || any(Symbol(first(k)) != _names(gd)[c] for (n, c) in zip(key, gd.cols))
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,14 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
+        throw(KeyError(key))
+    end
+    return Base.to_index(gd, ntuple(i -> last(key[i]), length(key)))
```
bkamins, line 331: this seems to have the same performance and reads a bit cleaner to me:
```suggestion
    return Base.to_index(gd, last.(key))
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,14 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function Base.to_index(gd::GroupedDataFrame, key::Tuple{Vararg{<:Pair}})
```
bkamins, line 326: I think it is better to be more restrictive with the signature here:
```suggestion
function Base.to_index(gd::GroupedDataFrame, key::Union{Tuple{Vararg{Pair{Symbol}}}), Tuple{Vararg{Pair{<:AbstractString}}})}
```

bkamins:
I have commented in the code - these are minor issues (they assume we stay with `Tuple` design and not switch to `All` design).

Now when reading the code I have realized we have one ambiguity. `("a" => 1, "b" => 2)` can mean two things:
* column `"a"` having `1` and column `"b"` having `2` and
* first column having `"a" => 1` and second column having `"b" => 2`

I think that the risk of such ambiguity is minimal, but still it is possible. So we can do two things:
* if we stick with the current design we should clearly note that if someone has columns holding pairs one has to always use `NamedTuple` or a `Tuple`;
* use some wrapper for tuples with col => value pairs; actually we could use `All`, so one could then write `All("a" => 1, "b" => 2)` - this seems to be OK for me as it is not contradicting the meaning of `All` and we already have this type anyway imported. In this way we would have no ambiguity.

What do you think?

bkamins:
An alternative to `All` we might  consider adding a special constructor to `GroupKey`. So it would look `GroupKey("a" => 1, "b" => 2)`. This maybe would be cleanest and there would be no ambiguity.

pdeffebach:
I think better constructors for `GroupKey` and exposing that API to the user, is the best way to go. I would even go so far as to abandon the Tuple indexing. 

1. `Tuple(s::String)` actually broadcasts, so to make an index you actually need `Tuple(Ref(s))` or `(s,)`. Both are a bit ugly. 
2. We can specialize the constructor to allow keyword arguments to work similar to named tuples
3. We can special-case putting in pairs to provide a convenient syntax for if there is a column for of pairs.

pdeffebach:
> An alternative to `All` we might consider adding a special constructor to `GroupKey`. So it would look `GroupKey("a" => 1, "b" => 2)`. This maybe would be cleanest and there would be no ambiguity.

@nalimilan I think that instead of this PR we could do the following

1. Deprecate indexing with tuples entirely
2. Make more robust constructors for `GroupKey` which allows 
  * Postitional arguments of non-pairs or `Ref`s of `Pair`s
  * Keyword arguments
  * `Pair`s 

Then the positional arguments are lighter weight. The order of the inputs for keyword arguments won't matter (it currently does). 

Does this seem like a way forward? Too breaking? 

One motivation for this is that it's unintuitive to turn a string into a tuple. 


```
julia> df = DataFrame(a = ["Group 1", "Group 2"], b = rand(2));

julia> gd = groupby(df, :a);

julia> gd[Tuple("Group 1")]
ERROR: KeyError: key ('G', 'r', 'o', 'u', 'p', ' ', '1') not found

julia> gd[("Group 1",)]; # ugly

```

bkamins:
> instead of this PR we could do the following

Whatever we decide let us keep in this PR (just overwrite the old changes with new changes).

> Deprecate indexing with tuples entirely

This would be breaking and it does not do much harm I think, so I would leave the current indexing options as they are.

> `julia> gd[Tuple("Group 1")]`

the syntax is `gd[tuple("Group 1")]` if you want to avoid `gd[("Group 1",)]`

> Make more robust constructors for `GroupKey` which allows.

The downside of this (I know I initially proposed it :)), is that you would have to write e.g. `gdf[GroupKey(gdf, :a=>1, :b=>2)]` - it is not the end of the world but still a bit verbose.

Now as we talk about it I have realized that we could use `Dict` (and you would have a benefit of ignoring the order as you wanted), then one could write `gdf[Dict("a" => 1, "b" => 2)]`.

The crucial benefit is that `GroupKey` is tied to the `GroupedDataFrame` and `Dict` is not.
We could allow `Union{Dict{AbstractString}, Dict{Symbol}}`.

pdeffebach:
> The downside of this (I know I initially proposed it :)), is that you would have to write e.g. `gdf[GroupKey(gdf, :a=>1, :b=>2)]` - it is not the end of the world but still a bit verbose.

Oh, sorry! I missed that `GroupKey`s needed the grouped data frame. 

I also somehow missed how easy the `tuple` constructor was. I will clean up this PR and move forward with it.

bkamins:
OK - but then do you think adding `Dict` indexing is a good idea?

pdeffebach:
No, I think that's probably overkill. A `tuple` is good. 

A first-best might be a `Key` type with the behavior I described above. But if it's just the same as a `tuple` it's not worth the hassle.

bkamins:
It is not the same as `Tuple`, tuple allows only `(1, 2)` not `(:a => 1, :b => 2)`.

pdeffebach:
Closed in favor of #2281, which uses a `Dict`.

PR #2270, opened 2020-05-26
Fix position of backticks in disallowmissing docstring

nalimilan:


bkamins:
Let us merge this after CI passes (to make sure docs are built correctly).

PR #2273, opened 2020-05-31
fix bugs in append! and push! corner cases

bkamins:
1. `append!` incorrectly tried to access a non-existent column in source in some cases.
2. `push!` incorrectly did not clean up the data frame on error in some cases

PS. and this is a lesson for me not to do copy-paste :cry:.

After this is merged it should be backported to 0.21.2.

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1273,15 +1273,19 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
                     _columns(df1)[j] = newcol
                 end
             else
-                # we can safely do it - if we got here this means that this is allowed
-                if Missing <: eltype(df1[!, j]) || !promote
-                    append!(df1[!, j], df2[!, n])
-                else
+                if Missing <: eltype(df1[!, j])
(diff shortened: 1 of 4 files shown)


src/dataframe/dataframe.jl
```diff
@@ -1273,15 +1273,19 @@ function Base.append!(df1::DataFrame, df2::AbstractDataFrame; cols::Symbol=:sete
...
+                elseif promote
                     newcol = similar(df1[!, j], Union{Missing, eltype(df1[!, j])},
                                      targetrows)
                     copyto!(newcol, 1, df1[!, j], 1, nrows)
                     newcol[nrows+1:targetrows] .= missing
                     _columns(df1)[j] = newcol
+                else
+                    throw(ArgumentError("promote=false and source data frame does " *
+                                        "not contain column :$n, while destination" *
+                                        " column does not allowmissing values"))
```
nalimilan, line 1288:
```suggestion
                                        "not contain column :$n, while destination " *
                                        "column does not allow for missing values"))
```

bkamins:
Thank you. I thought I check all cases, but I missed `cols=:union` and `promote=false` (which probably will be a rare case, but still).

PR #2274, opened 2020-05-31
[BREAKING] Keep naming ComposedFunction :function

bkamins:
Fixes change introduced by https://github.com/JuliaLang/julia/pull/35980.

I assume we prefer `function` rather than `ComposedFunction` to be used in auto-generated column names. Also this is non-breaking this way.

@nalimilan - I have checked and I think that https://github.com/JuliaLang/julia/pull/35980 does not break aggregation code for composed functions (we still should use a fast branch then), but maybe you will want to take a second look.

### DIFF
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -76,6 +76,6 @@ function funname(f)
     String(n)[1] == '#' ? :function : n
 end
 
-if isdefined(Base, :ComposedFunction)
+if VERSION >= v"1.6.0-DEV.85"
     funname(c::Base.ComposedFunction) = Symbol(funname(c.f), :_, funname(c.g))
 end
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -75,3 +75,7 @@ function funname(f)
     n = nameof(f)
     String(n)[1] == '#' ? :function : n
 end
+
+if isdefined(Base, :ComposedFunction)
+    funname(::Base.ComposedFunction) = :function
+end
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -77,5 +77,5 @@ function funname(f)
 end
 
 if isdefined(Base, :ComposedFunction)
-    funname(::Base.ComposedFunction) = :function
+    funname(c::Base.ComposedFunction) = Symbol(funname(c.f), :_, funname(c.g))
 end
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -76,7 +76,7 @@ function funname(f)
     String(n)[1] == '#' ? :function : n
 end
 
-if isdefined(Base, :ComposedFunction)
+if isdefined(Base, :ComposedFunction) # Julia >= 1.6.0-DEV.85
     using Base: ComposedFunction
 else
     const ComposedFunction = let h = identity ∘ convert
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -76,6 +76,15 @@ function funname(f)
     String(n)[1] == '#' ? :function : n
 end
 
-if VERSION >= v"1.6.0-DEV.85"
-    funname(c::Base.ComposedFunction) = Symbol(funname(c.f), :_, funname(c.g))
+if isdefined(Base, :ComposedFunction)
+    using Base: ComposedFunction
+else
+    const ComposedFunction = let h = identity ∘ convert
+        @assert h.f === identity
+        @assert h.g === convert
+        getfield(parentmodule(typeof(h)), nameof(typeof(h)))
+    end
+    @assert identity ∘ convert isa ComposedFunction
 end
+
+funname(c::ComposedFunction) = Symbol(funname(c.f), :_, funname(c.g))
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -2223,44 +2223,26 @@ end
     gdf = groupby_checked(df, :g)
     @test combine(gdf, :x => sum) == DataFrame(g=1:2, x_sum=[3.0, 4.5])
 
-    if VERSION >= v"1.6.0-DEV.85"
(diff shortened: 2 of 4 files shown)


nalimilan:
Thanks. Actually this is an opportunity to improve the printing of composed functions:
```julia
julia> repr(sum ∘ skipmissing)
"sum ∘ skipmissing"
```
Do you agree we'd better take advantage of this by adding a special method? One issue is that `repr` uses the full name, which isn't what we do know nor probably what we want (for conciseness):
```julia
julia> repr(Base.Libc.realloc ∘ sum)
"Base.Libc.realloc ∘ sum"
```
So I guess we have to reconstruct the composed name by hand.

bkamins:
I was considering it but the problem is that things like: `"sum ∘ skipmissing"` are long, contain spaces, and many novices will not know how to key-in `∘`. That is why I opted for printing what we currently do (i.e. `:function`). However, if we can find some concise way to print it I would be OK with it. A first idea would be to print e.g. `sum_skipmissing` - i.e. use `_` and do not use `∘` at all. What do you think?

nalimilan:
Ah, right, maybe using `_` is more practical! We could also keep only the first function, and indicate that the other(s) is abbreviated in some way. The most common use case is `f ∘ skipmissing` and there we don't really care about `skipmissing`.

bkamins:
so you would generate eg `:sum_` for `sum∘skipmissing`?

nalimilan:
I don't really know TBH. It's true that the trailing `_` isn't very evocative. Maybe `sum_skipmissing` isn't so bad.

bkamins:
OK - I will change it this way. Note that this will also affect multiply nested compositions (but this is rare, so it is probably not a problem).

src/other/utils.jl
```diff
@@ -75,3 +75,7 @@ function funname(f)
     n = nameof(f)
     String(n)[1] == '#' ? :function : n
 end
+
+if isdefined(Base, :ComposedFunction)
```
nalimilan, line 79: Maybe make this a `VERSION` check? That will be easier to clean when we drop support for older versions.

Also add a test?

bkamins: OK

bkamins: do you remember how to get the exact version number?

nalimilan: With contrib/commit_name.sh. This one is `1.6.0-DEV.85`.

src/other/utils.jl
```diff
@@ -75,3 +75,7 @@ function funname(f)
     n = nameof(f)
     String(n)[1] == '#' ? :function : n
 end
+
+if VERSION >= v"1.6.0-DEV.85"
+    funname(c::Base.ComposedFunction) = Symbol(funname(c.f), :_, funname(c.g))
+end
```
tkf, line 81: I'm not sure if you like a tricky code, but I think you can do this

```suggestion
if isdefined(Base, :ComposedFunction)
    using Base: ComposedFunction
else
    const ComposedFunction = let h = identity ∘ convert
        @assert h.f === identity
        @assert h.g === convert
        getfield(parentmodule(typeof(h)), nameof(typeof(h)))
    end
    @assert identity ∘ convert isa ComposedFunction
end

funname(c::ComposedFunction) = Symbol(funname(c.f), :_, funname(c.g))
```

to make `funname(sum ∘ skipmissing)` to work in all Julia versions back to 1.0.

tkf:
I wonder if it's useful here if `julia` uses `AnonymousFunction <: Function` for anonymous functions?

src/other/utils.jl
bkamins: @nalimilan - what do you think? The benefit is that we would then consistently work on every version of Julia we support. The trick uses an undocumented functionality but it seems it is valid :).

bkamins:
I think it would be useful so that we would not need to use the hack with `#`.

src/other/utils.jl
nalimilan: Why not. The behavior is unlikely to change in existing releases.

bkamins:
OK to merge?
Also we should make a decision if we backport this (the change is breaking, but on the other hand without it DataFrames.jl behaves differently on different versions of Julia).

src/other/utils.jl
```diff
@@ -75,3 +75,16 @@ function funname(f)
     n = nameof(f)
     String(n)[1] == '#' ? :function : n
 end
+
+if isdefined(Base, :ComposedFunction)
```
nalimilan, line 79:
```suggestion
if isdefined(Base, :ComposedFunction) # Julia >= 1.6.0-DEV.85
```

test/grouping.jl
```diff
@@ -2308,12 +2309,12 @@ end
           combine(gdf, :x => (x -> sum(x)) => :a, :x => (x -> prod(x)) => :b)
     df = DataFrame(g=[1, 1], x=[missing, "a"])
     gdf = groupby_checked(df, :g)
-    @test combine(gdf, :x => sum∘skipmissing => :a, :x => prod∘skipmissing => :b) ==
-          combine(gdf, :x => (x -> sum(skipmissing(x))) => :a, :x => (x -> prod(skipmissing(x))) => :b)
+    @test Matrix(combine(gdf, :x => sum∘skipmissing => :a, :x => prod∘skipmissing => :b)) ==
```
nalimilan, line 2312: Why is `Matrix` needed here? Output names are set explicitly.

bkamins: Ah - right. It was a leftover from earlier design. Fixed

bkamins:
OK - I have removed backport. But this means that maybe we should consider making a release 0.22 when Julia 1.5 is out.

bkamins:
Thank you!

Issue #1333, opened 2017-12-29
Handling of duplicate column names by join


bkamins:
Following the discussion in https://github.com/JuliaData/DataFrames.jl/pull/1308#discussion_r159054678 here is the satus what other packages do:
* Pandas has `lsuffix` and `rsuffix` arguments
* R base has `suffixes` argument
* `dplyr` has `suffix`

I guess we should add something similar. Any comments?

bkamins:
As a side note base R and Pandas may produce duplicates by adding a suffix, whereas dplyr is careful enough to avoid them.

ararslan:
How annoying would it be to simply throw an error for duplicate column names to force the user to handle them manually? We could also prefix duplicate column names with the name of the table they came from, e.g. joining `A` and `B` where both have a column `c` will result in a table with `A_c` and `B_c` or something like that.

bkamins:
Now we have `makeunique` that allows or disallows duplicates. I guess it is annoying to disallow it as every other framework has this functionality.

Here is a snippet that could be considered to fix the names by suffixing if `makeunique` is set to `true` (it seems to be a standard to use suffix not a prefix):
```
function fixnames(left, right; lsuffix="_left", rsuffix="_right")
    lsuffix == rsuffix && throw(ArgumentError("left and right suffix must be different"))
    common = Set(intersect(left, right))
    for (x, suffix) in [(left, lsuffix), (right, rsuffix)]
        for i in eachindex(x)
            if x[i] in common
                x[i] = Symbol(string(x[i], suffix))
            end
        end
    end
    pool = Set{Symbol}()
    for (x, suffix) in [(left, lsuffix), (right, rsuffix)]
        for (i, s) in enumerate(x)
            while s in pool
                s = Symbol(string(s, suffix))
            end
            x[i] = s
            push!(pool, s)
        end
    end
end
```

nalimilan:
@ararslan I think it's annoying to handle manually, because you have to find which columns are present in the two inputs, skipping columns you're joining on.

Regarding the API, why not use `suffix` as in dplyr, and make it a pair? That would be consistent with what we do with `on` (#1297).

bkamins:
CC @oxinabox - this is directly relevant to what you have asked. Do you have a preference for API?

bkamins:
Note that in some joins we allow more than 2 data frames.

bkamins:
When working on this the following case should be tested against as it is potentially confusing I think:
```
[...]

PR #2278, opened 2020-06-09
Added makeunique argument for Table -> DataFrame

bosonbaas:
This is a simple edit which adds the `makeunique` option for the `DataFrame` constructor from `Table` objects.
This PR applies the `makeunique` functionality introduced in #1333 to the constructor from `Table`s.

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -21,28 +21,29 @@ Tables.getcolumn(dfr::DataFrameRow, nm::Symbol) = dfr[nm]
 getvector(x::AbstractVector) = x
 getvector(x) = [x[i] for i = 1:length(x)]
 # note that copycols is ignored in this definition (Tables.CopiedColumns implies copies have already been made)
-fromcolumns(x::Tables.CopiedColumns, names; copycols::Bool=true) =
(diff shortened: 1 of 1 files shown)


src/other/tables.jl
```diff
@@ -21,28 +21,29 @@ Tables.getcolumn(dfr::DataFrameRow, nm::Symbol) = dfr[nm]
 getvector(x::AbstractVector) = x
 getvector(x) = [x[i] for i = 1:length(x)]
 # note that copycols is ignored in this definition (Tables.CopiedColumns implies copies have already been made)
-fromcolumns(x::Tables.CopiedColumns, names; copycols::Bool=true) =
+fromcolumns(x::Tables.CopiedColumns, names; makeunique::Bool=false, copycols::Bool=true) =
     DataFrame(AbstractVector[getvector(Tables.getcolumn(x, nm)) for nm in names],
```
bosonbaas, line 25: I have noted that this iteration through columns will break if names are not unique. This can be fixed by iterating by index instead.

bosonbaas:
I have now found that this PR will fail because `Table` objects are iterated through by name, so while column names will be unique, columns which originally had the same name will all resolve to the first column of that name.

Issue #1732, opened 2019-02-21
When `join(..., validate=(true,true))` fails, it should include a list of non-unique joinkeyrows in the error)


oxinabox:
If you turn on the `validate` kwarg when joining, it will throw an error if they columns being joined on are have non-unique elements.
Very handy for defensive programming.

It woud be even more handy if it would include a list of which nonunique elements they were.

We might have to be a little smart about this, because we don't want to flood the user by printing 10^6 rows into the REPL. 
But we have the technology.

cc @rofinn

rofinn:
At the very least we could probably include a `duplicates` function which returns the duplicate columns based on a key.

oxinabox:
> `duplicates` function which returns the duplicate columns based on a key.

Hmmm... wouldn't the existing  `nonunique(df::AbstractDataFrame, cols)` do that?

bkamins:
Currently `validate` returns the index of first offending row in each data frame. And then - as you have just noted, we can use `nonunique` to learn which rows are non-unique. I think what we have now is good enough (one function does one thing). What would be the benefit of combining `nonunique` in `join`?

rofinn:
> Hmmm... wouldn't the existing `nonunique(df::AbstractDataFrame, cols)` do that?

`nonunique` returns a bit array of the indices that have been duplicated. It doesn't return both the original and duplicated rows of the original dataframe.

> Currently `validate` returns the index of first offending row in each data frame. And then - as you have just noted, we can use `nonunique` to learn which rows are non-unique.

Both `validate` and `nonunique` return just the index of duplicated rows, but it's often valuable to see the original and duplicated rows together.

oxinabox:
> What would be the benefit of combining nonunique in join?


I would not think of this at all as that.

This is about giving the best error message we can.

bkamins:
> This is about giving the best error message we can.

The idea was to print the index of the first offending row so that the user can look it up.

> Both `validate` and `nonunique` return just the index of duplicated rows, but it's often valuable to see the original and duplicated rows together.

This could be probably added to `nonunique` with something like `verbose` keyword argument. @rofinn could you specify what exactly you would like to see?

rofinn:
I

rofinn:
Sorry, "Close and comment" is not "Cancel" :p

It might be easiest to explain in code. Our current `safe_join` function looks like this:
```
function safe_join(index, frame, join_key)
    try
[...]

PR #2279, opened 2020-06-09
add filter and filter! to GroupedDataFrame

bkamins:
@nalimilan - this is a follow up to your comment in #1732.

### DIFF
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -585,8 +585,41 @@ in which case a `NamedTuple` of these arguments is passed.
 and column duplicates are allowed if a vector of `Symbol`s, strings, or integers
 is passed.
 
+See also: [`filter!`](@ref)
+
 # Examples
 ```
+julia> df = DataFrame(g=1:2, x='a':'b');
+
+julia> gd = groupby(df, :g)
+GroupedDataFrame with 2 groups based on key: g
+First Group (1 row): g = 1
+│ Row │ g     │ x    │
+│     │ Int64 │ Char │
+├─────┼───────┼──────┤
+│ 1   │ 1     │ 'a'  │
+⋮
+Last Group (1 row): g = 2
+│ Row │ g     │ x    │
+│     │ Int64 │ Char │
+├─────┼───────┼──────┤
+│ 1   │ 2     │ 'b'  │
+
+julia> filter(x -> x.x[1] == 'a', gd)
+GroupedDataFrame with 1 group based on key: g
+First Group (1 row): g = 1
+│ Row │ g     │ x    │
+│     │ Int64 │ Char │
+├─────┼───────┼──────┤
+│ 1   │ 1     │ 'a'  │
+
+julia> filter(:x => x -> x[1] == 'a', gd)
+GroupedDataFrame with 1 group based on key: g
+First Group (1 row): g = 1
+│ Row │ g     │ x    │
+│     │ Int64 │ Char │
+├─────┼───────┼──────┤
+│ 1   │ 1     │ 'a'  │
 
 ```
 """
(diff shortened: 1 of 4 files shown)


src/groupeddataframe/groupeddataframe.jl
```diff
@@ -568,3 +568,177 @@ function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple}, default)
...
+    function mapper(i)
+        idxs = idx[starts[i]:ends[i]]
+        return map(x -> view(x, idxs), nt)
+    end
+
+    return gdf[[f(mapper(i))::Bool for i in 1:length(gdf)]]
+end
+
+"""
+    filter!(function, gdf::GroupedDataFrame)
```
nalimilan, line 673: Do we have any other in-place functions on `GroupedDataFrame` already, apart from `select!` and `transform!`? I'm asking because for consistency with the two latter, it would make sense for `filter!` to drop rows from `parent(gdf)`. Though that could be a unexpected. We need a general policy on this, which could also apply to in-place oeprations on `SubDataFrame`.

test/grouping.jl
```diff
@@ -2338,4 +2338,46 @@ end
...
+            gdf1  = groupby(df, gcols)
+            gdf2 = filter(predicate, gdf1)
+            if cutoff == 1
+                @test getindex.(keys(gdf2), 1) == 1:2
+            elseif cutoff == 0
+                @test gdf1 == gdf2
+            elseif cutoff == 10
+                @test isempty(gdf2)
+            end
+            filter!(predicate, gdf1)
```
nalimilan, line 2367:
```suggestion
            @test filter!(predicate, gdf1) === gdf1
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -568,3 +568,177 @@ function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple}, default)
         return default
     end
 end
+
+"""
+    filter(function, gdf::GroupedDataFrame)
```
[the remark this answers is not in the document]
bkamins: OK - changed to `fun`.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -568,3 +568,177 @@ function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple}, default)
...
+GroupedDataFrame with 1 group based on key: g
+First Group (1 row): g = 1
+│ Row │ g     │ x    │
+│     │ Int64 │ Char │
+├─────┼───────┼──────┤
+│ 1   │ 1     │ 'a'  │
+
+```
+"""
+Base.filter(f::Base.Callable, gdf::GroupedDataFrame) =
```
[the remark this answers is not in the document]
bkamins: no - removed.

src/groupeddataframe/groupeddataframe.jl
bkamins: This is a good question (actually I initially considered to add `filter` only for these reasons). It applies both to `filter` and `filter!`. I have not thought about it because what I proposed was most natural.

So the considerations are:
1. we do not have other functions that mutate `GroupedDataFrame` in-place
2. we do not have functions that mutate `SubDataFrame` schema in-place (of course we allow mutating the data, but not schema)
3. `select!` and `transform!` work only for some `GroupedDataFrames` (ones that have no groups dropped) and they guarantee to retain row count in the parent; also it is in general expected (and documented) that they mutate the parent as they define new columns in general
4. `filter!` and `filter` can work on an already "subsetted" `GroupDataFrame` so then it would be confusing what should happen with groups that are already not present.
5. it is natural to expect that `filter(predicate, collecion)` does the same as `collection[predicate.(collection)]`, which does not mutate the parent of `collection` if it is present.

All in all I think it is better to keep what we have.

bkamins:
I have also pushed updates to docstrings of `filter`/`filter!` for `AbstractDataFrame` to make them consistent.

bkamins:
Your comment made me realize that in the past we assumed `GroupedDataFrame` is immutable, and we used this assumption in transformation code. I have pushed a fix.

bkamins:
So the only downside of `filter!` is that it modifies a view. I have reviewed all the code and in https://github.com/JuliaData/DataFrames.jl/pull/2280 I make sure that we do not make assumptions that it is not modified, but still maybe it is better not to add `filter!`?

nalimilan:
Maybe let's leave it out until we have a clear use case for it? If nobody requests it maybe it's not needed, and that keeps the API more consistent regarding what `!` does.

bkamins:
OK - I have removed `filter!`. This PR is independent now of  #2280 as a consequence.

bkamins:
similar - only coverage fails

bkamins:
Thank you!

PR #2280, opened 2020-06-10
Fix various bugs in split/apply/combine in 0.21 release

bkamins:
In this PR I fix some corner cases that were bugs in 0.21 release. It should be merged before https://github.com/JuliaData/DataFrames.jl/pull/2279 as it should be backported.

What it covers:
* documentation improvements of what happens (and should happen in corner cases, especially when something is "empty", where something is: `GroupedDataFrame`, `DataFrame`, list of transformations, or `groupcols`)
* fix a bug with `GroupDataFrame` indexing that caused StackOverflow
* fix a bug in creation of `GroupedDataFrame` with `ungroup=false` (wrong columns could be shown as grouping columns in the past)
* fix cases of "empty" stuff passed (listed above), especially when `keepkeys=false` or `ungroup=false` (there are several corner cases here); most likely in real usage scenarios these cases do not happen, but still we should provide a correct result here.

In general correctly handling all these corner cases is tricky so probably if someone would be willing to review it it might be challenging (so thank you in advance for the efforts). I will also try to improve test coverage for this if I find other "corner cases" still not covered.

### DIFF
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -278,7 +278,8 @@ SELECT_ARG_RULES =
     select!(df::DataFrame, args...)
 
 Mutate `df` in place to retain only columns specified by `args...` and return it.
-The result is guaranteed to have the same number of rows as `df`.
+The result is guaranteed to have the same number of rows as `df`, except when no
+columns are selected.
 
 $SELECT_ARG_RULES
 
@@ -373,7 +374,8 @@ transform!(df::DataFrame, args...) = select!(df, :, args...)
     select(df::AbstractDataFrame, args...; copycols::Bool=true)
 
 Create a new data frame that contains columns from `df` specified by `args` and
-return it. The result is guaranteed to have the same number of rows as `df`.
+return it. The result is guaranteed to have the same number of rows as `df`,
+except when no columns are selected.
 
 If `df` is a `DataFrame` or `copycols=true` then column renaming and transformations
 are supported.
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -279,7 +279,7 @@ SELECT_ARG_RULES =
 
 Mutate `df` in place to retain only columns specified by `args...` and return it.
 The result is guaranteed to have the same number of rows as `df`, except when no
-columns are selected.
+columns are selected (in which case the result has zero rows).
 
 $SELECT_ARG_RULES
 
@@ -375,7 +375,7 @@ transform!(df::DataFrame, args...) = select!(df, :, args...)
 
 Create a new data frame that contains columns from `df` specified by `args` and
 return it. The result is guaranteed to have the same number of rows as `df`,
-except when no columns are selected.
+except when no columns are selected (in which case the result has zero rows)..
 
 If `df` is a `DataFrame` or `copycols=true` then column renaming and transformations
 are supported.
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -1,3 +1,17 @@
+# Implementation note
+# There are two important design  features of GroupedDataFrame
+# 1. idx, starts, ends and keymap are by default left uninitiated;
+#    they get populated only on demand; this means that every GroupedDataFrame
+#    has lazy_lock field which is used to make sure that two threads concurrently
+#    do not try to create them. The lock should be used in every function hat
+#    does a direct access to these fields via getfield.
+# 2. Except for point 1 above currently fields of GroupedDataFrame are never
+#    mutated after it is created. This means that internally when copying
+#    a GroupedDataFrame they are not copied for efficiency. If in the future
+#    operations that mutate GroupedDataFrame are introduced all non-copying
+#    passing of the internal fields to a new GroupedDataFrame should be
+#    updated. Currently this applies to `getindex` and `combine_helper` functions
+
 """
     GroupedDataFrame
 
@@ -202,15 +216,10 @@ end
 
 # Index with colon (creates copy)
 function Base.getindex(gd::GroupedDataFrame, idxs::Colon)
-    maybe_copy(x) = isnothing(x) ? x : copy(x)
     Threads.lock(gd.lazy_lock)
-    new_gd = GroupedDataFrame(gd.parent, copy(gd.cols), copy(gd.groups),
-                              maybe_copy(getfield(gd, :idx)),
-                              maybe_copy(getfield(gd, :starts)),
-                              maybe_copy(getfield(gd, :ends)),
-                              gd.ngroups,
-                              maybe_copy(getfield(gd, :keymap)),
-                              Threads.ReentrantLock())
+    new_gd = GroupedDataFrame(gd.parent, gd.cols, gd.groups, getfield(gd, :idx),
+                              getfield(gd, :starts), getfield(gd, :ends), gd.ngroups,
+                              getfield(gd, :keymap), Threads.ReentrantLock())
     Threads.unlock(gd.lazy_lock)
     return new_gd
 end
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -299,8 +299,10 @@ end
 # The full version (to_indices) is required rather than to_index even though
 # GroupedDataFrame behaves as a 1D array due to the behavior of Colon and Not.
 # Note that this behavior would be the default if it was <:AbstractArray
-Base.getindex(gd::GroupedDataFrame, idx...) =
-    getindex(gd, Base.to_indices(gd, idx)...)
+function Base.getindex(gd::GroupedDataFrame, idx...)
+    length(idx) == 1 || throw(ArgumentError("GroupedDataFrame requires a single index"))
+    return getindex(gd, Base.to_indices(gd, idx)...)
+end
 
 # The allowed key types for dictionary-like indexing
 const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple}
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -48,8 +48,8 @@ function Base.getproperty(gd::GroupedDataFrame, f::Symbol)
         if getfield(gd, f) === nothing
             gd.keymap = genkeymap(gd, ntuple(i -> parent(gd)[!, gd.cols[i]], length(gd.cols)))
         end
-        return getfield(gd, f)::Dict{Any,Int}
         Threads.unlock(gd.lazy_lock)
+        return getfield(gd, f)::Dict{Any,Int}
     else
         return getfield(gd, f)
     end
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -1,9 +1,9 @@
 # Implementation note
-# There are two important design  features of GroupedDataFrame
-# 1. idx, starts, ends and keymap are by default left uninitiated;
+# There are two important design features of GroupedDataFrame
+# 1. idx, starts, ends and keymap are by default left uninitialized;
 #    they get populated only on demand; this means that every GroupedDataFrame
 #    has lazy_lock field which is used to make sure that two threads concurrently
-#    do not try to create them. The lock should be used in every function hat
+#    do not try to create them. The lock should be used in every function that
 #    does a direct access to these fields via getfield.
 # 2. Except for point 1 above currently fields of GroupedDataFrame are never
 #    mutated after it is created. This means that internally when copying
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -15,6 +15,8 @@ mutable struct GroupedDataFrame{T<:AbstractDataFrame}
     ends::Union{Vector{Int},Nothing}     # ends of groups after permutation by idx
     ngroups::Int                         # number of groups
     keymap::Union{Dict{Any,Int},Nothing} # mapping of key tuples to group indices
+    lazy_lock::Threads.ReentrantLock     # lock is needed to make lazy operations
+                                         # thread safe
 end
 
 function genkeymap(gd, cols)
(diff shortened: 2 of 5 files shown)


src/abstractdataframe/selection.jl
```diff
@@ -278,7 +278,8 @@ SELECT_ARG_RULES =
     select!(df::DataFrame, args...)
 
 Mutate `df` in place to retain only columns specified by `args...` and return it.
-The result is guaranteed to have the same number of rows as `df`.
+The result is guaranteed to have the same number of rows as `df`, except when no
+columns are selected.
```
nalimilan, line 282:
```suggestion
columns are selected (in which case the result has zero rows).
```

src/abstractdataframe/selection.jl
```diff
@@ -373,7 +374,8 @@ transform!(df::DataFrame, args...) = select!(df, :, args...)
     select(df::AbstractDataFrame, args...; copycols::Bool=true)
 
 Create a new data frame that contains columns from `df` specified by `args` and
-return it. The result is guaranteed to have the same number of rows as `df`.
+return it. The result is guaranteed to have the same number of rows as `df`,
+except when no columns are selected.
```
nalimilan, line 378:
```suggestion
except when no columns are selected (in which case the result has zero rows)..
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -245,9 +249,13 @@ const KWARG_PROCESSING_RULES =
     in addition to those generated. In this case if the returned
     value contains columns with the same names as the grouping columns, they are
     required to be equal.
+    If `keepkeys=false` if generated columns have the same name as grouping columns
```
nalimilan, line 252:
```suggestion
    If `keepkeys=false` and some generated columns have the same name as grouping columns,
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -237,6 +239,8 @@ const F_ARGUMENT_RULES =
     If the first or last argument is a function `fun`, it is passed a [`SubDataFrame`](@ref)
     view for each group and can return any return value defined below.
     Note that this form is slower than `pair` or `args` due to type instability.
+
+    If grouped data frame has zero groups then no transformations are applied.
```
nalimilan, line 243:
```suggestion
    If `gd` has zero groups then no transformations are applied.
```

So this is an interesting corner case indeed. Maybe we'd better throw an error in future releases (not in this PR since it should be backported)? Code that works with a nonzero number of groups is likely to try to access new columns and therefore fail later. Or could we call the function on a fake group with empty vectors just to find out the types of expected columns, and create them? Throwing an error would allow implementing this later.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -245,9 +249,13 @@ const KWARG_PROCESSING_RULES =
     in addition to those generated. In this case if the returned
     value contains columns with the same names as the grouping columns, they are
     required to be equal.
+    If `keepkeys=false` if generated columns have the same name as grouping columns
+    they are kept and are not required to be equal to grouping columns.
 
     If `ungroup=true` (the default) a `DataFrame` is returned.
     If `ungroup=false` a `GroupedDataFrame` grouped using `keycols(gdf)` is returned.
+
+    If `gd` has zero groups then no additional columns are computed.
```
nalimilan, line 258:
```suggestion
    If `gd` has zero groups then no transformations are applied.
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1483,10 +1515,13 @@ end
            copycols::Bool=true, keepkeys::Bool=true, ungroup::Bool=true)
 
 Apply `args` to `gd` following the rules described in [`combine`](@ref) and return the
-result as a `DataFrame` if `ungroup=true` or `GroupedDataFrame` if `ungroup=false`.
+result as a `DataFrame` if `ungroup=true` or `GroupedDataFrame` if `ungroup=false`
+(in this case the returned value retains the order of groups of `gd`).
```
nalimilan, line 1519: I'd put this detail below: the first paragraph should just explain what's the purpose of the function.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1483,10 +1515,13 @@ end
...
 
-The `parent` of the returned value has as many rows as `parent(gd)`. If an operation
-in `args` returns a single value it is always broadcasted to have this number of rows.
+The `parent` of the returned value has as many rows as `parent(gd)`, except when
+`gd` has no grouping columns or `keepkeys=false` and no columns are selected.
+Also order of rows in the parent is preserved. If an operation in `args` returns
+a single value it is always broadcasted to have this number of rows.
```
nalimilan, line 1524:
```suggestion
The `parent` of the returned value has as many rows as `parent(gd)` and
in the same order, except when the returned value has no columns
(in which case it has zero rows). If an operation in `args` returns
a single value it is always broadcasted to have this number of rows.
```

test/grouping.jl
```diff
@@ -1320,13 +1320,17 @@ end
...
+    @test size(combine(gd)) == (0, 1)
+    @test names(combine(gd)) == ["a"]
 end
 
 @testset "GroupedDataFrame dictionary interface" begin
     df = DataFrame(a = repeat([:A, :B, missing], outer=4), b = repeat(1:2, inner=6), c = 1:12)
     gd = groupby_checked(df, [:a, :b])
 
+    @test gd[1] == DataFrame(a=[:A, :A], b=[1, 1], c=[1, 4])
+    @test_throws ArgumentError gd[1, 1]
```
nalimilan, line 1332: These are not really the dictionary interface, right? Don't we have another set of tests for integer indexing?

test/grouping.jl
```diff
@@ -2218,6 +2222,144 @@ end
...
+    @test df2′ == DataFrame(g=[3, 1, 1, 2, 3])
+end
+
+@testset "handling empty data frame / selectors / groupcols" begin
+    df = DataFrame(x=[], g=[])
+    gdf = groupby_checked(df, :g)
+
+    @test size(combine(gdf)) == (0, 1)
+    @test names(combine(gdf)) == ["g"]
+    # TODO: add tests for keepkeys and ungroup after deprecation
```
nalimilan, line 2257: Maybe write them know in a comment, as nobody will remember what this comment meant exactly. :-D

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -597,33 +606,55 @@ function combine_helper(f, gd::GroupedDataFrame,
...
-                                    Int[], Int[], Int[], 0, Dict{Any,Int}())
+                                    Int[], Int[], Int[], 0, Dict{Any,Int}(),
+                                    Threads.ReentrantLock())
         end
         if keeprows
+            @assert length(keys) > 0 || idx == gd.idx
+            @assert names(newparent, 1:length(gd.cols)) == names(parent(gd), gd.cols)
             # in this case we are sure that the result GroupedDataFrame has the
-            # same structure as the source
-            # we do not copy data as it should be safe - we never mutate fields of gd
```
nalimilan, line 612: When do we mutate fields now?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -597,33 +606,55 @@ function combine_helper(f, gd::GroupedDataFrame,
...
+                                    Threads.ReentrantLock())
         end
     else
-        if ungroup
-            return keepkeys ? parent(gd)[1:0, gd.cols] : DataFrame()
+        if keeprows
+            if nrow(parent(gd)) > 0
+                 throw(ArgumentError("select and transform do not support " *
+                                    "`GroupedDataFrame`s from which some groups have "*
+                                    "been dropped (including skipmissing=true)"))
```
nalimilan, line 645:
```suggestion
                                     "`GroupedDataFrame`s from which some groups have "*
                                     "been dropped (including skipmissing=true)"))
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -597,33 +606,55 @@ function combine_helper(f, gd::GroupedDataFrame,
...
 
-        if length(idx) == 0
+        if length(idx) == 0 && !(keeprows && length(keys) > 0)
             @assert nrow(newparent) == 0
             return GroupedDataFrame(newparent, collect(1:length(gd.cols)), Int[],
-                                    Int[], Int[], Int[], 0, Dict{Any,Int}())
+                                    Int[], Int[], Int[], 0, Dict{Any,Int}(),
+                                    Threads.ReentrantLock())
         end
         if keeprows
```
nalimilan, line 620:
```suggestion
        elseif keeprows
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -597,33 +606,55 @@ function combine_helper(f, gd::GroupedDataFrame,
...
+        if keeprows
+            if nrow(parent(gd)) > 0
+                 throw(ArgumentError("select and transform do not support " *
+                                    "`GroupedDataFrame`s from which some groups have "*
+                                    "been dropped (including skipmissing=true)"))
+            end
+            if ungroup
+                return keepkeys ? select(parent(gd), gd.cols, copycols=copycols) : DataFrame()
+            else
+                return groupby(select(parent(gd), gd.cols, copycols=copycols), 1:length(gd.cols))
```
nalimilan, line 650: Why not build the `GroupedDataFrame` directly? Same below.

src/groupeddataframe/splitapplycombine.jl
bkamins: we do not mutate fields now.
It is a preparation for `filter!`. But I prefer to make a copy here as it is not very expensive, and be safe that in the future changing if someone introduces mutation to `GroupedDataFrame` it does not get forgotten (as it is easy to forget that this line in such a large codebase assumes that it is not mutated).

src/groupeddataframe/splitapplycombine.jl
bkamins: OK - I will change it.

src/groupeddataframe/splitapplycombine.jl
bkamins: OK

test/grouping.jl
bkamins: This is testing for a bug introduced by dictionary interface addition.
Also essentially the error that user is likely to do is write `gd[1,1]` instead of `gd[(1,1)]`. But I will have a look if there is some better place for it.

test/grouping.jl
bkamins: OK

bkamins:
> Code that works with a nonzero number of groups is likely to try to access new columns and therefore fail later.

This is what I was afraid of exactly. I will open a separate issue or PR to track this.

src/groupeddataframe/splitapplycombine.jl
nalimilan: I'd rather avoid making copies preventively as long as we never mutate fields. Especially if we drop `filter!` for now.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1513,8 +1521,10 @@ end
     select(gd::GroupedDataFrame, args...;
            copycols::Bool=true, keepkeys::Bool=true, ungroup::Bool=true)
 
-Apply `args` to `gd` following the rules described in [`combine`](@ref) and return the
-result as a `DataFrame` if `ungroup=true` or `GroupedDataFrame` if `ungroup=false`
+Apply `args` to `gd` following the rules described in [`combine`](@ref).
+
+If `ungroup=true` the result is a `DataFrame`.
+If  `ungroup=false` thee result is a `GroupedDataFrame`
```
nalimilan, line 1527:
```suggestion
If  `ungroup=false` the result is a `GroupedDataFrame`
```

src/groupeddataframe/splitapplycombine.jl
bkamins: OK - changed. I have made an extensive implementation note, as `GroupedDataFrame` is currently quite tricky to implement new functionality for.

bkamins:
coverage only fails.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -1,3 +1,17 @@
+# Implementation note
+# There are two important design  features of GroupedDataFrame
```
nalimilan, line 2:
```suggestion
# There are two important design features of GroupedDataFrame
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -1,3 +1,17 @@
...
+# 1. idx, starts, ends and keymap are by default left uninitiated;
```
nalimilan, line 3:
```suggestion
# 1. idx, starts, ends and keymap are by default left uninitialized;
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -1,3 +1,17 @@
...
+#    they get populated only on demand; this means that every GroupedDataFrame
+#    has lazy_lock field which is used to make sure that two threads concurrently
+#    do not try to create them. The lock should be used in every function hat
```
nalimilan, line 6:
```suggestion
#    do not try to create them. The lock should be used in every function that
```

bkamins:
Thank you!

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -34,15 +50,19 @@ end
 function Base.getproperty(gd::GroupedDataFrame, f::Symbol)
     if f in (:idx, :starts, :ends)
         # Group indices are computed lazily the first time they are accessed
+        Threads.lock(gd.lazy_lock)
```
nalimilan, line 53: Thinking about this again, shouldn't this line be moved inside `if getfield(gd, f) === nothing`? Ideally threads wouldn't have to lock each other when accessing indices *once they have been computed*. To avoid threads from computing the fields multiple times (which would be wasteful, though probably not problematic), we could check `getfield(gd, f) === nothing` again after calling `lock`.

bkamins: Yes - I was also thinking about it. The solution is exactly as you say:

> we could check `getfield(gd, f) === nothing` again after calling lock.

I just wanted to have a simpler implementation as `lock`/`unlock` is cheap:
```
julia> function f(l)
       lock(l)
       unlock(l)
       end

julia> l = Threads.ReentrantLock()
ReentrantLock(nothing, Base.GenericCondition{Base.Threads.SpinLock}(Base.InvasiveLinkedList{Task}(nothing, nothing), Base.Threads.SpinLock(0)), 0)

julia> f(l)

julia> @benchmark f($l)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     26.066 ns (0.00% GC)
  median time:      26.087 ns (0.00% GC)
  mean time:        27.893 ns (0.00% GC)
  maximum time:     77.285 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     993

```
and with threads often things are more tricky than they seem.

I will open a PR with the change you propose and we can discuss there.

PR #2281, opened 2020-06-11
Index to grouped data frame using `Dict`s

pdeffebach:
Following discussion in #2265, this is the better strategy. 

This PR so far has the implementation and docs, but no tests.


src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,20 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function _dict_to_tuple(key::Dict{T, S}, gd) where {T, S}
+    t = ntuple(length(gd.cols)) do i 
+        key[T(_names(gd)[i])]
```
bkamins, line 328: We should access columns of appropriate index (also make sure this is tested when e.g. columns 5,3, and 7 are grouping columns in the order I have given)
```suggestion
        key[T(_names(gd)[gd.cols[i]])]
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,20 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
+    end
+end
+
+function Base.to_index(gd::GroupedDataFrame, key::Dict{T, S}) where {T<:Union{Symbol, <:AbstractString}, S}
```
bkamins, line 332: We should not allow mixing `Symbol`s and strings I think:
```suggestion
function Base.to_index(gd::GroupedDataFrame, key::Union{Dict{Symbol},Dict{<:AbstractString}})
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,20 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function _dict_to_tuple(key::Dict{T, S}, gd) where {T, S}
```
bkamins, line 326: As below I think it is better to be stricter just to be sure all works OK.
```suggestion
function _dict_to_tuple(key::Union{Dict{Symbol},Dict{<:AbstractString}} , gd)
    T = keytype(d)
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,20 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
+    t = ntuple(length(gd.cols)) do i 
+        key[T(_names(gd)[i])]
+    end
+end
+
+function Base.to_index(gd::GroupedDataFrame, key::Dict{T, S}) where {T<:Union{Symbol, <:AbstractString}, S}
+    if length(key) != length(gd.cols) || any(Symbol(n) ∉ _names(gd) for n in keys(key))
```
bkamins, line 333: The second test was incorrect as commented above
```suggestion
    if length(key) != length(gd.cols)
```
Also I think we do not need it as if the key is not found we will get an error trying to retrieve it in `_dict_to_tuple`. Finally I would move the `length`-check to `_dict_to_tuple` to avoid code duplication in `haskey`.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -508,6 +522,14 @@ function Base.haskey(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return haskey(gd, Tuple(key))
 end
 
+function Base.haskey(gd::GroupedDataFrame, key::Dict{T, S}) where {T<:Union{Symbol, <:AbstractString}, S}
+    if length(key) != length(gd.cols) || any(Symbol(n) ∉ _names(gd) for n in keys(key))
```
bkamins, line 526: As commented above - if we move this code to `_dict_to_tuple` then we will not have code duplication.

bkamins:
Thank you for the fix. The design looks good to me. I have left some comments. Also `function Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T}` needs to be updated to allow `Dict`.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,21 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
 
+function _dict_to_tuple(key::Union{Dict{Symbol},Dict{<:AbstractString}} , gd)
+    T = keytype(d)
+    t = ntuple(length(gd.cols)) do i 
+        key[T(_names(gd)[gd.cols[i]])]
+    end
+end
+
+function Base.to_index(gd::GroupedDataFrame, key::Union{Dict{Symbol},Dict{<:AbstractString}})
+    if length(key) != length(gd.cols)
```
bkamins, line 334: this check should go to `_dict_to_tuple`.

src/groupeddataframe/groupeddataframe.jl
bkamins: I am unresolving this because this check is incorrect. Also if you move the `if` in `to_index` to `_dict_to_tuple` it will not be needed at all.

bkamins:
As I have commented earlier this https://github.com/JuliaData/DataFrames.jl/pull/2281/files#diff-6349421a054bb7e74b7f27ae86304cbaR342 also should be fixed.

pdeffebach:
Sorry about that. Was fixing them but got distracted. 

Should be ready for another review. I added a `Dict` method where the keys are `String`s by default. Then I added a bunch of tests.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,21 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function _dict_to_tuple(key::Union{Dict{Symbol},Dict{<:AbstractString}} , gd)
+    if length(key) != length(gd.cols)
+        throw(KeyError(key))
+    end
+
+    T = keytype(key)
```
bkamins, line 331: This is needed as for strings we need to have a type that is able to perform the conversion described below
```suggestion
    T = keytype(key) <: Symbol ? Symbol : String
```
This assumes that a `Dict` with any string keys should be able to handle getting `String` input (@nalimilan - do you think it is safe to assume; maybe some strange custom `AbstractStrings` with non UTF-8 encoding would fail here?).

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -508,6 +525,14 @@ function Base.haskey(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return haskey(gd, Tuple(key))
 end
 
+function Base.haskey(gd::GroupedDataFrame, key::Union{Dict{Symbol},Dict{<:AbstractString}})
+    if any(Symbol(n) ∉ _names(gd) for n in keys(key))
```
bkamins, line 529: this is now not needed, as it is handled in `_dict_to_tuple`

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -517,7 +542,8 @@ Base.haskey(gd::GroupedDataFrame, key::Union{Signed,Unsigned}) =
 Get a group based on the values of the grouping columns.
 
 `key` may be a `NamedTuple` or `Tuple` of grouping column values (in the same
-order as the `cols` argument to `groupby`).
+order as the `cols` argument to `groupby`). It may also be a `Dict`, in which case the
```
bkamins, line 545: maybe add what types for keys of `Dict` are allowed?

test/grouping.jl
```diff
@@ -1339,6 +1339,13 @@ end
         @test gd[NamedTuple(key)] ≅ gd[i]
         # Plain tuple
         @test gd[Tuple(key)] ≅ gd[i]
+        # Dict, with `String` as default
```
bkamins, line 1342: Please add some tests for `Dict` where `keytype` is `AbstractString` and `SubString` (to check my fix above).

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -345,6 +360,8 @@ function Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T}
         Tuple
     elseif E1 <: NamedTuple
         NamedTuple
+    elseif E1 <: Dict
+        Dict
```
bkamins, line 364: This means we allow mixing `Dict`s with different signatures, but maybe it is OK? What do you think.
If we allow it please add test that something like: `[Dict(:a=>1), Dict("a"=>2)]` (appropriately adjusted) passes.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -303,7 +303,7 @@ Base.getindex(gd::GroupedDataFrame, idx...) =
     getindex(gd, Base.to_indices(gd, idx)...)
 
 # The allowed key types for dictionary-like indexing
-const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple}
+const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple, Dict}
```
bkamins, line 306: Defining it in this way should give a bit better error messages
```suggestion
const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple, Dict{Symbol}, Dict{<:AbstractString}}
```

src/groupeddataframe/groupeddataframe.jl
bkamins: But even if we keep it (i.e. allowing mixing) I would restrict it to `Union{Dict{Symbol}, Dict{<:AbstractString}}`. This will not change how the code works, but will produce a nicer error message `"Invalid index ...`. Now for invalid `Dict` you will get a `MethodError` when trying to call `to_index` and probably this will confuse users.

pdeffebach: I think it's better to be strict for now. I have added that code that will throw an error for different key types and added a test.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -561,7 +582,7 @@ julia> get(gd, (:baz,), nothing)
 julia> get(gd, (:qux,), nothing)
 ```
 """
-function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple}, default)
+function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple, Dict}, default)
```
bkamins, line 585:
```suggestion
function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple, Dict{Symbol}, Dict{<:AbstractString}}, default)
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -345,6 +360,8 @@ function Base.to_index(gd::GroupedDataFrame, idxs::AbstractVector{T}) where {T}
         Tuple
     elseif E1 <: NamedTuple
         NamedTuple
+    elseif E1 <: Dict
+        Dict{keytype(idx1)}
```
bkamins, line 364:
```suggestion
    elseif E1 <: Dict{Symbol}
        Dict{Symbol}
    elseif E1 <: Dict{<:AbstractString}
        Dict{<:AbstractString}
```

test/grouping.jl
```diff
@@ -1351,17 +1358,23 @@ end
...
     # Out of order
     @test_throws KeyError gd[(b=1, a=:A)]
     @test_throws KeyError gd[(1, :A)]
     # Empty
     @test_throws KeyError gd[()]
     @test_throws KeyError gd[NamedTuple()]
+
+    # Substring and abstractstring
+    @test gd[Dict(SubString("a",1,1) => :A, SubString("b",1,1) => 1)] ≅  gd[1]
+    @test gd[Dict(Test.GenericString("a") => :A, Test.GenericString("b") => 1)] ≅  gd[1]
```
bkamins, line 1377:
```suggestion
    @test gd[Dict(Test.GenericString("a") => :A, "b" => 1)] ≅  gd[1]
```

bkamins:
I have pushed some small things. If CI passes can you please have a final look at it and confirm all is clear and works as intended? Then I think we can approve it (but I will wait with merging for @nalimilan as he has not commented on this API change).

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -267,7 +267,7 @@ function Base.NamedTuple(key::GroupKey)
     N(_groupvalues(parent(key), getfield(key, :idx)))
 end
 Base.Tuple(key::GroupKey) = values(key)
-
+Base.Dict(key::GroupKey) = Dict(String(first(ki)) => last(ki) for ki in pairs(key))
```
nalimilan, line 270: This raises again the question of whether we should use symbols or strings by default... Another place where we had to choose is `eachcol`/`DataFrameColumns`. We should have a general policy about this.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,21 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function _dict_to_tuple(key::Union{Dict{Symbol},Dict{<:AbstractString}} , gd)
```
nalimilan, line 326:
```suggestion
function _dict_to_tuple(key::Union{Dict{Symbol}, Dict{<:AbstractString}}, gd)
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -303,7 +303,7 @@ Base.getindex(gd::GroupedDataFrame, idx...) =
     getindex(gd, Base.to_indices(gd, idx)...)
 
 # The allowed key types for dictionary-like indexing
-const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple}
+const GroupKeyTypes = Union{GroupKey, Tuple, NamedTuple, Dict{Symbol}, Dict{<:AbstractString}}
```
nalimilan, line 306: Maybe this should be `AbstractDict`? Same elsewhere.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -517,7 +539,8 @@ Base.haskey(gd::GroupedDataFrame, key::Union{Signed,Unsigned}) =
 Get a group based on the values of the grouping columns.
 
 `key` may be a `NamedTuple` or `Tuple` of grouping column values (in the same
-order as the `cols` argument to `groupby`).
+order as the `cols` argument to `groupby`). It may also be a `Dict`, with keys that are 
+either `Symbol`s or strings, in which case the order of the arguments does not matter.
```
nalimilan, line 543:
```suggestion
order as the `cols` argument to `groupby`). It may also be a `Dict` with
`Symbol` or string keys, in which case the order of the arguments does not matter. 
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -561,7 +584,7 @@ julia> get(gd, (:baz,), nothing)
 julia> get(gd, (:qux,), nothing)
 ```
 """
-function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple}, default)
+function Base.get(gd::GroupedDataFrame, key::Union{Tuple, NamedTuple, Dict{Symbol}, Dict{<:AbstractString}}, default)
```
nalimilan, line 587:
```suggestion
function Base.get(gd::GroupedDataFrame, key::GroupKeyTypes, default)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -33,7 +33,8 @@ and combines the result into a data frame).
 which can also be used to get the values of the grouping columns for each group.
 `Tuples` and `NamedTuple`s containing the values of the grouping columns (in the
 same order as the `cols` argument) are also accepted as indices, but this will
-be slower than using the equivalent `GroupKey`.
+be slower than using the equivalent `GroupKey`. Finally, a `Dict` can be used to index
+into a grouped data frame where the keys are column names of the data frame in any order.
```
nalimilan, line 37:
```suggestion
be slower than using the equivalent `GroupKey`. Finally, a `Dict` with
column names as keys can be used (order does not matter in this case).
Note that indexing with `GroupKey` is faster than other methods.
```
(And remove the mention of slower above for tuples and named tuples.)

test/grouping.jl
```diff
@@ -1351,17 +1358,23 @@ end
...
     @test_throws KeyError gd[(:A, 1, 1)]
+    @test_throws KeyError gd[Dict("a" => :A, "b" => 1, "c" => 1)]
     # Out of order
     @test_throws KeyError gd[(b=1, a=:A)]
     @test_throws KeyError gd[(1, :A)]
     # Empty
     @test_throws KeyError gd[()]
     @test_throws KeyError gd[NamedTuple()]
+
+    # Substring and abstractstring
```
nalimilan, line 1375:
```suggestion
    # SubString and AbstractString
```

src/groupeddataframe/groupeddataframe.jl
bkamins: why not. This probably requires a careful consideration if this does not introduce problems though (probably not).

For sure we can provide a conversion from `GroupKey` to a dictionary only for a concrete type (`Dict`).

src/groupeddataframe/groupeddataframe.jl
bkamins: The general policy is to use `Symbol` in this case (which was settled after I looked at this PR).
I know @pdeffebach voted otherwise :smile:, but we should be consistent.

bkamins: Ah - as @nalimilan prompted me - we have not settled the decision in https://github.com/JuliaData/DataFrames.jl/issues/2298#issuecomment-646703922. So first we should decide there and then update this PR accordingly.

src/groupeddataframe/groupeddataframe.jl
pdeffebach: this can be resolved, sorry.

pdeffebach:
I've added `AbstractDict` for this, but am waiting on #2298 for the `String` vs `Symbol` output.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -323,6 +323,21 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function _dict_to_tuple(key::Union{AbstractDict{Symbol}, AbstractDict{<:AbstractString}}, gd)
+    if length(key) != length(gd.cols)
+        throw(KeyError(key))
+    end
+
+    T = keytype(key) <: Symbol ? Symbol : String
+    t = ntuple(length(gd.cols)) do i
```
bkamins, line 332:
```suggestion
    return ntuple(length(gd.cols)) do i 
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -210,7 +210,7 @@ Base.getindex(gd::GroupedDataFrame, idxs::Colon) =
 Key for one of the groups of a [`GroupedDataFrame`](@ref). Contains the values
 of the corresponding grouping columns and behaves similarly to a `NamedTuple`,
 but using it to index its `GroupedDataFrame` is much more effecient than using the
-equivalent `Tuple` or `NamedTuple`.
+equivalent `Tuple`, `NamedTuple`, or `Dict`.
```
bkamins, line 213:
```suggestion
equivalent `Tuple`, `NamedTuple`, or `AbstractDict`.
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -399,7 +418,7 @@ end
 Get the set of keys for each group of the `GroupedDataFrame` `gd` as a
 [`GroupKeys`](@ref) object. Each key is a [`GroupKey`](@ref), which behaves like
 a `NamedTuple` holding the values of the grouping columns for a given group.
-Unlike the equivalent `Tuple` and `NamedTuple`, these keys can be used to index
+Unlike the equivalent `Tuple`, `NamedTuple`, and `Dict`, these keys can be used to index
```
bkamins, line 421:
```suggestion
Unlike the equivalent `Tuple`, `NamedTuple`, and `AbstractDict`, these keys can be used to index
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -517,7 +539,8 @@ Base.haskey(gd::GroupedDataFrame, key::Union{Signed,Unsigned}) =
 Get a group based on the values of the grouping columns.
 
 `key` may be a `NamedTuple` or `Tuple` of grouping column values (in the same
-order as the `cols` argument to `groupby`).
+order as the `cols` argument to `groupby`). It may also be a `Dict` with
```
bkamins, line 542:
```suggestion
order as the `cols` argument to `groupby`). It may also be a `AbstractDict` with
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -33,7 +33,9 @@ and combines the result into a data frame).
 which can also be used to get the values of the grouping columns for each group.
 `Tuples` and `NamedTuple`s containing the values of the grouping columns (in the
 same order as the `cols` argument) are also accepted as indices, but this will
-be slower than using the equivalent `GroupKey`.
+be slower than using the equivalent `GroupKey`. Finally, a `Dict` with
```
bkamins, line 36:
```suggestion
be slower than using the equivalent `GroupKey`. Finally, a `AbstractDict` with
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -33,7 +33,9 @@ and combines the result into a data frame).
...
+column names as keys can be used (order does not matter in this case).
```
bkamins, line 37:
```suggestion
column names as keys can be used (order does not matter in this case
even for ordered dictionaries).
```

bkamins:
I think in #2298 we will end up with `Symbol` as the default. But given your comment there let us wait for more feedback. This PR will be merged after the decision in #2298 is made

bkamins:
Given the decision in https://github.com/JuliaData/DataFrames.jl/issues/2298 I think this PR can be finalized following it if you have time. Thank you!

bkamins:
If you feel it would be good to have it can you tell me if you have time to finalize this PR so that we can include it in the coming release?

pdeffebach:
yes, sorry! I have time I will work on it today.

pdeffebach:
I think I may have borked this with a force push that I thought was working well after a `git reset --soft`. 

I will need a bit more time to write tests and make sure the diff is correct and incorporates all previous comments.

bkamins:
OK - thank you for working on it. This is indeed an issue when a PR is stalled for some time. I will review it carefully when you let me know it is ready.

pdeffebach:
This is ready for a review. 

I somehow messed up a lot of this via git. I think I have the commit ready to go now, however, and incorporated the entire previous discussion, whether or not there is a valid git history tying it all together. 

Apologies for this mistake. I know it makes a bit more work for you both.

bkamins:
CI fails on Julia 1.0.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -387,6 +391,29 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function _dict_to_tuple(key::AbstractDict{<:AbstractString} , gd)
+    T = keytype(key)
+    return ntuple(length(gd.cols)) do i 
+        # Most `<: AbstractString` types do not support constructors from symbols
+        key[T(string(gd.cols[i]))]
```
bkamins, line 398: I would assume this should be enough (If I am right then capturing `T` above is not needed)
```suggestion
        key[string(gd.cols[i])]
```
also why do you prefer `do` syntax here instead of just `i -> key[string(gd.cols[i])]`?

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -387,6 +391,29 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return Base.to_index(gd, Tuple(key))
 end
 
+function _dict_to_tuple(key::AbstractDict{<:AbstractString} , gd)
+    T = keytype(key)
```
bkamins, line 395: Also check if `key` does not contain other entries than these in `gd.cols`. It is enough to test length, as other mismatches will be captured by indexing.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -387,6 +391,29 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
+    return ntuple(length(gd.cols)) do i 
+        # Most `<: AbstractString` types do not support constructors from symbols
+        key[T(string(gd.cols[i]))]
+    end
+end
+
+function _dict_to_tuple(key::AbstractDict{Symbol}, gd)
+    T = keytype(key)
+    return ntuple(length(gd.cols)) do i 
+        key[T(gd.cols[i])]
```
bkamins, line 405: here using `T` is not needed for sure, as `gd.cols` contains `Symbols`. Other comments as above

src/groupeddataframe/groupeddataframe.jl
bkamins: ah - you check it below, but I think it would be better to check it in these methods and drop a check in the calling methods (I feel it would be less error prone in the long run).

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -387,6 +391,29 @@ function Base.to_index(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
+    end
+end
+
+function Base.to_index(gd::GroupedDataFrame, key::Union{AbstractDict{Symbol},AbstractDict{<:AbstractString}})
+    if length(key) != length(gd.cols)
+        throw(KeyError(key))
```
bkamins, line 411: if you drop this check here (in favor of `_dict_to_tuple` - this will be a simpler one liner)

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -572,6 +603,14 @@ function Base.haskey(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
     return haskey(gd, Tuple(key))
 end
 
+function Base.haskey(gd::GroupedDataFrame, key::AbstractDict{T, S}) where {T<:Union{Symbol, <:AbstractString}, S}
```
bkamins, line 606: it seems to me that this should be enough
```suggestion
function Base.haskey(gd::GroupedDataFrame, key::AbstractDict{<:Union{Symbol, AbstractString}})
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -572,6 +603,14 @@ function Base.haskey(gd::GroupedDataFrame, key::NamedTuple{N}) where {N}
...
+    if length(key) != length(gd.cols) || any(Symbol(n) ∉ _names(gd) for n in keys(key))
```
bkamins, line 607: this test will not be needed if you move it inside `_dict_to_tuple`.

src/groupeddataframe/groupeddataframe.jl
pdeffebach: It was longer before for some reason. Made in line now.

src/groupeddataframe/groupeddataframe.jl
pdeffebach: Yes. all gone noe.

pdeffebach:
Thanks for the review! Ready for another look.

bkamins:
CI fails on 1.0 and on nightly - can you please have a look what causes this (some `convert` errors, but you know your changes better).

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -328,6 +330,8 @@ Base.convert(::Type{Array{T}}, key::GroupKey) where {T} = Vector{T}(key)
 Base.Array(key::GroupKey) = Vector(key)
 Base.Array{T}(key::GroupKey) where {T} = Vector{T}(key)
 
+Base.Dict(key::GroupKey) = Dict(Symbol(first(ki)) => last(ki) for ki in pairs(key))
```
bkamins, line 333: This should be enough I think as `keys` will produce names as `Symbol` already (maybe this will resolve the CI errors?)
```suggestion
Base.Dict(key::GroupKey) = Dict(pairs(key))
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -581,7 +611,8 @@ Base.haskey(gd::GroupedDataFrame, key::Union{Signed,Unsigned}) =
 Get a group based on the values of the grouping columns.
 
 `key` may be a `NamedTuple` or `Tuple` of grouping column values (in the same
-order as the `cols` argument to `groupby`).
+order as the `cols` argument to `groupby`). It may also be an `AbstractDict`, in which case the
```
bkamins, line 614: also please add it can be `GroupKey` as in the implementation you allow it (and I think it is OK to allow it - just please make sure we test also this case)

bkamins:
Also can you please merge this branch with upstream master (it should be easier than rebasing :smile:) and add an entry to NEWS.md?

bkamins:
CI still fails - can you please check the PR on Julia 1.0 and on Julia nightly to check what causes a problem?

Thank you!

pdeffebach:
Works now!

bkamins:
OK - so can you please just merge it with master and add NEWS.md entry please?

@nalimilan - if you do not comment on this I am going to merge this PR after @pdeffebach adds NEWS.md entry and then work on implementing the conclusions from https://github.com/JuliaData/DataFrames.jl/issues/2305 including handling `Dict`.

bkamins (approved):

bkamins:
bump :smile: (I would finish it not to have constantly merge it with master)

pdeffebach:
Thanks for pinging me on this! Slipped through. Thanks for the feedback Milan.

bkamins:
also merge conflicts need resolving

bkamins (approved):

nalimilan (approved):

pdeffebach:
Conflict was just `News.md`. This can be merged.

bkamins:
Thank you!

PR #2282, opened 2020-06-11
improve error message when validating

### LINKED ISSUE #1732: When `join(..., validate=(true,true))` fails, it should include a list of non-unique joinkeyrows in the error)

bkamins:
Fix #1732

To see the result take out the test code and run it in REPL (just remove `@test_throws` and make sure that the loops go on working and do not stop after first error)

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -268,20 +268,77 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
     left_invalid = validate[1] ? any(nonunique(joiner.dfl, joiner.left_on)) : false
     right_invalid = validate[2] ? any(nonunique(joiner.dfr, joiner.right_on)) : false
 
+    if validate[1]
+        non_unique_left = nonunique(joiner.dfl, joiner.left_on)
+        if any(non_unique_left)
+            left_invalid = true
+            non_unique_left_df = unique(joiner.dfl[non_unique_left, joiner.left_on])
+            nrow_nul_df = nrow(non_unique_left_df)
+            @assert nrow_nul_df > 0
+            if nrow_nul_df == 1
+                left_invalid_msg = "df1 contains 1 duplicate key: " *
+                                   "$(NamedTuple(non_unique_left_df[1, :])). "
+            elseif nrow_nul_df == 2
+                left_invalid_msg = "df1 contains 2 duplicate keys: " *
+                                   "$(NamedTuple(non_unique_left_df[1, :])) and " *
+                                   "$(NamedTuple(non_unique_left_df[2, :])). "
+            else
+                left_invalid_msg = "df1 contains $(nrow_nul_df) duplicate keys: " *
+                                   "$(NamedTuple(non_unique_left_df[1, :])), ..., " *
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/join.jl
```diff
@@ -268,20 +268,77 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
     right_invalid = validate[2] ? any(nonunique(joiner.dfr, joiner.right_on)) : false
 
+    if validate[1]
+        non_unique_left = nonunique(joiner.dfl, joiner.left_on)
+        if any(non_unique_left)
+            left_invalid = true
+            non_unique_left_df = unique(joiner.dfl[non_unique_left, joiner.left_on])
+            nrow_nul_df = nrow(non_unique_left_df)
+            @assert nrow_nul_df > 0
+            if nrow_nul_df == 1
```
nalimilan, line 278: Would there be a way to just collect all duplicates in a vector and leave printing code decide how many entries can fit horizontally on screen? In my experiments, the best I could find is this:
```julia
julia> print(IOContext(stdout, :limit=>true), repeat([(a=1, b=2),], 1000))
NamedTuple{(:a, :b),Tuple{Int64,Int64}}[(a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2)  …  (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2), (a = 1, b = 2)]
```
That's not ideal as the eltype info is ugly and it doesn't really respect the screen width, but maybe we can improve that?

bkamins: The problem is that cannot rely on `stdout`/`sterr` and screen width. We do not know to what stream this message will be sent to. What I proposed is a "safe" approach. Alternatively, we can agree on some number of characters to be printed (but that will be much more complex to implement, as we would be almost reproducing the code printing data frames (calculating width of fields etc.)).

Therefore I would leave as I proposed as making it better will be much more complex.

nalimilan: Good point. A solution would be to throw a custom error type so that the formatting only happens in its `showerror` method. Not sure it's worth it, leaving it up to you.

bkamins: I would leave it as is not to complicate the codebase. Thank you!

bkamins:
Thank you. I will backport it so that we get a feedback from the users faster in case something needs fixing here (I guess we can treat the change of error message as non-breaking).

PR #2284, opened 2020-06-11
let indicator allow strings

### LINKED ISSUE #2283: `indicator` column in `join`s should allow `String`s
```
julia> df1 = DataFrame(a = [1, 2, 3]);

julia> df2 = DataFrame(a = [1, 2, 3, 4]);

julia> outerjoin(df1, df2, on = "a", indicator = "_merge")
ERROR: TypeError: in keyword argument indicator, expected Union{Nothing, Symbol}, got String
Stacktrace:
 [1] top-level scope at REPL[12]:1
```

pdeffebach:
Fixed #2283

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -242,6 +242,7 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
     _check_consistency(df2)
 
     if indicator !== nothing
+        indicator = Symbol(indicator)
         indicator_cols = ["_left", "_right"]
         for i in 1:2
             while (hasproperty(df1, Symbol(indicator_cols[i])) ||
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -488,9 +488,9 @@ the result. A left join includes all rows from `df1`.
   if duplicate names are found in columns not joined on;
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
-- `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
-   column with the given name for whether a row appeared in only `df1` (`"left_only"`),
-   only `df2` (`"right_only"`) or in both (`"both"`). If `Symbol` is already in use,
+- `indicator` : Default: `nothing`. If a `Symbol` or `String`, adds categorical indicator
+   column with the given name, for whether a row appeared in only `df1` (`"left_only"`),
+   only `df2` (`"right_only"`) or in both (`"both"`). If the name is already in use,
    the column name will be modified if `makeunique=true`.
 - `validate` : whether to check that columns passed as the `on` argument
    define unique keys in each input data frame (according to `isequal`).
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/join.jl
```diff
@@ -489,7 +489,7 @@ the result. A left join includes all rows from `df1`.
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
 - `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
-   column named `Symbol` for whether a row appeared in only `df1` (`"left_only"`),
+   column with the given name for whether a row appeared in only `df1` (`"left_only"`),
```
bkamins, line 492: In line 491 add that string is also allowed

src/abstractdataframe/join.jl
```diff
@@ -236,7 +236,7 @@ end
 
 function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
                on::Union{<:OnType, AbstractVector}, kind::Symbol, makeunique::Bool,
-               indicator::Union{Nothing, Symbol},
+               indicator::Union{Nothing, Symbol, <:AbstractString},
```
bkamins, line 239: `<:` is not needed in `Union` here and below.

src/abstractdataframe/join.jl
bkamins: The same below

bkamins:
Apart from the small comments there is a small problem.
You have left out a test:
```
Symbol(indicator_cols[i]) == indicator
```
which will always fail is `indicator` is a string.

However, now as I look at it it actually seems it can be removed if we move line:
```
select!(joined, Not([df1_ind, df2_ind]))
```
before line:
```
unique_indicator = indicator
```
(if it is not clear for you why then I can fix this after this PR is merged)

pdeffebach:
Thanks for the review. 

Indeed I cannot figure out this error you described. I can't figure out a way to write code that throws an error about comparing a symbol to a string. I think you may have to fix it after the merge.

src/abstractdataframe/join.jl
```diff
@@ -488,9 +488,9 @@ the result. A left join includes all rows from `df1`.
   if duplicate names are found in columns not joined on;
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
-- `indicator` : Default: `nothing`. If a `Symbol`, adds categorical indicator
-   column named `Symbol` for whether a row appeared in only `df1` (`"left_only"`),
-   only `df2` (`"right_only"`) or in both (`"both"`). If `Symbol` is already in use,
+- `indicator` : Default: `nothing`. If a `Symbol` or `String`, adds categorical indicator
```
bkamins, line 491:
```suggestion
- `indicator` : Default: `nothing`. If a `Symbol` or string, adds categorical indicator
```

src/abstractdataframe/join.jl
```diff
@@ -591,9 +591,9 @@ the result. A right join includes all rows from `df2`.
...
```
bkamins, line 594:
```suggestion
- `indicator` : Default: `nothing`. If a `Symbol` or string, adds categorical indicator
```

src/abstractdataframe/join.jl
```diff
@@ -699,9 +699,9 @@ of the passed data frames.
...
```
bkamins, line 702:
```suggestion
- `indicator` : Default: `nothing`. If a `Symbol` or string, adds categorical indicator
```

src/abstractdataframe/join.jl
```diff
@@ -817,9 +817,9 @@ match with the keys in `df2`.
...
```
bkamins, line 820:
```suggestion
- `indicator` : Default: `nothing`. If a `Symbol` or string, adds categorical indicator
```

bkamins:
OK - I can clean up the code later.

For now can you please just add after line https://github.com/JuliaData/DataFrames.jl/pull/2284/files#diff-3d9293bcc4eac4333160c4d203ae91d7R244 (i.e. as a first statement executed if `indicator !== nothing` is true)
the following line:
```
indicator = Symbol(indicator)
```
it will fix the bug (not in the cleanest way, but will do for now) as then we are sure that in `Symbol(indicator_cols[i]) == indicator)` we have symbols both on left and right hand sides.

I cannot fix it via web interface unfortunately.

pdeffebach:
Added! let me know what else to add

bkamins:
Looks good. Thank you!

bkamins:
Thank you!

Issue #2289, opened 2020-06-15
repeated "signal (11): Segmentation fault", but randomly happens


zhongyunde:
A simplified task is described like this: 
```
using DataStructures 
using DataFrames
vt = Union{Int64, Float64, Bool}
d = OrderedDict{Symbol, vt}(:a=>0.0, :b=>0.0, :c=>0.0)
df = DataFrame([vt, vt, vt], [:a, :b, :c], 0)
for i = 1:500
    d[:a] = rand()
    d[:b] = rand() 
    d[:c] = rand()
    push!(df, d) 
end 
```
There is no issues when run this task individually. But when it runs to process 2000 independent tasks on 12 threads for 4 hours (note: each task runs on its own thread), problem happens. It throws repeated "signal (11): Segmentation fault", but randomly. The exception message is below: 

```
signal (11): Segmentation fault
...
jl_unwrap_unionall at /buildworker/worker/package_linux64/build/src/jltypes.c:955
jl_typemap_assoc_exact at /buildworker/worker/package_linux64/build/src/julia_internal.h:1025 [inlined]
jl_lookup_generic_ at /buildworker/worker/package_linux64/build/src/gf.c:2270 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2319
iterate at /home/london/.julia/packages/OrderedCollections/P6ntV/src/ordered_dict.jl:417
iterate at ./abstractdict.jl:60 [inlined]
_all at ./reduce.jl:819 [inlined]
all at ./reduce.jl:815 [inlined]
#push!#186 at /home/london/.julia/packages/DataFrames/kwVTY/src/dataframe/dataframe.jl:1337
push! at /home/london/.julia/packages/DataFrames/kwVTY/src/dataframe/dataframe.jl:1315
```
I suspected the problem came from the code (dataframe.jl, Line 1337) 
```
all(x -> x isa AbstractString, keys(row))
```
So I made some changes to it today, and after that it worked very well. No more Segmentation fault for today. I do no know whether the problem is permanently fixed or not, but I can tell the changes helped.  
```
#dataframe.jl, Line 1337
- if row isa AbstractDict && all(x -> x isa AbstractString, keys(row))
+ if row isa AbstractDict && (eltype(row).types[1] <: AbstractString)
     row = (;(Symbol.(keys(row)) .=> values(row))...) 
  end 
...
#dataframe.jl, Line 1345 
[...]

PR #2290, opened 2020-06-15
add handling to cases when all is not optimized out

bkamins:
This should not change any functionality, but just speed up common cases, as it seems that `all` tests are not optimized out by the compiler.

Also it is a workaround for https://github.com/JuliaData/DataFrames.jl/issues/2289, although probably the bug there is unrelated to DataFrames.jl

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1344,12 +1344,9 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple};
     end
 
     old_row_type = typeof(row)
-    if row isa AbstractDict
-        if !(keytype(row) <: Symbol)
-            if keytype(row) <: AbstractString || all(x -> x isa AbstractString, keys(row))
-                row = (;(Symbol.(keys(row)) .=> values(row))...)
-            end
-        end
+    if row isa AbstractDict && keytype(row) !== Symbol &&
+        (keytype(row) <: AbstractString || all(x -> x isa AbstractString, keys(row)))
+        row = (;(Symbol.(keys(row)) .=> values(row))...)
     end
 
     # in the code below we use a direct access to _columns because
@@ -1357,7 +1354,7 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple};
     # inconsistent and normal data frame indexing would error.
     if cols == :union
         if row isa AbstractDict
-            if !(keytype(row) <: Symbol) && !all(x -> x isa Symbol, keys(row))
+            if keytype(row) !== Symbol && !all(x -> x isa Symbol, keys(row))
                 throw(ArgumentError("when `cols == :union` all keys of row must be Symbol"))
             end
         end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -1353,10 +1353,8 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple};
     # we resize the columns so temporarily the `DataFrame` is internally
     # inconsistent and normal data frame indexing would error.
     if cols == :union
-        if row isa AbstractDict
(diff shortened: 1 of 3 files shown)


bkamins:
only CI fails here

src/dataframe/dataframe.jl
```diff
@@ -1344,16 +1344,22 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple};
...
 
     old_row_type = typeof(row)
-    if row isa AbstractDict && all(x -> x isa AbstractString, keys(row))
-        row = (;(Symbol.(keys(row)) .=> values(row))...)
+    if row isa AbstractDict
+        if !(keytype(row) <: Symbol)
+            if keytype(row) <: AbstractString || all(x -> x isa AbstractString, keys(row))
+                row = (;(Symbol.(keys(row)) .=> values(row))...)
+            end
+        end
```
nalimilan, line 1352: Why not a single `if`?
```suggestion
    if row isa AbstractDict && keytype(row) !== Symbol &&
        (keytype(row) <: AbstractString || all(x -> x isa AbstractString, keys(row)))
        row = (;(Symbol.(keys(row)) .=> values(row))...)
```

src/dataframe/dataframe.jl
```diff
@@ -1344,16 +1344,22 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple};
...
     end
 
     # in the code below we use a direct access to _columns because
     # we resize the columns so temporarily the `DataFrame` is internally
     # inconsistent and normal data frame indexing would error.
     if cols == :union
-        if row isa AbstractDict && !all(x -> x isa Symbol, keys(row))
-            throw(ArgumentError("when `cols == :union` all keys of row must be Symbol"))
+        if row isa AbstractDict
+            if !(keytype(row) <: Symbol) && !all(x -> x isa Symbol, keys(row))
```
nalimilan, line 1360:
```suggestion
            if keytype(row) !== Symbol && !all(x -> x isa Symbol, keys(row))
```

src/dataframerow/dataframerow.jl
```diff
@@ -111,8 +111,10 @@ for T in (:AbstractVector, :Regex, :Not, :Between, :All, :Colon)
         end
 
         if v isa AbstractDict
```
nalimilan, line 113: Same here (but GitHub won't let me make a suggestion...).

bkamins: I accepted the suggestions and cleaned this proposal, though I am not sure if it is more readable (but I am OK with it).

bkamins:
only coverage fails here

src/dataframerow/dataframerow.jl
nalimilan: Yeah it's not really more readable, but having nested `if`s can be confusing IMHO, as they look like some code will be run outside of inner `if`s.

src/dataframe/dataframe.jl
```diff
@@ -1344,16 +1344,17 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple};
...
     end
 
     # in the code below we use a direct access to _columns because
     # we resize the columns so temporarily the `DataFrame` is internally
     # inconsistent and normal data frame indexing would error.
     if cols == :union
-        if row isa AbstractDict && !all(x -> x isa Symbol, keys(row))
-            throw(ArgumentError("when `cols == :union` all keys of row must be Symbol"))
+        if row isa AbstractDict && keytype(row) !== Symbol && !all(x -> x isa Symbol, keys(row))
+                throw(ArgumentError("when `cols == :union` all keys of row must be Symbol"))
```
nalimilan, line 1357:
```suggestion
            throw(ArgumentError("when `cols == :union` all keys of row must be Symbol"))
```

bkamins:
Thank you!

Issue #2229, opened 2020-05-06
findfirst/findlast/nextind/prevind not working for eachcol in v0.21.0


rasmushenningsson:
This code works in v0.20.2,
```julia
df = DataFrame(x=["a","b","c"], y=1:3)
findfirst(isequal(1:3),eachcol(df))
```
but errors in v0.21.0 with the error message.
```
ERROR: MethodError: no method matching nextind(::DataFrames.DataFrameColumns{DataFrame}, ::Symbol)
```

The cause seems to be that the underlying indices are now Symbols and not Ints.  I didn't see this in the list of breaking changes so I figured it might be unintentional. 
(A note, findall() works, but returns Symbols instead of Ints as before.)

Thanks for the new release and all the hard work!

bkamins:
This was intentional to make the following work consistently:
```
julia> keys(eachcol(df))
2-element Array{Symbol,1}:
 :x
 :y

julia> pairs(eachcol(df))
Iterators.Pairs(::DataFrames.DataFrameColumns{DataFrame}, ::Array{Symbol,1}) with 2 entries:
  :x => ["a", "b", "c"]
  :y => [1, 2, 3]
```
This was consider more useful than:
```
julia> x = rand(3)
3-element Array{Float64,1}:
 0.5557510873245723
 0.43710797460962514
 0.42471785049513144

julia> keys(x)
3-element LinearIndices{1,Tuple{Base.OneTo{Int64}}}:
 1
 2
 3

julia> pairs(x)
pairs(::Array{Float64,1}) with 3 entries:
  1 => 0.555751
  2 => 0.437108
  3 => 0.424718
```
but indeed it has a drawback you have highlighted.

I think the solution is to add custom implementations of the functions you mention. Would that work for you?

bkamins:
(of course breaking these functions was not intentional and I have forgotten to mention this in release notes as there were 102 PRs to annotate so it slipped through)

rasmushenningsson:
> I think the solution is to add custom implementations of the functions you mention. Would that work for you?

Absolutely. I don't think it's a common use case, but nice to have once in a while.

bkamins:
We should also fix:
```
julia> df = DataFrame(a=1,b=2)
1×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │

julia> hash(eachcol(df))
[...]

PR #2291, opened 2020-06-16
[BREAKING] Make DataFrameColumns stop being an AbstractVector

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2229
I have added functionalities for `DataFrameColumns` that I find useful, but please comment if you would like to see more.

### DIFF
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -110,8 +110,16 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
     DataFrameColumns{<:AbstractDataFrame}
 
 A vector-like object that allows iteration over columns of an `AbstractDataFrame`.
-Indexing into `DataFrameColumns` objects using integer or symbol indices
+
+Indexing into `DataFrameColumns` objects using integer, `Symbol` or string
 returns the corresponding column (without copying).
+Indexing into `DataFrameColumns` objects using multiple column selector
+returns a subsetted `DataFrameColumns` object with parent being a view of the
+original containg only the selected columns.
+
+It supports most of the `AbstractVector` API. The key differences are that it is
+read-only and is that the `keys` function returns a vector of `Symbol`s (and not
+integers as for normal vectors).
 """
 struct DataFrameColumns{T<:AbstractDataFrame}
     df::T
@@ -125,6 +133,13 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
 
 Return a `DataFrameColumns` object that is a vector-like
 that allows iterating an `AbstractDataFrame` column by column.
+
+Indexing into `DataFrameColumns` objects using integer, `Symbol` or string
+returns the corresponding column (without copying).
+Indexing into `DataFrameColumns` objects using multiple column selector
+returns a subsetted `DataFrameColumns` object with parent being a view of the
+original containg only the selected columns.
+
 It supports most of the `AbstractVector` API. The key differences are that it is
 read-only and is that the `keys` function returns a vector of `Symbol`s (and not
 integers as for normal vectors).
@@ -178,7 +193,7 @@ Base.iterate(itr::DataFrameColumns, i::Integer=1) =
 Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns, idx::ColumnIndex) =
     parent(itr)[!, idx]
 Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns, idx::MultiColumnIndex) =
-    eachcol(parent(itr)[!, idx])
+    eachcol(view(parent(itr), !, idx))
 Base.:(==)(itr1::DataFrameColumns, itr2::DataFrameColumns) =
     parent(itr1) == parent(itr2)
 Base.isequal(itr1::DataFrameColumns, itr2::DataFrameColumns) =
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -126,7 +126,7 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
 Return a `DataFrameColumns` that is an `AbstractVector`
 that allows iterating an `AbstractDataFrame` column by column.
 It supports most of `AbstractVector` API. The key differences are that it
-read-only and is that the `keys` function returns a vector of `Symbols` (and not
+read-only and is that the `keys` function returns a vector of `Symbol`s (and not
 integers as for normal vectors).
 
 # Examples
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -125,7 +125,7 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
 
 Return a `DataFrameColumns` that is an `AbstractVector`
 that allows iterating an `AbstractDataFrame` column by column.
-It supports most of `AbstractVector` API. The key differences are that it
+It supports most of `AbstractVector` API. The key differences are that it is
 read-only and is that the `keys` function returns a vector of `Symbol`s (and not
 integers as for normal vectors).
 
@@ -207,7 +207,7 @@ Base.keys(itr::DataFrameColumns) = propertynames(itr)
 """
     values(dfc::DataFrameColumns)
 
-Get a vector of columns of `dfc`.
+Get a vector of columns from `dfc`.
 """
 Base.values(itr::DataFrameColumns) = collect(itr)
 
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -109,7 +109,7 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
 """
     DataFrameColumns{<:AbstractDataFrame}
 
-A generator that allows iteration over columns of an `AbstractDataFrame`.
+A vector-like object that allows iteration over columns of an `AbstractDataFrame`.
 Indexing into `DataFrameColumns` objects using integer or symbol indices
 returns the corresponding column (without copying).
 """
@@ -123,9 +123,9 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
 """
     eachcol(df::AbstractDataFrame)
 
-Return a `DataFrameColumns` that is an `AbstractVector`
+Return a `DataFrameColumns` object that is a vector-like
 that allows iterating an `AbstractDataFrame` column by column.
-It supports most of `AbstractVector` API. The key differences are that it is
+It supports most of the `AbstractVector` API. The key differences are that it is
 read-only and is that the `keys` function returns a vector of `Symbol`s (and not
 integers as for normal vectors).
 
@@ -173,10 +173,11 @@ Base.length(itr::DataFrameColumns) = size(itr)[1]
 Base.eltype(::Type{<:DataFrameColumns}) = AbstractVector
 Base.firstindex(itr::DataFrameColumns) = 1
 Base.lastindex(itr::DataFrameColumns) = length(itr)
-Base.iterate(itr::DataFrameColumns, i=1) =
+Base.iterate(itr::DataFrameColumns, i::Integer=1) =
     i <= length(itr) ? (itr[i], i + 1) : nothing
-Base.getindex(itr::DataFrameColumns, idx::ColumnIndex) = parent(itr)[!, idx]
-Base.getindex(itr::DataFrameColumns, idx::MultiColumnIndex) =
+Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns, idx::ColumnIndex) =
+    parent(itr)[!, idx]
+Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns, idx::MultiColumnIndex) =
     eachcol(parent(itr)[!, idx])
 Base.:(==)(itr1::DataFrameColumns, itr2::DataFrameColumns) =
     parent(itr1) == parent(itr2)
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -125,8 +125,9 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
 
 Return a `DataFrameColumns` that is an `AbstractVector`
 that allows iterating an `AbstractDataFrame` column by column.
-Additionally it is allowed to index `DataFrameColumns` using column names,
-and convenience functions: `keys`, `values`, `pairs` are defined for it.
+It supports most of `AbstractVector` API. The key differences are that it
+read-only and is that the `keys` function returns a vector of `Symbols` (and not
+integers as for normal vectors).
 
 # Examples
 ```jldoctest
@@ -160,8 +161,18 @@ julia> sum.(eachcol(df))
 """
 eachcol(df::AbstractDataFrame) = DataFrameColumns(df)
 
-Base.length(itr::DataFrameColumns) = size(parent(itr), 2)
+Base.IteratorSize(::Type{<:DataFrameColumns}) = Base.HasShape{1}()
+Base.size(itr::DataFrameColumns) = (size(parent(itr), 2),)
+
+function Base.size(itr::DataFrameColumns, d::Integer)
+    d < 1 && throw(ArgumentError("dimension out of range"))
+    return d == 1 ? size(itr)[1] : 1
+end
+
+Base.length(itr::DataFrameColumns) = size(itr)[1]
 Base.eltype(::Type{<:DataFrameColumns}) = AbstractVector
+Base.firstindex(itr::DataFrameColumns) = 1
+Base.lastindex(itr::DataFrameColumns) = length(itr)
 Base.iterate(itr::DataFrameColumns, i=1) =
     i <= length(itr) ? (itr[i], i + 1) : nothing
 Base.getindex(itr::DataFrameColumns, idx::ColumnIndex) = parent(itr)[!, idx]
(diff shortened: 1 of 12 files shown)


bkamins:
@nalimilan - it should be good to have a look at.

bkamins:
As usual some corner cases of the changes popped up.
Good change:
`DataFrame(eachcol(df))` is now no-op (except copying of columns), previously it dropped column names.
Bad change (but maybe not that bad): `eachcol(df)[1,1]` now errors as it would produce duplicate column name.

src/abstractdataframe/iteration.jl
```diff
@@ -125,7 +125,9 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
 
 Return a `DataFrameColumns` that is an `AbstractVector`
 that allows iterating an `AbstractDataFrame` column by column.
-Additionally it is allowed to index `DataFrameColumns` using column names.
+It supports most of `AbstractVector` API. The key differences are that it
```
quinnj, line 128: `it is`

src/abstractdataframe/iteration.jl
```diff
@@ -190,6 +204,13 @@ Get a vector of column names of `dfc` as `Symbol`s.
 """
 Base.keys(itr::DataFrameColumns) = propertynames(itr)
 
+"""
+    values(dfc::DataFrameColumns)
+
+Get a vector of columns of `dfc`.
```
quinnj, line 210: "Get a vector of columns from `dfc`"

test/iteration.jl
```diff
@@ -22,13 +22,23 @@ using Test, DataFrames
...
+    @test size(eachcol(df), 2) == 1
+    @test_throws ArgumentError size(eachcol(df), 0)
     @test eachcol(df)[1] == df[:, 1]
+    @test eachcol(df)[:A] == df[:, :A]
+    @test eachcol(df)[All()] == eachcol(df)
+    @test isequal(eachcol(df)[[1]], eachcol(df[!, [1]]))
+    @test eachcol(df).A == df[:, :A]
+    @test eachcol(df)["A"] == df[:, "A"]
+    @test eachcol(df)."A" == df[:, "A"]
+    @test eachcol(df)[begin] == df[!, 1]
```
quinnj, line 40: I think you need to wrap this test in a version check for whenever `begin` was allowed in indexing (tests for 1.0 fail because of it)

bkamins: Indeed. Actually we have a separate file for these tests (as we have to conditionally `include` it), but I have forgotten about it :disappointed:.

src/abstractdataframe/iteration.jl
bkamins: fixed

src/abstractdataframe/iteration.jl
bkamins: fixed

bkamins:
Thank you for having a look at it!

rasmushenningsson:
This looks great!

src/abstractdataframe/iteration.jl
```diff
@@ -190,6 +204,13 @@ Get a vector of column names of `dfc` as `Symbol`s.
 """
 Base.keys(itr::DataFrameColumns) = propertynames(itr)
 
+"""
+    values(dfc::DataFrameColumns)
+
+Get a vector of columns from `dfc`.
+"""
+Base.values(itr::DataFrameColumns) = collect(itr)
```
nalimilan, line 212: Would this definition be better? It could make sense to use it instead so that the eltype is always the same. People can use `collect` if they prefer the narrowest type.
```suggestion
Base.values(itr::DataFrameColumns) = collect(AbstractVector, itr)
```

test/select.jl
```diff
@@ -619,12 +619,12 @@ end
 
     df = DataFrame(rand(10, 4))
     df2 = select(df, :, :x1 => :x3)
-    @test df2 == DataFrame(eachcol(df)[[1,2,1,4]])
+    @test df2 == DataFrame(collect(eachcol(df))[[1,2,1,4]])
```
nalimilan, line 622: What does this do without `collect`?

src/abstractdataframe/iteration.jl
```diff
@@ -159,15 +161,27 @@ julia> sum.(eachcol(df))
...
 
-Base.getindex(itr::DataFrameColumns, j::Symbol) = parent(itr)[!, j]
+Base.length(itr::DataFrameColumns) = size(itr)[1]
+Base.eltype(::Type{<:DataFrameColumns}) = AbstractVector
+Base.firstindex(itr::DataFrameColumns) = 1
+Base.lastindex(itr::DataFrameColumns) = length(itr)
+Base.iterate(itr::DataFrameColumns, i=1) =
+    i <= length(itr) ? (itr[i], i + 1) : nothing
+Base.getindex(itr::DataFrameColumns, idx::ColumnIndex) = parent(itr)[!, idx]
+Base.getindex(itr::DataFrameColumns, idx::MultiColumnIndex) =
```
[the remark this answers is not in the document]
bkamins: OK, but note that Base does not put this restriction.

src/abstractdataframe/iteration.jl
```diff
@@ -159,15 +161,27 @@ julia> sum.(eachcol(df))
...
+    d < 1 && throw(ArgumentError("dimension out of range"))
+    return d == 1 ? size(itr)[1] : 1
 end
 
-Base.getindex(itr::DataFrameColumns, j::Symbol) = parent(itr)[!, j]
+Base.length(itr::DataFrameColumns) = size(itr)[1]
+Base.eltype(::Type{<:DataFrameColumns}) = AbstractVector
+Base.firstindex(itr::DataFrameColumns) = 1
+Base.lastindex(itr::DataFrameColumns) = length(itr)
+Base.iterate(itr::DataFrameColumns, i=1) =
```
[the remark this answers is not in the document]
bkamins: OK, but note that Base does not put this restriction.

src/abstractdataframe/iteration.jl
bkamins: This is not needed, as `eltype(itr)` is already `AbstractVector`. To make the narrowest type one has to use comprehension `[v for v in itr]`

src/abstractdataframe/iteration.jl
```diff
@@ -159,15 +160,17 @@ julia> sum.(eachcol(df))
...
-Base.IndexStyle(::Type{<:DataFrameColumns}) = Base.IndexLinear()
-
-@inline function Base.getindex(itr::DataFrameColumns, j::Int)
-    @boundscheck checkbounds(itr, j)
-    @inbounds parent(itr)[!, j]
-end
-
-Base.getindex(itr::DataFrameColumns, j::Symbol) = parent(itr)[!, j]
+Base.length(itr::DataFrameColumns) = size(parent(itr), 2)
+Base.eltype(::Type{<:DataFrameColumns}) = AbstractVector
```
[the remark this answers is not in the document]
bkamins: This definition is not needed as it is already present by default:
```
julia> Base.IteratorEltype(typeof(eachcol(df)))
Base.HasEltype()

julia> Base.IteratorEltype(eachcol(df))
Base.HasEltype()
```

src/abstractdataframe/iteration.jl
```diff
@@ -198,6 +219,16 @@ with the corresponding column vector, i.e. `name => col`
 where `name` is the column name of the column `col`.
 """
 Base.pairs(itr::DataFrameColumns) = Base.Iterators.Pairs(itr, keys(itr))
+Base.findnext(f::Function, itr::DataFrameColumns, i::Integer) =
```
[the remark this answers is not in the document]
bkamins: I was thinking about it, but decided to leave it out, as ordering of column names is non-obvious (as opposed to integers) and it is easy enough to use the `columnindex` function to go from name to index.

src/subdataframe/subdataframe.jl
```diff
@@ -168,7 +168,7 @@ function DataFrame(sdf::SubDataFrame; copycols::Bool=true)
     if copycols
         sdf[:, :]
     else
-        DataFrame(eachcol(sdf), _names(sdf), copycols=false)
+        DataFrame(collect(eachcol(sdf)), _names(sdf), copycols=false)
```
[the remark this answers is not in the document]
bkamins: Not needed as commented above, eltype will be `AbstractVector` already.

test/select.jl
bkamins: `eachcol(df)[[1,2,1,4]]` will error, because multi column index on `eachcol(df)` will create a new data frame and return a `DataFrameColumns` wrapper of it, and in this case we have a duplicate index `1`, which will error when creating an internal data frame.

src/abstractdataframe/iteration.jl
```diff
@@ -159,15 +161,28 @@ julia> sum.(eachcol(df))
...
+Base.length(itr::DataFrameColumns) = size(itr)[1]
+Base.eltype(::Type{<:DataFrameColumns}) = AbstractVector
+Base.firstindex(itr::DataFrameColumns) = 1
+Base.lastindex(itr::DataFrameColumns) = length(itr)
+Base.iterate(itr::DataFrameColumns, i::Integer=1) =
+    i <= length(itr) ? (itr[i], i + 1) : nothing
+Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns, idx::ColumnIndex) =
+    parent(itr)[!, idx]
+Base.@propagate_inbounds Base.getindex(itr::DataFrameColumns, idx::MultiColumnIndex) =
+    eachcol(parent(itr)[!, idx])
```
bkamins, line 181: Your questions prompted me to discuss this line. When using `MultiColumnIndex` we create a new data frame, but **do not** copy its columns (so it contains aliases). It is faster this way, and consistent with `getindex` for `ColumnIndex` but it is probably a bit unsafe. Do you think we should mention this behaviour somewhere?

bkamins: Actually I have decided to change this to produce a `view` as it is cleaner.

bkamins:
@nalimilan - your questions were exactly the things I had to check against Base when implementing it. These interfaces are really non-obvious.

test/select.jl
bkamins: @nalimilan - have a look at last commit. Making `eachcol(df)[columns]` return a `DataFrameColumns` object based on a view of the parent is quite appealing, but it also has the consequence that the type of column when you iterate it changes (because we have a `SubDataFrame` underneath so its columns are views, not original columns).

I still like the view approach, as it is more "correct" in terms of design. However, it has a drawback of changing the type of the vector (and probably also some small performance penalty). The alternative would be to make a copy of the parent without copying the columns. Then all is fast and column types are preserved, but the cost is that we "detach" this new data frame from the source (which we typically try to avoid - but maybe in this case we can accept it?).

src/abstractdataframe/iteration.jl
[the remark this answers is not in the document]
nalimilan: Ah indeed. Though, should we set it to `EltypeUnknown()` because we don't know the concrete type of columns? `?Base.IteratorEltype` says:
>   This trait is generally used to select between algorithms that pre-allocate a specific type of result, and algorithms that pick a result type based on the types of yielded values.

Not saying we should do it, but it's worth thinking about it.

src/abstractdataframe/iteration.jl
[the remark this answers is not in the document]
nalimilan: OK. I still tend to think it could be nice, but there's no hurry.

src/abstractdataframe/iteration.jl
nalimilan: So maybe it would be useful to have `value` always return a `Vector{AbstractVector}`, and `collect` choose the most specific type? Or maybe not. Cf. the discussion above about `IteratorEltype`.

src/abstractdataframe/iteration.jl
```diff
@@ -110,8 +110,16 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
     DataFrameColumns{<:AbstractDataFrame}
 
 A vector-like object that allows iteration over columns of an `AbstractDataFrame`.
-Indexing into `DataFrameColumns` objects using integer or symbol indices
+
+Indexing into `DataFrameColumns` objects using integer, `Symbol` or string
 returns the corresponding column (without copying).
+Indexing into `DataFrameColumns` objects using multiple column selector
```
nalimilan, line 116:
```suggestion
Indexing into `DataFrameColumns` objects using a multiple column selector
```

src/abstractdataframe/iteration.jl
```diff
@@ -110,8 +110,16 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
...
+returns a subsetted `DataFrameColumns` object with parent being a view of the
+original containg only the selected columns.
```
nalimilan, line 118:
```suggestion
original containing only the selected columns.
```

src/abstractdataframe/iteration.jl
```diff
@@ -110,8 +110,16 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
     DataFrameColumns{<:AbstractDataFrame}
 
 A vector-like object that allows iteration over columns of an `AbstractDataFrame`.
-Indexing into `DataFrameColumns` objects using integer or symbol indices
+
+Indexing into `DataFrameColumns` objects using integer, `Symbol` or string
 returns the corresponding column (without copying).
+Indexing into `DataFrameColumns` objects using multiple column selector
+returns a subsetted `DataFrameColumns` object with parent being a view of the
```
nalimilan, line 117:
```suggestion
returns a subsetted `DataFrameColumns` object with parent being a `SubDataFrame` view of the
```

src/abstractdataframe/iteration.jl
```diff
@@ -110,8 +110,16 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
...
+original containg only the selected columns.
+
+It supports most of the `AbstractVector` API. The key differences are that it is
```
nalimilan, line 120:
```suggestion
`DataFrameColumns` supports most of the `AbstractVector` API. The key differences are that it is
```

src/abstractdataframe/iteration.jl
```diff
@@ -110,8 +110,16 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
...
+read-only and is that the `keys` function returns a vector of `Symbol`s (and not
```
nalimilan, line 121:
```suggestion
read-only and that the `keys` function returns a vector of `Symbol`s (and not
```

src/abstractdataframe/iteration.jl
```diff
@@ -125,6 +133,13 @@ Base.summary(io::IO, dfcs::DataFrameColumns) = print(io, summary(dfcs))
 
 Return a `DataFrameColumns` object that is a vector-like
 that allows iterating an `AbstractDataFrame` column by column.
+
+Indexing into `DataFrameColumns` objects using integer, `Symbol` or string
+returns the corresponding column (without copying).
+Indexing into `DataFrameColumns` objects using multiple column selector
+returns a subsetted `DataFrameColumns` object with parent being a view of the
```
nalimilan, line 140:
```suggestion
Indexing into `DataFrameColumns` objects using a multiple column selector
returns a subsetted `DataFrameColumns` object with parent being a `SubDataFrame` view of the
```

test/select.jl
nalimilan: That's an interesting design issue. :-)

I see the point of creating a `SubDataFrame` parent, but it's weird that the resulting `DataFrameColumns` object has `SubArray` elements: getting different types for `x[1]` and `y[1]` when doing e.g. `x = eachcol(df); y = x[1:2]` isn't ideal. Also, if you want to access the original parent, you have to do `parent(parent(y))`, so anyway you can't use the same code for `x` and `y`. Can you think of scenarios where this would be useful?

bkamins:
So the major point of discussion is this case:
```
julia> x = Any[[1,2,3], [1,2,3]]
2-element Array{Any,1}:
 [1, 2, 3]
 [1, 2, 3]

julia> collect(x)
2-element Array{Any,1}:
 [1, 2, 3]
 [1, 2, 3]
```
As you can see - I have implemented methods to follow what happens in Base with vectors. We can change it to follow your proposal though. What do you think?

nalimilan:
OK, makes sense. Then the only remaining issue to settle is whether to create a `SubDataFrame` parent with multi-indexing.

bkamins:
You convinced me to return a `DataFrame` :). I am working on it now and on more flexible `firstindex` per your comment

src/abstractdataframe/iteration.jl
```diff
@@ -106,20 +106,28 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
 
 # Iteration by columns
 
+const DATAFRAMECOLUMNS_DOCSTR = """
+Indexing into `DataFrameColumns` objects using integer, `Symbol` or string
+returns the corresponding column (without copying).
+Indexing into `DataFrameColumns` objects using a multiple column selector
+returns a subsetted `DataFrameColumns` object with parent containing only the
+selected columns (without copying).
```
nalimilan, line 114:
```suggestion
returns a subsetted `DataFrameColumns` object with a new parent containing
only the selected columns (without copying).
```

src/abstractdataframe/iteration.jl
```diff
@@ -106,20 +106,28 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
...
+
+`DataFrameColumns` supports most of the `AbstractVector` API. The key
+differences are that it is read-only and that the `keys` function returns a
+vector of `Symbol`s (and not integers as for normal vectors).
+
+In particular `findnext`, `findprev`, `findfirst`, `findlast`, and `findall`
+functions are supported, and in `findnext`, `findprev` functions it is allowed
```
nalimilan, line 121:
```suggestion
functions are supported, and in `findnext` and `findprev` functions it is allowed
```

src/abstractdataframe/iteration.jl
```diff
@@ -106,20 +106,28 @@ Base.propertynames(itr::DataFrameRows, private::Bool=false) = propertynames(pare
...
+to pass integer, string, or `Symbol` as a reference index.
```
nalimilan, line 122:
```suggestion
to pass an integer, string, or `Symbol` as a reference index.
```

bkamins:
Thank you!

PR #2292, opened 2020-06-16
[BREAKING] stricter isless for DataFrameRow

### LINKED ISSUE #2287: isless for DataFrameRow
`isless` for `DataFrameRow` ignores column names, but `==` checks them.

For `NamedTuple` `isless` is undefined so there is no inconsistency there. But we have it for `DataFrameRow`.

We have two options I think:
1. disallow `isless` for `DataFrameRow` (I think this is the way to go)
2. change `==` to ignore column names

bkamins:
Fixes #2287.

I do not do deprecation as it is a corner case and could be considered a bug in the old implementation.

### DIFF
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -334,7 +334,7 @@ end
 
 # lexicographic ordering on DataFrame rows, missing > !missing
 function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
-    length(r1) == length(r2) ||
+    (length(r1) == length(r2) && _names(r1) == _names(r2)) ||
         throw(ArgumentError("compared DataFrameRows must have the same number " *
                             "of columns (got $(length(r1)) and $(length(r2)))"))
     for (a,b) in zip(r1, r2)
--- a/src/dataframerow/dataframerow.jl
+++ b/src/dataframerow/dataframerow.jl
@@ -334,9 +334,16 @@ end
 
 # lexicographic ordering on DataFrame rows, missing > !missing
 function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
-    (length(r1) == length(r2) && _names(r1) == _names(r2)) ||
+    length(r1) == length(r2) ||
         throw(ArgumentError("compared DataFrameRows must have the same number " *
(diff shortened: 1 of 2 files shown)


src/dataframerow/dataframerow.jl
```diff
@@ -334,7 +334,7 @@ end
 
 # lexicographic ordering on DataFrame rows, missing > !missing
 function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
-    length(r1) == length(r2) ||
+    (length(r1) == length(r2) && _names(r1) == _names(r2)) ||
         throw(ArgumentError("compared DataFrameRows must have the same number " *
```
nalimilan, line 338: Better throw a different error mentioning names?

bkamins: Good point. I have separated them.

bkamins:
here only coverage fails.

src/dataframerow/dataframerow.jl
```diff
@@ -337,6 +337,13 @@ function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
     length(r1) == length(r2) ||
         throw(ArgumentError("compared DataFrameRows must have the same number " *
                             "of columns (got $(length(r1)) and $(length(r2)))"))
+    if _names(r1) != _names(r2)
+        mismatch = findfirst(i -> _names(r1)[i] != _names(r2)[i], 1:length(r1))
+        throw(ArgumentError("compared DataFrameRows must have the same colum " *
+                            "names and they differ in column number $mismatch" *
```
nalimilan, line 343:
```suggestion
                            "names but they differ in column number $mismatch" *
```

src/dataframerow/dataframerow.jl
```diff
@@ -337,6 +337,13 @@ function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
...
+                            " where the names are :$(names(r1)[mismatch]) and " *
+                            ":$(_names(r2)[mismatch]) respecitvely"))
```
nalimilan, line 345:
```suggestion
                            ":$(_names(r2)[mismatch]) respectively"))
```

bkamins:
Thank you!

PR #2296, opened 2020-06-18
improve handling of indicator in joins

bkamins:
This PR makes the code cleaner and makes the implementation follow the docstring.

@pdeffebach - this is the change I was talking about in the PR allowing strings as `indicator`.

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -242,12 +242,10 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
     _check_consistency(df2)
 
     if indicator !== nothing
-        indicator = Symbol(indicator)
         indicator_cols = ["_left", "_right"]
         for i in 1:2
-            while (hasproperty(df1, Symbol(indicator_cols[i])) ||
-                   hasproperty(df2, Symbol(indicator_cols[i])) ||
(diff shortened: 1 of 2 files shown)


pdeffebach:
I see, so we didn't need to worry about the uniqueness of the indicator column name until later. Moving the check down lower solved the problem. Thanks for this!

src/abstractdataframe/join.jl
```diff
@@ -242,12 +242,10 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
 
     if indicator !== nothing
-        indicator = Symbol(indicator)
         indicator_cols = ["_left", "_right"]
         for i in 1:2
-            while (hasproperty(df1, Symbol(indicator_cols[i])) ||
-                   hasproperty(df2, Symbol(indicator_cols[i])) ||
-                   Symbol(indicator_cols[i]) == indicator)
+            while (hasproperty(df1, indicator_cols[i]) ||
+                   hasproperty(df2, indicator_cols[i]))
```
nalimilan, line 248:
```suggestion
            while hasproperty(df1, indicator_cols[i]) ||
                   hasproperty(df2, indicator_cols[i])
```

bkamins:
Thank you!

PR #2299, opened 2020-06-19
fixes to follow breaking changes in printing in 1.6

bkamins:
Needed because of https://github.com/JuliaLang/julia/pull/36303.

(@nalimilan - the same problem is in CategoricalArrays.jl I if I recall correctly)


nalimilan:
Could we use `nameof` instead of accessing fields directly?

bkamins:
Indeed we could. I just reused the pattern from https://github.com/JuliaLang/julia/pull/36303, but I think `nameof` is better (it even simplifies code a bit).

bkamins:
Thank you!

Issue #2305, opened 2020-06-28
`GroupKey` and `DataFrameRow` to `NamedTuple`


goretkin:
https://juliadata.github.io/DataFrames.jl/stable/lib/types/#DataFrames.DataFrameRow mentions using the `copy` function to convert a row to a `NamedTuple`, and it just calls [`NamedTuple(::DataFrameRow)`](https://github.com/JuliaData/DataFrames.jl/blob/041229123157e28885dee22545b40e66b303c54d/src/dataframerow/dataframerow.jl#L274
)

To convert a `GroupKey` into a `NamedTuple`, there is no `copy` method, but there is [`NamedTuple(key::GroupKey)`](https://github.com/JuliaData/DataFrames.jl/blob/041229123157e28885dee22545b40e66b303c54d/src/groupeddataframe/groupeddataframe.jl#L295)

This strikes me as inconsistent, but perhaps there's a reason. My instinct is that the documentation should just suggest using the `NamedTuple(::T)` method in both cases.

goretkin:
xref: https://discourse.julialang.org/t/extracting-row-of-dataframe-directly-as-namedtuple/29380/3

bkamins:
The reason of the design is that `DataFrameRow` is a view, so `copy` should de-reference it (like in other views in Base).
The documentation uses `copy` to highlight the fact that `DataFrameRow` should be essentially treated as-if it were a `NamedTuple` except that it is a view.

So there are two action points to be decided:
1. we can update the docs to recommend `NamedTuple` instead of `copy`. I understand you think it would be better?
2. Actually we could add `copy` to `GroupKey`. This is a new addition so not API is present, as such additions are considered when needed. Are there some purposes where you would you need it?

goretkin:
>  we can update the docs to recommend `NamedTuple` instead of `copy`. I understand you think it would be better?

Yes, that feels more obvious to me.

> Actually we could add copy to GroupKey. This is a new addition so not API is present, as such additions are considered when needed. Are there some purposes where you would you need it?

The main surprise for me was:

```julia
df = DataFrame(
  a = repeat([1, 2, 3, 4], outer=[2]),
  b = repeat([2, 1], outer=[4]),
  c = 1:8)

gd = groupby(df, :a)

@show haskey(gd, (a=1,))
@show in((a=1,), keys(gd))

```

producing the inconsistent
```julia
haskey(gd, (a = 1,)) = true
(a = 1,) in keys(gd) = false

[...]

PR #2308, opened 2020-06-29
[BREAKING] make GroupKey API more consistent

bkamins:
Add missing methods for `GroupKey` to make it consistent with `DataFrameRow` which is similar.

See https://github.com/JuliaData/DataFrames.jl/issues/2305 for discussion. CC @goretkin.

Also this should be made consistent with https://github.com/JuliaData/DataFrames.jl/pull/2281 whichever gets merged first. CC @pdeffebach 

After this PR and #2281 another one with `==`, `isequal`, `in` etc. definitions should be done.

I hope I have not missed any definitions - could you please check? Thank you!

### DIFF
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -245,6 +245,10 @@ equivalent `Tuple` or `NamedTuple`.
 Instances of this type are returned by `keys(::GroupedDataFrame)` and are not
 meant to be constructed directly.
 
+Indexing is one-dimensional like specifying a column of a `DataFrame`.
+You can also access the data in a `GroupKey` using the `getproperty`
+function and convert it to a `Tuple`, `NamedTuple`, or `Vector`.
+
 See [`keys(::GroupedDataFrame)`](@ref) for more information.
 """
 struct GroupKey{T<:GroupedDataFrame}
@@ -267,6 +271,7 @@ Base.haskey(key::GroupKey, idx::Symbol) = idx in groupcols(parent(key))
 Base.haskey(key::GroupKey, idx::AbstractString) = haskey(key, Symbol(idx))
 Base.haskey(key::GroupKey, idx::Union{Signed,Unsigned}) = 1 <= idx <= length(key)
 Base.values(key::GroupKey) = Tuple(_groupvalues(parent(key), getfield(key, :idx)))
+Base.IteratorEltype(::Type{<:GroupKey}) = Base.EltypeUnknown()
 Base.iterate(key::GroupKey, i::Integer=1) =
     i <= length(key) ? (key[i], i + 1) : nothing
 Base.getindex(key::GroupKey, i::Integer) =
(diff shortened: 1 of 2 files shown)


src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +245,10 @@ equivalent `Tuple` or `NamedTuple`.
 Instances of this type are returned by `keys(::GroupedDataFrame)` and are not
 meant to be constructed directly.
 
+Indexing is one-dimensional like specifying a column of a `DataFrame`.
```
nalimilan, line 248: I don't understand the "like" part.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -296,8 +301,29 @@ function Base.NamedTuple(key::GroupKey)
...
+    copy(key::GroupKey)
+
+Construct a `NamedTuple` with the same contents as the [`GroupKey`](@ref).
+"""
+Base.copy(key::GroupKey) = NamedTuple(key)
+
+Base.convert(::Type{NamedTuple}, key::GroupKey) = NamedTuple(key)
+Base.convert(::Type{Tuple}, key::GroupKey) = Tuple(key)
+
+Base.convert(::Type{Vector}, key::GroupKey) = [v for v in key]
```
nalimilan, line 315: FWIW `NamedTuple` doesn't implement these methods. Maybe they should be defined for any iterator in Base. But I guess it's OK to provide them here.

bkamins: These are implemented, because `DataFrameRow` implements them.

src/groupeddataframe/groupeddataframe.jl
bkamins: I have copied it from `DataFrameRow` docstring, so we should fix it also there. It is meant that the rules are the same as for the second dimension in data frame. How do you think we could phrase it?

nalimilan: OK. Maybe something about accepted index types?

bkamins:
@nalimilan - I have updated the docstring

bkamins:
Only coverage fails

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -245,6 +245,10 @@ equivalent `Tuple` or `NamedTuple`.
 Instances of this type are returned by `keys(::GroupedDataFrame)` and are not
 meant to be constructed directly.
 
+Indexing fields of `GroupKey` is allowed using an integer, a `Symbol`, or a string.
+It is also possible to access the data in a `GroupKey` using the `getproperty`
+function and convert it to a `Tuple`, `NamedTuple`, or `Vector`.
```
nalimilan, line 250:
```suggestion
function. A `GroupKey` can be converted to a `Tuple`, `NamedTuple`, or `Vector`.
```

bkamins:
Thank you!

PR #2311, opened 2020-06-30
fix show of UnionAll

### LINKED ISSUE #2310: Error in showing DataFrame with Distribution column
in 0.21.3 and latest master, the following code results in an error when displaying the result (e.g. in the REPL). In 0.21.2 it's still working.

```
using DataFrames
using Distributions

df = DataFrame([Distribution], [:d])
```
```
Error showing value of type DataFrame:
ERROR: TypeError: in typeassert, expected DataType, got Type{Distribution}
Stacktrace:
 [1] compacttype(::Type, ::Int64, ::Bool) at C:\sw_source\JuliaDepot\packages\DataFrames\L8ekl\src\abstractdataframe\show.jl:100
...
```
This is on
Windows 10
Julia 1.4.2
Distributions 0.23.4

Thanks for looking into it!

bkamins:
Fixes #2310

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataFrames"
 uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
-version = "0.21.3"
+version = "0.21.4"
 
 [deps]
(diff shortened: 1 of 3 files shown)


bkamins:
@nalimilan - only coverage fails.

PR #2313, opened 2020-07-06
implementation proposal of column renaming

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/1333.

I have implemented the code of the logic. It would be good to validate it. Thank you for anyone agreeing to do this.

Now the key question is about the API. What keyword argument names we should provide and what forms of passed values should they accept (just note that we need to handle renaming of more than 2 data frames - that is why internally it will be a vector, but maybe other special syntaxes should be allowed in the simpler forms).

As usual CC: @nalimilan + @oxinabox + @pdeffebach :smile:.


oxinabox:
What is the advantage of accepting a string rather a helper function for `prefixwith(x)= y-> Symbol(x, y)` ?

oxinabox:
Are `leftjoin` and `rightjoin` still pending?

bkamins:
@oxinabox  - these are exactly the things I wanted to discuss 😄.

> What is the advantage of accepting a string rather a helper function for prefixwith(x)= y-> Symbol(x, y) ?

It is just shorter to write `"_left"` han `suffixwith("_left")`. Also in this case we have to export `prefixwith` and `suffixwith` (but maybe this is not that bad?).

> If we are accepting strings in this way shall we also accept Symbol s?

We could, but do you think anyone would want to specify the suffix as a `Symbol`?

> Are `leftjoin` and `rightjoin` still pending?

They can be easily added, but they are not fully general (we need to handle joining more than two data frames anyway).
Maybe for two-argument `join`s we should have `leftjoin` and `rightjoin` and for more than two argument `join`s some other kwarg.

I felt that having one kwarg was cleanest to start with, but this is exactly the point I was not sure what would be best.

oxinabox:
> > If we are accepting strings in this way shall we also accept Symbol s?
> We could, but do you think anyone would want to specify the suffix as a Symbol?

I think so, because people have a lot of habit of working with Symbols when thinking about DataFrames.
And basically everywhere we accept strings we accept symbols.
So consistancy with that is good.

nalimilan:
Sounds good. I agree it's nice to allow passing a string to suffix for simplicity (like other implementations to). Then we can discuss providing renaming functions like `suffix_with` separately.

I would also allow passing a pair when there are only two inputs (like `on`), and support symbols for consistency as @oxinabox noted.

bkamins:
I have fully updated the PR. For now I decided not to support the cases of passing more than 2 data frames (we can do it later), as for now for 2 data frames we can use `Pair` interface only like with `on`, and later decide how to handle more than 2 data frames case (which is rare anyway I think and we need to decide how we extend `on` for this case also as currently it is not fully defined).

This PR is relatively important because it turned out that I need to rewrite the `indicator` kwarg logic to make it possible to provide column renaming functionality. In the end what this PR proposes is faster and cleaner. This in turn forced me to understand how different joins order rows, which is by far non obvious. I have added the descriptions of the rules applied in the docstrings (and I think these rules are relevant as they are contracts that we should check against when redesigning joins to make them faster - I have added tests that check if ordering follows the rules in the most important cases). Note that in particular `rightjoin` uses a by far non-obvious ordering IMO.

A review would be welcome (if you find my English unpolished - which is likely - just please suggest corrections directly). Thank you!

test/join.jl
```diff
@@ -719,4 +719,128 @@ end
...
+
+    @test_throws ArgumentError outerjoin(df1, df2, on=[:id1, :id2 => :ID2],
+                                         rename = "_left" => "_right", indicator=:id1)
+    @test_throws ArgumentError outerjoin(df1, df2, on=[:id1, :id2 => :ID2],
+                                         rename = "_left" => "_right", indicator=:x_left)
+    @test outerjoin(df1, df2, on=[:id1, :id2 => :ID2],
+                    rename = "_left" => "_right", indicator=:ind) ≅
+          DataFrame(id1=[1,2,3,4], id2=[1,2,3,4], x_left=[1,2,3,missing],
+                    x_right=[1,2,missing,3], ind=["both", "both", "left_only", "right_only"])
+end
```
nalimilan, line 844: Can you test what happens if a column with the target name already exists, and when the two renaming functions return duplicate values (either duplicating its own names, or duplicating the other function's names)? This also includes passing twice the same suffix.

bkamins:
> but you noted somewhere that performance improvements could require changing it. Do you think that's likely? Should we mention that it might change in the future?

This note was after this PR was last updated. I would say that it is almost sure we will change the order (especially that e.g. what `rightjoin` now does is not make sense from the user perspective). I will change everywhere the docstring to stating that row order is undefined.

src/abstractdataframe/join.jl
```diff
@@ -611,38 +645,47 @@ julia> job2 = DataFrame(identifier = [1, 2, 4], Job = ["Lawyer", "Doctor", "Farm
...
-              indicator = nothing, validate = (false, false))
+    rightjoin(df1, df2; on, makeunique = false, indicator = nothing,
+              validate = (false, false), rename = identity => identity)
 
 Perform a right join on two data frame objects and return a `DataFrame` containing
-the result. A right join includes all rows from `df2`.
+the result. A right join includes all rows from `df2`. The result is composed
+of two groups of rows. The first group has the order of rows in `df2` for which
+there are matching rows in `df1`. The second group also has the order of rows
+in `df2` but contains these rows that do not have matching rows in `df1`.
```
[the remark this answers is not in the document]
bkamins: It does not follow the row order of `df2`, and this is a point - see my comment below (I have written it before reading this comment) :smile:. I will say for now that the row order is undefined.

src/abstractdataframe/join.jl
```diff
@@ -555,14 +583,20 @@ the result. A left join includes all rows from `df1`.
...
 - `validate` : whether to check that columns passed as the `on` argument
-   define unique keys in each input data frame (according to `isequal`).
-   Can be a tuple or a pair, with the first element indicating whether to
-   run check for `df1` and the second element for `df2`.
-   By default no check is performed.
+  define unique keys in each input data frame (according to `isequal`).
+  Can be a tuple or a pair, with the first element indicating whether to
+  run check for `df1` and the second element for `df2`.
+  By default no check is performed.
+- `rename` : a `Pair` specifying how columns of left and right data frames should
```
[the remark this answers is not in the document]
bkamins: fixed

src/abstractdataframe/join.jl
```diff
@@ -658,14 +701,20 @@ the result. A right join includes all rows from `df2`.
...
```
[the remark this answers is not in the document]
bkamins: fixed

bkamins:
For now I defensively say that the row order is undefined everywhere, but it is likely that for some of the joins we will be able to guarantee something in the future. In particular note that the key things that might influence row order in the future are:
* threading (if we split one of the data frames into chunks per thread then the row order might be affected unless we restore is in post-processing)
* imbalanced row count of tables (this might call for using different join algoirthms that might produce different row orders)

src/abstractdataframe/join.jl
```diff
@@ -340,30 +358,27 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
 
+    left_indicator, right_indicator = nothing, nothing
     if kind == :inner
-        joined = compose_joined_table(joiner, kind,
-                                      update_row_maps!(joiner.dfl_on, joiner.dfr_on,
-                                                       group_rows(joiner.dfr_on),
-                                                       true, false, true, false)...,
-                                      makeunique=makeunique)
+        joined, left_indicator, right_indicator = compose_joined_table(joiner, kind,
+            update_row_maps!(joiner.dfl_on, joiner.dfr_on, group_rows(joiner.dfr_on),
```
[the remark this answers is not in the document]
bkamins: ok - it was indeed inconsistent.

test/join.jl
bkamins: I have added tests with and without `makeunique`.

src/abstractdataframe/join.jl
```diff
@@ -360,25 +360,30 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
     left_indicator, right_indicator = nothing, nothing
     if kind == :inner
+        inner_row_maps = update_row_maps!(joiner.dfl_on, joiner.dfr_on,
+                                          group_rows(joiner.dfr_on),
+                                          true, false, true, false)
         joined, left_indicator, right_indicator = compose_joined_table(joiner, kind,
-            update_row_maps!(joiner.dfl_on, joiner.dfr_on, group_rows(joiner.dfr_on),
-                             true, false, true, false)...,
-            makeunique, left_rename, right_rename, nothing)
+            inner_row_maps..., makeunique, left_rename, right_rename, nothing)
```
nalimilan: And then how about doing this?
```julia
joined, left_indicator, right_indicator =
    compose_joined_table(joiner, kind, inner_row_maps...,
                         makeunique, left_rename, right_rename, nothing)
```

src/abstractdataframe/join.jl
```diff
@@ -447,7 +452,8 @@ end
 
 Perform an inner join of two or more data frame objects and return a `DataFrame`
 containing the result. An inner join includes rows with keys that match in all
-passed data frames. The order of rows in the result follows `df1`.
+passed data frames.
+The order of rows in the result undefined and may change in the future releases.
```
nalimilan, line 456:
```suggestion

The order of rows in the result is undefined and may change in the future releases.
```

Same below.

src/abstractdataframe/join.jl
```diff
@@ -471,10 +477,11 @@ passed data frames. The order of rows in the result follows `df1`.
   By default no check is performed.
 - `rename` : a `Pair` specifying how columns of left and right data frames should
   be renamed in the resulting data frame. Each element of the pair can be a
-  function in which case it is applied to the column name which is passed to it
-  as a `String`; alternatively a string or a `Symbol` can be passed in which case
-  it is appended to the original column name. Note that `rename` does not affect
-  `on` columns, whose are always taken from the left data frame and left unchanged.
+  string or a `Symbol` can be passed in which case it is appended to the original
+  column name; alternatively a function can be used  in which case it is applied
+  to each column name, which is passed to it as a `String`.; Note that `rename`
```
nalimilan, line 482:
```suggestion
  column name; alternatively a function can be passed in which case it is applied
  to each column name, which is passed to it as a `String`. Note that `rename`
```
Same below.

src/abstractdataframe/join.jl
bkamins: it is better indeed. fixed

src/abstractdataframe/join.jl
```diff
@@ -555,14 +595,21 @@ the result. A left join includes all rows from `df1`.
...
-   By default no check is performed.
+  define unique keys in each input data frame (according to `isequal`).
+  Can be a tuple or a pair, with the first element indicating whether to
+  run check for `df1` and the second element for `df2`.
+  By default no check is performed.
+- `rename` : a `Pair` specifying how columns of left and right data frames should
+  be renamed in the resulting data frame. Each element of the pair can be a
+  string or a `Symbol` can be passed in which case it is appended to the original
+  column name; alternatively a function can be passed in which case it is applied
+  to each column name, which is passed to it as a `String`.; Note that `rename`
```
nalimilan, line 610: Also need to fix this occurrence and the next ones.

bkamins: What is wrong here? I thought I have fixed all (I left the `.;` after `String` in these lines which I just pushed a fix for, but s there some problem with the wording?)

nalimilan: I was referring to ;.

bkamins:
resolved conflicts with "drop deprecations" PR.

bkamins:
Thank you!

PR #2315, opened 2020-07-10
Up Documenter.jl

bkamins:
Change documenter version to latest 0.25. This is mainly needed to allow for dark theme.


bkamins:
Thank you!

PR #2320, opened 2020-07-20
fix deprecated example in the docs for innerjoin

bkamins:


bkamins:
only coverage fails here

bkamins:
@quinnj can you please have a look at it as @nalimilan is OOO and it is an easy fix. Thank you!

bkamins:
Thank you!

Issue #2211, opened 2020-04-25
Creating new columns on a `view` should fill in missings everywhere else. 


pdeffebach:
This has been kicking around in my head for a while, but a `SubDataFrame` should allow creation of new columns. 

Current behavior is:

```
julia> df = DataFrame(
               id = rand(["a", "b", "c", "d"], 100),
               a = rand(100),
               b = 100 * randn(100),
               c = rand([100, 200, 300], 100)
       );

julia> sdf = view(df, 1:50, :);

julia> sdf[:, :d] = rand(nrow(sdf))
ERROR: ArgumentError: column name :d not found in the data frame; existing most similar names are: :a, :b, :c and :id
```

I propose we add a convenience function `if`, to enable workflow as follows (with `magittr` pipes for convenience)

```
df %>%
    if(:a => ByRow(!missing)) %>%
    transform(:b => (t -> t .- mean(x)) =`> :d)
```

The `transform` function only knows about rows where `:a` is not `missing`. This replicates the behavior of Stata's `if` syntax.

bkamins:
I do not know Stata well enough. Can you show a minimal example of expected input and output please? I understand that rows that do not meet the predicate are filled with `missing` or with some definable sentinel value?

bkamins:
I was thinking about it today. If we added it would you also request this to be added to `DataFrameRow` and both for `setindex!` and broadcasting? All this is technically doable but I would be reluctant to add this functionality without a strong reason.

Therefore it would be great if you gave some use-cases justifying the need, and in particular showing that just using broadcasted `ifelse` or a comprehension to create that column would not be enough.

pdeffebach:
> I was thinking about it today. If we added it would you also request this to be added to `DataFrameRow` and both for `setindex!` and broadcasting?

I would not advocate it for DataFrameRow but I would like it for `setindex!`. 

In Stata, it is very common to add `if` clauses to the ends of commands such that you only modify a subset of rows. 

```
gen income_low = income if filled_out_income_form & income < 10 
egen income_demeaned_low = income - mean(income) if !missing(form_id)
replace income = 100 if income > 100 // only replace a subset of observations
```

The first command will copy over values in column `income` to `income_low`, but only for observations that satisfy these conditions. `missing` is filled in elsewhere. The second command's `mean` function only applies to observations that are not missing `form_id` and fills in missing values similarly. 

[...]

PR #2329, opened 2020-07-24
comparison

matthieugomez:
Add a very simple table to compare the main functions in DataFrames with dplyr and Stata.
The focus is really on the main functions.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -36,6 +36,7 @@ Pages = ["man/getting_started.md",
          "man/sorting.md",
          "man/categorical.md",
          "man/missing.md",
+         "man/comparisons.md"]
          "man/querying_frameworks.md"]
 Depth = 2
 ```
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -1,0 +1,27 @@
+The table compares the main functions of DataFrames with R dplyr and Stata
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`combine(df, :x => mean)`  | `summarize(mean(x))`    | `collapse (mean) x `|
+|`transform(df, :x => mean)`   | `mutate(mean(x))`    | `egen x_mean = mean(x)`|
+|`select(df, :x, :y)`   | `select(df, x, y)`  | `keep x y` |
+|`sort(df, :x)`   | `arrange(df, x)`    | `sort x`|
+|`filter(:x => x -> x >= 1, df)`   | `filter(df, x >= 1)`  | `keep if x >= 1` |
+
+Here are more complicated examples for `combine`/`transform`/`select`:
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`combine(df, :x => mean => :x_mean)`   | `summarize(df, x_mean = mean(x))`    | `collapse (mean) x_mean = x`|
+|`combine(df, :x => x -> maximum(x) - minimum(x)`   | `summarize(df, max(x) - min(x))`    | |
+|`combine(df, [:x, :y] => cov)`   | `summarize(df, cov(x, y))`    | |
+|`combine(df, :x => mean,  :y => sd)`   | `summarize(df, mean(x), sd(y))`    | `collapse (mean) x (sd) y` |
+|`combine(df, [:x, :y] .=> mean)`   | `summarize(df, across(c(x, y), mean))`    | `collapse (mean) x y` |
+|`combine(df, names(df, r"^x.*) .=> mean)`   | `summarize(df, across(starts_with("x"), mean))`    | `collapse (mean) x*` |
+
+As in dplyr and, the functions `select`, `transform`, `combine` can also be used on grouped dataframes:
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`transform(groupby(df, :id), :x => mean)`   | `mutate(group_by(df, id), mean(x))`    | `egen x_mean = mean(x), by(id)`|
+|`combine(groupby(df, :id), :x => mean)`  | `summarize(group_by(df, id), mean(x))`    | `collapse (mean) x, by(id)`|
+|`select(groupby(df, :id), :x => mean)`   | `transmute(group_by(df, id), mean(x))`    | |
+
+


docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
+The table compares the main functions of DataFrames with R dplyr and Stata
```
bkamins, line 1: which versions of R/dplyr and Stata? If we do it also Pandas should be added I think.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
...
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`combine(df, :x => mean)`  | `summarize(mean(x))`    | `collapse (mean) x `|
```
bkamins, line 4: I think that in dplyr functions taka a data frame as a first argument

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
...
+|`transform(df, :x => mean)`   | `mutate(mean(x))`    | `egen x_mean = mean(x)`|
+|`select(df, :x, :y)`   | `select(df, x, y)`  | `keep x y` |
+|`sort(df, :x)`   | `arrange(df, x)`    | `sort x`|
+|`filter(:x => x -> x >= 1, df)`   | `filter(df, x >= 1)`  | `keep if x >= 1` |
```
bkamins, line 8:
```suggestion
|`filter(:x => >=(1), df)`   | `filter(df, x >= 1)`  | `keep if x >= 1` |
```

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
...
+
+Here are more complicated examples for `combine`/`transform`/`select`:
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`combine(df, :x => mean => :x_mean)`   | `summarize(df, x_mean = mean(x))`    | `collapse (mean) x_mean = x`|
+|`combine(df, :x => x -> maximum(x) - minimum(x))`   | `summarize(df, max(x) - min(x))`    | |
```
bkamins, line 14: you want to say that this is not achievable in Stata at all - if this is the case it is strange (but I do not know Stata)

docs/src/man/comparisons.md
bkamins: also later I would add DataFramesMeta.jl (when @pdeffebach is done with the work he is concentrating on now)

docs/src/man/comparisons.md
bkamins: also - out of couriosity - what would be the equivalent of:
```
combine(:x => x -> (extremum=["max", "min"], value=[maximum(x), minimum(x)]), df)
```
in dplyr and Stata?

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
...
+|`combine(df, [:x, :y] => cov)`   | `summarize(df, cov(x, y))`    | |
+|`combine(df, :x => maximum,  :y => minimum)`   | `summarize(df, max(x), min(y))`    | `collapse (max) x (min) y` |
+|`combine(df, [:x, :y] .=> mean)`   | `summarize(df, across(c(x, y), mean))`    | `collapse (mean) x y` |
+|`combine(df, names(df, r"^x.*") .=> mean)`   | `summarize(df, across(starts_with("x"), mean))`    | `collapse (mean) x*` |
```
bkamins, line 18:
```suggestion
|`combine(df, names(df, r"^x") .=> mean)`   | `summarize(df, across(starts_with("x"), mean))`    | `collapse (mean) x*` |
```

docs/src/man/comparisons.md
bkamins, line 1:
```suggestion
The table compares the main functions of DataFrames.jl with R dplyr and Stata
```

bkamins: also add that `df` is a data frame here.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
...
+
+As in dplyr and, the functions `select`, `transform`, `combine` can also be used on grouped dataframes:
```
bkamins, line 20: Stata is missing somewhere here I think.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
...
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`transform(groupby(df, :id), :x => mean)`   | `mutate(group_by(df, id), mean(x))`    | `egen x_mean = mean(x), by(id)`|
+|`combine(groupby(df, :id), :x => mean)`  | `summarize(group_by(df, id), mean(x))`    | `collapse (mean) x, by(id)`|
+|`select(groupby(df, :id), :x => mean)`   | `transmute(group_by(df, id), mean(x))`    | |
+
+
```
bkamins, line 27: I would add at the end that we support `select!` and `transform!` that work in place. How does that compare to dplyr (I guess it is not possible) and  Stata (here maybe this is what happens by default?).

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,27 @@
...
+|`combine(df, [:x, :y] => cov)`   | `summarize(df, cov(x, y))`    | |
+|`combine(df, :x => maximum,  :y => minimum)`   | `summarize(df, max(x), min(y))`    | `collapse (max) x (min) y` |
+|`combine(df, [:x, :y] .=> mean)`   | `summarize(df, across(c(x, y), mean))`    | `collapse (mean) x y` |
+|`combine(df, names(df, r"^x.*") .=> mean)`   | `summarize(df, across(starts_with("x"), mean))`    | `collapse (mean) x*` |
+
+As in dplyr and, the functions `select`, `transform`, `combine` can also be used on grouped dataframes:
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`transform(groupby(df, :id), :x => mean)`   | `mutate(group_by(df, id), mean(x))`    | `egen x_mean = mean(x), by(id)`|
+|`combine(groupby(df, :id), :x => mean)`  | `summarize(group_by(df, id), mean(x))`    | `collapse (mean) x, by(id)`|
```
bkamins, line 24: I would put the functions so that `combine` is first or last, as it does a different thing than `select` and `transform`.

bkamins:
Thank you for working on it. I have left some comments.

docs/src/man/comparisons.md
matthieugomez: I understand that it is a shorthand but it seems a bit harder to understand, and less generalizable, so I'm not sure I want to change it

docs/src/man/comparisons.md
matthieugomez: In Stata, `combine` only accepts a specific set of functions. To  do more complicated things, you would have to do a combination of `transform!`and `filter!`:
```
egen out = cov(x, y)
keep if _n == 1
```

docs/src/man/comparisons.md
pdeffebach: I don't think ungrouped sumaries are very common. Better to show off `describe` from data frames and `summary` from stata?

docs/src/man/comparisons.md
pdeffebach: again, not sure how frequent `combine` without grouping is

docs/src/man/comparisons.md
matthieugomez: I do not know Panda unfortunately. Maybe someone could add it after?

docs/src/man/comparisons.md
matthieugomez: added

docs/src/man/comparisons.md
matthieugomez: Adding `groupby` for each command makes it too verbose. To clarify, I've re-arranged the table after the paragraph explaining how to use `groupby`.

matthieugomez:
I've answered the comments and added two examples of functions that return dataframes rather than vectors.

docs/src/man/comparisons.md
bkamins: OK

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,30 @@
...
+| DataFrames       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`combine(df, :x => mean => :x_mean)`   | `summarize(df, x_mean = mean(x))`    | `collapse (mean) x_mean = x`|
+|`combine(df, :x => x -> maximum(x) - minimum(x))`   | `summarize(df, max(x) - min(x))`    | |
+|`combine(df, [:x, :y] => cov)`   | `summarize(df, cov(x, y))`    | |
+|`combine(df, :x => maximum,  :y => minimum)`   | `summarize(df, max(x), min(y))`    | `collapse (max) x (min) y` |
+|`combine(df, [:x, :y] .=> mean)`   | `summarize(df, across(c(x, y), mean))`    | `collapse (mean) x y` |
+|`combine(df, names(df, r"^x") .=> mean)`   | `summarize(df, across(starts_with("x"), mean))`    | `collapse (mean) x*` |
+|`combine(:x => x -> (name = ["minimum", "maximum"], value = [minimum(x), maximum(x)]), df)`   | `summarize(df, tibble(name = c("minimum", "maximum"), value = range(x)))`    | |
+|`combine(d -> head(d, 1), df)` | `summarize(df, head(across(), 1))`||
```
bkamins: we do not have `head` function in DataFrames.jl. Just use `first` as @pdeffebach suggested. If you want more than one row use `first(d, 2)`. `first` is the function in Base for this operation.

if you want to show a general row selection use e.g. `d -> d[1:2, :]`. The only difference is that this will fail when data frame has less than two rows, and `first(d, 2)` will work.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,28 @@
...
+
+The table compares more complicated syntaxes 
+Operations| DataFrames       | dplyr| Stata|
+|:------------|:------------|:------------|:------------|
+|Several columns |`combine(df, :x => maximum,  :y => minimum)`|`summarize(df, max(x), min(y))`|`collapse (max) x (min) y`|
+||`combine(df, [:x, :y] .=> mean)`|`summarize(df, across(c(x, y), mean))`|`collapse (mean) x y`|
+||`combine(df, names(df, r"^x") .=> mean)`|`summarize(df, across(starts_with("x"), mean))`|`collapse (mean) x*`|
+|Multivariate function|`combine(df, [:x, :y] => cov)`|`summarize(df, cov(x, y))`||
+|DataFrame as output|`combine(:x => x -> (name = ["minimum", "maximum"], value = [minimum(x), maximum(x)]), df)`|`summarize(df, tibble(name = c("minimum", "maximum"), value = range(x)))`||
+|DataFrame as input|`combine(d -> head(d, 1), df)`|`summarize(df, head(across(), 1))`||
```
bkamins, line 28: we do not have `head`. use `first`.

bkamins:
Thank you for the fixes. Let us wait for @nalimilan to have a look at the PR.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,31 @@
...
+
+The table compares more complicated syntaxes 
+Operations| DataFrames       | dplyr| Stata|
+|:------------|:------------|:------------|:------------|
+|Transform several columns |`combine(df, :x => maximum,  :y => minimum)`|`summarize(df, max(x), min(y))`|`collapse (max) x (min) y`|
+||`combine(df, [:x, :y] .=> mean)`|`summarize(df, across(c(x, y), mean))`|`collapse (mean) x y`|
+||`combine(df, names(df, r"^x") .=> mean)`|`summarize(df, across(starts_with("x"), mean))`|`collapse (mean) x*`|
+|Multivariate function|`transform(df, [:x, :y] => cov)`|`mutate(df, cov(x, y))`|`egen z = corr(x y)`|
+|Row-wise|`transform(df, [:x, :y] => ByRow(min))`|`mutate(rowwise(df), min(x, y))`|`egen z = rowmin(x y)`|
+||`transform(df, AsTable(names(df, r"^x")) => ByRow(sum))`|`mutate(rowwise(df), sum(c_across(starts_with("x"))))`|`egen z = rowtotal(x*)`|
```
bkamins, line 29:
```suggestion
||`transform(df, AsTable(r"^x") => ByRow(sum))`|`mutate(rowwise(df), sum(c_across(starts_with("x"))))`|`egen z = rowtotal(x*)`|
```
or shorter:
```
transform(df, r"^x" => ByRow(+))
```

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,31 @@
+The table compares the main functions of DataFrames.jl with the R package dplyr (v1) and Stata (>=v8)
+|Operations| DataFrames.jl       | dplyr | Stata|
+|:------------|:------------|:------------|:------------|
+|Reduce multiple values|`combine(df, :x => mean)`|`summarize(df, mean(x))`|`collapse (mean) x =`|
+|Add new columns|`transform(df, :x => mean => :x_mean)`|`mutate(df, x_mean = mean(x))`|`egen x_mean = mean(x)`|
+|Pick columns|`select(df, :x, :y)`|`select(df, x, y)`|`keep x y`|
+|Pick rows |`filter(:x => x -> x >= 1, df)`|`filter(df, x >= 1)`|`keep if x >= 1`|
```
bkamins, line 7: I was thinking about it - if you want to show how to filter using a custom defined predicate in Julia then probably it should be shown in R and Stata how to define such a predicate. Otherwise in Julia you should write `>=(1)` as I have suggested earlier I think. Or maybe show both in two consecutive lines?

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,31 @@
...
+
+The table compares more complicated syntaxes 
+Operations| DataFrames       | dplyr| Stata|
+|:------------|:------------|:------------|:------------|
+|Transform several columns |`combine(df, :x => maximum,  :y => minimum)`|`summarize(df, max(x), min(y))`|`collapse (max) x (min) y`|
+||`combine(df, [:x, :y] .=> mean)`|`summarize(df, across(c(x, y), mean))`|`collapse (mean) x y`|
+||`combine(df, names(df, r"^x") .=> mean)`|`summarize(df, across(starts_with("x"), mean))`|`collapse (mean) x*`|
+|Multivariate function|`transform(df, [:x, :y] => cov)`|`mutate(df, cov(x, y))`|`egen z = corr(x y)`|
+|Row-wise|`transform(df, [:x, :y] => ByRow(min))`|`mutate(rowwise(df), min(x, y))`|`egen z = rowmin(x y)`|
+||`transform(df, AsTable(r"^x") => ByRow(sum))`|`mutate(rowwise(df), sum(c_across(starts_with("x"))))`|`egen z = rowtotal(x*)`|
```
bkamins, line 29: I do not remember if we discussed this but you can also write `r"^x" => ByRow(+)`, maybe show both options?

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,31 @@
...
+|`combine(groupby(df, :id), :x => mean)`|`summarize(group_by(df, id), mean(x))`|`collapse (mean) x, by(id)`|
+|`transform(groupby(df, :id), :x => mean)`|`mutate(group_by(df, id), mean(x))`|`egen x_mean = mean(x), by(id)`|
+|`select(groupby(df, :id), :x => mean)`|`transmute(group_by(df, id), mean(x))`||
+
+
+The table compares more complicated syntaxes 
+Operations| DataFrames       | dplyr| Stata|
+|:------------|:------------|:------------|:------------|
+|Transform several columns |`combine(df, :x => maximum,  :y => minimum)`|`summarize(df, max(x), min(y))`|`collapse (max) x (min) y`|
+||`combine(df, [:x, :y] .=> mean)`|`summarize(df, across(c(x, y), mean))`|`collapse (mean) x y`|
```
bkamins, line 25: also recently several people asked about `([:x, :y] .=> [mean std length])...`, maybe also add this?

matthieugomez: Added

docs/src/man/comparisons.md
matthieugomez: I think the one I have now is the more straightforward comparison with dplyr.

docs/src/man/comparisons.md
matthieugomez: I don't think it's super easy to understand your suggested syntax. You have to know that >=(1) creates a Function which is really not obvious. I changed the predicate to ismissing to sidestep this.

bkamins: But this is exactly the point - we want to teach readers here that there is such an option I think; otherwise we will always have a problem that they do not know this syntax, because we do not show it :smile: .
I would keep `ismissing` example as it is nice

docs/src/man/comparisons.md
bkamins: I agree this is a straightforward comparison, but again - as written above, normally I would write `r"^x" => ByRow(+)` in my code, or even `r"^x" => +` in an interactive session, as this is much convenient to write. So the question is if we want to show users here typical patterns that work (I agree there is a slight difference between `+` and `sum` so if you feel `+` is too confusing I will not insist :smile: ).

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
...
+||`combine(df, [:x, :y] => [maximum minimum])`|`summarize(df, across(c(x, y), list(max, min)))`|`collapse (max) x y (min) x y`|
```
bkamins, line 26:
```suggestion
||`combine(df, [[:x, :y]] .=> [maximum minimum])`|`summarize(df, across(c(x, y), list(max, min)))`|`collapse (max) x y (min) x y`|
```

docs/src/man/comparisons.md
bkamins: you have added something else :smile:, but I think the example is also nice

docs/src/man/comparisons.md
matthieugomez: Ok so to sidestep this I use the function arg max as an example.

docs/src/index.md
```diff
@@ -36,6 +36,7 @@ Pages = ["man/getting_started.md",
          "man/sorting.md",
          "man/categorical.md",
          "man/missing.md",
+         "man/comparisons.md"]
```
nalimilan, line 39:
```suggestion
         "man/comparisons.md",
```

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
+The table compares the main functions of DataFrames.jl with the R package dplyr (v1) and Stata (>=v8)
```
nalimilan, line 1:
```suggestion
The following table compares the main functions of DataFrames.jl with the R package dplyr (version 1) and Stata (version 8 and above)
```

Also you'll have to add a heading at the top of the file.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
...
+|Operations| DataFrames.jl       | dplyr | Stata|
+|:------------|:------------|:------------|:------------|
+|Reduce multiple values|`combine(df, :x => mean)`|`summarize(df, mean(x))`|`collapse (mean) x =`|
+|Add new columns|`transform(df, :x => mean => :x_mean)`|`mutate(df, x_mean = mean(x))`|`egen x_mean = mean(x)`|
+|Pick columns|`select(df, :x, :y)`|`select(df, x, y)`|`keep x y`|
+|Pick rows |`filter(:x => ismissing, df)`|`filter(df, is.na(x))`|`keep if missing(x)`|
+|Sort rows|`sort(df, :x)`|`arrange(df, x)`|`sort x`|
+|Rename columns|`rename(df, :x => :v)`|`rename(df, v = x)`|`rename x v`|
```
nalimilan, line 9: How about putting this after "adding new columns"? That sounds more logical.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
...
+
+These functions create new dataframes (like in dplyr). To mutate dataframes in place (like in Stata), use the suffix `!` (e.g. `transform!`, `select!`, etc)
```
nalimilan, line 11:
```suggestion
These functions create new data frames (like in dplyr). To mutate data frames in place (like in Stata), use the suffix `!` (e.g. `transform!`, `select!`, etc)
```

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
...
+
+The functions `select`, `transform`, `combine` can be applied on grouped dataframes:
```
nalimilan, line 13:
```suggestion
The functions `select`, `transform` and `combine` can be applied on grouped data frames, in which case they operate by group:
```

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
...
+| DataFrames.jl       | dplyr | Stata|
```
nalimilan, line 14: Could you add a column describing the operation like in the previous table?

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+The table compares more complicated syntaxes 
+Operations| DataFrames       | dplyr| Stata|
+|:------------|:------------|:------------|:------------|
+|Transform several columns |`combine(df, :x => maximum,  :y => minimum)`|`summarize(df, max(x), min(y))`|`collapse (max) x (min) y`|
+||`combine(df, [:x, :y] .=> mean)`|`summarize(df, across(c(x, y), mean))`|`collapse (mean) x y`|
+||`combine(df, ([:x, :y] .=> [maximum minimum])...)`|`summarize(df, across(c(x, y), list(max, min)))`|`collapse (max) x y (min) x y`|
+||`combine(df, names(df, r"^x") .=> mean)`|`summarize(df, across(starts_with("x"), mean))`|`collapse (mean) x*`|
+|Multivariate function|`transform(df, [:x, :y] => cov)`|`mutate(df, cov(x, y))`|`egen z = corr(x y)`|
```
nalimilan, line 28:
```suggestion
|Multivariate function|`transform(df, [:x, :y] => cor)`|`mutate(df, cor(x, y))`|`egen z = corr(x y)`|
```

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
...
+
+The functions `select`, `transform`, `combine` can be applied on grouped dataframes:
+| DataFrames.jl       | dplyr | Stata|
+|:------------|:------------|:------------|
+|`combine(groupby(df, :id), :x => mean)`|`summarize(group_by(df, id), mean(x))`|`collapse (mean) x, by(id)`|
+|`transform(groupby(df, :id), :x => mean)`|`mutate(group_by(df, id), mean(x))`|`egen x_mean = mean(x), by(id)`|
+|`select(groupby(df, :id), :x => mean)`|`transmute(group_by(df, id), mean(x))`||
+
+
+The table compares more complicated syntaxes
```
nalimilan, line 21:
```suggestion
Finally, the table below compares more complicated syntaxes:
```

docs/src/man/comparisons.md
nalimilan: Maybe there can be another line below showing both `:x => x -> x > 1` and `:x => >(1)` in the same cell? I agree with @bkamins that it's a good occasion to show people how to do this. If we only show `ismissing`, people won't even now how to do more complex operations.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,32 @@
+The table compares the main functions of DataFrames.jl with the R package dplyr (v1) and Stata (>=v8)
+|Operations| DataFrames.jl       | dplyr | Stata|
+|:------------|:------------|:------------|:------------|
+|Reduce multiple values|`combine(df, :x => mean)`|`summarize(df, mean(x))`|`collapse (mean) x =`|
+|Add new columns|`transform(df, :x => mean => :x_mean)`|`mutate(df, x_mean = mean(x))`|`egen x_mean = mean(x)`|
+|Pick columns|`select(df, :x, :y)`|`select(df, x, y)`|`keep x y`|
```
nalimilan, line 6: I'd also show how to compute new columns using `select`. This is `transmute` in dplyr.

matthieugomez:
Ok. I still don't get why >=(1) is so important but I put it instead of missing. I did not put a multi-rows cell: one row per verb is good.

bkamins:
Thank you. `=>(1)` is IMO just an important idiom in Julia. E.g. would never write `x -> x  >= 1` in my code (but as @nalimilan commented showing both is also OK).

Anyway - once @nalimilan is happy I think we can merge this. We can always improve the table later (especially that we need to add pandas column to it after the PR is merged)

docs/src/man/comparisons.md
```diff
@@ -1,31 +1,34 @@
...
+
+The following table compares the main functions of DataFrames.jl with the R package dplyr (version 1) and Stata (version 8 and above)
 |Operations| DataFrames.jl       | dplyr | Stata|
 |:------------|:------------|:------------|:------------|
 |Reduce multiple values|`combine(df, :x => mean)`|`summarize(df, mean(x))`|`collapse (mean) x =`|
 |Add new columns|`transform(df, :x => mean => :x_mean)`|`mutate(df, x_mean = mean(x))`|`egen x_mean = mean(x)`|
+|Rename columns|`rename(df, :x => :v)`|`rename(df, v = x)`|`rename x v`|
 |Pick columns|`select(df, :x, :y)`|`select(df, x, y)`|`keep x y`|
-|Pick rows |`filter(:x => ismissing, df)`|`filter(df, is.na(x))`|`keep if missing(x)`|
+|Pick & transform columns|`select(df, :x => mean)`|`transmute(df, mean(x), y)`||
```
nalimilan, line 10:
```suggestion
|Pick & transform columns|`select(df, :x => mean, :y)`|`transmute(df, mean(x), y)`||
```

docs/src/man/comparisons.md
```diff
@@ -1,31 +1,34 @@
...
-|:------------|:------------|:------------|
-|`combine(groupby(df, :id), :x => mean)`|`summarize(group_by(df, id), mean(x))`|`collapse (mean) x, by(id)`|
-|`transform(groupby(df, :id), :x => mean)`|`mutate(group_by(df, id), mean(x))`|`egen x_mean = mean(x), by(id)`|
-|`select(groupby(df, :id), :x => mean)`|`transmute(group_by(df, id), mean(x))`||
+The functions `select`, `transform` and `combine` can be applied on grouped data frames, in which case they operate by group:
+|Operations| DataFrames.jl       | dplyr | Stata|
+|:------------|:------------|:------------|:------------|
+|Reduce multiple values|`combine(groupby(df, :id), :x => mean)`|`summarize(group_by(df, id), mean(x))`|`collapse (mean) x, by(id)`|
+|Add new columns|`transform(groupby(df, :id), :x => mean)`|`mutate(group_by(df, id), mean(x))`|`egen x_mean = mean(x), by(id)`|
+|Pick columns|`select(groupby(df, :id), :x => mean)`|`transmute(group_by(df, id), mean(x))`||
```
nalimilan, line 21:
```suggestion
|Pick & transform columns|`select(groupby(df, :id), :x => mean)`|`transmute(group_by(df, id), mean(x))`||
```

nalimilan:
Thanks. Looks good, though I think the use of anonymous functions should appear somewhere. Since the distinction between `:x => x -> x > 1` and `:x => >(1)` applies to several syntaxes, maybe we'd better show the former for `filter` (since anonymous functions work in all cases), but explain below that `>(1)` can be used as a shorthand? That would also be useful as we show how to compute `mean` but not more complex cases where an anonymous function is required.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,35 @@
+# Comparisons with other languages
+
+The following table compares the main functions of DataFrames.jl with the R package dplyr (version 1) and Stata (version 8 and above)
```
pdeffebach, line 3:
```suggestion
The following table compares the main functions of DataFrames.jl with the R package dplyr (version 1) and Stata (version 8 and above)

```

You need a new line for the html write up to work

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,35 @@
...
+
+The functions `select`, `transform` and `combine` can be applied on grouped data frames, in which case they operate by group:
+|Operations| DataFrames.jl       | dplyr | Stata|
+|:------------|:------------|:------------|:------------|
+|Reduce multiple values|`combine(groupby(df, :id), :x => mean)`|`summarize(group_by(df, id), mean(x))`|`collapse (mean) x, by(id)`|
+|Add new columns|`transform(groupby(df, :id), :x => mean)`|`mutate(group_by(df, id), mean(x))`|`egen x_mean = mean(x), by(id)`|
+|Pick & transform columns|`select(groupby(df, :id), :x => mean)`|`transmute(group_by(df, id), mean(x))`||
+
+
+Finally, the table below compares more complicated syntaxes:
```
pdeffebach, line 24:
```suggestion
Finally, the table below compares more complicated syntaxes:

```

bkamins:
Maybe we should switch from a table to a list of lists? Not sure what would look better though

matthieugomez:
We could split the comparison into a dplyr and stata one, which would make the table less wide. Moreover, in reality people tend to be interested in either comparison, not both at the same time.

bkamins:
Good point - agreed. It will create some duplication but I do not think it is a problem (we can make separate sections for them).

matthieugomez:
Done. This led me to add a row that is Stata specific about modifying certain rows. Makes me think that maybe it would be nice to add some kind of replace function to make it easier.

bkamins:
> Makes me think that it would be nice to define some kind of replace signature to make it easier.

Indeed, especially as `ifelse.(y .>= 1, 1, x))` will fail with `missing` values (our usual scenario). Can you open an Issue please with what you think would be good (or just a call to add it so that we keep track of it - I have a remote memory that we already discussed it, but it might have been lost so it is better to have an issue for it, as it is a common operation)

pdeffebach:
I think `passmissing(ifelse).(...)` is good enough to include here.

bkamins:
Unfortunately it is incorrect. You have to write `ifelse.(.!ismissing.(y) .& y .>= 1, 1, x)`.

Using `passmissing(ifelse)` will produce `missing` in the result instead of retaining the value stored in `x`.

bkamins:
This is the reason that I am on board with efforts to make handling `missing` easier (but I am too busy with other PRs now to sit down to it seriously). What were the latest ideas to solve this issue (it was a macro if I recall correctly - right?).

or in the original code:
```
ifelse.(coalesce.(y .>= 1, false), 1, x)
```
which feels more idiomatic.

matthieugomez:
I think it’s https://github.com/JuliaData/DataFrames.jl/issues/2211

bkamins:
@nalimilan - ok to merge?

bkamins:
@nalimilan - I would merge it if you do not have additional comments.

docs/src/man/comparisons.md
```diff
@@ -0,0 +1,75 @@
...
+
+Operations| DataFrames.jl | Stata|
+|:------------|:------------|:------------|
+|Transform certain rows |`transform(df, [:x, :y] => ((x, y)-> ifelse.(y .>= 1, 1, x)) => :x)`|`replace x = 1 if y >= 1`|
+|Transform several columns |`combine(df, :x => maximum,  :y => minimum)`|`collapse (max) x (min) y`|
+||`combine(df, [:x, :y] .=> mean)`|`collapse (mean) x y`|
+||`combine(df, names(df, r"^x") .=> mean)`|`collapse (mean) x*`|
+||`combine(df, ([:x, :y] .=> [maximum minimum])...)`|`collapse (max) x y (min) x y`|
+|Multivariate function|`transform!(df, [:x, :y] => cor => :z)`|`egen z = corr(x y)`|
+|Row-wise|`transform!(df, [:x, :y] => ByRow(min) => :z)`|`egen z = rowmin(x y)`|
```
nalimilan, line 75: Maybe it would be useful to add an example of a simple row-wise function using `gen` in Stata? For example it's not obvious that `ByRow` is also useful for things like `:x => ByRow(x -> x^2) => x²`.

PR #2331, opened 2020-07-26
make colnames type stable in broadcasting

bkamins:
I noticed this due to https://github.com/JuliaLang/julia/pull/36810, but even if that is merged we should fix the comprehension in DataFrames.jl.

PR #2333, opened 2020-07-27
fix DataFrameRow setindex!

### LINKED ISSUE #2332: Single row manipulation leads to wrong results
Create a new data frame using the following codes:
```
df =  DataFrame(A = [1, 1], B = 2, C = 3, D=4)
print(df)
print(df[1, [:D, :C]])
```

Everything seemed to work well until I tried to manipulate the first row of columns D and C by running the following line:
`df[1, [:D, :C]] = [40, 30]`
or
`df[1, [4, 3]] = [40, 30]`

After this, `print(df)` would give me the following result:

```
2×4 DataFrame
│ Row │ A     │ B     │ C     │ D     │
│     │ Int64 │ Int64 │ Int64 │ Int64 │
├─────┼───────┼───────┼───────┼───────┤
│ 1   │ 40    │ 30    │ 3     │ 4     │
│ 2   │ 1     │ 2     │ 3     │ 4     │
```

You can see that what I wanted was to change the data of D and C, but I ended up having A and B changed.

I am using **DataFrames.jl v0.21.4** and **Julia v1.0.5**.

bkamins:
Fix  #2332

Also it would be great to merge https://github.com/JuliaData/DataFrames.jl/pull/2331 today so that I can make a patch release before the workshop today (it is about indexing so it would be good to avoid having bugs in this area :)).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataFrames"
 uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
-version = "0.21.4"
+version = "0.21.5"
 
 [deps]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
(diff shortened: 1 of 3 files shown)


bkamins:
Thank you!

Issue #2326, opened 2020-07-22
Core dumped, unreachable reached


pdeffebach:
This is a scary error I get from the following operation

```
## Average population of republican states
pop_party = let d = votes_bills_legislators_states
	d2 = d[end-99:end, :]

	gd = groupby(d2, "state")
	# make sure all states represented
	@assert length(gd) == 50 && all(sdf -> nrow(sdf) == 2, gd)

	d3 = @pipe d2 |>
		groupby(_, "party") |>
		combine(_, "population" => mean) # errors on this last command
end
```

I don't have an MWE available. If I try to isloate this by saving `d2` to a CSV and reading it in, this works fine. 

```
Unreachable reached at 0x7fd0ec4c000d

signal (4): Illegal instruction
in expression starting at REPL[10]:1
groupreduce at /home/peterwd/.julia/packages/DataFrames/htZzm/src/groupeddataframe/splitapplycombine.jl:1023
unknown function (ip: 0x7fd0ec4c003a)
Reduce at /home/peterwd/.julia/packages/DataFrames/htZzm/src/groupeddataframe/splitapplycombine.jl:1032
unknown function (ip: 0x7fd0ec4bfef6)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2158 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2322
_combine at /home/peterwd/.julia/packages/DataFrames/htZzm/src/groupeddataframe/splitapplycombine.jl:1150
#combine_helper#391 at /home/peterwd/.julia/packages/DataFrames/htZzm/src/groupeddataframe/splitapplycombine.jl:589
combine_helper##kw at /home/peterwd/.julia/packages/DataFrames/htZzm/src/groupeddataframe/splitapplycombine.jl:585
```

```
 pkg> st
Status `~/Documents/Projects/Senate Voting/Project.toml`
  [336ed68f] CSV v0.7.3
  [a93c6f00] DataFrames v0.21.4
  [da1fdf0e] FreqTables v0.4.0
  [d96e819e] Parameters v0.12.1
  [08abe8d2] PrettyTables v0.9.1
  [b8865327] UnicodePlots v1.2.0
```

pdeffebach:
Tagging @quinnj I think this is an issue with `SentinelArrays`

```
julia> d2.population
[...]

PR #2335, opened 2020-07-27
Don't specialize groupreduce! on result array element type

quinnj:
Works around crash seen in #2326. The inferred return type of
`groupreduce_init` is `Union{Vector{Any}, SentinelVector{Float64}}` and
it seems the compiler then crashes when trying to correctly identify `U`
from that union of types. Part of my conclusion here is based on the
fact that if you remove all other argument type constraints and just
make `groupreduce!` return `res` directly, it still crashes; thus, by
deduction, the crash has something to do with the compiler having
trouble with the `::AbstractVector{U}` argument type
constraint/specialization.

The work-around is pretty uncontroversial; we were already calling
`eltype(res)` in several other places, and I've checked that it infers
the same. I didn't add a test since this seems like such an obscure
compiler bug that it doesn't really seem necessary for DataFrames to be
testing core compiler behavior.

Also note that this bug exists in Julia <= 1.5, so current Julia master
(pending 1.6), which includes a number of compiler refactorings/changes,
seems to have resolved whatever the issue was.

cc: @Keno, @vtjnash, @JeffBezanson

### DIFF
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -962,8 +962,8 @@ function copyto_widen!(res::AbstractVector{T}, x::AbstractVector) where T
     return res
 end
 
-function groupreduce!(res::AbstractVector{U}, f, op, condf, adjust, checkempty::Bool,
-                      incol::AbstractVector{T}, gd::GroupedDataFrame) where {U,T}
+function groupreduce!(res, f, op, condf, adjust, checkempty::Bool,
+                      incol::AbstractVector{T}, gd::GroupedDataFrame) where {T}
     n = length(gd)
     if adjust !== nothing || checkempty
         counts = zeros(Int, n)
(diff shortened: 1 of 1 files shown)


src/groupeddataframe/splitapplycombine.jl
```diff
@@ -962,8 +962,8 @@ function copyto_widen!(res::AbstractVector{T}, x::AbstractVector) where T
     return res
 end
 
-function groupreduce!(res::AbstractVector{U}, f, op, condf, adjust, checkempty::Bool,
-                      incol::AbstractVector{T}, gd::GroupedDataFrame) where {U,T}
+function groupreduce!(res, f, op, condf, adjust, checkempty::Bool,
```
bkamins, line 965: Just maybe leve `res::AbstractVector` for documentation reasons. I understand it should be OK, or this also crashes?

quinnj: done

bkamins:
I understand it would be good to backport it to 0.21 branch - right? If yes I will do it.

quinnj:
Yeah, I think that's a good idea.

bkamins:
(a small note: as a rule we typically squash-merge the PRs in DataFrames.jl)

PR #2338, opened 2020-07-28
Deprecate dataframe bang

### LINKED ISSUE #2317: remove `DataFrame!`?
As far as I can tell, this doesn't mutate anything, so the `!` is not appropriate. For the cases where it gives errors, `DataFrame` could instead give the error if `copycols=false` is passed.

bkamins:
Fixes #2317


bkamins:
@nalimilan - do you have any take on this? I think it is OK to remove `DataFrames!`

src/deprecated.jl
```diff
@@ -439,3 +439,10 @@ import Base: map
 @deprecate map(f::Base.Callable, gd::GroupedDataFrame) combine(f, gd, ungroup=false)
 @deprecate map(f::Pair{<:ColumnIndex}, gd::GroupedDataFrame) combine(f, gd, ungroup=false)
 @deprecate map(f::Pair, gd::GroupedDataFrame) combine(AsTable(first(f)) => last(f), gd, ungroup=false)
+
+@deprecate DataFrame!(args...; kwargs...) begin
+    if :copycols in keys(kwargs)
```
nalimilan, line 444: Isn't this going to be printed too? I think you need to call `depwarn` manually. Or just drop this check, people will get an error anyway.

test/constructors.jl
```diff
@@ -23,13 +23,13 @@ const ≅ = isequal
...
 
-    df2 = DataFrame!(collect(Any, vecvec), Index([:x1, :x2]))
+    df2 = DataFrame(collect(Any, vecvec), Index([:x1, :x2]), copycols=false)
     @test size(df2, 1) == 3
     @test size(df2, 2) == 2
     @test df2.x1 === vecvec[1]
     @test df2.x2 === vecvec[2]
 
-    for fun in (DataFrame, DataFrame!)
+    for fun in (DataFrame, (args...; kwargs...) -> DataFrame(args...; kwargs..., copycols=false))
```
nalimilan, line 32: It would be simpler to do `for copycols in (true, false)` and pass the argument directly..

test/constructors.jl
```diff
@@ -97,9 +97,9 @@ const ≅ = isequal
     @test df[!, "x1"] == df2[!, "x1"]
     @test df[!, "x2"] == df2[!, "x2"]
 
-    @test_throws ArgumentError DataFrame!([0.0 1.0;
-                                           0.0 1.0;
-                                           0.0 1.0])
+    @test_throws MethodError DataFrame([0.0 1.0;
```
nalimilan, line 100: Why `MethodError`? :-/

test/constructors.jl
```diff
@@ -443,11 +409,11 @@ end
     @test DataFrame(a = 1:5, b = 1) == DataFrame(a = collect(1:5), b = fill(1, 5))
     @test DataFrame(a = 1, b = 1:5) == DataFrame(a = fill(1, 5), b = collect(1:5))
     @test size(DataFrame(a=1, b=[])) == (0, 2)
-    @test size(DataFrame!(a=1, b=[])) == (0, 2)
+    @test size(DataFrame(a=1, b=[], copycols=false)) == (0, 2)
 end
 
 @testset "constructor thrown exceptions" begin
-    for f in [DataFrame, DataFrame!]
+    for f in (DataFrame, (args...; kwargs...) -> DataFrame(args...; kwargs..., copycols=false))
```
nalimilan, line 416: Same here.

src/deprecated.jl
bkamins: it is going to be printed. If we remove it then we change the behavior of `DataFrame!` in some corner cases. The issue is that when someone writes `DataFrame!(..., copycols=false)` with `depwarn` you will get:
1. a warning to use `copycols=false`
2. an error that you are not allowed to pass `copycols=false`

But now as I think of it - I can put `depwarn` after the error and it will be OK. I will change it

test/constructors.jl
bkamins: right - I will change it

test/constructors.jl
bkamins: because when converting form a matrix we do not allow `copycols` kwarg. Do you think we should add it (it would be ignored).

nalimilan: Ah OK that's fine.

src/deprecated.jl
bkamins: I the end I decided to skip the error part as it threw error earlier in order to bet a cleaner deprecation.

bkamins:
should be good now.

bkamins:
only Coverage fails

bkamins:
Thank you!

Issue #2269, opened 2020-05-26
describe on columns with e.g. Integers


KristofferC:
Right now, `describe` is documented with:

> If a column's base type derives from Real, :nunique will return nothings.

The worry is that, in general. computing the number of distinct elements requires memory proportional to the length of the column.
There are perhaps a few ways we could tweak things:

- Compute the number of unique values below some threshold
- Above some threshold, use a technique like HyperLogLog to get an approximate number.

The drawbacks of thresholds are that it could perhaps be surprising to write code that works for a given dataframe size and have it suddenly stop working when you load a bigger dataframe. For techniques like approximate counting (HyperLogLog) the drawback is that user might  think that the count is always exact and allocate datastructures with a certain size based on it (which might then fail when one tries to put in all the distinct values)

Ref https://github.com/JuliaData/DataFrames.jl/pull/1435

nalimilan:
Another advantage of not special-casing `Real` is that the high cardinality problem can also happen with other types (e.g. dates and times) so it would be more robust. I think we could start with the simple solution of stopping after some threshold, and later we could implement HyperLogLog (see the old [HyperLogLog.jl](https://github.com/johnmyleswhite/HyperLogLog.jl)).

bkamins:
My recommendation would be:
* handle everything in the same way when computing `:nunique`; I do not think presenting an approximate count is a good idea - we should do an exact calculation;
* disable computing of `:median` and `:nunique` by default (leave them as opt-in)

This will dramatically improve usability of `describe`. Currently for non-toy data sets `describe` is in practice very slow because of computing `:median` and `:nunique`. If someone really wants to see these statistics they can be opt-in.

piever:
Maybe relevant: HyperLogLog is implemented in OnlineStats [here](https://joshday.github.io/OnlineStats.jl/stable/api/#OnlineStats.HyperLogLog).
On the same vein, OnlineStats also has the [P square algorithm](https://joshday.github.io/OnlineStats.jl/stable/api/#OnlineStats.P2Quantile) for approximated quantiles.

StefanKarpinski:
I have an implementation of HyperLogLog with the best new corrections to the estimator that I can contribute. I should probably put it up in a new package.

bkamins:
[...]

PR #2339, opened 2020-07-31
remove median and nunique from describe by default

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2269

As decided there we just drop computing `:median` and `:nunique` by default.
This is simplest to change, and if someone wants them it is easy to opt-in.

Just to get a relative impact on performance of dropping this consider:
```
julia> x = rand(10^8);

julia> @time mean(x)
  0.055976 seconds (1 allocation: 16 bytes)
0.5000065333980973

julia> @time median(x)
  2.046998 seconds (3 allocations: 762.940 MiB, 8.01% gc time)
0.5000076592923275

julia> @time length(unique(x))
 24.142629 seconds (112 allocations: 4.903 GiB, 0.49% gc time)
100000000
```
which when you have even several dozens of variables only starts to be prohibitive.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -457,8 +457,8 @@ where each row represents a variable and each column a summary statistic.
   Arguments can be:
     - A symbol from the list `:mean`, `:std`, `:min`, `:q25`,
       `:median`, `:q75`, `:max`, `:eltype`, `:nunique`, `:first`, `:last`, and
-      `:nmissing`. The default statistics used are `:mean`, `:min`, `:median`,
-      `:max`, `:nunique`, `:nmissing`, and `:eltype`.
+      `:nmissing`. The default statistics used are `:mean`, `:min`,
+      `:max`, `:nmissing`, and `:eltype`.
     - `:all` as the only `Symbol` argument to return all statistics.
     - A `name => function` pair where `name` is a `Symbol` or string. This will
       create a column of summary statistics with the provided name.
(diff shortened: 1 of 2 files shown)


bkamins:
(in particular note that even if we start adding support to threading in DataFrames.jl - which is a plan according to the responses in https://discourse.julialang.org/t/dataframes-jl-development-survey/44022) still `:median` and `:nunique` will be problematic because of their memory usage (but this is another issue).

bkamins:
I have also switched `unique` to `Set`, which is faster, but still prohibitive:
```
julia> @time length(Set(x));
 13.259684 seconds (14 allocations: 3.375 GiB, 1.79% gc time)
```

nalimilan:
Dropping the number of unique values sounds fine, but I'm a bit more reluctant to drop the median. It's a more robust indicator than the mean and it would be too bad not to report it by default just because in some cases it will be slow. I assume in most cases it should be fast enough.

FWIW, the carefully designed [skimr](https://github.com/ropensci/skimr) R package has an interesting approach:
```
n_missing complete_rate  mean    sd    p0   p25   p50   p75  p100 hist 
```
Minimum and maximum are reported as being the 0% and 100% percentiles, so the median is between these two. Additionally, the 25% and 75% quantiles are reported. (The rate of complete observations is somewhat redundant IMO.) What do you think?

bkamins:
The problem is (for the same size of data):
```
julia> @time quantile(x, 0:0.25:1);
 20.659685 seconds (4 allocations: 762.940 MiB, 0.42% gc time)
```
so I would not include it for performance reasons.

I will revert the `median` as you propose, however it will mean that for 10^8 rows and 100 columns (probably the maximum size normally we should consider) it adds 5 minutes to computing time of the results.

nalimilan:
0% and 100% quantiles can be computed using minimum and maximum as currently, so I wouldn't include them in the timing. Also, it turns out that `quantile(x, [0.25, 0.5, 0.75])` is twice slower than `quantile(x, 0.25) +  quantile(x, 0.5) +  quantile(x, 0.75)`, which is completely absurd and needs to be investigated. But yes, computing two more quantiles would be somewhat slower.

bkamins:
Actually I thought that computing `quantile(x, 0:0.25:1)` would be fastest and then you would not compute min and max as you would get it for free. It definitely should be investigated why this slows down.

Anyway in this PR I would focus on dropping things (and I understand we agree to drop `nunique` and keep `median` by default). Let us come back to adding things when we understand better the issues with `quantile` in Statistics. OK?

bkamins:
re-introduced `:median`

bkamins:
Thank you!

PR #2346, opened 2020-08-03
Adding support for rich display of Markdown cells

NicholasWMRitchie:
With a fairly small change to the HTML, LaTeX and text display code, it is possible to add rich text output support to cells within a DataFrame.   Markdown strings are readily constructed using the default Markdown package and the @md_str macro.   It already represents a generalized format definition language that most Julia users are familiar with.

It is already possible to place Markdown.MD cells in DataFrames.  However, the display of these cells doesn't represent the richness of which Markdown is capable.  With markdown, it is possible to output bold, italicized, headers, links and even images into DataFrame cell outputs.

For me personally, the biggest win is the ability to display functional URLs in a table.  This is particularly useful when DataFrames are used in report-type documents like generated by Weave or other literate programming packages.  I don't think there is another way to produce functional URL links in DataFrame outputs and this is often useful to detail the provenance of the data in a particular row.

The one shortcoming I've run into has been displaying LaTeX encoded in the markdown string in HTML.  While the LaTeX displays correctly outside the table, it does not seem to display correctly in the HTML output table. It seems fine in the LaTeX table.


bkamins:
Thank you for this PR. It seems OK - I will have to experiment with it before commenting.

Also since in the long term we hope to change the default backend to PrettyTables.jl - @ronisbr can you please comment how you see this change from the perspective of that package?

ronisbr:
@bkamins IIUC, the idea is to support markdown notation inside the cells. In this case, this support must be added inside PrettyTables.jl. Currently, we can add all kind of decorations to the cell, but not add formats to individual words. The biggest problem (in Text backend) is to check the printed size of the string, but it should be possible using `Markdown`.

NicholasWMRitchie:
I've submitted a pull to PrettyTables with similar code.  It really is a lot of bang for the buck - particularly for URLs and images.

src/abstractdataframe/io.jl
```diff
@@ -417,7 +423,9 @@ function printtable(io::IO,
             elseif isnothing(df[i, j])
                 print(io, nothingstring)
             else
-                if ! (etypes[j] <: Real)
+                if etypes[j] isa Markdown.MD
+                    show(io, "text/plain", df[i,j])
```
bkamins, line 427: why is this branch required here and not required in standard output to `stdout`? (I guess there is a reason but I am not sure what so please comment)

bkamins:
Looks good to me apart from the comment I have left and of course please make sure that CI passes before I do a final review.

>  it does not seem to display correctly in the HTML output table

It works for me. Are you sure you are using `Markdown` style to denote LaTeX (i.e. double backquote)?

src/abstractdataframe/io.jl
NicholasWMRitchie: It probably isn't necessary.

bkamins: Actually it looks like it matters, and `"text/plain"` should not be used I think, as it will mess up the output (I might be missing something here as I never used `Markdown` in my codes so please check):
```
julia> x = md"# **2** + *2* + ### 2"

julia> show(stdout, x) # it is just unformatted
# **2** + *2* + ### 2

julia> show(stdout, "text/plain", x) # you get a newline in the middle of the output here, which we probably do not want, but maybe it is OK and should be just properly wrapped
  2 + 2 + ### 2
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡
```

NicholasWMRitchie: You are right that `show(stdout, "text/plain", x)` is a bad choice.  
However, there is another concern.  The `show(...)` methods for Markdown add a terminating `\n`.  This needs to be stripped.  Plus while it is possible to put a multi-line Markdown string in an HTML or LaTeX table, it doesn't work too well in a text table.

Use the DataFrames license as an example:
```julia
julia> md=Markdown.parse_file("C:\\Users\\nritchie\\.julia\\dev\\DataFrames\\LICENSE.md")
julia> df=DataFrame(a=[1,2,3],b=[3.0,4.0,5.0],c=[md"Link: [BBC](http://bbc.com)",md, md"Jump ![Julia Logo](https://docs.julialang.org/en/v1/assets/logo.svg) long"])
3×3 DataFrame
│ Row │ a     │ b       │ c                                                    │
│     │ Int64 │ Float64 │ Markdown.MD                                          │
├─────┼───────┼─────────┼──────────────────────────────────────────────────────┤
│ 1   │ 1     │ 3.0     │ [BBC](http://bbc.com)                                │
│ 2   │ 2     │ 4.0     │ DataFrames.jl is licensed under the MIT License:...  │
│ 3   │ 3     │ 5.0     │ ![Julia Logo](https://docs.julialang.org/en/v1/as... │
```
or as HTML
image


To address both of these problems maybe we could use:
```julia 
function ourshow(io::IO, x::Markdown.MD)
    r=repr(x)
    f = findfirst(c->c=='\n', r)  # I don't think this will ever return nothing but...
    f = min(50, isnothing(f) ? length(f)+1 : f)
    return print(io, f<length(r) ? r[1:f-1]*"..." : r[1:f-1])
end
``` 
to "show.jl" and then using it in place of `show(io, "text/plain", x)`.  This truncates the output in a "text/plain" table but at a linebreak or in the case of a very long line at ?50? characters. 
 
However, this truncates the CSV/TSV output so we might modify
```julia
                if df[i,j] isa Markdown.MD                             # add
                    print(io, quotemark)                                  # add
                    escapedprint(io, repr(df[i,j]), quotestr)      # add
                    print(io, quotemark)                                  # add
                elseif ! (etypes[j] <: Real)
                    print(io, quotemark)
                    escapedprint(io, df[i, j], quotestr)
                    print(io, quotemark)
                else
                    print(io, df[i, j])
                end
```
to maintain the whole MD as text in the file.

bkamins: Looks good. We should to be in sync here with PrettyTables.jl so @ronisbr should comment here also.

Small comments:
* Instead of `...` better use `…` (`\ldots`)
* Instead of `isnothing(f) ? length(f)+1 : f` use `something`
* `r[1:f-1]` is incorrect, use `nextind` (otherwise you will get errors for non-ASCII indexing) and `SubString` (for speed)

To be clear what I want to say with the last point:
```
julia> s = "∀1"
"∀1"

julia> i = findfirst(==('1'), s)
4

julia> s[1:i-1]
ERROR: StringIndexError("∀1", 3)
```

NicholasWMRitchie: Yes, I am fortunate enough to rarely work with text manipulation so my UTF-8 skills are a little rusty. :smiley:

I'm going to continue to work on this as I have time today and over the next few days.  I'll create more test cases for some edge cases.  I may have identified an issue with the empty markdown string `md""`.   I'll make your suggested changes, fix the test case failures (which resulted from differences in Int on 32-bit and 64-bit systems.)  When  I'm happy with the changes, I'll push them so you can give them a look over.

bkamins: Excellent (it is a pleasure working with a contributor that "just knows" what should be done 😄).

bkamins: (actually looking at your code using `prevind` should be simpler than `nextind`)

NicholasWMRitchie:
I think that the code is ready for review.  I've resolved all the Travis errors in my new test cases.

I will note that Markdown => latex is the weakest.  There are markdown items like headers that don't translate in a well in a LaTeX tabular environment.  If we want to fix this, we'd remap the #, ##, ### etc to font sizes rather then \section{} etc but this would involve reimplemening the markdown => latex code custom for this edge case.  It could be done but...

bkamins:
>  It could be done but...

I would leave it as is

src/abstractdataframe/show.jl
```diff
@@ -54,6 +54,11 @@ ourshow(io::IO, x::Symbol) = ourshow(io, string(x))
 ourshow(io::IO, x::Nothing; styled::Bool=false) = ourshow(io, "", styled=styled)
 ourshow(io::IO, x::SHOW_TABULAR_TYPES; styled::Bool=false) =
     ourshow(io, summary(x), styled=styled)
+function ourshow(io::IO, x::Markdown.MD)
+    r = repr(x)
+    len = min(something(findfirst(c->c=='\n', r), length(r) + 1) - 1, 50)
+    return print(io, len < length(r) - 1 ? first(r, len)*"…" : first(r, len))
```
bkamins, line 60: There is still a mix-up of byte and character indexing here:
```
julia> str = "∀"^4 * "1" * "∃"
"∀∀∀∀1∃"

julia> len = findfirst(==('1'), str)
13

julia> length(str)
6

julia> first(str, len)
"∀∀∀∀1∃"

julia> first(str, length(str, 1, len))
"∀∀∀∀1"
```

To be celar:
* `length`, `first` - use character indexing
* `findfirst` uses byteindexing

To fix this store in `len` not the result of `findfirst`, but use `length(str, i, j)` function to transform it to a character index.

Also please add a test that catches this corner case.

src/abstractdataframe/show.jl
```diff
@@ -56,7 +56,7 @@ ourshow(io::IO, x::SHOW_TABULAR_TYPES; styled::Bool=false) =
     ourshow(io, summary(x), styled=styled)
 function ourshow(io::IO, x::Markdown.MD)
     r = repr(x)
-    len = min(something(findfirst(c->c=='\n', r), length(r) + 1) - 1, 50)
+    len = min(length(r, 1, something(findfirst(c->c=='\n', r), lastindex(r)+1)-1), 50)
```
bkamins, line 59: A small suggestion as `==(...)` is already in 1.0 if I remember correctly:
```suggestion
    # here we use the fact that `repr` produces `String` and `\n` is stored on 1 byte in UTF-8
    len = min(length(r, 1, something(findfirst(==('\n'), r), lastindex(r)+1)-1), 50)
```
also do you thing that 50 is reasonable? Maybe we can cut it down to e.g. 32? Not sure here - @ronisbr - any opinon here?

bkamins: @StefanKarpinski - I will write a blog post on string indexing in Base soon I think 😄, as it is really tricky which functions take byte indices and which take character indices.

NicholasWMRitchie: I'd be happy to change to 32 or ????.  I'll wait for a consensus.   I've made and tested the `==('\n')` change on 1.5.

Coming from a C/C++ background, I'm used to characters being 1 unsigned byte.  It took me a while to acknowledge that, in Julia, they are 4 bytes.  However, when you assign them to a string, all of sudden they become 1,2,3 or 4 bytes depending.  Fine.  That is just UTF-8 encoding.  Then when I compare the unsigned byte at index i in a string to `'\n'` it sort of looks like a single byte comparison.  But it isn't.  When I do 
```julia
julia> s="∫αγ∞1∫αγ∞2"
"∫αγ∞1∫αγ∞2"

julia>  findfirst(==('∞'),"∫αγ∞1∫αγ∞2")
8

julia> s[8]
'∞': Unicode U+221E (category Sm: Symbol, math)
```
I'm clearly comparing against a 4-byte item but get an index into the 1 byte array.  So `findfirst(...)` must convert each valid character to a 4-byte and then perform the comparison but pass back the 1-byte index.  All this back and forth is a little mind-bending.  Particularly, when it isn't clear from the function names which indexing mode they operate in.  To a novice, `str[i]` would seem like it should return the `i-th` 4-byte character not the character at byte index  `i` if there is one or an error otherwise.   You just have to know.  A blog post that explained the underlying logic would really help.

That's not to say that my blunder wasn't foolish. :smiley:

bkamins: `==('\n')` for sure works on 1.5 and I think it works on 1.0, I just do not remember when it what introduced. Let us just pass it through CI :). And I propose to change the default width to 32 then, as wider output in the REPL is not very useful I think.

> That's not to say that my blunder wasn't foolish. 😃

Actually what I thought was that it was not foolish, but that the documentation of strings is not explicit enough here (I know we highlight the differences in indexing in https://docs.julialang.org/en/latest/manual/strings/#Unicode-and-UTF-8-1, but we do not cover in detail how functions behave, e.g. when you read a docsting for `findfirst` it does not say what kind of index you will get).

src/abstractdataframe/io.jl
```diff
@@ -417,7 +423,12 @@ function printtable(io::IO,
             elseif isnothing(df[i, j])
                 print(io, nothingstring)
             else
-                if ! (etypes[j] <: Real)
+                if df[i,j] isa Markdown.MD
+                    print(io, quotemark)
+                    r=repr(df[i,j])
+                    escapedprint(io, first(r,max(0,length(r)-1)), quotestr)
```
bkamins, line 429: I guess this is what you want to do here - right?
```suggestion
                    escapedprint(io, chomp(r), quotestr)
```

bkamins:
@nalimilan - OK to merge?

src/abstractdataframe/io.jl
```diff
@@ -417,7 +423,12 @@ function printtable(io::IO,
             elseif isnothing(df[i, j])
                 print(io, nothingstring)
             else
-                if ! (etypes[j] <: Real)
+                if df[i,j] isa Markdown.MD
```
nalimilan, line 426: Since accessing `df[i, j]` is relatively slow, better store it in `cell = df[i, j]` and use that afterwards as in the HTML method (not sure why we didn't do this here).

src/abstractdataframe/io.jl
```diff
@@ -417,7 +423,12 @@ function printtable(io::IO,
...
+                    print(io, quotemark)
+                    r=repr(df[i,j])
```
nalimilan, line 428:
```suggestion
                    r = repr(df[i,j])
```

src/abstractdataframe/show.jl
```diff
@@ -54,6 +54,11 @@ ourshow(io::IO, x::Symbol) = ourshow(io, string(x))
 ourshow(io::IO, x::Nothing; styled::Bool=false) = ourshow(io, "", styled=styled)
 ourshow(io::IO, x::SHOW_TABULAR_TYPES; styled::Bool=false) =
     ourshow(io, summary(x), styled=styled)
+function ourshow(io::IO, x::Markdown.MD)
+    r = repr(x)
+    len = min(length(r, 1, something(findfirst(==('\n'), r), lastindex(r)+1)-1), 32)
+    return print(io, len < length(r) - 1 ? first(r, len)*"…" : first(r, len))
```
nalimilan, line 60: Using chars should give a slightly more efficient code:
```suggestion
    return print(io, len < length(r) - 1 ? first(r, len)*'…' : first(r, len))
```

test/io.jl
```diff
@@ -9,20 +9,21 @@ using Test, DataFrames, CategoricalArrays, Dates
...
-        \t\\hline
-        \t& $(Int) & String & String & Float64? & Cat…? & String\\\\
-        \t\\hline
-        \t1 & 1 & \\\$10.0 & A & 1.0 & a & \\emph{\\#undef} \\\\
-        \t2 & 2 & M\\&F & B & 2.0 & \\emph{missing} & \\emph{\\#undef} \\\\
-        \t3 & 3 & A\\textasciitilde{}B & C & \\emph{missing} & c & \\emph{\\#undef} \\\\
-        \t4 & 4 & \\textbackslash{}\\textbackslash{}alpha & S & 3.0 & d & \\emph{\\#undef} \\\\
-        \\end{tabular}
-        """
+    str =
```
nalimilan, line 15: Can you keep the triple quoted form? It's more readable. Same below.

src/abstractdataframe/show.jl
```diff
@@ -54,6 +54,11 @@ ourshow(io::IO, x::Symbol) = ourshow(io, string(x))
 ourshow(io::IO, x::Nothing; styled::Bool=false) = ourshow(io, "", styled=styled)
 ourshow(io::IO, x::SHOW_TABULAR_TYPES; styled::Bool=false) =
     ourshow(io, summary(x), styled=styled)
+function ourshow(io::IO, x::Markdown.MD)
+    r = repr(x)
+    len = min(length(r, 1, something(findfirst(==('\n'), r), lastindex(r)+1)-1), 32)
```
nalimilan, line 59: We don't abbreviate in other formats, so why do it for Markdown? Regarding newlines, I would just print them as `\n` just like we do for strings.

We need a general mechanism to abbreviate too long cells, but this should be applied systematically disregarding the column type.

test/io.jl
```diff
@@ -180,6 +202,100 @@ end
...
+                            "  * 2∫αγ∞3∫αγ∞4\n"*
+                            "  * ∫αγ∞5∫αγ\n"*
+                            "  * ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0"),
+            Markdown.parse("∫αγ∞1∫αγ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0∫α\n"*
+                            "  * γ∞1∫α\n"*
+                            "  * γ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0"),
+        ]
+    )
+    @test sprint(show, "text/plain", df) == """
+8×2 DataFrame
```
nalimilan, line 225: Add 8-characters indentation. Same below.

src/abstractdataframe/show.jl
bkamins: I would keep it here for the time being as `md"` is likely to be overlong. Then maybe just add TODO to update it when we have a general mechanism.

In general maybe we will switch to PrettyTables.jl which actually handles this already.

These changes will be possible to implement after 1.0 release if needed, as display changes are not considered to be breaking.

nalimilan: OK, let's do that. Though do you think it would make sense to apply this 32-char limit to other types as well? I guess it can be wasteful when there are only a few columns and that more space is available, but it's hard to handle. How does PrettyTables handle it?

bkamins: But how wide terminal displays do you expect people have?
I have 210 characters width in my terminal, but having a column wider than 32 chars makes it very hard to follow what is in the column anyway. So I would say the rule - no column wider than 32 characters makes sense. If someone wants to check such a wide column in detail in REPL probably extracting it to a vector and investigating is the way to go anyway.

PrettyTables.jl behaves like this:
```
julia> df = DataFrame(a="a"^20, b="b"^20)
1×2 DataFrame. Omitted printing of 1 columns
│ Row │ a                    │
│     │ String               │
├─────┼──────────────────────┤
│ 1   │ aaaaaaaaaaaaaaaaaaaa │

julia> pretty_table(df)
┌──────────────────────┬─────────────────── ⋯
│                    a │                    ⋯
│               String │               Stri ⋯
├──────────────────────┼─────────────────── ⋯
│ aaaaaaaaaaaaaaaaaaaa │ bbbbbbbbbbbbbbbbbb ⋯
└──────────────────────┴─────────────────── ⋯
```

ronisbr: Hum, currently we can set the size of a column or its minimum size. However, I think it will be nice to have an option to set the maximum size also. In this case, we can set it to 32. Thus, columns that can be displayed with less characters will be, but columns larger than this will be cropped just like this:

```julia
julia> pretty_table(df, alignment = :l, columns_width = [10,10])
┌────────────┬────────────┐
│ a          │ b          │
│ String     │ String     │
├────────────┼────────────┤
│ aaaaaaaaa… │ bbbbbbbbb… │
└────────────┴────────────┘
```

bkamins: nice

ronisbr: Done! It is in `master`:

```julia
julia> df = DataFrame(a="a"^5, b="b"^10, c="c"^15, d = "d"^20);

julia> pretty_table(df, alignment = :l, maximum_columns_width = 10)
┌────────┬────────────┬────────────┬────────────┐
│ a      │ b          │ c          │ d          │
│ String │ String     │ String     │ String     │
├────────┼────────────┼────────────┼────────────┤
│ aaaaa  │ bbbbbbbbbb │ ccccccccc… │ ddddddddd… │
└────────┴────────────┴────────────┴────────────┘
```

Off-topic: I think I just need to add 2 more features before starting to testing good configurations to print DataFrames!

nalimilan: OK, then maybe we should set a limit to 32 characters for all types.

bkamins: @NicholasWMRitchie - would you be willing to make this change everywhere in this PR or we leave it for a follow up PR (I can do it)

bkamins:
@NicholasWMRitchie - following our discussion I have written down https://bkamins.github.io/julialang/2020/08/13/strings.html. I hope you will find it useful.

bkamins:
@NicholasWMRitchie - I would like to finalize this PR for 0.22 release if you have time to work on it. Thank you!

NicholasWMRitchie:
@bkamins I've been on vacation for 2 weeks but I'll get back to this ASAP.
The String write-up is really clear and helpful.   I'll be forwarding the link to a colleague of mine who had similar questions.

bkamins:
Thank you!. In the meantime we have merge conflicts, but they should be simple to resolve :smile:.

src/abstractdataframe/io.jl
NicholasWMRitchie: Done.

src/abstractdataframe/io.jl
NicholasWMRitchie: Done

src/abstractdataframe/show.jl
NicholasWMRitchie: Done

bkamins:
I have pushed merge conflicts resolution to your branch (also a small suggestion - it is safer not to make a from a `master` branch of your fork but rather a separate branch).

src/abstractdataframe/io.jl
nalimilan: I don't see it, nor the other changes you mention below. :-)

src/abstractdataframe/show.jl
nalimilan: So what's the conclusion on this?

test/io.jl
```diff
@@ -1,28 +1,28 @@
 module TestIO
 
-using Test, DataFrames, CategoricalArrays, Dates
+using Test, DataFrames, CategoricalArrays, Dates, Markdown
 
 # Test LaTeX export
 @testset "LaTeX export" begin
-    df = DataFrame(A = 1:4,
+    df = DataFrame(A = convert.( Int64, 1:4),
```
bkamins, line 7:
```suggestion
    df = DataFrame(A = Int64.(1:4),
```

test/io.jl
```diff
@@ -130,6 +130,27 @@ end
...
+    df = DataFrame(
+        A=Int64[1,4,9,16],
+        B = [
+            md"[DataFrames.jl](http://juliadata.github.io/DataFrames.jl)",
+            md"###A",
+            md"``\frac{A}{B}``",
+            md"*A*b**A**" ]
+    )
+
+    @test repr(MIME("text/html"), df) ==
```
nalimilan, line 143: Can you also use triple-quoted strings here and below?

test/io.jl
nalimilan: This one still applies.

test/io.jl
NicholasWMRitchie: As best I can tell, Julia 1.0.5 has issues with triple-quoted strings.  When I changed some of the test cases to use TQS, Julia 1.0.5 inserts spaces in front of each line.  So on one hand I'm being asked to use TQS and on the second hand I'm being asked to indent by 8 spaces.  

The DataFrames.jl/test/io.jl:26 works on Julia 1 but not Julia 1.0 because Julia 1.0 does not strip spaces before each TQS line. (At least that what it appears....)
```
LaTeX export: Test Failed at /home/travis/build/JuliaData/DataFrames.jl/test/io.jl:26

  Expression: repr(MIME("text/latex"), df) == str

   Evaluated: "\\begin{tabular}{r|ccccccc}\n\t& A & B & C & D & E & F & G\\\\\n\t\\hline\n\t& Int64 & String & String & Float64? & Cat…? & String & MD…\\\\\n\t\\hline\n\t1 & 1 & \\\$10.0 & A & 1.0 & a & \\emph{\\#undef} & \\href{http://juliadata.github.io/DataFrames.jl}{DataFrames.jl}\n\n \\\\\n\t2 & 2 & M\\&F & B & 2.0 & \\emph{missing} & \\emph{\\#undef} & \\#\\#\\#A\n\n \\\\\n\t3 & 3 & A\\textasciitilde{}B & C & \\emph{missing} & c & \\emph{\\#undef} & \$\\frac{A}{B}\$\n\n \\\\\n\t4 & 4 & \\textbackslash{}\\textbackslash{}alpha & S & 3.0 & d & \\emph{\\#undef} & \\emph{A}b\\textbf{A}\n\n \\\\\n\\end{tabular}\n" == "        \\begin{tabular}{r|ccccccc}\n        \t& A & B & C & D & E & F & G\\\\\n        \t\\hline\n\t& Int64 & String & String & Float64? & Cat…? & String & MD…\\\\\n        \t\\hline\n        \t1 & 1 & \\\$10.0 & A & 1.0 & a & \\emph{\\#undef} & \\href{http://juliadata.github.io/DataFrames.jl}{DataFrames.jl}\n\n \\\\\n        \t2 & 2 & M\\&F & B & 2.0 & \\emph{missing} & \\emph{\\#undef} & \\#\\#\\#A\n\n \\\\\n        \t3 & 3 & A\\textasciitilde{}B & C & \\emph{missing} & c & \\emph{\\#undef} & \$\\frac{A}{B}\$\n\n \\\\\n        \t4 & 4 & \\textbackslash{}\\textbackslash{}alpha & S & 3.0 & d & \\emph{\\#undef} & \\emph{A}b\\textbf{A}\n\n \\\\\n        \\end{tabular}\n"
```

NicholasWMRitchie: Second, triple quote strings are not ideal when the text represents a single line of text (not a multi-line block).  HTML output is formatted as a single line.

nalimilan: I find this surprising as I never encountered this problem. And the existing tests passed on Julia 1.0... But I know it can be hard to get spacing right with triple quotes. I think you need to put the closing `"""` back on its own line (it makes a difference).

It's OK to keep HTML as plain strings as indeed there aren't any line breaks there.

test/io.jl
```diff
@@ -180,6 +201,100 @@ end
...
+            md"This is *very*, **very**, very, very, very, very, very, very, very long line" ,
+            md"",
+            Markdown.parse("∫αγ∞1∫αγ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0∫αγ∞1∫αγ∞2∫αγ∞3"),
+            Markdown.parse("∫αγ∞1∫αγ∞\n"*
+                            "  * 2∫αγ∞3∫αγ∞4\n"*
+                            "  * ∫αγ∞5∫αγ\n"*
+                            "  * ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0"),
+            Markdown.parse("∫αγ∞1∫αγ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0∫α\n"*
+                            "  * γ∞1∫α\n"*
+                            "  * γ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0"),
```
nalimilan, line 220:
```suggestion
                           "  * 2∫αγ∞3∫αγ∞4\n"*
                           "  * ∫αγ∞5∫αγ\n"*
                           "  * ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0"),
            Markdown.parse("∫αγ∞1∫αγ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0∫α\n"*
                           "  * γ∞1∫α\n"*
                           "  * γ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0"),
```

test/io.jl
bkamins: Handling of triple quoted strings changed in 1.5 https://github.com/JuliaLang/julia/blob/v1.5.0/NEWS.md#language-changes (but I have not checked if this affects this case - probably not)

NicholasWMRitchie: @bkamins I think it is the issue you point to with the handling of escaped characters in  <1.5 that is giving me the grief.

nalimilan: I see. Then use actual empty lines in triple-quoted strings instead of `\n`. It's more logical anyway.

NicholasWMRitchie: It's `\t` that are the problem.

nalimilan: Not AFAICT. There are already uses of `\t` in the existing test. What's new is the use of `\n`. In my tests the following works on Julia 1.0:
```julia

str = """
    \\begin{tabular}{r|ccccccc}
    \t& A & B & C & D & E & F & G\\\\
    \t\\hline
    \t& Int64 & String & String & Float64? & Cat…? & String & MD…\\\\
    \t\\hline
    \t1 & 1 & \\\$10.0 & A & 1.0 & a & \\emph{\\#undef} & \\href{http://juliadata.github.io/DataFrames.jl}{DataFrames.jl}

     \\\\
    \t2 & 2 & M\\&F & B & 2.0 & \\emph{missing} & \\emph{\\#undef} & \\#\\#\\#A

     \\\\
    \t3 & 3 & A\\textasciitilde{}B & C & \\emph{missing} & c & \\emph{\\#undef} & \$\\frac{A}{B}\$

     \\\\
    \t4 & 4 & \\textbackslash{}\\textbackslash{}alpha & S & 3.0 & d & \\emph{\\#undef} & \\emph{A}b\\textbf{A}

     \\\\
    \\end{tabular}
    """
```

NicholasWMRitchie: You are correct.

bkamins:
@nalimilan - I would merge this PR and leave a decision to limit printing to 32 everywhere to another PR, I have opened an issue https://github.com/JuliaData/DataFrames.jl/issues/2389.
@NicholasWMRitchie - can you only merge this PR with master and add an entry to NEWS.md in  "Other relevant changes" please before merging?

NEWS.md
```diff
@@ -55,3 +55,5 @@
 
 * Documentation is now available also in *Dark* mode
   ([#2315](https://github.com/JuliaData/DataFrames.jl/pull/2315))
+* add rich display support for Markdown cell entries in HTML, Markdown and LaTeX
```
bkamins, line 58: what do you mean here by ", Markdown"? (we do not test shipping out data frame in Markdown anywhere - right?)

bkamins:
unfortunately you have performed a merge in a way that confused git (now some of the changes shown are not your changes + there are some duplicates). Can you please try fixing it?

If you do not have much experience with git the easiest is to:
* **rebase** against master
* squash all commits into one commit
* make sure all is clean (as now it is not unfortunately)

bkamins:
You are still mixing-in changes from other PRs. I do not want to merge it as I am not 100% sure if it would not confuse git (since it seems it confuses it now as these changes should not be shown).

NicholasWMRitchie:
Yes, I'm struggling a bit with git.  I fear that I'm only making things worse.  I suspect that my mental model of what git is doing is flawed.

While I can rebase my master against your master, I can't squash the result so... 
I have created a new branch (MarkdownBranch) on my repo with only my changes
I've rebased the MarkdownBranch against your master
This one commit seems to contain only the changes that I made relative to your master.
I'm currently rerunning the test suite against this on 1.5.1 and 1.0.5 to make sure I haven't broken anything. 

However, this isn't on the branch that I used to submit the pull request.

bkamins:
Thank you for caring so much. `git` is sometimes confusing. When tests pass can you then just close this PR and open a new PR with that fresh branch? Thank you!

NicholasWMRitchie:
Closing and will reopen as a new PR

PR #2349, opened 2020-08-04
fix locking of GroupedDataFrame

bkamins:
This is a minor fix, but it uses a proper pattern (in some extremely rare corner cases the old code could stall because exception thrown could make the lock not getting released).

@quinnj or @nalimilan - if we could merge this quickly I would backport it to 0.21.6 release that I hope to make soon.


bkamins:
Thank you!

bkamins:
I cannot bakcport it, but it is a very corner case, so it can wait for 0.22 release.

PR #2350, opened 2020-08-04
Drop deprecations

### LINKED ISSUE #2344: Deprecations in DataFrames.jl
As in Julia 1.5 the deprecations by default will not be printed I propose to stop deprecations for:
* `push!`
* `append!`
* `sort`
* `join` and new join functions related
* `getindex`
* `setindex!`
* `setproperty!`
* `view`
* `DataFrame`
* `show`
* `showall`
* `colwise`
* `deletecols`
* `deletecols!`
* `eltypes`
* `permutecols!`
* `names!`
* `describe`
* `stackdf`
* `meltdf`
* `lastindex`
* `insertcols!`
* `eachcol`
* `groupvars`
* `aggregate`
* `deleterows!`
* `by`
* `map`

Deprecations that will be kept for one release:
* `DataFrame!`

Also I would propose to introduce a policy: if in release `X` we deprecate something we remove it in release `X+1`.

As a side benefit the DataFrames.jl will be significantly "thinned down" by removal of the deprecations.

If we agree to this I will make a PR (but before doing this I would like to have an agreement that it should be done, as it will be 1000+ deletions PR and many additions of the tests that were pending have to be made). Thank you!

bkamins:
Fixes #2344

I turn depwarns to errors where it is reasonably easy to do.


bkamins:
One conclusion I came to is that:
* functions that we defined, but do not define any more should not have custom deprecations (like `melt` or `aggregate`) - as the point of removing deprecations is to free the name space of them in particular, so that the names can be used by users
* for `getindex`/`setindex`/`setproperty` I leave `MethodError` for unsupported functionality as writing proper error messages is very complex, the change that we remove deprecation is very old and `MethodError` actually prints good suggestions (i.e. to use two indices instead of only one)

bkamins:
as a special case note that `map` for `GroupedDataFrame` CHANGES behavior (which is OK I think).

bkamins:
OK - this is good to have a look at.

bkamins:
>  a lot of these cases just turn into `throw(ArgumentError(...`

Yes, but I did not want to go too crazy with this, and left it where it was natural to leave and did not impact the code in a negative way (in some cases the rules of Julia dispatch made it difficult to preserve, and in 2 cases actually the `depwarn` was a dead code :smile:, as later we made changes that made it impossible to hit that code path).

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1395,8 +1395,8 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame};
         end
     elseif cols === :setequal || cols === :equal
         if cols === :equal
-            Base.depwarn("`cols=:equal` is deprecated." *
-                         "Use `:setequal` instead.", :vcat)
+            throw(ArgumentError("`cols=:equal` is not supported. " *
```
nalimilan, line 1398: Below you removed this check in similar cases. Better be consistent.

src/dataframe/dataframe.jl
```diff
@@ -544,13 +536,7 @@ for T in (:AbstractVector, :Not, :Colon)
             return df
         end
         x = df[!, col_ind]
-        try
-            x[row_inds] = v
-        catch
-            insert_multiple_entries!(df, v, axes(df, 1)[row_inds], col_ind)
-            Base.depwarn("implicit vector broadcasting in setindex! is deprecated; " *
```
nalimilan, line 551: Maybe we could continue throwing an explicit error in the future? What error is printed without this?

src/dataframe/dataframe.jl
```diff
@@ -1370,17 +1361,13 @@ function Base.push!(df::DataFrame, row::Union{AbstractDict, NamedTuple};
...
+    elseif cols === :setequal
         # Only check for equal lengths if :setequal is selected,
         # as an error will be thrown below if some names don't match
         if length(row) != ncols
-            Base.depwarn("In the future `push!` with `cols` equal to `:setequal`" *
-                         "will require `row` to have the same number of elements " *
-                         "as is the number of columns in `df`.", :push!)
+            throw(ArgumentError("`push!` with `cols` equal to `:setequal`" *
+                                "requires `row` to have the same number of elements " *
+                                "as is the number of columns in `df`."))
```
nalimilan, line 1370:
```suggestion
                                "as the number of columns in `df`."))
```

src/other/index.jl
```diff
@@ -114,8 +114,8 @@ Base.haskey(x::Index, key::Integer) = 1 <= key <= length(x.names)
 Base.haskey(x::Index, key::Bool) =
     throw(ArgumentError("invalid key: $key of type Bool"))
 
-# TODO: If this should stay 'unsafe', perhaps make unexported
-function Base.push!(x::Index, nm::Symbol)
+function safe_push!(x::Index, nm::Symbol)
```
nalimilan, line 117: Why not keep this a method of `Base.push!`? The fact that it checks whether the name already exists seems consistent with `Base.push!`. If anything, the old method should have been called `unsafe_push!`.

(Also I assume you're confident it doesn't affect performance too much in practice.)

src/deprecated.jl
```diff
@@ -1,443 +1,3 @@
...
-                 " if functions in `fs` have unique names.", :aggregate)
-    gd = groupby(d, cols, sort=sort, skipmissing=skipmissing)
-    df = hcat([combine(gd, valuecols(gd) .=> [f], keepkeys=i==1) for (i, f) in enumerate(fs)]...,
-                       makeunique=true)
-    return df
-end
-
-@deprecate deleterows!(df::DataFrame, inds) delete!(df, inds)
-
-@deprecate by(f::Base.Callable, d::AbstractDataFrame, cols::Any;
```
nalimilan, line 410: Could you keep a method for `by` pointing to the new syntax at least? It used the be the main function to perform grouping and it's been deprecated only recently. So it's likely to confuse many users who read old blog posts or forum answers. (`aggregate` is less essential, though I would have done the same.)

src/abstractdataframe/abstractdataframe.jl
bkamins: I removed them below, because below they were dead code - the warning there was impossible to be reached (it was a leftover of many changes that was forgotten to be removed earlier). Here the warning could be printed, so I left to follow the rule to try turning warnings into informative errors.

src/dataframe/dataframe.jl
bkamins: It is a standard error that Base prints, which I think is pretty clear in this case (as vector dimensions do not match):
```
julia> df[1:2, 1] = [1]
ERROR: DimensionMismatch("tried to assign 1 elements to 2 destinations")
```

src/deprecated.jl
bkamins: OK - I will re-introduce `by` and `aggregate` but with a general error message (we have evolving rules of how we parse things, and these old blogs would not work anyway)

src/other/index.jl
bkamins: This is the overhead:
```
julia> df = DataFrame(ones(0, 10000));

julia> @benchmark hasproperty($df, :x)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.030 ns (0.00% GC)
  median time:      6.913 ns (0.00% GC)
  mean time:        7.308 ns (0.00% GC)
  maximum time:     25.230 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999

julia> @benchmark hasproperty($df, :x1000)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.279 ns (0.00% GC)
  median time:      7.212 ns (0.00% GC)
  mean time:        7.777 ns (0.00% GC)
  maximum time:     32.975 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999
```
and I would say it is negligible, given we do `safe_push!` in a very specific context - i.e. creating a new column in a data frame (which in particular most likely will be copied anyway which is orders of magnitude slower).

bkamins: I would not keep it as `Base.push!` because we do not export `Index` so I would prefer not to pollute `push!` with this method. This `safe_push!` has a very specific and narrow use case. The original code AFAICT was designed when it was assumed that `Index` would be exported, but it is not (and I think it does not have to be).

src/dataframe/dataframe.jl
bkamins: now if you see this error and know Base, it is obvious to write `df[1:2, 1] .= [1]` i.e.:
```
julia> x = ones(3,3)
3×3 Array{Float64,2}:
 1.0  1.0  1.0
 1.0  1.0  1.0
 1.0  1.0  1.0

julia> x[1:2, 1] = [1]
ERROR: DimensionMismatch("tried to assign 1-element array to 2×1 destination")

julia> x[1:2, 1] .= [2]
2-element view(::Array{Float64,2}, 1:2, 1) with eltype Float64:
 2.0
 2.0

julia> x
3×3 Array{Float64,2}:
 2.0  1.0  1.0
 2.0  1.0  1.0
 1.0  1.0  1.0
```
so I think it is enough to be as explicit as Base with error messages.

src/other/index.jl
nalimilan: Well we don't pollute anything since it's not exported. We also define `haskey`, `merge!`, `delete!`, `empty!`, `insert!`, `getindex`, `isequal`...

src/deprecated.jl
```diff
@@ -1,3 +1,10 @@
 import Base: @deprecate
 
 @deprecate DataFrame!(args...; kwargs...) DataFrame(args...; copycols=false, kwargs...)
+
+# TODO: remove these definitions in year 2021
+by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataFrames.jl. " *
+                                             "Use the `groupby` and `combine` functions instead."))
+
+aggregate(args...; kwargs...) = throw(ArgumentError("aggregate function was removed from DataFrames.jl. " *
+                                                    "Use the `combine` functions instead."))
```
nalimilan, line 10:
```suggestion
                                                    "Use the `combine` function instead."))
```

src/deprecated.jl
```diff
@@ -1,3 +1,10 @@
 import Base: @deprecate
 
 @deprecate DataFrame!(args...; kwargs...) DataFrame(args...; copycols=false, kwargs...)
+
+# TODO: remove these definitions in year 2021
+by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataFrames.jl. " *
+                                             "Use the `groupby` and `combine` functions instead."))
```
nalimilan, line 7: This may be more explicit?
```suggestion
                                             "Use the `combine(groupby(...), ...)` or `combine(f, groupby(...))` instead."))
```

src/other/index.jl
bkamins: OK - I have reverted `push!`.

nalimilan:
What do you think of the suggestion I made above?
> Maybe it would be worth leaving a comment to errors that are only there due to deprecations so that we can easily spot and remove them at some point in the future? Without comments, it could be hard to understand why we explicitly check for some unsupported patterns when you don't know the history of the package.

bkamins:
Ah - I do not know why I have missed this. I will add them.

bkamins:
added

bkamins:
Thank you!

PR #2352, opened 2020-08-05
[WIP] Started working on fixing fastpath combine

AliMalik9599:
Tried to alter code to fix the use of fast path in combine but needs to be tested and added upon

### DIFF
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -1048,7 +1048,7 @@ function (agg::Aggregate{typeof(length)})(incol::AbstractVector, gd::GroupedData
 end
 
 isagg(p::Pair) =
-    check_aggregate(last(p)) isa AbstractAggregate && first(p) isa ColumnIndex
+    check_aggregate(last(p)) isa AbstractAggregate && (first(p) isa Union{Missing, Number} || (first(p) == maximum(p) || first(p) == minimum(p)) && first(p) isa Union{Missing, Real})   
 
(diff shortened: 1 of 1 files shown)


bkamins:
I have quickly checked your proposal, and I would like to hint that this PR requires much more work to make it work.

The major thing that now `first(p)` is column identifier not element type of the column. Therefore to make this work you would have to refactor all the code that calls `isagg`. Then the check - as I have commented above should be conditional on the value of `last(p)`.

In general - I classify this PR as hard to make (that is why we did not implement it immediately). It digs really deeply into `combine` machinery which is the most complicated part of the DataFrames.jl (of course your contribution is very welcome, but I just wanted to give you a warning that it will be quite difficult to write this PR).

If you would like to get some practice with making simpler PRs to DataFrames.jl before going back to this one the usual area where people start is some documentation improvements (as this is relatively easy to make such a contribution, you can see how the review process works, and how to read CI results).

AliMalik9599:
Hi @bkamins, yes this is by no means complete that is why I marked it work in progress. Thank you for letting me know about the difficulty of this PR, I think it may be a little out of my depth. I hope what I added can help at all!

bkamins:
Thank you for contributing. When we have this implemented I will notify you in the relevant PR so that you can see the pattern that I have described. I will keep this PR open to keep track of your proposal until either the PR is finished here or some other PR is opened for it.

Thank you again.

bkamins:
See https://github.com/JuliaData/DataFrames.jl/pull/2357 to have a look at what is needed to fix this problem. Thank you for working on it!

Issue #2325, opened 2020-07-22
public API for accessing row number of `DataFrameRow`


ExpandingMan:
Internally this is called `row` but this symbol is probably far too common to export.  `rownumber` seems like one possibility.  I feel like it should probably be something from `Base`, but I'm not sure what, there isn't already an `indexof` or anything quite like that I think.

pdeffebach:
One corner case I just thought of for this

```
julia> df = DataFrame(a = [1, 1, 2,2], b = rand(4));

julia> gd = groupby(df, "a");

julia> t = gd[2][1,:];

julia> getfield(t, :row)
3
```

bkamins:
The corner case is that currently row always points at the parent `DataFrame` not at the row of `SubDataFrame`. This is due to performance reasons (as no intermediate index calculation is required). `DataFrameRow` is even not aware of existence of `SubDataFrame`. Since we have not exposed `row` in the past it was not a problem. However, if we were to add this functionality maybe this should be changed.

bkamins:
Note that the behavior of *unrolling* to the true parent is the same as in Base:
```
julia> x = rand(3,4)
3×4 Array{Float64,2}:
 0.50024   0.728962   0.0994949  0.989262
 0.644985  0.185798   0.695228   0.260861
julia> x = reshape(1:16, 4, 4)
4×4 reshape(::UnitRange{Int64}, 4, 4) with eltype Int64:
 1  5   9  13
 2  6  10  14
 3  7  11  15
 4  8  12  16

julia> y = view(x, 2:3, 2:3)
2×2 view(reshape(::UnitRange{Int64}, 4, 4), 2:3, 2:3) with eltype Int64:
 6  10
 7  11

julia> parent(y)
4×4 reshape(::UnitRange{Int64}, 4, 4) with eltype Int64:
 1  5   9  13
 2  6  10  14
 3  7  11  15
 4  8  12  16

julia> parentindices(y)
(2:3, 2:3)

julia> z = view(y, 2:2, 2:2)
1×1 view(reshape(::UnitRange{Int64}, 4, 4), 3:3, 3:3) with eltype Int64:
 11

julia> parent(z)
4×4 reshape(::UnitRange{Int64}, 4, 4) with eltype Int64:
 1  5   9  13
 2  6  10  14
 3  7  11  15
 4  8  12  16

julia> parentindices(z)
(3:3, 3:3)
```

bkamins:
[...]

PR #2356, opened 2020-08-07
add rownumber

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2325

I have also written a docstring that should explain the differences with `parentindices`


src/dataframerow/dataframerow.jl
```diff
@@ -76,17 +76,19 @@ julia> Vector(df[1, :])
 struct DataFrameRow{D<:AbstractDataFrame,S<:AbstractIndex}
     df::D
     colindex::S
-    row::Int
+    row::Int # row number in true parent DataFrame
```
nalimilan, line 79: Maybe we can find better names to distinguish these? Even before this PR, having a field called "row" in a row object is kind of weird. :-)

src/dataframerow/dataframerow.jl
```diff
@@ -111,6 +113,82 @@ Base.@propagate_inbounds DataFrameRow(df::AbstractDataFrame, row::Integer) =
...
 row(r::DataFrameRow) = getfield(r, :row)
+
+"""
+    rownumber(dfr::DataFrameRow)
+
+Return a row number in the `AbstractDataFrame` that `dfr` was created from.
+
+Note that this differs from the fist element in a tuple retuned by
+`parentindices` which gives a row number in the `parent(dfr)` which is
+a source `DataFrame` where data that `dfr` gives access to is stored.
```
nalimilan, line 124:
```suggestion
Note that this differs from the first element in the tuple returned by
[`parentindices`](@ref). The latter gives the row number in the `parent(dfr)`, which is
the source `DataFrame` where data that `dfr` gives access to is stored.
```

src/dataframerow/dataframerow.jl
```diff
@@ -129,8 +207,12 @@ Base.@propagate_inbounds Base.getindex(df::AbstractDataFrame, rowind::Integer, :
     DataFrameRow(df, rowind, :)
 Base.@propagate_inbounds Base.getindex(r::DataFrameRow, idx::ColumnIndex) =
     parent(r)[row(r), parentcols(index(r), idx)]
-Base.@propagate_inbounds Base.getindex(r::DataFrameRow, idxs::MultiColumnIndex) =
-    DataFrameRow(parent(r), row(r), parentcols(index(r), idxs))
+
+Base.@propagate_inbounds function Base.getindex(r::DataFrameRow, idxs::MultiColumnIndex)
+    dfr_tmp = DataFrameRow(parent(r), row(r), parentcols(index(r), idxs))
```
nalimilan, line 212: Why do we have to call the constructor twice?

docs/src/lib/functions.md
```diff
@@ -33,6 +33,11 @@ rename
 rename!
 ```
 
+## Working with views
+```@docs
+    rownumber
```
nalimilan, line 38:
```suggestion
rownumber
```

docs/src/lib/functions.md
```diff
@@ -33,6 +33,11 @@ rename
 rename!
 ```
 
+## Working with views
```
nalimilan, line 36: Maybe put this in "Summary information" instead?

src/dataframerow/dataframerow.jl
bkamins: Sure - maybe `parentrow` and change `df` to `parent`?

src/dataframerow/dataframerow.jl
bkamins: This should be inlined optimized out by the compiler I think. The reason why we do it is that the first constructor is an outer constructor that recalculates the indices and properly creates `index(dfr_tmp)`.

I did not want to copy the logic here (I could) not to duplicate the code (which might be later easy to forget to fix) - but if you prefer I can do this.

The second call calls the inner constructor that "does nothing" (ie. it is only a wrapper of passed values - no checking of values takes place so this should be zero overhead and since `DataFrameRow` is immutable it should be fully optimized out)

docs/src/lib/functions.md
bkamins: ok - moved.

src/dataframerow/dataframerow.jl
nalimilan: That or just `idx`.

src/dataframerow/dataframerow.jl
nalimilan: Maybe just add a comment explaining this.

bkamins: added

src/dataframerow/dataframerow.jl
bkamins: I have changed `row` to `dfrow` and added some comments

src/dataframerow/dataframerow.jl
```diff
@@ -209,6 +213,11 @@ Base.@propagate_inbounds Base.getindex(r::DataFrameRow, idx::ColumnIndex) =
     parent(r)[row(r), parentcols(index(r), idx)]
 
 Base.@propagate_inbounds function Base.getindex(r::DataFrameRow, idxs::MultiColumnIndex)
+    # we create a temporaty DataFrameRow object to compute the SubIndex
```
nalimilan, line 216:
```suggestion
    # we create a temporary DataFrameRow object to compute the SubIndex
```

src/dataframerow/dataframerow.jl
```diff
@@ -297,6 +306,11 @@ Base.view(r::DataFrameRow, col::ColumnIndex) =
     view(parent(r)[!, parentcols(index(r), col)], row(r))
 
 function Base.view(r::DataFrameRow, cols::MultiColumnIndex)
+    # we create a temporaty DataFrameRow object to compute the SubIndex
```
nalimilan, line 309:
```suggestion
    # we create a temporary DataFrameRow object to compute the SubIndex
```

bkamins:
Thank you!

PR #2360, opened 2020-08-11
[BREAKING] Return 0 missing count for columns of nonmissing type

nilshg:
This came up on Slack - as the `eltype` is printed in the returned table, there's little additional information in returning `nothing`, as the user can see from `eltype` whether the columns allows `missing` or not. Returning 0 here makes sorting the table on this column easier.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "julia.environmentPath": "c:\\Users\\ngudat\\.julia\\dev\\DataFrames.jl"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "julia.environmentPath": "c:\\Users\\ngudat\\.julia\\dev\\DataFrames.jl"
-}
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "julia.environmentPath": "c:\\Users\\ngudat\\.julia\\dev\\DataFrames.jl"
```
bkamins, line 2: please remove this file

nilshg: Thanks - goes to show how well my switch to VSCode is going 😄
The file showed as "untracked" in VSCode but apparently still was committed...

nilshg: Have deleted!

bkamins:
As a general note:
1. I would like to highlight that we now avoid making breaking changes. Also for future readers - please make sure to add `[BREAKING]` in PR name if it is breaking
2. In this case I think it is ok to be breaking, as it is mildly breaking and should not actually break any production workflow (unless someone relied on the old behaviour, but because of this change I will not backport it)
3. Still I will wait for @nalimilan to approve before merging, as it is breaking

Thank you for the PR.

bkamins:
Thank you!

Issue #2364, opened 2020-08-12
insertcols! taking name of column to insert before


oxinabox:
in most places we allow a column name or an index.
`insertcols!` is an exception to that.

I kind of wish insertcols!  let me specify ind as the name of the column that i want to insert before.
Right now it is only the number of the column I want to insert before.
Because I am making a column that is derived from another and i want to make sure they are next to each other.

bkamins:
Have a look at https://github.com/JuliaData/DataFrames.jl/pull/2365

PR #2365, opened 2020-08-12
allow to pass a reference to where to add a column by its name

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2364

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -603,7 +603,7 @@ end
 ##############################################################################
 
 """
-    insertcols!(df::DataFrame, [ind::Int], (name=>col)::Pair...;
+    insertcols!(df::DataFrame, [ref::Int], (name=>col)::Pair...;
                 makeunique::Bool=false, copycols::Bool=true)
 
 Insert a column into a data frame in place. Return the updated `DataFrame`.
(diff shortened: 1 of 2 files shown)


src/dataframe/dataframe.jl
```diff
@@ -603,7 +603,7 @@ end
 ##############################################################################
 
 """
-    insertcols!(df::DataFrame, [ind::Int], (name=>col)::Pair...;
+    insertcols!(df::DataFrame, [ref::Int], (name=>col)::Pair...;
```
bkamins, line 606:
```suggestion
    insertcols!(df::DataFrame, [ref], (name=>col)::Pair...;
```

src/dataframe/dataframe.jl
```diff
@@ -612,7 +612,8 @@ If `ind` is omitted it is set to `ncol(df)+1`
 
 # Arguments
 - `df` : the DataFrame to which we want to add columns
-- `ind` : a position at which we want to insert a column
+- `ref` : a position at which we want to insert a column, passed as an integer
```
oxinabox, line 615: line 610 still says `ind`.
Also we might want to mention something about being inserted to the left of `ref` or that the column currently at positions `ref` is shifted to the right.

bkamins: ok - I have proposed an improvement in the docs

nalimilan:
Good catch. Though we don't use `ref` to refer to a column selector anywhere else AFAIK -- we rather say `col` or `cols`. How about using something like `prevcol`, `colbefore`, etc.?

bkamins:
OK - I have changed `ref` to `col` and `col` to `val`.
Also NEWS.md is updated.

PR #2368, opened 2020-08-14
better error for rename! and improved docstring

bkamins:
Currently rename! on `SubDataFrame` either worked or failed with `MethodError` which was not very clear. This PR clarifies this.

I opt to allow `rename!` for `SubDataFrame` with `Index`, as this is consistent with what we discuss in https://github.com/JuliaData/DataFrames.jl/issues/2211 (i.e. to make `SubDataFrame` created with `:` special).

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -226,7 +226,7 @@ Each name is changed at most once. Permutation of names is allowed.
 
 # Arguments
 - `df` : the `AbstractDataFrame`; if it is a `SubDataFrame` then renaming is
-  only allowed if view column selector was a `:`.
+  only allowed if it was created using `:` as a column selector.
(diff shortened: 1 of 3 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -225,7 +225,8 @@ Create a new data frame that is a copy of `df` with changed column names.
 Each name is changed at most once. Permutation of names is allowed.
 
 # Arguments
-- `df` : the `AbstractDataFrame`
+- `df` : the `AbstractDataFrame`; if it is a `SubDataFrame` then renaming is
+  only allowed if view column selector was a `:`.
```
nalimilan, line 229:
```suggestion
  only allowed if it was created using `:` as a column selector.
```

nalimilan (approved):

bkamins:
Thank you!

PR #2373, opened 2020-08-20
 Add isapprox support for dataframes. 

Sov-trotter:
Based on #2169 
@bkamins For now I went on with the way `Base.isapprox` handles numeric values. And used broadcasting overall. Is this the right way to go?


Sov-trotter:
Also do we wish to return `missing` for missing values?

Sov-trotter:
In this latest commit I am using the isapprox method for individual columns defined https://github.com/JuliaLang/julia/blob/e58bc723ed7b1c10da28f55385c243f6df5499cf/stdlib/LinearAlgebra/src/generic.jl#L1631-L1642 here. One thing is that `norm::Function=abs` throws error since it's nowhere broadcasted, while in the commented out code(manual implementation), we were able to broadcast the `abs` function over to the elements.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -410,6 +410,29 @@ function Base.isequal(df1::AbstractDataFrame, df2::AbstractDataFrame)
     return true
 end
 
+function Base.isapprox(df1::AbstractDataFrame, df2::AbstractDataFrame;
+                  atol::Real=0, rtol::Real=atol>0 ? 0 : √eps(),
+                  nans::Bool=false, norm::Function=abs)
```
bkamins, line 415:
```suggestion
                       atol::Real=0, rtol::Real=atol>0 ? 0 : √eps(),
                       nans::Bool=false, norm::Function=abs)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -410,6 +410,29 @@ function Base.isequal(df1::AbstractDataFrame, df2::AbstractDataFrame)
...
+    size(df1, 2) == size(df2, 2) || return false
+    isequal(index(df1), index(df2)) || return false
```
bkamins, line 417:
```suggestion
    size(df1) == size(df2) || throw(DimensionMismatch("dimensions of passed data frames do not match"))
    isequal(index(df1), index(df2)) || throw(ArgumentError("column names of passed data frames do not match"))
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -410,6 +410,29 @@ function Base.isequal(df1::AbstractDataFrame, df2::AbstractDataFrame)
...
+            isapprox(col_1, col_2; atol, rtol, nans, norm) || return false
+            # if !(col_1 isa AbstractVector{<:Union{Missing, Number}}) # should we check for both || !df2[!, idx] isa AbstractVector{<:Union{Missing, Number}} 
+            #     ismissing(col_1 == col_2) && return missing
+            #     isequal(col_1, col_2) || return false
+            # else
+            #     ismissing(col_1 == col_2) && return missing
+            #     col_1 == col_2 || norm.(col_1-col_2) <= max.(atol, rtol*max.(norm.(col_1), norm.(col_2))) || (nans && isnan.(col_1) && isnan.(col_2)) || return false
+        end
+    end
+    return true
```
bkamins, line 434:
```suggestion
    return all(isapprox.(eachcol(df1), eachcol(df2), atol=atol, rtol=rtol, nans=nans, norm=norm))
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -410,6 +410,29 @@ function Base.isequal(df1::AbstractDataFrame, df2::AbstractDataFrame)
     return true
 end
 
+function Base.isapprox(df1::AbstractDataFrame, df2::AbstractDataFrame;
```
bkamins, line 413: Can you add a docstring please?

Sov-trotter: Yeah. Will do it. :)

src/abstractdataframe/abstractdataframe.jl
bkamins: Sorry for changing my mind. I was thinking what we should do and came to the conclusion we should be defensive. I.e. just fall back to `isapprox` from Base. If in the future Base starts to be more flexible then we will become more flexible automatically. Also now we error on "wrong" input, so we always can switch to not-error in a non-breaking way.

Note that I evaluate `isapprox` eagerly on purpose - to force an error if some columns are invalid.

In general - in order to use `isapprox` now the user should select numeric only columns in a data frame when doing a comparison which I think is not that bad.

src/abstractdataframe/abstractdataframe.jl
bkamins: Thank you!

src/abstractdataframe/abstractdataframe.jl
Sov-trotter: Thanks for the suggestion! :)
But one issue that still remains is the `abs` function that we use as default in the function definition

Sov-trotter: Ah. Seems valid. I am kind of thinking towards possible improvements that can directly go into the Base method for arrays from the things we have observed here for DataFrames. Like it's abit incomplete to check the strings  only for equalty where a check for hamming distance can bring it in line with the check for Numbers.

bkamins: Ah - sorry for not commenting on this. I think we should keep `norm` as it is used in Julia Base.
This will require a bit more work as you need:
1. add LinearAlgebra to Project.toml
2. add import LinearAlgebra: norm in the DataFrames.jl file

bkamins: > that can directly go into the Base

I think you can just open an issue in Base and CC @nalimilan and me to discuss it there. As I have said - if it goes into Base then automatically we will also have it in DataFrames.jl for free. Thank you!

Sov-trotter: Sorry for pestering you this way, but it seems like the Base methods will throw an error for `missing` values. Now it's alittle tricky to handle the `missing` case with the condition `return all(isapprox.(eachcol(df1), eachcol(df2), atol=atol, rtol=rtol, nans=nans, norm=norm))`.

bkamins: Yes - Base methods will error on `missing` and this is what I think is safe to do. As if we also throw an error in this case we can change this behavior in the future to whatever we like without being breaking. When you open an issue in Base then the case of `missing` should be raised I think.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -410,6 +410,29 @@ function Base.isequal(df1::AbstractDataFrame, df2::AbstractDataFrame)
...
+                    atol::Real=0, rtol::Real=atol>0 ? 0 : √eps(),
+                    nans::Bool=false, norm::Function=abs)
```
bkamins, line 415:
```suggestion
                    nans::Bool=false, norm::Function=norm)
```
As discussed I suggest to just use the default from `LinearAlgebra`.
Please just update Project.toml and import `norm` from `LinearAlgebra`.

bkamins:
Also can you please update NEWS.md with an entry about this addition (it should be mentioned as new functionality). Thank you!

Sov-trotter:
Yeah. Will do it. Thanks for helping out. :)

Sov-trotter:
Should I add `LinearAlgebra` into the dependency changes in `News.md`?

bkamins:
No, because it is a part of a standard library and in particular for `LinearAlgebra` you do not specify version number in `[compat]`.

The idea of "Dependency changes" is to inform users of possible dependency conflicts, but in this case there is no such risk (everyone has `LinearAlgebra` installed anyway).

PR #2378, opened 2020-08-24
Add pandas comparison

tk3369:
This PR adds a new section to compare with Python pandas. 

Couple things:
1. I have fixed some other typos in the dplyr section as well.
2. I have formatted the markdown table so it's easier on eyes when editing. If desired, I can apply formatting to dplyr and stata sections as well (I'm using a vscode extension so it's literally "1-click").
3. I have kept the same use cases as copied from the dplyr section; however, I'm uncertain if this is the "best" list. Maybe we should run the list by more R/Python/Stata users and get some feedback.

_Disclaimer_: I'm not really proficient with pandas so if anyone can suggest better ways to write the sample code, please feel free to suggest below.

PR #2388, opened 2020-08-27
Update comparisons.md

eloualiche:
Add a section on the comparison with R data.table that follows the dplyr template.

### DIFF
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -41,6 +41,46 @@ Operations|dplyr| DataFrames.jl       |
 |DataFrame as output|`summarize(df, tibble(value = min(x), max(x)))`|`combine(:x => x -> (value = [minimum(x), maximum(x)]), df)`|
 
 
+
+## Comparison with the R package data.table
+
+The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.13):
+
+|Operations              | data.table  | DataFrames.jl|  
+|:-----------------------|:------------|:-------------|
+|Reduce multiple values  |`df[, list(mean(x))]`       |`combine(df, :x => mean)`|
+|Add new columns         |`df[, x_mean := mean(x) ]`  |`transform(df, :x => mean => :x_mean)`|
+|Rename columns          |`setnames(df, "x", "x_new")`|`rename(df, :x => :x_new)`|
+|Pick columns            |`df[, list(x, y)]`          |`select(df, :x, :y)`|
+|Pick & transform columns|`df[, list(mean(x), y)]`    |`select(df, :x => mean, :y)`|
+|Pick rows               |`df[ x >= 1 ]`              |`filter(:x => >=(1), df)`|
+|Sort rows               |`setorder(df, x)`           |`sort(df, :x)`|
+
+As in data.table, some of these functions can be applied by groups:
+
+|Operations              | data.table                                      | DataFrames.jl     |
+|:-----------------------|:------------------------------------------------|:------------------|
+|Reduce multiple values  |`df[, list(mean(x)), by = list(id) ]`            |`combine(grouby(df, :id), :x => mean)`|
+|Add new columns         |`df[, x_mean := mean(x), by = list(id) ]`        |`transform(grouby(df, :id), :x => mean)`|
+|Pick & transform columns|`df[, list(x_mean = mean(x), y), by = list(id) ]`|`select(grouby(df, :id), :x => mean, :y)`|
+
+
+The table below compares more advanced commands:
+
+Operations        |data.table                                                   | DataFrames.jl       | 
+|:----------------|:------------------------------------------------------------|:--------------------|
+|Complex Function          |`df[, list(mean(x, na.rm = T)) ]`                   |`combine(df, :x => x -> mean(skipmissing(x)))`|
+|Transform several columns |`df[, list(max(x), min(y)) ]`                       |`combine(df, :x => maximum,  :y => minimum)`|
+|                          |`df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`   |`combine(df, [:x, :y] .=> mean)`|
+|                          |`df[, lapply(.SD, mean), .SDcols = patterns("x*") ]`|`combine(df, names(df, r"^x") .=> mean)`|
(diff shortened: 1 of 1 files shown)


bkamins:
Thank you for the submission!

Actually this area is quite busy recently 😄.

We will work on it when @tk3369 finishes https://github.com/JuliaData/DataFrames.jl/pull/2378 (also can you please have a look at the changes that are introduced there - mostly formatting and rebase and update your PR accordingly after that PR is merged).

eloualiche:
Thanks. Makes sense. It's great to see examples from some of the different data manipulation languages/frameworks.
I did not see https://github.com/JuliaData/DataFrames.jl/pull/2378;  I will update once this has been merged.

docs/src/man/comparisons.md
```diff
@@ -41,6 +41,46 @@ Operations|dplyr| DataFrames.jl       |
...
+
+|Operations              | data.table  | DataFrames.jl|  
+|:-----------------------|:------------|:-------------|
+|Reduce multiple values  |`df[, list(mean(x))]`       |`combine(df, :x => mean)`|
+|Add new columns         |`df[, x_mean := mean(x) ]`  |`transform(df, :x => mean => :x_mean)`|
+|Rename columns          |`setnames(df, "x", "x_new")`|`rename(df, :x => :x_new)`|
+|Pick columns            |`df[, list(x, y)]`          |`select(df, :x, :y)`|
+|Pick & transform columns|`df[, list(mean(x), y)]`    |`select(df, :x => mean, :y)`|
+|Pick rows               |`df[ x >= 1 ]`              |`filter(:x => >=(1), df)`|
+|Sort rows               |`setorder(df, x)`           |`sort(df, :x)`|
```
nalimilan, line 57: `setorder` changes the order in-place:
```suggestion
|Sort rows               |`setorder(df, x)`           |`sort!(df, :x)`|
```

docs/src/man/comparisons.md
```diff
@@ -41,6 +41,46 @@ Operations|dplyr| DataFrames.jl       |
...
+
+## Comparison with the R package data.table
+
+The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.13):
+
+|Operations              | data.table  | DataFrames.jl|  
+|:-----------------------|:------------|:-------------|
+|Reduce multiple values  |`df[, list(mean(x))]`       |`combine(df, :x => mean)`|
+|Add new columns         |`df[, x_mean := mean(x) ]`  |`transform(df, :x => mean => :x_mean)`|
+|Rename columns          |`setnames(df, "x", "x_new")`|`rename(df, :x => :x_new)`|
```
nalimilan, line 53: `setnames` renames columns in-place:
```suggestion
|Rename columns          |`setnames(df, "x", "x_new")`|`rename!(df, :x => :x_new)`|
```

docs/src/man/comparisons.md
```diff
@@ -41,6 +41,46 @@ Operations|dplyr| DataFrames.jl       |
...
+|Pick & transform columns|`df[, list(mean(x), y)]`    |`select(df, :x => mean, :y)`|
+|Pick rows               |`df[ x >= 1 ]`              |`filter(:x => >=(1), df)`|
+|Sort rows               |`setorder(df, x)`           |`sort(df, :x)`|
+
+As in data.table, some of these functions can be applied by groups:
+
+|Operations              | data.table                                      | DataFrames.jl     |
+|:-----------------------|:------------------------------------------------|:------------------|
+|Reduce multiple values  |`df[, list(mean(x)), by = list(id) ]`            |`combine(grouby(df, :id), :x => mean)`|
+|Add new columns         |`df[, x_mean := mean(x), by = list(id) ]`        |`transform(grouby(df, :id), :x => mean)`|
```
nalimilan, line 64:
```suggestion
|Add new columns         |`df[, x_mean := mean(x), by = list(id) ]`        |`transform!(grouby(df, :id), :x => mean)`|
```

nalimilan (commented):
Thanks. In the meantime I've spotted a few tricky things which are specific to how data.table works. Could you (now or once #2378 has been merged) add a mention of the difference between `select` and `select!` and so on, with a short comparison with data.table "reference" semantics? That's likely to confuse users.

docs/src/man/comparisons.md
eloualiche: Thanks a lot. 
I am dipping my toes into DataFames for a project where we hope to only use one language. So I actually figured out the change in place using `!` in the last couple of days. This is great.

I am not sure what you are referring to regarding the semantics in data.table?
 I might add things such as selection based on strings: `df[, c("x", "y"), with=FALSE ]` which are a little closer to julia. 

The one feature I miss from data.table is to be able to change in place specific rows such as:
```R
df[ x %in% Set, y := newvalue ] 
```
I know its technically doable in DataFrames but I think its a little tricky using just `filter` and `transform`. In any case if we could have something like this I would find it super valuable.

nalimilan: I just meant that it would be useful to explain that in DataFrames.jl, one needs to use the functions ending in `!` to modify the data frame in-place, and that indexing using `df[...]` will never modify `df`, while in data.table `:=` does that.

I think `df[ x %in% Set, y := newvalue ]` can be written as `df[in(Set).(df.x), :y] .= newvalue` or `df[in.(df.x, Ref(Set)), :y] .= newvalue` in Julia. Right?

eloualiche: Got it. I will add a preamble specifying how `:=` in data.table stands for a change in place and generally corresponds to `!` in julia. 

For the second one, I mostly agree. The `i` in the data.table framework of `df[i, j]` is a little more powerful as it allows merges and more general filtering (though again I am not all that familiar with the DataFrames syntax). 
But for now I will add this example and perhaps raise an issue when a more precise example where the julia idiom is not obvious shows up to me.

bkamins:
@eloualiche - hi, are you still willing to work on this PR?

eloualiche:
> @eloualiche - hi, are you still willing to work on this PR?

I am still on board with this. I was waiting for some of the syntax to settle. I will have another go at it in the next week or so.

bkamins:
Sure - the syntax is settled now. Probably it will be easiest when you open a new PR (against `main`) and then close this one which is against `master`.

Thank you!

RohitRathore1:
@bkamins has it solved?

bkamins:
No - as you can see this PR is work in progress

eloualiche:
I created a new PR on main for this. Hope this is a little better. 
https://github.com/JuliaData/DataFrames.jl/pull/2725

Issue #2321, opened 2020-07-20
remove CategoricalArrays dependency, for performance


vtjnash:
Because CategoricalArrays gets loaded, due to issues such as https://github.com/JuliaData/CategoricalArrays.jl/issues/177, everything you do with DataFrames can become a lot slower. For example:

---
With CategoricalArrays:
```
julia> @time begin
              @time using Plots
              @time using DelimitedFiles
              @time using HTTP
              @time using DataFrames
              @time file = HTTP.get("https://raw.githubusercontent.com/datasets/covid-19/master/data/countries-aggregated.csv")
              @time stats = DelimitedFiles.readdlm(IOBuffer(file.body), ',', header=true)
              @time statsdf = DataFrame(stats[1], map(Symbol, vec(stats[2])))
              @time stats2 = groupby(statsdf, :Country)
              @time ukstats=stats2[179]
              @time xy = [ukstats.Confirmed,ukstats.Deaths,ukstats.Recovered]
              @time plt = plot(ukstats.Date, xy, labels=["Confirmed" "Deaths" "Recovered"])
              @time display(plt)
              end;
  4.300588 seconds (9.47 M allocations: 622.370 MiB, 5.86% gc time)
  0.329027 seconds (588.77 k allocations: 33.346 MiB, 6.31% gc time)
  0.000152 seconds (293 allocations: 18.578 KiB)
  0.253159 seconds (745.07 k allocations: 53.301 MiB, 5.43% gc time)
  3.376627 seconds (9.46 M allocations: 539.943 MiB, 9.51% gc time)
  0.529758 seconds (2.02 M allocations: 103.027 MiB, 7.79% gc time)
  0.677973 seconds (2.58 M allocations: 172.079 MiB, 6.55% gc time)
  0.961566 seconds (4.36 M allocations: 245.538 MiB, 9.20% gc time)
  0.069549 seconds (148.21 k allocations: 10.844 MiB)
  0.034701 seconds (98.86 k allocations: 5.750 MiB)
  9.646682 seconds (37.82 M allocations: 2.197 GiB, 9.66% gc time)
  3.674751 seconds (7.06 M allocations: 423.162 MiB, 3.51% gc time)
 23.888708 seconds (74.41 M allocations: 4.358 GiB, 7.71% gc time)
```
---
Without CategoricalArrays:
```
  4.287277 seconds (9.47 M allocations: 622.460 MiB, 6.18% gc time)
[...]

PR #2391, opened 2020-08-29
[BREAKING] Change `stack` to create a `PooledArray{String}` column by default

nalimilan:
This is less likely to confuse users (it's the dplyr and pandas default), and it would allow dropping the dependency on CategoricalArrays if we wanted to.
The drawback of returning strings is that the order of old columns is only reflected by the order of rows, and no longer by the sorting order of levels.

Also return a `PooledArray{Symbol}` when `variable_eltype=Symbol` is passed. That's more consistent and more efficient, e.g. if you want to group on symbols later.

Any element type is now accepted, with `CategoricalValue{String}` being handled by the general path. In practice it would allow using a custom `AbstractString` type, though that's relatively unlikely.

See https://github.com/JuliaData/DataFrames.jl/issues/1947 for a previous discussion.


tk3369:
I like this as I found myself almost always having to unwrap the column after a `stack` call.

src/abstractdataframe/reshape.jl
```diff
@@ -75,17 +76,15 @@ function stack(df::AbstractDataFrame,
...
+        catnms = PooledArray(_names(df)[ints_measure_vars])
     elseif variable_eltype === String
         catnms = PooledArray(names(df, ints_measure_vars))
     else
-        throw(ArgumentError("`variable_eltype` keyword argument accepts only " *
-                            "`CategoricalValue{String}`, `String` or `Symbol` as a value."))
+        # this covers CategoricalArray{String} in particular,
+        # as copyto! inserts levels in their order of appearance
+        nms = names(df, ints_measure_vars)
+        catnms = copyto!(similar(nms, variable_eltype), nms)
```
bkamins, line 87: This relies on `similar` for `CategoricalValue{String}` producing a `CategoricalVector`. Is this something that will be kept being provided in CategoricalArrays.jl? (I guess you intend so, but just making sure.)

src/abstractdataframe/reshape.jl
```diff
@@ -32,9 +32,10 @@ that return views into the original data frame.
 - `view` : whether the stacked data frame should be a view rather than contain
   freshly allocated vectors.
 - `variable_eltype` : determines the element type of column `variable_name`.
-  By default a categorical vector of strings is created.
-  If `variable_eltype=Symbol` it is a vector of `Symbol`,
-  and if `variable_eltype=String` a vector of `String` is produced.
+  By default a `PooledArray{String}` is created.
+  If `variable_eltype=Symbol` it is a `PoolVector{Symbol}`,
+  and if `variable_eltype=CategoricalValue{String}`
```
bkamins, line 37: I guess we are more general here. `variable_eltype` can be any type as long as `copyto!` from `Vector{String}` will work. `CategoricalValue{String}` is just one particular case.

bkamins:
> but it would also be great to break the dependency.

This is the goal and I think if we want to go for it we should clean up all places where we have it in 0.22. The major things left are:
* `categorical` function - which I think we could just drop as I am not sure how much it is used
* fast grouping rows (for `groupby` and joins) - but here we probably can delegate things to DataAPI.jl - @nalimilan should have a clear vision of the details I assume
* reshape.jl - some more minor removals (this should not be very problematic I think)

src/abstractdataframe/reshape.jl
nalimilan: That's a good question. Given improvements at https://github.com/JuliaLang/julia/pull/37163, we may be able to keep that function in the end. Anyway if CategoricalArrays stops doing that, it will be in a major release so we'll be able to adapt. But yeah if we drop that we'll have to find another approach if we want to drop the CategoricalArrays dependency...

Concretely, for this PR, I could also simply go back to explicitly creating a `CategoricalArray`. That would probably be clearer.

bkamins: Maybe this is what we can do in this PR? I would go for the "smart" path at the moment we decide do drop CategoricalArrays.jl dependency of DataFrames.jl. I will post on #data to ask about it.

bkamins:
Given https://github.com/JuliaData/DataFrames.jl/issues/2321 I would go for removing CategoricalArrays.jl here as you proposed. Then in the docstring we should say that CategoricalArrays.jl should be loaded for what is recommended to work (to make sure we do not forget to add it later).

I have also noticed that I think it makes no sense to create `PooledVector{Symbol}`, as there is no benefit over `Vector{Symbol}`, so I would change this PR to create `Vector{Symbol}` if `Symbol` is passed.

nalimilan:
> * `categorical` function - which I think we could just drop as I am not sure how much it is used

Right, see https://github.com/JuliaData/DataFrames.jl/pull/2394.

> * fast grouping rows (for `groupby` and joins) - but here we probably can delegate things to DataAPI.jl - @nalimilan should have a clear vision of the details I assume

Yes, the API is already there, I just need a way to implement it. The issue I've discovered is that if I want `refpool` to return the  `CategoricalPool`, it has to allow passing zero to `getindex` for `missing` values. But until now `CategoricalPool` had now idea of missing values: only the parent array allowed or didn't allow for missing values. So I have to add a new type parameter to `CategoricalPool`, or to have `refpool` return a different object (which unfortunately cannot be a plain `Array` due to zeros...).

> * reshape.jl - some more minor removals (this should not be very problematic I think)

Hopefully not.

> Given #2321 I would go for removing CategoricalArrays.jl here as you proposed. Then in the docstring we should say that CategoricalArrays.jl should be loaded for what is recommended to work (to make sure we do not forget to add it later).

OK I'll have a look.

> I have also noticed that I think it makes no sense to create `PooledVector{Symbol}`, as there is no benefit over `Vector{Symbol}`, so I would change this PR to create `Vector{Symbol}` if `Symbol` is passed.

Actually I think the benefit is exactly the same as for `Vector{String}`, as in both cases the array holds 64-bit values pointing to shared objects. `PooledArray` uses 32-bit refs, and allows performing grouping much more efficiently as the pool is known.

bkamins:
> Actually I think the benefit is exactly the same

It is not the same, as `String` is not interned, but `Symbol` is, but agreed that the benefit is that the pool would be known for grouping/joins, so let us leave it as is.

nalimilan:
Well at least `repeat` reuses the same string:
```julia
julia> x = repeat(["a", "b"], outer=2)
4-element Array{String,1}:
 "a"
 "b"
 "a"
 "b"

julia> x[1] === x[3]
true
```
But yeah with symbols the difference is that you're sure that two different pointers refer to two different symbols, which isn't the case for strings (that's a big problem for grouping).

bkamins:
Well :smile: :
```
julia> s1 = "abcd"^100;

julia> s2 = "abcd"^100;

julia> s1 === s2
true

julia> pointer(s1) == pointer(s2)
false
```
so you see that `===` does not do an equal pointer test. But indeed `repeat` reuses the same string, so you are right:
```
julia> x = repeat([s1], 2);

julia> pointer(x[1]) == pointer(x[2])
true
```

src/abstractdataframe/reshape.jl
nalimilan: I've improved the docstring. I've also changed the code a bit so that it produces a `PooledArray` unless the provided type returns a special array.

nalimilan:
> Given #2321 I would go for removing CategoricalArrays.jl here as you proposed. Then in the docstring we should say that CategoricalArrays.jl should be loaded for what is recommended to work (to make sure we do not forget to add it later).

Actually after looking at it I'd rather merge this and #2394 first, and only them stop reexporting CategoricalArrays: that can't be done before merging either of these, and that's a logically independent step.

src/abstractdataframe/reshape.jl
```diff
@@ -100,17 +103,15 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
-    if variable_eltype <: CategoricalValue{String}
-        nms = names(df, measure_vars)
-        catnms = categorical(nms)
-        levels!(catnms, nms)
-    elseif variable_eltype <: Symbol
-        catnms = _names(df)[measure_vars]
-    elseif variable_eltype <: String
-        catnms = names(df, measure_vars)
+    if variable_eltype === Symbol
+        catnms = PooledArray(_names(df)[measure_vars])
```
bkamins, line 107: there is probably not much benefit from it here but we can have it for consistency.

src/abstractdataframe/reshape.jl
bkamins: maybe add to a docstring that this is for `view=false` as for `view=true` we have a `RepeatedVector`.

bkamins (approved):

bkamins:
I have left two minor comments - can you please merge this PR when you decide on them. Thank you!

src/abstractdataframe/reshape.jl
nalimilan: Good catch. Yeah that's probably counter-productive here so I've reverted the change. What could make sense is to define `copy(::RepeatedVector)` to return a `PooledArray`, but that's non-breaking so we can do that later.

src/abstractdataframe/reshape.jl
nalimilan: Indeed.

bkamins:
Looks good to merge after the CI passes

PR #2394, opened 2020-08-30
Deprecate `categorical` and `categorical!`

nalimilan:
These methods are the only remaining traces of CategoricalArrays in the public API.
Getting rid of them will allow dropping the dependency if we want.

`transform` and `transform!` now provide quite reasonable replacements for some uses.
What is lacking though is a way to select columns based on their types, which would be a useful addition in general. Also it could make sense to add a keyword argument to keep original names, which would be convenient to avoid repeating the selector.


bkamins:
> What is lacking though is a way to select columns based on their types, which would be a useful addition in general.

It should be relatively easy to add it everywhere. I will open an issue to keep track of it and later make a PR (which will not be complex I assume, but requires checking many places).

> add a keyword argument to keep original names, which would be convenient to avoid repeating the selector.

Similarly - it should be relatively easy, I will open an issue and make a PR. I would call it `suffix` and decide if we want to suffix a transformation to the column name (this is more general, as it will cover also the case when you pass multiple columns as an input)

bkamins:
Looks good (but let me sleep over it before merging 😄)

src/deprecated.jl
```diff
@@ -8,3 +8,102 @@ by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataF
...
+                                       compress::Bool=false)
+    if compress === nothing
+        compress = false
+        categoricalstr = "categorical"
+    else
+        categoricalstr = "categorical(x, compress=$compress)"
+    end
+    if cols === nothing
+        cols = Union{AbstractString, Missing}
+        Base.depwarn("`categorical(df, cols)` is deprecated. " *
```
bkamins, line 51:
```suggestion
        Base.depwarn("`categorical(df)` is deprecated. " *
```

src/deprecated.jl
```diff
@@ -8,3 +8,102 @@ by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataF
...
+                     "Use `cols = names(df)[map(c -> eltype(c) <: $cols, eachcol(df))]; transform(df, cols .=> $categoricalstr .=> cols)` instead.",
+                     :categorical)
+    else
+        Base.depwarn("`categorical(df, cols)` is deprecated. " *
+                     "Use `cols = names(df)[map(c -> eltype(c) <: cols, eachcol(df))]; transform(df, cols .=> $categoricalstr .=> cols)` instead.",
```
bkamins, line 56: `cols` variable is used twice here (overwriting the original `cols` vale) so it is probably better to change the name here.

src/deprecated.jl
```diff
@@ -8,3 +8,102 @@ by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataF
...
+                      compress::Bool=false)
+    if compress === nothing
+        compress = false
+        categoricalstr = "categorical"
+    else
+        categoricalstr = "(x -> categorical(x, compress=$compress))"
+    end
+    if cols === nothing
+        cols = Union{AbstractString, Missing}
+        Base.depwarn("`categorical!(df, cols)` is deprecated. " *
```
bkamins, line 99:
```suggestion
        Base.depwarn("`categorical!(df)` is deprecated. " *
```

src/deprecated.jl
```diff
@@ -8,3 +8,102 @@ by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataF
...
+                     "Use `cols = names(df)[map(c -> eltype(c) <: $cols, eachcol(df))]; transform!(df, cols .=> $categoricalstr .=> cols)` instead.",
+                     :categorical!)
+    else
+        Base.depwarn("`categorical!(df, cols)` is deprecated. " *
+                     "Use `cols = names(df)[map(c -> eltype(c) <: cols, eachcol(df))]; transform!(df, cols .=> $categoricalstr .=> cols)` instead.",
```
bkamins, line 104: same as above

test/reshape.jl
```diff
@@ -473,7 +473,7 @@ end
     @test d1s[:, 4] isa Vector{Symbol}
     @test levels(d1s[:, 4]) == [:c, :d]
 
-    d2 = categorical(d1, :)
+    d2 = transform(d1, names(d1) .=> categorical .=> names(d1))
```
bkamins, line 476: interestingly in this special case it can be just `mapcols(categorical, df)`.

nalimilan:
I've applied your comments and added a NEWS entry (we'll need a more general one if we stop reexporting CategoricalArrays).

src/deprecated.jl
```diff
@@ -8,3 +8,102 @@ by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataF
...
+        cols = Union{AbstractString, Missing}
+        Base.depwarn("`categorical(df)` is deprecated. " *
+                     "Use `cols = names(df)[map(c -> eltype(c) <: $cols, eachcol(df))]; transform(df, cols .=> $categoricalstr .=> cols)` instead.",
+                     :categorical)
+    else
+        Base.depwarn("`categorical(df, T)` is deprecated. " *
+                     "Use `cols = names(df)[map(c -> eltype(c) <: T, eachcol(df))]; transform(df, cols .=> $categoricalstr .=> cols)` instead.",
+                     :categorical)
+    end
+    cols = names(df)[map(c -> eltype(c) <: cols, eachcol(df))]
```
bkamins, line 59: you still have the "double use" of `cols` here (and in `categorical!`) - it is not end of the world, but maybe it would be good to fix it?

bkamins (approved):
OK - lets deprecate these methods

bkamins:
@nalimilan - please merge it when you are done. Thank you!

src/deprecated.jl
nalimilan: Ah right. Fixed.

Issue #2396, opened 2020-08-30
add a kwarg to reuse the input column names in cols => fun in select/transform/combine


bkamins:
This request is showing often as sometimes adding transformation function name is not desired.
A working idea is to add a `suffix` kwarg that would handle this (actually the suffix might be more flexible than just `true`/`false`, but a general suffix - I will propose something).

I will make a PR for this. This is an issue to keep track that this happens.

bkamins:
@nalimilan - the only problematic case to decide what to do is `[] => fun`. What should be the column name then when we drop adding function as a suffix? It can be `Symbol("")`,  as we allow it, but it seems to be obscure. However, maybe it is OK, as one has to be explicit to get this outcome.

nalimilan:
I'd say `Bool` is OK for a start, even if later we might want to allow passing any function. For `[] => fun` let's just throw an error?

Though for consistency with `join` (https://github.com/JuliaData/DataFrames.jl/pull/2313) shouldn't the argument be called `rename`?

bkamins:
Now I see this comment. So maybe in joins I also change it to `renamecols` to have the same kwarg name everywhere?

> For [] => fun let's just throw an error?

OK

nalimilan:
Yeah if you want to avoid breaking `unstack` then better use `renamecols` and change `join` to use that before a release includes it.

bkamins:
This is what I will do.

PR #2397, opened 2020-08-30
add suffix to select/transform/combine

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2396

I ended up deciding that `suffix` should be `Bool` (other uses did not make sense for me).

Before I update the docs and write tests a comment if we all agree that `suffix` kwarg name is OK (I am rather convinced) + that it is enough to make it `Bool` (here I am pretty convinced).


src/abstractdataframe/selection.jl
```diff
@@ -138,11 +142,19 @@ function normalize_selection(idx::AbstractIndex,
     end
     fun = last(sel)
     if length(c) > 3
-        newcol = Symbol(join(@views(_names(idx)[c[1:2]]), '_'), "_etc_", funname(fun))
+        if suffix
+            newcol = Symbol(join(@views(_names(idx)[c[1:2]]), '_'), "_etc_", funname(fun))
```
nalimilan, line 146: Maybe store the common part in a string to avoid duplicating the code? Same below.

bkamins: OK. But do you agree with the `suffix` name of the kwarg, if yes I would finalize the PR.

bkamins:
OK - I have added tests, updated docstrings, and started throwing an error when empty column selector is passed.

src/abstractdataframe/selection.jl
```diff
@@ -251,10 +274,13 @@ SELECT_ARG_RULES =
...
-    input columns and they are joined using `_`; if more than three columns are
-    passed then the name consists of the first two names and `etc` suffix then,
-    e.g. `[:a,:b,:c,:d] => fun` produces the new column name `:a_b_etc_fun`.
+    `\$(old_column)_\$(fun)` if `renamecols=true` and `\$(old_column)` if
+    `renamecols=false`. Up to three column names are used for multiple input
+    columns and they are joined using `_`; if more than three columns are passed
+    then the name consists of the first two names and `etc` suffix then, e.g.
+    `[:a,:b,:c,:d] => fun` produces the new column name `:a_b_etc_fun` if
+    `renamecols=true` and ``:a_b_etc` if `renamecols=false`.
+    It is not allowed to pass `renamecols=false` if `old_column` is empty.
```
nalimilan, line 283:
```suggestion
    It is not allowed to pass `renamecols=false` if `old_column` is empty
    as it would generate an empty column name.
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -227,13 +227,15 @@ const F_ARGUMENT_RULES =
...
-    specified to be selected, or a `NamedTuple` containing these `SubArray`s if
-    `source_cols` is an `AsTable` selector. It can return a vector or a single value
-    (defined precisely below).
+    Transformations allowed using `Pair`s follow the rules specified for
+    [`select`](@ref) and have the form `source_cols => fun`, `source_cols => fun
+    => target_col`, or `source_col => target_col`. Function `fun` is passed
+    `SubArray` views as positional arguments for each column specified to be
+    selected, or a `NamedTuple` containing these `SubArray`s if `source_cols` is
+    an `AsTable` selector. It can return a vector or a single value (defined
+    precisely below). In particular if automatic generation of target column
```
nalimilan, line 236: Why "in particular"?
```suggestion
    precisely below). If automatic generation of target column
```

nalimilan (approved):
Just two small suggestions.

bkamins:
I have added `renamecols` in the `categorical` deprecation

bkamins:
Thank you!

pdeffebach:
We should add a `news.md` entry for this!

bkamins:
We have, it is just called `renamecols` (the same for `joins`) - for consistency across different functions.

pdeffebach:
Ah thanks!

Issue #2395, opened 2020-08-30
Allow passing type as a column selector of a data frame


bkamins:
This request is showing often. I will make a PR for this. This is an issue to keep track that this happens.

PR #2400, opened 2020-08-30
add Type to names as cols

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2395

Now I understand why I have not added it in the past - `AbstractIndex` is not aware of column types.

Therefore I propose to add this option to `names` only. This is useful on its own right. In other places one will have to write something like `df[:, names(df, Int)]` which is a bit verbose, but not that bad. The benefits are twofold:
1. This way is explicit and easy to understand when reading the code
2. I do not have to change dozens of methods (as I would have to add a special method to every existing method that accepts `MultiColumnIndex`, as we cannot just extend `MultiColumnIndex` with `Type`)


nalimilan:
Thanks, looks good. I still think it would be nice to support this everywhere, but indeed it will require a deeper refactoring.

Could you update the `categorical`/`categorical!` deprecation to use this?

bkamins:
I updated `categorical` deprecation. When this is merged I will do another update in https://github.com/JuliaData/DataFrames.jl/pull/2397 and we are cleanly done then.

pdeffebach:
This is great! Love that we have `names` to avoid over-complicating indexing.

bkamins:
Now the deprecation of `categorical` is clean indeed.

nalimilan (approved):
Yeah, it's nice that the deprecation is reasonable. That also means many similar operations are now quite easy to do.

bkamins:
Thank you!

PR #2402, opened 2020-08-31
Fix typo in docstring of `stack`

Sh4pe:
The column holding the different :values was erroneously called
'variable_name' in the docstring.


bkamins (approved):

bkamins:
Thank you!

Issue #2389, opened 2020-08-28
do we want to limit column size to 32 characters in text/plain display


bkamins:
In https://github.com/JuliaData/DataFrames.jl/pull/2346 this is what we do for markdown. Maybe it would be good to do for all types?

nalimilan:
Makes sense, though we should check that other implementations do.

bkamins:
Also CC @ronisbr

ronisbr:
I totally agree with that! Anyway, with PrettyTables.jl, it will be very easy to let the user change this if they want.

bkamins:
OK - I will implement it when #2390 is merged.

PR #2403, opened 2020-08-31
allow to flexibly change the display width limit

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2389

I made it a bit different than Markdown rendering, but I think we do not have to be super cautious here as this is just handling corner cases of display.

Also fixes the problem of overwide first column (not as nicely as PrettyTabes.jl but again - this is a rare corner case now as we are trimming at 32 screen width anyway)

CC @nalimilan @ronisbr

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -68,5 +68,5 @@
 * add rich display support for Markdown cell entries in HTML and LaTeX
   ([#2346](https://github.com/JuliaData/DataFrames.jl/pull/2346))
 * limit display width in `text/plain` to approximately 32 screen characters
-  per column and fix a corner case when no columns are printed as they are
-  too wide ([2403](https://github.com/JuliaData/DataFrames.jl/pull/2403))
+  per column by default and fix a corner case when no columns are printed in situations
+  when they are too wide ([2403](https://github.com/JuliaData/DataFrames.jl/pull/2403))
--- a/NEWS.md
+++ b/NEWS.md
@@ -67,6 +67,7 @@
   ([#2315](https://github.com/JuliaData/DataFrames.jl/pull/2315))
 * add rich display support for Markdown cell entries in HTML and LaTeX
   ([#2346](https://github.com/JuliaData/DataFrames.jl/pull/2346))
-* limit display width in `text/plain` to approximately 32 screen characters
-  per column by default and fix a corner case when no columns are printed in situations
-  when they are too wide ([2403](https://github.com/JuliaData/DataFrames.jl/pull/2403))
+* limit the maximal display width the output can use in `text/plain` before
+  being truncated (in the `textwidth` sense, excluding `…`) to `32` per column
+  by default and fix a corner case when no columns are printed in situations when
+  they are too wide ([2403](https://github.com/JuliaData/DataFrames.jl/pull/2403))
--- a/NEWS.md
+++ b/NEWS.md
@@ -67,3 +67,6 @@
   ([#2315](https://github.com/JuliaData/DataFrames.jl/pull/2315))
 * add rich display support for Markdown cell entries in HTML and LaTeX
   ([#2346](https://github.com/JuliaData/DataFrames.jl/pull/2346))
+* limit display width in `text/plain` to approximately 32 screen characters
+  per column and fix a corner case when no columns are printed as they are
+  too wide ([2403](https://github.com/JuliaData/DataFrames.jl/pull/2403))
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -669,9 +669,9 @@ while `splitcols` defaults to `true`.
 - `rowlabel::Symbol = :Row`: The label to use for the column containing row numbers.
 - `summary::Bool = true`: Whether to print a brief string summary of the data frame.
 - `eltypes::Bool = true`: Whether to print the column types under column names.
-- `truncate::Int = 32`: The approximate number of text width characters to truncate
-  output in the rendered cell of a data frame. If `truncate` is non-positive
-  then no truncation is applied.
+- `truncate::Int = 32`: the maximal display width the output can use before
+  being truncated (in the `textwidth` sense, excluding `…`).
+  If `truncate` is 0 or less, no truncation is applied.
 
 # Examples
 ```jldoctest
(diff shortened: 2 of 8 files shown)


bkamins:
In the process I have noticed that we did an incorrect escaping of CSV and TSV output. I have fixed it. One debatable thing is what we should use in CSV/TSV for non-standard types (like `Symbol`, `Char` - in the tests, or e.g. `Date`) - should we use `print` or `repr` to show them. I opted for `repr` now (as it kind-of retains type information), but maybe `print` would be preferred. Comments are welcome here.

bkamins:
@nalimilan - tests will fail now. I will clean them when we decide if we want to use `repr` or `print` in CSV/TSV output.

A a comment - I thought of `repr` as `print` is used in CSV.jl, so here we could provide a more low-level output. Probably this CSV/TSV feature is not used much (if at all - given it had escaping bug no one reported) so we could make it have a different behavior than CSV.jl.

NEWS.md
```diff
@@ -67,3 +67,6 @@
   ([#2315](https://github.com/JuliaData/DataFrames.jl/pull/2315))
 * add rich display support for Markdown cell entries in HTML and LaTeX
   ([#2346](https://github.com/JuliaData/DataFrames.jl/pull/2346))
+* limit display width in `text/plain` to approximately 32 screen characters
```
nalimilan, line 70: Is it really approximate? AFAICT, you stop at 32 and take the previous character, so it's always the maximum width before going beyond 32, right?

src/abstractdataframe/show.jl
```diff
@@ -13,6 +13,17 @@ function ourstrwidth(io::IO, x::Any, buffer::IOBuffer)
     return textwidth(String(take!(buffer)))
 end
 
+function trimstring32(s::AbstractString)
```
nalimilan, line 16: `trim` in the context of strings usually means dropping whitespace at the beginning and end AFAIK. Maybe `truncate`?

nalimilan:
I'd prefer using `print`, as it gives something which is more likely to be usable and parseable. Even reading back a date into Julia will work better using `print`.

Regarding truncation, I've just checked, and dplyr only truncates when the output would be wider than the screen. Maybe we should do that too? One use case I can imagine is a column with short texts and one or two columns with variables describing them.

bkamins:
> I'd prefer using `print`

OK - I switch to `print`.

>  Maybe we should do that too?

If you try to do this it quickly starts to be very messy, it was tried in https://github.com/JuliaData/DataFrames.jl/pull/2087.
Actually even in your scenario I would prefer truncating (32 characters usually will be enough to get an understanding what is inside a string)

Also soon we will have a PrettyTables.jl backend to choose from that does what you propose (but it has a different design in general) - so someone will be able to go for that backend if needed.

NEWS.md
bkamins: No, because some characters have more than `1` text width (in which case we can get a bit different result)

src/abstractdataframe/show.jl
bkamins: OK - I will change to `truncate`.

nalimilan:
> If you try to do this it quickly starts to be very messy, it was tried in #2087.
> Actually even in your scenario I would prefer truncating (32 characters usually will be enough to get an understanding what is inside a string)
>
> Also soon we will have a PrettyTables.jl backend to choose from that does what you propose (but it has a different design in general) - so someone will be able to go for that backend if needed.

In the use case I have in mind, you have unique texts that you may want to read, so just having an idea of the contents isn't enough. But yeah it's difficult to handle this well in all cases. It's nice if PrettyTables aims to fill this gap.

bkamins:
I am changing the implementation to add `truncate` keyword argument that allows to specify number of characters at which the truncation happens, as this was needed anyway. Then also your use case will be handled

bkamins:
I had to add `truncate` keyword argument for generality (as things are unfortunately problematic in corner cases). I have decided to never truncate: 1) row column name, 2) column name, 3) column eltype if shown.

nalimilan:
We already abbreviate eltypes when they are too long, right?

NEWS.md
nalimilan: Yes but since you return the last character before going beyond 32, the result will always be less than 32, right? I think it would be worth mentioning in the docstring rather than just saying it's approximate.

bkamins:
> We already abbreviate eltypes when they are too long, right?

Yes, but now `truncate` can be less than we abbreviate (a corner case, but still).

NEWS.md
bkamins: The precise statement is that the result without `…` has maximal text width less or equal than 32 display characters, but how would you phrase it so that this would be readable?

bkamins:
@nalimilan - this should be good to have a look at. Thank you!

NEWS.md
nalimilan: Maybe something like this? Anyway that's internal, what matters is that it's correct even if it's not super easy to read. :-)
```
`truncstring` indicates the maximal width the output can use before being truncated
(in the `textwidth` sense, excluding `…`).
```

nalimilan:
> Yes, but now `truncate` can be less than we abbreviate (a corner case, but still).

Can you give an example? I don't get it. :-)

bkamins:
> Can you give an example? I don't get it. :-)

I explicitly test this case here:

https://github.com/JuliaData/DataFrames.jl/pull/2403/files#diff-be8288df383517cc1dabe74772b7535aR723

NEWS.md
bkamins: OK - I have proposed something along these lines :smile:

src/abstractdataframe/show.jl
```diff
@@ -643,6 +669,9 @@ while `splitcols` defaults to `true`.
 - `rowlabel::Symbol = :Row`: The label to use for the column containing row numbers.
 - `summary::Bool = true`: Whether to print a brief string summary of the data frame.
 - `eltypes::Bool = true`: Whether to print the column types under column names.
+- `truncate::Int = 32`: The approximate number of text width characters to truncate
+  output in the rendered cell of a data frame. If `truncate` is non-positive
+  then no truncation is applied.
```
nalimilan, line 674: Actually I was more concerned about the docstring than about the NEWS. :-)
```suggestion
- `truncate::Int = 32`: the maximal display width the output can use before
  being truncated (in the `textwidth` sense, excluding `…`).
  If `truncate` is 0 or less, no truncation is applied.
```

nalimilan (approved):
Looks good with the small suggestion for the docstring.

bkamins:
Thank you for fixing the docstring.

bkamins:
Thank you!

PR #2411, opened 2020-09-05
Replaced ParquetFiles.jl with Parquet.jl

xiaodaigh:
As there are long-standing issues that makes ParquetFiles unusable see https://github.com/queryverse/ParquetFiles.jl/pulls where ParquetFiles.jl has not updated to Parquet 0.6 and is still relying on Parquet 0.4.

Also, Parquet.jl can read and write Parquet files now and although it is not feature-complete it is a much better alternative to ParuqetFiles.jl now.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -107,8 +107,8 @@ integrated they are with DataFrames.jl.
       CSVs (using [CSV.jl](https://github.com/JuliaData/CSV.jl)),
       Stata, SPSS, and SAS files (using
       [StatFiles.jl](https://github.com/queryverse/StatFiles.jl)),
-      and reading (though not writing) parquet files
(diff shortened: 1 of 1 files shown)


bkamins:
Thank you for pointing this out.

@davidanthoff and @tanmaykm - can you just confirm this is the way to go before we merge (maybe we should point to both highlighting the differences)? Thank you!

quinnj (approved):

nalimilan:
Makes sense. Though tt would be nice if the Parquet.jl docs mentioned how to load a Parquet file into a DataFrame, or at least mention Tables.jl. Also, where did this weird idea of naming the type `ParFile` instead of `ParquetFile` come from? :-/

PR #2412, opened 2020-09-05
add SplitApplyCombine.jl reference in the docs

bkamins:


nalimilan:
I'm a bit hesitant to point to SplitApplyCombine here, especially if we present it as "high-level". It might be high-level compared with Base, but in the context of DataFrames it's rather low-level. Also the functions it provides don't really interoperate with DataFrames, and often compete with it (e.g. `group`).

I guess you wanted to mention it because of `invert`?

bkamins:
A good point. Yes - the reason is `invert` only. So what do you think would be the best approach (again - the description is just an adjustment of the description the package gives itself 😄)

bkamins:
closing this, as indeed we have to handle `invert` separately.

PR #2413, opened 2020-09-05
Added some info about PairAsPipe.jl and the `@pap` macro

xiaodaigh:


bkamins:
This is essentially https://github.com/matthieugomez/DataFramesMacros.jl created by @matthieugomez.
Also currently a very similar thing is added to DataFramesMeta.jl by @pdeffebach.

What is clear is that this syntax is widely wanted.
I propose to wait a bit with this PR to see how things settle down and mature and then I would prefer to have a single recommended way of doing this.

Thank you for working on this!

xiaodaigh:
I didn't know about those two. I might as well contribute to them

bkamins:
Thank you - combining the efforts is a great idea, as we have tons of things to add to make the DataFrames.jl and related packages feature complete.

pdeffebach:
@xiaodaigh One thing that would be super helpful for DataFramesMeta is getting the docs working as actual documentation so people can see the versions. I know nothing about docs! It would be great if you could help out with that!

Issue #2314, opened 2020-07-07
Skipping missing values more easily


nalimilan:
It seems that dealing with missing values is one of the most painful issues we have, which goes against the very powerful and convenient DataFrames API. Having to write things like `filter(:col => x -> coalesce(x > 1, false), df)` or `combine(gd, :col => (x -> sum(skipmissing(x)))` isn't ideal. One proposal to alleviate this is https://github.com/JuliaData/DataFrames.jl/issues/2258: add a `skipmissing` argument to functions like `filter`, `select`, `transform` and `combine` to unify the way one can skip missing values, instead of having to use different syntaxes which are hard to grasp for newcomers and make the code more complex to read.

That would be one step towards being more user-friendly, but one would still have to repeat `skipmissing=true` all the time when dealing with missing values. I figured two solutions could be considered to improve this:
- Have DataFramesMeta simplify the handling of missing values. This could be via a dedicated e.g. `@linqskipmissing` macro or a statement like `skipmissing` within a `@linq` block that would automatically pass `skipmissing=true` to all subsequent operations in a chain or block. This wouldn't really help with operations outside such blocks though.
- Have a field in `DataFrame` objects that would store the default value to use for the `skipmissing` argument. By default it would be `false`, so that you get the current behavior, which ensures safety via propagation of missing values. But when you know you are working with a data set with missing values, you would be able to call `skipmissing!(df, true)` once and then avoid repeating it.

Somewhat similar discussions have happened a long time ago (but at the array rather than the data frame level) at https://github.com/JuliaStats/DataArrays.jl/issues/39. I think it's fair to say that we know have enough experience now to make a decision. One argument against implementing this at the `DataFrame` level is that it will have no effect on operations applied directly to column vectors, like `sum(df.col)`. But that's better than nothing.

Cc: @bkamins, @matthieugomez, @pdeffebach, @mkborregaard

pdeffebach:
> * Have DataFramesMeta simplify the handling of missing values. This could be via a dedicated e.g. `@linqskipmissing` macro or a statement like `skipmissing` within a `@linq` block that would automatically pass `skipmissing=true` to all subsequent operations in a chain or block. This wouldn't really help with operations outside such blocks though.

This is a good idea, however I also like `skipmissing = true` at the level of a `transform` call or even at the level of argument because it's explicit. 

Perhaps DataFramesMeta could provide a block-level `skipmissing` option as well as a macro like

```
[...]

PR #2417, opened 2020-09-06
add predicate support for names and more tests

bkamins:
@nalimilan I added a predicate for `names` + added more tests (and keep them in one place so that it easier to make sure all is tested). In particular methods for `DataFrameRow` were missing.

One thing I was considering if the predicate should not take `String => Column` pair to allow for more flexibility, but decided that passing `String` is enough as I think the more complex cases can be relatively easily handled by `pairs(eachcol(df))` if neeeded and they are probably much rarer than the case of eg. `startswith`.


bkamins:
@nalimilan - I have felt that there is some corner case lurking, and it is `:`, which is a `Function` :smile:. Fixed now.

src/dataframerow/dataframerow.jl
```diff
@@ -272,6 +272,12 @@ function Base.names(r::DataFrameRow, cols)
     return [string(nms[i]) for i in idxs]
 end
 
+Base.names(r::DataFrameRow, T::Type) =
+    [String(n) for n in _names(index(r)) if eltype(parent(r)[!, n]) <: T]
```
nalimilan, line 276: Why not this (given that the method is defined just below)?

```suggestion
    [n for n in _names(r) if eltype(parent(r)[!, n]) <: T]
```
Though this is a case where `pairs` could be a bit simpler.

src/dataframerow/dataframerow.jl
```diff
@@ -272,6 +272,12 @@ function Base.names(r::DataFrameRow, cols)
...
+
+Base.names(r::DataFrameRow, fun::Function) =
+    [String(n) for n in _names(index(r)) if fun(String(n))]
```
nalimilan, line 279:
```suggestion
    filter(fun, _names(r))
```

nalimilan:
> One thing I was considering if the predicate should not take `String => Column` pair to allow for more flexibility, but decided that passing `String` is enough as I think the more complex cases can be relatively easily handled by `pairs(eachcol(df))` if neeeded and they are probably much rarer than the case of eg. `startswith`.

Yes, that occurred to me too. On the one hand, since this is the `names` function, it makes sense that it would operate on names and not columns in general. On the other hand, if we allow passing a type we already go beyond pure names.

Something that isn't ideal in dplyr is that you can do `across(starts_with("sepal"), mean)` but you have to write `across(where(is.numeric), mean)`. The `where` doesn't sound like a good term to mean "refer to the contents of columns and not just their names". So I hope we can find something better. We already do since we allow passing types, which is probably the most common case (apart from conditions on names).

It's great that you can write `names(df, startswith("a"))`, which is very concise and easy to understand. If we passed `(name, column)` as an argument; having to write `names(df, (n, col) -> startswith(col, "a"))` wouldn't be as nice. But then what can we recommend to select e.g. all columns without missing values? `names(df)[.!any.(ismissing, eachcol(df))]` is relatively short but quite tricky. Though even if we passed `(name, column)` as an argument, `names(df, (n, col) -> !any(ismissing, col))` wouldn't be much better. So maybe the PR is OK as it is...

test/dataframe.jl
```diff
@@ -1881,16 +1881,32 @@ end
...
+                        b1 = 1.0:3.0, b2 = [1.0, missing, 3.0],
+                        c1 = '1':'3', c2 = ['1', missing, '3'], x=1:3)
+    for x in (df_long[:, Not(end)], @view(df_long[:, Not(end)]),
+              groupby(df_long[:, Not(end)], :a1), groupby(@view(df_long[:, Not(end)]), :a1),
+              eachrow(df_long[:, Not(end)]), eachrow(@view(df_long[:, Not(end)])),
+              eachcol(df_long[:, Not(end)]), eachcol(@view(df_long[:, Not(end)])),
+              df_long[1, Not(end)])
+        @test names(x, 1) == ["a1"]
+        @test names(x, "a1") == ["a1"]
+        @test names(x, :a1) == ["a1"]
```
tk3369, line 1895: Is it actually useful to support these? Why would I ask for names if I already know the names? 🤔

test/dataframe.jl
```diff
@@ -1881,16 +1881,32 @@ end
...
+        @test names(x, [2, 1]) == ["a2", "a1"]
+        @test names(x, ["a2", "a1"]) == ["a2", "a1"]
+        @test names(x, [:a2, :a1]) == ["a2", "a1"]
+        @test names(x, Int) == ["a1"]
+        @test names(x, Int) == ["a1"]
```
tk3369, line 1900: Repeated.

test/dataframe.jl
```diff
@@ -1881,16 +1881,32 @@ end
...
+        @test names(x, Union{Missing, Int}) == ["a1", "a2"]
```
tk3369, line 1901: As a user, when I ask for `names(x, Int)`, my intuition tells me that all `Int` columns should be considered regardless of whether it has missing values or not. Maybe it's better to pick a default behavior, and let the user choose the different behavior via another argument or a kwarg?

test/dataframe.jl
bkamins: it is a test that all works correctly - not an intended use.

test/dataframe.jl
bkamins: @nalimilan - what do you think? I can see the usefulness of doing the union with `Missing` by default (most likely this is what you want typically) but then we need to add a kwarg to allow to opt-out of it.

nalimilan: Well it's a difficult choice. Ideally we would be able to write `Int?` and everything would be nice. :-)

I agree that in most cases people probably don't care about whether the columns allow for missing values. But automatically assuming `Missing` makes things less obvious, so I don't know...

bkamins:
> `names(df)[.!any.(ismissing, eachcol(df))]`

`names(df, Not(findall(any.(ismissing, eachcol(df))))` is longer but I would say natural and relatively easy to understand.

test/dataframe.jl
bkamins: I will go for what @tk3369 suggests and add `skipmissing` kwarg with default to `false`. The reason is that `Union{Missing, Int}` is not ideal, as it allows bare `Missing`. I will propose something in the update of the PR that I find reasonable.

bkamins:
> How about allowing regexes while you're at it? :-)

What do you mean by this? We already allow regexes now (this is on 0.21):
```
julia> using DataFrames

julia> df = DataFrame(x=1,y=2)
names(df, 1×2 DataFrame
│ Row │ x     │ y     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 2     │

julia> names(df, r"x")
1-element Array{String,1}:
 "x"
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -85,6 +88,9 @@ end
 Base.names(df::AbstractDataFrame, T::Type) =
     [String(n) for (n, c) in pairs(eachcol(df)) if eltype(c) <: T]
 
+Base.names(df::AbstractDataFrame, fun::Function) =
+    [String(n) for (n, c) in pairs(eachcol(df)) if fun(String(n))]
```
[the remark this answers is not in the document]
bkamins: This is incorrect, as `_names` return `Symbol`s, and we want to pass `String`s to `fun` (and I want this function to be efficient - using `names` will allocate twice). But indeed `_names` is better to use - I will change it.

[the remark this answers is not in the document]
bkamins: Ah - I have benchmarked that actually `String` on `Symbol` copies. Then your suggestion is OK but with `names` not `_names`.

src/dataframerow/dataframerow.jl
bkamins: again - we need to return `String`s not `Symbol`s here. I will update it accordingly.

test/dataframe.jl
bkamins: fixed.

test/dataframe.jl
bkamins: Done - the rule is complex, but I assume it reflects the intuitive expectation. I have listed all relevant cases in tests.

bkamins:
CI passes - it just a time out of the worker.

test/dataframe.jl
nalimilan: Actually I think we should consider this in the context of https://github.com/JuliaData/DataFrames.jl/issues/2314. If we don't skip missing values by default anywhere, should we really assume `Union{T, Missing}` by default here? Maybe we should rather be strict and explicit here, but find a more concise syntax for DataFramesMeta so that by default it skips missing values and assumes `T` means `Union{T, Missing}`?

bkamins: Let us discuss it, but my point is that I do not implement `Union{T, Missing}` here. I implement something more careful, that is not easy to express using type system. The issue is that `Missing <: Union{T, Missing}` and we want to exclude this case, unless `T` is `Missing` (and some more corner cases like `Union{}`).

Anyway - there is no rush with merging this PR so we can think what is best.

nalimilan (commented):
Sorry I totally forgot this.

bkamins:
> Sorry I totally forgot this.

There is no rush with this PR. It is not scheduled for 0.22 as I am still not sure what is the best design here.

src/other/utils.jl
```diff
@@ -88,3 +88,15 @@ else
 end
 
 funname(c::ComposedFunction) = Symbol(funname(c.f), :_, funname(c.g))
+
+testtype(t::Type, ct::Type, unionmissing::Bool) =
```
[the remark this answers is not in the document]
bkamins: fixed

src/other/utils.jl
```diff
@@ -88,3 +88,15 @@ else
...
+    if t == Missing
+        return ct === Missing
+    else
+        if unionmissing
+            ct === Missing && return t === Union{} || Missing <: t
```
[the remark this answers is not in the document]
bkamins: It is good we have left some time to sleep over this PR.

The core question is (do not look at the implementation as it is just a consequence). If `eltype` of the column is `Missing` and user passes `unionmissing` equal to `true` what should happen then? The implementation was doing special casing of this case. Now I think if `CT === Missing` and `unionmissing===true` then we should always return `true`. Do you agree? (if yes - the above implementation will change)

test/dataframe.jl
bkamins: > The issue is that `Missing <: Union{T, Missing}` and we want to exclude this case, unless `T` is `Missing` (and some more corner cases like `Union{}`).

@nalimilan - this was my earlier assumption, but maybe we should change this as I described below.

The core of the question is if column contains only `missing` values and has `eltype` `Missing` should it match type e.g. `Int` but with `unionmissing=true`? (the thing is that this column does not allow the originaly passed type, e.g. `Int`)

bkamins:
@nalimilan - I was thinking about this PR and `unionmissing` option. Maybe we should not introduce it and ask what is required now - to give an explicit signature. This is not that much more writing and in the future it is possible it will be only adding `?`. On the other hand adding `unionmissing` might be too much magic - this is not the way we try to do things (e.g. the discussion on `where` shows that in an even more obvious case we have doubts about being flexible).

src/other/utils.jl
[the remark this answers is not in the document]
nalimilan: The simplest and most obvious approach would be to check `ct <: Union{t, Missing}` when `unionmissing=true`. Otherwise things get too complex IMHO.

nalimilan:
Yes I think we should decide on a general policy which would also cover `where` (and maybe other cases).

src/other/utils.jl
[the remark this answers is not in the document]
bkamins: The only problem is that if someone writes `names(df, Float64)` and has a column with `eltype` equal to `Missing` then one would get that column although it does not allow `Float64` values at all.

bkamins:
I have thought of it and commented in #2496. I think here we do not need this kwarg and in `where` `dropmissing=false` should be the default.

I think it is better to keep DataFrames.jl close to Julia Base behavior, as otherwise it will lead to the confusion.

bkamins:
Given we have released 0.22 and now will be non-breaking `unionmissing` proposal is out of discussion as it would be breaking (the decision slipped through - my bad).
So in this PR what remains is the "predicate" part and "cleanup" part.

Cleanup of the functionality should for sure be done. Adding predicate also makes sense. Then the only question is to finally settle does the predicate take only column (values) or a pair `name => column`. I think passing column only is enough.

nalimilan:
> Given we have released 0.22 and now will be non-breaking `unionmissing` proposal is out of discussion as it would be breaking (the decision slipped through - my bad).

If we really think the behavior should be changed I think it would be OK to have some limited breakage in 1.0, especially since it's a very recent feature. But it's hard to decide what's best.

> So in this PR what remains is the "predicate" part and "cleanup" part.
> 
> Cleanup of the functionality should for sure be done. Adding predicate also makes sense. Then the only question is to finally settle does the predicate take only column (values) or a pair `name => column`. I think passing column only is enough.

Do you mean passing column or passing name? :-D

bkamins:
> But it's hard to decide what's best.

Exactly - that is why I would leave it out. I think what we have now is also OK, and it will be possibly even more convenient in the future if `Union{Missing, T}` syntax would be made easier to write.

> Do you mean passing column or passing name? :-D

OK to be clear we potentially could have three behaviors:
1. passing name
2. passing column (vector)
3. passing "name => column" pair

I propose to use the option #2 (passing column). I think that this is most often what users would want.

In particular option #1 is easy enough to achieve using `filter!(predictate, names(df))`. Option #3 which can be done made with `eachcol` and `pairs`.

nalimilan:
OK, so you changed your mind since you filed this PR. :-)

I'm really torn as both options are useful, and both have less user-friendly alternatives. But I tend to think conditions on names are more common than conditions on columns. For example, in dplyr's help, `where` is documented only for `where(is.numeric)` and `where(is.factor)`, and we cover both using the method taking a type.

Also, a criterion to help making our decision is that it could make sense to have `names(df, cols)` be consistent with `Cols(cols)` (and possibly even with `df[!, cols]` if we want to make it more flexible later). It would be nice to be able to do e.g. `select(df, Cols(startswith("x")) => +)`. Of course this is already possible with `r"^x"` but `startswith` is probably nicer for many users -- in particular those who would appreciate the most to have a simple syntax.

bkamins:
:smile: - good to have you here. Indeed I have changed my mind. Which means - as you say that probably both options would be nice to have and both have "less convenient" alternatives.

I understand you propose the predicate to work on column names? Maybe the solution would be the following:
* `names(df, predicate)` and in the future other options would take column name
* `names(df, Cols(predicate))` would pass columns to a predicate (I propose `Cols` as it avoids introducing a new name and it seems to have a natural meaning)

What do you think?

bkamins:
@nalimilan - I propose to make a decision on this PR. My current thinking, to keep the API simple, that we just add:
```
names(df::AbstractDataFrame, predicate::Function) = filter(predicate, _names(df))
```
to keep `names` be column name focused and this is probably a more common case.

This would finalize the API of `names` and I would avoid adding more complexity to it (it is already quite large).

nalimilan:
Sounds good. That would indeed keep things relatively simple.

bkamins:
@nalimilan - this should be good to have a look at

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -68,12 +68,16 @@ Return a freshly allocated `Vector{String}` of names of columns contained in `df
...
 If `cols` is passed then restrict returned column names to those matching the
 selector (this is useful in particular with regular expressions, `Cols`, `Not`, and `Between`).
-`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR)
-or a `Type`, in which case columns whose `eltype` is a subtype of `cols` are returned.
+`cols` can be:
+* any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR)
+* a `Type`, in which case names of columns whose `eltype` is a subtype of `T`
+  are returned
+* a `Function` predicate, in which case names of columns for which the predicate, taking a
+  `String` containg column name, returns `true`
```
nalimilan, line 76:
```suggestion
* a `Function` predicate taking the column name as a string and returning `true`
  for columns that should be kept
```

src/dataframerow/dataframerow.jl
```diff
@@ -272,6 +272,10 @@ function Base.names(r::DataFrameRow, cols)
     return [string(nms[i]) for i in idxs]
 end
 
+Base.names(r::DataFrameRow, T::Type) =
+    [String(n) for n in _names(r) if eltype(parent(r)[!, n]) <: T]
+Base.names(r::DataFrameRow, fun::Function) = filter(fun, names(r))
```
nalimilan, line 277: Likewise, why not use `filter!` or `_names` here?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -84,6 +88,7 @@ end
 
 Base.names(df::AbstractDataFrame, T::Type) =
     [String(n) for (n, c) in pairs(eachcol(df)) if eltype(c) <: T]
+Base.names(df::AbstractDataFrame, fun::Function) = filter(fun, names(df))
```
nalimilan, line 91: IIRC `names` always return a copy so `filter!` is safe, right?
```suggestion
Base.names(df::AbstractDataFrame, fun::Function) = filter!(fun, names(df))
```

src/dataframerow/dataframerow.jl
bkamins: Right. The problem is that I used `_names` and it returns a vector of `Symbol` and I have forgotten to add `!`. `names` is guaranteed to freshly allocate a `Vector{String}`.

bkamins, line 277:
```suggestion
Base.names(r::DataFrameRow, fun::Function) = filter!(fun, names(r))
```

nalimilan (approved):

PR #2418, opened 2020-09-08
Add canonical URL to manual

nalimilan:
This prevents search engines from pointing to outdated releases. The same change should be applied to HTML files for older releases.

Currently Google seems to point to the latest release, but this is safer.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -14,7 +14,7 @@ makedocs(
     doctest = false,
     clean = false,
     sitename = "DataFrames.jl",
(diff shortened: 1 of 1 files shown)


bkamins (approved):
I assume you know what you are doing 😄

nalimilan:
Let's hope that's the case. I just copied the CSV.jl PR. :-p

quinnj:
> Let's hope that's the case. I just copied the CSV.jl PR. :-p

Where I also assumed people knew what they were doing! 🤣

PR #2427, opened 2020-09-12
remove dependency on CategoricalArrays.jl in legacy show

bkamins:
Using reflection is the best what could come up with unless there is a solution to https://stackoverflow.com/questions/63860020/how-to-get-a-list-of-modules-imported-by-a-given-module (which seems unlikely - I have checked the source code, and it seems that this part of the functionality is not exposed to Julia code and is only available in C).

The approach is assuming that if something is pretending to be CategoricalArrays.jl we can just treat it in display as if it were CategoricalArrays.jl (and in general such pretending is unlikely).


src/abstractdataframe/show.jl
```diff
@@ -113,7 +119,11 @@ function compacttype(T::Type, maxwidth::Int=8, initial::Bool=true)
 
     maxwidth -= 1 # we will add "…" at the end
 
-    if T <: CategoricalValue
+    # assume that if module of T is CategoricalArrays its name is CategoricalValue
+    # the T is an instance of CategoricalValue from CategoricalArrays.jl
+    if hasproperty(T, :name) &&
+        nameof(T.name.module) === :CategoricalArrays &&
+        T.name.name === :CategoricalValue
```
quinnj, line 126: Do we not care if the `T.parameters[1] <: AbstractString` here?

quinnj (approved):
LGTM; hacky, but whatever, it does the job. The alternative route is we put something in DataAPI.jl like `iscategorical(x)` or something that CategoricalArrays could overload.

src/abstractdataframe/show.jl
bkamins: bacause we are replacing the behaviour for `CategoricalValue{<:AbstractString}` defined earlier in line https://github.com/JuliaData/DataFrames.jl/pull/2427/files#diff-38a6c4f0cb7743ccc6d4e8470ebdbd11L69. I just want to ensure I am matching functionality 1-2-1 exactly.

bkamins:
> The alternative route is we put something in DataAPI.jl like `iscategorical(x)` or something that CategoricalArrays could overload.

I was also thinking about it, but I think it is best for @nalimilan to decide what is best. In particular if we have `iscategorical` in `DataAPI.jl` we probably should handle both: `CategoricalValue` (for use cases like here) and `CategoricalArray` (to allow e.g. using it in /src/dataframerow/utils.jl to fire a fast path for `CategoricalVector` without having to change the code much - AFAICT @nalimilan is working on it now).

nalimilan:
Mmm, this is definitely very hacky... I see two other possibly clearner solutions:
- use an `IOContext` property that `CategoricalValue` would dected
- use `print` everywhere

I wonder what the drawbacks of using `print` would be. Besides omitted quotes when printing string and `:` when printing symbols, it also represents `Date("2020-09-13")` as `2020-09-13` and `Day(1)` as `1 day`. Maybe that's what we're actually looking for? It's not clear why we do something special for strings but not for other types. What do you think?

bkamins:
If we are at it I would actually prefer to switch to `print` if everyone is OK with this (I thought there was some reason to use `show` but maybe there was none). PrettyTables.jl uses `print`.

If we use `print` we do not need to be hacky here.

bkamins:
OK - I have switched `show` to `print`. I think we are OK to do this as we show `eltype` in the header (so e.g. not escaping `Char` is OK as you see in the changed tests).

@rofinn - could you please have a look at it as you use `print` in PrettyTables.jl

bkamins:
CC @ronisbr not @rofinn - sorry for a typo.

bkamins:
TODO. Check what is printed with this PR if you call:
```
julia> DataFrame(x=[true,false], y=[0x1, 0x2])
2×2 DataFrame
│ Row │ x    │ y     │
│     │ Bool │ UInt8 │
├─────┼──────┼───────┤
│ 1   │ 1    │ 0x01  │
│ 2   │ 0    │ 0x02  │
```
(this is the output from legacy code)

matthieugomez:
Stata prints strings without quotes but, for clarity, it also uses a different color (red). Maybe that would be useful to do the same.

bkamins:
CC @ronisbr, as this PR will not get merged. #2429 is the way to go, and when that is merged I will close this. I keep it open only to make sure #2429 correctly handles CategoricalArrays.jl without introducing a dependency.

bkamins:
@nalimilan: given the discussion with @ronisbr we should merge this PR anyway (we need this to get rid of CategoricalArrays.jl dependency for HTML and LaTeX even if we start using PrettyTables.jl for text/plain)

src/abstractdataframe/show.jl
```diff
@@ -113,7 +91,9 @@ function compacttype(T::Type, maxwidth::Int=8, initial::Bool=true)
 
     maxwidth -= 1 # we will add "…" at the end
 
-    if T <: CategoricalValue
+    # This is only type display shortening so we
+    # are OK with any T whose name starts with Categorical here
+    if startswith(sT, "Categorical")
```
nalimilan, line 96: We only need to handle `CategoricalValue`, right. so maybe better strictly check for `"CategoricalValue"`? Abbreviating other types could be weird.

bkamins: OK - I have written it when `CategoricalString` was present.

bkamins, line 96:
```suggestion
    if startswith(sT, "CategoricalValue")
```

src/abstractdataframe/show.jl
```diff
@@ -28,47 +28,25 @@ end
...
 
-Render a value to an `IO` object compactly and omitting type information, by
-calling 3-argument `show`, or 2-argument `show` if the former contains line breaks.
-Unlike `show`, render strings without surrounding quote marks.
+Render a value to an `IO` object compactly using print.
 `truncstring` indicates the approximate number of text characters width to truncate
 the output (if it is a non-positive value then no truncation is applied).
 """
-function ourshow(io::IO, x::Any, truncstring::Int; styled::Bool=false)
+function ourshow(io::IO, x::T, truncstring::Int; styled::Bool=false) where T
```
nalimilan, line 35: `T` isn't used right?

bkamins: ah - yes; it was earlier, but we ended up not using it. Fixing.

bkamins, line 35:
```suggestion
function ourshow(io::IO, x::Any, truncstring::Int; styled::Bool=false)
```

src/abstractdataframe/show.jl
```diff
@@ -113,7 +91,9 @@ function compacttype(T::Type, maxwidth::Int=8, initial::Bool=true)
 
     maxwidth -= 1 # we will add "…" at the end
 
-    if T <: CategoricalValue
+    # This is only type display shortening so we
+    # are OK with any T whose name starts with Categorical here
```
nalimilan, line 95:
```suggestion
    # are OK with any T whose name starts with CategoricalValue here
```

CI fails due to `CategoricalValue` not being abbreviated to `Cat`. Not sure why.

bkamins:
Now I remember why I used `"Categorical"` not `"CategoricalValue"` :smile:. It is because CategoricalArrays.jl might not be in scope in which case `CategoricalArray.CategoricalValue` is `eltype` name. I have added `||` now to handle this case.

src/abstractdataframe/show.jl
```diff
@@ -113,7 +91,9 @@ function compacttype(T::Type, maxwidth::Int=8, initial::Bool=true)
 
     maxwidth -= 1 # we will add "…" at the end
 
-    if T <: CategoricalValue
+    # This is only type display shortening so we
+    # are OK with any T whose name starts with CategoricalValue here
+    if startswith(sT, "CategoricalValue") || startswith(sT, "CategoricalArray.CategoricalValue")
```
nalimilan, line 96:
```suggestion
    if startswith(sT, "CategoricalValue") || startswith(sT, "CategoricalArrays.CategoricalValue")
```

nalimilan (approved):

PR #2430, opened 2020-09-14
make schema(df) faster by using less allocation

OkonSamuel:
```julia
julia> using DataFrames, Tables, BenchmarkTools

julia> X = DataFrame(randn(1000, 10000));

julia> @btime Tables.schema($X); #master
  10.189 ms (9505 allocations: 774.17 KiB)

julia> @btime Tables.schema($X); # this PR
  1.336 ms (12 allocations: 469.45 KiB)
```

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -21,7 +21,7 @@ end
 Tables.columnindex(df::Union{AbstractDataFrame, DataFrameRow}, idx::AbstractString) =
     columnindex(df, Symbol(idx))
 
-Tables.schema(df::AbstractDataFrame) = Tables.Schema(propertynames(df), eltype.(eachcol(df)))
+Tables.schema(df::AbstractDataFrame) = Tables.Schema{Tuple(_names(df)), Tuple{[eltype(col) for col in eachcol(df)]...}}()
 Tables.materializer(df::AbstractDataFrame) = DataFrame
 
 Tables.getcolumn(df::AbstractDataFrame, i::Int) = df[!, i]


quinnj (approved):
somethingisweirdandgithubwontletmetypeanynonlettercharacterssorry

src/other/tables.jl
```diff
@@ -21,7 +21,7 @@ end
 Tables.columnindex(df::Union{AbstractDataFrame, DataFrameRow}, idx::AbstractString) =
     columnindex(df, Symbol(idx))
 
-Tables.schema(df::AbstractDataFrame) = Tables.Schema(propertynames(df), eltype.(eachcol(df)))
+Tables.schema(df::AbstractDataFrame) = Tables.Schema{Tuple(_names(df)), Tuple{[eltype(col) for col in eachcol(df)]...}}()
```
nalimilan, line 24: I'm a bit late, but wouldn't a generator be even faster?
```julia
Tuple{(eltype(col) for col in eachcol(df))...}
```

bkamins: It seems it would be slower (I think I already checked it back then):
```
julia> f(y) = Tuple{[x for x in y]}

julia> g(y) = Tuple{(x for x in y)...}
g (generic function with 1 method)

julia> @time f(1:5);
  0.000007 seconds (4 allocations: 304 bytes)

julia> @time g(1:5);
  0.000012 seconds (8 allocations: 336 bytes)

julia> @time f(1:100);
  0.000011 seconds (4 allocations: 1.781 KiB)

julia> @time g(1:100);
  0.000017 seconds (106 allocations: 6.594 KiB)

julia> @time f(1:1000);
  0.000076 seconds (494 allocations: 31.562 KiB)

julia> @time g(1:1000);
  0.000115 seconds (1.99 k allocations: 85.375 KiB)
```
(and in general - probably it is a minimal overhead I think)

PR #2434, opened 2020-09-16
Remove mutating examples for adding new columns

tk3369:
The `rename` function in pandas creates a copy and so it would more appropriate to provide the equivalent DataFrames.jl example without any mutation as well. Previous discussions at https://github.com/JuliaData/DataFrames.jl/pull/2378#issuecomment-691733622

### DIFF
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -74,6 +74,8 @@ rows having the index value of `'c'`.
 | Pick & transform columns | `df.assign(x_mean = df['x'].mean())[['x_mean', 'y']]`          | `select(df, :x => mean, :y)`                |
 | Sort rows                | `df.sort_values(by = 'x')`                                     | `sort(df, :x)`                              |
 |                          | `df.sort_values(by = ['grp', 'x'], ascending = [True, False])` | `sort(df, [:grp, order(:x, rev = true)])`   |
+| Drop missing rows        | `df.dropna()`                                                  | `dropmissing(df)`                           |
+| Select unique rows       | `df.drop_duplicates()`                                         | `unique(df)`                                |
 
 Note that pandas skips `NaN` values in its analytic functions by default. By contrast,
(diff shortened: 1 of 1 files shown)


bkamins:
Agreed - can you then just add another line where the two removed Julia codes would stay and in pandas you would show how to add a column to an existing data frame (without copying)? Thank you!

tk3369:
Looks like 3 examples in that table have mutating versions. In pandas:
```python
df['z1'] = df['z'] + 1
df.rename(columns = {'x': 'x_new'}, inplace = True)
df.sort_values(by = 'x', inplace = True)
```

How about we add a new table in the same section for these?

bkamins:
OK - then we have a clearly separated "mutating" part.

tk3369:
* Added `dropmissing` and `unique` to Common Operations section
* Added Mutating Operations section

docs/src/man/comparisons.md
```diff
@@ -93,6 +93,20 @@ examples above do not synchronize the column names between pandas and DataFrames
 (you can pass `renamecols=false` keyword argument to `select`, `transform` and
 `combine` functions to retain old column names).
 
+### Mutating operations
+
+| Operation          | pandas                                                | DataFrames.jl                                |
+| :----------------- | :---------------------------------------------------- | :------------------------------------------- |
+| Add new columns    | `df['z1'] = df['z'] + 1`                              | `df.z1 = df.z .+ 1`                          |
+|                    |                                                       | `transform!(df, :z => (x -> x .+ 1) => :z1)` |
```
bkamins, line 101: maybe add another row showing how to use `insertcols!`, e.g. `insertcols!(df, 2, :const=1)` to insert a column `:const` as a second column of a data frame that consists of `1`s.

tk3369: Sure, updated PR.

bkamins (approved):

bkamins:
Thank you!

PR #2442, opened 2020-09-19
[BREAKING] Use DataAPI.refpool for optimized grouping

nalimilan:
Generalize existing optimized `row_group_slots` method for `CategoricalArray` and `PooledArray` so that it can be used for other array types for which `DataAPI.refpool` returns an `AbstractVector`. This allows dropping the dependency on CategoricalArrays in this part of the code.

Also refactor the method to be faster when not sorting. In that case, we do not need to build a map between reference codes and groups (indexing into it is slow when the number of groups is very large). `CategoricalArray` is no longer special cased: when `sort=false`, levels are still sorted, but `missing` appears first.

Add more tests to cover weird combinations.

Some benchmarks:
```julia
df = DataFrame(x=PooledArray(rand(1:ngroups, N)), y=rand(N));

## 10M rows and 1k groups
# Current
  51.310 ms (35 allocations: 76.31 MiB)

julia> @btime groupby(df, :x, sort=true);
  62.972 ms (71 allocations: 76.34 MiB)

julia> @btime combine(groupby(df, :x), :y => sum);
  67.693 ms (216 allocations: 76.39 MiB)

julia> @btime combine(groupby(df, :x, sort=true), :y => sum);
  78.816 ms (252 allocations: 76.42 MiB)

# This PR
  43.987 ms (50 allocations: 76.35 MiB)

julia> @btime groupby(df, :x, sort=true);
  48.425 ms (55 allocations: 76.38 MiB)

julia> @btime combine(groupby(df, :x), :y => sum);
  59.312 ms (231 allocations: 76.43 MiB)

julia> @btime combine(groupby(df, :x, sort=true), :y => sum);
  63.025 ms (236 allocations: 76.46 MiB)


## 1G rows and 10M groups
# Current
julia> @time groupby(df, :x);
129.936916 seconds (7.60 M allocations: 7.907 GiB, 0.54% gc time)

julia> @time groupby(df, :x, sort=true);
163.077585 seconds (1.76 M allocations: 8.575 GiB)

julia> @time combine(groupby(df, :x), :y => sum);
165.301710 seconds (8.78 M allocations: 8.456 GiB, 0.39% gc time)

julia> @time combine(groupby(df, :x, sort=true), :y => sum);
193.600132 seconds (1.03 M allocations: 9.014 GiB)


# This PR
julia> @time groupby(df, :x);
 39.474096 seconds (96 allocations: 7.740 GiB)

julia> @time groupby(df, :x, sort=true);
134.739279 seconds (76.63 k allocations: 8.042 GiB, 0.51% gc time)

julia> @time combine(groupby(df, :x), :y => sum);
 72.631447 seconds (5.17 M allocations: 8.475 GiB)

julia> @time combine(groupby(df, :x, sort=true), :y => sum);
167.185490 seconds (293 allocations: 8.511 GiB, 0.97% gc time)
```

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -94,10 +94,20 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
 # 4) whether groups are already sorted
 # Optional `groups` vector is set to the group indices of each row (starting at 1)
 # With skipmissing=true, rows with missing values are attributed index 0.
+row_group_slots(cols::Tuple{Vararg{AbstractVector}},
+                hash::Val = Val(true),
+                groups::Union{Vector{Int}, Nothing} = nothing,
+                skipmissing::Bool = false,
+                sort::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} =
+    row_group_slots(cols, DataAPI.refpool.(cols), hash, groups, skipmissing, sort)
+
+# Generic fallback method based on open adressing hash table
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
+                         refpools::Any,
                          hash::Val = Val(true),
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool}
+                         skipmissing::Bool = false,
+                         sort::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool}
     @assert groups === nothing || length(groups) == length(cols[1])
     rhashes, missings = hashrows(cols, skipmissing)
     # inspired by Dict code from base cf. https://github.com/JuliaData/DataTables.jl/pull/17#discussion_r102481481
@@ -140,21 +150,29 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
     return ngroups, rhashes, gslots, false
 end
 
-nlevels(x::PooledArray) = length(x.pool)
-nlevels(x) = length(levels(x))
-
-function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}},
+# Optimized method for arrays for which DataAPI.refpool is defined and returns an AbstractVector
+function row_group_slots(cols::NTuple{N,<:AbstractVector},
+                         refpools::NTuple{N,<:AbstractVector},
                          hash::Val{false},
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} where N
+                         skipmissing::Bool = false,
+                         sort::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} where N
     # Computing neither hashes nor groups isn't very useful,
     # and this method needs to allocate a groups vector anyway
     @assert groups !== nothing && all(col -> length(col) == length(groups), cols)
 
+    refpools = map(DataAPI.refpool, cols)
+    foreach(refpool -> @assert(allunique(refpool)), refpools)
+
     # If skipmissing=true, rows with missings all go to group 0,
     # which will be removed by functions down the stream
-    ngroupstup = map(cols) do c
-        nlevels(c) + (!skipmissing && eltype(c) >: Missing)
+    ngroupstup = map(refpools) do refpool
+        len = length(refpool)
+        if skipmissing && eltype(refpool) >: Missing && any(ismissing, refpool)
+            return len - 1
+        else
+            return len
+        end
     end
     ngroups = prod(ngroupstup)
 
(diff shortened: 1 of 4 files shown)


src/dataframerow/utils.jl
```diff
@@ -140,21 +150,29 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} where N
+                         skipmissing::Bool = false,
+                         sort::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} where N
     # Computing neither hashes nor groups isn't very useful,
     # and this method needs to allocate a groups vector anyway
     @assert groups !== nothing && all(col -> length(col) == length(groups), cols)
 
+    refpools = map(DataAPI.refpool, cols)
+    foreach(refpool -> @assert(allunique(refpool)), refpools)
```
nalimilan, line 165: This should probably be turned into a check, with a fallback to the standard method if it fails, but I wanted to discuss that are the expectations for a `refpool`. Should we require it to contain unique values or not? @quinnj @piever

src/dataframerow/utils.jl
```diff
@@ -167,43 +185,83 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
...
+                end
+                if sort
+                    nm = missingind === nothing ? eachindex(refpool) :
+                        setdiff(eachindex(refpool), missingind)
+                    perm = sortperm(view(refpool, nm))
+                    invpermute!(view(refmap, nm .- fi .+ 1), perm)
+                end
+            elseif sort
+                # FIXME: collect is needed for CategoricalRefPool
+                perm = sortperm(collect(refpool))
```
nalimilan, line 219: Without this I get:
```julia
ERROR: MethodError: no method matching similar(::Type{Array{Int64,1}}, ::Tuple{UnitRange{Int64}})
```

I'm afraid it could be hard to fix in CategoricalArrays, since we would need to return an `OffsetArray`. Also since our `refmap` uses 1-based indexing, we actually don't want to preserve custom indices. Calling `collect` is OK in terms of API but it's a bit of a waste in terms of copying...

src/dataframerow/utils.jl
quinnj: Hmmmm....in the arrow format, duplicate values are specifically allowed in the "dict encoding" pool, though it's recommended to "normalize" them to avoid duplicates, so I'm not sure; probably a fine requirement in practice, but could potentially cause an issue for weird arrow files.

quinnj (approved):
Yay! This is awesome; love the speedups and making the code more generic with `DataAPI` calls.

src/dataframerow/utils.jl
nalimilan: OK, so I'll turn this into a simple check with a fallback to the generic method.

bkamins:
Adding [BREAKING] as it changes return value order when `sort=false` (minor but still...).

src/dataframerow/utils.jl
bkamins: I think it is not a problem as it should be very fast - right?

src/dataframerow/utils.jl
```diff
@@ -167,43 +185,83 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
...
-        refmap
-    end
-    strides = (cumprod(collect(reverse(ngroupstup)))[end-1:-1:1]..., 1)::NTuple{N,Int}
-    @inbounds for i in eachindex(groups)
-        local refs
-        let i=i # Workaround for julia#15276
-            refs = map(c -> c.refs[i], cols)
+        @inbounds for i in eachindex(groups)
+            local refs_i
+            let i=i # Workaround for julia#15276
```
bkamins, line 226: as discussed `getindex.(refs, i)` should be a bit faster I think.

src/dataframerow/utils.jl
```diff
@@ -167,43 +185,83 @@ function row_group_slots(cols::NTuple{N,<:Union{CategoricalVector,PooledVector}}
...
+    else
+        missinginds = map(refpools) do refpool
+            something(findfirst(ismissing, refpool), lastindex(refpool)+1)
+        end
+        @inbounds for i in eachindex(groups)
+            local refs_i
+            let i=i # Workaround for julia#15276
+                refs_i = map(refs, missinginds) do ref, missingind
+                    r = Int(ref[i])
+                    if skipmissing
```
bkamins, line 248: can we have `skipmissing` branch outside? In this way for the most common case of `skipmissing=false` we have an unbranching code here (or you leave it as is as you assume that the compiler should optimize this out?).

nalimilan: In my tests this doesn't make any difference to performance. In general I think LLVM is pretty good at hoisting this kind of check.

bkamins: Great - thank you!

bkamins (approved):

bkamins:
Looks good thank you - the open comments are minor. I am OK to merge the PR when you resolve them.

src/dataframerow/utils.jl
nalimilan: I just tested it, it doesn't make any difference here. That's probably because the loop does many different things.

For reference, I've also tried using a generated function, but the result isn't faster (actually even a bit slower). SIMD isn't used unless I remove the `seen[gid] = true` line (note that without the generated function we don't get SIMD even without it). I kind of expected this. Apparently scatter instructions are available on AVX512, but I don't have access to such a machine.

What's weird is that remove `seen` makes things a bit *slower* on my machine (Skylake with AVX2), despite (or because?) the use of SIMD. Maybe this code is memory-bound so SIMD doesn't make it any faster? (If removing `seen` was faster, we could have kept track of the number of groups we have seen, and when we have seen all of them switch to a loop without it.)

```julia
@generated function loop_helper(groups::AbstractVector{<: Integer},
                                seen::AbstractVector{Bool},
                                refs::NTuple{N, <: AbstractVector{<: Integer}},
                                strides::NTuple{N, Integer},
                                firstinds::NTuple{N, Integer},
                                missinginds::NTuple{N, Integer},
                                skipmissing::Bool) where N
    ex = :()
    for j in 1:N
        ex = :(begin
            $ex
            r = Int(refs[$j][i])
            missingind = missinginds[$j]
            if skipmissing
                if r == missingind
                    r = -1
                    anymissing = true
                elseif r > missingind
                    r -= 1
                end
            end
            fi = firstinds[$j]
            gid += (r-fi) * strides[$j]
        end)
    end
    quote
        @inbounds for i in eachindex(groups)
            gid = 1
            anymissing = false
            $ex
            if skipmissing && anymissing
                gid = 0
            else
                seen[gid] = true
            end
            groups[i] = gid
        end
    end
end
```

bkamins: Then let us have it as is for now. I think *skylake* is currently what we should optimize for. Thank you for looking into it.

src/dataframerow/utils.jl
nalimilan: Yeah it's not too bad. If there are many levels it's a bit silly to make a copy of them, but not a big deal. I've removed the FIXME.

nalimilan:
I've added a commit to use the generic fallback when the pools are not unique, plus a few small things.

src/dataframerow/utils.jl
```diff
@@ -140,70 +150,133 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
                          hash::Val{false},
                          groups::Union{Vector{Int}, Nothing} = nothing,
-                         skipmissing::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} where N
+                         skipmissing::Bool = false,
+                         sort::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} where N
     # Computing neither hashes nor groups isn't very useful,
     # and this method needs to allocate a groups vector anyway
     @assert groups !== nothing && all(col -> length(col) == length(groups), cols)
 
+    refpools = map(DataAPI.refpool, cols)
```
bkamins, line 164: why do you compute it there if you already were passed it in line 155?

src/dataframerow/utils.jl
```diff
@@ -140,70 +150,133 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+    # or if the pool does not contain only unique values
     # The first check ensures the computation of ngroups did not overflow.
     # The rationale for the 2 threshold is that while the fallback method is always slower,
     # it allocates a hash table of size length(groups) instead of the remap vector
     # of size ngroups (i.e. the number of possible combinations) in this method:
     # so it makes sense to allocate more memory for better performance,
     # but it needs to remain reasonable compared with the size of the data frame.
-    if prod(Int128.(ngroupstup)) > typemax(Int) || ngroups > 2 * length(groups)
+    anydups = !all(allunique, refpools)
+    if prod(Int128.(ngroupstup)) > typemax(Int) ||
```
bkamins, line 192: maybe better use `big` here instead? It should not be that expensive, and `Int128` can be overflowed in practice (7 columns of 10^6 unique values - unlikely but possible). Using `big` here should not be prohibitive I think

src/dataframerow/utils.jl
```diff
@@ -140,70 +150,133 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+    # Fall back to hashing if there would be too many empty combinations
+    # or if the pool does not contain only unique values
     # The first check ensures the computation of ngroups did not overflow.
     # The rationale for the 2 threshold is that while the fallback method is always slower,
     # it allocates a hash table of size length(groups) instead of the remap vector
     # of size ngroups (i.e. the number of possible combinations) in this method:
     # so it makes sense to allocate more memory for better performance,
     # but it needs to remain reasonable compared with the size of the data frame.
-    if prod(Int128.(ngroupstup)) > typemax(Int) || ngroups > 2 * length(groups)
+    anydups = !all(allunique, refpools)
```
bkamins, line 191: how expensive is this line?

src/dataframerow/utils.jl
```diff
@@ -140,70 +150,133 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
...
+    if prod(Int128.(ngroupstup)) > typemax(Int) ||
+       ngroups > 2 * length(groups) ||
+       anydups
+        # In the simplest case, we can work directly with the reference codes
```
bkamins, line 195: indentation

bkamins:
Looks good. I left some small remarks. Additionally - a general question - I hope that the tests cover all code paths (in particular the path when we have too many groups so we should invoke a slow path).

src/dataframerow/utils.jl
nalimilan:
```suggestion
```

src/dataframerow/utils.jl
nalimilan:
```suggestion
    if prod(big.(ngroupstup)) > typemax(Int) ||
```

src/dataframerow/utils.jl
nalimilan: This was discussed in another thread. Luckily that's only 1-2%.

nalimilan:
> Additionally - a general question - I hope that the tests cover all code paths (in particular the path when we have too many groups so we should invoke a slow path).

Yes, that's covered by the new tests (there was already one but quite limited). We don't have an example of an array for which `refpool` contains duplicates, but since the path is the same it should be OK.

src/dataframerow/utils.jl
nalimilan: I could check, and even with a version of Julia built from source to target the native CPU, I don't get SIMD on a machine that supports AVX512.

bkamins:
So - I understand we merge it once CI passes. Right?

bkamins:
Thank you!

src/dataframerow/utils.jl
nalimilan: Actually, in some extreme cases it can take a lot of time. For example, the following takes 65s currently, but only 27s if I get rid of the check. So we probably should add an API to check whether the pool may contain duplicates. For `PooledArray` and `CategoricalArray` it would always be false, for others it could either always be true or vary at runtime.
```julia
using DataFrames, PooledArrays
N = 1_000_000_000;
k = N ÷ 5
df = DataFrame(x=PooledArray(rand(1:k, N)), y=rand(N));
@time gd = groupby(df, :x);
```

quinnj: `Arrow.DictEncoded` allows duplicates and would need a runtime check

nalimilan: Yes that's why I mentioned that possibility.

matthieugomez: I wanted to use refpool for a different project, but the fact that refpool could have non-unique element made the code too complicated to be worth it. I agree it would be nice to have an API to check pools are unique.

nalimilan: See also https://github.com/JuliaData/DataFrames.jl/issues/2558.

PR #2449, opened 2020-09-22
add only

bkamins:
Following discussion on Slack I think it is good to add `only` to complement `first` and `last`.

### DIFF
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -80,6 +80,13 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
+if VERSION >= v"1.4"
+    import Base.only
+else
+    import Compact.only
+    export only
+end
+
 include("other/utils.jl")
 include("other/index.jl")
 
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -83,7 +83,7 @@ end
 if VERSION >= v"1.4"
     import Base.only
 else
-    import Compact.only
+    import Compat.only
     export only
 end
 
--- a/src/DataFrames.jl
+++ b/src/DataFrames.jl
@@ -80,7 +80,7 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
-if VERSION >= v"1.4"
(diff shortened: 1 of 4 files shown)


src/DataFrames.jl
```diff
@@ -80,6 +80,13 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
+if VERSION >= v"1.4"
+    import Base.only
+else
+    import Compact.only
```
ararslan, line 86:
```suggestion
    import Compat.only
```

bkamins: indeed - it is even more compact

bkamins:
only CI and nightly fails now

src/DataFrames.jl
```diff
@@ -80,6 +80,13 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
+if VERSION >= v"1.4"
```
ararslan, line 83: I'd probably do this as
```suggestion
if isdefined(Base, :only)  # Introduced in 1.4.0
```
Probably matters much less nowadays than it used to in the Julia 0.x days. You're welcome to ignore this suggestion if you prefer what's currently there.

bkamins: Actually I was considering this myself, but then I saw what we had for other methods. I am OK to change to what you propose.

ararslan (approved):
I was wanting this just yesterday! :smile:

bkamins:
@nalimilan - OK to merge?

src/DataFrames.jl
```diff
@@ -80,6 +80,13 @@ if VERSION < v"1.2"
     export hasproperty
 end
 
+if isdefined(Base, :only)  # Introduced in 1.4.0
```
nalimilan, line 83: I don't to prefer explicit VERSION checks, which Femtocleaner can automatically remove once we drop support for 1.4, but as you like.

nalimilan (approved):

bkamins:
Thank you! I leave it as is. When we will move to a new LTS I will have to review the whole package anyway.

PR #2452, opened 2020-09-25
Add canonical links to old docs

nalimilan:
To apply https://github.com/JuliaData/DataFrames.jl/pull/2418 to older releases.


bkamins (approved):
What is the benefit of having this?

nalimilan:
#2418 is a good change for the future, but if we don't fix old releases manually search engines will keep returning them forever. Unfortunately I'm not even sure that's enough: we may have to apply the same change to each page for each release... For example, if I run a Google search for "DataFrames.jl read CSV" in private mode, I get old docs about `readtable` as the first result, and it's in the Getting Started page. Let's see...

nalimilan:
Actually while I'm at it I've applied the change to all pages. It's quite tedious and easy to get wrong as each page needs its own URL (this is what Documenter does for latest docs). See https://github.com/JuliaData/DataFrames.jl/pull/2453.

Issue #1502, opened 2018-09-05
Precompilation of DataFrames


bkamins:
I think it would be good to put some most common cases of functions used in DataFrames.jl to the module to force their precompilation as this will improve user experience (and DataFrames.jl is probably one of the packages that new users start with). Any opinions if we should do it and how (I could manually list the cases I think are relevant and code it by hand in the worst case).

nalimilan:
Good idea. We should also do this for dependencies, in particular CategoricalArrays. A reasonable way of doing that would be to run the full test suite under SnoopCompile.jl.

quinnj:
Honestly I haven't found using `precompile` statements to be very helpful (or at least worth the effort to generate them). The problem is that `precompile` still doesn't generate machine code, so it's really just the lowering cost that you're avoiding. What _has_ been more useful is to include a simple couple calls in the top-level `__init__` function that will get called every time the module is loaded. Yes, it adds a little extra time to the initial module loading, but IMO, it can be worth it to have the penalty there instead of the first time a user calls a function.

nalimilan:
Given how big is the DataFrames API, it doesn't really sound possible to call all functions from `__init__`. At least setting up precompilation will be useful in preparation of the future Julia version where native code will be stored.

KristofferC:
+1 for `precompile` statements not helping much unless it is paired with sysimg building (and then they help tremendously, c.f the Julia REPL).

bkamins:
I think it would be worth to do it at least to understand what should be compiled and how long this compilation takes. Then I guess people who want to build a custom sysimg instead of using `.juliarc` (which is now `~/.julia/config/startup.jl` right?) also would have an easier life.

nalimilan:
Right. Let's try using SnoopCompile while running the full test suite.

bkamins:
> +1 for `precompile` statements not helping much unless it is paired with sysimg building

@KristofferC Is this still true for Julia 1.4?

Here is a comparison (I am annotating `by` with `@nospecialize` - not sure if this matters)

Without precompile:
```
julia> @time using DataFrames
  0.924471 seconds (1.39 M allocations: 88.058 MiB)

julia> df = DataFrame(rand(10,4));

julia> @time by(df, :x1, [:x2, :x3] => +);
  4.523086 seconds (15.01 M allocations: 759.657 MiB, 8.18% gc time)
```

With precompile:
```
julia> @time using DataFrames
[...]

PR #2456, opened 2020-09-27
Enable precompilation

nalimilan:
`DataFrames.precompile()` precompiles all methods that are used when running tests.
This is useful to run benchmarks like H2Oai where the time of the first run should include the time needed to specialize on a particular operation on a particular dataset, but not the time needed for the more general compilation of common parts.
It could also be useful for users who would like to generate a precompiled image.

To avoid making the package slower to precompile or load, the precompilation code is only included when actually used.

Some timings on a clean Julia 1.5 session:
```julia
# Without precompilation
julia> df = DataFrame(x=[1, 1, 2, 2], y=rand(4));

julia> @time combine(groupby(df, :x), :y => sum);
  7.392566 seconds (16.59 M allocations: 850.477 MiB, 8.03% gc time)

# With precompilation
julia> @time DataFrames.precompile()
277.652629 seconds (494.20 M allocations: 24.648 GiB, 3.38% gc time)

julia> df = DataFrame(x=[1, 1, 2, 2], y=rand(4));

julia> @time combine(groupby(df, :x), :y => sum);
  0.527538 seconds (329.43 k allocations: 17.070 MiB)
```

@jangorecki Does it sound OK to you to spend 277 seconds on precompilation before running the H2Oai benchmarks? We could precompile only the methods which are relevant for you if you prefer.

See https://github.com/JuliaData/DataFrames.jl/issues/1502.


jangorecki:
Looks good like this. Please do not restrict to functions only from db-benchmark. Code should be as general as possible. Specializing it for particular cases would be little bit like cheating. As long as there is no "startup" task in db-benchmark you will be good, and such task won't happen anytime soon :)
Eventually there is a plan for a "full workflow" task, that includes reading, processing and writing, but IMO we can precompile only when it is worth. As long as this is an option, and not installing different package it is fine.

bkamins:
@nalimilan - please let me know if we want to add it to 0.22 release goal. Thank you!

nalimilan:
I've updated the PR to include two levels of precompilation:
- a first level which takes 37s and is applied automatically when installing the package
- a second level which takes a very long time and is only applied when calling `DataFrames.precompiled` (just like the original state of the PR)


Current master:
```julia
julia> @time using DataFrames
  7.056511 seconds (3.10 M allocations: 180.010 MiB, 1.52% gc time)

julia> df = DataFrame(x=[1, 1, 2, 2], y=rand(4));

julia> @time combine(groupby(df, :x), :y => sum);
  5.426770 seconds (14.92 M allocations: 760.778 MiB, 7.97% gc time)

# New clean session
julia> @time using DataFrames
  1.370335 seconds (1.80 M allocations: 114.414 MiB, 0.91% gc time)
```

This PR:
```julia
julia> @time using DataFrames
 37.902554 seconds (3.56 M allocations: 218.251 MiB, 0.47% gc time)

julia> df = DataFrame(x=[1, 1, 2, 2], y=rand(4));

julia> @time combine(groupby(df, :x), :y => sum);
  2.215117 seconds (2.82 M allocations: 143.991 MiB, 1.27% gc time)

# New clean session
julia> @time using DataFrames
  2.200718 seconds (2.26 M allocations: 152.705 MiB, 1.47% gc time)

julia> DataFrames.precompile(true)

julia> df = DataFrame(x=[1, 1, 2, 2], y=rand(4));

julia> @time combine(groupby(df, :x), :y => sum);
  0.591102 seconds (482.27 k allocations: 24.417 MiB)
```

bkamins:
I will test it before merging but it looks excellent.

bkamins:
The downside is that we should update these codes each time we make a release - is this correct?

bkamins:
You need to update the `all=true` version because of the removal `Symbol` support in `CategoricalValue` (the case we have just fixed on master), as now:
```
julia> @time DataFrames.precompile(true)
ERROR: TypeError: in CategoricalValue, in T, expected T<:Union{AbstractChar, AbstractString, Number}, got Type{Symbol}
```

bkamins:
Results of my benchmarks (and in general I think it is really nice; the only strange thing is that `DataFrame` constructor takes longer to execute, but it is fast so it is not a problem - it might be just noise):

### This PR:
```
julia> @time using DataFrames
 25.607132 seconds (2.87 M allocations: 181.689 MiB, 0.50% gc time)

julia> @time df = DataFrame(x=[1, 1, 2, 2], y=rand(4))
  0.039932 seconds (91.70 k allocations: 5.010 MiB)
4×2 DataFrame
 Row │ x      y        
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.470576
   2 │     1  0.370083
   3 │     2  0.560629
   4 │     2  0.794492

julia> @time combine(groupby(df, :x), :y => sum)
  1.450754 seconds (2.48 M allocations: 126.870 MiB, 2.82% gc time)
2×2 DataFrame
 Row │ x      y_sum    
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.840659
   2 │     2  1.35512
```
fresh session:
```
julia> @time using DataFrames
  1.435035 seconds (1.66 M allocations: 120.235 MiB, 6.26% gc time)

julia> @time df = DataFrame(x=[1, 1, 2, 2], y=rand(4))
  0.038710 seconds (91.70 k allocations: 5.010 MiB)
4×2 DataFrame
 Row │ x      y        
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.397124
   2 │     1  0.423037
   3 │     2  0.591869
   4 │     2  0.362487

julia> @time combine(groupby(df, :x), :y => sum)
  1.414719 seconds (2.48 M allocations: 126.876 MiB, 1.54% gc time)
2×2 DataFrame
 Row │ x      y_sum    
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.820161
   2 │     2  0.954356
```

### master
```
julia> @time using DataFrames
  4.429047 seconds (2.41 M allocations: 142.747 MiB, 2.76% gc time)

julia> @time df = DataFrame(x=[1, 1, 2, 2], y=rand(4))
  0.011212 seconds (3.61 k allocations: 261.883 KiB)
4×2 DataFrame
 Row │ x      y         
     │ Int64  Float64   
─────┼──────────────────
   1 │     1  0.992411
   2 │     1  0.0578703
   3 │     2  0.745452
   4 │     2  0.201506

julia> @time combine(groupby(df, :x), :y => sum)
  3.445880 seconds (14.16 M allocations: 722.170 MiB, 7.72% gc time)
2×2 DataFrame
 Row │ x      y_sum    
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  1.05028
   2 │     2  0.946958
```
fresh session:
```
julia> @time using DataFrames
  0.684325 seconds (1.19 M allocations: 81.294 MiB)

julia> @time df = DataFrame(x=[1, 1, 2, 2], y=rand(4))
  0.011208 seconds (3.61 k allocations: 261.883 KiB)
4×2 DataFrame
 Row │ x      y         
     │ Int64  Float64   
─────┼──────────────────
   1 │     1  0.335812
   2 │     1  0.339437
   3 │     2  0.0230391
   4 │     2  0.156288

julia> @time combine(groupby(df, :x), :y => sum)
  3.572502 seconds (14.16 M allocations: 722.197 MiB, 10.75% gc time)
2×2 DataFrame
 Row │ x      y_sum    
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.675249
   2 │     2  0.179327
```

nalimilan:
> The downside is that we should update these codes each time we make a release - is this correct?

Well we don't have to, as `precompile` will fail silently if the method doesn't exist. But it would be better to update it after making large changes, otherwise precompilation will have less effect.

> You need to update the `all=true` version because of the removal `Symbol` support in `CategoricalValue` (the case we have just fixed on master), as now:

Ah right, I had run this with the old version waiting for the test fix. I've pushed a commit to fix it.

I'm not sure why the constructor is slightly slower with precompilation. Maybe the method table is larger, or the GC has more work to do. I guess it's still worth it since taking 0.02s longer isn't really noticeable for users.

bkamins:
> I guess it's still worth it since taking 0.02s longer isn't really noticeable for users.

agreed

bkamins:
Another test with `precompile(true)`. As you can see it has an impact on time of constructor execution

### First run

```
julia> @time using DataFrames
 24.461958 seconds (2.86 M allocations: 180.183 MiB, 0.46% gc time)

julia> @time DataFrames.precompile(true)
184.200906 seconds (429.83 M allocations: 21.758 GiB, 4.35% gc time)

julia> @time df = DataFrame(x=[1, 1, 2, 2], y=rand(4))
  0.011488 seconds (180 allocations: 32.188 KiB)
4×2 DataFrame
 Row │ x      y        
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.11737
   2 │     1  0.340595
   3 │     2  0.304888
   4 │     2  0.210472

julia> @time combine(groupby(df, :x), :y => sum)
  0.433025 seconds (482.27 k allocations: 24.415 MiB)
2×2 DataFrame
 Row │ x      y_sum    
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.457965
   2 │     2  0.51536

```

### Second run

```
julia> @time using DataFrames
  1.390826 seconds (1.65 M allocations: 118.753 MiB, 6.34% gc time)

julia> @time df = DataFrame(x=[1, 1, 2, 2], y=rand(4))
  0.127018 seconds (91.68 k allocations: 5.008 MiB, 57.73% gc time)
4×2 DataFrame
 Row │ x      y         
     │ Int64  Float64   
─────┼──────────────────
   1 │     1  0.0532204
   2 │     1  0.35766
   3 │     2  0.936712
   4 │     2  0.842535

julia> @time combine(groupby(df, :x), :y => sum)
  1.387860 seconds (2.48 M allocations: 126.842 MiB, 1.79% gc time)
2×2 DataFrame
 Row │ x      y_sum   
     │ Int64  Float64 
─────┼────────────────
   1 │     1  0.41088
   2 │     2  1.77925
```

### Third run

```
julia> @time using DataFrames
  1.393187 seconds (1.65 M allocations: 118.755 MiB, 6.31% gc time)

julia> @time df = DataFrame(x=[1, 1, 2, 2], y=rand(4))
  0.126041 seconds (91.68 k allocations: 5.008 MiB, 58.16% gc time)
4×2 DataFrame
 Row │ x      y        
     │ Int64  Float64  
─────┼─────────────────
   1 │     1  0.875997
   2 │     1  0.648122
   3 │     2  0.961639
   4 │     2  0.386705

julia> @time combine(groupby(df, :x), :y => sum)
  1.450556 seconds (2.48 M allocations: 126.842 MiB, 1.96% gc time)
2×2 DataFrame
 Row │ x      y_sum   
     │ Int64  Float64 
─────┼────────────────
   1 │     1  1.52412
   2 │     2  1.34834
```

src/other/precompile.jl
```diff
@@ -0,0 +1,3532 @@
...
+# using SnoopCompileCore
+# inf_timing = @snoopi tmin=0.01 include("test/runtests.jl")
+# using SnoopCompile
+# pc = SnoopCompile.parcel(inf_timing)
+# SnoopCompile.write("src/other/precompile_tmp.jl", pc[:DataFrames], always=true)
+function precompile(all=false)
+    if !all
+        ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
+
+        Base.precompile(Tuple{typeof(DataFrames._combine_tables_with_first!),DataFrame,Tuple{CategoricalArrays.CategoricalArray{String,1,UInt32,String,CategoricalArrays.CategoricalValue{String,UInt32},Union{}}},Array{Int,1},Int,Int,Function,GroupedDataFrame{DataFrame},Nothing,Tuple{Symbol},Val{true}})
```
bkamins, line 14: we have many precompilations including CategoricalArrays.jl. I understand they will have to be removed when we drop CategoricalArrays.jl dependency in 1.0. Right?

So maybe better drop them now?

nalimilan: Yeah, why not. It's a bit sad that we cannot precompile these without depending on CategoricalArrays, but...

bkamins:
The benefit of removing CategoricalArrays.jl precompilation is ~2 sec faster first precompile and 0.1 sec faster startup

src/other/precompile.jl
```diff
@@ -0,0 +1,3216 @@
+# precompile.jl contains precompilation directives for all methods whose compilation
+# is triggered by running DataFrames tests and takes more than a given threshold
+# For all=true, the threshold is 0.01, for all=false it is 0.2.
+# `precompile` calls are generated using SnoopCompile via:
+# using SnoopCompileCore
+# inf_timing = @snoopi tmin=0.01 include("test/runtests.jl")
+# using SnoopCompile
+# pc = SnoopCompile.parcel(inf_timing)
+# SnoopCompile.write("src/other/precompile_tmp.jl", pc[:DataFrames], always=true)
```
bkamins, line 9: maybe pls. add comments how you manually tweaked it so that later it is easier to repeat the process? Thank you!

bkamins (approved):

bkamins:
After this is merged I will have to start using Revise.jl for development :smile:.

bkamins:
@nalimilan - do we have PrettyTables.jl related precompilations in your list? (hopefully yes - just making sure as it is relevant to reduce "time to first plot" in data frame display)

bkamins:
Another thing (sorry for random comments) - we probably do not need to precompile the deprecated functionality. Avoiding this will save some time in initial precompilation and load time.

nalimilan:
> @nalimilan - do we have PrettyTables.jl related precompilations in your list? (hopefully yes - just making sure as it is relevant to reduce "time to first plot" in data frame display)

Yes there seem to be `precompile` calls for `show` on all kinds of objects defined in the package (`DataFrame`, `DataFrameRow`, `SubDataFrame`, etc.). Though IIUC only precompiled code for methods in the package will be saved: the parts that live in PrettyTables and are not inlined will have to be compiled from scratch again. We could add precompilation to PrettyTables for these.

> Another thing (sorry for random comments) - we probably do not need to precompile the deprecated functionality. Avoiding this will save some time in initial precompilation and load time.

Have you spotted deprecated methods in the list? I've removed CategoricalArrays uses, and I can't see `by` nor `aggregate`. I think there's one call for a deprecated constructor, but that's not a big deal (I can remove it if I push another commit).

> After this is merged I will have to start using Revise.jl for development smile.

Yes we're going to be the first victims of this. :-p Maybe there should be a way to disable precompilation when doing development.

bkamins:
> We could add precompilation to PrettyTables for these.

I have thought about this also. @ronisbr - you can consider having precompilation in PrettyTables.jl to decrease "time to first table output" latency.

src/other/precompile.jl
```diff
@@ -0,0 +1,3219 @@
...
+        Base.precompile(Tuple{typeof(DataFrames._combine_with_first),NamedTuple{(:x1,),Tuple{Int}},Function,GroupedDataFrame{DataFrame},Tuple{Array{Union{Missing, Rational{Int}},1}},Val{false},Array{Int,1}})
+        Base.precompile(Tuple{typeof(Base.Broadcast.broadcasted),Function,Base.Broadcast.Broadcasted{DataFrames.DataFrameStyle,Nothing,typeof(+),Tuple{DataFrame,DataFrame}},Array{Float64,2}})
+        Base.precompile(Tuple{Type{DataFrame},Array{UnitRange,1},Symbol})
+        Base.precompile(Tuple{typeof(DataFrames._sortperm),SubDataFrame{DataFrame,DataFrames.Index,Array{Int,1}},Base.Sort.MergeSortAlg,DataFrames.DFPerm{Base.Order.ForwardOrdering,Tuple{SubArray{Union{Missing, String},1,PooledArrays.PooledArray{Union{Missing, String},UInt8,1,Array{UInt8,1}},Tuple{Array{Int,1}},false},SubArray{Union{Missing, String},1,Array{Union{Missing, String},1},Tuple{Array{Int,1}},false},SubArray{Union{Missing, String},1,Array{Union{Missing, String},1},Tuple{Array{Int,1}},false}}}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.DataFrame!)),NamedTuple{(:x, :y, :copycols),Tuple{Array{Int,1},Array{Int,1},Bool}},typeof(DataFrame!)})
```
bkamins, line 258: here is an example of `DataFrame!`. It is not the end of the world to have these deprecations. We will simply have to remove them later. The general procedure should be to remove test/deprecated.jl from test suite for generation of precompile statements.

I think this is minor so we can leave it as is for now.

nalimilan: Yeah, as you want. If you spot more like these then maybe it's worth dropping them.

bkamins: OK - I will try to mark them (I will do suggestions)

bkamins, line 258:
```suggestion
```

src/other/precompile.jl
```diff
@@ -0,0 +1,3219 @@
...
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.combine)),NamedTuple{(:ungroup,),Tuple{Bool}},typeof(combine),GroupedDataFrame{DataFrame},Pair{Symbol,Pair{typeof(identity),Symbol}}})
+        Base.precompile(Tuple{typeof(DataFrames._combine_rows_with_first!),NamedTuple{(:x1,),Tuple{Int}},Tuple{Array{Union{Missing, Int},1}},Int,Int,Function,GroupedDataFrame{DataFrame},Tuple{Array{Union{Missing, Int},1}},Tuple{Symbol},Val{false}})
+        Base.precompile(Tuple{typeof(copy),DataFrameRow{DataFrame,DataFrames.SubIndex{DataFrames.Index,Array{Int,1},Array{Int,1}}}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:x1,),Tuple{Array{Int,1}}},Type{DataFrame}})
+        Base.precompile(Tuple{typeof(Base.Broadcast.materialize),Base.Broadcast.Broadcasted{DataFrames.DataFrameStyle,Nothing,typeof(+),Tuple{SubDataFrame{DataFrame,DataFrames.Index,Base.OneTo{Int}},Int}}})
+        Base.precompile(Tuple{typeof(DataFrames.row_group_slots),Tuple{Array{String,1},PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}}},Val{false},Array{Int,1},Bool,Bool})
+        Base.precompile(Tuple{typeof(DataFrame!),Dict{String,Any}})
```
bkamins, line 1070:
```suggestion
```

src/other/precompile.jl
```diff
@@ -0,0 +1,3219 @@
...
+        Base.precompile(Tuple{DataFrames.Reduce{typeof(Base.add_sum),Nothing,Nothing},Array{Union{Missing, Bool},1},GroupedDataFrame{DataFrame}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:a, :b, :c),Tuple{Array{Int,1},Array{Array{Int,1},1},Array{Array{Int,1},1}}},Type{DataFrame}})
+        Base.precompile(Tuple{typeof(stack),DataFrame,InvertedIndex{InvertedIndex{InvertedIndex{Regex}}}})
+        Base.precompile(Tuple{DataFrames.Aggregate{typeof(Statistics.var),Nothing},Array{Union{Missing, Complex{Float64}},1},GroupedDataFrame{DataFrame}})
+        Base.precompile(Tuple{typeof(getindex),DataFrame,Colon,UnitRange{Int}})
+        Base.precompile(Tuple{typeof(show),IOContext{Base.GenericIOBuffer{Array{UInt8,1}}},GroupedDataFrame{DataFrame}})
+        Base.precompile(Tuple{DataFrames.Reduce{typeof(Base.mul_prod),Nothing,Nothing},Array{Union{Missing, Float64},1},GroupedDataFrame{DataFrame}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:e,),Tuple{Array{String,1}}},Type{DataFrame}})
+        Base.precompile(Tuple{typeof(DataFrame!),DataFrame})
```
bkamins, line 1327:
```suggestion
```

src/other/precompile.jl
```diff
@@ -0,0 +1,3219 @@
...
+        isdefined(DataFrames, Symbol("#67#74")) && Base.precompile(Tuple{getfield(DataFrames, Symbol("#67#74")),Array{Any,1}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:id, :fid, :id_1),Tuple{Array{Union{Missing, Int},1},Array{Int,1},Array{Union{Missing, Int},1}}},Type{DataFrame}})
+        Base.precompile(Tuple{typeof(view),DataFrame,UnitRange{UInt},Colon})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.rightjoin)),NamedTuple{(:on,),Tuple{Array{Pair{String,String},1}}},typeof(rightjoin),DataFrame,DataFrame})
+        Base.precompile(Tuple{typeof(DataFrames._combine_with_first),NamedTuple{(:x1,),Tuple{Float64}},Function,GroupedDataFrame{DataFrame},Tuple{Array{Union{Missing, Int, Int8},1}},Val{false},Array{Int,1}})
+        Base.precompile(Tuple{typeof(push!),DataFrame,Tuple{Int,String,Int}})
+        Base.precompile(Tuple{typeof(DataFrames.genkeymap),GroupedDataFrame{DataFrame},Tuple{Array{Symbol,1},Array{Int,1}}})
+        Base.precompile(Tuple{typeof(findall),Function,DataFram
```
bkamins, line 1991:
```suggestion
```

bkamins, line 2397:
```suggestion
```

bkamins, line 2980:
```suggestion
```

bkamins, line 1991:
```suggestion
```

bkamins: for some reason GitHub presents an incorrect text but in a correct line. So 😕.

bkamins: same here - GitHub messes up something

bkamins, line 3048:
```suggestion
```

bkamins: same here

bkamins, line 2596: this is deprecated
```suggestion
```

nalimilan: This one (line 1594) seems to be the real
```julia
         Base.precompile(Tuple{typeof(findall),Function,DataFrames.DataFrameColumns{DataFrame}})
```
right?

bkamins, line 2231: this is deprecated
```suggestion
```

src/other/precompile.jl
```diff
@@ -0,0 +1,3219 @@
...
+        Base.precompile(Tuple{typeof(isequal),Array{Pair{DataFrames.GroupKey{GroupedDataFrame{DataFrame}},SubDataFrame{DataFrame,DataFrames.Index,Array{Int,1}}},1},Array{Pair{DataFrames.GroupKey{GroupedDataFrame{DataFrame}},SubDataFrame{DataFrame,DataFrames.Index,Array{Int,1}}},1}})
+        Base.precompile(Tuple{typeof(DataFrames._combine_tables_with_first!),DataFrame,Tuple{Array{Int,1},Array{String,1}},Array{Int,1},Int,Int,Function,GroupedDataFrame{DataFrame},Nothing,Tuple{Symbol,Symbol},Val{true}})
+        Base.precompile(Tuple{DataFrames.Reduce{typeof(Base.add_sum),Nothing,Nothing},Array{Complex{Int},1},GroupedDataFrame{DataFrame}})
+        Base.precompile(Tuple{Type{DataFrame},Array{Pair{Symbol,Int},1}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:a, :b),Tuple{Array{Int,1},Array{Array{Int,1},1}}},Type{DataFrame}})
+        Base.precompile(Tuple{typeof(combine),GroupedDataFrame{DataFrame},Array{Pair{typeof(nrow),Symbol},1}})
+        Base.precompile(Tuple{Type{DataFrame},Base.ReshapedArray{Float64,2,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},Tuple{}},Symbol})
```
bkamins, line 174:
```suggestion
```

src/other/precompile.jl
```diff
@@ -0,0 +1,3219 @@
...
+        Base.precompile(Tuple{DataFrames.Reduce{typeof(Base.add_sum),Nothing,Nothing},Array{BigInt,1},GroupedDataFrame{DataFrame}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:c, :b, :d, :a, :e),Tuple{Array{String,1},Array{Float64,1},Array{Bool,1},Array{Char,1},Array{String,1}}},Type{DataFrame}})
+        Base.precompile(Tuple{typeof(DataFrames._combine_process_pair_symbol),Bool,GroupedDataFrame{DataFrame},Dict{Symbol,Tuple{Bool,Int}},Array{DataFrames.TransformationResult,1},Nothing,Symbol,Bool,NamedTuple{(:z,),Tuple{Int}},Union{Function, Type},Tuple{Array{Int,1}}})
+        Base.precompile(Tuple{typeof(DataFrames.do_call),typeof(prod),Array{Int,1},Array{Int,1},Array{Int,1},GroupedDataFrame{DataFrame},Tuple{Array{Any,1}},Int})
+        Base.precompile(Tuple{Type{DataFrame},Dict{Symbol,UnitRange{Int}}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.manipulate)),NamedTuple{(:copycols, :keeprows, :renamecols),Tuple{Bool,Bool,Bool}},typeof(DataFrames.manipulate),DataFrame,Pair{Array{Symbol,1},typeof(-)}})
+        Base.precompile(Tuple{Type{DataFrame},Array{Union{Missing, Float64},2}})
```
bkamins, line 285:
```suggestion
```

src/other/precompile.jl
```diff
@@ -0,0 +1,3219 @@
...
+        Base.precompile(Tuple{typeof(DataFrames.update_row_maps!),DataFrame,DataFrame,DataFrames.RowGroupDict{DataFrame},DataFrames.RowIndexMap,Nothing,DataFrames.RowIndexMap,Nothing,Tuple{Array{Any,1},Array{Int,1}},Tuple{Array{Any,1},Array{Int,1}}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.show)),NamedTuple{(:allrows, :allcols, :rowlabel, :summary, :eltypes, :truncate),Tuple{Bool,Bool,Symbol,Bool,Bool,Int}},typeof(show),Base.PipeEndpoint,DataFrame})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.select)),NamedTuple{(:copycols, :keepkeys, :ungroup, :renamecols),NTuple{4,Bool}},typeof(select),GroupedDataFrame{DataFrame},Function,Pair{Symbol,typeof(sum)},Vararg{Any,N} where N})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:a, :b),Tuple{UnitRange{Int},Int}},Type{DataFrame}})
+        Base.precompile(Tuple{Core.kwftype(typeof(DataFrames.Type)),NamedTuple{(:id1, :id2, :x_left, :x_left_1),NTuple{4,UnitRange{Int}}},Type{DataFrame}})
+        Base.precompile(Tuple{Type{Array{T,2} where T},DataFrame})
+        Base.precompile(Tuple{Type{DataFrame},Tuple{UnitRange{Int},UnitRange{Int}}})
```
bkamins, line 514:
```suggestion
```

bkamins:
I am done adding the suggestions. GitHub messes up the display, but I hope it will be clear for you what I suggest to remove.

src/other/precompile.jl
nalimilan: I didn't expect you'd find so many of them. I should have looked more carefully. :-)

nalimilan:
I've just realized that this should probably be tested, to avoid making coverage too low, and because it could break (e.g. when changing types). It takes about 100s in a session after running tests. I guess that's OK?

bkamins:
> I guess that's OK?

It is OK. We have ~15 minutes testing time anyway. It takes 180 seconds on my machine to test it.

Please add this test to a separate file (so that it can be easily disabled when developing).

Thank you!

ronisbr:
I really want to add precompilation in PrettyTables.jl, but I might need a help, since I have never done something like this for a package (except for small tests).

nalimilan:
> It is OK. We have ~15 minutes testing time anyway. It takes 180 seconds on my machine to test it.
> 
> Please add this test to a separate file (so that it can be easily disabled when developing).

@bkamins OK done.

> I really want to add precompilation in PrettyTables.jl, but I might need a help, since I have never done something like this for a package (except for small tests).

@ronisbr I'm not too familiar either, I just took inspiration from [FixedPointNumbers](https://github.com/JuliaMath/FixedPointNumbers.jl/pull/148/) and [SnoopCompile docs](https://timholy.github.io/SnoopCompile.jl/stable/snoopi/). You can also use the same sequence of calls that I put in a comment at the top of src/other/precompile.jl in this PR.

FWIW, here's the `precompile` calls that SnoopCompile generated for PrettyTables when running the DataFrames tests (with a time threshold of 0.01 or 0.02, I don't remember exactly). You'll probably get similar ones with PrettyTables tests, but if not it could make sense to include them anyway to make DataFrames faster.
```julia
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Bool})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Char})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Float64})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Int32})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Int64})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Irrational{:π}})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Markdown.MD})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),String})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),Symbol})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),UndefInitializer})
    Base.precompile(Tuple{Core.kwftype(typeof(PrettyTables._parse_cell_text)),NamedTuple{(:autowrap, :cell_data_type, :cell_first_line_only, :column_width, :compact_printing, :has_color, :linebreaks, :renderer),Tuple{Bool,DataType,Bool,Int64,Bool,Bool,Bool,Val{:print}}},typeof(PrettyTables._parse_cell_text),UnitRange{Int64}})
    Base.precompile(Tuple{typeof(PrettyTables._fill_row_name_column!),Array{String,2},Array{Int64,2},Array{Array{String,1},2},Array{Array{Int64,1},2},Array{Int64,1},Array{String,1},Array{Int64,1},Array{Int64,1},Int64,Bool,Val{:print},String})
    Base.precompile(Tuple{typeof(PrettyTables._print_table_data),IOContext{Base.GenericIOBuffer{Array{UInt8,1}}},PrettyTables.Display,PrettyTables.ColumnTable,Array{Array{String,1},2},Array{Array{Int64,1},2},Array{Int64,1},Array{Int64,1},Int64,Array{Int64,1},Array{Int64,1},Array{Union{NTuple{4,Int64}, Symbol},1},Array{Union{Int64, Symbol},1},Array{Symbol,1},NTuple{4,Char},Tuple{},Symbol,Int64,Tuple{PrettyTables.Highlighter},Bool,Bool,Bool,PrettyTables.TextFormat,Crayons.Crayon,Crayons.Crayon,Crayons.Crayon})
    Base.precompile(Tuple{typeof(PrettyTables._print_table_header!),IOContext{Base.GenericIOBuffer{Array{UInt8,1}}},PrettyTables.Display,SubArray{String,1,Array{String,1},Tuple{StepRange{Int64,Int64}},true},Array{String,2},Array{Int64,2},Array{Int64,1},Array{Int64,1},Int64,Int64,Array{Int64,1},Array{Int64,1},Array{Symbol,1},Array{Symbol,1},Tuple{},Bool,Bool,PrettyTables.TextFormat,Crayons.Crayon,Array{Crayons.Crayon,1},Array{Crayons.Crayon,1},Crayons.Crayon,Crayons.Crayon})
    Base.precompile(Tuple{typeof(copy),Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1},Tuple{Base.OneTo{Int64}},typeof(PrettyTables.compact_type_str),Tuple{Array{Any,1}}}})
```

ronisbr:
@nalimilan normally precompile statements are divided by Julia version and architecture in the packages I looked in the past. Why this is not required here?

bkamins:
> Why this is not required here?

I would say it would be desirable, but it is simply hard do to. We removed 32-bit specific precompilation statements and things that fail on Julia 1.0 manually.

Indeed some things might fail on e.g. Julia 1.3 which we do not check, but if we get such a report I will quickly release patch.

Let us hope that for 1.0 release we will have a more robust workflow for this.

nalimilan:
I'm not super familiar with precompilation, but if you look at the FixedPointNumbers example that's not the case. AFAICT the only issue with changing Julia versions is that the `precompile` calls may use unexported Base functions (my last commit), but in theory this could be fixed by dropping them automatically (or checking that they exist).

BTW, I've just noticed that Plots.jl uses [CompileBot](https://github.com/aminya/CompileBot.jl/), which appears to automate what I did here. I guess we could use that in the future (not necessarily via a GitHub Action as running locally works too).

ronisbr:
@bkamins and @nalimilan one question. I added precompilation locally here to PrettyTables.jl and I could reduce the time to render the first table from 1.8s to 1.0s. However, the loading time (`using`), increase from 0.16 to 0.38. Is it ok?

bkamins:
I guess it is OK as it reduces total time to first plot :smile:. The question is by how much `using DataFrames` time is increased.

ronisbr:
I think it is probably linear, like, it will increase 0,22s here.

bkamins:
Not necessarily. From my experience there can be some differences as packages might have the same dependencies for instance.

ronisbr:
Ok! Let me try.

ronisbr:
Without precompilation of PrettyTables:

1. Load time: 1.21s
2. Time to first table: 3.02s

With precompilation of PrettyTables:

1. Load time: 1.44s
2. Time to first table: 2.65s

bkamins:
yeah - it is roughly the same

nalimilan:
Yeah it doesn't seem great. What did you include in the precompilation phase?

ronisbr:
> Yeah it doesn't seem great. What did you include in the precompilation phase?

I only used the default options of CompileBot. Maybe should I use SnoopCompile directly?

nalimilan:
I've never used CompileBot. Did it generate many `precompile` statements? You could also copy the ones I posted above, and which should particularly help DataFrames.

ronisbr:
> I've never used CompileBot. Did it generate many `precompile` statements? You could also copy the ones I posted above, and which should particularly help DataFrames.

Yes, there are a lot of statements! All those lines you pasted were also generated by CompileBot.

nalimilan:
OK. So I guess there's nothing to do at this point then. (Precompilation only stores the lowered Julia code currently, so LLVM still has to compile it for the specific CPU, and depending on cases that may be the larger share of the work.)

Issue #2457, opened 2020-09-28
ByRow without columns


matthieugomez:
Would it be possible at some point to use DataFrames `ByRow` without columns? This would allow one to do things like:
```julia
transform(df, Any[] => ByRow(rand) => :z)
```

bkamins:
It was considered.

The reason why it is not allowed is that `ByRow` is essentially a shorthand for broadcasting currently. To keep consistency here broadcasting over an empty collection would give us an empty collection. If you pass at least one column then you are good as this column defines you the dimensions of the output.

Actually your example was the only use-case I could come-up with that would be useful and I thought that it is not worth to break the design to handle it.

Technically it would be possible to work around this and add this support but I was not sure if it is worth it. What do you think?

pdeffebach:
I would like it. I was working on a problem set the other day where I wanted to generate random numbers for error terms. 

This actually seems quite hard to fix! we have to catch the combination of `Any[]` and `ByRow` and then use some sort sort of struct for dispatch that constains the number of rows in a DataFrame to call `[f() for i in 1:N]`

pdeffebach:
It will also make DataFramesMeta a lot easier in case you see something with the yet-to-be-implemented `@row` macro

```
@transform(df, @row y = 7)
```

This will get lowered to 

```
transform(df, [] => ByRow(_f) => :y)
```

which will error.

matthieugomez:
Yes, exactly. Ideally macro-packages will make it as easy to work row-wise
and col-wise (that's the point of PairsMacros at least), and so people do
end up writing `@rows y = 1`. I don't think that needs to be done before
1.0, but it's certainly something that would be nice to have someday.

bkamins:
I have added it to the PR fixing #2041.
In a few days I will open an issue for #2041 for `AbstractDataFrame`. Then looking at it will be welcome.
When we finalize that one I will open another PR for `GroupedDataFrame` (where we also need to change this in particular) and `filter`.

Issue #2410, opened 2020-09-04
[BREAKING] Handling of ByRow


bkamins:
In preparation for allowing producing multiple columns in `select` and `transform` we have the following decision to be made, as indicated to me by @nalimilan (I also give the options we can consider).

Preamble: currently `ByRow` produces a result "just as passed" (no pseudo-broadcasting happens). We disallowed "tabular types" like `NamedTuple` for future use (envisioning that some day we might want to allow multiple columns to be returned).

Now the problem is how to distinguish if we want to pass a `NamedTuple` as a single column vs. turn it into several columns. In pseudo-broadcasting (which does not apply here) this is achieved by e.g. `Ref` but this is not something that `ByRow` supports.

We need to be breaking to fill this gap and possible options are:
* if `Ref` or `0`-dimensional array is returned unwrap it (just like in pseudo-broadcasting)
* by default allow everything and produce a single column out of it always, but add e.g. `:cols => ByRow(fun) => AsTable` form (or something similar - to be discussed) - that would signal that the returned value is to be unwrapped into several columns

What do you think would be better?

CC @nalimilan, @pdeffebach, @matthieugomez

pdeffebach:
> * by default allow everything and produce a single column out of it always, but add e.g. `:cols => ByRow(fun) => AsTable` form (or something similar - to be discussed) - that would signal that the returned value is to be unwrapped into several columns

I had this exact idea yesterday! So I think it's a good solution. 

Something similar is coming up in DataFramesMeta with `@based_on`. 

In combine you can do `combine([:x, :y] => fun)`. 

To do similar behavior in a `@combine` macro you would write `@combined(fun(:x, :y), df)`. 

This poses a major problem for inspecting the expression since it doesn't have a set structure the way `y = :x + :z` does. I was thinking of a `@astable` flag like 

```
@combine(@astable myfunction(:a, :b))
```

So the user says "this function returns a table object". I think an `@astable` flag that would transform expressions to `[:x, :y] => fun => asTable` would be a great parallel between DataFrames and DataFramsMeta. 

I guess in this world we would also deprecate the table-as-return value in `combine`. This isn't strictly necessary for DataFrames, but might be good for consistency.

nalimilan:
Whatever solution we choose, I think we should be consistent with and without `ByRow`:
- either always require `AsTable` to return multiple columns, or never
- either always unwrap `Ref` and 0-dimensional arrays, or never

[...]

PR #2461, opened 2020-10-01
Allow multicolumn transformations for AbstractDataFrame

bkamins:
This PR partially addresses https://github.com/JuliaData/DataFrames.jl/issues/2410 and https://github.com/JuliaData/DataFrames.jl/issues/2457.

It covers `select` etc. for `AbstractDataFrame`.

If we are OK with the functionality I will update the documentation.

TODO:
* do the same for `select` etc. for `GroupedDataFrame` (this will be a separate PR to keep PRs more atomic)
* add support for `ByRow` with no columns passed to `filter` (also a separate PR)

CC @nalimilan @pdeffebach @matthieugomez - this is a rather complex PR so independent testing (especially for corner cases) would be welcome (if you would have suggestions for types of tests to add please comment and I will add them).

### DIFF
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -224,11 +224,15 @@ _expand_to_table(res::Union{AbstractDataFrame, NamedTuple, DataFrameRow, Abstrac
 function _expand_to_table(res::AbstractVector)
     isempty(res) && return Tables.columntable(res)
     kp1 = keys(res[1])
+    prepend = all(x -> x isa Integer, kp1)
+    if !(prepend || all(x -> x isa Symbol, kp1) || all(x -> x isa AbstractString, kp1))
+        throw(ArgumentError("keys of the returned elements must be " *
+                            "`Symbol`s, strings or integers"))
+    end
     if any(x -> !isequal(keys(x), kp1), res)
         throw(ArgumentError("keys of the returned elements must be identical"))
     end
     newres = DataFrame()
-    prepend = all(x -> x isa Integer, kp1)
     for n in kp1
         newres[!, prepend ? Symbol("x", n) : Symbol(n)] = [x[n] for x in res]
     end
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -373,7 +373,7 @@ function select_transform!(@nospecialize(nc::Union{Base.Callable, Pair{<:Union{I
 
     if res isa Union{AbstractDataFrame, NamedTuple, DataFrameRow, AbstractMatrix}
         if newname isa Symbol
-            throw(ArgumentError("Table returned while a single column return value was requested"))
+            throw(ArgumentError("Table returned but a single output column was expected"))
         end
         colnames = _gen_colnames(res, newname)
         isempty(colnames) && return # nothing to do
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -275,7 +275,7 @@ end
 function _add_col_check_copy(newdf::DataFrame, df::AbstractDataFrame,
                              col_idx::Union{Nothing, Int, AbstractVector{Int}, AsTable},
                              copycols::Bool, @nospecialize(fun),
-                             newname::Symbol, @nospecialize(v))
+                             newname::Symbol, v::AbstractVector)
     cdf = eachcol(df)
     vpar = parent(v)
     parent_cols = col_idx isa AsTable ? col_idx.cols : something(col_idx, 1:ncol(df))
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -195,10 +195,10 @@ function select_transform!(nc::Union{Base.Callable, Pair{<:Union{Int, AbstractVe
         tbl = Tables.columntable(select(df, col_idx.cols, copycols=false))
         if isempty(tbl) && fun isa ByRow
             if isempty(df)
-                T = Base.return_types(fun.fun, ())[1]
+                T = Base.return_types(fun.fun, (NamedTuple{(),Tuple{}},))[1]
                 res = T[]
             else
-                res = [fun.fun() for _ in 1:nrow(df)]
+                res = [fun.fun(NamedTuple()) for _ in 1:nrow(df)]
             end
         else
             res = fun(tbl)
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -68,8 +68,9 @@ normalize_selection(idx::AbstractIndex, sel::Pair{<:ColumnIndex, <:AbstractStrin
 
 function normalize_selection(idx::AbstractIndex,
                              sel::Pair{<:Any,<:Pair{<:Base.Callable,
-                             <:Union{Symbol, AbstractString, DataType,
-                                     AbstractVector{Symbol}, AbstractVector{<:AbstractString}}}},
+                                                    <:Union{Symbol, AbstractString, DataType,
+                                                            AbstractVector{Symbol},
+                                                            AbstractVector{<:AbstractString}}}},
                              renamecols::Bool)
     lls = last(last(sel))
     if lls isa DataType
@@ -169,7 +170,7 @@ function normalize_selection(idx::AbstractIndex,
     return (wanttable ? AsTable(c) : c) => fun => newcol
 end
 
-function _transformation_helper(df, col_idx, @nospecialize(fun))
+function _transformation_helper(df::AbstractDataFrame, col_idx, @nospecialize(fun))
     if col_idx === nothing
         return fun(df)
     elseif col_idx isa Int
@@ -178,7 +179,7 @@ function _transformation_helper(df, col_idx, @nospecialize(fun))
         tbl = Tables.columntable(select(df, col_idx.cols, copycols=false))
         if isempty(tbl) && fun isa ByRow
             if isempty(df)
-                T = Base.return_types(fun.fun, (NamedTuple{(),Tuple{}},))[1]
+                T = Core.Compiler.return_type(fun.fun, (NamedTuple{(),Tuple{}},))
                 return T[]
             else
                 return [fun.fun(NamedTuple()) for _ in 1:nrow(df)]
@@ -211,7 +212,7 @@ function _gen_colnames(@nospecialize(res), newname)
         colnames = propertynames(res)
     end
 
-    if !(newname === AsTable || newname === nothing)
+    if newname !== AsTable && newname !== nothing
         if length(colnames) != length(newname)
             throw(ArgumentError("Number of returned columns does not match the " *
                                 "length of requested output"))
@@ -273,7 +274,7 @@ end
 function _add_col_check_copy(df, newdf, col_idx, copycols, @nospecialize(fun), newname, @nospecialize(v))
     cdf = eachcol(df)
     vpar = parent(v)
-    parent_cols = col_idx isa AsTable ? col_idx.cols : (col_idx === nothing ? (1:ncol(df)) : col_idx)
+    parent_cols = col_idx isa AsTable ? col_idx.cols : something(col_idx, 1:ncol(df))
     if copycols && !(fun isa ByRow) && (v isa SubArray || any(i -> vpar === parent(cdf[i]), parent_cols))
         newdf[!, newname] = copy(v)
     else
@@ -313,7 +314,8 @@ function select_transform!(@nospecialize(nc::Union{Base.Callable, Pair{<:Union{I
         isempty(colnames) && return # nothing to do
 
         if any(in(transformed_cols), colnames)
-            throw(ArgumentError("Duplicate column name returned"))
+            throw(ArgumentError("Duplicate column name(s) returned: :" *
+                                "$(join(intersect(colnames, transformed_cols), ", :"))"))
         else
             startlen = length(transformed_cols)
             union!(transformed_cols, colnames)
@@ -354,7 +356,7 @@ function select_transform!(@nospecialize(nc::Union{Base.Callable, Pair{<:Union{I
             newname = :x1
         end
         if newname in transformed_cols
-            throw(ArgumentError("duplicate name of a transformed column"))
+            throw(ArgumentError("duplicate output column name: :$newname"))
         else
             push!(transformed_cols, newname)
         end
@@ -366,7 +368,7 @@ function select_transform!(@nospecialize(nc::Union{Base.Callable, Pair{<:Union{I
             newname = :x1
         end
         if newname in transformed_cols
-            throw(ArgumentError("duplicate name of a transformed column"))
+            throw(ArgumentError("duplicate output column name: :$newname"))
         else
             push!(transformed_cols, newname)
         end
@@ -535,8 +537,7 @@ select!(df::DataFrame, args...; renamecols::Bool=true) =
 
 function select!(arg::Base.Callable, df::AbstractDataFrame; renamecols::Bool=true)
     if arg isa Colon
-        throw(ArgumentError("Only transformations are allowed when function is a " *
-                            "frist argument to select!"))
+        throw(ArgumentError("First argument must be a transformation if the second argument is a data frame"))
     end
     return select!(df, arg)
 end
@@ -555,8 +556,7 @@ transform!(df::DataFrame, args...; renamecols::Bool=true) =
 
 function transform!(arg::Base.Callable, df::AbstractDataFrame; renamecols::Bool=true)
     if arg isa Colon
-        throw(ArgumentError("Only transformations are allowed when function is a " *
-                            "frist argument to transform!"))
+        throw(ArgumentError("First argument must be a transformation if the second argument is a data frame"))
     end
     return transform!(df, arg)
 end
@@ -686,8 +686,7 @@ select(df::AbstractDataFrame, args...; copycols::Bool=true, renamecols::Bool=tru
 
 function select(arg::Base.Callable, df::AbstractDataFrame; renamecols::Bool=true)
     if arg isa Colon
-        throw(ArgumentError("Only transformations are allowed when function is a " *
-                            "frist argument to select"))
+        throw(ArgumentError("First argument must be a transformation if the second argument is a data frame"))
     end
     return select(df, arg)
 end
@@ -707,8 +706,7 @@ transform(df::AbstractDataFrame, args...; copycols::Bool=true, renamecols::Bool=
 
 function transform(arg::Base.Callable, df::AbstractDataFrame; renamecols::Bool=true)
     if arg isa Colon
-        throw(ArgumentError("Only transformations are allowed when function is a " *
-                            "frist argument to transform"))
+        throw(ArgumentError("First argument to must be a transformation if the second argument is a data frame"))
     end
     return transform(df, arg)
 end
@@ -750,8 +748,7 @@ combine(df::AbstractDataFrame, args...; renamecols::Bool=true) =
 
 function combine(arg::Base.Callable, df::AbstractDataFrame; renamecols::Bool=true)
     if arg isa Colon
-        throw(ArgumentError("Only transformations are allowed when function is a " *
-                            "frist argument to combine"))
+        throw(ArgumentError("First argument to select! must be a transformation if the second argument is a data frame"))
     end
     return combine(df, arg)
 end
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -70,8 +70,9 @@ function normalize_selection(idx::AbstractIndex,
                              <:Union{Symbol, AbstractString, DataType,
                                      AbstractVector{Symbol}, AbstractVector{<:AbstractString}}}},
                              renamecols::Bool)
-    if last(last(sel)) isa DataType
-        last(last(sel)) === AsTable || throw(ArgumentError("Only DataType supported as target is AsTable"))
+    lls = last(last(sel))
+    if lls isa DataType
+        lls === AsTable || throw(ArgumentError("Only DataType supported as target is AsTable"))
     end
     if first(sel) isa AsTable
         rawc = first(sel).cols
@@ -99,15 +100,17 @@ function normalize_selection(idx::AbstractIndex,
         throw(ArgumentError("at least one column must be passed to a " *
                             "`ByRow` transformation function"))
     end
-    ls = last(sel)
-    if ls isa AbstractString
-        r = Symbol(ls)
-    elseif ls isa AbstractVector{<:AbstractString}
-        r = Symbol.(ls)
+    if lls isa AbstractString
+        r = Symbol(lls)
+    elseif lls isa AbstractVector{<:AbstractString}
+        r = Symbol.(lls)
     else
-        r = ls
+        r = lls
     end
-    return (wanttable ? AsTable(c) : c) => r
+    if r isa AbstractVector{Symbol}
+        allunique(r) || throw(ArgumentError("target column names must be unique"))
+    end
+    return (wanttable ? AsTable(c) : c) => first(last(sel)) => r
 end
 
 function normalize_selection(idx::AbstractIndex,
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -432,35 +432,41 @@ SELECT_ARG_RULES =
 
     * Any index that is allowed for column indexing
       ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+    * A function or a type
     * Column transformation operations using the `Pair` notation that is
-      described below and vectors of such pairs.
+      described below and vectors or matrices of such pairs.
 
     Columns can be renamed using the `old_column => new_column_name` syntax, and
     transformed using the `old_column => fun => new_column_name` syntax.
-    `new_column_name` must be a `Symbol` or a string, and `fun` a function or a
-    type. If `old_column` is a `Symbol`, a string, or an integer then `fun` is
-    applied to the corresponding column vector. Otherwise `old_column` can be
-    any column indexing syntax, in which case `fun` will be passed the column
-    vectors specified by `old_column` as separate arguments. The only exception
-    is when `old_column` is an `AsTable` type wrapping a selector, in which case
-    `fun` is passed a `NamedTuple` containing the selected columns.
+    `new_column_name` must be a `Symbol` or a string, a vector of `Symbol` or
+    string, or `AsTable`, and `fun` a function or a type. If `old_column` is a
+    `Symbol`, a string, or an integer then `fun` is applied to the corresponding
+    column vector. Otherwise `old_column` can be any column indexing syntax, in
+    which case `fun` will be passed the column vectors specified by `old_column`
+    as separate arguments. The only exception is when `old_column` is an
+    `AsTable` type wrapping a selector, in which case `fun` is passed a
+    `NamedTuple` containing the selected columns.
+
+    Column renaming and transformation operations can be passed wrapped in
+    vectors or matrices (this is useful when combined with broadcasting).
+
+    # Rules when `new_column_name` is a `Symbol` or a string or is missing
 
     If `fun` returns a value of type other than `AbstractVector` then it will be
     broadcasted into a vector matching the target number of rows in the data
     frame, unless its type is one of `AbstractDataFrame`, `NamedTuple`,
-    `DataFrameRow`, `AbstractMatrix`, in which case an error is thrown as
-    currently these return types are not allowed. As a particular rule, values
-    wrapped in a `Ref` or a `0`-dimensional `AbstractArray` are unwrapped and
-    then broadcasted.
+    `DataFrameRow`, `AbstractMatrix`, in which case an error is thrown. As a
+    particular rule, values wrapped in a `Ref` or a `0`-dimensional
+    `AbstractArray` are unwrapped and then broadcasted.
 
     To apply `fun` to each row instead of whole columns, it can be wrapped in a
     `ByRow` struct. In this case if `old_column` is a `Symbol`, a string, or an
     integer then `fun` is applied to each element (row) of `old_column` using
     broadcasting. Otherwise `old_column` can be any column indexing syntax, in
     which case `fun` will be passed one argument for each of the columns
-    specified by `old_column`. If `ByRow` is used it is not allowed for
-    `old_column` to select an empty set of columns nor for `fun` to return a
-    `NamedTuple` or a `DataFrameRow`.
+    specified by `old_column`. If `ByRow` is used it is allowed for
+    `old_column` to select an empty set of columns, in which case no arguments
+    are passed to `fun` for each row.
 
     Column transformation can also be specified using the short `old_column =>
     fun` form. In this case, `new_column_name` is automatically generated as
@@ -473,8 +479,52 @@ SELECT_ARG_RULES =
     It is not allowed to pass `renamecols=false` if `old_column` is empty
     as it would generate an empty column name.
 
-    Column renaming and transformation operations can be passed wrapped in
-    vectors (this is useful when combined with broadcasting).
+    # Rules when `new_column_name` is a vector of `Symbol` or a string or is `AsTable`
+
+    In this case it is assumed that `fun` returns multiple columns.
+
+    If `fun` returns one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
+    `AbstractMatrix` then rules described below for `args` being a function or
+    a type apply.
+
+    If `fun` returns an `AbstractVector` then each element of this vector must
+    support `keys` function that must return a collection of `Symbol`s, strings
+    or integers; the return value of `keys` must be identical for all elements.
+    Then as many columns are created as there are elements in the return value
+    of the `keys` function and their names are set to be equal to the key names,
+    except if `keys` returns integers, in which case they are prefixed by `x`
+    (so the column names are e.g. `x1`, `x2`, ...)
+
+    If `fun` returns a value of any other type then it is assumed that it is
+    a table conforming to Tables.jl API and the `Tables.columntable` function is
+    called on it to get the resulting columns and their names.
+
+    Additionally if `new_column_name` is a vector of `Symbol` or string then column
+    names produced using the rules above are ignored and replaced by `new_column_name`
+    (the number of columns must be the same as the length `new_column_name` in this case).
+
+    # Rules when element of `args` is a function or a type
+
+    In this case a transformaton is passed `df` as a single argument.
+
+    If the return value of the transformation is of `AbstractDataFrame`,
+    `NamedTuple`, `DataFrameRow` or `AbstractMatrix` then it is treated as
+    containing multiple columns. For `AbstractMatrix` column names are generated
+    as `x1`, `x2`, etc. For `AbstractDataFrame`, `NamedTuple` of vectors and
+    `AbstractMatrix` the columns are taken as is from the returned value. For
+    `DataFrameRow` and` NamedTuple` not containing any vectors the returned
+    value is broadcasted a vector matching the target number of rows in the data
+    frame.
+
+    If the return value is an `AbstractVector` then it is used as-is. The resulting
+    column gets a name `x1`.
+
+    In all other cases the return value is broadcasted into a vector matching
+    the target number of rows in the data frame. As a particular rule, values
+    wrapped in a `Ref` or a `0`-dimensional `AbstractArray` are unwrapped and
+    then broadcasted. The resulting column gets a name `x1`.
+
+    # Special rules
 
     As a special rule passing `nrow` without specifying `old_column` creates a
     column named `:nrow` containing a number of rows in a source data frame, and
@@ -493,6 +543,7 @@ SELECT_ARG_RULES =
 
 """
     select!(df::DataFrame, args...; renamecols::Bool=true)
+    select!(args::Callable, df::DataFrame; renamecols::Bool=true)
 
 Mutate `df` in place to retain only columns specified by `args...` and return it.
 The result is guaranteed to have the same number of rows as `df`, except when no
@@ -547,16 +598,16 @@ julia> select!(df, :, [:c, :b] => (c,b) -> c .+ b .- sum(b)/length(b))
 
 julia> df = DataFrame(a=1:3, b=4:6);
 
-julia> select!(df, names(df) .=> sum);
+julia> select!(df, names(df) .=> [minimum maximum]);
 
 julia> df
-3×2 DataFrame
-│ Row │ a_sum │ b_sum │
-│     │ Int64 │ Int64 │
-├─────┼───────┼───────┤
-│ 1   │ 6     │ 15    │
-│ 2   │ 6     │ 15    │
-│ 3   │ 6     │ 15    │
+3×4 DataFrame
+│ Row │ a_minimum │ b_minimum │ a_maximum │ b_maximum │
+│     │ Int64     │ Int64     │ Int64     │ Int64     │
+├─────┼───────────┼───────────┼───────────┼───────────┤
+│ 1   │ 1         │ 4         │ 3         │ 6         │
+│ 2   │ 1         │ 4         │ 3         │ 6         │
+│ 3   │ 1         │ 4         │ 3         │ 6         │
 
 julia> df = DataFrame(a=1:3, b=4:6);
 
@@ -570,6 +621,36 @@ julia> select!(df, AsTable(:) => ByRow(mean), renamecols=false)
 │ 1   │ 2.5     │
 │ 2   │ 3.5     │
 │ 3   │ 4.5     │
+
+julia> df = DataFrame(a=1:3, b=4:6);
+
+julia> select!(first, df)
+3×2 DataFrame
+│ Row │ a     │ b     │
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 4     │
+│ 2   │ 1     │ 4     │
+│ 3   │ 1     │ 4     │
+
+julia> df = DataFrame(a=1:3, b=4:6, c=7:9)
+3×3 DataFrame
+│ Row │ a     │ b     │ c     │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 4     │ 7     │
+│ 2   │ 2     │ 5     │ 8     │
+│ 3   │ 3     │ 6     │ 9     │
+
+julia> select!(df, AsTable(:) => ByRow(x -> (mean=mean(x), std=std(x))) => :stats,
+               AsTable(:) => ByRow(x -> (mean=mean(x), std=std(x))) => AsTable)
+3×3 DataFrame
+│ Row │ stats                   │ mean    │ std     │
+│     │ NamedTuple…             │ Float64 │ Float64 │
+├─────┼─────────────────────────┼─────────┼─────────┤
+│ 1   │ (mean = 4.0, std = 3.0) │ 4.0     │ 3.0     │
+│ 2   │ (mean = 5.0, std = 3.0) │ 5.0     │ 3.0     │
+│ 3   │ (mean = 6.0, std = 3.0) │ 6.0     │ 3.0     │
 ```
 
 """
@@ -585,6 +666,7 @@ end
 
 """
     transform!(df::DataFrame, args...; renamecols::Bool=true)
+    transform!(args::Callable, df::DataFrame; renamecols::Bool=true)
 
 Mutate `df` in place to add columns specified by `args...` and return it.
 The result is guaranteed to have the same number of rows as `df`.
@@ -604,6 +686,7 @@ end
 
 """
     select(df::AbstractDataFrame, args...; copycols::Bool=true, renamecols::Bool=true)
+    select(args::Callable, df::DataFrame; renamecols::Bool=true)
 
 Create a new data frame that contains columns from `df` specified by `args` and
 return it. The result is guaranteed to have the same number of rows as `df`,
@@ -648,7 +731,7 @@ julia> df = DataFrame(a=1:3, b=4:6)
 │ 2   │ 2     │ 5     │
 │ 3   │ 3     │ 6     │
 
-julia> select(df, :b)
+julia> select(df, 2)
 3×1 DataFrame
 │ Row │ b     │
 │     │ Int64 │
@@ -656,24 +739,6 @@ julia> select(df, :b)
 │ 1   │ 4     │
 │ 2   │ 5     │
 │ 3   │ 6     │
-
-julia> select(df, Not(:b)) # drop column :b from df
-3×1 DataFrame
-│ Row │ a     │
-│     │ Int64 │
-├─────┼───────┤
-│ 1   │ 1     │
-│ 2   │ 2     │
-│ 3   │ 3     │
-
-julia> select(df, :a => :c, :b)
-3×2 DataFrame
-│ Row │ c     │ b     │
-│     │ Int64 │ Int64 │
-├─────┼───────┼───────┤
-│ 1   │ 1     │ 4     │
-│ 2   │ 2     │ 5     │
-│ 3   │ 3     │ 6     │
 
 julia> select(df, :a => ByRow(sin) => :c, :b)
 3×2 DataFrame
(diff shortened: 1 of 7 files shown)


test/select.jl
```diff
@@ -720,20 +720,32 @@ end
...
-            @test select(df, :x => ByRow(x -> retval)) == DataFrame(x_function = [retval])
-            cdf = copy(df)
-            select!(cdf, :x => ByRow(x -> retval))
-            @test cdf == DataFrame(x_function = [retval])
+            @test select(df, :x => ByRow(x -> retval) => AsTable) == DataFrame(;retval...)
+        elseif retval isa DataFrame
+            @test_throws MethodError select(df, :x => ByRow(x -> retval) => AsTable)
+        else # Matrix; surprising but following the API
+            @test select(df, :x => ByRow(x -> retval) => AsTable) ==
+                  DataFrame(["CartesianIndex($i, $j)" => 1.0 for i in 1:2, j in 1:2]...)
```
nalimilan, line 735: Can you comment on what happens here? :-)

test/select.jl
```diff
@@ -1306,4 +1327,246 @@ end
     @test df == DataFrame(a=1:3, b=4:6, c=7:9, d=10:12, a_b=5:2:9, a_b_etc=22:4:30)
 end
 
+@testset "additional tests for new rules" begin
+#    select select! transform transform! combine
+#    Union{Type{AsTable}, Symbol, AbstractVector{Symbol}, AbstractString, AbstractVector{<:AbstractString}}
+#    DataFrame, SubDataFrame
+    @testset "SELECT(FUN, DF)" begin
```
nalimilan, line 1334: Find a more general placeholder name than `select`, or split into separate testsets? Same below.

test/select.jl
```diff
@@ -1306,4 +1327,246 @@ end
...
+            @test select(df, [:b, :a], :a => (x -> (a=[11,12], b=[13,14])) => [:b, :a], :) ==
+                  DataFrame(b=[11, 12], a=[13, 14], c=[5, 6])
+        end
+    end
+end
+
+@testset "empty ByRow" begin
+    df = DataFrame(a=1:3)
+
+    @test select(df, [] => ByRow(() -> 1)) == DataFrame("function" => [1, 1, 1])
```
nalimilan, line 1511: Ah, so the default name we choose may not be the best idea. Should we use e.g. "fun" instead?

test/select.jl
```diff
@@ -1306,4 +1327,246 @@ end
...
+
+    for df in (DataFrame(), DataFrame(a=[]))
+        @test select(df, [] => ByRow(() -> 1)) == DataFrame("function" => [])
+        @test combine(df, [] => ByRow(() -> 1)) == DataFrame("function" => [])
+        if ncol(df) == 0
+            @test transform(df, [] => ByRow(() -> 1)) == DataFrame("function" => [])
+        else
+            @test transform(df, [] => ByRow(() -> 1)) == DataFrame("a" => [], "function" => [])
+        end
+        @test eltype(select(df, [] => ByRow(() -> 1)).function) == Int
```
nalimilan, line 1523: Maybe define a helper and use it everywhere to check both the contents and the type in a single line everywhere? IIRC we have something like this somewhere already.

nalimilan (commented):
Thanks, that's impressive. I haven't looked at the tests in detail yet, feel free to point me at interesting cases that I may have missed.

pdeffebach:
Could you please clarify what `AsTable` is? It's odd to see both `AsTable([:a, :b])` and `AsTable` that isn't called also. Is this a common pattern? Is it odd to dispatch on a data type and not an instance? Is there a better way to make an instance?

bkamins:
> Could you please clarify what `AsTable` is?

it is a type. types in Julia are also values, and we use this fact here

> It's odd to see both `AsTable([:a, :b])` and `AsTable` that isn't called also.

The first is instance of a type the second is a type (`AsTable` is not instantiated - so it kind-of says it is yet undecided what column names will actually be)

>  Is this a common pattern?

What do you mean by "pattern"? In general the transformation mini-language is DataFrames.jl specific. What is important is that `AsTable` does not collide with other column selectors.

> Is it odd to dispatch on a data type and not an instance?

We do not dispatch on type (actually if you look at the implementation there is a problem with this - we have to dynamically check for `AsTable` as a value)

> Is there a better way to make an instance?

We could use `AsTable()` or `astable` that would be `AsTable()`, but the question is if it would be more readable. Actually it was not me who invented using `AsTable` as the `DST` but I came to the conclusion that it is not a problem to use it.

---

In summary: it is not a common pattern, but do you have a better proposal what to use instead?

The benefit of this approach is:
* it is easy to type (which I guess Stata users like :smile:)
* it does not introduce a new symbol to learn

(and just to stress - we do not dispatch on `AsTable` as `DST` - it is a run-time check)

This situation is kind-of similar to `:` which is `Colon()` which is a function (and actually we have to also special case it), but how many times when you write `:` you think of it as a function?, so when you write `vector[:]` if Julia designer would be fancy it could be `vector[sin]` (fortunately it is not :smile:).

bkamins:
As an aferthought: we could use `:` instead of `AsTable` there - as `:` also has a meaning "take everything". Do you think `:` would be clearer there than `AsTable` (for sure it would be shorter).

pdeffebach:
Thank you for the clarification. I'm glad I have understanding of your thought process, here. 

I think `AsTable()` would be more readable, but I'm not sure. I'm not a fan of mixing instances and types like this, but I don't know if others share my viewpoint. It's also not clear what the empty parenthesis mean in this context, as clearly there are new variables.

bkamins:
@nalimilan - what do you think? I dislike `AsTable()` as it normally indicates zero columns in `SRC`.

nalimilan:
I prefer `AsTable` over `AsTable()`, as the latter could mean that you expect zero columns. `:` is an interesting option too, though it's less explicit and it could mean "replace all existing columns with the output".

bkamins:
Let us keep `AsTable` then.

pdeffebach:
Sounds good. Perhaps the best mental model is for it to be a `Callable` that acts as function barrier for correct types. We already have a similar logic with the `fun` in `src => fun => dest`.

src/abstractdataframe/selection.jl
```diff
@@ -70,8 +67,14 @@ normalize_selection(idx::AbstractIndex, sel::Pair{<:ColumnIndex, <:AbstractStrin
     normalize_selection(idx, first(sel) => Symbol(last(sel)), renamecols::Bool)
 
 function normalize_selection(idx::AbstractIndex,
-                             sel::Pair{<:Any,<:Pair{<:Base.Callable, Symbol}},
+                             sel::Pair{<:Any,<:Pair{<:Base.Callable,
+                             <:Union{Symbol, AbstractString, DataType,
+                                     AbstractVector{Symbol}, AbstractVector{<:AbstractString}}}},
                              renamecols::Bool)
+    lls = last(last(sel))
+    if lls isa DataType
```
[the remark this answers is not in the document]
bkamins: No, we just want to catch `AsTable` here - which unfortunately (as I commented to @pdeffebach) does not allow a better dispatch as:
```
julia> typeof(1 => sin => AsTable)
Pair{Int64,Pair{typeof(sin),DataType}}
```

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
     return (wanttable ? AsTable(c) : c) => fun => newcol
 end
 
-function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}, AsTable},
-                                    <:Pair{<:Base.Callable, Symbol}},
+function _transformation_helper(df, col_idx, @nospecialize(fun))
```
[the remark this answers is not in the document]
bkamins: I will fix the signatures. I will try to follow what you suggest.

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
+        return fun(df[!, col_idx])
+    elseif col_idx isa AsTable
+        tbl = Tables.columntable(select(df, col_idx.cols, copycols=false))
+        if isempty(tbl) && fun isa ByRow
+            if isempty(df)
+                T = Base.return_types(fun.fun, (NamedTuple{(),Tuple{}},))[1]
+                return T[]
+            else
+                return [fun.fun(NamedTuple()) for _ in 1:nrow(df)]
+            end
```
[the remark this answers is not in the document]
bkamins: > since `collect` will use inference to choose the element type of the empty vector?

It was failing for some reason - I will recheck and remove `if` it it is possible

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
+        else
+            cdf = eachcol(df)
+            return fun(map(c -> cdf[c], col_idx)...)
+        end
+    end
+    throw(ErrorException("unreachable reached"))
+end
+
+function _gen_colnames(@nospecialize(res), newname)
```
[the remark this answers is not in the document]
bkamins: yes - will do (I expected this comment after pushing the PR :smile:)

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
+            throw(ArgumentError("Number of returned columns does not match the " *
+                                "length of requested output"))
+        end
+        colnames = newname
+    end
+
+    return colnames
+end
+
+function _expand_to_table(@nospecialize(res))
```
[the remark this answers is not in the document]
bkamins: OK - I will remove it and check the timing.

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
+    if ncol(newdf) == 0
+        # if allow_resizing_newdf[] is false we know this is select or transform
+        rows = allow_resizing_newdf[] ? 1 : nrow(df)
+    else
+        # allow squashing a scalar to 0 rows
+        rows = nrow(newdf)
+    end
+    @assert length(colnames) == length(res)
+    for (newname, v) in zip(colnames, res)
+        # note that newdf potentially can contain c in general
```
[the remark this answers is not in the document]
bkamins: yes - leftover from ugly initial code

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
-                           transformed_cols::Dict{Symbol, Any}, copycols::Bool,
+                           transformed_cols::Set{Symbol}, copycols::Bool,
                            allow_resizing_newdf::Ref{Bool})
-    col_idx, (fun, newname) = nc
+    if nc isa Base.Callable
+        col_idx, fun, newname = nothing, nc, nothing
+    else
+        col_idx, (fun, newname) = nc
+    end
+    if newname isa DataType
```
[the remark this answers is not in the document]
bkamins: As commented above - I only allow `DataType` in the function signature when defining what `Pair`s are allowed.

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
-        res = fun(Tables.columntable(select(df, col_idx.cols, copycols=false)))
-    else
-        # it should be fast enough here as we do not expect to do it millions of times
-        @assert col_idx isa AbstractVector{Int}
-        res = fun(map(c -> cdf[c], col_idx)...)
+    res = _transformation_helper(df, col_idx, fun)
+
+    if (newname === AsTable || newname isa AbstractVector{Symbol}) &&
+        !(res isa Union{AbstractDataFrame, NamedTuple, DataFrameRow, AbstractMatrix})
+        res = _expand_to_table(res)
```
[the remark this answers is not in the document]
bkamins: OK

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
-    end
-    if res isa AbstractVector
-        # allow shortening to 0 rows
-        if allow_resizing_newdf[] && nrow(newdf) == 1
-            newdfcols = _columns(newdf)
-            for (i, col) in enumerate(newdfcols)
-                newdfcols[i] = fill!(similar(col, length(res)), first(col))
-            end
+        if newname isa Symbol
+            throw(ArgumentError("Table returned while a single column return value was requested"))
```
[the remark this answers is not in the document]
bkamins: let us discuss about the proper wording. We have to cover two cases:
```src => fun => :a```
and
```src => fun```
(as in the second we implicitly assume that one column is returned when doing automatic column name generation, if one wants to pass a table either `AsTable` or multiple column names are required)

File: src/abstractdataframe/selection.jl [hunk 14]
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
+        isempty(colnames) && return # nothing to do
 
-        # !allow_resizing_newdf[] && ncol(newdf) == 0
-        # means that we use `select` or `transform` not `combine`
-        if !allow_resizing_newdf[] && ncol(newdf) == 0 && length(res) != nrow(df)
-            throw(ArgumentError("length $(length(res)) of vector returned from " *
-                                "function $fun is different from number of rows " *
-                                "$(nrow(df)) of the source data frame."))
+        if any(in(transformed_cols), colnames)
+            throw(ArgumentError("Duplicate column name returned"))
```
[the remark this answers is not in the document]
bkamins: you got it right :)

src/abstractdataframe/selection.jl
```diff
@@ -386,6 +533,14 @@ julia> select!(df, AsTable(:) => ByRow(mean), renamecols=false)
 select!(df::DataFrame, args...; renamecols::Bool=true) =
     _replace_columns!(df, select(df, args..., copycols=false, renamecols=renamecols))
 
+function select!(arg::Base.Callable, df::AbstractDataFrame; renamecols::Bool=true)
+    if arg isa Colon
+        throw(ArgumentError("Only transformations are allowed when function is a " *
+                            "frist argument to select!"))
```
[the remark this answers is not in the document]
bkamins: it is more complex unfortunately - see the corrected error message. (we also have to handle `GroupedDataFrame`)

test/select.jl
bkamins: `keys` for `Matrix` returns `CartesianIndex` values, which get converted to `Symbol`.

test/select.jl
bkamins: OK

test/select.jl
bkamins: not sure - I think this case is very rare in practice.

@pdeffebach - any opinion?

test/select.jl
bkamins: OK

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
+    parent_cols = col_idx isa AsTable ? col_idx.cols : (col_idx === nothing ? (1:ncol(df)) : col_idx)
+    if copycols && !(fun isa ByRow) && (v isa SubArray || any(i -> vpar === parent(cdf[i]), parent_cols))
+        newdf[!, newname] = copy(v)
+    else
+        newdf[!, newname] = v
+    end
+end
+
+function select_transform!(@nospecialize(nc::Union{Base.Callable, Pair{<:Union{Int, AbstractVector{Int}, AsTable},
+                                                    <:Pair{<:Base.Callable,
```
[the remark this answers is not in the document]
bkamins: fixed

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +169,207 @@ function normalize_selection(idx::AbstractIndex,
...
-                                "function $fun is different from number of rows " *
-                                "$(nrow(df)) of the source data frame."))
+        if any(in(transformed_cols), colnames)
+            throw(ArgumentError("Duplicate column name returned"))
+        else
+            startlen = length(transformed_cols)
+            union!(transformed_cols, colnames)
+            @assert startlen + length(colnames) == length(transformed_cols)
+        end
+        if res isa AbstractDataFrame
```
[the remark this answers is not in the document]
bkamins: done

test/select.jl
bkamins: OK - will add (we have it for split-apply-combine)

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
bkamins: and the complexity of the signatures shows how many options we support (which prompted me to write about it in https://bkamins.github.io/julialang/2020/10/02/performance.html).

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +170,208 @@ function normalize_selection(idx::AbstractIndex,
     return (wanttable ? AsTable(c) : c) => fun => newcol
 end
 
-function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}, AsTable},
-                                    <:Pair{<:Base.Callable, Symbol}},
+function _transformation_helper(df::AbstractDataFrame, col_idx, @nospecialize(fun))
+    if col_idx === nothing
+        return fun(df)
```
pdeffebach, line 175: Okay so we are keeping this in transform? I voted against [here](https://github.com/JuliaData/DataFrames.jl/issues/2410#issuecomment-695823805) but I don't know if there was ever a consensus. 

It's fine to add! But just to confirm.

pdeffebach: Also, the `do` syntax here is really nice, I didnt realize that was added. I approve this.

pdeffebach:
Is the following expected behavior? 

```
julia> df = DataFrame(a = [1, 2], b = [3, 4]);

julia> transform(df; renamecols = true) do sdf
           sdf
       end
2×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 3     │
│ 2   │ 2     │ 4     │
```

I would have thought with the `renamecols = true` we would have columns `:a, :b, :a1, :b1`

pdeffebach:
Additionally, should the following work?

```
julia> transform(df, foo)
2×3 DataFrame
│ Row │ a     │ b     │ x1    │
│ 1   │ 1     │ 3     │ 1     │
│ 2   │ 2     │ 4     │ 1     │
julia> df = DataFrame(a = [1, 2], b = [3, 4]);

julia> foo(df) = fill(1, nrow(df));

julia> transform(df, foo => :a)
ERROR: ArgumentError: Unrecognized column selector: foo => :a
```

src/abstractdataframe/selection.jl
bkamins: We allow `select(fun, df)` and `select(df, fun)` but not allow `select(fun => AsTable, df)` and `select(df, fun => AsTable)` (or anything else in place of `AsTable`). In short only a stand-alone `::Callable` is allowed as a special case.

pdeffebach: Thanks. That takes care of my 2nd comment, but I think my first one (on the main discussion) still applies.

bkamins:
> I would have thought with the `renamecols = true` we would have columns `:a, :b, :a1, :b1`

No - this would be `makeunique`, but we do not support it in transformations (it would be ambiguous if you want to overwrite columns vs add new columns with de-duplicated names).

`renamecols` handles the way how automatic column generation works (if a suffix with function name is added or not)

> julia> transform(df, foo)
2×3 DataFrame
│ Row │ a     │ b     │ x1    │
│ 1   │ 1     │ 3     │ 1     │
│ 2   │ 2     │ 4     │ 1     |

What is `foo` here (you define it below this call)? If it is `foo(x) = 1` or `foo(x) = [1,1]` or `foo(x) = ones(Int, 2, 1)` then this is expected

> julia> transform(df, foo => :a)
ERROR: ArgumentError: Unrecognized column selector: foo => :a

This is also expected - and follows your request to disallow `FUN => DST` syntax. Only `FUN` or `SRC => FUN => DST` or `SRC => FUN` syntaxes are allowed.

In general `FUN` syntax is mainly intended for `SELECT(FUN, DF)` style, but we support it in `SELECT(DF, FUN)` style as it is not a problem to support it.

pdeffebach:
Thanks, this is all very clear. 

> What is `foo` here (you define it below this call)? If it is `foo(x) = 1` or `foo(x) = [1,1]` or `foo(x) = ones(Int, 2, 1)` then this is expected

Yes that is expected. 

This is a really impressive work! Really appreciate it and the thought you've put into this.

pdeffebach:
I think a `makeunique` option where all the columns are over-written by default would be useful in the future. But this is fine for now, especially since we don't support `makeunique` in existing transform.

bkamins:
> I think a `makeunique` option where all the columns are not over-written by default would be useful in the future.

OK - this can be added easily, but let us keep it for the future (this PR is already complex enough :smile:).

test/select.jl
bkamins: @pdeffebach - in case you missed it. Would you prefer `function` or `fun` in DataFramesMeta.jl (and in general) as a default in this case? For me `function` is more explicit (that is why I have chosen it), but it has a shortcoming that `function` is a reserved word (which is mildly inconvenient, as you can write `df.function` the only thing you cannot write is `DataFrame(function  = 1)` as it will error)

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
nalimilan: Ah, indeed. Then maybe just say "expected" instead of "requested"?

test/select.jl
nalimilan: OK. Should we really allow this? It doesn't seem to be of any practical use, and it could be confusing (e.g. if you thought you returned a vector but returned a matrix instead).

test/select.jl
nalimilan: Yeah it shouldn't be a very common use.

test/select.jl
nalimilan: OK. Should we really allow this? It doesn't seem to be of any practical use, and it could be confusing (e.g. if you thought you returned a vector but returned a matrix instead).

src/abstractdataframe/selection.jl
```diff
@@ -204,40 +197,47 @@ function _transformation_helper(df, col_idx, @nospecialize(fun))
...
+    if newname !== AsTable && newname !== nothing
         if length(colnames) != length(newname)
             throw(ArgumentError("Number of returned columns does not match the " *
                                 "length of requested output"))
         end
         colnames = newname
     end
 
-    return colnames
+    # fix the type to avoid unnecesarry compilations of methods
```
nalimilan, line 216:
```suggestion
    # fix the type to avoid unnecessary compilations of methods
```

src/abstractdataframe/selection.jl
```diff
@@ -270,20 +272,85 @@ function _fix_existing_columns_for_vector(newdf, df, allow_resizing_newdf, lr, @
     allow_resizing_newdf[] = false
 end
 
-function _add_col_check_copy(df, newdf, col_idx, copycols, @nospecialize(fun), newname, @nospecialize(v))
+function _add_col_check_copy(newdf::DataFrame, df::AbstractDataFrame,
+                             col_idx::Union{Nothing, Int, AbstractVector{Int}, AsTable},
+                             copycols::Bool, @nospecialize(fun),
+                             newname::Symbol, @nospecialize(v))
```
nalimilan, line 278: `v` supposed to be a vector, right? IIRC `@nospecialize` works with type annotations.

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
bkamins: OK

test/select.jl
bkamins: But then the problem is that we take `keys`. Should we then check if `keys` is either all `Integer` or all `Symbol` and allow only these two cases?

src/abstractdataframe/selection.jl
bkamins: you are right - thank you for being so careful, actually we can specialize on `v` I think.

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +170,245 @@ function normalize_selection(idx::AbstractIndex,
...
+                            "function $fun is different from number of rows " *
+                            "$(nrow(df)) of the source data frame."))
+    end
+    allow_resizing_newdf[] = false
+end
+
+function _add_col_check_copy(newdf::DataFrame, df::AbstractDataFrame,
+                             col_idx::Union{Nothing, Int, AbstractVector{Int}, AsTable},
+                             copycols::Bool, @nospecialize(fun),
+                             newname::Symbol, @nospecialize(v))
```
bkamins, line 278:
```suggestion
                             newname::Symbol, v::AbstractVector)
```

src/abstractdataframe/selection.jl
```diff
@@ -170,56 +170,245 @@ function normalize_selection(idx::AbstractIndex,
...
-        res = fun(df[!, col_idx])
-    elseif col_idx isa AsTable
-        res = fun(Tables.columntable(select(df, col_idx.cols, copycols=false)))
-    else
-        # it should be fast enough here as we do not expect to do it millions of times
-        @assert col_idx isa AbstractVector{Int}
-        res = fun(map(c -> cdf[c], col_idx)...)
+    res = _transformation_helper(df, col_idx, fun)
+
+    if (newname === AsTable || newname isa AbstractVector{Symbol})
```
nalimilan, line 370:
```suggestion
    if newname === AsTable || newname isa AbstractVector{Symbol}
```

test/select.jl
nalimilan: Since matrices have a shape, it would make sense to allow returning only single-row matrices with `ByRow`, and generate column names like we do for vectors. What happens when you return a multiple-row data frame with `ByRow`?

nalimilan: Since matrices have a shape, it would make sense to allow returning only single-row matrices with `ByRow`, and generate column names like we do for vectors. What happens when you return a multiple-row data frame with `ByRow`?

bkamins: > What happens when you return a multiple-row data frame with ByRow?

1. if it is just `src => ByRow(x -> DataFrame(...))` then nothing - it is just accepted.
2. if it is `src => ByRow(x -> DataFrame(...)) => AsTable` then it is an error, because `DataFrame` does not support `keys`

I think both behaviours are OK.

> Since matrices have a shape, it would make sense to allow returning only single-row matrices with ByRow, and generate column names like we do for vectors.

I would prefer to avoid such a special case for something that is not really needed often. What I expect to be returned most often are: 1) vectors, 2) `Tuple`s, 3) `NamedTuple`s, 4) `Dict`s, 5) `DataFrameRow`. And the rule "use `keys` to get default column names" for them works correctly. I would handle other cases just as an extension of the rule that correctly works for these 5 cases. That is why I proposed to check if `keys` returns `Integer` or `Symbol` (maybe also `AbstractString`) - as these are only useful cases. An alternative to `keys` could be `eachindex` (which for a `Matrix` would give integer indices).

Also if you have a matrix it is easy to wrap it in `vec` to cheaply get a vector without copying anything.

Finally - the fact that matrices have a shape actually should disallow them as in "vector of something" form this "something" should be 1-dimenional (technically we accept an iterable and disregard a dimension).

nalimilan: Then what about disallowing returning matrices with `src => ByRow(...) => AsTable` completely? That way we would only accept 1-dimensional or shapeless objects.

bkamins: requiring `keys` to return `Integer` or `Symbol` or `AbstractString` will have exactly this effect (as I think it is an easier rule to understand).

pdeffebach:
Why allow returning matrices at all if we are deprecating the `matrix` constructor? For the `transform` case, if you can't call `DataFrame` on it, then we shouldn't support it. That seems like the most consistent approach.

bkamins:
> Why allow returning matrices at all

Only for backward compatibility reasons. Note that we will not disallow returning them. The only question is what happens with them and we have two options:
1. leave them "as is" when there is no `AsTable` (i.e. repeat as many times as there are rows without copying) and throw an error if there is `AsTable`
2. throw an error when there is no `AsTable` and expand to multiple columns when there is `AsTable`

I was thinking which behavior the user would prefer when returning a matrix and I thought that the second is more natural. Would you prefer the first?

In general - under current rules the only case when we throw an error is `NamedTuple` that mixes scalars and vectors.

---

Note that this is a different case from what we discuss with @nalimilan, as he has raised a case when `ByRow` returns a matrix and there is `AsTable` at the end.

pdeffebach:
> * leave them "as is" when there is no `AsTable` (i.e. repeat as many times as there are rows without copying) and throw an error if there is `AsTable`
> * throw an error when there is no `AsTable` and expand to multiple columns when there is `AsTable`

The first option reminds me of `R` too much. The silently repetition is a source of bugs. The user can always pass it as a `Ref` if they want it to be repeated, right? I prefer option 2.

bkamins:
So option 2 is what we currently have 😄.

test/select.jl
bkamins: I will implement this rule (i.e. restricting allowed types of keys)

bkamins:
I have updated the documentation (so essentially when we accept this this should be good to be merged).

@nalimilan - as usual - feel free to rewrite the docstrings :smile: (and sorry for mistakes, as for sure there will be some).

src/abstractdataframe/selection.jl
```diff
@@ -428,35 +432,41 @@ SELECT_ARG_RULES =
...
+    column vector. Otherwise `old_column` can be any column indexing syntax, in
+    which case `fun` will be passed the column vectors specified by `old_column`
+    as separate arguments. The only exception is when `old_column` is an
+    `AsTable` type wrapping a selector, in which case `fun` is passed a
+    `NamedTuple` containing the selected columns.
+
+    Column renaming and transformation operations can be passed wrapped in
+    vectors or matrices (this is useful when combined with broadcasting).
+
+    # Rules when `new_column_name` is a `Symbol` or a string or is missing
```
nalimilan, line 453:
```suggestion
    # Rules when `new_column_name` is a `Symbol` or a string or is absent
```

Can you check that this gives a readable result in the REPL/manual given that it's a long line?

src/abstractdataframe/selection.jl
```diff
@@ -428,35 +432,41 @@ SELECT_ARG_RULES =
...
     `ByRow` struct. In this case if `old_column` is a `Symbol`, a string, or an
     integer then `fun` is applied to each element (row) of `old_column` using
     broadcasting. Otherwise `old_column` can be any column indexing syntax, in
     which case `fun` will be passed one argument for each of the columns
-    specified by `old_column`. If `ByRow` is used it is not allowed for
-    `old_column` to select an empty set of columns nor for `fun` to return a
-    `NamedTuple` or a `DataFrameRow`.
+    specified by `old_column`. If `ByRow` is used it is allowed for
+    `old_column` to select an empty set of columns, in which case no arguments
+    are passed to `fun` for each row.
```
nalimilan, line 469:
```suggestion
    `old_column` to select an empty set of columns, in which case `fun`
     is called for each row without any arguments.
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
     It is not allowed to pass `renamecols=false` if `old_column` is empty
     as it would generate an empty column name.
 
-    Column renaming and transformation operations can be passed wrapped in
-    vectors (this is useful when combined with broadcasting).
+    # Rules when `new_column_name` is a vector of `Symbol` or a string or is `AsTable`
```
nalimilan, line 482:
```suggestion
    # Rules when `new_column_name` is a vector of `Symbol`s or strings or is `AsTable`
```

src/abstractdataframe/selection.jl
```diff
@@ -428,35 +432,41 @@ SELECT_ARG_RULES =
...
-    `new_column_name` must be a `Symbol` or a string, and `fun` a function or a
-    type. If `old_column` is a `Symbol`, a string, or an integer then `fun` is
-    applied to the corresponding column vector. Otherwise `old_column` can be
-    any column indexing syntax, in which case `fun` will be passed the column
-    vectors specified by `old_column` as separate arguments. The only exception
-    is when `old_column` is an `AsTable` type wrapping a selector, in which case
-    `fun` is passed a `NamedTuple` containing the selected columns.
+    `new_column_name` must be a `Symbol` or a string, a vector of `Symbol` or
+    string, or `AsTable`, and `fun` a function or a type. If `old_column` is a
+    `Symbol`, a string, or an integer then `fun` is applied to the corresponding
```
nalimilan, line 443:
```suggestion
    `new_column_name` must be a `Symbol` or a string, a vector of `Symbol`s or
    strings, or `AsTable`. `fun` must be a function or a type. If `old_column` is a
    `Symbol`, a string, or an integer then `fun` is applied to the corresponding
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    `AbstractMatrix` then rules described below for `args` being a function or
+    a type apply.
+
+    If `fun` returns an `AbstractVector` then each element of this vector must
+    support `keys` function that must return a collection of `Symbol`s, strings
+    or integers; the return value of `keys` must be identical for all elements.
+    Then as many columns are created as there are elements in the return value
+    of the `keys` function and their names are set to be equal to the key names,
+    except if `keys` returns integers, in which case they are prefixed by `x`
+    (so the column names are e.g. `x1`, `x2`, ...)
```
nalimilan, line 496:
```suggestion
    (so the column names are e.g. `x1`, `x2`, ...).
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
-    vectors (this is useful when combined with broadcasting).
+    # Rules when `new_column_name` is a vector of `Symbol` or a string or is `AsTable`
+
+    In this case it is assumed that `fun` returns multiple columns.
+
+    If `fun` returns one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
+    `AbstractMatrix` then rules described below for `args` being a function or
+    a type apply.
+
+    If `fun` returns an `AbstractVector` then each element of this vector must
```
nalimilan, line 490: `keys` is called both on the vector itself and on its entries, so it's confusing.

Also, it would be useful to mention here that automatic names are ignored when provided manually. Otherwise one may wonder whether the description applies to this case.

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    support `keys` function that must return a collection of `Symbol`s, strings
+    or integers; the return value of `keys` must be identical for all elements.
+    Then as many columns are created as there are elements in the return value
+    of the `keys` function and their names are set to be equal to the key names,
+    except if `keys` returns integers, in which case they are prefixed by `x`
+    (so the column names are e.g. `x1`, `x2`, ...)
+
+    If `fun` returns a value of any other type then it is assumed that it is
+    a table conforming to Tables.jl API and the `Tables.columntable` function is
```
nalimilan, line 499:
```suggestion
    a table conforming to the Tables.jl API and the `Tables.columntable` function is
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
 
-    Column renaming and transformation operations can be passed wrapped in
-    vectors (this is useful when combined with broadcasting).
+    # Rules when `new_column_name` is a vector of `Symbol` or a string or is `AsTable`
+
+    In this case it is assumed that `fun` returns multiple columns.
+
+    If `fun` returns one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
+    `AbstractMatrix` then rules described below for `args` being a function or
+    a type apply.
```
nalimilan, line 488: Maybe put this at the end so that it's clearer what "below" means.

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    except if `keys` returns integers, in which case they are prefixed by `x`
+    (so the column names are e.g. `x1`, `x2`, ...)
+
+    If `fun` returns a value of any other type then it is assumed that it is
+    a table conforming to Tables.jl API and the `Tables.columntable` function is
+    called on it to get the resulting columns and their names.
+
+    Additionally if `new_column_name` is a vector of `Symbol` or string then column
+    names produced using the rules above are ignored and replaced by `new_column_name`
+    (the number of columns must be the same as the length `new_column_name` in this case).
```
nalimilan, line 504:
```suggestion
    (the number of columns must be the same as the length of `new_column_name` in this case).
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    Then as many columns are created as there are elements in the return value
+    of the `keys` function and their names are set to be equal to the key names,
+    except if `keys` returns integers, in which case they are prefixed by `x`
+    (so the column names are e.g. `x1`, `x2`, ...)
+
+    If `fun` returns a value of any other type then it is assumed that it is
+    a table conforming to Tables.jl API and the `Tables.columntable` function is
+    called on it to get the resulting columns and their names.
+
+    Additionally if `new_column_name` is a vector of `Symbol` or string then column
```
nalimilan, line 502:
```suggestion
    Additionally if `new_column_name` is a vector of `Symbol`s or strings then column
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    names produced using the rules above are ignored and replaced by `new_column_name`
+    (the number of columns must be the same as the length `new_column_name` in this case).
+
+    # Rules when element of `args` is a function or a type
+
+    In this case a transformaton is passed `df` as a single argument.
```
nalimilan, line 508:
```suggestion
    In this case the function or type is called with `df` as a single argument.
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+
+    If the return value of the transformation is of `AbstractDataFrame`,
```
nalimilan, line 510:
```suggestion
    If the return value of the transformation is one of `AbstractDataFrame`,
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    `NamedTuple`, `DataFrameRow` or `AbstractMatrix` then it is treated as
+    containing multiple columns. For `AbstractMatrix` column names are generated
+    as `x1`, `x2`, etc. For `AbstractDataFrame`, `NamedTuple` of vectors and
+    `AbstractMatrix` the columns are taken as is from the returned value. For
+    `DataFrameRow` and` NamedTuple` not containing any vectors the returned
+    value is broadcasted a vector matching the target number of rows in the data
```
nalimilan, line 516:
```suggestion
    value is broadcasted to a vector matching the target number of rows in the data
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    frame.
+
+    If the return value is an `AbstractVector` then it is used as-is. The resulting
+    column gets a name `x1`.
```
nalimilan, line 520:
```suggestion
    column gets the name `x1`.
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+
+    In all other cases the return value is broadcasted into a vector matching
+    the target number of rows in the data frame. As a particular rule, values
+    wrapped in a `Ref` or a `0`-dimensional `AbstractArray` are unwrapped and
+    then broadcasted. The resulting column gets a name `x1`.
```
nalimilan, line 525:
```suggestion
    then broadcasted. The resulting column gets the name `x1`.
```

src/abstractdataframe/selection.jl
```diff
@@ -469,8 +479,52 @@ SELECT_ARG_RULES =
...
+    `DataFrameRow` and` NamedTuple` not containing any vectors the returned
+    value is broadcasted a vector matching the target number of rows in the data
+    frame.
+
+    If the return value is an `AbstractVector` then it is used as-is. The resulting
+    column gets a name `x1`.
+
+    In all other cases the return value is broadcasted into a vector matching
+    the target number of rows in the data frame. As a particular rule, values
+    wrapped in a `Ref` or a `0`-dimensional `AbstractArray` are unwrapped and
```
nalimilan, line 524: Is this a particular rule or just the application of broadcasting rules?

src/abstractdataframe/selection.jl
```diff
@@ -778,6 +869,68 @@ julia> combine(df, :a => sum, nrow, renamecols=false)
...
+3×3 DataFrame
+│ Row │ a     │ b     │ c     │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 4     │ 7     │
+│ 2   │ 2     │ 5     │ 8     │
+│ 3   │ 3     │ 6     │ 9     │
+
+julia> combine(df, AsTable(:) => ByRow(x -> (mean=mean(x), std=std(x))) => :stats,
+                      AsTable(:) => ByRow(x -> (mean=mean(x), std=std(x))) => AsTable)
```
nalimilan, line 926: Weird indentation.

src/abstractdataframe/selection.jl
bkamins: It is OK (the text is 70 character long, so it would not look nice only for narrower terminals, but I think it is better to leave a longer description)

src/abstractdataframe/selection.jl
bkamins: So here maybe we need to change a wording if we want to be precise. I have written "broadcasted into" which suggests we use broadcasting from Base, but actually we do not. We rather "reuse the value", and with respect to this `Ref` and `0`-dimensional `AbstractArray` are special cases.

We do not do broadcasting, as it has a different set of corner cases (e.g. a 1-element vector would be expanded and we do not do it; if a returned value does not support broadcasting an error would be thrown and we reuse it etc.)

What do you think would be best here?

src/abstractdataframe/selection.jl
```diff
@@ -564,18 +869,78 @@ julia> combine(df, :a => sum, nrow, renamecols=false)
...
```
bkamins, line 926:
```suggestion
julia> combine(df, AsTable(:) => ByRow(x -> (mean=mean(x), std=std(x))) => :stats,
               AsTable(:) => ByRow(x -> (mean=mean(x), std=std(x))) => AsTable)
```

src/abstractdataframe/selection.jl
bkamins: fixed

src/abstractdataframe/selection.jl
bkamins: In other words: we here use "pseduo-broadcasting" not "broadcasting" and it would be best to discuss how to document it in general (it is documented well I think for `DataFrame` and `insertcols!` where we also use "pseudo-broadcasting", but there additional rules apply (not all values are accepted), that we do not impose here)

src/abstractdataframe/selection.jl
bkamins: I made it more precise what we reference to

src/abstractdataframe/selection.jl
bkamins: >  it would be useful to mention here that automatic names are ignored when provided manually

It was there below. Now I have moved it up, but it means I have to repeat the description twice (but I think it is OK, so I followed your suggestion)

> `keys` is called both on the vector itself and on its entries, so it's confusing.

`keys` is not called on a vector. It is only called on all of its entries.

Essentially what we do is making sure that each entry of the vector has the same schema (to avoid the problem that people repeatedly report in Tables.jl that the schema is taken form the first entry and reused, so wrong results are silently produced if the schema is not constant across all entries of a row table)

Can you please highlight which part of the description is confusing?

Also there is a special case when the vector is empty, in which case we do not call `keys` at all but fall back to `Tables.columntable`, but I would not mention it as it is a corner case that probably is not very important (though it leads to inconsistencies in some corner cases, see https://github.com/JuliaData/Tables.jl/issues/204).

src/abstractdataframe/selection.jl
nalimilan: Maybe just say "repeated"?

src/abstractdataframe/selection.jl
```diff
@@ -283,8 +479,52 @@ SELECT_ARG_RULES =
...
+    # Rules when `new_column_name` is a vector of `Symbol` or a string or is `AsTable`
+
+    In this case it is assumed that `fun` returns multiple columns.
+
+    If `fun` returns one of `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`,
+    `AbstractMatrix` then rules described below for `args` being a function or
+    a type apply.
+
+    If `fun` returns an `AbstractVector` then each element of this vector must
+    support `keys` function that must return a collection of `Symbol`s, strings
```
nalimilan, line 491:
```suggestion
    support the `keys` function, which must return a collection of `Symbol`s, strings
```

src/abstractdataframe/selection.jl
nalimilan: Sorry I guess I misread.

src/abstractdataframe/selection.jl
```diff
@@ -283,8 +479,54 @@ SELECT_ARG_RULES =
...
+    value is broadcasted to a vector matching the target number of rows in the data
+    frame.
+
+    If the return value is an `AbstractVector` then it is used as-is. The resulting
+    column gets the name `x1`.
+
+    In all other cases the return value is broadcasted into a vector matching
+    the target number of rows in the data frame. As a particular rule, values
+    wrapped in a `Ref` or a `0`-dimensional `AbstractArray` are unwrapped and
+    then broadcasted. The resulting column gets the name `x1`.
```
bkamins, line 527:
```suggestion
    In all other cases the return value is repeated in a vector matching
    the target number of rows in the data frame. As a particular rule, values
    wrapped in a `Ref` or a `0`-dimensional `AbstractArray` are unwrapped and
    then repeated. The resulting column gets the name `x1`.
```

src/abstractdataframe/selection.jl
bkamins: ok

src/abstractdataframe/selection.jl
```diff
@@ -242,35 +432,41 @@ SELECT_ARG_RULES =
...
+    `AsTable` type wrapping a selector, in which case `fun` is passed a
+    `NamedTuple` containing the selected columns.
+
+    Column renaming and transformation operations can be passed wrapped in
+    vectors or matrices (this is useful when combined with broadcasting).
+
+    # Rules when `new_column_name` is a `Symbol` or a string or is absent
 
     If `fun` returns a value of type other than `AbstractVector` then it will be
     broadcasted into a vector matching the target number of rows in the data
```
bkamins, line 456:
```suggestion
    repeated in a vector matching the target number of rows in the data
```

src/abstractdataframe/selection.jl
```diff
@@ -242,35 +432,41 @@ SELECT_ARG_RULES =
...
     frame, unless its type is one of `AbstractDataFrame`, `NamedTuple`,
-    `DataFrameRow`, `AbstractMatrix`, in which case an error is thrown as
-    currently these return types are not allowed. As a particular rule, values
-    wrapped in a `Ref` or a `0`-dimensional `AbstractArray` are unwrapped and
-    then broadcasted.
+    `DataFrameRow`, `AbstractMatrix`, in which case an error is thrown. As a
+    particular rule, values wrapped in a `Ref` or a `0`-dimensional
+    `AbstractArray` are unwrapped and then broadcasted.
```
bkamins, line 460:
```suggestion
    `AbstractArray` are unwrapped and then repeated.
```

nalimilan (approved):

bkamins:
Thank you for all the comments. If there are no more issues with this proposal I will merge the PR tomorrow and follow up with a small `filter` and a big `GroupedDataFrame` PRs.

bkamins:
Thank you!

Issue #2433, opened 2020-09-14
Deprecate `DataFrame(::AbstractMatrix)`


rofinn:
I feel like it's rare that folks actually want to call this constructor without specifying column names. Also, it isn't that hard to do this with the tables interface via `DataFrame(Tables.table(X))`. 

```
julia> DataFrame(Tables.table(x))
10×5 DataFrame
│ Row │ Column1  │ Column2  │ Column3   │ Column4  │ Column5   │
│     │ Float64  │ Float64  │ Float64   │ Float64  │ Float64   │
├─────┼──────────┼──────────┼───────────┼──────────┼───────────┤
│ 1   │ 0.829388 │ 0.100101 │ 0.733667  │ 0.591796 │ 0.538582  │
│ 2   │ 0.573906 │ 0.402706 │ 0.41509   │ 0.899583 │ 0.482383  │
│ 3   │ 0.703773 │ 0.729495 │ 0.375752  │ 0.782216 │ 0.425253  │
│ 4   │ 0.5717   │ 0.93305  │ 0.0922584 │ 0.211547 │ 0.123723  │
│ 5   │ 0.927349 │ 0.654163 │ 0.11306   │ 0.452749 │ 0.921814  │
│ 6   │ 0.514157 │ 0.891617 │ 0.481489  │ 0.368915 │ 0.718185  │
│ 7   │ 0.717906 │ 0.627349 │ 0.371058  │ 0.958308 │ 0.925878  │
│ 8   │ 0.104714 │ 0.995778 │ 0.671143  │ 0.410917 │ 0.547495  │
│ 9   │ 0.373802 │ 0.147139 │ 0.425674  │ 0.893018 │ 0.937846  │
│ 10  │ 0.26242  │ 0.30546  │ 0.750857  │ 0.196515 │ 0.0470533 │

julia> DataFrame(x)
10×5 DataFrame
│ Row │ x1       │ x2       │ x3        │ x4       │ x5        │
│     │ Float64  │ Float64  │ Float64   │ Float64  │ Float64   │
├─────┼──────────┼──────────┼───────────┼──────────┼───────────┤
│ 1   │ 0.829388 │ 0.100101 │ 0.733667  │ 0.591796 │ 0.538582  │
│ 2   │ 0.573906 │ 0.402706 │ 0.41509   │ 0.899583 │ 0.482383  │
[...]

PR #2464, opened 2020-10-01
[BREAKING] deprecate DataFrame constructors

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2433. Before I change tests and documentation please review if we agree with what is proposed to get deprecated. Thank you.


src/dataframe/dataframe.jl
```diff
@@ -180,15 +180,11 @@ function DataFrame(pairs::Pair{<:AbstractString,<:Any}...; makeunique::Bool=fals
                      copycols=copycols)
 end
 
-# these two are needed as a workaround Tables.jl dispatch
+# this is needed as a workaround for Tables.jl dispatch
 DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false,
           copycols::Bool=true) =
     DataFrame(pairs..., makeunique=makeunique, copycols=copycols)
```
oxinabox, line 186: Can we get ride of this one also?
All it is doing is calling `DataFrame(pairs...)` 
can't the user do that themselves?

oxinabox: I guess this one is handy so you can do:
```
DataFrame(colnames .=> cols)
```
etc

oxinabox (approved):

src/dataframe/dataframe.jl
bkamins: Exactly, in other places we got recurring requests from the users to keep allowing vectors, where one could splat them (e.g. in  `select` etc.) as it is easy to construct vectors programatically.

src/deprecated.jl
```diff
@@ -105,3 +105,40 @@ function categorical!(df::DataFrame, cols::Union{Type, Nothing}=nothing;
     end
     return transform!(df, names(df, cols) .=> (x -> categorical(x, compress=compress)), renamecols=false)
 end
+
+@deprecate DataFrame(pairs::NTuple{N, Pair}; makeunique::Bool=false,
+          copycols::Bool=true) where {N} DataFrame(pairs..., makeunique=makeunique, copycols=copycols)
+@deprecate DataFrame(columns::NTuple{N, AbstractVector}, cnames::NTuple{N, Symbol}; makeunique::Bool=false,
+          copycols::Bool=true) where {N} DataFrame(collect(AbstractVector, columns), collect(Symbol, cnames);
```
nalimilan, line 112: Maybe give the simpler syntax? It will make a copy, but that was also the case with the tuple syntax. Same below.
```suggestion
          copycols::Bool=true) where {N} DataFrame(collect(columns), collect(cnames);
```

src/deprecated.jl
```diff
@@ -105,3 +105,40 @@ function categorical!(df::DataFrame, cols::Union{Type, Nothing}=nothing;
...
+                                                              gennames(length(columns)), copycols=copycols)
+@deprecate DataFrame(columns::AbstractMatrix, cnames::AbstractVector{Symbol} = gennames(size(columns, 2));
+                     makeunique::Bool=false) DataFrame(AbstractVector[columns[:, i] for i in 1:size(columns, 2)],
+                                                       cnames; makeunique=makeunique, copycols=false)
+
+@deprecate DataFrame(columns::AbstractMatrix, cnames::AbstractVector{<:AbstractString};
+                     makeunique::Bool=false) DataFrame(AbstractVector[columns[:, i] for i in 1:size(columns, 2)],
+                                                       Symbol.(cnames); makeunique=makeunique, copycols=false)
+
+function DataFrame(column_eltypes::AbstractVector{T}, cnames::AbstractVector{Symbol},
```
nalimilan, line 128: I'd be tempted to keep this one. It doesn't seem to make any harm, and the replacement is quite long if you want to allocate a nonzero number of rows. (Sorry for not commenting before.)

src/deprecated.jl
```diff
@@ -105,3 +105,40 @@ function categorical!(df::DataFrame, cols::Union{Type, Nothing}=nothing;
...
+          copycols::Bool=true) where {N} DataFrame(collect(AbstractVector, columns), collect(Symbol, cnames);
+              makeunique=makeunique, copycols=copycols)
+@deprecate DataFrame(columns::NTuple{N, AbstractVector}, cnames::NTuple{N, AbstractString}; makeunique::Bool=false,
+                     copycols::Bool=true) where {N} DataFrame(collect(AbstractVector, columns), [Symbol(c) for c in cnames];
+                                                              makeunique=makeunique, copycols=copycols)
+@deprecate DataFrame(columns::NTuple{N, AbstractVector};
+                     copycols::Bool=true) where {N} DataFrame(collect(AbstractVector, columns),
+                                                              gennames(length(columns)), copycols=copycols)
+@deprecate DataFrame(columns::AbstractMatrix, cnames::AbstractVector{Symbol} = gennames(size(columns, 2));
+                     makeunique::Bool=false) DataFrame(AbstractVector[columns[:, i] for i in 1:size(columns, 2)],
```
nalimilan, line 121: Doesn't this work? Same below.
```suggestion
                     makeunique::Bool=false) DataFrame(collect(eachcol(columns)),
```

src/dataframe/dataframe.jl
nalimilan: Yes splatting should be avoided when possible as it forces compilation of methods specialized on the number of arguments.

src/deprecated.jl
bkamins: This will not work on Julia 1.0 as `eachcol` is not defined there.

src/deprecated.jl
bkamins: I am largely indifferent. @oxinabox wanted to remove them (but maybe it was a mild preference?)

src/deprecated.jl
nalimilan: Ah too bad.

bkamins: I can use `Tables.table` instead. What do you think?

src/deprecated.jl
bkamins: this should be OK

src/deprecated.jl
nalimilan: Ah yeah that's probably better for users that can't rely on Julia > 1.0.

bkamins: OK. Then we need to wait a bit till Tables.jl gets a new release. But I will fix it then.

src/deprecated.jl
oxinabox: > It doesn't seem to make any harm

Every additional way to do the same thing makes it harder to learn to do the thing.
This one is just very out of place vs the others.
A particular problem is if people find this one first and think it is the normal one.
(In-fact, i think something like that happened to me years ago when i was first learning julia.
The only way to construct a `DataFrame` i knew was the `DataFrame(cnames, ctypes)` and I kept writing awkward code to use it.
Eventually this resulted in https://github.com/JuliaData/DataFrames.jl/pull/1416)
I might not have the history quite right, but I know I was doing something awkward and was confused by constructors.


This operation seems really obscure to me.
I am not sure exactly what circumstances  you encounter this, but the one i can kind of imagine is:
where `column_eltypes` and `cnames` and `nrows` are known to the author (not computed).
and for that it is clearer to write it out.
e.g for `column_eltypes=[Int, Bool, String]` and `cnames=["A","B","C"]` and `nrows=9`
```
DataFame(
    "A"=>Vector{Int}(undef, 4),
    "B"=>Vector{Bool}(undef, 4),
    "C"=>Vector{String}(undef, 4),
)
```
I feel like if you are computing `column_eltypes` and `cnames`  you are not using this constructor in the first place, are instead doing:
```
df = DataFrame()
col_names, col_types = compute_names_and_types()
for cname, ctype in zip(col_names, col_types)
    df[!, cname] = Vector{ctype}(undef, nrows)
end
```
Both of my code examples here seem cleaner to read than the
`DataFrame(cnames, ctypes; nrows)` constructor.

src/deprecated.jl
oxinabox: Compat.jl has `eachcol`

src/deprecated.jl
bkamins: @nalimilan - I am leaning towards removing them. The reason is that usually doing:
```
DataFame(
    "A"=>Vector{Int}(undef, 4),
    "B"=>Vector{Bool}(undef, 4),
    "C"=>Vector{String}(undef, 4),
)
```
or similar is a bad idea, as most likely later the user will start doing inefficient operations.

So as @oxinabox said:
1. experienced users will not cry when they are gone
2. new users - might think it is something that is recommended, while actually it should be discouraged (one should use other patterns, like: `push!`, `append!`, transformation functions etc. to populate columns; as likely if someone writes such code the next thing will be a loop over rows with `df[i,j] = something` inside)

nalimilan: OK, I'm not willing to die on that hill. Though should we really recommend using `push!` over allocating the full data frame upfront? The latter will generally overallocate.

src/deprecated.jl
nalimilan: Can we tag a new Tables.jl release now or are we waiting on a fix?

bkamins: It was tagged yesterday, so we are good to go.

I was waiting for https://github.com/queryverse/TableTraits.jl/pull/21 which is waiting for https://github.com/queryverse/TableTraits.jl/pull/21 (as discussed yesterday 😄), but since we deprecate `AbstractMatrix` constructor we do not have to wait for it now.

src/deprecated.jl
bkamins: Yes this is true that `push!` will be slower than preallocating and iterating.

@pdeffebach, @matthieugomez, @quinnj - can you comment if you think we should keep or drop the `eltype` constructor?

I have never (except for writing tests 😄) needed to use this constructor. And in cases I need an empty data frame I always prefer something like `DataFrame(a=Int[], b=Float64[])` as it reads more easily than `DataFrame([Int, Float64], [:a, :b])` as you clearly see which column gets what type.

bkamins: @matthieugomez - is :+1: for dropping it? (just to be sure as I tried to be neutral with the question)

matthieugomez:
Yes, my view is that it’s best to reduce the number of methods before 1.0.
It can always be added back later.

src/deprecated.jl
laborg: FWIW: I agree with oxinabox!

bkamins:
OK - so the decision is (unless I grossly misunderstood something 😄): we keep the current proposal (including deprecating `eltype` constructors).

I will wait for 1-2 days in case more comments come in and then update the whole package not to use deprecated functionality (mainly tests).

bkamins:
It should be good for a review.

bkamins:
@rofinn - can you please check if with this PR your use case works as expected?

rofinn:
Yep, looks good to me. I only really use the tables or pairs constructors anyways.

bkamins:
Last commit follows the discussion in https://github.com/JuliaData/Tables.jl/issues/204 and fixes an embarrassing case we had:
```
julia> DataFrame(Pair[])
1×1 DataFrame
│ Row │ makeunique │
│     │ Bool       │
├─────┼────────────┤
│ 1   │ 0          │
```

bkamins:
This should be good for a final check before merging.

test/broadcasting.jl
```diff
@@ -39,18 +39,18 @@ end
         @test (df .+ df) ./ 2 !== df
         @test df .+ Matrix(df) == 2 .* df
         @test Matrix(df) .+ df == 2 .* df
-        @test (Matrix(df) .+ df .== 2 .* df) == DataFrame(trues(size(df)), names(df))
+        @test (Matrix(df) .+ df .== 2 .* df) == DataFrame(Tables.table(trues(size(df)), header=names(df)))
```
nalimilan, line 42: Looks at these examples, it feels unfortunate to lose the symmetry between `Matrix(df)` and `DataFrame(mat)`, especially given the fact that broadcasting between data frames and matrices is allowed.

I think it makes sense to have some asymmetry, since `Matrix(df)` can just drop names, but `DataFrame(mat)` needs to generate names: so requiring `DataFrame(mat, colnames)` isn't illegitimate. But `DataFrame(Tables.table(df, header=colnames))` really breaks that symmetry (and it's really verbose as we see here). Maybe we could keep `DataFrame(mat, colnames)`, since in that case there's no ambiguity (`mat` is treated as a matrix, not as a Table.jl object).

(Sorry for suggesting this now. I can adjust the tests if you like.)

src/dataframe/dataframe.jl
```diff
@@ -12,19 +12,11 @@ DataFrame(columns::AbstractVector, names::AbstractVector{Symbol};
...
-          makeunique::Bool=false)
 DataFrame(kwargs...)
 DataFrame(pairs::Pair{Symbol,<:Any}...; makeunique::Bool=false, copycols::Bool=true)
 DataFrame(pairs::Pair{<:AbstractString,<:Any}...; makeunique::Bool=false,
           copycols::Bool=true)
 DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false, copycols::Bool=true)
-DataFrame(pairs::NTuple{N, Pair}; makeunique::Bool=false, copycols::Bool=true) where {N}
 DataFrame() # an empty DataFrame
 DataFrame(column_eltypes::AbstractVector, names::AbstractVector{Symbol},
           nrows::Integer=0; makeunique::Bool=false)
```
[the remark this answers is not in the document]
bkamins: we are. fixed

src/deprecated.jl
```diff
@@ -105,3 +105,43 @@ function categorical!(df::DataFrame, cols::Union{Type, Nothing}=nothing;
...
+@deprecate DataFrame(columns::NTuple{N, AbstractVector};
+                     copycols::Bool=true) where {N} DataFrame(collect(columns),
+                                                              Symbol.(:x, 1:length(columns)), copycols=copycols)
+
+# this is not a 100% correct deprecation as it does not support makeunique, but
+# we leave it as is to show users a recommended way to create a DataFrame from a matrix
+@deprecate DataFrame(columns::AbstractMatrix, cnames::AbstractVector{Symbol} = gennames(size(columns, 2));
+                     makeunique::Bool=false) DataFrame(Tables.table(columns, header=cnames))
+@deprecate DataFrame(columns::AbstractMatrix, cnames::AbstractVector{<:AbstractString};
+                     makeunique::Bool=false) DataFrame(Tables.table(columns, header=cnames))
```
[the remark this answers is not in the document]
bkamins: right - changed do Tables 1.1

src/other/tables.jl
```diff
@@ -43,11 +43,13 @@ fromcolumns(x, names; copycols::Bool=true) =
               copycols=copycols)
 
 function DataFrame(x::T; copycols::Bool=true) where {T}
-    if !Tables.istable(x)
-        if x isa AbstractVector && all(col -> isa(col, AbstractVector), x)
+    if !Tables.istable(x) && x isa AbstractVector && !isempty(x)
+        # here we handle the cases that are accepted by standard DataFrame constructors
+        # but eltype(x) is more flexible than assumed in these methods
```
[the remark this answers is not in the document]
bkamins: Exactly. What we handle is essentially `Any[...]` case (or some weird `Union{...}[...]`) as it will not be handled by standard constructors.

It was not very clear earlier and this is essentially what this code path is for only.

test/broadcasting.jl
bkamins: I am OK with this (I can fix the tests - it should be quick this way as we have `Tables.table` in all places we need to fix :smile:).

@oxinabox + @pdeffebach : can you comment here please?

The proposal is to have `DataFrame(::AbstractMatrix, ::Union{AbstractVector{Symbol}, AbstractVector{<:AbstractString})` signature (with no default for the second positional argument, so it would be required).

Then there would be no ambiguity with `DataFrame(table)` constructor that can handle matrices as tables.

src/other/tables.jl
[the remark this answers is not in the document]
nalimilan: OK. I think what confused me is "that are accepted by standard DataFrame constructors", since these are not really accepted, or they wouldn't have to be dealt with in this method. :-)

test/broadcasting.jl
oxinabox: Once again, I stand by my claim that every extra constructor adds confusion.
For this case i think nicer than using Tables.jl is:
```julia
DataFrame(names .=> eachcol(mat))
```
that is a nice clear explict statement about what is going on, and is I think clearer than
`DataFrame(mat, names)` since we are explictly stating that we are going to use the columns one per name.
Also naturally extends to:
`DataFrame(names .=> vec.(eachrow(mat))`
for constructing dataframes based on the rows of the matrix.
Particularly relevent as sometimes we do observations as columns, e.g. Distances.jl, the Models.jl interface, some others.

bkamins: @nalimilan - @oxinabox has convinced me. Writing `DataFrame(names .=> eachcol(mat))` is indeed clean. We have to use `Tables.table` for now because `eachcol` is not supported in Julia 1.0, but once we move the support to new Julia LTS we can clean it up.

nalimilan: What I find problematic is that by supporting `Matrix(df)` and `df .+ mat`, we consider that data frames and matrices are objects with compatible shapes. This alone means that we exclude the view of observations as columns (which BTW is the exception in the ecosystem, even in Distances you now have to explicitly give `dims`). Yet `DataFrame(names .=> eachcol(mat))` conceptually requires you to decompose the matrix into columns to create a data frame. The notion that shapes are compatibles is completely lost.

src/other/tables.jl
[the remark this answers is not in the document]
bkamins: Can you - as usual - propose a better wording (or what we have now is OK?). Thank you!

test/broadcasting.jl
bkamins: The shape is lost, but in `df .+ mat` we explicitly treat `mat` as not being a table (i.e. we assume that it does not have column names). What I mean by this is that `df .+ DataFrame(mat)` currently (i.e. before this PR) would generally fail because column names do not match, so `df .+ mat` does not differ conceptually here from `df .+ vector` or `df .+ scalar` (and we do not expect that `vector` or `scalar` can be converted to a `DataFrame`). In other words:
* broadcasting is about shapes (and that is why it works)
* `DataFrame` constructor is about being a table (which in particular implies having column names)

pdeffebach: I agree with this. note you can also do `df[1:4, 5:10] .= mat`, which is inconsistent.

src/deprecated.jl
Nosferican: I use the `DataFrame(::AbstractVector{DataType}, ::AbstractVector{<:Union{Symbol,String}}, ::Integer)` constructor quite a bit. I am a bit conflicted... Would have to update quite a bit of codebase but if convinced it helps I am okay with it. However, maybe I can force myself to write it more descriptive. So consider myself convinced and +1 on it.

test/broadcasting.jl
nalimilan: >     * `DataFrame` constructor is about being a table (which in particular implies having column names)

I don't think this rule holds in general, does it? It's only true for the one-argument constructor (and even in that case, not e.g. for the `::GroupedDataFrame` one). For the two-argument constructor, we don't actually allow tables as the first argument. So it would be OK to accept matrices (just like we accept vectors of vectors).

bkamins: > I don't think this rule holds in general, does it?

We have to have some minimal set of constructors. For instance I hate `DataFrame(a=[1,2], b=[2,3])` constructor as it is inconsistent with the general design we have, but I think we should keep it as it is convenient. How I understand @oxinabox is that he wants to limit the number of special cases as much as possible, and it seems that for matrices `DataFrame(names .=> eachcol(matrix))` is an easy enough pattern.

oxinabox: > For instance I hate DataFrame(a=[1,2], b=[2,3]) constructor as it is inconsistent with the general design we have

I was going to suggest deprecating that as well; for `DataFrame(:a=>[1,2], ...)` 😂 
because getting rid of it would mean we are free to add what ever keywords we wanted to the constructor later without clashing with column names.

bkamins: we have this clash already with `copycols` and `makeunique` kwargs which are handled inconsistently:
```
julia> DataFrame(makeunique=2, copycols=false)
1×1 DataFrame
│ Row │ makeunique │
│     │ Int64      │
├─────┼────────────┤
│ 1   │ 2          │
```

mtsch:
I'm just chiming in as requested on slack. In my use case, I have a `MyType<:AbstractVector` and I want a vector of those to act as a table. If I implement the Tables interface for `AbstractVector{<:MyType}` and pass it to the constructor, it dispatches on `::AbstractVector{<:AbstractVector}`. That fails as it's trying to treat each instance of `MyType` as a column.

To make this use case work, the constructor should check if the input implements Tables before dispatching on to other constructors.

bkamins:
Which means that in:
```
DataFrame(columns::AbstractVector{<:AbstractVector},
          cnames::AbstractVector{Symbol}=gennames(length(columns));
          makeunique::Bool=false, copycols::Bool=true)::DataFrame
```
we should not make `cnames` an optional argument but it should be required. Right? (this would be in line with @nalimilan proposal for matrices which goes along the same line of thinking)

nalimilan:
> I'm just chiming in as requested on slack. In my use case, I have a `MyType<:AbstractVector` and I want a vector of those to act as a table. If I implement the Tables interface for `AbstractVector{<:MyType}` and pass it to the constructor, it dispatches on `::AbstractVector{<:AbstractVector}`. That fails as it's trying to treat each instance of `MyType` as a column.

I'm not sure whether it's OK to implement the Tables.jl interface for `AbstractVector{<:MyType}` in general. That's not type piracy strictly speaking, but it's still quite broad.

@quinnj Have you considered this case?

If this kind of thing is supposed to be allowed, then this problem is the exact equivalent of the `AbstractMatrix` ambiguity issue that this PR intends to solve, but for `AbstractVector`. As @bkamins said we should drop automatic generations of column names, and require passing two arguments in addition to columns. That way when a single argument is passed it would always be a Tables.jl object (or a `GroupedDataFrame` but there's no ambiguity).

> To make this use case work, the constructor should check if the input implements Tables before dispatching on to other constructors.

Unfortunately that's not really enough, as this just amounts to solving the ambiguity by making the opposite, but still somewhat arbitrary choice. What if the user really intended to call the constructor which interprets the first argument as a vector of vectors? We need to handle all `AbstractVector`s consistently to avoid surprising behaviors, which break generic programming.

bkamins:
After some discussions with @nalimilan I came to this conclusion (I know this will not make everyone happy, but there are conflicting opinions here and at some point we have to make some decisions; I also know I have changed my mind several times here, but this is only because our aim is to provide a consistent yet convenient API and in as non-breaking way as possible, and these are conflicting objectives):
* If a single positional argument is passed to `DataFrame` then it must be a table and is handled consistently with Tables.jl; exceptions are:
    - `DataFrame(::GroupedDataFrame)`
    - `DataFrame(::Pair)`
    - `DataFrame(::AbstractVector{<:Pair})`
    - `DataFrame(::AbstractDict)` (here the only difference is that DataFrames.jl is more flexible than Tables.jl in what is considered to be a table)
* we leave kwarg form `DataFrame(a=1, b=2)` for legacy reasons (there is way to much code that uses this; I will only make a change that disallows `makeunique` as column name as it is confusing; also R users will expect such thing to work)
* if there are more than one positional argument passed to `DataFrame` constructor then DataFrames.jl is free to define what is accepted and in what way. In particular we will accept (I am aware that this is a bit inconsistent but we have to weigh different voices and in particular user convenience here - especially people coming e.g. from R, when such conversion is a normal thing):
    - `DataFrame(::Pair...)`
    - `DataFrame(::AbstractVector, colnames)`
    - `DataFrame(::AbstractMatrix, colnames)` (sorry for this, but again - retaining it is intended to make the life easier for people that used this constructor; we can consider dropping it in 2.0 release)
* expose `gennames` function (that is currently internal) so that people can easily change their legacy code e.g. `DataFrame(matrix)` into `DataFrame(matrix, gennames(size(matrix, 2))`
* update the documentation to clearly indicate the logic of the design I have described here

If you feel like it please :+1:  or :-1: this proposal under the comment, but unless there are some really good reasons raised I am inclined to follow the rules described above (as said at the top - at some point we need to make a tie-braeaker decision; at least I am convinced that we can clearly lay down the rules we use).

laborg:
My arguments against the kwarg constructor:
- Yes, there is a lot of code out there that uses the kwarg constructor, but my bet is that this code examples will fail due to other syntax changes. I don't see much benefit in being "slightly" breaking. (I'm unclear if a nice deprecation is possible). 
- R users expecting a 1:1 dplyr interface will be "dissappointed" either way so being "mildly" compatible with R doesn't count much in my book. On the contrary maybe R users will want to switch because they want to have a fresh take on things and for those historical bag of inconsistencies might be a downer. 
- It has been stated a couple of times that other packages will offer different/extended APIs for `DataFrames.jl` (e.g. `DataFramesMeta.jl`) - following this argument, wouldn't it be fitting to have a concise and clean set of constructors in `DataFrames.jl` and let other packages go crazy on the construction site?

jkrumbiegel:
As the `DataFrame(pairs...)` constructor is syntactically really close to the keyword syntax, is it really so bad if people have to change occurances of the keyword syntax after this breaking change? It's only a symbol replacement. Additionally, for people coming over from R and just trying things out without reading docs, we could add a helpful error message whenever unknown keyword arguments are given to DataFrame. Not just error because no method with those keywords could be found, but instead 
```
Unknown keywords `mycol`, `mycol2`,... encountered.
If you want to create a DataFrame with keywords for columns, use the pair syntax instead:
`DataFrame(keyword => column, keyword2 => column2...)`
```

oxinabox:
I don't entirely agree with the proposal, but, it will do. :shipit:

test/broadcasting.jl
rofinn: > I was going to suggest deprecating that as well; for DataFrame(:a=>[1,2], ...)

Oh, is that an option here? I'm 100% on board with that change 😸

rofinn:
> (I'm unclear if a nice deprecation is possible).

I don't see why not. Just throw a `depwarn` for unknown kwargs and then call the pairs constructor with them.

pdeffebach:
Bikeshed: 

I would prefer 

```
DataFrame([1, 2, 3] => :a, [4, 5, 6] => :b)
```

for it's consistency with `transform`.

Nosferican:
What about the `zip(colnames, values)`?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

quinnj:
I agree (if I understood correctly) @bkamins proposal of keeping the kwargs constructor; it's super convenient and very easy to use and very familiar to users of other languages. I don't see a strong reason for removing it.

bkamins:
Thank you all for commenting.

As I have said earlier there are many different angles to this problem, but some decision has to be made to move forward.

Given the feedback collected I will implement what is proposed in https://github.com/JuliaData/DataFrames.jl/pull/2464#issuecomment-708262422 and this will be the functionality for 1.0 release.

Let me summarize it in a shorter way: one-argument `DataFrame` constructor is consistent with Tables.jl with a few exceptions that are desirable to introduce and are not going to be conflicting. Zero or two or more argument `DataFrame` constructors are reserved for DataFrames.jl decision what to do and for now we opt to leave some of them for user convenience.

To explain what I mean by "not going to be conflicting" is essentially this case, which is often found confusing by the users:
```
julia> DataFrame([:a => [1,2], :b => [3,4]]) # this is what users find intuitive to get
2×2 DataFrame
│ Row │ a     │ b     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 3     │
│ 2   │ 2     │ 4     │

julia> DataFrame(Tables.columntable([:a => [1,2], :b => [3,4]])) # this is what users find very confusing
2×2 DataFrame
│ Row │ first  │ second │
│     │ Symbol │ Array… │
├─────┼────────┼────────┤
│ 1   │ a      │ [1, 2] │
│ 2   │ b      │ [3, 4] │
```

bkamins:
@nalimilan - I have finished the updates of implementation, tests and documentation for this PR. It should be good for a review. Thank you!

As noted in https://github.com/JuliaData/Tables.jl/issues/208 deprecation of `DataFrame(AbstractVector{<:AbstractVector})` is the most problematic thing, as in some cases this will fall back to `DataFrame(table)` constructor producing a significantly different results after we drop the deprecation.

bkamins:
@nalimilan - CI passes here

bkamins:
Any more comments on this PR?

bkamins:
This is the last major PR regarding the design before we tag 0.22 release. So if you have any comments here (or in general) please raise them now.

The plan is that 0.22 will mean breaking-changes freeze. I am not sure how fast 1.0 will go out, as we might add new functionalities, but it will be relatively soon after it and between 0.22 and 1.0 no breaking changes (unless related to bugs) will be introduced.

Thank you!

src/dataframe/dataframe.jl
```diff
@@ -239,55 +282,44 @@ DataFrame(columns::AbstractVector, cnames::AbstractVector{<:AbstractString};
...
-
-DataFrame(columns::NTuple{N, AbstractVector}; copycols::Bool=true) where {N} =
-    DataFrame(collect(AbstractVector, columns), gennames(length(columns)),
-              copycols=copycols)
+function DataFrame(columns::AbstractVector, cnames::Symbol; copycols::Bool=true)
+    if cnames !== :gennames
+        throw(ArgumentError("if the first positional argument to DataFrame " *
+                            "constructor is a vector of vectors and a second" *
+                            " positional argument is passed then the second " *
+                            "argument must be a list of column names or :gennames"))
```
nalimilan, line 300:
```suggestion
        throw(ArgumentError("if the first argument to DataFrame " *
                            "constructor is a vector of vectors, the second " *
                            "argument must either be omitted or be a " *
                            "vector of column names or :gennames"))
```

src/dataframe/dataframe.jl
```diff
@@ -239,55 +282,44 @@ DataFrame(columns::AbstractVector, cnames::AbstractVector{<:AbstractString};
...
-                             Tables.allocatecolumn(elty, nrows)
-                             for elty in column_eltypes]
-    return DataFrame(columns, Index(convert(Vector{Symbol}, cnames),
-                     makeunique=makeunique), copycols=false)
+function DataFrame(columns::AbstractMatrix, cnames::Symbol)
+    if cnames !== :gennames
+        throw(ArgumentError("if the first positional argument to DataFrame " *
+                            "constructor is a matrix and a second" *
+                            " positional argument is passed then the second " *
+                            "argument must be a list of column names or :gennames"))
```
nalimilan, line 318:
```suggestion
        throw(ArgumentError("if the first argument to DataFrame " *
                            "constructor is a vector of vectors, the second " *
                            "argument must either be omitted or be a " *
                            "vector of column names or :gennames"))
```

src/dataframe/dataframe.jl
```diff
@@ -881,10 +913,10 @@ hcat!(df1::DataFrame, df2::DataFrame;
            makeunique=makeunique, copycols=copycols)::DataFrame
 
 hcat!(df::DataFrame, x::AbstractVector; makeunique::Bool=false, copycols::Bool=true) =
-    hcat!(df, DataFrame(AbstractVector[x], copycols=copycols),
+    hcat!(df, DataFrame(:x1 => x, copycols=copycols),
```
nalimilan, line 916: Wouldn't it be more direct to pass `AbstractVector[x], [:x1]`?

test/broadcasting.jl
```diff
@@ -4,10 +4,10 @@ using Test, DataFrames, PooledArrays, Random, CategoricalArrays
 
 const ≅ = isequal
 
-refdf = DataFrame(reshape(1.5:15.5, (3,5)))
+refdf = DataFrame(reshape(1.5:15.5, (3,5)), [:x1, :x2, :x3, :x4, :x5])
```
nalimilan, line 7: What about using `:gennames` here and elsewhere?

src/other/tables.jl
```diff
@@ -43,11 +43,11 @@ fromcolumns(x, names; copycols::Bool=true) =
...
 
 function DataFrame(x::T; copycols::Bool=true) where {T}
-    if !Tables.istable(x)
-        if x isa AbstractVector && all(col -> isa(col, AbstractVector), x)
-            return DataFrame(Vector{AbstractVector}(x), copycols=copycols)
-        elseif (x isa AbstractVector || x isa Tuple) &&
-            all(v -> v isa Pair{Symbol, <:AbstractVector}, x)
+    if !Tables.istable(x) && x isa AbstractVector && !isempty(x)
+        # here we handle the case that is accepted a DataFrame constructor
+        # defined for a vector of `Pair`s but eltype(x) is less specific
```
nalimilan, line 48:
```suggestion
        # here we handle eltypes not specific enough to be dispatched
        # to other DataFrames constructors taking vector of `Pair`s
```

test/broadcasting.jl
```diff
@@ -1152,7 +1152,7 @@ end
         end
         df6 = copy(df2)
         df7 = copy(df3)
-        df4 = DataFrame(sin.(df1[1,1] .+ copy(df1[!, 1]) .+ Matrix(df2) ./ Matrix(df3)))
+        df4 = DataFrame(sin.(df1[1,1] .+ copy(df1[!, 1]) .+ Matrix(df2) ./ Matrix(df3)), names(df3))
```
nalimilan, line 1155: Here and below `:gennames` would also be more obvious.

src/dataframe/dataframe.jl
```diff
@@ -8,105 +8,144 @@ particularly a Vector or CategoricalVector.
...
+DataFrame(kwargs..., copycols::Bool=true)
+
+DataFrame(columns::AbstractVecOrMat, names::Union{AbstractVector, Symbol};
+          makeunique::Bool=false, copycols::Bool=true)
+
+DataFrame(::DataFrameRow)
 DataFrame(::GroupedDataFrame; keepkeys::Bool=true)
 ```
 
-# Arguments
```
nalimilan, line 39: IMHO it's useful to have a list of arguments to make it easier to find relevant information, at least for keyword arguments.

src/dataframe/dataframe.jl
```diff
@@ -8,105 +8,144 @@ particularly a Vector or CategoricalVector.
...
+
+julia> DataFrame([[1, 2], [0, 0]], [:a, :b]) # vector of vectors constructor
+2×2 DataFrame
+│ Row │ a     │ b     │
+│     │ Int64 │ Int64 │
+├─────┼───────┼───────┤
+│ 1   │ 1     │ 0     │
+│ 2   │ 2     │ 0     │
+
+julia> DataFrame([1 0; 2 0], :gennames) # matrix constructor
```
nalimilan, line 142: Final bikeshedding session: I wonder whether `:gennames` is the best name. "gen" is not a super standard abbreviation and the double "n" isn't terribly nice. Maybe just `:auto` would be enough? Since the second argument is always for names, repeating "names" might be redundant. Opinions? :-)

nalimilan (commented):
Thanks! Looks mostly good I'd say.

src/dataframe/dataframe.jl
```diff
@@ -182,14 +221,15 @@ function DataFrame(pairs::Pair{<:AbstractString,<:Any}...; makeunique::Bool=fals
...
-    DataFrame(pairs..., makeunique=makeunique, copycols=copycols)
+# this is needed as a workaround for Tables.jl dispatch
+function DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false,
+                   copycols::Bool=true)
+    if isempty(pairs)
+        return DataFrame()
+    else
+        return DataFrame(pairs..., makeunique=makeunique, copycols=copycols)
+    end
+end
```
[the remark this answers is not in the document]
bkamins: no, because `DataFrame(makeunique=makeunique, copycols=copycols)` as it will dispatch to kwarg constructor in which  `makeunique` or `copycols` keyword arguments are not allowed. I will fix the splatting.

src/dataframe/dataframe.jl
bkamins: I will word it differently. I would not comment here on omitting the second argument (as user passed the second argument). It would lead to a confusion as passing one argument triggers Tables.jl that will lead to a completely different behavior.

src/dataframe/dataframe.jl
bkamins: same here.

src/dataframe/dataframe.jl
bkamins: OK - I change it to `:auto`

src/dataframe/dataframe.jl
```diff
@@ -8,105 +8,144 @@ particularly a Vector or CategoricalVector.
...
-DataFrame(columns::NTuple{N,AbstractVector}, names::NTuple{N,<:AbstractString};
-          makeunique::Bool=false, copycols::Bool=true)
-DataFrame(columns::Matrix, names::AbstractVector{Symbol}; makeunique::Bool=false)
-DataFrame(columns::Matrix, names::AbstractVector{<:AbstractString};
-          makeunique::Bool=false)
-DataFrame(kwargs...)
-DataFrame(pairs::Pair{Symbol,<:Any}...; makeunique::Bool=false, copycols::Bool=true)
-DataFrame(pairs::Pair{<:AbstractString,<:Any}...; makeunique::Bool=false,
-          copycols::Bool=true)
+DataFrame(table; copycols::Bool=true)
```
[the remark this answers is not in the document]
bkamins: ok

src/dataframe/dataframe.jl
```diff
@@ -8,105 +8,144 @@ particularly a Vector or CategoricalVector.
...
-`DataFrame(pairs::AbstractVector{<:Pair})`, and `DataFrame(pairs::NTuple{N, Pair})` form
-constructors which additionally allow a column to be of any other type that is
-not an `AbstractArray`, in which case the passed value is automatically repeated
+If a single positional argument is passed to a `DataFrame` constructor then it
+is assumed to be of type that implements the
+[Tables.jl](https://github.com/JuliaData/Tables.jl) interface using which the
+returned `DataFrame` is materialized.
+
+In addition to this general constructor there are several other convenience
+constructors provided. They are described below.
```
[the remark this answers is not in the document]
bkamins: ok - rewritten.

src/dataframe/dataframe.jl
```diff
@@ -8,105 +8,144 @@ particularly a Vector or CategoricalVector.
...
+
+It is allowed to pass a vector of `Pair`s, a list of `Pair`s as positional
+arguments, or a list of keyword arguments. In this case each pair is considered
+to represent a column name to column value mapping and column name must be a
+`Symbol` or string. Alternatively a dictionary can be passed to the constructor
+in which case its entries are considered to define the column name and column
+value pairs with the exception that in this case column names can be of any type
+and will be always converted to a `Symbol`. In all these constructors column
```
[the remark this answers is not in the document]
bkamins: I have fixed the description. Actually it was related with a small bug in the implementation of this constructor that I have now fixed.

src/dataframe/dataframe.jl
```diff
@@ -8,105 +8,144 @@ particularly a Vector or CategoricalVector.
...
-using `setindex!` then its reference is stored in the `DataFrame`. An exception
-to this rule is assignment of an `AbstractRange` as a column, in which case the
-range is collected to a `Vector`.
+`DataFrame` object is designed to allow column types to vary and to be
+dynamically changed also after it is constructed. Therefore `DataFrame`s are not
+type stable.
 
-Because column types can vary, a `DataFrame` is not type stable. For
-performance-critical code, do not index into a `DataFrame` inside of loops.
+For performance-critical code that requires type-stability either use the
```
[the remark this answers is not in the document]
bkamins: done

src/dataframe/dataframe.jl
bkamins: I have added kwargs description

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
bkamins: fixed the constructor.

src/dataframe/dataframe.jl
bkamins: I left the original wording (fixing only the stylistic corrections).

src/dataframe/dataframe.jl
bkamins: ok - fixed

test/broadcasting.jl
bkamins: ok - I will change this

test/broadcasting.jl
bkamins: where I use `names` I test if the names from some concrete data frame are reused, I remove the explicit `[:x1, :x2]` and similar with `:auto`.

bkamins:
I have applied all the suggestions. It should be good for another look at.

bkamins:
CI passes cleanly (I am testing against current master with all the changes made since this PR was started)

src/dataframe/dataframe.jl
```diff
@@ -227,19 +236,23 @@ function DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false,
     if isempty(pairs)
         return DataFrame()
     else
-        return DataFrame(pairs..., makeunique=makeunique, copycols=copycols)
+        if !(all(((k,v),) -> k isa Symbol, pairs) || all(((k,v),) -> k isa AbstractString, pairs))
+            throw(ArgumentError("All column names must be Symbols or strings"))
+        end
+        colnames = [Symbol(k) for (k,v) in pairs]
+        columns = Any[v for (k,v) in pairs]
```
nalimilan, line 243: Doesn't this avoid a copy?
```suggestion
        columns = AbstractVector[v for (k,v) in pairs]
```

src/dataframe/dataframe.jl
```diff
@@ -227,19 +236,23 @@ function DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false,
...
-        colnames = sort!(collect(keys(d)))
-    else
-        colnames = keys(d)
-    end
-    colindex = Index([Symbol(k) for k in colnames])
-    columns = Any[d[c] for c in colnames]
-    DataFrame(columns, colindex, copycols=copycols)
+    colnames = [Symbol(k) for k in keys(d)]
+    colindex = Index(colnames)
+    columns = Any[v for v in values(d)]
```
nalimilan, line 252:
```suggestion
    columns = AbstractVector[v for v in values(d)]
```

test/constructors.jl
```diff
@@ -337,4 +344,11 @@ end
     end
 end
 
+@testset "Dict constructor corner case" begin
+    @test DataFrame(Dict('a' => 1, true => 2)) == DataFrame("a" => 1, "true" => 2)
```
nalimilan, line 348: Should we really allow this?

bkamins: This was the original design. Actually I thought that maybe it is best to allow only string or `Symbol` as keys - again. It is last time to change it :). Do we go this way?

src/dataframe/dataframe.jl
bkamins: This and below is incorrect as `v` does not have to be a vector - we allow pseudo broadcasting here. Here is an example on 0.21:
```
julia> using DataFrames

julia> DataFrame([:A => 1, :B => 11:12])
2×2 DataFrame
│ Row │ A     │ B     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 1     │ 11    │
│ 2   │ 1     │ 12    │
```
I think we should keep it so that `DataFrame(vec_of_pairs)` and `DataFraem(vec_of_pairs...)` behave the same way.

src/dataframe/dataframe.jl
bkamins: same here - pseudo broadcasting is allowed.

test/constructors.jl
nalimilan: That sounds safer -- as usual. :-)

bkamins: OK - I will change this then. Thank you!

bkamins: done - now we are as strict as with `Pair` constructors (except for sorting when `Dict` is passed).

bkamins:
CI passes, only coverage fails

src/dataframe/dataframe.jl
```diff
@@ -247,7 +244,14 @@ function DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false,
 end
 
 function DataFrame(d::AbstractDict; copycols::Bool=true)
-    colnames = [Symbol(k) for k in keys(d)]
+    if all(k -> k isa Symbol, keys(d))
+        colnames = collect(Symbol, keys(d))
+    elseif all(k -> k isa AbstractString, keys(d))
+        colnames = [Symbol(k) for k in keys(d)]
+    else
+        throw(ArgumentError("All column names must be Symbols or strings"))
```
nalimilan, line 252:
```suggestion
        throw(ArgumentError("All column names must be either Symbols or strings (mixing is not allowed)"))
```

nalimilan (approved):

src/dataframe/dataframe.jl
```diff
@@ -182,24 +227,36 @@ function DataFrame(pairs::Pair{<:AbstractString,<:Any}...; makeunique::Bool=fals
...
-          copycols::Bool=true) where {N} =
-    DataFrame(pairs..., makeunique=makeunique, copycols=copycols)
+# this is needed as a workaround for Tables.jl dispatch
+function DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false,
+                   copycols::Bool=true)
+    if isempty(pairs)
+        return DataFrame()
+    else
+        if !(all(((k,v),) -> k isa Symbol, pairs) || all(((k,v),) -> k isa AbstractString, pairs))
+            throw(ArgumentError("All column names must be Symbols or strings"))
```
bkamins, line 237:
```suggestion
            throw(ArgumentError("All column names must be either Symbols or strings"))
```

bkamins:
Thank you! I will merge this once the CI passes.

src/dataframe/dataframe.jl
nalimilan: Did you leave out "(mixing is not allowed)" on purpose?

bkamins: No - I just reused error message from above (as it was identical). I now fixed them both. Actually in the tests I added a check against disallowing mixing.

src/dataframe/dataframe.jl
```diff
@@ -182,24 +227,36 @@ function DataFrame(pairs::Pair{<:AbstractString,<:Any}...; makeunique::Bool=fals
...
-          copycols::Bool=true) where {N} =
-    DataFrame(pairs..., makeunique=makeunique, copycols=copycols)
+# this is needed as a workaround for Tables.jl dispatch
+function DataFrame(pairs::AbstractVector{<:Pair}; makeunique::Bool=false,
+                   copycols::Bool=true)
+    if isempty(pairs)
+        return DataFrame()
+    else
+        if !(all(((k,v),) -> k isa Symbol, pairs) || all(((k,v),) -> k isa AbstractString, pairs))
+            throw(ArgumentError("All column names must be either Symbols or strings"))
```
nalimilan, line 237: Still missing?
```suggestion
            throw(ArgumentError("All column names must be either Symbols or strings (mixing is not allowed)"))
```

bkamins: I thought "either" meant that :smile:, but I will add your comment in both places then.

bkamins:
Thank you!

0kto:
Hi all, I am late to this discussion.
I updated yesterday, and with v0.22 I receive the deprecation warning:
```
DataFrame` constructor with passed eltypes is deprecated.
[...]
```
Reading through the docs and the source code, I was not able to find a suitable replacement:
In my scripts I initialize a DataFrame object by defining column-names and types (and zero rows), to which I append data read from (rather messy) instrument data files.  Having the eltypes defined before appending data to the DataFrame, bad data (where the Type is not matching) fails and immediately helps figuring out if there is a problem.  Also, adding a column is as easy as adding a Dictionary entry, making my code easily usable for non-developers.

So, my question is if there is a suitable replacement I haven't seen, or if there is a recommended workflow I could implement instead. Or if at least one constructor, that accepts `Dict{Symbol,Type}( :colname => eltype)`.

pdeffebach:
How about this? I think it's decently intuitive

```
julia> d = Dict(:id => String, :income => Float64, :education => Int64);

julia> df = DataFrame([n => T[] for (n, T) in d])
0×3 DataFrame
```

bkamins:
or just e.g.:
```
DataFrame([:a, :b] .=> [T[] for T in [Int, Float64]])
```

`Dict` is not 100% correct as it will reorder columns.

0kto:
Thanks, that is really helpful.
Since I am using an `OrderedDict` from the DataStructures.jl package, order should be preserved.
Cheers!

bkamins:
Yes only `Dict` is sorted.

PR #2466, opened 2020-10-03
fix setproperty! for data frame

bkamins:
Without it if you pass incorrect types a default `setproperty!` method is invoked (user would have to try very hard to exploit this, but still I think it is better to throw an informative error in this case). The problem was hidden earlier by deprecations.

Before:
```
julia> df = DataFrame(a=1);

julia> dfv = @view df[:, :];

julia> @edit dfv.parent = DataFrame();

julia> df.a = 1
ERROR: type DataFrame has no field a
Stacktrace:
 [1] setproperty!(::DataFrame, ::Symbol, ::Int64) at ./Base.jl:34
 [2] top-level scope at REPL[12]:1

julia> dfv.parent = DataFrame();
ERROR: setfield! immutable struct of type SubDataFrame cannot be changed
Stacktrace:
 [1] setproperty!(::SubDataFrame{DataFrame,DataFrames.Index,Base.OneTo{Int64}}, ::Symbol, ::DataFrame) at ./Base.jl:34
 [2] top-level scope at REPL[5]:1
```

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -499,10 +499,12 @@ Base.setproperty!(df::DataFrame, col_ind::Symbol, v::AbstractVector) =
     (df[!, col_ind] = v)
 Base.setproperty!(df::DataFrame, col_ind::AbstractString, v::AbstractVector) =
     (df[!, col_ind] = v)
-Base.setproperty!(::AbstractDataFrame, ::Symbol, ::Any) =
-    throw(ArgumentError("it is only allowed to pass a vector as a column of a DataFrame"))
-Base.setproperty!(::AbstractDataFrame, ::AbstractString, ::Any) =
-    throw(ArgumentError("it is only allowed to pass a vector as a column of a DataFrame"))
+Base.setproperty!(::DataFrame, col_ind::Symbol, v::Any) =
+    throw(ArgumentError("It is only allowed to pass a vector as a column of a DataFrame." *
+                        "Instead use `df[!, col_ind] .= v` if you want to use broadcasting."))
+Base.setproperty!(::DataFrame, col_ind::AbstractString, v::Any) =
+    throw(ArgumentError("It is only allowed to pass a vector as a column of a DataFrame." *
+                        "Instead use `df[!, col_ind] .= v` if you want to use broadcasting."))
 
 # df[SingleRowIndex, SingleColumnIndex] = Single Item
 function Base.setindex!(df::DataFrame, v::Any, row_ind::Integer, col_ind::ColumnIndex)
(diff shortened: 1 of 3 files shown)


src/dataframe/dataframe.jl
```diff
@@ -499,6 +499,10 @@ Base.setproperty!(df::DataFrame, col_ind::Symbol, v::AbstractVector) =
     (df[!, col_ind] = v)
 Base.setproperty!(df::DataFrame, col_ind::AbstractString, v::AbstractVector) =
     (df[!, col_ind] = v)
+Base.setproperty!(::AbstractDataFrame, ::Symbol, ::Any) =
+    throw(ArgumentError("it is only allowed to pass a vector as a column of a DataFrame"))
+Base.setproperty!(::AbstractDataFrame, ::AbstractString, ::Any) =
+    throw(ArgumentError("it is only allowed to pass a vector as a column of a DataFrame"))
```
nalimilan, line 505: Would it make sense to point people to `df[:, col] .= v`? This is a very common mistake.

bkamins: Right. I will have to add more methods then. For `DataFrame` it should be `df!, col] .= v` (replacing a column), and for `SubDataFrame` either it should keep erroring or use `df:, col] .= v` (updating in place).

What do you think would be a better hint for `SubDataFrame`?

nalimilan: `SubDataFrame` doesn't support `setproperty!` at all, right? So we can point to both `sdf[:, col] = v` and  `sdf[!, col] .= v`, noting that these are in-place.

BTW, methods for `AbstractDataFrame` should probably got to abstractdataframe.jl (or only define them for `DataFrame` if we don't define working methods for `AbstractDataFrame`).

src/dataframe/dataframe.jl
```diff
@@ -499,6 +499,12 @@ Base.setproperty!(df::DataFrame, col_ind::Symbol, v::AbstractVector) =
     (df[!, col_ind] = v)
 Base.setproperty!(df::DataFrame, col_ind::AbstractString, v::AbstractVector) =
     (df[!, col_ind] = v)
+Base.setproperty!(::DataFrame, col_ind::Symbol, v::Any) =
```
nalimilan, line 502: `col_ind` and `v` aren't used.

nalimilan (approved):

src/dataframe/dataframe.jl
bkamins: I added them because they are referenced in the error message, so I thought to add them, as when the user later goes to the method definition it is clear what refers to what (but if you prefer I can remove them).

nalimilan: I don't really mind.

bkamins:
Thank you!

Issue #2354, opened 2020-08-05
allow filter to return a view


bkamins:
This has been discussed in several places I create a separate issue for this to keep track of it as it is an important functionality I think. What we want is `filter(predicate, df, view=true)` to return a view (so then we can conveniently update this view for example).

bkamins:
Now given https://github.com/JuliaData/DataFrames.jl/issues/2211#issuecomment-669730638 and https://github.com/JuliaData/DataFrames.jl/issues/2211#issuecomment-669730638

my question is if we should not also add `dropmissing` keyword argument that would:
* if `nothing` do what we have now
* if `true` then missings would be excluded from the selection
* if `false` then missings would be included in the selection

This is essentially as adding `coalesce` wrapper to the filtering function (so maybe we feel that it is not needed), but maybe you will judge that:
```
filter(:x => >(1), df, dropmissing=yes)
```
reads better than:
```
filter(:x => x -> coalesce(x > 1, false), df)
```
we do not save typing here, but maybe you will find it more readable?

CC @nalimilan @matthieugomez @pdeffebach

matthieugomez:
I think I would prefer skipmissing, where true means to drop missing values, and false errors if there are missing value. I am not sure it is worth introducing another kind of missing argument.

bkamins:
A good point. But I would not alter the defaults (sorry for this position here but I do want to avoid breaking changes of this kind till 1.0).

bkamins:
The only issue is that `skipmissing` in `select` etc. will have a bit different behavior than `skipmissing` here.

matthieugomez:
Personally, I prefer an argument that always have the same name (skipmissing), even if it has slightly different meanings in different contexts, rather than a different keyword argument everytime. I am not a fan on how R has na.rm, na.exclude, na.omit etc.

bkamins:
(ok - and I would appreciate a comment in #2314 what you think should be the results on the cases I have listed there :smile:)

bkamins:
I was thinking about these things and I would close #2211, #2314 and #2258 in favour of this issue which I would aim to solve all the issues raised in one consistent design.

So my proposal would be the following:
1. in `filter` and `dropmissing` functions if they take `DataFrame` or `SubDataFrame` with index `Index` (not just any `AbstractDataFrame`, and not for `filter!` and `dropmissing!`) add `view::Bool=flase` keyword argument. Of course `view` also allows to create it
[...]

Issue #2323, opened 2020-07-21
`filter` on an `AbstractDataFrame` should take in columns


pdeffebach:
I think it's a bit frustrating that `select`, `transform`, etc. operate on columns as the default, but `filter` operates on rows. I would be find with broadcasting. Plus, it makes it easier to filter with operations like `x -> x .> mean(x)`

bkamins:
Let us discuss it, but I think it is not likely to change. It is actually `select`, `transform` and `combine` that break the default behavior that data frame is a collection of rows. Also what you write is achievable by `getindex`:
```
df[df.x .> mean(df.x), :]
```

pdeffebach:
Thanks for considering this. 

I like piping, and when I (eventually) run a lab I will try to make piping part of my style guide because it "reads well". That is to say, there are benefits to allowing this kind of behavior in `filter` even though it is achievable with `getindex`. 

We can add `ByRow` wrappers to `filter` that would allow individuals to opt in to ByRow behaviors. We can also provide easier syntax for `ByRow` behavior in DataFramesMeta. 

I guess maybe I would argue consistency within piping is more important than consistency with the general concept of a data frame as a collection of rows.

matthieugomez:
I see the benefit in having the same grammar for `select`/`transform` and `filter` (whether it's about taking rows or columns by default). W
hy were `select/`transform` decided to be column-wise? Wouldn't the same arguments apply to `filter`? (or conversely, if there is a strong reason for `filter` to be row-wise, should not `select` and `transform` be row-wise by default too?)

pdeffebach:
I confess I'm the one who made the biggest stink about `transform` acting on columns as noted in #1952. In my mind this was the best "one-to-one" mapping for working with `df.x` etc. It also makes it easier to do `x .- mean(x)`. 

After working with the new syntax I do think that adding `.` everywhere gets kind of annoying and having `ByRow` be the default and a `ByCols` wrapper would also have been a good decision.

matthieugomez:
It makes sense that `select` and `transform` use columns by default — esp. now that they are used with GroupedDataFrame.

Could `filter` be extended to accept the same syntax as `select`/`transform` when a DataFrame is given in the first argument?

As for the annoyance about using `.` everywhere, maybe `map` could be used for a row-wise `transform`.
```
map(:y => y + 1, df)
# gives the same thing as
transform(df, :y => y .+ 1)
```

pdeffebach:
[...]

Issue #2465, opened 2020-10-01
Add method to filter on `Bool` column symbols


jkbest2:
Would it be reasonable to define a method `filter(b::Symbol, df::DataFrame)` that attempts to coerce `df.b` to `Bool` and then `filter`s on that?

bkamins:
it is an interesting idea. I think it should rather ho to `where`, see https://github.com/JuliaData/DataFrames.jl/issues/2323, as we tend toward leaving only `filter(fun, data_frame)` in the long run. Let us see how things go.

Note that already you can write `df[df.b, :]` to get what you want.

bkamins:
Closing it, as it is implemented as `subset(df, :b)` now.

PR #2467, opened 2020-10-03
add WhereDataFrame

matthieugomez:
This pull request fleshes out a `WhereDataFrame` type, constructed using `where`.

 `filter/filter!`, `delete!`, `transform/transform!`, `combine`,  `view` operate on `WhereDataFrames`

It is similar in spirit to SQL where, data.table i, or Stata if. 
PR would solve #2354, #2323, #2211

### DIFF
--- a/src/wheredataframe/wheredataframe.jl
+++ b/src/wheredataframe/wheredataframe.jl
@@ -10,7 +10,7 @@ struct WhereDataFrame{D<:AbstractDataFrame, T<:AbstractVector{Int}}
     rows::T
 end
 
-Base.@propagate_inbounds function WhereDataFrame(parent::AbstractDataFrame, rows::AbstractVector{Bool})
+function WhereDataFrame(parent::AbstractDataFrame, rows::AbstractVector{Bool})
     if length(rows) != nrow(parent)
         throw(ArgumentError("invalid length of `AbstractVector{Bool}` row index" *
                             " (got $(length(rows)), expected $(nrow(parent)))"))
@@ -127,7 +127,7 @@ end
 
 ##############################################################################
 ##
-## Common Operations
+## Operations
 ##
 ##############################################################################
 
@@ -135,30 +135,16 @@ Base.filter(wdf::WhereDataFrame) = parent(wdf)[rows(wdf), :]
 Base.filter!(wdf::WhereDataFrame) = delete!(parent(wdf), setdiff(1:nrow(parent(wdf)), rows(wdf)))
 Base.delete!(wdf::WhereDataFrame) = delete!(parent(wdf), rows(wdf))
 Base.view(wdf::WhereDataFrame) = view(parent(wdf), rows(wdf), :)
-#function nonunique(wdf::WhereDataFrame, args...)
-#  x = falses(size(parent(wdf), 1))
-#  idx = rows(wdf)[findall(nonunique(view(wdf), args...))]
-#  x[idx] .= true
-#  return x
-#end
-#Base.unique!(wdf::WhereDataFrame, args...) = delete!(parent(wdf), findall(nonunique(wdf, args...)))
-#Base.unique(wdf::WhereDataFrame, args...) = parent(wdf)[(!).(nonunique(wdf, args...)), :]
 combine(wdf::WhereDataFrame, args...; kwargs...) = combine(view(wdf), args...; kwargs...)
-
 DataFrame(wdf::WhereDataFrame; copycols::Bool=true) = DataFrame(view(wdf); copycols = copycols)
-Base.first(df::WhereDataFrame, args...) = first(view(df), args...)
-Base.last(df::WhereDataFrame, args...) = last(view(df), args...)
 DataAPI.describe(wdf::WhereDataFrame, args...; kwargs...) = describe(view(wdf), args...; kwargs...)
 
 ##############################################################################
 ##
-## Select/transform
+## transform/transform!
 ##
 ##############################################################################
-#select!(wdf::WhereDataFrame, args...; renamecols::Bool=true) =
-    #_replace_columns!(parent(wdf), select(wdf, args..., copycols=false, renamecols=renamecols))
-#select(wdf::WhereDataFrame, args...; copycols::Bool=true, renamecols::Bool=true) =
-    #manipulate(wdf, args..., copycols=copycols, renamecols=renamecols)
+
 transform!(df::WhereDataFrame, args...; renamecols::Bool=true) =
     _replace_columns!(parent(wdf), transform(wdf, args..., copycols=false, renamecols=renamecols))
 transform(wdf::WhereDataFrame, args...; copycols::Bool=true, renamecols::Bool=true) =
--- a/src/wheredataframe/wheredataframe.jl
+++ b/src/wheredataframe/wheredataframe.jl
@@ -47,6 +47,8 @@ julia> filter(where(df, :a => ByRow(>(1)), :b => x -> x .< 2))
 # transform only certain rows
 julia> transform(where(df, :a => ByRow(ismissing)), :a => (x -> 0) => :a)
 julia> transform(where(df, :a => ByRow(!ismissing)), :a => cumsum)
+# this returns an error since type different
+julia> transform(where(df, :a => ByRow(ismissing)), :a => (x -> "1") => :a)
 
 # combine using certain rows
 julia> combine(where(df, :a => ByRow(!ismissing)), :a => sum)
@@ -256,14 +258,14 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}, AsTable},
         if copycols && !(fun isa ByRow) &&
             (res isa SubArray || any(i -> respar === parent(cdf[i]), parent_cols))
             if newname ∈ propertynames(df)
-                newdf[!, newname] = convert(Vector{Union{eltype(res), eltype(df[!, newname])}}, df[:, newname])
+                newdf[!, newname] = copycols ? df[:, newname] : df[!, newname]
             else
                 newdf[!, newname] = Vector{Union{eltype(res), Missing}}(missing, size(df, 1))
             end
             newdf[rows(wdf), newname] = copy(res)
         else
             if newname ∈ propertynames(df)
-                newdf[!, newname] = convert(Vector{Union{eltype(res), eltype(df[!, newname])}}, df[:, newname])
+                newdf[!, newname] = copycols ? df[:, newname] : df[!, newname]
             else
                 newdf[!, newname] = Vector{Union{eltype(res), Missing}}(missing, size(df, 1))
             end
@@ -272,7 +274,7 @@ function select_transform!(nc::Pair{<:Union{Int, AbstractVector{Int}, AsTable},
     else
         res_unwrap = res isa Union{AbstractArray{<:Any, 0}, Ref} ? res[] : res
           if newname ∈ propertynames(df)
-              newdf[!, newname] = convert(Vector{Union{eltype(res), eltype(df[!, newname])}}, df[:, newname])
+              newdf[!, newname] = copycols ? df[:, newname] : df[!, newname]
           else
              newdf[!, newname] = Vector{Union{eltype(res), Missing}}(missing, size(df, 1))
         end
--- a/src/wheredataframe/wheredataframe.jl
+++ b/src/wheredataframe/wheredataframe.jl
@@ -1,0 +1,300 @@
+"""
+    WhereDataFrame{<:AbstractDataFrame,<:AbstractIndex}
+
+    The result of a [`where`](@ref) operation on an `AbstractDataFrame`; a
+    subset of a `AbstractDataFrame`
+    Not meant to be constructed directly, see `where`.
+"""
+struct WhereDataFrame{D<:AbstractDataFrame, T<:AbstractVector{Int}}
+    parent::D
+    rows::T
+end
+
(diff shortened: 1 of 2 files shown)


src/wheredataframe/wheredataframe.jl
```diff
@@ -0,0 +1,287 @@
+"""
+    WhereDataFrame{<:AbstractDataFrame,<:AbstractIndex}
+
+    The result of a [`where`](@ref) operation on an `AbstractDataFrame`; a
+    subset of a `AbstractDataFrame`
```
bkamins, line 5: can you please highlight that this is a wrapper (as opposed to `SubDataFrame` which is `AbstractDataFrame`?)

src/wheredataframe/wheredataframe.jl
```diff
@@ -0,0 +1,287 @@
...
+
+"""
+    where(d::AbstractDataFrame, args...)
+
+`where` introduces a `where` clause, which will be applied in the next function. Returns a `WhereDataFrame`.
+
+`args...` obey the same syntax as `select(d, args...)`
+Rows that return missing are understood as false
+
+- `filter`/`filter!` returns an AbstractDataFrame after filtering (resp. deleting) specified rows
```
bkamins, line 33: please add that this is a list of functions that support `where` wrapper. In general - in the long term almost all functions we define could support it I think (but we can make this change gradually)

bkamins: in particular you omit `select`/`select!` now and e.g. `insertcols!`.

src/wheredataframe/wheredataframe.jl
matthieugomez: Yes. Just to be clear, this draft is just a way to start a discussion about this potential syntax — I figured it may be interesting for people to try it out a bit.

src/wheredataframe/wheredataframe.jl
```diff
@@ -0,0 +1,287 @@
...
+
+# transform only certain rows
+julia> transform(where(df, :a => ByRow(ismissing)), :a => (x -> 0) => :a)
+julia> transform(where(df, :a => ByRow(!ismissing)), :a => cumsum)
+# combine using certain rows
+julia> combine(where(df, :a => ByRow(!ismissing)), :a => sum)
+```
+"""
+function where(df::AbstractDataFrame, args...)
+    dfr = select(df, args...)
```
bkamins, line 55: probably `copycols=false` should be passed here

bkamins: also check the case when `dfr` has `0` columns

src/wheredataframe/wheredataframe.jl
```diff
@@ -0,0 +1,287 @@
...
+##
+##############################################################################
+function Base.summary(wdf::WhereDataFrame)
+    @sprintf("%d×%d %s", size(parent(wdf))..., "WhereDataFrame")
+end
+
+Base.summary(io::IO, wdf::WhereDataFrame) = summary(wdf)
+
+
+function Base.show(io::IO,
```
bkamins, line 79: All mime types should be supported (in particular HTML and LaTeX). Also I would make sure that we very clearly visually indicate that this is NOT an `AbstractDataFrame`.

Finally - standard `show` has to be updated to avoid `StackOverflow` when trying to display a data frame with `WhereDataFrame` in a circular reference.

bkamins: Also in general - I would wait with merging this PR after PrettyTables.jl integration is done and then update it accordingly. c.f. #2429

src/wheredataframe/wheredataframe.jl
```diff
@@ -0,0 +1,287 @@
...
+end
+
+
+##############################################################################
+##
+## Operations
+##
+##############################################################################
+
+Base.filter(wdf::WhereDataFrame) = parent(wdf)[rows(wdf), :]
```
bkamins, line 135: we should support `view` kwarg in `filter`.

src/wheredataframe/wheredataframe.jl
```diff
@@ -0,0 +1,287 @@
...
+## transform/transform!
+##
+##############################################################################
+
+transform!(df::WhereDataFrame, args...; renamecols::Bool=true) =
+    _replace_columns!(parent(wdf), transform(wdf, args..., copycols=false, renamecols=renamecols))
+transform(wdf::WhereDataFrame, args...; copycols::Bool=true, renamecols::Bool=true) =
+    manipulate(wdf, :, args..., copycols=copycols, renamecols=renamecols)
+
+function manipulate(wdf::WhereDataFrame, cs...; copycols::Bool, renamecols::Bool)
```
bkamins, line 154: this should wait till #2461 is merged - in case there are some changes to the logic on this level.

bkamins:
This would also solve  #2465.

bkamins:
This would not solve #2211, but I think #2211 could be closed - as the case would be handled via `transform!` or `insertcols!` - right?

bkamins:
Thank you for submitting this PR (it is really great to have you contribute).

The approach you propose is a valid alternative to making `SubDataFrame` on `Index` I think. It is less flexible, but I can see its benefits - which are consequences of lower flexibility. The crucial thing is to decide on the list of functions that should support `WhereDataFrame` (and make sure this design is easily understood by the users). Let us see what @nalimilan and @pdeffebach think (especially with relation to DataFramesMeta.jl).

For me - the big point that I am not sure will work nicely with the proposed design - is how do you want to support `groupby` here? Do you want to define `where` for `GroupedDataFrame` similarly, or you want to pass `where` wrapper of a data frame to `groupby`, or you do not want to support it at all (I would assume you want to support it)?

(also at some point please remember to update the manual, the docstrings and NEWS.md - but this should be done at the end when we have design settled)

On a development side let me make a small comment (as I am not sure you are aware of this - and with large PRs this can easily become a problem). Each push to GitHub triggers CI. For DataFrames.jl CI takes over 1h for each push. CI is shared across JuliaData. This means that doing a lot of pushes to DataFrames.jl easily stalls CI for all JuliaData projects for e.g. one day. Therefore, for DataFrames.jl, it is strongly encouraged to limit the number of pushes as much as possible (i.e. preferably there should be one push from local repo to GitHub when you are happy with what you have and when you get reviewing suggestions - they should be put in a batch and committed together).

src/wheredataframe/wheredataframe.jl
```diff
@@ -0,0 +1,287 @@
...
+##
+##############################################################################
+
+Base.filter(wdf::WhereDataFrame) = parent(wdf)[rows(wdf), :]
+Base.filter!(wdf::WhereDataFrame) = delete!(parent(wdf), setdiff(1:nrow(parent(wdf)), rows(wdf)))
+Base.delete!(wdf::WhereDataFrame) = delete!(parent(wdf), rows(wdf))
+Base.view(wdf::WhereDataFrame) = view(parent(wdf), rows(wdf), :)
+combine(wdf::WhereDataFrame, args...; kwargs...) = combine(view(wdf), args...; kwargs...)
+DataFrame(wdf::WhereDataFrame; copycols::Bool=true) = DataFrame(view(wdf); copycols = copycols)
+DataAPI.describe(wdf::WhereDataFrame, args...; kwargs...) = describe(view(wdf), args...; kwargs...)
```
bkamins, line 141: since `where` is a wrapper (so `WhereDataFrame` is not an `AbstractDataFrame`) I think I would be better to avoid these methods and require `view` to be called by the user explicitly. Not sure - but this would better highlight that `where` is a lightweight wrapper only.

The point is that e.g. `transform(wdf)` and `combine(wdf, :)` will return a completely different result with your proposed design and I think it would be better to avoid this. Intuitively I would support only these methods for `WhereDataFrame` where its behavior differs from `SubDataFrame` and for other methods require to call `view` on it. Then the message to the user would be consistent.

src/wheredataframe/wheredataframe.jl
bkamins: Understood - I highlighted in the comments the major design considerations I see with this proposal.

bkamins:
Just to add what I just commented to @nalimilan. If we accept this design (which is one of the two options - the other is to make `SubDataFrame` more flexible) then I think `WhereDataFrame` should be designed to highlight that it should be a short-lived object not used stand-alone (so maybe it even should not have `show` implemented). So `where(df, ...)` would be a kind-of modifier like `ByRow` or `AsTable` in a selected list of functions (maybe it should be `Where` then not `where` and the object should be called `Where`, as it is not an `AbstractDataFrame`, so `WhereDataFrame` is actually a confusing name).

The point is that `Where(df, ...)` is a syntax should be meant to modify how a limited number of functions should work (as opposed to `SubDataFrame` which is a fully operational `AbstractDataFrame`). This is a key decision to make - do we prefer a modifier syntax (`Where`) or add functionality to `SubDataFrame`.

Seeing your PR I actually think that maybe `Where` would actually be a better approach, but it requires a very careful thought:
* which functions and how should support it
* are we happy with what we get as a result (i.e. if all operations that we think are common and important can be conveniently handled using `Where`)

nalimilan:
Thanks for making a PR. Though I'd prefer to settle the discussion on the best design before diving into the implementation. Otherwise out time may be wasted and/or we may not consider all available options because we are too focused on a particular approach.

More specifically, my concern with a lazy `where` is that in general in Julia and in DataFrames functions are eager by default, and we wouldn't have an exact equivalent to `where` which wouldn't return a view. This could be confusing as filtering is a very common operation, and currently `filter` isn't perfectly convenient as you noted at https://github.com/JuliaData/DataFrames.jl/issues/2323 and https://github.com/JuliaData/DataFrames.jl/issues/2211. If we added a lazy `where`, then we still wouldn't have an eager variant with convenient features (like skipping missings and selecting rows inside `GroupedDataFrame`) -- and I don't like `filter(where(...))` as a replacement as it seems redundant. So I think we should decide first whether we want an eager `where`, a lazy `where` as in this PR, or both (with different names), and how they should work.

bkamins:
Some more comments to keep in mind:
* `SubDataFrame` does not remember its source, only its true parent (this is something that makes `Where` preferable over `SubDataFrame`)
* still we would have to decide how `where` works with `GroupedDataFrame` (should it be `where(gdf, ...)` or `groupby(where(df, ...), ...)` in particular)
* per @nalimilan comment I think mixing lazy and non lazy variants in one function (differentiated by kwarg) is not desirable as it would lead to `transform(where(df, ..., lazy=true), ...)` and `transform(where(df, ..., lazy=false), ...)` produce significantly different results

pdeffebach:
@matthieugomez thanks for this. I'm glad someone finally went and implemented it. 

A few "big picture" questions

1. This can cause anti-patterns where people port their stata workflow to Julia. In Stata, since you can only have one data set in RAM at a time, people often store multiple versions of their data in the same data-set and work with them separately with `if` statements. 

I don't think any of us want people using `where` when they should use a new data frame all together. Can you give some examples of use-cases that *aren't* this? 

2. In general, I would say this is no substitute for making skipping values easier. If you have missing values in `:long_column_name` and want to, for example, normalize by mean and SD, your options are (in DataFramesMeta)

```
@pipe df |>
    transform(_, :long_column_name = (:long_column_name .- mean(skipmissing(:long_column_name)) .- std(skipmissing(:long_column_name) )
```

or 

```
@pipe df 
    where(_, .!missing.(:long_column_name)) |>
    transform(_, :long_column_name = :long_column_name .- mean(:long_column_name) ./ std(:long_column_name)) 
```

Either way that's a lot of typing. Given that we still want, in the future, an easier way to skip missing values, is this feature still worth it? 

Basically, if we take away an anti-pattern and the benefits of skipping missing values, what are the remaining benefits of this PR?

matthieugomez:
It is probably too complex. I think a simpler `where` that simply creates a dataframe (or a view with view = true) would be easier to reason about. While it does not solve the problem of updating a variable if a condition is satisfied, there could be an `update` verb down the line to do so (I don’t think using views for that is a good idea).

pdeffebach:
I will start experimenting with an `@if` macro in DataFramesMeta and see what I can come up with.

pdeffebach:
I no longer think the Stata-esque dataset-in-a-dataset is such a bad anti-pattern. 

I wonder if we can implement this but having a `mutable = true` flag for a `SubDataFrame`? Should behave the same in all instances, except `setindex!`.

bkamins:
Can you open an issue for this, since this is closed so it is easier to track?

I think a flag could be OK (we could even allow for mutation of this flag post creation). But it requires a careful consideration of the design (maybe repeating what we already discussed to refresh memory and include recent experiences). Still I would not allow it by default I think (as views should be immutable in general)

In general - the more I think about this issue the more I feel we need to resolve the `where` issue in a nice way (i.e. something that will be user friendly in most of the cases even at the expense of some inconsistency in corner cases - which would be e.g. disallowed)

bkamins:
Before you opened a new issue let me write down the thoughts here.

I have reviewed the implementation and it should be possible to implement it without much redesign. We might even consider adding this feature by default (i.e. without requiring `mutable=true`). The only caveat is that mutating via a view might invalidate other views, but we already allow this in `select!` and `transform!` for `GroupedDataFrame` so maybe it is OK (we just need to make sure to clearly signal the risks in the documentation).

The key thing to think of is how to design:
* `setindex!` (easy)
* broadcasting assignment
* `select!` (the question is if we need to support `select!` as it potentially drops columns)
* `transform!`

I assume that what we should allow is ADDING new columns only, and not replacing old columns (`setindex!` for old columns already works and has a defined API - in particular it does not allow changing of column `eltype`)

bkamins:
Actually maybe we could allow replacing columns - we need to list all possible operations in the new issue and decide which should work and how.

PR #2469, opened 2020-10-05
remove internals.md

bkamins:
This file is not maintained (and is outdated) and is not referenced anyway.


bkamins:
Sorry for mixing things up. I have forgotten that we have this page to avoid throwing error by documenter.

PR #2475, opened 2020-10-09
mapcols!(): exit early if no columns

### LINKED ISSUE #2474: mapcols!() fails when DataFrame is empty
```
julia> mapcols!(identity, DataFrame())
ERROR: ArgumentError: collection must be non-empty
Stacktrace:
 [1] _extrema_itr(::typeof(identity), ::Base.Generator{Array{AbstractArray{T,1} where T,1},typeof(length)}) at ./operators.jl:481
 [2] extrema at ./operators.jl:460 [inlined]
 [3] mapcols!(::typeof(identity), ::DataFrame) at /home/ge68wan/.julia/packages/DataFrames/cdZCk/src/abstractdataframe/iteration.jl:385
```

The fix would be to skip the column lengths check if there are no columns.

alyst:
fixes #2474

### DIFF
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -411,6 +411,8 @@ julia> df
 """
 function mapcols!(f::Union{Function,Type}, df::DataFrame)
     # note: `f` must return a consistent length
+    (ncol(df) == 0) && return df # skip if no columns
+
     vs = AbstractVector[]
     seenscalar = false
     seenvector = false
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -411,7 +411,7 @@ julia> df
 """
 function mapcols!(f::Union{Function,Type}, df::DataFrame)
     # note: `f` must return a consistent length
-    (ncol(df) == 0) && return df # skip if no columns
+    ncol(df) == 0 && return df # skip if no columns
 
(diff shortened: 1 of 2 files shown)


bkamins (approved):

src/abstractdataframe/iteration.jl
```diff
@@ -411,6 +411,8 @@ julia> df
 """
 function mapcols!(f::Union{Function,Type}, df::DataFrame)
     # note: `f` must return a consistent length
+    (ncol(df) == 0) && return df # skip if no columns
```
nalimilan, line 414:
```suggestion
    ncol(df) == 0 && return df # skip if no columns
```

nalimilan (approved):

bkamins:
Thank you!

src/abstractdataframe/iteration.jl
alyst: Just for me to know for the future: is it a recommended codestyle for dataframes?
I'm fine with it,  just with all these &, &&, == precedence subtleties I prefer to be on the safe side in my scripts.

bkamins: You probably saw that I left the same comment earlier, but I am usually more flexible than @nalimilan.
The general coding style we try to follow is to use what Julia Base recommends (and in this case this is the recommended style, though with `&` indeed parentheses would be needed)

PR #2481, opened 2020-10-12
[BREAKING] Multicolumn transformations for GoupedDataFrame

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2410

I am opening now to make sure that 0.22 branch of https://github.com/JuliaData/DataFrames.jl/commit/a13db50a0cd8115691fe92e67b1301422c664401 does not get lost.

This PR is not finished. I have implemented all (in particular allowing returning multiple columns from functions) except handling `AsTable` and multipe `Symbol`s as destination columns (this still needs to be implemented as it requires new logic)


bkamins:
OK - this PR is not cleaned, tested, nor documented yet, but the promised functionality seems (remember - not tested :smile:)) to work, so please feel free to experiment with it and comment if you catch something surprising. Thank you!

bkamins:
@pdeffebach - with this PR we ensure the following invariant (which I think is relevant for DataFramesMeta.jl design). The result of:
```
transform(df, args...)
```
is always the same as the result of
```
transform(groupby(df, cols), args...)
```
(also up to errors - if one errors the other also errors)

pdeffebach:
Wait, why would we want that? 

What if I want to perform the operation `x .- mean(x)`? We put a lot of effort into ensuring `transform` works by groups.

pdeffebach:
Oh good. You scared me! Yes having the same with `ByRow` is the correct behavior. Thanks!

bkamins:
I still have a lot of tests to write (but now the PR should pass the tests) and documentation to update.
However, @nalimilan - can you please review the changes in docs/src/man/split_apply_combine.md as correcting the documentation is usually a lengthy process, and probably we can reuse things we work-out there in the docstrings later. Thank you!

(all others: feel free to have a look at docs/src/man/split_apply_combine.md if you are interested as it specifies how new rules work)

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
-`function` can return multiple columns in the form of an `AbstractDataFrame`,
-`AbstractMatrix`, `NamedTuple` or `DataFrameRow`.
+All functions have two types of signatures. One of them takes a `GroupedDataFrame`
+as a first argument and an arbitrary number of transfomations described above
+as following arguments. The second type of signature is when `Function` or `Type`
+is passed as a first argument and `GroupedDataFrame` is a second argument (in a
+similar fashion like it is passed in e.g. `map` function).
+
+As a special rule that applies to `cols => function` and `cols => function =>
+target_cols` syntaxes is the following. If `cols` is wrapped in an `AsTable`
```
nalimilan, line 62:
```suggestion
As a special rule, with the `cols => function` and `cols => function =>
target_cols` syntaxes, if `cols` is wrapped in an `AsTable`
```

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
+object then a `NamedTuple` containing columns selected by `cols` is passed to
+`function`.
+
+What is allowed for `function` to return is determined by the `target_cols` value
+in the following way:
```
nalimilan, line 67:
```suggestion
What is allowed for `function` to return is determined by the `target_cols` value:
```

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
+1. If just a `function` is passed as an argument then returning a data frame,
```
nalimilan, line 68:
```suggestion
1. If `target_cols` is omitted, then returning a data frame,
```

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
-- a `DataFrame` or a matrix produces as many rows and columns as it contains;
-  note that this option should be avoided due to its poor performance when the number
-  of groups is large
+with the same number and order of rows as the source (even if `GroupedDataFrame`
+had its groups reordered).
 
-The kind of return value and the number and names of columns must be the same for all groups.
+For `combine` return value is ordered by the order of groups in `GroupedDataFrame`
+and for each group the functions can return an arbibrary number of rows (provided
+that these numbers are consistent).
```
nalimilan, line 105:
```suggestion
For `combine`, rows in the returned object appear in the order of
groups in the `GroupedDataFrame`. The functions can return an arbitrary number
of rows for each group, but the kind of returned object and the number
and names of columns must be the same for all groups.
```

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
 
-If a single value or a vector is returned by the `function` and `target_col` is not
-provided, it is generated automatically, by concatenating source column name and
-`function` name where possible (see examples below).
+To apply `function` to each row instead of whole columns, it can be wrapped in a
+`ByRow` struct. In this case if `cols` is a `Symbol`, a string, or an
+integer then `function` is applied to each element (row) of `cols` using
+broadcasting. Otherwise `cols` can be any column indexing syntax, in
+which case `function` will be passed one argument for each of the columns
+specified by `cols`. If `ByRow` is used it is allowed for
```
nalimilan, line 116: Do we really need to distinguish the 1-column case from the general case? This was needed back when a named tuple was passed, but now both cases follow the same rule.

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
+`cols` to select an empty set of columns, in which case `function`
+ is called for each row without any arguments.
 
-We show several examples of the `by` function applied to the `iris` dataset below:
+The kind of return value and the number and names of columns must be the same for all groups.
+
```
nalimilan, line 121: Remove if you accept my suggestion to mention it above.
```suggestion
```

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
+(not all keyword arguments are supported in all cases; in general they are allowed
+in situations when they are meaningful, see the documentation of the specific functions
+for details):
+- `keepkeys` : if grouping columns should be kept in the returned data frame.
+- `ungroup` : if the retun value of the operation should be a data frame or a
+  `GroupedDataFrame`.
+- `copycols` : if columns of the source data frame should be copied if no transformation
+  is applied to them.
+- `renamecols` : if in `cols => funcion` form the automatically generated column name
+  should include the name of transformation function or not.
```
nalimilan, line 132:
```suggestion
- `keepkeys` : whether grouping columns should be kept in the returned data frame.
- `ungroup` : whether the return value of the operation should be a data frame or a
  `GroupedDataFrame`.
- `copycols` : whether columns of the source data frame should be copied if no transformation
  is applied to them.
- `renamecols` : whether in the `cols => function` form automatically generated column names
  should include the name of transformation functions or not.
```

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
+2. a `cols => function => target_cols` form additionally specifying the target column or columns
+3. a `cols => function` pair indicating that `function` should be called with
+   positional arguments holding columns `cols`, which can be a any valid column selector;
+   in this case target column name is automatically generated and it is assumed that
+   `function` returns a single value or a vector; the generated name is created by concatenating
+   source column name and `function` name where possible (see examples below).
+4. a `col => target_cols` pair, which renames the column `col` to `target_cols`
+5. a `nrow` or `nrow => target_cols` form which efficiently computes the number of rows
+   in a group (without `target_cols` the new column is called `:nrow`)
+6. vectors or matrices transformations specified by `Pair` syntax described in points 2 to 5
```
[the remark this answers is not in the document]
bkamins: We allow vectors of symbols etc. (this is covered in point 1.)

The description of vararg is rewritten and now is below (https://github.com/JuliaData/DataFrames.jl/pull/2481/files#diff-45f87ad2a77d5bb7ac1969a14bb11079c730996f6904a82f9ae00222d1ae5f72R56)

docs/src/man/split_apply_combine.md
bkamins: This is not correct. I mean specifically "a `function`" (so neither `cols` nor `target_cols`). If `cols` is passed but `target_cols` is omitted then we assume that **a single column is returned** (and we automatically generate a name for it). If you specify `cols` and want multiple columns you must be explicit (e.g. by adding `AsTable`) and in this case different rules apply.

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
-5. a `nrow` or `nrow => target_col` form which efficiently computes the number of rows
-   in a group (without `target_col` the new column is called `:nrow`)
-6. several arguments of the forms given above, or vectors thereof
-7. a function which will be called with a `SubDataFrame` corresponding to each group;
+2. a `cols => function => target_cols` form additionally specifying the target column or columns
+3. a `cols => function` pair indicating that `function` should be called with
+   positional arguments holding columns `cols`, which can be a any valid column selector;
+   in this case target column name is automatically generated and it is assumed that
+   `function` returns a single value or a vector; the generated name is created by concatenating
+   source column name and `function` name where possible (see examples below).
```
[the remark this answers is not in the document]
bkamins: fixed

docs/src/man/split_apply_combine.md
```diff
@@ -26,59 +35,103 @@ Operations can then be applied on each group using one of the following function
...
+4. a `col => target_cols` pair, which renames the column `col` to `target_cols`
```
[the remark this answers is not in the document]
bkamins: fixed (but I keep the name, as we reference it below)

docs/src/man/split_apply_combine.md
bkamins: right. fixed

docs/src/man/split_apply_combine.md
```diff
@@ -176,8 +229,8 @@ julia> combine(gdf, nrow, :PetalLength => mean => :mean)
 │ 2   │ Iris-versicolor │ 50    │ 4.26    │
 │ 3   │ Iris-virginica  │ 50    │ 5.552   │
 
-julia> combine([:PetalLength, :SepalLength] => (p, s) -> (a=mean(p)/mean(s), b=sum(p)),
-               gdf) # multiple columns are passed as arguments
+julia> combine(gdf, [:PetalLength, :SepalLength] => ((p, s) -> (a=mean(p)/mean(s), b=sum(p))) =>
+       AsTable) # multiple columns are passed as arguments
```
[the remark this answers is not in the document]
bkamins: fixed

bkamins:
@nalimilan - documentation updates part 2 (and hopefully final) is pushed here. I made adjustments to the manual per the comments given + I have refactored the docstrings. Now they are consistent with the manual and reuse the same template. This has three benefits I believe:
1. it will be easier to maintain
2. the docstrings take less space (which is relevant as the files are very large anyway)
3. the docstrings show a comparison of similarities and differences between available functions

Another review of documentation would be appreciated (after it is done I will review the tests of the functionality to make sure we cover everything and the PR then will be good for a final review).

bkamins:
I have added additional tests of correctness of the functionality we expose. This should be good for a whole code review.

bkamins:
@nalimilan - the code is now after refactoring. The functionality is split into functions. I also rearranged the source code into more files. The relevant changes in this are in splitapplycombine.jl (the fastaggregates.jl, complextransforms.jl and callprocessing.jl are just moved out from splitapplycombine.jl as it became unmaneageably large).

Also because of the complexity of changes it is best to review this PR for splitapplycombine.jl file (even without the moving-out the sub-files it was almost unreadable) not in the browser but in the editor. The testing code and the docstrings (that are now in abstractdataframe/selection.jl) are probably clean enough to check in the browser.

Sorry for this complexity, but adding the requested functionality required a massive rewrite of splitapplycombine.jl.

bkamins:
This PR is holding implementation of `where` PR.

test/string.jl
```diff
@@ -169,18 +169,18 @@ end
...
+          combine(gdf, nrow => "n", ungroup=false)
+
+    @test combine(gdf, "a" => identity) == combine(gdf, :a => identity) ==
           combine(gdf, "a" => identity) == combine(gdf, :a => identity)
-    @test combine(["a"] => identity, gdf) == combine([:a] => identity, gdf) ==
+    @test combine(gdf, ["a"] => identity) == combine(gdf, [:a] => identity) ==
           combine(gdf, ["a"] => identity) == combine(gdf, [:a] => identity)
-    @test combine(nrow => :n, gdf) == combine(nrow => "n", gdf) ==
+    @test combine(gdf, nrow => :n) == combine(gdf, nrow => "n") ==
           combine(gdf, nrow => :n) == combine(gdf, nrow => "n")
```
nalimilan, line 184: These include duplicates calls.

BTW, we need deprecations for this, right?

docs/src/man/split_apply_combine.md
```diff
@@ -13,6 +13,14 @@ In order to perform operations by groups you first need to create a `GroupedData
 object from your data frame using the `groupby` function that takes two arguments:
 (1) a data frame to be grouped, and (2) a set of columns to group by.
 
+!!! note
+
+    All operations described for `GroupedDataFrame` in this section of the manual
+    are also supported for `AbstractDataFrame` in which case it is considered as
+    being grouped on no columns (meaning it has a single group, or zero groups if it is empty).
+    The only difference is that in this case the `keepkeys` and `ungroup` keyword
+    arguments are not supported and a data frame is always returned.
```
nalimilan, line 22: Not to be addressed by this PR, but it's weird that `select` and `transform` are documented in detail only in this section, which is about grouped operations.

test/grouping.jl
```diff
@@ -770,128 +770,123 @@ end
...
+        @test combine(gd, cols => ((b,c) -> sum(b) + sum(c)) => :xyz, ungroup=ungroup) ==
             combine(d -> (xyz=sum(d.b) + sum(d.c),), gd, ungroup=ungroup)
-        if eltype(cols) === Bool
-            cols2 = [[false, true, false], [false, false, true]]
-            @test_throws MethodError combine((xyz = cols[1] => sum, xzz = cols2[2] => sum),
-                                             gd, ungroup=ungroup)
-            @test_throws MethodError combine((xyz = cols[1] => sum, xzz = cols2[1] => sum),
-                                             gd, ungroup=ungroup)
-            @test_throws MethodError combine((xyz = cols[1] => sum, xzz = cols2[2] => x -> first(x)),
-                                             gd, ungroup=ungroup)
```
nalimilan, line 873: Why drop these?

test/grouping.jl
```diff
@@ -3020,4 +3014,156 @@ end
     @test_throws MethodError select(gdf, AsTable([]) => ByRow(inc0) => :bin)
 end
 
+@testset "aggregation of reordered groups" begin
+    df = DataFrame(id = [1, 2, 3, 1, 3, 2], x=1:6)
```
nalimilan, line 3018:
```suggestion
    df = DataFrame(id=[1, 2, 3, 1, 3, 2], x=1:6)
```
(For consistency with below.)

test/grouping.jl
```diff
@@ -3020,4 +3014,156 @@ end
...
+    @test combine(gdf, :x => x -> 2x) ==
+          DataFrame(id=[1, 1, 3, 3, 2, 2], x_function=[6, 10, 2, 8, 4, 12])
+    @test combine(gdf, identity) == DataFrame(id=[1, 1, 3, 3, 2, 2], x=[3, 5, 1, 4, 2, 6])
+    @test combine(gdf, x -> (a=x.x, b=x.x)) ==
+          DataFrame(id=[1, 1, 3, 3, 2, 2], a=[3, 5, 1, 4, 2, 6], b=[3, 5, 1, 4, 2, 6])
+end
+
+@testset "new rules tests" begin
+    @testset "basic tests" begin
+        df = DataFrame(id = [1, 2, 3, 1, 3, 2], x=1:6)
```
nalimilan, line 3074:
```suggestion
        df = DataFrame(id=[1, 2, 3, 1, 3, 2], x=1:6)
```

test/grouping.jl
```diff
@@ -3020,4 +3014,156 @@ end
...
+    @test transform(df, x -> (a=x.x, b=x.x)) == transform(gdf, x -> (a=x.x, b=x.x))
+    @test combine(gdf, :x => x -> 2x) ==
+          DataFrame(id=[1, 1, 3, 3, 2, 2], x_function=[6, 10, 2, 8, 4, 12])
+    @test combine(gdf, identity) == DataFrame(id=[1, 1, 3, 3, 2, 2], x=[3, 5, 1, 4, 2, 6])
+    @test combine(gdf, x -> (a=x.x, b=x.x)) ==
+          DataFrame(id=[1, 1, 3, 3, 2, 2], a=[3, 5, 1, 4, 2, 6], b=[3, 5, 1, 4, 2, 6])
+end
+
+@testset "new rules tests" begin
+    @testset "basic tests" begin
```
nalimilan, line 3073: I'd use a name that still makes sense in the future. Also, probably better avoid nesting testsets

docs/src/man/split_apply_combine.md
bkamins: Agreed, but the point is that `select`, `transform`, and `combine` are now transparently handling data frames and grouped data frames in the same way - so maybe we just need to change the name of this section? As this note is about the case when "split" step is skipped.

src/abstractdataframe/selection.jl
```diff
@@ -10,6 +10,148 @@
...
+       in a group; without `target_cols` the new column is called `:nrow`, otherwise
+       it must be single column (a `Symbol` or a string).
+    6. vectors or matrices containing transformations specified by the `Pair` syntax
+       described in points 2 to 5
+    8. a function which will be called with a `SubDataFrame` corresponding to each group;
+       this form should be avoided due to its poor performance unless a very large
+       number of columns are processed (in which case `SubDataFrame` avoids excessive
+       compilation)
+
+    All functions have two types of signatures. One of them takes a `GroupedDataFrame`
```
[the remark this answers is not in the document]
bkamins: I will check what would be the best approach (if I recall correctly there was some problem with inconsistent indentation between the two - but I might be wrong)

src/abstractdataframe/selection.jl
```diff
@@ -434,233 +576,24 @@ function select_transform!(@nospecialize(nc::Union{Base.Callable, Pair{<:Union{I
...
-│ 3   │ 4.5     │
+    select!(args::Base.Callable, df::DataFrame; renamecols::Bool=true)
+    select!(gd::GroupedDataFrame{DataFrame}, args...; ungroup::Bool=true, renamecols::Bool=true)
+    select!(f::Base.Callable, gd::GroupedDataFrame; ungroup::Bool=true, renamecols::Bool=true)
 
-julia> df = DataFrame(a=1:3, b=4:6);
+Mutate `df` or `gd` in place to retain only columns specified by `args...` and
+return it. The result is guaranteed to have the same number of rows as `df` or
+parent of `gd`, except when no columns are selected (in which case the result
+has zero rows).
```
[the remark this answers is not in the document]
bkamins: You mean to disallow `select(df)`?, `select(groupby(df, cols), keepkeys=false)` and `select(groupby(df, []))`?

Then `select(groupby(df, cols))` would pass or fail depending on the contents of `cols`.

I am not sure what is best - internally we treat data frame with 0 columns as having an undefined number of rows. We return `0` only because we have to return something when calling `nrow`.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1178,87 +503,76 @@ function _combine(f::AbstractVector{<:Pair},
...
-            firstres = length(gd) > 0 ?
-                       do_call(fun, gd.idx, gd.starts, gd.ends, gd, incols, 1) :
-                       do_call(fun, Int[], 1:1, 0:0, gd, incols, 1)
-            firstmulticol = firstres isa MULTI_COLS_TYPE
-            if firstmulticol
-                throw(ArgumentError("a single value or vector result is required when " *
-                                    "passing multiple functions (got $(typeof(res)))"))
+    for i in eachindex(cs_norm, optional_transform)
+        cs_i = cs_norm[i]
+        ot_i = optional_transform[i]
```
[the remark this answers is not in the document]
bkamins: it is right, but it is unsafe (i.e. assuming that the code is correct this should be equivalent, but the code is so complex that I prefer to use safe style). I avoid using `zip` as it silently accepts vectors of uneven lengths.

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
nalimilan: Yeah. Calling `select` without any transformation doesn't sound terribly useful. Though I guess it's good to have for full generality. Indeed, otherwise you wouldn't be able to call `select` on a zero-column data frame...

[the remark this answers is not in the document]
bkamins: I will leave it as is.

test/grouping.jl
bkamins: done

test/string.jl
bkamins: These deprecations will be extremely hard to do because of how much the rules have changed. I will add a method throwing an informative error.

test/grouping.jl
bkamins: this is just an obscure case that is completely irrelevant now (it was relevant when we accepted something else than `Base.Callable` as a first argument to `combine` - but now it is just a random argument that could be anything against which we would check if `MethodError` is thrown by dispatch). I added tests for passing `Pair` as a first argument (throwing an error).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -640,219 +107,8 @@ function combine_helper(f, gd::GroupedDataFrame,
...
-check_aggregate(f::typeof(last),
-                ::AbstractVector{<:Union{Missing, MULTI_COLS_TYPE, AbstractVector}}) = f
-check_aggregate(f::typeof(last∘skipmissing), v::AbstractVector) =
-    eltype(v) === Any ? f : Aggregate(last, !ismissing)
-check_aggregate(f::typeof(last∘skipmissing),
-                ::AbstractVector{<:Union{Missing, MULTI_COLS_TYPE, AbstractVector}}) = f
-check_aggregate(f::typeof(length), ::AbstractVector) = Aggregate(length)
-
-# SkipMissing does not support length
+# implementation of execution of transformation function calls and return value wrapping
```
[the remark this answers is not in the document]
bkamins: I am not sure how you propose to fix this. I understand that `include` follows the rule:

> a task-local include path is set to the directory containing the file

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -911,226 +167,299 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
...
+                                   parentdf::AbstractDataFrame,
+                                   gd::GroupedDataFrame,
+                                   seen_cols::Dict{Symbol, Tuple{Bool, Int}},
+                                   trans_res::Vector{TransRes},
+                                   idx_agg::Union{Nothing, AbstractVector{Int}})
+    firstres = length(gd) > 0 ? cs_i(gd[1]) : cs_i(similar(parentdf, 0))
+    idx, outcols, nms = _combine_multicol(firstres, cs_i, gd, nothing)
+
+    if !(firstres isa Union{AbstractVecOrMat, AbstractDataFrame,
+        NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}})
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1178,87 +503,76 @@ function _combine(f::AbstractVector{<:Pair},
...
+            _combine_process_agg(cs_i, ot_i, parentdf, gd, seen_cols, trans_res, idx_agg)
+        elseif keeprows && cs_i isa Pair && first(last(cs_i)) === identity &&
+               !(first(cs_i) isa AsTable) && (last(last(cs_i)) isa Symbol)
+            # this is a fast path used when we pass a column or rename a column in select or transform
+            _combine_process_noop(cs_i, ot_i, parentdf, seen_cols, trans_res, idx_keeprows, copycols)
+        elseif cs_i isa Base.Callable
+            idx_callable = _combine_process_callable(cs_i, ot_i, parentdf, gd, seen_cols, trans_res, idx_agg)
+            if idx_callable !== nothing
+                if idx_agg === nothing
+                        idx_agg = idx_callable
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -1178,87 +503,76 @@ function _combine(f::AbstractVector{<:Pair},
...
-            # in this case if we are not passed a vector compute it.
-            if !(firstres isa AbstractVector) && isnothing(idx_agg)
-                idx_agg = Vector{Int}(undef, length(gd))
-                fillfirst!(nothing, idx_agg, 1:length(gd.groups), gd)
+        else
+            @assert cs_i isa Pair
+            idx_pair = _combine_process_pair(cs_i, ot_i, parentdf, gd, seen_cols, trans_res, idx_agg)
+            if idx_pair !== nothing
+                if idx_agg === nothing
+                        idx_agg = idx_pair
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -911,226 +167,299 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
...
-            end
-            # It is safe to use a non-missing init value
-            # since missing will poison the result if present
-            # we assume here that groups are non-empty (current design assures this)
-            # + workaround for https://github.com/JuliaLang/julia/issues/36978
-            if isconcretetype(S) && hasmethod($initf, Tuple{S}) && !(S <: Irrational)
-                fill!(outcol, $initf(S))
-            else
-                fillfirst!(condf, outcol, incol, gd)
+struct TransRes
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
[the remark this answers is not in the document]
nalimilan: I'm not sure what that excerpt means TBH. :-D But currently all `include` calls are in src/DataFrames.jl, so what's special about these?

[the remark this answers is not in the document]
bkamins: ah - now I see what you mean. I can move them there. 

> I'm not sure what that excerpt means TBH

It just says that the path is relative to the file where `include` is present (so if I move `include` to src/DataFrames.jl I will have to change the paths, but this is not a problem).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -911,226 +167,299 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
...
+    col_idx::Vector{Int} # index for a column
+    col::AbstractVector # computed value of a column
+    name::Symbol # name of a column
+    optional::Bool # if a column is allowed to be replaced in the future
+end
+
+function _combine_process_agg(@nospecialize(cs_i::Any),
```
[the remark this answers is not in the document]
bkamins: fixed

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -911,226 +167,299 @@ function fillfirst!(condf, outcol::AbstractVector, incol::AbstractVector,
...
+                              ot_i::Bool,
```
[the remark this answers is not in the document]
bkamins: done

src/groupeddataframe/splitapplycombine.jl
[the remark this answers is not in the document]
bkamins: done

src/abstractdataframe/selection.jl
```diff
@@ -815,6 +731,129 @@ julia> select(df, AsTable(:) => ByRow(x -> (mean=mean(x), std=std(x))) => :stats
...
+⋮
+Last Group (3 rows): a = 2
+│ Row │ a     │ c_sum │ nrow  │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 2     │ 17    │ 3     │
+│ 2   │ 2     │ 17    │ 3     │
+│ 3   │ 2     │ 17    │ 3     │
+
+julia> select(gd, :c => (x -> sum(log, x)) => :sum_log_c) # specifying a name for target column
```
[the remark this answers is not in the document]
bkamins: done

src/abstractdataframe/selection.jl
```diff
@@ -10,6 +10,148 @@
...
+
+    If `df` is a `SubDataFrame` and `copycols=true` then a `DataFrame` is
+    returned and the same copying rules apply as for a `DataFrame` input: this
+    means in particular that selected columns will be copied. If
+    `copycols=false`, a `SubDataFrame` is returned without copying columns.
+
+    There the following keyword arguments are supported by the transformation functions
+    (not all keyword arguments are supported in all cases; in general they are allowed
+    in situations when they are meaningful, see the documentation of the specific functions
+    for details):
```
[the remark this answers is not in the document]
bkamins: changed to "signatures" as I show it there. I would leave it in one place. It will be a nightmare to maintain it in several places (you have already requested above to merge the documentation between docstrings and manual :smile:).

src/abstractdataframe/selection.jl
```diff
@@ -10,6 +10,148 @@
...
+    names in target data frame are accepted (e.g. `select!(df, [:a], :, r"a")`
+    is allowed) and only the first occurrence is used. In particular a syntax to
+    move column `:col` to the first position in the data frame is
+    `select!(df, :col, :)`. On the contrary, output column names of renaming,
+    transformation and single column selection operations must be unique, so e.g.
+    `select!(df, :a, :a => :a)` or `select!(df, :a, :a => ByRow(sin) => :a)` are not allowed.
+
+    Note that including the same column several times in the data frame via renaming
+    or transformations that return the same object without copying may create
+    column aliases even if `copycols=true`. An example of such a situation is
```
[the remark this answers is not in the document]
bkamins: ok. I have expanded the description.

docs/src/man/split_apply_combine.md
bkamins: I have just renamed this section of the manual.

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
bkamins: I have checked. Unfortunately markdown does not allow for including files I am afraid (if you know how to do it please let me know). So we have to copy this in two places. Let us agree that we finalize the description in the manual, and then I copy it to the docstrings. OK?

bkamins:
@nalimilan I have incorportated all the recommended changes apart from syncing manual and docs - I will do it when manual is finalized. It should be ready for another round of reviews. Thank you!

docs/src/man/split_apply_combine.md
```diff
@@ -1,26 +1,34 @@
-# The Split-Apply-Combine Strategy
+# Transforming data frames
```
nalimilan, line 1: This is probably too general. Better keep something more specific to grouping, at least for now, and maybe later reorganise docs so that we present the `select`/`transform`/`combine` API elsewhere in detail outside of the context of grouping.

src/abstractdataframe/selection.jl
```diff
@@ -1137,6 +1145,11 @@ function combine(arg::Base.Callable, df::AbstractDataFrame; renamecols::Bool=tru
     return combine(df, arg)
 end
 
+combine(f::Pair, gd::AbstractDataFrame; renamecols::Bool=true) =
+    throw(ArgumentError("First argument must be a transformation if the second argument is a data frame. " *
+                        "You can pass a `Pair` as a second argument of the transformation. If you want the return " *
```
nalimilan, line 1150:
```suggestion
                        "You can pass a `Pair` as the second argument of the transformation. If you want the return " *
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -45,7 +45,11 @@ into row groups.
   ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
 - `sort` : whether to sort groups according to the values of the grouping columns
   `cols`; if all `cols` are `CategoricalVector`s then groups are always sorted
-  irrespective of the value of `sort`
+  irrespective of the value of `sort`; if `sort=false` then the order of groups
+  is undefined (a typical case is that they follow the order of appreance of
+  respecive values in the grouping columns, but a notable exception is when the
+  columns are `PooledVector`s, in which case they are ordered accoring to the `pool`
+  field in these vectors)
```
nalimilan, line 52: I don't think that's true: in general the order is really undefined due to the dict-like grouping fallback. And the pool of `PooledDataArray` isn't exposed to users so it's not really useful to tell give them this information.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -618,11 +622,17 @@ end
...
-        throw(ArgumentError("First argument must be a transformation if the second argument is a grouped data frame"))
+        throw(ArgumentError("First argument must be a transformation if the second argument is a GroupedDataFrame"))
     end
     return combine(gd, f, keepkeys=keepkeys, ungroup=ungroup, renamecols=renamecols)
 end
 
+combine(f::Pair, gd::GroupedDataFrame;
+        keepkeys::Bool=true, ungroup::Bool=true, renamecols::Bool=true) =
+    throw(ArgumentError("First argument must be a transformation if the second argument is a GroupedDataFrame. " *
+                        "You can pass a `Pair` as a second argument of the transformation. If you want the return " *
```
nalimilan, line 633:
```suggestion
                        "You can pass a `Pair` as the second argument of the transformation. If you want the return " *
```

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
nalimilan: But how about making this part specific to each function's docstring so that we never mention an argument that a function doesn't support?

(BTW I don't see the change to "signature".)

test/string.jl
nalimilan: OK, but there are still duplicates. :-)

docs/src/man/split_apply_combine.md
bkamins: OK - I have reverted.

src/groupeddataframe/groupeddataframe.jl
bkamins: I have updated the description to make it more precise.

test/string.jl
bkamins: I removed the duplicates

src/abstractdataframe/selection.jl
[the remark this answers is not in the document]
bkamins: OK - I moved them to specific functions.

bkamins:
@nalimilan - I have incorporated the comments. Let me know if it is OK to move the descriptions from the manual to the docstring. Thank you!

bkamins:
TODO: add NEWS.md entry (when @nalimilan approves moving the manual text into the docstrings).

src/abstractdataframe/selection.jl
```diff
@@ -912,6 +964,14 @@ by the values returned by passed transformations.
 
 $TRANSFORMATION_COMMON_RULES
 
+There the following keyword arguments are supported by `combine`:
```
nalimilan, line 967:
```suggestion
# Keyword arguments
```

src/abstractdataframe/selection.jl
```diff
@@ -596,6 +584,12 @@ the same parent data frame they might get corrupt.
 
 $TRANSFORMATION_COMMON_RULES
 
+There the following keyword arguments are supported by `select!`:
+- `renamecols` : whether in the `cols => function` form automatically generated
```
nalimilan, line 588: I think in general the convention is to repeat the argument type and default value. That's especially useful here because the signature is separated from this section by a large block of text.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -33,6 +33,190 @@ mutable struct GroupedDataFrame{T<:AbstractDataFrame}
...
+
+Return a `GroupedDataFrame` representing a view of an `AbstractDataFrame` split
+into row groups.
+
+# Arguments
+- `df` : an `AbstractDataFrame` to split
+- `cols` : data frame columns to group by. Can be any column selector
+  ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+- `sort` : whether to sort groups according to the values of the grouping columns
+  `cols`; if `sort=false` then the order of groups follows the order of appereance
```
nalimilan, line 47: This phrasing implies we commit to keeping this order at least until 2.x. I think we should say that this is the current implementation but stress that it is undefined. Indeed it's possible that radix sort-based implementations could be faster  in some cases and it would be nice not to close that possibility.

Also:
```suggestion
  `cols`; if `sort=false` then the order of groups follows the order of appearance
```

src/abstractdataframe/selection.jl
bkamins: ok - fixed

src/groupeddataframe/groupeddataframe.jl
bkamins: OK - added a passage that it is undefined

bkamins:
All suggestions are applied, manual and dosstrings are synchronized, and NEWS.md is updated.

src/abstractdataframe/selection.jl
```diff
@@ -15,13 +15,19 @@ const TRANSFORMATION_COMMON_RULES =
...
-    All operations described below are supported both for `GroupedDataFrame` and
-    `AbstractDataFrame`. In the latter case, the data frame is considered as being grouped on no
-    columns (meaning it has a single group, or zero groups if it is empty). The
-    only difference is that in this case the `keepkeys` and `ungroup` keyword
-    arguments are not supported and a data frame is always returned.
-
-    Operations can be applied on each group using one of the following functions:
+    All operations described in this section of the manual are supported both for
+    `AbstractDataFrame` (when split and combine steps are skipped) and
+    `GroupedDataFrame`. Technically, `AbstractDataFrame` is just considered as being
```
nalimilan, line 20:
```suggestion
    All these operations are supported both for
    `AbstractDataFrame` (when split and combine steps are skipped) and
    `GroupedDataFrame`. Technically, `AbstractDataFrame` is just considered as being
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -44,11 +44,13 @@ into row groups.
 - `cols` : data frame columns to group by. Can be any column selector
   ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
 - `sort` : whether to sort groups according to the values of the grouping columns
-  `cols`; if `sort=false` then the order of groups follows the order of appereance
-  of values in the grouping columns, except when all grouping columns provide
-  non-`nothing` `DataAPI.refpool` in which case the order of groups follows the
-  order of values returned by `DataAPI.refpool`. As a particular application of
-  this rule if all `cols` are `CategoricalVector`s then groups are always sorted
+  `cols`; if `sort=false` then the order of groups in the result is undefined
+  and may change in the future releases. In the current implementation the
```
nalimilan, line 48:
```suggestion
  and may change in future releases. In the current implementation
```

nalimilan (approved):

bkamins:
Thank you!

PR #2447, opened 2020-09-21
Implement permutedims

kescobo:
Figured I'd get a jumpstart on Hacktoberfest, and this is something I've wanted in DataFrames for a longtime. This is an attempt to address #2420 

## todo
- [x] correct function signature
- [x] docs
- [x] tests

## So far

I implemented 3 versions of `transpose`:

- `transpose1` copies @tbeason 's stack/unstack method from the issue above
- `transpose2` creates a `Matrix` from everything _except_ the indexing column, and does `permutedims` on it
- `transpose3` builds column-by-column by iterating rows

## Benchmarking

```
julia> builddf(n) = DataFrame(a = [randstring(6) for _ in 1:n], b = rand(n), c = rand(Int, n), d = repeat(["x", "y"], inner=n÷2))
builddf (generic function with 1 method)

julia> df1 = builddf(10);

julia> df2 = builddf(100);

julia> df3 = builddf(1000);

julia> @btime DataFrames.transpose1($df1);
  33.565 μs (398 allocations: 32.61 KiB)

julia> @btime DataFrames.transpose2($df1);
  12.637 μs (179 allocations: 17.81 KiB)

julia> @btime DataFrames.transpose3($df1);
  14.723 μs (172 allocations: 16.58 KiB)

julia> @btime DataFrames.transpose1($df2);
  144.708 μs (1083 allocations: 115.92 KiB)

julia> @btime DataFrames.transpose2($df2);
  50.667 μs (650 allocations: 69.75 KiB)

julia> @btime DataFrames.transpose3($df2);
  137.353 μs (1261 allocations: 119.12 KiB)

julia> @btime DataFrames.transpose1($df3);
  1.168 ms (7924 allocations: 976.30 KiB)

julia> @btime DataFrames.transpose2($df3);
  469.847 μs (5678 allocations: 643.34 KiB)

julia> @btime DataFrames.transpose3($df3);
  1.513 ms (20035 allocations: 1.27 MiB)
```

## First impressions

- `transpose1` (stack/unstack) is slower on small `DataFrame`s but scales well.
     - This also has the possible disadvantage that the columns end up sorted, rather than preserving the row order
- `transpose2` is the fastest, but might be harder to fix type issues (see below)
- `transpose3` is pretty fast for small `DataFrame`s but doesn't scale well at all

## Other thoughts

[...]

Issue #2484, opened 2020-10-13
Release 0.22 tracking


bkamins:
Here I list things to do before 0.22 can be released.

A must (breaking):
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2481(returning multiple columns in transformation functions)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2464 (data frame constructors)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2494 (clean up `unstack`)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2505 (categorical in joins)
- [x] https://github.com/JuliaData/DataFrames.jl/issues/2499 (missings in joins)
- [x] #2495 (add `Cols` support)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2514 (up DataAPI.jl dependency)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2429 (change printing backend - it is breaking as allowed kwargs will change)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2520 (bump coverage tests)
- [x] https://github.com/JuliaData/DataFrames.jl/issues/2506, https://github.com/JuliaData/DataFrames.jl/pull/2519, https://github.com/JuliaData/DataFrames.jl/pull/2518 (remaining categorical dependencies)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2521 (up CategoricalArrays.jl)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2515 (clean up deprecated tests)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2523 (CategoricalArrays.jl integration)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2530 (sync with Tables 1.2)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2456 (improve precompilation)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2536 (code layout)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2538 (clean up tests)
- [x] https://github.com/JuliaData/DataFrames.jl/issues/2525 (improve float printing in text/plain)
- [x] https://github.com/JuliaData/DataFrames.jl/pull/2522 (update docstrings and manual to new printing backend)
- [x] bump DataFrames.jl version
[...]

PR #2488, opened 2020-10-15
Style fixes

kescobo:
You didn't ask for this, but I thought I'd throw it out there anyway. I am 100% fine with closing this because the effort/reward ratio for reviewing is too high (or for any other reason).

This PR does not add or change any functionality, it just fixes some inconsistent style things that came up in #2447. I tried to keep the following things separate, so if any of them are not desired, I should be able to cut them out.

1. The biggest thing is spaces after commas, which I accomplished primarily using regex find-and-replace in VS-Code (there were nearly 3000 instances). Specifically, I did `,([^\s)])` => `, $1`. I initially searched for `,\S`, but there were a bunch of instances of 1-element tuples that I assumed should not be changed (that is, I wanted to leave `(a='b',)` alone).
    1. There were a few file exceptions to global find-and replace: `iris.csv` (for obvious reasons); `test/dataframerow.jl` due to a couple of `@sprint` lines that seemed like they shouldn't be changed - I went in and changed some other stuff in this file manually; `test/io.jl` that had a bunch of output stuff that I didn't want to mess with.
    1. [In `test/grouping.jl`](https://github.com/JuliaData/DataFrames.jl/compare/master...kescobo:style?expand=1#diff-5dcf6d0cf7bc3b35e4d99277738b1a186a2d3b1540141f8b11ee1108defe7199L2094-R2095), there were a couple of examples of spaces _before_ commas that I removed (I only noticed them because there were also no spaces after the commas)
    1. A few lines in `test/cat.jl` with error messages were inconsistent, I tried to unify
    1. I tried to scan through everything else, but it's definitely possible I missed something. I tested locally, but I'm still getting a segfault - I'll fix anything that pops up in CI
1. A few instaces of optional  arguments in docstrings, I tried to unify the style here with `julia` itself (that is, do `function(arg1 [, optional_arg1 [, optional_arg2]]`
1. Zeros after decimal for floats - I used regex to search, but couldn't sort out a good find-replace, so I just scanned through the results for `\d\.\D` and fixed them manually


bkamins:
This is much appreciated. We will review and merge this when things listed in https://github.com/JuliaData/DataFrames.jl/issues/2484 are done.

The issue is that complex PRs (as mentioned there) take hundreds of lines and such PR as this one will lead to merge conflicts.

However, I hope you understand that we put a priority to "functionality" PRs over "clean-up" PRs. So the policy is that it is "clean-up" PRs should wait for "functionality" PRs to be merged and then rebase appropriately. If we have a proposal like this one I stall opening new significant PRs for several days to allow "clean-up"s to be merged in this "waiting period".

(I would have told you the above policy earlier if I knew you started working on this)

Are you OK with this approach (I assume that it is ~1 week to finish the relevant open PRs as there are three of them: a) multi-column return values in transforms, b) changing printing to PrettyTables.jl, c) deprecation of `DataFrame` constructors).

bkamins:
(also note that probably these PRs will introduce some new style inconsistencies - feel free to fix them also then if you would be willing to do this)

kescobo:
>However, I hope you understand that we put a priority to "functionality" PRs over "clean-up" PRs. So the policy is that it is "clean-up" PRs should wait for "functionality" PRs to be merged and then rebase appropriately.

:100: understand and agree

>(I would have told you the above policy earlier if I knew you started working on this)

No worries - it only took me about an 90 min to figure stuff out (and 20 min of that was because github ate my first draft of the description :laughing:), and I bet I could repeat the process in much less time now that I know what to look for.

>Are you OK with this approach (I assume that it is ~1 week to finish the relevant open PRs as there are three of them: a) multi-column return values in transforms, b) changing printing to PrettyTables.jl, c) deprecation of DataFrame constructors).

Completely OK with it - it may be more sensible to just start from scratch at that point. Let's just leave this open for now, and then you can ping me when it would be worth taking another look (though, if it's going to be after Oct 31st and you're willing to mark this as `hacktoberfest-approved`, that would be appreciated :smile:)

bkamins:
Actually the key PR to finalize is the "constructor deprecation" PR (as it touches all test files). When it is merged I can tell you which test files are safe to be touched (other PRs will have a major impact but only at selected files)

bkamins:
Actually looking at your changes - when we go back to this PR - I would split it into 3 smaller PRs:
1. spaces after `,`
2. fixing signatures in docstrings
3. proper float formatting

It will be much easier to review them this way (as when reviewing you only have to check one thing when looking at diff).

kescobo:
Makes sense to me :+1: - I'll close this now, you can ping me on slack or here when that's ready.

bkamins:
OK

PR #2491, opened 2020-10-17
Support multithreading in groupreduce

nalimilan:
Keep the default to a single thread until we find a reliable way of predicting a reasonably optimal number of threads.


src/groupeddataframe/splitapplycombine.jl
```diff
@@ -453,18 +453,22 @@ julia> combine(gd, :, AsTable(Not(:a)) => sum, renamecols=false)
 ```
 """
 function combine(f::Base.Callable, gd::GroupedDataFrame;
-                 keepkeys::Bool=true, ungroup::Bool=true, renamecols::Bool=true)
+                 keepkeys::Bool=true, ungroup::Bool=true, renamecols::Bool=true,
+                 nthreads::Int=NTHREADS)
```
nalimilan, line 457: Ideally we should use a keyword argument name that is standard across packages, but for now I haven't found examples of similar cases elsewhere.

We should also anticipate what it will look like once we switch the default to  the automatic optimal number of threads. Would it be `nthreads=nothing`, `nthreads=:auto`, `nthreads=0`? We could also use another name: `threads=true` looks good at it also works with e.g. `threads=5`. Though using a different type forces recompilation, so maybe `0` is better as a default, even though it's not super logical.

bkamins:
Note that after https://github.com/JuliaData/DataFrames.jl/pull/2481 method signatures will change so this PR will have merge conflicts.

nalimilan:
I've rebased against master. Should be good for a review now. I just hope CI really used 2 threads but it's not visible in the logs AFAICT.

bkamins:
> EDIT: seeing Codecov, probably not.

@quinnj - could you please have a quick look at the CI set-up? Thank you!

src/groupeddataframe/fastaggregates.jl
```diff
@@ -157,24 +157,84 @@ function copyto_widen!(res::AbstractVector{T}, x::AbstractVector) where T
...
-    @inbounds for i in eachindex(incol, groups)
-        gix = groups[i]
-        x = incol[i]
-        if gix > 0 && (condf === nothing || condf(x))
-            # this check should be optimized out if U is not Any
-            if eltype(res) === Any && !isassigned(res, gix)
-                res[gix] = f(x, gix)
-            else
-                res[gix] = op(res[gix], f(x, gix))
+    nt = min(nthreads, Threads.nthreads())
```
bkamins, line 166: Did you intentionally allow 0 or negative values of `nt` here? Maybe we should throw an error? Similarly you allow `true` and `false` passed as `nthreads`. Is this intentional?

nalimilan: Yes we could be stricter, and that would prepare the possibility of switching to `0` as the default to choose automatically the optimal number of threads.

Not sure about `true` and `false`. We could restrict the argument to `Int` if that's simpler. (Otherwise since `true` would be equivalent to `1` it could be counter-intuitive.)

bkamins: We traditionally do `Union{Signed, Unsigned}` in other places to disallow `Bool` 😄. And I agree to disallow `<1` as then `0` can be used later.

nalimilan:
Wait, I just did something completely stupid when setting the environment variable. I've committed a probable fix.

nalimilan:
OK that worked!

src/groupeddataframe/fastaggregates.jl
nalimilan: OK I've moved to `Int` for simplicity. I've also added a check.

nalimilan:
I've added `DataFrames.nthreads` (called to choose the default `nthreads`) and `DataFrames.nthreads!` to set it, to make it easier to choose the number of threads once for a session, as requested by Jan.

tk3369:
I was just thinking about `nthreads!` sounding a little weird because it's not a verb. Given that it's a global setting, would it make sense to create a struct that holds on to all global settings? Then, the user can configure it via `setproperty!`. Just a thought...

nalimilan:
> I was just thinking about `nthreads!` sounding a little weird because it's not a verb. Given that it's a global setting, would it make sense to create a struct that holds on to all global settings? Then, the user can configure it via `setproperty!`. Just a thought...

Yes, I also thought about `setnthreads!`. That's an eternal question about naming conventions for setters in Julia. But note that we have `Random.seed!` already in the stdlib, which is quite similar.

Adding a struct for global settings sounds overkill at this point since it's not clear we're going to add more settings in the future. I'd rather keep this minimal, so that it can be extended if needed in the future, e.g. using Preferences.jl for storage across sessions.

src/other/utils.jl
```diff
@@ -83,3 +83,28 @@ else
...
+"""
+    DataFrames.nthreads!(n::Int)
+
+Set to `n` the default value for the `nthreads` argument, which determines the number
+of CPU threads used by functions that support it when not specified explicitly.
+
+Use [`DataFrames.nthreads`](@ref) to access the value.
+"""
+function nthreads!(n::Int)
+    n > 0 || throw(ArgumentError("n must be equal to or greater than 1 (got $n)"))
```
bkamins, line 108: we might also cap this at `Threads.nthreads()` here so that later `DataFrames.nthreads` does report the number of threads that is going to be actually used

bkamins:
> I'd rather keep this minimal

I am OK with the current design as it is an implementation detail. Though probably having a `mutable struct` instead of `Ref` is more natural (but we can always change it in the future).

Regarding the naming - I am OK with `nthreads!`, but in general I do not have a super strong opinion here.

nalimilan:
The `Ref` is an implementation detail, but we have to decide whether `DataFrames.nthreads` and `DataFrames.nthreads!` are part of the public API. If we're hesitant we could avoid mentioning these in the docs and only use them for benchmarking for now (like `DataFrames.precompile`).

bkamins:
I am OK to have them as a part of the public API.

@StefanKarpinski - if Julia Base were to add an option to change number of available threads dynamically during the Julia session, would `Threads.nthreads!` would be a name for this operation?

NEWS.md
```diff
@@ -1,10 +1,28 @@
...
+
+## Bug fixes
+
 ## New functionalities
 
-* `combine`, `select` and `transform` with `GroupedDataFrame` now accept
-  a `nthreads` argument which enables multithreading for some optimized
-  grouped reductions ([#2491](https://github.com/JuliaData/DataFrames.jl/pull/2491)).
+* `combine`, `select` and `transform` with `GroupedDataFrame` now have
+  support multithreading for some optimized  grouped reductions.
```
bkamins, line 12:
```suggestion
  support multithreading for some optimized grouped reductions.
```

src/DataFrames.jl
```diff
@@ -91,6 +91,8 @@ else
     export only
 end
 
+const NTHREADS = Ref(1)
```
bkamins, line 94: maybe add a docstring?

test/grouping.jl
```diff
@@ -3272,4 +3354,13 @@ end
     @test df == df2
 end
 
+@testset "invalid nthreads" begin
+    gdf = groupby(DataFrame(x=1:10, y=1:10), :y)
+    DataFrames.NTHREADS[] = 0
+    @test_throws ArgumentError select(gdf, :x => sum)
+    @test_throws ArgumentError transform(gdf, :x => sum)
+    @test_throws ArgumentError select!(gdf, :x => sum)
+    @test_throws ArgumentError transform!(gdf, :x => sum)
```
bkamins, line 3363: make sure we do not leave an invalid value after the test
```suggestion
    @test_throws ArgumentError transform!(gdf, :x => sum)
    DataFrames.NTHREADS[] = 1
```

test/grouping.jl
```diff
@@ -98,6 +98,32 @@ function groupby_checked(df::AbstractDataFrame, keys, args...; kwargs...)
     return ogd
 end
 
+function combine_checked(gd::GroupedDataFrame, args...; kwargs...)
```
bkamins, line 101: nice

src/groupeddataframe/fastaggregates.jl
```diff
@@ -163,18 +163,80 @@ function groupreduce!(res::AbstractVector, f, op, condf, adjust, checkempty::Boo
...
-        gix = groups[i]
-        x = incol[i]
-        if gix > 0 && (condf === nothing || condf(x))
-            # this check should be optimized out if U is not Any
-            if eltype(res) === Any && !isassigned(res, gix)
-                res[gix] = f(x, gix)
-            else
-                res[gix] = op(res[gix], f(x, gix))
+    nt = min(NTHREADS[], Threads.nthreads())
+    if nt <= 1 || axes(incol) != axes(groups)
```
bkamins, line 167: when `axes` might be different?

test/grouping.jl
```diff
@@ -2334,6 +2360,13 @@ end
         @test parent(gdf2).y ≅ df.y
         @test parent(gdf2).g === df.g
 
+        # Test that nthreads argument is accepted
```
bkamins, line 2363: should be removed now - right?

nalimilan: Yes. I'm not sure how tests passed locally. There's something weird going on. :-/

src/groupeddataframe/fastaggregates.jl
nalimilan: Since `groups` is a `Vector`, that could happen only when `incol` doesn't use 1-based indexing. It's not completely clear whether we support it but it's probably safer the check that.

bkamins:
This can be closed given the other approach we took? Or we keep it open for the future?

nalimilan:
This PR is complementary with the other threaded grouping PRs, since without it optimized reductions are not multithreaded. We should check whether we can identify conditions under which multithreading is faster than single threading.

bkamins:
I have totally forgotten about this PR. I thought you have merged it :). So what is the status here - are you working on the condition or I should review it as is?

Thank you!

nalimilan:
Given that we now automatically spawn as many tasks as there are threads in other places, requiring users to set `NTHREADS` for this function may be a bit weird. I guess we should try to automatically compute a reasonable number of threads -- but that turned out to be quite tricky.

bkamins:
> equiring users to set `NTHREADS` for this function may be a bit weird

I agree

> but that turned out to be quite tricky.

Is it possible to give some safe condition though? (I do not remember the details of our earlier discussion)

nalimilan:
> Is it possible to give some safe condition though? (I do not remember the details of our earlier discussion)

Probably. I need to look at the data again. What's difficult is that there are several dimensions to cross: number of groups, number of rows, number of threads. And IIRC the overhead due to threading is larger when Julia was started with many threads.

Finally, there's always the possibility that all CPUs are already busy with another threaded operation, so that spawning multiple tasks is counter-productive. Maybe that's not a serious issue.

bkamins:
Should I mark it as 1.0, or keep it 1.x and we will merge it when we are comfortable?

nalimilan:
I think this doesn't have to happen before 1.0.

src/groupeddataframe/fastaggregates.jl
```diff
@@ -163,18 +163,80 @@ function groupreduce!(res::AbstractVector, f, op, condf, adjust, checkempty::Boo
...
+            end
+        end
+        @sync for tid in 1:nt
+            Threads.@spawn begin
+                res′ = res_vec[tid]
+                if adjust !== nothing || checkempty
+                    counts′ = counts_vec[tid]
+                end
+                start = 1 + ((tid - 1) * length(groups)) ÷ nt
+                stop = (tid * length(groups)) ÷ nt
```
bkamins, line 204: avoid overflow on 32-bit machines, on 64-bit machine this is a no-op
```suggestion
                start = 1 + ((tid - 1) * Int64(length(groups))) ÷ nt
                stop = (tid * Int64(length(groups))) ÷ nt
```

bkamins:
@nalimilan - what do we do with this PR? Close it (as it is likely outdated)?

In general - for the future we should rather add a kwarg allowing to choose if users wants to use multithreading c.f. https://github.com/JuliaData/DataFrames.jl/issues/2992.

bkamins:
bump

@nalimilan - what do you think we should do with this PR?

nalimilan:
I'm not sure. There are two aspects to address:
- If we decide to add keyword arguments to choose the number of threads (https://github.com/JuliaData/DataFrames.jl/issues/2992, https://github.com/JuliaData/DataFrames.jl/issues/2988), it would make sense to (rebase and) merge this.
- Otherwise, we would need to find conditions under which using multiple threads is always faster. I haven't retried.

bkamins:
@nalimilan - following our discussion I recommend to do what is written in https://github.com/JuliaData/DataFrames.jl/issues/2988#issuecomment-1046664288.

We would then implement this proposed change. There are the following things to be done to implement this change:

1. implement the feature of this PR
2. implement enabling/disabling threading of what you propose in this PR
3. implement enabling/disabling threading in general in `_combine` (across transformation operations)
4. implement `ntasks` keyword argument in all relevant functions


@nalimilan - can you please comment which parts of this list you would be willing to have a look at (I understand that 1, 2, and 4 must be covered in this PR, but 3 can be done in a separate follow-up PR that I can do if you prefer after this PR is finished but before 1.4 release)

Additional maintenance tasks:
- add tests for all cases
- create and run benchmarks
- update docstrings
- update documentation
- update NEWS.md

nalimilan:
OK I'll rebase this PR, but I'd rather leave point 3 for another one.

bkamins:
Sure. Thank you!

bkamins:
@nalimilan - what are your plans with this PR 😄? (I am asking because it is getting so old that there is a risk that it is better to re-implement it from scratch if we want to have it given the recent changes)

nalimilan:
Now that we have the `threads` keyword argument, we could allow passing an integer to enable multithreaded `groupreduce`. But it would be kind of weird that the default `threads=true` wouldn't enable multithreading here, while e.g. `threads=2` would. Also `threads=2` wouldn't actually limit the number of tasks to 2 as the existing code doesn't really allow doing that (nor should it). And anyway many people wouldn't benefit from it as they will probably expect threading to be automatic, just like it is in other cases.

So it would probably be worth experimenting more to find conditions under which we can be sure starting multiple tasks is faster.

bkamins:
I converted this PR to draft to avoid accidental merging.

PR #2494, opened 2020-10-21
[BREAKING] Refactor unstack

### LINKED ISSUE #2485: clean-up unstack
While reviewing `unstack` to remove CategoricalArrays.jl dependency I noticed:
```
julia> struct A
       x
       end

julia> df = DataFrame(x=[A(1), A(2), A(3)], y=[A(1), A(2), A(3)], variable=1, value=1)
3×4 DataFrame
│ Row │ x    │ y    │ variable │ value │
│     │ A    │ A    │ Int64    │ Int64 │
├─────┼──────┼──────┼──────────┼───────┤
│ 1   │ A(1) │ A(1) │ 1        │ 1     │
│ 2   │ A(2) │ A(2) │ 1        │ 1     │
│ 3   │ A(3) │ A(3) │ 1        │ 1     │

julia> unstack(df)
ERROR: MethodError: no method matching isless(::A, ::A)
```
my plan is to:
* add `sort` kwarg that allows user to decide if sorting should be performed (with default to `true`)
* add kwarg to handle cases when we currently print warnings (on duplicates and on `missing`s) - it will be more consistent to either throw an error or silently accept it (which would be managed by the kwarg)

bkamins:
Fixes #2485


bkamins:
In the end I try to sort and reshape without sorting if it fails.
This PR adds two kwargs which change warning behaviour we had before to error or silently allowing the result (the default is to error).

I have removed CategoricalArrays.jl dependency, but there are 2 places that are problematic because of this (both are corner cases so they can be resolved later).

src/abstractdataframe/reshape.jl
```diff
@@ -377,8 +554,10 @@ end
 
 Base.parent(v::RepeatedVector) = v.parent
 DataAPI.levels(v::RepeatedVector) = levels(parent(v))
-CategoricalArrays.isordered(v::RepeatedVector{<:Union{CategoricalValue, Missing}}) =
-    isordered(parent(v))
+
+# TODO: uncomment when DataAPI.jl supports this
```
bkamins, line 558: here we need https://github.com/JuliaData/DataAPI.jl/issues/23

src/abstractdataframe/reshape.jl
```diff
@@ -394,12 +573,12 @@ Base.reverse(v::RepeatedVector) = RepeatedVector(reverse(parent(v)), v.inner, v.
 Base.similar(v::RepeatedVector, T::Type, dims::Dims) = similar(parent(v), T, dims)
 Base.unique(v::RepeatedVector) = unique(parent(v))
 
-function CategoricalArrays.CategoricalArray(v::RepeatedVector)
-    res = CategoricalArray(parent(v), levels=levels(parent(v)))
-    res.refs = repeat(res.refs, inner = [v.inner], outer = [v.outer])
-    res
-end
-
+# TODO: @nalimilan: is there a generic way to support this?
```
bkamins, line 576: @nalimilan - and here I am not sure what would be best. Maybe you can suggest something?

nalimilan: I don't see a solution here, except defining an interface in DataAPI. But it's probably not worth it: I suspect very few people use `RepeatedVector`, and even fewer call `CategoricalVector` on it.

One way to make this faster would be to implement the `refarray`/`refvalue`/`refpool` API, which would be useful for other cases too. But this can be done later if somebody cares about it.

test/reshape.jl
```diff
@@ -419,23 +420,25 @@ end
...
-
-    v = categorical(["a", "b", "c"])
-    levels!(v, ["b", "c", "a"])
-    rv = DataFrames.RepeatedVector(v, 1, 1)
-    @test !isordered(v)
-    @test !isordered(categorical(v))
-    @test levels(v) == ["b", "c", "a"]
-    @test levels(categorical(v)) == ["b", "c", "a"]
-end
+# TODO: uncomment these tests when we improve handling of categorical here
```
nalimilan, line 423: At least do tests on `categorical(rv)` pass if you drop test on `rv` itself?

src/abstractdataframe/reshape.jl
```diff
@@ -121,137 +198,240 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
-    g = groupby(df, rowkey_ints, sort=true)
-    keycol = categorical(df[!, colkey])
-    droplevels!(keycol)
+                                               renamecols=renamecols,
+                                               allowmissing=allowmissing,
+                                               allowduplicates=allowduplicates)
+    local g
+    try
+        g = groupby(df, rowkey_ints, sort=true)
+    catch
```
nalimilan, line 406: For unordered types, this is going to do most of the work and fail only at the end, right? So that's going to be twice slower as it could. Maybe you could pass `sort=true` only when we know it's going to be more efficient than sorting manually at the end, i.e. when the grouping column implements `refpool` (see exact conditions in `row_group_slots`; would be worth putting in a common function to keep them in sync).

src/abstractdataframe/reshape.jl
```diff
@@ -121,137 +198,240 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
+                    keycol, valuecol, refkeycol, renamecols, allowmissing, allowduplicates)
+end
+
+function _unstack_preprocess_vector(v::AbstractVector)
+    v_unique = unique(v)
+    had_missing = any(ismissing, v_unique)
+    v_unique = intersect(levels(v), v_unique)
+    had_missing && (v_unique = vcat(v_unique, [missing]))
+    len_v = length(v_unique)
+    v_map = Dict([x => i for (i,x) in enumerate(v_unique)])
```
nalimilan, line 351: How hard do you think it would be to use `refpool` here when it's defined, and work with `refarray` for the rest of the processing? That should be much faster for these types (and avoid a regression for `CategoricalArray` and `PooledArrays`).

src/abstractdataframe/reshape.jl
```diff
@@ -121,137 +198,240 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
+    # both unique and Dict should use isequal to test for identity of values
+    @assert length(v_map) == length(v_unique)
+    # if there are no missings in v then set reference index of missing to 0
+    col = similar(v, length(v_unique))
+    copyto!(col, v_unique)
+    return col, v_map, get(v_map, missing, 0)
```
nalimilan, line 357: This assumes `col` has 1-based indexing. Maybe worth `@assert`ing it? Or is this assumption too widespread in the codebase to be worth it?

src/abstractdataframe/reshape.jl
```diff
@@ -121,137 +198,240 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
+`renamecols` is a function called on each unique value in `colkey` which must
+return the name of the column to be created (typically as a string or a
+`Symbol`). Duplicates in resulting names when converted to `Symbol` are not allowed.
+
+If `colkey` contains `missing` values then they will be included  if
+`allowmissing=true` and an error will be thrown otherwise (the default).
+
+If combination of `rowkeys` and `colkey` contains duplicate entries then last
+`value` will be retained and a warning will be printed if `allowduplicates=true`
+and an error will be thrown otherwise (the default).
```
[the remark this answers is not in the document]
bkamins: fixed

src/abstractdataframe/reshape.jl
bkamins: Earlier we did:
```
    refkeycol = categorical(df[!, rowkey])
    droplevels!(refkeycol)
    keycol = categorical(df[!, colkey])
    droplevels!(keycol)
```
which I think is also expensive (please correct me if I am wrong - you essentially have to do a very similar lookup to the one that I propose now).

The point is that we have to make sure that the mapping is to values `1,2,...` in the right order (and that is why `droplevels!` is needed and in general `PooledArray` is not useful as it can contain "holes" in level numbers and no order of them is guaranteed).

The only case when it could be potentially faster if when we already have a `CategoricalVector` that has dropped levels (but even then I am not sure if `droplevels!` is cost-free - you know the internals better here).

src/abstractdataframe/reshape.jl
bkamins: We use `intersect` above, which internally uses `filter` which normally always returns a `Vector` in this case (though it is not documented this way). E.g. OffsetArrays.jl works this way:
```
julia> OA = OffsetArray([1,2,3], -1:1)
3-element OffsetArray(::Array{Int64,1}, -1:1) with eltype Int64 with indices -1:1:
 1
 2
 3

julia> intersect(OA, OA)
3-element Array{Int64,1}:
 1
 2
 3
```
but I will add an assertion as it does not hurt to have it

src/abstractdataframe/reshape.jl
```diff
@@ -121,137 +198,240 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
+    v_unique = unique(v)
+    had_missing = any(ismissing, v_unique)
+    v_unique = intersect(levels(v), v_unique)
+    had_missing && (v_unique = vcat(v_unique, [missing]))
+    len_v = length(v_unique)
+    v_map = Dict([x => i for (i,x) in enumerate(v_unique)])
+    # both unique and Dict should use isequal to test for identity of values
+    @assert length(v_map) == length(v_unique)
+    # if there are no missings in v then set reference index of missing to 0
+    col = similar(v, length(v_unique))
```
bkamins, line 355:
```suggestion
    col = similar(v, length(v_unique))
    @assert firstindex(col) == 1 # we do not support e.g. OffsetArrays.jl yet
```

src/abstractdataframe/reshape.jl
nalimilan: The point is that when the input was a `CategoricalArray` or `PooledArray`, the `categorical` call operated directly on the reference codes, which are much faster to deal with than performing a dict lookup for each value (we look for the new reference code in a vector indexed by old reference codes). And in the ideal case where all levels were used and in the right order (which is quite common at least for `CategoricalArray`), the call only implied doing a (very fast) copy.

src/abstractdataframe/reshape.jl
bkamins: I have checked it, but I am not sure how to extract it out to make everything efficient. I proposed an alternative solution assuming https://github.com/JuliaData/DataAPI.jl/pull/26 is merged, but I am open for suggestions.

test/reshape.jl
bkamins: They will not pass. This is the point that `categorical(rv)` forgets about the ordering:
```
julia> isordered(v)
true

julia> levels(v)
3-element Array{String,1}:
 "b"
 "c"
 "a"

julia> isordered(categorical(rv))
false

julia> levels(categorical(rv))
3-element Array{String,1}:
 "a"
 "b"
 "c"
```

src/abstractdataframe/reshape.jl
bkamins: OK - let us leave it "as is". I expect that `RepeatedVector` is rare enough in practice (and the only negative effect is different order which is not the end of the world I think).

src/abstractdataframe/reshape.jl
bkamins: Here is a benchmark showing the performance penalty of what I proposed:
```
wide = DataFrame(id = 1:6000000,
                        a  = repeat([1:3;], inner = [2000000]),
                        b  = repeat([1:2;], inner = [3000000]),
                        c  = randn(6000000),
                        d  = randn(6000000));

long = stack(wide);
@time unstack(long, :id, :variable, :value); # master: 9.839218 seconds (6.00 M allocations: 2.726 GiB, 8.00% gc time)
                                             # this PR: 15.094627 seconds (132.00 M allocations: 4.858 GiB, 4.90% gc time)
categorical!(long, 1:4);
@time unstack(long, :id, :variable, :value); # master: 9.277382 seconds (18.00 M allocations: 3.044 GiB, 32.15% gc time)
                                             # this PR: 15.921799 seconds (126.99 M allocations: 4.883 GiB, 10.74% gc time)
```

test/reshape.jl
nalimilan: Ah, right. That would require a new function in DataAPI to know when `levels` is O(1). Then we could always preserve the order of levels, and `isordered`. But that's quite low priority.

nalimilan: At any rate probably better remove these tests.

NEWS.md
```diff
@@ -41,6 +41,9 @@
 * in `describe` the specification of custom aggregation is now `function => name`;
   old `name => function` order is now deprecated
   ([#2401](https://github.com/JuliaData/DataFrames.jl/pull/2401))
+* `unstack` now does not depend on CategoricalArrays.jl and has two new keyword
```
nalimilan, line 44: The dependency is a package-level property, for users it doesn't matter whether the implementation of `unstack` depends on CategoricalArrays. This formulation could be interpreted as meaning that the behavior changed.

src/abstractdataframe/reshape.jl
bkamins: Given https://github.com/JuliaData/DataFrames.jl/pull/2494/files#r511616549 maybe we should leave this out and return `false`?

src/abstractdataframe/reshape.jl
nalimilan: I don't think `isordered` is relevant here: `isordered(categorical(["a", "b", "c"]))` is false yet it's the most efficient case.

Maybe the cleanest implementation is to extract these lines to two helper functions so that you can call the first one to group, and then if the second one throws you still have the result.

Or maybe we should reconsider the sorting behavior? I've just checked, and `tidyr::pivot_wider` doesn't sort: it keeps the order of appearance, contrary to its predecessor `reshape2::spread`. That kind of makes sense since if you cared about order you would have sorted the input too. So maybe we should do the same and have `sort=false` by default, so that we can simply throw an error if sorting fails when `sort=true`?

src/abstractdataframe/reshape.jl
nalimilan: Yeah so about 80% slower.

Do you think it would be hard to use `refpool`? If `_unstack_preprocess_vector` returned `keycol_map = refpool(keycol)`, and `rvkeycol = refvalue(keycol)` to be used as `rvkeycol[k]` instead of `keycol[k]`, wouldn't everything work?

NEWS.md
bkamins: I will update the description when we settle if we want to sort or not

src/abstractdataframe/reshape.jl
bkamins: > If `_unstack_preprocess_vector` returned `keycol_map = refpool(keycol)`

It would work if `refpool(keycol)` guaranteed that the values that are present are mapped to numbers 1,2,... in a continuous way. Unfortunately neither `PolledArray` nor `CategoricalArray` guarantee this. But freshly allocated `PooledArray` ensures such behavior and this is what I will try to implement.

bkamins:
> So maybe we should do the same and have `sort=false` by default, so that we can simply throw an error if sorting fails when `sort=true`?

I will implement `sort=false` and not even allow `sort=true`. The reason is that if someone wants to sort columns or rows later it can be done easily enough as the second step. Let us discuss how you like it when I push the changes.

src/abstractdataframe/reshape.jl
bkamins: @nalimilan - I have pushed a `PooledVector` implementation. It is much faster than what we have on master (the first non-categorical test is ~2 seconds). Can you please have a look if you think it is OK? I have still one problem with CategoricalArrays.jl though. I will comment in the code.

src/abstractdataframe/reshape.jl
```diff
@@ -121,169 +198,263 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
+                    renamecols, allowmissing, allowduplicates)
+end
+
+function _unstack_preprocess_vector(v::AbstractVector)
+    # make sure we re-allocate if v is already a PooledArray
+    vp = PooledArray{eltype(v)}(v)
+    rp = DataAPI.refpool(vp)
+    pool = similar(v, length(rp))
+    @assert firstindex(pool) == 1
+    copyto!(pool, rp)
```
bkamins, line 353: this `copyto!` is extremely slow if `v` is a `CategoricalArray` with very many levels. @nalimilan - can we do something to improve it here.

bkamins: I was thinking if we can use a generic fallback using `DataAPI.refarray`, but the problem is that we cannot, because `DataAPI.refarray` does not provide any guarantees on the values that it can hold (all standard implementations hold integers - normally starting from 1), but this is not guaranteed.

The line `copyto!(pool, rp)` is especially difficult, because in my implementation `rp` is a `Vector` of `CategoricalValue` and this is kind of hard to optimize.

@nalimilan - do you have any idea how to generically handle this? (actually in these codes it hurts us that we have `CategoricalArray` type - but the problem is that there is no generic way to unwrap values held in `CategoricalArray` unfortunately AFAICT).

nalimilan: The solution I used in the grouping code is to check whether `DataAPI.refarray(vp) isa AbstractArray{Integer}` and if not bail out and just use the original array instead. You can also bail out if `firstindex(DataAPI.refpool(vp)) != 1`, but it's generally not too hard to handle this by adding `+firstindex(DataAPI.refpool(vp))-1` to `getindex` calls.

But here you call `DataAPI.refpool` and `DataAPI.refarray` on `vp`, which is always a `PooledArray`. Is that intended? In that case you can just work with the internal fields. Also why do you allocate `pool` instead of just using `rp`?

bkamins:
@nalimilan - I have thought on it more and decided to recommend a different path. I propose use `groupby` both for rows and for columns. The overhead should be negligible, and if there is some - we should concentrate to make `groupby` faster. The benefit is that we have one place to maintain the complexity of `refpool` etc. . Also the above test is just faster with `groupby`:
```
wide = DataFrame(id = 1:6000000,
                        a  = repeat([1:3;], inner = [2000000]),
                        b  = repeat([1:2;], inner = [3000000]),
                        c  = randn(6000000),
                        d  = randn(6000000));

long = stack(wide);
@time unstack(long, :id, :variable, :value); # master: 9.839218 seconds (6.00 M allocations: 2.726 GiB, 8.00% gc time)
                                             # with groupby: 2.086908 seconds (6.00 M allocations: 1.702 GiB, 10.32% gc time)
categorical!(long, 1:4);
@time unstack(long, :id, :variable, :value); # master: 9.277382 seconds (18.00 M allocations: 3.044 GiB, 32.15% gc time)
                                             # with groupby: 6.419417 seconds (12.00 M allocations: 2.222 GiB, 6.74% gc time)
```

src/abstractdataframe/reshape.jl
```diff
@@ -121,169 +198,212 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
+    unstack(df, :variable, :value, renamecols=renamecols, allowmissing=allowmissing,
+            allowduplicates=allowduplicates)
 
-function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Int},
-                  colkey::Int, keycol::CategoricalVector,
-                  valuecol::AbstractVector, g::GroupedDataFrame,
-                  renamecols::Function)
-    idx, starts, ends = g.idx, g.starts, g.ends
+function getrefs(g::GroupedDataFrame)
+    idx::Vector{Int}, starts::Vector{Int}, ends::Vector{Int} = g.idx, g.starts, g.ends
```
nalimilan, line 363: Why are type assertions needed?

src/abstractdataframe/reshape.jl
```diff
@@ -121,169 +198,212 @@ function _stackview(df::AbstractDataFrame, measure_vars::AbstractVector{Int},
...
     groupidxs = [idx[starts[i]:ends[i]] for i in 1:length(starts)]
-    rowkey = zeros(Int, size(df, 1))
+    ref = zeros(Int, size(parent(g), 1))
```
nalimilan, line 365: Why not this?
```suggestion
    ref = zeros(Int, length(groupidxs))
```
Iterating over `starts` and `ends` directly to avoid storing the intermediate `groupidxs` would make sense.

But isn't this just recomputing the `gd.groups` field? What am I missing?

nalimilan:
I'm all for it if that's simpler and faster than master. Maybe just check with different shapes to be sure we don't miss something.

bkamins:
> Maybe just check with different shapes to be sure we don't miss something.

Sure.

Also this change will allow in the future to easily add support for multiple key columns - something that was requested.

src/abstractdataframe/reshape.jl
bkamins: as discussed we will stop sorting

src/abstractdataframe/reshape.jl
bkamins: I went to use `groupby` everywhere.

src/abstractdataframe/reshape.jl
bkamins: right - they are not needed. There is an assertion in `getproperty` already.

src/abstractdataframe/reshape.jl
bkamins: I will check - this was the original code I reused

bkamins:
I went down to:
```
julia> @time unstack(long, :id, :variable, :value);
  1.088316 seconds (183 allocations: 598.652 MiB)
```
when I optimized the use of `gdf.groups`

bkamins:
This should be good for a review

test/reshape.jl
```diff
@@ -507,6 +489,97 @@ end
...
+@testset "additional unstack tests" begin
+    df = DataFrame(id=repeat(1:3, inner=3),
+                   id2=repeat(1:3, inner=3),
+                   var=repeat('a':'c', 3),
+                   val=1:9)
+    @test unstack(df, :id, :var, :val) == DataFrame(id=1:3, a=1:3:7, b=2:3:8, c=3:3:9)
+    @test unstack(df, [:id, :id2], :var, :val) == unstack(df, :var, :val) ==
+          DataFrame(id=1:3, id2=1:3, a=1:3:7, b=2:3:8, c=3:3:9)
+
+    # an exertice on current unstack invariants
```
bkamins, line 501: @nalimilan - these tests hopefully cover hard cases.

test/reshape.jl
```diff
@@ -507,6 +489,97 @@ end
...
+        @test unstack(df, :id, :var, :val)[sortperm(unique(df.id)), [1; 1 .+ sortperm(unique(df.var))]] ==
+              DataFrame(id=1:3, a=1:3:7, b=2:3:8, c=3:3:9)
+        @test unstack(df, [:id, :id2], :var, :val) == unstack(df, :var, :val)
+        @test unstack(df, :var, :val)[sortperm(unique(df.id)), [1:2; 2 .+ sortperm(unique(df.var))]] ==
+              DataFrame(id=1:3, id2=1:3, a=1:3:7, b=2:3:8, c=3:3:9)
+
+        df2 = copy(df)
+        df2.id = PooledArray(df.id)
+        df2.var = PooledArray(df.var)
+        @test unstack(df2, :id, :var, :val)[sortperm(df2.id.pool), [1; 1 .+ sortperm(df2.var.pool)]] ==
```
bkamins, line 514: in particular here we have an undocumented behaviour of `groupby`. I have pushed https://github.com/JuliaData/DataFrames.jl/pull/2481/commits/4014fb4552c676f036d309b07d3c074020ab1618 in #2481 to cover this case.

src/abstractdataframe/reshape.jl
```diff
@@ -207,8 +207,8 @@ end
 
 Unstack data frame `df`, i.e. convert it from long to wide format.
 
-Row keys and values from value column will be ordered in the order of their
-appearance in the respective vectors.
+Row keys and values from value column will be ordered in the order produced
+by `groupby`.
```
nalimilan, line 211: That's not very explicit, you have to read the full `groupby` docstring to find out what's the order. I'd just say that it's undefined.

src/abstractdataframe/reshape.jl
```diff
@@ -359,29 +359,40 @@ unstack(df::AbstractDataFrame; renamecols::Function=identity,
...
+    while filled < length(gdf)
+        group = gdf.groups[i]
+        if rows[group] == 0
+            rows[group] = i
+            filled += 1
+        end
+        i += 1
     end
-    return ref
+    return rows # return row index of first occurence of each group in gdf
```
nalimilan, line 380:
```suggestion
    return rows # return row index of first occurrence of each group in gdf
```

src/abstractdataframe/reshape.jl
```diff
@@ -359,29 +359,40 @@ unstack(df::AbstractDataFrame; renamecols::Function=identity,
...
+# this function is tailor made for unstack so it does assume that no groups were
+# dropped (i.e. gdf.groups does not contain 0 entries)
+function find_group_row(gdf::GroupedDataFrame)
+    rows = zeros(Int, length(gdf))
+    isempty(rows) && return rows
+
+    filled = 0
+    i = 1
+    while filled < length(gdf)
+        group = gdf.groups[i]
```
nalimilan, line 373:
```suggestion
    groups = gdf.groups
    while filled < length(gdf)
        group = groups[i]
```

test/reshape.jl
nalimilan:
```suggestion
    # an exercise on current unstack invariants
```

test/reshape.jl
```diff
@@ -499,12 +498,36 @@ end
     @test unstack(df, [:id, :id2], :var, :val) == unstack(df, :var, :val) ==
           DataFrame(id=1:3, id2=1:3, a=1:3:7, b=2:3:8, c=3:3:9)
 
+    # an exertice on current unstack invariants
     Random.seed!(1234)
-    for i in 1:3
+    for i in 1:16
```
nalimilan, line 503: Why so high?

test/reshape.jl
```diff
@@ -507,6 +489,97 @@ end
...
+@testset "additional unstack tests not sortable" begin
+    df = DataFrame(id=repeat(A_TYPE.([2, 1, 3]), inner=3),
+                   id2=repeat(A_TYPE.([2, 1, 3]), inner=3),
+                   var=repeat(A_TYPE.([3, 2, 1]), 3),
+                   val=1:9)
+    @test unstack(df, :id, :var, :val, renamecols=x -> Symbol(:x, x.x)) ==
+          DataFrame(id=A_TYPE.([2, 1, 3]), x3=1:3:7, x2=2:3:8, x1=3:3:9)
+    @test unstack(df, [:id, :id2], :var, :val, renamecols=x -> Symbol(:x, x.x)) ==
+          DataFrame(id=A_TYPE.([2, 1, 3]), id2=A_TYPE.([2, 1, 3]), x3=1:3:7, x2=2:3:8, x1=3:3:9)
+end
```
nalimilan, line 581: We don't need this if we don't sort, right?

nalimilan:
For some reason I hadn't realized that using groupby makes it impossible to use the order of appearance. That's annoying, in particular as it means that even if you sort rows before calling `unstack` the order will be lost. Though I guess it's OK for now as we can switch to a defined order later.

bkamins:
> Though I guess it's OK for now as we can switch to a defined order later.

I think we should resolve this issue in `groupby` in general if we want, as mostly people will think they get the order of appearance there also, which will be not the case for `PooledArray`.

bkamins:
As an additional comment. By default `PooledArray` correctly sorts `pool` by the order of appearance. The only problem is when later you reorder the values:
```
julia> x = PooledArray([1,2,3])
3-element PooledArray{Int64,UInt8,1,Array{UInt8,1}}:
 1
 2
 3

julia> x.pool
3-element Array{Int64,1}:
 1
 2
 3

julia> x[1:3] = 3:-1:1
3:-1:1

julia> x.pool
3-element Array{Int64,1}:
 1
 2
 3

julia> groupby(DataFrame(x=x), :x)
GroupedDataFrame with 3 groups based on key: x
First Group (1 row): x = 1
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
⋮
Last Group (1 row): x = 3
│ Row │ x     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 3     │

```
I think this is rare.

test/reshape.jl
bkamins: to ensure that `[1,1,1,2,2,2,3,3,3]` is reordered in such a way that we test several first orders of appearance other than 1,2,3, as e.g. reordering to `[1,2,3,1,2,3,1,2,3]` does not hit the "hard case" where order of first appearance is broken. I have checked that this is very fast (sub second) anyway (the whole testset for reshape takes 1 minute so this is negligible).

test/reshape.jl
bkamins: we test in this case we order things in the order of first appearance.

src/abstractdataframe/reshape.jl
bkamins: I think we can (and should) make it precise. I have proposed something.

bkamins:
@nalimilan - I have pushed the fix that uses the order of first appearance both for rows and for columns.

test/reshape.jl
```diff
@@ -16,10 +16,10 @@ const ≅ = isequal
     @test levels(df[!, 2]) == ["YYY", "Color", "Mass"] # make sure we did not mess df[!, 2] levels
     #Unstack without specifying a row column
     df3 = unstack(df, :Key, :Value)
-    #The expected output, XXX level should be dropped as it has no rows with this key
+    #The expected output is in odred of appereance
```
nalimilan, line 19:
```suggestion
    # The expected output is in orded of appearance
```

nalimilan (approved):

bkamins:
Thank you!

Issue #2203, opened 2020-04-22
Decide if we want to rename All to Cols


bkamins:
When https://github.com/JuliaData/DataAPI.jl/issues/16 is decided. But we should have this decision before 0.21 release as it is potentially breaking.

bkamins:
Moving to 1.x, as `All` will stay anyway, and `Cols` can be added any time.

CameronBieganek:
I vote for adding `Cols`.

Is there an argument for keeping both `All` and `Cols`, aside from being non-breaking? It seems like it would be a little confusing to have both if they do exactly the same thing.

bkamins:
Well - they do not do exactly the same thing and that is the point I guess :).

`Cols` will be added. Probably we will keep support for `All` as it does not hurt, but just stop advertising it. As long as `DataAPI.jl` has `All` we will support it (as other packages might support it and if people write cross-Tables.jl code it should work).

CameronBieganek:
Ohhhh, right. I had to look really closely at the docstrings to discern the difference:

**All:**

```julia
If `cols == ()`, select *all* columns.
```

**Cols:**

```julia
If `cols == ()`, select *no* columns.
```

(Emphasis mine.)

bkamins:
Closing as the semantics is set. Only `All()` is allowed and it is the same as `:`. Otherwise use `Cols`.

PR #2495, opened 2020-10-21
add Cols support

bkamins:
We have forgotten to add it after DataAPI.jl release


pdeffebach:
Should we deprecate `All()` in favor of `Cols(:)`?

bkamins:
I was thinking about it. DataAPI.jl supports `All` so I thought we can leave it, but we can also deprecate it.

pdeffebach:
x-ref discussions 

* DataFrames: https://github.com/JuliaData/DataFrames.jl/issues/2203
* DataAPI: https://github.com/JuliaData/DataAPI.jl/issues/16

New discussion in https://github.com/JuliaData/DataAPI.jl/issues/27

src/other/index.jl
```diff
@@ -219,6 +219,8 @@ end
 @inline Base.getindex(x::AbstractIndex, idx::Between) = x[idx.first]:x[idx.last]
 @inline Base.getindex(x::AbstractIndex, idx::All) =
     isempty(idx.cols) ? (1:length(x)) : union(getindex.(Ref(x), idx.cols)...)
+@inline Base.getindex(x::AbstractIndex, idx::Cols) =
+    isempty(idx.cols) ? Int[] : union(getindex.(Ref(x), idx.cols)...)
```
nalimilan, line 223: This might be enough. At least something like `getindex.(Ref(x), Symbol[])` gives `Int[]`.
```suggestion
    union(getindex.(Ref(x), idx.cols)...)
```

nalimilan (approved):
Looks good!

src/other/index.jl
bkamins: It is not OK, as:
```
julia> getindex.(Ref(index(df)), DataAPI.Cols().cols)
()
```
(in short `Cols()` stores an empty tuple not `Symbol[]`)

bkamins: and:
```
julia> union(getindex.(Ref(idx), DataAPI.Cols().cols))
Union{}[]
```

bkamins:
OK - I understand that we deprecate `All` in DataAPI.jl. I will make a PR for this.

bkamins:
Thank you! After (if) `All` is removed from DataAPI.jl we need to update DataFrames.jl.

PR #2496, opened 2020-10-22
Add subset

bkamins:
This follows https://github.com/JuliaData/DataFrames.jl/issues/2314 discussion.

CC @pdeffebach @matthieugomez

### DIFF
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -204,8 +204,13 @@ df <- tibble(grp = rep(1:2, 3), x = 6:1, y = 4:9,
 | Rename columns           | `rename(df, x_new = x)`        | `rename(df, :x => :x_new)`             |
 | Pick columns             | `select(df, x, y)`             | `select(df, :x, :y)`                   |
 | Pick & transform columns | `transmute(df, mean(x), y)`    | `select(df, :x => mean, :y)`           |
-| Pick rows                | `filter(df, x >= 1)`           | `subset(df, :x => ByRow(>=(1)))`       |
+| Pick rows                | `filter(df, x >= 1)`           | `subset(df, :x => ByRow(x -> x >= 1))` |
 | Sort rows                | `arrange(df, x)`               | `sort(df, :x)`                         |
+
+Note that in the table above the `x -> x >= 1` predicate can be more compactly
+written as `=>(1)`. The latter form has an additional benefit that it is
+compiled only once per Julia session (as opposed to `x -> x >= 1` which defines
+a new anonymous function every time it is introduced).
 
 As in dplyr, some of these functions can be applied to grouped data frames, in which case they operate by group:
 
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -245,7 +245,7 @@ The following table compares the main functions of DataFrames.jl with Stata:
 | Add new columns        | `egen x_mean = mean(x)` | `transform!(df, :x => mean => :x_mean)` |
 | Rename columns         | `rename x x_new`        | `rename!(df, :x => :x_new)`             |
 | Pick columns           | `keep x y`              | `select!(df, :x, :y)`                   |
-| Pick rows              | `keep if x >= 1`        | `subset!(df, :x => ByRow(>=(1))`        |
+| Pick rows              | `keep if x >= 1`        | `subset!(df, :x => ByRow(x -> x >= 1)`        |
 | Sort rows              | `sort x`                | `sort!(df, :x)`                         |
 
 Note that the suffix `!` (i.e. `transform!`, `select!`, etc) ensures that the operation transforms the dataframe in place, as in Stata
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -204,7 +204,7 @@ df <- tibble(grp = rep(1:2, 3), x = 6:1, y = 4:9,
 | Rename columns           | `rename(df, x_new = x)`        | `rename(df, :x => :x_new)`             |
 | Pick columns             | `select(df, x, y)`             | `select(df, :x, :y)`                   |
 | Pick & transform columns | `transmute(df, mean(x), y)`    | `select(df, :x => mean, :y)`           |
-| Pick rows                | `filter(df, x >= 1)`           | `filter(:x => >=(1), df)`              |
+| Pick rows                | `filter(df, x >= 1)`           | `subset(df, :x => ByRow(>=(1)))`       |
 | Sort rows                | `arrange(df, x)`               | `sort(df, :x)`                         |
 
 As in dplyr, some of these functions can be applied to grouped data frames, in which case they operate by group:
@@ -240,7 +240,7 @@ The following table compares the main functions of DataFrames.jl with Stata:
 | Add new columns        | `egen x_mean = mean(x)` | `transform!(df, :x => mean => :x_mean)` |
 | Rename columns         | `rename x x_new`        | `rename!(df, :x => :x_new)`             |
 | Pick columns           | `keep x y`              | `select!(df, :x, :y)`                   |
-| Pick rows              | `keep if x >= 1`        | `filter!(:x => >=(1), df)`              |
+| Pick rows              | `keep if x >= 1`        | `subset!(df, :x => ByRow(>=(1))`        |
 | Sort rows              | `sort x`                | `sort!(df, :x)`                         |
 
 Note that the suffix `!` (i.e. `transform!`, `select!`, etc) ensures that the operation transforms the dataframe in place, as in Stata
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -12,6 +12,11 @@ df = DataFrame(grp = repeat(1:2, 3), x = 6:-1:1, y = 4:9, z = [3:7; missing], id
 df2 = DataFrame(grp = [1, 3], w = [10, 11])
 ```
 
+Note that in the comparisons presented below predicates like `x -> x >= 1` can
+be more compactly written as `=>(1)`. The latter form has an additional benefit
+that it is compiled only once per Julia session (as opposed to `x -> x >= 1`
+which defines a new anonymous function every time it is introduced).
+
 ## Comparison with the Python package pandas
 
 The following table compares the main functions of DataFrames.jl with the Python package pandas (version 1.1.0):
@@ -206,11 +211,6 @@ df <- tibble(grp = rep(1:2, 3), x = 6:1, y = 4:9,
 | Pick & transform columns | `transmute(df, mean(x), y)`    | `select(df, :x => mean, :y)`           |
 | Pick rows                | `filter(df, x >= 1)`           | `subset(df, :x => ByRow(x -> x >= 1))` |
 | Sort rows                | `arrange(df, x)`               | `sort(df, :x)`                         |
-
-Note that in the table above the `x -> x >= 1` predicate can be more compactly
-written as `=>(1)`. The latter form has an additional benefit that it is
-compiled only once per Julia session (as opposed to `x -> x >= 1` which defines
-a new anonymous function every time it is introduced).
 
 As in dplyr, some of these functions can be applied to grouped data frames, in which case they operate by group:
 
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -1,0 +1,231 @@
+# subset allows a transformation specification without a target column name or a column
+
+_process_subset_pair(i::Int, a::ColumnIndex) = a => Symbol(:x, i)
+_process_subset_pair(i::Int, @nospecialize(a::Pair{<:Any, <:Base.Callable})) =
+    first(a) => last(a) => Symbol(:x, i)
+_process_subset_pair(i::Int, a) =
+    throw(ArgumentError("condition specifier $a is not supported by `subset`"))
+
+function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
+                               @nospecialize(args), skipmissing::Bool)
+    conditions = Any[_process_subset_pair(i, a) for (i, a) in enumerate(args)]
+
+    isempty(conditions) && throw(ArgumentError("at least one condition must be passed"))
+
+    if df isa AbstractDataFrame
+        df_conditions = select(df, conditions..., copycols=!(df isa DataFrame))
+    else
+        df_conditions = select(df, conditions...,
+                               copycols=!(parent(df) isa DataFrame), keepkeys=false)
+    end
+    test_type = skipmissing ? Union{Missing, Bool} : Bool
+    for col in eachcol(df_conditions)
+        if !(eltype(col) <: test_type)
+            throw(ArgumentError("each transformation must produce a vector whose " *
+                                "eltype is subtype of $test_type"))
+        end
+    end
+    @assert ncol(df_conditions) == length(conditions)
+    if ncol(df_conditions) == 1
+        return .===(df_conditions[!, 1], true)
+    else
+        return .===(.&(eachcol(df_conditions)...), true)
+    end
+end
+
+"""
+    subset(df::AbstractDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+    subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
+which all values produced by transformation(s) `args` for a given row is `true`.
+
+If `skipmissing=true` returing `missing` in `args` is allowed and these rows are
+dropped. If `skipmissing=false` (the default) an error is thrown if `missing` is
+present.
+
+Each argument passed in `args` can be either a single column selector or a
+`source_columns => function` transformation specifier following the rules
+described for [`select`](@ref).
+
+Note that as opposed to [`filter`](@ref) the `subset` function works on whole
+columns (or all rows in groups for `GroupedDataFrame`) and by default drops rows
+for which contition is false.
+
+If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
+
+If `GroupedDataFrame` is subsetted then it must include all groups present in the
+`parent` data frame, like in [`select!`](@ref).
+
+See also: [`subset!`](@ref), [`filter`](@ref), [`filter!`](@ref)
+
+# Examples
+
+```
+julia> df = DataFrame(id=1:4, x=[true, false, true, false], y=[true, true, false, false],
+                      z=[true, true, missing, missing], v=[1, 2, 11, 12])
+4×5 DataFrame
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 1     │ 1    │ 1    │ 1       │ 1     │
+│ 2   │ 2     │ 0    │ 1    │ 1       │ 2     │
+│ 3   │ 3     │ 1    │ 0    │ missing │ 11    │
+│ 4   │ 4     │ 0    │ 0    │ missing │ 12    │
+
+julia> subset(df, :x)
+2×5 DataFrame
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 1     │ 1    │ 1    │ 1       │ 1     │
+│ 2   │ 3     │ 1    │ 0    │ missing │ 11    │
+
+julia> subset(df, :v => x -> x .> 3)
+2×5 DataFrame
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 3     │ 1    │ 0    │ missing │ 11    │
+│ 2   │ 4     │ 0    │ 0    │ missing │ 12    │
+
+julia> subset(df, :x, :y => ByRow(!))
+1×5 DataFrame
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 3     │ 1    │ 0    │ missing │ 11    │
+
+julia> subset(df, :x, :z, skipmissing=true)
+1×5 DataFrame
+│ Row │ id    │ x    │ y    │ z     │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool? │ Int64 │
+├─────┼───────┼──────┼──────┼───────┼───────┤
+│ 1   │ 1     │ 1    │ 1    │ 1     │ 1     │
+
+julia> subset(df, :x, :z)
+ERROR: ArgumentError: each transformation must return a vector whose eltype is subtype of Bool
+
+julia> subset(groupby(df, :y), :v => x -> x .> minimum(x))
+2×5 DataFrame
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 2     │ 0    │ 1    │ 1       │ 2     │
+│ 2   │ 4     │ 0    │ 0    │ missing │ 12    │
+```
+"""
+@inline function subset(df::AbstractDataFrame, @nospecialize(args...);
+                        skipmissing::Bool=false, view::Bool=false)
+    row_selector = _get_subset_conditions(df, args, skipmissing)
+    return view ? Base.view(df, row_selector, :) : df[row_selector, :]
+end
+
+@inline function subset(gdf::GroupedDataFrame, @nospecialize(args...);
+                        skipmissing::Bool=false, view::Bool=false)
+    row_selector = _get_subset_conditions(gdf, args, skipmissing)
+    df = parent(gdf)
+    return view ? Base.view(df, row_selector, :) : df[row_selector, :]
+end
+
+"""
+    subset!(df::AbstractDataFrame, args...; skipmissing::Bool=false)
+    subset!(gdf::GroupedDataFrame{DataFrame}, args..., skipmissing::Bool=false)
+
+Update data frame `df` or the parent of `gdf` in place to contain only rows for
+which all values produced by transformation(s) `args` for a given row is `true`.
+
+If `skipmissing=true` returing `missing` in `args` is allowed and these rows are
+dropped. If `skipmissing=false` (the default) an error is thrown if `missing` is
+present.
+
+Each argument passed in `args` can be either a single column selector or a
+`source_columns => function` transformation specifier following the rules
+described for [`select`](@ref).
+
+Note that as opposed to [`filter!`](@ref) the `subset!` function works on whole
+columns (or all rows in groups for `GroupedDataFrame`) and by default drops rows
+for which contition is false.
+
+If `GroupedDataFrame` is subsetted then it must include all groups present in the
+`parent` data frame, like in [`select!`](@ref).
+
+See also: [`subset`](@ref), [`filter`](@ref), [`filter!`](@ref)
+
+# Examples
+
+```
+julia> df = DataFrame(id=1:4, x=[true, false, true, false], y=[true, true, false, false])
+4×3 DataFrame
+│ Row │ id    │ x    │ y    │
+│     │ Int64 │ Bool │ Bool │
+├─────┼───────┼──────┼──────┤
+│ 1   │ 1     │ 1    │ 1    │
+│ 2   │ 2     │ 0    │ 1    │
+│ 3   │ 3     │ 1    │ 0    │
+│ 4   │ 4     │ 0    │ 0    │
+
+julia> subset!(copy(df), :x, :y => ByRow(!));
+
+julia> df
+4×3 DataFrame
+│ Row │ id    │ x    │ y    │
+│     │ Int64 │ Bool │ Bool │
+├─────┼───────┼──────┼──────┤
+│ 1   │ 1     │ 1    │ 1    │
+│ 2   │ 2     │ 0    │ 1    │
+│ 3   │ 3     │ 1    │ 0    │
+│ 4   │ 4     │ 0    │ 0    │
+
+julia> df = DataFrame(id=1:4, y=[true, true, false, false], v=[1, 2, 11, 12]);
+
+julia> subset!(groupby(df, :y), :v => x -> x .> minimum(x));
+
+julia> df
+2×3 DataFrame
+│ Row │ id    │ y    │ v     │
+│     │ Int64 │ Bool │ Int64 │
+├─────┼───────┼──────┼───────┤
+│ 1   │ 2     │ 1    │ 2     │
+│ 2   │ 4     │ 0    │ 12    │
+
+julia> df = DataFrame(id=1:4, x=[true, false, true, false],
+                             z=[true, true, missing, missing], v=1:4);
+
+julia> subset!(df, :x, :z)
+ERROR: ArgumentError: each transformation must return a vector whose eltype is subtype of Bool
+
+julia> subset!(df, :x, :z, skipmissing=true);
+
+julia> df
+1×4 DataFrame
+│ Row │ id    │ x    │ z     │ v     │
+│     │ Int64 │ Bool │ Bool? │ Int64 │
+├─────┼───────┼──────┼───────┼───────┤
+│ 1   │ 1     │ 1    │ 1     │ 1     │
+
+julia> df = DataFrame(id=1:4, x=[true, false, true, false], y=[true, true, false, false],
+                             z=[true, true, missing, missing], v=[1, 2, 11, 12]);
+
+julia> subset!(groupby(df, :y), :v => x -> x .> minimum(x));
+
+julia> df
+2×5 DataFrame
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 2     │ 0    │ 1    │ 1       │ 2     │
+│ 2   │ 4     │ 0    │ 0    │ missing │ 12    │
+```
+"""
+function subset!(df::AbstractDataFrame, @nospecialize(args...); skipmissing::Bool=false)
+    row_selector = _get_subset_conditions(df, args, skipmissing)
+    return delete!(df, findall(!, row_selector))
+end
+
+function subset!(gdf::GroupedDataFrame, @nospecialize(args...);
+                skipmissing::Bool=false)
+    row_selector = _get_subset_conditions(gdf, args, skipmissing)
+    df = parent(gdf)
+    return delete!(df, findall(!, row_selector))
+end
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -38,17 +38,17 @@ _process_subset_pair(i::Int, a) =
     if skipmissing
         return coalesce.(cond, false)
     else
-        # actually currently the inner condition is only evaluated if actually
+        # currently the inner condition is only evaluated if actually
         # we have some missings in cond, but it might change in the future so
         # I leave this check
         if Missing <: eltype(cond)
             pos = findfirst(ismissing, cond)
             if pos !== nothing
                 # note that the user might have passed a GroupedDataFrame but we
-                # then referer to the parent data frame of this GroupedDataFrame
-                throw(ArgumentError("skipmissing=false and in row $pos of the" *
+                # then refer to the parent data frame of this GroupedDataFrame
+                throw(ArgumentError("skipmissing=false but at row $pos the" *
                                     " passed conditions were evaluated to" *
-                                    " missing value"))
+                                    " a missing value"))
             end
         end
         return cond
@@ -62,9 +62,12 @@ end
 Return a copy of data frame `df` or parent of `gdf` containing only rows for
 which all values produced by transformation(s) `args` for a given row are `true`.
 
-If `skipmissing=true`, producing `missing` in conjunction of results produced by
-`args` is allowed and corresponding rows are dropped. If `skipmissing=false`
-(the default) an error is thrown if `missing` is produced.
+If `skipmissing=false` (the default), an error is thrown if the conjunction (i.e. `&`)
+of results produced by `args` is `missing`. If `skipmissing=true`, `missing`
+is allowed and corresponding rows are dropped. Note that if some transformations
+in `args` return `false` and others `missing` for a given row, the conjunction is `false`,
+but that if some return `true` and others `missing`, the conjunction is `missing`
+([three-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic)).
 
 Each argument passed in `args` can be either a single column selector or a
 `source_columns => function` transformation specifier following the rules
@@ -126,7 +129,7 @@ julia> subset(df, :x, :z, skipmissing=true)
    1 │     1  true  true   true      1
 
 julia> subset(df, :x, :z)
-ERROR: ArgumentError: skipmissing=false and in row 3 of the passed conditions were evaluated to missing value
+ERROR: ArgumentError: skipmissing=false but at row 3 the passed conditions were evaluated to a missing value
 
 julia> subset(groupby(df, :y), :v => x -> x .> minimum(x))
 2×5 DataFrame
@@ -220,7 +223,7 @@ julia> df = DataFrame(id=1:4, x=[true, false, true, false],
    4 │     4  false  missing      4
 
 julia> subset!(df, :x, :z)
-ERROR: ArgumentError: skipmissing=false and in row 3 of the passed conditions were evaluated to missing value
+ERROR: ArgumentError: skipmissing=false but at row 3 the passed conditions were evaluated to a missing value
 
 julia> subset!(df, :x, :z, skipmissing=true);
 
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -53,12 +53,6 @@ _process_subset_pair(i::Int, a) =
         end
         return cond
     end
-
-    if ncol(df_conditions) == 1
-        return .===(df_conditions[!, 1], true)
-    else
-        return .===(.&(eachcol(df_conditions)...), true)
-    end
 end
 
 """
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -42,7 +42,7 @@ end
 
 # Note that _get_subset_conditions will have a large compilation time
 # if more than 32 conditions are passed as `args`.
-@noinline function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
+function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
                                 @nospecialize(args), skipmissing::Bool)
     conditions = Any[_process_subset_pair(i, a) for (i, a) in enumerate(args)]
 
@@ -92,7 +92,7 @@ If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
 If a `GroupedDataFrame` is passed then it must include all groups present in the
 `parent` data frame, like in [`select!`](@ref).
 
-See also: [`subset!`](@ref), [`filter`](@ref), [`filter!`](@ref),  [`select`](@ref)
+See also: [`subset!`](@ref), [`filter`](@ref), [`select`](@ref)
 
 # Examples
 
@@ -186,7 +186,7 @@ returns `missing` are skipped).
 If `GroupedDataFrame` is subsetted then it must include all groups present in the
 `parent` data frame, like in [`select!`](@ref).
 
-See also: [`subset`](@ref), [`filter`](@ref), [`filter!`](@ref)
+See also: [`subset`](@ref), [`filter!`](@ref), [`select!`](@ref)
 
 # Examples
 
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -69,7 +69,8 @@ end
 
 """
     subset(df::AbstractDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
-    subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+    subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false,
+           ungroup::Bool=true)
 
 Return a copy of data frame `df` or parent of `gdf` containing only rows for
 which all values produced by transformation(s) `args` for a given row are `true`.
@@ -79,8 +80,7 @@ Each argument passed in `args` can be either a single column selector or a
 described for [`select`](@ref).
 
 Note that as opposed to [`filter`](@ref) the `subset` function works on whole
-columns (or all rows in groups for `GroupedDataFrame`) and by default skips rows
-for which condition is false.
+columns (or all rows in groups for `GroupedDataFrame`).
 
 If `skipmissing=false` (the default) `args` are required to produce vectors
 containing only `Bool` values. If `skipmissing=true`, additionally `missing` is
@@ -88,6 +88,9 @@ allowed and it is treated as `false` (i.e. rows for which one of the conditions
 returns `missing` are skipped).
 
 If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
+
+If `ungroup=false` the return value of the operation on `gdf` is re-grouped
+based on the same grouping columns and `GroupedDataFrame` is returned.
 
 If a `GroupedDataFrame` is passed then it must include all groups present in the
 `parent` data frame, like in [`select!`](@ref).
@@ -157,15 +160,18 @@ julia> subset(groupby(df, :y), :v => x -> x .> minimum(x))
 end
 
 @inline function subset(gdf::GroupedDataFrame, @nospecialize(args...);
-                        skipmissing::Bool=false, view::Bool=false)
+                        skipmissing::Bool=false, view::Bool=false,
+                        ungroup::Bool=true)
     row_selector = _get_subset_conditions(gdf, args, skipmissing)
     df = parent(gdf)
-    return view ? Base.view(df, row_selector, :) : df[row_selector, :]
+    res = view ? Base.view(df, row_selector, :) : df[row_selector, :]
+    return ungroup ? res : groupby(res, groupcols(gdf))
 end
 
 """
     subset!(df::AbstractDataFrame, args...; skipmissing::Bool=false)
-    subset!(gdf::GroupedDataFrame{DataFrame}, args..., skipmissing::Bool=false)
+    subset!(gdf::GroupedDataFrame{DataFrame}, args..., skipmissing::Bool=false,
+            ungroup::Bool=true)
 
 Update data frame `df` or the parent of `gdf` in place to contain only rows for
 which all values produced by transformation(s) `args` for a given row is `true`.
@@ -175,13 +181,15 @@ Each argument passed in `args` can be either a single column selector or a
 described for [`select`](@ref).
 
 Note that as opposed to [`filter!`](@ref) the `subset!` function works on whole
-columns (or all rows in groups for `GroupedDataFrame`) and by default skips rows
-for which contition is false.
+columns (or all rows in groups for `GroupedDataFrame`).
 
 If `skipmissing=false` (the default) `args` are required to produce vectors
 containing only `Bool` values. If `skipmissing=true`, additionally `missing` is
 allowed and it is treated as `false` (i.e. rows for which one of the conditions
 returns `missing` are skipped).
+
+If `ungroup=false` the return value of the operation on `gdf` is re-grouped
+based on the same grouping columns and `GroupedDataFrame` is returned.
 
 If `GroupedDataFrame` is subsetted then it must include all groups present in the
 `parent` data frame, like in [`select!`](@ref).
@@ -264,8 +272,10 @@ function subset!(df::AbstractDataFrame, @nospecialize(args...); skipmissing::Boo
     return delete!(df, findall(!, row_selector))
 end
 
-function subset!(gdf::GroupedDataFrame, @nospecialize(args...); skipmissing::Bool=false)
+function subset!(gdf::GroupedDataFrame, @nospecialize(args...); skipmissing::Bool=false,
+                 ungroup::Bool=true)
     row_selector = _get_subset_conditions(gdf, args, skipmissing)
     df = parent(gdf)
-    return delete!(df, findall(!, row_selector))
+    res = delete!(df, findall(!, row_selector))
+    return ungroup ? res : groupby(res, groupcols(gdf))
 end
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -165,6 +165,7 @@ end
     row_selector = _get_subset_conditions(gdf, args, skipmissing)
     df = parent(gdf)
     res = view ? Base.view(df, row_selector, :) : df[row_selector, :]
+    # TODO: in some cases it might be faster to groupby gdf.groups[row_selector]
     return ungroup ? res : groupby(res, groupcols(gdf))
 end
 
@@ -277,5 +278,6 @@ function subset!(gdf::GroupedDataFrame, @nospecialize(args...); skipmissing::Boo
     row_selector = _get_subset_conditions(gdf, args, skipmissing)
     df = parent(gdf)
     res = delete!(df, findall(!, row_selector))
+    # TODO: in some cases it might be faster to groupby gdf.groups[row_selector]
     return ungroup ? res : groupby(res, groupcols(gdf))
 end
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -7,7 +7,7 @@ _process_subset_pair(i::Int, a) =
     throw(ArgumentError("condition specifier $a is not supported by `subset`"))
 
 function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
-                               @nospecialize(args), skipmissing::Bool)
+                                @nospecialize(args), skipmissing::Bool)
     conditions = Any[_process_subset_pair(i, a) for (i, a) in enumerate(args)]
 
     isempty(conditions) && throw(ArgumentError("at least one condition must be passed"))
@@ -38,9 +38,9 @@ end
     subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
 
 Return a copy of data frame `df` or parent of `gdf` containing only rows for
-which all values produced by transformation(s) `args` for a given row is `true`.
+which all values produced by transformation(s) `args` for a given row are `true`.
 
-If `skipmissing=true` returing `missing` in `args` is allowed and these rows are
+If `skipmissing=true`, returning `missing` in `args` is allowed and corresponding rows are
 dropped. If `skipmissing=false` (the default) an error is thrown if `missing` is
 present.
 
@@ -50,14 +50,14 @@ described for [`select`](@ref).
 
 Note that as opposed to [`filter`](@ref) the `subset` function works on whole
 columns (or all rows in groups for `GroupedDataFrame`) and by default drops rows
-for which contition is false.
+for which condition is false.
 
 If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
 
-If `GroupedDataFrame` is subsetted then it must include all groups present in the
+If a `GroupedDataFrame` is passed then it must include all groups present in the
 `parent` data frame, like in [`select!`](@ref).
 
-See also: [`subset!`](@ref), [`filter`](@ref), [`filter!`](@ref)
+See also: [`subset!`](@ref), [`filter`](@ref), [`filter!`](@ref),  [`select`](@ref)
 
 # Examples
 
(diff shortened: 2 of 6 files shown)


bkamins:
Let us confirm if the functionality is as we like it before I write the tests. Thank you!

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,181 @@
+using DataFrames: normalize_selection, ColumnIndex
+# where allows a transformation specification without a target column name or a column
+
+_process_where_pair(i::Int, a::Pair{<:Any, <:Base.Callable}) = first(a) => last(a) => Symbol(:x, i)
```
pdeffebach, line 4: This `Base.Callable` dispatch is probably causing our latency, right? Good to see the implementation.

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,181 @@
...
+                               @nospecialize(args), dropmissing::Bool)
+    conditions = Any[_process_where_pair(i, a) for (i, a) in enumerate(args)]
+
+    isempty(conditions) && throw(ArgumentError("At least one condition must be passed"))
+
+    if df isa AbstractDataFrame
+        df_conditions = select(df, conditions..., copycols=!(df isa DataFrame))
+    else
+        df_conditions = select(df, conditions...,
+                               copycols=!(parent(df) isa DataFrame), keepkeys=false)
```
pdeffebach, line 18: I'm not sure I understand this. If `sdf isa SubDataFrame`, then `copycols` becomes `false`. But somehow this doesn't error.

pdeffebach (commented):
Thanks for implementing this!

I don't think you've gotten all the `missing` logic worked out yet. 

```
julia> df = DataFrame(a = [1, 1, 1, 2, 2, 2, 2], b = [1, 1, 2, 101, 101, 102, missing]);

julia> where(df, :b => (t -> t .> mean(t)))
ERROR: MethodError: no method matching &(::Missing)
```

I think this only shows up when we return exactly one column that contains missing values. Proabably a missing method for `(&)(missing)`

matthieugomez:
This looks great! I just have two comments about missing values. 

 I'm not a fan of introducing a new kwarg `dropmissing = true`. As I said in another thread, I don't like R multiplication of `na.rm, na.omit` etc.  I think it'd be better to have no keyword argument at all, or to have `skipmissing = true`.

Second, @nalimilan  was concerned about the default of dropping rows with missing values. As I said, I support dropping them because that is why what dplyr/sql/stata do. But maybe it would be worth having a vote somewhere. I'd be ok with erroring on them as long as we can change this behavior with `skipmissing = true`.

docs/src/man/comparisons.md
```diff
@@ -240,7 +240,7 @@ The following table compares the main functions of DataFrames.jl with Stata:
 | Add new columns        | `egen x_mean = mean(x)` | `transform!(df, :x => mean => :x_mean)` |
 | Rename columns         | `rename x x_new`        | `rename!(df, :x => :x_new)`             |
 | Pick columns           | `keep x y`              | `select!(df, :x, :y)`                   |
-| Pick rows              | `keep if x >= 1`        | `filter!(:x => >=(1), df)`              |
+| Pick rows              | `keep if x >= 1`        | `where!(df, :x => >=(1))`               |
```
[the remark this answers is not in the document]
bkamins: Right :smile:

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,181 @@
...
+        end
+    end
+    return isequal.(.&(eachcol(df_conditions)...), true)
+end
+
+"""
+    where(df::AbstractDataFrame, args...; dropmissing::Bool=true, view::Bool=false)
+    where(gdf::GroupedDataFrame, args...; dropmissing::Bool=true, ungroup::Bool=true)
+
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
```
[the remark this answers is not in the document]
bkamins: agreed

src/abstractdataframe/where.jl
bkamins: this should compile pretty fast, but I will check (I can add `@nospecialize` if it causes latency)

src/abstractdataframe/where.jl
bkamins: I think it is correct `parent(df) isa DataFrame` is `false` if the parent is a `SubDataFrame` and we negate it with `!` so:
* I copy columns for `SubDataFrame`
* do not copy columns for `DataFrame`

bkamins:
Thank you both for looking into it. Indeed there are some rough edges - thank you for catching them (such things happen if tests are not written yet)

> `(&)(missing)` error

This is exactly the reason. I will fix this.

> Also, I support dropping rows with missing values because that is why what R/sql/stata do.

I think we should drop them (but clearly indicate that we behave here differently than expected).

> I'm not a fan of introducing a new kwarg `dropmissing = true`.

I would assume that it would be hardly used, so it is not that problematic hopefully. On the other hand having it makes, in particular, sure that users notice more easily that we have a non-standard behavior here.

---

In general - thank you for the feedback. I will fix the mistake, and let us wait for @nalimilan to comment when he has time (I would not put a vote on it - I think us four have enough understanding of the ecosystem to decide).

pdeffebach:
w.r.t `&`. I used `reduce(and, eachcol(df))` which I think might be better because it doesn't specialize on number of arguments.

bkamins:
> w.r.t &. I used `reduce(and, eachcol(df))`

This is less efficient as it will not be able to do broadcasting fusion (you will have allocations grow linearly with the number of conditions).

bkamins:
see https://github.com/JuliaLang/julia/pull/38157

bkamins:
I have thought of pros and cons of dropping row vs erroring on `missing`.

The question is - if we error on missing how do we cleanly handle something like:
```
where(df, :x => ByRow(>=(1)), :y => ByRow(>=(2)))
```
assuming both `:x` and `:y` may contain `missing`s.

You would have to write something like:
```
where(df, :x => x -> coalesce.(x .>= 1, false), :y => y -> coalesce.(y .>= 1, false))
```
which seems verbose (but indeed it is not the end of the world and at least it is explicit).
If we chose to error on missing by default then I think we do not need `dropmissing` kwarg, as then using `coalesce` seems fine for me.

By the way. Can you please comment what would be the recommended style of writing simple filters:
```
where(df, :x => ByRow(>=(1)), :y => ByRow(>=(2)))
```
or
```
where(df, :x => x -> x .>= 1, :y => y -> y .>= 1)
```
?

The benefit of:
```
where(df, :x => ByRow(>=(1)), :y => ByRow(>=(2)))
```
is that it is compiled once, while both:
```
where(df, :x => x -> x .>= 1, :y => y -> y .>= 1)
where(df, :x => x -> coalesce.(x .>= 1, false), :y => y -> coalesce.(y .>= 1, false))
```
would be compiled anew on every call.

bkamins:
@pdeffebach + @matthieugomez : if you could spare some time it would be great if you could comment with examples similar to my above that you find typical use-cases, so that we have more hands-on material to make an informed decision what to do. Thank you!

bkamins:
@nalimilan - do you have an opinion what we should do with handling of `missing` by `where` given the discussion?

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
+using DataFrames: normalize_selection, ColumnIndex
```
nalimilan, line 1: ?

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+            throw(ArgumentError("each transformation must return a vector whose " *
+                                "eltype is subtype of $test_type"))
+        end
+    end
+    @assert ncol(df_conditions) == length(conditions)
+    if ncol(df) == 1
+        return isequal.(df_conditions[!, 1], true)
+    else
+        return isequal.(.&(eachcol(df_conditions)...), true)
+    end
```
nalimilan, line 34: It would probably be better to use `===` instead of `isequal`, as core devs strongly discourage the introduction of "thruthiness" in Julia (that is, objects which are treated as `true` but aren't `Bool`). It shouldn't make a difference in existing cases but it prevents non-standard uses that would have a different behavior than `if`.

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+is present.
+
+Each argument passed in `args` can be either a single column selector or a
+`source_columns => function` transformation specifier following the rules
+described for [`select`](@ref).
+
+Note that as opposed to [`filter`](@ref) the `where` function works on whole columns
+and by default drops rows for which contition is false.
+
+If `view=true` a view instead of a `DataFrame` is returned.
```
nalimilan, line 54:
```suggestion
If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
```

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+
+If `ungroup=true` then a `GroupedDataFrame` is returned using the same columns as
+`gdf` to group by.
```
nalimilan, line 57:
```suggestion
If `ungroup=false` then a `GroupedDataFrame` grouped on the same columns as
`gdf` is returned.
```

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+│ 1   │ 1     │ 1    │ 1    │ 1       │ 1     │
+│ 2   │ 3     │ 1    │ 0    │ missing │ 11    │
+
+julia> where(df, :x, :y => ByRow(!))
+1×5 DataFrame
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 3     │ 1    │ 0    │ missing │ 11    │
+
```
nalimilan, line 89: Also show a simple example like `:v => (x -> x > 2)`?

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 2     │ 0    │ 1    │ 1       │ 2     │
+│ 2   │ 4     │ 0    │ 0    │ missing │ 12    │
+```
+"""
+function where(df::AbstractDataFrame, @nospecialize(args...);
+               dropmissing::Bool=true, view::Bool=false)
+    row_selector = _get_where_conditions(df, args, dropmissing)
+    return view ? Base.view(df, row_selector, :) : df[row_selector, :]
+end
```
nalimilan, line 113: Probably worth using `@inline` to avoid the type instability, with `@noinline` in `_get_where_conditions` (with inference tests to prevent regressions). Same for the `GroupedDataFrame` method.

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+
+"""
+    where!(df::AbstractDataFrame, @nospecialize(args...); dropmissing::Bool=true)
+    where!(gdf::GroupedDataFrame{DataFrame}, args...; ungroup::Bool=true, dropmissing::Bool=true)
+
+Update a data frame `df` or a parent of `gdf` in place to contain only rows for
+which all values produced by `args` transformation in a given row is `true`. If
+`dropmissing=true` (the default) returing `missing` in `args` is allowed and
+these rows are dropped. If `dropmissing=false` an error is thrown if `missing`
+is present.
```
nalimilan, line 123:
```suggestion
Update data frame `df` or the parent of `gdf` in place to contain only rows for
which all values produced by transformation(s) `args` for a given row is `true`.

If `dropmissing=true` (the default) returning `missing` from `args` is allowed and
these rows are dropped. If `dropmissing=false` an error is thrown if `missing`
is present.
```

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+
+"""
+    where(df::AbstractDataFrame, args...; dropmissing::Bool=true, view::Bool=false)
+    where(gdf::GroupedDataFrame, args...; dropmissing::Bool=true, ungroup::Bool=true)
+
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
+which all values produced by `args` transformation in a given row is `true`. If
+`dropmissing=true` (the default) returing `missing` in `args` is allowed and
+these rows are dropped. If `dropmissing=false` an error is thrown if `missing`
+is present.
```
nalimilan, line 45:
```suggestion
which all values produced by transformation(s) `args` for a given row is `true`.

If `dropmissing=true` (the default) returing `missing` in `args` is allowed and
these rows are dropped. If `dropmissing=false` an error is thrown if `missing`
is present.
```

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+
+Each argument passed in `args` can be either a single column selector or a
+`source_columns => function` transformation specifier following the rules
+described for [`select`](@ref).
+
+If `ungroup=true` then a `GroupedDataFrame` is returned using the same columns as
+`gdf` to group by.
```
nalimilan, line 130:
```suggestion
If `ungroup=false` then a `GroupedDataFrame` grouped on the same columns as
`gdf` is returned.
```

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,191 @@
...
+    delete!(df, findall(!, row_selector))
+    return df
+end
+
+function where(gdf::GroupedDataFrame, @nospecialize(args...);
+               dropmissing::Bool=true, ungroup::Bool=true)
+    row_selector = _get_where_conditions(gdf, args, dropmissing)
+    df = parent(gdf)[row_selector, :]
+    return ungroup ? df : groupby(df, groupcols(gdf))
+end
```
nalimilan, line 183: I'd put this above with the other `where` method, since the docstrings separate `where` from `where!` quite clearly.

nalimilan:
Sorry I've given the priority to breaking issues. I think this one could wait for post-0.22 so that we have time to find the best design.

I think having a `dropmissing`/`skipmissing` keyword argument is nice as it's simpler to use than `coalesce`. I'm still not completely sure whether the default should be `true` or `false` (as I said I think we need to discuss a general plan for missing values first).

Regarding the name of that argument, `dropmissing` already exists in our API so it's not completely new, but I agree with @matthieugomez that since other functions take a `skipmissing` keyword argument it would be easier for users to be consistent.

src/abstractdataframe/where.jl
bkamins, line 1:
```suggestion
```

bkamins: leftover from testing

src/abstractdataframe/where.jl
bkamins: OK - changed to `===`.

src/abstractdataframe/where.jl
bkamins: This will not work. But I will add a working example.

src/abstractdataframe/where.jl
bkamins: OK - I have added `@inline`.

src/abstractdataframe/where.jl
bkamins: moved

bkamins:
> I think this one could wait for post-0.22 so that we have time to find the best design.

Sure - we can skip it for now to gather more feedback what would be best (especially concrete examples so that we can judge between both designs).

src/abstractdataframe/where.jl
```diff
@@ -0,0 +1,203 @@
...
+    test_type = dropmissing ? Union{Missing, Bool} : Bool
+    for col in eachcol(df_conditions)
+        if !(eltype(col) <: test_type)
+            throw(ArgumentError("each transformation must return a vector whose " *
+                                "eltype is subtype of $test_type"))
+        end
+    end
+    @assert ncol(df_conditions) == length(conditions)
+    if ncol(df) == 1
+        return .===(df_conditions[!, 1], true)
```
pdeffebach, line 30: We should use `isequal` instead of `===`. It results in a significant compilation speedup for interactive use.

bkamins: Let us discuss it as it was what I originally preferred, but @nalimilan wanted to switch to `===`.

nalimilan: That's surprising. In general `===` should be faster as it's very simple (no dispatch). Can you post a benchmark?

bkamins: maybe it is because it is broadcasted and some bad things (invalidations) happen when this mechanics is triggered?

pdeffebach: Hmm... yes I think I was mistaken. I can't replicate it now.

bkamins:
Given the discussion in https://github.com/JuliaData/DataFrames.jl/pull/2417 I have changed my mind and think that `dropmissing=false` should be the default in DataFrames.jl.

However, what we should discuss I think is @pdeffebach thinks that DataFramesMeta.jl should use the `dropmissing=true` by default (and also then in other operations `missings` would be propagated). In this way we would have a different default in DataFrames.jl and DataFramesMeta.jl consistently (and users could choose what they prefer). Would that work?

pdeffebach:
I have to admit that #2417 is a bit of a niche and tangential issue to base this decision on. 

I am okay with a bifurcation in DataFramesMeta. I suppose I will probably start out with mimicking DataFrames and then change it if enough people complain? 

This really underscores how we need to 

1) Use :x` instead of `x` for lhs assignment, which opens the door for 
2) Allowing keyword arguments to be passed.

bkamins:
I would like to revive this PR and make a decision:
1. do we need `where` (I vote yes); here the only discussion could be that `where` is also a keyword in Julia Base, so e.g. some users might be confused and it is very likely that some IDEs might be confused with highlighting
2. if we add `where` what should be the default for `dropmissing` kwarg (I think it should be `false`)

pdeffebach:
I'm fine with a different name for `where`, not sure what it would be though. 

I'm okay with `dropmissing = true` by default, I guess. One reasoning is that if users want to use `&&` and `||`, which I think is pretty common inside those expressions, they will need to handle `missing`s anyways. Hopefully something like `spreadmissings` can make help this.

matthieugomez:
The only thing I feel strongly about is that dropmissing should be called skipmissing if it defaults to false.

nalimilan:
As I said somewhere before I'm not a fan of the name `where` since all other functions in the API are verbs, and there's a (small) confusion with the `where T` syntax. I'd be inclined to use something else, like `subset` (R term) instead. Anyway our terminology is quite different from SQL since the only term in common is `select`.

Regarding missings, I think we should be consistent everywhere, and since we are strict in `select` and `transform` by default (no propagation of missing values) I think we shouldn't drop missing values automatically here either. But DataFramesMeta should probably choose a more permissive approach in all functions.

bkamins:
OK, so I vote for:
1. `subset` as a verb (this is OK as we treat data frame as a collection of rows)
2. `skipmissing=false` as a kwarg name and default value

bkamins:
This should be good for a review

matthieugomez:
Thanks @bkamins ! I love that it has a `!` form, as well as `skipmissing` and `view` kwargs. 

I don't want to start a debate again, but I just wanted to mention I still like the idea of using `filter`, since it's not a bad application of multiple dispatch, and it would avoid exporting a new name.

pdeffebach:
+1 on the `!`. Stata users will rely like this. It means you can get through a whole scrips without ever creating a second dataframe, which is often a good standard to enforce.

bkamins:
> I still like the idea of using `filter`, since it's not a bad application of multiple dispatch

@nalimilan - what do you think?

pdeffebach:
Wait what would we be dispatching on? by column if `filter(df, :x => fun)` and by row if `filter(:x => fun, df)`. That seems inconsistent / confusing.

bkamins:
This is a good point - the distinction in name (not only on dispatch level) is important as a clear signal of different behavior.

matthieugomez:
Ok! subset is fine anyway.

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
+# subset allows a transformation specification without a target column name or a column
+
+_process_subset_pair(i::Int, a::ColumnIndex) = a => Symbol(:x, i)
+_process_subset_pair(i::Int, @nospecialize(a::Pair{<:Any, <:Base.Callable})) =
+    first(a) => last(a) => Symbol(:x, i)
+_process_subset_pair(i::Int, a) =
+    throw(ArgumentError("condition specifier $a is not supported by `subset`"))
+
+function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
+                               @nospecialize(args), skipmissing::Bool)
```
nalimilan, line 10:
```suggestion
                                @nospecialize(args), skipmissing::Bool)
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+            throw(ArgumentError("each transformation must produce a vector whose " *
+                                "eltype is subtype of $test_type"))
+        end
+    end
+    @assert ncol(df_conditions) == length(conditions)
+    if ncol(df_conditions) == 1
+        return .===(df_conditions[!, 1], true)
+    else
+        return .===(.&(eachcol(df_conditions)...), true)
+    end
```
nalimilan, line 33: Usually we treat `Union{T, Missing}` like `T` when there are no missing values. How about just doing `.&(eachcol(df_conditions)...)` when `skipmissing=false` and let the error be thrown later if a `missing` appears?

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+end
+
+"""
+    subset(df::AbstractDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+    subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
+which all values produced by transformation(s) `args` for a given row is `true`.
```
nalimilan, line 41:
```suggestion
which all values produced by transformation(s) `args` for a given row are `true`.
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+
+If `skipmissing=true` returing `missing` in `args` is allowed and these rows are
```
nalimilan, line 43:
```suggestion
If `skipmissing=true`, returning `missing` in `args` is allowed and corresponding rows are
```
Also, maybe better mention the default behavior first?

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+dropped. If `skipmissing=false` (the default) an error is thrown if `missing` is
+present.
+
+Each argument passed in `args` can be either a single column selector or a
+`source_columns => function` transformation specifier following the rules
+described for [`select`](@ref).
+
+Note that as opposed to [`filter`](@ref) the `subset` function works on whole
+columns (or all rows in groups for `GroupedDataFrame`) and by default drops rows
+for which contition is false.
```
nalimilan, line 53: I don't understand "and by default drops rows for which condition is false".

Also typo:
```suggestion
columns (or all rows in groups for `GroupedDataFrame`) and by default drops rows
for which condition is false.
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+
+If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
+
+If `GroupedDataFrame` is subsetted then it must include all groups present in the
```
nalimilan, line 57:
```suggestion
If a `GroupedDataFrame` is passed then it must include all groups present in the
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+`parent` data frame, like in [`select!`](@ref).
+
+See also: [`subset!`](@ref), [`filter`](@ref), [`filter!`](@ref)
```
nalimilan, line 60:
```suggestion
See also: [`subset!`](@ref), [`filter`](@ref), [`filter!`](@ref),  [`select`](@ref)
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+
+julia> df
+4×3 DataFrame
+│ Row │ id    │ x    │ y    │
+│     │ Int64 │ Bool │ Bool │
+├─────┼───────┼──────┼──────┤
+│ 1   │ 1     │ 1    │ 1    │
+│ 2   │ 2     │ 0    │ 1    │
+│ 3   │ 3     │ 1    │ 0    │
+│ 4   │ 4     │ 0    │ 0    │
```
nalimilan, line 178: It's weird to `copy(df)` to illustrate in-place mutation. :-D

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+julia> df
+2×3 DataFrame
+│ Row │ id    │ y    │ v     │
+│     │ Int64 │ Bool │ Int64 │
+├─────┼───────┼──────┼───────┤
+│ 1   │ 2     │ 1    │ 2     │
+│ 2   │ 4     │ 0    │ 12    │
+
+julia> df = DataFrame(id=1:4, x=[true, false, true, false],
+                             z=[true, true, missing, missing], v=1:4);
```
nalimilan, line 193:
```suggestion
julia> df = DataFrame(id=1:4, x=[true, false, true, false],
                      z=[true, true, missing, missing], v=1:4);
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+
+julia> df
+1×4 DataFrame
+│ Row │ id    │ x    │ z     │ v     │
+│     │ Int64 │ Bool │ Bool? │ Int64 │
+├─────┼───────┼──────┼───────┼───────┤
+│ 1   │ 1     │ 1    │ 1     │ 1     │
+
+julia> df = DataFrame(id=1:4, x=[true, false, true, false], y=[true, true, false, false],
+                             z=[true, true, missing, missing], v=[1, 2, 11, 12]);
```
nalimilan, line 208:
```suggestion
julia> df = DataFrame(id=1:4, x=[true, false, true, false], y=[true, true, false, false],
                      z=[true, true, missing, missing], v=[1, 2, 11, 12]);
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+│ Row │ id    │ x    │ y    │ z       │ v     │
+│     │ Int64 │ Bool │ Bool │ Bool?   │ Int64 │
+├─────┼───────┼──────┼──────┼─────────┼───────┤
+│ 1   │ 2     │ 0    │ 1    │ 1       │ 2     │
+│ 2   │ 4     │ 0    │ 0    │ missing │ 12    │
+```
+"""
+function subset!(df::AbstractDataFrame, @nospecialize(args...); skipmissing::Bool=false)
+    row_selector = _get_subset_conditions(df, args, skipmissing)
+    return delete!(df, findall(!, row_selector))
```
nalimilan, line 223: Doesn't `delete!` call `findall` automatically? Same below.

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,231 @@
...
+end
+
+function subset!(gdf::GroupedDataFrame, @nospecialize(args...);
+                skipmissing::Bool=false)
```
nalimilan, line 227:
```suggestion
function subset!(gdf::GroupedDataFrame, @nospecialize(args...); skipmissing::Bool=false)
```

test/grouping.jl
```diff
@@ -3272,4 +3272,166 @@ end
...
+        @test subset(gdf, :x, :id => ByRow(<(4))) ≅
+              filter([:x, :id] => (x, id) -> x && id < 4, df)
+        @test subset(gdf, :x, :id => ByRow(<(4)), view=true) ≅
+              filter([:x, :id] => (x, id) -> x && id < 4, df)
+        @test_throws ArgumentError subset(gdf)
+        @test isempty(subset(gdf, :x, :x => ByRow(!)))
+    end
+
+    df = copy(refdf)
+    @test subset!(df, :x) ≅ df ≅ filter(:x => identity, refdf)
```
nalimilan, line 3331: Impressive set of tests! I think we only miss the most trivial one: check that `subset!(df, ...) === df` at least once for each method.

docs/src/man/comparisons.md
```diff
@@ -204,7 +204,7 @@ df <- tibble(grp = rep(1:2, 3), x = 6:1, y = 4:9,
 | Rename columns           | `rename(df, x_new = x)`        | `rename(df, :x => :x_new)`             |
 | Pick columns             | `select(df, x, y)`             | `select(df, :x, :y)`                   |
 | Pick & transform columns | `transmute(df, mean(x), y)`    | `select(df, :x => mean, :y)`           |
-| Pick rows                | `filter(df, x >= 1)`           | `filter(:x => >=(1), df)`              |
+| Pick rows                | `filter(df, x >= 1)`           | `subset(df, :x => ByRow(>=(1)))`       |
```
nalimilan, line 207: Maybe we should also show `x -> x >= 1`? It may be simpler to grasp for many users. Or does that take too much space?

bkamins: changed. I also added a description why `>=(1)` is preferred from the perspective of compilation cost.

src/abstractdataframe/subset.jl
bkamins: this is tricky. But maybe it is better? I will allow what you propose. Essentially the thing is that `missing & false` is `false`. Which means that even if `skipmissing=false` we will not error in such case, but I think it is OK.

I also added a test for this specific case

src/abstractdataframe/subset.jl
bkamins: of course - fixed.

src/abstractdataframe/subset.jl
bkamins: it does, but `delete!` does the reverse operation (i.e. it requires passing rows to be deleted, and in `select` we select rows to be kept, so I have to negate the result before passing it to `delete!`)

test/grouping.jl
bkamins: added

src/abstractdataframe/where.jl
bkamins: The code is type unstable even with `@inline` and `@noinline` in `_get_where_conditions`. Do you think it can be resolved somehow? (and do you think it is so problematic given `DataFrame` is type unstable anyway?)

bkamins:
@nalimilan - than you for the review. I have applied all your suggestions. The only change in the logic is that now the following passes without an error (but I think it is OK):
```
julia> subset(DataFrame(a=1), :a => x -> false, :a => x -> missing)
0×1 DataFrame
```
(the reason is that `missing` is overridden by `false` so we can determine the result even with this `missing`)

I had also to update the docstring to reflect this change (I have marked the location).

docs/src/man/comparisons.md
```diff
@@ -204,9 +204,14 @@ df <- tibble(grp = rep(1:2, 3), x = 6:1, y = 4:9,
 | Rename columns           | `rename(df, x_new = x)`        | `rename(df, :x => :x_new)`             |
 | Pick columns             | `select(df, x, y)`             | `select(df, :x, :y)`                   |
 | Pick & transform columns | `transmute(df, mean(x), y)`    | `select(df, :x => mean, :y)`           |
-| Pick rows                | `filter(df, x >= 1)`           | `filter(:x => >=(1), df)`              |
+| Pick rows                | `filter(df, x >= 1)`           | `subset(df, :x => ByRow(x -> x >= 1))` |
 | Sort rows                | `arrange(df, x)`               | `sort(df, :x)`                         |
 
+Note that in the table above the `x -> x >= 1` predicate can be more compactly
```
bkamins, line 210: @nalimilan - I added an extended description why `>=(1)` is preferable.

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,263 @@
...
+end
+
+"""
+    subset(df::AbstractDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+    subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
+which all values produced by transformation(s) `args` for a given row are `true`.
+
+If `skipmissing=true`, producing `missing` in conjunction of results produced by
```
bkamins, line 71: @nalimilan - this is a new description of `skipmissing` handling.

src/abstractdataframe/subset.jl
```diff
@@ -18,14 +18,42 @@ function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
...
+    if ncol(df_conditions) == 1
+        cond = df_conditions[!, 1]
+    else
+        cond = .&(eachcol(df_conditions)...)
+    end
+
+    @assert eltype(cond) <: Union{Bool, Missing}
+
+    if skipmissing
+        return coalesce.(cond, false)
```
nalimilan, line 39: Better fuse this with `.&` above?

docs/src/man/comparisons.md
```diff
@@ -240,7 +245,7 @@ The following table compares the main functions of DataFrames.jl with Stata:
 | Add new columns        | `egen x_mean = mean(x)` | `transform!(df, :x => mean => :x_mean)` |
 | Rename columns         | `rename x x_new`        | `rename!(df, :x => :x_new)`             |
 | Pick columns           | `keep x y`              | `select!(df, :x, :y)`                   |
-| Pick rows              | `keep if x >= 1`        | `filter!(:x => >=(1), df)`              |
+| Pick rows              | `keep if x >= 1`        | `subset!(df, :x => ByRow(>=(1))`        |
```
nalimilan, line 248:
```suggestion
| Pick rows              | `keep if x >= 1`        | `subset!(df, :x => ByRow(x -> x >= 1)`        |
```
Not sure what to do about the explanation you added. Maybe put it at the top of the page so that it applies to all section?

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+    else
+        # actually currently the inner condition is only evaluated if actually
```
nalimilan, line 41:
```suggestion
        # currently the inner condition is only evaluated if actually
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+        # we have some missings in cond, but it might change in the future so
+        # I leave this check
+        if Missing <: eltype(cond)
+            pos = findfirst(ismissing, cond)
+            if pos !== nothing
+                # note that the user might have passed a GroupedDataFrame but we
+                # then referer to the parent data frame of this GroupedDataFrame
```
nalimilan, line 48:
```suggestion
                # then refer to the parent data frame of this GroupedDataFrame
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+                throw(ArgumentError("skipmissing=false and in row $pos of the" *
+                                    " passed conditions were evaluated to" *
+                                    " missing value"))
```
nalimilan, line 51:
```suggestion
                throw(ArgumentError("skipmissing=false but at row $pos the" *
                                    " passed conditions were evaluated to" *
                                    " a missing value"))
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+"""
+    subset(df::AbstractDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+    subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
+which all values produced by transformation(s) `args` for a given row are `true`.
+
+If `skipmissing=true`, producing `missing` in conjunction of results produced by
+`args` is allowed and corresponding rows are dropped. If `skipmissing=false`
+(the default) an error is thrown if `missing` is produced.
```
nalimilan, line 67: Maybe worth being even more explicit:
```suggestion
If `skipmissing=false` (the default), an error is thrown if the conjunction (i.e. `&`)
of results produced by `args` is `missing`. If `skipmissing=true`, `missing`
is allowed and corresponding rows are dropped. Note that if some transformations
in `args` return `false` and others `missing` for a given row, the conjunction is `false`,
but that if some return `true` and others `missing`, the conjunction is `missing`
([three-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic)).
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+"""
+    subset!(df::AbstractDataFrame, args...; skipmissing::Bool=false)
+    subset!(gdf::GroupedDataFrame{DataFrame}, args..., skipmissing::Bool=false)
+
+Update data frame `df` or the parent of `gdf` in place to contain only rows for
+which all values produced by transformation(s) `args` for a given row is `true`.
+
+If `skipmissing=true`, producing `missing` in conjunction of results produced by
+`args` is allowed and corresponding rows are dropped. If `skipmissing=false`
+(the default) an error is thrown if `missing` is produced.
```
nalimilan, line 162: If `skipmissing=false` (the default), an error is thrown if the conjunction (i.e. `&`)
of results produced by `args` is `missing`. If `skipmissing=true`, `missing`
is allowed and corresponding rows are dropped. Note that if some transformations
in `args` return `false` and others `missing` for a given row, the conjunction is `false`,
but that if some return `true` and others `missing`, the conjunction is `missing`
([three-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic)).

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+ Row │ id     x      z        v
+     │ Int64  Bool   Bool?    Int64
+─────┼──────────────────────────────
+   1 │     1   true     true      1
+   2 │     2  false     true      2
+   3 │     3   true  missing      3
+   4 │     4  false  missing      4
+
+julia> subset!(df, :x, :z)
+ERROR: ArgumentError: skipmissing=false and in row 3 of the passed conditions were evaluated to missing value
```
nalimilan, line 223:
```suggestion
ERROR: ArgumentError: skipmissing=false but at row 3 the passed conditions were evaluated to a missing value
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+
+julia> subset(df, :x, :z, skipmissing=true)
+1×5 DataFrame
+ Row │ id     x     y     z      v
+     │ Int64  Bool  Bool  Bool?  Int64
+─────┼─────────────────────────────────
+   1 │     1  true  true   true      1
+
+julia> subset(df, :x, :z)
+ERROR: ArgumentError: skipmissing=false and in row 3 of the passed conditions were evaluated to missing value
```
nalimilan, line 129:
```suggestion
ERROR: ArgumentError: skipmissing=false but at row 3 the passed conditions were evaluated to a missing value
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,257 @@
...
+
+"""
+    subset(df::AbstractDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+    subset(gdf::GroupedDataFrame, args...; skipmissing::Bool=false, view::Bool=false)
+
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
+which all values produced by transformation(s) `args` for a given row are `true`.
+
+If `skipmissing=true`, producing `missing` in conjunction of results produced by
+`args` is allowed and corresponding rows are dropped. If `skipmissing=false`
```
matthieugomez, line 66: and corresponding rows are skipped

nalimilan:
Actually, maybe we should throw an error if any of the transformations returns missing. People who keep skipmissing=false probably don’t expect 3VL to used. Also it can be made less strict if needed without breaking.

bkamins:
> Actually, maybe we should throw an error if any of the transformations returns missing.

Let us settle this issue before I process your comments (as essentially all of them are related).

The argument that in the future we can be more flexible is good. In order to clearly expose the decision to be made we should decide if we want to follow this:
```
julia> if missing && false
       1
       else
       2
       end
ERROR: TypeError: non-boolean (Missing) used in boolean context
Stacktrace:
 [1] top-level scope at REPL[1]:1
```
or this
```
julia> if missing & false
       1
       else
       2
       end
2
```
behavior. Actually we might define:

```
_and(x::Bool) = x
_and(x::Bool, y...) = x && _and(y...)

_and_missing(x::Bool) = x
_and_missing(x::Bool, y...) = x && _and_missing(y...)
_and_missing(x::Missing, y...) = false
```
and let Julia Base handle the thing automatically for us by appropriately calling broadcasted versions of these functions (if we went for this probably it might make sense to unroll the definitions a bit for performance).

nalimilan:
Yes, defining helpers that call `&&` might be enough.  Though if we want to throw an even more explicit error pointing to `skipmissing=true` we could use e.g. `_checknotmissing(x:::Any) = ismissing(x) ? throw(...) : x`.

bkamins:
I will propose some design, but the question is if we want `&&` or `&` behavior? @pdeffebach, @matthieugomez - do you have any opinion here?

bkamins:
@pdeffebach, @matthieugomez - do you have any preference here? If not I will change the code to use the `&&` (strict) rule.

docs/src/man/comparisons.md
bkamins: OK - moved.

src/abstractdataframe/subset.jl
bkamins: OK.

src/abstractdataframe/subset.jl
bkamins: this comment is outdated.

bkamins:
I have updated the PR following the suggestions to use `&&` and be strict (the issue is that `&&` is not strict by default as it does short circuit evaluation).

We are slightly slower with `_and` than with `&`:
```
julia> v = [rand(Bool, 10^6) for i in 1:10];
julia> @btime .&($v...);
  4.170 ms (26 allocations: 127.33 KiB)

julia> @time DataFrames._and.($v...);
  5.452 ms (26 allocations: 127.33 KiB)
```
but I think it is acceptable.

The only problem is that if we pass more than 32 arguments to `subset` it slows down significantly because of the compilation in Julia:
```
julia> df = DataFrame([rand(Bool, 10^6) for i in 1:40], :auto);

julia> @time subset(df, 1:40...);
 20.440743 seconds (316.00 M allocations: 15.945 GiB, 3.82% gc time)

julia> @time subset(df, 1:40...);
 21.541057 seconds (316.00 M allocations: 15.945 GiB, 2.95% gc time)

julia> @time subset(df, 1:33...);
  2.470614 seconds (36.00 M allocations: 1.833 GiB, 3.90% gc time)

julia> @time subset(df, 1:33...);
  2.494020 seconds (36.00 M allocations: 1.833 GiB, 3.78% gc time)

julia> @time subset(df, 1:32...);
  0.026481 seconds (1.91 k allocations: 229.094 KiB)

julia> @time subset(df, 1:32...);
  0.022544 seconds (1.91 k allocations: 229.094 KiB)

julia> @time subset(df, 1:31...);
  0.524459 seconds (2.22 M allocations: 85.684 MiB, 1.20% gc time)

julia> @time subset(df, 1:31...);
  0.019008 seconds (1.86 k allocations: 226.172 KiB)
```
(I am not sure if it can be easily resolved - or maybe we should add a separate path for many conditions? - the question is if we feel passing many conditions is a realistic use-case?)

src/abstractdataframe/subset.jl
```diff
@@ -62,23 +49,26 @@ end
...
 described for [`select`](@ref).
 
 Note that as opposed to [`filter`](@ref) the `subset` function works on whole
-columns (or all rows in groups for `GroupedDataFrame`) and by default drops rows
+columns (or all rows in groups for `GroupedDataFrame`) and by default skips rows
 for which condition is false.
 
+If `skipmissing=false` (the default) vectors containing only `Bool` values are
+allowed to be produced by `args`. If `skipmissing=true`, additionally `missing`
+is allowed and it is treated as `false` (corresponding rows are skipped).
```
nalimilan, line 62:
```suggestion
If `skipmissing=false` (the default) `args` are required to produce vectors containing
only `Bool` values.
If `skipmissing=true`, additionally `missing` is allowed and it is treated as `false`
(i.e. rows for which one of the conditions returns `missing` are skipped).
```

nalimilan:
TBH I'm not really concerned about supporting more than 32 arguments efficiently. :-D I would even avoid mentioning it.

src/abstractdataframe/subset.jl
```diff
@@ -6,6 +6,17 @@ _process_subset_pair(i::Int, @nospecialize(a::Pair{<:Any, <:Base.Callable})) =
 _process_subset_pair(i::Int, a) =
     throw(ArgumentError("condition specifier $a is not supported by `subset`"))
 
+_and(x::Bool) = x
+_and(x::Bool, y::Bool...) = x && _and(y...)
+_and(x::Any...) =
+    throw(ArgumentError("set of non-boolean values $x passed in boolean context"))
```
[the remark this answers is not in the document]
bkamins: This would not be correct (it would problems with dispatch). But I will use your proposal to print it (along with condition number).

bkamins:
> I would even avoid mentioning it.

OK - I will add a comment in the code though.

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,271 @@
...
+    xv = x[loc]
+    throw(ArgumentError("value $xv was returned in condition number $loc" *
+                        "but only true, false, or missing are allowed"))
+end
+
+
+# Note that _get_subset_conditions will have a large compilation time
+# if more than 32 conditions are passed as `args`.
+@noinline function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
+                                @nospecialize(args), skipmissing::Bool)
```
nalimilan, line 46: Actually, is `@noinline` really necessary? I know I suggested that, but maybe `@inline` for `@subset` is enough and doesn't have to be complemented with `@noinline` here: I don't see anything that requires a function barrier.

Also:
```suggestion
@noinline function _get_subset_conditions(df::Union{AbstractDataFrame, GroupedDataFrame},
                                          @nospecialize(args), skipmissing::Bool)
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,271 @@
...
+```
+"""
+@inline function subset(df::AbstractDataFrame, @nospecialize(args...);
+                        skipmissing::Bool=false, view::Bool=false)
+    row_selector = _get_subset_conditions(df, args, skipmissing)
+    return view ? Base.view(df, row_selector, :) : df[row_selector, :]
+end
+
+@inline function subset(gdf::GroupedDataFrame, @nospecialize(args...);
+                        skipmissing::Bool=false, view::Bool=false)
```
nalimilan, line 160: Doesn't have to be implemented right now, but I guess at some point it will make sense to support `ungroup=false` to get a `GroupedDataFrame` like with `select`/`transform`?

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,271 @@
...
+containing only `Bool` values. If `skipmissing=true`, additionally `missing` is
+allowed and it is treated as `false` (i.e. rows for which one of the conditions
+returns `missing` are skipped).
+
+If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
+
+If a `GroupedDataFrame` is passed then it must include all groups present in the
+`parent` data frame, like in [`select!`](@ref).
+
+See also: [`subset!`](@ref), [`filter`](@ref), [`filter!`](@ref),  [`select`](@ref)
```
nalimilan, line 95:
```suggestion
See also: [`subset!`](@ref), [`filter`](@ref), [`select`](@ref)
```

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,271 @@
...
+Return a copy of data frame `df` or parent of `gdf` containing only rows for
+which all values produced by transformation(s) `args` for a given row are `true`.
+
+Each argument passed in `args` can be either a single column selector or a
+`source_columns => function` transformation specifier following the rules
+described for [`select`](@ref).
+
+Note that as opposed to [`filter`](@ref) the `subset` function works on whole
+columns (or all rows in groups for `GroupedDataFrame`) and by default skips rows
+for which condition is false.
```
nalimilan, line 83: (I vaguely remember spotting this before but I can't find a previous comment.) Why "and by default skips rows
for which condition is false"? This is consistent with `filter` AFAICT.

src/abstractdataframe/subset.jl
```diff
@@ -0,0 +1,271 @@
...
+
+If `skipmissing=false` (the default) `args` are required to produce vectors
+containing only `Bool` values. If `skipmissing=true`, additionally `missing` is
+allowed and it is treated as `false` (i.e. rows for which one of the conditions
+returns `missing` are skipped).
+
+If `GroupedDataFrame` is subsetted then it must include all groups present in the
+`parent` data frame, like in [`select!`](@ref).
+
+See also: [`subset`](@ref), [`filter`](@ref), [`filter!`](@ref)
```
nalimilan, line 189:
```suggestion
See also: [`subset`](@ref), [`filter!`](@ref), [`select!`](@ref)
```

src/abstractdataframe/where.jl
nalimilan: That's weird. We use this kind of pattern elsewhere and it works, right? Is `row_selector` correctly inferred?

src/abstractdataframe/subset.jl
bkamins: OK - I will remove it.

src/abstractdataframe/subset.jl
bkamins: I have not seen the earlier comment 😄 but this comment is correct. It is left-over of the old design. I will correct the description.

src/abstractdataframe/subset.jl
bkamins: This is what I had in an initial implementation. Then I thought that the problem is that `ungroup=false` does not give us any benefit (as in `select`/`transform` as we have to run `groupby` anyway) and complicates the API. But given you raised it I think it is OK to have it for consistency. I will add it back.

src/abstractdataframe/where.jl
bkamins: @nalimilan - given we add `ungroup` kwarg I am not sure we can get type stability here, do we? (as now we have yet another option of the return value related to the second kwarg)

nalimilan: `ungroup` shouldn't really make things worse as long as the function is inlined and the value of arguments know at compile time. But if that doesn't work in the first place...

src/abstractdataframe/subset.jl
```diff
@@ -89,10 +89,13 @@ returns `missing` are skipped).
 
 If `view=true` a `SubDataFrame` view  is returned instead of a `DataFrame`.
 
+If `ungroup=false` the return value of the operation on `gdf` is re-grouped
+based on the same grouping columns and `GroupedDataFrame` is returned.
```
nalimilan, line 93:
```suggestion
If `ungroup=false` the data frame is re-grouped based on the same grouping columns
as `gdf` and a `GroupedDataFrame` is returned.
```

src/abstractdataframe/subset.jl
nalimilan: We could apply the grouping code directly to group indices, using the same path as `refarray`/`refpool`. It could often be faster than using the original columns. Maybe leave a TODO?

src/abstractdataframe/where.jl
bkamins: This is what I have now:
```
julia> @code_warntype subset(DataFrame(a=1:10), :a => ByRow(<(4)), view=false, skipmissing=false)
Variables
  #unused#::Core.Compiler.Const(DataFrames.var"#subset##kw"(), false)
  @_2::NamedTuple{(:view, :skipmissing),Tuple{Bool,Bool}}
  @_3::Core.Compiler.Const(DataFrames.subset, false)
  df::DataFrame
  args::Tuple{Pair{Symbol,ByRow{Base.Fix2{typeof(<),Int64}}}}
  skipmissing::Bool
  view::Bool
  @_8::Bool
  @_9::Bool

Body::Any
1 ── %1  = Base.haskey(@_2, :skipmissing)::Core.Compiler.Const(true, false)
│          %1
│    %3  = Base.getindex(@_2, :skipmissing)::Bool
│    %4  = (%3 isa DataFrames.Bool)::Core.Compiler.Const(true, false)
│          %4
└───       goto #3
2 ──       Core.Compiler.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :skipmissing, DataFrames.Bool, %3)), false)
└───       Core.Compiler.Const(:(Core.throw(%7)), false)
3 ┄─       (@_8 = %3)
└───       goto #5
4 ──       Core.Compiler.Const(:(@_8 = false), false)
5 ┄─       (skipmissing = @_8)
│    %13 = Base.haskey(@_2, :view)::Core.Compiler.Const(true, false)
│          %13
│    %15 = Base.getindex(@_2, :view)::Bool
│    %16 = (%15 isa DataFrames.Bool)::Core.Compiler.Const(true, false)
│          %16
└───       goto #7
6 ──       Core.Compiler.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :view, DataFrames.Bool, %15)), false)
└───       Core.Compiler.Const(:(Core.throw(%19)), false)
7 ┄─       (@_9 = %15)
└───       goto #9
8 ──       Core.Compiler.Const(:(@_9 = false), false)
9 ┄─       (view = @_9)
│    %25 = (:skipmissing, :view)::Core.Compiler.Const((:skipmissing, :view), false)
│    %26 = Core.apply_type(Core.NamedTuple, %25)::Core.Compiler.Const(NamedTuple{(:skipmissing, :view),T} where T<:Tuple, false)
│    %27 = Base.structdiff(@_2, %26)::Core.Compiler.Const(NamedTuple(), false)
│    %28 = Base.pairs(%27)::Core.Compiler.Const(Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}(), false)
│    %29 = Base.isempty(%28)::Core.Compiler.Const(true, false)
│          %29
└───       goto #11
10 ─       Core.Compiler.Const(:(Core.tuple(@_2, @_3, df)), false)
└───       Core.Compiler.Const(:(Core._apply_iterate(Base.iterate, Base.kwerr, %32, args)), false)
11 ┄ %34 = Core.tuple(skipmissing, view, @_3, df)::Core.Compiler.PartialStruct(Tuple{Bool,Bool,typeof(subset),DataFrame}, Any[Bool, Bool, Core.Compiler.Const(DataFrames.subset, false), DataFrame])
│    %35 = Core._apply_iterate(Base.iterate, DataFrames.:(var"#subset#406"), %34, args)::Any
└───       return %35

julia> f() = subset(DataFrame(a=1:10), :a => ByRow(<(4)), view=false, skipmissing=false)

julia> @code_warntype f()
Variables
  #self#::Core.Compiler.Const(f, false)

Body::Any
1 ─ %1  = (1:10)::Core.Compiler.Const(1:10, false)
│   %2  = (:a,)::Core.Compiler.Const((:a,), false)
│   %3  = Core.apply_type(Core.NamedTuple, %2)::Core.Compiler.Const(NamedTuple{(:a,),T} where T<:Tuple, false)
│   %4  = Core.tuple(%1)::Core.Compiler.Const((1:10,), false)
│   %5  = (%3)(%4)::Core.Compiler.Const((a = 1:10,), false)
│   %6  = Core.kwfunc(Main.DataFrame)::Core.Compiler.Const(Core.var"#Type##kw"(), false)
│   %7  = (%6)(%5, Main.DataFrame)::DataFrame
│   %8  = (<)(4)::Core.Compiler.Const(Base.Fix2{typeof(<),Int64}(<, 4), false)
│   %9  = Main.ByRow(%8)::Core.Compiler.Const(ByRow{Base.Fix2{typeof(<),Int64}}(Base.Fix2{typeof(<),Int64}(<, 4)), false)
│   %10 = (:a => %9)::Core.Compiler.Const(:a => ByRow{Base.Fix2{typeof(<),Int64}}(Base.Fix2{typeof(<),Int64}(<, 4)), false)
│   %11 = (:view, :skipmissing)::Core.Compiler.Const((:view, :skipmissing), false)
│   %12 = Core.apply_type(Core.NamedTuple, %11)::Core.Compiler.Const(NamedTuple{(:view, :skipmissing),T} where T<:Tuple, false)
│   %13 = Core.tuple(false, false)::Core.Compiler.Const((false, false), false)
│   %14 = (%12)(%13)::Core.Compiler.Const((view = false, skipmissing = false), false)
│   %15 = Core.kwfunc(Main.subset)::Core.Compiler.Const(DataFrames.var"#subset##kw"(), false)
│   %16 = (%15)(%14, Main.subset, %7, %10)::Any
└──       return %16
```
(not passing kwargs does not change this)

Maybe the reason is that we have a vararg function with `@nospecialize`?

bkamins:
I have applied the same fix in the other docstring.

bkamins:
> We could apply the grouping code directly to group indices

You mean to use `gdf.groups[row_selector]` and based on this create groups faster?
The code would be something like:

```
res.newuniquetempcol = gdf.groups[row_selector]
gdf2 = groupby(res, :newuniquetempcol)
copy!(gdf2.cols, gdf.cols)
select!(res, 1:ncol(res)-1)
return res
```

Do you think it would be worth it (I expect this code-path will not be used much).

nalimilan:
Yes, something like this. That's not super important, but it doesn't hurt (at least as a TODO).

bkamins:
I will add it as TODO, as I do not want to overcomplicate the code now.

src/abstractdataframe/where.jl
bkamins: removing `@nospecialize` did not change anything.

bkamins:
I left a TODO, as:
1. in some cases it will be slower than doing `groupby` on the original column (e.g. when grouping by `CategoricalArray` it is faster)
2. in some cases it will be the same
3. in some cases of course it will be faster

In order to do a fully fast implementation we should treat `gdf.groups[row_selector]` as categorical (which it is in fact), but this would complicate code more than what I proposed above (we would have to use a special grouping path as you proposed).

bkamins:
OK - so regarding this PR is there anything we should discuss or it is good to be merged?

nalimilan (approved):
It would be nice to fix the return type stability issue at some point, but this doesn't have to block the PR.

src/abstractdataframe/subset.jl
```diff
@@ -175,18 +181,20 @@ Each argument passed in `args` can be either a single column selector or a
...
-for which contition is false.
+columns (or all rows in groups for `GroupedDataFrame`).
 
 If `skipmissing=false` (the default) `args` are required to produce vectors
 containing only `Bool` values. If `skipmissing=true`, additionally `missing` is
 allowed and it is treated as `false` (i.e. rows for which one of the conditions
 returns `missing` are skipped).
 
+If `ungroup=false` the return value of the operation on `gdf` is re-grouped
+based on the same grouping columns and `GroupedDataFrame` is returned.
```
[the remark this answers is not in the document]
bkamins: This comment was already applied - I do not know why it shows up.

bkamins:
@pdeffebach, @matthieugomez - I will merge this PR later this week. Please comment if you want to add something to the discussion before merging.

pdeffebach:
This PR looks good to me. I have a few suggestions. 

1. Is it too late to rename `skipmissing` to `missingasfalse`? The second is more clear and doesn't conflict with `Base.skipmissing`
2. There is a section in the docs called "Taking a Subset" which makes no mention of this. Perhaps we should add a line in the docs? We also don't mention `filter` there. (Could be a separate PR)
3. Compile time is not great. I'm getting about .25 seconds for every call with an anonymous function, compared to about .15 seconds with an equivalent transform call. This can also be fixed in another PR, or more realistically a massive 1.X PR. 
4. Do we have an opinionated way to do "OR" operations? Particularly with broadcasting and avoiding `||` as that doesn't play well with `missing`.

bkamins:
These are good points. Thank you for raising them.


> Is it too late to rename `skipmissing` to `missingasfalse`?

I thought @matthieugomez wanted `skipmissing` exactly to reduce number of words to learn. Also `groupby` uses `skipmissing` to mean the same what we mean here (i.e. drop missings). I am not sure what is best here (naming is always super hard).

> There is a section in the docs called "Taking a Subset" which makes no mention of this.

I will add `filter` and `subset` in this PR to this section.

> Compile time is not great.

This is strange. I would think that it should be the same as the most compilation work should be with `select`. I will try to investigate (but as you know these things are hard; c.f. https://github.com/JuliaData/DataFrames.jl/issues/2566 which I will probably want to follow up before 1.0 release).

> Do we have an opinionated way to do "OR" operations?

Probably one would have to do it manually (which means it would be not very convenient). My thinking was that "OR" is needed much less frequently than "AND" (and if you need "OR" you probably want it on the same column, which can be expressed using an anonymous function).

bkamins:
I started updating the manual, but given the discussion we had today on Slack I gave up - it was too ad hoc. The manual needs to be updated. I have opened a separate PR for this: https://github.com/JuliaData/DataFrames.jl/issues/2595.

nalimilan:
IMO using `skipmissing` everywhere is indeed better for simplicity.

bkamins:
I have investigated the performance related to compilation and I could not improve things here (the compilation time is mostly `select` except for the first call where we compile also other methods):
```
~/Desktop/Dev/DF_dev$ julia --banner=no
 Activating environment at `~/Desktop/Dev/DF_dev/Project.toml`
julia> using DataFrames

julia> df = DataFrame(a=true)
1×1 DataFrame
 Row │ a    
     │ Bool 
─────┼──────
   1 │ true

julia> @time subset(df, :a => ByRow(x -> x))
  1.042957 seconds (2.91 M allocations: 149.359 MiB, 2.55% gc time)
1×1 DataFrame
 Row │ a    
     │ Bool 
─────┼──────
   1 │ true

julia> @time subset(df, :a => ByRow(x -> x))
  0.248947 seconds (603.93 k allocations: 30.169 MiB, 5.64% gc time)
1×1 DataFrame
 Row │ a    
     │ Bool 
─────┼──────
   1 │ true

julia> @time subset(df, :a => ByRow(x -> x))
  0.225462 seconds (603.94 k allocations: 30.172 MiB)
1×1 DataFrame
 Row │ a    
     │ Bool 
─────┼──────
   1 │ true


~/Desktop/Dev/DF_dev$ julia --banner=no
 Activating environment at `~/Desktop/Dev/DF_dev/Project.toml`
julia> using DataFrames

julia> df = DataFrame(a=true)
1×1 DataFrame
 Row │ a    
     │ Bool 
─────┼──────
   1 │ true

julia> @time select(df, :a => ByRow(x -> x))
  0.609927 seconds (1.77 M allocations: 90.192 MiB, 4.28% gc time)
1×1 DataFrame
 Row │ a_function 
     │ Bool       
─────┼────────────
   1 │       true

julia> @time select(df, :a => ByRow(x -> x))
  0.213393 seconds (563.55 k allocations: 27.732 MiB, 5.39% gc time)
1×1 DataFrame
 Row │ a_function 
     │ Bool       
─────┼────────────
   1 │       true

julia> @time select(df, :a => ByRow(x -> x))
  0.214110 seconds (563.56 k allocations: 27.735 MiB, 5.01% gc time)
1×1 DataFrame
 Row │ a_function 
     │ Bool       
─────┼────────────
   1 │       true
```

I will wait for a feedback and if there are no more comments I am going to merge this PR soon as is now.

pdeffebach (approved):

bkamins:
Thank you!

PR #2497, opened 2020-10-22
avoid allocation when negating BitArray

OkonSamuel:
avoid allocation when negating BitArray in `dropmissing!` method


bkamins (approved):

bkamins:
Looks good. Can you submit here some performance benchmarks for comparison before we merge? Thank you!

OkonSamuel:
> Looks good. Can you submit here some performance benchmarks for comparison before we merge? Thank you!

sure.

OkonSamuel:
@bkamins done

Issue #2502, opened 2020-10-29
innerjoin not working if one df is a SubDataFrame or item of GroupedDataFrame


lungben:
innerjoin is not working if one of the input dataframes is a DataFrame and the other one is a SubDataFrame. This error occurs both on the current release version and on master.

From the documentation / signature of the `innerjoin` function I think it should accept arbitrary AbstractDataFrames, correct?

MWE:

```julia
using Pkg
Pkg.activate(mktempdir())
# Pkg.add("DataFrames")
Pkg.add(name="DataFrames", rev="master")
using DataFrames
df1 = DataFrame(a=[1, 2, 3], b=[4, 5, 6])
df1_view = @view df1[df1.a .> 1, :]
df2 = DataFrame(a=[1, 2, 3], c=[7, 8, 9])
innerjoin(df1, df2, on=:a) # works correctly
innerjoin(df1_view, df2, on=:a) # does not work
```

Error message:
```
ERROR: MethodError: Cannot `convert` an object of type
  SubDataFrame{DataFrame{},DataFrames.SubIndex{DataFrames.Index,Array{Int64,1},Array{Int64,1}},Array{Int64{},1}} to an object of type
  SubDataFrame{DataFrame{},DataFrames.Index,Array{Int64{},1}}
Closest candidates are:
  convert(::Type{S}, ::T) where {S, T<:CategoricalValue} at C:\Users\d90394\.julia\packages\CategoricalArrays\0ZAbp\src\value.jl:68
  convert(::Type{T}, ::T) where T at essentials.jl:171
  SubDataFrame{DataFrame,DataFrames.Index,Array{Int64,1}}(::Any, ::Any, ::Any) where {D<:AbstractDataFrame, S<:DataFrames.AbstractIndex, T<:AbstractArray{Int64,1}} at C:\Users\d90394\.julia\packages\DataFrames\GtZ1l\src\subdataframe\subdataframe.jl:26
Stacktrace:
 [1] DataFrames.DataFrameJoiner{SubDataFrame{DataFrame,DataFrames.Index,Array{Int64,1}},DataFrame}(::SubDataFrame{DataFrame,DataFrames.Index,Array{Int64,1}}, ::DataFrame, ::Symbol) at C:\Users\xxx\.julia\packages\DataFrames\GtZ1l\src\abstractdataframe\join.jl:46
[...]

PR #2503, opened 2020-10-29
Remove type parameters in DataFrameJoiner

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2502

The problem is that `SubDataFrame` is `UnionAll`. Now I propose to have a container of abstract types, but I think it is OK, as it will save some compilation time + data frame is type unstable anyway.

### DIFF
--- a/test/join.jl
+++ b/test/join.jl
@@ -887,7 +887,7 @@ end
     @test_throws ArgumentError join(df1, df2, on=:id, kind=:inner)
 end
 
-@testset "Make DataFrameJoiner rely on AbstractDataFrame" begin
+@testset "join mixing DataFrame and SubDataFrame" begin
     df1 = DataFrame(a=[1, 2, 3], b=[4, 5, 6])
     df1_copy = df1[df1.a .> 1, :]
     df1_view1 = @view df1[df1.a .> 1, :]
@@ -896,7 +896,6 @@ end
     @test innerjoin(df1_copy, df2, on=:a) ==
           innerjoin(df1_view1, df2, on=:a) ==
           innerjoin(df1_view2, df2, on=:a)
-
 end
 
 end # module
--- a/test/join.jl
+++ b/test/join.jl
@@ -887,4 +887,16 @@ end
     @test_throws ArgumentError join(df1, df2, on=:id, kind=:inner)
 end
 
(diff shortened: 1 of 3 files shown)


test/join.jl
```diff
@@ -887,4 +887,16 @@ end
     @test_throws ArgumentError join(df1, df2, on=:id, kind=:inner)
 end
 
+@testset "Make DataFrameJoiner rely on AbstractDataFrame" begin
```
nalimilan, line 890:
```suggestion
@testset "join mixing DataFrame and SubDataFrame" begin
```
(Would it make sense to test other joins?)

test/join.jl
```diff
@@ -887,4 +887,16 @@ end
...
+    df1 = DataFrame(a=[1, 2, 3], b=[4, 5, 6])
+    df1_copy = df1[df1.a .> 1, :]
+    df1_view1 = @view df1[df1.a .> 1, :]
+    df1_view2 = @view df1[df1.a .> 1, 1:2]
+    df2 = DataFrame(a=[1, 2, 3], c=[7, 8, 9])
+    @test innerjoin(df1_copy, df2, on=:a) ==
+          innerjoin(df1_view1, df2, on=:a) ==
+          innerjoin(df1_view2, df2, on=:a)
+
```
nalimilan, line 899:
```suggestion
```

nalimilan (approved):
If that doesn't affect performance in critical places I guess that's OK.

test/join.jl
bkamins: no - it is OK to test only this

bkamins:
1. Actually before this PR `SubDataFrame` was not supported at all (see the last commit with `findrow`).
2. The timing is improved with this PR.

Here is the result of calling `@time include("DataFrames/test/join.jl")` twice in a fresh session.

Before this PR:
```
63.439274 seconds (112.52 M allocations: 5.469 GiB, 2.52% gc time)
22.607366 seconds (31.93 M allocations: 1.479 GiB, 1.58% gc time)
```

After this PR (we do one more test set here additionally):
```
52.542405 seconds (102.13 M allocations: 5.092 GiB, 2.78% gc time)
17.194671 seconds (21.85 M allocations: 1.080 GiB, 1.50% gc time)
```

Issue #2499, opened 2020-10-24
Explicitly handling missingness in join columns


adkabo:
Julia has a great philosophy of [taking missingness seriously](https://julialang.org/blog/2018/06/missing/#the_missing_object_safe_and_generic_missing_values). For example, unlike in Pandas or Postgres, `sum([1,2,missing])` gives `missing`. However, this philosophy hasn't yet been applied to all of the functions in the JuliaData ecosystem. I'll give an example to illustrate.

My goal is to find the _relationship between age and salary_. To find out, I will combine observations from two datasets.

One with age,

```
DataFrame([(name="John", employer=missing, age=40)])
```
and one with salary.
```
DataFrame([(name="John", employer="Julia Computing", salary=9999)])
```

In a complete-data environment, if these observations correspond to the same person, I want one row in the joined dataframe; on the other hand, if they correspond to different people, I want zero rows in the join.

Given the missingness in the "employer" column, we don't know if that's the same person or not.  So when I join them on `("name", "employer")`, we _cannot know the right answer_. Yet

```
innerjoin(
    DataFrame([(name="John", employer=missing, age=40)]),
    DataFrame([(name="John", employer="Julia Computing", salary=9999)]), 
    on=[:name, :employer]
)
```

makes a decision implicitly, returning an empty dataframe. If the observations correspond to the same person, this result -- failing to match the two observations -- is a _false negative_. 

To avoid drawing mistaken conclusions from analysis, I would like to extend the practice of enforcing explicit handling of missing values. So I would like to get an error message in this case by default, and to actively tell `innerjoin()` how it should handle missing values. 

The `passmissing()` and `skipmissing()` patterns used elsewhere in JuliaData are a great reassurance that Julia is looking out for missing data problems. When applied to joins, I would like to consider: 

- In an outer join, should rows with missingness all be dropped from the output or all be kept in? 
- In an inner join, should they be considered matches or nonmatches? 
- What if there is missingness on both sides or one side only? 

I'm not sure if something like `passmissing(innerjoin)(a,b)` would work, or if it should be more like `innerjoin(a,b, missingrule=:drop)` or something else. But I do want to start the conversation about it.

---
https://github.com/JuliaData/DataFrames.jl/issues/2243 has some discussion of missingness and joins, mainly focused on the `a.fillna(b)` use case.

bkamins:
[...]

PR #2504, opened 2020-10-29
[BREAKING] add matchmissing kwarg to joins

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2499.

This is the core implementation and update of docstrings.

I will add tests later, as it first requires https://github.com/JuliaData/DataFrames.jl/pull/2503 to be merged and rebased against.

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -55,6 +55,13 @@ struct DataFrameJoiner
         elseif matchmissing !== :equal
             throw(ArgumentError("matchmissing allows only :error or :equal"))
         end
+
+        for df in (dfl_on, dfr_on), col in eachcol(df)
+            if any(x -> (x isa Union{Complex, Real}) & isnan(x), col)
+                throw(ArgumentError("NaN values in key columns are not allowed"))
+            end
+        end
+
         new(dfl, dfr, dfl_on, dfr_on, left_on, right_on)
     end
 end
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -57,8 +57,13 @@ struct DataFrameJoiner
         end
 
         for df in (dfl_on, dfr_on), col in eachcol(df)
-            if any(x -> (x isa Union{Complex, Real}) && isnan(x), col)
-                throw(ArgumentError("NaN values in key columns are not allowed"))
+            if any(x -> (x isa Union{Complex, Real}) &&
+                        (isnan(x) || real(x) == -0.0 || imag(x) == -0.0), col)
+                throw(ArgumentError("currently for numeric values NaN and `-0.0` " *
+                                    "in their real or imaginary components are not" *
+                                    " allowed. Use CategoricalArrays.jl to wrap" *
+                                    "these values into CategoricalVector to perform" *
+                                    "the requested join."))
             end
         end
 
@@ -507,6 +512,11 @@ The order of rows in the result is undefined and may change in the future releas
 - `matchmissing` : if equal to `:error` throw an error if `missing` is present
   in `on` columns; if equal to `:equal` then `missing` is allowed and missings are
   matched (`isequal` is used for comparisons of rows for equality)
+
+It is not allowed to join on columns that contain `NaN` or `-0.0` in real or
+imaginary part of the number. If you need to perform a join on such values use
+CategoricalArrays.jl and transform a column containing such values into a
+`CategoricalVector`.
 
 When merging `on` categorical columns that differ in the ordering of their
 levels, the ordering of the left data frame takes precedence over the ordering
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -19,7 +19,8 @@ struct DataFrameJoiner
     right_on::Vector{Symbol}
 
     function DataFrameJoiner(dfl::AbstractDataFrame, dfr::AbstractDataFrame,
-                             on::Union{<:OnType, AbstractVector})
+                             on::Union{<:OnType, AbstractVector},
+                             matchmissing::Symbol)
         on_cols = isa(on, AbstractVector) ? on : [on]
         left_on = Symbol[]
         right_on = Symbol[]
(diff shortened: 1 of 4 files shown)


nalimilan (commented):
Looks good. Though I think we should do something about `NaN` (which dplyr also handles via the equivalent of `matchmissing`). Maybe just throw an error if there are `NaN`s for now and see whether people complain.

bkamins:
> Maybe just throw an error if there are `NaN`s for now and see whether people complain.

OK. Doing joins on floats is in general not advisable though:
```
julia> outerjoin(DataFrame(x=0.0,a=1), DataFrame(x=-0.0,b=2), on=:x)
2×3 DataFrame
│ Row │ x       │ a       │ b       │
│     │ Float64 │ Int64?  │ Int64?  │
│ 1   │ 0.0     │ 1       │ missing │
│ 2   │ -0.0    │ missing │ 2       │

```
(which means in particular that I am OK to error on `NaN`)

nalimilan:
Yeah `-0.0` is also a tricky case. We could also throw an error if we find `-0.0`.

bkamins:
@StefanKarpinski - do you have an opinion how `NaN` and `-0.0` should be handled in joins when they are present in columns you join on? (currently we accept them and use `isequal` to match - just like `Dict`s)

As you can see @nalimilan considers if we should not just error on them. In general I understand his point, but I am not 100% sure what would be best. Given your experiences with designing `isequal` in Base what do you think would be best?

kescobo:
> Doing joins on floats is in general not advisable though

:100: If people are doing joins on floats, something has already gone seriously wrong. But kudos for trying to get it right. How are other `==` but not `===` things handled? Eg.

```
a = [1,2]
b = [1,2]
c = [1,3]

df1 = DataFrame(x = [a,b,c], y = rand(3))
df2 = DataFrame(x = [a], z = [rand()])
```
```
julia> leftjoin(df1,df2, on=:x)
3×3 DataFrame
│ Row │ x      │ y        │ z        │
│     │ Array… │ Float64  │ Float64? │
├─────┼────────┼──────────┼──────────┤
│ 1   │ [1, 2] │ 0.213345 │ 0.614537 │
│ 2   │ [1, 2] │ 0.70363  │ 0.614537 │
│ 3   │ [1, 3] │ 0.181217 │ missing  │
```

So for consistency, seems like `-0.0` and `0.0` should be treated the same, no?

kescobo:
Oh, but the original question was about `NaN` - that one is weird since it's `===` but not `==`... That I have no idea aboutz

tbeason:
I think `NaN` could more or less follow whatever decision was made about `missing`, right? Why treat them differently with regards to matching?

I am indifferent about `0.0` and `-0.0` being special cased.

bkamins:
I have thought about it and I think I would leave floating point columns as they are now. One normally should not join on such columns anyway, as it is error prone in general. Instead I propose to add a comment in the docstrings warning about doing joins on floating point columns.

nalimilan:
Since we can't start throwing an error without breaking, I'd rather throw an error with `NaN`s (current state of PR), and see later whether we should add an argument to disable it (or even disable it altogether). It would be safer to do that also for `-0.0`.

src/abstractdataframe/join.jl
```diff
@@ -41,7 +42,27 @@ struct DataFrameJoiner
...
+                    throw(ArgumentError("missing values in key columns are not allowed " *
+                                        "when matchmissing == :error"))
+                end
+            end
+        elseif matchmissing !== :equal
+            throw(ArgumentError("matchmissing allows only :error or :equal"))
+        end
+
+        for df in (dfl_on, dfr_on), col in eachcol(df)
+            if any(x -> (x isa Union{Complex, Real}) & isnan(x), col)
```
nalimilan, line 60: Short-circuiting is needed AFAICT:
```suggestion
            if any(x -> (x isa Union{Complex, Real}) && isnan(x), col)
```

bkamins: indeed

bkamins:
OK - I will disallow `-0.0` also then and throw an informative error message.

bkamins:
I have updated the implementation and added appropriate tests. I think that the solution I propose (to use `CategoricalVector` as a wrapper) is clean enough (ie. if someone really wants to make such a join there is a way to do it).

The PR should be ready for a review.

src/abstractdataframe/join.jl
```diff
@@ -57,8 +57,13 @@ struct DataFrameJoiner
...
         for df in (dfl_on, dfr_on), col in eachcol(df)
-            if any(x -> (x isa Union{Complex, Real}) & isnan(x), col)
-                throw(ArgumentError("NaN values in key columns are not allowed"))
+            if any(x -> (x isa Union{Complex, Real}) &&
+                        (isnan(x) || real(x) === -0.0 || imag(x) === -0.0), col)
+                throw(ArgumentError("currently for numeric values NaN and `-0.0` " *
+                                    "in their real or imaginary components are not" *
+                                    " allowed. Use CategoricalArrays.jl to wrap" *
+                                    "these values into CategoricalVector to perform" *
+                                    "the requested join."))
```
nalimilan, line 66:
```suggestion
                                    "in their real or imaginary components are not " *
                                    "allowed. Use CategoricalArrays.jl to wrap " *
                                    "these values in a CategoricalVector to perform " *
                                    "the requested join."))
```

src/abstractdataframe/join.jl
```diff
@@ -57,8 +57,13 @@ struct DataFrameJoiner
         end
 
         for df in (dfl_on, dfr_on), col in eachcol(df)
-            if any(x -> (x isa Union{Complex, Real}) & isnan(x), col)
-                throw(ArgumentError("NaN values in key columns are not allowed"))
+            if any(x -> (x isa Union{Complex, Real}) &&
+                        (isnan(x) || real(x) === -0.0 || imag(x) === -0.0), col)
```
nalimilan, line 61:
```suggestion
                        (isnan(x) || isequal(real(x), -0.0) || isequal(imag(x), -0.0)), col)
```

test/join.jl
```diff
@@ -573,25 +573,57 @@ end
...
-        @test innerjoin(name_w_special, job, on=:ID, validate=(true, false)) == inner
+        @test_throws ArgumentError innerjoin(name_w_special, job, on=:ID)
+        @test_throws ArgumentError leftjoin(name_w_special, job, on=:ID)
+        @test_throws ArgumentError rightjoin(name_w_special, job, on=:ID)
+        @test_throws ArgumentError outerjoin(name_w_special, job, on=:ID)
+        @test_throws ArgumentError semijoin(name_w_special, job, on=:ID)
+        @test_throws ArgumentError antijoin(name_w_special, job, on=:ID)
+    end
+
+    for special in [missing, 0.0]
```
nalimilan, line 587: `0.0` works even without `matchmissing=:equal`, right?

src/abstractdataframe/join.jl
bkamins: ah - of course. I was over influenced by your comment not to use `isequal`, but here it is the opposite and it has to be used 😄.

test/join.jl
bkamins: right, but I wanted to avoid adding the second set of the same tests. I think it is OK, as `:equal` has no influence on `0.0` handling

nalimilan (approved):

bkamins:
Thank you!

PR #2505, opened 2020-10-31
[BREAKING] remove CategoricalArrays dependency from joins

bkamins:
I have found a dependency on CategoricalArrays in joins also.

### DIFF
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -122,9 +122,9 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
         # assign 0x1 to it anyway and these rows are guaranteed to come first
         # (even if they are permuted)
         left_indicator = zeros(UInt32, nrow)
-        left_indicator[axes(all_orig_left_ixs, 1)] .= UInt32(1)
+        left_indicator[axes(all_orig_left_ixs, 1)] .= 1
         right_indicator = zeros(UInt32, nrow)
-        right_indicator[axes(all_orig_right_ixs, 1)] .= UInt32(2)
+        right_indicator[axes(all_orig_right_ixs, 1)] .= 2
         permute!(right_indicator, right_perm)
     end
 
@@ -418,9 +418,9 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
     if indicator !== nothing
         left_indicator .+= right_indicator
         pool = ["left_only", "right_only", "both"]
-        invpool = Dict{String, UInt8}("left_only" => UInt32(1),
-                                      "right_only" => UInt32(2),
-                                      "both" => UInt32(3))
+        invpool = Dict{String, UInt32}("left_only" => 1,
+                                       "right_only" => 2,
+                                       "both" => 3)
         indicatorcol = PooledArray(PooledArrays.RefArray(left_indicator),
                                    invpool, pool)
 
--- a/src/abstractdataframe/join.jl
+++ b/src/abstractdataframe/join.jl
@@ -416,10 +416,11 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
     end
 
     if indicator !== nothing
-        refs = left_indicator + right_indicator
-        refs_short, invpool, pool = PooledArrays._label(["left_only", "right_only", "both"], String)
-        @assert refs_short == [1, 2, 3]
-        indicatorcol = PooledArray(PooledArrays.RefArray(refs), invpool, pool)
+        left_indicator .+= right_indicator
+        pa_base = PooledArray(["left_only", "right_only", "both"])
+        indicatorcol = PooledArray(PooledArrays.RefArray(left_indicator),
+                                   Dict{String, UInt8}("left_only" => 0x1, "right_only" => 0x2, "both" => 0x3),
+                                   ["left_only", "right_only", "both"])
 
         unique_indicator = indicator
         if makeunique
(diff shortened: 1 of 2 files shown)


src/abstractdataframe/join.jl
```diff
@@ -417,8 +417,9 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
 
     if indicator !== nothing
         refs = left_indicator + right_indicator
-        pool = CategoricalPool{String,UInt8}(["left_only", "right_only", "both"])
-        indicatorcol = CategoricalArray{String,1}(refs, pool)
+        refs_short, invpool, pool = PooledArrays._label(["left_only", "right_only", "both"], String)
```
nalimilan, line 420: How about creating the dicts manually to avoid relying on an internal function?

bkamins: I was considering this. What I do checks if what PooledArrays.jl produces and uses is consistent (i.e. if PooledArrays.jl changes its internal design we will get an error here and this is good; if we create dicts manually and PooledArrays.jl is changed we might silently start producing wrong results).

What do you think we should do?

nalimilan: AFAICT passing dicts and refs is part of the public API, so we're safe.

bkamins: OK - I will rewrite it.

bkamins: done

src/abstractdataframe/join.jl
```diff
@@ -416,9 +416,11 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
     end
 
     if indicator !== nothing
-        refs = left_indicator + right_indicator
-        pool = CategoricalPool{String,UInt8}(["left_only", "right_only", "both"])
-        indicatorcol = CategoricalArray{String,1}(refs, pool)
+        left_indicator .+= right_indicator
+        pa_base = PooledArray(["left_only", "right_only", "both"])
```
nalimilan, line 420: Not used?
```suggestion
```

src/abstractdataframe/join.jl
```diff
@@ -416,9 +416,11 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
+        indicatorcol = PooledArray(PooledArrays.RefArray(left_indicator),
+                                   Dict{String, UInt8}("left_only" => 0x1, "right_only" => 0x2, "both" => 0x3),
```
nalimilan, line 422: BTW, I've just checked an CSV uses `UInt32`:
https://github.com/JuliaData/CSV.jl/blob/2e8a8f38aba186161ff4eeaeada9756457f010ea/src/file.jl#L424

It could make sense to use that even if that wastes space since it will reduce latency. Though not a big deal.

nalimilan (approved):

src/abstractdataframe/join.jl
bkamins: right - sorry for forgetting to delete it.

src/abstractdataframe/join.jl
bkamins: I thought about it, and decided that having smaller footprint is more important. But OK, I will change it.

src/abstractdataframe/join.jl
```diff
@@ -121,10 +121,10 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol,
...
         # assign 0x1 to it anyway and these rows are guaranteed to come first
         # (even if they are permuted)
-        left_indicator = zeros(UInt8, nrow)
-        left_indicator[axes(all_orig_left_ixs, 1)] .= 0x1
-        right_indicator = zeros(UInt8, nrow)
-        right_indicator[axes(all_orig_right_ixs, 1)] .= 0x2
+        left_indicator = zeros(UInt32, nrow)
+        left_indicator[axes(all_orig_left_ixs, 1)] .= UInt32(1)
+        right_indicator = zeros(UInt32, nrow)
+        right_indicator[axes(all_orig_right_ixs, 1)] .= UInt32(2)
```
nalimilan, line 127:
```suggestion
        left_indicator[axes(all_orig_left_ixs, 1)] .= 1
        right_indicator = zeros(UInt32, nrow)
        right_indicator[axes(all_orig_right_ixs, 1)] .= 2
```

src/abstractdataframe/join.jl
```diff
@@ -416,9 +416,13 @@ function _join(df1::AbstractDataFrame, df2::AbstractDataFrame;
...
 
     if indicator !== nothing
-        refs = left_indicator + right_indicator
-        pool = CategoricalPool{String,UInt8}(["left_only", "right_only", "both"])
-        indicatorcol = CategoricalArray{String,1}(refs, pool)
+        left_indicator .+= right_indicator
+        pool = ["left_only", "right_only", "both"]
+        invpool = Dict{String, UInt8}("left_only" => UInt32(1),
+                                      "right_only" => UInt32(2),
+                                      "both" => UInt32(3))
```
nalimilan, line 423:
```suggestion
        invpool = Dict{String, UInt32}("left_only" => 1,
                                       "right_only" => 2,
                                       "both" => 3)
```

src/abstractdataframe/join.jl
bkamins: Why change this?

`UInt32(1)` will get optimized during compilation and I thought it is more explicit.

src/abstractdataframe/join.jl
bkamins: same here

src/abstractdataframe/join.jl
nalimilan: Does it really make a difference? I would guess LLVM will optimize it away.

src/abstractdataframe/join.jl
nalimilan: Likewise. IMHO the presence of the type parameter is explicit enough.

bkamins:
OK - changed.

bkamins:
Thank you!

Issue #2506, opened 2020-10-31
remaining dependencies on CategoricalArrays


bkamins:
Apart from open PRs we have the following remaining dependencies on CategoricalArrays.jl:
1. https://github.com/JuliaData/DataFrames.jl/blob/master/src/groupeddataframe/splitapplycombine.jl#L963
2. https://github.com/JuliaData/DataFrames.jl/blob/master/src/groupeddataframe/splitapplycombine.jl#L1040
3. https://github.com/JuliaData/DataFrames.jl/blob/master/src/dataframerow/utils.jl#L38

This should be worked on after https://github.com/JuliaData/DataFrames.jl/pull/2481 is merged.

@nalimilan - it can be done post 0.22 I think but before 1.0 (as this seems only optimization related). But could you please check (you know this code better), as maybe removing the dependency will have consequences like changing type or order of resulting computations (as we had in several cases recently). In general - it would be great if you could make an appropriate PR as you know this code much better than me. Thank you!

nalimilan:
Yes since it's not breaking it doesn't need to be in 0.22.

For the first ones, and ugly hack checking the name of the type could work. For the latter we can just use `DataAPI.refpool`.

bkamins:
I have looked into it and my conclusion is:

For
```
function hashrows_col!(h::Vector{UInt},
                       n::Vector{Bool},
                       v::AbstractCategoricalVector,
                       firstcol::Bool)
```
it is problematic to make it generic because `CategoricalVector` encodes `missing` as `0` as opposed to `PooledVector`.
Actually I think it would be best to add `hasharray(array, h=0)` function to DataAPI.jl that would delegate to the packages doing fast hashing of all their elements. What do you think?

---

for fast aggregates I was not fully sure how to do it cleanly (we need to assure pool sharing, but calling `outcol = CategoricalArray{U, 1}(outcol.refs, incol.pool)` seems to require having CategoricalArrays.jl as a dependency)

nalimilan:
AFAICT we don't really need to know where `missing` is: just do `hashes = hash.(refpool(v))`. I think I wrote the code in its current form because `refpool` didn't exist until recently.

For aggregates it's harder. Need to think about it.

bkamins:
The point is that what you propose does not work:
```
julia> x = categorical([1,2,missing])
3-element CategoricalArray{Union{Missing, Int64},1,UInt32}:
 1
 2
 missing

[...]

PR #2518, opened 2020-11-06
make hashrows_col! not depend on CategoricalArrays.jl

bkamins:
@nalimilan - this follows your suggestion in https://github.com/JuliaData/DataFrames.jl/issues/2506#issuecomment-723066223.

It is not fully in line with DataAPI.jl API (but I propose - as already mentioned to make that API stricter and require `DataAPI.refpool` to be `AbstractVector`).

If we agree on the proposal I will add more tests.

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -51,19 +51,8 @@ function hashrows_col_pool!(h::Vector{UInt},
         end
 
         fi = firstindex(rp)
-        if fi == 1
-            @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
-                h[i] = hashes[ref]
-            end
-        elseif fi == 0
-            @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
-                h[i] = hashes[ref+1]
-            end
-        else
-            # currently no implementation of DataAPI.jl interface hits this branch
-            @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
-                h[i] = hashes[ref+1-fi]
-            end
+        @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
+            h[i] = hashes[ref+1-fi]
         end
     else
         @inbounds for (i, x) in enumerate(v)
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -43,8 +43,8 @@ function hashrows_col!(h::Vector{UInt},
                        firstcol::Bool)
     # When hashing the first column, no need to take into account previous hash,
     # which is always zero
-    # also when there are more than 90% of refs in the pool than the length of the
-    # vector avoid using this path. 90% is picked heuristically
+    # also when the number of values in the pool is more than half the length
+    # of the vector avoid using this path. 50% is roughly based on benchmarks
     if firstcol && 2 * length(rp) < length(v)
         hashes = Vector{UInt}(undef, length(rp))
         @inbounds for (i, v) in zip(eachindex(hashes), rp)
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -23,6 +23,7 @@ end
 function hashrows_col!(h::Vector{UInt},
                        n::Vector{Bool},
                        v::AbstractVector{T},
+                       rp::Nothing,
                        firstcol::Bool) where T
     @inbounds for i in eachindex(h)
         el = v[i]
@@ -35,11 +36,11 @@ function hashrows_col!(h::Vector{UInt},
 end
 
 # should give the same hash as AbstractVector{T}
-function hashrows_col_pool!(h::Vector{UInt},
-                            n::Vector{Bool},
-                            v::AbstractVector,
-                            rp::AbstractVector, # this condition is currently met in all implementations, but is not part of the API
-                            firstcol::Bool)
+function hashrows_col!(h::Vector{UInt},
+                       n::Vector{Bool},
+                       v::AbstractVector,
+                       rp::Any,
+                       firstcol::Bool)
     # When hashing the first column, no need to take into account previous hash,
     # which is always zero
     # also when there are more than 90% of refs in the pool than the length of the
@@ -51,6 +52,8 @@ function hashrows_col_pool!(h::Vector{UInt},
         end
 
         fi = firstindex(rp)
+        # here we rely on the fact that `DataAPI.refpool` supports a continuous
+        # block of indices
         @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
             h[i] = hashes[ref+1-fi]
         end
@@ -72,12 +75,8 @@ function hashrows(cols::Tuple{Vararg{AbstractVector}}, skipmissing::Bool)
     rhashes = zeros(UInt, len)
     missings = fill(false, skipmissing ? len : 0)
     for (i, col) in enumerate(cols)
-        rp = DataAPI.refpool(x)
-        if rp === nothing
-            hashrows_col!(rhashes, missings, col, i == 1)
-        else
-            hashrows_col_pool!(rhashes, missings, col, rp, i == 1)
-        end
+        rp = DataAPI.refpool(col)
+        hashrows_col!(rhashes, missings, col, rp, i == 1)
     end
     return (rhashes, missings)
 end
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -35,19 +35,35 @@ function hashrows_col!(h::Vector{UInt},
 end
 
 # should give the same hash as AbstractVector{T}
-function hashrows_col!(h::Vector{UInt},
-                       n::Vector{Bool},
-                       v::AbstractCategoricalVector,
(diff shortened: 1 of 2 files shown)


src/dataframerow/utils.jl
```diff
@@ -67,7 +83,12 @@ function hashrows(cols::Tuple{Vararg{AbstractVector}}, skipmissing::Bool)
     rhashes = zeros(UInt, len)
     missings = fill(false, skipmissing ? len : 0)
     for (i, col) in enumerate(cols)
-        hashrows_col!(rhashes, missings, col, i == 1)
+        rp = DataAPI.refpool(x)
+        if rp === nothing
+            hashrows_col!(rhashes, missings, col, i == 1)
+        else
+            hashrows_col_pool!(rhashes, missings, col, rp, i == 1)
+        end
```
nalimilan, line 91: I'd typically rely on dispatch for that: `hashrows_col_pool!` can be a method of `hashrows_col!` when the pool is `AbstractVector` (or even `Any` given my comment above, with `Nothing` being caught by the fallback).

src/dataframerow/utils.jl
```diff
@@ -35,19 +35,35 @@ function hashrows_col!(h::Vector{UInt},
...
 # should give the same hash as AbstractVector{T}
-function hashrows_col!(h::Vector{UInt},
-                       n::Vector{Bool},
-                       v::AbstractCategoricalVector,
-                       firstcol::Bool)
-    levs = levels(v)
+function hashrows_col_pool!(h::Vector{UInt},
+                            n::Vector{Bool},
+                            v::AbstractVector,
+                            rp::AbstractVector, # this condition is currently met in all implementations, but is not part of the API
```
nalimilan, line 41: Do we really need that requirement? AFAICT we only need `length` and iteration, which are now required by the API.

src/dataframerow/utils.jl
```diff
@@ -35,19 +35,35 @@ function hashrows_col!(h::Vector{UInt},
...
+        elseif fi == 0
+            @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
+                h[i] = hashes[ref+1]
+            end
+        else
+            # currently no implementation of DataAPI.jl interface hits this branch
+            @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
+                h[i] = hashes[ref+1-fi]
+            end
         end
```
nalimilan, line 67: AFAIK the compiler is smart enough to statically resolve `fi`:
```suggestion
        @inbounds for (i, ref) in enumerate(DataAPI.refarray(v))
            h[i] = hashes[ref+1-fi]
        end
```

bkamins: this is something I was not sure.

src/dataframerow/utils.jl
bkamins: we need `firstindex` and that indexing is continuous starting with `firstindex` and ending with `lastindex`.

src/dataframerow/utils.jl
bkamins: ok

src/dataframerow/utils.jl
nalimilan: Right, the continuity requirement isn't documented in the docstring right now. That's a reasonable assumption to make, but we could add it for clarity. (It doesn't make a lot of sense to have non contiguous indices when you are required to implement `length`, `eachindex`, `firstindex` and `lastindex`. Basically we want array- or tuple-like objects.)

bkamins: Agreed. I have forgotten that we have added these requirements.

bkamins:
Here are the benchmarks of the change:

```
using DataFrames, PooledArrays, CategoricalArrays, DataAPI

x = [1:9000000; fill(1, 1000001)];
x1 = PooledArray(x);
x2 = categorical(x);
y = [1:9000002; fill(1, 1000000-1)];
y1 = PooledArray(y);
y2 = categorical(y);
length(DataAPI.refpool(x1))/length(x)
length(DataAPI.refpool(y1))/length(y)
for (i,n) in enumerate((x, x1, x2, y, y1, y2))
    @info i
    GC.gc()
    @time DataFrames.hashrows((n,), false)
end


julia> using DataFrames, PooledArrays, CategoricalArrays, DataAPI

julia> x = [1:9000000; fill(1, 1000001)];

julia> x1 = PooledArray(x);
^[[C
julia> x2 = categorical(x);

julia> y = [1:9000002; fill(1, 1000000-1)];

julia> y1 = PooledArray(y);

julia> y2 = categorical(y);

julia> length(DataAPI.refpool(x1))/length(x)
0.899999910000009

julia> length(DataAPI.refpool(y1))/length(y)
0.900000109999989

julia> for (i,n) in enumerate((x, x1, x2, y, y1, y2))
           @info i
           GC.gc()
           @time DataFrames.hashrows((n,), false)
       end
[ Info: 1
  0.051783 seconds (5 allocations: 76.294 MiB, 5.32% gc time)
[ Info: 2
  0.101282 seconds (7 allocations: 144.959 MiB, 2.91% gc time)
[ Info: 3
  0.158293 seconds (7 allocations: 144.959 MiB, 2.19% gc time)
[ Info: 4
  0.061556 seconds (5 allocations: 76.294 MiB, 2.48% gc time)
[ Info: 5
  0.087488 seconds (5 allocations: 76.294 MiB, 1.94% gc time)
[ Info: 6
  0.112878 seconds (5 allocations: 76.294 MiB, 1.38% gc time)

julia> for (i,n) in enumerate((x, x1, x2, y, y1, y2))
           @info i
           GC.gc()
           @time DataFrames.hashrows((n,), false)
       end
[ Info: 1
  0.055569 seconds (5 allocations: 76.294 MiB, 2.99% gc time)
[ Info: 2
  0.112733 seconds (7 allocations: 144.959 MiB, 2.90% gc time)
[ Info: 3
  0.166712 seconds (7 allocations: 144.959 MiB, 2.05% gc time)
[ Info: 4
  0.062511 seconds (5 allocations: 76.294 MiB, 2.40% gc time)
[ Info: 5
  0.081028 seconds (5 allocations: 76.294 MiB, 1.94% gc time)
[ Info: 6
  0.117411 seconds (5 allocations: 76.294 MiB, 1.64% gc time)
```

@nalimilan - So it seems that the 90% threshold is OK (probably it could be even a bit lower, but it is hard to tune it optimally).
Also - we can see that when there are so many levels it is better not to do pooling.

@quinnj - do you still disable creation of a `PooledArray` in CSV.jl if there are too many levels in a categorical column or not?

I will add a test to make sure that all these cases produce the same hashes.

nalimilan:
Thanks for benchmarking! So as I suspected (I was going to comment) the pooled hashing is a bit slower at 90%. I think I'd go with a lower threshold, e.g. 50% or even 10%.

bkamins:
> So as I suspected

I also suspected this, but the previous code did not use this optimization. 10% is too low (copying data is faster than calculating hashes). I will change it to 50% then (it will not be optimal if we hash very long strings though - in that case something closer to 90% is better, but this is probably rare).

src/dataframerow/utils.jl
```diff
@@ -37,17 +38,24 @@ end
...
     # which is always zero
-    if firstcol
-        hashes = Vector{UInt}(undef, length(levs)+1)
-        hashes[1] = hash(missing)
-        hashes[2:end] .= hash.(levs)
-        @inbounds for (i, ref) in enumerate(v.refs)
-            h[i] = hashes[ref+1]
+    # also when there are more than 90% of refs in the pool than the length of the
+    # vector avoid using this path. 90% is picked heuristically
+    if firstcol && length(rp) < 0.9length(v)
```
bkamins, line 48:
```suggestion
    if firstcol && 2 * length(rp) < length(v)
```

src/dataframerow/utils.jl
```diff
@@ -37,17 +38,24 @@ end
...
     # When hashing the first column, no need to take into account previous hash,
     # which is always zero
-    if firstcol
-        hashes = Vector{UInt}(undef, length(levs)+1)
-        hashes[1] = hash(missing)
-        hashes[2:end] .= hash.(levs)
-        @inbounds for (i, ref) in enumerate(v.refs)
-            h[i] = hashes[ref+1]
+    # also when there are more than 90% of refs in the pool than the length of the
+    # vector avoid using this path. 90% is picked heuristically
```
nalimilan, line 47:
```suggestion
    # also when the number of values in the pool is more than half the length
    # of the vector avoid using this path. 50% is roughly based on benchmarks
```

test/grouping.jl
```diff
@@ -3173,4 +3173,19 @@ end
                     :min => min.(df.y, df.z), :max => max.(df.y, df.z), :y => df.y) |> sort
 end
 
+@testset "hashing of pooled vectors" begin
+    # test both hashrow calculation paths - the of pool length thereshold is 50%
```
nalimilan, line 3177:
```suggestion
    # test both hashrow calculation paths - the of pool length threshold is 50%
```

nalimilan (approved):

bkamins:
Thank you!

PR #2522, opened 2020-11-07
update manual and docstrings to PrettyTables.jl

bkamins:
I have done the manual. I will continue with docstrings tomorrow.


bkamins:
This should be good for a review. In general you can expect that this PR does not change anything. What I do:
1. update manual
2. update all docstrings (and fix some of them that were not working)
3. remove methods that were used for printing but are not used any more (@ronisbr - here I would appreciate if you confirmed this) and remove them from internals.md
4. do the same changes as https://github.com/JuliaData/DataFrames.jl/pull/2523 (just to make sure CI passes)
5. small update of groupby tests as now we have trailing whitespace in some printed strings

In general I have used this PR to make sure all doctests pass before the release.

docs/src/man/getting_started.md
```diff
@@ -891,45 +980,45 @@ julia> x = [3, 1, 2];
...
-│ 2   │ 2     │
-│ 3   │ 3     │
+ Row │ x
+     │ Int64
+─────┼───────
+   1 │   100
+   2 │     2
+   3 │     3
 
 julia> x
```
nalimilan, line 1017: While we're at it, probably better not print `x` here as it's not mutated, which is kind of the opposite of what the text above says. Or we should explain that `DataFrame(x=x)` makes a copy.

docs/src/man/reshaping_and_pivoting.md
```diff
@@ -5,89 +5,77 @@ Reshape data from wide to long format using the `stack` function:
...
-│ 3   │ Iris-virginica │ PetalWidth │ 1.9     │
-│ 4   │ Iris-virginica │ PetalWidth │ 2.0     │
-│ 5   │ Iris-virginica │ PetalWidth │ 2.3     │
-│ 6   │ Iris-virginica │ PetalWidth │ 1.8     │
+julia> iris = DataFrame(CSV.File(joinpath(dirname(pathof(DataFrames)), "../docs/src/assets/iris.csv")))
+150×5 DataFrame
+ Row │ SepalLength  SepalWidth  PetalLength  PetalWidth  Species
+     │ Float64      Float64     Float64      Float64     String
+─────┼──────────────────────────────────────────────────────────────────
+   1 │ 5.1          3.5         1.4          0.2         Iris-setosa
```
bkamins, line 13: @ronisbr - @nalimilan that floating point columns seem aligned to the left not to the right. Why it is so?

docs/src/man/sorting.md
```diff
@@ -58,71 +58,72 @@ In sorting `DataFrame`s, you may want to sort different columns with different o
...
+ 145 │ 6.3          3.3         6.0          2.5         Iris-virginica
+ 146 │ 6.3          3.4         5.6          2.4         Iris-virginica
+ 147 │ 6.2          3.4         5.4          2.3         Iris-virginica
+ 148 │ 7.2          3.6         6.1          2.5         Iris-virginica
+ 149 │ 7.9          3.8         6.4          2.0         Iris-virginica
+ 150 │ 7.7          3.8         6.7          2.2         Iris-virginica
+                                                        136 rows omitted
+
+julia> sort!(iris, [order(:Species, by=length), order(:SepalLength, rev=true)])
+
```
nalimilan, line 106:
```suggestion
```

src/abstractdataframe/selection.jl
```diff
@@ -1064,139 +1050,137 @@ julia> gd = groupby(df, :a);
...
-│ 2   │ 2     │ (a = [2, 2], b = [1, 1], c = [2, 6]) │
-│ 3   │ 3     │ (a = [3, 3], b = [2, 2], c = [3, 7]) │
-│ 4   │ 4     │ (a = [4, 4], b = [1, 1], c = [4, 8]) │
+ Row │ a      a_b_c_Ref
+     │ Int64  NamedTup…
+─────┼──────────────────────────────────────────
+   1 │     1  (a = [1, 1], b = [2, 2], c = [1,…
+   2 │     2  (a = [2, 2], b = [1, 1], c = [2,…
+   3 │     3  (a = [3, 3], b = [2, 2], c = [3,…
+   4 │     4  (a = [4, 4], b = [1, 1], c = [4,…
```
nalimilan, line 1169: Why isn't this printed in full?

test/grouping.jl
```diff
@@ -1448,10 +1448,11 @@ end
           groupby_checked(DataFrame(x2_identity=[1,1,2]), [])
     @test isequal_typed(DataFrame(gdf), df)
 
-    @test sprint(show, groupby_checked(df, [])) == """
-        GroupedDataFrame with 1 group based on key: 
-        Group 1 (3 rows): 
-         Row │ x1     x2     y
+    # work around automatic trimming of trailing whitespace in most editors
```
nalimilan, line 1451: Is this the only place where we have this problem?

bkamins: we do not need to hunt for it. My editor will catch it the next time I edit some file that has this problem and I will fix it.

src/abstractdataframe/selection.jl
bkamins: Ah - good catch. This is not due to terminal width but the 32 character limit for the width of the column. Before I change it (to produce something narrower) let us confirm that we like this limit.

I am OK to crop in this case.

docs/src/man/getting_started.md
bkamins: I think this was the point of the original example. I have added a comment below.

bkamins:
@ronisbr - there are two things to dicsuss left in this PR. Can you please have a look at them? Thank you!

docs/src/man/getting_started.md
```diff
@@ -1020,6 +1020,8 @@ julia> x
  1
  2
 ```
+Note, that in the above example the original `x` vector is not mutated in the
+process as `DataFrame(x=x)` constructor stores its copy in a data frame.
```
nalimilan, line 1024:
```suggestion
process as the `DataFrame(x=x)` constructor makes a copy by default.
```

src/abstractdataframe/selection.jl
nalimilan: Ah, indeed. This is for this kind of case where we have plenty of space that I wasn't a big fan of the hard limit at 32-char. But that's beyond the scope of this PR.

docs/src/man/reshaping_and_pivoting.md
ronisbr: Hum, this is because the algorithm to align the decimal points. In this case, I need them to be aligned to the left, otherwise it will not work.

bkamins: OK - so we leave it as is.

ronisbr: @bkamins I am not entirely sure, but maybe I can fix this (my concern is adding a lot of processing). Can I try in another PR?

bkamins:
I have updated the example with overwide column that caused cropping

ronisbr:
@bkamins beside that question of float pointing alignment, is there anything else? Everything looks good to me!

bkamins:
> is there anything else?

No - all is clean. I have opened https://github.com/JuliaData/DataFrames.jl/issues/2525 to keep track of to-do, but it is for later (you are probably best to lead these changes).

Also @nalimilan commented that the 32 char column width limit might be reconsidered, but this is for later. The question is if it is possible to do something like:
1. Have 32 char limit by default
2. But if this limit is triggered and we see that we have horizontal space then loosen it

(I assume this is tricky to do correctly, especially if you have more than one column that hits this problem, so I guess both @nalimilan and I think that it is for later)

ronisbr:
> Also @nalimilan commented that the 32 char column width limit might be reconsidered, but this is for later. The question is if it is possible to do something like:
> 
> 1. Have 32 char limit by default
> 2. But if this limit is triggered and we see that we have horizontal space then loosen it
> 
> (I assume this is tricky to do correctly, especially if you have more than one column that hits this problem, so I guess both @nalimilan and I think that it is for later)

I am guessing here, since I really did not analyze the code, but maybe an `autoexpand` option for PrettyTables.jl is not that hard to implement. The column is always processed entirely and cropped when printing. Hence, I just need to change a number. The really tricky part will be to make this compatible with the rest of PrettyTables.jl API, especially when you have multiple lines in a cell or when you have omitted rows. I think this can be done, but not very quickly. I will open an issue to remind me of this.

ronisbr (approved):

bkamins:
> I think this can be done, but not very quickly.

Thank you. I think it is not a super pressing feature. Adding HTML and LaTeX for consistency are more of the priority (if you have the stamina for this). I hope you enjoyed the work - all JuliaData community really appreciates what you have done.

ronisbr:
> Thank you. I think it is not a super pressing feature. Adding HTML and LaTeX for consistency are more of the priority (if you have the stamina for this).

Ok! I will focus in HTML and LaTeX. We are not too far from a stable version, but there is a lot of tests to add. 

>  I hope you enjoyed the work - all JuliaData community really appreciates what you have done.

I really did! I would like also to thank you and @nalimilan, because PrettyTables.jl is much better now after all the modifications :)

docs/src/man/reshaping_and_pivoting.md
nalimilan: That would be nice. Since we're changing printing, which will force dependent packages to update their doctests, better break everything at the same time in 0.22.

nalimilan:
Yes the 32-char limit is much lower priority than aligning float columns to the right IMHO. If possible I'd like to get the latter in 0.22, but the former can wait.

docs/src/man/reshaping_and_pivoting.md
bkamins: @ronisbr - so please check it and let us know. From my perspective this is not a critical thing to have. We are aligning floats to `.` anyway.

The point of right alignment of integers is that they can be easily compared, but this goal is already met with `.` alignment.

bkamins: Maybe the only question is how numbers in scientific notation are aligned? (and do we consistently make sure that all is scientific or non-scientific notation)

Have you tested this?

bkamins:
> aligning float columns to the right

@nalimilan - why do you think it is that important?

nalimilan:
For consistency with integers, and because that's what all other software does (Excel, dplyr...). But the main reason I think it deserves to be in 0.22 is that it's better to avoid breaking doctests of other packages too frequently (other than that it could perfectly wait for a later release).

bkamins:
OK - so if this is reasonably easily doable let us wait for @ronisbr to do the update and then I will rebase and update this PR.

@ronisbr - so please let us know what is your conclusion here.

docs/src/man/reshaping_and_pivoting.md
ronisbr: > @ronisbr - so please check it and let us know. From my perspective this is not a critical thing to have. We are aligning floats to `.` anyway.

OK, I will do some tests.

> Maybe the only question is how numbers in scientific notation are aligned? (and do we consistently make sure that all is scientific or non-scientific notation)
> 
> Have you tested this?

We are using `print` to convert the number. Hence, if the number has order less than -5 or high than +5, then it will be printed in scientific notation. This is taken care and everything is aligned to the decimal point.

bkamins: > This is taken care and everything is aligned to the decimal point.

The same is done in Base, so I guess it is OK.

ronisbr:
To keep the right alignment, we need to compute the field size. In the current implementation, we would need to convert the float columns to string twice (inside DataFrames and inside PrettyTables). This can decrease the performance.

The solution to this is adding an API to post process the table after the conversion of the matrix to strings. However, this will take time to design.

Hence, we have three possible solutions:

1. Process the column twice to right align the numbers properly, even it decrease the performance;
2. Accept it as it is until I can think about something to be implemented in PrettyTables; or
3. Remove the decimal alignment of floating point columns until I can solve this issue.

bkamins:
> Accept it as is until I can think about something to be implemented in PrettyTables

@nalimilan - I vote for this option. I would prefer @ronisbr to design a general solution for alignment in PrettyTables.jl first so that we do not have to use hacks in DataFrames.jl. As for option 3 - I think that decimal alignment is a priority, so I would rule it out.

Two side comments:
1. this issue is mostly visible for columns with very wide labels. for narrow labels it is almost invisible.l
2. Also when we do cropping of the last column actually having the output left aligned is better (as you see "something" at least)

Finally - we consider display changes non breaking, and such a change in the future, while requiring depending packages to update their docstrings is minor (even if these packages do not update the docstrings the difference will be minimal visually).

So in summary - given the situation I vote for leaving things as they are now, as @ronisbr already put massive work into these things. We can update it later when @ronisbr updates PrettyTables.jl to allow for flexible alignment changes.

@nalimilan - but if you would still prefer solution #1 then I understand it is doable, and I am OK with waiting till the workaround is implemented before 0.22 release, so please let us know how you see this.

nalimilan:
Thanks for investigating this @ronisbr. With the goal of minimizing disruptions for users (because doctests can make tests fail even if printing is not part of the API), I'd be inclined to try solution 1 if you feel like it's doable reasonably soon and without doing too much work that you'll have to undo once we have a better solution. But if you feel it's not worth it then we can change that after 0.22.

bkamins:
OK - so if @ronisbr agrees let us go for option 1 that is less performant (doing floating processing twice) and hope that in the future PrettyTables.jl is upgraded to do if faster.

@ronisbr - is this OK with you?

If yes - can you please open a PR with the fix? Then I will rebase this PR after the other is merged (this means that in your PR only update tests and docstrings for `show` - all other things will be handled in this PR). 

(an alternative would be to merge this PR and then you could update only the outputs that get changed after your fix - we can go also this way if you would prefer it)

ronisbr:
I think I can take a look by this weekend! Then I let you know if the downsides are that bad 😄

bkamins:
> I prefer submit a PR and then you update this one, if you don’t mind.

Sure - no problem. So I am waiting for your PR. Thank you!

bkamins:
@nalimilan + @ronisbr I have updated this PR after floating point alignment change. I have caught another case of `NaN` that is special, but it is aligned like `missing` and I think it is OK.

Hopefully this should be good to merge. I have bumped version to 0.22 here so when this PR is merged I will make a release (unless you object :smile:).

nalimilan (approved):

ronisbr (approved):

bkamins:
Thank you! We are going for 0.22 release now!

PR #2524, opened 2020-11-08
fix coverage badge

pdeffebach:
Small oversight fixed.


nalimilan (approved):
Good catch!

(I've cancelled CI since it's useless.)

bkamins:
Thank you!

PR #2527, opened 2020-11-10
Update TagBot.yml

quinnj:
As recommended [here](https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249)

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,9 +1,12 @@
 name: TagBot
 on:
-  schedule:
-    - cron: 0 * * * *
+  issue_comment:
+    types:
+      - created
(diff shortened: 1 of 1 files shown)


bkamins (approved):

bkamins:
Thank you. CI fails for some strange reason - I will investigate it separately.

bkamins:
CI fails because of some breaking changes in Tables.jl.
Surprisingly different CI runs seemed to use different state of Tables.jl.
I will report separately when I investigate it in detail.

bkamins:
> CI runs seemed to use different state of Tables.jl

Nice case 😄 - some of the runs were started before Tables 1.2 was released and some after it was released.

PR #2530, opened 2020-11-10
Update tests to Tables.jl v1.2

### LINKED ISSUE #2529: Sync tests with Tables 1.2
Some tests are broken under Tables.jl 1.2.
This also means that Tables.jl dependency should be bumped.

I will open a PR when I finish updating it.

bkamins:
Fixes #2529

### DIFF
--- a/test/select.jl
+++ b/test/select.jl
@@ -1542,10 +1542,12 @@ end
                                DataFrame(p=Int[], q=String[]))
 
         # here this follows Tables.jl behavior
-        for res in ([1, "1"], (1, "1"))
-            @test select(df, [] => ByRow(() -> res) => AsTable) == DataFrame()
-            @test_throws ArgumentError select(df, [] => ByRow(() -> res) => [:p, :q])
-        end
+        @test select(df, [] => ByRow(() -> [1, "1"]) => AsTable) == DataFrame()
+        @test_throws ArgumentError select(df, [] => ByRow(() -> [1, "1"]) => [:p, :q])
+        @test isequal_coltyped(select(df, [] => ByRow(() -> (1, "1")) => AsTable),
+                               DataFrame(Column1=Int[], Column2=String[]))
+        @test isequal_coltyped(select(df, [] => ByRow(() -> (1, "1")) => [:p, :q]),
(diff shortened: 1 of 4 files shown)


test/select.jl
```diff
@@ -1542,10 +1542,12 @@ end
...
-        for res in ([1, "1"], (1, "1"))
-            @test select(df, [] => ByRow(() -> res) => AsTable) == DataFrame()
-            @test_throws ArgumentError select(df, [] => ByRow(() -> res) => [:p, :q])
-        end
+        @test select(df, [] => ByRow(() -> [1, "1"]) => AsTable) == DataFrame()
+        @test_throws ArgumentError select(df, [] => ByRow(() -> [1, "1"]) => [:p, :q])
+        @test isequal_coltyped(select(df, [] => ByRow(() -> (1, "1")) => AsTable),
+                               DataFrame(Column1=Int[], Column2=String[]))
+        @test isequal_coltyped(select(df, [] => ByRow(() -> (1, "1")) => [:p, :q]),
+                               DataFrame(p=Int[], q=String[]))
```
nalimilan, line 1550: Can you explain why `(1, "1")` is treated differently from `[1, "1"]` now?

bkamins: Because in Tables.jl 1.2 tuples have a schema (they have a fixed number of elements), while vectors do not have it (as they can have varying number of elements).

bkamins: `Tuple`s now behave like `NamedTuple`s but with auto-generated column names.

nalimilan (approved):

test/select.jl
quinnj: Hmmmm, I don't quite follow what's going on here; how are `(1, "1")` and `[1, "1"]` being treated here? as an entire table? or as a single row? What (should have) changed in Tables.jl 1.2 is only the "empty" table case, where we will preserve the input schema if possible.

bkamins: And this is exactly what is tested here - if you look above these lines you will see that we are testing source data frames with `0` rows.

bkamins:
Thank you!

PR #2533, opened 2020-11-10
Reduce number of tests in the CI

bkamins:
Currently CI takes ~75 minutes to finish. Also the practice has shown that what we essentially need is:
- test all options on current release
- test all platforms on x64 nightly (to report bugs back to Julia Base, as they happen on different platforms)
- just run one test on LTS 1.0 (as things that stop working are not platform specific)

I would like to open a discussion if we go this way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,12 +25,23 @@ jobs:
   exclude:
     - os: osx
       arch: i386
+    - julia 1.0
+      os: linux
+      arch: i386
+    - julia 1.0
+      os: osx
+    - julia 1.0
+      os: windows
+    - julia nightly
+      os: linux
+      arch: i386
   allow_failures:
     - julia: nightly
   include:
     - stage: "Documentation"
-      julia: 1.0
+      julia: 1
       os: linux
+      arch: amd64
       script:
         - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
                                     Pkg.instantiate()'


nalimilan:
Ideally we would have a way to run CI on all platforms from time to time (notably before releases) to ensure that everything is OK, while checking only the minimum platforms for PRs. But I have no idea whether that's possible.

We could also consider moving to GitHub Actions for CI. Even with fewer platforms tested, do you think [the new time budget](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) that Travis CI is going to offer us will be enough? I'm mentioning this because it [might be possible](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/manually-running-a-workflow) to run CI on more platforms manually using a separate GitHub Action, which would allow us to make the standard CI lighter.

bkamins:
> But I have no idea whether that's possible.

I can apply this patch after the release each time and revert it before the release. Cumbersome, but doable.

> We could also consider moving to GitHub Actions for CI.

The new Travis policy is also my concern here. But do we know who can correctly set up GitHub Actions for CI? I have no experience with this unfortunately.

I will ask on Slack.

bkamins:
@fredrikekre / @aviks : maybe you please could suggest who has experience with setting up GH actions, as neither @nalimilan nor me have experience with it. Thank you!

bkamins:
@nalimilan - as @quinnj commented on Slack - we probably should move go GH Actions instead.

@quinnj https://github.com/quinnj/JSON3.jl/blob/master/.github/workflows/ci.yml - looking at it I understand that we would drop OSx also. So the question is if you think that we should keep Win, OSx and 32-bit tests on Travis? And if doing 32-bit tests is feasible on GH Actions.

Sorry for not being more well educated on CI topics.

As a comment: we have caught many bugs in Julia Base, especially on nightly, thanks to doing tests on Win and OSx and on 32-bit arch in the past.

nalimilan:
Does GitHub Actions support macOS? If so, we could add another set of tests that we would run from time to time, and notably before releases.

bkamins:
I think it does not include it. Running extra tests before releases is a good idea. The question is how to set up such "on demand" tests?

quinnj:
Superseded by #2552

quinnj:
Do we have a good sense for how "efficient" our `runtests.jl` are? Are we perhaps over testing in some cases? I know in CSV.jl, the tests have gotten a little out of hand over the years and it could use a cleanup since the tests are starting to take quite a long time. Perhaps there's room for improvement here as well.

nalimilan:
A lot of time is spent testing the grouping code, but it's so tricky that simplifying it while still covering all cases would be difficult.

quinnj:
Github action runners (vms) have 2 cores, so maybe we could do some parallelization; like do a `Threads.@spawn` for each test file and max the cpu out some more.

bkamins:
It would be good to have it. The question is how to make sure that then we have a clean output from tests to the terminal when we use 2 threads?

bkamins:
Also - in the future we will be adding multithreading support to operations, so this will be actually good thing to have per operation. @nalimilan has recently been working on this

nalimilan:
Waiting for 20 minutes to get the result of CI is OK IMO. (Yes, we'll have to find a way to test threaded operation, which will make things a bit more complex.)

quinnj:
That's easy, we just add `env: JULIA_NUM_THREADS: 2` to our CI; I can do that now and we won't have to worry about it.

nalimilan:
What will be complex is to ensure we test the relevant code paths both with a single and multiple threads. The risk is that we have to run lots of tests twice, with a loop, taking even more time.

bkamins:
But looking at the current PRs that I have just switched to use GH Actions things look much better already.

PR #2537, opened 2020-11-14
Improve floating point alignment

ronisbr:
@bkamins and @nalimilan,

As we say here, prontinho :)

I could add a good algorithm to align floating points to the right in all occasions.

Now, I do not store a vector with some information about padding for each row. I compute the column inside the cell that the decimal point must be aligned and pass just one number per column to the formatter. The downside is that I need to convert the cell to string twice. However, the good side, which @nalimilan pointed out in the last PR, is that it is **way** faster to print very large DataFrames because of the lower allocation. For example, here, a 100_000 x 2 DataFrame (Float64) took 8s to be printed and now it takes only 1.7s.

Captura de Tela 2020-11-13 às 23 45 13


bkamins:
This looks fantastic. One test needs to be updated but this is a minor thing (difference in formatting of `1.00003e5` vs `100003.0`).

src/abstractdataframe/prettytables.jl
```diff
@@ -65,24 +65,31 @@ end
...
-            return " "^pad * str
-        end
+        # We want to align everything at '.'.
+        id_dp = findfirst('.', str)
+
+        # Compute the require padding to align the cell.
+        pad = id_dp !== nothing ? align_col_i - id_dp : align_col_i - 1
+
+        # Pad cannot be negative.
+        # NOTE: This is just a failsafe, `pad` will be negative only if we have
```
bkamins, line 88: This is a nice safeguard. I understand that already `id_dp` is `nothing` only if we have a bug - right?

src/abstractdataframe/show.jl
```diff
@@ -604,6 +604,11 @@ function _show(io::IO,
 
     compact_printing::Bool = get(io, :compact, true)
 
+    # For consistency, if `kwargs` has `compact_printng`, we must use it.
+    if haskey(kwargs, :compact_printing)
+        compact_printing = kwargs[:compact_printing]
+    end
+
```
bkamins, line 611: maybe this will be easier to read? (it is a soft suggestion)
```suggestion
    # For consistency, if `kwargs` has `compact_printng`, we must use it.
    if haskey(kwargs, :compact_printing)
        compact_printing = kwargs[:compact_printing]
    else
        compact_printing::Bool = get(io, :compact, true)
    end

```

src/abstractdataframe/show.jl
```diff
@@ -614,10 +619,9 @@ function _show(io::IO,
     # the printed numbers will be aligned on the decimal point.
     float_cols = Int[]
 
-    # These vectors contain the number of the row and the padding that must be
-    # applied so that the float number is aligned with on the decimal point
-    indices = Vector{Int}[]
-    padding = Vector{Int}[]
+    # This vector contains the column with respect to the beginning of the cell
+    # that the decimal point must be aligned.
```
bkamins, line 623:
```suggestion
    # that the decimal point must be aligned to.
```

src/abstractdataframe/show.jl
```diff
@@ -657,13 +661,15 @@ function _show(io::IO,
...
         for i in float_cols
-            # Analyze the order of the number to compute the maximum padding
-            # that must be applied to align the numbers at the decimal
-            # point.
 
-            max_order_i = 0
-            order_i = zeros(Δr)
-            indices_i = zeros(Δr)
+            # This variable stores the column with respect the beginning of the
+            # cell that the decimal point must be aligned.
```
bkamins, line 666:
```suggestion
            # cell that the decimal point must be aligned to.
```

src/abstractdataframe/show.jl
```diff
@@ -675,46 +681,42 @@ function _show(io::IO,
...
-                    # scientific notation.
-                    order_v = log_v > 5 ? 0 : floor(Int, log_v)
-
-                    # If the number is negative (including -0.0), we need to add an additional
-                    # padding to print the sign.
-                    isless(v, 0) && (order_v += 1)
+                if id_dp !== nothing
+                    (align_col_i < id_dp) && (align_col_i = id_dp)
+                else
+                    id_dp = 0
```
bkamins, line 692: I understand this should not happen in theory - right?

src/abstractdataframe/show.jl
```diff
@@ -675,46 +681,42 @@ function _show(io::IO,
...
                 end
 
-                order_i[k] = order_v
-                indices_i[k] = kr
+                size_after_align_col = length(v_str) - id_dp
 
-                order_v > max_order_i && (max_order_i = order_v)
+                max_size_after_align_col < size_after_align_col &&
+                    (max_size_after_align_col = size_after_align_col)
```
bkamins, line 698:
```suggestion
                if max_size_after_align_col < size_after_align_col
                    max_size_after_align_col = size_after_align_col
                end
```

bkamins (approved):
Looks good. I have left some minor comments + one test needs to be updated. In general I understand that `id_dp` should never be `nothing`, and you handle this case not to throw an error in some unexpected situations - right?

src/abstractdataframe/prettytables.jl
ronisbr: No. I can be nothing at least in two occasions:

1. The value is `missing`.
2. The user defines a structure that is `<: AbstractFloat` in which the printing method does not contain a decimal point.

src/abstractdataframe/show.jl
ronisbr: Yes, it can happen if the value is `missing` for example.

bkamins:
Ah - `missing` makes sense :) as a special case. I understand that you used the rule that for `missing` we have that `m` is aligned to the column where `.` is stored? But it would be better to special case `missing` I think and just print it as is (left or right aligned - as you prefer, but not influencing the rest). Is this hard?

Also - what would be alignment of a float that does not print `.`. It should be like this:
```
 123.0
1234
```
(assuming 1234 were a "strange float")
but it seems that the algorithm would do something different in this case.

You can use such type for testing:
```
struct F<:AbstractFloat
       i::Int
       end

Base.show(io::IO, f::F) = show(io, f.i)
```
(should be added to tests I think)

PS. Sorry for throwing more requirements but if we fix this in general then it would be best to clean it up till the end.

ronisbr:
OK, I will do this. The only problem is that failing test. Here, in my machine, `10.0^5 + pi` is converted to `100003.0` instead of `1.00003e5` as in the Travis testing. Do you have any idea why this is happening?

bkamins:
I think the reason is `compact` option in IO:
```
julia> show(IOContext(stdout, :compact => true), [10.0^i + pi for i = -5:1:5])
[3.1416, 3.14169, 3.14259, 3.15159, 3.24159, 4.14159, 13.1416, 103.142, 1003.14, 10003.1, 1.00003e5]
julia> show(IOContext(stdout, :compact => false), [10.0^i + pi for i = -5:1:5])
[3.141602653589793, 3.1416926535897933, 3.142592653589793, 3.151592653589793, 3.241592653589793, 4.141592653589793, 13.141592653589793, 103.1415926535898, 1003.1415926535898, 10003.14159265359, 100003.14159265358]
```
but I am not 100% sure.

bkamins:
ah - regarding `missing` alignment probably it is best to align it the same way it is aligned in other cases (how do we align `missing` when they are mixed in `Union{Int, Missing}` eltype?)

ronisbr:
What!? Something is really weird here:

```julia
julia> show(IOContext(stdout, :compact => true), [10.0^i + pi for i = -5:1:5])
[3.1416, 3.14169, 3.14259, 3.15159, 3.24159, 4.14159, 13.1416, 103.142, 1003.14, 10003.1, 100003.0]
```

ronisbr:
> ah - regarding `missing` alignment probably it is best to align it the same way it is aligned in other cases (how do we align `missing` when they are mixed in `Union{Int, Missing}` eltype?)

We align to the right. I am following what Base do:

```julia
julia> [i == 3 ? missing : 10.0^i for i = -5:1:5]
11-element Array{Union{Missing, Float64},1}:
      1.0e-5
      0.0001
      0.001
      0.01
      0.1
      1.0
     10.0
    100.0
       missing
  10000.0
 100000.0
```

Should I change?

bkamins:
>  I am following what Base do:

then this is OK, but it is strange that `missing` is more to the right than it would need to be. Still - if this is what base does then it is OK.

bkamins:
> What!? Something is really weird here:

Really strange. Maybe just replace the test with something that pasess then :smile: .

ronisbr:
> then this is OK, but it is strange that `missing` is more to the right than it would need to be. Still - if this is what base does then it is OK.

I really prefer to align missing before the decimal point. I just did this because it is was `Base` does. Should I change?

bkamins:
Also is https://github.com/ronisbr/PrettyTables.jl/issues/91 something that should hold us off as it needs to be fixed in PrettyTables.jl or it is not a problem?

bkamins:
> I really prefer to align missing before the decimal point. I just did this because it is was Base does. Should I change?

I would leave it as is if we follow that Base does (but it was also strange for me).

ronisbr:
> Also is [ronisbr/PrettyTables.jl#91](https://github.com/ronisbr/PrettyTables.jl/issues/91) something that should hold us off as it needs to be fixed in PrettyTables.jl or it is not a problem?

That example is using the old API. Did I miss something in the manual?

ronisbr:
> Really strange. Maybe just replace the test with something that pasess then 😄 .

Or just remember to update Julia before submitting PR... I was using 1.5.0 and it changed when I updated to 1.5.3 🤦🏻‍♂️

bkamins:
Yes - display changes are defined to be a "non-breaking" and you chose to work in the most muddy waters :smile: (and thank you for doing this as I can imagine how hard it is to maintain the whole PrettyTables.jl stuff). Great work!

ronisbr:
I think I did everything :) If the user defines a float without a decimal point, this can handle as you proposed.

src/abstractdataframe/show.jl
```diff
@@ -602,7 +602,12 @@ function _show(io::IO,
         crop = :both
     end
 
-    compact_printing::Bool = get(io, :compact, true)
+    # For consistency, if `kwargs` has `compact_printng`, we must use it.
+    if haskey(kwargs, :compact_printing)
+        compact_printing = kwargs[:compact_printing]
+    else
+        compact_printing::Bool = get(io, :compact, true)
+    end
```
nalimilan, line 610: Or maybe even this? :-)
```suggestion
        compact_printing::Bool = get(kwargs, :compact_printing, get(io, :compact, true))
```

src/abstractdataframe/show.jl
```diff
@@ -675,46 +681,49 @@ function _show(io::IO,
...
+                else
+                    # We want to align everything at '.'.
+                    id_dp = findfirst('.', v_str)
 
-                    # If the order is higher than 5, then we print using
-                    # scientific notation.
-                    order_v = log_v > 5 ? 0 : floor(Int, log_v)
+                    # If a decimal point is not found, then assume that the
+                    # entire should be aligned before the alignment column. This
+                    # can happen with a custom `AbstractFloat` structure.
```
nalimilan, line 695:
```suggestion
                    # entire text should be aligned before the alignment column. This
                    # can happen with a custom `AbstractFloat` type.
```

bkamins:
Wow - you are really skilled with this. Let us wait for @nalimilan to have a look at this PR before merging.

test/show.jl
```diff
@@ -511,24 +517,139 @@ end
...
+
+    df = DataFrame(😄😄😄😄😄😄😄 = [10.0^i for i = -5:1:5],
+                   🚀🚀🚀🚀🚀 = 1.0:2:22)
+
+    io = IOBuffer()
+    show(io, df)
+    str = String(take!(io))
+    @test str == """
+        11×2 DataFrame
+         Row │ 😄😄😄😄😄😄😄  🚀🚀🚀🚀🚀
```
nalimilan, line 571: This is a case where you're smarter than GitHub or Firefox, as it doesn't show the alignment correctly. :-D

src/abstractdataframe/prettytables.jl
```diff
@@ -65,24 +65,40 @@ end
...
+            pad = align_col_i - 1
+        else
+            # We want to align everything at '.'.
+            id_dp = findfirst('.', str)
+
+            # Compute the require padding to align the cell.
+            #
+            # If a decimal point is not found, then assume that the
+            # entire should be aligned before the alignment column. This
+            # can happen with a custom `AbstractFloat` structure.
```
nalimilan, line 91:
```suggestion
            # entire text should be aligned before the alignment column. This
            # can happen with a custom `AbstractFloat` type.
```

nalimilan:
Actually Base also aligns `missing` on the left to the "invisible decimal dot" for integers:
```julia
julia> [1.03585, missing]
2-element Vector{Union{Missing, Float64}}:
 1.03585
  missing

julia> [1, missing]
2-element Vector{Union{Missing, Int64}}:
 1
  missing
```
What we do is the same for floats, but different for integers, where `missing` is aligned on the right to the "invisible decimal dot" (cf. the screenshots you posted). I agree this makes more sense as it saves a lot of space. Maybe we should do the same for floats, i.e. align them to the decimal dot *on the right*? Two thoughts:
- it would be better for consistency between floats and integers
- but it's hard to know which solution saves more space: it depends on whether you store large numbers

What do you think?

nalimilan (commented):
Looks great!

ronisbr:
> What we do is the same for floats, but different for integers, where `missing` is aligned on the right to the "invisible decimal dot" (cf. the screenshots you posted). I agree this makes more sense as it saves a lot of space. Maybe we should do the same for floats, i.e. align them to the decimal dot _on the right_? Two thoughts:
> 
> * it would be better for consistency between floats and integers
> * but it's hard to know which solution saves more space: it depends on whether you store large numbers
> 
> What do you think?

I think it is better to align `missing` to the right of the decimal point, just as we do with `Int`. It is different from Base, but seems much better IMHO.

src/abstractdataframe/show.jl
ronisbr: This is awesome!

nalimilan:
> > 
> 
> I think it is better to align `missing` to the right of the decimal point, just as we do with `Int`. It is different from Base, but seems much better IMHO.

OK, sounds good to me too. FWIW it's what dplyr does. Can you implement that if @bkamins agrees?

src/abstractdataframe/show.jl
nalimilan: I hope it works, I haven't checked. :-p

ronisbr:
> OK, sounds good to me too. FWIW it's what dplyr does. Can you implement that if @bkamins agrees?

Yes, sure, it is very easy:

Captura de Tela 2020-11-14 às 12 06 04

bkamins:
> Can you implement that if @bkamins agrees?

Agreed and let us move forward 😄.

src/abstractdataframe/show.jl
bkamins: seems OK

src/abstractdataframe/show.jl
```diff
@@ -602,7 +602,8 @@ function _show(io::IO,
         crop = :both
     end
 
-    compact_printing::Bool = get(io, :compact, true)
+    # For consistency, if `kwargs` has `compact_printng`, we must use it.
+        compact_printing::Bool = get(kwargs, :compact_printing, get(io, :compact, true))
```
bkamins, line 606:
```suggestion
    compact_printing::Bool = get(kwargs, :compact_printing, get(io, :compact, true))
```

bkamins: alignment fix

ronisbr:
Done!

bkamins:
@nalimilan - OK to merge?

nalimilan (approved):
Yes !

bkamins:
Thank you!

PR #2546, opened 2020-11-16
Spaces after commas

kescobo:
This is the first of 3 PRs replacing #2488 now that the big reorg has happened. These should not change or add any functionality, but are instead intended to unify some style conventions.

Here, I deal with spaces after commas. Original post:

> 1. The biggest thing is spaces after commas, which I accomplished primarily using regex find-and-replace in VS-Code (there were nearly 3000 instances). Specifically, I did `,([^\s)])` => `, $1`. I initially searched for `,\S`, but there were a bunch of instances of 1-element tuples that I assumed should not be changed (that is, I wanted to leave `(a='b',)` alone).
>     1. There were a few file exceptions to global find-and replace: `iris.csv` (for obvious reasons); `test/dataframerow.jl` due to a couple of `@sprint` lines that seemed like they shouldn't be changed - I went in and changed some other stuff in this file manually; `test/io.jl` that had a bunch of output stuff that I didn't want to mess with.
>     1. [In `test/grouping.jl`](https://github.com/JuliaData/DataFrames.jl/compare/master...kescobo:style?expand=1#diff-5dcf6d0cf7bc3b35e4d99277738b1a186a2d3b1540141f8b11ee1108defe7199L2094-R2095), there were a couple of examples of spaces _before_ commas that I removed (I only noticed them because there were also no spaces after the commas)
>     1. A few lines in `test/cat.jl` with error messages were inconsistent, I tried to unify
>     1. I tried to scan through everything else, but it's definitely possible I missed something.

This is still largely true, but I added `src/abstractdataframe/io.jl` to the list of file exclusions

Another thing to note is that in `categorical.md` (and elsewhere), there are doctests where type-signatures in the output are eg

```jl
julia> cv = CategoricalArray(v)
6-element CategoricalArray{String,1,UInt32}:
```

This is indeed how the `CategoricalArrays` type signature looks, but in other cases (eg `2-element Vector{Union{Missing, Int64}}`), there are spaces. I thought at first to click through manually and just leave anything that's in a doctest output alone
(assuming that someone copy-pasted), but I noticed that on julia 1.6, `Array{T, 1}` is now `Vector{T}` in the REPL, and other type-signature things have different structure, eg

```julia
# julia 1.5
julia> reshape([1,missing,3,4,5,6], 1,2,3) |> collect
1×2×3 Array{Union{Missing, Int64},3}

# vs julia 1.6
julia> reshape([1,missing,3,4,5,6], 1,2,3) |> collect
1×2×3 Array{Union{Missing, Int64}, 3}
#...

# julia 1.5
julia> v = [(a=1, b=2), (a=3, b=4)]
2-element Array{NamedTuple{(:a, :b),Tuple{Int64,Int64}},1}:
 (a = 1, b = 2)
 (a = 3, b = 4)

# vs julia 1.6
julia> v = [(a=1, b=2), (a=3, b=4)]
2-element Vector{NamedTuple{(:a, :b), Tuple{Int64, Int64}}}: # note spaces in Tuple
 (a = 1, b = 2)
 (a = 3, b = 4)
 ```

but doctests seem to succeed on `nightly` regardless. This may be worth fixing later, but I think I managed to leave them all alone.

Tests pass locally, let's see what happens!

------

PS - I feel a little bit bad that, especially because of the new `precompile` file, which this PR changes almost every line of,
that it's going to look like I'm much more of a contributor than I actually am.

Not sure what to do about it, but if there's any remedy for that, but if there is, let me know. Perhaps that file in particular should not actually follow this rule?

### CODE REVIEW

**Comment by bkamins** [comment]
> that it's going to look like I'm much more of a contributor than I actually am.

It is not a problem for me.

File: Project.toml [hunk 1]
```diff
@@ -37,6 +36,7 @@ Reexport = "0.1, 0.2"
 SortingAlgorithms = "0.1, 0.2, 0.3"
 TableTraits = "0.4, 1"
 Tables = "1.2"
+julia = "1"
```
bkamins, line 39: why put it last?

docs/src/man/comparisons.md
```diff
@@ -40,7 +40,7 @@ as row indices rather than a separate `id` column.
 | Column slicing by location | `df.iloc[:, 1:]`       | `df[:, 2:end]`                     |
 | Row indexing by label      | `df.loc['c']`          | `df[findfirst(==('c'), df.id), :]` |
 | Column indexing by label   | `df.loc[:, 'x']`       | `df[:, :x]`                        |
-| Column slicing by label    | `df.loc[:, ['x','z']]` | `df[:, [:x, :z]]`                  |
+| Column slicing by label    | `df.loc[:, ['x', 'z']]` | `df[:, [:x, :z]]`                  |
```
bkamins, line 43: can you please double check that if we loose alignment of `|` in markdown is not required in tables (I think it is not required).

docs/src/man/missing.md
```diff
@@ -1,177 +1,177 @@
...
- missing  missing  missing
-
-julia> missings(Int, 1, 3)
-1×3 Array{Union{Missing, Int64},2}:
- missing  missing  missing
-
-```
-
-See the [Julia manual](https://docs.julialang.org/en/v1/manual/missing/) for more information about missing values.
+# Missing Data
```
bkamins, line 1: why is it all changed here?

bkamins: especially that it seems that this file is almost unchanged.

src/other/precompile.jl
```diff
@@ -16,2236 +16,2236 @@ function precompile(all=false)
     if !all
         ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
 
-        let fbody = try __lookup_kwbody__(which(DataFrames.combine, (GroupedDataFrame{DataFrame},Array{Bool,1},Vararg{Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractArray{T,1} where T, Type, All, Between, Cols, InvertedIndex},N} where N,))) catch missing end
+        let fbody = try __lookup_kwbody__(which(DataFrames.combine, (GroupedDataFrame{DataFrame}, Array{Bool, 1}, Vararg{Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractArray{T, 1} where T, Type, All, Between, Cols, InvertedIndex}, N} where N,))) catch missing end
```
bkamins, line 19: I would leave-out precompile.jl from fixing formatting as it is auto-generated.

test/broadcasting.jl
```diff
@@ -453,15 +453,15 @@ end
                          4.5  7.5  9.5  12.5  15.5]
 
     df = copy(refdf)
-    df[:, [1,2]] .= Matrix(df[:, [1,2]]) .+ [1 4
+    df[:, [1, 2]] .= Matrix(df[:, [1, 2]]) .+ [1 4
```
bkamins, line 456: alignment is broken here (and below).

bkamins:
I have went through all changes and left the comments I had. Other than that it looks good.
In this case the changes are consistent with https://github.com/JuliaData/DataFrames.jl/blob/master/CONTRIBUTING.md.

In next PRs (but wait with them till this one is finished please) please check https://github.com/JuliaData/DataFrames.jl/blob/master/CONTRIBUTING.md and make sure that the changes are consistent with the recommendations there (maybe it needs updating then please update it).

Thank you!

Project.toml
kescobo: I did not change this intentionally :-/. Not sure how that happened, but I had to `] instantiate` and did it on both julia 1.5 and 1.6, maybe that did something? I'll change it back

docs/src/man/comparisons.md
kescobo: Almost certain that it's not required, but I will fix it anyway. No sense fixing one style thing and breaking another :laughing:

src/other/precompile.jl
kescobo: Ahh, :+1:

docs/src/man/comparisons.md
bkamins: great - and now you see why your work is that much appreciated 😄.

docs/src/man/missing.md
kescobo: Yeah, this is weird. When I look in VS code's diff view, the only thing changed is line 105, but when I use `git diff`, it shows the same thing as this.

I think this is a git bug. If you checkout master, add that one comma and then do `git diff`, it says the whole file is changed.

bkamins: OK - just wanted to make sure as this is hard to review.

docs/src/man/comparisons.md
kescobo: In fixing this, I noticed an inconsistency with table formatting (specifically the alignment rows) in this file. Roughly half of them have spaces after and before the bars, while the others don't. That is

```
| :------ | ------- |
vs
|:--------|---------|
```

Would you like me to fix these up as well? I prefer the later style.

bkamins: I also prefer the latter. It would be great if you fixed these.

kescobo:
Alright, I've got these fixes locally, waiting to [resolve this](https://github.com/JuliaData/DataFrames.jl/pull/2546#pullrequestreview-531583677) before pushing.

bkamins:
I replied earlier - use the latter please.

kescobo:
Ah, looks like I needed a refresh. Should be all set now!

bkamins (approved):

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -24,11 +24,11 @@ mutable struct GroupedDataFrame{T<:AbstractDataFrame}
...
-    starts::Union{Vector{Int},Nothing}   # starts of groups after permutation by idx
-    ends::Union{Vector{Int},Nothing}     # ends of groups after permutation by idx
+    idx::Union{Vector{Int}, Nothing}      # indexing vector sorting rows into groups
+    starts::Union{Vector{Int}, Nothing}   # starts of groups after permutation by idx
+    ends::Union{Vector{Int}, Nothing}     # ends of groups after permutation by idx
     ngroups::Int                         # number of groups
-    keymap::Union{Dict{Any,Int},Nothing} # mapping of key tuples to group indices
+    keymap::Union{Dict{Any, Int}, Nothing} # mapping of key tuples to group indices
     lazy_lock::Threads.ReentrantLock     # lock is needed to make lazy operations
                                          # thread safe
```
nalimilan, line 33: Need to adjust spacing to align comments.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -716,7 +716,7 @@ julia> k = keys(gd)[1]
 GroupKey: (a = :foo, b = 2)
 
 julia> keys(k)
-2-element Array{Symbol,1}:
+2-element Array{Symbol, 1}:
```
nalimilan, line 719: This is the actual output:
```suggestion
2-element Array{Symbol,1}:
```

bkamins: ah - right! We should not change the docstrings.

bkamins:
@kescobo - I have merged the other two PRs. This one requires another pass on your side. Thank you!

src/groupeddataframe/groupeddataframe.jl
kescobo: On it!

src/groupeddataframe/groupeddataframe.jl
kescobo: Whoops! Yeah, I was trying to avoid those, good catch!

kescobo:
Rebased and addressed @nalimilan 's review.

bkamins:
Looks good. @nalimilan - can you please approve or merge when CI passes and you are also OK with the changes? Thank you!

nalimilan:
Thanks @kescobo! Would you be familiar with automated code formatting bots like https://github.com/davidanthoff/julia-pkgbutler and https://github.com/julia-actions/julia-format by chance? I wonder whether it would be workable to enable these.

bkamins:
@kescobo - let me also voice the appreciation.

In general (this is a soft suggestion) - I think it would be great if we had you as a "code layout janitor" (given you already have invested time in it). In particular - if you wanted to invest time into it - maybe some automated tools like @nalimilan suggest might be used for it so that it can be taken care of without human intervention (unfortunately neither @nalimilan nor me have much experience with these things).

Please treat it as a soft suggestion of course. I just feel what you did is great and it would be good to have a way to keep the style consistent in the future (and believe me - for me as a person who implements things mostly - it is really hard even if I try; @nalimilan e.g. does a fantastic job making sure that documentation is properly written, as I am also really lousy in this dimension also unfortunately).

kescobo:
I am only familiar with `JuliaFormatter.jl` in the sense that it comes with VS code. It looks like `julia-format` action does a lot less than `JuliaFormatter.jl` is capable of (it might be possible to expand), but based on a short skim, neither deals with the stuff that I did in these 3 PRs. It might be worth making a pass through the entire code-base periodically (I can give that a shot), but I think trying to use it automatically on a repo of this size is asking for trouble.

I am much better at binge cleaning than maintenance cleaning (in my house and in my code), as evidenced by how many things @nalimilan has found to fix in my PRs. That said, I also have some weird places where I can engage anal-retentiveness, and I'd be happy to try to put those skills to work here, depending on what that entails. Were you thinking that I would review PRs looking for this kind of thing?

bkamins:
> Were you thinking that I would review PRs looking for this kind of thing?

If you would be willing I can mention you in bigger PRs so that you can have a look if you would have time. Would this work with you?

kescobo:
@bkamins Sure, happy to!

PR #2548, opened 2020-11-16
Zero after decimal

kescobo:
This is the third of 3 PRs replacing #2488 now that the big reorg has happened. These should not change or add any functionality, but are instead intended to unify some style conventions.

Here, I add zeros after decimals when floats are manually written. These were found by doing find/replace for `(\d)\.(\D)` => `$1.0$2`, though I had to manually go through and accept the replacements because there are a lot of things like `df1.col`.

In addition, this PR includes addition to `CONTRIBUTING.md` for #2547


bkamins (approved):

nalimilan (approved):

bkamins:
Thank you!

Issue #2550, opened 2020-11-18
Two small typos in docs


roualdes:
1. There is a missing period, to end the sentence, immediately following the only right parenthesis of the paragraph beginning with `DataFrameRow` in the Type hierarchy design section of the [Types page](https://dataframes.juliadata.org/stable/lib/types/).

2. In the paragraph beginning with "All these functions take a specification" on the [The Split-Apply-Combine Strategy page](https://dataframes.juliadata.org/stable/man/split_apply_combine/), and within the second item of the ordered list, the phrase "which can be a any valid column selector" should not include the article a and should be changed to "which can be any valid column selector."

I'm happy to submit a PR for these, if that's easier for you.  Just let me know.

Thanks for such great documentation.

bkamins:
If you would be willing to it would be easiest to review if you made a PR. I am not a native speaker, so usually @nalimilan is also involved 😄. Thank you!

roualdes:
Great.  Just submit PR #2551.  Thanks.

PR #2551, opened 2020-11-19
issue #2550 fix two small typos in docs

roualdes:
Fixes issue #2550.

I looked through the contributor's guide, but didn't see anything about local testing before submitting a PR.  Please let me know if there is documentation that I missed and should follow.  Thanks.

### DIFF
--- a/docs/src/lib/types.md
+++ b/docs/src/lib/types.md
@@ -28,8 +28,7 @@ It is intended to be created as a result of a call to the `groupby` function.
 
 `DataFrameRow` is a view into a single row of an `AbstractDataFrame`. It stores only a reference
 to a parent `DataFrame` and information about which row and columns from the parent are selected
-(both as integer indices referring to the parent)
-The `DataFrameRow` type supports iteration over columns of the row and is similar in functionality to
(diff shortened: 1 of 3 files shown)


docs/src/lib/types.md
```diff
@@ -28,8 +28,7 @@ It is intended to be created as a result of a call to the `groupby` function.
 
 `DataFrameRow` is a view into a single row of an `AbstractDataFrame`. It stores only a reference
 to a parent `DataFrame` and information about which row and columns from the parent are selected
-(both as integer indices referring to the parent)
-The `DataFrameRow` type supports iteration over columns of the row and is similar in functionality to
+(both as integer indices referring to the parent). The `DataFrameRow` type supports iteration over columns of the row and is similar in functionality to
```
bkamins, line 31:
```suggestion
(both as integer indices referring to the parent).
The `DataFrameRow` type supports iteration over columns of the row and is similar in functionality to
```

bkamins:
Thank you. While we are at it could you also fix https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/selection.jl#L47 as there is the same error there.

bkamins (approved):

bkamins:
Thank you!

PR #2553, opened 2020-11-20
error when using one dimension for indexing

bkamins:
This was asked for by @nalimilan and @pdeffebach.

I think we should be restrictive and cover only the most common cases. I was also unsure how to best formulate the error (and let us hope these definitions do not run into method ambiguities, because we use `AbstractDataFrame` in both, and `Any` in the second one). It seems they are OK.

The only change is that we throw `ArgumentError` not `MethodError` but it is considered non-breaking.


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1995,3 +1995,11 @@ function repeat_lengths!(longnew::AbstractVector, shortold::AbstractVector,
...
     end
 end
+
+# Disallowed operations that are a common mistake
+
+Base.getindex(::AbstractDataFrame, ::Symbol) =
+    throw(ArgumentError("syntax df[column] is not supported use df[!, column] instead"))
+
+Base.setindex!(::AbstractDataFrame, ::Any, ::Symbol) =
+    throw(ArgumentError("syntax df[column] is not supported use df[!, column] instead"))
```
nalimilan, line 2005: How about also accepting strings? They are as likely to be used as symbols.

bkamins: This is one of the things that is disputable. The time when `df[:col]` was accepted we did not accept strings for column indexing, so old code would be not affected by your proposal.

However, maybe pandas users might try this. So the error with `AbstractStrings` would be not for legacy code fixing but for new users. What do you think given this?

nalimilan: Yes pandas and R users will likely try that. I'd say let's bite the bullet: if we add this, better cover all cases. We could also do the same for `Integer`.

bkamins: OK - changed. Let us see how it goes

nalimilan (approved):

bkamins:
@pdeffebach - are you OK with this? After I merge this PR I am going to make a patch release.

pdeffebach (approved):

bkamins:
Thank you!

PR #2554, opened 2020-11-20
remove deprecations and CategoricalArrays.jl dependency

bkamins:


bkamins:
It seems that removing CategoricalArrays.jl dependency does not have much impact on compilation times now:
### Master
```
julia> @time using DataFrames;
  1.388877 seconds (2.29 M allocations: 147.759 MiB, 7.78% gc time)

julia> @time df = DataFrame(:a => [1,2], :b => [3,4]);
  0.192164 seconds (190.26 k allocations: 10.219 MiB)

julia> @time gdf = groupby(df, :a);
  0.504278 seconds (840.73 k allocations: 42.139 MiB, 3.94% gc time)

julia> @time combine(gdf, nrow);
  0.943988 seconds (1.80 M allocations: 92.772 MiB, 2.72% gc time)

julia> @time combine(gdf, :b => sum);
  0.185167 seconds (642.57 k allocations: 33.666 MiB, 4.39% gc time)

julia> @time transform(gdf);
  0.751447 seconds (2.82 M allocations: 148.360 MiB, 3.53% gc time)
```

### This PR:

```
julia> @time using DataFrames;
  0.829970 seconds (916.92 k allocations: 63.014 MiB)

julia> @time df = DataFrame(:a => [1,2], :b => [3,4]);
  0.211642 seconds (157.83 k allocations: 8.434 MiB)

julia> @time gdf = groupby(df, :a);
  0.449700 seconds (459.70 k allocations: 23.189 MiB, 14.18% gc time)

julia> @time combine(gdf, nrow);
  0.944649 seconds (1.79 M allocations: 92.054 MiB, 2.79% gc time)

julia> @time combine(gdf, :b => sum);
  0.186040 seconds (642.18 k allocations: 33.630 MiB, 4.17% gc time)

julia> @time transform(gdf);
  0.781098 seconds (2.77 M allocations: 145.770 MiB, 3.76% gc time)
```
except for package loading time (which was expected to decrease).

@timholy - where would you expect the biggest gains. Thank you!

CC @nalimilan

timholy:
First, CategoricalArrays no longer invalidates much stuff, so there may be no urgency in getting rid of it. Version 0.9 only invalidates 212 methods, and most of these are probably inconsequential.

Second, when you do invalidate stuff, it's often the things that build on you, not you yourself, that are the big winners. For example, I bet that if Gadfly added a precompile file (using the tricks in SnoopCompile) they'd now have vastly more success.
Basically, what happens is that if loading PkgA invalidates method1, but PkgA doesn't need/user method1, it has literally no impact on PkgA. However, if PkgB uses both PkgA and method1, then PkgB might get hammered by depending on PkgA. If PkgA stops invalidating method1, PkgB wins even if it doesn't help PkgA at all.

bkamins:
Thank you for the explanation.

> so there may be no urgency in getting rid of it.

As you have commented on Slack - we have already removed the dependency in 0.22 release. Here we only remove the actual `using` statement essentially as we needed one round of releases before doing such a change due to our deprecation policy.

Still - even if CategoricalArrays.jl is much more compiler friendly now (which I agree it is) it is better anyway to have no dependence between the packages and make them interoperate via DataAPI.jl interface.

src/deprecated.jl
```diff
@@ -1,147 +1,10 @@
...
-                     copycols::Bool=true) where {N} DataFrame(collect(columns), [Symbol(c) for c in cnames];
-                                                              makeunique=makeunique, copycols=copycols)
-@deprecate DataFrame(columns::NTuple{N, AbstractVector};
-                     copycols::Bool=true) where {N} DataFrame(collect(columns),
-                                                              Symbol.(:x, 1:length(columns)), copycols=copycols)
-
-# this deprecation is very important, becuase without it users will
-# get strange results with old code as described in https://github.com/JuliaData/Tables.jl/issues/208
-@deprecate DataFrame(columns::AbstractVector{<:AbstractVector}; makeunique::Bool=false,
-                     copycols::Bool=true) DataFrame(columns, :auto, copycols=copycols)
```
nalimilan, line 124: As usual I'd be tempted to keep as many deprecations as possible for some time, turning them into errors. That shouldn't be too costly and that will help people who haven't noticed the deprecations introduced in 0.22. That's likely to be the case for many people as you usually don't use strict version bounds when developing interactive scripts (as opposed to packages).

(Of course that doesn't apply to `categorical` since it would conflict with CategoricalArrays.)

NEWS.md
```diff
@@ -1,3 +1,28 @@
+# DataFrames v1.0 Release Notes
+
+## Breaking changes
+
+* No breaking changes are planned for v1.0 release
+
+## Bug fixes
+
+* fixed bug when displaying floating point columns with eltype turned off in
```
nalimilan, line 9: Shouldn't we make a 0.22.1 release with small fixes like this before moving on to 1.0? It will take some time until packages and users have tested 0.22 to make sure we're totally OK with the API, and during that period it would be nice to be able to fix small bugs.

bkamins: We will make 0.22.1 as a separate branch into which I will cherry-pick the commits that should go there. This is what we did in 0.21 release.

bkamins: But I thought that making 0.22.1 only with this minor printing fix was not justified

src/deprecated.jl
bkamins: OK - this essentially applies to `DataFrame` constructors. I will re-introduce them.

NEWS.md
nalimilan: As you like.

bkamins: Actually I came to the conclusion that after merging #2553 we can make a patch release with everything merged to that point.

src/dataframe/dataframe.jl
```diff
@@ -337,6 +337,30 @@ function DataFrame(columns::AbstractMatrix, cnames::Symbol)
     return DataFrame(columns, gennames(size(columns, 2)), makeunique=false)
 end
 
+# Discontinued constructors
```
bkamins, line 340: I have reverted the constructors that "made sense" and were common-enough.

The left out tuple-constructors now are:

```
julia> DataFrame((:a => [1,2], :b => [3,4]))
2×2 DataFrame
 Row │ first   second 
     │ Symbol  Array… 
─────┼────────────────
   1 │ a       [1, 2]
   2 │ b       [3, 4]

julia> DataFrame(([1,2], [3,4]))
ERROR: ArgumentError: 'Tuple{Array{Int64,1},Array{Int64,1}}' iterates 'Array{Int64,1}' values, which doesn't satisfy the Tables.jl `AbstractRow` interface
```
but I do not think anyone used them.

src/dataframe/dataframe.jl
```diff
@@ -337,6 +337,30 @@ function DataFrame(columns::AbstractMatrix, cnames::Symbol)
...
+
+DataFrame(matrix::Matrix) =
+    throw(ArgumentError("`DataFrame` constructor from a `Matrix` requires " *
+                        "passing :auto as a second argument to automatically " *
+                        "generate column names: `DataFrame(matrix, :auto)`"))
+
+DataFrame(matrix::Vector{<:AbstractVector}) =
+    throw(ArgumentError("`DataFrame` constructor from a `Vector` of vectors requires " *
+                        "passing :auto as a second argument to automatically " *
+                        "generate column names: `DataFrame(matrix, :auto)`"))
```
nalimilan, line 350:
```suggestion
DataFrame(vecs::AbstractVector{<:AbstractVector}) =
    throw(ArgumentError("`DataFrame` constructor from a `Vector` of vectors requires " *
                        "passing :auto as a second argument to automatically " *
                        "generate column names: `DataFrame(vecs, :auto)`"))
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -556,19 +556,14 @@ julia> describe(df, :min, sum => :sum, cols=:x)
...
-end
+DataAPI.describe(df::AbstractDataFrame,
+                 stats::Union{Symbol, Pair{<:Base.Callable, <:SymbolOrString}}...;
+                 cols=:) =
+    _describe(select(df, cols, copycols=false), Any[s for s in stats])
+
 DataAPI.describe(df::AbstractDataFrame; cols=:) =
     _describe(select(df, cols, copycols=false),
-              [:mean, :min, :median, :max, :nmissing, :eltype])
+              Any[:mean, :min, :median, :max, :nmissing, :eltype])
```
nalimilan, line 566: Is `Any`  really better? Users are quite likely to pass `Vector{Symbol}` if they select some stats.

NEWS.md
```diff
@@ -1,3 +1,28 @@
...
+  text/plain ([2542](https://github.com/JuliaData/DataFrames.jl/pull/2542))
+
+## New functionalities
+
+
+## Deprecated
+
+* all old deprecations now throw an error
+  ([]())
```
nalimilan, line 18: Need to update this.

src/dataframe/dataframe.jl
nalimilan: Any idea why Codecov thinks this one isn't covered?

bkamins: I will check as I added tests for this. However, note that it must be a `Vector` not `AbstractVector` (as `AbstractVector`s are Tables.jl tables sometimes - just like above I use `Matrix` not `AbstractMatrix` for the same reason).

src/abstractdataframe/abstractdataframe.jl
bkamins: I just wanted to be consistent with the other method (as it passes `Any`) - so that we do not recompile `_describe`. but it is not a big deal.

NEWS.md
bkamins: sure - but I can update this only after creation of the PR (so this is a chicken and egg issue) :smile:. Still - thank you for spotting this.

bkamins, line 18:
```suggestion
  ([#2554](https://github.com/JuliaData/DataFrames.jl/pull/2554))
```

src/abstractdataframe/abstractdataframe.jl
nalimilan: Ah OK. Not sure what's best then.

bkamins: it probably does not matter much :smile:

src/dataframe/dataframe.jl
bkamins: this is tested

src/abstractdataframe/abstractdataframe.jl
bkamins: fixed

NEWS.md
bkamins: fixed

src/abstractdataframe/abstractdataframe.jl
bkamins, line 566:
```suggestion
              [:mean, :min, :median, :max, :nmissing, :eltype])
```

bkamins:
@nalimilan - this should be good for a review

src/deprecated.jl
```diff
@@ -1,147 +1,10 @@
...
 
-@deprecate DataFrame!(args...; kwargs...) DataFrame(args...; copycols=false, kwargs...)
+# import Base: @deprecate
 
 # TODO: remove these definitions in year 2021
 by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataFrames.jl. " *
                                              "Use the `combine(groupby(...), ...)` or `combine(f, groupby(...))` instead."))
 
 aggregate(args...; kwargs...) = throw(ArgumentError("aggregate function was removed from DataFrames.jl. " *
                                                     "Use the `combine` function instead."))
```
nalimilan, line 10: Are these still tested?

NEWS.md
nalimilan: So at any rate this bug won't appear in 1.0 release notes since it will have been fixed in 0.22.1, right?

src/deprecated.jl
bkamins: no - I have re-introduded the tests

NEWS.md
bkamins: OK - removed

bkamins:
updated after the review

bkamins:
@nalimilan - I have just checked that we have a significant regression on Julia nightly (the timings are the for the same operations as ones above):
```
julia> @time using DataFrames
  0.288947 seconds (747.34 k allocations: 52.368 MiB, 14.73% compilation time)

julia> @time df = DataFrame(:a => [1,2], :b => [3,4]);
  0.220673 seconds (266.01 k allocations: 16.396 MiB, 5.58% gc time, 99.95% compilation time)

julia> @time gdf = groupby(df, :a);
  1.435309 seconds (6.04 M allocations: 373.123 MiB, 5.68% gc time, 99.99% compilation time)

julia> @time combine(gdf, nrow);
  3.053012 seconds (9.99 M allocations: 610.595 MiB, 7.72% gc time, 99.96% compilation time)

julia> @time combine(gdf, :b => sum);
  0.305407 seconds (542.59 k allocations: 34.092 MiB, 3.56% gc time, 99.79% compilation time)

julia> @time transform(gdf);
  0.173324 seconds (241.60 k allocations: 15.545 MiB, 5.98% gc time, 99.43% compilation time)
```

nalimilan:
Wow, with 99% spent in compilation time, that's a serious regression in Julia. Can you file a bug there?

nalimilan (approved):

bkamins:
Thank you!

simeonschaub:
Is there a reason `CategoricalArrays` is still listed as a dependency in the `Project.jl` file? IIUC, it should now only be a test dependency.

bkamins:
This must have been a mistake when rebasing PR. It should be only test dependency. I will fix it.

bkamins:
see https://github.com/JuliaData/DataFrames.jl/pull/2646 (as I am not 100% certain about Project.toml semantics)

PR #2559, opened 2020-11-24
Improve inference in `vcat`

timholy:
Master:
```julia
julia> @time include("cat.jl")
Test Summary: | Pass  Total
...
 13.921540 seconds (30.55 M allocations: 1.835 GiB, 3.81% gc time, 97.72% compilation time)
```

This PR:
```julia
 12.296551 seconds (28.88 M allocations: 1.728 GiB, 3.05% gc time, 97.42% compilation time)
```

This is of course from a fresh session in both cases, and all the gains are to inference time on first use.

The idea is that improvements in inference quality let your precompile statements do you more good because they reach deeper into the call stack. It may be helpful to outline some of the principles that guide these changes:
- the change to `reduce` tries to "standardize" the types that get passed to `_vcat`; any `DF` that is fully concrete can be specialized, otherwise we standardize on `AbstractDataFrame` (which is already used in some places) as the fallback. This reduces the number of times you have to infer `_vcat` for abstract types.
- `_vcat` had some inference problems from the infamous julia#15276
- methods that get called for `AbstractDataFrame` are likely to return poorly-inferred types (they may be well-inferred for any concrete subtype, but poorly inferred for the abstract type). However, in some cases you can fix them, as exemplified by annotating the return type of `names`. This stabilizes a lot of `_vcat`'s body when `dfs::Vector{AbstractDataFrame}`.

There is one part of `_vcat` that is still poorly inferred: https://github.com/JuliaData/DataFrames.jl/blob/8645651e30785fbedca82a9f125d21d1f27a726e/src/abstractdataframe/abstractdataframe.jl#L1617-L1623

Basically anything that uses `newcols` is uninferrable. I didn't know enough to fix it. However, if that `Iterators.repeated` is just going to be expanded to a `Vector{Missing}` anyway (I'm not sure it will), then just writing it in terms of `fill` instead might fix it. Another option would be to annotate the type of `newcols` itself, or to annotate `lens` as a `Vector{Int}`. However, I expect these to be relatively minor gains.

I discovered these through a combination of `@snoopi`,`@snoopi_deep` (which still needs documentation), and of course Cthulhu (which sadly seems to have broken on 1.6 in the last few days, see https://github.com/JuliaDebug/Cthulhu.jl/issues/100. For the moment one has to diagnose problems on 1.5 and analyze the impact of improvements on 1.6).

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1552,16 +1552,10 @@ function Base.reduce(::typeof(vcat),
                                 Tuple{Vararg{AbstractDataFrame}}};
                      cols::Union{Symbol, AbstractVector{Symbol},
                      AbstractVector{<:AbstractString}}=:setequal)
-    isempty(dfs) && return _vcat(AbstractDataFrame[]; cols=cols)
-    DF = eltype(dfs)
-    if isconcretetype(DF)
-        return _vcat(DF[df for df in dfs if ncol(df) != 0]; cols=cols)
-    else
-        return _vcat(AbstractDataFrame[df for df in dfs if ncol(df) != 0]; cols=cols)
-    end
+    return _vcat(AbstractDataFrame[df for df in dfs if ncol(df) != 0]; cols=cols)
 end
 
-function _vcat(dfs::AbstractVector{<:AbstractDataFrame};
+function _vcat(dfs::AbstractVector{AbstractDataFrame};
                cols::Union{Symbol, AbstractVector{Symbol},
                            AbstractVector{<:AbstractString}}=:setequal)
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1547,16 +1547,24 @@ Base.vcat(dfs::AbstractDataFrame...;
                       AbstractVector{<:AbstractString}}=:setequal) =
     reduce(vcat, dfs; cols=cols)
 
-function Base.reduce(::typeof(vcat),
-            dfs::Union{AbstractVector{DF},
-                       Tuple{Vararg{DF}}};
-            cols::Union{Symbol, AbstractVector{Symbol},
-                        AbstractVector{<:AbstractString}}=:setequal) where DF<:AbstractDataFrame
-    if @isdefined(DF) && isconcretetype(DF)
-        return _vcat(DF[df for df in dfs if ncol(df) != 0]; cols=cols)
-    else
-        return _vcat(AbstractDataFrame[df for df in dfs if ncol(df) != 0]; cols=cols)
+if Base.VERSION >= v"1.5"
+    function Base.reduce(::typeof(vcat),
+                         dfs::Union{AbstractVector{DF},Tuple{Vararg{DF}}};
+                         cols::Union{Symbol, AbstractVector{Symbol},
+                                     AbstractVector{<:AbstractString}}=:setequal) where DF<:AbstractDataFrame
+        if @isdefined(DF) && isconcretetype(DF)
+            return _vcat(DF[df for df in dfs if ncol(df) != 0]; cols=cols)
+        else
+            return _vcat(AbstractDataFrame[df for df in dfs if ncol(df) != 0]; cols=cols)
+        end
     end
+else
+    Base.reduce(::typeof(vcat),
+                dfs::Union{AbstractVector{<:AbstractDataFrame},
+                           Tuple{Vararg{AbstractDataFrame}}};
+                cols::Union{Symbol, AbstractVector{Symbol},
+                AbstractVector{<:AbstractString}}=:setequal) =
+        _vcat([df for df in dfs if ncol(df) != 0]; cols=cols)
 end
 
 function _vcat(dfs::AbstractVector{<:AbstractDataFrame};
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1547,12 +1547,17 @@ Base.vcat(dfs::AbstractDataFrame...;
                       AbstractVector{<:AbstractString}}=:setequal) =
     reduce(vcat, dfs; cols=cols)
 
-Base.reduce(::typeof(vcat),
-            dfs::Union{AbstractVector{<:AbstractDataFrame},
-                       Tuple{Vararg{AbstractDataFrame}}};
+function Base.reduce(::typeof(vcat),
+            dfs::Union{AbstractVector{DF},
+                       Tuple{Vararg{DF}}};
             cols::Union{Symbol, AbstractVector{Symbol},
-                        AbstractVector{<:AbstractString}}=:setequal) =
-    _vcat([df for df in dfs if ncol(df) != 0]; cols=cols)
+                        AbstractVector{<:AbstractString}}=:setequal) where DF<:AbstractDataFrame
+    if @isdefined(DF) && isconcretetype(DF)
+        return _vcat(DF[df for df in dfs if ncol(df) != 0]; cols=cols)
+    else
+        return _vcat(AbstractDataFrame[df for df in dfs if ncol(df) != 0]; cols=cols)
+    end
+end
 
 function _vcat(dfs::AbstractVector{<:AbstractDataFrame};
                cols::Union{Symbol, AbstractVector{Symbol},
@@ -1586,13 +1591,15 @@ function _vcat(dfs::AbstractVector{<:AbstractDataFrame};
 
         if !isempty(coldiff)
             # if any DataFrames are a full superset of names, skip them
-            filter!(u -> !issetequal(u, header), uniqueheaders)
+            let header=header     # julia #15276
+                filter!(u -> !issetequal(u, header), uniqueheaders)
+            end
             estrings = map(enumerate(uniqueheaders)) do (i, head)
                 matching = findall(h -> head == h, allheaders)
                 headerdiff = setdiff(coldiff, head)
-                cols = join(headerdiff, ", ", " and ")
+                badcols = join(headerdiff, ", ", " and ")
                 args = join(matching, ", ", " and ")
-                return "column(s) $cols are missing from argument(s) $args"
+                return "column(s) $badcols are missing from argument(s) $args"
             end
             throw(ArgumentError(join(estrings, ", ", ", and ")))
         end
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1547,24 +1547,18 @@ Base.vcat(dfs::AbstractDataFrame...;
                       AbstractVector{<:AbstractString}}=:setequal) =
     reduce(vcat, dfs; cols=cols)
 
-if Base.VERSION >= v"1.5"
-    function Base.reduce(::typeof(vcat),
-                         dfs::Union{AbstractVector{DF},Tuple{Vararg{DF}}};
-                         cols::Union{Symbol, AbstractVector{Symbol},
-                                     AbstractVector{<:AbstractString}}=:setequal) where DF<:AbstractDataFrame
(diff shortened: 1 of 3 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1547,12 +1547,17 @@ Base.vcat(dfs::AbstractDataFrame...;
...
-            dfs::Union{AbstractVector{<:AbstractDataFrame},
-                       Tuple{Vararg{AbstractDataFrame}}};
+function Base.reduce(::typeof(vcat),
+            dfs::Union{AbstractVector{DF},
+                       Tuple{Vararg{DF}}};
             cols::Union{Symbol, AbstractVector{Symbol},
-                        AbstractVector{<:AbstractString}}=:setequal) =
-    _vcat([df for df in dfs if ncol(df) != 0]; cols=cols)
+                        AbstractVector{<:AbstractString}}=:setequal) where DF<:AbstractDataFrame
+    if @isdefined(DF) && isconcretetype(DF)
```
bkamins, line 1555: why calling `@isfefined(DF)` is required here?

src/other/index.jl
```diff
@@ -422,7 +422,7 @@ Base.@propagate_inbounds SubIndex(parent::AbstractIndex, cols) =
     SubIndex(parent, parent[cols])
 
 Base.length(x::SubIndex) = length(x.cols)
-Base.names(x::SubIndex) = string.(_names(x))
+Base.names(x::SubIndex) = string.(_names(x))::Vector{String}
```
bkamins, line 425: do you know why Julia fails to infer this without an annotation?

src/abstractdataframe/abstractdataframe.jl
timholy: Tuple inputs with 2 or more different types.

src/other/index.jl
timholy: SubIndex has `I<:AbstractIndex, S<:AbstractVector{Int}, T<:AbstractVector{Int}` as type parameters. When it doesn't know what `I, S, T` are (which is what happens when it's a `Vector{AbstractDataFrame}`, inference knows almost nothing of value. In principle an `AbstractIndex` could do *anything*. You/we could consider adding an interface: https://timholy.github.io/SnoopCompile.jl/stable/snoopr/#Inferrable-field-access-for-abstract-types-1.

timholy:
That's interesting: versions older than Julia 1.5 don't dispatch to the new version of `reduce`. I pushed a commit restricting that change to versions 1.5 and above.

src/abstractdataframe/abstractdataframe.jl
bkamins: But then `isconcretetype(DF)` should return `false` - right? (or would it error)

timholy: If I take the `@isdefined` check out and preface that conditional with a `@show typeof(dfs)`, here's what I get from running the `cat.jl` test:
```julia
julia> include("cat.jl")
Test Summary: | Pass  Total
hcat          |   10     10
Test Summary: | Pass  Total
hcat: copying |   26     26
Test Summary:            | Pass  Total
hcat ::AbstractDataFrame |    2      2
Test Summary:            | Pass  Total
hcat ::AbstractDataFrame |    2      2
Test Summary:          | Pass  Total
hcat ::AbstractVectors |   11     11
Test Summary:  | Pass  Total
hcat: copycols |   76     76
typeof(dfs) = Tuple{DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = NTuple{4, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.DataFrame, DataFrames.DataFrame}
typeof(dfs) = Tuple{DataFrames.SubDataFrame{DataFrames.DataFrame, DataFrames.Index, UnitRange{Int64}}, DataFrames.SubDataFrame{DataFrames.DataFrame, DataFrames.SubIndex{DataFrames.Index, Vector{Int64}, Vector{Int64}}, UnitRange{Int64}}}
vcat: Error During Test at /home/tim/.julia/dev/DataFrames/test/cat.jl:269
  Test threw exception
  Expression: vcat(view(df, 1:2, :), view(df, 3:5, [3, 2, 1])) == df
  UndefVarError: DF not defined
  Stacktrace:
   [1] reduce(::typeof(vcat), dfs::Tuple{DataFrames.SubDataFrame{DataFrames.DataFrame, DataFrames.Index, UnitRange{Int64}}, DataFrames.SubDataFrame{DataFrames.DataFrame, DataFrames.SubIndex{DataFrames.Index, Vector{Int64}, Vector{Int64}}, UnitRange{Int64}}}; cols::Symbol)
     @ DataFrames ~/.julia/dev/DataFrames/src/abstractdataframe/abstractdataframe.jl:1556
...
```
So as soon as it gets a heterogeneous tuple, it throws an `UndefVarError`.

bkamins: But what I do not understand is the fact that you use `DF <: AbstractDataFrame` (instead of the original implementation). And if I try to do a MWE I get:
```
julia> using DataFrames

julia> f(::Tuple{Vararg{DF}}) where {DF<:AbstractDataFrame} = "matched"

julia> df = DataFrame(a=1,b=2)
1×2 DataFrame
 Row │ a      b
     │ Int64  Int64
─────┼──────────────
   1 │     1      2

julia> df1 = @view df[:, :]
1×2 SubDataFrame
 Row │ a      b
     │ Int64  Int64
─────┼──────────────
   1 │     1      2

julia> df2 = @view df[:, 1:1]
1×1 SubDataFrame
 Row │ a
     │ Int64
─────┼───────
   1 │     1

julia> f((df, df))
"matched"

julia> f((df1, df1))
"matched"

julia> f((df2, df2))
"matched"

julia> f((df, df1))
ERROR: MethodError: no method matching f(::Tuple{DataFrame,SubDataFrame{DataFrame,DataFrames.Index,Base.OneTo{Int64}}})

julia> f((df1, df2))
ERROR: MethodError: no method matching f(::Tuple{SubDataFrame{DataFrame,DataFrames.Index,Base.OneTo{Int64}},SubDataFrame{DataFrame,DataFrames.SubIndex{DataFrames.Index,UnitRange{Int64},UnitRange{Int64}},Base.OneTo{Int64}}})

julia> h(::Tuple{Vararg{AbstractDataFrame}}) = "matched"
h (generic function with 1 method)

julia> h((df, df))
"matched"

julia> h((df, df1))
"matched"

julia> h((df2, df1))
"matched"
```
so it seems that the pattern `Tuple{Vararg{DF}} where DF <: AbstractDataFrame` is not the same as the original and intended `Tuple{Vararg{AbstractDataFrame}}`

timholy: Yes, this seems to exploit a relatively recent & subtle change in how type parameters. Basically the `Tuple{Vararg{DF}}` is requiring a single concrete value to *set* `DF`, but signature *matching* is the same as `<:AbstractDataFrame`.

Another way we could do this is nominal-runtime:
```julia
DF = typeof(first(dfs))
if isconcretetype(DF) && all(df->typeof(df) === DF, dfs) ... end
```
Do you like that better?

My workflow for these things isn't very deliberative; I typically write it first without the `@isdefined`, and then when I get an error like the one above I add `@isdefined`. If you `grep` this in `Base` you'll see quite a few similar uses.

timholy: Oh dear, now I see your point with the two different `df`s. Very interesting. I'll do the runtime version perhaps.

bkamins: I would separate `AbstractVector` and `Tuple` implementations then.
Also for `Tuple` please make sure that in the run-time version a case when the tuple is empty is correctly handled.

I have now noticed a subtle bug that should be fixed:
```
julia> reduce(vcat, ())

julia> using DataFrames

julia> reduce(vcat, ())
0×0 DataFrame
```

To be decided if we keep:
```
julia> reduce(vcat, AbstractDataFrame[])
0×0 DataFrame
```
or also error here.

timholy: That is weird. It does match *in situ*, even though your example fails for me. :confused:

bkamins: can you show me the output of the same commands in the fresh Julia session (and what Julia are you using I am on 1.5.3). Thank you! (extending functions from Base is hard)

nalimilan (approved):
Thanks!

bkamins (changes requested):
We still have the problem:
```
julia> reduce(vcat, ())
0×0 DataFrame
```
in this branch

timholy:
An even simpler approach seems to work pretty well. I was worried about the lack of specialization hurting performance, but given that a single return-type annotations fixs inference on `AbstractDataFrame` for this function, that proves unfounded. This branch:
```julia

julia> df = DataFrame(a=1,b=2);

julia> df1 = @view df[:, :];

julia> df2 = @view df[:, 1:1];

julia> cols = [:a];

julia> @btime vcat($df1, $df2; cols=$cols);
  4.527 μs (88 allocations: 6.69 KiB)

julia> @btime vcat($df, $df);
  5.719 μs (79 allocations: 5.98 KiB)
```
vs 0.22.1:
```julia
julia> @btime vcat($df1, $df2; cols=$cols);
  5.697 μs (97 allocations: 6.78 KiB)

julia> @btime vcat($df, $df);
  7.484 μs (88 allocations: 6.39 KiB)
```

And this implementation gives approximately the same speedup of running the `cat.jl` tests (which is much more about compile-time latency than runtime performance).

I threw in a couple of changes to signatures that I noticed during other perusals, though I can strip them out if you'd prefer:
- `AsTable` is concrete, so `<:AsTable` is the same as `AsTable`
- `Vector{Any}` is actually a fully-specialized type (its elements are not, but the container is). Hence `@nospecialize(v::Vector{Any})` doesn't do anything.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1547,14 +1547,15 @@ Base.vcat(dfs::AbstractDataFrame...;
...
 
-Base.reduce(::typeof(vcat),
-            dfs::Union{AbstractVector{<:AbstractDataFrame},
-                       Tuple{Vararg{AbstractDataFrame}}};
-            cols::Union{Symbol, AbstractVector{Symbol},
-                        AbstractVector{<:AbstractString}}=:setequal) =
-    _vcat([df for df in dfs if ncol(df) != 0]; cols=cols)
+function Base.reduce(::typeof(vcat),
+                     dfs::Union{AbstractVector{<:AbstractDataFrame},
+                                Tuple{Vararg{AbstractDataFrame}}};
```
bkamins, line 1552: I think this should ensure that we do not overwrite the implementation from base for `reduce(vcat, ())`:
```suggestion
                                Tuple{AbstractDataFrame, Vararg{AbstractDataFrame}}};
```
(for vectors it is OK to accept 0-length vectors as they have `eltype`)

bkamins (approved):

bkamins:
Thank you! I will add tests in a separate PR.

PR #2563, opened 2020-11-28
Split methods to reduce compile cost of argument diversity

timholy:
The cost of compilation is very roughly proportional to the product
"method complexity" * "number of specializations." Reducing either
factor can therefore reduce compile latency. This splits five methods:

- _combine_with_first
- _combine_rows_with_first
- _combine_tables_with_first
- groupreduce!
- _combine_prepare

into two pieces, one of which receives lower argument diversity and thus
has fewer specializations. The remaining piece is therefore smaller
and thus has lower "method complexity." Together these changes reduce
the time needed to run the "grouping.jl" test file by about 15%,
from 283s to 248s.

There might be even better ways to do this (e.g., refactor so that
argument diversity is reduced at an early stage in the call chain),
but as a neophyte unfamiliar with the package internals this seemed
like a safe and easy way to make improvements, with the only cost being
a slight "uglification."

--------------------

This is one of the steps (#2559 was another) in a [blog post](https://github.com/JuliaLang/www.julialang.org/pull/1093) I'm working on describing new tools in Julia 1.6 and SnoopCompile to diagnose and address sources of compiler latency (direct link to preview: https://julialang.netlify.app/previews/pr1093/blog/2020/12/package_latency/). This is different from, but complementary to, the invalidations work. I hope you don't mind being used as a guinea pig; some of the reasons why I chose DataFrames are described in the blog.

There will be one or two more PRs that I'll submit later.

### DIFF
--- a/src/groupeddataframe/complextransforms.jl
+++ b/src/groupeddataframe/complextransforms.jl
@@ -22,6 +22,23 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
                              f::Base.Callable, gd::GroupedDataFrame,
                              incols::Union{Nothing, AbstractVector, Tuple, NamedTuple},
                              firstmulticol::Val, idx_agg::Union{Nothing, AbstractVector{<:Integer}})
+    first, n, eltys, extrude, idx, initialcols, targetcolnames = _combine_with_first_prelude(first, f, gd, firstmulticol, idx_agg)
+    if !extrude && first isa Union{AbstractDataFrame,
+                                   NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}}
+        outcols, finalcolnames = _combine_tables_with_first!(first, initialcols, idx, 1, 1,
+                                                             f, gd, incols, targetcolnames,
+                                                             firstmulticol)
+    else
+        outcols, finalcolnames = _combine_rows_with_first!(first, initialcols, 1, 1,
+                                                           f, gd, incols, targetcolnames,
+                                                           firstmulticol)
+    end
+    return idx, outcols, collect(Symbol, finalcolnames)
+end
+
+function _combine_with_first_prelude(first::Union{NamedTuple, DataFrameRow, AbstractDataFrame},
+                                     f::Base.Callable, gd::GroupedDataFrame,
+                                     firstmulticol::Val, idx_agg::Union{Nothing, AbstractVector{<:Integer}})
     extrude = false
 
     if first isa AbstractDataFrame
@@ -51,17 +68,8 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
         initialcols = ntuple(i -> Tables.allocatecolumn(eltys[i], n), _ncol(first))
     end
     targetcolnames = tuple(propertynames(first)...)
-    if !extrude && first isa Union{AbstractDataFrame,
-                                   NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}}
-        outcols, finalcolnames = _combine_tables_with_first!(first, initialcols, idx, 1, 1,
-                                                             f, gd, incols, targetcolnames,
-                                                             firstmulticol)
-    else
-        outcols, finalcolnames = _combine_rows_with_first!(first, initialcols, 1, 1,
-                                                           f, gd, incols, targetcolnames,
-                                                           firstmulticol)
-    end
-    return idx, outcols, collect(Symbol, finalcolnames)
+
+    return first, n, eltys, extrude, idx, initialcols, targetcolnames
 end
 
 function fill_row!(row, outcols::NTuple{N, AbstractVector},
@@ -110,7 +118,20 @@ function _combine_rows_with_first!(first::Union{NamedTuple, DataFrameRow},
     # Handle first group
     j = fill_row!(first, outcols, rowstart, colstart, colnames)
     @assert j === nothing # eltype is guaranteed to match
-    # Handle remaining groups
+    # Handle remaining groups (eliminate the argument type-diversity of `first`)
+    _combine_rows_with_first(len, outcols, rowstart, f, gd, incols, colnames, firstmulticol)
+end
+
+function _combine_rows_with_first(len::Int,
+                                  outcols::NTuple{N, AbstractVector},
+                                  rowstart::Integer,
+                                  f::Base.Callable, gd::GroupedDataFrame,
+                                  incols::Union{Nothing, AbstractVector, Tuple, NamedTuple},
+                                  colnames::NTuple{N, Symbol},
+                                  firstmulticol::Val) where N
+    gdidx = gd.idx
+    starts = gd.starts
+    ends = gd.ends
     @inbounds for i in rowstart+1:len
         row = wrap_row(do_call(f, gdidx, starts, ends, gd, incols, i), firstmulticol)
         j = fill_row!(row, outcols, i, 1, colnames)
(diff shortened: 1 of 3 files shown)


timholy:
As I say in the blog post, you don't have to accept this PR: it's just one approach for reducing compile-time latency, but you may be able to think of much better ways to go about it. Really, the most useful part of this is the detective work identifying which methods have the highest cost, and why.

bkamins:
Thank you for working on this. Indeed, in the past we were mostly blind-guessing what changes to method splitting or `@nospecialize` annotations could help. Your work on tooling allowing to dig into these things is fantastic.

src/groupeddataframe/complextransforms.jl
```diff
@@ -110,7 +118,20 @@ function _combine_rows_with_first!(first::Union{NamedTuple, DataFrameRow},
     # Handle first group
     j = fill_row!(first, outcols, rowstart, colstart, colnames)
     @assert j === nothing # eltype is guaranteed to match
-    # Handle remaining groups
+    # Handle remaining groups (eliminate the argument type-diversity of `first`)
```
bkamins, line 121: lines https://github.com/JuliaData/DataFrames.jl/pull/2563/files#diff-fe9f0d31012cded8c66d568e600c72b8b7a55ad5e4826bf645d885c7aed0a76eR111-R113 above can be removed as they are not used in this method now.

src/groupeddataframe/complextransforms.jl
```diff
@@ -110,7 +118,20 @@ function _combine_rows_with_first!(first::Union{NamedTuple, DataFrameRow},
...
+    _combine_rows_with_first(len, outcols, rowstart, f, gd, incols, colnames, firstmulticol)
```
bkamins, line 122:
```suggestion
    return _combine_rows_with_first(len, outcols, rowstart, f, gd, incols, colnames, firstmulticol)
```

src/groupeddataframe/complextransforms.jl
```diff
@@ -110,7 +118,20 @@ function _combine_rows_with_first!(first::Union{NamedTuple, DataFrameRow},
...
+end
+
+function _combine_rows_with_first(len::Int,
```
bkamins, line 125: @nalimilan - are you OK with this function name? Maybe `_combine_rows_with_first_filled!` would be better?

src/groupeddataframe/complextransforms.jl
```diff
@@ -188,16 +209,27 @@ function _combine_tables_with_first!(first::Union{AbstractDataFrame,
...
         j = append_rows!(first, outcols, colstart, colnames)
         @assert j === nothing # eltype is guaranteed to match
         append!(idx, Iterators.repeated(gdidx[starts[rowstart]], _nrow(first)))
     end
-    # Handle remaining groups
+    # Handle remaining groups (eliminate the argument type-diversity of `first`)
+    return _combine_tables_with_first(len, outcols, idx, rowstart, f, gd, incols, colnames, firstmulticol)
+end
+
+function _combine_tables_with_first(len::Int, outcols::NTuple{N, AbstractVector},
```
bkamins, line 224: @nalimilan - same here maybe `_combine_tables_with_first_filled!`?

src/groupeddataframe/fastaggregates.jl
```diff
@@ -193,6 +194,12 @@ function groupreduce!(res::AbstractVector, f, op, condf, adjust, checkempty::Boo
             end
         end
     end
+    # Eliminate the argument-type diversity that we no longer need
+    return _groupreduce!(res, counts, n, op === min || op === max, adjust, checkempty, incol)
+end
+
+function _groupreduce!(res::AbstractVector, counts::Union{Nothing,Vector{Int}}, n::Int, opminormax::Bool, adjust, checkempty::Bool,
+                      incol::AbstractVector)
```
nalimilan, line 202: If you just pass `op`, no specialization should happen on it since it's not called by the function, right? I'd find it clearer.

Also I'd use a more explicit name  like `groupreduce_fixres` or something.

src/groupeddataframe/fastaggregates.jl
```diff
@@ -159,6 +159,7 @@ end
 function groupreduce!(res::AbstractVector, f, op, condf, adjust, checkempty::Bool,
                       incol::AbstractVector, gd::GroupedDataFrame)
     n = length(gd)
+    counts = nothing
```
nalimilan, line 162: Have you checked that this doesn't affect performance? Normally I'd expect inference to work, but you never know...

src/groupeddataframe/fastaggregates.jl
```diff
@@ -193,6 +194,12 @@ function groupreduce!(res::AbstractVector, f, op, condf, adjust, checkempty::Boo
...
     end
+    # Eliminate the argument-type diversity that we no longer need
+    return _groupreduce!(res, counts, n, op === min || op === max, adjust, checkempty, incol)
+end
+
+function _groupreduce!(res::AbstractVector, counts::Union{Nothing,Vector{Int}}, n::Int, opminormax::Bool, adjust, checkempty::Bool,
+                      incol::AbstractVector)
     if adjust !== nothing
         res .= adjust.(res, counts)
     end
```
nalimilan, line 205: Putting this block and the next one in this method doesn't sound great, as `adjust` and `counts` are tightly linked to `op` (they are used for `mean` basically). If you keep them in the parent method this one won't have to be specialized on the particular operation.

nalimilan (commented):
Thanks @timholy!

The case of `_combine_with_first` is tricky. Indeed the types of `first` and `incols` do not vary independently in real life: when they are `NamedTuples` (which is the problem), the types are determined by the input and output column names. So avoiding recompiling code when the type of `first` changes but not that of `first` will only be useful when applying a different operation to input columns with the same names. The fact that it has a large effect on tests is probably due to their looping over functions and to our lack of imagination when naming columns. :-)

I guess it's still good to have, but I wonder whether we could avoid recompiling much more code by using `NamedTuples` less. The fact that `_combine_with_first` takes `incols` and `colnames` separately is precisely to avoid specializing on names. In some cases we pass a `Tuple` of vectors and a `Tuple` of names to it rather than a `NamedTuple` of vectors.  OTC in some cases (when `AsTable` is used) we pass a `NamedTuple`. @bkamins Maybe we could avoid that, by passing always two tuples, and reconstructing a `NamedTuple` only in the very last step (in `do_call`)?

timholy:
> I wonder whether we could avoid recompiling much more code by using NamedTuples less

I wondered the same thing. I'm sure you know the answer, but I even wondered whether passing column types as a `Vector{Any}` would make sense, simply because the types are not inferrable from `DataFrame` itself. Then a NamedTuple would become a `Vector{Symbol}` and a `Vector{Any}`, and you get no type-diversity.

bkamins:
@timholy - a question to you in respect of your time.

Usually (as you see) - we have a lot of discussions with @nalimilan even for relatively small PRs as this one. So my question is - do you prefer to work on this PR (which is welcome) and we together resolve the issues or you prefer us to "take over" the proposal (essentially meaning I would open an new PR and close this one).

To be clear - I would love to have your contribution reflected in the repository, but I can imagine that given the amount of Julia-related projects you are involved you might prefer us to just follow your suggestions and make the changes on our own.

Thank you once more for looking into DataFrames.jl!

timholy:
I'm happy to have you take it over. I am hoping that within 72 hours all the necessary changes will be merged in Julia and SnoopCompile, and then I'll update some of the text of the blog post. That way if you need to look at any of the data yourself, go right ahead.

I'll close this, but steal from it freely if anything here is useful. (Both of us probably expect the answer to be "no.")

bkamins:
Thank you! Actually reading your WIP blog-post already helped me a lot (i.e. having a better understanding how things work actually allows us to design the code better in the first place)

PR #2571, opened 2020-12-05
Make GroupedDataFrame show tests more robust to context

nalimilan:
`show` always says `GroupedDataFrame`, even when the type would be printed as `DataFrames.GroupedDataFrame`. We do the
same for `DataFrame`. So it makes no sense to use complex constructs such as `$(nameof(typeof(gd)))` nor `$GroupedDataFrame`. This fixes an error when calling `include("test/grouping.jl")` without `using DataFrames`.


bkamins (approved):
Do you understand why "nightly" fails?

nalimilan:
One of them should be fixed by https://github.com/JuliaData/DataFrames.jl/pull/2572. The one about `firstindex` is more intriguing. I've restarted CI on master to see whether it's also present there.

PR #2572, opened 2020-12-05
Fix Markdown in test

nalimilan:
Since JuliaLang/julia#38502, Markdown printed has changed slightly, making tests fail on nightly.


bkamins (approved):

test/io.jl
```diff
@@ -222,7 +222,7 @@ end
                            "  * γ∞2∫αγ∞3∫αγ∞4∫αγ∞5∫αγ∞6∫αγ∞7∫αγ∞8∫αγ∞9∫αγ∞0"),
         ]
     )
-    if VERSION > v"1.6.0"
+    if VERSION < v"1.6.0"
```
bkamins, line 225: :smile:

nalimilan: It's even worse than that. VScode had messed with the indentation, and I forgot adding two new chars before the ellipsis! One should never trying to fix things without checking...

bkamins: Looks good now!

PR #2573, opened 2020-12-06
implement firstindex, lastindex, axes, size, and ndims consistently

bkamins:
This makes sure we have a complete indexing API for all types we expose. I hope I have not missed anything.

@nalimilan - if you are OK with this approach I will add tests and make it ready for a review.


src/groupeddataframe/groupeddataframe.jl
```diff
@@ -380,7 +380,19 @@ function Base.iterate(gd::GroupedDataFrame, i=1)
     end
 end
 
-Compat.lastindex(gd::GroupedDataFrame) = gd.ngroups
```
nalimilan, line 383: Why did we need Compat before and not now?

nalimilan (commented):
Sure, looks good. Though if Base adds a fallback maybe we won't need all of these?

src/groupeddataframe/groupeddataframe.jl
bkamins: probably it was needed before Julia 1.0. I have checked that `lastindex` is defined in Julia 1.0.

bkamins:
> Though if Base adds a fallback maybe we won't need all of these?

We need them, because earlier versions of Julia will not have them so this way we maintain consistency and make sure DataFrames.jl codes work on all versions of Julia post 1.0.

---

Also the major decision is if we really want all those methods defined. What I mean is that in Julia Base you now have for example:
```
julia> size((1,2,3))
ERROR: MethodError: no method matching size(::Tuple{Int64,Int64,Int64})

julia> size((a=1,b=2,c=3))
ERROR: MethodError: no method matching size(::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}})

julia> axes((1,2,3))
(Base.OneTo(3),)

julia> axes((a=1,b=2,c=3))
ERROR: MethodError: no method matching size(::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}})
```

so the question is if we define them now for objects in DataFrames.jl - can we have some problems with consistency with Julia Base in the future.

@mbauman, @StefanKarpinski - could you please give some guidance when non-array types are allowed to safely `size` and `axes`? In particular the tricky case is types that allow indexing with multiple types of values (like `NamedTuple` in Julia Base - we have several such types in DataFrames.jl that allow indexing using column number but also column name for instance).

Thank you!

bkamins:
@nalimilan - this should be good for a review.

I have come to the following conclusion that for array interface (`axes` etc.) we should return always integer-based objects.

The issue is that Julia Base sometimes is not clear in distinction between using term "index" and "key" for `v` in the context of `collection[v]`. I recommend the following (and now consistently use it):
* "index" is always integer in DataFrames.jl
* "key" is always a natural key for the collection (so it can be an integer, but can be e.g. `Symbol` for column names)

Also I am strict with `size` and `axes` for invalid dimension unless the type is `AbstractVector` in which case too large dimension produces `1` and `OneTo(1)` respectively, as this is what Julia Base does.

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
     @test_throws ArgumentError df[1] = [2]
 end
 
+@testset "array interface tests for all types" begin
+    df = DataFrame(reshape(1:12, 3, 4), :auto)
+    @test_throws MethodError length(df)
+    @test ndims(df) == 2
```
bkamins, line 1865:
```suggestion
    @test ndims(df) == ndims(typeof(df)) == 2
```

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
...
+    @test_throws ArgumentError firstindex(df, 0)
+    @test_throws MethodError lastindex(df)
+    @test lastindex(df, 1) == 3
+    @test lastindex(df, 2) == 4
+    @test_throws ArgumentError lastindex(df, 3)
+    @test_throws ArgumentError lastindex(df, 0)
+
+    dfr = df[1, 1:3]
+    @test length(dfr) == 3
+    @test ndims(dfr) == 1
```
bkamins, line 1889:
```suggestion
    @test ndims(dfr) == ndims(typeof(dfr)) == 1
```

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
...
+    @test_throws BoundsError firstindex(dfr, 2)
+    @test_throws BoundsError firstindex(dfr, 0)
+    @test lastindex(dfr) == 3
+    @test lastindex(dfr, 1) == 3
+    @test_throws BoundsError lastindex(dfr, 2)
+    @test_throws BoundsError lastindex(dfr, 0)
+
+    er = eachrow(df)
+    @test length(er) == 3
+    @test ndims(er) == 1
```
bkamins, line 1909:
```suggestion
    @test ndims(er) == ndims(typeof(er)) == 1
```

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
...
+    @test firstindex(er, 2) == 1
+    @test_throws BoundsError firstindex(er, 0)
+    @test lastindex(er) == 3
+    @test lastindex(er, 1) == 3
+    @test lastindex(er, 2) == 1
+    @test_throws BoundsError lastindex(er, 0)
+
+    ec = eachcol(df)
+    @test length(ec) == 4
+    @test ndims(ec) == 1
```
bkamins, line 1929:
```suggestion
    @test ndims(ec) == ndims(typeof(ec)) == 1
```

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
...
+    @test_throws ArgumentError firstindex(ec, 2)
+    @test_throws ArgumentError firstindex(ec, 0)
+    @test lastindex(ec) == 4
+    @test lastindex(ec, 1) == 4
+    @test_throws ArgumentError lastindex(ec, 2)
+    @test_throws ArgumentError lastindex(ec, 0)
+
+    gdf = groupby(df, [:x1, :x2, :x3])
+    @test length(gdf) == 3
+    @test ndims(gdf) == 1
```
bkamins, line 1949:
```suggestion
    @test ndims(gdf) == == ndims(typeof(gdf)) == 1
```

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
...
+    @test_throws BoundsError firstindex(gdf, 2)
+    @test_throws BoundsError firstindex(gdf, 0)
+    @test lastindex(gdf) == 3
+    @test lastindex(gdf, 1) == 3
+    @test_throws BoundsError lastindex(gdf, 2)
+    @test_throws BoundsError lastindex(gdf, 0)
+
+    kgdf = keys(gdf)
+    @test length(kgdf) == 3
+    @test ndims(kgdf) == 1
```
bkamins, line 1969:
```suggestion
    @test ndims(kgdf) == ndims(typeof(kgdf)) == 1
```

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
...
+    @test firstindex(kgdf, 2) == 1
+    @test_throws BoundsError firstindex(kgdf, 0)
+    @test lastindex(kgdf) == 3
+    @test lastindex(kgdf, 1) == 3
+    @test lastindex(kgdf, 2) == 1
+    @test_throws BoundsError lastindex(kgdf, 0)
+
+    gk = kgdf[1]
+    @test length(gk) == 3
+    @test ndims(gk) == 1
```
bkamins, line 1989:
```suggestion
    @test ndims(gk) == ndims(typeof(gk)) == 1
```

test/indexing.jl
```diff
@@ -1859,4 +1859,150 @@ end
...
+    @test_throws ArgumentError firstindex(ec, 2)
+    @test_throws ArgumentError firstindex(ec, 0)
+    @test lastindex(ec) == 4
+    @test lastindex(ec, 1) == 4
+    @test_throws ArgumentError lastindex(ec, 2)
+    @test_throws ArgumentError lastindex(ec, 0)
+
+    gdf = groupby(df, [:x1, :x2, :x3])
+    @test length(gdf) == 3
+    @test ndims(gdf) == == ndims(typeof(gdf)) == 1
```
bkamins, line 1949:
```suggestion
    @test ndims(gdf) == ndims(typeof(gdf)) == 1
```

bkamins:
CI passes

docs/src/lib/indexing.md
```diff
@@ -238,12 +238,27 @@ The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
 This table presents return value types of calling `names`, `propertynames` and `keys`
```
nalimilan, line 238:
```suggestion
This table presents return value types of calling `names`, `propertynames`, `keys`, `length` and `ndims`
```

docs/src/lib/indexing.md
```diff
@@ -238,12 +238,27 @@ The elements of a `GroupedDataFrame` are [`SubDataFrame`](@ref)s of its parent.
...
+| `GroupedDataFrame`  | `Vector{String}` | tuple of fields  | `GroupKeys`      | `Int`     | `1`     |
+| `GroupKeys`         | undefined        | tuple of fields  | vector of `Int`  | `Int`     | `1`     |
+| `GroupKey`          | `Vector{String}` | `Vector{Symbol}` | `Vector{Symbol}` | `Int`     | `1`     |
+
+Additionally the above types `T` (i.e. `AbstractDataFrame`, `DataFrameRow`, `DataFrameRows`,
+`DataFrameColumns`, `GroupedDataFrame`, `GroupKeys`, `GroupKey`) the following methods are defined:
+* `size(::T)` returning a `Tuple` of `Int`.
+* `size(::T, ::Integer)` returning an `Int`.
+* `axes(::T)` returning a `Tuple` of `Int` vectors.
+* `axes(::T, ::Integer)` returning an `Int` vector.
```
nalimilan, line 256: When the dimension is invalid an error is thrown, at least for `AbstractDataFrame`, right?

src/abstractdataframe/iteration.jl
```diff
@@ -176,14 +176,23 @@ Base.IteratorSize(::Type{<:DataFrameColumns}) = Base.HasShape{1}()
...
+
 Base.length(itr::DataFrameColumns) = size(itr)[1]
 Base.eltype(::Type{<:DataFrameColumns}) = AbstractVector
+
 Base.firstindex(itr::DataFrameColumns) = 1
 Base.lastindex(itr::DataFrameColumns) = length(itr)
+
+Base.firstindex(itr::DataFrameColumns, i::Integer) = first(axes(itr, i))
+Base.lastindex(itr::DataFrameColumns, i::Integer) = last(axes(itr, i))
+Base.axes(itr::DataFrameColumns, i::Integer) = Base.OneTo(size(itr, i))
```
nalimilan, line 194: Do we need these in the end given that a fallback was added to Base?

nalimilan:
Makes sense. Just a few comments.

src/abstractdataframe/iteration.jl
bkamins: this fallback is only in 1.6. I added it to have it also e.g. in 1.0 in case someone in the future uses these functions in user code explicitly.

docs/src/lib/indexing.md
bkamins: Yes, it is thrown except for `DataFrameRows` and `GroupKeys` which are `AbstractVector`s so:
```
julia> axes(df, 3)
ERROR: ArgumentError: DataFrames only have two dimensions

julia> axes(eachrow(df), 3)
Base.OneTo(1)

julia> axes(keys(groupby(df, names(df, 1))), 3)
Base.OneTo(1)
```

I find this a bit inconsistent in Base, but this is how it works.

nalimilan: Sorry, my point was that this should probably be mentioned like for `firstindex`.

src/abstractdataframe/iteration.jl
nalimilan: Maybe add a `VERSION` check? It's always nice to be able to remove code at some point in the future. :-)

docs/src/lib/indexing.md
bkamins: added

src/abstractdataframe/iteration.jl
bkamins: added - checking if I did it correctly is welcome ;). Thank you!

bkamins:
Does this PR look OK now?

src/abstractdataframe/iteration.jl
bkamins: @nalimilan - does the version check look good to you now?

docs/src/lib/indexing.md
```diff
@@ -253,7 +253,9 @@ Additionally the above types `T` (i.e. `AbstractDataFrame`, `DataFrameRow`, `Dat
 * `size(::T)` returning a `Tuple` of `Int`.
 * `size(::T, ::Integer)` returning an `Int`.
 * `axes(::T)` returning a `Tuple` of `Int` vectors.
-* `axes(::T, ::Integer)` returning an `Int` vector.
+* `axes(::T, ::Integer)` returning an `Int` vector for a valid dimension (except
+   `DataFrameRows` and `GroupKeys` for which `Base.OneTo(1)` is also returned for
+   a dimension higher than a valid one because they are `AbstractVector`)..
```
nalimilan, line 258:
```suggestion
   a dimension higher than a valid one because they are `AbstractVector`).
```

nalimilan (approved):
Yes, sorry, for some reason I had missed the notifications.

bkamins:
I have added NEWS.md entry.

bkamins:
Thank you!

PR #2574, opened 2020-12-06
Enable multithreading with several operations in combine/select/transform

nalimilan:


nalimilan:
Benchmarks indicate a relatively small overhead for small data frames and a large gain with large ones when there are multiple operations. Maybe the overhead could be reduced by not using anonymous functions.
```julia
using Revise, DataFrames, BenchmarkTools, Random, PooledArrays
Random.seed!(1);
for N in (1_000, 10_000, 1_000_000, 100_000_000)
    @show N
    df = DataFrame(x=rand(1:10, N), y=rand(N));
    gd = groupby(df, :x);
    @btime combine($gd, :y => sum);
    @btime combine($gd, :y => sum, :y => maximum);
    @btime combine($gd, :y => (y -> sum(y)) => :sum, :y => (y -> maximum(y)) => :maximum);
end

# master, 1 thread (similar with 2 threads)
N = 1000
  18.067 μs (159 allocations: 12.72 KiB)
  28.645 μs (194 allocations: 14.31 KiB)
  43.973 μs (400 allocations: 35.06 KiB)
N = 10000
  26.693 μs (159 allocations: 12.72 KiB)
  55.812 μs (194 allocations: 14.31 KiB)
  103.867 μs (400 allocations: 176.94 KiB)
N = 1000000
  1.382 ms (159 allocations: 12.72 KiB)
  3.970 ms (194 allocations: 14.31 KiB)
  16.699 ms (420 allocations: 15.28 MiB)
N = 100000000
  147.389 ms (159 allocations: 12.72 KiB)
  405.261 ms (194 allocations: 14.31 KiB)
  2.523 s (420 allocations: 1.49 GiB)

# PR, 2 threads
N = 1000
  27.866 μs (179 allocations: 14.03 KiB)
  39.069 μs (219 allocations: 16.31 KiB)
  49.763 μs (445 allocations: 38.00 KiB)
N = 10000
  42.318 μs (179 allocations: 14.03 KiB)
  61.622 μs (219 allocations: 16.31 KiB)
  100.182 μs (443 allocations: 179.81 KiB)
N = 1000000
  1.412 ms (179 allocations: 14.03 KiB)
  2.957 ms (219 allocations: 16.31 KiB)
  8.945 ms (463 allocations: 15.28 MiB)
N = 100000000
  147.132 ms (179 allocations: 14.03 KiB)
  285.011 ms (220 allocations: 16.34 KiB)
  2.379 s (464 allocations: 1.49 GiB)

# PR, 1 thread
N = 1000
  22.249 μs (177 allocations: 13.97 KiB)
  32.287 μs (215 allocations: 16.19 KiB)
  55.203 μs (440 allocations: 37.84 KiB)
N = 10000
  31.565 μs (177 allocations: 13.97 KiB)
  60.558 μs (215 allocations: 16.19 KiB)
  115.944 μs (439 allocations: 179.69 KiB)
N = 1000000
  1.394 ms (177 allocations: 13.97 KiB)
  3.845 ms (215 allocations: 16.19 KiB)
  15.734 ms (459 allocations: 15.28 MiB)
N = 100000000
  146.931 ms (177 allocations: 13.97 KiB)
  408.077 ms (215 allocations: 16.19 KiB)
  2.506 s (459 allocations: 1.49 GiB)
```

bkamins:
Yes - this is very nice. I would not be overly concerned about the overhead given its size (which is very small). Thank you for working on this!

Could you please check if the following produce a correct result (if you prefer I can write a test-set for this):
```
using Combinatorics
trans = [:id => (y -> sum(y)) => :v1, :id => (y -> 10maximum(y)) => :v2, y -> (v3=100y.id[1],), y -> (v4=fill(1000y.id[1],y.id[1]+1),)]
for p in permutations(1:4), i in 1:4
    @show combine(groupby(DataFrame(id=1:2), :id), trans[p[1:i]]...)
end
```
this test is made to make sure we correctly expand the columns if needed.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -194,24 +194,27 @@ function _combine_process_agg(@nospecialize(cs_i::Pair{Int, <:Pair{<:Function, S
...
-        # we have seen this col but it is not allowed to replace it
-        optional || throw(ArgumentError("duplicate output column name: :$out_col_name"))
-        @assert trans_res[loc].optional && trans_res[loc].name == out_col_name
-        trans_res[loc] = TransformationResult(idx_agg, outcol, out_col_name, optional_i)
-        seen_cols[out_col_name] = (optional_i, loc)
-    else
-        push!(trans_res, TransformationResult(idx_agg, outcol, out_col_name, optional_i))
-        seen_cols[out_col_name] = (optional_i, length(trans_res))
+    return function()
+        if haskey(seen_cols, out_col_name)
```
bkamins, line 206: A small question to check (sorry for being picky) - have you checked if we do not have problems with boxing in this anonymous function (same below). It is just curiosity, as even if we do boxing it is not a problem as we do not do any expensive operations inside.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -194,24 +194,27 @@ function _combine_process_agg(@nospecialize(cs_i::Pair{Int, <:Pair{<:Function, S
                               gd::GroupedDataFrame,
                               seen_cols::Dict{Symbol, Tuple{Bool, Int}},
                               trans_res::Vector{TransformationResult},
-                              idx_agg::Union{Nothing, AbstractVector{Int}})
+                              idx_agg::Ref{Union{Nothing, Vector{Int}}})
```
bkamins, line 197: `idx_agg` should never be `nothing` here AFAICT (I know it is my code, but `idx_agg` is the most tricky part here so I think it is better to narrow down the signature to catch bugs if they are present).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -523,58 +541,66 @@ function _combine(gd::GroupedDataFrame,
...
-                    idx_agg = idx_pair
-                else
-                    @assert idx_agg === idx_pair
-                end
+            _combine_process_pair(cs_i, optional_i, parentdf, gd, seen_cols, trans_res, idx_agg)
+        end
+    end
+    for t in tasks
+        local postprocessf
+        try
```
bkamins, line 566: what is the purpose of this `try`? Just to unwrap one level of stack trace?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -523,58 +541,66 @@ function _combine(gd::GroupedDataFrame,
...
+            postprocessf = fetch(t)
+        catch e
+            if e isa TaskFailedException
+                rethrow(t.exception)
+            else
+                rethrow()
+            end
+        end
+        idx = postprocessf()
```
bkamins, line 575: I think given the `Ref` logic you have implemented we:
1. do not need this step with `idx` (it is enough to call `postprocessf()`)
2. we do not have to return `idx` from the anonymous functions above (you correctly update `idx_agg` in the codes with `lock` part - this is a side benefit of having `Ref` here)

Agreed?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -591,23 +617,25 @@ function _combine(gd::GroupedDataFrame,
...
-            newcol = similar(col)
-            # we can probably make it more efficient, but I leave it as an optimization for the future
-            gd_idx = gd.idx
-            k = 0
-            # consider adding @inbounds later
-            for (s, e) in zip(gd.starts, gd.ends)
-                for j in s:e
-                    k += 1
-                    newcol[gd_idx[j]] = col[k]
+    @sync for i in eachindex(trans_res)
```
bkamins, line 620: This should be properly tested (as it is not tested with `combine`, only `select` or `transform` would invoke this code path and only if `GroupedDataFrame` was reordered).

(maybe it is, but I am just making it to remember to ensure this)

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -523,58 +541,66 @@ function _combine(gd::GroupedDataFrame,
     # and if a given column can be replaced in the future
     seen_cols = Dict{Symbol, Tuple{Bool, Int}}()
 
+    tasks = similar(cs_norm, Task)
+
     parentdf = parent(gd)
-    for i in eachindex(cs_norm, optional_transform)
+    for i in eachindex(cs_norm, optional_transform, tasks)
```
bkamins, line 547: if you agree with comment https://github.com/JuliaData/DataFrames.jl/pull/2574/files#r537099264 then `@sync` should be added here and below we should just call the post-processing anonymous function when everything is done being executed (so that we are sure that `idx_agg` after this loop and before next steps is properly set).

src/groupeddataframe/splitapplycombine.jl
nalimilan: I've pushed an alternative version which returns a function an arguments rather than an anonymous function to nl/threadedops2. It's actually slightly slower. It could be considered as a bit cleaner but it's more verbose. Let me know what you think.

src/groupeddataframe/splitapplycombine.jl
nalimilan: That's to avoid throwing a `TaskFailedException`. Without it:
```julia
julia> combine(gd, :y => (x -> nonexistent(x)))
ERROR: TaskFailedException
Stacktrace:
 [1] wait
   @ ./task.jl:317 [inlined]
 [2] fetch
   @ ./task.jl:332 [inlined]
 [3] _combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:604
 [4] _combine_prepare(gd::GroupedDataFrame{DataFrame}, cs::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractVector{T} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:66
 [5] #combine#519
   @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703 [inlined]
 [6] combine(gd::GroupedDataFrame{DataFrame}, cs::Pair{Symbol, var"#17#18"})
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:696
 [7] top-level scope
   @ REPL[14]:1

    nested task error: UndefVarError: nonexistent not defined
    Stacktrace:
     [1] (::var"#17#18")(x::SubArray{Float64, 1, Vector{Float64}, Tuple{Vector{Int64}}, false})
       @ Main ./REPL[14]:1
     [2] do_call(f::var"#17#18", idx::Vector{Int64}, starts::Vector{Int64}, ends::Vector{Int64}, gd::GroupedDataFrame{DataFrame}, incols::Tuple{Vector{Float64}}, i::Int64)
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/callprocessing.jl:95
     [3] _combine_process_pair(cs_i::Pair, optional_i::Bool, parentdf::DataFrame, gd::GroupedDataFrame{DataFrame}, seen_cols::Dict{Symbol, Tuple{Bool, Int64}}, trans_res::Vector{DataFrames.TransformationResult}, idx_agg::Base.RefValue{Union{Nothing, Vector{Int64}}})
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:508
     [4] macro expansion
       @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:598 [inlined]
     [5] (::DataFrames.var"#806#812"{GroupedDataFrame{DataFrame}, Bool, Bool, DataFrame, Dict{Symbol, Tuple{Bool, Int64}}, Vector{DataFrames.TransformationResult}, Base.RefValue{Union{Nothing, Vector{Int64}}}, Bool, Pair{Int64, Pair{var"#17#18", Symbol}}})()
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:541
```

With it:
```julia
julia> combine(gd, :y => (x -> nonexistent(x)))
ERROR: UndefVarError: nonexistent not defined
Stacktrace:
 [1] wait
   @ ./task.jl:317 [inlined]
 [2] fetch
   @ ./task.jl:332 [inlined]
 [3] _combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:604
 [4] _combine_prepare(gd::GroupedDataFrame{DataFrame}, cs::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractVector{T} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:66
 [5] #combine#519
   @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703 [inlined]
 [6] combine(gd::GroupedDataFrame{DataFrame}, cs::Pair{Symbol, var"#19#20"})
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703
 [7] top-level scope
   @ REPL[14]:1
```

Another option would be to use `throw(t.exception)` instead of `rethrow(t.exception)`. The gives three exceptions, which is too bad, but the last one gives the line where the error really happened. But we probably don't care since that's not useful for users. Ideally we would only print the last one but I don't know how to do that.
```julia
julia> combine(gd, :y => (x -> nonexistent(x)))
ERROR: UndefVarError: nonexistent not defined
Stacktrace:
 [1] _combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:607
 [2] _combine_prepare(gd::GroupedDataFrame{DataFrame}, cs::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractVector{T} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:66
 [3] #combine#519
   @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703 [inlined]
 [4] combine(gd::GroupedDataFrame{DataFrame}, cs::Pair{Symbol, var"#21#22"})
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703
 [5] top-level scope
   @ REPL[14]:1

caused by: TaskFailedException
Stacktrace:
 [1] wait
   @ ./task.jl:317 [inlined]
 [2] fetch
   @ ./task.jl:332 [inlined]
 [3] _combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:604
 [4] _combine_prepare(gd::GroupedDataFrame{DataFrame}, cs::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractVector{T} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:66
 [5] #combine#519
   @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703 [inlined]
 [6] combine(gd::GroupedDataFrame{DataFrame}, cs::Pair{Symbol, var"#21#22"})
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703
 [7] top-level scope
   @ REPL[14]:1

    nested task error: UndefVarError: nonexistent not defined
    Stacktrace:
     [1] (::var"#21#22")(x::SubArray{Float64, 1, Vector{Float64}, Tuple{Vector{Int64}}, false})
       @ Main ./REPL[14]:1
     [2] do_call(f::var"#21#22", idx::Vector{Int64}, starts::Vector{Int64}, ends::Vector{Int64}, gd::GroupedDataFrame{DataFrame}, incols::Tuple{Vector{Float64}}, i::Int64)
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/callprocessing.jl:95
     [3] _combine_process_pair(cs_i::Pair, optional_i::Bool, parentdf::DataFrame, gd::GroupedDataFrame{DataFrame}, seen_cols::Dict{Symbol, Tuple{Bool, Int64}}, trans_res::Vector{DataFrames.TransformationResult}, idx_agg::Base.RefValue{Union{Nothing, Vector{Int64}}})
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:508
     [4] macro expansion
       @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:598 [inlined]
     [5] (::DataFrames.var"#940#946"{GroupedDataFrame{DataFrame}, Bool, Bool, DataFrame, Dict{Symbol, Tuple{Bool, Int64}}, Vector{DataFrames.TransformationResult}, Base.RefValue{Union{Nothing, Vector{Int64}}}, Bool, Pair{Int64, Pair{var"#21#22", Symbol}}})()
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:541
```

nalimilan: The issue with the current `rethrow` approach is that it hides all the actual stacktrace, for example if you do:
```julia
julia> f(x) = nonexistent(x)

julia> combine(gd, :y => (x -> f(x)))
ERROR: UndefVarError: nonexistent not defined
Stacktrace:
 [1] wait
   @ ./task.jl:317 [inlined]
 [2] fetch
   @ ./task.jl:332 [inlined]
 [3] _combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:604
 [4] _combine_prepare(gd::GroupedDataFrame{DataFrame}, cs::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractVector{T} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:66
 [5] #combine#519
   @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703 [inlined]
 [6] combine(gd::GroupedDataFrame{DataFrame}, cs::Pair{Symbol, var"#29#30"})
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:703
 [7] top-level scope
   @ REPL[18]:1
```
That can be really problematic as it makes debugging impossible. Not sure whether there's a way to print only the error we want to show to users.

src/groupeddataframe/splitapplycombine.jl
bkamins: I prefer this one 😄. It actually seems easier to read and follow the logic. Even better that it is faster.

src/groupeddataframe/splitapplycombine.jl
bkamins: Since with your approach:
```
_combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
```
is still printed (and it is a crucial line) I am OK to have it as you proposed. But I would be also OK not to do the `try-catch` block and just leave what is printed by default.

bkamins:
Ah - one thing. So in this PR you propose to use all threads that Julia was started with to use? I would be OK with this, but wanted to double check with you if this is the intent (as earlier we considered a different strategy - still now it is probably OK to do as you propose).

src/groupeddataframe/splitapplycombine.jl
nalimilan: Yes, good catch. At least I hope so. Do you think there's anything to check to ensure that's the case (beyond seeing tests pass)?

bkamins: The tests I proposed in https://github.com/JuliaData/DataFrames.jl/pull/2574#issuecomment-739541672 are intended to make sure that all works correctly.
+
we have to double check each update and use of `idx_agg` is consistent (I did this and it looks all is fine, but a second pair of eyes would not hurt - but I assume you also already did this)

src/groupeddataframe/splitapplycombine.jl
nalimilan: I wouldn't say that's the crucial line: the user doesn't care where the error happened in DataFrames, since that's not an error in DataFrames. Users care about where in their code the error happened.

I've pushed a commit to just use the default error handling. This gives something a bit verbose but with all the information. What's annoying is that the first exception is `TaskFailedException`, so I've added a comment at https://github.com/JuliaLang/julia/pull/37224#issuecomment-743804791.
```julia
julia> f(x) = nonexistent(x)

julia> combine(gd, :x => (x -> f(x)))
ERROR: TaskFailedException
Stacktrace:
 [1] wait
   @ ./task.jl:317 [inlined]
 [2] fetch
   @ ./task.jl:332 [inlined]
 [3] _combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:562
 [4] _combine_prepare(gd::GroupedDataFrame{DataFrame}, cs::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractVector{T} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:66
 [5] #combine#529
   @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:645 [inlined]
 [6] combine(gd::GroupedDataFrame{DataFrame}, cs::Pair{Symbol, var"#21#22"})
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:647
 [7] top-level scope
   @ REPL[34]:1

    nested task error: UndefVarError: nonexistent not defined
    Stacktrace:
     [1] f(x::SubArray{Int64, 1, Vector{Int64}, Tuple{Vector{Int64}}, false})
       @ Main ./REPL[6]:1
     [2] #21
       @ ./REPL[34]:1 [inlined]
     [3] do_call(f::var"#21#22", idx::Vector{Int64}, starts::Vector{Int64}, ends::Vector{Int64}, gd::GroupedDataFrame{DataFrame}, incols::Tuple{Vector{Int64}}, i::Int64)
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/callprocessing.jl:95
     [4] _combine_process_pair(cs_i::Pair, optional_i::Bool, parentdf::DataFrame, gd::GroupedDataFrame{DataFrame}, seen_cols::Dict{Symbol, Tuple{Bool, Int64}}, trans_res::Vector{DataFrames.TransformationResult}, idx_agg::Base.RefValue{Union{Nothing, Vector{Int64}}})
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:466
     [5] macro expansion
       @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:556 [inlined]
     [6] (::DataFrames.var"#1871#1877"{GroupedDataFrame{DataFrame}, Bool, Bool, DataFrame, Dict{Symbol, Tuple{Bool, Int64}}, Vector{DataFrames.TransformationResult}, Base.RefValue{Union{Nothing, Vector{Int64}}}, Bool, Pair{Int64, Pair{var"#21#22", Symbol}}})()
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:499
```

src/groupeddataframe/splitapplycombine.jl
nalimilan: Actually `@sync` isn't needed here since we call `fetch` on each task below (and we have to since we need to store tasks until they return).

nalimilan:
> Ah - one thing. So in this PR you propose to use all threads that Julia was started with to use? I would be OK with this, but wanted to double check with you if this is the intent (as earlier we considered a different strategy - still now it is probably OK to do as you propose).

Yes, here the overhead of starting multiple tasks is much lower so I'm not sure it's necessary to allow tweaking that. Even if other threads are busy it shouldn't be a problem. Though we'll probably have to address this issue in other places where the tradeoff is less clear.

src/groupeddataframe/splitapplycombine.jl
bkamins: It is needed I think because you have to make sure that `idx_agg` HAS been computed before finishing this loop. Otherwise it is possible that you would start post-processing and `idx_agg` there would not be computed yet in this loop. So:
1) post processing would think that `idx_agg` is `nothing`
2) later actual processing would compute it

Because of this using `fetch` sequentially is not enough. You have to be sure you `fetch`ed everything before moving forward.

However - let us both review my line of reasoning again (it is for sure safe, but maybe in our actual case it is overly safe and the `idx_agg` issue will not hit us).

I am raising this because my experience is that these kinds of bugs (if I am right and there is a bug - I might be wrong) are heisenbugs (as if you start monitoring the process to diagnose them they disappear because main loop has enough time to finish its operations).

src/groupeddataframe/splitapplycombine.jl
bkamins: OK - thank you!

src/groupeddataframe/splitapplycombine.jl
nalimilan: OK I've added a test. Yes I tried to check, but I'm not super familiar with the code logic.

src/groupeddataframe/splitapplycombine.jl
nalimilan: OK, let's add it, anyway it shouldn't cost anything and it makes the code clearer.

src/groupeddataframe/splitapplycombine.jl
nalimilan: Actually with `@sync` the error is better:
```julia
julia> f(x) = nonexistent(x)

julia> combine(gd, :x => (x -> f(x)))
ERROR: TaskFailedException

    nested task error: UndefVarError: nonexistent not defined
    Stacktrace:
     [1] f(x::SubArray{Int64, 1, Vector{Int64}, Tuple{Vector{Int64}}, false})
       @ Main ./REPL[115]:1
     [2] #145
       @ ./REPL[116]:1 [inlined]
     [3] do_call(f::var"#145#146", idx::Vector{Int64}, starts::Vector{Int64}, ends::Vector{Int64}, gd::GroupedDataFrame{DataFrame}, incols::Tuple{Vector{Int64}}, i::Int64)
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/callprocessing.jl:95
     [4] _combine_process_pair(cs_i::Pair, optional_i::Bool, parentdf::DataFrame, gd::GroupedDataFrame{DataFrame}, seen_cols::Dict{Symbol, Tuple{Bool, Int64}}, trans_res::Vector{DataFrames.TransformationResult}, idx_agg::Base.RefValue{Union{Nothing, Vector{Int64}}})
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:466
     [5] macro expansion
       @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:556 [inlined]
     [6] (::DataFrames.var"#3079#3085"{GroupedDataFrame{DataFrame}, Bool, Bool, DataFrame, Dict{Symbol, Tuple{Bool, Int64}}, Vector{DataFrames.TransformationResult}, Base.RefValue{Union{Nothing, Vector{Int64}}}, Bool, Pair{Int64, Pair{var"#145#146", Symbol}}})()
       @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:499
Stacktrace:
 [1] sync_end(c::Channel{Any})
   @ Base ./task.jl:364
 [2] macro expansion
   @ ./task.jl:383 [inlined]
 [3] _combine(gd::GroupedDataFrame{DataFrame}, cs_norm::Vector{Any}, optional_transform::Vector{Bool}, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:542
 [4] _combine_prepare(gd::GroupedDataFrame{DataFrame}, cs::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractVector{T} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:66
 [5] #combine#529
   @ ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:645 [inlined]
 [6] combine(gd::GroupedDataFrame{DataFrame}, cs::Pair{Symbol, var"#145#146"})
   @ DataFrames ~/.julia/dev/DataFrames/src/groupeddataframe/splitapplycombine.jl:647
 [7] top-level scope
   @ REPL[116]:1
```

src/groupeddataframe/splitapplycombine.jl
nalimilan: It seems to be properly tested, as it failed on Julia 1.0 when I got the implementation of `@spawn` wrong yesterday (so that `@sync` didn't have any effect).

bkamins: OK. Thank you!

src/groupeddataframe/splitapplycombine.jl
bkamins: I will re-check this when the CI starts to pass and let you know.

src/groupeddataframe/splitapplycombine.jl
bkamins: OK

bkamins:
CI fails because with threading the exception type changes. I would unwrap `CompositeException` in tests and check the wrapped exception. OK?

src/groupeddataframe/splitapplycombine.jl
nalimilan: Actually given that `@sync` changes the type of the thrown exception, I suggest we go with the approach at https://github.com/JuliaData/DataFrames.jl/pull/2574#discussion_r541750910 instead. It's quite verbose, but the exception type is correct so we won't have to break that in the future, and the full backtrace is printed in the last exception. I've pushed a commit to do that.

nalimilan:
> CI fails because with threading the exception type changes. I would unwrap `CompositeException` in tests and check the wrapped exception. OK?

Actually as I replied above I suggest we return the right exception type, and then we don't need to change tests. What do you think?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -523,58 +536,62 @@ function _combine(gd::GroupedDataFrame,
...
         end
     end
+    # Workaround JuliaLang/julia#38931:
+    # we want to preserve the exception type thrown in user code,
+    # and print the backtrace corresponding to it
+    for t in tasks
+        try
+            wait(t)
+        catch e
+            throw(t.exception)
```
bkamins, line 566: maybe add `@assert e isa TaskFailedException`. This is consistent with `wait` docstring.

Actually I would say that it is needed to do so, as otherwise we have:
```
julia> Threads.@spawn sqrt(-1)
Task (failed) @0x00007f5ceb93b0d0
DomainError with -1.0:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
throw_complex_domainerror(::Symbol, ::Float64) at ./math.jl:33
sqrt at ./math.jl:573 [inlined]
sqrt(::Int64) at ./math.jl:599
(::var"#11#12")() at ./threadingconstructs.jl:169

julia> t = Threads.@spawn begin
       sleep(5)
       sqrt(-1)
       end
Task (runnable) @0x00007f5cebce24a0

julia> wait(t)
ERROR: TaskFailedException:
DomainError with -1.0:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(::Symbol, ::Float64) at ./math.jl:33
 [2] sqrt at ./math.jl:573 [inlined]
 [3] sqrt(::Int64) at ./math.jl:599
 [4] macro expansion at ./REPL[10]:3 [inlined]
 [5] (::var"#13#14")() at ./threadingconstructs.jl:169
Stacktrace:
 [1] wait(::Task) at ./task.jl:267
 [2] top-level scope at REPL[11]:1
```

nalimilan: Good idea. Actually we can even be more general and call `rethrow(e)` if we get an unexpected type, that can't hurt.

What do you mean with this example? Just that the exception type isn't `DomainError` as one would expect?

bkamins: By this example I mean that when you do `@spawn` you might get an exception before `wait` is called.

So in general - you could have two different exception types thrown from some code (as what I passed in `begin`-`end` block could be anything - either something that throws immediately or something that will throw only when `wait` is called)

bkamins:
Thank you for working on it. It would be great (if possible) to make sure that codecov/patch does not produce :x: (it seems related to the `v"1.3"` checks so maybe it is not possible).

Apart of this I will now do some testing of this branch and try to break it :smile:!

nalimilan:
> Thank you for working on it. It would be great (if possible) to make sure that codecov/patch does not produce x (it seems related to the `v"1.3"` checks so maybe it is not possible).

Yeah I 'm not sure why Codecov thinks the VERSION line isn't tested, since we definitely cover both branches. Another issue is due to the fact that to emulate `@spawn` on Julia < 1.3, which I copyied the definition of `@async` from Base: the lines that handle `@sync` are not used, but it's probably better to keep them anyway. I guess could add a test for it. What do you think?

The last uncovered line is https://github.com/JuliaData/DataFrames.jl/pull/2574/files#annotation_739358955, which isn't covered currently. Looks like something like `select(gd, :x, :x)` should be added?

bkamins:
> I guess could add a test for it. What do you think?

If you could it would be welcome. Thank you!

> Looks like something like select(gd, :x, :x) should be added?

I thought I had a test for it, but I will double check and add it if it is missing.

bkamins:
Also please remember to change target branch to `main` in this PR (it might require making sure that nothing gets messed up, but it should be OK in this case I think). Thank you!

src/DataFrames.jl
```diff
@@ -91,6 +91,23 @@ else
     export only
 end
 
+if VERSION >= v"1.3"
+    using Base.Threads: @spawn
+else
+    # This is the definition of @async in Base
+    macro spawn(expr)
```
bkamins, line 98: the definition of `@spawn` is:
```
macro spawn(expr)
    letargs = Base._lift_one_interp!(expr)

    thunk = esc(:(()->($expr)))
    var = esc(Base.sync_varname)
    quote
        let $(letargs...)
            local task = Task($thunk)
            task.sticky = false
            if $(Expr(:islocal, var))
                put!($var, task)
            end
            schedule(task)
            task
        end
    end
end
```
and I understand we do not need the `letargs` variable, but why the other things are changed?

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -523,58 +536,74 @@ function _combine(gd::GroupedDataFrame,
...
+        end
+    end
+    # Workaround JuliaLang/julia#38931:
+    # we want to preserve the exception type thrown in user code,
+    # and print the backtrace corresponding to it
+    for t in tasks
+        try
+            wait(t)
+        catch e
+            @static if VERSION > v"1.3"
```
bkamins, line 566: maybe add a test when there is an error to check this path?

test/grouping.jl
```diff
@@ -3272,4 +3273,24 @@ end
     @test df == df2
 end
 
+@testset "permutations of operations with combine" begin
+    df = DataFrame(id=rand(1:10, 20))
+    gd = groupby(df, :id)
```
bkamins, line 3278:
```suggestion
    gd = groupby_checked(df, :id)
```

test/grouping.jl
```diff
@@ -3272,4 +3273,24 @@ end
     @test df == df2
 end
 
+@testset "permutations of operations with combine" begin
+    df = DataFrame(id=rand(1:10, 20))
```
bkamins, line 3277: we usually set seed of PRNG.

test/grouping.jl
```diff
@@ -3272,4 +3273,24 @@ end
...
+    gd = groupby(df, :id)
+    trans = [:id => (y -> sum(y)) => :v1,
+             :id => (y -> 10maximum(y)) => :v2,
```
bkamins, line 3280: we should add a standard aggregation (as it creates `idx_agg` in a different moment)
```suggestion
             :id => sum => :v3,
```

bkamins: ah - also a no-op like `:id => :v4` might be added

bkamins:
I am running the following test:
```
julia> gd = groupby(DataFrame(id=rand(1:10, 10^7)), :id);

julia> trans = [:id => (y -> sum(y)) => :v1,
                   :id => (y -> 10maximum(y)) => :v2,
                   :id => sum => :v6,
                      y -> (v3=100y.id[1],),
                    y -> (v4=fill(1000y.id[1],y.id[1]+1),),
                    :id => :v5]

julia> @time for p in permutations(1:4), i in 1:4
             res = combine(gd, trans[p[1:i]]...)
             for j in 1:i
                 expected = nrow(res) <= 10 ? combine(gd, trans[p[j]]) :
                     # Second operation is there only to generate as many rows as in res
                     combine(gd, trans[p[j]], y -> (xxx=fill(1000y.id[1],y.id[1]+1),))
                 nms = intersect(names(expected), names(res))
                 @test res[!, nms] == expected[!, nms]
             end
         end
```
on `main` and on this branch with threads ranging from 1 to 8. I do not see significant differences (there is no clear pattern what is faster) - do you see the same?

bkamins:
I am getting the following problem on Ubuntu Linux (machine with 8 virtual cores, 4 physical cores):
```
~/Desktop/Dev/DF_dev$ julia -t 4
 Activating environment at `~/Desktop/Dev/DF_dev/Project.toml`
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.3 (2020-11-09)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using DataFrames, Random

julia> function f()
           Random.seed!(1);
           N = 100_000_000
           df = DataFrame(x=rand(1:10, N), y=rand(N));
           gd = groupby(df, :x);
           @time combine(gd, :y => sum, :y => maximum);
           @time combine(gd, :y => (y -> sum(y)) => :sum, :y => (y -> maximum(y)) => :maximum);
           @time combine(gd, :y => sum, :y => maximum, :y => (y -> sum(y)) => :sum2, :y => (y -> maximum(y)) => :maximum2);
           nothing
       end

julia> f()
  2.070578 seconds (4.44 M allocations: 229.045 MiB, 7.12% gc time)
  3.568066 seconds (3.58 M allocations: 2.416 GiB, 5.25% gc time)
  2.419049 seconds (680.06 k allocations: 1.524 GiB, 2.81% gc time)

julia> f()
  0.317690 seconds (211 allocations: 16.797 KiB)
  3.026716 seconds (465 allocations: 2.235 GiB, 9.06% gc time)
  2.661975 seconds (535 allocations: 1.490 GiB, 2.52% gc time)

julia> f()
Killed
```
(for sure I have enough RAM, and current `main` works smoothly)

nalimilan:
Mmm that's worrying. Could you run it with `--check-bounds=yes` or within `gdb` to see whether we can get more information?

bkamins:
With `--check-bounds=yes` it is the same

bkamins:
I am not sure how to run the `julia -t 4` and my code under `gdb`. Could you please help me with this? (also - can you reproduce this or not?)

Also: it does not crash with 4 threads under:
```
julia> versioninfo()
Julia Version 1.6.0-beta1
Commit b84990e1ac (2021-01-08 12:42 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.0 (ORCJIT, skylake)
```
on the same machine

nalimilan:
I'd have used `gdb julia` and then `run -t4`, but I've never tried that actually (it seems to work here). I couldn't reproduce it on Julia 1.5.3 (Fedora) or 1.6. Though I've noticed gets relatively high and my machine started to swap. Have you checked the actual memory use of the process? If GC is disabled maybe things gets really bad...

bkamins:
Yes, it seems to be the GC issue and running out of memory. I will as on Slack.

bkamins:
Regarding the core of this PR - my tests (and it seems this is also what you get) is that the benefit is only when one uses several fast-aggregation functions (in other cases the difference is minimal even for large data sets). Do you confirm this.

nalimilan:
Yes, actually I was going to answer this:

> I am running the following test:
> 
> ```
> julia> gd = groupby(DataFrame(id=rand(1:10, 10^7)), :id);
> 
> julia> trans = [:id => (y -> sum(y)) => :v1,
>                    :id => (y -> 10maximum(y)) => :v2,
>                    :id => sum => :v6,
>                       y -> (v3=100y.id[1],),
>                     y -> (v4=fill(1000y.id[1],y.id[1]+1),),
>                     :id => :v5]
> 
> julia> @time for p in permutations(1:4), i in 1:4
>              res = combine(gd, trans[p[1:i]]...)
>              for j in 1:i
>                  expected = nrow(res) <= 10 ? combine(gd, trans[p[j]]) :
>                      # Second operation is there only to generate as many rows as in res
>                      combine(gd, trans[p[j]], y -> (xxx=fill(1000y.id[1],y.id[1]+1),))
>                  nms = intersect(names(expected), names(res))
>                  @test res[!, nms] == expected[!, nms]
>              end
>          end
> ```
> 
> on `main` and on this branch with threads ranging from 1 to 8. I do not see significant differences (there is no clear pattern what is faster) - do you see the same?

I don't see clear differences either. In the benchmarks I posted [above](https://github.com/JuliaData/DataFrames.jl/pull/2574#issuecomment-739517789), the only case where multithreading has a clear effect is with two operations that use the fast path (sum and maximum). So that's consistent with your benchmark. I'm not sure why that's the case. Though the good news is that https://github.com/JuliaData/DataFrames.jl/pull/2588 makes operations that do *not* use the fast path faster, so the two PRs are kind of complementary.

bkamins:
> so the two PRs are kind of complementary.

But how would they play-together? I.e. this PR uses threads for different aggregations and #2588 uses them for single aggregation. So the question is if they are combined what would happen.

nalimilan:
Actually #2588 is on top of this PR, so it includes both. The idea is that each operation gets a task, and then inside operations those that use the fast path only get one task, but those that use the slow path get one task per CPU. Since operations that use the fast path are, well, faster, this should be quite to optimal when mixing fast path and slow path operations.

bkamins:
Makes sense. I will look into the other PR tomorrow.

nalimilan:
Could you try with the latest commit? It should allow the GC to run every 100,000 rows, i.e. 1000 times in this example. Performance doesn't seem to be affected too much.

src/groupeddataframe/fastaggregates.jl
```diff
@@ -163,18 +163,30 @@ function groupreduce!(res::AbstractVector, f, op, condf, adjust, checkempty::Boo
...
-                counts[gix] += 1
+    if VERSION >= v"1.4" && Threads.nthreads() > 1
+        batchsize = 100_000
+    else
+        batchsize = typemax(Int)
+    end
+    for batch in Iterators.partition(eachindex(incol, groups), batchsize)
+        # Allow other tasks to do garbage collection while this one runs
+        @static if VERSION >= v"1.4"
+            GC.safepoint()
```
bkamins, line 174: The only issue is that `GC.safepoint` requires that the thread in which we call it is non-allocating. Are we sure this is the case here? (maybe we should ask on Slack - as I do not know the internals of how `GC.safepoint` works, but it seems that if the thread is allocating it would be not safe to add it)

nalimilan: I don't think that's right. The docstring says "Inserts a point in the program where garbage collection may run. [...] Calling this function periodically in non-allocating threads allows garbage collection to run." It's just that it's redundant if the thread allocates, as the GC can already run automatically at these points if needed.

src/DataFrames.jl
bkamins: Ah - you mean that on 1.0 we just do @async but without threading?

src/groupeddataframe/fastaggregates.jl
bkamins: OK - probably you are right. I will test it then again tomorrow.

bkamins:
What is our strategy? You want to merge this PR first and then the other threading PR separately?

nalimilan:
I'd say yes.

src/DataFrames.jl
nalimilan: Yes. That's the best we can do there.

bkamins:
OK - I think it is good to merge, except the minor comments I have left and resolving merge conflicts

nalimilan:
Do you mean the GC issue is fixed with the last commit? :-)

bkamins:
Yes, I cannot reproduce it now.

src/groupeddataframe/splitapplycombine.jl
nalimilan: AFAICT it's covered, there's just a bug with coverage of `@static`: at https://codecov.io/gh/JuliaData/DataFrames.jl/pull/2574/diff the lines inside both branches are covered.

bkamins:
I have run some more stress tests on larger data. The system chokes (even System Monitor stalls for a second in extreme memory usage cases), but I cannot get Julia "Killed". Fantastic job as usual!

Do you think we should ask someone with experience in threading for another review, or you are confident with what we have? (I have checked the "logic", but I do not know internals of threading implementation)

bkamins:
Can you please also add information about using multithreading in the docstings and in the manual? I think more advanced users would want to know the "way" we do it and use the threads as it might affect their workflows and decision with how many threads to start Julia.

bkamins:
I also asked for opinion on Slack on using all available threads.

nalimilan:
Cool!

I'm reasonably confident with the implementation, but of course it's always good to have more eyes if somebody familiar with threading is willing to look at the PR.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -591,23 +620,25 @@ function _combine(gd::GroupedDataFrame,
...
-            newcol = similar(col)
-            # we can probably make it more efficient, but I leave it as an optimization for the future
-            gd_idx = gd.idx
-            k = 0
-            # consider adding @inbounds later
-            for (s, e) in zip(gd.starts, gd.ends)
-                for j in s:e
-                    k += 1
-                    newcol[gd_idx[j]] = col[k]
+    @sync for i in eachindex(trans_res)
```
bkamins, line 623: can you please add a comment on the logic how we handle the threading here? (i.e. a high-level logic how we ensure that we do not have race-conditions and correctly process things). Thank you!

bkamins (approved):
Looks good. I will test the other PR when it is rebased against the current state of this one

src/groupeddataframe/splitapplycombine.jl
nalimilan: I'm not sure what to say about this loop actually. Each entry can be processed in parallel because we extract its fields, do some computing based on them, and replace the original entry with the result. We don't update any external state. Is there something in particular that you feel is worth mentioning?

bkamins: I do not mean this loop but the logic in general. I.e. that we have to:
1. sync `idx_agg` across threads
2. do actual inserting of columns in post-processing as we have to do it sequentially (although we compute them in parallel)

maybe something more that might be considered hard to understand.

docs/src/man/split_apply_combine.md
```diff
@@ -122,6 +122,10 @@ It is allowed to mix single values and vectors if multiple transformations
 are requested. In this case single value will be repeated to match the length
 of columns specified by returned vectors.
 
+A separate task is spawned for each specified transformation, allowing for
+parallel operation when several transformations are requested and Julia was
+started with more than one thread.
```
bkamins, line 127: Maybe add that this means that the transformation functions passed should not modify the same state of the Julia program?

nalimilan: Good idea. I've also mentioned that this may be extended in the future so that people are not caught by surprise.

nalimilan:
I've realized I forgot to actually enable multithreading on CI. I chose 4 threads to maximize the chances of catching synchronization bugs, and a warning so that we detect issues with settings instead of things silently succeeding.

bkamins:
Thank you for the changes. I was testing it on 1 to 8 threads and did not catch a bug. I am OK with merging it once CI passess.

bkamins:
OK - please let me know when you rebase the other PR and I will review. Thank you!

PR #2594, opened 2021-01-08
add firstindex check and fix a bug in mapcols!

bkamins:
We currently assume that vectors are 1-based in DataFrames.jl. This PR makes sure we check for that. Not checking this will lead to bugs. Also I found some small bug in `mapcols!` in the process related to `AbstractRange` handling.

I think we should decide on this before 1.0 release (it is better to be strict I think now; we can be more flexible later).

I will add tests, NEWS.md and update documentation with this PR after we agree that we want to add these changes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,11 @@
 
 ## Bug fixes
 
+* DataFrames.jl now checks that passed columns are 1-based as this is a current
+  design assumption; `mapcols!` makes sure not to create columns being
+  `AbstractRange` consistently with other methods that add columns to a `DataFrame`
+  ([#2594](https://github.com/JuliaData/DataFrames.jl/pull/2594))
+
 ## New functionalities
 
 * `firstindex`, `lastindex`, `size`, `ndims`, and `axes` are now consistently defined
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,8 +7,9 @@
 ## Bug fixes
 
 * DataFrames.jl now checks that passed columns are 1-based as this is a current
-  design assumption; `mapcols!` makes sure not to create columns being
-  `AbstractRange` consistently with other methods that add columns to a `DataFrame`
+  design assumption ([#2594](https://github.com/JuliaData/DataFrames.jl/pull/2594))
+* `mapcols!` makes sure not to create columns being `AbstractRange` consistently
+  with other methods that add columns to a `DataFrame`
   ([#2594](https://github.com/JuliaData/DataFrames.jl/pull/2594))
 
 ## New functionalities
--- a/Project.toml
+++ b/Project.toml
@@ -52,4 +52,4 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
 test = ["Combinatorics", "DataStructures", "DataValues", "Dates", "Logging",
-        "OffsetArrays", Random", "Test", "Unitful"]
+        "OffsetArrays", "Random", "Test", "Unitful"]
--- a/Project.toml
+++ b/Project.toml
@@ -45,9 +45,11 @@ DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 DataValues = "e7dc6d0d-1eca-5fa6-8ad6-5aecde8b7ea5"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = ["Combinatorics", "DataStructures", "DataValues", "Dates", "Logging", "Random", "Test", "Unitful"]
+test = ["Combinatorics", "DataStructures", "DataValues", "Dates", "Logging",
+        "OffsetArrays", Random", "Test", "Unitful"]
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -456,8 +456,7 @@ function mapcols!(f::Union{Function, Type}, df::DataFrame)
     end
 
     for (i, col) in enumerate(vs)
-        fic = firstindex(col)
-        fic != 1 && _onebased_check_error(i, fic)
+        firstindex(col) != 1 && _onebased_check_error(i, col)
     end
 
     @assert length(vs) == ncol(df)
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -440,7 +440,7 @@ function mapcols!(f::Union{Function, Type}, df::DataFrame)
                 throw(ArgumentError("mixing scalars and vectors in mapcols not allowed"))
             end
             seenvector = true
-            push!(vs, fv)
+            push!(vs, fv isa AbstractRange ? collect(fv) : fv)
         else
             if seenvector
                 throw(ArgumentError("mixing scalars and vectors in mapcols not allowed"))
@@ -454,6 +454,16 @@ function mapcols!(f::Union{Function, Type}, df::DataFrame)
     if len_min != len_max
         throw(DimensionMismatch("lengths of returned vectors must be identical"))
     end
+
+    for (i, col) in enumerate(vs)
+        fic = firstindex(col)
+        if fic != 1
+            throw(ArgumentError("Currently DataFrames.jl supports only " *
+                                "columns that use 1-based indexing and " *
+                                "column $i has starting index equal to $fic"))
+        end
+    end
+
     _columns(df) .= vs
 
     return df
--- a/src/abstractdataframe/iteration.jl
+++ b/src/abstractdataframe/iteration.jl
@@ -457,14 +457,14 @@ function mapcols!(f::Union{Function, Type}, df::DataFrame)
 
     for (i, col) in enumerate(vs)
         fic = firstindex(col)
-        if fic != 1
-            throw(ArgumentError("Currently DataFrames.jl supports only " *
-                                "columns that use 1-based indexing and " *
-                                "column $i has starting index equal to $fic"))
-        end
+        fic != 1 && _onebased_check_error(i, fic)
     end
 
-    _columns(df) .= vs
+    @assert length(vs) == ncol(df)
+    raw_columns = _columns(df)
+    for i in 1:ncol(df)
+        raw_columns[i] = vs[i]
+    end
 
     return df
 end
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -383,7 +383,7 @@ index(df::DataFrame) = getfield(df, :colindex)
 # make sure that:
 # 1. `AbstractRange` columns are not added to a `DataFrame`
 # 2. all inserted columns use 1-based indexing
-# 3. that after several mutating operations on the vector are performed
+# 3. after several mutating operations on the vector are performed
 #    each element (column) has the same length
 # 4. if length of the vector is changed that the index of the `DataFrame`
 #    is adjusted appropriately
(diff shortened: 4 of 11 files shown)


src/abstractdataframe/iteration.jl
```diff
@@ -454,6 +454,16 @@ function mapcols!(f::Union{Function, Type}, df::DataFrame)
...
         throw(DimensionMismatch("lengths of returned vectors must be identical"))
     end
+
+    for (i, col) in enumerate(vs)
+        fic = firstindex(col)
+        if fic != 1
+            throw(ArgumentError("Currently DataFrames.jl supports only " *
+                                "columns that use 1-based indexing and " *
+                                "column $i has starting index equal to $fic"))
+        end
```
nalimilan, line 464: How about moving this to a separate function to avoid repeating the error message?

bkamins: OK - makes sense. Actually I think we should even remove all uses of the `_columns` function as using it leads to unsafe code. I will work on this. I understand you agree with a general direction I propose here.

bkamins:
@nalimilan - are you OK with these proposed changes. If yes I would finalize the PR

bkamins:
@nalimilan - the issue came up again in #2604. Are you OK to finalize this PR in the form I have proposed?

src/abstractdataframe/iteration.jl
bkamins: I will leave `_columns` for now as we have different use-cases for it. I have added a developers comment to it though

bkamins:
@nalimilan - this should be good to have a look at (especially corner cases in the tests).

test/indexing.jl
```diff
@@ -1,6 +1,6 @@
 module TestIndexing
 
-using Test, DataFrames
+using Test, DataFrames. OffsetArrays
```
bkamins, line 3:
```suggestion
using Test, DataFrames, OffsetArrays
```

bkamins:
I had to split the tests into a separate file as OffsetArrays.jl does not run on Julia 1.0 (due to its dependency).

test/indexing.jl
```diff
@@ -2005,4 +2005,8 @@ end
     @test_throws BoundsError lastindex(gk, 0)
 end
 
+if VERSION >= v"1.5"
+    include("indexing_offest.jl")
```
bkamins, line 2009:
```suggestion
    include("indexing_offset.jl")
```

bkamins:
CI and coverage passes cleanly now.

NEWS.md
```diff
@@ -6,6 +6,11 @@
 
 ## Bug fixes
 
+* DataFrames.jl now checks that passed columns are 1-based as this is a current
+  design assumption; `mapcols!` makes sure not to create columns being
+  `AbstractRange` consistently with other methods that add columns to a `DataFrame`
```
nalimilan, line 11: This part is probably worth a separate bullet.

Project.toml
```diff
@@ -45,9 +45,10 @@ DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
...
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = ["Combinatorics", "DataStructures", "DataValues", "Dates", "Logging", "Random", "Test", "Unitful"]
+test = ["Combinatorics", "DataStructures", "DataValues", "Dates", "Logging", "OffsetArrays", "Random", "Test", "Unitful"]
```
nalimilan, line 54: 92 chars limit ? :-p

src/abstractdataframe/iteration.jl
```diff
@@ -454,7 +454,17 @@ function mapcols!(f::Union{Function, Type}, df::DataFrame)
     if len_min != len_max
         throw(DimensionMismatch("lengths of returned vectors must be identical"))
     end
-    _columns(df) .= vs
```
nalimilan, line 457: What was the problem with this syntax? Compilation overhead?

src/dataframe/dataframe.jl
```diff
@@ -369,8 +377,26 @@ DataFrame(column_eltypes::AbstractVector{<:Type}, cnames::AbstractVector{<:Abstr
...
+# 4. if length of the vector is changed that the index of the `DataFrame`
+#    is adjusted appropriately
 _columns(df::DataFrame) = getfield(df, :columns)
 
+_onebased_check_error() =
+    throw(ArgumentError("Currently DataFrames.jl supports only columns " *
+                        "that use 1-based indexing"))
+_onebased_check_error(i, fic) =
+    throw(ArgumentError("Currently DataFrames.jl supports only " *
+                        "columns that use 1-based indexing and " *
```
nalimilan, line 397:
```suggestion
                        "columns that use 1-based indexing, but " *
```

src/dataframe/dataframe.jl
```diff
@@ -369,8 +377,26 @@ DataFrame(column_eltypes::AbstractVector{<:Type}, cnames::AbstractVector{<:Abstr
 ##############################################################################
 
 index(df::DataFrame) = getfield(df, :colindex)
+
+# this function grants the access to the internal storage of columns of the
+# `DataFrame` and its use is unsafe. If the returned vector is mutated then
+# make sure that:
+# 1. `AbstractRange` columns are not added to a `DataFrame`
+# 2. all inserted columns use 1-based indexing
+# 3. that after several mutating operations on the vector are performed
```
nalimilan, line 386:
```suggestion
# 3. after several mutating operations on the vector are performed
```

src/dataframe/dataframe.jl
```diff
@@ -205,6 +208,11 @@ struct DataFrame <: AbstractDataFrame
             end
         end
 
+        for (i, col) in enumerate(columns)
+            fic = firstindex(col)
+            fic != 1 && _onebased_check_error(i, fic)
```
nalimilan, line 213: Given that this pattern happens in lots of places and that we don't care about performance when an error happens, I would suggest changing `_onebased_check_error` to take `col`:
```suggestion
            firstindex(col) != 1 && _onebased_check_error(i, col)
```

src/abstractdataframe/iteration.jl
```diff
@@ -454,7 +454,17 @@ function mapcols!(f::Union{Function, Type}, df::DataFrame)
...
+
+    for (i, col) in enumerate(vs)
+        fic = firstindex(col)
+        fic != 1 && _onebased_check_error(i, fic)
```
nalimilan, line 460: Given that this pattern happens in lots of places and that we don't care about performance when an error happens, I would suggest changing `_onebased_check_error` to take `col`:
```suggestion
        firstindex(col) != 1 && _onebased_check_error(i, col)
```

test/indexing_offset.jl
```diff
@@ -0,0 +1,35 @@
...
+    DataFrames._check_consistency(df)
+
+    @test_throws ArgumentError df.b = ov1
+    @test_throws ArgumentError insertcols!(df, :b => ov1)
+    @test_throws DimensionMismatch df[!, :b] .= ov1
+
+    # this is consequence of the fact that OffsetArrays wrap AbstractRange in this case
+    @test_throws ErrorException df[:, :a] = ov1
+
+    # this inconsistency is the consequence how setindex! for vector is deifned in Base
```
nalimilan, line 28:
```suggestion
    # this inconsistency is the consequence how setindex! for vector is defined in Base
```

yurivish:
@bkamins I just learned about the function `Base.require_one_based_indexing` – are you aware of it? It's mentioned in the devdocs [here](https://docs.julialang.org/en/v1/devdocs/offset-arrays/) and I'm not sure if it's sanctioned for use in user code, but the devdocs are written as if it is.

It is defined here: https://github.com/JuliaLang/julia/blob/ae53238c45a0cd6dafc6e121f4daaa93143bf627/base/abstractarray.jl#L103

bkamins:
@yurivish - thank you for looking into this. I preferred custom checks for the following reasons (although the function seems to be a part of "official" API even if it is not exported):
* it is not available in Julia 1.0 which we support
* we can throw custom error message if we handle the check ourselves
* it is, easy to make this check

src/abstractdataframe/iteration.jl
bkamins: yes - I try to remove broadcasting in the internal code to reduce latency a bit (also I thought that the syntax might seem too magic for casual reader).

src/abstractdataframe/iteration.jl
bkamins: But `col` is a vector. What would you then print in the error message?

NEWS.md
bkamins: fixed

Project.toml
bkamins: fixed

src/abstractdataframe/iteration.jl
bkamins: The difference is:

current release:
```
julia> using DataFrames

julia> df = DataFrame(a=1)
1×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │     1

julia> @time mapcols!(length, df)
  0.003640 seconds (2.75 k allocations: 146.932 KiB)
1×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │     1
```
this PR:
```
julia> using DataFrames

julia> df = DataFrame(a=1)
1×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │     1

julia> @time mapcols!(length, df)
  0.000015 seconds (4 allocations: 256 bytes)
1×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │     1
```
so it is not a big deal (but we have to iterate `vs` anyway to check `firstindex`).

src/dataframe/dataframe.jl
```diff
@@ -369,8 +377,26 @@ DataFrame(column_eltypes::AbstractVector{<:Type}, cnames::AbstractVector{<:Abstr
...
+#    is adjusted appropriately
 _columns(df::DataFrame) = getfield(df, :columns)
 
+_onebased_check_error() =
+    throw(ArgumentError("Currently DataFrames.jl supports only columns " *
+                        "that use 1-based indexing"))
+_onebased_check_error(i, fic) =
+    throw(ArgumentError("Currently DataFrames.jl supports only " *
+                        "columns that use 1-based indexing, but " *
+                        "column $i has starting index equal to $fic"))
```
bkamins, line 398:
```suggestion
_onebased_check_error(i, col) =
    throw(ArgumentError("Currently DataFrames.jl supports only " *
                        "columns that use 1-based indexing, but " *
                        "column $i has starting index equal to $(firstindex(col))"))
```

src/dataframe/dataframe.jl
```diff
@@ -390,6 +416,12 @@ corrupt_msg(df::DataFrame, i::Integer) =
 
 function _check_consistency(df::DataFrame)
     cols, idx = _columns(df), index(df)
+
+    for (i, col) in enumerate(cols)
+        fic = firstindex(col)
+        fic != 1 && _onebased_check_error(i, fic)
```
bkamins, line 422:
```suggestion
        firstindex(col) != 1 && _onebased_check_error(i, col)
```

src/abstractdataframe/iteration.jl
nalimilan: Wow. I wouldn't have expected such a large difference in the number of allocations.

nalimilan (approved):

nalimilan:
Mmm, is the CI failure related to recent changes in Tables?

src/abstractdataframe/iteration.jl
bkamins: broadcasting is heavy (when doing the PRs that Tim Holy prompted I will review the whole codebase against using broadcasting). Of course here the difference is one-time and not that big.

bkamins:
Yes, I will rebase and merge after CI passes.

bkamins:
Thank you!

PR #2327, opened 2020-07-24
add `filter` example to docs on taking subsets

floswald:
* this refers to #2318 
* I also added an additional example from the `append!` docs

PR #2598, opened 2021-01-11
rebased version for #2327 add filter subsetting to docs

floswald:
rebased try!

### DIFF
--- a/docs/src/man/getting_started.md
+++ b/docs/src/man/getting_started.md
@@ -645,6 +645,41 @@ Equivalently, the `in` function can be called with a single argument to create
 a function object that tests whether each value belongs to the subset
 (partial application of `in`): `df[in([1, 5, 601]).(df.A), :]`.
 
+#### Selecting Rows with `filter`
+
+We have seen above how to subset a `DataFrame` to several criteria, involving multiple columns, by supplying a logical vector to the first dimension. For instance, in the following we want to subset to all rows where `x > 2` and where `a == 'c'`:
+
+```jldoctest dataframe
+julia> df = DataFrame(:x => 1:4, :y => "a", :a => 'a':'d', :b => exp(1))
+4×4 DataFrame
+│ Row │ x     │ y      │ a    │ b       │
+│     │ Int64 │ String │ Char │ Float64 │
+├─────┼───────┼────────┼──────┼─────────┤
+│ 1   │ 1     │ a      │ 'a'  │ 2.71828 │
+│ 2   │ 2     │ a      │ 'b'  │ 2.71828 │
+│ 3   │ 3     │ a      │ 'c'  │ 2.71828 │
+│ 4   │ 4     │ a      │ 'd'  │ 2.71828 │
+
+julia> df[(df.x .> 2) .& (df.a .== 'c'), : ]
+1×4 DataFrame
+│ Row │ x     │ y      │ a    │ b       │
+│     │ Int64 │ String │ Char │ Float64 │
+├─────┼───────┼────────┼──────┼─────────┤
+│ 1   │ 3     │ a      │ 'c'  │ 2.71828 │
+```
+
+An alternative formulation, which notably saves on the need to use
+broadcasting syntax via `.` prefixes, uses [`filter`](@ref) or [`filter!`](@ref):
+
+```jldoctest dataframe
+julia> filter([:x, :a] => (x1, x2) -> x1 > 2 && x2 == 'c', df)
+1×4 DataFrame
+│ Row │ x     │ y      │ a    │ b       │
+│     │ Int64 │ String │ Char │ Float64 │
+├─────┼───────┼────────┼──────┼─────────┤
+│ 1   │ 3     │ a      │ 'c'  │ 2.71828 │
+```
+
 !!! note
 
     As with matrices, subsetting from a data frame will usually return a copy of


docs/src/man/getting_started.md
```diff
@@ -645,6 +645,41 @@ Equivalently, the `in` function can be called with a single argument to create
 a function object that tests whether each value belongs to the subset
 (partial application of `in`): `df[in([1, 5, 601]).(df.A), :]`.
 
+#### Selecting Rows with `filter`
```
bkamins, line 648:
```suggestion
#### Selecting rows with `filter`
```

bkamins (approved):
@nalimilan - looks good to me (except for comments), but please kindly review it from the language perspective. I would merge it as is, although soon we should add the information about `subset` to it.

docs/src/man/getting_started.md
```diff
@@ -645,6 +645,41 @@ Equivalently, the `in` function can be called with a single argument to create
...
+
+We have seen above how to subset a `DataFrame` to several criteria, involving multiple columns, by supplying a logical vector to the first dimension. For instance, in the following we want to subset to all rows where `x > 2` and where `a == 'c'`:
+
+```jldoctest dataframe
+julia> df = DataFrame(:x => 1:4, :y => "a", :a => 'a':'d', :b => exp(1))
```
bkamins, line 653: it would be better to make columns `y` and `b` to contain heterogenous data, or just drop them.

Also - your example was produced using DataFrames.jl 0.21, and now we have DataFrames.jl 0.22, which produces a different display.

floswald: dropped `b` but not `y` (changed to `rand`). it's important to realize that we filter the _entire_ `DataFrame` in this example, so `y` should be there.

bkamins (approved):

bkamins:
We have just merged https://github.com/JuliaData/DataFrames.jl/pull/2496. I will amend your PR with an example how it can be used.

bkamins:
I have proposed something (subject to correction most likely as I am not a native speaker).

docs/src/man/getting_started.md
```diff
@@ -877,6 +877,58 @@ package provides interfaces similar to LINQ and [dplyr](https://dplyr.tidyverse.
 
 See the [Data manipulation frameworks](@ref) section for more information.
 
+#### Selecting rows with `filter` and `subset`
```
nalimilan, line 880: I'd put this section above the one presenting `select` and `transform`, which is more complex and goes beyond taking subsets. (Transformations probably deserve a separate section so that we only present columns selection, but that's another story.)

That way you don't need to repeat the example: you can just continue with the ones given at the end of the Indexing section.

docs/src/man/getting_started.md
```diff
@@ -877,6 +877,58 @@ package provides interfaces similar to LINQ and [dplyr](https://dplyr.tidyverse.
...
+
+julia> df[(df.x .> 2) .& (df.a .== 'c'), : ]
+1×3 DataFrame
+ Row │ x      y       a    
+     │ Int64  String  Char 
+─────┼─────────────────────
+   1 │     3  Claire  c
+```
+
+An alternative formulation, which notably saves on the need to use
```
nalimilan, line 905: Now that we provide `subset` with a syntax closer to `select`, I'm not sure we should mention `filter` at all in Getting Started. Or maybe just mention it at the end for reference. What do you think @bkamins?

docs/src/man/getting_started.md
```diff
@@ -877,6 +877,58 @@ package provides interfaces similar to LINQ and [dplyr](https://dplyr.tidyverse.
...
+     │ Int64  String  Char 
+─────┼─────────────────────
+   1 │     3  Claire  c
+```
+
+You can also use the [`subset`](@ref) or [`subset!`](@ref) functions that are more
+flexible. They use the same syntax as [`select`](@ref) that was discussed in
+the previous section). The requirement is that each column transformation
+must produce a vector of `Bool` and only the rows from the source data frame
+for which all transformations produce `true` are inluded in the result:
```
nalimilan, line 921:
```suggestion
for which all transformations produce `true` are included in the result:
```

docs/src/man/getting_started.md
```diff
@@ -877,6 +877,58 @@ package provides interfaces similar to LINQ and [dplyr](https://dplyr.tidyverse.
...
+
+```jldoctest dataframe
+julia> subset(df, :x => ByRow(>(2)), :a => ByRow(==('c')))
```
nalimilan, line 924: Maybe start showing the standard `x -> x > 2` syntax? That's easier to grasp for newcomers, and more general.

nalimilan (commented):
Sorry for the delay!

docs/src/man/getting_started.md
bkamins: We cannot, as `subset` has the same syntax as `select` so we need to introduce it after `filter`. We could split the discussion of `filter` and `subset` though.

docs/src/man/getting_started.md
bkamins: We could do this - then it should be after `select` section as commented above. So - should I make this change?

docs/src/man/getting_started.md
bkamins: OK - I will update it when we settle on the general layout.

docs/src/man/getting_started.md
nalimilan: Ah indeed. Let's keep it that way then. But "Taking a Subset" isn't actually a good name for the parent section as it also shows complex transformations. Maybe we should just split each sub-section into its own section, i.e. "Indexing", "Selecting and transforming columns" and "Selecting rows".

bkamins: Agreed.

bkamins:
@floswald - would you want to update this PR to take into account the current state of the documentation? Thank you!

bkamins:
closed given https://github.com/JuliaData/DataFrames.jl/pull/2900 is merged (and anything here would have to be heavily rewritten so probably it is better to open an new PR if needed).

Thank you for raising the issue.

PR #2601, opened 2021-01-16
Fix groupreduce with var and std for Unitful types

### LINKED ISSUE #2600: combine() error with std() on GroupedDataFrame with Unitful columns
```
using DataFrames, Statistics, Unitful
df = DataFrame(a=[:a, :a, :b, :b], b=[0, 1, 2, 3]u"m")
```

```
julia> combine(df, :b => std)
1×1 DataFrame
 Row │ b_std
     │ Quantity…
─────┼───────────
   1 │ 1.29099 m
```

```
julia> combine(groupby(df, :a), :b => std)
ERROR: DimensionError: 0.0 m and 0.25 m² are not dimensionally compatible.
Stacktrace:
 [1] +(::Quantity{Float64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}}, ::Quantity{Float64,𝐋²,Unitful.FreeUnits{(m²,),𝐋²,nothing}}) at /Users/tomyun/.julia/packages/Unitful/1t88N/src/quantities.jl:137
 [2] groupreduce!(::Array{Quantity{Float64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}, ::DataFrames.var"#535#538"{Array{Quantity{Float64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}}, ::typeof(+), ::Nothing, ::Function, ::Bool, ::Array{Quantity{Int64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}, ::GroupedDataFrame{DataFrame}) at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/fastaggregates.jl:174
 [3] (::DataFrames.Aggregate{typeof(var),Nothing})(::Array{Quantity{Int64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}, ::GroupedDataFrame{DataFrame}) at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/fastaggregates.jl:248
 [4] (::DataFrames.Aggregate{typeof(std),Nothing})(::Array{Quantity{Int64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}, ::GroupedDataFrame{DataFrame}) at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/fastaggregates.jl:254
 [5] _combine_process_agg(::Pair{Int64,var"#s151"} where var"#s151"<:(Pair{var"#s150",Symbol} where var"#s150"<:Function), ::Bool, ::DataFrame, ::GroupedDataFrame{DataFrame}, ::Dict{Symbol,Tuple{Bool,Int64}}, ::Array{DataFrames.TransformationResult,1}, ::Array{Int64,1}) at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/splitapplycombine.jl:203
 [6] _combine(::GroupedDataFrame{DataFrame}, ::Array{Any,1}, ::Array{Bool,1}, ::Bool, ::Bool, ::Bool) at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/splitapplycombine.jl:532
 [7] _combine_prepare(::GroupedDataFrame{DataFrame}, ::Union{Regex, AbstractString, Function, Signed, Symbol, Unsigned, Pair, AbstractArray{T,1} where T, Type, All, Between, Cols, InvertedIndex}; keepkeys::Bool, ungroup::Bool, copycols::Bool, keeprows::Bool, renamecols::Bool) at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/splitapplycombine.jl:66
 [8] #combine#509 at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/splitapplycombine.jl:638 [inlined]
 [9] combine(::GroupedDataFrame{DataFrame}, ::Pair{Symbol,typeof(std)}) at /Users/tomyun/.julia/packages/DataFrames/yqToF/src/groupeddataframe/splitapplycombine.jl:638
 [10] top-level scope at REPL[16]:1
```

Tested on Julia 1.5.3 with DataFrames 0.22.2 and Unitful 1.5.0.

nalimilan:
For `Number`s for which squaring changes the type, we need to use different types for the mean, the variance and the standard deviation.

Fixes #2600.

@bkamins Do you think we should add Unitful as a test dependency to test this? I wonder whether there are types in Base that would allow reproducing the problem. Otherwise we could create a custom type just for this.

### DIFF
--- a/src/groupeddataframe/fastaggregates.jl
+++ b/src/groupeddataframe/fastaggregates.jl
@@ -251,7 +251,7 @@ end
 
 function (agg::Aggregate{typeof(var)})(incol::AbstractVector, gd::GroupedDataFrame)
     means = groupreduce((x, i) -> x, Base.add_sum, agg.condf, /, false, incol, gd)
-    z = zero(eltype(incol))
+    z = zero(eltype(incol)) - zero(eltype(means))
     S = typeof((abs2(z) + abs2(z))/2)
     # !ismissing check is purely an optimization to avoid a copy later
     T = eltype(incol) >: Missing && agg.condf !== !ismissing ?
--- a/src/groupeddataframe/fastaggregates.jl
+++ b/src/groupeddataframe/fastaggregates.jl
@@ -258,7 +258,7 @@ function (agg::Aggregate{typeof(var)})(incol::AbstractVector, gd::GroupedDataFra
         T = Union{Missing, S} : S
     res = zeros(T, length(gd))
     return groupreduce!(res, (x, i) -> @inbounds(abs2(x - means[i])), +, agg.condf,
-                        (x, l) -> l <= 1 ? oftype(x / (l-1), NaN) : x / (l-1),
+                        (x, l) -> l <= 1 ? x/0 : x/(l-1),
                         false, incol, gd)
 end
 
--- a/src/groupeddataframe/fastaggregates.jl
+++ b/src/groupeddataframe/fastaggregates.jl
@@ -251,12 +251,11 @@ end
 
 function (agg::Aggregate{typeof(var)})(incol::AbstractVector, gd::GroupedDataFrame)
     means = groupreduce((x, i) -> x, Base.add_sum, agg.condf, /, false, incol, gd)
+    z = zero(eltype(incol))
+    S = typeof((abs2(z) + abs2(z))/2)
     # !ismissing check is purely an optimization to avoid a copy later
-    if eltype(means) >: Missing && agg.condf !== !ismissing
-        T = Union{Missing, real(eltype(means))}
-    else
-        T = real(eltype(means))
-    end
+    T = eltype(incol) >: Missing && agg.condf !== !ismissing ?
+        T = Union{Missing, S} : S
     res = zeros(T, length(gd))
     return groupreduce!(res, (x, i) -> @inbounds(abs2(x - means[i])), +, agg.condf,
                         (x, l) -> l <= 1 ? oftype(x / (l-1), NaN) : x / (l-1),
@@ -265,10 +264,10 @@ end
 
 function (agg::Aggregate{typeof(std)})(incol::AbstractVector, gd::GroupedDataFrame)
     outcol = Aggregate(var, agg.condf)(incol, gd)
-    if eltype(outcol) <: Union{Missing, Rational}
-        return sqrt.(outcol)
-    else
+    if typeof(sqrt(zero(eltype(outcol)))) === eltype(outcol)
         return map!(sqrt, outcol, outcol)
+    else
+        return map(sqrt, outcol)
     end
 end
 
--- a/test/grouping.jl
+++ b/test/grouping.jl
@@ -1,7 +1,7 @@
 module TestGrouping
 
 using Test, DataFrames, Random, Statistics, PooledArrays, CategoricalArrays, DataAPI,
-    Combinatorics
+    Combinatorics, Unitful
 const ≅ = isequal
 
 """Check if passed data frames are `isequal` and have the same element types of columns"""
@@ -906,7 +906,8 @@ Base.isless(::TestType, ::TestType) = false
     skip in (false, true), sort in (false, true), indices in (false, true)
     Random.seed!(1)
     df = DataFrame(a = rand([1:5;missing], 20), x1 = rand(1:100, 20),
-                   x2 = rand(1:100, 20) +im*rand(1:100, 20))
+                   x2 = rand(1:100, 20) +im*rand(1:100, 20),
+                   x4 = rand(1:100, 20) .* u"m")
 
     for f in (sum, prod, maximum, minimum, mean, var, std, first, last, length)
         gd = groupby_checked(df, :a, skipmissing=skip, sort=sort)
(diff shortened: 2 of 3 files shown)


bkamins:
I think adding Unitful.jl as [extras] dependency is not a problem.

src/groupeddataframe/fastaggregates.jl
```diff
@@ -238,12 +238,11 @@ end
 
 function (agg::Aggregate{typeof(var)})(incol::AbstractVector, gd::GroupedDataFrame)
     means = groupreduce((x, i) -> x, Base.add_sum, agg.condf, /, false, incol, gd)
+    z = zero(eltype(incol))
```
bkamins, line 241: This should be more precise I think:
```suggestion
    z = zero(eltype(incol)) - zero(eltype(means))
```

bkamins:
@nalimilan - when we merge this we should backport it to https://github.com/JuliaData/DataFrames.jl/tree/0.22_patches and make a release. Therefore can you please change the version to 0.22.3 in Project.toml in this PR?

(I can do the backport and the release later, unless you would prefer to do it)

bkamins:
So now we have:
```
m^2 and NaN are not dimensionally compatible.
```
On Julia 1.0 😢.

nalimilan:
Actually that's a legitimate failure that must be fixed on other versions too. The random values just happened to hit that case on Julia 1.0. I've pushed a fix and an adaptation of tests to more reliably cover the situation where a group contains a single row.

test/grouping.jl
```diff
@@ -1,6 +1,7 @@
 module TestGrouping
 
-using Test, DataFrames, Random, Statistics, PooledArrays, CategoricalArrays, DataAPI
+using Test, DataFrames, Random, Statistics, PooledArrays, CategoricalArrays, DataAPI,
```
bkamins, line 3: I would not make a line break here :)

test/grouping.jl
```diff
@@ -904,8 +905,11 @@ Base.isless(::TestType, ::TestType) = false
 @testset "combine with aggregation functions (skipmissing=$skip, sort=$sort, indices=$indices)" for
     skip in (false, true), sort in (false, true), indices in (false, true)
     Random.seed!(1)
-    df = DataFrame(a = rand([1:5;missing], 20), x1 = rand(1:100, 20),
-                   x2 = rand(1:100, 20) +im*rand(1:100, 20))
+    # 5 is there to ensure we test a single-row group
+    df = DataFrame(a = [rand([1:4;missing], 19); 5],
+                   x1 = rand(1:100, 20),
+                   x2 = rand(1:100, 20) + im*rand(1:100, 20),
+                   x4 = rand(1:100, 20) .* u"m")
```
bkamins, line 912: can we also add `x5` column that would be like `x4` but also contain `missing` value and then below, in particular we would test also `var∘skipmissing` etc. (and make sure that we have a group that has only `missing` values in it).

In general I would want to test what happens if fast aggregation functions get a vector of zero length (which is only possible with `missing`s as currently our groups must have a positive length).

bkamins (approved):
Wow - that was fast and precise. I would just add one more test as commented (hopefully it will pass so I approve the PR).

test/grouping.jl
nalimilan: This is tested below. Do you mean we should also test it for Unitful? I'm afraid that will require duplicating the whole block, as `prod` doesn't apply to these so we can't just loop over different types of columns. And then there's also `Complex` and `CategoricalArray`...

test/grouping.jl
nalimilan: Well without it the line takes 94 chars... :-)

test/grouping.jl
bkamins: Yes - I meant a test with `Unitful`. Maybe then just add a small test with e.g. `var∘skipmissing` on a concrete test DataFrame like: one group only, column is union of missing and `Unitful` and contains only `missing` values?

test/grouping.jl
bkamins: as you like then :)

test/grouping.jl
nalimilan: OK, let's bite the bullet and add systematic tests like for floats. I've pushed a commit.

test/grouping.jl
```diff
@@ -965,29 +968,95 @@ Base.isless(::TestType, ::TestType) = false
...
+        @test typeof(res.y) == typeof(expected.y)
+        res = combine(gd, :x2m => f∘skipmissing => :y)
+        expected = combine(gd, :x2m => (x -> f(collect(skipmissing(x)))) => :y)
+        @test res ≅ expected
+        @test typeof(res.y) == typeof(expected.y)
+
+        # Test reduction over group with only missing values
+        gd = groupby_checked(df, :a, skipmissing=skip, sort=sort)
+        indices && @test gd.idx !== nothing # Trigger computation of indices
+        gd[1][:, :x2m] .= missing
```
bkamins, line 990: nice 😄

bkamins (approved):

bkamins:
will you make a patch release or should I do it?

nalimilan:
Please go ahead. :-)

PR #2602, opened 2021-01-17
add Gallery to the documentation

bkamins:
We should first discuss how the gallery should be structured, but this PR is to keep track of creating it.

If would be great to have the gallery filled with something for 1.0 release.


bkamins:
closing it. We will manage upcoming examples as they are added

PR #2603, opened 2021-01-20
Added HTML + stylized output

emmettgb:
Hi! I found output to be rather tedious in IJulia, so I decided to go ahead and create something like this. That being said, I do think that this will need further implementation to work with some of the other display methods in DataFrames. It does add the dependency IJulia, but I think it also makes the package generally a lot more usable. Feel free to let me know what you think, or if these changes might not suite the package best in your opinion!


bkamins:
Thank you for the contribution, which is an useful idea - we just need to decide how to implement it.
Two comments:
* I do not think you need `IJulia` dependency in the code, the methods you import from `IJulia` are just in `Base`
* we will switch HTML and LaTeX printing to PrettyTables.jl backend soon; @ronisbr is working on it now; therefore this proposal should be coordinated with this bigger task.

@ronisbr - could you please comment how you see it?

emmettgb:
That's funny, I guess I am not sure why I thought show() was from IJulia, I use in the REPL all the time haha.
I'll go ahead and get rid of that change to the Project.toml.
I will look through the code on this new Tex/HTML printing when it is up! Thanks!

bkamins:
Thank you!

ronisbr:
Hi!

Yes, all this code would be removed when I finished the HTML backend of PrettyTables so that DataFrames can use it. However, we can add this style as a predefined format of PrettyTables :)

@bkamins due to some activities in my work, I really cannot say how much time it would take to finish HTML backend of PrettyTables. Hence, I will let you decide whether or not merge this. Notice that this `style` keyword would have to be modified (or adapted) to work with PrettyTables.

bkamins:
We will wait till PrettyTables.jl is out, as we cannot change the API all the time as we want to hit 1.0 release soon (it would be a one-shot change when PrettyTables.jl is out and then having ability to define styles there would be great). Thank you!

@emmettgb - in summary: I understand you will get what you want but unfortunately not immediately.

emmettgb:
Understandable! I am excited to see your implementation! It is no big deal, I can just write this into an extension and use it as my preference until the PrettyTables release :)

bkamins:
I would close this as @ronisbr is working on adding HTML via PrettyTables.jl (hopefully soon). If you feel it should be reopened please comment. Thank you!

PR #2607, opened 2021-01-22
bumped Tables.jl to 1.3

ExpandingMan:
I know this would eventually get created by CompatHelper, but I don't know when that will be, so I made this.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -35,7 +35,7 @@ PrettyTables = "0.10"
 Reexport = "0.1, 0.2, 1.0"
 SortingAlgorithms = "0.1, 0.2, 0.3"
 TableTraits = "0.4, 1"
-Tables = "1.3"
+Tables = "1.2, 1.3"
 Unitful = "1"
 julia = "1"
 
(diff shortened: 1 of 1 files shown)


quinnj:
Why is this needed? DataFrames.jl isn't _restricting_ the version to 1.2, so it's not like it would hold back other packages, right? And there's no functionality in DataFrames.jl that seems to require the new functionality?

ExpandingMan:
I'm confused... doesn't having `Tables = "1.2` in `compat` restrict DataFrames.jl to use `[1.2, 1.3)`?

quinnj:
I thought it still meant `[1.2, 2.0)`

ExpandingMan:
Ohhh... shit, I think I'm an idiot.  I was getting it confused with < 1.0 which works differently.  Ugh, it's so confusing.

quinnj:
No worries; I thought the same until someone pointed this out to me too.

ExpandingMan:
On the bright side, it's nice to be able to relax somewhat when dealing with packages that have tagged a 1.0.

quinnj:
Oh definitely; that's why I've made a huge push lately to tag a bunch of 1.0s everywhere. Like, if the code is expected to be use-able at all by people, I say tag a 1.0, just to make the package maintenance/future releases/dependents lives much easier.

PR #2609, opened 2021-01-23
Fix DataFrame(Any[]) test

bkamins:
@quinnj - I guess it is intended in Tables.jl 1.3.1 change?

### DIFF
--- a/test/tables.jl
+++ b/test/tables.jl
@@ -265,7 +265,7 @@ end
 end
 
 @testset "test constructor with vectors" begin
-    @test_throws ArgumentError DataFrame(Any[])
+    @test DataFrame(Any[]) == DataFrame()
     df = DataFrame(typeof((1, 1))[])
     @test names(df) == ["Column1", "Column2"]
     @test size(df) == (0, 2)


quinnj (approved):
LGTM

bkamins:
Thank you!

strickek:
@bkamins - I have a test error (current Julia installation 1.6.0) because this commit is in master, but not part of the actual release v0.22.7. Is this intended or an oversight?

bkamins:
Yes - this commit is in `main`, but it was not backported, as it only fixes tests and has no effect on the functionality. Why do you need 0.22.7 tests to pass? (if we make 0.22.8 release I will backport it)

strickek:
no need. I just tested the few packages I use regularly after installing Julia 1.6.0. and was not sure about.

bkamins:
OK - thank you for reporting then. It is really appreciated.

This test fails because Tables.jl version changed underneath. On `main` all works correctly.

PS. We regularly have a similar issue with Julia Base 😄, as tests in DataFrames.jl are extensive, so if something changes in Julia usually DataFrames.jl tests start to fail. We keep track of this, but do not make release for every test that has to be updated as this is not something exposed to the user.

PR #2610, opened 2021-01-23
Use refpool optimized method for integer grouping

nalimilan:
main:
```julia

julia> df = DataFrame(x=rand(1:10, 100_000), y=rand(10:20, 100_000));

  730.118 μs (31 allocations: 2.53 MiB)

julia> @btime groupby(df, [:x, :y]);
  1.091 ms (38 allocations: 2.53 MiB)

julia> df = DataFrame(x=rand(1:1000, 100_000), y=rand(1000:2000, 100_000));

  1.009 ms (31 allocations: 2.53 MiB)

julia> @btime groupby(df, [:x, :y]);
  2.690 ms (38 allocations: 2.53 MiB)
```

PR (updated to latest commits):
```julia
jjulia> using DataFrames, BenchmarkTools

julia> df = DataFrame(x=rand(1:10, 100_000), y=rand(10:20, 100_000));

  186.219 μs (58 allocations: 784.46 KiB)

julia> @btime groupby(df, [:x, :y]);
  296.452 μs (85 allocations: 785.84 KiB)

julia> df = DataFrame(x=rand(1:1000, 100_000), y=rand(1000:2000, 100_000));

  187.802 μs (58 allocations: 785.43 KiB)

# Too many combinations, falling back to generic method
julia> @btime groupby(df, [:x, :y]);
  2.982 ms (98 allocations: 2.53 MiB)
```

### DIFF
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -45,7 +45,7 @@ function hashrows_col!(h::Vector{UInt},
     # which is always zero
     # also when the number of values in the pool is more than half the length
     # of the vector avoid using this path. 50% is roughly based on benchmarks
-    if firstcol && 2 * length(rp) < length(v)
+    if firstcol && Int64(2) * length(rp) < length(v)
         hashes = Vector{UInt}(undef, length(rp))
         @inbounds for (i, v) in zip(eachindex(hashes), rp)
             hashes[i] = hash(v)
@@ -119,6 +119,7 @@ Base.IndexStyle(::Type{<:IntegerRefarray{T}}) where {T} = Base.IndexStyle(T)
     if eltype(x.x) >: Missing && v === missing
         return x.replacement
     else
+        # Overflow is guaranteed not to happen by checks before calling the constructor
         return Int(v - x.offset)
     end
 end
@@ -162,7 +163,7 @@ function refpool_and_array(x::AbstractArray)
         # (note that it would be possible to allow minval and maxval to be outside of the
         # range supported by Int by adding a type parameter for minval to IntegerRefarray)
         if typemin(Int) < minval <= maxval < typemax(Int) &&
-            ngroups + 1 <= 2 * length(x) <= typemax(Int)
+            ngroups + 1 <= Int64(2) * length(x) <= typemax(Int)
             T = eltype(x) >: Missing ? Union{Int, Missing} : Int
             refpool′ = IntegerRefpool{T}(Int(ngroups))
             refarray′ = IntegerRefarray(x, Int(minval) - 1, Int(ngroups) + 1)
@@ -287,7 +288,7 @@ function row_group_slots(cols::NTuple{N, AbstractVector},
     # but it needs to remain reasonable compared with the size of the data frame.
     anydups = !all(allunique, refpools)
     if prod(big.(ngroupstup)) > typemax(Int) ||
-       ngroups > 2 * length(groups) ||
+       ngroups > Int64(2) * length(groups) ||
        anydups
         # In the simplest case, we can work directly with the reference codes
         newcols = (skipmissing && any(refpool -> eltype(refpool) >: Missing, refpools)) ||
--- a/src/dataframerow/utils.jl
+++ b/src/dataframerow/utils.jl
@@ -95,39 +95,55 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
         isequal_row(Base.tail(cols1), r1, Base.tail(cols2), r2)
 
 # Simple vector type for internal use which represents a virtual DataAPI.refpool
-# so that an integer vector can be its own DataAPI.refarray:
+# so that a vector of reals can be its own DataAPI.refarray:
 # this just allows telling the generic row_group_slots method for generic refpools
 # what is the minimum index and the number of (potential) groups
-struct IntegerRefpool{T} <: AbstractVector{T}
+# missing values are supported and represented used index max+1
+struct IntegerRefpool{T<:Union{Real, Missing}} <: AbstractVector{T}
     min::Int
     max::Int
 end
 
-Base.size(x::IntegerRefpool) = (x.max - x.min + 1,)
-Base.axes(x::IntegerRefpool) = (x.min:x.max,)
+Base.size(x::IntegerRefpool{T}) where {T} = (x.max - x.min + 1 + (T >: Missing),)
+Base.axes(x::IntegerRefpool{T}) where {T} = (x.min:(x.max + (T >: Missing)),)
 Base.IndexStyle(::Type{<:IntegerRefpool}) = Base.IndexLinear()
-@inline function Base.getindex(x::IntegerRefpool, i::Integer)
+@inline function Base.getindex(x::IntegerRefpool{T}, i::Integer) where T
     @boundscheck checkbounds(x, i)
-    return i - x.min + 1
+    if T >: Missing && i == x.max+1
+        return missing
+    else
+        return i - x.min + 1
+    end
 end
 Base.allunique(::IntegerRefpool) = true
 Base.issorted(::IntegerRefpool) = true
 
-function _refpool(x::AbstractArray)
+function refpool_and_array(x::AbstractArray)
     refpool = DataAPI.refpool(x)
+    refarray = DataAPI.refarray(x)
+
     if refpool !== nothing
-        return refpool
-    elseif x isa AbstractArray{<:Real} && !isempty(x) && all(isinteger, x)
-        minval, maxval = extrema(x)
+        return refpool, refarray
+    elseif x isa AbstractArray{<:Union{Real, Missing}} &&
+        all(v -> ismissing(v) | isinteger(v), x)
+        isempty(skipmissing(x)) && return nothing, nothing
+        minval, maxval = extrema(skipmissing(x))
         # Threshold chosen with the same rationale as the row_group_slots refpool method:
         # refpool approach is faster but we should not allocate too much memory either
         if maxval - minval + 1 <= 2 * length(x)
-            return IntegerRefpool{eltype(x)}(minval, maxval)
+            refpool′ = IntegerRefpool{eltype(x)}(minval, maxval)
+            if eltype(x) >: Missing
+                # Missing values go to last group with code maxval+1
+                refarray′ = Missings.replace(x, maxval+1)
+            else
+                refarray′ = x
+            end
+            return refpool′, refarray′
         else
-            return nothing
+            return nothing, nothing
         end
     else
-        return nothing
+        return nothing, nothing
     end
 end
 
@@ -140,18 +156,21 @@ end
 # 4) whether groups are already sorted
 # Optional `groups` vector is set to the group indices of each row (starting at 1)
 # With skipmissing=true, rows with missing values are attributed index 0.
-row_group_slots(cols::Tuple{Vararg{AbstractVector}},
-                hash::Val = Val(true),
-                groups::Union{Vector{Int}, Nothing} = nothing,
-                skipmissing::Bool = false,
-                sort::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool} =
-    row_group_slots(cols, _refpool.(cols), DataAPI.refarray.(cols),
-                    hash, groups, skipmissing, sort)
+function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
+                         hash::Val = Val(true),
+                         groups::Union{Vector{Int}, Nothing} = nothing,
+                         skipmissing::Bool = false,
+                         sort::Bool = false)::Tuple{Int, Vector{UInt}, Vector{Int}, Bool}
+    rpa = refpool_and_array.(cols)
+    refpools = first.(rpa)
+    refarrays = last.(rpa)
+    row_group_slots(cols, refpools, refarrays, hash, groups, skipmissing, sort)
+end
 
 # Generic fallback method based on open adressing hash table
 function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
-                         refpools::Any,
-                         refs::Any,
+                         refpools::Any,  # Ignored
+                         refarrays::Any, # Ignored
                          hash::Val = Val(true),
                          groups::Union{Vector{Int}, Nothing} = nothing,
                          skipmissing::Bool = false,
@@ -202,9 +221,11 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
 end
 
 # Optimized method for arrays for which DataAPI.refpool is defined and returns an AbstractVector
-function row_group_slots(cols::NTuple{N, <:AbstractVector},
-                         refpools::NTuple{N, <:AbstractVector},
-                         refs::NTuple{N, <:AbstractVector},
+function row_group_slots(cols::NTuple{N, AbstractVector},
+                         refpools::NTuple{N, AbstractVector},
+                         refs::NTuple{N, Union{AbstractVector{<:Real},
+                                               Missings.EachReplaceMissing{
+                                                   <:AbstractVector{<:Union{Real, Missing}}}}},
                          hash::Val{false},
                          groups::Union{Vector{Int}, Nothing} = nothing,
                          skipmissing::Bool = false,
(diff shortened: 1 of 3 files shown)


src/dataframerow/utils.jl
```diff
@@ -94,6 +94,43 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+
+function _refpool(x::AbstractArray)
+    refpool = DataAPI.refpool(x)
+    if refpool !== nothing
+        return refpool
+    elseif x isa AbstractArray{<:Real} && !isempty(x) && all(isinteger, x)
+        minval, maxval = extrema(x)
+        # Threshold chosen with the same rationale as the row_group_slots refpool method:
+        # refpool approach is faster but we should not allocate too much memory either
+        if maxval - minval + 1 <= 2 * length(x)
```
nalimilan, line 124: It seems that even above this threshold the `refpool` approach is faster. But I'm not sure how much memory we're willing to allocate for speed.

nalimilan:
Last commit will fail until JuliaData/Missings.jl#126.

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,59 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
     isequal(cols1[1][r1], cols2[1][r2]) &&
         isequal_row(Base.tail(cols1), r1, Base.tail(cols2), r2)
 
+# Simple vector type for internal use which represents a virtual DataAPI.refpool
+# so that a vector of reals can be its own DataAPI.refarray:
+# this just allows telling the generic row_group_slots method for generic refpools
+# what is the minimum index and the number of (potential) groups
+# missing values are supported and represented used index max+1
+struct IntegerRefpool{T<:Union{Real, Missing}} <: AbstractVector{T}
```
bkamins, line 102: we should disallow bare `Missing` I think

bkamins: or not? (I just started reading this PR 😄 )

bkamins: Also `Union{}` meets the restriction you specified.

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,59 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+    min::Int
+    max::Int
+end
+
+Base.size(x::IntegerRefpool{T}) where {T} = (x.max - x.min + 1 + (T >: Missing),)
+Base.axes(x::IntegerRefpool{T}) where {T} = (x.min:(x.max + (T >: Missing)),)
```
bkamins, line 108: we should check then upon construction that `x.max + 1` does not overflow. The same for `x.max - x.min + 1 + (T >: Missing)`. Alternatively here we can perform `checked_add` here (it is slower than normal one, but it probably does not matter here)

bkamins: same with `cheked_sub`

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,59 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+Base.allunique(::IntegerRefpool) = true
+Base.issorted(::IntegerRefpool) = true
+
+function refpool_and_array(x::AbstractArray)
+    refpool = DataAPI.refpool(x)
+    refarray = DataAPI.refarray(x)
+
+    if refpool !== nothing
+        return refpool, refarray
+    elseif x isa AbstractArray{<:Union{Real, Missing}} &&
```
bkamins, line 127: same here we allow `Union{}` and `Missing` eltype. Do we want this?

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,59 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+        all(v -> ismissing(v) | isinteger(v), x)
+        isempty(skipmissing(x)) && return nothing, nothing
+        minval, maxval = extrema(skipmissing(x))
+        # Threshold chosen with the same rationale as the row_group_slots refpool method:
```
bkamins, line 131: I think we can leave this threshold for now.

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,59 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+        # refpool approach is faster but we should not allocate too much memory either
+        if maxval - minval + 1 <= 2 * length(x)
+            refpool′ = IntegerRefpool{eltype(x)}(minval, maxval)
```
bkamins, line 134: we should check here if `minval` and `maxval` can be cast to `Int`. If not we should fall back to a slow path.

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,59 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+
+    if refpool !== nothing
+        return refpool, refarray
+    elseif x isa AbstractArray{<:Union{Real, Missing}} &&
+        all(v -> ismissing(v) | isinteger(v), x)
+        isempty(skipmissing(x)) && return nothing, nothing
+        minval, maxval = extrema(skipmissing(x))
+        # Threshold chosen with the same rationale as the row_group_slots refpool method:
+        # refpool approach is faster but we should not allocate too much memory either
+        if maxval - minval + 1 <= 2 * length(x)
```
bkamins, line 133: `maxval - minval + 1` can overflow.

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,59 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+            refpool′ = IntegerRefpool{eltype(x)}(minval, maxval)
+            if eltype(x) >: Missing
+                # Missing values go to last group with code maxval+1
+                refarray′ = Missings.replace(x, maxval+1)
```
bkamins, line 137: again - we should earlier check if `maxval+1` does not overflow `Int`.

bkamins:
Looks very nice. We only need to make sure we correctly disable trying this path in corner cases (like `BigInt` vector that has very large values, but with small range) as otherwise we will have overflow.

src/dataframerow/utils.jl
nalimilan: I don't think that's a problem. I'll add tests.

src/dataframerow/utils.jl
nalimilan: Yeah, `Missing` is caught by `isempty(skipmissing(x))`, and `Union{}` is either caught by the same check when empty, or triggers an error that will happen anyway when trying to access a value.

src/dataframerow/utils.jl
bkamins: OK - it is just my second reflex that in such `Union` cases `Missing` and `Union{}` are corner case that should be considered. If it is not a problem here then it is great.

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,68 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+# so that a vector of reals can be its own DataAPI.refarray:
+# this just allows telling the generic row_group_slots method for generic refpools
+# what is the minimum index and the number of (potential) groups
+# missing values are represented using index max+1
+struct IntegerRefpool{T<:Union{Int, Missing}} <: AbstractVector{T}
+    min::Int
+    max::Int
+    function IntegerRefpool{T}(min::Real, max::Real) where T<:Union{Int, Missing}
+        @assert max < typemax(Int) - 1
+        @assert typemin(Int) <= widen(max) - widen(min) + 1 < typemax(Int)
```
bkamins, line 107: Small suggestions:
1. check that `min` is less of equal than `max`
2. then no need to check `typemin(Int)`
3. also `+2` is needed as we potentially have `Missing` to handle
```suggestion
        @assert min <= max < typemax(Int) - 1
        @assert widen(max) - widen(min) + 2 < typemax(Int)
```

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,68 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+        new{T}(min, max)
+    end
+end
+
+Base.size(x::IntegerRefpool{T}) where {T} = (x.max - x.min + 1 + (T >: Missing),)
+Base.axes(x::IntegerRefpool{T}) where {T} = (x.min:(x.max + (T >: Missing)),)
+Base.IndexStyle(::Type{<:IntegerRefpool}) = Base.IndexLinear()
+@inline function Base.getindex(x::IntegerRefpool{T}, i::Real) where T
```
bkamins, line 115: just to be sure - do we have a test when `i` is e.g. `1.0` or `true` to make sure that we error in this case? In general why do you want to allow `i` to be `Real` rather than `Union{Signed, Unsigned}`? (actually the same question is to line 105 with inner constructor)

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,68 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+Base.allunique(::IntegerRefpool) = true
+Base.issorted(::IntegerRefpool) = true
+
+function refpool_and_array(x::AbstractArray)
+    refpool = DataAPI.refpool(x)
+    refarray = DataAPI.refarray(x)
+
+    if refpool !== nothing
+        return refpool, refarray
+    elseif x isa AbstractArray{<:Union{Real, Missing}} &&
```
bkamins, line 132: again - why do we allow `Real` here?

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,68 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+        all(v -> ismissing(v) | isinteger(v), x)
+        isempty(skipmissing(x)) && return nothing, nothing
+        minval, maxval = extrema(skipmissing(x))
+        # Threshold chosen with the same rationale as the row_group_slots refpool method:
+        # refpool approach is faster but we should not allocate too much memory either
+        # We also have to avoid overflow
+        if typemin(Int) <= maxval + 1 < typemax(Int) &&
+            typemin(Int) <= minval <= typemax(Int) &&
+            widen(maxval) - widen(minval) + 1 <= 2 * length(x) < typemax(Int)
```
bkamins, line 141: as above you have to use `+2` to allow to store `missing`
```suggestion
            widen(maxval) - widen(minval) + 2 <= 2 * length(x) < typemax(Int)
```

src/dataframerow/utils.jl
```diff
@@ -163,8 +230,12 @@ function row_group_slots(cols::Tuple{Vararg{AbstractVector}},
 end
 
 # Optimized method for arrays for which DataAPI.refpool is defined and returns an AbstractVector
-function row_group_slots(cols::NTuple{N, <:AbstractVector},
-                         refpools::NTuple{N, <:AbstractVector},
+function row_group_slots(cols::NTuple{N, AbstractVector},
+                         refpools::NTuple{N, AbstractVector},
+                         refarrays::NTuple{N,
```
bkamins, line 235: These line breaks are ugly, but probably we cannot help it.

bkamins:
I would also add a test that on 32-bit machine uses slow path and on 64-bit machine uses fast path (as `Int` is defined differently on both) and another test that on 64-bit machine uses slow path because the values are too large.

nalimilan:
I've added checks and tests for overflows. I've also bumped into an already existing bug with `skipmissing=true` when there are only missing values so I've added more tests for that.

bkamins:
Have you seen my comments I left today in the morning?

nalimilan:
> I would also add a test that on 32-bit machine uses slow path and on 64-bit machine uses fast path (as `Int` is defined differently on both) and another test that on 64-bit machine uses slow path because the values are too large.

I've added tests that the fast path is used just below `typemax(Int)`, and correctness tests just above it.

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,84 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+Base.size(x::IntegerRefarray) = size(x.x)
+Base.axes(x::IntegerRefarray) = axes(x.x)
+Base.IndexStyle(::Type{<:IntegerRefarray{T}}) where {T} = Base.IndexStyle(T)
+@inline function Base.getindex(x::IntegerRefarray, i)
+    @boundscheck checkbounds(x.x, i)
+    @inbounds v = x.x[i]
+    if eltype(x.x) >: Missing && v === missing
+        return x.replacement
+    else
+        return Int(v - x.offset)
```
bkamins, line 122: are we sure `v - x.offset` will not overflow here?

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,84 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+        # Threshold chosen with the same rationale as the row_group_slots refpool method:
+        # refpool approach is faster but we should not allocate too much memory either
+        # We also have to avoid overflow, including with ngroups + 1 for missing values
+        # (note that it would be possible to allow minval and maxval to be outside of the
+        # range supported by Int by adding a type parameter for minval to IntegerRefarray)
+        if typemin(Int) < minval <= maxval < typemax(Int) &&
+            ngroups + 1 <= 2 * length(x) <= typemax(Int)
+            T = eltype(x) >: Missing ? Union{Int, Missing} : Int
+            refpool′ = IntegerRefpool{T}(Int(ngroups))
+            refarray′ = IntegerRefarray(x, Int(minval) - 1, Int(ngroups) + 1)
```
bkamins, line 168: OK - I guess this line answers my question above. If we initialize `offset` to what you propose then we are safe. Maybe just add a comment above what `offset` means for `IntegerRefarray` type?

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,84 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+        !isempty(skipmissing(x))
+        minval, maxval = extrema(skipmissing(x))
+        ngroups = big(maxval) - big(minval) + 1
+        # Threshold chosen with the same rationale as the row_group_slots refpool method:
+        # refpool approach is faster but we should not allocate too much memory either
+        # We also have to avoid overflow, including with ngroups + 1 for missing values
+        # (note that it would be possible to allow minval and maxval to be outside of the
+        # range supported by Int by adding a type parameter for minval to IntegerRefarray)
+        if typemin(Int) < minval <= maxval < typemax(Int) &&
+            ngroups + 1 <= 2 * length(x) <= typemax(Int)
```
bkamins, line 165: it is interesting here that `2*length(x)` can overflow if we are on 32 bit machine, but then it will be negative so the condition is correct (simply we will not use a fast path then). But if you want to be correct maybe better write `Int64(2) * length(x)`?

src/dataframerow/utils.jl
```diff
@@ -94,6 +94,84 @@ isequal_row(cols1::Tuple{Vararg{AbstractVector}}, r1::Int,
...
+Base.issorted(::IntegerRefpool) = true
+
+function refpool_and_array(x::AbstractArray)
+    refpool = DataAPI.refpool(x)
+    refarray = DataAPI.refarray(x)
+
+    if refpool !== nothing
+        return refpool, refarray
+    elseif x isa AbstractArray{<:Union{Real, Missing}} &&
+        all(v -> ismissing(v) | isinteger(v), x) &&
```
bkamins, line 155: Have you checked that using `|` here is is optimized by the compiler correctly? What I mean is that when `eltype(x) <: Integer` is this check a no-op?

bkamins (approved):

bkamins:
Looks good. Thank you. Can you just please make sure that in the points I have just raised we are clear what happens?

src/dataframerow/utils.jl
nalimilan: Yes, that's a good point, but I had tested it and it works (which is really nice).

src/dataframerow/utils.jl
nalimilan: I've added a comment.

src/dataframerow/utils.jl
nalimilan: Woops, that's indeed not super clean. I've switched to `Int64` (including in two similar occurrences).

bkamins:
Thank you. Looks good now. These overflow things are tricky.

PR #2615, opened 2021-01-31
Fix missing mention of transform in manual

nalimilan:
Also change heading to match the pattern used for other sections.

Spotted while looking at https://github.com/JuliaData/DataFrames.jl/pull/2598.


bkamins (approved):

bkamins:
Thank you!

PR #2635, opened 2021-02-28
Some breaking changes expected

PallHaraldsson:
@bkamins. On Nov 27th you planned for no breaking changes, but seem to have changed your mind with e.g.:

https://github.com/JuliaData/DataFrames.jl/issues/2620

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Breaking changes
 
-* No breaking changes are planned for v1.0 release
+* A few breaking changes are planned for v1.0 release
 
 ## Bug fixes
 


bkamins:
This issue does not mean we want to make these changes but that people are asking to make these changes. It is labelled 1.0, as if we want to break something it has to be considered before making 1.0 release.

See my comment in https://github.com/JuliaData/DataFrames.jl/pull/2633, where I oppose making breaking changes.

However, if we decide to merge #2633 then the change you propose will be made there. My current thinking is that we should not introduce breaking changes.

Because of this reasons I am closing this PR, as what you propose is tracked in #2633.

bkamins:
CC @nalimilan

PallHaraldsson:
It wasn't about that issue only, I just thought I had seen a pattern, and the changelog not updated to reflect.

There are 7 issues marked breaking, thereof 3 new after Nov. including one from "14 hours ago".

Not all of them from you, but if you're possibly considering some changes then maybe at least be silent on it in the changelog, to not promise too much?

I'm ok with either, but one question. If none of these breaking changes are merged, will they just be closed, or still be on the table for 2.0? I don't think it's on the horizon, would be strange with 1.0 not out. I didn't look closely at any proposal, so I'm not taking sides if any of them is needed or a good idea.

bkamins:
>  if you're possibly considering some changes

This is the point - I am not considering making breaking changes.

However, there are several points:
1. I am not the only person involved in DataFrames.jl and other people might outvote me
2. We issue issues and pool requests to keep track of valid ideas. They might be merged now, in the future or never. However, if someone submits a proposal it goes through evaluation. If we label something `breaking` it means that implementing is "as proposed" at the moment of proposal would be breaking (but we might eventually come to some other solution that is not breaking or reject the proposal). In particular - some things are on the "border" of being considered breaking (there are always discussions about what is breaking and what is not in corner cases)
3. I prefer to firmly state "no breaking changes" as then it is clear that if someone wants a breaking change the proposal will face a significant objection before being accepted.
4. The phrase you refer to is in `main` branch not in released branch so it is meant for developers not for users. What is binding for users is presented in released versions of DataFrames.jl.

Still - thank you fro raising this point, as it is easy to open a flood gate with breaking proposals which I want to avoid.

PR #2641, opened 2021-03-04
Add faster semi and antijoin

### LINKED ISSUE #2340: Faster joins meta issue

bkamins:
Fixes #2340 

This is the last PR for making joins faster.
Here are the benchmarks against 0.22.5 release. As before we are either much faster or comparable.
[semi_022_5.txt](https://github.com/JuliaData/DataFrames.jl/files/6085850/semi_022_5.txt)
[semi_pr.txt](https://github.com/JuliaData/DataFrames.jl/files/6085852/semi_pr.txt)

The only small problem is code duplication. I would prefer to avoid it, but first I wanted to make sure we have thing correct and fast. The problem is that duplicate methods get different arguments and are minimally different (in other words: we use exactly the same high level logic, but collect and pass around different information as the resulting join is different since it involves only left table).

### DIFF
--- a/src/join/core.jl
+++ b/src/join/core.jl
@@ -1,3 +1,5 @@
+### Common preprocessing
+
 struct OnColRow{T}
     row::Int
     cols::T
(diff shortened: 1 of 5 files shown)


src/join/core.jl
```diff
@@ -115,6 +117,88 @@ function map2refs(x::AbstractVector, invrefpool)
...
 
+function preprocess_columns(joiner::DataFrameJoiner)
+    right_len = length(joiner.dfr_on[!, 1])
+    left_len = length(joiner.dfl_on[!, 1])
+    right_shorter = right_len <= left_len
+
+    left_cols = collect(eachcol(joiner.dfl_on))
+    right_cols = collect(eachcol(joiner.dfr_on))
+
+    # if column of a longer table supports DataAPI.refpool and DataAPI.invrefpool
```
nalimilan, line 128:
```suggestion
    # if column of the longer table supports DataAPI.refpool and DataAPI.invrefpool
```

src/join/core.jl
```diff
@@ -115,6 +117,88 @@ function map2refs(x::AbstractVector, invrefpool)
...
+    end
+
+    # TODO:
+    # If DataAPI.invrefpool vectors are found in the "on" columns
+    # then potentially the following optimizations can be done:
+    # 1. identify rows in shorter table that should be dropped
+    # 2. develop custom _innerjoin_sorted and _innerjoin_unsorted that
+    #    drop rows from shorter table that do not match rows from longer table based on
+    #    PooledArray refpool check
+    # This optimization significantly complicates the code (especially sorted path).
```
nalimilan, line 187:
```suggestion
    # This optimization would significantly complicate the code (especially sorted path).
```

src/join/core.jl
```diff
@@ -428,84 +509,306 @@ function _innerjoin_postprocess_int(left::AbstractVector{<:Union{Integer, Missin
...
 
-        # we disallow using sorted branch for some columns that theoretically
-        # could be safely sorted (e.g. having Any eltype but holding strings)
-        # for safety reasons assuming that such cases will be rare in practice
-        disallow_sorted = true
+    @inbounds for (idx_l, val_l) in enumerate(left)
+        if ismissing(val_l)
+            idx_r = dict[end]
+            if idx_r > 0
+                if right_shorter
```
nalimilan, line 656: Did you make `right_shorter` a `Val` just because of this? I think LLVM should be able to hoist the check so using `Val` shouldn't be necessary.

Also, maybe something like `r  = idx_r > 0; seen_rows[idx_l] |= r` could be slightly faster if that avoids a branch? `val_l === missing` might also be faster than `ismissing(val_l)` (until Julia 1.7).

src/join/core.jl
```diff
@@ -428,84 +509,306 @@ function _innerjoin_postprocess_int(left::AbstractVector{<:Union{Integer, Missin
...
+    if right_shorter
+        @assert length(left) == length(seen_rows)
+    else
+        @assert length(right) == length(seen_rows)
+    end
+
+    dict = Dict{T, Int}()
+
+    right_len = length(right)
+    sizehint!(dict, 2 * min(right_len, typemax(Int) >> 2))
```
nalimilan, line 602: Can you add a comment explaining how the size is chosen? Why the × 2 in particular?

src/join/core.jl
```diff
@@ -428,84 +509,306 @@ function _innerjoin_postprocess_int(left::AbstractVector{<:Union{Integer, Missin
...
+            end
+            n = newn
+        end
+    end
+
+    return seen_rows
+end
+
+function find_semi_rows(joiner::DataFrameJoiner)
+
```
nalimilan, line 805:
```suggestion
```

src/join/core.jl
```diff
@@ -428,84 +509,306 @@ function _innerjoin_postprocess_int(left::AbstractVector{<:Union{Integer, Missin
...
 
-    for (lc, rc) in zip(left_cols, right_cols)
-        @assert length(lc) == left_len
-        @assert length(rc) == right_len
-        lc_et = nonmissingtype(eltype(lc))
-        rc_et = nonmissingtype(eltype(rc))
+    val_range = big(maxv) - big(minv)
+    if val_range > typemax(Int) - 3 || val_range ÷ 2 > max(64, length(right)) ||
+       minv < typemin(Int) + 2 || maxv > typemax(Int) - 3
+       return _semijoin_unsorted(left, right, seen_rows, Val(right_shorter))
```
nalimilan, line 636: This isn't tested currently.

nalimilan (commented):
Impressive timings!

src/join/core.jl
bkamins: I have added a comment

src/join/core.jl
bkamins: added

src/join/core.jl
bkamins: Thank you for the comment:
* indeed `=== missing` is faster; I have changed this
* using `|= r` can avoid branch in only one of the cases, in the other we use `idx_r` also for indexing so it had to stay; I have changed this
* regarding `Val` - I will refactor the code so that we do not need it

bkamins:
I have refactored the code following your suggestions. The code is now faster in some specific cases (as I was able to switch from `Dict` to `Set` for some code paths so less memory is used). This mostly affects `semijoin` and `antijoin` when left and right tables have approximately similar length.

CI fails because of breaking changes in 1.7. I will fix it in separate PRs.

src/join/core.jl
```diff
@@ -274,6 +274,14 @@ function _innerjoin_unsorted(left::AbstractArray, right::AbstractArray{T}) where
     dict = Dict{T, Int}()
 
     right_len = length(right)
+    # we make sure that:
+    # * we do not preallocate dict to size larger than half of size of Int
+    #   (this is relevant in 32 bit architectures)
```
nalimilan, line 279:
```suggestion
    # * we do not preallocate dict of size larger than half of size of Int
    #   (this is relevant in 32 bit architectures)
```
(I don't want to imagine allocating a 2^63 length vector on 64-bit. :-D)

src/join/core.jl
```diff
@@ -587,98 +595,108 @@ end
...
     val_range = big(maxv) - big(minv)
     if val_range > typemax(Int) - 3 || val_range ÷ 2 > max(64, length(right)) ||
        minv < typemin(Int) + 2 || maxv > typemax(Int) - 3
-       return _semijoin_unsorted(left, right, seen_rows, Val(right_shorter))
+       return _semijoin_unsorted(left, right, seen_rows, right_shorter)
     end
 
     offset = 1 - Int(minv) # we are now sure it does not overflow
     len = Int(maxv) - Int(minv) + 2
     dict = zeros(Int, len)
```
nalimilan, line 652: Calling this "dict" is a bit confusing given that it's a vector. :-)

nalimilan (approved):

src/join/core.jl
bkamins: OK - I have changed it to `group_map`. Is this better?

nalimilan: Yes!

bkamins:
Thank you!

PR #2646, opened 2021-03-06
Fix CategoricalArrays.jl in Project.toml

bkamins:
@KristofferC - I have a question regarding Project.toml.

If a package is in `[extras]` then is `[compat]` entry allowed for it?

In https://julialang.github.io/Pkg.jl/v1/toml-files/#The-[compat]-section it is stated that only `[deps]` entries are allowed in `[compat]`, but maybe also `[extras]` are allowed and the description given there should be fixed? Thank you!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,6 +23,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
+CategoricalArrays = "0.9.3"
 Compat = "3.17"
 DataAPI = "1.4"
 InvertedIndices = "1"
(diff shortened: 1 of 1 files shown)


simeonschaub:
> If a package is in [extras] then is [compat] entry allowed for it?

Yes, that should work.

bkamins:
So I understand the documentation in Pkg.jl should be fixed - right?

simeonschaub:
Yes. Could you open an issue?

bkamins:
https://github.com/JuliaLang/Pkg.jl/issues/2424

Project.toml
```diff
@@ -24,7 +23,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
-CategoricalArrays = "0.8.3, 0.9"
```
nalimilan, line 27: Do we really need 0.9?

bkamins: we need 0.9.3 for `unwrap`

nalimilan (approved):

bkamins:
Thank you

PR #2648, opened 2021-03-08
fix repeat and some other minor issues

bkamins:
Some fixes after https://github.com/JuliaLang/julia/pull/31829.
Mainly restricting the signature of `repeat`.

Additionally one dangling `get` to `unwrap` for CategoricalArrays.jl compatibility I missed earlier.

@mbauman

The https://github.com/JuliaLang/julia/pull/31829 broke the following thing in Julia Base that was not covered by tests (fortunately we had appropriate tests in DataFrames.jl):
```
julia> x = [1,2,3]
3-element Vector{Int64}:
 1
 2
 3

julia> repeat(x, inner=1,outer=1)
3-element Vector{Int64}:
 1
 2
 3

julia> repeat(x, inner=1,outer=false) # it works but I think it should be an error
Int64[]

julia> repeat(x, inner=false,outer=1) # previously it worked
ERROR: ArgumentError: invalid index: false of type Bool
```

### DIFF
--- a/test/subdataframe.jl
+++ b/test/subdataframe.jl
@@ -295,6 +295,9 @@ end
 
     if VERSION > v"1.6"
         @test_throws ArgumentError SubDataFrame(sdf, Integer[true, true, true], :)
+    else
+        @test SubDataFrame(sdf, Integer[true, true, true], :) ==
+              SubDataFrame(sdf, [1, 1, 1], :)
     end
 end
 
--- a/test/subdataframe.jl
+++ b/test/subdataframe.jl
@@ -293,7 +293,9 @@ end
     @test_throws ArgumentError SubDataFrame(sdf, true, :)
     @test_throws ArgumentError SubDataFrame(sdf, Integer[true], 1)
 
-    @test_throws ArgumentError SubDataFrame(sdf, Integer[true, true, true], :)
+    if VERSION > v"1.6"
+        @test_throws ArgumentError SubDataFrame(sdf, Integer[true, true, true], :)
+    end
 end
 
 end # module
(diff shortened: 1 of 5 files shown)


bkamins:
Actually maybe `Bool` in `repeat` should be allowed as it is not an indexing context and it is clear what will be the result of the operation. Let us wait what @mbauman comments for Julia Base.

bkamins:
As commented above - I have thought about it and I think that we should accept `Bool` as in this context it should be treated as `Integer`.

test/subdataframe.jl
```diff
@@ -293,8 +293,9 @@ end
     @test_throws ArgumentError SubDataFrame(sdf, true, :)
     @test_throws ArgumentError SubDataFrame(sdf, Integer[true], 1)
 
-    # this is an error in Julia Base
-    SubDataFrame(sdf, Integer[true, true, true], :) == SubDataFrame(sdf, [1, 1, 1], :)
+    if VERSION > v"1.6"
+        @test_throws ArgumentError SubDataFrame(sdf, Integer[true, true, true], :)
+    end
```
nalimilan, line 298: And what happens before 1.6? :-)

bkamins: Before (and including 1.6) 1.7 it works - see the old test that got removed.

nalimilan: I mean, shouldn't this be tested too?

bkamins: OK - I have reverted it (although it was kept mainly to keep track of the bug in Julia Base).

A follow up PR in Julia Base that should be fixed is https://github.com/JuliaLang/julia/pull/39962 (I am pinging you as it seem no one commented on it).

nalimilan (approved):

Issue #659, opened 2014-07-27
Add a function for combine/vcat with source names


johnmyleswhite:
Should we add a function like `vcat(df1, df2, df3, source = ["X", "Y", "Z"])` which concatenates DataFrames while also labeling their origins? In this example, the output would be equivalent to doing:

```
df1_alt = copy(df1)
df1_alt[:source] = "X"
df2_alt = copy(df2)
df2_alt[:source] = "Y"
df3_alt = copy(df3)
df3_alt[:source] = "Z"

vcatted = vcat(df1_alt, df2_alt, df3_alt)
```

The idea is to preserve the semantics of `Base.vcat`, but allow the introduction of a custom column (whose name comes from a keyword arg) that denotes the origin of each subset of data.

quinnj:
Anyone still interested in this? Could be useful if someone wants to take a stab at it. @cjprybol ?

bkamins:
@oxinabox seems to also want it in `join` 😄.

bkamins:
> Anyone still interested in this?

I have added it in #2649

PR #2649, opened 2021-03-09
add vcat with source; deprecate indicator in joins in favor of source

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/659

I was prompted by https://discourse.julialang.org/t/would-it-help-to-have-a-tool-that-automatically-determines-which-issues-should-be-closed/56274/16 :).

The PR is relatively simple so we can add it in 1.0 release I think.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1439,9 +1439,9 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
 """
     vcat(dfs::AbstractDataFrame...;
          cols::Union{Symbol, AbstractVector{Symbol},
-                     AbstractVector{<:AbstractString}}=:setequal),
+                     AbstractVector{<:AbstractString}}=:setequal,
          source::Union{Nothing, SymbolOrString,
-                       Pair{<:SymbolOrString, <:AbstractVector}}=nothing
+                       Pair{<:SymbolOrString, <:AbstractVector}}=nothing)
 
 Vertically concatenate `AbstractDataFrame`s.
 
@@ -1459,13 +1459,13 @@ The `cols` keyword argument determines the columns of the returned data frame:
 * A vector of `Symbol`s or strings: only listed columns are kept.
   Columns not present in some data frames are filled with `missing` where necessary.
 
-The `source` keyword argument, if not `nothing` (the default) specifies the additional
-column to be added as a first column in the resulting data frame that will identify
-the source data frame. In this case `source` can be a `Symbol` or an `AbstractString`
-in which case the identifier will be the number of the passed source data frame.
-Alternatively `source` can be a `Pair` consisting of a `Symbol` or an `AbstractString`
-and a vector specifying the data frame identifiers (the identifiers do not have to be
-unique). The name of the source column name is not allowed to be present in any
+The `source` keyword argument, if not `nothing` (the default), specifies the additional
+column to be added in the first position in the resulting data frame that will identify
+the source data frame. It can be a `Symbol` or an `AbstractString`,
+in which case the identifier will be the number of the passed source data frame, or
+a `Pair` consisting of a `Symbol` or an `AbstractString`
+and of a vector specifying the data frame identifiers (which do not have to be unique).
+The name of the source column is not allowed to be present in any
 source data frame.
 
 The order of columns is determined by the order they appear in the included data
@@ -1615,13 +1615,13 @@ function Base.reduce(::typeof(vcat),
 
         idx = findfirst(df -> columnindex(df, col) > 0, dfs)
         if idx !== nothing
-            throw(ArgumentError("source column name exists in data frame " *
+            throw(ArgumentError("source column name :$source exists in data frame " *
                                 " passed in position $idx"))
         end
 
         if len != length(vals)
-            throw(ArgumentError("number of passed source column identifiers " *
-                                "does not match the number of passed columns"))
+            throw(ArgumentError("number of passed source identifiers ($(length(vals)))" *
+                                "does not match the number of data frames ($len)"))
         end
 
         dfs′ = Vector{AbstractDataFrame}(undef, len)
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1440,8 +1440,8 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
     vcat(dfs::AbstractDataFrame...;
          cols::Union{Symbol, AbstractVector{Symbol},
                      AbstractVector{<:AbstractString}}=:setequal,
-         indicator::Union{Nothing, SymbolOrString,
-                          Pair{<:SymbolOrString, <:AbstractVector}}=nothing)
+         source::Union{Nothing, SymbolOrString,
+                       Pair{<:SymbolOrString, <:AbstractVector}}=nothing)
 
 Vertically concatenate `AbstractDataFrame`s.
 
@@ -1459,13 +1459,13 @@ The `cols` keyword argument determines the columns of the returned data frame:
 * A vector of `Symbol`s or strings: only listed columns are kept.
   Columns not present in some data frames are filled with `missing` where necessary.
 
-The `indicator` keyword argument, if not `nothing` (the default), specifies the additional
-column to be added in the first position in the resulting data frame that will identify
-the source data frame. It can be a `Symbol` or an `AbstractString`,
-in which case the identifier will be the number of the passed source data frame, or
-a `Pair` consisting of a `Symbol` or an `AbstractString`
-and of a vector specifying the data frame identifiers (which do not have to be unique).
-The name of the source column is not allowed to be present in any
+The `source` keyword argument, if not `nothing` (the default), specifies the
+additional column to be added in the last position in the resulting data frame
+that will identify the source data frame. It can be a `Symbol` or an
+`AbstractString`, in which case the identifier will be the number of the passed
+source data frame, or a `Pair` consisting of a `Symbol` or an `AbstractString`
+and of a vector specifying the data frame identifiers (which do not have to be
+unique). The name of the source column is not allowed to be present in any
 source data frame.
 
 The order of columns is determined by the order they appear in the included data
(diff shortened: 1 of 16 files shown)


bkamins:
CI error is unrelated and is fixed in https://github.com/JuliaData/DataFrames.jl/pull/2648

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1457,6 +1459,15 @@ The `cols` keyword argument determines the columns of the returned data frame:
 * A vector of `Symbol`s or strings: only listed columns are kept.
   Columns not present in some data frames are filled with `missing` where necessary.
 
+The `source` keyword argument, if not `nothing` (the default) specifies the additional
+column to be added as a first column in the resulting data frame that will identify
+the source data frame. In this case `source` can be a `Symbol` or an `AbstarctString`
```
ericphanson, line 1464:
```suggestion
the source data frame. In this case `source` can be a `Symbol` or an `AbstractString `
```

pdeffebach:
Should this be called `indicator` for consistency with `join`s? See [here](https://dataframes.juliadata.org/latest/man/joins/)

bkamins:
> Should this be called `indicator` for consistency with `join`s?

I am OK with `indicator`. @nalimilan - what do you think?

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1439,7 +1439,9 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
 """
     vcat(dfs::AbstractDataFrame...;
          cols::Union{Symbol, AbstractVector{Symbol},
-                     AbstractVector{<:AbstractString}}=:setequal)
+                     AbstractVector{<:AbstractString}}=:setequal),
+         source::Union{Nothing, SymbolOrString,
+                       Pair{<:SymbolOrString, <:AbstractVector}}=nothing
```
nalimilan, line 1444:
```suggestion
                     AbstractVector{<:AbstractString}}=:setequal,
         source::Union{Nothing, SymbolOrString,
                       Pair{<:SymbolOrString, <:AbstractVector}}=nothing)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1457,6 +1459,15 @@ The `cols` keyword argument determines the columns of the returned data frame:
...
   Columns not present in some data frames are filled with `missing` where necessary.
 
+The `source` keyword argument, if not `nothing` (the default) specifies the additional
+column to be added as a first column in the resulting data frame that will identify
+the source data frame. In this case `source` can be a `Symbol` or an `AbstractString`
+in which case the identifier will be the number of the passed source data frame.
+Alternatively `source` can be a `Pair` consisting of a `Symbol` or an `AbstractString`
+and a vector specifying the data frame identifiers (the identifiers do not have to be
+unique). The name of the source column name is not allowed to be present in any
+source data frame.
```
nalimilan, line 1469:
```suggestion
The `source` keyword argument, if not `nothing` (the default), specifies the additional
column to be added in the first position in the resulting data frame that will identify
the source data frame. It can be a `Symbol` or an `AbstractString`,
in which case the identifier will be the number of the passed source data frame, or
a `Pair` consisting of a `Symbol` or an `AbstractString`
and of a vector specifying the data frame identifiers (which do not have to be unique).
The name of the source column is not allowed to be present in any
source data frame.
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,94 @@ julia> vcat(df1, df3, cols=:intersect)
...
+        len = length(dfs)
+        if source isa SymbolOrString
+            col, vals = source, 1:len
+        else
+            @assert source isa Pair{<:SymbolOrString, <:AbstractVector}
+            col, vals = source
+        end
+
+        idx = findfirst(df -> columnindex(df, col) > 0, dfs)
+        if idx !== nothing
```
nalimilan, line 1617:
```suggestion
        if any(df -> columnindex(df, col) > 0, dfs)
```
Also would `haskey` work too? Just for the style. :-)

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,94 @@ julia> vcat(df1, df3, cols=:intersect)
...
+            throw(ArgumentError("source column name exists in data frame " *
```
nalimilan, line 1618:
```suggestion
            throw(ArgumentError("source column name :$source exists in data frame " *
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,94 @@ julia> vcat(df1, df3, cols=:intersect)
...
+                                " passed in position $idx"))
+        end
+
+        if len != length(vals)
+            throw(ArgumentError("number of passed source column identifiers " *
+                                "does not match the number of passed columns"))
```
nalimilan, line 1624:
```suggestion
            throw(ArgumentError("number of passed source identifiers ($(length(vals))" *
                                "does not match the number of data frames ($len)"))
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,94 @@ julia> vcat(df1, df3, cols=:intersect)
...
+        end
+
+        dfs′ = Vector{AbstractDataFrame}(undef, len)
+        for (i, (v, df)) in enumerate(zip(vals, dfs))
+            dfs′[i] = insertcols!(copy(df, copycols=false), 1, col => Ref(v))
+        end
```
nalimilan, line 1630: Probably not a big deal, but wouldn't it be more efficient (and not really more complex) to create the column only after concatenating?

test/dataframe.jl
```diff
@@ -1723,6 +1723,30 @@ end
...
+    df2 = DataFrame(A=4:6, B=4:6)
+    df3 = DataFrame(A=7:9, C=7:9)
+    df4 = DataFrame()
+
+    for col in [:source, "source"]
+        @test vcat(df1, df2, df3, df4, cols=:union, source=col) ≅
+              DataFrame(source=[1, 1, 1, 2, 2, 2, 3, 3, 3, 4],
+                        A=[1:9; missing], B=[1:6; fill(missing, 4)],
+                        C=[fill(missing, 6); 7:9; missing])
+        res = vcat(df1, df2, df3, df4, cols=:union, source=col => categorical(-4:-1))
```
nalimilan, line 1737: You really like taking risks! :-D

test/dataframe.jl
```diff
@@ -1723,6 +1723,30 @@ end
...
 end
 
+@testset "vcat with source" begin
+    df1 = DataFrame(A=1:3, B=1:3)
+    df2 = DataFrame(A=4:6, B=4:6)
+    df3 = DataFrame(A=7:9, C=7:9)
+    df4 = DataFrame()
+
+    for col in [:source, "source"]
+        @test vcat(df1, df2, df3, df4, cols=:union, source=col) ≅
```
nalimilan, line 1733: Just so that `CategoricalArray` isn't the only tested type:
```suggestion
        @test vcat(df1, df2, df3, df4, cols=:union, source=col) ≅
            vcat(df1, df2, df3, df4, cols=:union, source=[1, 2, 3, 4]) ≅
```

nalimilan:
Consistency with joins sounds like a good reason to use `indicator`, even though it isn't super explicit. For now these are a bit different, but I imagine at some point we could allow passing a collection of names to `*join` to choose the values to use, and extend support to more than two data frames, which would make it more similar to this PR.

nalimilan:
Regarding the API, have you considered something like `vcat(:A => df1, :B => df2, :C => df3, indicator=:source)` instead? I guess which version is most convenient depends on whether you have all the names in a vector already or whether you write them by hand. We could support both if needed.

bkamins:
I analyzed it before making the PR.

We could allow:
```
vcat(:A => df1, :B => df2, :C => df3, indicator=:source)
```
but I am hesitant to add it as currently:
```
julia> vcat(:a => DataFrame(a=1), :b => DataFrame(b=2))
2-element Vector{Pair{Symbol, DataFrame}}:
 :a => 1×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │     1
 :b => 1×1 DataFrame
 Row │ b     
     │ Int64 
─────┼───────
   1 │     2
```
(and this makes sense and is expected)

So adding a kwarg would significantly affect the produced result, which I think is not desirable.

src/abstractdataframe/abstractdataframe.jl
bkamins: 1. I cannot use `any` as I need `idx` variable in the error message.
2. `AbstractDataFrame` does not support `AbstractDict` API and it does not have `haskey` method in particular

src/abstractdataframe/abstractdataframe.jl
bkamins: This is a design decision related with an empty data frame. As you can see in tests and examples, if you pass `DataFrame()` to `vcat` now a single row is created with all `missing` values except indicator column. Otherwise we would drop such a data frame.

But maybe dropping it is preferable. What do you think?

test/dataframe.jl
bkamins: well - I wanted to make sure we correctly instantiate the source column.

bkamins:
> I imagine at some point we could allow passing a collection of names to *join to choose the values to use, and extend support to more than two data frames, which would make it more similar to this PR.

Not doing this in `join` was not laziness but the issue that it is just different. If you pass e.g. three columns you can have the following outcomes:
* only 1
* only 2
* only 3
* 1 and 2
* 1 and 3
* 1, 2, and 3

Which would all have to be produced, which I thought would be more confusing than clarifying.

Similarly for two columns passed we have three values (left, right, and both) so it is not the same.

But of course the function of the kwarg is similar (clearly indicate the source data frame) so I would not oppose to syncing it between `join` and `vcat`.

Given these considerations what do you think. Should we make them consistent?

src/abstractdataframe/abstractdataframe.jl
bkamins: I have thought about it and concluded that it is better to drop it. I will change the implementation.

bkamins:
and I will change the kwarg do `indicator` as suggested.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1439,7 +1439,9 @@ Base.hcat(df1::AbstractDataFrame, df2::AbstractDataFrame, dfn::AbstractDataFrame
 """
     vcat(dfs::AbstractDataFrame...;
          cols::Union{Symbol, AbstractVector{Symbol},
-                     AbstractVector{<:AbstractString}}=:setequal)
+                     AbstractVector{<:AbstractString}}=:setequal,
+         indicator::Union{Nothing, SymbolOrString,
+                          Pair{<:SymbolOrString, <:AbstractVector}}=nothing)
```
nalimilan, line 1444: We don't export `SymbolOrString`:
```suggestion
         indicator::Union{Nothing, Symbol, AbstractString,
                          Pair{<:Union{Symbol, AbstractString}, <:AbstractVector}}=nothing)
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,100 @@ julia> vcat(df1, df3, cols=:intersect)
...
+            col, vals = indicator, 1:len
+        else
+            @assert indicator isa Pair{<:SymbolOrString, <:AbstractVector}
+            col, vals = indicator
+        end
+
+        if columnindex(res, col) > 0
+            idx = findfirst(df -> columnindex(df, col) > 0, dfs)
+            @assert idx !== nothing
+            throw(ArgumentError("indicator column name :$col exists in data frame " *
```
nalimilan, line 1616:
```suggestion
            throw(ArgumentError("indicator column name :$col already exists in data frame " *
```

nalimilan (approved):
Looks good! I kind of wish we would have used `source` instead of `indicator` for joins, as that sounds more explicit, but well...

bkamins:
> I kind of wish we would have used `source` instead of `indicator` for joins

Same with me. Maybe we can:
* deprecate `indicator` in favor of `source` in `joins`, but keep it till 2.0 release? It should be non-problematic from performance perspective and would be clearer? (@oxinabox - could you comment as if I recall correctly I used `indicator` in `join` because you have mentioned this name at some point - so maybe you have some opinion here)
* in this PR use only `source`?

oxinabox:
I have no recollection.
Maybe @rofinn ?

bkamins:
Sorry for pinging then. Do you have an opinion though :smile:?

rofinn:
I also don't recall this discussion. I'm also in favour of `source` over `indicator`.

pdeffebach:
My only comment is that `source` should go at the end of the data frame, not at the front.

bkamins:
> My only comment is that `source` should go at the end of the data frame, not at the front.

OK. I will change it.

bkamins:
OK - I have traced the original issue here https://github.com/JuliaData/DataFrames.jl/issues/1412. Thank you all for responding.

src/join/composer.jl
```diff
@@ -636,22 +636,35 @@ julia> leftjoin(name, job2, on = [:ID => :identifier], renamecols = uppercase =>
...
+        if indicator !== nothing
+            source = indicator
+            Base.depwarn("`indicator` keyword argument is deprecated and " *
+                         "will be removed in 2.0 release of DataFrames.jl. " *
+                         "Use `source` keyword argument instead.", :leftjoin)
+        end
+    elseif indicator !== nothing
+        throw(ArgumentError("`indicator` keyword argument is deprecated. " *
+                            "it is not allow to pass while also passing `source` " *
+                            "keyword argument."))
```
nalimilan, line 658:
```suggestion
                            "It is not allowed to pass both `indicator` and `source` " *
                            "keyword arguments at the same time."))
```

src/join/composer.jl
```diff
@@ -677,7 +690,7 @@ change in future releases.
   if duplicate names are found in columns not joined on;
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
-- `indicator` : Default: `nothing`. If a `Symbol` or string, adds categorical indicator
+- `soruce` : Default: `nothing`. If a `Symbol` or string, adds indicator
```
nalimilan, line 693:
```suggestion
- `source` : Default: `nothing`. If a `Symbol` or string, adds indicator
```

src/join/composer.jl
```diff
@@ -770,21 +783,34 @@ julia> rightjoin(name, job2, on = [:ID => :identifier], renamecols = uppercase =
...
+        if indicator !== nothing
+            source = indicator
+            Base.depwarn("`indicator` keyword argument is deprecated and " *
+                         "will be removed in 2.0 release of DataFrames.jl. " *
+                         "Use `source` keyword argument instead.", :rightjoin)
+        end
+    elseif indicator !== nothing
+        throw(ArgumentError("`indicator` keyword argument is deprecated. " *
+                            "it is not allow to pass while also passing `source` " *
+                            "keyword argument."))
```
nalimilan, line 804:
```suggestion
                            "It is not allowed to pass both `indicator` and `source` " *
                            "keyword arguments at the same time."))
```

src/join/composer.jl
```diff
@@ -814,7 +840,7 @@ This behavior may change in future releases.
   if duplicate names are found in columns not joined on;
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
-- `indicator` : Default: `nothing`. If a `Symbol` or string, adds categorical indicator
+- `soruce` : Default: `nothing`. If a `Symbol` or string, adds indicator
```
nalimilan, line 843:
```suggestion
- `source` : Default: `nothing`. If a `Symbol` or string, adds indicator
```

src/join/composer.jl
```diff
@@ -913,16 +939,29 @@ julia> rightjoin(name, job2, on = [:ID => :identifier], renamecols = uppercase =
...
+        if indicator !== nothing
+            source = indicator
+            Base.depwarn("`indicator` keyword argument is deprecated and " *
+                         "will be removed in 2.0 release of DataFrames.jl. " *
+                         "Use `source` keyword argument instead.", :outerjoin)
+        end
+    elseif indicator !== nothing
+        throw(ArgumentError("`indicator` keyword argument is deprecated. " *
+                            "it is not allow to pass while also passing `source` " *
+                            "keyword argument."))
```
nalimilan, line 961:
```suggestion
                            "It is not allowed to pass both `indicator` and `source` " *
                            "keyword arguments at the same time."))
```

src/join/composer.jl
```diff
@@ -544,7 +544,7 @@ change in future releases.
   if duplicate names are found in columns not joined on;
   if `true`, duplicate names will be suffixed with `_i`
   (`i` starting at 1 for the first duplicate).
-- `indicator` : Default: `nothing`. If a `Symbol` or string, adds categorical indicator
+- `soruce` : Default: `nothing`. If a `Symbol` or string, adds indicator
```
nalimilan, line 547:
```suggestion
- `source` : Default: `nothing`. If a `Symbol` or string, adds indicator
```

nalimilan (approved):

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,172 @@ julia> vcat(df1, df3, cols=:intersect)
...
 Base.vcat(dfs::AbstractDataFrame...;
           cols::Union{Symbol, AbstractVector{Symbol},
-                      AbstractVector{<:AbstractString}}=:setequal) =
-    reduce(vcat, dfs; cols=cols)
+                      AbstractVector{<:AbstractString}}=:setequal,
+          source::Union{Nothing, SymbolOrString,
+                           Pair{<:SymbolOrString, <:AbstractVector}}=nothing) =
+    reduce(vcat, dfs; cols=cols, source=source)
+
+"""
```
bkamins, line 1596: @nalimilan - I have added a docstring for `reduce` to make this option more discoverable. Could you please have a look at it before I merge? Thank you!

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,172 @@ julia> vcat(df1, df3, cols=:intersect)
...
+    reduce(::typeof(vcat),
+           dfs::Union{AbstractVector{<:AbstractDataFrame},
+                      Tuple{AbstractDataFrame, Vararg{AbstractDataFrame}}};
+           cols::Union{Symbol, AbstractVector{Symbol},
+                       AbstractVector{<:AbstractString}}=:setequal,
+           source::Union{Nothing, Symbol, AbstractString,
+                         Pair{<:Union{Symbol, AbstractString}, <:AbstractVector}}=nothing)
+
```
nalimilan, line 1604:
```suggestion
```

nalimilan (approved):

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1535,29 +1546,171 @@ julia> vcat(df1, df3, cols=:intersect)
...
+        end
+
+        @assert start == nrow(res) + 1
+        insertcols!(res, col => source_vec)
+    end
+
+    return res
 end
 
+
```
bkamins, line 1713:
```suggestion
```

bkamins:
Thank you!

Issue #2651, opened 2021-03-11
add predicate/transformation version of unique?


pcjentsch:
This issue is a place for continuing discussion of additional `unique` methods that accept predicates for comparing rows.

A simple way would be to follow `Base.unique(f, iter)`, which simply maps columns, and then checks for equality with `isequal`. 

Another would to specify an equivalence relation that accepts two rows as an argument and returns whether they are equivalent.

bkamins:
I think we already support it:
```
julia> df = DataFrame(a=1:10, b=10:-1:1)
10×2 DataFrame
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1     10
   2 │     2      9
   3 │     3      8
   4 │     4      7
   5 │     5      6
   6 │     6      5
   7 │     7      4
   8 │     8      3
   9 │     9      2
  10 │    10      1

julia> unique(df, [:a, :b] => (a,b) -> a .+ b)
1×2 DataFrame
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1     10
```

Is this enough for you? If yes please close the issue.

pcjentsch:
ah, nice!

bkamins:
We need to update the docsting though so I will keep it open. Thank you for reporting

MatthewRGonzalez:
The [documentation](https://dataframes.juliadata.org/stable/lib/functions/#Base.unique) for `unique` indicates, "cols can be any column selector or transformation accepted by select." 

In the following example,

```
julia> df = DataFrames.DataFrame(x=[1, 1, 3, 4], y=[2, 2, 6, 8])
4×2 DataFrame
 Row │ x      y     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2
   2 │     1      2
   3 │     3      6
   4 │     4      8

julia> unique(select(df, :x => (x -> (0 .* x) .+ 3) => :x, :y => (y -> y .+ 1) => :y))
3×2 DataFrame
 Row │ x      y     
     │ Int64  Int64 
─────┼──────────────
   1 │     3      3
   2 │     3      7
   3 │     3      9

julia> unique(df, :x => (x -> (0 .* x) .+ 3) => :x, :y => (y -> y .+ 1) => :y)
[...]

Issue #2385, opened 2020-08-26
describe doesn't follow the source => function => name convention


nalimilan:
I've just realized that `describe` takes `name => function` pairs, which is totally inconsistent with `select`, `transform` and `combine`. I think we should deprecate the current syntax and move to `function => name`.

bkamins:
Well - we were aware that we left this case.

`describe` is consistent with `DataFrame` and `insertcols!` syntax. Also see https://bkamins.github.io/julialang/2020/07/17/pair.html for a review of all uses of `=>` we have right now.

Having said that I would not oppose to change it prior to 1.0 release as this is a non-core functionality.

nalimilan:
OK, I must have forgotten that then. I'd be inclined to change this, as it's quite similar to `select`/`transform`/`combine` (function associated with output column name). `insertcols!` is a bit different as there's no function. Also, the fewer exceptions, the better.

bkamins:
As I have said - I would not oppose it.

@oxinabox, @quinnj, @ararslan - do you have any opinion here? (I am mostly asking about the approach to "breaking things" in cases like here - we have a case of a relatively unimportant feature that is a bit inconsistent with the rest of the design; so - it would be better to change it but it is breaking).

In general this is a last call to break it in 0.22 release, as in general I will oppose any breaking changes to ensure the stability of the API of DataFrames.jl as I believe this is what we need to start ensuring if we want to have a wider adoption.

nalimilan:
If we want to avoid breakage, we could also allow both orders (since there's no ambiguity), but recommend `function => name`.

bkamins:
Actually I would prefer to be breaking, but want to confirm the policy (i.e. allow small breakages for 0.22 like this one to clean things up; do not allow breaking changes after 1.0 release at all 😄)

tk3369:
My preference is to change the syntax to make it consistent with `select`/`transform`/`combine`.

matthieugomez:
If you’re going to be breaking, would not it be nicer to have cols as the default arg, and stats as the kwarg? That would be nice to be able to do describe(df, :x) similarly to groupby(df, :x) or select(df, :x)

bkamins:
`cols` kwarg is a standard approach to restrict columns to which a function is applied in DataFrames.jl.

nalimilan:
Is that really the case? We have lots of cases where it's a positional argument (`allowmissing`, `completecases`, `unique`...).

[...]

PR #2652, opened 2021-03-11
update describe, unique and nonunique docs

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2651

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1282,9 +1282,9 @@ end
 
 Return a data frame containing only the first occurrence of unique rows in `df`.
 When `cols` is specified, the returned `DataFrame` contains complete rows,
-retaining in each case the first instance of unique selected columns or their
-transformations. `cols` can be any column selector or transformation accepted by
-[`select`](@ref).
+retaining in each case the first occurrence of a given combination of values
+in selected columns or their transformations. `cols` can be any column
+selector or transformation accepted by [`select`](@ref).
 
 
 For `unique`, if `view=false` a freshly allocated `DataFrame` is returned,
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -503,8 +503,9 @@ where each row represents a variable and each column a summary statistic.
     - `:all` as the only `Symbol` argument to return all statistics.
     - A `function => name` pair where `name` is a `Symbol` or string. This will
       create a column of summary statistics with the provided name.
-- `cols` : a keyword argument allowing to select only a subset of columns from `df`
-  to describe. Can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+- `cols` : a keyword argument allowing to select only a subset or transformation
+  of columns from `df` to describe using `select(df, cols, copycols=false)`.
+  Can be any column selector accepted by `select`.
 
 # Details
 For `Real` columns, compute the mean, standard deviation, minimum, first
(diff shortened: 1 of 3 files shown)


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -503,8 +503,9 @@ where each row represents a variable and each column a summary statistic.
     - `:all` as the only `Symbol` argument to return all statistics.
     - A `function => name` pair where `name` is a `Symbol` or string. This will
       create a column of summary statistics with the provided name.
-- `cols` : a keyword argument allowing to select only a subset of columns from `df`
-  to describe. Can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+- `cols` : a keyword argument allowing to select only a subset or transformation
+  of columns from `df` to describe using `select(df, cols, copycols=false)`.
+  Can be any column selector accepted by `select`.
```
nalimilan, line 508:
```suggestion
  Can be any column selector or transformation accepted by [`select`](@ref).
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -503,8 +503,9 @@ where each row represents a variable and each column a summary statistic.
     - `:all` as the only `Symbol` argument to return all statistics.
     - A `function => name` pair where `name` is a `Symbol` or string. This will
       create a column of summary statistics with the provided name.
-- `cols` : a keyword argument allowing to select only a subset of columns from `df`
-  to describe. Can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+- `cols` : a keyword argument allowing to select only a subset or transformation
+  of columns from `df` to describe using `select(df, cols, copycols=false)`.
```
nalimilan, line 507: Maybe not worth giving the implementation details? Saying that selectors supported by `describe` are accepted is probably enough.
```suggestion
  of columns from `df` to describe.
```
If you agree this can also apply to other places.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1187,8 +1188,9 @@ See also [`unique`](@ref) and [`unique!`](@ref).
 
 # Arguments
 - `df` : `AbstractDataFrame`
-- `cols` : a selector specifying the column(s) to compare. Can be any column
-  selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+- `cols` : a selector specifying the column(s) to compare using
```
nalimilan, line 1191: Also mention transformations (here and below)?

pdeffebach:
Just to confirm, changing `describe` at this point, to do `descibe(df, :x, :y, funs = ...)` is a no-go?

bkamins:
> changing describe at this point, to do `descibe(df, :x, :y, funs = ...)` is a no-go?

I would say so. Also note that you can always write `describe(select(df, ...), ...)`.

pdeffebach:
Perhaps a `summarize` method can be added in DataFramesMeta to alleviate this.

nalimilan:
> Perhaps a `summarize` method can be added in DataFramesMeta to alleviate this.

Please don't! We managed to keep the API relatively small so that it's easy to master, and we've almost reached the goal to have a parallel between DataFrames functions and DataFramesMeta macros. Also `summarize` is used in dplyr for something totally different.

For reference, the previous discussion was at https://github.com/JuliaData/DataFrames.jl/issues/2385.

pdeffebach:
Sounds good! Thankfully we can export the `@col` macro in DataFramesMeta to make those transformations really seamless even if they are in a keyword argument.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1280,10 +1281,11 @@ end
...
 
 Return a data frame containing only the first occurrence of unique rows in `df`.
-When `cols` is specified, the returned `DataFrame` contains
-complete rows, retaining in each case the first instance for which `df[cols]` is
-unique. `cols` can be any column selector ($COLUMNINDEX_STR;
-$MULTICOLUMNINDEX_STR).
+When `cols` is specified, the returned `DataFrame` contains complete rows,
+retaining in each case the first instance of unique selected columns or their
+transformations. `cols` can be any column selector or transformation accepted by
+[`select`](@ref).
```
nalimilan, line 1287:
```suggestion
retaining in each case the first occurrence of a given combination of values
in selected columns or their transformations. `cols` can be any column
selector or transformation accepted by [`select`](@ref).
```

nalimilan (approved):

bkamins:
Thank you!

PR #2655, opened 2021-03-12
add support for getproperty broadcasting

bkamins:
Now we will be able to do:
```
df.a .= 1
```
even if `:a` is not present in `df`.

The only decision to be made is if we want to allow:

`dfv.a .= 1` if `dfv` is a `SubDataFrame`. Currently this is allowed, but it is inconsistent with `dfv[!, :a] .= 1` so I would disallow it. The problem is that then DataFrames.jl would work differently under Julia 1.7 and prior to 1.7 in this area - thus it is a hard decision to make.


quinnj (approved):
LGTM; I agree with disallowing for now; better to error for now and we can allow things later.

bkamins:
The problem is that it works under Julia 1.6 although we wanted to disallow it, but we could not because of the limitations of Julia Base.

bkamins:
todo: tests and docs need an update after @nalimilan comments

bkamins:
@nalimilan - I have summarized the consequences of this PR in the docstrings. It is super unfortunate we cannot make it consistent. What do you think we should do?

bkamins:
Unfortunately following the comment by @KristofferC in https://github.com/JuliaLang/julia/pull/39473#issuecomment-797339888 the change will not go into 1.6 LTS.

bkamins:
I really have a mental problem here. If we introduce this PR now then DataFrames.jl will have a different behavior under different versions of Julia and we will not be able to even give a warning.

Maybe a lesser evil is only to add support for creating columns via `df.a .= 1` and leave the rest of the functionality "as is" and only print deprecation warning in Julia 1.7 that things will change in the future (before Julia 1.7 it is even impossible to print a deprecation warning in this case).

pdeffebach:
This is exciting! I didn't realize this would get added so soon. 

I am fine with disallowing for a while. It's a "nice to have" and I'm not sure how many users will be burned by this, at least until we get a new LTS and can be more aggressive with version requirements.

bkamins:
The problem is that new LTS will be 1.6 most likely. And in 1.6 this feature will not be available, but only in 1.7.

Just to clarify the issue: In Julia 1.6 and earlier we have no control over `df.a .= 1` as it gets transformed before it gets passed to DataFrames.jl. Therefore we have two roads:

1. have Julia 1.7 and Julia 1.6 have different behaviors, but the behavior in Julia 1.7 would be a correct one (this might be acceptable as as @pdeffebach noted - the cases when behaviors differ are super rare - essentially broadcasting to `SubDataFrame` and when you would want to change `eltype` of a column in `DataFrame`); please vote :rocket: for this option under this post
2. have Julia 1.7 copy the incorrect behavior of Julia 1.6 with a deprecation warning + only allow `df.a .= 1` for new columns in Julia 1.7 (which is an error in Julia 1.6 so it is not a huge problem) and make the change in some future release of Julia that becomes LTS which would support `getproperty` broadcasting (probably 2.0); please vote :eyes: for this option under this post

nalimilan:
I'd say option 2 is safer, especially since we're likely to release DataFrames 1.0 before Julia 1.7: people won't have had the occasion to test their code against Julia 1.7. But I think we should feel free to remove the deprecation warning and switch to the new, correct behavior relatively quickly, i.e. even before we drop support for Julia 1.6 (anyway the behavior won't change there). Otherwise I'd rather take option 1.

bkamins:
OK. I will go for option 2 then and add comments in the code what has to be changed.

To be clear, what is incorrect in Julia 1.6 are the following things (in general the incorrect things are where `getproperty` now works differently than `!` row selector):
* `sub_data_frame.col .= 1` is allowed and is in-place (technically this should not be allowed and throw an error and we should require `sub_data_frame[:, :col] .= 1` for this case); the reason is that `getproperty` should match `!` column selector which replaces columns (and thus is disallowed for `SubDataFrame`)
* if `col` is a vector of e.g. `Int` then `data_frame.col .= 'a'` works but converts `a` to integer and stores it in-place in `col`; what it should do is to replace the column and create column of `Char` filled with `a`.

As noted above it is impossible to fix this behavior in Julia 1.6 because the resolution of the operation happens at parse time before DataFrames.jl mechanics is able to intervene and change this behavior.

src/other/broadcasting.jl
```diff
@@ -121,6 +121,28 @@ end
 Base.dotview(df::SubDataFrame, ::typeof(!), idxs) =
     throw(ArgumentError("broadcasting with ! row selector is not allowed for SubDataFrame"))
 
+
+# TODO: remove the deprecations when Julia 1.7 functionality is commonly used
```
bkamins, line 125: @nalimilan - can you please check this. I will update tests and documentation once we decide what we like.

bkamins: bump (no rush of curse).

The key question is if when we write:

```
df.col .= value
```
and `col` is present in the data frame do we want this operation to always work (even for `SubDataFrame`) and be in-place - this is a current behavior

or

we want this to replace a column and be allowed only for `DataFrame`.

The reason I am asking is that when someone writes:
```
df.col .= value
```
maybe it is natural for broadcasting to work on the column `col` extracted from `df` (this is our current behavior)

and the only thing we should change is make `df.col .= value` work if `col` is not present (and allocate it then).

The tension is that if we follow the reasoning above `df.col .= value` is not consistent with `df[!, :col] .= value`. But maybe it is not a problem?

bkamins: Just to highlight the crucial tension in the decision to be made here. Currently we have:
```
julia> using DataFrames

julia> df = DataFrame(a=1:4)
4×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │     1
   2 │     2
   3 │     3
   4 │     4

julia> df.a .= 'a':'d'
4-element Vector{Int64}:
  97
  98
  99
 100

julia> df
4×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │    97
   2 │    98
   3 │    99
   4 │   100
```
and this is what happens if `:a` is present in `df`. If `:a` were not present in `df` then after this PR it would hold `['a','b','c','d']`. The question is if it is better to:
* keep an in-place operation we have now (with the inconsistent behaviors between existing and new column operations, but it is a corner case, and probably normally users might want an in-place behavior)
* deprecate in-place operation and in the future replace column (but this will then also mean that `df.col .= value` will not be allowed for `SubDataFrame`)

Both options have pros and cons. @pdeffebach - maybe you can comment from user's perspective which one would seem more convenient in daily work?.

bkamins: OK - then I will finalize this PR with deprecations injected into code. Thank you!

NEWS.md
```diff
@@ -26,12 +26,23 @@
   additional column to be added in the last position in the resulting data frame
   that will identify the source data frame.
   ([#2649](https://github.com/JuliaData/DataFrames.jl/pull/2649))
+* since Julia 1.7 using broadcasting assignment on a `DataFrames` column
+  selected as a property is allowed when it does not exist in a data frame
+  and it allocates a fresh column
```
nalimilan, line 31:
```suggestion
* since Julia 1.7 using broadcasted assignment on a `DataFrame` column
  selected as a property (e.g. `df.col .= 1`) is allowed when column does not
  exist and it allocates a fresh column
```

NEWS.md
```diff
@@ -26,12 +26,23 @@
...
+  ([#2655](https://github.com/JuliaData/DataFrames.jl/pull/2655))
 
 ## Deprecated
 
 * in `leftjoin`, `rightjoin`, and `outerjoin` the `indicator` keyword argument
   is deprecated in favor of `source` keyword argument; `indicator` will be removed
   in 2.0 release ([2649](https://github.com/JuliaData/DataFrames.jl/pull/2649))
+* Using broadcasting assignment on a `SubDataFrames` column selected as a property
+  is deprecated; it will be disallowed in the future.
```
nalimilan, line 40:
```suggestion
* Using broadcasting assignment on a `SubDataFrames` column selected as a property
  (e.g. `sdf.col .= 1`) is deprecated; it will be disallowed in the future.
```

NEWS.md
```diff
@@ -26,12 +26,23 @@
...
+  ([#2655](https://github.com/JuliaData/DataFrames.jl/pull/2655))
+* Broadcasting assignment to an existing column of a `DataFrame`
+  selected as a property being an in-place operation is deprecated. It will
+  allocate a fresh column in the future
```
nalimilan, line 44:
```suggestion
* Broadcasting assignment to an existing column of a `DataFrame`
  selected as a property (e.g. `df.col .= 1`) being an in-place
  operation is deprecated. It will allocate a fresh column in the future
```

docs/src/lib/indexing.md
```diff
@@ -183,19 +183,24 @@ In such an operation `AbstractDataFrame` is considered as two-dimensional and `D
...
+  `df` is performed in-place; if `rows` is `:` and `col` is `Symbol` or `AbstractString`
+  and it is missing from `df` then a new column is allocated and added;
   the length of the column is always the value of `nrow(df)` before the assignment takes place;
 * in the `df[!, col] .= v` syntax column `col` is replaced by a freshly allocated vector;
   if `col` is `Symbol` or `AbstractString` and it is missing from `df` then a new column is allocated added;
   the length of the column is always the value of `nrow(df)` before the assignment takes place;
 * the `df[!, cols] .= v` syntax replaces existing columns `cols` in data frame `df` with freshly allocated vectors;
-* `df.col .= v` syntax is allowed and performs in-place assignment to an existing vector `df.col`.
+* `df.col .= v` currently syntax performs in-place assignment to an existing vector `df.col`;
+  this behavior is deprecated and a freshcolumn will be allocated in the future.
```
nalimilan, line 198:
```suggestion
* `df.col .= v` syntax currently performs in-place assignment to an existing vector `df.col`;
  this behavior is deprecated and a new column will be allocated in the future.
```

test/broadcasting.jl
```diff
@@ -1842,4 +1847,17 @@ end
...
 end
 
+@testset "broadcasting of getproperty" begin
+    if isdefined(Base, :dotgetproperty)
+        df = DataFrame(a=1:4)
+        df.b .= 1
+        df.c .= 4:-1:1
+        # TODO: enable this in the future when the deprecation period is finished
+        # df.a .= 'a':'d'
+        # @test df == DataFrame(a='a':'d', b=1, c=4:-1:1)
```
nalimilan, line 1857:
```suggestion
        # df.a .= 'a':'d'
        # @test df.a isa Vector{Char}
        # @test df == DataFrame(a='a':'d', b=1, c=4:-1:1)
```

docs/src/lib/indexing.md
```diff
@@ -192,10 +192,10 @@ Additional rules:
   if `col` is `Symbol` or `AbstractString` and it is missing from `df` then a new column is allocated added;
   the length of the column is always the value of `nrow(df)` before the assignment takes place;
 * the `df[!, cols] .= v` syntax replaces existing columns `cols` in data frame `df` with freshly allocated vectors;
-* `df.col .= v` syntax is allowed and performs in-place assignment to an existing vector `df.col`.
+* `df.col .= v` syntax is allowed up to Julia 1.6 and performs in-place assignment to an existing vector `df.col`. Since Julia 1.7 column gets replaced.
 * in the `sdf[CartesianIndex(row, col)] .= v`, `sdf[row, col] .= v` and `sdf[row, cols] .= v` syntaxes the assignment to `sdf` is performed in-place;
 * in the `sdf[rows, col] .= v` and `sdf[rows, cols] .= v` syntaxes the assignment to `sdf` is performed in-place;
-* `sdf.col .= v` syntax is allowed and performs in-place assignment to an existing vector `sdf.col`.
+* `sdf.col .= v` syntax is allowed and up to Julia 1.6 performs an in-place assignment to an existing vector `sdf.col`. Since Julia 1.7 this is deprecated.
```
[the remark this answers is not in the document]
bkamins: agreed

bkamins:
@nalimilan - OK to merge?

nalimilan (approved):

bkamins:
Thank you!

PR #2656, opened 2021-03-12
Respect copycols when building DataFrame from Tables.CopiedColumns

quinnj:
Fixes https://github.com/JuliaData/Arrow.jl/issues/146. The issue here
was we were just ignoring `copycols` entirely when building DataFrame
from `Tables.CopiedColumns`. Which is mostly fine because
`CopiedColumns` implies DataFrames can assume ownership of the columns.
The wrinkle comes in when you consider cases like `Arrow.Table`, which
wraps columns in `Tables.CopiedColumns`, not necessarily because copies
have been made, but because the data is immutable, so it should be safe
for other tables to assume ownership. The problem is when users want to
_mutate_ those columns, they need to make copies of them, so naturally
users try `DataFrame(::Arrow.Table; copycols=true)`, but then that
doesn't actually make copies.

The proposal here just intercepts the top-level
`DataFrame(::CopiedColumns)` constructor and passes `copycols=false` as
the default, so if users pass `copycols=true`, it will still be
respected.

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -32,15 +32,16 @@ Tables.getcolumn(dfr::DataFrameRow, nm::Symbol) = dfr[nm]
 
 getvector(x::AbstractVector) = x
 getvector(x) = [x[i] for i = 1:length(x)]
-# note that copycols is ignored in this definition (Tables.CopiedColumns implies copies have already been made)
-fromcolumns(x::Tables.CopiedColumns, names; copycols::Bool=true) =
-    DataFrame(AbstractVector[getvector(Tables.getcolumn(x, nm)) for nm in names],
-              Index(names),
-              copycols=false)
+
 fromcolumns(x, names; copycols::Bool=true) =
     DataFrame(AbstractVector[getvector(Tables.getcolumn(x, nm)) for nm in names],
               Index(names),
               copycols=copycols)
+
+# note that copycols is false by default in this definition (Tables.CopiedColumns implies copies have already been made)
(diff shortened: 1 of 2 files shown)


src/other/tables.jl
```diff
@@ -32,16 +32,17 @@ Tables.getcolumn(dfr::DataFrameRow, nm::Symbol) = dfr[nm]
...
+
 fromcolumns(x, names; copycols::Bool=true) =
     DataFrame(AbstractVector[getvector(Tables.getcolumn(x, nm)) for nm in names],
               Index(names),
               copycols=copycols)
 
+# note that copycols is false by default in this definition (Tables.CopiedColumns implies copies have already been made)
+# but if `copycols=true`, a copy will still be made; this is useful for scenarios where the input is immutable
+# so avoiding copies is desirable, but you may still want a copy for mutation (Arrow.Table is like this)
+DataFrame(x::Tables.CopiedColumns; copycols::Bool=false) = DataFrame(Tables.source(x); copycols=copycols)
```
bkamins, line 44: code width fix
```suggestion
# note that copycols is false by default in this definition (Tables.CopiedColumns
# implies copies have already been made) but if `copycols=true`, a copy will still be
# made; this is useful for scenarios where the input is immutable so avoiding copies
# is desirable, but you may still want a copy for mutation (Arrow.Table is like this)
DataFrame(x::Tables.CopiedColumns; copycols::Bool=false) =
    DataFrame(Tables.source(x); copycols=copycols)
```

bkamins (approved):

PR #2661, opened 2021-03-14
Use multithreading in row_group_slots refarray method

nalimilan:
main:
```julia
julia> using Revise, DataFrames, PooledArrays, BenchmarkTools

julia> x = PooledArray(rand(1:10, 10_000_000));

julia> df = DataFrame(x=x);

  40.742 ms (38 allocations: 76.30 MiB)

julia> df.x[df.x .== 5] .= 6; # Empty group

  52.488 ms (39 allocations: 76.30 MiB)

julia> x = PooledArray(rand([missing; 1:10], 10_000_000));

julia> df = DataFrame(x=x);

  42.955 ms (42 allocations: 76.30 MiB)

julia> @btime groupby(df, :x, skipmissing=true);
  63.951 ms (42 allocations: 76.30 MiB)
```

PR with two threads:
```julia
julia> using Revise, DataFrames, PooledArrays, BenchmarkTools

julia> x = PooledArray(rand(1:10, 10_000_000));

julia> df = DataFrame(x=x);

  25.053 ms (113 allocations: 76.30 MiB)

julia> df.x[df.x .== 5] .= 6; # Empty group

  41.185 ms (114 allocations: 76.30 MiB)

julia> x = PooledArray(rand([missing; 1:10], 10_000_000));

julia> df = DataFrame(x=x);

  26.842 ms (116 allocations: 76.30 MiB)

julia> @btime groupby(df, :x, skipmissing=true);
  36.391 ms (116 allocations: 76.30 MiB)
```

### DIFF
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -88,7 +88,7 @@ funname(c::ComposedFunction) = Symbol(funname(c.outer), :_, funname(c.inner))
     tforeach(f, x::AbstractArray; basesize::Integer)
 
 Apply function `f` to each entry in `x` in parallel, spawning
-one separate task for each block of `basesize` entries.
+one separate task for each block of at least `basesize` entries.
 
 A number of task higher than `Threads.nthreads()` may be spawned,
 since that can allow for a more efficient load balancing in case
@@ -97,8 +97,12 @@ some threads are busy (nested parallelism).
 function tforeach(f, x::AbstractArray; basesize::Integer)
     @static if VERSION >= v"1.4"
         nt = Threads.nthreads()
-        if nt > 1 && length(x) > basesize
-            @sync for p in partition(x, basesize)
+        len = length(x)
+        if nt > 1 && len > basesize
+            # Round size up to ensure all chunks have at least `basesize` entries
+            # This ensures balanced sizes by avoiding a small last chunk
+            basesize′ = cld(len, fld(len, basesize))
+            @sync for p in Iterators.partition(x, basesize′)
                 Threads.@spawn begin
                     for i in p
                         f(@inbounds x[i])
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -83,3 +83,37 @@ else
 end
 
 funname(c::ComposedFunction) = Symbol(funname(c.outer), :_, funname(c.inner))
+
+"""
+    tforeach(f, x::AbstractArray; basesize::Integer)
+
+Apply function `f` to each entry in `x` in parallel, spawning
+one separate task for each block of `basesize` entries.
+
+A number of task higher than `Threads.nthreads()` may be spawned,
+since that can allow for a more efficient load balancing in case
+some threads are busy (nested parallelism).
+"""
+function tforeach(f, x::AbstractArray; basesize::Integer)
+    @static if VERSION >= v"1.4"
+        nt = Threads.nthreads()
+        if nt > 1 && length(x) > basesize
+            @sync for p in partition(x, basesize)
+                Threads.@spawn begin
+                    for i in p
+                        f(@inbounds x[i])
+                    end
+                end
+            end
+        else
+            for i in eachindex(x)
+                f(@inbounds x[i])
+            end
+        end
+    else
+        for i in eachindex(x)
+            f(@inbounds x[i])
+        end
+    end
+    return
+end
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -88,8 +88,8 @@ funname(c::ComposedFunction) = Symbol(funname(c.outer), :_, funname(c.inner))
 # This method ensures balanced sizes by avoiding a small last chunk
 function split_indices(len::Integer, basesize::Integer)
     len′ = Int64(len) # Avoid overflow on 32-bit machines
-    np = max(1, div(len′, basesize))
-    return ((1 + ((i - 1) * len′) ÷ np):((i * len′) ÷ np) for i in 1:np)
+    np = max(1, div(len, basesize))
+    return (Int(1 + ((i - 1) * len′) ÷ np):Int((i * len′) ÷ np) for i in 1:np)
 end
 
 """
@@ -103,7 +103,7 @@ since that can allow for a more efficient load balancing in case
 some threads are busy (nested parallelism).
 """
 function tforeach(f, x::AbstractArray; basesize::Integer)
-    Base.require_one_based_indexing(x)
+    @assert firstindex(x) == 1
 
     @static if VERSION >= v"1.4"
         nt = Threads.nthreads()
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -84,6 +84,14 @@ end
 
 funname(c::ComposedFunction) = Symbol(funname(c.outer), :_, funname(c.inner))
 
+# Compute chunks of indices, each with at least `basesize` entries
+# This method ensures balanced sizes by avoiding a small last chunk
+function split_indices(len::Integer, basesize::Integer)
+    len′ = Int64(len) # Avoid overflow on 32-bit machines
+    np = max(1, div(len′, basesize))
+    return ((1 + ((i - 1) * len′) ÷ np):((i * len′) ÷ np) for i in 1:np)
+end
+
 """
     tforeach(f, x::AbstractArray; basesize::Integer)
 
@@ -95,14 +103,13 @@ since that can allow for a more efficient load balancing in case
 some threads are busy (nested parallelism).
 """
 function tforeach(f, x::AbstractArray; basesize::Integer)
+    Base.require_one_based_indexing(x)
+
     @static if VERSION >= v"1.4"
         nt = Threads.nthreads()
         len = length(x)
         if nt > 1 && len > basesize
-            # Round size up to ensure all chunks have at least `basesize` entries
-            # This ensures balanced sizes by avoiding a small last chunk
-            basesize′ = cld(len, fld(len, basesize))
-            @sync for p in Iterators.partition(x, basesize′)
+            @sync for p in split_indices(len, basesize)
                 Threads.@spawn begin
                     for i in p
                         f(@inbounds x[i])
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -101,4 +101,17 @@ end
     @test fetch(t) === true
 end
 
+@testset "split_indices" begin
+    for len in 0:12
+        basesize = 10
+        x = DataFrames.split_indices(len, basesize)
+
+        @test length(x) == max(1, div(len, basesize))
+        @test reduce(vcat, x) == 1:len
+        vmin, vmax = extrema(length, x)
+        @test vmin + 1 == vmax || vmin == vmax
+        @test len < basesize || vmin >= basesize
+    end
+end
+
 end # module
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -108,7 +108,7 @@ end
 
         @test length(x) == max(1, div(len, basesize))
         @test reduce(vcat, x) === 1:len
-        vmin, vmax = extrema(length, x)
+        vmin, vmax = extrema(length(v) for v in x)
         @test vmin + 1 == vmax || vmin == vmax
         @test len < basesize || vmin >= basesize
     end
@@ -120,7 +120,7 @@ end
     @test length(x) == div(len, basesize)
     @test x[1][1] === 1
     @test x[end][end] === Int(len)
-    vmin, vmax = extrema(length, x)
+    vmin, vmax = extrema(length(v) for v in x)
     @test vmin + 1 == vmax || vmin == vmax
     @test len < basesize || vmin >= basesize
 end
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -107,11 +107,22 @@ end
         x = DataFrames.split_indices(len, basesize)
 
         @test length(x) == max(1, div(len, basesize))
-        @test reduce(vcat, x) == 1:len
+        @test reduce(vcat, x) === 1:len
         vmin, vmax = extrema(length, x)
         @test vmin + 1 == vmax || vmin == vmax
         @test len < basesize || vmin >= basesize
     end
+
+    # Check overflow on 32-bit
+    len = typemax(Int32)
+    basesize = 100_000_000
+    x = collect(DataFrames.split_indices(len, basesize))
+    @test length(x) == div(len, basesize)
+    @test x[1][1] === 1
+    @test x[end][end] === Int(len)
+    vmin, vmax = extrema(length, x)
+    @test vmin + 1 == vmax || vmin == vmax
+    @test len < basesize || vmin >= basesize
 end
 
 end # module
--- a/benchmarks/grouping_performance.jl
+++ b/benchmarks/grouping_performance.jl
@@ -1,0 +1,23 @@
+using DataFrames
+using CategoricalArrays
+using PooledArrays
+using BenchmarkTools
+
+
+# `refpool`/`refarray` optimized grouping method
+for k in (10, 10_000), n in (100, 100_000, 10_000_000)
+    for x in (PooledArray(rand(1:k, n)),
+              CategoricalArray(rand(1:n, 10_000_000)),
+              PooledArray(rand([missing; 1:n], 10_000_000)),
+              CategoricalArray(rand([missing; 1:n], 10_000_000)))
+        df = DataFrame(x=x)
+        @btime groupby($df, :x)
+
+        # Skipping missing values
+        @btime groupby($df, :x, skipmissing=true)
+
+        # Empty group which requires adjusting group indices
+        replace!(df.x, 5 => 6)
+        @btime groupby($df, :x)
+    end
+end
--- a/benchmarks/grouping_performance.jl
+++ b/benchmarks/grouping_performance.jl
@@ -2,22 +2,51 @@ using DataFrames
 using CategoricalArrays
 using PooledArrays
 using BenchmarkTools
+using Random
 
+Random.seed!(1)
+
+grouping_benchmarks = BenchmarkGroup()
 
 # `refpool`/`refarray` optimized grouping method
+refpool_benchmarks = grouping_benchmarks["refpool"] = BenchmarkGroup()
+
 for k in (10, 10_000), n in (100, 100_000, 10_000_000)
     for x in (PooledArray(rand(1:k, n)),
               CategoricalArray(rand(1:n, 10_000_000)),
               PooledArray(rand([missing; 1:n], 10_000_000)),
               CategoricalArray(rand([missing; 1:n], 10_000_000)))
         df = DataFrame(x=x)
-        @btime groupby($df, :x)
+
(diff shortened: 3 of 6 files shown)


src/other/utils.jl
```diff
@@ -83,3 +83,37 @@ else
...
+    tforeach(f, x::AbstractArray; basesize::Integer)
+
+Apply function `f` to each entry in `x` in parallel, spawning
+one separate task for each block of `basesize` entries.
+
+A number of task higher than `Threads.nthreads()` may be spawned,
+since that can allow for a more efficient load balancing in case
+some threads are busy (nested parallelism).
+"""
+function tforeach(f, x::AbstractArray; basesize::Integer)
```
nalimilan, line 97: This docstring makes Documentation build fail as it's not included in the manual. I guess I can just turn it into a comment if we don't have a better solution.

bkamins: We have a solution for this.
Add this function to docs/src/lib/internals.md

bkamins (approved):
Thank you. Looks good.
I assume you have mostly copy-pasted the implementation. Right?

I have left some small comments.

src/other/utils.jl
```diff
@@ -83,3 +83,37 @@ else
...
+    @static if VERSION >= v"1.4"
+        nt = Threads.nthreads()
+        if nt > 1 && length(x) > basesize
+            @sync for p in partition(x, basesize)
```
bkamins, line 101: is `partition(x, basesize)` really optimal?

I ask because I have noticed:
```
julia> collect(DataFrames.partition(1:21, 10))
3-element Vector{UnitRange{Int64}}:
 1:10
 11:20
 21:21
```
which means that we would have one small dangling iteration while it would be probably better to consume it (but maybe it is OK?).

nalimilan: Very good point. I've pushed a commit which increases the size to ensure all chunks have at least `basesize` entries. In your example, this would give only two chunks, which is more in line with the docstring. Please check that it's correct as I always get these things wrong. :-)

nalimilan:
>  I assume you have mostly copy-pasted the implementation. Right?

`tforeach` is a very simplified version inspired by `ThreadsX.foreach`, which was itself inspired by ThreadTools. But it's so simplified that it doesn't have a lot in common:
https://github.com/tkf/ThreadsX.jl/blob/6a69562f0fb5663e5661881698d9f11f7d45dc06/src/foreach.jl#L8

src/other/utils.jl
nalimilan: BTW, the current approach is quite good when the number of tasks is large compared with the number of threads, as work will be distributed quite evenly over threads. But it's not completely optimal if there are few tasks. For example, with 3M rows and 2 threads, 3 tasks will be started with each 1M rows, so one thread will finish much before the other one. It would be more efficient to start two tasks with 1.5M rows each.

But there are more complex cases: imagine we have 4M rows and 3 threads, but one thread is actually busy doing something else. If we start 3 tasks instead of 4, we end up in the situation described in the previous paragraph. Not sure how to fix that in general. Maybe we shouldn't care as the 1M threshold is chosen precisely because it takes very little time to process a chunk of that size. If it's a problem that one thread gets one more chunk than other threads, maybe we should simply decrease the threshold...

bkamins: This fix is incorrect:
```
julia> len = 122
122

julia> basesize = 10
10

julia> basesize′ = cld(len, fld(len, basesize))
11

julia> collect(Iterators.partition(1:len, basesize′))
12-element Vector{UnitRange{Int64}}:
 1:11
 12:22
 23:33
 34:44
 45:55
 56:66
 67:77
 78:88
 89:99
 100:110
 111:121
 122:122
```
The way to fix it is something like:
```
function getranges(len, batchsize)
    partitions = div(len, batchsize)
    d , r = divrem(len, partitions)
    [((i - 1) * d + min(r, i - 1) + 1) .+ (0:d - 1 + (i ≤ r ? 1 : 0)) for i in 1:partitions]
end
```
but it is complex so I am not sure if it is worth the effort?

bkamins: > If it's a problem that one thread gets one more chunk than other threads, maybe we should simply decrease the threshold...

That is the point I think (and also why I am considering my fix overly complex) - I assume that 1M rows is small enough to be negligible.

nalimilan: I think it's worth it. But can't we find a simplest solution (i.e. one I can understand :-p)? At #2491 I had copied a solution I found on Discourse. Do you think it's correct?
```julia
julia> len = 122
122

julia> basesize = 10
10

julia> npartitions = div(len, basesize)
12

julia> collect((1 + ((i - 1) * len) ÷ npartitions):((i * len) ÷ npartitions) for i in 1:npartitions)
12-element Vector{UnitRange{Int64}}:
 1:10
 11:20
 21:30
 31:40
 41:50
 51:61
 62:71
 72:81
 82:91
 92:101
 102:111
 112:122
```

src/other/utils.jl
```diff
@@ -83,3 +83,41 @@ else
...
+since that can allow for a more efficient load balancing in case
+some threads are busy (nested parallelism).
+"""
+function tforeach(f, x::AbstractArray; basesize::Integer)
+    @static if VERSION >= v"1.4"
+        nt = Threads.nthreads()
+        len = length(x)
+        if nt > 1 && len > basesize
+            # Round size up to ensure all chunks have at least `basesize` entries
+            # This ensures balanced sizes by avoiding a small last chunk
```
bkamins, line 103: As noted in the main thread this comment is incorrect unfortunately.

src/other/utils.jl
bkamins: The https://github.com/JuliaData/DataFrames.jl/pull/2491/files#diff-b452a777f1ba151f78b9d66f93f4bdb7a8148c5cbc5329fcbcea56954fe7c7a5R203 implementation is OK, but it only spawns as many tasks as you have threads so using `÷` (which is expensive) is not a problem.

My implementation has the same approach as the one you reference to but avoids `÷`. They essentially do the following:
1. calculate the required number of partitions (rounding down so that we are sure each group will get enough elements)
2. then each group gets either `k` or `k+1` elements (as number of groups might not be a divisor of length)
3. the complexity in the codes is to ensure that this `+1` is added as many times as the reminder from the division

However, both your code and my code are incorrect as they fail if `len < basesize`. Here is your corrected code along with correctness tests:

```
function fun(len::Int64, basesize) # Int64 is relevant here to avoid overflow on 32-bit machines
    npartitions = max(1, div(len, basesize))
    x = collect((1 + ((i - 1) * len) ÷ npartitions):((i * len) ÷ npartitions) for i in 1:npartitions)

    # correctness tests
    @assert length(x) == npartitions
    @assert reduce(vcat, x) == 1:len
    vmin, vmax = extrema(length, x)
    @assert vmin + 1 == vmax || vmin == vmax
    @assert len < basesize || vmin >= basesize

    return x
end

for len in 1:1000, basesize in 1:100
    fun(len, basesize)
end
```

src/other/utils.jl
```diff
@@ -83,3 +83,41 @@ else
...
+            basesize′ = cld(len, fld(len, basesize))
+            @sync for p in Iterators.partition(x, basesize′)
```
quinnj, line 105: It'd be nice to have a good sense for "common column types" when doing operations like this. For example, any `<:AbstractArray` ends up using `view` for each partition, but for generic iterators, it actually accumulates the partitions manually, which is less efficient.

All that to say, I think we're fine here, it's just something that I pause on because I don't always have a good mental model of what the most common kinds of column types are and whether there might be hidden bottlenecks in things like ths.

nalimilan: But columns are always `AbstractArray`?

src/other/utils.jl
nalimilan: Thanks. I've pushed a commit using your implementation.

FWIW I've realized that `Threads.@threads` also has similar code: https://github.com/JuliaLang/julia/blob/c19d389a00b4a15058c16ac46a991aa312c6fc40/base/threadingconstructs.jl#L47-L83

src/other/utils.jl
quinnj: Yes, that's why I think this is really a non-issue; but just gave me a pause to consider if we're really efficient here.

nalimilan: OK. BTW, we don't use views are all, we just iterate over relevant indices.

test/utils.jl
```diff
@@ -101,4 +101,28 @@ end
...
 end
 
+@testset "split_indices" begin
+    for len in 0:12
+        basesize = 10
+        x = DataFrames.split_indices(len, basesize)
+
+        @test length(x) == max(1, div(len, basesize))
+        @test reduce(vcat, x) === 1:len
+        vmin, vmax = extrema(length, x)
```
nalimilan, line 111: To fix 1.0:
```suggestion
        vmin, vmax = extrema(length(v) for v in x)
```

test/utils.jl
```diff
@@ -101,4 +101,28 @@ end
...
+    end
+
+    # Check overflow on 32-bit
+    len = typemax(Int32)
+    basesize = 100_000_000
+    x = collect(DataFrames.split_indices(len, basesize))
+    @test length(x) == div(len, basesize)
+    @test x[1][1] === 1
+    @test x[end][end] === Int(len)
+    vmin, vmax = extrema(length, x)
```
nalimilan, line 123:
```suggestion
    vmin, vmax = extrema(length(v) for v in x)
```

test/utils.jl
bkamins: ah - this pesky 1.0 LTS :)

bkamins:
@nalimilan - could you please in this PR start adding something to /benchmarks/ folder? (maybe also for the other threading PRs)? I fear that if we do not start adding the benchmarks incrementally we will never get them. Thank you!

nalimilan:
I've added benchmarks. I ended up using BenchmarkTools more systematically, as doing the comparisons manually was really hard. This should be easier to automate using PkgBenchmarks. Here are the results. I'm taking the median rather than the minimum as for some reason results were quite inconsistent with the latter.

PR 10 threads versus main 1 thread:
```julia
julia> leaves(judge(median(grouping_results_main), median(grouping_results)))
36-element Vector{Any}:
 (Any["refpool", (10000, 100, "PooledArray", "skipmissing=false")], TrialJudgement(-67.02% => improvement))
 (Any["refpool", (10, 100000, "CategoricalArray", "skipmissing=true")], TrialJudgement(-0.85% => invariant))
 (Any["refpool", (10000, 100000, "CategoricalArray", "skipmissing=false")], TrialJudgement(+7.02% => regression))
 (Any["refpool", (10, 100, "PooledArray", "skipmissing=true")], TrialJudgement(-66.93% => improvement))
 (Any["refpool", (10000, 100, "PooledArray", "empty group")], TrialJudgement(-67.29% => improvement))
 (Any["refpool", (10000, 10000000, "CategoricalArray", "skipmissing=false")], TrialJudgement(+3.44% => invariant))
 (Any["refpool", (10, 100000, "PooledArray", "skipmissing=false")], TrialJudgement(+1.72% => invariant))
 (Any["refpool", (10, 10000000, "CategoricalArray", "skipmissing=true")], TrialJudgement(-28.75% => improvement))
 (Any["refpool", (10000, 100, "CategoricalArray", "empty group")], TrialJudgement(-68.22% => improvement))
 (Any["refpool", (10000, 100000, "CategoricalArray", "skipmissing=true")], TrialJudgement(-1.00% => invariant))
 (Any["refpool", (10, 100000, "PooledArray", "empty group")], TrialJudgement(-4.63% => invariant))
 (Any["refpool", (10000, 100, "PooledArray", "skipmissing=true")], TrialJudgement(-65.23% => improvement))
 (Any["refpool", (10, 100, "CategoricalArray", "skipmissing=false")], TrialJudgement(-68.95% => improvement))
 (Any["refpool", (10, 10000000, "PooledArray", "skipmissing=false")], TrialJudgement(-42.16% => improvement))
 (Any["refpool", (10, 100000, "CategoricalArray", "empty group")], TrialJudgement(+6.91% => regression))
 (Any["refpool", (10000, 100000, "PooledArray", "skipmissing=false")], TrialJudgement(+5.11% => regression))
 (Any["refpool", (10000, 10000000, "CategoricalArray", "skipmissing=true")], TrialJudgement(-29.63% => improvement))
 (Any["refpool", (10, 10000000, "PooledArray", "empty group")], TrialJudgement(-39.39% => improvement))
 (Any["refpool", (10, 100000, "PooledArray", "skipmissing=true")], TrialJudgement(-1.25% => invariant))
 (Any["refpool", (10, 10000000, "CategoricalArray", "empty group")], TrialJudgement(-51.84% => improvement))
 (Any["refpool", (10000, 100000, "PooledArray", "empty group")], TrialJudgement(+1.60% => invariant))
 (Any["refpool", (10000, 100, "CategoricalArray", "skipmissing=false")], TrialJudgement(-71.81% => improvement))
 (Any["refpool", (10, 100, "CategoricalArray", "skipmissing=true")], TrialJudgement(-65.20% => improvement))
 (Any["refpool", (10000, 100000, "CategoricalArray", "empty group")], TrialJudgement(+4.60% => invariant))
 (Any["refpool", (10000, 10000000, "PooledArray", "skipmissing=false")], TrialJudgement(-37.56% => improvement))
 (Any["refpool", (10, 10000000, "PooledArray", "skipmissing=true")], TrialJudgement(-39.75% => improvement))
 (Any["refpool", (10000, 10000000, "PooledArray", "empty group")], TrialJudgement(-33.09% => improvement))
 (Any["refpool", (10000, 100000, "PooledArray", "skipmissing=true")], TrialJudgement(+0.40% => invariant))
 (Any["refpool", (10, 100, "PooledArray", "skipmissing=false")], TrialJudgement(-66.95% => improvement))
 (Any["refpool", (10, 100000, "CategoricalArray", "skipmissing=false")], TrialJudgement(-10.23% => improvement))
 (Any["refpool", (10000, 10000000, "CategoricalArray", "empty group")], TrialJudgement(+5.86% => regression))
 (Any["refpool", (10000, 100, "CategoricalArray", "skipmissing=true")], TrialJudgement(-69.29% => improvement))
 (Any["refpool", (10, 100, "PooledArray", "empty group")], TrialJudgement(-69.21% => improvement))
 (Any["refpool", (10000, 10000000, "PooledArray", "skipmissing=true")], TrialJudgement(-33.82% => improvement))
 (Any["refpool", (10, 10000000, "CategoricalArray", "skipmissing=false")], TrialJudgement(-52.23% => improvement))
 (Any["refpool", (10, 100, "CategoricalArray", "empty group")], TrialJudgement(-72.82% => improvement))
```

PR 1 thread versus main 1 thread:
```julia
julia> leaves(judge(median(grouping_results_main), median(grouping_results)))
36-element Vector{Any}:
 (Any["refpool", (10000, 100, "PooledArray", "skipmissing=false")], TrialJudgement(+0.09% => invariant))
 (Any["refpool", (10, 100000, "CategoricalArray", "skipmissing=true")], TrialJudgement(+0.96% => invariant))
 (Any["refpool", (10000, 100000, "CategoricalArray", "skipmissing=false")], TrialJudgement(+0.97% => invariant))
 (Any["refpool", (10, 100, "PooledArray", "skipmissing=true")], TrialJudgement(-5.04% => improvement))
 (Any["refpool", (10000, 100, "PooledArray", "empty group")], TrialJudgement(-0.09% => invariant))
 (Any["refpool", (10000, 10000000, "CategoricalArray", "skipmissing=false")], TrialJudgement(-2.39% => invariant))
 (Any["refpool", (10, 100000, "PooledArray", "skipmissing=false")], TrialJudgement(+1.42% => invariant))
 (Any["refpool", (10, 10000000, "CategoricalArray", "skipmissing=true")], TrialJudgement(-0.47% => invariant))
 (Any["refpool", (10000, 100, "CategoricalArray", "empty group")], TrialJudgement(-1.51% => invariant))
 (Any["refpool", (10000, 100000, "CategoricalArray", "skipmissing=true")], TrialJudgement(+0.87% => invariant))
 (Any["refpool", (10, 100000, "PooledArray", "empty group")], TrialJudgement(+1.44% => invariant))
 (Any["refpool", (10000, 100, "PooledArray", "skipmissing=true")], TrialJudgement(-5.27% => improvement))
 (Any["refpool", (10, 100, "CategoricalArray", "skipmissing=false")], TrialJudgement(-1.57% => invariant))
 (Any["refpool", (10, 10000000, "PooledArray", "skipmissing=false")], TrialJudgement(+0.16% => invariant))
 (Any["refpool", (10, 100000, "CategoricalArray", "empty group")], TrialJudgement(+1.10% => invariant))
 (Any["refpool", (10000, 100000, "PooledArray", "skipmissing=false")], TrialJudgement(+1.75% => invariant))
 (Any["refpool", (10000, 10000000, "CategoricalArray", "skipmissing=true")], TrialJudgement(-0.28% => invariant))
 (Any["refpool", (10, 10000000, "PooledArray", "empty group")], TrialJudgement(+0.11% => invariant))
 (Any["refpool", (10, 100000, "PooledArray", "skipmissing=true")], TrialJudgement(+1.96% => invariant))
 (Any["refpool", (10, 10000000, "CategoricalArray", "empty group")], TrialJudgement(-0.64% => invariant))
 (Any["refpool", (10000, 100000, "PooledArray", "empty group")], TrialJudgement(+1.52% => invariant))
 (Any["refpool", (10000, 100, "CategoricalArray", "skipmissing=false")], TrialJudgement(-1.30% => invariant))
 (Any["refpool", (10, 100, "CategoricalArray", "skipmissing=true")], TrialJudgement(-7.19% => improvement))
 (Any["refpool", (10000, 100000, "CategoricalArray", "empty group")], TrialJudgement(+0.84% => invariant))
 (Any["refpool", (10000, 10000000, "PooledArray", "skipmissing=false")], TrialJudgement(+1.45% => invariant))
 (Any["refpool", (10, 10000000, "PooledArray", "skipmissing=true")], TrialJudgement(-0.32% => invariant))
 (Any["refpool", (10000, 10000000, "PooledArray", "empty group")], TrialJudgement(+1.32% => invariant))
 (Any["refpool", (10000, 100000, "PooledArray", "skipmissing=true")], TrialJudgement(+1.96% => invariant))
 (Any["refpool", (10, 100, "PooledArray", "skipmissing=false")], TrialJudgement(+0.26% => invariant))
 (Any["refpool", (10, 100000, "CategoricalArray", "skipmissing=false")], TrialJudgement(+0.90% => invariant))
 (Any["refpool", (10000, 10000000, "CategoricalArray", "empty group")], TrialJudgement(-0.04% => invariant))
 (Any["refpool", (10000, 100, "CategoricalArray", "skipmissing=true")], TrialJudgement(-7.07% => improvement))
 (Any["refpool", (10, 100, "PooledArray", "empty group")], TrialJudgement(+0.25% => invariant))
 (Any["refpool", (10000, 10000000, "PooledArray", "skipmissing=true")], TrialJudgement(+3.22% => invariant))
 (Any["refpool", (10, 10000000, "CategoricalArray", "skipmissing=false")], TrialJudgement(-1.97% => invariant))
 (Any["refpool", (10, 100, "CategoricalArray", "empty group")], TrialJudgement(-1.45% => invariant))
```

bkamins:
Looks good to merge. Thank you!

PR #2664, opened 2021-03-19
Support threading in joined column creation

bkamins:
A minor PR to speed up column creation in joins using threading.

Its major benefit is for left, right and outerjoins having many extra columns (and producing over 1_000_000 rows)

### DIFF
--- a/src/join/composer.jl
+++ b/src/join/composer.jl
@@ -97,11 +97,16 @@ function compose_inner_table(joiner::DataFrameJoiner,
                              right_rename::Union{Function, AbstractString, Symbol})
     left_ixs, right_ixs = find_inner_rows(joiner)
 
-    if VERSION >= v"1.4" && Threads.nthreads() > 1 && length(left_ixs) >= 1_000_000
-        dfl_task = Threads.@spawn joiner.dfl[left_ixs, :]
-        dfr_noon_task = Threads.@spawn joiner.dfr[right_ixs, Not(joiner.right_on)]
-        dfl = fetch(dfl_task)
-        dfr_noon = fetch(dfr_noon_task)
+    @static if VERSION >= v"1.4"
+        if Threads.nthreads() > 1 && length(left_ixs) >= 1_000_000
+            dfl_task = Threads.@spawn joiner.dfl[left_ixs, :]
+            dfr_noon_task = Threads.@spawn joiner.dfr[right_ixs, Not(joiner.right_on)]
+            dfl = fetch(dfl_task)
+            dfr_noon = fetch(dfr_noon_task)
+        else
+            dfl = joiner.dfl[left_ixs, :]
+            dfr_noon = joiner.dfr[right_ixs, Not(joiner.right_on)]
+        end
     else
         dfl = joiner.dfl[left_ixs, :]
         dfr_noon = joiner.dfr[right_ixs, Not(joiner.right_on)]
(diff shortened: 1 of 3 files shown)


src/join/composer.jl
```diff
@@ -96,8 +96,16 @@ function compose_inner_table(joiner::DataFrameJoiner,
                              left_rename::Union{Function, AbstractString, Symbol},
                              right_rename::Union{Function, AbstractString, Symbol})
     left_ixs, right_ixs = find_inner_rows(joiner)
-    dfl = joiner.dfl[left_ixs, :]
-    dfr_noon = joiner.dfr[right_ixs, Not(joiner.right_on)]
+
+    if VERSION >= v"1.4" && Threads.nthreads() > 1 && length(left_ixs) >= 1_000_000
+        dfl_task = Threads.@spawn joiner.dfl[left_ixs, :]
```
bkamins, line 101: since `getindex` already uses threading this is only to ensure we are fast in case left and right table have very different processing times.

src/join/composer.jl
```diff
@@ -207,21 +215,35 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol, makeunique:
 
     @assert col_idx == ncol(joiner.dfl_on) + 1
 
-    for col in eachcol(dfl_noon)
```
bkamins, line 210: I did not add threading for on-columns creation as in practice I expect that on-columns are minority of created columns and I avoid adding too much boilerplate code this way.

bkamins:
Example run.

Setup:
```
julia> Random.seed!(1234);

julia> df1 = sort!(DataFrame(rand(10^7, 10), :auto));

julia> df2 = sort!(DataFrame(rand(10^7, 10), :auto));

```

Execution on `main`:
```
julia> @benchmark outerjoin($df1, $df2, on=:x1, renamecols="left"=>"right")
BenchmarkTools.Trial: 
  memory estimate:  3.47 GiB
  allocs estimate:  479
  --------------
  minimum time:     2.890 s (27.14% GC)
  median time:      2.938 s (28.04% GC)
  mean time:        2.938 s (28.04% GC)
  maximum time:     2.986 s (28.92% GC)
  --------------
  samples:          2
  evals/sample:     1
```

Execution on this PR:
```
julia> @benchmark outerjoin($df1, $df2, on=:x1, renamecols="left"=>"right")
BenchmarkTools.Trial: 
  memory estimate:  3.47 GiB
  allocs estimate:  636
  --------------
  minimum time:     1.380 s (7.04% GC)
  median time:      1.632 s (19.31% GC)
  mean time:        1.640 s (18.85% GC)
  maximum time:     1.915 s (23.12% GC)
  --------------
  samples:          4
  evals/sample:     1
```

(using 8 threads in both cases)

src/join/composer.jl
```diff
@@ -207,21 +220,49 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol, makeunique:
...
+            @sync begin
+                for col in eachcol(dfl_noon)
+                    cols_i = left_idxs[col_idx]
+                    Threads.@spawn _noon_compose_helper(cols, _similar_left, cols_i,
+                                                        col, target_nrow, left_ixs, lil + 1, leftonly_ixs, loil)
+                    col_idx += 1
+                end
+                @assert col_idx == ncol(joiner.dfl) + 1
+                for col in eachcol(dfr_noon)
+                    let cols_i = col_idx
```
nalimilan, line 234: Why do you use `let` here but not above?

src/join/composer.jl
```diff
@@ -233,6 +274,13 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol, makeunique:
     return res, src_indicator
 end
 
+function _noon_compose_helper(cols, similar_col, cols_i, col, target_nrow,
+                              side_ixs, offset, sideonly_ixs, tocopy)
```
nalimilan, line 278: Can you add types to the signature? Maybe also add a `!` at the end of the function name.

nalimilan (commented):
Cool! I just wish we could find a way to avoid putting `if` checks everywhere in the long term.

bkamins:
> find a way to avoid putting `if` checks everywhere in the long term.

When we stop supporting Julia 1.0 it will be possible to significantly simplify the codes.

src/join/composer.jl
bkamins: You are right - `let` is not needed, as just writing `cols_i = col_idx` will ensure we get a new binding for `cols_i`. I will change this.

src/join/composer.jl
```diff
@@ -207,21 +220,49 @@ function compose_joined_table(joiner::DataFrameJoiner, kind::Symbol, makeunique:
...
+                    Threads.@spawn _noon_compose_helper(cols, _similar_left, cols_i,
+                                                        col, target_nrow, left_ixs, lil + 1, leftonly_ixs, loil)
+                    col_idx += 1
+                end
+                @assert col_idx == ncol(joiner.dfl) + 1
+                for col in eachcol(dfr_noon)
+                    let cols_i = col_idx
+                        Threads.@spawn _noon_compose_helper(cols, _similar_right, cols_i, col, target_nrow,
+                                                            right_ixs, lil + loil + 1, rightonly_ixs, roil)
+                    end
```
bkamins, line 237:
```suggestion
                    cols_i = col_idx
                    Threads.@spawn _noon_compose_helper(cols, _similar_right, cols_i, col, target_nrow,
                                                        right_ixs, lil + loil + 1, rightonly_ixs, roil)
```

src/join/composer.jl
bkamins: OK - fixed

nalimilan (approved):

bkamins:
Thank you!

Issue #2639, opened 2021-03-04
`GroupKey` should be comparable between DataFrames


ExpandingMan:
The `GroupKey` grouped dataframe keys are distinct from a `NamedTuple` or other "row-like" objects  for the sake of efficiency, but currently they are not comparable across dataframes.  This is not consistent with other "row-like" objects, all of which are comparable regardless of their parent.

This forbids some reasonable patterns such as
```julia
gs1 = groupby(df1, [:A, :B])
gs2 = groupby(df2, [:A, :B])

for k ∈ keys(gs1)
    # currently this would fail
    f(gs2[k])
end
```
As well as the direct creation of any other type of dictionary using `GroupKey` as a key.

Now, conversion of the `GroupKey` is quite simple with `NamedTuple`, but the question is whether this conversion should be left up to the user.

Note that, if we did change the behavior, some sort of safe but efficient reference check would be need on `getindex(::GroupedDataFrame, ::GroupKey)` to ensure that a compatible key is being used and if not, to perform the proper conversion.

ExpandingMan:
By the way, it might be worth doing some quick performance tests before seriously considering changing this.  While I think the inability to compare `GroupKey` is rather ugly, the possibility of converting to `NamedTuple` means this isn't such a big deal.  Even though I think it would be nice to change the behavior, I do not think it would be worth it if there were any kind of significant performance barrier.

bkamins:
Performance is not a problem. We can just change current:
```
function Base.to_index(gd::GroupedDataFrame, key::GroupKey)
    gd === parent(key) && return getfield(key, :idx)
    throw(ErrorException("Cannot use a GroupKey to index a GroupedDataFrame " *
                         "other than the one it was derived from."))
end
```
to
```
Base.to_index(gd::GroupedDataFrame, key::GroupKey) =
    gd === parent(key) ? return getfield(key, :idx) : Base.to_index(gd, NamedTuple(key))
```
the only thing is that it will be slow when mixing data frames (but will be as fast as currently for the same data frame).

Then we would need to define `==`, `isequal`, and `hash`, and think how `haskey` and `in` should work (and make sure we use consistent rules here).

@nalimilan - do you think we should consider these potential changes as breaking?

bkamins:
As a general comment - it should be easy to change once we decide what we want.

nalimilan:
[...]

PR #2669, opened 2021-03-22
add ==, isequal <, and isless for DataFrameRow and GroupKey

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2639.

As usual, before I add tests and update documentation please have a look at the implementation.

It would be good to have a decision on https://github.com/JuliaLang/julia/issues/40142 before finalizing this PR.

In particular we fix a bug according to which two `DataFrameRow` were `true` in `==` if they were the same row from the same data frame even if they contained `missing` (`missing` should be produced then).


bkamins:
Ah - and it is mildly breaking as we now define `hash` for `DataFrameRow` and `GroupKey` to be consistent with respective `NamedTuple` hash.

This means in particular that this hash is different than `rowhash` used in grouping (we could make `rowhash` consistent with the current `hash` - @nalimilan - do you think it is worth doing?).

src/other/utils.jl
```diff
@@ -128,3 +128,14 @@ function tforeach(f, x::AbstractArray; basesize::Integer)
     end
     return
 end
+
+function _nt_like_hash(v, h::UInt)
```
nalimilan, line 132: Is this faster than just doing `hash(NamedTuple(v), h)`?

src/dataframerow/dataframerow.jl
```diff
@@ -488,16 +483,40 @@ function Base.isequal(r1::DataFrameRow, r2::DataFrameRow)
...
+    return false
+end
+
+function Base.isless(r1::NamedTuple, r2::DataFrameRow)
+    length(r1) == length(r2) ||
+        throw(ArgumentError("compared objects must have the same number " *
+                            "of columns (got $(length(r1)) and $(length(r2)))"))
+    if propertynames(r1) != _names(r2)
+        mismatch = findfirst(i -> propertynames(r1)[i] != _names(r2)[i], 1:length(r1))
+        throw(ArgumentError("compared objects must have the same property names " *
```
nalimilan, line 517:
```suggestion
        throw(ArgumentError("compared objects must have the same property " *
```

src/dataframerow/dataframerow.jl
```diff
@@ -506,6 +525,23 @@ function Base.isless(r1::DataFrameRow, r2::DataFrameRow)
     return false
 end
 
+function Base.isless(r1::DataFrameRow, r2::NamedTuple)
+    length(r1) == length(r2) ||
+        throw(ArgumentError("compared objects must have the same number " *
+                            "of columns (got $(length(r1)) and $(length(r2)))"))
+    if propertynames(r1) != _names(r2)
+        mismatch = findfirst(i -> names(r1)[i] != propertynames(r2)[i], 1:length(r1))
+        throw(ArgumentError("compared objects must have the same property names " *
```
nalimilan, line 534:
```suggestion
        throw(ArgumentError("compared objects must have the same property " *
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -521,6 +521,131 @@ end
...
+end
+
+Base.isequal(k1::DataFrameRow, k2::GroupKey) = ==(k2, k1)
+
+function Base.isequal(k1::GroupKey, k2::NamedTuple)
+    parent(k1).cols == _propertynames(k2) || return false
+    return all(((a, b),) -> isequal(a, b), zip(k1, k2))
+end
+
+Base.isequal(k1::NamedTuple, k2::GroupKey) = ==(k2, k1)
```
nalimilan, line 562: Maybe this duplication could be avoided by using an `@eval` loop? The only difference is the function which is used with `all`.

nalimilan:
> This means in particular that this hash is different than `rowhash` used in grouping (we could make `rowhash` consistent with the new `hash` - @nalimilan - do you think it is worth doing?).

We probably don't care about this since it's internal, right?

bkamins:
> We probably don't care about this since it's internal, right?

Yes - I have checked that we use rowhash only in one place in the code so it should be OK.

src/groupeddataframe/groupeddataframe.jl
bkamins: I will remove most of these codes anyway (and then I will look at how to reduce duplication).
However I am not super fan of `@eval` as then the stacktrace users receive when they make some error are very hard to understand for non-experts.

src/other/utils.jl
bkamins: doing `NamedTuple(v)` can have a prohibitive compilation cost for wide data frames. Also it is faster. Here are timings for 5-column data frame:
```
julia> @benchmark _nt_like_hash($df[1, :], UInt(0))
BenchmarkTools.Trial:
  memory estimate:  368 bytes
  allocs estimate:  14
  --------------
  minimum time:     589.655 ns (0.00% GC)
  median time:      598.276 ns (0.00% GC)
  mean time:        654.433 ns (1.49% GC)
  maximum time:     9.713 μs (89.57% GC)
  --------------
  samples:          10000
  evals/sample:     174

julia> @benchmark hash(NamedTuple($df[1, :]))
BenchmarkTools.Trial:
  memory estimate:  416 bytes
  allocs estimate:  11
  --------------
  minimum time:     1.310 μs (0.00% GC)
  median time:      1.450 μs (0.00% GC)
  mean time:        1.537 μs (0.00% GC)
  maximum time:     5.290 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     10
```

bkamins:
Also `rowhash` ignores column names for speed, so indeed they do not have to match.

src/dataframerow/dataframerow.jl
```diff
@@ -455,55 +455,78 @@ Base.merge(a::DataFrameRow, b::DataFrameRow) = merge(NamedTuple(a), NamedTuple(b
...
 end
 
-function Base.isequal(r1::DataFrameRow, r2::DataFrameRow)
-    if parent(r1) === parent(r2)
-        parentcols(index(r1)) == parentcols(index(r2)) || return false
-        row(r1) == row(r2) && return true
-    else
-        _names(r1) == _names(r2) || return false
+
+for (fun1, fun2) in ((Base.isequal, Base.isless), (Base.:(==), Base.:(<)))
```
nalimilan, line 479: Maybe call these `eqfun` and `cmpfun` or something explicit like that?

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -521,6 +521,117 @@ end
...
+            $fun1(a, b) || return $fun2(a, b)
+        end
+        return false
+    end
+
+    function $fun2(k1::GroupKey, k2::DataFrameRow)
+        length(k1) == length(k2) ||
+            throw(ArgumentError("compared objects must have the same number " *
+                                "of columns (got $(length(k1)) and $(length(k2)))"))
+        if parent(k1).cols != _names(k2)
```
nalimilan, line 570: You defined a helper like this, the same definition would work for both types, right?
```julia
_getnames(x::GroupKey) = parent(x).cols 
_getnames(x::DataFrameRow) = _names(x)
```

src/dataframerow/dataframerow.jl
bkamins: OK - will change.

src/groupeddataframe/groupeddataframe.jl
bkamins: I will change this. I was thinking about it, but was too lazy ;(.

bkamins:
OK - I hope I fixed everything :). This was hard (but hopefully now Julia Base will be also more consistent with `<` handling).

src/dataframerow/dataframerow.jl
```diff
@@ -460,12 +460,23 @@ Base.hash(r::DataFrameRow, h::UInt) = _nt_like_hash(r, h)
 _getnames(x::DataFrameRow) = _names(x)
 _getnames(x::NamedTuple) = propertynames(x)
 
+# this is required as == does not allow for comparison of tuples and vectors
```
nalimilan, line 463:
```suggestion
# this is required as == does not allow for comparison between tuples and vectors
```

test/dataframerow.jl
```diff
@@ -586,4 +586,68 @@ end
...
+    gk = [keys(groupby(df, [:a, :b])); keys(groupby(df, :a))]
+
+    for l in (dfr[1], nt[1], gk[1]), r in (dfr[1], nt[1], gk[1])
+        @test ismissing(l == r)
+        @test ismissing(l == (a=1, b=2))
+        @test l ≅ r
+        @test l ≇ (a=1, b=2)
+        # work around https://github.com/JuliaLang/julia/pull/40147
+        if !(l isa NamedTuple && r isa NamedTuple)
+            @test ismissing(l < r)
```
nalimilan, line 602: It could be interesting to test the equivalent of `(1, missing) < (2, 1)`, which is a corner case that shouldn't give `missing` but is easy to get wrong.

bkamins: I will add it

nalimilan (approved):

bkamins:
@pdeffebach - this is what we concluded is a good design. Could you please comment on it before merging (as you were against it)?

pdeffebach:
Sounds good! I'm fine with this.

src/dataframerow/utils.jl
```diff
@@ -450,6 +450,16 @@ function group_rows(df::AbstractDataFrame)
     return RowGroupDict(df, rhashes, gslots, groups, rperm, starts, stops)
 end
 
+# hash of DataFrame rows based on its values
+# so that duplicate rows would have the same hash
+# table columns are passed as a tuple of vectors to ensure type specialization
+rowhash(cols::Tuple{AbstractVector}, r::Int, h::UInt = zero(UInt))::UInt =
+    hash(cols[1][r], h)
+function rowhash(cols::Tuple{Vararg{AbstractVector}}, r::Int, h::UInt = zero(UInt))::UInt
```
bkamins, line 458: it is interesting that we do not have this covered by tests. I will add some.

bkamins: it turns out we do not use these functions. I have removed them. @nalimilan - could you please have a look if we would ever need them? (I have never used them)

test/dataframerow.jl
```diff
@@ -171,42 +171,26 @@ end
     @test_throws ArgumentError df[1, 1:2] < df[1, 2:3]
 end
 
-@testset "hashing" begin
+@testset "hashing of DataFrameRow and GroupKey" begin
     df = deepcopy(ref_df)
+    gks = keys(groupby(df, :))
```
bkamins, line 176:
```suggestion
```

test/dataframerow.jl
```diff
@@ -171,42 +171,26 @@ end
...
-    # grouping empty frame
-    gd = DataFrames.group_rows(DataFrame(x=Int[]))
-    @test length(unique(gd.groups)) == 0
-
-    # grouping single row
-    gd = DataFrames.group_rows(df1[1:1, :])
-    @test length(unique(gd.groups)) == 1
+    for i in axes(df, 1), h in UInt(0):UInt(10)
+        @test hash(DataFrameRow(df, i, :), h) ==
+              hash(NamedTuple(DataFrameRow(df, i, :)), h)
```
bkamins, line 192:
```suggestion
        @test hash(DataFrameRow(df, i, :), h) ==
              hash(gks[i], h) ==
              hash(NamedTuple(DataFrameRow(df, i, :)), h)
```

src/dataframerow/utils.jl
nalimilan: AFAICT one guy removed the last use of `findrow` and `group_rows` in a recent PR. :-D https://github.com/JuliaData/DataFrames.jl/pull/2641

It's weird that coverage didn't notice this. Maybe we only looked at the changed code and not at unrelated parts?

src/dataframerow/utils.jl
```diff
@@ -440,57 +440,3 @@ function compute_indices(groups::AbstractVector{<:Integer}, ngroups::Integer)
...
     return rperm, starts, stops
 end
-
-# Build RowGroupDict for a given DataFrame, using all of its columns as grouping keys
-function group_rows(df::AbstractDataFrame)
-    groups = Vector{Int}(undef, nrow(df))
-    ngroups, rhashes, gslots, sorted =
-        row_group_slots(ntuple(i -> df[!, i], ncol(df)), Val(true), groups, false, false)
-    rperm, starts, stops = compute_indices(groups, ngroups)
-    return RowGroupDict(df, rhashes, gslots, groups, rperm, starts, stops)
```
nalimilan, line 450: I think `RowGroupDict` can also be removed.

bkamins: Indeed - I am  running tests to double check. It is astonishing how much we have reworked internally in this release.

src/dataframerow/utils.jl
bkamins: There were tests that checked correctness of internal functions.

src/dataframerow/utils.jl
bkamins: OK - all seems clean. I will merge the PR after CI passes.

nalimilan: Now that this file contains only grouping code, we will be able to move it to the corresponding folder and rename it without touching anything else. :-)

bkamins: OK will move it in this PR as otherwise we will forget. Also `nonunique` uses it, but I think it is not a problem.

bkamins:
Thank you!

pdeffebach:
This allows 

```
(a = 1, b = 2) in eachrow(df)
```

right?

bkamins:
Right, and even more e.g.:
```
julia> df = DataFrame(a=1:5, b=2:6)
5×2 DataFrame
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2
   2 │     2      3
   3 │     3      4
   4 │     4      5
   5 │     5      6

julia> (a=1, b=2) in eachrow(df)
true

julia> Ref((a=4, b=4)) .< eachrow(df)
5-element BitVector:
 0
 0
 0
 1
 1
```

PR #2670, opened 2021-03-23
Fix small typo in docs

eirikbrandsaas:
Maybe I'm misunderstanding the manual here, but shouldn't the `row` be `df2` as in the other options for `cols`?


bkamins (approved):

bkamins:
Thank you!

PR #2679, opened 2021-03-25
remove deprecations for 1.0 release

bkamins:


bkamins:
re-trigger CI, as on nightly strange `OutOfMemory` error happens in multi-threaed code that I cannot reproduce

src/other/index.jl
```diff
@@ -218,7 +218,7 @@ end
     setdiff(1:length(x), getindex(x, notidx.skip))
 @inline Base.getindex(x::AbstractIndex, idx::Between) = x[idx.first]:x[idx.last]
 @inline Base.getindex(x::AbstractIndex, idx::All) =
-    isempty(idx.cols) ? (1:length(x)) : union(getindex.(Ref(x), idx.cols)...)
+    isempty(idx.cols) ? (1:length(x)) : throw(ArgumentError("All(args...) is not supported"))
```
nalimilan, line 221:
```suggestion
    isempty(idx.cols) ? (1:length(x)) : throw(ArgumentError("All(args...) is not supported: use Cols(args...) instead"))
```

nalimilan (approved):

bkamins:
This out-of-memory error https://github.com/JuliaData/DataFrames.jl/pull/2679/checks?check_run_id=2209199932 randomly happens on nightly.

@nalimilan - to whom do you think we should report it? It is related to multi-threading.

nalimilan:
Maybe it's just that the VM has a limited amount of RAM? We didn't had tests with several million rows until recently. We could use 3 or even 2 columns instead of 4 there.

bkamins:
OK. I will merge this and open a PR reducing the memory use.

PR #2680, opened 2021-03-26
make Tables.columns(df) return eachcol(df)

### LINKED ISSUE #2636: Usage of `eachcol` seems to be a pun?
I'm opening this in response to https://github.com/JuliaData/TypedTables.jl/issues/68, with the hope of improving consistency across the ecosystem.

It seems that DataFrames.jl overloads `Base.eachcol` to return the column interator `DataFrameColumns`. To me, this seems like a bit of a pun, and the `columns` function from `Tables.jl` seems to be a much better fit for this behavior.

While this seems harmless at first, tables (in the Tables.jl sense) like `Matrix{NamedTuple}` or a `TypedTables.Table <: AbstractMatrix` are required to follow `Base` behavior. I see `Base.eachcol` as an `AbstractArray` interface - and one that is actively being built upon in `Base` (soon it will return a lazy vector-of-vectors `SliceArray` rather than just an iterator - see https://github.com/JuliaLang/julia/pull/32310).

This leaves the data table ecosystem a little bit split. Even ignoring TypedTables, users may first learn to use a `CSV.File` and a `Vector{<:NamedTuple}` and will need to use `columns` there and switch to `eachcol` later when they learn DataFrames.jl. Do you think it would be reasonable to use the Tables.jl interface `columns` to return a `DataFrameColumns` instead?

bkamins:
Fixes #2636 

In the end I decided only to change the return value of `Tables.columns` to `DataFrameColumns` to ensure iteration is supported for it.
I did not go for `DataFrameColumns <: Tables.AbstractColumns` as it did not seem necessary (it would only add `haskey` and `get` methods - we can always add this support later if needed).

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -1,6 +1,6 @@
 Tables.istable(::Type{<:AbstractDataFrame}) = true
 Tables.columnaccess(::Type{<:AbstractDataFrame}) = true
-Tables.columns(df::AbstractDataFrame) = df
+Tables.columns(df::AbstractDataFrame) = eachcol(df)
 Tables.rowaccess(::Type{<:AbstractDataFrame}) = true
 Tables.rows(df::AbstractDataFrame) = eachrow(df)
 Tables.rowtable(df::AbstractDataFrame) = Tables.rowtable(Tables.columntable(df))


quinnj (approved):
SGTM; yeah, the real benefit of `<: Tables.AbstractColumns` is to get the `getindex`, `getproperty`, `haskey`, `get`, etc. methods for free, so if those are mostly all defined anyway, there's not much benefit.

So just to make sure I understand completely: `eachcol(df)` returns `DataFrameColumns`, and that supports `Tables.columnnames` and `Tables.getcolumn` methods, right? Just want to ensure I understand that correctly.

bkamins:
Yes `DataFrameColumns` supports the API. The only things that are not supported are `get` and `haskey`.
`Tables.columnames` are supported by the default fall-back, and `Tables.getcolumn` is defined here https://github.com/JuliaData/DataFrames.jl/blob/main/src/other/tables.jl#L88.

`Tables.getcolumn(table, ::Type{T}, i::Int, nm::Symbol)` is not implemented and falls back to the default.

quinnj:
Ok, that sounds good to me. `get` and `haskey` aren't required for result of `Tables.columns`, just `Tables.columnnames` and `Tables.getcolumn`, so we're good.

bkamins:
Thank you!

Issue #2423, opened 2020-09-11
make sure aggregation uses fast path if a single column is used


bkamins:
I will fix it. Currently `:x => sum` uses a fast path, but `[:x] => sum` does not.

sa-:
Looking into this. Pasting some of your Slack messages here in case the messages get deleted: 

```
fast path is implemented here: https://github.com/JuliaData/DataFrames.jl/blob/main/src/groupeddataframe/fastaggregates.jl

10:30
In the fast path you do not have to calculate grouping views. You can just pre-allocate the target container (whose eltype you know upfront) and just iterate groups and perform reduction.
```

bkamins:
The key thing to do here is to rewrite code like:
```
@chain df begin
    groupby(:a)
    combine([:b] => sum)
end
```
to:
```
@chain df begin
    groupby(:a)
    combine(:b => sum)
end
```
before executing it.

We should check in the syntax `src => fun` and `src => fun => dst` if `src` is `Union{AbstractVector{<:AbstractString}}, AbstractVector{Symbol}, AbstractVector{<:Union{Signed, Unsigned}}` of length 1 and if this is so its first element should be fetched with `only` before further processing.

sa-:
Can this be closed now?

bkamins:
yes

PR #2681, opened 2021-03-26
Small no-op refactor before functionality change

sa-:
In preparation for #2423 I'm making a no-op refactor.

I would feel comfortable going through the process of contributing with a small no-op change like this one before making any functionality changes.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -52,16 +52,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = [
-  "CategoricalArrays", 
-  "Chain, 
-  "Combinatorics", 
-  "DataStructures", 
-  "DataValues", 
-  "Dates", 
-  "Logging", 
-  "OffsetArrays", 
-  "Random", 
-  "Test", 
-  "Unitful"
-]
+test = ["CategoricalArrays", "Chain, "Combinatorics", "DataStructures", "DataValues", "Dates", "Logging", "OffsetArrays", "Random", "Test", "Unitful"]
--- a/Project.toml
+++ b/Project.toml
@@ -52,4 +52,15 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = ["CategoricalArrays", "Chain, "Combinatorics", "DataStructures", "DataValues", "Dates", "Logging", "OffsetArrays", "Random", "Test", "Unitful"]
+test = [
+  "CategoricalArrays", 
+  "Chain", 
+  "Combinatorics", 
+  "DataStructures", 
+  "DataValues", 
+  "Dates", 
+  "Logging", 
+  "OffsetArrays", 
+  "Random", 
+  "Test", 
+  "Unitful"]
--- a/Project.toml
+++ b/Project.toml
@@ -40,6 +40,7 @@ julia = "1"
 
 [extras]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
+Chain = "8be319e6-bccf-4806-a6f7-6fae938471bc"
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 DataValues = "e7dc6d0d-1eca-5fa6-8ad6-5aecde8b7ea5"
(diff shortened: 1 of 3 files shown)


src/groupeddataframe/splitapplycombine.jl
```diff
@@ -19,34 +19,31 @@ function gen_groups(idx::Vector{Int})
...
                                         ColumnIndex, MultiColumnIndex}...);
                           keepkeys::Bool, ungroup::Bool, copycols::Bool,
                           keeprows::Bool, renamecols::Bool)
     if !ungroup && !keepkeys
         throw(ArgumentError("keepkeys=false when ungroup=false is not allowed"))
     end
 
-    cs_vec = []
-    for p in cs
+    processed_combine_ops = map(combine_operations) do p
```
bkamins, line 30: we do not want to use `map` here to avoid excessive specialization. Please leave the `Vector{Any}` as a container for `processed_combine_ops`.

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -65,7 +62,9 @@ function _combine_prepare(gd::GroupedDataFrame,
 
     idx, valscat = _combine(gd, cs_norm, optional_transform, copycols, keeprows, renamecols)
 
-    !keepkeys && ungroup && return valscat
+    if !keepkeys && ungroup
```
bkamins, line 65: I think the previous code was idiomatic Julia style. Why do you want to change it?

Project.toml
```diff
@@ -40,6 +40,7 @@ julia = "1"
 
 [extras]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
+Chain = "8be319e6-bccf-4806-a6f7-6fae938471bc"
```
bkamins, line 43: I do not see it used in this PR

bkamins:
Also looking by the CI errors you have changed the logic somewhere (I did not dig into it as I understand this is still a draft).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -19,34 +19,31 @@ function gen_groups(idx::Vector{Int})
...
+        end
+
         if p === nrow
-            push!(cs_vec, nrow => :nrow)
-        elseif p isa AbstractVecOrMat{<:Pair}
-            append!(cs_vec, p)
+            nrow => :nrow
         else
-            push!(cs_vec, p)
+            p
```
bkamins, line 40: this line is incorrect as you mix vector and scalar values in it. Note the distinction between `append!` and `push!` in the original code.

bkamins:
In summary: thank you very much for working on it. I think that the refactorizations that improve code readability are very useful and we should go for them.

However, I would recommend you to start with refactorizations that do not change the logic of the code (like here you tried to change `for` loop to `map` and this change has a significant impact on the code logic).

Of course I am not opposed to such changes in general in the future - if they would improve things, but they are much more tricky than just no-op refactorizations, so it is probably better to do them when you get more experience with the codebase.

bkamins:
Also - in general - especially to start with - it is best if you push PR that pass tests locally with:
```
]test DataFrames
```
In the future you will notice that this is not a strict rule as sometimes one pushes a PR that fails to discuss some design ideas, but for start I think it is useful to build a habit that commits are pushed to GitHub when they pass tests locally.

sa- (commented):
Hey, I wasn't expecting anyone to look at this but I appreciate that you did anyway!

Some thoughts - 

- Makes sense that the refactoring would be a good idea after I'm comfortable with the codebase
- I was hoping that my refactoring changes would not make any functional changes but that seems harder to do than I initially thought e.g. I didn't think changing for to map would make a difference but I've clearly broken something.

sa-:
Hey, thanks for looking through this PR!

bkamins:
> I didn't think changing for to map would make a difference but I've clearly broken something.

Here is an example. Current code:
```
julia> cs = (nrow, [1,2], 3, [1 => 4,  2 => 5])
(DataFrames.nrow, [1, 2], 3, [1 => 4, 2 => 5])

julia> vec_cs = Any[]
Any[]

julia> for p in cs
           if p === nrow
               push!(vec_cs, nrow => :nrow)
           elseif p isa AbstractVecOrMat{<:Pair}
               append!(vec_cs, p)
           else
               push!(vec_cs, p)
           end
       end

julia> vec_cs
5-element Vector{Any}:
 DataFrames.nrow => :nrow
                  [1, 2]
                 3
               1 => 4
               2 => 5
```
your code:
```
julia> map(cs) do p
           if p === nrow
               nrow => :nrow
           else
               p
           end
       end
(DataFrames.nrow => :nrow, [1, 2], 3, [1 => 4, 2 => 5])
```

So you can see that:
1. type of the return value changed
2. structure of the collection changed

A more advanced comment was about specialization. Methods for `Vector{Any}` are precompiled in Julia Base, so they do not trigger specialization. If you use `map` a new `Tuple` is created. As `Tuple` has a type that depends on all its elements the whole function has to be recompiled on each call, while the original function had to be compiled only once (on the first call).

> I wasn't expecting anyone to look at this but I appreciate that you did anyway!

I assumed you wanted to get some feedback to learn a bit about the design concepts behind DataFrames.jl internals.

Issue #2424, opened 2020-09-11
Resolve type instability issues in aggregation


bkamins:
See https://gist.github.com/nalimilan/b801d48ab2931b6a671f05081fd94ba3 and in particular
https://gist.github.com/nalimilan/b801d48ab2931b6a671f05081fd94ba3#file-splitapplycombine-jl-mem-L1402
and
https://gist.github.com/nalimilan/b801d48ab2931b6a671f05081fd94ba3#file-splitapplycombine-jl-mem-L1438
and
https://gist.github.com/nalimilan/b801d48ab2931b6a671f05081fd94ba3#file-splitapplycombine-jl-mem-L1439

nalimilan:
In the following functions, the first two allocate, but not the last one:
```julia
function fill_row2!(row, col::AbstractVector,
                    i::Integer,
                    colnames::NTuple{N, Symbol}) where N
    @inbounds for j in 1:length(colnames)
        cn = colnames[j]
        val = row[cn]
        col[i] = val
    end
    return nothing
end

function fill_row2b!(row, col::AbstractVector,
                     i::Integer,
                     colnames::NTuple{N, Symbol}) where N
    @inbounds for cn in colnames
        val = row[cn]
        col[i] = val
    end
    return nothing
end

function fill_row3!(row, col::AbstractVector,
                    i::Integer,
                    colname::Symbol) where N
    val = row[colname]
    col[i] = val
    return nothing
end
```

A possible workaround is to require column names to be always in the same order: then we could index using an integer, which I just checked doesn't allocate.

bkamins:
This is a bug. We have to fix it and insist that column names always have the same order. I will fix it. Otherwise we have:
```
julia> df = DataFrame(x = 1:4)
4×1 DataFrame
 Row │ x     
     │ Int64 
─────┼───────
   1 │     1
   2 │     2
   3 │     3
   4 │     4

julia> gdf = groupby(df, :x)
GroupedDataFrame with 4 groups based on key: x
First Group (1 row): x = 1
 Row │ x     
     │ Int64 
─────┼───────
   1 │     1
⋮
Last Group (1 row): x = 4
 Row │ x     
     │ Int64 
─────┼───────
   1 │     4

julia> combine(gdf, :x => f => AsTable)
4×3 DataFrame
 Row │ x      b      a     
     │ Int64  Int64  Int64 
─────┼─────────────────────
   1 │     1      1     11
   2 │     2     12      2
[...]

PR #2682, opened 2021-03-27
check if column order is correct in aggregations

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2424

I still need to update tests and docs (which will also make sure we have no bug in my code :smile:). I will do it later this weekend.

The benchmarks are.

Setup:
```
julia> df = DataFrame(x = 1:10^6);

julia> gdf = groupby(df, :x);

julia> f(x) = return (a=x[1], b=x[1]+1)
```

this PR:
```
julia> @benchmark combine($gdf, :x => f => AsTable)
BenchmarkTools.Trial: 
  memory estimate:  129.72 MiB
  allocs estimate:  1000280
  --------------
  minimum time:     225.386 ms (1.09% GC)
  median time:      242.938 ms (2.17% GC)
  mean time:        252.185 ms (5.65% GC)
  maximum time:     349.955 ms (27.36% GC)
  --------------
  samples:          20
  evals/sample:     1
```

0.22.6 release:
```
julia> @benchmark combine($gdf, :x => f => AsTable)
BenchmarkTools.Trial: 
  memory estimate:  358.56 MiB
  allocs estimate:  10998173
  --------------
  minimum time:     261.659 ms (6.93% GC)
  median time:      300.361 ms (7.92% GC)
  mean time:        337.227 ms (15.66% GC)
  maximum time:     443.614 ms (23.25% GC)
  --------------
  samples:          15
  evals/sample:     1
```

so we:
1. improve performance a bit
2. significantly cut down allocations
3. avoid a bug described in #2424

### DIFF
--- a/src/groupeddataframe/complextransforms.jl
+++ b/src/groupeddataframe/complextransforms.jl
@@ -4,9 +4,6 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
 _ncol(df::AbstractDataFrame) = ncol(df)
 _ncol(x::Union{NamedTuple, DataFrameRow}) = length(x)
 
-_getnames(x::AbstractDataFrame) = _names(x)
-_getnames(x::Any) = throw(ArgumentError("names not defined for argument of type $(typeof(x))"))
-
 function _combine_multicol(firstres, fun::Base.Callable, gd::GroupedDataFrame,
                            incols::Union{Nothing, AbstractVector, Tuple, NamedTuple})
     firstmulticol = firstres isa MULTI_COLS_TYPE
--- a/src/groupeddataframe/complextransforms.jl
+++ b/src/groupeddataframe/complextransforms.jl
@@ -82,7 +82,7 @@ function fill_row!(row, outcols::NTuple{N, AbstractVector},
                    colnames::NTuple{N, Symbol}) where N
     if !_names_match(row, colnames)
         throw(ArgumentError("return value must have the same column names " *
-                           "for all groups (got $colnames and $(propertynames(row)))"))
+                            "for all groups (got $colnames and $(propertynames(row)))"))
     end
     @inbounds for j in colstart:length(outcols)
         col = outcols[j]
@@ -315,7 +315,7 @@ function append_rows!(rows, outcols::NTuple{N, AbstractVector},
                       colstart::Integer, colnames::NTuple{N, Symbol}) where N
 if !_names_match(rows, colnames)
         throw(ArgumentError("return value must have the same column names " *
-                           "for all groups (got $colnames and $(propertynames(row)))"))
+                            "for all groups (got $colnames and $(propertynames(rows)))"))
     end
     @inbounds for j in colstart:length(outcols)
         col = outcols[j]
--- a/src/groupeddataframe/complextransforms.jl
+++ b/src/groupeddataframe/complextransforms.jl
@@ -304,7 +304,6 @@ end
 
 _get_col(rows::AbstractDataFrame, j::Int) = rows[!, j]
 _get_col(rows::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}, j::Int) = rows[j]
-_get_col(rows::Any, j::Int) = throw(ArgumentError(ERROR_ROW_COUNT))
 
 function append_rows!(rows, outcols::NTuple{N, AbstractVector},
                       colstart::Integer, colnames::NTuple{N, Symbol}) where N
--- a/src/groupeddataframe/complextransforms.jl
+++ b/src/groupeddataframe/complextransforms.jl
@@ -4,6 +4,9 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
 _ncol(df::AbstractDataFrame) = ncol(df)
 _ncol(x::Union{NamedTuple, DataFrameRow}) = length(x)
 
+_getnames(x::AbstractDataFrame) = _names(x)
+_getnames(x::Any) = throw(ArgumentError("names not defined for argument of type $(typeof(x))"))
+
 function _combine_multicol(firstres, fun::Base.Callable, gd::GroupedDataFrame,
                            incols::Union{Nothing, AbstractVector, Tuple, NamedTuple})
     firstmulticol = firstres isa MULTI_COLS_TYPE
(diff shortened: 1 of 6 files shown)


bkamins:
The same benchmark but for `df = DataFrame(x = 1:10^7)` is:

this PR:
```
julia> @time combine(gdf, :x => f => AsTable);
  2.380177 seconds (10.00 M allocations: 1.267 GiB, 4.92% gc time)
```

0.22.6 release:
```
julia> @time combine(gdf, :x => f => AsTable);
  3.461393 seconds (110.00 M allocations: 3.502 GiB, 12.10% gc time)
```

src/groupeddataframe/complextransforms.jl
```diff
@@ -300,24 +306,21 @@ else
...
+
+
 function append_rows!(rows, outcols::NTuple{N, AbstractVector},
                       colstart::Integer, colnames::NTuple{N, Symbol}) where N
-    if !isa(rows, Union{AbstractDataFrame, NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}})
-        throw(ArgumentError(ERROR_ROW_COUNT))
-    elseif _ncol(rows) != N
-        throw(ArgumentError("return value must have the same number of columns " *
-                            "for all groups (got $N and $(_ncol(rows)))"))
+if !_names_match(rows, colnames)
```
nalimilan, line 316:
```suggestion
    if !_names_match(rows, colnames)
```

src/groupeddataframe/complextransforms.jl
```diff
@@ -300,24 +306,21 @@ else
...
+        throw(ArgumentError("return value must have the same column names " *
+                           "for all groups (got $colnames and $(propertynames(row)))"))
```
nalimilan, line 318:
```suggestion
                            "for all groups (got $colnames and $(propertynames(row)))"))
```

src/groupeddataframe/complextransforms.jl
```diff
@@ -300,24 +306,21 @@ else
     end
 end
 
+_get_col(rows::AbstractDataFrame, j::Int) = rows[!, j]
+_get_col(rows::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}, j::Int) = rows[j]
+_get_col(rows::Any, j::Int) = throw(ArgumentError(ERROR_ROW_COUNT))
+
+
```
nalimilan, line 313:
```suggestion
```

src/groupeddataframe/complextransforms.jl
```diff
@@ -64,23 +67,26 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
...
 
+function _names_match(r1, t::NTuple{N, Symbol}) where N
+    n1 = _getnames(r1)
+    length(n1) == length(t) || return false
+    for (a, b) in zip(n1, t)
+        a == b || return false
+    end
+    return true
+end
+
```
nalimilan, line 78:
```suggestion
```

src/groupeddataframe/complextransforms.jl
```diff
@@ -4,6 +4,9 @@ _nrow(x::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}) =
 _ncol(df::AbstractDataFrame) = ncol(df)
 _ncol(x::Union{NamedTuple, DataFrameRow}) = length(x)
 
+_getnames(x::AbstractDataFrame) = _names(x)
+_getnames(x::Any) = throw(ArgumentError("names not defined for argument of type $(typeof(x))"))
```
nalimilan, line 8: When can this happen?

src/groupeddataframe/complextransforms.jl
```diff
@@ -64,23 +67,26 @@ function _combine_with_first(first::Union{NamedTuple, DataFrameRow, AbstractData
     return idx, outcols, collect(Symbol, finalcolnames)
 end
 
+function _names_match(r1, t::NTuple{N, Symbol}) where N
+    n1 = _getnames(r1)
```
nalimilan, line 71: Maybe have the function take names directly? Than would avoid specializing it on different row types.

bkamins: OK

src/groupeddataframe/complextransforms.jl
bkamins: It should not happen. But I thought to add it to avoid `MethodError` but rather get an informative message. It will try to come up with a case when it is actually triggered :) (hopefully I will not find such).

src/groupeddataframe/complextransforms.jl
bkamins: But wait - different row types will produce different `_getnames(r1)` type anyway, so I do not think it is much benefit of this.

src/groupeddataframe/complextransforms.jl
bkamins: I have another question - should I group the `_getnames` definitions in one place?

The issue is that they are now spread around the code-base as they are defined where they were needed. The issue is that they refer to types that are introduced in different parts of the code base (a pain of not having DataFrames.jl having the code layout that we first define all types and then use them).

nalimilan: Ah yeah if these overload the same function it would be clearer to have them in the same place.

src/groupeddataframe/complextransforms.jl
nalimilan: Don't e.g. `DataFrameRow` and `AbstractDataFrame` give the same type?

bkamins: No, `DataFrameRow` produces a `view`:
```
julia> using DataFrames

julia> df = DataFrame(x=1:2)
2×1 DataFrame
 Row │ x     
     │ Int64 
─────┼───────
   1 │     1
   2 │     2

julia> DataFrames._names(df)
1-element Vector{Symbol}:
 :x

julia> DataFrames._names(df[1, :])
1-element view(::Vector{Symbol}, Base.OneTo(1)) with eltype Symbol:
 :x
```

src/groupeddataframe/complextransforms.jl
bkamins: OK - I moved them, and I removed the `::Any` case as I could not think of the scenario when it would be triggered.

src/other/names.jl
```diff
@@ -0,0 +1,17 @@
+# common utilities for getting and comparing column names
```
nalimilan, line 1: Is it really worth a separate file? Maybe utils.jl is OK?

bkamins: It is not OK. It must be placed after all types are defined. The last type introduced is `GroupKey` in `groupeddataframe/groupeddataframe.jl`, but I did not want to put it there as it is also used in `DataFrameRow` operations.

bkamins: OK - so what do we with this? I think we can leave it as is for now.

nalimilan (approved):

src/groupeddataframe/complextransforms.jl
```diff
@@ -300,24 +302,20 @@ else
     end
 end
 
+_get_col(rows::AbstractDataFrame, j::Int) = rows[!, j]
+_get_col(rows::NamedTuple{<:Any, <:Tuple{Vararg{AbstractVector}}}, j::Int) = rows[j]
+_get_col(rows::Any, j::Int) = throw(ArgumentError(ERROR_ROW_COUNT))
```
bkamins, line 307:
```suggestion
```

bkamins:
Thank you!

PR #2685, opened 2021-03-28
Enable multithreading in hashrows_col!

nalimilan:
This is somewhat faster for large number of rows (by about 20% for 10M rows and 10 threads), but a bit slower when the number of rows is lower than the threshold (so that a single thread is used). Not sure whether something can be done about it, nor whether the improvement in some cases is worth the loss in others.

```julia
julia> using Revise, DataFrames, BenchmarkTools

julia> for n in (1000, 10_000_000)
           df = DataFrame(x=rand(string.(1000000:1000010), n));
           @btime groupby($df, :x);

          # Benchmark directly the changes
           h = similar(df.x, UInt);
           n = similar(df.x, Bool);
           @btime DataFrames.hashrows_col!($h, $n, $df.x, nothing, true);
       end

# main
  21.105 μs (29 allocations: 34.05 KiB)
  12.211 μs (0 allocations: 0 bytes)
  335.166 ms (32 allocations: 280.59 MiB)
  122.653 ms (0 allocations: 0 bytes)

# PR
  26.041 μs (29 allocations: 34.05 KiB)
  16.291 μs (0 allocations: 0 bytes)
  258.437 ms (107 allocations: 280.60 MiB)
  18.463 ms (74 allocations: 5.72 KiB)
```

### DIFF
--- a/src/groupeddataframe/utils.jl
+++ b/src/groupeddataframe/utils.jl
@@ -5,7 +5,7 @@ function hashrows_col!(h::Vector{UInt},
                        v::AbstractVector{T},
                        rp::Nothing,
                        firstcol::Bool) where T
-    @splittasks 1_000_000 for i in eachindex(h)
+    @spawn_for_chunks 1_000_000 for i in eachindex(h)
         @inbounds begin
             el = v[i]
             h[i] = hash(el, h[i])
@@ -33,18 +33,18 @@ function hashrows_col!(h::Vector{UInt},
         fira = firstindex(ra)
 
         hashes = Vector{UInt}(undef, length(rp))
-        @splittasks 1_000_000 for i in eachindex(hashes)
+        @spawn_for_chunks 1_000_000 for i in eachindex(hashes)
             @inbounds hashes[i] = hash(rp[i+firp-1])
         end
 
         # here we rely on the fact that `DataAPI.refpool` has a continuous
         # block of indices
-        @splittasks 1_000_000  for i in eachindex(h)
+        @spawn_for_chunks 1_000_000  for i in eachindex(h)
             @inbounds ref = ra[i+fira-1]
             @inbounds h[i] = hashes[ref+1-firp]
         end
     else
-        @splittasks 1_000_000 for i in eachindex(h, v)
+        @spawn_for_chunks 1_000_000 for i in eachindex(h, v)
             @inbounds h[i] = hash(v[i], h[i])
         end
     end
--- a/src/groupeddataframe/utils.jl
+++ b/src/groupeddataframe/utils.jl
@@ -5,11 +5,13 @@ function hashrows_col!(h::Vector{UInt},
                        v::AbstractVector{T},
                        rp::Nothing,
                        firstcol::Bool) where T
-    @inbounds for i in eachindex(h)
-        el = v[i]
-        h[i] = hash(el, h[i])
-        if length(n) > 0
-            n[i] |= ismissing(el)
+    tforeach(eachindex(h), basesize=1_000_000) do i
+        @inbounds begin
+            el = v[i]
+            h[i] = hash(el, h[i])
+            if length(n) > 0
+                n[i] |= ismissing(el)
+            end
         end
     end
     h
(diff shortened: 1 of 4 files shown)


bkamins:
I was testing the same when implementing hashing in `OnCols` for joins and had similar conclusions. We probably can have it, but the threshold for turning "this on" should probably be large.

nalimilan:
> I was testing the same when implementing hashing in `OnCols` for joins and had similar conclusions. We probably can have it, but the threshold for turning "this on" should probably be large.

The point is, the regression happens even under the threshold. It looks like we pay the cost even when multithreading isn't used.

bkamins:
My guess is that we would have to split it into several methods. Maybe (I have not checked, but seen it in other cases), when you have very long methods that branch compiles is not always doing the right thing. If you will not find the root cause of this I can look at the issue later this week.

nalimilan:
I've switched to using a macro (inspired by `@threads` but using `@spawn`) instead of a function, and the overhead is negligible now. Actually the syntax is even nicer, it's almost identical to the non-threaded version. I guess it could replace `tforeach`.

Updated benchmark (with 4 threads):
main:
```julia
  16.520 μs (29 allocations: 34.05 KiB)
  9.406 μs (0 allocations: 0 bytes)
  207.426 ms (32 allocations: 280.59 MiB)
  100.835 ms (0 allocations: 0 bytes)
```

PR:
```julia
  17.076 μs (29 allocations: 34.05 KiB)
  10.564 μs (0 allocations: 0 bytes)
  139.823 ms (93 allocations: 280.60 MiB)
  33.247 ms (61 allocations: 5.16 KiB)
```

src/groupeddataframe/utils.jl
```diff
@@ -5,11 +5,13 @@ function hashrows_col!(h::Vector{UInt},
...
-        h[i] = hash(el, h[i])
-        if length(n) > 0
-            n[i] |= ismissing(el)
+    @splittasks 1_000_000 for i in eachindex(h)
+        @inbounds begin
+            el = v[i]
+            h[i] = hash(el, h[i])
+            if length(n) > 0
+                n[i] |= ismissing(el)
+            end
```
bkamins, line 14: Changing these lines to:
```
 if eltype(v) >: Missing && length(n) > 0
        n .|= ismissing.(v)
    end
```
test and moving it oudide might make sense in general. What do you think?

src/groupeddataframe/utils.jl
```diff
@@ -5,11 +5,13 @@ function hashrows_col!(h::Vector{UInt},
                        v::AbstractVector{T},
                        rp::Nothing,
                        firstcol::Bool) where T
-    @inbounds for i in eachindex(h)
-        el = v[i]
-        h[i] = hash(el, h[i])
-        if length(n) > 0
```
bkamins, line 11: Can you remind me why did we do `length(n) > 0` check here? (is it that `length(n) > 0` if `skipmissing` is used?)

src/groupeddataframe/utils.jl
bkamins: Also I have confirmed that only this line causes problem, ie. `n[i] |= ismissing(el)` does not get executed properly. I will report the details.

bkamins: it happens if we work with `CategoricalVector`.

src/groupeddataframe/utils.jl
nalimilan: Yes, we pass an empty vector when `skipmissing=false`.

src/other/utils.jl
```diff
@@ -129,6 +129,61 @@ function tforeach(f, x::AbstractArray; basesize::Integer)
...
+        lidx = iter.args[1]
+        range = iter.args[2]
+        quote
+            let x = $(esc(range))
+                for i in eachindex(x)
+                    local $(esc(lidx)) = @inbounds x[i]
+                    $(esc(lbody))
+                end
+            end
+            return nothing
```
bkamins, line 172:
```suggestion
            nothing
```

src/other/utils.jl
```diff
@@ -129,6 +129,61 @@ function tforeach(f, x::AbstractArray; basesize::Integer)
...
+                             for p in split_indices(len, basesize)]
+                    foreach(wait, tasks)
+                else
+                    for i in eachindex(x)
+                        local $(esc(lidx)) = @inbounds x[i]
+                        $(esc(lbody))
+                    end
+                end
+            end
+            return nothing
```
bkamins, line 158:
```suggestion
            nothing
```

nalimilan:
I confirm that benchmarks give similar results with the fix.

nalimilan:
CI passed on the previous commit except docs, I've disabled it for the latest commit to avoid wasting resources but docs build fine locally.

src/groupeddataframe/utils.jl
bkamins: Let us keep it as is for now - it is not likely a big difference anyway

bkamins (approved):

bkamins:
Looks good now. I leave up to you if you want to investigate https://github.com/JuliaData/DataFrames.jl/pull/2685/files#r611769396 still or merge it as is.

nalimilan:
> Looks good now. I leave up to you if you want to investigate https://github.com/JuliaData/DataFrames.jl/pull/2685/files#r611769396 still or merge it as is.

Given that the current approach allows running both steps in parallel I'd say it's OK (and I had benchmarked it when it was introduced IIRC).

PR #2687, opened 2021-03-29
Use fast path for single column aggs

sa-:
PR is still a draft because there are no tests yet

Using this script I was able to verify that it works much faster now

```
using DataFrames
##
df = DataFrame(a = rand(1:100, 10^8), b=rand(10^8));

##
gdf = groupby(df, :a);

##
@time combine(gdf, :a => sum);
@time combine(gdf, [:a] => sum);
```

Output: 
```
  0.133490 seconds (173 allocations: 15.250 KiB)
  0.133525 seconds (184 allocations: 15.781 KiB)
```


src/abstractdataframe/selection.jl
```diff
@@ -254,17 +257,28 @@ function normalize_selection(idx::AbstractIndex,
                 end
             end
     end
+
     if lls isa AbstractString
-        r = Symbol(lls)
+        combine_target_col = Symbol(lls)
     elseif lls isa AbstractVector{<:AbstractString}
-        r = Symbol.(lls)
+        combine_target_col = Symbol.(lls)
```
bkamins, line 264: when you add tests can you please add a test that will cover this line? It is unrelated to the PR, but since you change the line we should cover it with tests.

src/abstractdataframe/selection.jl
```diff
@@ -254,17 +257,28 @@ function normalize_selection(idx::AbstractIndex,
...
+    if combine_target_col isa AbstractVector{Symbol}
+        allunique(combine_target_col) || throw(ArgumentError("target column names must be unique"))
     end
-    return (wanttable ? AsTable(c) : c) => first(last(sel)) => r
+
+    combine_src = if wanttable 
+        AsTable(c) 
+    else 
+        (length(c) == 1 ? only(c) : c)
+    end
```
bkamins, line 277:
```suggestion
    combine_src = wanttable ? AsTable(c) : (length(c) == 1 ? only(c) : c)
```

src/abstractdataframe/selection.jl
```diff
@@ -323,7 +337,14 @@ function normalize_selection(idx::AbstractIndex,
             newcol = Symbol(prefix)
         end
     end
-    return (wanttable ? AsTable(c) : c) => fun => newcol
+
+    combine_src = if wanttable 
+        AsTable(c) 
+    else 
+        (length(c) == 1 ? only(c) : c)
+    end
```
bkamins, line 345:
```suggestion
    combine_src = wanttable ? AsTable(c) : (length(c) == 1 ? only(c) : c)
```

bkamins:
Looks good apart from small comments. Just please make sure to add tests as you commented. Thank you!

src/abstractdataframe/selection.jl
sa-: In terms of readability, I think clarity is more important than conciseness here.

src/abstractdataframe/selection.jl
sa-: Sure

src/abstractdataframe/selection.jl
bkamins: In my opinion your version is less readable (of course this is subjective). Also normally users do not expect `if-else` to return a value (it does in Julia, but not e.g. in Python). If you insist on your version then I would do:
```
    if wanttable 
        combine_src = AsTable(c) 
    else 
        combine_src = (length(c) == 1 ? only(c) : c)
    end
```

bkamins:
The PR looks good. If you could just add tests to `test/select.jl` as discussed on Slack I think we can finalize this (in particular to cover the line that is not covered currently). Thank you!

src/abstractdataframe/selection.jl
sa-: Added

bkamins (approved):

bkamins:
@nalimilan - can you please have a quick look on this before merging? Thank you!

test/select.jl
```diff
@@ -1608,4 +1608,25 @@ end
     @test_throws ArgumentError select(df, :a => (x -> 1) => :a1, :a => (x -> [2]) => :a2, [:a])
 end
 
+@testset "normalize_selection" begin
+    @test DataFrames.normalize_selection(
+            DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]),
+            [:a] => sum,
+            false) ==
+        (1 => (sum => :a))
```
nalimilan, line 1616: We don't use this style in DataFrames: rather keep the first argument on the first line, and align other arguments to it.

nalimilan (commented):
I'm not familiar with that part of the code, so if that's OK for you...

test/select.jl
```diff
@@ -1608,4 +1608,23 @@ end
...
```
bkamins, line 1616:
```suggestion
            DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]),
            [:a] => sum, false) == (1 => (sum => :a))
```

test/select.jl
```diff
@@ -1608,4 +1608,23 @@ end
...
+
+    @test DataFrames.normalize_selection(DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]), 
+                                         [:a] => sum => [:new], 
+                                         false) ==
+        (1 => (sum => [:new]))
```
bkamins, line 1621:
```suggestion
    @test DataFrames.normalize_selection(DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]), 
                                         [:a] => sum => [:new],  false) == (1 => (sum => [:new]))
```

test/select.jl
```diff
@@ -1608,4 +1608,23 @@ end
...
+
+    # Test that target col Strings are converted to symbols
+    @test DataFrames.normalize_selection(DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]), 
+                                         [:a] => sum => ["new"], 
+                                         false) ==
+        (1 => (sum => [:new]))
```
bkamins, line 1627:
```suggestion
    @test DataFrames.normalize_selection(DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]), 
                                         [:a] => sum => ["new"],  false) == (1 => (sum => [:new]))
```

test/select.jl
bkamins, line 1616:
```suggestion
    @test DataFrames.normalize_selection(DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]),
                                         [:a] => sum, false) == (1 => (sum => :a))
```

test/select.jl
```diff
@@ -1608,4 +1608,16 @@ end
...
 end
 
+@testset "normalize_selection" begin
+    @test DataFrames.normalize_selection(DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]),
+                                         [:a] => sum, false) == (1 => (sum => :a))
+
+    @test DataFrames.normalize_selection(DataFrames.Index(Dict(:a => 1, :b => 2), [:a, :b]), 
+                                         [:a] => sum => [:new],  false) == (1 => (sum => [:new]))
+
+    # Test that target col Strings are converted to symbols
```
bkamins, line 1618:
```suggestion
    # Test that target col strings are converted to Symbols
```

bkamins:
Thank you!

PR #2690, opened 2021-03-30
check for data frame corruption in delete!

bkamins:
check for data frame corruption in delete!

### DIFF
--- a/src/dataframe/dataframe.jl
+++ b/src/dataframe/dataframe.jl
@@ -966,9 +966,18 @@ function Base.delete!(df::DataFrame, inds)
     if !isempty(inds) && size(df, 2) == 0
         throw(BoundsError(df, (inds, :)))
     end
+    
     # we require ind to be stored and unique like in Base
     # otherwise an error will be thrown and the data frame will get corrupted
-    foreach(col -> deleteat!(col, inds), _columns(df))
+    n = nrow(df)
+    for col in _columns(df)
+        if length(col) != n
+            throw(ArgumentError("Unequal lengths of columns detected. Most likely the " *
+                                "same column was stored in the data frame several times. 
+                                 The source data frame is now corrupted."))
+        end
+        deleteat!(col, inds)
+    end
     return df
 end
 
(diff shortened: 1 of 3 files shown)


src/dataframe/dataframe.jl
```diff
@@ -966,9 +966,18 @@ function Base.delete!(df::DataFrame, inds)
...
+    
     # we require ind to be stored and unique like in Base
     # otherwise an error will be thrown and the data frame will get corrupted
-    foreach(col -> deleteat!(col, inds), _columns(df))
+    n = nrow(df)
+    for col in _columns(df)
+        if length(col) != n
+            throw(ArgumentError("Unequal lengths of columns detected. Most likely the " *
+                                "same column was stored in the data frame several times. 
+                                 The source data frame is now corrupted."))
```
nalimilan, line 977: Let's continue the user-friendliness contest! :-)
```suggestion
    for i in 1:ncol(df)
        col = _columns(df)[i]
        if length(col) != n
            throw(ArgumentError("Unequal lengths of columns detected at column :$(_names(df)[i]). " *
                                "Most likely the same column was stored in the data frame " *
                                "several times. The source data frame is now corrupted."))
```
The next step is to check which other column is `===` to the current one, and print that too. Oh, and the next one is to skip this column but continue resizing other columns so that the result is *not* corrupt.

bkamins: OK - no rush to merge it. I will make these improvements

bkamins: I have added it :).

src/dataframe/dataframe.jl
```diff
@@ -966,23 +966,46 @@ function Base.delete!(df::DataFrame, inds)
...
+    for i in 2:length(cols)
+        col = cols[i]
+        if length(col) == n
+            deleteat!(col, drop)
+        end
+    end
+
+    for i in 1:length(cols)
+        # this should never happen, but we add it for safety
+        @assert length(cols[i]) == newn corrupt_msg(df, i)
```
bkamins, line 1003: can you think of a scenario when this could be triggered (I cannot unless there is some bug in `deleteat!` for some special vector type, but who knows)?

bkamins:
CI fails for unrelated reasons

src/dataframe/dataframe.jl
nalimilan: That would be a serious bug in a custom array implementation. Probably not worth checking here as it's not particular to data frames.

bkamins: But it is very cheap in comparison to all other things we do. We have found many serious bugs even in Julia Base when developing DataFrames.jl. Do you think it is a problem to have this check? (as you know I prefer to have an overly safe code)

nalimilan: If you're willing to write tests that cover these lines, no problem. :-p

bkamins: This is easy:
```
    df = DataFrame(a=1, b=2)
    push!(df.b, 3)
    @test_throws AssertionError delete!(df, 1)
```
I have added this test :smile:.

nalimilan (approved):

bkamins:
Thank you!

Issue #2693, opened 2021-03-30
describe(...,:eltype) takes forever to complete on categorical columns


pgagarinov:
`describe` takes forever to complete when run on categorical columns of Union{Missing, String} type

Let us create a dataframe of random repeated strings.

```
using DataFrames, Random
rand_addresses = [randstring(20) for k in 1:6500];
data_mat=(rand([rand_addresses;missing],500000,30));
df = DataFrame(data_mat);
@time describe(df,:eltype);
`>  0.274541 seconds (813 allocations: 150.049 MiB)`
```
`@time collect(skipmissing(df.x1));`
`> 0.008496 seconds (20 allocations: 5.001 MiB)`

now let us make all columns categorical

`categorical!(df);`

and now this takes forever!:

`@time collect(skipmissing(df.x1));`

and as well as this:

`@time describe(df,:eltype);`

Julia 1.6
DataFrames v0.22.6

pdeffebach:
Can you test whether `collect(skipmissing(x))` has different timing for categorical vs. non-categorical columns?

pgagarinov:
@pdeffebach `@time collect(skipmissing(df.x1));` takes forever if x1 is a categorical column in the example above. For plain columns there is no such problem. But why does `@describe` have to call `collect` when I just request the element type? Isn't the element type known in advance?

bkamins:
I think we should skip `collect` altogether.

pdeffebach:
I think we need collect because a lot of the functions used allocate anyway, like `median`, we also calculate `last`, which needs a `Vector`. Plus we allow people to  use their own functions, and want things to "just work". 

We only do this is there are missing values, which is the case in your example.

pdeffebach:
interesting! I can explore this in a PR.

pdeffebach:
Regardless, this is still likely a CategoricalArrays bug, right? `collect(skipmissing(x))` should not be exceptionally slow.

bkamins:
@pdeffebach - right, it has been fixed AFAICT

@pgagarinov 

Please update CategoricalArrays.jl and DataFrames.jl, currently you will see for your data:
```
julia> @time describe(df, :eltype);
  0.162089 seconds (813 allocations: 150.049 MiB, 4.00% gc time)

julia> @time describe(df, :eltype);
  0.136962 seconds (813 allocations: 150.049 MiB, 3.13% gc time)

julia> @time collect(skipmissing(df.x1));
[...]

PR #2694, opened 2021-03-30
do not use collect in describe

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2693

@pdeffebach - some tests (correctness and performance) would be welcome if you can spare some time.

In my opinion this change should be OK, as rather the called functions should add support for `skipmissing` value passed - and I would rather fix this limitation if possible.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -522,9 +522,8 @@ column `:nmissing` will report the number of missing values of that variable.
 
 If custom functions are provided, they are called repeatedly with the vector
 corresponding to each column as the only argument. For columns allowing for
-missing values, the vector is wrapped in a call to `skipmissing`: custom
-functions must therefore support such objects (and not only vectors), and cannot
-access missing values.
+missing values, the vector is wrapped in a call to `skipmissing` and then
+`collect`ed: therefore custom functions cannot access missing values.
 
 # Examples
 ```julia
@@ -610,7 +609,7 @@ function _describe(df::AbstractDataFrame, stats::AbstractVector)
     # An array of Dicts for summary statistics
     col_stats_dicts = map(eachcol(df)) do col
         if eltype(col) >: Missing
-            t = skipmissing(col)
+            t = collect(skipmissing(col))
             d = get_stats(t, predefined_funs)
             get_stats!(d, t, custom_funs)
         else
@@ -649,7 +648,7 @@ end
 # Compute summary statistics
 # use a dict because we dont know which measures the user wants
 # Outside of the `describe` function due to something with 0.7
-function get_stats(col::Union{AbstractVector,Base.SkipMissing}, stats::AbstractVector{Symbol})
+function get_stats(col::AbstractVector, stats::AbstractVector{Symbol})
     d = Dict{Symbol, Any}()
 
     if :q25 in stats || :median in stats || :q75 in stats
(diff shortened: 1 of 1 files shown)


nalimilan:
> `median` will call `collect` anyway (it does it always).

Yeah actually we could call `median!` if we do that after calling all other functions. That would avoid a copy.

bkamins:
if we decided to keep `collect`. In this PR I have removed `collect`.

pdeffebach:
I see that there is this warning in `describe`:

> If custom functions are provided, they are called repeatedly with the
  vector corresponding to each column as the only argument. For columns
  allowing for missing values, the vector is wrapped in a call to
  skipmissing: custom functions must therefore support such objects (and
  not only vectors), and cannot access missing values.

To be honest I'm not sure people will read this or follow it. Given there isn't a huge performance improvement, I would err on the side of safety and just use `collect` and get rid of that note.

bkamins:
OK - @nalimilan - if you agree to add custom `collect` for `skipmissing` in CategoricalArrays.jl I would just update the docstring in this PR and retain `collect`.

bkamins:
I have reverted the commit and just updated the docstring.

pdeffebach (approved):

nalimilan:
Sorry, actually I wonder whether the best solution would be to call `collect` only internally as an optimization when we compute the median (via `median!`), but always pass the `SkipMissing` iterators to custom functions. That way we would still be able to avoid making copies at all when we don't compute the median or quantiles. Otherwise we would be locked into a suboptimal implementation in terms of performance. What do you think?

Here's a simple benchmark:
```julia
julia> df = DataFrame(rand([1:10; missing], 500000, 30), :auto);

# With collect
julia> @btime describe(df, :mean, :min, :max, :nmissing, :eltype);
  210.184 ms (896 allocations: 150.05 MiB)

# Without collect
julia> @btime describe(df, :mean, :min, :max, :nmissing, :eltype);
  76.155 ms (356 allocations: 30.94 KiB)
```

bkamins:
Let us do the following, we do `collect` if we:
1. calculate median or quantiles
2. user passes a custom function (which, as @pdeffebach might not support `SkipMissings`)

(i.e. when we use only internal functions that do not need to collect we would not collect)

I will do an update to the PR.

nalimilan:
Though that would mean that users cannot benefit from maximum performance if they pass custom functions. Do we really expect that calling `collect` manually would be annoying for users? They should be used to handling `skipmissing` as it's basically needed everywhere.

bkamins:
The tension is between:
1. calling `collect` once and then all functions can use `collect`ed value which is faster and collects only once
2. not `calling` collect which means that functions that do not need collect can allocate less memory

Let us wait for @pdeffebach to comment what he things as an end user :smile:.

bkamins:
@pdeffebach - could you comment what you feel is preferable here given the discussion we had above? Thank you!

pdeffebach:
Oh, sorry I missed this!

I'm fine with not using collect, I guess. I would hate for very large datasets to be super laggy at `describe`. 

I do think that the issue with custom functions is important. People should be able to do other things than just with `skipmissing`. Here's what I propose

1. Don't use `collect` in general. Remove `median` from the default output and only `collect` unless `median` or `:q25` etc. are called. 
2.  When people pass custom functions, pass the whole vector. The user can deal with `missing`s on their own.

pdeffebach:
Sorry for bikeshedding with a new proposal. I guess, I don't have a super strong opinion on this. I think reporting the median is a good idea, but understand the appeal of not allocating. 

As for the custom functions, I'm fine with requiring that they work with a `SkipMissing`. You are right, Milan, that they should know how to work with this.

bkamins:
OK - so I have just removed `collect` (as `median` will collect it automatically anyway), reverted the docstring (as now it matches what we do) and despecialized some methods to avoid inference time hit of the change we do.

nalimilan (approved):

bkamins:
Thank you!

PR #2697, opened 2021-04-02
Tutorial

RohitRathore1:


docs/src/assets/german_credit.csv
```diff
@@ -0,0 +1,1001 @@
+,Age,Sex,Job,Housing,Saving accounts,Checking account,Credit amount,Duration,Purpose
```
bkamins, line 1: why rowid column does not have a name?

docs/src/assets/readme.txt
```diff
@@ -0,0 +1,34 @@
+# Source
```
bkamins, line 1: if you use markdown then better name it readme.md (this is even better)

docs/src/assets/german_credit.csv
RohitRathore1: Oh, yeah. Fixing it. In the kaggle they had mentioned using #.

docs/src/assets/readme.txt
RohitRathore1: Yeah, fixing it.

bkamins: I think it is enough to do a list of files we provide in assets. And for each file add information about the original source, the source from which we have taken the file and a link to the license we use (like Kaggle - they just give a link to the license)

docs/src/assets/readme1.txt
```diff
@@ -0,0 +1,487 @@
+# Source
```
bkamins, line 1: as per comment above then one file would be enough

docs/src/assets/readme.md
```diff
@@ -0,0 +1,439 @@
+# Source
```
bkamins, line 1: As commented - I would avoid adding this much detail.

Just write something like

```
# Introduction

In this folder we store the following data sets:

* german_credit.csv
* iris.csv

# German Credit data set

## License:
link

## Source:

both info that the original data is from UCI, and that this data is from Kaggle (as this information is missing)
```

If you feel uncomfortable with something just leave an empty space and I will fill it

RohitRathore1: Oh okay, i had not read your above comment.

docs/src/assets/readme.md
```diff
@@ -0,0 +1,25 @@
...
+Professor Dr. Hans Hofmann
+Institut f"ur Statistik und "Okonometrie
+Universit"at Hamburg
+FB Wirtschaftswissenschaften
+Von-Melle-Park 5
+2000 Hamburg 13
+
+The original data is from [UCI](https://archive.ics.uci.edu/ml/datasets/statlog+(german+credit+data)), and that this data is from [Kaggle](https://www.kaggle.com/uciml/german-credit)
+
+
```
bkamins, line 24:
```suggestion
# Iris data set

## License
https://creativecommons.org/publicdomain/zero/1.0/

## Source:
https://archive.ics.uci.edu/ml/datasets/Iris
Creator:
R.A. Fisher
```

bkamins:
Given you have opened https://github.com/JuliaData/DataFrames.jl/pull/2700 please make sure to correctly handle rebasing of this PR before submitting it for a review (as I assume in this PR you will add tutorial sections and in https://github.com/JuliaData/DataFrames.jl/pull/2700 you propose to change the legacy structure of the tutorial).

docs/src/assets/readme.md
bkamins: in the future you can just clik "Commit suggestion" (or edit it before accepting and then commit)

bkamins:
Here I meant rebasing because I assumed that most likely this PR would have conflicts with other changes you make. Therefore it is better to open it again against latest `main`. I assume this is what you are planning to do. Right?

Issue #2623, opened 2021-02-16
Promote Documentation on Working with Data Frames


jessebett:
The [Working with Data Frames](https://dataframes.juliadata.org/v0.17.0/man/getting_started.html#Working-with-Data-Frames-1) sub-section of the documentation contains important sub-sub-sections.

However, these are very difficult to discover because the left menu navigation bar does not display down to the sub-sub-section. 

Consider removing the Working with Data Frames section and promoting the sub sections so they can appear in the menu.

bkamins:
Thank you for reporting this. I will announce today on Slack a call for proposal to re-organize DataFrames.jl manual.

bkamins:
https://discourse.julialang.org/t/call-for-proposals-dataframes-jl-documentation/55418

PR #2700, opened 2021-04-03
promoted documentation on working with DataFrames

RohitRathore1:
It resolves issue #2623 where our aim was to promote documentation on working with DataFrames as the sub-sub section of working with DataFrames was not appearing in the left menu navigation.


bkamins:
OK - we can make this change (though this part of the manual is likely to get replaced soon by the new manual - right?).

If we go for it then also the sections on "Replacing data" and "Importing and Exporting Data (I/O)" should be moved to this separate file (as now you have extracted one section in the middle of the manual and applying this change would change the order in which the user would read it).

RohitRathore1:
> OK - we can make this change (though this part of the manual is likely to get replaced soon by the new manual - right?).
> Yeah, but for now it was solving the current open issue. 
> If we go for it then also the sections on "Replacing data" and "Importing and Exporting Data (I/O)" should be moved to this separate file (as now you have extracted one section in the middle of the manual and applying this change would change the order in which the user would read it).
Hmm, i am adding some commits.

RohitRathore1:
Because of typo error in file name building was failing. So fixed it

bkamins:
> Because of typo error in file name building was failing. So fixed it

As commented on Slack - please make sure that before you push a PR to GitHub you are able to build the documentation locally and all looks good.

RohitRathore1:
> > Because of typo error in file name building was failing. So fixed it
> 
> As commented on Slack - please make sure that before you push a PR to GitHub you are able to build the documentation locally and all looks good.

yes, I am following this now.

docs/make.jl
```diff
@@ -22,6 +22,9 @@ makedocs(
         "Introduction" => "index.md",
         "User Guide" => Any[
             "Getting Started" => "man/getting_started.md",
+            "Working with DataFrames" => "man/Working_with_DataFrames.md",
+            "Replacing Data" => "man/replacing_data.md",
```
bkamins, line 26: do you feel (as a user) it is worth to promote "Replacing Data" to top level here? Or it is better to have it as a section in "Working with DataFrames"?

RohitRathore1: I don't think so because when a user will review 'Working with DataFrames' then after completing that section he can directly jump to the 'Replacing Data' section in the series. But both things will be fine(show it as a section in 'working with DataFrames' or promote it to the top level).

bkamins: I would then include "Replacing Data" in "Working with DataFrames"

bkamins (approved):

docs/make.jl
```diff
@@ -22,6 +22,8 @@ makedocs(
         "Introduction" => "index.md",
         "User Guide" => Any[
             "Getting Started" => "man/getting_started.md",
+            "Working with DataFrames" => "man/Working_with_DataFrames.md",
```
bkamins, line 25: before merging please do not use capital letters in file names for consistency.

RohitRathore1: Resolved.

bkamins:
Thank you!

Issue #2564, opened 2020-11-29
Weird slowdowns when grouping on PooledArray


nalimilan:
It seems `row_group_slots` for arrays implementing `DataAPI.refpool` is much slower than it should. Profiling reveals that most of the time is spent on the `groups[i] = j` line. Yet, calling `row_group_slots` directly is twice faster than using `groupby`. The difference seems to be due to the allocation of the `groups` vector, but allocating the array separately is very fast. Not sure whether that's a benchmarking artifact.
```julia
julia> using DataFrames, PooledArrays, BenchmarkTools

julia> N = 10_000_000;

julia> df = DataFrame(x=PooledArray(rand(1:10, N)), y=rand(N));

julia> @btime groupby(df, :x);
  43.736 ms (40 allocations: 76.30 MiB)

julia> @btime DataFrames.row_group_slots($((df.x,)), Val(false), Vector{Int}(undef, nrow(df)), false, false);
  43.233 ms (17 allocations: 76.30 MiB)

julia> @btime DataFrames.row_group_slots($((df.x,)), Val(false), $(Vector{Int}(undef, nrow(df))), false, false);
  16.615 ms (14 allocations: 1.07 KiB)

julia> @btime Vector{Int}(undef, nrow(df));
  2.784 μs (3 allocations: 76.29 MiB)

# Slow (!)
julia> f(df) = DataFrames.row_group_slots((df.x,), Val(false), Vector{Int}(undef, nrow(df)), false, false)
f (generic function with 1 method)

julia> @btime f(df);
  44.118 ms (20 allocations: 76.30 MiB)
```

 This is not new as it also happens both on master and on 0.21.8, and both with Julia 1.5 and 1.6.

bkamins:
I think it is a benchmarking artifact.

A small thing is that we could replace `DataAPI.refpool.(cols)` with `map(DataAPI.refpool, col)` as it should reduce the compilation time. But this is minor. Also as @timholy suggests - this function can be probably split up into several smaller functions depending on `sort` and `skipmissing` arguments.

Finally - and I believe this is a major potential gain: if `sort=false`, `skipmissing=false` and a single column is passed when `refpool <: AbstractVector` then we can just check `allunique(only(refpools))` and if all levels are used (most of the time they will be used in practice) and then have a fast path function.

[...]

PR #2708, opened 2021-04-10
Fix grouping aggregation slowdown

bkamins:
This is related to https://github.com/JuliaData/DataFrames.jl/issues/2564#issuecomment-809177160 problem. I propose to produce a `Vector` always when `add_sum` or `mul_prod` is used. The slowdown was due to the fact that output vector was `similar` to input vector, so if input were `PooledArray` a big slowdown was observed.

### DIFF
--- a/src/groupeddataframe/fastaggregates.jl
+++ b/src/groupeddataframe/fastaggregates.jl
@@ -102,7 +102,7 @@ function groupreduce_init(op, condf, adjust,
         end
         # here we are sure that only Base.add_sum or Base.mul_prod are performed
         # so we always fall back to Vector as output column type
-        v = Vector{V}(undef, length(gd))
+        v = Tables.allocatecolumn(V, length(gd))
         fill!(v, x)
         return v
     else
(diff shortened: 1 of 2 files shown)


src/groupeddataframe/fastaggregates.jl
```diff
@@ -100,7 +100,9 @@ function groupreduce_init(op, condf, adjust,
         else
             V = U >: Missing ? Union{typeof(x), Missing} : typeof(x)
         end
-        v = similar(incol, V, length(gd))
+        # here we are sure that only Base.add_sum or Base.mul_prod are performed
+        # so we always fall back to Vector as output column type
+        v = Vector{V}(undef, length(gd))
```
nalimilan, line 105: Maybe use `Tables.allocatecolumn`? In practice I don't know of types where this would make a difference, but that sounds more consistent.

bkamins: OK

nalimilan (approved):

PR #2710, opened 2021-04-11
Fix default copycols for Tables.jl constructor

quinnj:


bkamins (approved):

bkamins:
Looks good. Can you add a test please? Thank you!

bkamins:
@quinnj - apart from adding tests for new functionality also old tests need updating as we tested against the old behavior.

quinnj:
Alright, fixed an issue w/ dispatching to `fromcolumns` for `CopiedColumns` and added an additional test that actually uses the internal `CopiedColumns` code path; we had a test when `CopiedColumns` was passed directly to `DataFrame` constructor, but, as is the case for `Arrow.Table` and `CSV.File`, they return `CopiedColumns` for the result of `Tables.columns`. I think this is in good shape now and will merge once CI passes.

PR #2709, opened 2021-04-11
Despecialization part 2

bkamins:
I changed `grouped_reduce!` by extracting out the part of code that does have less arguments (this is a relatively clean change and gives an improvement in compilation - but only a small one).

Further despecialization of `combine_with_first!` and `combine_rows_with_first!` (but not `combine_tables_with_first!` as it contains a hot loop). This second change is heavy handed, but gives the following benefits:

this PR
```
julia> using DataFrames

julia> df = DataFrame(a=[1,2],b=[3,4]);

julia> gdf = groupby(df, :a);

julia> @time combine(gdf, :a => x -> (b=1,) => AsTable);
  2.130731 seconds (6.10 M allocations: 353.595 MiB, 4.26% gc time, 99.98% compilation time)

julia> @time combine(gdf, :a => x -> (b=1,) => AsTable);
  0.101837 seconds (106.97 k allocations: 6.511 MiB, 99.33% compilation time)

julia> @time combine(gdf, :a => x -> (c=1,) => AsTable);
  0.229507 seconds (311.18 k allocations: 19.438 MiB, 99.48% compilation time)

julia> @time combine(gdf, :a => x -> (c=1,) => AsTable);
  0.115103 seconds (106.96 k allocations: 6.512 MiB, 10.16% gc time, 99.41% compilation time)

julia> @time combine(gdf, [:a, :b] => (x,y) -> (b=1, d=1) => AsTable);
  0.566155 seconds (988.28 k allocations: 60.440 MiB, 1.70% gc time, 99.65% compilation time)

julia> @time combine(gdf, [:a, :b] => (x,y) -> (b=1, d=1) => AsTable);
  0.097503 seconds (97.75 k allocations: 5.922 MiB, 99.25% compilation time)

julia> @time combine(gdf, [:a, :b] => (x,y) -> (c=1, d=1) => AsTable);
  0.260902 seconds (302.07 k allocations: 18.859 MiB, 9.65% gc time, 99.52% compilation time)

julia> @time combine(gdf, [:a, :b] => (x,y) -> (c=1, d=1) => AsTable);
  0.094826 seconds (97.75 k allocations: 5.920 MiB, 99.21% compilation time)
```

main
```
julia> using DataFrames

julia> df = DataFrame(a=[1,2],b=[3,4]);

julia> gdf = groupby(df, :a);

julia> @time combine(gdf, :a => x -> (b=1,) => AsTable);
[...]

PR #2716, opened 2021-04-12
Replace `tforeach` with `@spawn_for_chunks`

nalimilan:
It allows keeping the syntax even closer to a normal loop,
and it avoids duplicating functionality. Also simplify `@inbounds`
uses since it work when put in front of `@spawn_for_chunks`.

I've checked that performance is the same.


bkamins (approved):

bkamins:
Looks good. I am OK to merge this when the CI passes. After you merge this PR and #2709 is merged I will generate precompile statements.

PR #2721, opened 2021-04-15
make renaming perform copy in transform and transform!

bkamins:
in `transform!(df, :a => :b)` we want to make sure that `:b` column in the resulting data frame is not an alias of `:a` column as this can lead to hard-to-catch bugs.


src/abstractdataframe/selection.jl
```diff
@@ -656,8 +658,17 @@ $TRANSFORMATION_COMMON_RULES
 
 See [`select`](@ref) for examples.
 """
-transform!(df::DataFrame, @nospecialize(args...); renamecols::Bool=true) =
-    select!(df, :, args..., renamecols=renamecols)
+function transform!(df::DataFrame, @nospecialize(args...); renamecols::Bool=true)
+    idx = index(df)
+    newargs = Any[if sel isa Pair{<:ColumnIndex, Symbol}
```
bkamins, line 663: I was not sure about indentation here and below.

test/select.jl
```diff
@@ -1620,4 +1620,26 @@ end
...
+    @test df2.b !== df2.a
+    @test df2.b !== df.a
+    @test df2.a === df.a
+
+    a = df.a
+    transform!(df, :a => :b)
+    @test df.b == df.a == a
+    @test df.b !== df.a
+    @test df.b !== a
+    @test df.a === a
```
bkamins, line 1642:
```suggestion
    for oldcol in (:a, "a", 1), newcol in (:b, "b")
        df = DataFrame(a=1)
        df2 = transform(df, oldcol => newcol)
        @test df2.b == df2.a == df.a
        @test df2.b !== df2.a
        @test df2.b !== df.a
        @test df2.a !== df.a

        df2 = transform(df, oldcol => newcol, copycols=false)
        @test df2.b == df2.a == df.a
        @test df2.b !== df2.a
        @test df2.b !== df.a
        @test df2.a === df.a

        a = df.a
        transform!(df, oldcol => newcol)
        @test df.b == df.a == a
        @test df.b !== df.a
        @test df.b !== a
        @test df.a === a
    end
```

test/select.jl
```diff
@@ -1086,7 +1086,9 @@ end
     x2 = df.x2
     transform!(df, [:x1, :x2] => +, :x2 => :x3)
     @test df == df2
-    @test x2 === df.x2 === df.x3
+    @test x2 == df.x2 == df.x3
```
bkamins, line 1089: @nalimilan: as usual - we already were aware of this decision and had a test against the past behavior :).

src/abstractdataframe/selection.jl
bkamins: ok - aligned this way.

src/abstractdataframe/selection.jl
```diff
@@ -978,8 +989,17 @@ ERROR: ArgumentError: column :x in returned data frame is not equal to grouping
...
-    select(df, :, args..., copycols=copycols, renamecols=renamecols)
+function transform(df::AbstractDataFrame, @nospecialize(args...); copycols::Bool=true, renamecols::Bool=true)
+    idx = index(df)
+    newargs = Any[if sel isa Pair{<:ColumnIndex, Symbol}
+            idx[first(sel)] => copy => last(sel)
+        elseif sel isa Pair{<:ColumnIndex, <:AbstractString}
+            idx[first(sel)] => copy => Symbol(last(sel))
+        else
+            sel
+        end for sel in args]
```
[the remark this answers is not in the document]
bkamins: This is needed for `copycols=false`. For `copycols=true` it is indeed not required as we have `_add_col_check_copy` logic for this, where we check:
```
if copycols && !(fun isa ByRow) && (v isa SubArray || any(i -> vpar === parent(cdf[i]), parent_cols))
```
but this check will fail if we do the rewrite to `copy` so we will not copy twice.

The problem we want to solve for `copycols=false` is (this is before this PR):
```
julia> df = DataFrame(a=1)
df21×1 DataFrame
 Row │ a     
     │ Int64 
─────┼───────
   1 │     1

julia> df2 = transform(df, :a=>:b, copycols=false)
1×2 DataFrame
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      1

julia> df2.a === df2.b === df.a
true
```
Now you can assure yourself that we do not make copy twice like this. On main (after compilation):
```

julia> df = DataFrame(a=1:10^7);

julia> @allocated transform(df, :a => :b)
160005536
```
And the same in this PR:
```
julia> df = DataFrame(a=1:10^7);

julia> @allocated transform(df, :a => :b)
160007344
```
(the difference is because of the additional logic)

I could add an additional `if` and use the new logic only if `copycols=false`, and the old logic if `copycols=true`. What would you prefer?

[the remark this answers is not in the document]
nalimilan: Got it. Maybe just a comment would be useful.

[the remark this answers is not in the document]
bkamins: I have now pushed a comment. Also I have added a comment in NEWS.md that we will consider not-making a copy in similar situations a bug (if we identify them).

src/abstractdataframe/selection.jl
```diff
@@ -978,8 +989,20 @@ ERROR: ArgumentError: column :x in returned data frame is not equal to grouping
 
 See [`select`](@ref) for more examples.
 """
-transform(df::AbstractDataFrame, @nospecialize(args...); copycols::Bool=true, renamecols::Bool=true) =
-    select(df, :, args..., copycols=copycols, renamecols=renamecols)
+function transform(df::AbstractDataFrame, @nospecialize(args...); copycols::Bool=true, renamecols::Bool=true)
+    idx = index(df)
+    # when using the copy function the copy of source data frame
+    # is made exactly once even if copycols=true
+    # (copycols=true makes a copy only if the column were not copied previously)
```
nalimilan, line 996:
```suggestion
    # (copycols=true makes a copy only if the column was not copied previously)
```

nalimilan (approved):

bkamins:
Thank you!

Issue #2650, opened 2021-03-10
Add `matchmissing = :notequal` option


nilshg:
Consider two DataFrames:

```
julia> df = DataFrame(a = [1, 2, missing, missing])
4×1 DataFrame
 Row │ a       
     │ Int64?  
─────┼─────────
   1 │       1
   2 │       2
   3 │ missing 
   4 │ missing 

julia> df2 = DataFrame(a = [1, 3, missing, missing], b = rand(4))
4×2 DataFrame
 Row │ a        b        
     │ Int64?   Float64  
─────┼───────────────────
   1 │       1  0.459054
   2 │       3  0.649346
   3 │ missing  0.875563
   4 │ missing  0.709856
```

currently to join them I would need to set `matchmissing = :equal`, which produces duplicates:

```
julia> leftjoin(df, df2, on = :a, matchmissing = :equal)
6×2 DataFrame
 Row │ a        b
     │ Int64?   Float64?       
─────┼─────────────────────────
   1 │       1        0.459054
   2 │ missing        0.875563
   3 │ missing        0.875563
   4 │ missing        0.709856
   5 │ missing        0.709856
   6 │       2  missing     
```

I would like an option `matchmissing = :ignore` (or whatever other name) that preserves the left table exactly, and only adds information on the right side where non-missing values match. Currently I think this would be achieved via

```
julia> leftjoin(df, dropmissing(df2), on = :a, matchmissing = :equal)
4×2 DataFrame
 Row │ a        b
     │ Int64?   Float64?       
─────┼─────────────────────────
   1 │       1        0.459054
   2 │       2  missing        
   3 │ missing  missing        
   4 │ missing  missing       
```

which is a bit counterintuitive from an API perspective (I need to set `matchmissing` to `:equal` even though I don't want to match missings!), and might also suboptimal from an efficiency perspective.

bkamins:
I think we should add it.

pstorozenko:
I believe
`leftjoin(df, df2, on = :a, matchmissing = :ignore)`
should work like
`leftjoin(df, dropmissing(df2, :a), on = :a, matchmissing = :equal)`
instead of proposed
`leftjoin(df, dropmissing(df2), on = :a, matchmissing = :equal)`
right?
It seems not working properly for `df2` with missings in `b` column.

How about other types of join?
Rightjoin is symmetric to leftjoin.
With inner and outer I think we should drop missings on key in both dataframes.
[...]

PR #2724, opened 2021-04-17
Matchmissing == :notequal

pstorozenko:
Not documented, available for testing purpose.

As discussed in #2650 I create PR with my changes.


src/join/composer.jl
```diff
@@ -55,6 +56,42 @@ struct DataFrameJoiner
                                         "when matchmissing == :error"))
                 end
             end
+        elseif matchmissing === :notequal_copy
+            if kind in [:left, :semi, :anti]
```
bkamins, line 60: no need to allocate here, this is minor:
```suggestion
            if kind in (:left, :semi, :anti)
```

src/join/composer.jl
```diff
@@ -55,6 +56,42 @@ struct DataFrameJoiner
...
+                dfr = dropmissing(dfr, right_on)
+                dfr_on = select(dfr, right_on, copycols = false)
+            elseif kind === :right
+                dfl = dropmissing(dfl, left_on)
+                dfl_on = select(dfl, left_on, copycols = false)
+            elseif kind === :inner
+                @show kind, matchmissing
```
bkamins, line 67:
```suggestion
```

src/join/composer.jl
```diff
@@ -55,6 +56,42 @@ struct DataFrameJoiner
...
+                dfl = dropmissing(dfl, left_on)
+                dfl_on = select(dfl, left_on, copycols = false)
+                dfr = dropmissing(dfr, right_on)
+                dfr_on = select(dfr, right_on, copycols = false)
+            elseif kind === :outer 
+                throw(ArgumentError("matchmissing == :notequal with kind == :outer not allowed"))
+            else
+                throw("matchmissing == :notequal not implemented for kind == $kind")
```
bkamins, line 75:
```suggestion
                throw(ArgumentError("matchmissing == :notequal not implemented for kind == $kind"))
```

src/join/composer.jl
```diff
@@ -55,6 +56,42 @@ struct DataFrameJoiner
...
+                dfl = dropmissing(dfl, left_on)
+                dfl_on = select(dfl, left_on, copycols = false)
+            elseif kind === :inner
+                @show kind, matchmissing
+                dfl = dropmissing(dfl, left_on)
+                dfl_on = select(dfl, left_on, copycols = false)
+                dfr = dropmissing(dfr, right_on)
+                dfr_on = select(dfr, right_on, copycols = false)
+            elseif kind === :outer 
+                throw(ArgumentError("matchmissing == :notequal with kind == :outer not allowed"))
```
bkamins, line 73:
```suggestion
                throw(ArgumentError("matchmissing == :notequal for `outerjoin` is not allowed"))
```

src/join/composer.jl
```diff
@@ -55,6 +56,42 @@ struct DataFrameJoiner
...
+            else
+                throw("matchmissing == :notequal not implemented for kind == $kind")
+            end
+        elseif matchmissing === :notequal_view
+            if kind in [:left, :semi, :anti]
```
bkamins, line 78:
```suggestion
            if kind in (:left, :semi, :anti)
```

src/join/composer.jl
```diff
@@ -55,6 +56,42 @@ struct DataFrameJoiner
...
+                dfl_on = select(dfl, left_on)
+                dfr = dropmissing(dfr, right_on, view = true)
+                dfr_on = select(dfr, right_on)
+            elseif kind === :outer 
+                throw(ArgumentError("matchmissing == :notequal with kind == :outer not allowed"))
+            else
+                throw("matchmissing == :notequal not implemented for kind == $kind")
+            end
         elseif matchmissing !== :equal
             throw(ArgumentError("matchmissing allows only :error or :equal"))
```
bkamins, line 96: this error will need to be updated later

bkamins:
So things to do to improve the performance are the following (this is a general comment without running of the code):
1. in `completecases(df::AbstractDataFrame, col::Colon=:)` process only columns that potentially allow missing values
2. in `dropmissing` and `dropmissing!` do not invoke `completecases` if no consider columns potentially allow for `missing`
3. in `dropmissing` and `dropmissing!` if we invoked `completecases`
4. in `SubDataFrame(parent::DataFrame, rows::AbstractVector{Bool}, cols)` first check if `rows` creates one continuous block of rows; if yes then instead of `findall` use a range to avoid allocations (actually probably a custom `findall` should be written to do it in one shot)

Regarding memory profiling - I will run my test and report here later.

bkamins:
Regarding the performance of view vs copy the offending line is: https://github.com/JuliaData/DataFrames.jl/blob/a671a3b86537a253d461ff54b87ae2c26e4e8a04/src/subdataframe/subdataframe.jl#L171

When doing `getindex` in a `SubDataFrame` we allocate one temporary vector (both for left and right) that has the size comparable to a data frame. So we do "as if" we had 2 extra columns.

Therefore a tentative conclusion is:
1. in the `dropmissing` even if we use `view` we should check if actually any rows were dropped and if not pass through the original `DataFrame` (if it were a `DataFrame` already) - this will save some time.
2. we probably should branch conditional on the number of columns in a data frame (the best cut-off should be identified by benchmarking)
   * if number of columns is low use a copy
   * if number of columns is high use a view (as otherwise it is extremely easy to "Kill" the process due to out-of-memory)
3. an even more advanced optimization would be to split line https://github.com/JuliaData/DataFrames.jl/blob/a671a3b86537a253d461ff54b87ae2c26e4e8a04/src/join/composer.jl#L111 into two parts: `left_noon` and `left_on` (as `left_on` is always materialized) - but I am not sure if it is worth the effort

Also maybe we should just always use a `view` if benchmarking shows that the speed differences are not very big, as joins are very memory hungry, so in this case it is better not to do too much temporary allocations.

In summary:
1. in the first round I would go for the `view` option - even if we know it is slower - it is safer (or a mix of `copy` and `view` with low threshold for column count when we switch to `view`)
2. then in consecutive PRs work on improving the performance (and my experience is that these things are usually a deep hole where one optimization leads to further requests for changes, but this is how it works :))

bkamins:
Just to be clear on my position - I have "Killed" my Julia session several times when benchmarking with "copy" option, while "view" option worked each time.

bkamins:
@pstorozenko - please let me know on which things listed here you would be willing to work, so that I can plan better the development. Thank you!

Also in general after 1.0 release of DataFrames.jl there are four streams of potential things that maybe you would be interested in to contribute after we are done with this PR/series of PRs:
* performance (things like discussed in this PR + improving multi threading support); here - as commented - PRs should be small and focus on a single thing that is improved;
* adding new functionality (here the key discussions are usually API design considerations);
* documentation improvements;
* test coverage improvement.

pstorozenko:
I'd propose the following:

> 1. in `completecases(df::AbstractDataFrame, col::Colon=:)` process only columns that potentially allow missing values
> 2. in `dropmissing` and `dropmissing!` do not invoke `completecases` if no considered columns potentially allow for `missing`

I'll do PRs for those two.
Is there a nicer way of getting a vector of _missingable_ columns than `Missing .<: eltype.(eachcol(df))`? 

> 3. in `SubDataFrame(parent::DataFrame, rows::AbstractVector{Bool}, cols)` first check if `rows` creates one continuous block of rows; if yes then instead of `findall` use a range to avoid allocations (actually probably a custom `findall` should be written to do it in one shot)

Do you have any suggestions on naming a custom `findall` function and it's full signature? 
I'll to PR then.

> Therefore a tentative conclusion is: ...

In this PR I'll do some quick benchmarks on number of columns and then write solution with mixed view / copy approach for `:notequal` as well as write documentation and tests.

> an even more advanced optimization would be to split line

I think, we can benchmark it later, but I'll not do it now.

----------------

> Also in general after 1.0 release of DataFrames.jl there are four streams of potential things that maybe you would be interested in to contribute after we are done with this PR/series of PRs:

Thanks a lot! We will see later.

bkamins:
> `Missing .<: eltype.(eachcol(df))`

is OK, but do not do this this way, as it will add compilation latency without any benefit. better just iterate columns of `df` and compare their `eltype` to `Missing`.

> Do you have any suggestions on naming a custom `findall` function and it's full signature?

maybe just `_findall(v::AbstractVector{Bool}))`

> In this PR I'll do some quick benchmarks on number of columns and then write solution with mixed view / copy approach

OK, just please keep in mind that I prefer a solution that is not memory hungry in general. Also note when benchmarking that your original benchmark was kind of "worst case" as the join produced many more rows than the input tables because of the way how you generated data. This is not a typical scenario (normally one expects no dupilcates or at most few duplicates). 

> I think, we can benchmark it later, but I'll not do it now.

Yes - this is something that is probably not going to give much benefit. The point is to avoid materializing `left_on` twice (as we materialize it anyway earlier)

Thank you!

pstorozenko:
> 2. in dropmissing and dropmissing! do not invoke completecases if no considered columns potentially allow for missing

With #2726 optimization to `completecases` I don't see a point in writing additional ifs in `dropmissing` as it will only obscure code and not result in any benefits, don't you think?

bkamins:
> it will only obscure code and not result in any benefits, don't you think?

The corner case is when there are no `missing`s. In this case you probably want to avoid allocations and just make a view with `:` as row selector.

---

Also as a general comment - all my proposals are speculative, i.e. I have not implemented and benchmarked them. We will make these changes if we can see the benefit.

bkamins:
Given https://github.com/JuliaData/DataFrames.jl/pull/2727 is merged could you please review what should be done in this PR? Thank you!

pstorozenko:
Yes, I'll look on it today.

pstorozenko:
I've tested versions in a more realistic scenario as you suggested.
I took `Posts` and `Votes` from stackoverflow dataset, _missified_ p percent of values in merge column and innerjoined frames.

Version with view is faster here and allocates less memory, so let's leave this version.

```julia
posts = GZip.open("match/Posts.csv.gz") do f
    CSV.read(f, DataFrame)
end

votes = GZip.open("match/Votes.csv.gz") do f
    CSV.read(f, DataFrame)
end


function testp(posts, votes, p)
    Nv = nrow(votes)
    v2 = copy(votes)
    allowmissing!(v2, :PostId)
    iv = sample(1:Nv, trunc(Int, Nv * p), replace=false)
    v2[iv, :PostId] .= missing

    Np = nrow(posts)
    p2 = copy(posts)
    allowmissing!(p2, :Id)
    ip = sample(1:Np, trunc(Int, Np * p), replace=false)
    p2[ip, :Id] .= missing

    @btime innerjoin($p2, $v2, on = :Id => :PostId, makeunique = true, matchmissing = :notequal_view);
    @btime innerjoin($p2, $v2, on = :Id => :PostId, makeunique = true, matchmissing = :notequal_copy);
end

for p in [0.0, 0.001, 0.002, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 0.75]
    @show p
    testp(posts, votes, p);
end
```

```
p = 0.0
  86.736 ms (623 allocations: 116.23 MiB)
  96.405 ms (720 allocations: 145.00 MiB)
p = 0.001
  88.957 ms (625 allocations: 120.90 MiB)
  103.818 ms (701 allocations: 149.64 MiB)
p = 0.002
  93.594 ms (625 allocations: 120.68 MiB)
  119.779 ms (701 allocations: 149.40 MiB)
p = 0.005
  105.603 ms (625 allocations: 120.00 MiB)
  116.058 ms (701 allocations: 148.66 MiB)
p = 0.01
  103.518 ms (625 allocations: 118.92 MiB)
  119.578 ms (701 allocations: 147.48 MiB)
p = 0.02
  103.239 ms (625 allocations: 116.80 MiB)
  115.631 ms (701 allocations: 145.15 MiB)
p = 0.05
  95.703 ms (625 allocations: 110.15 MiB)
  111.279 ms (701 allocations: 137.88 MiB)
p = 0.1
  87.453 ms (625 allocations: 102.36 MiB)
  102.669 ms (701 allocations: 128.97 MiB)
p = 0.2
  71.221 ms (625 allocations: 82.82 MiB)
  83.093 ms (701 allocations: 107.16 MiB)
p = 0.5
  39.548 ms (628 allocations: 36.54 MiB)
  48.013 ms (704 allocations: 52.86 MiB)
p = 0.75
  12.758 ms (626 allocations: 11.03 MiB)
  17.022 ms (702 allocations: 19.71 MiB)
```

bkamins:
Thank you for checking this. Given your past PRs you probably know what needs to be done (please let me know if you are willing to do this):
1. merge `main` into the PR (to make sure we are on the latest version)
2. finalize the code
3. write comprehensive tests with full coverage of corner cases
4. add an appropriate NEWS.md entry
5. add appropriate updates to docstrings and to the manual
6. then go through the reviews 😄

Thank you!

pstorozenko:
Sure thing!
I rebased my code onto main, should I rather merge?

bkamins:
Well - I used to tell people to rebase, squash to one commit and force push the changes (this is what I normally do if I push a rewrite; merging is preferable if there are significant comments to the implementation pending in the old code, but in this case we do not have such, as we have discussed the design already, so I need to anyway just review the final implementation).

However, most contributors had problems with following this workflow so I started suggesting merging, which is usually simpler to handle properly in git 😄.

In short - if you rebased this is preferable.

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # DataFrames.jl changes on main since last release notes
 
+## New functionalities
+
+* add option `matchmissing=:notequal` in joins;
+in `leftjoin`, `semijoin` and `antijoin` `missing`s are dropped in right df,
+but preserved in left; in `rightjoin` `missing`s are dropped in left df,
+but preserved in right df; in `innerjoin` `missing`s are dropped in both dfs;
+in `outerjoin` method errors
+([#2724](https://github.com/JuliaData/DataFrames.jl/pull/2724))
```
bkamins, line 10:
```suggestion
  in `leftjoin`, `semijoin` and `antijoin` `missing`s are dropped in right data frame,
  but preserved in left; in `rightjoin` `missing`s are dropped in left data frame,
  but preserved in right df; in `innerjoin` `missing`s are dropped in both data frames;
  in `outerjoin` this value of keyword argument is not supported
  ([#2724](https://github.com/JuliaData/DataFrames.jl/pull/2724))
```

src/join/composer.jl
```diff
@@ -55,8 +56,25 @@ struct DataFrameJoiner
                                         "when matchmissing == :error"))
                 end
             end
+        elseif matchmissing === :notequal
+            if kind in (:left, :semi, :anti)
+                dfr = dropmissing(dfr, right_on, view = true)
```
bkamins, line 61:
```suggestion
                dfr = dropmissing(dfr, right_on, view=true)
```

src/join/composer.jl
```diff
@@ -55,8 +56,25 @@ struct DataFrameJoiner
...
+                dfr_on = select(dfr, right_on)
+            elseif kind === :right
+                dfl = dropmissing(dfl, left_on, view = true)
```
bkamins, line 64:
```suggestion
                dfl = dropmissing(dfl, left_on, view=true)
```

src/join/composer.jl
```diff
@@ -55,8 +56,25 @@ struct DataFrameJoiner
...
+                dfl_on = select(dfl, left_on)
+            elseif kind === :inner
+                dfl = dropmissing(dfl, left_on, view = true)
```
bkamins, line 67:
```suggestion
                dfl = dropmissing(dfl, left_on, view=true)
```

src/join/composer.jl
```diff
@@ -55,8 +56,25 @@ struct DataFrameJoiner
...
+                dfl_on = select(dfl, left_on)
+                dfr = dropmissing(dfr, right_on, view = true)
```
bkamins, line 69:
```suggestion
                dfr = dropmissing(dfr, right_on, view=true)
```

src/join/composer.jl
bkamins, line 67: as a thought - for `:innejrojn` it is enough to `dropmissing` only for left or right table, as in the other table we will not find a match anyway. But maybe it is faster to perform `dropmissing` on both first as it lowers the cost of joining later. Have you checked this?

(probably this will not make a huge difference, so I am asking just to make sure we have considered this)

src/join/composer.jl
```diff
@@ -55,8 +56,25 @@ struct DataFrameJoiner
...
+                dfr_on = select(dfr, right_on)
+            elseif kind === :outer
+                throw(ArgumentError("matchmissing == :notequal for `outerjoin` is not allowed"))
+            else
+                throw(ArgumentError("matchmissing == :notequal not implemented for kind == $kind"))
+            end
         elseif matchmissing !== :equal
-            throw(ArgumentError("matchmissing allows only :error or :equal"))
+            throw(ArgumentError("matchmissing allows only :error, :notequal and :equal"))
```
bkamins, line 77:
```suggestion
            throw(ArgumentError("matchmissing allows only :error, :equal, or :notequal"))
```

test/join.jl
```diff
@@ -1557,4 +1559,27 @@ end
                     c="c", d="d")
 end
 
+@testset "matchmissing :notequal correctness" begin
+    name = DataFrame(ID = Union{Int, Missing}[1, 2, missing],
+                    Name = Union{String, Missing}["John Doe", "Jane Doe", "Joe Blogs"])
+    noid = DataFrame(ID = Union{Int, Missing}[], Name = String[])
+    missid = DataFrame(ID = Union{Int, Missing}[missing, missing, missing],
+                    Name = String["John Doe", "Jane Doe", "Joe Blogs"])
+    job = DataFrame(ID = Union{Int, Missing}[missing, 2, 2, 4],
```
bkamins, line 1568: maybe for `job` add several options (so that we have a loop both on left and right table)

test/join.jl
```diff
@@ -1557,4 +1559,27 @@ end
                     c="c", d="d")
 end
 
+@testset "matchmissing :notequal correctness" begin
+    name = DataFrame(ID = Union{Int, Missing}[1, 2, missing],
```
bkamins, line 1563: we should also check:
1. joining on more than 1 column
2. having `missing` in non-key column that is not dropped (to make sure we do not accidentally drop it)
3. location of `missing` not only at the end but also in the front, middle + several scattered `missing`s (just to make sure we cover several different `_findall` cases - generating range or vector)

Thank you!

test/join.jl
pstorozenko: Done

test/join.jl
pstorozenko: Done

src/join/composer.jl
pstorozenko: Hmm, good catch. I will take a look at it tomorrow.

bkamins: As noted - it probably does not matter much, but it would be worth to double check. What you do now is a safe default (e.g. it might speed up things if data frame is sorted, but has `missing` sprinkled around as then faster join algorithm would be used).

If it does not matter much I would leave things as is. The reason is that in `innerjoin` dropping `missing`s might affect row order (as it will change which table - left or right is shorter), so in case it does not matter much the approach: we drop missings before joining is a good rule making sure that an advanced user will know what to expect in the output.

bkamins (approved):
I guess it it is not a draft any more :).

@nalimilan - the PR is ready for you to have a look at. Thank you!

src/join/composer.jl
```diff
@@ -55,8 +56,25 @@ struct DataFrameJoiner
                                         "when matchmissing == :error"))
                 end
             end
+        elseif matchmissing === :notequal
+            if kind in (:left, :semi, :anti)
+                dfr = dropmissing(dfr, right_on, view=true)
+                dfr_on = select(dfr, right_on)
```
nalimilan, line 62: Wouldn't `copycols=false` make sense here?

src/join/composer.jl
```diff
@@ -55,8 +56,25 @@ struct DataFrameJoiner
...
+                dfl_on = select(dfl, left_on)
+            elseif kind === :inner
+                dfl = dropmissing(dfl, left_on, view=true)
+                dfl_on = select(dfl, left_on)
+                dfr = dropmissing(dfr, right_on, view=true)
+                dfr_on = select(dfr, right_on)
+            elseif kind === :outer
+                throw(ArgumentError("matchmissing == :notequal for `outerjoin` is not allowed"))
+            else
+                throw(ArgumentError("matchmissing == :notequal not implemented for kind == $kind"))
```
nalimilan, line 74:
```suggestion
                throw(ArgumentError("matchmissing == :notequal not implemented for kind == :$kind"))
```

src/join/composer.jl
```diff
@@ -485,7 +503,8 @@ change in future releases.
   data frame and left unchanged.
 - `matchmissing` : if equal to `:error` throw an error if `missing` is present
   in `on` columns; if equal to `:equal` then `missing` is allowed and missings are
-  matched (`isequal` is used for comparisons of rows for equality)
+  matched; if equal to `:notequal` then missings are dropped in `df1` and `df2` 
+  `on` columns (`isequal` is used for comparisons of rows for equality)
```
nalimilan, line 507:
```suggestion
  `on` columns (in the two latter cases `isequal` is used for comparisons of rows for equality)
```
(and same below)

test/join.jl
```diff
@@ -1557,4 +1559,163 @@ end
                     c="c", d="d")
 end
 
+@testset "matchmissing :notequal correctness" begin
+    Random.seed!(1337)
+    names = [
+        DataFrame(ID = Union{Int, Missing}[1, 2, missing],
+            Name = Union{String, Missing}["John Doe", "Jane Doe", "Joe Blogs"]),
```
nalimilan, line 1566: Align with `ID`.

test/join.jl
```diff
@@ -1557,4 +1559,163 @@ end
...
+            antijoin(name, job, on=:ID, matchmissing=:notequal)
+        @test rightjoin(dropmissing(name, :ID), job, on=:ID, matchmissing=:equal) ≅
+            rightjoin(name, job, on=:ID, matchmissing=:notequal)
+        @test innerjoin(dropmissing(name, :ID), dropmissing(job, :ID), on=:ID, matchmissing=:equal) ≅
+            innerjoin(name, job, on=:ID, matchmissing=:notequal)
+    end
+
+    rl(n) = rand(["a", "b", "c"], n)
+    names2 = [
+        DataFrame(
```
nalimilan, line 1622: Put `ID1` on this line for style consistency (and align `ID2` with it).

test/join.jl
```diff
@@ -1557,4 +1559,163 @@ end
                     c="c", d="d")
 end
 
+@testset "matchmissing :notequal correctness" begin
+    Random.seed!(1337)
+    names = [
+        DataFrame(ID = Union{Int, Missing}[1, 2, missing],
```
nalimilan, line 1565: `Union{Int, Missing}` isn't needed here. Also, below, what's the point of using `Union{T, Missing}` even in the absence of `missing` values?

nalimilan (commented):
Thanks! Just a few minor points.

src/join/composer.jl
bkamins: definitely - same below. Thank you for spotting this.

Similarly lines 49 and 50 above duplicate work so it is better to do this operation only once (this is a minor issue), https://github.com/JuliaData/DataFrames.jl/pull/2724/files#diff-00fe5953fad7e5572a8a4ae57ab3724fd3a3895cae501c40bc60a9fed12cf60fR49-R50.

src/join/composer.jl
bkamins: `isequal` is used in all cases to test for equality so I think what we had before was correct; maybe just it should be a stand alone sentence - not something put in parentheses

test/join.jl
bkamins: we check for `Missing` in the eltype in `completecases` so it matters (but it is not super crucial as we have `completecases` test coverage done properly).

test/join.jl
bkamins: while we are at it - we do not put spaces for kwargs, so e.g. it should be `DataFrame(ID=Union{Int, Missing}[], Name=String[])`

bkamins:
@nalimilan - thank you for the review. You always have a keen eye for the details.

src/join/composer.jl
pstorozenko: I made some benchmarks with `innerjoin` and it's hard for me to decide on the best approach.
`innerjoin($p2, $v2, on = :Id => :PostId, makeunique = true, matchmissing = :notequal)`
I rerun experiments multiple times and this bump was consistent.

memory
time

bkamins: Interesting. As I have feared the best trade-off is not obvious (and probably depends on many factors) - let us keep "both drop" as it seems most robust. Thank you for checking this.

pstorozenko: Also, as you commented above, dropping only left/right changes the order of result and it took me some time to realize that I have to sort both dfs to compare them :D
I'll add comment in the code regarding possible gain in the future.

src/join/composer.jl
pstorozenko: Regarding lines 49-50
I noticed it but I had trouble with coming up with some elegant solution without obscuring ifs.
Since always view is returned here I decided to leave it as is.
After you comments I'll change it for `select(dfl, left_on, copycols=true)` for consistency.

test/join.jl
pstorozenko: I tried to mimic other tests, what is preferred by the way?
1)
```julia
DataFrame(ID=[1:100; missings(100)],
          Name=repeat(["Jane Doe"], 200)),
```
or
2)
```julia
DataFrame(ID=Union{Int, Missing}[1:100; missings(100)],
          Name=Union{String, Missing}[repeat(["Jane Doe"], 200)...]),
```
?

bkamins:
```
DataFrame(ID=[1:100; missings(100)],
          Name=repeat(["Jane Doe"], 200))
```
is preferred unless you want to explicitly test what happens if you have a column that does not have `missing`, but has `Missing` in `eltype` union. As I have commented in this case actually it has some impact as `completecases` is sensitive to this difference (and it is called at some point of the processing), but since we test `completecases` elsewhere we do not need to do it here.

src/join/composer.jl
bkamins: > I noticed it but I had trouble with coming up with some elegant solution without obscuring ifs.

The way would be:
```
if matchmissing === :notequal
    # do the stuff you have added and in every place explicitly define `dfl_on` and dfr_on`
else
    if matchmissing === :error
        for df in (dfl_on, dfr_on), col in eachcol(df)
                if any(ismissing, col)
                    throw(ArgumentError("missing values in key columns are not allowed " *
                                        "when matchmissing == :error"))
                end
            end
    end
    dfl_on = dfl[!, left_on]
    dfr_on = dfr[!, right_on]
end
```
in this way you define `dfl_on` and `dfr_on` only once.

test/join.jl
pstorozenko: Ok, makes a perfect sense. I felt like we want to write explicit types when possible for some reason (that is important, but I don't get it yet :D).

src/join/composer.jl
pstorozenko: But since it is used later in line 82, I will have to add `local dfl_on` before `if`, right?

bkamins: no, because `if` does not introduce a new scope, see https://docs.julialang.org/en/v1/manual/variables-and-scoping/#man-scope-table

src/join/composer.jl
```diff
@@ -45,8 +46,8 @@ struct DataFrameJoiner
                                     "Symbol or Pair{Symbol, Symbol}."))
             end
         end
-        dfl_on = dfl[!, left_on]
-        dfr_on = dfr[!, right_on]
+        dfl_on = select(dfl, left_on, copycols=true)
+        dfr_on = select(dfr, right_on, copycols=true)
```
bkamins, line 50:
```suggestion
        dfl_on = select(dfl, left_on, copycols=false)
        dfr_on = select(dfr, right_on, copycols=false)
```

src/join/composer.jl
```diff
@@ -55,8 +56,27 @@ struct DataFrameJoiner
                                         "when matchmissing == :error"))
                 end
             end
+        elseif matchmissing === :notequal
+            if kind in (:left, :semi, :anti)
+                dfr = dropmissing(dfr, right_on, view=true)
+                dfr_on = select(dfr, right_on, copycols=true)
```
bkamins, line 62:
```suggestion
                dfr_on = select(dfr, right_on, copycols=false)
```

src/join/composer.jl
```diff
@@ -55,8 +56,27 @@ struct DataFrameJoiner
...
+            elseif kind === :right
+                dfl = dropmissing(dfl, left_on, view=true)
+                dfl_on = select(dfl, left_on, copycols=true)
```
bkamins, line 65:
```suggestion
                dfl_on = select(dfl, left_on, copycols=false)
```

src/join/composer.jl
```diff
@@ -55,8 +56,27 @@ struct DataFrameJoiner
...
+            elseif kind === :inner
+                # it possible to drop only left or right df
+                # to gain some performance but needs more testing, see #2724
+                dfl = dropmissing(dfl, left_on, view=true)
+                dfl_on = select(dfl, left_on, copycols=true)
+                dfr = dropmissing(dfr, right_on, view=true)
+                dfr_on = select(dfr, right_on, copycols=true)
```
bkamins, line 72:
```suggestion
                dfr_on = select(dfr, right_on, copycols=false)
```

src/join/composer.jl
bkamins: This is normal (really) and that is why review process is so important.

When I see comments by @nalimilan of my code I often ask myself how I could have written and committed something.

pstorozenko: And that is why I decided to involve in open source project at least a bit!

src/join/composer.jl
bkamins: As an example why not having `if` create a new scope is useful consider e.g. this part of source: https://github.com/JuliaData/DataFrames.jl/blob/f56982d063254e947ca98247e5f471f8a095ecb0/src/DataFrames.jl#L92.

As you can see we either import or define `@spawn` conditionally, but it all happens in global scope. If `if` created a local scope then such things would be much harder to write.

pstorozenko:
Is there a way of _telling_ `codecov` that some lines should not be reached?

src/join/composer.jl
```diff
@@ -46,37 +46,36 @@ struct DataFrameJoiner
...
+                if any(ismissing, col)
+                    throw(ArgumentError("missing values in key columns are not allowed " *
+                                        "when matchmissing == :error"))
+                end
+            end
+        elseif !(matchmissing in (:equal, :notequal))
             throw(ArgumentError("matchmissing allows only :error, :equal, or :notequal"))
         end
-
+
```
nalimilan, line 78:
```suggestion

```

test/join.jl
```diff
@@ -1562,47 +1562,48 @@ end
...
-        DataFrame(ID = Union{Int, Missing}[rand([1:1000; missing], 10000)...],
-            Job = Union{String, Missing}[rand(["Lawyer", "Doctor", "Florist", missing], 10000)...]),
+        DataFrame(ID=[1, 2, 2, 4],
+                  Job=["Lawyer", "Doctor", "Florist", "Farmer"]),
+        DataFrame(ID=[missing, 2, 2, 4],
+                  Job=["Lawyer", "Doctor", "Florist", "Farmer"]),
+        DataFrame(ID=[missing, 2, 2, 4],
+                  Job=["Lawyer", "Doctor", missing, "Farmer"]),
+        DataFrame(ID=Union{Int, Missing}[],
+                  Job=Union{String, Missing}[]),
```
nalimilan, line 1594:
```suggestion
        DataFrame(ID=[],
                  Job=[]),
```

test/join.jl
```diff
@@ -1619,90 +1620,69 @@ end
...
+                  Name=["John Doe", "Jane Doe", "Joe Blogs"]),
+        DataFrame(ID1=[1, 1, 2, missing],
+                  ID2=["a", "b", "a", missing],
+                  Name=["John Doe", "Jane Doe", "Joe Blogs", missing]),
+        DataFrame(ID1=[missing, 1, 2, missing],
+                  ID2=["a", "b", missing, missing],
+                  Name=[missing, "Jane Doe", "Joe Blogs", missing]),
+        DataFrame(ID1=[missing, 1, 2, missing],
+                  ID2=["a", "b", missing, missing],
+                  Name=Union{String, Missing}[missings(4)...]),
```
nalimilan, line 1634:
```suggestion
                  Name=missings(4)),
```

test/join.jl
```diff
@@ -1619,90 +1620,69 @@ end
...
+        DataFrame(ID1=[missing, 1, 2, missing],
+                  ID2=Union{String, Missing}[missings(4)...],
```
nalimilan, line 1636:
```suggestion
                  ID2=missings(4),
```

test/join.jl
```diff
@@ -1562,47 +1562,48 @@ end
...
-            Name = Union{String, Missing}[repeat(["Jane Doe"], 200)...]),
-        DataFrame(ID = Union{Int, Missing}[1:63; missings(65); 129:200],
-            Name = Union{String, Missing}[repeat(["Jane Doe"], 200)...]),
-        DataFrame(ID = Union{Int, Missing}[rand([1:1000; missing], 10000)...],
-            Name = Union{String, Missing}[rand(["John Doe", "Jane Doe", "Joe Blogs", missing], 10000)...]),
+        DataFrame(ID=[1, 2, missing],
+                  Name=["John Doe", "Jane Doe", "Joe Blogs"]),
+        DataFrame(ID=Int[],
+                  Name=String[]),
+        DataFrame(ID=Union{Int, Missing}[missings(3)...],
```
nalimilan, line 1569:
```suggestion
        DataFrame(ID=missings(3),
```

test/join.jl
```diff
@@ -1562,47 +1562,48 @@ end
...
+                  Name=String["John Doe", "Jane Doe", "Joe Blogs"]),
```
nalimilan, line 1570:
```suggestion
                  Name=["John Doe", "Jane Doe", "Joe Blogs"]),
```

test/join.jl
```diff
@@ -1557,4 +1559,143 @@ end
                     c="c", d="d")
 end
 
+@testset "matchmissing :notequal correctness" begin
+    Random.seed!(1337)
+    names = [
+        DataFrame(ID=[1, 2, missing],
+                  Name=["John Doe", "Jane Doe", "Joe Blogs"]),
+        DataFrame(ID=Int[],
+                  Name=String[]),
```
nalimilan, line 1568:
```suggestion
        DataFrame(ID=[],
                  Name=[]),
```

test/join.jl
pstorozenko: I wanted to hint that this is empty array of type `Union{Int, Missing}[]`, not `Any[]`.
Same with missings bellow.
Do you think I should remove types all the way?

src/join/composer.jl
pstorozenko: Very interesting! I assumed it's like in c++ for some reason, I need to expand my knowledge on this topic.
Thanks

test/join.jl
nalimilan: Yes, unless there's a specific reason not to. For example, it would be appropriate if it's needed to cover a particular code path, test a particular case, or when checking that types are equal. But here it's not clear to me that it's the case.

pstorozenko: Ok, thanks

bkamins:
> Is there a way of telling codecov that some lines should not be reached?

No. If you have unreachable code (I have not looked at your commits yet) and want to make sure it is not reached throw an error there. See e.g. https://github.com/JuliaData/DataFrames.jl/blob/67ccc074796fb35e8b8879b2dfa160c682442eba/src/join/core.jl#L562

bkamins:
The PR looks good to me (if @nalimilan is OK with the code formatting, especially in the test section)

nalimilan (approved):

pstorozenko:
Thank you for reviews!

bkamins:
Thank you!

PR #2725, opened 2021-04-17
Update comparisons with data.table info

eloualiche:
Update to #2388, with comparison with data.table syntax. Focus on in place changes (!)

### DIFF
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -283,8 +283,8 @@ df2 <- data.table(grp=c(1,3), w = c(10,11))
 | Multivariate function             | `df[, .(cor(x,y)) ]`                                                                       | `transform(df, [:x, :y] => cor)`                                            |
 | Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform(df, [:x, :y] => ByRow(min))`                                     |
 |                                   | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`           | `transform(df, AsTable(r"^x") => ByRow(argmax))`                            |
-| DataFrame as output               | `df[, .SD[1], by=grp]`                                                                     | `combine(groupby(df, :grp), names(df) .=> (x->x[1]) .=> names(df)`          |
-| DataFrame as output               | `df[, .SD[which.max(x)], by=grp]`                                                          | `subset(groupby(df, :grp), :x => (x->(x.==maximum(x))) )`                   |
+| DataFrame as output               | `df[, .SD[1], by=grp]`                                                                     | `combine(groupby(df, :grp), first)`                                         |
+| DataFrame as output               | `df[, .SD[which.max(x)], by=grp]`                                                          | `combine(groupby(df, :grp), sdf -> sdf[argmax(sdf.x), :])`                  |
 
 ### Joining data frames
 
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -239,6 +239,13 @@ The table below compares more advanced commands:
 
 The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). 
 
+```R
+library(data.table)
+df  <- data.table(grp = rep(1:2, 3), x = 6:1, y = 4:9,
+                  z = c(3:7, NA), id = letters[1:6])
+df2 <- data.table(grp=c(1,3), w = c(10,11))                 
+```
+
 |Operations                         | data.table                                      | DataFrames.jl                                | 
 |:----------------------------------|:------------------------------------------------|:---------------------------------------------|
 |Reduce multiple values             | `df[, list(mean(x))]`                           | `combine(df, :x => mean)`                    |
@@ -267,7 +274,7 @@ The following table compares the main functions of DataFrames.jl with the R pack
 |:---------------------------------|:-------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------|
 |Complex Function                  | `df[, list(mean(x, na.rm = T)) ]`                                                          | `combine(df, :x => x -> mean(skipmissing(x)))`                             |
 |Transform certain rows (in place) | `df[x<=0, x:=0 ]`                                                                          | `df.x[df.x .<= 0] .= 0`                                                    |
-|Transform several columns         | `df[, list(max(x), min(y)) ]`                                                              | `combine(df, :x => maximum, :y => minimum)`                               |
+|Transform several columns         | `df[, list(max(x), min(y)) ]`                                                              | `combine(df, :x => maximum, :y => minimum)`                                |
 |                                  | `df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`                                          | `combine(df, [:x, :y] .=> mean)`                                           |
 |                                  | `df[, lapply(.SD, mean), .SDcols = patterns("x*") ]`                                       | `combine(df, names(df, r"^x") .=> mean)`                                   |
 |                                  | `df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]` | `combine(df, ([:x, :y] .=> [maximum minimum])...)`                         |
@@ -275,7 +282,6 @@ The following table compares the main functions of DataFrames.jl with the R pack
 |Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform(df, [:x, :y] => ByRow(min))`                                    |
 |                                  | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`           | `transform(df, AsTable(r"^x") => ByRow(argmax))`                           |
 |DataFrame as output               | `df[, .SD[, .(value = min(x), max(x))] ]`                                                  | `combine(df, :x => (x -> (value = [minimum(x), maximum(x)],)) => AsTable)` |
-
 
 ### Joining data frames
 
@@ -286,6 +292,7 @@ The following table compares the main functions of DataFrames.jl with the R pack
 | Left join             | `merge(df, df2, all.x = TRUE, on = "grp")`      | `leftjoin(df, df2, on = :grp)`  |
 | Right join            | `merge(df, df2, all.y = TRUE, on = "grp")`      | `rightjoin(df, df2, on = :grp)` |
 | Anti join (filtering) | `df[!df2, on = "grp" ]`                         | `antijoin(df, df2, on = :grp)`  |
+| Semi join (filtering) | `merge(df1, df2[, list(grp)])`                  | `semijoin(df, df2, on = :grp)`  |
 
 
 ## Comparison with Stata (version 8 and above)
@@ -298,7 +305,7 @@ The following table compares the main functions of DataFrames.jl with Stata:
 | Add new columns        | `egen x_mean = mean(x)` | `transform!(df, :x => mean => :x_mean)` |
 | Rename columns         | `rename x x_new`        | `rename!(df, :x => :x_new)`             |
 | Pick columns           | `keep x y`              | `select!(df, :x, :y)`                   |
-| Pick rows              | `keep if x >= 1`        | `subset!(df, :x => ByRow(x -> x >= 1)`        |
+| Pick rows              | `keep if x >= 1`        | `subset!(df, :x => ByRow(x -> x >= 1)`  |
 | Sort rows              | `sort x`                | `sort!(df, :x)`                         |
 
 Note that the suffix `!` (i.e. `transform!`, `select!`, etc) ensures that the operation transforms the dataframe in place, as in Stata
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -235,6 +235,7 @@ The table below compares more advanced commands:
 | DataFrame as input        | `summarize(df, head(across(), 2))`                        | `combine(d -> first(d, 2), df)`                               |
 | DataFrame as output       | `summarize(df, tibble(value = c(min(x), max(x))))`        | `combine(:x => x -> (value = [minimum(x), maximum(x)],), df)` |
 
+
 ## Comparison with the R package data.table
 
 The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). 
@@ -248,14 +249,15 @@ df2 <- data.table(grp=c(1,3), w = c(10,11))
 
 | Operation                          | data.table                                       | DataFrames.jl                                | 
 |:-----------------------------------|:-------------------------------------------------|:---------------------------------------------|
-| Reduce multiple values             | `df[, list(mean(x))]`                            | `combine(df, :x => mean)`                    |
-| Add new columns                    | `df[, x_mean := mean(x) ]`                       | `transform(df, :x => mean => :x_mean)`       |
+| Reduce multiple values             | `df[, .(mean(x))]`                               | `combine(df, :x => mean)`                    |
+| Add new columns                    | `df[, x_mean:=mean(x) ]`                         | `transform(df, :x => mean => :x_mean)`       |
 | Rename column (in place)           | `setnames(df, "x", "x_new")`                     | `rename!(df, :x => :x_new)`                  |
 | Rename multiple columns (in place) | `setnames(df, c("x", "y"), c("x_new", "y_new"))` | `rename!(df, [:x, :y] .=> [:x_new, :y_new])` |
-| Pick columns                       | `df[, list(x, y)]`                               | `select(df, :x, :y)`                         |
-| Remove columns                     | `df[, -"x" ]`                                    | `select(df, Not(:x))`                        |
-| Remove columns (in place)          | `df[, c("x") := NULL ]`                          | `select!(df, Not(:x))`                       |
-| Pick & transform columns           | `df[, list(mean(x), y)]`                         | `select(df, :x => mean, :y)`                 |
+| Pick columns                       | `df[, .(x, y)]`                                  | `select(df, :x, :y)`                         |
+| Remove columns                     | `df[, -"x"]`                                     | `select(df, Not(:x))`                        |
+| Remove columns (in place)          | `df[, x:=NULL]`                                  | `select!(df, Not(:x))`                       |
+| Remove columns (in place)          | `df[, c("x", "y"):=NULL]`                        | `select!(df, Not([:x, :y]))`                 |
+| Pick & transform columns           | `df[, .(mean(x), y)]`                            | `select(df, :x => mean, :y)`                 |
 | Pick rows                          | `df[ x >= 1 ]`                                   | `filter(:x => >=(1), df)`                    |
 | Sort rows (in place)               | `setorder(df, x)`                                | `sort!(df, :x)`                              |
 | Sort rows                          | `df[ order(x) ]`                                 | `sort(df, :x)`                               |
@@ -264,24 +266,25 @@ df2 <- data.table(grp=c(1,3), w = c(10,11))
 
 | Operation                   | data.table                                        | DataFrames.jl                             |
 |:----------------------------|:--------------------------------------------------|:------------------------------------------|
-| Reduce multiple values      | `df[, list(mean(x)), by = list(id) ]`             | `combine(groupby(df, :id), :x => mean)`   |
-| Add new columns (in place)  | `df[, x_mean := mean(x), by = list(id) ]`         | `transform!(groupby(df, :id), :x => mean)`|
-| Pick & transform columns    | `df[, list(x_mean = mean(x), y), by = list(id) ]` | `select(groupby(df, :id), :x => mean, :y)`|
+| Reduce multiple values      | `df[, mean(x), by=id ]`                           | `combine(groupby(df, :id), :x => mean)`   |
+| Add new columns (in place)  | `df[, x_mean:=mean(x), by=id]`                    | `transform!(groupby(df, :id), :x => mean)`|
+| Pick & transform columns    | `df[, .(x_mean = mean(x), y), by=id]`             | `select(groupby(df, :id), :x => mean, :y)`|
 
 ### More advanced commands
 
 | Operation                         | data.table                                                                                 | DataFrames.jl                                                               |
 |:----------------------------------|:-------------------------------------------------------------------------------------------|:----------------------------------------------------------------------------|
-| Complex Function                  | `df[, list(mean(x, na.rm = T)) ]`                                                          | `combine(df, :x => x -> mean(skipmissing(x)))`                              |
+| Complex Function                  | `df[, .(mean(x, na.rm=TRUE)) ]`                                                            | `combine(df, :x => x -> mean(skipmissing(x)))`                              |
 | Transform certain rows (in place) | `df[x<=0, x:=0 ]`                                                                          | `df.x[df.x .<= 0] .= 0`                                                     |
-| Transform several columns         | `df[, list(max(x), min(y)) ]`                                                              | `combine(df, :x => maximum, :y => minimum)`                                 |
+| Transform several columns         | `df[, .(max(x), min(y)) ]`                                                                 | `combine(df, :x => maximum, :y => minimum)`                                 |
 |                                   | `df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`                                          | `combine(df, [:x, :y] .=> mean)`                                            |
 |                                   | `df[, lapply(.SD, mean), .SDcols = patterns("x*") ]`                                       | `combine(df, names(df, r"^x") .=> mean)`                                    |
 |                                   | `df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]` | `combine(df, ([:x, :y] .=> [maximum minimum])...)`                          |
-| Multivariate function             | `df[, list(cor(x,y)) ]`                                                                    | `transform(df, [:x, :y] => cor)`                                            |
+| Multivariate function             | `df[, .(cor(x,y)) ]`                                                                       | `transform(df, [:x, :y] => cor)`                                            |
 | Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform(df, [:x, :y] => ByRow(min))`                                     |
 |                                   | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`           | `transform(df, AsTable(r"^x") => ByRow(argmax))`                            |
-| DataFrame as output               | `df[, .SD[, list(value = c(min(x), max(x)) )] ]`                                           | `combine(df, :x => (x -> (value = [minimum(x), maximum(x)],)) => AsTable)`  |
+| DataFrame as output               | `df[, .SD[1], by=grp]`                                                                     | `combine(groupby(df, :grp), names(df) .=> (x->x[1]) .=> names(df)`          |
+| DataFrame as output               | `df[, .SD[which.max(x)], by=grp]`                                                          | `subset(groupby(df, :grp), :x => (x->(x.==maximum(x))) )`                   |
 
 ### Joining data frames
 
@@ -292,7 +295,7 @@ df2 <- data.table(grp=c(1,3), w = c(10,11))
 | Left join             | `merge(df, df2, all.x = TRUE, on = "grp")`      | `leftjoin(df, df2, on = :grp)`  |
 | Right join            | `merge(df, df2, all.y = TRUE, on = "grp")`      | `rightjoin(df, df2, on = :grp)` |
 | Anti join (filtering) | `df[!df2, on = "grp" ]`                         | `antijoin(df, df2, on = :grp)`  |
-| Semi join (filtering) | `merge(df1, df2[, list(grp)])`                  | `semijoin(df, df2, on = :grp)`  |
+| Semi join (filtering) | `merge(df1, df2[, .(grp)])   `                  | `semijoin(df, df2, on = :grp)`  |
 
 
 ## Comparison with Stata (version 8 and above)
--- a/docs/src/man/comparisons.md
+++ b/docs/src/man/comparisons.md
@@ -237,43 +237,44 @@ The table below compares more advanced commands:
 
 ## Comparison with the R package data.table
 
-The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). To contrast data.table with dplyr, the focus has been put on transformations in place.
+The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). 
 
-|Operations              | data.table  | DataFrames.jl|  
-|:-----------------------|:------------|:-------------|
-|Reduce multiple values  |`df[, list(mean(x))]`       |`combine(df, :x => mean)`|
-|Add new columns         |`df[, x_mean := mean(x) ]`  |`transform(df, :x => mean => :x_mean)`|
-|Rename column (in place) |`setnames(df, "x", "x_new")`|`rename!(df, :x => :x_new)`|
-|Rename multiple columns (in place) |`setnames(df, c("x", "y"), c("x_new", "y_new")`|`rename!(df, [:x, :y] .=> [:x_new, :y_new])`|
-|Pick columns            |`df[, list(x, y)]`          |`select(df, :x, :y)`|
-|Remove columns          |`df[, -"x" ]`               |`select(df, Not(:x))`|
-|Remove columns (in place) |`df[, c("x") := NULL ]`   |`select!(df, Not(:x))`|
-|Pick & transform columns |`df[, list(mean(x), y)]`    |`select(df, :x => mean, :y)`|
-|Pick rows               |`df[ x >= 1 ]`              |`filter(:x => >=(1), df)`|
-|Sort rows (in place)    |`setorder(df, x)`           |`sort!(df, :x)`|
-|Sort rows               |`df[ order(x) ]`            |`sort(df, :x)`|
+|Operations                         | data.table                                      | DataFrames.jl                                | 
+|:----------------------------------|:------------------------------------------------|:---------------------------------------------|
+|Reduce multiple values             | `df[, list(mean(x))]`                           | `combine(df, :x => mean)`                    |
+|Add new columns                    | `df[, x_mean := mean(x) ]`                      | `transform(df, :x => mean => :x_mean)`       |
+|Rename column (in place)           | `setnames(df, "x", "x_new")`                    | `rename!(df, :x => :x_new)`                  |
+|Rename multiple columns (in place) | `setnames(df, c("x", "y"), c("x_new", "y_new")` | `rename!(df, [:x, :y] .=> [:x_new, :y_new])` |
+|Pick columns                       | `df[, list(x, y)]`                              | `select(df, :x, :y)`                         |
+|Remove columns                     | `df[, -"x" ]`                                   | `select(df, Not(:x))`                        |
+|Remove columns (in place)          | `df[, c("x") := NULL ]`                         | `select!(df, Not(:x))`                       |
+|Pick & transform columns           | `df[, list(mean(x), y)]`                        | `select(df, :x => mean, :y)`                 |
+|Pick rows                          | `df[ x >= 1 ]`                                  | `filter(:x => >=(1), df)`                    |
+|Sort rows (in place)               | `setorder(df, x)`                               | `sort!(df, :x)`                              |
+|Sort rows                          | `df[ order(x) ]`                                | `sort(df, :x)`                               |
 
 ### Grouping data and aggregation
 
-|Operations              | data.table                                      | DataFrames.jl     |
-|:-----------------------|:------------------------------------------------|:------------------|
-|Reduce multiple values  |`df[, list(mean(x)), by = list(id) ]`            |`combine(groupby(df, :id), :x => mean)`|
-|Add new columns (in place) |`df[, x_mean := mean(x), by = list(id) ]`        |`transform!(groupby(df, :id), :x => mean)`|
-|Pick & transform columns|`df[, list(x_mean = mean(x), y), by = list(id) ]`|`select(groupby(df, :id), :x => mean, :y)`|
+|Operations                 | data.table                                        | DataFrames.jl                             |
+|:--------------------------|:--------------------------------------------------|:------------------------------------------|
+|Reduce multiple values     | `df[, list(mean(x)), by = list(id) ]`             | `combine(groupby(df, :id), :x => mean)`   |
+|Add new columns (in place) | `df[, x_mean := mean(x), by = list(id) ]`         | `transform!(groupby(df, :id), :x => mean)`|
+|Pick & transform columns   | `df[, list(x_mean = mean(x), y), by = list(id) ]` | `select(groupby(df, :id), :x => mean, :y)`|
 
 ### More advanced commands
 
-|Operations              | data.table                                      | DataFrames.jl             |
-|:-----------------------|:------------------------------------------------|:--------------------------|
-|Complex Function          |`df[, list(mean(x, na.rm = T)) ]`              |`combine(df, :x => x -> mean(skipmissing(x)))`|
-|Transform certain rows (in place) |`df[x<=0, x:=0 ]`                      |`df.x[df.x .<= 0] .= 0`|
-|Transform several columns |`df[, list(max(x), min(y)) ]`                  |`combine(df, :x => maximum,  :y => minimum)`|
-|                          |`df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`   |`combine(df, [:x, :y] .=> mean)`|
-|                          |`df[, lapply(.SD, mean), .SDcols = patterns("x*") ]`|`combine(df, names(df, r"^x") .=> mean)`|
-|                          |`df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]`|`combine(df, ([:x, :y] .=> [maximum minimum])...)`|
-|Multivariate function     |`df[, list(cor(x,y)) ]`                        |`transform(df, [:x, :y] => cor)`|
-|Row-wise                  |`df[, min_xy := min(x, y), by = 1:nrow(df)]`   |`transform(df, [:x, :y] => ByRow(min))`|
-|                          |`df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`|`transform(df, AsTable(r"^x") => ByRow(argmax))`|
+|Operations                        | data.table                                                                                 | DataFrames.jl                                                              |
+|:---------------------------------|:-------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------|
+|Complex Function                  | `df[, list(mean(x, na.rm = T)) ]`                                                          | `combine(df, :x => x -> mean(skipmissing(x)))`                             |
+|Transform certain rows (in place) | `df[x<=0, x:=0 ]`                                                                          | `df.x[df.x .<= 0] .= 0`                                                    |
+|Transform several columns         | `df[, list(max(x), min(y)) ]`                                                              | `combine(df, :x => maximum, :y => minimum)`                               |
+|                                  | `df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`                                          | `combine(df, [:x, :y] .=> mean)`                                           |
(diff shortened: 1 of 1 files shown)


docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
 | DataFrame as input        | `summarize(df, head(across(), 2))`                        | `combine(d -> first(d, 2), df)`                               |
 | DataFrame as output       | `summarize(df, tibble(value = c(min(x), max(x))))`        | `combine(:x => x -> (value = [minimum(x), maximum(x)],), df)` |
 
+## Comparison with the R package data.table
+
+The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). To contrast data.table with dplyr, the focus has been put on transformations in place.
+
+|Operations              | data.table  | DataFrames.jl|  
+|:-----------------------|:------------|:-------------|
+|Reduce multiple values  |`df[, list(mean(x))]`       |`combine(df, :x => mean)`|
```
bkamins, line 244: this is not the same as the resulting variable name does not match.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|Add new columns         |`df[, x_mean := mean(x) ]`  |`transform(df, :x => mean => :x_mean)`|
+|Rename column (in place) |`setnames(df, "x", "x_new")`|`rename!(df, :x => :x_new)`|
+|Rename multiple columns (in place) |`setnames(df, c("x", "y"), c("x_new", "y_new")`|`rename!(df, [:x, :y] => [:x_new, :y_new])`|
```
bkamins, line 247:
```suggestion
|Rename multiple columns (in place) |`setnames(df, c("x", "y"), c("x_new", "y_new")`|`rename!(df, [:x, :y] .=> [:x_new, :y_new])`|
```

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|Pick columns            |`df[, list(x, y)]`          |`select(df, :x, :y)`|
+|Remove columns          |`df[, -"x" ]`               |`select(df, Not(:x))`|
+|Remove columns (in place) |`df[, c("x") := NULL ]`   |`select!(df, Not(:x))`|
+|Pick & transform columns|`df[, list(mean(x), y)]`    |`select(df, :x => mean, :y)`|
```
bkamins, line 251:
```suggestion
|Pick & transform columns |`df[, list(mean(x), y)]`    |`select(df, :x => mean, :y)`|
```
also the same as above - column names do not match

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|Pick rows               |`df[ x >= 1 ]`              |`filter(:x => >=(1), df)`|
+|Sort rows (in place)    |`setorder(df, x)`           |`sort!(df, :x)`|
+|Sort rows               |`df[ order(x) ]`            |`sort(df, :x)`|
+
+### Grouping data and aggregation
+
+|Operations              | data.table                                      | DataFrames.jl     |
+|:-----------------------|:------------------------------------------------|:------------------|
+|Reduce multiple values  |`df[, list(mean(x)), by = list(id) ]`            |`combine(groupby(df, :id), :x => mean)`|
```
bkamins, line 260: same as above - names do not match

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|Add new columns (in place) |`df[, x_mean := mean(x), by = list(id) ]`        |`transform!(groupby(df, :id), :x => mean)`|
+|Pick & transform columns|`df[, list(x_mean = mean(x), y), by = list(id) ]`|`select(groupby(df, :id), :x => mean, :y)`|
+
+### More advanced commands
+
+|Operations              | data.table                                      | DataFrames.jl             |
+|:-----------------------|:------------------------------------------------|:--------------------------|
+|Complex Function          |`df[, list(mean(x, na.rm = T)) ]`              |`combine(df, :x => x -> mean(skipmissing(x)))`|
```
bkamins, line 268: same

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|Transform certain rows (in place) |`df[x<=0, x:=0 ]`                      |`transform!(df, :x => (x -> ifelse.(x .<= 0, 0, x)) => :x)`|
```
bkamins, line 269: why not just:
```
df.x[df.x .<= 0] .= 0
```
?

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|Transform several columns |`df[, list(max(x), min(y)) ]`                  |`combine(df, :x => maximum,  :y => minimum)`|
+|                          |`df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`   |`combine(df, [:x, :y] .=> mean)`|
```
bkamins, line 271: this errors

bkamins: (maybe I am on a wrong data.table version?)

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|                          |`df[, lapply(.SD, mean), .SDcols = patterns("x*") ]`|`combine(df, names(df, r"^x") .=> mean)`|
+|                          |`df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]`|`combine(df, ([:x, :y] .=> [maximum minimum])...)`|
+|Multivariate function     |`df[, list(cor(x,y)) ]`                        |`transform(df, [:x, :y] => cor)`|
+|Row-wise                  |`df[, min_xy := min(x, y), by = 1:nrow(df)]`   |`transform(df, [:x, :y] => ByRow(min))`|
+|                          |`df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`|`transform(df, AsTable(r"^x") => ByRow(argmax))`|
+|DataFrame as input        |`df[, head(.SD, 2) ]`                          |`combine(d -> first(d, 2), df)`|
```
bkamins, line 277: I am not sure what we want to show with this line.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+|DataFrame as output       |`df[, .SD[, .(value = min(x), max(x))] ]`      |`combine(:x => x -> (value = [minimum(x), maximum(x)]), df)`|
```
bkamins, line 278: same here - there is something wrong with this example (for sure DataFrames.jl example is not working)

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,59 @@ The table below compares more advanced commands:
...
+
+### Joining data frames
+
+| Operation             | data.table                                      | DataFrames.jl                   |
+|:----------------------|:------------------------------------------------|:--------------------------------|
+| Inner join            | `merge(df, df2, on = "grp")`                    | `innerjoin(df, df2, on = :grp)` |
+| Outer join            | `merge(df, df2, all = TRUE, on = "grp")`        | `outerjoin(df, df2, on = :grp)` |
+| Left join             | `merge(df, df2, all.x = TRUE, on = "grp")`      | `leftjoin(df, df2, on = :grp)`  |
+| Right join            | `merge(df, df2, all.y = TRUE, on = "grp")`      | `rightjoin(df, df2, on = :grp)` |
+| Anti join (filtering) | `df[!df2, on = "grp" ]`                         | `antijoin(df, df2, on = :grp)`  |
```
bkamins, line 288: probably it would be good to add then semijoin and crossjoin for completeness.

bkamins:
Thank you!

docs/src/man/comparisons.md
eloualiche: I have not found a good standard for semi join in data.table. I can add cross join though. 
I am restricting myself to what I think is idiomatic to data.table. Perhaps there ought to be a join section within the dplyr section as well?

docs/src/man/comparisons.md
eloualiche: I think they are conceptually the same. data.table automatically names variable V1, V2, ... and dataframes appends the function name. 
I dont think anyone would ever write: `df[, list(x_mean = mean(x)) ]` if we just wanted to visualize the mean?

Happy to change it if you think the match has to be exactly the same, but this was made to illustrate a case where you create a new variable and let the framework pick the name.

bkamins: OK - but the difference is that `V1` is a meaningless name and I would never want such a name in my data frame :).
But this is probably a matter of style, so I am OK to leave it as is.

docs/src/man/comparisons.md
bkamins: OK - let us keep with what is idiomatic for data.table. This section is for people coming from other ecosystems to DataFrames.jl.

docs/src/man/comparisons.md
eloualiche: I copied the stata example. I have asked this question previously and the ifelse was the one recommended. I think we ought to include one such example where x might include missing (in which case we have to add a coalesce?). 

Tbh, this is where the base data frames syntax shows its limit and I hope that this sort of complicated things will be easier with dataframesmeta down the road.

docs/src/man/comparisons.md
eloualiche: Which one errors. I am running data.table 1.14.1 under R 4.03.

The following seems to be ok:

```R
library(data.table)
df = data.table(mtcars)
df[, lapply(.SD, mean), .SDcols = c("mpg", "cyl") ]
df[, lapply(.SD, mean), .SDcols = patterns("mp.*") ]
df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("mpg", "cyl") ]
```

docs/src/man/comparisons.md
eloualiche: Take it to Matt Dowle ;)

docs/src/man/comparisons.md
eloualiche: I lost track of my former self. I will remove for now. 
I think I wanted to show how DataFrames can easily manipulate the table itself like the idiomatic `.SD` of data.table. Since data.table people like me think this is one of the thing that makes data.table really nice and complete, I wanted to illustrate some sort of feature parity. This example is not that.

docs/src/man/comparisons.md
eloualiche: Same as above. I think this is useful when doing grouped operation. 
When I use DataFrames though I prefer to iterate over the subgroups and run my operation in a loop for clarity (at the cost of conciseness). So perhaps this is not really the place for such example and a separate long and detailed post could go into such details.

docs/src/man/comparisons.md
eloualiche: There is one option in a data.table outer join that allows multiple match on each side with the allow.cartesian=T [flag] (https://jangorecki.gitlab.io/-/data.table/-/jobs/640724/artifacts/public/html/data.table.html).
It is more a convenience feature (merge first and filter later) than a proper merge though.

One thing that we could discuss post 1.0 is whether rolling joins is a feature that we should aim for. (I will leave this for another issue...)

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,58 @@ The table below compares more advanced commands:
 | DataFrame as input        | `summarize(df, head(across(), 2))`                        | `combine(d -> first(d, 2), df)`                               |
 | DataFrame as output       | `summarize(df, tibble(value = c(min(x), max(x))))`        | `combine(:x => x -> (value = [minimum(x), maximum(x)],), df)` |
 
+## Comparison with the R package data.table
+
+The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). To contrast data.table with dplyr, the focus has been put on transformations in place.
+
+|Operations              | data.table  | DataFrames.jl|
```
nalimilan, line 242: Can you add a space before and after bars for consistency with above? Could you also align bars to make the code easier to read?

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,58 @@ The table below compares more advanced commands:
 | DataFrame as input        | `summarize(df, head(across(), 2))`                        | `combine(d -> first(d, 2), df)`                               |
 | DataFrame as output       | `summarize(df, tibble(value = c(min(x), max(x))))`        | `combine(:x => x -> (value = [minimum(x), maximum(x)],), df)` |
 
+## Comparison with the R package data.table
+
+The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). To contrast data.table with dplyr, the focus has been put on transformations in place.
```
nalimilan, line 240: "To contrast data.table with dplyr" is a bit of a weird goal here: the point of comparisons with other implementations is to allow their users to understand DataFrames.jl easily, not to compare them with other implementations. Does data.table support non-in-place operations?

docs/src/man/comparisons.md
bkamins: I agree that currently we do not have a super nice syntax for working on subsets of rows. I guess the difference between `df.x[df.x .<= 0] .= 0` and `transform!(df, :x => (x -> ifelse.(x .<= 0, 0, x)) => :x)` is that:

```
df.x[df.x .<= 0] .= 0
```
does not play way with piping, but as `data.table` is not piping oriented either then I thought it is not a problem. I assume @pdeffebach prefers `transform!` because of piping.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,58 @@ The table below compares more advanced commands:
...
+
+|Operations              | data.table  | DataFrames.jl|  
+|:-----------------------|:------------|:-------------|
+|Reduce multiple values  |`df[, list(mean(x))]`       |`combine(df, :x => mean)`|
+|Add new columns         |`df[, x_mean := mean(x) ]`  |`transform(df, :x => mean => :x_mean)`|
+|Rename column (in place) |`setnames(df, "x", "x_new")`|`rename!(df, :x => :x_new)`|
+|Rename multiple columns (in place) |`setnames(df, c("x", "y"), c("x_new", "y_new")`|`rename!(df, [:x, :y] .=> [:x_new, :y_new])`|
```
nalimilan, line 247: Not sure whether it fits, but it would be nice to show the more idiomatic approach:
```suggestion
|Rename multiple columns (in place) |`setnames(df, c("x", "y"), c("x_new", "y_new")`|`rename!(df, :x => :x_new, :y => :y_new)` or `rename!(df, [:x, :y] .=> [:x_new, :y_new])`|
```

docs/src/man/comparisons.md
bkamins: I probably made some error as indeed it works. Again - column names do not match but I guess it is ok.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,58 @@ The table below compares more advanced commands:
...
+|Add new columns (in place) |`df[, x_mean := mean(x), by = list(id) ]`        |`transform!(groupby(df, :id), :x => mean)`|
+|Pick & transform columns|`df[, list(x_mean = mean(x), y), by = list(id) ]`|`select(groupby(df, :id), :x => mean, :y)`|
+
+### More advanced commands
+
+|Operations              | data.table                                      | DataFrames.jl             |
+|:-----------------------|:------------------------------------------------|:--------------------------|
+|Complex Function          |`df[, list(mean(x, na.rm = T)) ]`              |`combine(df, :x => x -> mean(skipmissing(x)))`|
+|Transform certain rows (in place) |`df[x<=0, x:=0 ]`                      |`df.x[df.x .<= 0] .= 0`|
+|Transform several columns |`df[, list(max(x), min(y)) ]`                  |`combine(df, :x => maximum,  :y => minimum)`|
```
nalimilan, line 270:
```suggestion
|Transform several columns |`df[, list(max(x), min(y)) ]`                  |`combine(df, :x => maximum, :y => minimum)`|
```

docs/src/man/comparisons.md
bkamins: I meant that:
```
combine(:x => x -> (value = [minimum(x), maximum(x)]), df)
```
is not a valid syntax any more. You probably meant to write (note the comma after `]` otherwise `value = ` is just an assignment to a local variable that will get ignored):
```
julia> combine(df, :x => (x -> (value = [minimum(x), maximum(x)],)) => AsTable)
```
or
```
julia> combine(df, :x => (x -> [minimum(x), maximum(x)]) => :value)
```
btw. this specific case can be also written as:
```
julia> combine(df, :x => collect∘extrema => :value)
```

docs/src/man/comparisons.md
eloualiche: Correct me if I am wrong. The `setnames` function in data.table is nice because it is convenient for programmatic use. `old_var = ...; new_var = ...; setnames(df, old_var, new_var)`.

I am thinking that the form I have in the text is easiest to use in this way?

docs/src/man/comparisons.md
eloualiche: I just nixed it. I am assuming the guide is for people coming over from data.table rather than the opposite. so they know what they generally know that data.table has lots of operations that are changes in place.

docs/src/man/comparisons.md
eloualiche: Done (I think). This is easier to read (if you have a wide screen!)

docs/src/man/comparisons.md
grantmcdermott: Two minor comments that @eloualiche is almost certainly aware of, but just in case it's useful...

1. Much of a muchness, but most data.table users prefer the more concise `.()` instead of `list()` in my experience. I seem to recall that the former is also the preferred syntax in the vignettes.
2. On that note, you don't need to coerce to `list()` / `.()` for single variable reduction. E.g. `df[, mean(x)]` and df[, mean(x), by = grp] would both work.

I agree that renaming aggregation variables in place is probably a matter of style and, if the main purpose is to help data.table users navigate in DF.jl, then keeping things simple is best. FWIW the idiomatic way of doing this is via `.SD(cols)`, which Erik documents further down in the PR.

docs/src/man/comparisons.md
grantmcdermott: > I have not found a good standard for semi join in data.table. 

Wouldn't it just be `merge(df1, df2[, .(grp)])`? Or, wrap with with `unique()` if you want to limit to remove dups.

docs/src/man/comparisons.md
bkamins: In DataFrames.jl both forms are easy enough to use programatically (but I agree that vector form is more natural and it would have lower compilation latency).
In fact `rename!(df, :x => :x_new, :y => :y_new)` can be obtained from your form by writing `rename!(df, ([:x, :y] .=> [:x_new, :y_new])...)`.

bkamins:
@eloualiche + @grantmcdermott : as a general comment my preference would be to use the syntax that data.table users would find natural (I guess this follows your recommendations, but I just wanted to highlight this). This section is intended to help users who know data.tables understand how to write an equivalent code in DataFrames.jl. Thank you for working on this!

docs/src/man/comparisons.md
eloualiche: Thanks Grant. I thought having the dot here might be confusing. As I said this is mostly for people coming from data.table so I erred on the side of being explicit rather than short (though sometimes it's a tradeoff).

docs/src/man/comparisons.md
eloualiche: I have to be honest. I don't get the point of semijoins? Is this just a filtering on the rows of df2? I have added it but it sounds a lot like more of a convenient wrapper than an actual fundamental join?

I am not a database expert though, so educate me!

docs/src/man/comparisons.md
nalimilan: Yeah it really depends on the use case, which is why showing both sounds useful if that fits.

nalimilan: But since the single-argument case is shown above maybe that's enough and people will guess that they can pass multiple pairs if needed.

docs/src/man/comparisons.md
```diff
@@ -237,44 +237,51 @@ The table below compares more advanced commands:
...
+The following table compares the main functions of DataFrames.jl with the R package data.table (version 1.14.1). 
+
+```R
+library(data.table)
+df  <- data.table(grp = rep(1:2, 3), x = 6:1, y = 4:9,
+                  z = c(3:7, NA), id = letters[1:6])
+df2 <- data.table(grp=c(1,3), w = c(10,11))                 
+```
+
+|Operations                         | data.table                                      | DataFrames.jl                                |
```
nalimilan, line 249: Sill missing a space after `|`.

docs/src/man/comparisons.md
grantmcdermott: TBH I'm not much of fan either. I don't think I've ever used a semi join in any real-life project. Still, the above suggestion should get you there if you want it (not saying you do!)

docs/src/man/comparisons.md
```diff
@@ -1,4 +1,4 @@
-# Comparisons
+# comparisons
```
nalimilan, line 1:
```suggestion
# Comparisons
```

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,66 @@ The table below compares more advanced commands:
...
+```R
+library(data.table)
+df  <- data.table(grp = rep(1:2, 3), x = 6:1, y = 4:9,
+                  z = c(3:7, NA), id = letters[1:6])
+df2 <- data.table(grp=c(1,3), w = c(10,11))                 
+```
+
+| Operation                         | data.table                                      | DataFrames.jl                                | 
+|:----------------------------------|:------------------------------------------------|:---------------------------------------------|
+|Reduce multiple values             | `df[, list(mean(x))]`                           | `combine(df, :x => mean)`                    |
```
nalimilan, line 251: Also add space to this line and similar ones?

pstorozenko:
Have you considered writing e.g. `.(m = mean(x))` instead of `list(m = mean(x))` in `data.table`'s `j` and `by`? 
Even though `.()` is an alias for `list()`, I find it more how I'd write it and that's how they use it in [tutorial](https://cran.r-project.org/web/packages/data.table/vignettes/datatable-intro.html).
Just an idea :)

eloualiche:
> Have you considered writing e.g. `.(m = mean(x))` instead of `list(m = mean(x))` in `data.table`'s `j` and `by`?
> Even though `.()` is an alias for `list()`, I find it more how I'd write it and that's how they use it in [tutorial](https://cran.r-project.org/web/packages/data.table/vignettes/datatable-intro.html).
> Just an idea :)

We discussed this above. I just want to avoid aliases. This is not a guide for people wanting to switch to data.table, but rather targeted at people knowledgeable with data.table who would like to also use to DataFrames. 
I personally never use list, but I also find the dot syntax confusing when I am not the one writing it.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,66 @@ The table below compares more advanced commands:
...
+|                                   | `df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]` | `combine(df, ([:x, :y] .=> [maximum minimum])...)`                          |
+| Multivariate function             | `df[, list(cor(x,y)) ]`                                                                    | `transform(df, [:x, :y] => cor)`                                            |
+| Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform(df, [:x, :y] => ByRow(min))`                                     |
+|                                   | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`           | `transform(df, AsTable(r"^x") => ByRow(argmax))`                            |
+| DataFrame as output               | `df[, .SD[, list(value = min(x), max(x))] ]`                                               | `combine(df, :x => (x -> (value = [minimum(x), maximum(x)],)) => AsTable)`  |
```
bkamins, line 284: these operations do not produce the same result. Also I think that the DataFrames.jl part is overly complex, you could have just written `combine(df, :x => (x -> [minimum(x), maximum(x)]) => :value)` to get the same.

bkamins:
Looks good apart from the comments. Some `data.table` experts comments would be welcome.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,66 @@ The table below compares more advanced commands:
...
+                  z = c(3:7, NA), id = letters[1:6])
+df2 <- data.table(grp=c(1,3), w = c(10,11))                 
+```
+
+| Operation                          | data.table                                      | DataFrames.jl                                | 
+|:-----------------------------------|:------------------------------------------------|:---------------------------------------------|
+| Reduce multiple values             | `df[, list(mean(x))]`                           | `combine(df, :x => mean)`                    |
+| Add new columns                    | `df[, x_mean := mean(x) ]`                      | `transform(df, :x => mean => :x_mean)`       |
+| Rename column (in place)           | `setnames(df, "x", "x_new")`                    | `rename!(df, :x => :x_new)`                  |
+| Rename multiple columns (in place) | `setnames(df, c("x", "y"), c("x_new", "y_new")` | `rename!(df, [:x, :y] .=> [:x_new, :y_new])` |
```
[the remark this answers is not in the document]
bkamins: I am OK with what you proposed if you find it more readable.

docs/src/man/comparisons.md
eloualiche: I think you told me earlier to write `:x => (x -> (value = [minimum(x), maximum(x)],))` 
The goal of this line is to emulate how data.table can process sub-tables inside of a data.table. I thought this was similar to the written code in dataframes. 
I think writing `combine(df, :x => (x -> [minimum(x), maximum(x)]) => :value)` is a little different in spirit. 

I changed the code so it does give the right answer though.

bkamins: OK - we can leave it as is (the code is clear for me - I was unsure if it will be easy to read for people who learn DataFrames.jl)

bkamins:
OK - looks good, so let us wait for `data.table` experts to comment if they have anything to add. Thank you!

eloualiche:
> OK - looks good, so let us wait for `data.table` experts to comment if they have anything to add. Thank you!

Maybe the main developers of data.table will like chime in. 

cc: @mattdowle @ arunsrinivasan @jangorecki @MichaelChirico

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,66 @@ The table below compares more advanced commands:
...
+| Transform certain rows (in place) | `df[x<=0, x:=0 ]`                                                                          | `df.x[df.x .<= 0] .= 0`                                                     |
+| Transform several columns         | `df[, list(max(x), min(y)) ]`                                                              | `combine(df, :x => maximum, :y => minimum)`                                 |
+|                                   | `df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`                                          | `combine(df, [:x, :y] .=> mean)`                                            |
+|                                   | `df[, lapply(.SD, mean), .SDcols = patterns("x*") ]`                                       | `combine(df, names(df, r"^x") .=> mean)`                                    |
+|                                   | `df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]` | `combine(df, ([:x, :y] .=> [maximum minimum])...)`                          |
```
mattdowle, line 280: Looks like we have some work to do on this one. I can't think of an easier way right now. There may be an outstanding issue or pull request, maybe @jangorecki @MichaelChirico recall. I never wanted to encourage wide data, so my focus was on long. But I know people like to go wide like this, perhaps for presenting results in a paper or web page, so this task should be easier.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,66 @@ The table below compares more advanced commands:
...
+| Multivariate function             | `df[, list(cor(x,y)) ]`                                                                    | `transform(df, [:x, :y] => cor)`                                            |
+| Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform(df, [:x, :y] => ByRow(min))`                                     |
+|                                   | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`           | `transform(df, AsTable(r"^x") => ByRow(argmax))`                            |
+| DataFrame as output               | `df[, .SD[, list(value = c(min(x), max(x)) )] ]`                                           | `combine(df, :x => (x -> (value = [minimum(x), maximum(x)],)) => AsTable)`  |
```
mattdowle, line 284: `df[, .(c(min(x), max(x)))]`

mattdowle (commented):
Many thanks for pinging us. LGTM. Just a few suggested changes in this review.

Aside: on spacing, I tend to think that operators like `=` already visually separate alphanumeric characters so adding space around operators isn't necessary. In fact, to my eyes at least, the spaces lose some utility when used too much (i.e. in the way that is popular these days). I increase the utility of spaces by placing them between logical parts, and not always following a single rule. How I use spaces depends on the particular line of code, but I'm trying to group together parts. And sometimes vertically aligning common parts of two lines so you can see the differences between the two lines. So I've removed some spaces in the suggestions in this review, but the same thinking could be applied to the other queries too.
For example `df[, x:=NULL]` uses one space to separate the filter part (before the comma) from the _doing_ part (the `j`). There isn't any need for spaces around `:=` here because the non-alphanumeric characters in `:=` are already are distinguished from `x` and `NULL`. I think the whole spaces around operators thing came about because "use `<-`" was promoted, but I really prefer `=`.
Anyway, the spaces thinking is why I prefer `.()` to `list()`. I saw your comments about `list` and `.()` so I'm adding a bit more background. `list` is a set of 4 alphabetic characters which my brain has to process and realize that it's just there to permit a set of things; i.e., it's not a function or anything import. Whereas `.()` is more like a space that groups things together. When `list()` is used in all three parts inside `[...]` then that's 9 extra characters vs using `.()`. Those extra characters are alphabetic characters too, which make it slightly harder for the actual functions to jump out at the reader.
I made the changes in this review from `list()` to `.()` in these suggestions where there is another suggestion too, but all the `list()` could be changed to `.()`.

Thanks again for pinging us! Much appreciated.

docs/src/man/comparisons.md
eloualiche: I was trying to show a typical use case for .SD
I typically find it useful for non-standard function that I have to execute by groups (a regression for example). But that's the best example I could come up with. 
If you have an idea, let me know; if not I will likely remove this row because this has generated some confusion on the goal before.

mattdowle: Ok I see. One use-case for `.SD` that springs to mind is `DT[, .SD[1], by=group]`;  i.e. return the first row of every group.  Or, `DT[, .SD[which.max(someCol)], by=group]`; i.e. return the row in each group that has the biggest value in some column.

docs/src/man/comparisons.md
grantmcdermott: It's cheating a bit &mdash; or maybe not &mdash; but I'd probably use `dcast` here.

```r
dcast(df, .~., fun=list(min, max), value.var = c('x', 'y'))
```

The advantage of this approach is that it also scales well to cases where you want to collapse by group. I think the 'unlist' approach would struggle here.

```r
dcast(df, grp~., fun=list(min, max), value.var = c('x', 'y'))
```

Mind you, grouping is something that the DataFrames.jl implementation automatically supports (and, to @mattdowle's point, might be conceptually simpler my `dcast` workflow).

```julia
combine(groupby(df, :grp), ([:x, :y] .=> [minimum maximum])...)
```

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,66 @@ The table below compares more advanced commands:
...
+|:-----------------------------------|:-------------------------------------------------|:---------------------------------------------|
+| Reduce multiple values             | `df[, list(mean(x))]`                            | `combine(df, :x => mean)`                    |
+| Add new columns                    | `df[, x_mean := mean(x) ]`                       | `transform(df, :x => mean => :x_mean)`       |
+| Rename column (in place)           | `setnames(df, "x", "x_new")`                     | `rename!(df, :x => :x_new)`                  |
+| Rename multiple columns (in place) | `setnames(df, c("x", "y"), c("x_new", "y_new"))` | `rename!(df, [:x, :y] .=> [:x_new, :y_new])` |
+| Pick columns                       | `df[, list(x, y)]`                               | `select(df, :x, :y)`                         |
+| Remove columns                     | `df[, -"x" ]`                                    | `select(df, Not(:x))`                        |
+| Remove columns (in place)          | `df[, c("x") := NULL ]`                          | `select!(df, Not(:x))`                       |
```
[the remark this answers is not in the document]
eloualiche: I fixed this thanks. 
I always think about the case when you want to remove two variables. So I also added
`df[, c("x", "y"):=NULL ]`
for 
`select!(df, Not([:x, :y]))`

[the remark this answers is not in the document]
bkamins: OK. Please let me know when all is resolved. Thank you!

docs/src/man/comparisons.md
eloualiche: Thanks. I think I confused myself with the dataframes syntax. Here are your examples, it would be great if some of the dataframes could tell me if this is idiomatic

First element by group (usually comes after a sort)
```R
df[, .SD[1], by=grp]
```
```Julia
combine(groupby(df, :grp), names(df) .=> (x->x[1]) .=> names(df))
```

Or max by group:
```R
df[, .SD[which.max(x)], by=grp]
```
```Julia
subset(groupby(df, :grp), :x => (x->(x.==maximum(x))) )
```

I am not sure the last one is the best way of doing this. But that is also a use case for subset by group (which @matthieugomez raised [here](https://github.com/JuliaData/DataFrames.jl/issues/2740) recently)

bkamins: > `combine(groupby(df, :grp), names(df) .=> (x->x[1]) .=> names(df))`

is just `combine(groupby(df, :grp), first)`

> `subset(groupby(df, :grp), :x => (x->(x.==maximum(x))) )`
is probably more natural to write as `combine(groupby(df, :grp), sdf -> sdf[argmax(sdf.x), :])`

(note that in both cases the approach in data.table and in DataFrames.jl is conceptually the same (I have learned something 😄) only the syntax is a bit different)

eloualiche: That's amazing. I had no idea about `sdf`. Is this a reserved variable that stands for the df for each group within the scope of combine? I did not see it documented. 
That's exactly the kind of things I wanted to show. This will make the guide extra useful!

pdeffebach: No, the syntax `x -> f(x)` is used for creating an anonymous function. So `sdf` is argument to a lamdba.

bkamins: Just to expand on what @pdeffebach commented. We often use `sdf` to signal that this is a view of an original data frame, which is of `SubDataFrame` type.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,69 @@ The table below compares more advanced commands:
...
+df2 <- data.table(grp=c(1,3), w = c(10,11))                 
+```
+
+| Operation                          | data.table                                       | DataFrames.jl                                | 
+|:-----------------------------------|:-------------------------------------------------|:---------------------------------------------|
+| Reduce multiple values             | `df[, .(mean(x))]`                               | `combine(df, :x => mean)`                    |
+| Add new columns                    | `df[, x_mean:=mean(x) ]`                         | `transform(df, :x => mean => :x_mean)`       |
+| Rename column (in place)           | `setnames(df, "x", "x_new")`                     | `rename!(df, :x => :x_new)`                  |
+| Rename multiple columns (in place) | `setnames(df, c("x", "y"), c("x_new", "y_new"))` | `rename!(df, [:x, :y] .=> [:x_new, :y_new])` |
+| Pick columns                       | `df[, .(x, y)]`                                  | `select(df, :x, :y)`                         |
```
pstorozenko, line 256: How about adding a way of taking a single column as a vector?
I think it is sometimes useful and nicely corresponds between `data.table` and `DataFrames.jl`.
I'm only not sure either we should use here `!` or `:`, but I think view should be preferred. 
```
| Pick columns as dataframe | `df[, .(x, y)]` | `select(df, :x, :y)` |
| Pick column as a vector   | `df[, x]`       | `df[!, :x]`          |
```

docs/src/man/comparisons.md
eloualiche: Thanks. I got totally confused by the lambda argument name and thought somehow this was a special variable much like `.SD` 
I am still amazed that this ends up being so easy. Kudos to DataFrames!

docs/src/man/comparisons.md
eloualiche: Great suggestion! Implemented.

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,70 @@ The table below compares more advanced commands:
...
+library(data.table)
+df  <- data.table(grp = rep(1:2, 3), x = 6:1, y = 4:9,
+                  z = c(3:7, NA), id = letters[1:6])
+df2 <- data.table(grp=c(1,3), w = c(10,11))                 
+```
+
+| Operation                          | data.table                                       | DataFrames.jl                                | 
+|:-----------------------------------|:-------------------------------------------------|:---------------------------------------------|
+| Reduce multiple values             | `df[, .(mean(x))]`                               | `combine(df, :x => mean)`                    |
+| Add new columns                    | `df[, x_mean:=mean(x) ]`                         | `transform(df, :x => mean => :x_mean)`       |
```
MichaelChirico, line 253: elsewhere the in-place distinction is drawn -- IINM, it should be `transform!` for this row?

Two `data.table` ways to do it not in-place:

```
copy(df)[, x_mean:=mean(x)]
df[ , c(.SD, list(x_mean = mean(x))]
```

docs/src/man/comparisons.md
MichaelChirico: I would do `df[, c(lapply(.SD, max), lapply(.SD, min)), .SDcols = c("x", "y")]`. That _should_ `GForce` as well where the `unlist` one will not.

docs/src/man/comparisons.md
bkamins: the intention is that `data.table` and DataFrames.jl operations are doing the same. If `df[, x_mean:=mean(x) ]` is in-place, then we should use `transform!`, additional alternatives are:
* `df.x_mean = mean(df.x)`
* `insertcols!(df, :x_mean => mean(df.x))`

docs/src/man/comparisons.md
MichaelChirico: I guess it doesn't use GForce, and also, it results in duplicate names! ouch

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,70 @@ The table below compares more advanced commands:
...
+|                                   | `df[, lapply(.SD, mean), .SDcols = patterns("x*") ]`                                       | `combine(df, names(df, r"^x") .=> mean)`                                    |
+|                                   | `df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]` | `combine(df, ([:x, :y] .=> [maximum minimum])...)`                          |
+| Multivariate function             | `df[, .(cor(x,y)) ]`                                                                       | `transform(df, [:x, :y] => cor)`                                            |
+| Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform(df, [:x, :y] => ByRow(min))`                                     |
+|                                   | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("x*"), by = 1:nrow(df) ]`           | `transform(df, AsTable(r"^x") => ByRow(argmax))`                            |
```
MichaelChirico, line 286: left uses `x*`, right uses `^x`

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,70 @@ The table below compares more advanced commands:
...
+| DataFrame as output               | `df[, .SD[1], by=grp]`                                                                     | `combine(groupby(df, :grp), first)`                                         |
+| DataFrame as output               | `df[, .SD[which.max(x)], by=grp]`                                                          | `combine(groupby(df, :grp), sdf -> sdf[argmax(sdf.x), :])`                  |
```
MichaelChirico, line 288: FWIW I would usually do this as `df[order(-x), .SD[1], by=grp]`

docs/src/man/comparisons.md
MichaelChirico: yes, `:=` is always in-place

eloualiche: Thanks. That was a typo, it should have been `transform!`

docs/src/man/comparisons.md
eloualiche: Agreed. For this use case, it is probably more idiomatic. 
However, the goal here is to showcase a function that uses subdataframes. I am afraid that if we only use `first` and .SD[1]`, this might seem more limited than using actual function on `.SD`.

MichaelChirico: great! FWIW I ran into the exact same issue writing the .SD vignette:

https://stackoverflow.com/a/47406952/3576984

bkamins:
@eloualiche - I think it is high time to merge this PR and let the community review the merged docs on `main` (so that other contributors can work on their proposals without introducing merge conflicts).

Given this if you could please kindly:
1. resolve merge conflicts
2. comment if you agree to add the `lag` example (and if yes add it)

Then I will do a final review and merge. OK?

eloualiche:
> _How to add a lagged variable by group to a dataframe_ 

Thanks for chiming in, @floswald.

I don't use data.table shift because I was never sure how it handles different date formats (monthly/quarterly etc.), something that @matthieugomez solved in statar. 
Please correct me if I am wrong here as I have not used shift for dates for a long while and this might have changed.
I think we had a PR on ShiftedArrays at some point to work around dates but I don't think this was incorporated. 

I am happy to include this but we need to solve two things:
  1. dates: perhaps we should not insist on the panel feature because panels commonly involve dates. 
  2. how to include this new feature: a new section with new dataset, in which case it would probably best to include another language as well (surely adding this to stata can't be hard).

bkamins:
Given the complications - let us leave this out. I propose to resolve merge conflicts and then I merge this PR as is. Then I propose to shift a discussion on lagging to a separate PR as it looks as a more complex issue. OK?

floswald:
sounds good ! will you open an issue @bkamins or how shall this proceed?
thanks

bkamins:
Yes - please open a new PR after this is merged (so that you do not have to rebase it later). Thank you.

eloualiche:
I think I merged it. This should be all set now. Waiting for @floswald PR to revisit it!

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,70 @@ The table below compares more advanced commands:
...
+|                                   | `df[, lapply(.SD, mean), .SDcols = c("x", "y") ]`                                          | `combine(df, [:x, :y] .=> mean)`                                            |
+|                                   | `df[, lapply(.SD, mean), .SDcols = patterns("*x") ]`                                       | `combine(df, names(df, r"^x") .=> mean)`                                    |
+|                                   | `df[, unlist(lapply(.SD, function(x) c(max=max(x), min=min(x)))), .SDcols = c("x", "y") ]` | `combine(df, ([:x, :y] .=> [maximum minimum])...)`                          |
+| Multivariate function             | `df[, .(cor(x,y)) ]`                                                                       | `transform(df, [:x, :y] => cor)`                                            |
+| Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform(df, [:x, :y] => ByRow(min))`                                     |
```
bkamins, line 285:
```suggestion
| Row-wise                          | `df[, min_xy := min(x, y), by = 1:nrow(df)]`                                               | `transform!(df, [:x, :y] => ByRow(min))`                                    |
```

docs/src/man/comparisons.md
```diff
@@ -235,6 +235,70 @@ The table below compares more advanced commands:
...
+|                                   | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("*x"), by = 1:nrow(df) ]`           | `transform(df, AsTable(r"^x") => ByRow(argmax))`                            |
```
bkamins, line 286:
```suggestion
|                                   | `df[, argmax_xy := which.max(.SD) , .SDcols = patterns("*x"), by = 1:nrow(df) ]`           | `transform!(df, AsTable(r"^x") => ByRow(argmax))`                            |
```

bkamins:
I have made two fixes as I assume the `data.table` operation is in-place in these places. If this is correct please commit the changes and I will merge. Thank you!

bkamins (approved):

bkamins:
Thank you!

bkamins:
@floswald @pdeffebach @nalimilan - I have merged this to move forward. If you find something on the "comparisons" page on `main` that should be fixed could you please open a PR fixing this? Thank you!

PR #2727, opened 2021-04-18
Run `findall(rows)` only if `rows` are not all true

pstorozenko:
PR that resulted from #2724 disussion.
Addition of `_findall(rows::AbstractVector{Bool})`
to not run `findall(rows)` if `all(rows .== true)` in
`function SubDataFrame(parent::DataFrame, rows::AbstractVector{Bool}, cols)`

I'm not sure if `src/other/utils.jl` is a good location for this function. 

------------------

Benchmarks:
```julia
df = DataFrame(
    a = rand(1000000),
    b = allowmissing(rand(1000000)),
    c = [allowmissing(rand(1000000 - 1)); missing],
    d = repeat([1:9; missing], 100000)
)
```

Main branch
```
 julia> @btime dropmissing(df, :a, view=true);
  2.040 ms (15 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :b, view=true);
  2.071 ms (15 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :c, view=true);
  2.098 ms (15 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :d, view=true);
  1.911 ms (15 allocations: 6.99 MiB)

julia> @btime dropmissing(df, view=true);
  7.677 ms (37 allocations: 7.00 MiB)
```

Main + this fix
```
julia> @btime dropmissing(df, :a, view=true);
  388.342 μs (13 allocations: 126.64 KiB)

julia> @btime dropmissing(df, :b, view=true);
  399.634 μs (13 allocations: 126.64 KiB)

julia> @btime dropmissing(df, :c, view=true);
  2.103 ms (15 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :d, view=true);
  1.897 ms (15 allocations: 6.99 MiB)

julia> @btime dropmissing(df, view=true);
  7.679 ms (37 allocations: 7.00 MiB)
  ```
  
After `completecases` fix #2726 
```
julia> @btime dropmissing(df, :a, view=true);
  1.883 ms (10 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :b, view=true);
  2.075 ms (15 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :c, view=true);
  2.095 ms (15 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :d, view=true);
  1.905 ms (15 allocations: 6.99 MiB)

julia> @btime dropmissing(df, view=true);
  2.351 ms (28 allocations: 7.12 MiB)
```

After `completecases` fix #2726 + this fix

```
julia> @btime dropmissing(df, :a, view=true);
  251.012 μs (8 allocations: 122.38 KiB)

julia> @btime dropmissing(df, :b, view=true);
  399.373 μs (13 allocations: 126.64 KiB)

julia> @btime dropmissing(df, :c, view=true);
  2.109 ms (15 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :d, view=true);
  1.903 ms (15 allocations: 6.99 MiB)

julia> @btime dropmissing(df, view=true);
  2.355 ms (28 allocations: 7.12 MiB)
```

------------------

Summary:

* #2726 Reduces time of `dropmissing(df, view=true)`
* This fix reduces average time for `dropmissing(df, :a, view=true)`. In the worst case scenario (as with `:c` column) where there is only one missing in the end it is a bit slower, but I think that it is very rarely case. Usually we'll find missings/falses earlier.

### CODE REVIEW

File: src/other/utils.jl [hunk 1]
```diff
@@ -168,3 +168,11 @@ function _nt_like_hash(v, h::UInt)
 
     return xor(objectid(Tuple(propertynames(v))), h)
 end
+
+function _findall(v::AbstractVector{Bool})
+    if all(v)
```
bkamins, line 173: this is not what I have suggested.

The algorithm I think we should investigate is:
1. for safety reasons check that `firstindex(v) == 1` (which should always be the case)
2. locate first `true` in `v`
3. locate end of continuous block of `true` in `v` since start
4. check if there exists any `true` beyond this end identified in point 2

Now we have three scenarios:
1. the block of `true` is continuous and spans from `1` to `length(v)` - then return `Base.OneTo(length(v))`.
2. the block of `true` is continuous but has a shorter span then return `start:stop` of the block (identified above)
3. otherwise materialize `start:stop` range and start adding to it `trues` identified later (for this first finish doing `count(v)` from the point where we have started and then reuse the code from `findall` but taking into account the fact that we should not re-do work already done)

Is the algorithm clear?

(of course like with `completecases` caution must be taken to make sure that the operation we make is fast as maybe some optimizations can be made - e.g. if we want to use `_findall` only for `BitVector` maybe it will be better to process the array in chunks taking advantage of the storage mode of `BitVector` - this would have to be benchmarked)

If anything of what I write above is not clear please comment.

In general the point is that e.g. in your code you call `all(v)` and later `findall` will call `count(v)` anyway, so it is better to do `count(v)` and compare it to the length of `v` rather than redoing the work (what I propose is even more complex but I make this comment to show you the way we should think while designing the algorithm - avoid doing the same work twice).

bkamins: Just to be clear what I mean by taking advantage of low level storage of `BitVector`. If you write:
```
julia> x = trues(1000);

julia> x.chunks
16-element Vector{UInt64}:
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0xffffffffffffffff
 0x000000ffffffffff
```
you can super fast identify continuous blocks of `true`s or `false`s as you have 64 element (or 32 element on 32-bit machine) blocks that are very easy and fast to check. All falses are just `0` and all `trues` are `typemax`. (care should be taken with the first and last block in the continuous range which can be partially filled, but again - it should be easy to check using bit operations)

pstorozenko: Hmm, ok, now I get it, you meant fully custom `findall`.
I thought about `all(v)` as it's short-circuiting i.e. returns on the first encountered false so it's not that bad.
I can play with doing this optimizations, let's see what they bring.

bkamins: In general up to 1.0 release we mostly focused the development effort on API design in DataFrames.jl.

In post 1.0 development performance will take more priority. Therefore I would prefer to work on new implementations of functions in detail (so that we can have the best design we can think of and not have to come back to it later).

bkamins: Also - my approach does not have to be best. If you can come up with something else that is good then of course please go ahead and propose something. We do not need to rush with the changes.

bkamins: As a reference I have just checked how `all` is implemented for `BitArray` and it uses the approach using `chunks` I suggested above.

pstorozenko: I'm working on this custom `findall` function.
IINM depending on the case we want to return one of `Base.oneTo{Int64}`, `UnitRange{Int64}`, `Vector{Int64}`, right?

bkamins: Thinking about it now probably `UnitRange{Int}` and `Vector{Int}` should be enough. The performance benefit of `Base.OneTo{Int}` is probably negligible (if any) and it is better to have a smaller union (2 types) in return type, as it will be better handled by the compiler.

pstorozenko: Sure thing, thanks

pstorozenko:
I think this is more what you were talking about :)
I took `Base.findall(B::BitArray)` function and rewrote/adapted it for our needs.

In the first `while true` we process `B` in the optimized way.
After it we either returned solution or materialized result `I`.

In the second `while true` we process `B` in _almost_ the same way as in `Base.findall(B::BitArray)`.

I've tried to comment code a lot as operations are not obvious.
I've used one line ifs like `Bi == length(Bc) && return start:length(B)` because it makes code much cleaner, but I don't know if you like it.
Tell me what do you think about this whole approach.

In general this solution is sometimes a bit slower in case of random `B`.
This is because `Base.findall(B::BitArray)` does not try to _jump_ over blocks with trues only.
In case of bigger blocks of trues it is faster.
Of course in case of single, continuous block of trues it is much, much faster.

Bellow I attach some benchmarks:

```julia
BD = OrderedDict(
    "T Big" => trues(100000),
    "F Big" => falses(100000),
    "T64 F64" => [trues(64); falses(64)],
    "F64 T64" => [falses(64); trues(64)],
    "F80 T100" => [falses(85); trues(100)],
    "F256 T32" => [falses(256); trues(32)],
    "F260 T32" => [falses(260); trues(32)],
    "TF Big" => [trues(100000); falses(100000)],
    "FT Big" => [falses(100000);trues(100000)] ,

    # some edge cases
    "TFT small" => [trues(85); falses(100); trues(85)],
    "FTFFT small" => [falses(64 + 32); trues(32); falses(128); trues(32)],
    "TFTF small" => [falses(64); trues(64); falses(64); trues(64)],
    "TFT small" => [trues(64); falses(10); trues(100)],

    "FTF Big" => [falses(8500); trues(100000); falses(65000)],
    "TFT Big" => [trues(8500); falses(100000); trues(65000)],
    "FTFTFTF Big" => [falses(65000); trues(65000); falses(65000); trues(65000); falses(65000); trues(65000); falses(65000)],

    "FTFR small" => [falses(85); trues(100); falses(65); rand([true, false], 20)],
    "R Big" => BitVector(rand([true, false], 200000)),
    "RF Big" => [BitVector(rand([true, false], 100000)) ; falses(100000)],
    "RT Big" => [BitVector(rand([true, false], 100000)) ; trues(100000)],
    "FTFR Big" => [falses(65000);  trues(65000);  falses(65000); rand([true, false], 20000)],
    "T256 R100" => [trues(256);  rand([true, false], 100)],
    "F256 R100" => [falses(256); rand([true, false], 100)],
)

for (l, B) in BD
    println("Processing $l")
    @show Base.findall(B) == DataFrames._findall(B)
    @btime Base.findall($B)
    @btime DataFrames._findall($B)
end
```

```
Processing T Big
Base.findall(B) == DataFrames._findall(B) = true
  109.098 μs (2 allocations: 781.33 KiB)
  398.537 ns (0 allocations: 0 bytes)
Processing F Big
Base.findall(B) == DataFrames._findall(B) = true
  430.095 ns (1 allocation: 80 bytes)
  421.095 ns (1 allocation: 80 bytes)
Processing T64 F64
Base.findall(B) == DataFrames._findall(B) = true
  104.586 ns (1 allocation: 624 bytes)
  41.959 ns (1 allocation: 80 bytes)
Processing F64 T64
Base.findall(B) == DataFrames._findall(B) = true
  104.734 ns (1 allocation: 624 bytes)
  41.498 ns (1 allocation: 80 bytes)
Processing F80 T100
Base.findall(B) == DataFrames._findall(B) = true
  143.548 ns (1 allocation: 896 bytes)
  46.941 ns (1 allocation: 80 bytes)
Processing F256 T32
Base.findall(B) == DataFrames._findall(B) = true
  73.466 ns (1 allocation: 336 bytes)
  45.835 ns (1 allocation: 80 bytes)
Processing F260 T32
Base.findall(B) == DataFrames._findall(B) = true
  73.387 ns (1 allocation: 336 bytes)
  45.847 ns (1 allocation: 80 bytes)
Processing TF Big
Base.findall(B) == DataFrames._findall(B) = true
  105.403 μs (2 allocations: 781.33 KiB)
  1.467 μs (1 allocation: 80 bytes)
Processing FT Big
Base.findall(B) == DataFrames._findall(B) = true
  105.608 μs (2 allocations: 781.33 KiB)
  1.987 μs (1 allocation: 80 bytes)
Processing TFT small
Base.findall(B) == DataFrames._findall(B) = true
  217.728 ns (1 allocation: 1.45 KiB)
  188.086 ns (2 allocations: 1.41 KiB)
Processing FTFFT small
Base.findall(B) == DataFrames._findall(B) = true
  106.897 ns (1 allocation: 624 bytes)
  112.246 ns (2 allocations: 624 bytes)
Processing TFTF small
Base.findall(B) == DataFrames._findall(B) = true
  215.025 ns (1 allocation: 1.14 KiB)
  160.212 ns (2 allocations: 1.14 KiB)
Processing FTF Big
Base.findall(B) == DataFrames._findall(B) = true
  127.988 μs (2 allocations: 781.33 KiB)
  1.767 μs (1 allocation: 80 bytes)
Processing TFT Big
Base.findall(B) == DataFrames._findall(B) = true
  95.470 μs (2 allocations: 574.33 KiB)
  66.360 μs (2 allocations: 574.39 KiB)
Processing FTFTFTF Big
Base.findall(B) == DataFrames._findall(B) = true
  253.018 μs (2 allocations: 1.49 MiB)
  194.054 μs (2 allocations: 1.49 MiB)
Processing FTFR small
Base.findall(B) == DataFrames._findall(B) = true
  189.834 ns (1 allocation: 1008 bytes)
  153.568 ns (2 allocations: 976 bytes)
Processing R Big
Base.findall(B) == DataFrames._findall(B) = true
  153.825 μs (2 allocations: 778.89 KiB)
  119.013 μs (2 allocations: 778.95 KiB)
Processing RF Big
Base.findall(B) == DataFrames._findall(B) = true
  78.821 μs (2 allocations: 389.64 KiB)
  59.672 μs (2 allocations: 389.70 KiB)
Processing RT Big
Base.findall(B) == DataFrames._findall(B) = true
  204.643 μs (2 allocations: 1.15 MiB)
  165.110 μs (2 allocations: 1.15 MiB)
Processing FTFR Big
Base.findall(B) == DataFrames._findall(B) = true
  100.556 μs (2 allocations: 586.64 KiB)
  72.579 μs (2 allocations: 586.70 KiB)
Processing T256 R100
Base.findall(B) == DataFrames._findall(B) = true
  659.301 ns (1 allocation: 2.56 KiB)
  519.550 ns (2 allocations: 2.58 KiB)
Processing F256 R100
Base.findall(B) == DataFrames._findall(B) = true
  120.717 ns (1 allocation: 544 bytes)
  131.407 ns (2 allocations: 528 bytes)
```

test/utils.jl
```diff
@@ -125,4 +125,41 @@ end
...
+        "TFT Big" => ([trues(8500); falses(100000); trues(65000)], Vector{Int}),
+        "FTFTFTF Big" => ([falses(65000); trues(65000); falses(65000); trues(65000); falses(65000); trues(65000); falses(65000)], Vector{Int}),
+
+        "FTFR small" => ([falses(85); trues(100); falses(65); rand([true, false], 40)], Vector{Int}),
+        "R Big" => (BitVector(rand([true, false], 200000)), Vector{Int}),
+        "RF Big" => ([BitVector(rand([true, false], 100000)) ; falses(100000)], Vector{Int}),
+        "RT Big" => ([BitVector(rand([true, false], 100000)) ; trues(100000)], Vector{Int}),
+        "FTFR Big" => ([falses(65000);  trues(65000);  falses(65000); rand([true, false], 20000)], Vector{Int}),
+        "T256 R100" => ([trues(256);  rand([true, false], 100)], Vector{Int}),
+        "F256 R100" => ([falses(256); rand([true, false], 100)], Vector{Int}),
```
bkamins, line 156: it would be good to have some more randomized tests I think. I will review the code in the meantime. Thank you!

bkamins:
Could you please also resolve merge conflicts so that we can run the CI? Thank you!

test/utils.jl
pstorozenko: I added some more random tests, tell me if that's what you meant.
`_findall` is constantly a faster here.
```
Processing R Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  1.350 ms (2 allocations: 7.63 MiB)
  1.317 ms (2 allocations: 7.63 MiB)
Processing RF Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  830.955 μs (2 allocations: 3.81 MiB)
  659.672 μs (2 allocations: 3.81 MiB)
Processing RT Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  2.221 ms (2 allocations: 11.44 MiB)
  1.958 ms (2 allocations: 11.44 MiB)
Processing FR Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  819.751 μs (2 allocations: 3.81 MiB)
  644.985 μs (2 allocations: 3.81 MiB)
Processing TR Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  2.193 ms (2 allocations: 11.44 MiB)
  1.922 ms (2 allocations: 11.44 MiB)
Processing FRT Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  2.225 ms (2 allocations: 11.45 MiB)
  1.960 ms (2 allocations: 11.45 MiB)
Processing TRF Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  2.224 ms (2 allocations: 11.45 MiB)
  1.932 ms (2 allocations: 11.45 MiB)
Processing FRF Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  840.763 μs (2 allocations: 3.82 MiB)
  650.147 μs (2 allocations: 3.82 MiB)
Processing TRT Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  3.611 ms (2 allocations: 19.07 MiB)
  3.254 ms (2 allocations: 19.07 MiB)
Processing RFR Big
(Base.findall(B) == res, typeof(res) == T) = (true, true)
  1.659 ms (2 allocations: 7.63 MiB)
  1.320 ms (2 allocations: 7.63 MiB)
```

bkamins: Yes - this is what I meant - to be sure that all works in a "scattered data" case correctly and fast.

bkamins:
Yes - I have checked this also :). It seems that probably the compiler somehow optimizes these lines out (not sure though).

src/other/utils.jl
```diff
@@ -178,3 +178,133 @@ function _nt_like_hash(v, h::UInt)
 
     return xor(objectid(Tuple(propertynames(v))), h)
 end
+
+_findall(B) = findall(B)
+
+@inline _blsr(x)= x & (x-1) 
+const _msk64 = ~UInt64(0)
```
bkamins, line 185: Use capital letters for constants.

Also this seems to be a bit more readable (but if you feel the `~` is better please revert)
```suggestion
const _MSK64 = typemax(UInt64)
```

src/other/utils.jl
```diff
@@ -178,3 +178,133 @@ function _nt_like_hash(v, h::UInt)
...
+            if start != -1 && stop == -1
+                stop = i1 - 1
+            end
+            while c == 0 # no need to return here as we returned above
+                i1 += 64
+                Bi += 1
+                c = Bc[Bi]
+            end
+        end
+        if c == _msk64
```
bkamins, line 216:
```suggestion
        if c == _MSK64
```

src/other/utils.jl
```diff
@@ -178,3 +178,133 @@ function _nt_like_hash(v, h::UInt)
...
+        while c == _msk64
+            I[i:i+64-1] .= i1:(i1+64-1)
+            Bi == length(Bc) && return I
+            i += 64
+            i1 += 64
+            Bi += 1
+            c = Bc[Bi]
+        end
+
+        while c!= 0
```
bkamins, line 303:
```suggestion
        while c != 0
```

src/subdataframe/subdataframe.jl
```diff
@@ -96,7 +96,7 @@ Base.@propagate_inbounds function SubDataFrame(parent::DataFrame, rows::Abstract
         throw(ArgumentError("invalid length of `AbstractVector{Bool}` row index " *
                             "(got $(length(rows)), expected $(nrow(parent)))"))
     end
-    return SubDataFrame(parent, findall(rows), cols)
+    return SubDataFrame(parent, _findall(rows), cols)
```
bkamins, line 99: Can you also test performance and corrected of `view` (it should be OK, but just to be sure as this is the point of the change we make so I prefer to double check). Thank you!

src/other/utils.jl
```diff
@@ -178,3 +178,133 @@ function _nt_like_hash(v, h::UInt)
 
     return xor(objectid(Tuple(propertynames(v))), h)
 end
+
+_findall(B) = findall(B)
+
+@inline _blsr(x)= x & (x-1)
```
bkamins, line 184: Also please make sure you do not leave trailing whitespace in other places
```suggestion
@inline _blsr(x)= x & (x-1)

```

src/other/utils.jl
```diff
@@ -178,3 +178,133 @@ function _nt_like_hash(v, h::UInt)
...
+@inline _blsr(x)= x & (x-1) 
+const _msk64 = ~UInt64(0)
+
+# findall optimized for B::BitVector
+# the main idea from Base.findall(B::BitArray)
+function _findall(B::BitVector)::Union{UnitRange{Int}, Vector{Int}}
+    nnzB = count(B)
+    nnzB == 0 && return Int[]
+    nnzB == length(B) && return 1:length(B)
+    I = Vector{Int}()
```
bkamins, line 193: you can also move it up to return `I` in line 191 for consistency.
```suggestion
    I = Vector{Int}(undef, nnzB)
```

bkamins, line 193: why not delaying defining it until we know we need it and then do `I = Vector{Int}(undef, nnzB)`?

src/other/utils.jl
```diff
@@ -178,3 +178,133 @@ function _nt_like_hash(v, h::UInt)
...
+        i > nnzB && return start:start + i - 2 # all ones in B found
+
+        if c == 0
+            if start != -1 && stop == -1
+                stop = i1 - 1
+            end
+            while c == 0 # no need to return here as we returned above
+                i1 += 64
+                Bi += 1
+                c = Bc[Bi]
```
bkamins, line 213: are you sure this is always `@inbounds`?

bkamins: ah - you are sure as if we went out of bounds we would return in line 204 - right?

pstorozenko: Exactly, if we already found all trues, we leave in 204, otherwise we will find some trues later, not in this loop.

test/utils.jl
```diff
@@ -149,4 +149,49 @@ end
     @test_throws AssertionError DataFrames.split_to_chunks(10, 11)
 end
 
+@testset "_findall(B::BitVector)" begin
```
bkamins, line 152:
```suggestion
@testset "_findall(B::BitVector)" begin
    Random.seed!(1234)
```

bkamins:
@nalimilan - after all my comments are resolved and CI passes could you please have a look (I have checked the logic of the whole code in detail and we have good test coverage so it is rather safe, but you always have a keen eye for problems). I am OK to merge the PR.

@pstorozenko - a very good PR

src/other/utils.jl
```diff
@@ -178,3 +178,133 @@ function _nt_like_hash(v, h::UInt)
...
+    Bi = 1 # block index
+    i1 = 1 # index of current block beginng in B
+    i = 1  # index of the _next_ one in I
+    c = Bc[1] # current block
+
+    start = -1 # the begining of ones block
+    stop = -1  # the end of ones block
+    
+    @inbounds while true # I not materialized
+        i > nnzB && return start:start + i - 2 # all ones in B found
```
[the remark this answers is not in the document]
pstorozenko: Regarding `@assert`, I'd change all
`Bi == length(Bc) && return start:stop`
into
``` 
if Bi == length(Bc) 
    res = start:stop
    @assert length(res) == nnzB
    return res
end
```
and `Bi == length(Bc) && return I`
into
```
if Bi == length(Bc) 
    @assert nnzB == i - 1
    return I
end
```
Right?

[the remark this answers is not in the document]
bkamins: yes - something like this (if this covers all cases :)). Just make sure that `res` is not a name used for something else in the code (probably not, but such all purpose names are a common source of type instabilities).

test/utils.jl
```diff
@@ -149,4 +149,68 @@ end
...
+    Random.seed!(1234)
```
bkamins, line 153: we need to add `using Random` on top of the module - I thought was already added.

src/subdataframe/subdataframe.jl
pstorozenko: Fortunately it is faster :)
```
df = DataFrame(
    a = rand(1000000),
    b = allowmissing(rand(1000000)),
    c = [allowmissing(rand(1000000 - 1)); missing],
    d = repeat([1:9; missing], 100000),
    e = repeat([rand(50000); missings(50000)], 10),
    f = [missings(300000); ones(400000); missings(300000)],
    g = missings(1000000)
)
```

Main:
```
julia> @btime dropmissing(df, :a, view=true);
  1.872 ms (9 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :b, view=true);
  2.075 ms (14 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :c, view=true);
  2.111 ms (14 allocations: 7.75 MiB)

julia> @btime dropmissing(df, :d, view=true);
  1.913 ms (14 allocations: 6.99 MiB)

julia> @btime dropmissing(df, :e, view=true);
  1.117 ms (14 allocations: 3.94 MiB)

julia> @btime dropmissing(df, :f, view=true);
  909.713 μs (14 allocations: 3.18 MiB)
 
julia> @btime dropmissing(df, :g, view=true);
  152.216 μs (13 allocations: 126.67 KiB)

julia> @btime dropmissing(df, view=true);
  1.432 ms (38 allocations: 1.63 MiB)
```


Main + this PR:

```
julia> @btime dropmissing(df, :a, view=true);
  17.093 μs (7 allocations: 122.34 KiB)

julia> @btime dropmissing(df, :b, view=true);
  162.656 μs (12 allocations: 126.59 KiB)

julia> @btime dropmissing(df, :c, view=true);
  168.988 μs (12 allocations: 126.59 KiB)

julia> @btime dropmissing(df, :d, view=true);
  1.890 ms (14 allocations: 6.99 MiB)

julia> @btime dropmissing(df, :e, view=true);
  1.100 ms (14 allocations: 3.94 MiB)

julia> @btime dropmissing(df, :f, view=true);
  166.613 μs (12 allocations: 126.59 KiB)

julia> @btime dropmissing(df, :g, view=true);
  188.653 μs (13 allocations: 126.67 KiB)
  
julia> @btime dropmissing(df, view=true);
  1.410 ms (38 allocations: 1.63 MiB)
```

test/utils.jl
pstorozenko: Sure

src/other/utils.jl
```diff
@@ -178,3 +178,155 @@ function _nt_like_hash(v, h::UInt)
...
+            end
+        end
+        if c == _msk64
+            if stop != -1
+                I = Vector{Int}(undef,nnzB)
+                I[1:i-1] .= start:stop
+                break
+            end
+            if start == -1
+                start = i1
```
bkamins, line 227: this should be covered by the tests where first `true` is in the boundary of the chunk - right?

bkamins (approved):

src/other/utils.jl
```diff
@@ -178,3 +178,155 @@ function _nt_like_hash(v, h::UInt)
 
     return xor(objectid(Tuple(propertynames(v))), h)
 end
+
+_findall(B) = findall(B)
+
+@inline _blsr(x)= x & (x-1)
+const _msk64 = ~UInt64(0)
```
nalimilan, line 185: Is it really useful to define this constant? I would think using `~UInt64(0)` directly in the function is OK as the compiler should easily optimize it out.

src/other/utils.jl
```diff
@@ -178,3 +178,155 @@ function _nt_like_hash(v, h::UInt)
...
+
+# findall optimized for B::BitVector
+# the main idea from Base.findall(B::BitArray)
```
nalimilan, line 188:
```suggestion
# findall returning a range when possible (all true indices are contiguous), and optimized for B::BitVector
# the main idea is taken from Base.findall(B::BitArray)
```

src/other/utils.jl
```diff
@@ -178,3 +178,155 @@ function _nt_like_hash(v, h::UInt)
...
+function _findall(B::BitVector)::Union{UnitRange{Int}, Vector{Int}}
+    nnzB = count(B)
+    nnzB == 0 && return Int[]
```
nalimilan, line 191: How about this instead?
```suggestion
    nnzB == 0 && return 1:0
```

src/other/utils.jl
```diff
@@ -178,3 +178,155 @@ function _nt_like_hash(v, h::UInt)
...
+        end
+
+        while c != 0
+            tz = trailing_zeros(c)
+            c = _blsr(c) # zeros last nonzero bit
+            I[i] = i1 + tz
+            i += 1
+        end
+    end
+end
```
nalimilan, line 332:
```suggestion
    @assert false "should not be reached"
end
```

src/other/utils.jl
```diff
@@ -178,3 +178,155 @@ function _nt_like_hash(v, h::UInt)
 
     return xor(objectid(Tuple(propertynames(v))), h)
 end
+
+_findall(B) = findall(B)
+
+@inline _blsr(x)= x & (x-1)
```
nalimilan, line 184:
```suggestion
@inline _blsr(x) = x & (x-1)
```

nalimilan (commented):
Thanks! I haven't checked the logic as the body of the function is scary. Can you just fix the uncovered line?

src/other/utils.jl
bkamins: I also thought this, however, this is the way Julia Base does this (but yeah - let us move it)

pstorozenko: Ok, sure :)

src/other/utils.jl
pstorozenko: I felt `Int[]` is not a good way, but I couldn't find better, thanks :)

pstorozenko (commented):
Asserts added

pstorozenko:
@nalimilan What do you mean by _fix the uncovered line_?
Do you mean code coverage? I'm sure all lines (except of `@assert false "should not be reached"`) are covered by tests (I tested with good old `@show`), but for some reason codecov is showing differently.

bkamins:
> but for some reason codecov is showing differently.

I have also checked the logic that the important lines are covered.

nalimilan (approved):
Thanks!

bkamins:
Thank you!

PR #2749, opened 2021-05-05
fix performance issue in multirow split-apply-combine

bkamins:
Resolves https://discourse.julialang.org/t/allocations-and-slow-perf-for-transform-on-groupeddataframes/60594

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+# DataFrames.jl v1.1.1 Patch Release Notes
+
+## Performance improvements
+
+* fix performance issue when aggregation function produces multiple rows
+  in split-apply-combine
+  ([XXXX](https://github.com/JuliaData/DataFrames.jl/pull/XXXX))
+
 # DataFrames.jl v1.1 Release Notes
 
 ## Functionality changes
--- a/src/groupeddataframe/callprocessing.jl
+++ b/src/groupeddataframe/callprocessing.jl
@@ -90,7 +90,7 @@ end
 function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
                  starts::AbstractVector{<:Integer}, ends::AbstractVector{<:Integer},
                  gd::GroupedDataFrame, incols::Tuple{AbstractVector}, i::Integer)
-    idx = idx[starts[i]:ends[i]]
+    idx = view(idx, starts[i]:ends[i])
     return f(view(incols[1], idx))
 end
 
--- a/src/groupeddataframe/callprocessing.jl
+++ b/src/groupeddataframe/callprocessing.jl
@@ -97,21 +97,21 @@ end
 function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
                  starts::AbstractVector{<:Integer}, ends::AbstractVector{<:Integer},
                  gd::GroupedDataFrame, incols::NTuple{2, AbstractVector}, i::Integer)
-    idx = idx[starts[i]:ends[i]]
+    idx = view(idx, starts[i]:ends[i])
     return f(view(incols[1], idx), view(incols[2], idx))
 end
 
 function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
                  starts::AbstractVector{<:Integer}, ends::AbstractVector{<:Integer},
                  gd::GroupedDataFrame, incols::NTuple{3, AbstractVector}, i::Integer)
-    idx = idx[starts[i]:ends[i]]
+    idx = view(idx, starts[i]:ends[i])
     return f(view(incols[1], idx), view(incols[2], idx), view(incols[3], idx))
 end
 
 function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
                  starts::AbstractVector{<:Integer}, ends::AbstractVector{<:Integer},
                  gd::GroupedDataFrame, incols::NTuple{4, AbstractVector}, i::Integer)
-    idx = idx[starts[i]:ends[i]]
+    idx = view(idx, starts[i]:ends[i])
     return f(view(incols[1], idx), view(incols[2], idx), view(incols[3], idx),
              view(incols[4], idx))
 end
@@ -119,7 +119,7 @@ end
 function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
                  starts::AbstractVector{<:Integer}, ends::AbstractVector{<:Integer},
                  gd::GroupedDataFrame, incols::Tuple, i::Integer)
-    idx = idx[starts[i]:ends[i]]
+    idx = view(idx, starts[i]:ends[i])
     return f(map(c -> view(c, idx), incols)...)
 end
 
@@ -129,7 +129,7 @@ function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
     if f isa ByRow && isempty(incols)
         return [f.fun(NamedTuple()) for _ in 1:(ends[i] - starts[i] + 1)]
     else
-        idx = idx[starts[i]:ends[i]]
+        idx = view(idx, starts[i]:ends[i])
         return f(map(c -> view(c, idx), incols))
     end
 end
@@ -137,6 +137,6 @@ end
 function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
                  starts::AbstractVector{<:Integer}, ends::AbstractVector{<:Integer},
                  gd::GroupedDataFrame, incols::Nothing, i::Integer)
-    idx = idx[starts[i]:ends[i]]
+    idx = view(idx, starts[i]:ends[i])
     return f(view(parent(gd), idx, :))
 end
--- a/src/groupeddataframe/complextransforms.jl
+++ b/src/groupeddataframe/complextransforms.jl
@@ -30,6 +30,7 @@ function _combine_with_first((first,)::Ref{Any},
     @assert first isa Union{NamedTuple, DataFrameRow, AbstractDataFrame}
     extrude = false
 
+    lgd = length(gd)
     if first isa AbstractDataFrame
         n = 0
         eltys = eltype.(eachcol(first))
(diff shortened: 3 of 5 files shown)


src/groupeddataframe/splitapplycombine.jl
```diff
@@ -682,6 +667,23 @@ function _combine(gd::GroupedDataFrame,
     return idx, DataFrame(outcols, nms, copycols=false)
 end
 
+function reorder_cols!(trans_res, i, col, col_idx, keeprows, idx_keeprows, gd)
```
nalimilan, line 670: Make signature more specific?

NEWS.md
```diff
@@ -1,3 +1,11 @@
+# DataFrames.jl v1.1.1 Patch Release Notes
+
+## Performance improvements
+
+* fix performance issue when aggregation function produces multiple rows
+  in split-apply-combine
+  ([XXXX](https://github.com/JuliaData/DataFrames.jl/pull/XXXX))
```
bkamins, line 7:
```suggestion
  ([2749](https://github.com/JuliaData/DataFrames.jl/pull/2749))
```

src/groupeddataframe/complextransforms.jl
```diff
@@ -30,31 +30,37 @@ function _combine_with_first((first,)::Ref{Any},
...
         end
     end
     idx = idx_agg === NOTHING_IDX_AGG ? Vector{Int}(undef, n) : idx_agg
+    sizehint!(idx, lgd)
     local initialcols
     let eltys=eltys, n=n # Workaround for julia#15276
-        initialcols = ntuple(i -> Tables.allocatecolumn(eltys[i], n), _ncol(first))
+        initialcols = ntuple(_ncol(first)) do i
+            res = Tables.allocatecolumn(eltys[i], n)
+            sizehint!(res, lgd)
```
bkamins, line 61: most likely we will have at least 1 row per group

src/groupeddataframe/callprocessing.jl
```diff
@@ -90,7 +90,7 @@ end
 function do_call(f::Base.Callable, idx::AbstractVector{<:Integer},
                  starts::AbstractVector{<:Integer}, ends::AbstractVector{<:Integer},
                  gd::GroupedDataFrame, incols::Tuple{AbstractVector}, i::Integer)
-    idx = idx[starts[i]:ends[i]]
+    idx = view(idx, starts[i]:ends[i])
```
bkamins, line 93: saves allocations. I left allocating code for more passed columns as I was unsure what is faster - allocate indices or have a doubly nested `view`.

bkamins:
I was unable to go lower than 8 allocations per group :sob: (which is bad for many groups).

NEWS.md
```diff
@@ -1,3 +1,11 @@
+# DataFrames.jl v1.1.1 Patch Release Notes
+
+## Performance improvements
+
+* fix performance issue when aggregation function produces multiple rows
+  in split-apply-combine
+  ([2749](https://github.com/JuliaData/DataFrames.jl/pull/2749))
```
bkamins, line 7:
```suggestion
  ([2749](https://github.com/JuliaData/DataFrames.jl/pull/2749))
* `completecases` is now optimized and only processes columns that
   can contain missing values; additionally it is now type stable and
   always returns a `BitVector`
   ([#2726](https://github.com/JuliaData/DataFrames.jl/pull/2726))
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -543,7 +543,7 @@ function _combine(gd::GroupedDataFrame,
         end
         idx_keeprows = prepare_idx_keeprows(gd.idx, gd.starts, gd.ends, nrow(parent(gd)))
     else
-        idx_keeprows = nothing
+        idx_keeprows = Int[]
```
bkamins, line 546: this reduces specialization; if `keeprows` is `false` we never use `idx_keeprows` in processing.

NEWS.md
```diff
@@ -1,3 +1,15 @@
...
+* `completecases` is now optimized and only processes columns that
+   can contain missing values; additionally it is now type stable and
+   always returns a `BitVector`
+   ([#2726](https://github.com/JuliaData/DataFrames.jl/pull/2726))
```
bkamins, line 11:
```suggestion
   ([#2726](https://github.com/JuliaData/DataFrames.jl/pull/2726))
* fix performance bottleneck when displaying wide tables
  ([#2750](https://github.com/JuliaData/DataFrames.jl/pull/2750))
```

NEWS.md
```diff
@@ -1,3 +1,17 @@
...
```
bkamins, line 11:
```suggestion
* `completecases` is now optimized and only processes columns that
  can contain missing values; additionally it is now type stable and
  always returns a `BitVector`
  ([#2726](https://github.com/JuliaData/DataFrames.jl/pull/2726))
```

bkamins:
do not merge this PR before https://github.com/JuliaData/DataFrames.jl/pull/2750 is merged (as I update NEWS.md in this PR to avoid having to resolve merge conflicts later)

src/groupeddataframe/callprocessing.jl
bkamins: I did some more benchmarking and I change the code to use `view` in all cases.

nalimilan: I'd think that views based on a range of indices are quite efficient, but only benchmarking can tell.

src/groupeddataframe/complextransforms.jl
```diff
@@ -30,31 +30,36 @@ function _combine_with_first((first,)::Ref{Any},
...
     end
     idx = idx_agg === NOTHING_IDX_AGG ? Vector{Int}(undef, n) : idx_agg
+    sizehint!(idx, lgd)
     local initialcols
     let eltys=eltys, n=n # Workaround for julia#15276
-        initialcols = ntuple(i -> Tables.allocatecolumn(eltys[i], n), _ncol(first))
+        initialcols = ntuple(_ncol(first)) do i
+            res = Tables.allocatecolumn(eltys[i], n)
+            return res
+        end
```
nalimilan, line 62: What's the point of this change? The new form is much more verbose.

src/groupeddataframe/complextransforms.jl
```diff
@@ -397,7 +402,16 @@ function _combine_tables_with_first!(first::Union{AbstractDataFrame,
             return _combine_tables_with_first!(rows, newcols, idx, i, j,
                                                f, gd, incols, colnames, firstmulticol)
         end
-        append!(idx, Iterators.repeated(gdidx[starts[i]], _nrow(rows)))
+        append_const!(idx, gdidx[starts[i]], _nrow(rows))
     end
     return outcols, colnames
 end
+
+@inline function append_const!(idx::Vector{Int}, val, growsize::Int)
```
nalimilan, line 410: Isn't `val` always `::Integer`?

Also, maybe `@inlined` isn't worth it given the cost of resizing? That could reduce the cost of compiling the caller function, which is specialized on column types.

src/groupeddataframe/callprocessing.jl
bkamins: The performance difference is small, but it will be more GC friendly (with a large number of groups it is a bit better to use `view` with a small number of groups it is a bit better to materialize the `idx`).

PR after the change:
```
julia> df = DataFrame(x=rand(1:10^6, 10^7), v1=rand(10^7), v2=rand(10^7));

julia> gdf = groupby(df, :x);

julia> f(v1,v2) = cor(v1, v2)^2

julia> @btime combine($gdf, [:v1, :v2] => f => :r2);
  932.859 ms (1000270 allocations: 179.30 MiB)

julia> df = DataFrame(x=rand(1:10, 10^7), v1=rand(10^7), v2=rand(10^7));

julia> gdf = groupby(df, :x);

julia> f(v1,v2) = cor(v1, v2)^2

julia> @btime combine($gdf, [:v1, :v2] => f => :r2);
  287.085 ms (321 allocations: 76.31 MiB)
```

current release:
```
julia> df = DataFrame(x=rand(1:10^6, 10^7), v1=rand(10^7), v2=rand(10^7));

julia> gdf = groupby(df, :x);

julia> f(v1,v2) = cor(v1, v2)^2

julia> @btime combine($gdf, [:v1, :v2] => f => :r2);
  810.070 ms (315 allocations: 22.91 MiB)

julia> df = DataFrame(x=rand(1:10, 10^7), v1=rand(10^7), v2=rand(10^7));

julia> gdf = groupby(df, :x);

julia> f(v1,v2) = cor(v1, v2)^2

julia> @btime combine($gdf, [:v1, :v2] => f => :r2);
  269.337 ms (302 allocations: 18.81 KiB)
```

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -653,24 +653,9 @@ function _combine(gd::GroupedDataFrame,
...
-                for (s, e) in zip(gd.starts, gd.ends)
-                    for j in s:e
-                        k += 1
-                        newcol[gd_idx[j]] = col[k]
-                    end
-                end
-                @assert k == length(gd_idx)
-                trans_res[i] = TransformationResult(col_idx, newcol, trans_res[i].name, trans_res[i].optional)
-            end
+        let i=i
```
nalimilan, line 656: How about using `$i` instead?

src/groupeddataframe/splitapplycombine.jl
bkamins: done

src/groupeddataframe/complextransforms.jl
bkamins: I had to remove this as CategoricalArrays.jl does not support `sizehint!`.

src/groupeddataframe/splitapplycombine.jl
bkamins: I was not sure the Julia 1.0 version of the macro supports `$i`. And for sure Julia 1.1, 1.2 and 1.3 do not support it.

src/groupeddataframe/complextransforms.jl
bkamins: yes - now it is not needed; I will remove (I wanted to do `sizehint!` but it does not work always)

src/groupeddataframe/complextransforms.jl
bkamins: OK - changed. `val` is always `Int` now.

src/groupeddataframe/complextransforms.jl
```diff
@@ -30,28 +30,30 @@ function _combine_with_first((first,)::Ref{Any},
...
     else # other NamedTuple giving a single row
-        n = length(gd)
+        n = lgd
         eltys = map(typeof, first)
         if any(x -> x <: AbstractVector, eltys)
             throw(ArgumentError("mixing single values and vectors in a named tuple is not allowed"))
         end
     end
     idx = idx_agg === NOTHING_IDX_AGG ? Vector{Int}(undef, n) : idx_agg
+    sizehint!(idx, lgd)
```
nalimilan, line 56: This assumes that there's going to be at least one row per group, right? Maybe add a comment explaining this? That's not necessarily the case if groups are dropped, though that's probably quite common.

nalimilan (approved):

src/groupeddataframe/complextransforms.jl
bkamins: Added. What you write is exactly the thinking. `sizehint!` is cheap and dropping groups is uncommon (actually in the past it was even super hard to do :smile:):
```
julia> x = Int[];

julia> @time sizehint!(x, 10^7);
  0.000018 seconds (2 allocations: 76.294 MiB)

julia> x = Int[];

julia> @time sizehint!(x, 10^7);
  0.000021 seconds (2 allocations: 76.294 MiB)
```

bkamins:
Thank you!

PR #2756, opened 2021-05-10
Alignment of complex numbers

### LINKED ISSUE #2754: Display complex numbers - alignment
Actual for Complex[Float64] the decimal of the real part is used as anchor (whole imaginary part right of it), complex without decimal are printed like pre-decimal places:
grafik

For complex numbers I suggest to use the sign of the imaginary part as anchor, so the example above would look this way:
grafik

Code for example:
`DataFrame(Test_Integer=[123+5im, 45-56im, 1-234im], Test_Float=[1.234+2.3im, 12.1-23.4im, 1.3-1.456im], Test_Number=Number[12.1-23.4im, 1//3+4//5im, 45-56im])`

ronisbr:
Hi!

This PR should fix #2754 . However, it adds somewhat complex regexes, which can decrease performance, and have some corner case that I did not think of. Thus, I need help testing this algorithm :)


test/show.jl
```diff
@@ -687,6 +687,56 @@ end
...
+       4 │          1.123+123.0im                   -1000                   1.0            0.01+10.0im
+       5 │          -2.323-32.0im                  -10000             missing               0.1-1.0im
+       6 │          -1//3+4//5*im                 -100000                 100.0            -1.0+0.1im
+       7 │                1028.23                       1                1000.0            10.0-0.01im
+       8 │                missing                      10               10000.0           100.0+0.001im
+       9 │              -0.304105                     100              100000.0         -1000.0-0.0001im
+      10 │                2123123                    1000                   1.0e6       10000.0+1.0e-5im
+      11 │               0.304105                   10000                   1.0e7      100000.0-1.0e-6im
+      12 │  1.3123e-10+1.123e-5im                  100000                   1.0e8        -1.0e6+1.0e-7im"""
+end
```
bkamins, line 738: @ronisbr can you please check how the following `DataFrame` would get printed?
```
DataFrame(x=[false, im, true], y=[im, im, im])
```

bkamins (approved):

bkamins:
@nalimilan - can you please have a quick look if you agree with the rules used?

test/show.jl
ronisbr: Yes, it is!

test/show.jl
```diff
@@ -687,6 +687,56 @@ end
...
+        -1//3 - 4//5im,
+        1028.23123,
+        missing,
+        -0.304105,
+        2123123,
+        +0.304105,
+        1.3123e-10 + 1.123e-5im
+    ]
+    b = Int64[i ≤ 6 ? -1 * 10^(i - 1) : 10^(i - 7) for i = 1:12];
+    c = Union{Missing, Float64}[i == 5 ? missing : 10.0^(i-4) for i = 1:12];
```
nalimilan, line 706:
```suggestion
    b = Int64[i ≤ 6 ? -1 * 10^(i - 1) : 10^(i - 7) for i = 1:12]
    c = Union{Missing, Float64}[i == 5 ? missing : 10.0^(i-4) for i = 1:12]
```

test/show.jl
```diff
@@ -687,6 +687,56 @@ end
...
+    d = ComplexF64[(-1)^(i % 3 == 0) * 10.0^(i-6) + (-1)^i * (10.0)^(11-i-6) * im for i = 1:12]
+    df = DataFrame(
+        very_big_column_name_1 = a,
+        very_big_column_name_2 = b,
+        very_big_column_name_3 = c,
+        very_big_column_name_4 = d
+    )
```
nalimilan, line 713:
```suggestion
    df = DataFrame(very_big_column_name_1 = a,
                   very_big_column_name_2 = b,
                   very_big_column_name_3 = c,
                   very_big_column_name_4 = d)
```

test/show.jl
```diff
@@ -687,6 +687,56 @@ end
...
+
+    io = IOContext(IOBuffer())
+    show(io, df)
+    str = String(take!(io.io))
+
+    str_complex = VERSION < v"1.1" ? "Complex{Float64}" : "ComplexF64"
+
+    @test str == """
+    12×4 DataFrame
```
nalimilan, line 722: Add four-space indent?

test/show.jl
```diff
@@ -687,6 +687,56 @@ end
...
+     Row │ very_big_column_name_1  very_big_column_name_2  very_big_column_name_3  very_big_column_name_4
+         │ Union{Missing, Number}  Int64                   Float64?                $str_complex
```
nalimilan, line 724: Isn't this enough?
```suggestion
         │ Union{Missing, Number}  Int64                   Float64?                $(Complex{Float64})
```

nalimilan (commented):
Makes sense. Just a few style comments.

(Ideally I would say that in addition to aligning on the sign of the imaginary parts, the real parts should be aligned like real numbers, and imaginary parts should be aligned like real numbers, but Base doesn't do this either since that would require adding spaces in the middle of the number...)

test/show.jl
ronisbr: Maybe, let's see if the test passes.

test/show.jl
ronisbr: Ok!

ronisbr:
> (Ideally I would say that in addition to aligning on the sign of the imaginary parts, the real parts should be aligned like real numbers, and imaginary parts should be aligned like real numbers, but Base doesn't do this either since that would require adding spaces in the middle of the number...)

Yes, this would be good but involve some changes in PrettyTables.jl. What we need to implement is the support for multiple anchors, so that we can have multiple alignments. It is not an easy task.

I think I resolved all @nalimilan's comments!

bkamins:
> It is not an easy task.

This is what I would imagine. I would leave it for later. I think we have more pivotal functionalities (like HTML/LaTeX support) that should get a higher priority.

nalimilan (approved):

bkamins:
@ronisbr - OK to merge?

ronisbr:
Yes!

bkamins:
Thank you!

PR #2774, opened 2021-05-24
Put longer type into th title argument in HTML show

mortenpi:
Adds a `title` argument to the types in the HTML output, so that we would get the full type (up to 256 characters) when hovering over the shortened type with a mouse:

Screenshot from 2021-05-24 14-15-24

### DIFF
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -157,6 +157,10 @@ function html_escape(cell::AbstractString)
     cell = replace(cell, "&"=>"&amp;")
     cell = replace(cell, "<"=>"&lt;")
     cell = replace(cell, ">"=>"&gt;")
+    # Replace quotes so that the resulting string could also be used in the attributes of
+    # HTML tags
+    cell = replace(cell, "\""=>"&quot;")
+    cell = replace(cell, "'"=>"&apos;")
     return cell
 end
 
--- a/src/abstractdataframe/io.jl
+++ b/src/abstractdataframe/io.jl
@@ -195,10 +195,15 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
     if eltypes
         write(io, "<tr>")
         write(io, "<th></th>")
-        ct = batch_compacttype(Any[eltype(df[!, idx]) for idx in 1:mxcol])
+        # We put a longer string for the type into the title argument of the <th> element,
+        # which the users can hover over. The limit of 256 characters is arbitrary, but
+        # we want some maximum limit, since the types can sometimes get really-really long.
+        types = Any[eltype(df[!, idx]) for idx in 1:mxcol]
+        ct, ct_title = batch_compacttype(types), batch_compacttype(types, 256)
         for j in 1:mxcol
             s = html_escape(ct[j])
-            write(io, "<th>$s</th>")
+            title = html_escape(ct_title[j])
+            write(io, "<th title=\"$title\">$s</th>")
         end
         write(io, "</tr>")
     end
--- a/test/io.jl
+++ b/test/io.jl
@@ -52,7 +52,7 @@ end
     str = String(take!(io))
     @test str == "<table class=\"data-frame\"><thead><tr><th>" *
                  "</th><th>Fish</th><th>Mass</th></tr>" *
-                 "<tr><th></th><th>String</th><th>Float64?</th></tr></thead><tbody>" *
+                 "<tr><th></th><th title=\"String\">String</th><th title=\"Union{Missing, Float64}\">Float64?</th></tr></thead><tbody>" *
                  "<p>2 rows × 2 columns</p>" *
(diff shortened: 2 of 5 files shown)


src/abstractdataframe/io.jl
```diff
@@ -157,6 +157,10 @@ function html_escape(cell::AbstractString)
     cell = replace(cell, "&"=>"&amp;")
     cell = replace(cell, "<"=>"&lt;")
     cell = replace(cell, ">"=>"&gt;")
+    # Replace quotes so that the resulting string could also be used in the attributes of
+    # HTML tags
+    cell = replace(cell, "\""=>"&quot;")
+    cell = replace(cell, "'"=>"&apos;")
```
mortenpi, line 163: I don't think adding additional escapes here should break anything for existing uses of `html_escape`. As a test, this `DataFrame` seems to `show` fine:

```
struct Foo{T} end
DataFrame(
    xs = ["'", "\"", "<foo>'</bar>"],
    ys = [Foo{'\''}(), Foo{'"'}, Foo{Symbol("\"'")}()],
    zs = Foo{'"'}[Foo{'"'}(), Foo{'"'}(), Foo{'"'}()],
)
```

mortenpi:
The test failures are coming from the string comparisons of the `show` output. Will happily fix them if the implementation looks good to you.

src/abstractdataframe/io.jl
bkamins: yes - I think it should be OK. Although it would be good to have a test if all is OK.

src/abstractdataframe/io.jl
```diff
@@ -195,10 +199,15 @@ function _show(io::IO, ::MIME"text/html", df::AbstractDataFrame;
     if eltypes
         write(io, "<tr>")
         write(io, "<th></th>")
-        ct = batch_compacttype(Any[eltype(df[!, idx]) for idx in 1:mxcol])
+        # We put a longer string for the type into the title argument of the <th> element,
```
bkamins, line 202: I think it is OK

bkamins:
Looks very good. Thank you!

@ronisbr - it would be good to have it in PrettyTables.jl also.

CC @nalimilan

ronisbr:
> @ronisbr - it would be good to have it in PrettyTables.jl also.

This feature is awesome! I will add an option into PrettyTables.jl to create comments for each HTML cell. Thus, the packages can use it as they like.

src/abstractdataframe/io.jl
mortenpi: I added this example in as an additional test case.

Of course, string comparison doesn't actually check if the resulting HTML makes sense (although in this case I've inspected and tested the output). As I was working through the tests, I did wonder whether, instead of comparing strings, it might be a good idea to run the outputs through a HTML parser and see (1) if it parses as valid HTML, and (2) check that the attributes and elements have the correct values.

mortenpi:
After a bit of whack-a-mole, I think I got all the tests fixed :slightly_smiling_face:

src/abstractdataframe/io.jl
bkamins: > although in this case I've inspected and tested the output

This is what I always do. I always feared that setting up the dependencies required to:

> run the outputs through a HTML parser and see (1) if it parses as valid HTML, and (2) check that the attributes and elements have the correct values.

would be too hard to maintain and too expensive (it is cheaper for me - as I am not an expert in the HTML parsing field - to just manually check it).

Agreed?

test/io.jl
```diff
@@ -153,6 +159,47 @@ end
...
+    df = DataFrame(
+        xs = ["'", "\"", "<foo>'</bar>"],
+        ys = [QuoteTestType{'\''}(), QuoteTestType{'"'}, QuoteTestType{Symbol("\"'")}()],
+        zs = QuoteTestType{'"'}[QuoteTestType{'"'}(), QuoteTestType{'"'}(), QuoteTestType{'"'}()],
+    )
+    io = IOBuffer()
+    show(io, "text/html", df)
+    str = String(take!(io))
+    @test str ==
+        "<table class=\"data-frame\"><thead>"*
```
bkamins, line 172: can you add a space before `*` everywhere for consistency? Thank you!

bkamins (approved):

bkamins:
@nalimilan - can you please have a quick look before I merge (there is a small code layout fix that I asked but it does not affect the solution). Thank you!

test/io.jl
mortenpi: Done!

nalimilan (approved):
Nice! That's especially useful to explain that `T?` means `Union{T, Missing}`.

bkamins:
@mortenpi - I have pushed a NEWS.md entry for your PR. Please let me know if it is OK. Then I will merge. Thank you!

mortenpi:
LGTM!

bkamins:
Thank you!

PR #2778, opened 2021-06-03
remove escape in Char

bkamins:
`'$'` and `'\$'` are the same, but the latter confuses code parser in VS Code.


nalimilan (approved):

bkamins:
Thank you!

PR #2782, opened 2021-06-07
allow Matrices in transformations of GroupedDataFrame

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2781

### DIFF
--- a/src/groupeddataframe/splitapplycombine.jl
+++ b/src/groupeddataframe/splitapplycombine.jl
@@ -27,7 +27,8 @@ function _combine_prepare(gd::GroupedDataFrame,
                           keepkeys::Bool, ungroup::Bool, copycols::Bool,
                           keeprows::Bool, renamecols::Bool)
     for cei in cs
-        @assert cei isa Union{Pair, Base.Callable, ColumnIndex, MultiColumnIndex}
+        @assert cei isa Union{Pair, Base.Callable, ColumnIndex, MultiColumnIndex,
+                              AbstractVecOrMat{<:Pair}}
     end
     if !ungroup && !keepkeys
         throw(ArgumentError("keepkeys=false when ungroup=false is not allowed"))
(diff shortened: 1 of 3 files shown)


pdeffebach (commented):
It looks like all the action with `select(df::AbstractDataFrame...)` is happening inside `manipulate` [here](https://github.com/JuliaData/DataFrames.jl/blob/a671a3b86537a253d461ff54b87ae2c26e4e8a04/src/abstractdataframe/selection.jl#L1288). 

Why not use the same structure / logic with SplitApplyCombine?

bkamins:
We essentially use the same structure just the function is not called `manipulate`, but `_combine_prepare` and then super complex logic.

Originally I preferred to add type assertions in `select` etc. in `GroupedDataFrame` case to be on a safe side - if I am not able to write a correct signature then it means that the design is too complex :) (and here what happened is that we have added support for matrices, but forgotten to add signatures).

pdeffebach:
Okay. It sounds like we might want to re-factor in the future, but shouldn't use this PR to do it.

bkamins:
Yes - refactoring is something I wanted to do for a long time, but you know how it is: we add features, the code gest more complex, at some point it is super hard to change anything unless you rewrite it from scratch (that is why I put so much attention to tests and documentation).

src/groupeddataframe/splitapplycombine.jl
```diff
@@ -27,7 +27,8 @@ function _combine_prepare(gd::GroupedDataFrame,
                           keepkeys::Bool, ungroup::Bool, copycols::Bool,
                           keeprows::Bool, renamecols::Bool)
     for cei in cs
-        @assert cei isa Union{Pair, Base.Callable, ColumnIndex, MultiColumnIndex}
+        @assert cei isa Union{Pair, Base.Callable, ColumnIndex, MultiColumnIndex,
+                              AbstractVecOrMat{<:Pair}}
```
nalimilan, line 31: How were vectors supported before this PR?

bkamins: I could add `AbstractMatrix{<:Pair}` in the union, but I thought that it is cleaner to use `AbstractVecOrMat{<:Pair}` to signal the design intention.

bkamins: @nalimilan - do we agree on this approach?

bkamins: @nalimilan - ok to merge?

nalimilan (approved):

bkamins:
Thank you!

Issue #2790, opened 2021-06-18
inconsistency of groupby() for -0.0


sl-solution:
The following examples demonstrate the issue,

```julia
df = DataFrame(a = [0.0,-0.0,1.0])
groupby(df,1)
df = DataFrame(a = [0.0,-0.0,1.1])
groupby(df,1)
```

bkamins:
Thank you for reporting.

PR #2791, opened 2021-06-18
Fix float grouping

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2790

### DIFF
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -75,7 +75,7 @@ not matter in this case.
 
 In the current implementation if `sort=false` groups are ordered following the
 order of appearance of values in the grouping columns, except when all grouping
-columns provide non-`nothing` `DataAPI.refpool` in which case the order of groups
+columns provide non-`nothing` `DataAPI.refpool`, in which case the order of groups
 follows the order of values returned by `DataAPI.refpool`. As a particular application
 of this rule if all `cols` are `CategoricalVector`s then groups are always sorted.
 Integer columns with a narrow range also use this this optimization, so to the
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -45,14 +45,8 @@ into row groups.
   ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
 - `sort` : whether to sort groups according to the values of the grouping columns
   `cols`; if `sort=false` then the order of groups in the result is undefined
-  and may change in future releases. In the current implementation
-  groups are ordered following the order of appearance of values in the grouping
-  columns, except when all grouping columns provide non-`nothing`
-  `DataAPI.refpool` in which case the order of groups follows the order of
-  values returned by `DataAPI.refpool`. As a particular application of this rule
-  if all `cols` are `CategoricalVector`s then groups are always sorted
-  Integer columns with a narrow range also use this this optimization, so
-  to the order of groups when grouping on integer columns is undefined.
+  and may change in future releases; below a description of the current
+  implementation is provided.
 - `skipmissing` : whether to skip groups with `missing` values in one of the
   grouping columns `cols`
 
@@ -79,10 +73,17 @@ an `AbstractDict` can be used to index into a grouped data frame where
 the keys are column names of the data frame. The order of the keys does
 not matter in this case.
 
+In the current implementation if `sort=false` groups are ordered following the
+order of appearance of values in the grouping columns, except when all grouping
+columns provide non-`nothing` `DataAPI.refpool` in which case the order of groups
+follows the order of values returned by `DataAPI.refpool`. As a particular application
+of this rule if all `cols` are `CategoricalVector`s then groups are always sorted.
+Integer columns with a narrow range also use this this optimization, so to the
+order of groups when grouping on integer columns is undefined.
 A column is considered to be an integer column when deciding on the grouping
-algorithm choice if its `eltype` is a subtype of `Union{Missing, Real}`, all its
-elements are either `missing` or pass `isinteger` test, and none of them is
-equal to `-0.0`.
+algorithm choice if its `eltype` is a subtype of `Union{Missing, Real}`,
+all its elements are either `missing` or pass `isinteger` test,
+and none of them is equal to `-0.0`.
 
 # See also
 
--- a/src/groupeddataframe/groupeddataframe.jl
+++ b/src/groupeddataframe/groupeddataframe.jl
@@ -79,6 +79,11 @@ an `AbstractDict` can be used to index into a grouped data frame where
 the keys are column names of the data frame. The order of the keys does
 not matter in this case.
 
+A column is considered to be an integer column when deciding on the grouping
+algorithm choice if its `eltype` is a subtype of `Union{Missing, Real}`, all its
+elements are either `missing` or pass `isinteger` test, and none of them is
+equal to `-0.0`.
+
 # See also
 
 [`combine`](@ref), [`select`](@ref), [`select!`](@ref), [`transform`](@ref), [`transform!`](@ref)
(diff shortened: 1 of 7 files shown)


nalimilan:
Woops. Though maybe it would be worth checking whether the data contains `-0.0`, and use the fast path if not? Is there also a problem with `NaN` and `Inf`?

bkamins:
>  Is there also a problem with `NaN` and `Inf`?

No, it is OK, because:
```
julia> isinteger(Inf)
false

julia> isinteger(NaN)
false
```

We could add a check against `-0.0`. However, I feel that doing `groupby` floats is rare (and in general discouraged) therefore I thought we can skip the fast path in this case.

If you would prefer to add special handling of `-0.0` instead please let me know and I can add this instead of what I propose now.

nalimilan:
As you prefer, but I tend to think that it's good to support the fast path in as many cases as possible unless that's hard to do. For example, R uses floats by default when reading CSV files, so people could end up with float columns with only integers instead of `Integer` columns when transferring from or loading Arrow files written in R. That may even end up in a blog post comparing DataFrames.jl's performance against R. It's hard to anticipate what weird things people might do...

test/grouping.jl
```diff
@@ -3710,6 +3710,14 @@ end
     end
 end
 
+@testset "grouping floats" begin
+    @test length(groupby_checked(DataFrame(a=[0.0, -0.0]), :a)) == 2
+    @test getindex.(keys(groupby_checked(DataFrame(a=[3.0, 2.0, 0.0]), :a)), 1) ==
+          [0, 2, 3]
+    @test getindex.(keys(groupby_checked(DataFrame(a=[3.0, 2.0, -0.0]), :a)), 1) ==
+          [3, 2, 0]
+end
```
bkamins, line 3719: in the tests above you see the consequence of allowing floats in fast path. The issue is that then the order of groups (being a consequence of switching of algorithm) depends on the values not only on types.

bkamins:
> As you prefer, but I tend to think that it's good to support the fast path in as many cases as possible unless that's hard to do.

I have implemented it and added tests showing the consequences. Let us decide which approach is better.

src/groupeddataframe/utils.jl
```diff
@@ -143,10 +143,28 @@ function refpool_and_array(x::AbstractArray)
         else
             return nothing, nothing
         end
-    elseif x isa AbstractArray{<:Union{Real, Missing}} &&
-        all(v -> ismissing(v) | isinteger(v), x) &&
-        !isempty(skipmissing(x))
-        minval, maxval = extrema(skipmissing(x))
+    elseif x isa AbstractArray{<:Union{Real, Missing}} && length(x) > 0
+        if !(x isa AbstractArray{<:Union{Integer, Missing}})
+            if !all(v -> (ismissing(v) | isinteger(v)) & (!isequal(v, -0.0)), x)
```
nalimilan, line 148:
```suggestion
            if !all(v -> (ismissing(v) | isinteger(v)) & !isequal(v, -0.0), x)
```

src/groupeddataframe/utils.jl
```diff
@@ -143,10 +143,28 @@ function refpool_and_array(x::AbstractArray)
...
+            y === nothing && return nothing, nothing
+            (v, s) = y
+            minval = maxval = v
+            while true
+                y = iterate(smx, s)
+                y === nothing && break
+                (v, s) = y
+                maxval = max(v, maxval)
+                minval = min(v, minval)
+            end
```
nalimilan, line 164: Why not just keep using `extrema`? Is this just to avoid checking `isempty(skipmissing(x))` first? If so I'm not sure it's worth it TBH, as it should be relatively cheap compared with the cost of grouping. Also if we wanted to be maximally efficient we should use a single loop in which we would check both whether there's at least one non-missing value and `(ismissing(v) | isinteger(v)) & !isequal(v, -0.0)`.

test/grouping.jl
nalimilan: Ah, right. We don't provide any guarantees on the order, but it's true that the change could be confusing.

src/groupeddataframe/utils.jl
bkamins: The reason is performance. It is not a huge difference but still noticeable for sizes like 10^8.

> `(ismissing(v) | isinteger(v)) & !isequal(v, -0.0)`

This check is done only for non-`Integer` eltype (and I decided it is OK to run a loop twice, as I am still convinced that it is a rare use-case - people should not `groupby` float values I think).

test/grouping.jl
bkamins: Before I move forward with the changes you suggested above let us make this decision:
1. do we want to be flexible (and **sometimes** handle floats using fast path),
2. or we want to use fast path only to `Integer`s?

I prefer option 2. (i.e. do not handle floats using fast path at all), but I also understand the reasons why you would prefer it, so I am not sure which way to go.

bkamins:
I will finalize this PR with the current design (treat float as integer if possible). After thinking I think it is less breaking, as we just have to add only the following condition to the current rules:

> `-0.0` is considered not to be an integer

(so doing it this way is less breaking)

src/groupeddataframe/utils.jl
nalimilan: > The reason is performance. It is not a huge difference but still noticeable for sizes like 10^8.

You mean the performance of `isempty(skipmissing(x))`? That cost should be pretty low unless you have 10^8 missing values at the beginning, right?

bkamins: OK - I will revert.

pdeffebach (approved):

bkamins:
@nalimilan - I have made the requested cleanups. This should be good to be merged. Thank you!

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -79,6 +79,11 @@ an `AbstractDict` can be used to index into a grouped data frame where
 the keys are column names of the data frame. The order of the keys does
 not matter in this case.
 
+A column is considered to be an integer column when deciding on the grouping
+algorithm choice if its `eltype` is a subtype of `Union{Missing, Real}`, all its
+elements are either `missing` or pass `isinteger` test, and none of them is
+equal to `-0.0`.
```
nalimilan, line 85: It would make sense to regroup all information regarding algorithm and order in the same place. Currently it's discussed under the `sort` keyword argument. Maybe shorten its description and move its contents here, with a pointer to here?

src/groupeddataframe/utils.jl
```diff
@@ -143,10 +143,22 @@ function refpool_and_array(x::AbstractArray)
...
-    elseif x isa AbstractArray{<:Union{Real, Missing}} &&
-        all(v -> ismissing(v) | isinteger(v), x) &&
-        !isempty(skipmissing(x))
-        minval, maxval = extrema(skipmissing(x))
+    elseif x isa AbstractArray{<:Union{Real, Missing}} && length(x) > 0
+        if !(x isa AbstractArray{<:Union{Integer, Missing}})
+            if !all(v -> (ismissing(v) | isinteger(v)) & !isequal(v, -0.0), x)
+                return nothing, nothing
+            end
+        end
```
nalimilan, line 151:
```suggestion
    elseif x isa AbstractArray{<:Union{Real, Missing}} && length(x) > 0 &&
        (x isa AbstractArray{<:Union{Integer, Missing}} ||
         all(v -> (ismissing(v) | isinteger(v)) & !isequal(v, -0.0), x))
```

src/groupeddataframe/groupeddataframe.jl
bkamins: OK - I moved it as suggested.

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -79,6 +73,18 @@ an `AbstractDict` can be used to index into a grouped data frame where
 the keys are column names of the data frame. The order of the keys does
 not matter in this case.
 
+In the current implementation if `sort=false` groups are ordered following the
+order of appearance of values in the grouping columns, except when all grouping
+columns provide non-`nothing` `DataAPI.refpool` in which case the order of groups
```
nalimilan, line 78:
```suggestion
columns provide non-`nothing` `DataAPI.refpool`, in which case the order of groups
```

nalimilan (approved):

bkamins:
Thank you!

Issue #2785, opened 2021-06-10
Assignment to SubDataFrame


bkamins:
@pdeffebach following our discussion on `WhereDataFrame` I propose the following extra new rules for assignment to `SubDataFrame`.

They add complexity unfortunately, but the benefit is that they are doing what probably user would want (I hope).
The crucial decision is the following:
* for adding new columns the issue is simple I think and it is clear we should add the functionality;
* Stata and `data.table` DO NOT allow replacing existing columns; however, if we want we can (and my proposal assumes this); the point is that with `df[:, col] = v` and `df[!, col] = v` distinction we can support both functionalities cleanly. I thought users might find it useful to replace existing columns with filtered-out values rather than only being able to create a new column this way. But maybe you will feel that this is too flexible. Then we can only focus on adding functionalities for adding columns
* also this means that what I propose now allows to remove columns from the parent via `select!`

In summary - this is a "full option" proposal. I am OK to limit its scope to only add things that we find really useful in practice (but I thought it is better to have at this stage all things laid out on a table and decide what we allow and what we disallow).

The extra rules are only for `SubDataFrame` that does not subset columns (uses `:` as column selector - the point is to use `:` as selector not to select all columns in their original order as this is a less strict requirements).
This is easily identified using the `AbstractIndex` in the type. Example:
```
julia> df = DataFrame(a=1,b=2,c=3)
1×3 DataFrame
 Row │ a      b      c     
     │ Int64  Int64  Int64 
─────┼─────────────────────
   1 │     1      2      3

julia> typeof(view(df, 1:1, 1:2)) # NOT ALLOWED
SubDataFrame{DataFrame, DataFrames.SubIndex{DataFrames.Index, UnitRange{Int64}, UnitRange{Int64}}, UnitRange{Int64}}   

julia> typeof(view(df, 1:1, :)) # ALLOWED
SubDataFrame{DataFrame, DataFrames.Index, UnitRange{Int64}}

julia> typeof(view(df, 1:1, All())) # NOT ALLOWED
SubDataFrame{DataFrame, DataFrames.SubIndex{DataFrames.Index, UnitRange{Int64}, UnitRange{Int64}}, UnitRange{Int64}} 
```

Functions to cover:
* `insertcols!` - natural (it does not allow column overwriting already)
[...]

PR #2794, opened 2021-06-21
Support adding columns to views

bkamins:
This partially implements https://github.com/JuliaData/DataFrames.jl/issues/2785 for:
* `setindex!` with `:` as row selector
* broadcasted assignment with `:` as row selector
* `insertcols!`

as only these operations, as discussed, ensure that we ADD columns (and not replace them).

Before I add tests let us discuss what we think about this functionality.

@matthieugomez - in particular could you please comment if you would find it useful? (or your fundamental use case is `transform!` with `groupby` run on `SubDataFrame` and having only these methods is of not much value?)

The problematic point is that supporting `select!` and `transfrorm!`, as commented earlier creates a slight ambiguity (of course possible to resolve) when if we write `:a => identity => :a` as a transformation it is not clear if `:a` should be filled with `missing` in filtered-out rows or it should be left as is.


src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2173,3 +2173,209 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
 
 Base.setindex!(::AbstractDataFrame, ::Any, ::Union{Symbol, Integer, AbstractString}) =
     throw(ArgumentError("syntax df[column] is not supported use df[!, column] instead"))
+
+# insertcols!
```
nalimilan, line 2177: Can you use two separate commits, one just moving the code to the new place and the next one making actual changes? That will help reviewing and if we  keep both when merging it will also make the history cleaner.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2173,3 +2173,209 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
...
+  if it is `false` an error will be thrown; if it is `true` a new unique name will
+  be generated by adding a suffix
+- `copycols` : whether vectors passed as columns should be copied
+
+If `val` is an `AbstractRange` then the result of `collect(val)` is inserted.
+
+If `df` is a `SubDataFrame` then it must be created with `:` as column selector
+(otherwise an error is thrown). In this case the `copycols` keyword argument
+is ignored an added column is always copied and the parent data frame is
+filled with `missing` in rows that are filtered out by `df`.
```
nalimilan, line 2207:
```suggestion
If `df` is a `SubDataFrame` then it must have been created with `:` as column selector
(otherwise an error is thrown). In this case the `copycols` keyword argument
is ignored (i.e. the added column is always copied) and the parent data frame's
column is filled with `missing` in rows that are filtered out by `df`.
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2173,3 +2173,209 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
...
+─────┼───────────────────────────
+   1 │ a         2      3      1
+   2 │ b         3      4      2
+   3 │ c         4      5      3
+```
+"""
+function insertcols!(df::AbstractDataFrame, col::ColumnIndex, name_cols::Pair{Symbol, <:Any}...;
+                     makeunique::Bool=false, copycols::Bool=true)
+    if !(df isa DataFrame || (df isa SubDataFrame && getfield(df, :colindex) isa Index))
+        throw(ArgumentError("insertcols! is only supported for DataFrame or" *
```
nalimilan, line 2252:
```suggestion
        throw(ArgumentError("insertcols! is only supported for DataFrame or " *
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2173,3 +2173,209 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
...
+                x = item[]
+                item_new = fill!(Tables.allocatecolumn(typeof(x), target_row_count), x)
+            else
+                @assert !(item isa AbstractArray)
+                item_new = fill!(Tables.allocatecolumn(typeof(item), target_row_count), item)
+            end
+        elseif item isa AbstractRange
+            item_new = collect(item)
+        elseif copycols
+            item_new = copy(item)
```
nalimilan, line 2315: These steps are not needed when `df isa SubDataFrame`, right?

src/other/broadcasting.jl
```diff
@@ -108,6 +108,18 @@ function Base.dotview(df::DataFrame, ::Colon, cols::ColumnIndex)
...
 end
 
+function Base.dotview(df::SubDataFrame, ::Colon, cols::ColumnIndex)
+    haskey(index(df), cols) && return view(df, :, cols)
+    if !(cols isa SymbolOrString)
+        throw(ArgumentError("creating new columns using an integer index is disallowed"))
+    end
+    if !(getfield(df, :colindex) isa Index)
+        throw(ArgumentError("creating new columns in a SubDataFrame that subsets " *
+                            "columns of a parent data frame is disallowed"))
```
nalimilan, line 118:
```suggestion
                            "columns of its parent data frame is disallowed"))
```

src/subdataframe/subdataframe.jl
```diff
@@ -181,7 +181,16 @@ Base.@propagate_inbounds function Base.setindex!(sdf::SubDataFrame, val::Any, id
     setindex!(sdf, val, idx[1], idx[2])
 end
 Base.@propagate_inbounds function Base.setindex!(sdf::SubDataFrame, val::Any, ::Colon, colinds::Any)
-    parent(sdf)[rows(sdf), parentcols(index(sdf), colinds)] = val
+    if colinds isa SymbolOrString && getfield(sdf, :colindex) isa Index &&
+       val isa AbstractVector && columnindex(sdf, colinds) == 0 && nrow(sdf) == length(val)
```
nalimilan, line 185: When `nrow(sdf) == length(val)` isn't true, an error will eventually be thrown by the other branch, right? Maybe it would be clearer to check this here. Taking a completely different path depending on the length makes me feel uneasy.

Also:
```suggestion
        val isa AbstractVector && columnindex(sdf, colinds) == 0 && nrow(sdf) == length(val)
```

src/other/broadcasting.jl
```diff
@@ -144,15 +156,22 @@ if isdefined(Base, :dotgetproperty)
...
         copyto!(col, bc)
     else
         col = Base.Broadcast.materialize(bc)
     end
-    lazydf.df[!, lazydf.col] = col
+    if df isa DataFrame
+        return df[!, lazydf.col] = col
+    else
+        @assert df isa SubDataFrame && getfield(df, :colindex) isa Index
+        return df[:, lazydf.col] = col
```
nalimilan, line 173: Why use `:` here?

docs/src/lib/indexing.md
```diff
@@ -190,6 +192,10 @@ Additional rules:
   `df` is performed in-place; if `rows` is `:` and `col` is `Symbol` or `AbstractString`
   and it is missing from `df` then a new column is allocated and added;
   the length of the column is always the value of `nrow(df)` before the assignment takes place;
+* in the `sdf[:, col] .= v` if `sdf` was created with `:` as column selector
+  and `col` is a `Symbol` or `AbstractString` that is not present in `df` then a new column in `df`
+  is created and holds contents of `v` broadcasted onto rows selected in `sdf`
+  and `missing` in all rows present in `parent(sdf)` but not present in `sdf`.
```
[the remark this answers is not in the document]
bkamins: it is allowed, see line below:

> in the `sdf[rows, col] .= v` and `sdf[rows, cols] .= v` syntaxes the assignment to `sdf` is performed in-place;

I will clean up these soon (adding the rules for `!`)

src/abstractdataframe/abstractdataframe.jl
bkamins: OK - so for now I will not move this code to abstractdataframe.jl but keep it in dataframe.jl (and move it later). The reason is that we squash merge PRs anyway.

src/abstractdataframe/abstractdataframe.jl
bkamins: yes - copying is not needed then. I will fix this

src/other/broadcasting.jl
bkamins: because `df` here is a `SubDataFrame`, and we want to create a new column (and in the current version of the PR `!` was not supported - I will change it when we add newely agreed rules).

src/subdataframe/subdataframe.jl
bkamins: OK

bkamins:
@nalimilan @pdeffebach @matthieugomez 

This PR should be good for review.

The rules in short are:
* `SubDataFrame` with `:` column selector allow any mutation of the parent. values from the parent in filtered-out rows are used, except when a new column is created when `missing` is used for filling
* `SubDataFrame` with other column selector allows replacing existing columns but not creating new columns

Operations that are supported:
* `setindex!`
* broadcasted assignment
* `insertcols!`
* `select!` and `transform!` both for `SubDataFrame` and `GroupedDataFrame`

@pdeffebach @jkrumbiegel this new functionality is aimed - in particular - to allow a flexible `where` clause for transformations in macro-packages.

bkamins:
@nalimilan + @pdeffebach : no rush of course, but it would be excellent if you had a look at this PR at some not very distant future (it is relatively complex so it would be good to work on it while I still remember what I have coded 😄). Thank you!

NEWS.md
```diff
@@ -17,7 +21,27 @@
...
+
+  Assignment to existing columns retains allocates a new column. Values already stored in filtered-out rows are retained.
+
+* Allow `SubDataFrame` to be passed as argument of , `select!` and `transform!`
+  (also on `GroupedDataFrame` created a `SubDataFrame`)
+  ([#2794](https://github.com/JuliaData/DataFrames.jl/pull/2794)).
+
+  Assignment to existing columns retains allocates a new column. Values already stored in filtered-out rows are retained.
+  In case of creation of new columns `missing` values stored in filtered-out rows;
+  If `SubDataFrame` is not created with `:` as column selector the resulting operation
```
pdeffebach, line 43: I think this is a good rule.

docs/src/lib/indexing.md
```diff
@@ -138,15 +138,26 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
...
-Note that `sdf[!, col] = v`, `sdf[!, cols] = v` and `sdf.col = v` are not allowed as `sdf` can be only modified in-place.
+* `sdf[!, col] = v` -> replaces `col` with `v` with copying; if `col` is present if `sdf`
+                       then filtered-out rows in newly created vector are filled with
+                       values already present in that column;
+                       if `col` is not present in `sdf` then the operation is only allowed
+                       if `sdf` was created with `:` as column selector. In this case
+                       filtered-out rows are filled with `missing`;
+                      equivalent to `sdf.col = v` if `col` is a valid identifier;
+                      operation is allowed if `length(v) == nrow(sdf)`;
+* `sdf[!, cols] = v` -> replaces existing columns `cols` in data frame `sdf` with copying;
```
pdeffebach, line 156: Do you mean `sdf[:, cols]` rather than `sdf[!, cols]`?

I'm not sure what's being copied here. is `v` needs to get new memory allocated regardless, right? Is it that the entire column `df.v` is replaced? That can't be fast, if someone is doing a loop with multiple groups.

src/subdataframe/subdataframe.jl
```diff
@@ -176,32 +176,106 @@ Base.@propagate_inbounds Base.getindex(df::SubDataFrame, row_ind::typeof(!),
...
+                        ::typeof(!), col_ind::ColumnIndex)
+    if col_ind isa Union{Signed, Unsigned} && !(1 <= col_ind <= ncol(sdf))
+        throw(ArgumentError("Cannot assign to non-existent column: $col_ind"))
+    end
+    if col_ind isa SymbolOrString && columnindex(sdf, col_ind) == 0
+        if !is_column_adding_allowed(sdf)
+            throw(ArgumentError("creating new columns in a SubDataFrame that subsets " *
+                                "columns of its parent data frame is disallowed"))
+        end
+        sdf[:, col_ind] = v
```
pdeffebach, line 214: Should we disallow either `sdf[:, col_ind]` or `sdf[!, col_ind]`? Given that the implementations are so similar and the logic about memory access doesn't apply as closely?

pdeffebach (commented):
Thanks!

I will leave Milan to comment more on the technical details. 

On the API, I wonder if `transform!(sdf, ...)` should return `parent(sdf)` after the transformation. This is more in-line with what we do for `transfrorm(gd::GroupedDataFrames, ...)`. We decided to automatically combine because the `ungroup` stickiness can lead to bugs in R. Maybe we should add a keyword argument to `transform!` about this?

src/subdataframe/subdataframe.jl
```diff
@@ -176,32 +176,106 @@ Base.@propagate_inbounds Base.getindex(df::SubDataFrame, row_ind::typeof(!),
...
+            throw(ArgumentError("creating new columns in a SubDataFrame that subsets " *
+                                "columns of its parent data frame is disallowed"))
+        end
+        if !(val isa AbstractVector && nrow(sdf) == length(val))
+            throw(ArgumentError("Assigned value must be a vector with length " *
+                                "equal to number of rows in the SubDataFrame"))
+        end
+        T = eltype(val)
+        newcol = Tables.allocatecolumn(Union{T, Missing}, nrow(parent(sdf)))
+        fill!(newcol, missing)
```
nalimilan, line 195: `similar` is even better when we have a reference column to use:
```suggestion
        newcol = similar(old_col, promote_type(T, S))
        copyto!(newcol, old_col)
```

src/subdataframe/subdataframe.jl
```diff
@@ -176,32 +176,106 @@ Base.@propagate_inbounds Base.getindex(df::SubDataFrame, row_ind::typeof(!),
...
+        end
+        sdf[:, col_ind] = v
+    else
+        pdf = parent(sdf)
+        p_col_ind = parentcols(index(sdf), col_ind)
+        old_col = pdf[!, p_col_ind]
+        T = eltype(old_col)
+        S = eltype(v)
+        newcol = Tables.allocatecolumn(promote_type(T, S), length(old_col))
+        newcol .= old_col
```
nalimilan, line 222:
```suggestion
        newcol = similar(old_col, promote_type(T, S))
        copyto!(newcol, old_col)
```

src/subdataframe/subdataframe.jl
nalimilan: Both have their uses. The only case in which they are identical is when creating a new column, but this is the same as for `DataFrame`.

docs/src/lib/indexing.md
nalimilan: I think `!` is what is meant here. But indeed docs for `:` are missing AFAICT.

nalimilan (commented):
Here are some comments. I haven't looked at tests yet.

> On the API, I wonder if `transform!(sdf, ...)` should return `parent(sdf)` after the transformation. This is more in-line with what we do for `transfrorm(gd::GroupedDataFrames, ...)`. We decided to automatically combine because the `ungroup` stickiness can lead to bugs in R. Maybe we should add a keyword argument to `transform!` about this?

Given that `SubDataFrame` is an ` AbstractDataFrame` I think it should continue behaving as much as possible like a `DataFrame`. But we could add a keyword argument later if that turns out to be useful (notably for piping).

docs/src/lib/indexing.md
bkamins: > But indeed docs for `:` are missing AFAICT.

In this case `:` is just `rows` and it is covered in `sdf[rows, cols] = v` rule.

> I'm not sure what's being copied here. (...) That can't be fast, if someone is doing a loop with multiple groups.

This is the point of difference of `:` and `!`:
* `!` makes a copy and makes a promotion.
* `:` is in-place and makes a conversion.

See this example that hopefully makes it clear:
```
julia> df = DataFrame(ones(3,2), :auto)
3×2 DataFrame
 Row │ x1       x2      
     │ Float64  Float64 
─────┼──────────────────
   1 │     1.0      1.0
   2 │     1.0      1.0
   3 │     1.0      1.0

julia> dfv = @view df[2:2, :]
1×2 SubDataFrame
 Row │ x1       x2      
     │ Float64  Float64 
─────┼──────────────────
   1 │     1.0      1.0

julia> dfv[:, :] = ['a' 'a']
1×2 Matrix{Char}:
 'a'  'a'

julia> df
3×2 DataFrame
 Row │ x1       x2      
     │ Float64  Float64 
─────┼──────────────────
   1 │     1.0      1.0
   2 │    97.0     97.0
   3 │     1.0      1.0

julia> df = DataFrame(ones(3,2), :auto)
3×2 DataFrame
 Row │ x1       x2      
     │ Float64  Float64 
─────┼──────────────────
   1 │     1.0      1.0
   2 │     1.0      1.0
   3 │     1.0      1.0

julia> dfv = @view df[2:2, :]
1×2 SubDataFrame
 Row │ x1       x2      
     │ Float64  Float64 
─────┼──────────────────
   1 │     1.0      1.0

julia> dfv[!, :] = ['a' 'a']
1×2 Matrix{Char}:
 'a'  'a'

julia> df
3×2 DataFrame
 Row │ x1   x2  
     │ Any  Any 
─────┼──────────
   1 │ 1.0  1.0
   2 │ a    a
   3 │ 1.0  1.0
```

src/subdataframe/subdataframe.jl
bkamins: Exactly. See my comment above. It was for `cols` selector, but it also applies to `col_ind` selector. Sometimes you want in-place operation with conversion (then do `:`) and sometimes you want new allocation with promotion (then do `!`).

docs/src/lib/indexing.md
pdeffebach: Ah I see that makes sense. 

This may end up being a real pitfall, performance wise. Someone might do 

```
for ai in unique(df.a)
    sdf = subset(df, :a => ==(ai))
    sdf[!, :b] = 100
end
```

If `unique(df.a)` is large and `nrow(df)` is large, this will cause horrendous allocations, right?

src/subdataframe/subdataframe.jl
pdeffebach: > Both have their uses. The only case in which they are identical is when creating a new column, but this is the same as for `DataFrame`.

Wait this isn't true. `df[:, :x] = y` and `df[!, :x] = y` mean different things. The first copies `y` and the second doesn't. Usually this is more performant, but in the case of `sdf[:, :x] = y`, using `!` is actually *less* performant. I'm worried this will trip up new users.

docs/src/lib/indexing.md
bkamins: Yes - it will allocate a lot. If one is sure using in-place operation is safe then `sdf[:, :b]` should be used.
The timing will not be super bad, but it will be noticeably worse than `sdf[:, :b]`.

src/subdataframe/subdataframe.jl
bkamins: The mental model should be the following:
- `df[:, :x] = y` is an in-place operation; and this is true with this PR both for `DataFrame` and `SubDataFrame` (except when `:x` is not present in `df`, but I guess you are not worried about this case)
- `df[:, :x] .= y` follows the same rules
- `df[!, :x] = y` is a column replacement operation. A new column `:x` is stored in `df`; and again with this PR this is true both for `DataFrame` and for `SubDataFrame`; the difference is that for `DataFrame` we do not copy `y`, but store it "as is", while for `SubDataFrame` we allocate a new column (it is impossible not to allocate)
- `df.x = y` the same as above
- `df[!, :x] .= y` is a column replacement operation with guaranteed copying of a column. A new column `:x` is stored in `df`; and again with this PR this is true both for `DataFrame` and for `SubDataFrame`; and here in both cases new column is allocated and data is copied
- `df.x .= y` will be the same as above when Julia 1.7 is out

This is the way we should explain the difference between `:`, `!`, and `getproperty`. The design is not about performance, but about allowing different kind of operations. The performance is just a consequence.

In summary - I think new users should not be taught what is fast and what is slow, but rather what is the objective of the operation (and in https://github.com/JuliaData/DataFrames.jl/pull/2794#discussion_r693241259 I have shown an exact case when in-place and replace differ; this is also the same for `DataFrame`).

bkamins:
> Also, could you clarify why `df.newcol = v` is not allowed? I'm sure there is a reason but I always forget.

It is allowed.

What is not allowed is `df.newcol .= v`. It will be allowed in Julia 1.7. In Julia 1.6 it is not possible, as `df.newcol` is resolved before broadcasting is invoked. Only Julia 1.7 has feature allowing to delay resolving `getproperty`.

bkamins:
> On the API, I wonder if `transform!(sdf, ...)` should return `parent(sdf)` after the transformation. (...) Maybe we should add a keyword argument to transform! about this?

In `GroupedDataFrame` we add an option to group/ungroup for performance reasons. Here, we could add a kwarg as you propose, but writing:
```
transform!(sdf, ..., parent=true)
```
is longer than just
```
parent(transform!(sdf, ...))
```

In general I would prefer to keep `transform` and `transform!` consistent in their return value.

src/subdataframe/subdataframe.jl
pdeffebach: Okay. Thanks for the summary / reminder. 

I'm still on the fence about allowing `df[!, :x]`. I think the scenario where someone would want to avoid type promotion, as in your `Int` and `Char`  example, is relatively uncommon. Comparead to people thinking `!` means speed. 

But the proposed design is internally consistent, so overall I think it's okay to move ahead.

bkamins: > is relatively uncommon

I think the crucial common case is if you want to introduce `missing` in a column that does not allow it.

---

Still, before tackling @nalimilan comments I think it is important to work out this case. I would prefer not to add `!` (and in consequence also we will not add `sdf.newcol = v`) if we feel it is not needed.

I designed a full proposal here, as my experience in the past was that when I proposed a limited functionality I was immediately asked to propose a full one 😄.

@nalimilan - can you please comment what you think here? Thank you!

bkamins: Also @matthieugomez - maybe you have an opinion here?

pdeffebach: For more background, a common Stata idiom is to go 

```
foreach i in numlist 1/100 {
    replace x = y_`i' if x == `i'
}
```

Which I would imagine a lot of users would be doing here.

bkamins: @pdeffebach the discussion prompted me with two things:
* maybe `sdf[!, :col] = v` if `eltype(v) <: eltype(sdf.col)` should not allocate the column but be in-place if `:col` is present in `sdf`? This would improve efficiency at the cost of docstring complexity. I think it is OK not co copy if not needed as we assume `parent(sdf)` owns `:col`. @nalimilan - what do you think?
* your example currently would be written (if I read it correctly):
```
for i in 1:100
    transform!(view(df, df.x .== i, :), Symbol("y_", i) => :x)
    # or
    transform!(filter(:x => ==(i), df, view=true), Symbol("y_", i) => :x)
    # or
    sdf = view(df, df.x .== i, :x)
    sdf.x = sdf[!, Symbol("y_", i)]
end
```
but probably you meant to make it even more convenient as indeed it is more verbose than Stata (do you have any specific syntax in mind)?

nalimilan: I would write this specific example as `df.x = string.("y_", df.x)`. But probably this illustrates a more general idiom? :-D

Anyway, I feel like we don't really have the choice if we want to remain consistent. The rules are already quite complex. Making `df[!, col] .= x` replace columns but `sdf[!, col] .= x` fill in-place (always or only in some particular cases) would be really hard to explain.

pdeffebach: I think I'm in favor of auto-promoting only when needed. though you are right that can lead to some difficult-to-explain rules and perhaps some performance problems.

bkamins: > I think I'm in favor of auto-promoting only when needed

This is the rule that is currently implemented (we do only promote when needed).

AFAICT what @nalimilan writes is that `sdf[!, col] = x` should always replace `col` in `parent(sdf)` no matter if promotion is needed or not.

What we need to do is:
1. clearly explain in the docs the consequences of `!` on LHS in `sdf`.
2. change the implementation of `select!` and `transform!` so that they do not copy if not needed (I will change this in this PR)

docs/src/lib/indexing.md
```diff
@@ -138,15 +138,26 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
                        `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
                        (in the latter case column names must match);
 
-`setindex!` on `SubDataFrame`:
+`setindex!` on `SubDataFrame` (not created with `:` as column selector):
```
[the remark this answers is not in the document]
bkamins: yes, fixed.

src/dataframe/dataframe.jl
```diff
@@ -801,8 +813,12 @@ julia> insertcols!(df, 2, :c => 2:4, :c => 3:5, makeunique=true)
    3 │ c         4      5      3
 ```
 """
-function insertcols!(df::DataFrame, col::ColumnIndex, name_cols::Pair{Symbol, <:Any}...;
+function insertcols!(df::AbstractDataFrame, col::ColumnIndex, name_cols::Pair{Symbol, <:Any}...;
                      makeunique::Bool=false, copycols::Bool=true)
+    if !is_column_adding_allowed(df)
```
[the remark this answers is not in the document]
bkamins: OK - changed.

src/dataframe/dataframe.jl
```diff
@@ -862,52 +878,63 @@ function insertcols!(df::DataFrame, col::ColumnIndex, name_cols::Pair{Symbol, <:
...
 
+        if df isa DataFrame
+            dfp = df
+        else
+            dfp = parent(df)
+            T = eltype(item_new)
+            newcol = Tables.allocatecolumn(Union{T, Missing}, nrow(dfp))
+            fill!(newcol, missing)
+            newcol[rows(df)] = item_new
+            item_new = newcol
```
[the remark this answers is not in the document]
bkamins: `item_new` is used in a lot of places in this function already in the current code so I have left it. I refactored this part - hopefully it is better now - please comment. I also switched `Tables.allocatecolumn` to `similar`.

src/dataframe/dataframe.jl
```diff
@@ -862,52 +878,63 @@ function insertcols!(df::DataFrame, col::ColumnIndex, name_cols::Pair{Symbol, <:
...
+            insert!(_columns(dfp), col_ind, item_new)
         end
         col_ind += 1
     end
     return df
 end
 
-insertcols!(df::DataFrame, col::ColumnIndex, name_cols::Pair{<:AbstractString, <:Any}...;
+insertcols!(df::AbstractDataFrame, col::ColumnIndex, name_cols::Pair{<:AbstractString, <:Any}...;
                      makeunique::Bool=false, copycols::Bool=true) =
```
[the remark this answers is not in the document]
bkamins: I have added this check. It is safer to throw an error if `copycols=false` and `df isa SubDataFrame`. We can always change this later.

src/subdataframe/subdataframe.jl
bkamins: Changed to `similar`, but you probably copy-pasted wrong code here.

src/subdataframe/subdataframe.jl
bkamins: Right. The only question is if we should use `promote_type` or `Base.promote_typejoin`. What do you think?

In particular let us discuss a corner case of when `old_col` is a `CategoricalVector` holding e.g. `String` and `v` is a vector of `String`. In this case we have:
```
julia> Base.promote_typejoin(CategoricalValue{String, UInt32}, String)
Any

julia> promote_type(CategoricalValue{String, UInt32}, String)
String
```
so:
1. we get a significantly different result in both options.
2. in BOTH cases the result of `similar` will not be a `CategoricalVector` but just a `Vector`

src/subdataframe/subdataframe.jl
```diff
@@ -176,32 +176,106 @@ Base.@propagate_inbounds Base.getindex(df::SubDataFrame, row_ind::typeof(!),
...
+        idxs = index(sdf)[col_inds]
+        if size(mx, 2) != length(idxs)
+            throw(DimensionMismatch("number of selected columns ($(length(idxs))) " *
+                                    "and number of columns in " *
+                                    "matrix ($(size(mx, 2))) do not match"))
+        end
+        for (j, col) in enumerate(idxs)
+            sdf[!, col] = view(mx, :, j)
+        end
+        return sdf
```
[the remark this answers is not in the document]
bkamins: They are minimally different. The difference is that for `SubDataFrame` we will make a copy anyway when doing an assignment, so on RHS I do not make a copy.

src/subdataframe/subdataframe.jl
```diff
@@ -223,3 +297,38 @@ function DataFrame(sdf::SubDataFrame; copycols::Bool=true)
...
+    end
+
+    if !is_column_adding_allowed(sdf)
+        throw(ArgumentError("changing the sequence of column names in a SubDataFrame " *
+                            "that subsets columns of its parent data frame is disallowed"))
+    end
+
+    psdf = parent(sdf)
```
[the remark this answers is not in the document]
bkamins: ok - fixed

src/subdataframe/subdataframe.jl
```diff
@@ -223,3 +297,38 @@ function DataFrame(sdf::SubDataFrame; copycols::Bool=true)
...
+    @assert psdf isa DataFrame
+    for colname in _names(newdf)
+        sdf[!, colname] = newdf[!, colname]
+    end
```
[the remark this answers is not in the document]
bkamins: fixed and added extra explanation in the code

bkamins:
@nalimilan I have applied all the discussions we had. I have resolved all conversations that I thought I clear and left only three open.
The main issue is if we should use `promote_type` or `Base.promote_typejoin` when doing `sdf[!, col] = v`. I have discussed in the comments and in the code the potential consequences.

src/subdataframe/subdataframe.jl
```diff
@@ -223,3 +298,55 @@ function DataFrame(sdf::SubDataFrame; copycols::Bool=true)
...
+
+    for colname in _names(newdf)
+        oldcol = sdf[!, colname]
+        newcol = newdf[!, colname]
+        # We perform an in-place operation if possible for performance.
+        # This has an additional effect that for CategoricalVector levels
+        # and ordering will be retained or not depending on which code patch is taken.
+
+        # TODO: add tests when promote_type vs Base.promote_typejoin decision is made
```
bkamins, line 330: @nalimilan - this is the performance optimization we have discussed for `select!` and `transform!`. As commented - it has a side effect for CategoricalArrays.jl but I think it is OK.

docs/src/lib/indexing.md
```diff
@@ -144,9 +144,28 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
 * `sdf[row, cols] = v` -> the same as `dfr = df[row, cols]; dfr[:] = v` in-place;
 * `sdf[rows, col] = v` -> set rows `rows` of column `col`, in-place; `v` must be an abstract vector;
 * `sdf[rows, cols] = v` -> set rows `rows` of columns `cols` in-place;
-                           `v` can be an `AbstractMatrix` or `v` can be `AbstractDataFrame` when column names must match;
+                           `v` can be an `AbstractMatrix` or `v` can be `AbstractDataFrame`
+                           when column names must match;
```
nalimilan, line 148:
```suggestion
                           in which case column names must match;
```

docs/src/lib/indexing.md
```diff
@@ -144,9 +144,28 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
...
+                       `v` must be an `AbstractMatrix` or an `AbstractDataFrame`
+                       (in the latter case column names must match);
+                       filtered-out rows in newly created vectors are filled with
+                       values already present in respective columns;
+
+!!! note
+
+    The rules above mean that `sdf[:, col] = v` is an in-place operation if `col` is present in `sdf`,
+    therefore it will be fast in general. On the other hand using `sdf[!, col] = v`
+    or `sdf.col = v` will always allocate a new vector which is more expensive computationally.
```
nalimilan, line 167:
```suggestion
    or `sdf.col = v` will always allocate a new vector, which is more expensive computationally.
```

docs/src/lib/indexing.md
```diff
@@ -144,9 +144,28 @@ so it is unsafe to use it afterwards (the column length correctness will be pres
 * `sdf[row, cols] = v` -> the same as `dfr = df[row, cols]; dfr[:] = v` in-place;
 * `sdf[rows, col] = v` -> set rows `rows` of column `col`, in-place; `v` must be an abstract vector;
 * `sdf[rows, cols] = v` -> set rows `rows` of columns `cols` in-place;
-                           `v` can be an `AbstractMatrix` or `v` can be `AbstractDataFrame` when column names must match;
+                           `v` can be an `AbstractMatrix` or `v` can be `AbstractDataFrame`
+                           when column names must match;
+* `sdf[!, col] = v` -> replaces `col` with `v` with copying; if `col` is present in `sdf`
```
nalimilan, line 149: Why isn't `:` also mentioned here?

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
nalimilan: OK, but `item_new_df` sounds like a weird name. How about something like `item_new_orig`?

src/subdataframe/subdataframe.jl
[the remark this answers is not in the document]
nalimilan: I see. Why not merge them and just use a branch for each type?

src/subdataframe/subdataframe.jl
nalimilan: I think `promote_type` is better. We use it in other places IIRC, it's better to get homogeneous types in general. In the CategoricalArrays example, `Vector{String}` is better than `Vector{Any}` (which is why `promote_type` methods are defined that way).

src/subdataframe/subdataframe.jl
nalimilan: Ah, but this isn't just an optimization, as the difference can be observed from the outside, right? I think we should only apply tricks that are completely invisible for users, or the behavior will be to complex. Apparently I was wrong in thinking that `select!` and `transform` are amenable to such optimizations.

src/subdataframe/subdataframe.jl
bkamins: OK - I keep `promote_type` then. I will ammend the documentation to reflect this.

src/subdataframe/subdataframe.jl
bkamins: Yes - it is visible from outside if the column has metadata. I will restrict the optimization to `Vector`, as `Vector` has no metadata (as opposed to `PooledVector` or `CategoricalVector`)

src/subdataframe/subdataframe.jl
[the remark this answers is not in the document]
bkamins: I propose to leave it as-is for now. Otherwise we should move these definitions to abstractdataframe/abstractdataframe.jl where we avoided defining `setindex!` currently.

In the future, if we decide to refactor codebase so that we first define types in a single file e.g. `types.jl`, and then define methods for all types arranged cleanly (e.g. a separate file for `setindex!`) I would merge them.

I left a comment in the file that these methods should be merged in the future.

bkamins:
> Why isn't `:` also mentioned here?

It is mentioned at the top in the following part:

> The rules for a valid type of index into a row are the following:
> * a value, later denoted as `row`:
>     * an `Integer` that is not `Bool`;
> * a vector, later denoted as `rows`:
>     * a vector of `Integer` other than `Bool` (does not have to be a subtype of `AbstractVector{<:Integer}`);
>     * a vector of `Bool` that has to be a subtype of `AbstractVector{Bool}`;
>     * a `Not` expression;
>     * a colon literal `:`;
> * an exclamation mark `!`.

src/dataframe/dataframe.jl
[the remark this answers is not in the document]
bkamins: OK - changed.

test/subdataframe_mutation.jl
```diff
@@ -0,0 +1,1647 @@
...
+end
+
+@testset "promote_type tests" begin
+    df = DataFrame(a=1:4)
+    sdf = @view df[1:1, :]
+    sdf[!, 1] = [1.5]
+    @test df.a == [1.5, 2, 3, 4]
+    @test eltype(df.a) === Float64
+
+    # note that CategoricalVector is dropped as
```
bkamins, line 1600: @nalimilan - this part of tests probably deserves your special attention. Thank you!

bkamins:
@nalimilan - this should be good to have a look at again. Thank you!

src/subdataframe/subdataframe.jl
nalimilan: Isn't it visible also for `Vector`? That is, if one holds a reference to the column, mutating it rather than replacing it can have consequences. (I'm not overly concerned about `PooledVector` and `CategoricalVector` if that only makes a difference in weird cases.)

test/broadcasting.jl
```diff
@@ -148,7 +148,11 @@ end
     @test df[:, 2:end] == refdf[:, 2:end]
 
     dfv = @view df[1:2, 2:end]
-    @test_throws ArgumentError dfv[!, 1] .+= 1
+    dfv[!, 1] .+= 100
+    @test df.x2 == [104.5, 105.5, 6.5]
+    df.x1 -= [1, 1, 1]
+    df.x2 -= [100, 100, 0]
+    @test df == refdf
```
nalimilan, line 155: Why not simply compare `df` with a literal `DataFrame`?

test/broadcasting.jl
```diff
@@ -220,7 +224,8 @@ end
     @test df[:, Not("x1")] == refdf[:, 2:end]
 
     dfv = @view df[1:2, 2:end]
-    @test_throws ArgumentError dfv[!, 1] .+= [0, 1] .+ 1
+    dfv[!, 1] .+= [0, 1] .+ 1
+    @test df.x2 == [5.5, 7.5, 6.5]
```
nalimilan, line 228: Check the contents of the whole `df` while you're at it (here and elsewhere)? For example, the column could be re-added in the wrong place...

test/broadcasting.jl
```diff
@@ -1599,8 +1609,13 @@ end
     @test v1 == [100.0, 100.0, 100.0]
 
     df = view(copy(refdf), :, :)
-    @test_throws ArgumentError df.newcol .= 'd'
-    @test df == refdf
+    if VERSION >= v"1.7"
+        df.newcol .= 'd'
+        df.newcol == fill('d', 3)
```
nalimilan, line 1614:
```suggestion
        @test df.newcol == fill('d', 3)
```

test/indexing.jl
```diff
@@ -1299,10 +1299,17 @@ end
         sdf[:, 1] = 10:12
         @test df == DataFrame(a=10:12, b=4:6, c=7:9)
         @test_throws MethodError sdf[:, 1] = ["a", "b", "c"]
-        @test_throws ArgumentError sdf[:, :z] = ["a", "b", "c"]
         @test_throws BoundsError sdf[:, 4] = ["a", "b", "c"]
         @test_throws DimensionMismatch sdf[:, 1] = [1]
         @test_throws MethodError sdf[:, 1] = 1
+        if DataFrames.is_column_insertion_allowed(sdf)
```
nalimilan, line 1305: Any simple way to check this without relying on the internal function? That would allow checking that it's correct too.

test/indexing.jl
```diff
@@ -1843,8 +1853,11 @@ end
     @test_throws ArgumentError df.a = 1
     @test_throws ArgumentError df."a" = 1
     dfv = @view df[:, :]
-    @test_throws ArgumentError dfv.a = [1]
-    @test_throws ArgumentError dfv."a" = [1]
+    dfv.a = [5]
+    @test df == DataFrame(a=5)
```
nalimilan, line 1857:
```suggestion
    @test df == DataFrame(a=5)
    @test eltype(df.a) === Int
```

test/subdataframe_mutation.jl
```diff
@@ -0,0 +1,1647 @@
...
+end
+
+@testset "mutating SubDataFrame with assignment to [:, cols]" begin
+    for sel in (:, 1:3)
+        df = DataFrame(a=1:5, b=11:15, c=21:25)
+        sdf = @view df[[3, 2], sel]
+        tmpa = df.a
+        sdf[:, [:c, :b, :a]] = DataFrame(c=[5, 6], b=[1.0, 2.0], a=[13, 12])
+        @test df == DataFrame(a=[1, 12, 13, 4, 5],
+                            b=[11.0, 2.0, 1.0, 14.0, 15.0],
```
nalimilan, line 1149: Incorrect indentation here and below.

nalimilan (commented):
Impressive set of tests!

src/subdataframe/subdataframe.jl
bkamins: It can have consequences if there are aliases to this vector. But OK - let us just copy always and stick to "safety first" rule (if someone wants it fast it is always easy-enough to achieve with indexing).

bkamins:
> Impressive set of tests!

Frankly - we are at the level of complexity of the logic and number of corner cases where I need the tests to convince myself that I can make a PR public.

test/broadcasting.jl
bkamins: ok - changed

test/broadcasting.jl
bkamins: done

test/indexing.jl
bkamins: added

test/subdataframe_mutation.jl
bkamins: fixed

test/subdataframe_mutation.jl
```diff
@@ -1590,4 +1590,58 @@ end
...
+    @test df.a isa CategoricalVector{String}
+    @test df.a == ["a", "2", "3", "4"]
+    # we first copy old data and then add new data so "1" is in levels
+    # although it is not present in df.a
+    @test levels(df.a) == ["1", "2", "3", "4", "a"]
+
+    df = DataFrame(a=1:4)
+    a = df.a
+    sdf = @view df[1:1, :]
+    select!(sdf, :a => (x -> x) => :a)
```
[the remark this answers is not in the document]
bkamins: done

bkamins:
All comments should be now resolved. Thank you!

nalimilan (approved):

NEWS.md
```diff
@@ -18,6 +22,30 @@
...
   ([#2812](https://github.com/JuliaData/DataFrames.jl/pull/2812))
+
+* Allow adding new columns to a `SubDataFrame` created with `:` as column selector
+  ([#2794](https://github.com/JuliaData/DataFrames.jl/pull/2794)).
+
+  If `sdf` is a `SubDataFrame` created with `:` as a column selector then
+  `insertcols!`, `setindex!`, and broadcasted assignment allow for creation
+  of new columns, automatically filling filtered-out rows with `missing` values;
+
+* Allow replacing existing columns in a `SubDataFrame` with `!` as row selector in assignment and broadcasted assignment
```
bkamins, line 33:
```suggestion
* Allow replacing existing columns in a `SubDataFrame` with `!` as row selector
  in assignment and broadcasted assignment
```

bkamins:
Thank you!

@pdeffebach - now it is probably the time to experiment with this functionality in combination with what DataFramesMeta.jl could provide.

Issue #2792, opened 2021-06-19
Clean up precompile statements


bkamins:
Before 1.2 release precompile statements need to be re-generated and tests should be cleaned up not to throw warnings under Julia 1.8.

I will do it once https://github.com/JuliaData/DataFrames.jl/pull/2791 is merged.

Issue #2788, opened 2021-06-13
Sync with DataAPI.jl 1.7 release


bkamins:
See https://github.com/JuliaData/DataAPI.jl/issues/37 (this is related to `join*` methods)

bkamins:
see https://github.com/JuliaData/DataAPI.jl/pull/38

Issue #2783, opened 2021-06-09
Add a method to add/insert empty columns


hhaensel:
Currently adding empty columns must be done by hand AFAIK, e.g.
```
df.new = Vector{Union{Missing, Int}}(missing, nrow(df))
```
This is difficult to remember and certainly not easily accessible for beginners.
I propose to define
```
insertemptycols!(df::DataFrame, col::ColumnIndex, name_cols::Pair{<:Union{Symbol, AbstractString}, <:Union{DataType, Union}}...; makeunique::Bool=false) = 
    insertcols!(df, col, (name => Vector{Union{Missing, T}}(missing, nrow(df)) for (name, T) in name_cols)...; makeunique=makeunique)
    
insertemptycols!(df::DataFrame, name_cols::Pair{<:Union{Symbol, AbstractString}, <:Union{DataType, Union}}...; makeunique::Bool=false) = 
    insertemptycols!(df, ncol(df) + 1, name_cols...; makeunique=makeunique)
```
in order to provide, e.g.

```
julia> df = DataFrame(a=1:10);

julia> insertemptycols!(df, :new => Int)
10×2 DataFrame
 Row │ a      new    
     │ Int64  Int64?  
─────┼────────────────
   1 │     1  missing 
   2 │     2  missing 
   3 │     3  missing 
   4 │     4  missing 
   5 │     5  missing 
   6 │     6  missing 
   7 │     7  missing 
   8 │     8  missing 
   9 │     9  missing 
  10 │    10  missing 
```
I'll file a respective PR for testing. If this sounds like a reasonable feature, I'd be happy to also provide a docstring 😉

bkamins:
```
insertcols!(hh, 2, [col => missings(Float64, nrow(hh)) for col in [:x, :y, :z]]...)
```
is indeed longer, but doable.

The key question is: do we want to add another function into the ecosystem? We try to minimize the number of verbs we provide to only these that are really needed and are not very niche. The issue is that we should answer ourselves if `insertemptycols!` would be:
1. used frequently enough (i.e. do people really create such columns in practice often? In what usage scenarios does it happen? as maybe the need comes from not knowing other ways to work with DataFrames.jl)
2. discovered by the users

to warrant its addition. Note that e.g. you were unaware of the `missings` function - this is exactly the problem we face.

I am leaving this question open, as I would some input from you and the community to judge this issue.

hhaensel:
Fair enough!

[...]

PR #2797, opened 2021-06-22
Use standard Tables.Schema constructor instead of constructing directly

quinnj:
This is part of fixing errors like
https://github.com/JuliaData/CSV.jl/issues/635 in addition to the
changes to support really wide tables in
https://github.com/JuliaData/Tables.jl/pull/241. Luckily, there aren't
many cases I've found across Tables.jl implementations that make working
with really wide tables impossible, but this was a key place where for
really wide tables, we want the names/types to be stored as `Vector`s
instead of `Tuple`/`Tuple{}` in `Tables.Schema`. This shouldn't have any
noticeable change/affect for non-wide DataFrames and should be covered
by existing tests.

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -21,7 +21,7 @@ end
 Tables.columnindex(df::Union{AbstractDataFrame, DataFrameRow}, idx::AbstractString) =
     columnindex(df, Symbol(idx))
 
-Tables.schema(df::AbstractDataFrame) = Tables.Schema{Tuple(_names(df)), Tuple{[eltype(col) for col in eachcol(df)]...}}()
+Tables.schema(df::AbstractDataFrame) = Tables.Schema(_names(df), [eltype(col) for col in eachcol(df)])
 Tables.materializer(df::AbstractDataFrame) = DataFrame
 
 Tables.getcolumn(df::AbstractDataFrame, i::Int) = df[!, i]


bkamins (approved):
I understand this change does not need https://github.com/JuliaData/Tables.jl/pull/241 (it works both on the current release of Tables.jl). Right?

src/other/tables.jl
```diff
@@ -21,7 +21,7 @@ end
 Tables.columnindex(df::Union{AbstractDataFrame, DataFrameRow}, idx::AbstractString) =
     columnindex(df, Symbol(idx))
 
-Tables.schema(df::AbstractDataFrame) = Tables.Schema{Tuple(_names(df)), Tuple{[eltype(col) for col in eachcol(df)]...}}()
+Tables.schema(df::AbstractDataFrame) = Tables.Schema(_names(df), [eltype(col) for col in eachcol(df)])
```
bkamins, line 24: `_names(df)` returns a `SubArray` if `df` is a `SubDataFrame`. Is this acceptable by `Tables.schema`?

Also `_names(df)` for a `DataFrame` returns an internal vector (without copying). Is this safe?

In short. Maybe it should be ` propertynames(df)` instead (which guarantees a `Vector{Symbol}` that is always a copy)

quinnj: Nah, Tables.jl will make its own copy; it accepts any iterable and makes its own `Vector{Symbol}`.

bkamins: Then I just "approve" :).

quinnj:
> I understand this change does not need JuliaData/Tables.jl#241 (it works both on the current release of Tables.jl). Right?

Correct. The goal in the Tables.jl change was no user-visible change. In practice this should affect a very small % of tables. But yes, `Tables.Schema(names, types)` has always been the preferred `Schema` constructor; I think we've just supported Tables.jl in DataFrames.jl for so long that we originally did the shortcut before the constructor was standardized.

quinnj:
Not urgent to release, but should be safe to include in a patch release at any time.

bkamins:
The 1.2 release is on the way. It depends on the availability of @nalimilan (I know he is busy).

What I need is merging:
* https://github.com/JuliaData/DataFrames.jl/pull/2791 (here a decision by @nalimilan is needed)
* https://github.com/JuliaData/DataAPI.jl/pull/38 (and a release) and a follow up https://github.com/JuliaData/DataFrames.jl/issues/2788
* https://github.com/JuliaData/DataFrames.jl/issues/2792

and we can make 1.2 release. Optionally we could include (as these are relatively simple things to implement; the only thing is decision, but we can add them 1.3 equally well, so they are non-blocking):
* https://github.com/JuliaData/DataFrames.jl/pull/2794 (unfinished and probably will need discussion)
* https://github.com/JuliaData/DataFrames.jl/issues/2783 (here a decision is needed if we want pseudo-broadcasting on 0-dimensional objects to retain their `eltype` - I am pretty convinced we should go this way, but the discussion is open)

PR #2807, opened 2021-06-26
up DataAPI.jl to 1.7 and CategoricalArrays.jl to 0.10.0

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2788


test/join.jl
```diff
@@ -355,8 +355,8 @@ end
     @test levels(leftjoin(A, B, on=:b).b) == ["d", "c", "b", "a"]
     @test levels(leftjoin(B, A, on=:b).b) == ["a", "b", "c"]
     @test levels(rightjoin(A, B, on=:b).b) == ["a", "b", "c"]
-    @test levels(rightjoin(B, A, on=:b).b) == ["d", "c", "b", "a"]
```
bkamins, line 358: @nalimilan - these changes follow CategoricalArrays.jl 0.10.0 release, so a check would be welcome.

Project.toml
```diff
@@ -23,9 +23,9 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
-CategoricalArrays = "0.9.3"
+CategoricalArrays = "0.10.0"
```
nalimilan, line 26: How about allowing both since we support both? That will help ensuring that people don't use old DataFrames versions due to conflicts.

test/join.jl
```diff
@@ -355,8 +355,8 @@ end
...
-    @test levels(outerjoin(B, A, on=:b).b) == ["a", "b", "d", "c"]
+    @test levels(rightjoin(B, A, on=:b).b) ==  ["d", "c", "b", "a"]
```
nalimilan, line 358:
```suggestion
    @test levels(rightjoin(B, A, on=:b).b) == ["d", "c", "b", "a"]
```

test/join.jl
nalimilan: Interesting. Why aren't levels the same for inner and outer joins? `["d", "c", "b", "a"]` would sound like a good choice since `levels(A.b)` is a superset of `levels(B.b)`.

That said, since the old behavior wasn't ideal either, the new one is OK -- both are kind of arbitrary.

Project.toml
bkamins: I assumed that this will not be conflicting as we use CategoricalArrays.jl only in `[extras]` so it would be used only when testing the package.

@KristofferC - could you please confirm this is correct.

Also I assume soon we will have a new Julia LTS and then we will drop 1.0 support, when the `[extras]` section will be gone.

bkamins:
> Why aren't levels the same for inner and outer joins? 

The reason is (under CategoricalArrays.jl 0.10):
```
julia> x = categorical(["a", "b", "c", "d"])
4-element CategoricalArray{String,1,UInt32}:
 "a"
 "b"
 "c"
 "d"

julia> levels!(x, ["d", "c", "b", "a"])
4-element CategoricalArray{String,1,UInt32}:
 "a"
 "b"
 "c"
 "d"

julia> y = categorical(["a", "b", "c"])
 "a"
 "b"
 "c"

julia> levels(x)
4-element Vector{String}:
 "d"
 "c"
 "b"
 "a"

julia> levels(y)
3-element Vector{String}:
 "a"
 "b"
 "c"

julia> levels([x; y])
4-element Vector{String}:
 "d"
 "c"
 "b"
 "a"

julia> levels([y; x])
4-element Vector{String}:
 "d"
 "a"
 "b"
 "c"
```
so as you can see the order of vectors matters (in this example in `vcat`).

nalimilan:
Ah right that's because the orders are not compatible, so we have to make a choice.

Project.toml
nalimilan: Indeed. I've just checked, it's perfectly fine to install CategoricalArrays 0.8 and DataFrames 1.1 in the same environment.

nalimilan (approved):

bkamins:
Thank you!

PR #2805, opened 2021-06-24
Add some clarifying comments on copycols for Tables.jl inputs

quinnj:
Should help address some confusion in #2799 around `copycols=nothing`
default for `DataFrame` constructor when Tables.jl input is provided.

PR #2809, opened 2021-06-28
improve subset! for GroupedDataFrame

### LINKED ISSUE #2808: subset!(gd::GroupedDataFrame, ...) should make sure `gd` still works after
Here's an example

```
julia> df = DataFrame(a = [1, 1, 2, 2], b = [1, 2, 3, 4]);

julia> gd = groupby(df, :a);

julia> subset!(gd, :b => (b -> b .> first(b)))
2×2 DataFrame
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2
   2 │     2      4

julia> gd
GroupedDataFrame with 2 groups based on key: a
First Group (2 rows): a = 1
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2
   2 │     2      4Error showing value of type GroupedDataFrame{DataFrame}:
ERROR: BoundsError: attempt to access 2×2 DataFrame at index [[3, 4], :]
```

`gd` is worthless after a call to `subset!` maybe we can repair it? If it isn't too costly.

bkamins:
Fixes #2808

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,8 @@
 * correctly handle selectors of the form `:col => AsTable` and `:col => cols`
   by expanding a single column into multiple columns
   ([#2780](https://github.com/JuliaData/DataFrames.jl/pull/2780))
-* if `subset!`
+* if `subset!` is passed a `GroupedDataFrame` the grouping in the passed object
+  gets updated to reflect rows removed from the parent data frame
   ([#2809](https://github.com/JuliaData/DataFrames.jl/pull/2809))
 
 ## Bug fixes
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@
 * correctly handle selectors of the form `:col => AsTable` and `:col => cols`
   by expanding a single column into multiple columns
   ([#2780](https://github.com/JuliaData/DataFrames.jl/pull/2780))
+* if `subset!`
+  ([#2809](https://github.com/JuliaData/DataFrames.jl/pull/2809))
 
 ## Bug fixes
 
--- a/src/abstractdataframe/subset.jl
+++ b/src/abstractdataframe/subset.jl
@@ -211,10 +211,8 @@ If `ungroup=false` the resulting data frame is re-grouped based on the same
 grouping columns as `gdf` and a `GroupedDataFrame` is returned.
 
 If `GroupedDataFrame` is subsetted then it must include all groups present in the
-`parent` data frame, like in [`select!`](@ref). Also note that in general
-the parent of the passed `GroupedDataFrame` is mutated in place, which means
-that the `GroupedDataFrame` is not safe to use as most likely it will
-get corrupted due to row removal in the parent.
+`parent` data frame, like in [`select!`](@ref). In this case the passed
+`GroupedDataFrame` is updated to have correct groups after its parent is updated.
 
 See also: [`subset`](@ref), [`filter!`](@ref), [`select!`](@ref)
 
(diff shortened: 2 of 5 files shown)


bkamins:
I want to include this fix in 1.2 release. So apart from this PR we need #2805 and #2792 and we are good to make the release.

NEWS.md
```diff
@@ -11,6 +11,8 @@
 * correctly handle selectors of the form `:col => AsTable` and `:col => cols`
   by expanding a single column into multiple columns
   ([#2780](https://github.com/JuliaData/DataFrames.jl/pull/2780))
+* if `subset!`
```
nalimilan, line 14: Seems like a few words are missing. :-D

bkamins: I hoped it was clear what the PR does 😆.

BTW: it turns out that this change was needed. The check I added to `GroupedDataFrame` consistency already caught one bug in our testsets :).

nalimilan (approved):

bkamins:
Thank you!

raphaelsaavedra:
Hi @bkamins! I'm hitting the recently added AssertionError because I do *intentionally* append stuff to the GroupedDataFrame with the intention of modifying the parent DataFrame. I can rewrite my code of course, but I think the current way was pretty simple and elegant.

Short example:

```julia
    datetimes = unique(df.datetime)
    for gr in groupby(df, :id)
        # If the data is missing entries for some hours of the day, append some stuff / do imputation
        if nrow(gr) < 24
            missing_hours = setdiff(datetimes, df.datetime)
            # Append stuff to `gr` for the missing hours
        end
    end
```

I wonder if it would make sense to have the option to disable this check in case the user is intentionally modifying the parent DataFrame.

raphaelsaavedra:
Also, it seems this is a breaking change, so the semver should have been changed to 2.0, right?

pdeffebach:
No. this only resulted in errors before, and taking something from an error to a non-error does not merit a breaking release.

nickrobinson251:
>  it seems this is a breaking change

It seems the addition of the `@assert` was accidentally breaking, since this seems to have been allowed in v1.1 e.g.
```jl
julia> df = DataFrame(x = [1, 2, 3], y = [1, 2, 2])
3×2 DataFrame
 Row │ x      y
     │ Int64  Int64
─────┼──────────────
   1 │     1      1
   2 │     2      2
   3 │     3      2

julia> for g in groupby(df, :y)
           nrow(g) == 1 && append!(df, DataFrame(x=[99], y=first(g.y)))
       end

julia> df
4×2 DataFrame
 Row │ x      y
     │ Int64  Int64
─────┼──────────────
   1 │     1      1
   2 │     2      2
   3 │     3      2
   4 │    99      1

```
Perhaps the `@assert` can be removed? (or possibly changed to an `@warn`?)

nickrobinson251:
This same code in v1.2 thows an error (which makes it breaking) ...but also it throws _after_ the DataFrame is mutated (so it's no "safer" ad the DataFrame gets "corrupted" anyway)

```jl
julia> df = DataFrame(x = [1, 2, 3], y = [1, 2, 2])
3×2 DataFrame
 Row │ x      y
     │ Int64  Int64
─────┼──────────────
   1 │     1      1
   2 │     2      2
   3 │     3      2

julia> for g in groupby(df, :y)
           nrow(g) == 1 && append!(df, DataFrame(x=[99], y=first(g.y)))
       end
ERROR: AssertionError: The current number of rows in the parent data frame is 4 and it does not match the number of rows it contained when GroupedDataFrame was created which was 3. The number of rows in the parent data frame has likely been changed unintentionally (e.g. using subset!, filter!, delete!, push!, or append! functions).
Stacktrace:
 [1] getproperty(gd::GroupedDataFrame{DataFrame}, f::Symbol)
   @ DataFrames ~/.julia/packages/DataFrames/vQokV/src/groupeddataframe/groupeddataframe.jl:252
 [2] length
   @ ~/.julia/packages/DataFrames/vQokV/src/groupeddataframe/groupeddataframe.jl:389 [inlined]
 [3] iterate(gd::GroupedDataFrame{DataFrame}, i::Int64)
   @ DataFrames ~/.julia/packages/DataFrames/vQokV/src/groupeddataframe/groupeddataframe.jl:392
 [4] top-level scope
   @ ./REPL[19]:2

julia> df
4×2 DataFrame
 Row │ x      y
     │ Int64  Int64
─────┼──────────────
   1 │     1      1
   2 │     2      2
   3 │     3      2
   4 │    99      1
```

raphaelsaavedra:
For the reasons @nickrobinson251 presented, I think the warn suggestion makes a lot of sense, and has the bonus of making this nonbreaking.

bkamins:
> Also, it seems this is a breaking change, so the semver should have been changed to 2.0, right?

and

> It seems the addition of the @assert was accidentally breaking, since this seems to have been allowed in v1.1 e.g.

I consider it non breaking, as we explicitly warn in the documentation that if parent of view is modified the view might get invalidated. But this is a linguistic discussion. Note that the fact that your code works is using undocumented design of `GroupedDataFrame`; if we used a different way to index it (e.g. counting from the last not the first row - I know it makes no sense, but I just want to highlight the fact) the changes you propose would invalidate the loop. Essentially you are using the fact that appending to a parent does not invalidate the `GroupedDataFrame`.

> Append stuff to `gr` for the missing hours

I am not sure what you mean by this, as `gr` is a `SubDataFrame`, so you cannot append anything to it.

> ...but also it throws after the DataFrame is mutated (so it's no "safer" as the DataFrame gets "corrupted" anyway)

`DataFrame` is not corrupted. Only `GroupedDataFrame` gets corrupted.

---

In general the decision to add `@assert` was conscious as I believed that case:

> `GroupedDataFrame` got corrupted because user modified its parent by accident

is more frequent than

> user intentionally modifies the parent and is sure that it will not corrupt `GroupedDataFrame`

In particular note that the former case can be easily handled using:
```
groupby(copy(df), :y)
```
I am aware that it uses more memory, so in some cases it might be prohibitive to perform this operation, but in "typical cases" `copy(df)` is fast and non problematic.

---

Having said that I have opened https://github.com/JuliaData/DataFrames.jl/pull/2811 reverting the change.
I am OK to remove this assertion, but I personally feel that it is not safe to use `GropedDataFrame` after its parent is mutated.
However, I always try to think that the users know what they want - and clearly you both have hit an issue in this case, so there is a merit to your comments.
Could you please comment in the PR (so that @nalimilan also gets notified and can comment + we can make an informed decision - I hope we can decide on this issue quickly)?
Thank you!

PR #2811, opened 2021-07-02
Document GroupedDataFrame consistency check

bkamins:
Following the discussion in https://github.com/JuliaData/DataFrames.jl/pull/2809 this PR removes the `@assert` checking consistency of `GroupedDataFrame`.

### DIFF
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -562,8 +562,7 @@ julia> push!(df, [3])
    2 │     2
    3 │     3
 
-julia> gd
-Error showing value of type GroupedDataFrame{DataFrame}:
+julia> gd[1]
 ERROR: AssertionError: The current number of rows in the parent data frame is 3 and it does not match the number of rows it contained when GroupedDataFrame was created which was 2. The number of rows in the parent data frame has likely been changed unintentionally (e.g. using subset!, filter!, delete!, push!, or append! functions).
 ```
 
@@ -604,17 +603,10 @@ julia> push!(df, [3])
    2 │     2
    3 │     3
 
-julia> gd
-GroupedDataFrame with 2 groups based on key: id
-First Group (1 row): id = 1
+julia> gd[1]
+1×1 SubDataFrame
  Row │ id
      │ Int64
 ─────┼───────
    1 │     1
-⋮
-Last Group (1 row): id = 2
- Row │ id
-     │ Int64
-─────┼───────
-   1 │     2
 ```
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -526,3 +526,98 @@ julia> combine(gd, valuecols(gd) .=> mean)
    3 │ Iris-virginica              6.588            2.974             5.552
                                                                 1 column omitted
 ```
+
+Note that `GroupedDataFrame` is a view and therefore grouping columns of its
+parent data frame should not be mutated. In particular, if number or rows
+of the parent changes then an error is thrown when `GroupedDataFrame` is used:
+```jldoctest sac
+julia> df = DataFrame(id=1:2)
+2×1 DataFrame
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     1
+   2 │     2
+
+julia> gd = groupby(df, :id)
+GroupedDataFrame with 2 groups based on key: id
+First Group (1 row): id = 1
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     1
+⋮
+Last Group (1 row): id = 2
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     2
+
+julia> push!(df, [3])
+3×1 DataFrame
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     1
+   2 │     2
+   3 │     3
+
+julia> gd
+Error showing value of type GroupedDataFrame{DataFrame}:
+ERROR: AssertionError: The current number of rows in the parent data frame is 3
+and it does not match the number of rows it contained when GroupedDataFrame was
+created which was 2. The number of rows in the parent data frame has likely been
+changed unintentionally (e.g. using subset!, filter!, delete!, push!, or append! functions).
+```
+
+Sometimes it is useful to append the rows to the source data frame of
+`GroupedDataFrame`, without affecting the rows used for grouping.
+In such a scenario you can create the grouped data frame using a `view`
+of the parent data frame to avoid the error:
+
+```jldoctest sac
+julia> df = DataFrame(id=1:2)
+2×1 DataFrame
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     1
+   2 │     2
+
+julia> gd = groupby(view(df, :, :), :id)
+GroupedDataFrame with 2 groups based on key: id
+First Group (1 row): id = 1
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     1
+⋮
+Last Group (1 row): id = 2
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     2
+
+julia> push!(df, [3])
+3×1 DataFrame
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     1
+   2 │     2
+   3 │     3
+
+julia> gd
+GroupedDataFrame with 2 groups based on key: id
+First Group (1 row): id = 1
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     1
+⋮
+Last Group (1 row): id = 2
+ Row │ id
+     │ Int64
+─────┼───────
+   1 │     2
+```
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -564,10 +564,7 @@ julia> push!(df, [3])
 
 julia> gd
 Error showing value of type GroupedDataFrame{DataFrame}:
-ERROR: AssertionError: The current number of rows in the parent data frame is 3
-and it does not match the number of rows it contained when GroupedDataFrame was
-created which was 2. The number of rows in the parent data frame has likely been
-changed unintentionally (e.g. using subset!, filter!, delete!, push!, or append! functions).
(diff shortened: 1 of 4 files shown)


bkamins:
As an afterthought - maybe instead of removing the assertion we should change it to:
```
@assert length(getfield(gd, :groups)) <= nrow(getfield(gd, :parent)) corrupt_msg(gd)
```
so that appending to the parent does not trigger the assertion, but removing them does.
@raphaelsaavedra + @nickrobinson251: what do you think (I do not see a scenario when dropping rows might be useful).

nickrobinson251:
Thanks!

> so that appending to the parent does not trigger the assertion, but removing them does

I don't think i have a strong opinion on this add/remove difference in particular, although I _think_ i'd find it surprising to discover adding rows allowed but removing them not.

bkamins:
> i'd find it surprising to discover adding rows allowed but removing them not.

The reason is that we use `@inbounds` in the code. When you ADD rows you will stay within bounds. When you remove rows you will be out of bounds, ans with `@inbounds` it might lead to crashing Julia.

nickrobinson251:
Oh, yes I think it is better to remove the assertion.
Whilst I agree it's not a particularly safe pattern, I don't think it should be an error condition (not least because it's useful, and apparently relied upon by some internal code at Invenia).

Thanks for being so fast to respond on this!

bkamins:
> and apparently relied upon by some internal code at Invenia

OK - then I will merge this PR and make a release after I get an approval. I will just push one more commit to this PR improving a documentation of `GroupedDataFrame`.

pdeffebach:
I want to chime in and say that I don't think this should be merged. 

I think the docs stating "Modifying the underlying data frame may corrupt a grouped data frame" implies that we don't support grouped data frames that become out of sync with their parent data frame. Given this implication, we can throw an error when they get out of sync. Appending to a data frame past its grouping indices should be considered undefined behavior.

raphaelsaavedra:
Just to be clear, I think it would be pretty simple to rewrite our code not to rely on this particular interaction, so I don't think that should be the deciding factor. I commented on the PR because I thought it might have been an unintentional effect of the assertion and possibly an unintentional breaking change. You guys are certainly more qualified to decide on the matters of this package, so feel free to choose what you prefer :) and thanks for the quick response!

bkamins:
Thank you all for the comments. We have the following tension here:

On one side we do not strictly need the `@assert` - all will work without it and I can understand that advanced users, like Invenia, might be relying on undefined behavior to e.g. improve performance of operations.

On the other side the reason why the `@assert` was added was to safeguard the users, mostly newcomers, from unintentionally mutating the parent without learning that the `GroupedDataFrame` got corrupted.

So we have flexibility vs safety tension here. Let us wait to hear what @nalimilan thinks. Maybe @oxinabox would also have an opinion here.

Just to summarize what I normally would recommend (independent on what we decide to do in this PR). Instead of:
```
julia> for g in groupby(df, :y)
           nrow(g) == 1 && append!(df, DataFrame(x=[99], y=first(g.y)))
       end
```
One should write (simplest):
```
julia> for g in groupby(copy(df), :y)
           nrow(g) == 1 && append!(df, DataFrame(x=[99], y=first(g.y)))
       end
```
or (more efficient)
```
julia> aux_df = DataFrame()
julia> for g in groupby(df, :y)
           nrow(g) == 1 && append!(aux_df, DataFrame(x=[99], y=first(g.y)))
       end
julia> append!(df, aux_df) # after the loop is done append the rows
```
Both patterns are strictly preferred to just doing an update of the parent within the loop.

nalimilan:
> As an afterthought - maybe instead of removing the assertion we should change it to:
> 
> ```
> @assert length(getfield(gd, :groups)) <= nrow(getfield(gd, :parent)) corrupt_msg(gd)
> ```
> 
> so that appending to the parent does not trigger the assertion, but removing them does.

Maybe we could adopt a hybrid solution: use the above `@assert` since that can prevent invalid memory accesses (and therefore a crash), but only warn when rows have been added? That would be less disruptive for users, at least as a temporary solution.

Regarding the alternative patterns, is this approach OK?
```julia
df = DataFrame(x = [1, 2, 3], y = [1, 2, 2])
sdf = view(df, 1:nrow(df), :)
for g in groupby(sdf, :y)
    nrow(g) == 1 && append!(df, DataFrame(x=[99], y=first(g.y)))
end
```

oxinabox:
>  but only warn when rows have been added? 

Warnings would be very disruptive.
Because if processing 100 files with this pattern you get spammed 100 times.
Sure I could write a LoggingExtras filter to mute that but easier maybe to rewrite to avoid this pattern.
So it might as well be an error.

bkamins:
> sdf = view(df, 1:nrow(df), :)
for g in groupby(sdf, :y)

Yes this approach is OK and efficient. Even `groupby(view(df, :, :), :y)` is OK and it is easier to write, as `:` does not get any special treatment for row selection (as opposed to column selection). The reason is that now we have:
```
julia> df = DataFrame(a=[1,2,3]);

julia> sdf = view(df, :, :);

julia> getfield(sdf, :rows)
Base.OneTo(3)
```
so as you can see such `sdf` almost does not allocate and will be very fast to index (even it is possible that compiler might optimize out the `Base.OneTo` part - though I have not checked it, but potentially it should do it).

Given the proposal by @nalimilan I would keep the error and recommend the `view` approach. The good thing is that as we are throwing an error now it will be easy to catch and fix the undefined behavior. Just replace any `df` in `groupby(df, cols)` with `view(df, :, :)`.

Let us then wait a few days - if there will be no more comments on this approach I will rewrite this PR so that we keep the `@assert` part and only improve the `GroupedDataFrame` docstring.

bkamins:
Ah - and I would like to thank everyone for reporting the issue and the discussion. I think the solution we have reached for this use case is quite good and have a better understanding of the design intentions and available options in general.

bkamins:
@raphaelsaavedra - are descriptions I have added to the documentation OK?

bkamins:
@mortenpi - could you please kindly have a look at the doctest error? I do not understand why `AssertionError` causes an error in doctests (I would expect it should be handled as described in https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Exceptions). Thank you!

raphaelsaavedra:
The additions to the documentation are great and super clear, thanks a lot! I have already fixed our internal code by using the suggested method.

docs/src/man/split_apply_combine.md
```diff
@@ -526,3 +526,95 @@ julia> combine(gd, valuecols(gd) .=> mean)
    3 │ Iris-virginica              6.588            2.974             5.552
                                                                 1 column omitted
 ```
+
+Note that `GroupedDataFrame` is a view and therefore grouping columns of its
+parent data frame should not be mutated. In particular, if number or rows
+of the parent changes then an error is thrown when `GroupedDataFrame` is used:
```
nalimilan, line 532:
```suggestion
Note that `GroupedDataFrame` is a view: therefore 
grouping columns of its parent data frame must not be mutated, and
rows must not be added nor removed from it. If the number or rows
of the parent changes then an error is thrown when a child `GroupedDataFrame`
is used:
```

docs/src/man/split_apply_combine.md
```diff
@@ -526,3 +526,95 @@ julia> combine(gd, valuecols(gd) .=> mean)
...
+   1 │     1
+   2 │     2
+   3 │     3
+
+julia> gd
+Error showing value of type GroupedDataFrame{DataFrame}:
+ERROR: AssertionError: The current number of rows in the parent data frame is 3 and it does not match the number of rows it contained when GroupedDataFrame was created which was 2. The number of rows in the parent data frame has likely been changed unintentionally (e.g. using subset!, filter!, delete!, push!, or append! functions).
+```
+
+Sometimes it is useful to append the rows to the source data frame of
```
nalimilan, line 570:
```suggestion
Sometimes it is useful to append rows to the source data frame of a
```

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -18,7 +18,20 @@
...
 
-Not meant to be constructed directly, see `groupby`.
+Not meant to be constructed directly, see [`groupby`](@ref).
+
+One can get the names of columns used to create `GroupedDataFrame`
+using the [`groupcols`](@ref) function. Similarly the [`groupindices`](@ref)
+function returns a vector of group indices for each row of the parent data frame.
+
+After creation of `GroupedDataFrame` it reflects the grouping of rows that was
+valid at the creation time of `GroupedDataFrame`. Therefore making changes to
```
nalimilan, line 28:
```suggestion
After its creation, a `GroupedDataFrame` reflects the grouping of rows that was
valid at the its creation time. Therefore making changes to
```

mortenpi:
> @mortenpi - could you please kindly have a look at the doctest error? I do not understand why `AssertionError` causes an error in doctests (I would expect it should be handled as described in https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Exceptions). Thank you!

Hmm, I think it's because Documenter only handles errors thrown when executing the code, but not when `show`ing the resulting object (the actual error happens here https://github.com/JuliaDocs/Documenter.jl/blob/v0.27.3/src/DocTests.jl#L319). https://github.com/JuliaDocs/Documenter.jl/issues/1631. But for the time being, I would suggest turning that code snippet into a non-executing `julia-repl` block.

bkamins:
Thank you. I will then force an error by some other operation :)

src/groupeddataframe/groupeddataframe.jl
```diff
@@ -18,7 +18,20 @@
...
-Not meant to be constructed directly, see `groupby`.
+Not meant to be constructed directly, see [`groupby`](@ref).
+
+One can get the names of columns used to create `GroupedDataFrame`
+using the [`groupcols`](@ref) function. Similarly the [`groupindices`](@ref)
+function returns a vector of group indices for each row of the parent data frame.
+
+After its creation, a `GroupedDataFrame` reflects the grouping of rows that was
+valid at the its creation time. Therefore making changes to
+grouping columns of the parent data frame can lead to hard to catch bugs.
```
nalimilan, line 29:
```suggestion
valid at the its creation time. Therefore
grouping columns of its parent data frame must not be mutated, and
rows must not be added nor removed from it.
```

nalimilan (approved):

bkamins:
Thank you!

Issue #2762, opened 2021-05-15
add a keyword to specify group order in groupby


bkamins:
Currently we essentially do not guarantee the group order (unless `sort=true` is passed). One can of course sort them afterwards, but I think we should provide an option to control this more.

PR #2812, opened 2021-07-07
make sort kwarg in groupby more flexible

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2762.

If we accept it then:
* by default we use the fastest algorithm
* if user passes `sort=false` explicitly then groups are created in order of appearance
* if user passes `sort=true` then groups are sorted

If we agree to add this extension I will add tests and NEWS.md entry and finalize this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,12 +8,15 @@
   - `true` sorts groups by key columns;
   - `false` creates groups in the order of their appearance in the parent data
     frame;
-  The old behavior was that the `sort` keyword argument allowed only `Bool` values
-  and if `false` was passed (which was the default) it corresponded to current
-  behavior when `nothing` is passed. Therefore only the user visible change is
-  when `sort=false` is passed explicitly as now instead of the order being
-  undefined it is guaranteed that the groups are in order of their appearance
-  in the parent data frame.
+  In previous versions, the `sort` keyword argument allowed only `Bool` values
+  and `false` (which was the default) corresponded to the new
+  behavior when `nothing` is passed. Therefore only the user visible change
+  affecting existing code is when `sort=false` is passed explicitly.
+  The order of groups was undefined in that case, but in practice
+  groups were already created in their order of appearance, *except*
+  when grouping columns implemented the `DataAPI.refpool` API
+  (notably `PooledArray` and `CategoricalArray`) or when they contained only
+  integers in a small range.
   ([#2812](https://github.com/JuliaData/DataFrames.jl/pull/2812))
 
 # DataFrames.jl v1.2 Release Notes
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,21 @@
+# DataFrames.jl v1.3 Release Notes
+
+## New functionalities
+
+* in the `groupby` function the `sort` keyword argument now allows three values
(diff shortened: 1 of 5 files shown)


bkamins:
@nalimilan - this PR should be good for review.

@pdeffebach - you might want to have a look at it in the context of DataFramesMeta.jl. Thank you!

NEWS.md
```diff
@@ -1,3 +1,21 @@
...
+    `groupby` to pick the fastest available grouping algorithm;
+  - `true` sorts groups by key columns;
+  - `false` creates groups in the order of their appearance in the parent data
+    frame;
+  The old behavior was that the `sort` keyword argument allowed only `Bool` values
+  and if `false` was passed (which was the default) it corresponded to current
+  behavior when `nothing` is passed. Therefore only the user visible change is
+  when `sort=false` is passed explicitly as now instead of the order being
+  undefined it is guaranteed that the groups are in order of their appearance
+  in the parent data frame.
```
nalimilan, line 16:
```suggestion
  In previous versions, the `sort` keyword argument allowed only `Bool` values
  and `false` (which was the default) corresponded to the new
  behavior when `nothing` is passed. Therefore only the user visible change
  affecting existing code is when `sort=false` is passed explicitly.
  The order of groups was undefined in that case, but in practice
  groups were already created in their order of appearance, *except*
  when grouping columns implemented the `DataAPI.refpool` API
  (notably `PooledArray` and `CategoricalArray`) or when they contained only
  integers in a small range.
```

test/grouping.jl
```diff
@@ -3844,9 +3844,107 @@ end
...
+
+    df.id = PooledArray(df.id)
+    gd = groupby_checked(df, :id, sort=nothing)
+    @test Vector.(keys(gd)) == [["3"], ["1"], ["2"]]
+    gd = groupby_checked(df, :id, sort=true)
+    @test Vector.(keys(gd)) == [["1"], ["2"], ["3"]]
+    gd = groupby_checked(df, :id, sort=false)
+    @test Vector.(keys(gd)) == [["3"], ["1"], ["2"]]
+
+    # for PooledVector sorted=true is not the same as sorted=nothing
```
nalimilan, line 3885:
```suggestion
    # for PooledVector sort=true is not the same as sort=nothing
```

test/grouping.jl
```diff
@@ -3844,9 +3844,107 @@ end
...
+    df.id[1] = "300"
+    df.id[3] = "200"
+    gd = groupby_checked(df, :id, sort=nothing)
+    @test Vector.(keys(gd)) == [["1"], ["300"], ["200"]]
+    gd = groupby_checked(df, :id, sort=true)
+    @test Vector.(keys(gd)) == [["1"], ["200"], ["300"]]
+    gd = groupby_checked(df, :id, sort=false)
+    @test Vector.(keys(gd)) == [["300"], ["1"], ["200"]]
+
+    # for CategoricalVector sorted=true is the same as sorted=nothing
```
nalimilan, line 3895:
```suggestion
    # for CategoricalVector sort=true is the same as sort=nothing
```

nalimilan (approved):

bkamins:
Thank you!

PR #2818, opened 2021-07-19
Add makeunique to Tables.jl DataFrame constructor

bkamins:
This is WIP as it requires changes to `Tables.columns` (so that it can accept `makeunique` kwarg).

See https://discourse.julialang.org/t/sqlite-query-with-duplicate-column-names-dataframe/64518 for a related discussion.

@quinnj - do you think it is feasible to change `Tables.columns` implementation? In particular the challenge is that we are post 1.0 in Tables.jl and some custom packages might already have implemented `Tables.columns` without this kwarg.
Or maybe there is some other better solution? You know Tables.jl design best.

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -45,17 +45,18 @@ fromcolumns(x, names; copycols::Union{Nothing, Bool}=nothing) =
 fromcolumns(x::Tables.CopiedColumns, names; copycols::Union{Nothing, Bool}=nothing) =
     fromcolumns(Tables.source(x), names; copycols=something(copycols, false))
 
-function DataFrame(x::T; copycols::Union{Nothing, Bool}=nothing) where {T}
+function DataFrame(x::T; copycols::Union{Nothing, Bool}=nothing,
+                   makeunique::Bool=false) where {T}
     if !Tables.istable(x) && x isa AbstractVector && !isempty(x)
         # here we handle eltypes not specific enough to be dispatched
         # to other DataFrames constructors taking vector of `Pair`s
         if all(v -> v isa Pair{Symbol, <:AbstractVector}, x) ||
             all(v -> v isa Pair{<:AbstractString, <:AbstractVector}, x)
             return DataFrame(AbstractVector[last(v) for v in x], [first(v) for v in x],
-                             copycols=something(copycols, true))
+                             copycols=something(copycols, true), makeunique=makeunique)
         end
     end
-    cols = Tables.columns(x)
+    cols = Tables.columns(x, makeunique=true)
     names = collect(Symbol, Tables.columnnames(cols))
     return fromcolumns(cols, names, copycols=copycols)
 end
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -56,7 +56,7 @@ function DataFrame(x::T; copycols::Union{Nothing, Bool}=nothing,
                              copycols=something(copycols, true), makeunique=makeunique)
         end
     end
-    cols = Tables.columns(x, makeunique=true)
+    cols = makeunique ? Tables.columns(x, makeunique=true) : Tables.columns(x)
     names = collect(Symbol, Tables.columnnames(cols))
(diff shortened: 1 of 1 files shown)


viraltux:
As an example, when using the SQLite.jl implemenation of Tables.jl the following query with duplicate column names works however when converting to a DataFrame we have a "not unique" error.

```
using SQLite
db = SQLite.DB();
Q = q -> DBInterface.execute(db, q);

DataFrame(rand(10,3),:auto) |> SQLite.load!(db, "a", ifnotexists=false);
DataFrame(rand(10,2),:auto) |> SQLite.load!(db, "b", ifnotexists=false);

Q("select * from a join b on a.x1 > b.x2")  |> DataFrame

ERROR: duplicate field name in NamedTuple: "x1" is not unique
```
This forces us to name each duplicate in the query which they can be potentially in the hundreds.

Ideally something like ```DataFrame(Q(…); makeunique=true)``` would handle it.

bkamins:
@quinnj - do you have any opinion how we should proceed? Thank you!

CC @nalimilan

nalimilan:
> In particular the challenge is that we are post 1.0 in Tables.jl and some custom packages might already have implemented `Tables.columns` without this kwarg.

That's not an issue for Tables.jl IMO: currently, passing `makeunique=true/false` to `Tables.columns` is an error. So adding that argument won't break anything. It's true that some types may have implemented `Tables.columns` without that argument, and they will error, but getting an error for some types is strictly better than getting an error for all of them. They can be improved later.

What would be problematic is if the `DataFrame` construct started to throw an error by default when the particular `Tables.columns` method doesn't support `makeunique`. To avoid this we can do `makeunique ? Tables.columns(x) : Tables.columns(x, makeunique=true)`. That's a bit weird but OK I think.

src/other/tables.jl
```diff
@@ -45,17 +45,18 @@ fromcolumns(x, names; copycols::Union{Nothing, Bool}=nothing) =
...
         # to other DataFrames constructors taking vector of `Pair`s
         if all(v -> v isa Pair{Symbol, <:AbstractVector}, x) ||
             all(v -> v isa Pair{<:AbstractString, <:AbstractVector}, x)
             return DataFrame(AbstractVector[last(v) for v in x], [first(v) for v in x],
-                             copycols=something(copycols, true))
+                             copycols=something(copycols, true), makeunique=makeunique)
         end
     end
-    cols = Tables.columns(x)
+    cols = Tables.columns(x, makeunique=true)
```
bkamins, line 59:
```suggestion
    cols = makeunique ? Tables.columns(x, makeunique=true) : Tables.columns(x)
```

bkamins:
> To avoid this we can do `makeunique ? Tables.columns(x) : Tables.columns(x, makeunique=true)`. That's a bit weird but OK I think.

This is a sensible workaround (I have implemented it - but the condition has to be reversed). The problem here is that one might pass `makeunique=false` in case when `Tables.columns` would support `makeunique` and have `makeunique=true` as a default. I think we should agree in Tables.jl that `makeunique=false` must the default and then all will be OK.

Thank you!

bkamins:
@quinnj - bump for opinion

quinnj:
Sorry for such a slow response here; I've been swamped w/ things lately.

Having read through the discourse post and the comments here, I don't like the idea of adding `makeunique` to `Tables.columns`. `Tables.columns` tries to be a very light, non-fussy way to just get columns from a source, and it adds a lot of complexity to introduce an optional keyword arg like this. Most sources won't end up implementing it; many callers won't use it, the lack of consistency in supporting it will only cause problems when people try to use it and it doesn't work.

I think it's more sensible to require that `Tables.columnnames(source)` return a _unique_ column name list. I think the ultimate fix should be in SQLite.jl here and I can try to get that fixed quickly.

bkamins:
OK, so I am closing this PR.

I have opened https://github.com/JuliaData/Tables.jl/pull/248 and adding a x-link to https://github.com/JuliaDatabases/SQLite.jl/pull/254.

Issue #2795, opened 2021-06-22
DataFrames not threadsafe 


smjaberl:
Hi,

I have a case, where I use a push to a DataFrame on multiprocessing. When the rare case occurs, that two processes are pushing on the same time, I get this error:

┌ Error: Error adding value to column :t.
└ @ DataFrames ~/.julia/packages/DataFrames/nxjiD/src/dataframe/dataframe.jl:1644

Is this a known bug? I found no documentation about it. 

regards
jan

bkamins:
This is not a bug, but a design decision in Julia Base, which is just propagated to DataFrames.jl. Operations like `push!` (I assume you are doing this) are not thread safe in Julia in general (not only in DataFrames.jl).

I will keep this issue open, as maybe in the future the design assumptions will change.

smjaberl:
Well, thanks for adding the feature label to the issue.  I think it make sense to think about this design decision. A modern fast programming language like julia should be threadsafe in such elementary functions. On the other hand I know, it would be more complex.

bkamins:
I agree, but this is a request to Julia Base mainly. The issue is that adding thread safety to such elementary operations would make them slower.

See a comment by @StefanKarpinski here https://discourse.julialang.org/t/can-dicts-be-threadsafe/27172/6.

StefanKarpinski:
> A modern fast programming language like julia should be threadsafe in such elementary functions. On the other hand I know, it would be more complex.

It's not just a matter of complexity. For many data structures and operations, the threadsafe version is significantly slower that the threadunsafe version. So it is precisely because Julia is a fast programming language that it cannot simply make everything threadsafe. The core data structures and operations provided by the language are fast and threadunsafe (unless thread safety is necessary or can be provided without loss of performance) and the tools (like atomics) are provided for packages to build thread safe algorithms and data structures on top of that.

nalimilan:
I agree that we can't ensure Julia is threadsafe everywhere without killing performance. However, given that appending new rows to a data frame is quite slower than appending entries to a single vector (due to type instability and column lookup), it might make sense to ensure thread safety if we can confirm the overhead is negligible.

bkamins:
We could ensure thread safety by:
* having an internal lock (like in `GroupedDataFrame`)
* using it in such operations

However, it would not have any performance benefit (i.e. the speed of threaded `push!` would be worse than speed of sequential `push!`) - the only benefit would be thread safety at the cost of non deterministic row order.

[...]

PR #2823, opened 2021-07-24
Add muli-threading support description to the manual

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2795

### DIFF
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -17,11 +17,11 @@ This is a list of operations that currently make use of multi-threading:
   is used).
 - `join*` functions for composing output data frame (but currently not for finding
   matching rows in joined data frames).
-- `combine`, `select[!]`, and `transform[!]` on `GroupedDataFrame` when performing
-  multiple transformations (each transformation is spawned in a separate task)
-- `combine` when a transformation produces one row per group and the passed transformation
-  is a custom function (i.e. not for standard reductions, which use
-  optimized single-threaded methods).
+- `combine`, `select[!]`, and `transform[!]` on `GroupedDataFrame` when either of the conditions below is met:
+  * multiple transformations are performed (each transformation is spawned in a separate task)
+  * a transformation produces one row per group and the passed transformation
+    is a custom function (i.e. not for standard reductions, which use
+    optimized single-threaded methods).
 
 In general at least Julia 1.4 is required to ensure that multi-threading is used
 and the Julia process must be started with more than one thread. Some operations
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -4,6 +4,31 @@ CurrentModule = DataFrames
 
 # Functions
 
+## Multi-threading support
+
+Selected operations in DataFrames.jl support muliti-threading.
+It is task-based and implemented using the `@spawn` macro form Julia Base.
+
+This is a list of operations that currently support multi-threading:
+- `DataFrame` constructor with `copycols=true`; also recursively all functions
+  that call this constructor, e.g. `copy`.
+- `getindex` when multiple columns are selected.
+- `groupby` (both when hashing is required and when fast path using `DataAPI.refpool`
+  is used).
+- `join*` functions for composing output data frame (but currently not for finding
+  matching rows in joined data frames).
+- in `combine`, `select[!]``, and `transform[!]` on `GroupedDataFrame` when performing
+  multiple transformations (each transformation is spawned in a separate task)
+- in `combine` when a transformation produces one row per group and the passed tranformation
+  is a custom function (i.e. when reduction based aggregation is not performed).
+
+In general at least Julia 1.4 is required to ensure that multi-threading is used
+and the Julia process must be started with more than one thread. Some operations
+turn on multi-threading only if enough rows in the processed data frame is present
+(the exact threshold when multi-threading is enabled is considered to be undefined
+and might change in the future).
+
+## Index
 ```@index
 Pages = ["functions.md"]
 ```
--- a/docs/src/lib/functions.md
+++ b/docs/src/lib/functions.md
@@ -6,10 +6,10 @@ CurrentModule = DataFrames
 
 ## Multi-threading support
 
-Selected operations in DataFrames.jl support muliti-threading.
-It is task-based and implemented using the `@spawn` macro form Julia Base.
+Selected operations in DataFrames.jl automatically use multiple threads when available.
+It is task-based and implemented using the `@spawn` macro from Julia Base.
 
-This is a list of operations that currently support multi-threading:
+This is a list of operations that currently make use of multi-threading:
 - `DataFrame` constructor with `copycols=true`; also recursively all functions
   that call this constructor, e.g. `copy`.
 - `getindex` when multiple columns are selected.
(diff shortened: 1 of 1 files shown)


docs/src/lib/functions.md
```diff
@@ -4,6 +4,31 @@ CurrentModule = DataFrames
...
+  multiple transformations (each transformation is spawned in a separate task)
+- in `combine` when a transformation produces one row per group and the passed tranformation
+  is a custom function (i.e. when reduction based aggregation is not performed).
+
+In general at least Julia 1.4 is required to ensure that multi-threading is used
+and the Julia process must be started with more than one thread. Some operations
+turn on multi-threading only if enough rows in the processed data frame is present
+(the exact threshold when multi-threading is enabled is considered to be undefined
+and might change in the future).
+
```
bkamins, line 30:
```suggestion

Except for the list above, where multi-threading is used automatically, all functions provided by DataFrames.jl that update a data frame are not thread safe.
```

docs/src/lib/functions.md
```diff
@@ -4,6 +4,32 @@ CurrentModule = DataFrames
...
+Except for the list above, where multi-threading is used automatically, all functions provided by DataFrames.jl that update a data frame are not thread safe.
```
bkamins, line 31:
```suggestion
Except for the list above, where multi-threading is used automatically,
all functions provided by DataFrames.jl that update a data frame are not thread safe.

```

docs/src/lib/functions.md
```diff
@@ -4,6 +4,34 @@ CurrentModule = DataFrames
...
+  that call this constructor, e.g. `copy`.
+- `getindex` when multiple columns are selected.
+- `groupby` (both when hashing is required and when fast path using `DataAPI.refpool`
+  is used).
+- `join*` functions for composing output data frame (but currently not for finding
+  matching rows in joined data frames).
+- in `combine`, `select[!]``, and `transform[!]` on `GroupedDataFrame` when performing
+  multiple transformations (each transformation is spawned in a separate task)
+- in `combine` when a transformation produces one row per group and the passed tranformation
+  is a custom function (i.e. when reduction based aggregation is not performed).
```
nalimilan, line 23:
```suggestion
  is a custom function (i.e. not for standard reductions, which use
  optimized single-threaded methods).
```

docs/src/lib/functions.md
```diff
@@ -4,6 +4,34 @@ CurrentModule = DataFrames
...
+
+In general at least Julia 1.4 is required to ensure that multi-threading is used
+and the Julia process must be started with more than one thread. Some operations
+turn on multi-threading only if enough rows in the processed data frame is present
```
nalimilan, line 27:
```suggestion
turn on multi-threading only if enough rows in the processed data frame are present
```

docs/src/lib/functions.md
```diff
@@ -4,6 +4,34 @@ CurrentModule = DataFrames
...
+(the exact threshold when multi-threading is enabled is considered to be undefined
+and might change in the future).
+
+Except for the list above, where multi-threading is used automatically,
+all functions provided by DataFrames.jl that update a data frame are not thread safe.
```
nalimilan, line 32:
```suggestion
all functions provided by DataFrames.jl that update a data frame are not thread safe.
This means that while they can be called from any thread, the caller is responsible
for ensuring that a given `DataFrame` object is never modified by one thread while
others are using it (either for reading or writing). Using the same `DataFrame`
at the same time from different threads is safe as long as it is not modified.
```

docs/src/lib/functions.md
```diff
@@ -4,6 +4,39 @@ CurrentModule = DataFrames
...
+
+This is a list of operations that currently make use of multi-threading:
+- `DataFrame` constructor with `copycols=true`; also recursively all functions
+  that call this constructor, e.g. `copy`.
+- `getindex` when multiple columns are selected.
+- `groupby` (both when hashing is required and when fast path using `DataAPI.refpool`
+  is used).
+- `join*` functions for composing output data frame (but currently not for finding
+  matching rows in joined data frames).
+- `combine`, `select[!]``, and `transform[!]` on `GroupedDataFrame` when performing
```
bkamins, line 20:
```suggestion
- `combine`, `select[!]`, and `transform[!]` on `GroupedDataFrame` when performing
```

bkamins:
Thank you - very good corrections.

docs/src/lib/functions.md
```diff
@@ -4,6 +4,34 @@ CurrentModule = DataFrames
...
+- `DataFrame` constructor with `copycols=true`; also recursively all functions
+  that call this constructor, e.g. `copy`.
+- `getindex` when multiple columns are selected.
+- `groupby` (both when hashing is required and when fast path using `DataAPI.refpool`
+  is used).
+- `join*` functions for composing output data frame (but currently not for finding
+  matching rows in joined data frames).
+- in `combine`, `select[!]``, and `transform[!]` on `GroupedDataFrame` when performing
+  multiple transformations (each transformation is spawned in a separate task)
+- in `combine` when a transformation produces one row per group and the passed tranformation
```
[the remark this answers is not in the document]
nalimilan: Any comment on this?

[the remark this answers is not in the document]
bkamins: Ah - I missed the question. I was thinking if we should include this. This can happen only if `select` or `transform` are passed a `GroupedDataFrame` that has exactly one row per group and when aggregation also produces one row per group. This is an unlikely scenario (in this scenario it does make sense to do `groupby` and then `select`/`transform` but rather one should just run `select`/`transform` on a data frame with `ByRow`). Do you think we should add `select` and `transform` to the list (as technically they could, but the condition when they would is unlikely).

In essence since this is "one row per group" case it means that we are `combine`ing rows unless each group has exactly one row to start with 😄.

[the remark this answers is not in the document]
bkamins: @nalimilan - what do you think given this comment?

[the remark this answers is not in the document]
nalimilan: Sorry, I had missed your reply. Isn't multithreading also used with `select` and `transform` when it returns a single row for each group (disregarding the number of rows in the group)? We repeat the returned row to match the number of rows in the input, but the step before that is multithreaded IIRC, right?

docs/src/lib/functions.md
```diff
@@ -4,6 +4,39 @@ CurrentModule = DataFrames
...
+- `getindex` when multiple columns are selected.
+- `groupby` (both when hashing is required and when fast path using `DataAPI.refpool`
+  is used).
+- `join*` functions for composing output data frame (but currently not for finding
+  matching rows in joined data frames).
+- `combine`, `select[!]`, and `transform[!]` on `GroupedDataFrame` when performing
+  multiple transformations (each transformation is spawned in a separate task)
+- `combine` when a transformation produces one row per group and the passed transformation
+  is a custom function (i.e. not for standard reductions, which use
+  optimized single-threaded methods).
```
bkamins, line 24:
```suggestion
- `combine`, `select[!]`, and `transform[!]` on `GroupedDataFrame` when either of the conditions below is met:
  * performing multiple transformations (each transformation is spawned in a separate task)
  * a transformation produces one row per group and the passed transformation
    is a custom function (i.e. not for standard reductions, which use
    optimized single-threaded methods).
```

docs/src/lib/functions.md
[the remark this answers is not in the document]
bkamins: You are right (as usual - I am glad to have you :smile:). I have fixed this.

nalimilan (approved):

bkamins:
Thank you!

PR #2826, opened 2021-07-31
correct signature of merge for AbstractIndex

bkamins:
`merge` implementation was super old and it turns out we even did not have proper tests of it. I have added them now. `merge` is used only in `crossjoin` now, and the problem appeared when one tries to do a `crossjoin` of two `SubDataFrame`s with `makeunique=true`. This is also added to the tests

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,14 @@
   integers in a small range.
   ([#2812](https://github.com/JuliaData/DataFrames.jl/pull/2812))
 
+# DataFrames.jl v1.2.2 Patch Release Notes
+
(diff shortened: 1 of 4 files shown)


bkamins:
Currently I have prepared this PR so that I can easily backport it and make a 1.2.2 patch release. However, equally well we can wait with it till 1.3 release as the bug will only cause inconvenience (something that should work currently errors) and the scenario when this happens is super unlikely. Please comment what you prefer.

pdeffebach (approved):

bkamins:
Thank you!

PR #2828, opened 2021-08-01
feat: `unstack` receives kwarg `fillvalue`

### LINKED ISSUE #2698: control fill value for missing cells in `unstack`
Currently, when `unstack` encounters combinations of column and row keys that are not present in the DataFrame, the corresponding cell in the output is filled with `missing`.  It would be nice to be able to control the fill value used with a kwarg.  

For context, I hit this using `unstack` as a convenient way to make a confusion matrix (for which it's especially nice since you can add additional grouping variables on the rows), but cells that are missing really correspond to `0` in that situation.

pstorozenko:
Closes #2698 

I thought about checking every column separately for `promote_type`, but it would be a breaking change since right now value column with eltype `T` is always converted to `Vector{Union{Missing, T}}`. We may think about changing it in another PR.


bkamins:
>  but it would be a breaking change

Why do you think so? I would assume that if `fillvalue` is `missing` which is the default and old behavior in your PR we would get the same (so this means that no old code would be broken - the result would stay the same).

NEWS.md
```diff
@@ -18,6 +18,11 @@
   (notably `PooledArray` and `CategoricalArray`) or when they contained only
   integers in a small range.
   ([#2812](https://github.com/JuliaData/DataFrames.jl/pull/2812))
+* the `unstack` function receives new keyword argument `fillvalue`
+  (with `missing` default) that fills combination of not encountered
```
bkamins, line 22:
```suggestion
  (with `missing` default) that is used to fill combinations of not encountered
```

NEWS.md
```diff
@@ -18,6 +18,11 @@
...
+  rows and columns. This feature allows to distinguish between missings in
+  value column and just missing row/column combinations.
```
bkamins, line 24: I would say that the main use case is `fillvalue=0` if you have count data.

test/reshape.jl
```diff
@@ -654,4 +654,27 @@ end
...
+    dfu = unstack(df, :variable, :value)
+    @test dfu ≅ dfu1
+    @test eltype(dfu[!, :Var1]) == Union{Missing, Int}
+
+    delete!(df, 4)
+    for (sentinel, fcoltype) in zip([1, 1., "1", nothing], [Int, Float64, Any, Union{Int, Nothing}])
+        dfu_m = unstack(df, :variable, :value, fillvalue=sentinel)
+        @test dfu_m[2, :Var2] == sentinel
+        @test eltype(dfu_m[!, :Var2]) == fcoltype
+    end
```
bkamins, line 677: It is very important here to add tests for `CategoricalVector`, `ChainedVector` and maybe also `PooledVector`.
The reason is to check if the line https://github.com/JuliaData/DataFrames.jl/pull/2828/files#diff-9b09ee01805b1e911ca9c715199a11e9c872da2dfb66250a5ae8097cf8e67216R429 will produce what you expect in different combinations of cases. E.g. if `CategoricalVector` will be retained given different types of `fillvalue`.

bkamins:
> I thought about checking every column separately

Ah - after reviewing I see what you mean. What you have implemented is OK I think. One can run `identity.(df)` to do type narrowing later if needed.

However, it is important to test the additional scenarios I have mentioned I think. Especially for `CategoricalVector` check how `levels` behave in different combinations of inputs and in cases when original `CategoricalVector` is ordered/unordered.

Other than that - the PR looks good.

bkamins:
@pstorozenko - merge conflicts require fixing + have you had time to add the tests I have suggested?

bkamins:
@pstorozenko bump

pstorozenko:
Sorry for not answering.
There are a few things I found when trying to create tests for Categorical Vector:

1. When dealing with vectors that are subtypes of `CategoricalArrays.AbstractCategoricalArray` we should create `unstacked_val` with `fillvalue` value and levels from `valuecol` + `fillvalue`.
2. When `fillvalue` is `missing` then it is not added to levels (that's how CategoricalArrays work AFAIK).

bkamins:
Point 2 is correct.

Point 1 - there is a corner case when `fillvalue` is not allowed by `CategoricalArray`. As discussed - I will update the PR to check for corner cases. Thank you!

test/reshape.jl
bkamins: currently what we produce is disputable. I will comment above.

src/abstractdataframe/reshape.jl
```diff
@@ -398,13 +421,15 @@ function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Int},
...
     colref_map = df[col_group_row_idxs, colkey]
-
     if any(ismissing, colref_map) && !allowmissing
         throw(ArgumentError("Missing value in variable :$(_names(df)[colkey]). " *
                             "Pass `allowmissing=true` to skip missings."))
     end
+    unstacked_val = [fill!(similar(valuecol,
+                                   promote_type(eltype(valuecol), typeof(fillvalue)),
+                                   Nrow),
+                           fillvalue) for _ in 1:Ncol]
```
bkamins, line 431: @nalimilan - this definition is problematic for categorical columns the problem is that:
```
julia> using CategoricalArrays

julia> valuecol = categorical(["a"])
1-element CategoricalArray{String,1,UInt32}:
 "a"

julia> fillvalue = ""
""

julia> Nrow = 1
1

julia> similar(valuecol, promote_type(eltype(valuecol), typeof(fillvalue)), Nrow)
1-element Vector{String}:
 #undef

julia> promote_type(eltype(valuecol), typeof(fillvalue))
String
```

Is there a generic idiom that would (without introducing dependency on CategoricalArrays.jl) allow us to take the union of types and if it is OK (like it should be in the example above) produce `CategoricalArray`? Or maybe you think we should change the definition of `similar` in CategoricalArrays.jl?

bkamins: OK - I have thought about it. Given we currently allow:
```
julia> x = categorical([1,2,3])
3-element CategoricalArray{Int64,1,UInt32}:
 1
 2
 3

julia> CategoricalValue(3, x)
CategoricalValue{Int64, UInt32} 3
```
the behavior we have now should be OK.

test/reshape.jl
```diff
@@ -654,6 +654,74 @@ end
...
+    df = DataFrame(factory=["Fac1", "Fac1", "Fac2"],
+                   variable=["Var1", "Var2", "Var1"],
+                   value=categorical([1, 2, 3]))
+    dfu = unstack(df, :variable, :value, fillvalue=CategoricalValue(0, categorical([0])))
+    @test dfu.Var1 == [1, 3]
+    @test typeof(dfu.Var1) <: CategoricalVector{Int}
+    @test dfu.Var2 ≅ [2, 0]
+    @test typeof(dfu.Var2) <: CategoricalVector{Int}
+    @test levels(dfu.Var1) == levels(dfu.Var2) == 0:3
+    dfu = unstack(df, :variable, :value, fillvalue=CategoricalValue("0", categorical(["0"])))
```
bkamins, line 717: @nalimilan - this test in particular is something we should think of. My judgement is that it is OK to produce such an union.

bkamins (approved):

src/abstractdataframe/reshape.jl
```diff
@@ -331,36 +335,55 @@ julia> unstack(long, :id, :variable, :value, renamecols=x->Symbol(:_, x))
    4 │     4       2.0       1.0       2.0
    5 │     5       2.0       1.0       3.0
    6 │     6       2.0       1.0       3.0
+
+julia> df = DataFrame(id= ["1", "1", "2"],
```
nalimilan, line 339:
```suggestion
julia> df = DataFrame(id=["1", "1", "2"],
```

src/abstractdataframe/reshape.jl
nalimilan: I think it's OK. If we wanted to change this we could adapt `promote_type` in CategoricalArrays, but it's orthogonal to this PR. There would certainly be advantages in promoting to `CategoricalValue`, though it could also create problems (e.g. when concatenating `Vector{Int} and `CategoricalArray{Union{String, Missing}}` the resulting type would be `Any` but CategoricalArrays doesn't handle that well).

NEWS.md
```diff
@@ -22,6 +22,12 @@
   (notably `PooledArray` and `CategoricalArray`) or when they contained only
   integers in a small range.
   ([#2812](https://github.com/JuliaData/DataFrames.jl/pull/2812))
+* the `unstack` function receives new keyword argument `fillvalue`
```
nalimilan, line 25: Maybe we should call this just `fill`? Or is that name useful in other contexts as a Boolean argument?

src/abstractdataframe/reshape.jl
bkamins: Yes - I think there would be more problems than benefits. We only need an easy way to create a "stand alone" `CategoricalValue`, as now it is `CategoricalValue(scalar, categorical([scalar]))` which is quite inconvenient.

NEWS.md
bkamins: No - this name is specific only to this function. I will rename it to `fill`.

bkamins:
OK - I have renamed `fillvalue` to `fill`. The only potential problem would be if we used the `fill` function from Julia Base in the body of the function, but we do not use it. We only use `fill!` so all is OK.

bkamins:
@nalimilan - any additional comments here?

src/abstractdataframe/reshape.jl
```diff
@@ -229,6 +229,10 @@ Row and column keys will be ordered in the order of their first appearance.
 - `allowduplicates`: if `false` (the default) then an error an error will be thrown
   if combination of `rowkeys` and `colkey` contains duplicate entries; if `true`
   then  then the last encountered `value` will be retained.
+- `fill`: missing row/column combinations are filled with this value; The default
+  is `missing`. If the `value` column is a `CategoricalVector` and `fill`
```
nalimilan, line 233:
```suggestion
- `fill`: missing row/column combinations are filled with this value. The default
  is `missing`. If the `value` column is a `CategoricalVector` and `fill`
```

nalimilan (approved):

bkamins:
Thank you!

PR #2837, opened 2021-08-10
more careful test of ByRow for PooledArray

### LINKED ISSUE #2834: Should ByRow use map or not
Following https://discourse.julialang.org/t/dataframes-byrow-fails-in-transform-with-pooledarrays-after-csv-read/65957/3.

Maybe we should use either broadcasting or comprehension instead of `map` - unfortunately it will have performance consequences.

CC @nalimilan @pdeffebach

bkamins:
Fixes #2834

I think the fast path for `map` should be asked for explicitly. Otherwise users will not understand the following result:
```
julia> let
           id = 0
           df = DataFrame(a=PooledArray([1, 1, 1]))
           function f(x)
               id += 1
               return id
           end
           select(df, :a => ByRow(f) => :a)
       end
3×1 DataFrame
 Row │ a
     │ Int64
─────┼───────
   1 │     2
   2 │     2
   3 │     2
```


pdeffebach:
In the future, in order to support easier out-of-memory operations, we should maybe give this to DataAPI and see if Dagger-esque arrays want to overload the relevant functions.

bkamins:
Yes - I think that out-of-core functionality in DataFrames.jl should be one of the priorities to investigate after 1.3 release (where I want to finish polishing the API).

Hopefully MIT JuliaLab can join the design here - I have had some preliminary discussions with @ViralBShah about it.
Also @quinnj has this issue on the radar.

(I am mentioning all them as adding out-of-core support for DataFrames.jl was not the original goal of the package and it will be a significant effort to do it correctly and efficiently).

ViralBShah:
@shashi spent quite a bit of time on out-of-core with IndexedTables.jl on JuliaDB. All that code is in the JuliaDB repo for reviewing from a design perspective. It was built on Dagger, which has since made substantial progress thanks to @jpsamaroo.

bkamins:
I have opened a pool https://discourse.julialang.org/t/future-directions-for-dataframes-jl/66247 for this. Can you all please vote? Thank you!

My personal perspective is that we should focus on out-of-core processing. The reason is the following. If the user has some data to process one may hit two bottlenecks:
* data does not fit RAM (and thus DataFrames.jl is hard to use) -> solved by out-of-core processing capabilities
* processing is too slow (and thus DataFrames.jl is inconvenient to use) -> solved by GPU processing capabilities

Given these two I would prefer to concentrate on harder constraint. The fact that computing is too slow is not such a big problem as we currently are already quite fast for normal processing tasks (all operations take roughly "seconds", so even if we can be faster this will not be that noticeable, especially as GPU support will probably significantly increase compilation latency).

nalimilan:
This seems like the wrong place to fix #2834 to me. If we think the current behavior of `map` for `PooledArray` is misleading, we should change it in PooledArrays rather than work around it in DataFrames. Let's continue the discussion at https://github.com/JuliaData/PooledArrays.jl/issues/63.

bkamins:
@nalimilan - this should be good to merge. CI failed because of out of memory on server side that sporadically happens.

test/select.jl
```diff
@@ -1665,4 +1665,14 @@ end
     @test_throws ArgumentError select(gdf, [:c1, :c2] => AsTable)
 end
 
+@testset "fix ByRow on PooledArray" begin
```
nalimilan, line 1668:
```suggestion
@testset "ByRow on PooledArray calls function on each entry" begin
```

nalimilan (approved):

bkamins:
Thank you!

Issue #2833, opened 2021-08-06
implement `Tables.materializer(::Type{<:AbstractDataFrame})`?


kleinschmidt:
The `Tables.materializer` function spec is unclear; the docs suggest that it should accept a *value* and return a function to materialize a generic table to a different value of the same type, which are the methods implemented in DataFrames.  But it would also be useful to be able to get the materializer given the *type*.  That's straightforward for DataFrames:

```julia
Tables.materializer(::Type{<:AbstractDataFrame}) = DataFrame
```

The only uncertainty is whether this is something that should work for all tables (if not, it's not particularly useful to implement here), so this is more of a Tables.jl issue: https://github.com/JuliaData/Tables.jl/issues/246

pdeffebach:
I don't think any Tables.jl stuff works on types, right? Just instances?

kleinschmidt:
```julia
julia> using Tables, DataFrames

julia> Tables.istable(DataFrame)
true
```

pdeffebach:
That's interesting, because the Tables.jl docs don't mention that at all. It doesn't seem to be a requirement for the interface.

bkamins:
@quinnj - can you please recommend a decision? Thank you!

kleinschmidt:
> That's interesting, because the Tables.jl docs don't mention that at all. It doesn't seem to be a requirement for the interface.

It's actually not clear what the `table` in the (ahem) table of interface functions refers to.  For instance, the brief description for `istable(table)` reads "Declare that your table type implements the interface", which suggests that `table` is (or could be) a Type.  In fact, most of the descriptions of those functions refer to "your table type".

krynju:
Current Tables.jl docs make sense (although they're little ambigous on this) and the typical usage of materializer is fine with it only taking a table instance as an arg.

Adjusted example from TableOperations.jl:
```
ctable = (A=[1, missing, 3], B=[1.0, 2.0, 3.0], C=["hey", "there", "sailor"])

table_subset = ctable |> TableOperations.select(:C, :A) |> Tables.materializer(ctable)
```

I raised this here https://github.com/JuliaData/Tables.jl/issues/246 some time ago, because I think this is more of a topic for all Tables compatible sources, since only having DataFrames implement it is not useful overall.

[...]

PR #2839, opened 2021-08-12
add Tables.materializer for types methods

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2833

Implement `Tables.materializer` on types rather than on values, as when passed values it will fall back to types anyway.

### DIFF
--- a/src/other/tables.jl
+++ b/src/other/tables.jl
@@ -22,7 +22,7 @@ Tables.columnindex(df::Union{AbstractDataFrame, DataFrameRow}, idx::AbstractStri
     columnindex(df, Symbol(idx))
 
 Tables.schema(df::AbstractDataFrame) = Tables.Schema(_names(df), [eltype(col) for col in eachcol(df)])
-Tables.materializer(df::AbstractDataFrame) = DataFrame
+Tables.materializer(::Type{<:AbstractDataFrame}) = DataFrame
 
 Tables.getcolumn(df::AbstractDataFrame, i::Int) = df[!, i]
 Tables.getcolumn(df::AbstractDataFrame, nm::Symbol) = df[!, nm]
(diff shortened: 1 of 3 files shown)


src/other/tables.jl
```diff
@@ -82,8 +82,8 @@ Tables.schema(itr::Union{DataFrameRows, DataFrameColumns}) = Tables.schema(paren
 Tables.rowtable(itr::Union{DataFrameRows, DataFrameColumns}) = Tables.rowtable(parent(itr))
 Tables.namedtupleiterator(itr::Union{DataFrameRows, DataFrameColumns}) =
     Tables.namedtupleiterator(parent(itr))
-Tables.materializer(itr::Union{DataFrameRows, DataFrameColumns}) =
-    Tables.materializer(parent(itr))
```
nalimilan, line 86: Any idea why we used this complex definition?

nalimilan (approved):

src/other/tables.jl
bkamins: Because we thought earlier that methods should be defined for instances not for types.
Also the old definition (with `parent`) tried to be more general in case someone would implement a custom `AbstractDataFrame`, but I do not think we need to be that careful. Currently - essentially - DataFrames.jl has the assumption that we only support `DataFrame` and `SubDataFrame` baked in in many places. Also falling back to `DataFrame` always is a relatively safe option.

bkamins: And we have `Tables.materializer(df::AbstractDataFrame) = DataFrame` anyway now, so mostly this does not change anything.

bkamins:
Thank you!

@quinnj - I assume you were OK with this change?

quinnj:
Yes, this looks good by me

Issue #2841, opened 2021-08-22
Error for `unstack`ing an empty dataframe


kollch:
When attempting to run the `unstack` function on an empty dataframe, the error produced references an internal `MethodError`:

`ERROR: no method matching DataFrame(::Vector{Vector{Any}}, ::Vector{Any}; copycols=false)`

How to replicate:
```
df = DataFrame(a = [], b = [], c = [])
unstack(df, :b, :c)
```

I think it'd be nice to have a check for this, especially considering `stack` doesn't appear to have a similar problem.

bkamins:
Thank you for reporting. This is a problem with broadcasting in base. Fixed in https://github.com/JuliaData/DataFrames.jl/pull/2842

PR #2842, opened 2021-08-23
Fix empty unstack on empty data frame

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2841

### DIFF
--- a/src/abstractdataframe/reshape.jl
+++ b/src/abstractdataframe/reshape.jl
@@ -418,10 +418,11 @@ function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Int},
         mask_filled[row_id, col_id] = true
     end
 
-    # note that Symbol.(renamecols.(colref_map)) must produce unique column names
+    # note that Symbol(renamecols(colref_map)) must produce unique column names
     # and names between df1 and df2 must be unique
     df1 = df[row_group_row_idxs, g_rowkey.cols]
-    df2 = DataFrame(unstacked_val, Symbol.(renamecols.(colref_map)), copycols=false)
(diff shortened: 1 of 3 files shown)


bkamins:
@nalimilan - here only coverage fails (and it is false positive)

src/abstractdataframe/reshape.jl
```diff
@@ -418,10 +418,11 @@ function _unstack(df::AbstractDataFrame, rowkeys::AbstractVector{Int},
         mask_filled[row_id, col_id] = true
     end
 
-    # note that Symbol.(renamecols.(colref_map)) must produce unique column names
+    # note that Symbol(renamecols(colref_map)) must produce unique column names
```
nalimilan, line 421:
```suggestion
    # note that Symbol(renamecols(x)) must produce unique column names
```

test/reshape.jl
```diff
@@ -654,4 +654,12 @@ end
     @test IndexStyle(DataFrames.StackedVector) == IndexLinear()
 end
 
+@testset "empty unstack" begin
+    df = DataFrame(a = [], b = [], c = [])
+    dfu = unstack(df, :b, :c)
+    @test isempty(dfu)
+    @test names(dfu) == ["a"]
+    @test dfu.a isa Vector{Any}
```
nalimilan, line 662: Any particular reason to check this type?

nalimilan (approved):

test/reshape.jl
bkamins: For current and future safety I guess. I was surprised that:
```
julia> Symbol.(Any[])
Any[]
```
because I knew that:
```
julia> map(Symbol, Any[])
Symbol[]
```
so I thought it does not hurt to be extra careful with empty collections and add a test.

bkamins:
Thank you!

Issue #2846, opened 2021-08-29
The `test/show.jl` tests fail when Julia is started with `julia --color=no`


DilumAluthge:
The `test/show.jl` tests fail when Julia is started with `julia --color=no`

## Minimum Working Example (MWE)

All of the commands should be run in the same shell session so that the `JULIA_DEPOT_PATH` and `JULIA_PROJECT` environment variables have the same values across all steps and all commands.

### Step 1: Setup a temporary depot, `dev` DataFrames, and `add` the test dependencies

Run the following commands: 

```bash
export JULIA_DEPOT_PATH=$(mktemp -d)
echo $JULIA_DEPOT_PATH
export JULIA_PROJECT=$(mktemp -d)
echo $JULIA_PROJECT
git clone https://github.com/JuliaData/DataFrames.jl.git
cd DataFrames.jl
git checkout master
julia -e 'import Pkg; Pkg.develop(path = pwd())'
julia -e 'import Pkg; Pkg.add(["CategoricalArrays", "Dates", "Random", "Test"])' # test dependencies
```

### Step 2: Run the tests with `julia --color=yes`

Run the following command: 
```bash
JULIA_LOAD_PATH="@" julia --color=yes -e 'using Dates; include("test/show.jl")'
```

For me, the tests pass. Here is the output:
```
Test Summary:                            | Pass  Total
Basic show test with allrows and allcols |    9      9
Test Summary:    | Pass  Total
displaysize test |    4      4
Test Summary:             | Pass  Total
IOContext parameters test |    2      2
Test Summary:          | Pass  Total
SubDataFrame show test |    1      1
Test Summary:                                 | Pass  Total
Test showing StackedVector and RepeatedVector |    3      3
Test Summary:                                            | Pass  Total
Test colors and non-standard values: missing and nothing |    3      3
Test Summary:                  | Pass  Total
Test correct width computation |    1      1
Test Summary:              | Pass  Total
Test showing special types |    5      5
Test Summary:                              | Pass  Total
Test using :compact parameter of IOContext |    2      2
Test Summary:                              | Pass  Total
Test of DataFrameRows and DataFrameColumns |    4      4
Test Summary:                          | Pass  Total
Test empty data frame and DataFrameRow |    3      3
Test Summary: | Pass  Total
consistency   |    2      2
Test Summary:  | Pass  Total
wide type name |    6      6
Test Summary:  | Pass  Total
wide type name |    2      2
Test Summary: | Pass  Total
[...]

PR #2847, opened 2021-08-29
Disable color testing when color is not supported

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2846

I have tested these changes both with and without color.

### DIFF
--- a/test/io.jl
+++ b/test/io.jl
@@ -630,23 +630,25 @@ end
        8 │     8
        9 │     9  #undef"""
 
-    io = IOBuffer()
-    show(IOContext(io, :color => true), df)
-    str = String(take!(io))
-    @test str == """
-    \e[1m9×2 DataFrame\e[0m
-    \e[1m Row \e[0m│\e[1m A     \e[0m\e[1m B                                 \e[0m
-    \e[1m     \e[0m│\e[90m Int64 \e[0m\e[90m Any                               \e[0m
-    ─────┼──────────────────────────────────────────
-       1 │     1 \e[90m 9×2 DataFrame                     \e[0m
-       2 │     2 \e[90m 2-element DataFrameRow            \e[0m
-       3 │     3 \e[90m 1×2 SubDataFrame                  \e[0m
-       4 │     4 \e[90m 9-element DataFrameRows           \e[0m
-       5 │     5 \e[90m 2-element DataFrameColumns        \e[0m
-       6 │     6 \e[90m GroupedDataFrame with 9 groups b… \e[0m
-       7 │     7 \e[90m missing                           \e[0m
-       8 │     8 \e[90m                                   \e[0m
-       9 │     9 \e[90m #undef                            \e[0m"""
+    if Base.get_have_color()
+        io = IOBuffer()
+        show(IOContext(io, :color => true), df)
+        str = String(take!(io))
+        @test str == """
+        \e[1m9×2 DataFrame\e[0m
+        \e[1m Row \e[0m│\e[1m A     \e[0m\e[1m B                                 \e[0m
+        \e[1m     \e[0m│\e[90m Int64 \e[0m\e[90m Any                               \e[0m
+        ─────┼──────────────────────────────────────────
+           1 │     1 \e[90m 9×2 DataFrame                     \e[0m
+           2 │     2 \e[90m 2-element DataFrameRow            \e[0m
+           3 │     3 \e[90m 1×2 SubDataFrame                  \e[0m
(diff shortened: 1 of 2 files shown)


bkamins:
@nalimilan - let us put this on hold, as maybe https://github.com/KristofferC/Crayons.jl/issues/47 will be resolved and then it should not be required.

bkamins:
`Base.get_have_color` is not available in Julia 1.0, but I will keep it as is, since maybe this PR will not be needed given https://github.com/KristofferC/Crayons.jl/issues/47.

DilumAluthge:
In 1.0, instead of `Base.get_have_color()`, I think you can look for the `Base.have_color` book.

I would pursue this as a short-term solution. I'm guessing that https://github.com/KristofferC/Crayons.jl/issues/47 will require some time, and possible some breaking changes.

test/io.jl
```diff
@@ -630,23 +630,25 @@ end
...
-       1 │     1 \e[90m 9×2 DataFrame                     \e[0m
-       2 │     2 \e[90m 2-element DataFrameRow            \e[0m
-       3 │     3 \e[90m 1×2 SubDataFrame                  \e[0m
-       4 │     4 \e[90m 9-element DataFrameRows           \e[0m
-       5 │     5 \e[90m 2-element DataFrameColumns        \e[0m
-       6 │     6 \e[90m GroupedDataFrame with 9 groups b… \e[0m
-       7 │     7 \e[90m missing                           \e[0m
-       8 │     8 \e[90m                                   \e[0m
-       9 │     9 \e[90m #undef                            \e[0m"""
+    if Base.get_have_color()
```
bkamins, line 633:
```suggestion
    if VERSION >= v"1.6" && Base.get_have_color()
```

test/show.jl
```diff
@@ -215,26 +215,28 @@ end
...
-                   C = [:missing, "missing", missing])
-    @test sprint(show, df, context=:color=>true) == """
-        \e[1m3×3 DataFrame\e[0m
-        \e[1m Row \e[0m│\e[1m A       \e[0m\e[1m B       \e[0m\e[1m C       \e[0m
-        \e[1m     \e[0m│\e[90m Symbol? \e[0m\e[90m String? \e[0m\e[90m Any     \e[0m
-        ─────┼───────────────────────────
-           1 │ Symbol  \e[90m missing \e[0m missing
-           2 │\e[90m missing \e[0m String   missing
-           3 │ missing  missing \e[90m missing \e[0m"""
+    if Base.get_have_color()
```
bkamins, line 218:
```suggestion
    if if VERSION >= v"1.6" && Base.get_have_color()
```

bkamins:
Thank you. I just restricted these tests to Julia 1.6 or newer. I assume we will drop 1.0 support anyway soon when new LTS is out.

@nalimilan - given @DilumAluthge comment I mark it as ready for review.

Thank you!

test/show.jl
```diff
@@ -215,26 +215,29 @@ end
...
-    @test sprint(show, df, context=:color=>true) == """
-        \e[1m3×3 DataFrame\e[0m
-        \e[1m Row \e[0m│\e[1m A       \e[0m\e[1m B       \e[0m\e[1m C       \e[0m
-        \e[1m     \e[0m│\e[90m Symbol? \e[0m\e[90m String? \e[0m\e[90m Any     \e[0m
-        ─────┼───────────────────────────
-           1 │ Symbol  \e[90m missing \e[0m missing
-           2 │\e[90m missing \e[0m String   missing
-           3 │ missing  missing \e[90m missing \e[0m"""
+    # TODO: update when https://github.com/KristofferC/Crayons.jl/issues/47 is resolved
+    if if VERSION >= v"1.6" && Base.get_have_color()
```
bkamins, line 219:
```suggestion
    if VERSION >= v"1.6" && Base.get_have_color()
```

nalimilan (approved):

DilumAluthge:
Before you merge this, you may want to double-check it fixes the issue by doing e.g.

```
julia --color=no test/runtests.jl
```

Note that you probably don't want to do `Pkg.test`. `Pkg.test` runs the tests in a different process, and I'm not sure if the `--color` flag will get propagated to the subprocess.

bkamins:
> Before you merge this, you may want to double-check that it fixes the issue

As commented in the original post I have done this. Though thank you for double-checking, as these things are indeed tricky (I have run the test again now to be sure 😄).

bkamins:
Thank you!

Issue #2850, opened 2021-09-01
AsTable docstring doesn't mention it can be used as a target for select etc.


kleinschmidt:
I was flailing around trying to figure out how to expand a column full of named tuples into columns, and remembered to look at `AsTable` but the docstring makes it sound like it's only mean to be used to wrap input to the transform/select/combine functions.  What I needed as `:col => AsTable`; a mention of this possibility in the docstring would have helped avoid my confusion!  If I have a bit of time I'll make a PR but wanted to open an issue at least to track the confusion.

bkamins:
It is covered in https://github.com/JuliaData/DataFrames.jl/blob/3a71ae58d9b47d215a023ac4886f78b7f15b674c/src/abstractdataframe/selection.jl#L50:

> a `col => target_cols` pair, which renames the column `col` to `target_cols`, which must be single name (as a `Symbol` or a string), a vector of names or `AsTable`.

I agree this is a bit on a terse side. If you have a proposal how to improve this please comment or make a PR.

bkamins:
(it has to be updated in the linked source + in the split-apply-combine section of the manual that has the same passage)

BTW. I personally find `:cols => AsTable` a super nice addition. I have blogged on it recently: https://bkamins.github.io/julialang/2021/07/02/dataframes-1.2.0.html.

kleinschmidt:
Yup, agree all around.  Having that in the `AsTable` docstring would help with discoverability at the REPL.

bkamins:
Ah - good point. I will make a PR for this.

PR #2851, opened 2021-09-01
Improve docstring of AsTable

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2850

### DIFF
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -40,7 +40,7 @@ julia> select(df2, :a_b_identity => AsTable)
    2 │     2     12
    3 │     3     13
 
-julia> select(df, AsTable([:a, :b]) => ByRow(nt -> map(x -> x^2, nt)) => AsTable)
+julia> select(df1, AsTable([:a, :b]) => ByRow(nt -> map(x -> x^2, nt)) => AsTable)
 3×2 DataFrame       
  Row │ a      b     
      │ Int64  Int64 
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -8,9 +8,10 @@ selection operations supported by [`combine`](@ref), [`select`](@ref), [`select!
 If `AsTable(cols)` is used in `source` position it signals that the columns selected
 by the wrapped selector `cols` should be passed as a `NamedTuple` to the function.
 
-If `AsTable` is used in `destination` position it means that the result of the `transformation` operation
-is a vector of containers (or a single container if `ByRow` is used) that should be expanded 
-into multiple columns using `keys` to get column names.
+If `AsTable` is used in `destination` position it means that the result of
+the `transformation` operation is a vector of containers
+(or a single container if `ByRow(transformation)` is used)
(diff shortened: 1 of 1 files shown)


src/other/utils.jl
```diff
@@ -1,8 +1,54 @@
...
+julia> select(df2, :a_b_identity => AsTable)
+3×2 DataFrame
+ Row │ a      b     
+     │ Int64  Int64 
+─────┼──────────────
+   1 │     1     11
+   2 │     2     12
+   3 │     3     13
+
+julia> select(df, AsTable([:a, :b]) => ByRow(nt -> map(x -> x^2, nt)) => AsTable)
```
bkamins, line 43:
```suggestion
julia> select(df1, AsTable([:a, :b]) => ByRow(nt -> map(x -> x^2, nt)) => AsTable)
```

src/other/utils.jl
```diff
@@ -1,8 +1,54 @@
...
-wrapped selector should be passed as a `NamedTuple` to the function.
+A type having a special meaning in `source => transformation => destination`
+selection operations supported by [`combine`](@ref), [`select`](@ref), [`select!`](@ref),
+[`transform`](@ref), [`transform!`](@ref), [`subset`](@ref), and [`subset!`](@ref).
+
+If `AsTable(cols)` is used in `source` position it signals that the columns selected
+by the wrapped selector `cols` should be passed as a `NamedTuple` to the function.
+
+If `AsTable` is used in `destination` position it means that the result of the operation
+is a vector of containers that be expanded into multiple columns using `keys` to get column names.
```
kleinschmidt, line 12:
```suggestion
If `AsTable` is used in `destination` position it means that the result of the operation
is a vector of containers (or a single container if `ByRow` is used) that should be expanded 
into multiple columns using `keys` to get column names.
```

kleinschmidt (approved):
Generally looks great, thanks for the quick fix!!  One tiny suggestion

nalimilan (approved):

src/other/utils.jl
```diff
@@ -1,8 +1,55 @@
...
+A type having a special meaning in `source => transformation => destination`
+selection operations supported by [`combine`](@ref), [`select`](@ref), [`select!`](@ref),
+[`transform`](@ref), [`transform!`](@ref), [`subset`](@ref), and [`subset!`](@ref).
+
+If `AsTable(cols)` is used in `source` position it signals that the columns selected
+by the wrapped selector `cols` should be passed as a `NamedTuple` to the function.
+
+If `AsTable` is used in `destination` position it means that the result of the `transformation` operation
+is a vector of containers (or a single container if `ByRow` is used) that should be expanded 
+into multiple columns using `keys` to get column names.
```
bkamins, line 13:
```suggestion
If `AsTable` is used in `destination` position it means that the result of
the `transformation` operation is a vector of containers
(or a single container if `ByRow(transformation)` is used)
that should be expanded  into multiple columns using `keys` to get column names.
```

bkamins:
Thank you!

PR #2852, opened 2021-09-02
Fix three uses of "data table"

nalimilan:
These were likely missed when renaming back DataTables.jl to DataFrames.jl.


bkamins (approved):

bkamins:
Thank you!

Issue #2853, opened 2021-09-02
delete! in DataFrames.jl


bkamins:
We currently have `delete!` to remove rows from a `DataFrame`. However, `delete!` is defined for collections in Julia Base REMOVING A MAPPING for a passed key. Clearly this is not what `delete!` in DataFrames.jl does.

I propose to:
* deprecate `delete!` (it is a long term deprecation - it does not hurt to keep it)
* what `delete!` does should be `deleteat!`
* additionally implement `keepat!` to match Julia 1.7 behavior

(the issue is prompted by the fact that when we add `keepat!` it is strange to have `delete!` as its counterpart)

CC @nalimilan and @oxinabox (for end-user opinion as usual 😄)

nalimilan:
For reference, I introduced `delete!` at https://github.com/JuliaData/DataFrames.jl/pull/2189 but it should clearly have been called `deleteat!`.

PR #2854, opened 2021-09-02
deprecate delete!, define deleteat!

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2853

@nalimilan: the issue I was not sure about what would be best to do is if passed indices contain duplicates or are not sorted.
Intuitively I would prefer `deleteat!` to perform sorting and deduplication instead of throwing an error (the cost of this should not be super big in typical cases). Alternatively I could require in `keepat!` sorted and unique indices.


nalimilan:
Thanks. It's indeed a bit weird to have different requirements for `deleteat!` and `keepat!`. How costly is sorting and deduplication (when it's needed but more importantly when the input is already unique and sorted)? Do you have any idea whether implementing `keepat!` by calling `keepat!` on columns would make a difference for performance?

bkamins:
>  Do you have any idea whether implementing keepat! by calling keepat! on columns would make a difference for performance?

For this `keepat!` would have to be made available for vectors. I would do this this way assuming https://github.com/JuliaLang/Compat.jl/issues/750 is approved for adding. I would switch to using `keepat!` then as it should be faster.

nalimilan:
I guess it will be accepted in Compat, but we'll have to do it ourselves if we want it to happen.

bkamins:
OK - I will implement `keepat!` in Compat.jl when its design is settled in Julia Base (which means probably after Julia 1.7 release) then get back to this issue.

bkamins:
we should also add `insert!` method. I wait with this PR till `keepat!` is stabilized in Julia 1.7.

bkamins:
Also add `splice!`.

bkamins:
@nalimilan - I have decided to split `keepat!` to a separate issue (#2936). This PR is just about cleaning up `delete!` and `deleteat!`. Notably the fact that `keepat!` is missing is not a huge problem as one can easily do `deleteat!(df, Not(inds))`.

This PR should be good for a review.

src/dataframe/dataframe.jl
```diff
@@ -999,12 +999,13 @@ function Base.copy(df::DataFrame; copycols::Bool=true)
...
 """
-    delete!(df::DataFrame, inds)
+    deleteat!(df::DataFrame, inds)
 
 Delete rows specified by `inds` from a `DataFrame` `df` in place and return it.
 
 Internally `deleteat!` is called for all columns so `inds` must be:
-a vector of sorted and unique integers, a boolean vector, an integer, or `Not`.
+a vector of sorted and unique integers, a boolean vector, an integer.
+Additionally `Not` any valid selector is accepted.
```
nalimilan, line 1008:
```suggestion
a vector of sorted and unique integers, a boolean vector, an integer,
or `Not` wrapping any valid selector.
```

src/deprecated.jl
```diff
@@ -1,7 +1,11 @@
...
+# TODO: remove definitions in 2.0 release
 by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataFrames.jl. " *
                                              "Use the `combine(groupby(...), ...)` or `combine(f, groupby(...))` instead."))
-
 aggregate(args...; kwargs...) = throw(ArgumentError("aggregate function was removed from DataFrames.jl. " *
                                                     "Use the `combine` function instead."))
+
+# TODO: remove deprecation in 2.0 release
+import Base.delete!
+@deprecate delete!(df::DataFrame, inds) deleteat!(df::DataFrame, inds)
```
nalimilan, line 11: Add a test for the deprecated method?

test/dataframe.jl
```diff
@@ -582,74 +582,150 @@ end
...
+    @test_throws BoundsError keepat!(df, 0)
+    @test_throws BoundsError keepat!(df, [0])
+
+    df = DataFrame(a=[])
+    @test_throws BoundsError keepat!(df, 10)
+    @test_throws BoundsError keepat!(df, 0)
+    # the exception type changed between Julia 1.0.2 and Julia 1.1
+    # so we use their supertype below
+    @test_throws Exception keepat!(df, [10])
+    @test_throws Exception keepat!(df, [0])
```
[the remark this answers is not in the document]
bkamins: OK. I will add it. I was just lazy because tired of fixing such things since Julia Base is not considering changing of exception type as breaking.

src/deprecated.jl
bkamins: ok - added

bkamins:
This should be good for another review.

nalimilan (approved):

bkamins:
Thank you!

PR #2865, opened 2021-09-08
update docs following CSV.jl 0.9 release

### LINKED ISSUE #2864: update docs to CSV.jl 0.9

bkamins:
Fixes #2864

### DIFF
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -106,12 +106,12 @@ function compacttype(T::Type, maxwidth::Int=8)
     T === Any && return "Any"
     T === Missing && return "Missing"
 
-    sT = string(nameof(T))
+    sT = string(T isa Union ? T : nameof(T))
     textwidth(sT) ≤ maxwidth && return sT
 
     if T >: Missing
         T = nonmissingtype(T)
-        sT = string(nameof(T))
+        sT = string(T isa Union ? T : nameof(T))
         suffix = "?"
         textwidth(sT) ≤ maxwidth && return sT * suffix
     else
--- a/src/abstractdataframe/show.jl
+++ b/src/abstractdataframe/show.jl
@@ -106,12 +106,12 @@ function compacttype(T::Type, maxwidth::Int=8)
     T === Any && return "Any"
     T === Missing && return "Missing"
 
-    sT = string(T isa Union ? T : nameof(T))
+    sT = string(T)
     textwidth(sT) ≤ maxwidth && return sT
 
     if T >: Missing
         T = nonmissingtype(T)
-        sT = string(T isa Union ? T : nameof(T))
+        sT = string(T)
         suffix = "?"
         textwidth(sT) ≤ maxwidth && return sT * suffix
     else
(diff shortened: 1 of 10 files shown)


bkamins:
waiting for a decision in https://github.com/JuliaData/WeakRefStrings.jl/issues/85

bkamins:
@nalimilan - I have updated the docs. I left `String[X]` types and added a paragraph explaining how they work.

quinnj (approved):
I restarted the jobs now that WeakRefStrings has released

nalimilan (approved):
Looks much better now!

nalimilan:
Though doctests fail due to that `String15…` weirdness:
```julia
│ Evaluated output:
│ 
│ 3×5 DataFrame
│  Row │ Species          SepalLength_mean  SepalWidth_mean  PetalLength_mean  P ⋯
│      │ String15…        Float64           Float64          Float64           F ⋯
│ ─────┼──────────────────────────────────────────────────────────────────────────
│    1 │ Iris-setosa                 5.006            3.418             1.464    ⋯
│    2 │ Iris-versicolor             5.936            2.77              4.26
│    3 │ Iris-virginica              6.588            2.974             5.552
│                                                                 1 column omitted
│ 
│ Expected output:
│ 
│ 3×5 DataFrame
│  Row │ Species          SepalLength_mean  SepalWidth_mean  PetalLength_mean  P ⋯
│      │ String15         Float64           Float64          Float64           F ⋯
│ ─────┼──────────────────────────────────────────────────────────────────────────
│    1 │ Iris-setosa                 5.006            3.418             1.464    ⋯
│    2 │ Iris-versicolor             5.936            2.77              4.26
│    3 │ Iris-virginica              6.588            2.974             5.552
│                                                                 1 column omitted
```

Cc: @ronisbr

bkamins:
> Though doctests fail due to that String15… weirdness:

I will check what is the reason of failure and fix it - it is most likely on DataFrames.jl side. The strange thing is that what I put into this PR passes on my local machine.

bkamins:
The problem was the following:
```
julia> using DataFrames

julia> import Dates

julia> DataFrame(d=Dates.Date(1))
1×1 DataFrame
 Row │ d
     │ Date…
─────┼────────────
   1 │ 0001-01-01

julia> using Dates

julia> DataFrame(d=Dates.Date(1))
1×1 DataFrame
 Row │ d
     │ Date
─────┼────────────
   1 │ 0001-01-01
```

Now I have fixed it to never try printing type module name as a prefix (so the latter output is presented consistently).

In this way data frame is displayed in the same way no matter if the module is loaded with `using` or with `import` or only indirectly by some other module.

src/abstractdataframe/show.jl
```diff
@@ -106,12 +106,12 @@ function compacttype(T::Type, maxwidth::Int=8)
     T === Any && return "Any"
     T === Missing && return "Missing"
 
-    sT = string(T)
+    sT = string(T isa Union ? T : nameof(T))
```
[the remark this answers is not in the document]
bkamins: It is even more complicated 😞. I will try to propose something reasonable. The problem is that `nameof` drops parameters from parametric types.

nalimilan (approved):

bkamins:
The patches lead me to uncover that we printed `MD` type incorrectly previously which got uncaught and that we inconsistently printed type information in text vs HTML/LaTeX. Now I made it more consistent (use 9 text width characters by default - @ronisbr - this might affect your work on HTML and LaTeX output, but these are minor changes)

bkamins:
Thank you!

test/show.jl
```diff
@@ -301,8 +301,8 @@ end
     @test sprint(show, df) == """
         1×3 DataFrame
          Row │ a           b                    c
-             │ Date        DateTime             Day
-        ─────┼────────────────────────────────────────
+             │ Date        DateTime             Dates.Day
```
nalimilan: Just curious: why don't we print `Dates.DateTime`?

Anyway it would make sense to never print the module name to avoid this kind of variation (which is hard to understand for users).

nalimilan: If that's doable, I'd say that taking into account the final size width of the column (including contents) would be better. Otherwise there's some wasted space which in some cases could be used to print relevant information.

bkamins: @ronisbr - is this EASILY doable? (if it is hard it is probably better to focus on HTML/LaTeX support first). Thank you!

Issue #2867, opened 2021-09-08
Better error messages when frame is empty


jtrakk:
```jl
julia> df = DataFrame()
0×0 DataFrame

julia> rename(df, :a => :b)
ERROR: ArgumentError: Tried renaming :a to :b, when :a does not exist in the Index.
```

Many functions fail when the DataFrame is empty but they don't mention that it was empty in the error. It would help with debugging if it did mention it was empty.

bkamins:
The error tells you that `:a` is not present in `df`. Can you please suggest what other error you would like to have as for me the error message seems clear (i.e. I understand it is confusing you, but I need to know the exact wording of the message that would be clearer to you).

jtrakk:
This normally comes up when I construct a dataframe dynamically like

```jl
julia> xs = [];

julia> DataFrame((;a=x) for x in xs);

julia> df = DataFrame((;a=x) for x in xs);

julia> rename(df, :a => :b)
ERROR: ArgumentError: Tried renaming :a to :b, when :a does not exist in the Index.
```

"Tried renaming :a to :b, when :a does not exist in the Index." is nice, but it can supplement that with more information. For example "The DataFrame had 0 columns and 0 rows."

jtrakk:
This also applies to other functions like `sort` which says `ArgumentError: column name :x not found in the data frame`. I'd want to have a hint at the reason -- there are no columns in the data frame. It could even just say 
```
ERROR: ArgumentError: column name :x not found in the (0x0) data frame
```

PR #2868, opened 2021-09-09
code cleanup to improve error messages

bkamins:
In this PR I want to collect changes that are not breaking but just aimed at improving error messages.
I will make it not-draft after some more suggestions what should be fixed are collected.

CC @pdeffebach

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -1124,7 +1124,7 @@ julia> filter!(AsTable(:) => nt -> nt.x == 1 || nt.y == "b", df)
    3 │     1  b
 ```
 """
-Base.filter!(f, df::AbstractDataFrame) = delete!(df, findall(!f, eachrow(df)))
+Base.filter!(f::Function, df::AbstractDataFrame) = delete!(df, findall(!f, eachrow(df)))
 Base.filter!((col, f)::Pair{<:ColumnIndex}, df::AbstractDataFrame) =
     _filter!_helper(df, f, df[!, col])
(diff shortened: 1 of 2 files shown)


bkamins:
TODO: improve error message in `sort` per https://github.com/JuliaData/DataFrames.jl/issues/2867#issuecomment-917183223 suggestion.

bkamins:
I have cleaned up several things per user requests. In particular I avoid throwing errors mentioning Index, as they are probably not clear for users.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1124,7 +1124,7 @@ julia> filter!(AsTable(:) => nt -> nt.x == 1 || nt.y == "b", df)
    3 │     1  b
 ```
 """
-Base.filter!(f, df::AbstractDataFrame) = delete!(df, findall(!f, eachrow(df)))
+Base.filter!(f::Function, df::AbstractDataFrame) = delete!(df, findall(!f, eachrow(df)))
```
bkamins, line 1127: we can add this restriction here, as it is required later anyway. It simplifies error messages

src/other/index.jl
```diff
@@ -285,6 +290,10 @@ end
     if i === nothing
         candidates = fuzzymatch(l, idx)
         if isempty(candidates)
+            if isempty(l)
+                throw(ArgumentError("column name :$idx not found in the " *
+                                    "data frame with no columns"))
```
nalimilan, line 295:
```suggestion
                                    "data frame since it has no columns"))
```

nalimilan (approved):

bkamins:
Thank you!

Issue #2439, opened 2020-09-18
All() should return a tuple instead of positional arguments


ppalmes:
- All() operates on collection of columns and typically, you need to operate in many columns so it is easier to process vector or tuple of rows instead of using positional arguments.
- All() -> ByRow(sum or mean) won't work including other statistical metrics because All returns positional arguments and most functions process either vectors or tuples.
- All()=>ByRow((x...) -> sum(x)) or AsTable(All()) => (x->sum(x)) is needed but it is messy because why do you have to transform All() into a table because from the user perspective, dataframe is a table already. It is not clear why you need AsTable with All(). This happens because All() returns a positional value instead of a collection like vector or tuple
- most functions return scalar, or tuple, or vector. it is confusing to find All() returning positional argument for typical user. it is convenient to deal with two or three columns but many use-cases require operating each row across several columns

bkamins:
> `All() -> ByRow(sum or mean)` won't work

`All() => ByRow(+)` works, as for mean indeed you have to write `combine(df, AsTable(All()) => ByRow(mean))`

> `AsTable(All()) => (x->sum(x))` is needed

It is enough to write:
```
All() => +
```
or
```
AsTAble(All()) => sum
```
in this case

> but many use-cases require operating each row across several columns

As a side note. These operations will be inefficient using a `DataFrame`, if you have such a use case probably a different data structure should be used (DataFrames.jl is optimized for column access not for row access).

---

Having said that, we could have another wrapper similar to `AsTable`, e.g. `AsTuple` that would produce a `Tuple` instead.

pdeffebach:
Thanks for the issue report! 

I acknowledge that `All()` is awkward in this context since it has an undefined number of inputs so it mirrors splatting more. However note that

1. Plenty of functions allow `spat`ting, i.e. `+`, `string`, `*`, so these things will just work with current behavior.
2. I disagree that "working with many many columns" is the dominant use-case. See some code I wrote for a personal project [here](https://gitlab.com/pdeffebach/senate-voting/-/blob/master/src/import_clean_funs.jl). Plenty of uses of `transform` with positional arguments. 
3.   > into a table because from the user perspective, dataframe is a table already. It is not clear why you need AsTable with All(). This happens because All() returns a positional value instead of a collection like vector or tuple

note that you aren't advocating making a DataFrame the default output of `All()`, so this problem wouldn't be fixed by your proposal. 

[...]

Issue #2768, opened 2021-05-19
Fast row aggregation in DataFrames.jl


sl-solution:
As you might know `DataFrame` is optimised for column operations, and row operations are not efficient. There are some solutions for this and has been previously discussed (#2440, #2757,  #2439,  #952, ...).

Based on my knowledge, the most efficient way (similar-to-work-with-matrix performance) to do this when the problem is fitted into map and reduce, is using `mapreduce`,  e.g.

``` 
df = DataFrame(rand(10^5, 100), :auto)
op(x, y) = x .+= y
mapreduce(identity, op, eachcol(df), init = zeros(nrow(df)))
```

At the beginning I thought this is very trivial and just having some documentations about `mapreduce` should be enough for `DataFrames.jl` users. However, thinking about it for a while, I guess this is not that much trivial as I thought (particularly if `missings` are present). Thus, I think having a bunch of common row operations inside `DataFrames.jl` would be helpful, particularly, the operations which take care of `missing` automatically. Since I know this may be controversial, at the moment, I develop a package `DFRowOperation.jl` to define and store a set of common row operations. The users may use, contribute and evaluate this package and if it make sense, it would be great to add its functionality into `DataFrames.jl` in future.

you may access the package at

https://github.com/sl-solution/DFRowOperation.jl

and currently contains the following functions
```
row_sum,
row_prod,
row_mean,
row_count,
row_minimum,
row_maximum,
row_var,
row_std,
row_stdze!,
row_stdze,
row_sort!,
row_sort,
row_anymissing
```

bkamins:
This makes a lot of sense. However, if you allow 😄 (given you have already put a lot of work into it) in DataFrames.jl I would do the following (this is easy to do, and is more extensible). I will give an example of only `sum`, but this extends to other functions.

When user writes:
```
select(df, cols => ByRow(sum) => outcol)
```
or (if handling missing values is required)
```
select(df, cols => ByRow(sum∘skipmissing) => outcol)
```
then we can intercept `ByRow(sum)` (and other). The key benefit of our current design of `ByRow` (and it was intentional) is that:
```
julia> typeof(ByRow(sum))
ByRow{typeof(sum)}
```
so we can dispatch on it and provide an efficient aggregation path both for `DataFrame` and `GroupedDataFrame` (similarly to how we currently do efficient aggregations in for `GroupedDataFrame`).

[...]

Issue #2440, opened 2020-09-18
Investigate performance of aggregations


bkamins:
The case that should be improved (or checked why it is problematic):
```
julia> df = DataFrame(rand(10^5, 40));

julia> @time combine(df, AsTable(1:40) => ByRow(sum));
  0.031775 seconds (144 allocations: 31.294 MiB, 15.38% gc time)

julia> @time combine(df, 1:40 => ByRow(+));
  2.719783 seconds (69.70 M allocations: 2.410 GiB, 9.69% gc time)
```

bkamins:
This is the reason:
```
julia> f(a) = +(a,a,a,a,a,a,a,a,a,a,a,a,a,a,a,a,a);

julia> @code_warntype f(a)
Variables
  #self#::Core.Compiler.Const(f, false)
  a::Array{Float64,1}

Body::Array{Float64,1}
1 ─ %1 = (a + a + a + a + a + a + a + a + a + a + a + a + a + a + a + a + a)::Array{Float64,1}
└──      return %1

julia> f(a) = +(a,a,a,a,a,a,a,a,a,a,a,a,a,a,a,a,a,a);

julia> @code_warntype f(a)
Variables
  #self#::Core.Compiler.Const(f, false)
  a::Array{Float64,1}

Body::Any
1 ─ %1 = (a + a + a + a + a + a + a + a + a + a + a + a + a + a + a + a + a + a)::Any
└──      return %1
```

bkamins:
@nalimilan - so it seems we have an inherent limitation to splatting as at some point compiler gives up with optimization.

bkamins:
So the conclusion is:
1. we should add an information in the documentation that having more than 16 columns in the default (splatting) syntax is likely to be slow;
2. we should add `AsVector` or `AsTuple` wrapper for cases when you want to aggregate across e.g. 1000 columns (most likely of a homogenous type). This is tracked in #2439. The reason is that currently there is no way to perform it fast without huge run-time or compile-time cost. I will investigate both options and give some recommendation what should be added.

bkamins:
The solution is the following type:
```
"""
    VectorIterator(::AbstractDataFrame)

Return a `Vector` iterator over rows of a passed.
The returned vector is reused between iterations for efficiency.
This iterator is not thread safe.
"""
struct VectorIterator{S, T}
    v::Vector{S}
    x::Vector{T}

    function VectorIterator(df::AbstractDataFrame)
[...]

PR #2869, opened 2021-09-09
Add fast reductions

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2439 https://github.com/JuliaData/DataFrames.jl/issues/2768 https://github.com/JuliaData/DataFrames.jl/issues/2440

I have implemented a proof of concept for `sum`. Let us discuss if we like this approach. If we do we can extend the list of supported reductions (and comments what reductions we would like to support are welcome).

CC @nalimilan @pdeffebach

### DIFF
--- a/src/abstractdataframe/selection.jl
+++ b/src/abstractdataframe/selection.jl
@@ -376,12 +376,17 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
 
 _empty_astable_helper(fun, len) = [fun(NamedTuple()) for _ in 1:len]
 
+@noinline _table_transformation(df_sel::AbstractDataFrame, fun) = fun(Tables.columntable(df_sel))
+
 function _transformation_helper(df::AbstractDataFrame, col_idx::AsTable, (fun,)::Ref{Any})
-    tbl = Tables.columntable(select(df, col_idx.cols, copycols=false))
-    if isempty(tbl) && fun isa ByRow
+    df_sel = select(df, col_idx.cols, copycols=false)
+    if ncol(df_sel) == 0 && fun isa ByRow
         return _empty_astable_helper(fun.fun, nrow(df))
+    elseif typeof(fun) === typeof(Base.sum) || typeof(fun) === ByRow{typeof(Base.sum)}
+        # apply map to get consistency of results with the method in the non-AsTable case
+        return sum(map(identity, eachcol(df_sel)))
     else
-        return fun(tbl)
+        return _table_transformation(df_sel, fun)
     end
 end
 
@@ -392,6 +397,9 @@ function _transformation_helper(df::AbstractDataFrame, col_idx::AbstractVector{I
         return _empty_selector_helper(fun.fun, nrow(df))
     else
         cdf = eachcol(df)
+        if typeof(fun) === typeof(Base.:+) || typeof(fun) === ByRow{typeof(Base.:+)}
+            return sum(map(c -> cdf[c], col_idx))
+        end
         return fun(map(c -> cdf[c], col_idx)...)
     end
 end
--- a/test/select.jl
+++ b/test/select.jl
@@ -1678,4 +1678,26 @@ end
     @test select(df, :a => ByRow(f) => :a) == DataFrame(a=1:3)
 end
 
+@testset "wide reductions" begin
+    Random.seed!(1234)
+    df = DataFrame(rand(Int, 2, 20_000), :auto)
+    df.x100 = [1, 2]
+    df.x1000 = Union{Int, Missing}[1, 2]
+    df.x10000 = Union{Float64, Missing}[1, missing]
+
+    @test @elapsed(select(df, All() => (+) => :res)) < 10.0
+    @test @elapsed(select(df, Between(:x2,:x19999) => ByRow(+) => :res)) < 10.0
+    @test @elapsed(select(df, AsTable(All()) => sum => :res)) < 10.0
+    @test @elapsed(select(df, AsTable(Not(:x99)) => ByRow(sum) => :res)) < 10.0
+
+    for sel in (All(), Between(:x2,:x19999), Not(:x99))
+        @show sel
+        @test select(df, sel => (+) => :res) ≅
+              select(df, sel => ByRow(+) => :res) ≅
+              select(df, AsTable(sel) => sum => :res) ≅
+              select(df, AsTable(sel) => ByRow(sum) => :res) ≅
+              DataFrame(res=sum(collect(eachcol(select(df, sel)))))
+    end
+end
+
 end # module


bkamins:
TODO: when we decide on what we do update documentation and NEWS.md

bkamins:
@nalimilan The thing to discuss is the following tension:
This is what we currently do:
```
julia> df = DataFrame(x1=[1,2,missing], x2=[4.0,missing,missing])
3×2 DataFrame
 Row │ x1       x2        
     │ Int64?   Float64?  
─────┼────────────────────
   1 │       1        4.0
   2 │       2  missing   
   3 │ missing  missing   

julia> select(df, AsTable(:) => ByRow(sum∘skipmissing))
3×1 DataFrame
 Row │ x1_x2_sum_skipmissing 
     │ Real
─────┼───────────────────────
   1 │                   5.0
   2 │                   2
   3 │                   0

julia> df = DataFrame(x1=Any[missing,2,3], x2=Any[4.0,missing,missing])
3×2 DataFrame
 Row │ x1       x2      
     │ Any      Any     
─────┼──────────────────
   1 │ missing  4.0     
   2 │ 2        missing 
   3 │ 3        missing 

julia> select(df, AsTable(:) => ByRow(sum∘skipmissing))
3×1 DataFrame
 Row │ x1_x2_sum_skipmissing 
     │ Real
─────┼───────────────────────
   1 │                   4.0 
   2 │                   2   
   3 │                   3   
```
Which I think is not very nice. I would rather prefer to do `promote_type` of `eltype` of all columns and based on this determine the eltype of the output. In the first case above it would nicely produce a vector of `Float64` - which I think is preferable.
However, the consequence is that having vectors of `Any` would be problematic as we would not be able to establish `0` in this case. What I would propose to do is to internally narrow-down the eltype of such vectors. If it is still `Any` - error (as it means we have non numeric or missing values in the column). Otherwise perform the calculation using the columns with narrowed down eltype. What do you think?

nalimilan:
Use `promote_type` would indeed make sense. That would be a difference between `ByRow` and plain broadcasting, right?

Regarding vector with eltype `Any`, I'd rather not make any particular efforts to make them work. Ideally we would do the same thing as without the fast path, even if that involves taking the slow path for them.

bkamins:
> That would be a difference between `ByRow` and plain broadcasting, right?

It is unrelated to `ByRow`/broadcasting. Neither `map` nor broadcasting do `promote_type` the way I propose (as you see in my examples the problem is that each row is treated differently based on mixture of information about both eltypes of columns and values stored in a given row). What I propose is essentially to determine the result type based ONLY on eltypes of columns (which is superior to what what we have now except for the case of columns with `Any` type).

> Ideally we would do the same thing as without the fast path, even if that involves taking the slow path for them.

The problem is that taking a slow path will not yield any result as for wide inputs Julia crashes - this PR is meant to avoid these crashes in the first place.
Instead of narrowing of eltype we could also just error if the column has `Any` eltype and produce an informative error message. I think in general we should discourage users from using `Any` eltype in columns.

bkamins:
I have pushed the proposal for summation with skipping missings. It works as follows:
```
julia> df = DataFrame(x1=[1,2,missing], x2=[4.0,missing,missing])
3×2 DataFrame
 Row │ x1       x2        
     │ Int64?   Float64?  
─────┼────────────────────
   1 │       1        4.0 
   2 │       2  missing   
   3 │ missing  missing   

julia> select(df, AsTable(:) => ByRow(sum∘skipmissing))
3×1 DataFrame
 Row │ x1_x2_sum_skipmissing 
     │ Float64
─────┼───────────────────────
   1 │                   5.0 
   2 │                   2.0 
   3 │                   0.0 

julia> df = DataFrame(x1=Any[missing,2,3], x2=Any[4.0,missing,missing])
3×2 DataFrame
 Row │ x1       x2      
     │ Any      Any     
─────┼──────────────────
   1 │ missing  4.0     
   2 │ 2        missing 
   3 │ 3        missing 

julia> select(df, AsTable(:) => ByRow(sum∘skipmissing))
ERROR: ArgumentError: The reduced element type Any does not support zero that is required to perform summation. Narrowing down element types of passed columns should be performed first.    
```

It is possible to try harder to deduce the summation result in the second case, but I am not sure it is useful (it would add a lot of complexity to the code in the case that I am not sure we want to support anyway).

nalimilan:
I see two difficult points to address:
- Changing the behavior of `ByRow` is breaking. Minor changes to the eltype are probably fine (especially if we make them more narrow). But throwing an error in cases that don't is problematic, and that would be the case for `Any`: even though it crashes with a large number of columns, it works in other cases. That's why I think that falling back to the current slow path would make sense to avoid any regressions.
- The fast path for aggregations should be an implementation detail. `ByRow(f)` should be equivalent to `ByRow(x -> f(x))`, modulo minor differences like rounding errors or possibly the former working in cases where the latter fails. So if we use `promote_type` for reductions, we should probably use it for all functions  (even if that implies converting the vector returned by `map` manually).

bkamins:
> That's why I think that falling back to the current slow path would make sense to avoid any regressions.

OK - I will change it this way

> The fast path for aggregations should be an implementation detail. `ByRow(f)` should be equivalent to `ByRow(x -> f(x))`

I do not think it is possible. This is exactly the same problem with as fast aggegations for `GroupedDataFrame`.

If I write `AsTable(some_cols) => ByRow(f)` I do not know how `f` would use the `NamedTuple` passed. E.g. it might rely on column names etc.

The other problem is that even if we did some way to signal that `f` is a reduction the general approach will be slower than what is proposed here as in e.g. `sum` or `sum∘skipmissing` we explicitly use the knowledge of the way the reduction should be performed to make it efficient.

nalimilan:
> I do not think it is possible. This is exactly the same problem with as fast aggegations for `GroupedDataFrame`.
> 
> If I write `AsTable(some_cols) => ByRow(f)` I do not know how `f` would use the `NamedTuple` passed. E.g. it might rely on column names etc.
> 
> The other problem is that even if we did some way to signal that `f` is a reduction the general approach will be slower than what is proposed here as in e.g. `sum` or `sum∘skipmissing` we explicitly use the knowledge of the way the reduction should be performed to make it efficient.

What do you mean with "not possible"? What I suggest is to only enable optimizations for known aggregation functions for which we know the equivalence holds -- just like for grouped reduction. Isn't that possible here, or do you just mean that it's not possible in general for any function?

bkamins:
Ah, you mean with:

> `ByRow(f)` should be equivalent to `ByRow(x -> f(x))`

That the output should be identical - not the code path. Then it is possible but it will be slow and I am not sure if it is desirable. See the first example:
```
julia> df = DataFrame(x1=[1,2,missing], x2=[4.0,missing,missing])
3×2 DataFrame
 Row │ x1       x2        
     │ Int64?   Float64?  
─────┼────────────────────
   1 │       1        4.0
   2 │       2  missing   
   3 │ missing  missing   

julia> select(df, AsTable(:) => ByRow(sum∘skipmissing))
3×1 DataFrame
 Row │ x1_x2_sum_skipmissing 
     │ Real
─────┼───────────────────────
   1 │                   5.0
   2 │                   2
   3 │                   0
```
Retaining this behavior is possible, but will be complex to achieve and I would argue that the result is not something I think most people would expect or want.

nalimilan:
I'd rather do the reverse actually: always choose the return type using `promote_type`, even if that implies making a copy with a more narrow type after the fact in the general case. That's doable, right?

bkamins:
What is doable is the following (and this is the correct way to do it in fast path, as my current implementation is incorrect):
* fast path:
    - try doing `zero` on `eltype` of each column, except if some column has `Missing` eltype - then ignore such column; if it fails - fallback to slow path; if it works - sum these zeros. Use the sum of zeros to create the `init` for reduction;
* slow path (i.e. some column eltype does not support `zero`): do what we do (and accept that in corner cases it will error), then do `promote_type` on all values in the produced vector; convert the produced vector to the vector with the determined type.

I will now implement this to show how it will look (as I assume this is what you agree with 😄, but of course if you do not like it we can change it).

bkamins:
OK - I have pushed some tentative implementation. I need to check its correctness and performance.

bkamins:
It does the following now:
```
julia> df = DataFrame(x1=[1,2,missing], x2=[4.0,missing,missing])
3×2 DataFrame
 Row │ x1       x2        
     │ Int64?   Float64?  
─────┼────────────────────
   1 │       1        4.0 
   2 │       2  missing   
   3 │ missing  missing   

julia> select(df, AsTable(:) => ByRow(sum∘skipmissing))
3×1 DataFrame
 Row │ x1_x2_sum_skipmissing 
     │ Float64
─────┼───────────────────────
   1 │                   5.0 
   2 │                   2.0 
   3 │                   0.0 

julia> df = DataFrame(x1=Any[missing,2,3], x2=Any[4.0,missing,missing])
3×2 DataFrame
 Row │ x1       x2      
     │ Any      Any     
─────┼──────────────────
   1 │ missing  4.0     
   2 │ 2        missing 
   3 │ 3        missing 

julia> select(df, AsTable(:) => ByRow(sum∘skipmissing))
3×1 DataFrame
 Row │ x1_x2_sum_skipmissing 
     │ Float64
─────┼───────────────────────
   1 │                   4.0 
   2 │                   2.0 
   3 │                   3.0 
```

bkamins:
And the timing is as follows:
```
julia> df = DataFrame(rand(10_000, 10_000), :auto);

julia> @time select(df, AsTable(:) => ByRow(sum∘skipmissing));
  0.082187 seconds (19.66 k allocations: 1.142 MiB)
```

in comparison to:
```
julia> @time sum(eachcol(df));
  0.461784 seconds (20.00 k allocations: 763.626 MiB, 41.38% gc time)
```
and
```
julia> x = collect(eachcol(df));

julia> @time sum(x);
  0.400074 seconds (20.00 k allocations: 763.626 MiB, 32.77% gc time)

julia> m = Matrix(df);

julia> @time sum(m, dims=2);
  0.054808 seconds (6 allocations: 78.281 KiB)
```

So we have some overhead over `Matrix` but I think it is acceptable.

bkamins:
If I get an approval for this design I will update docs and tests. Then I will move forward with other reductions.
@nalimilan - can you please list which reductions you think it is worth to add? (as I would prefer to focus on selected that are really worth adding and not do any conceivable one)

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+
 function _transformation_helper(df::AbstractDataFrame, col_idx::AsTable, (fun,)::Ref{Any})
-    tbl = Tables.columntable(select(df, col_idx.cols, copycols=false))
-    if isempty(tbl) && fun isa ByRow
+    df_sel = select(df, col_idx.cols, copycols=false)
+    if ncol(df_sel) == 0 && fun isa ByRow
         return _empty_astable_helper(fun.fun, nrow(df))
+    elseif typeof(fun) === typeof(Base.sum) || typeof(fun) === ByRow{typeof(Base.sum)}
+        return _sum_fast(map(identity, eachcol(df_sel)))
+    elseif typeof(fun) === ByRow{typeof(Base.:∘(Base.sum, Base.skipmissing))}
```
nalimilan, line 422: Isn't this equivalent?
```suggestion
    elseif fun === sum || fun === ByRow(sum)
        return _sum_fast(map(identity, eachcol(df_sel)))
    elseif fun === ByRow(sum∘skipmissing)
```

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+        fastsum = _sum_skipmissing_fast(map(identity, eachcol(df_sel)))
```
nalimilan, line 423:
```suggestion
        fastsum = _sum_skipmissing_fast(collect(eachcol(df_sel)))
```

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+        if isnothing(fastsum)
+            slowsum = _table_transformation(df_sel, fun)
+            T = mapreduce(typeof, promote_type, slowsum)
+            return convert(Vector{T}, slowsum)
```
nalimilan, line 427:
```suggestion
            return convert(AbstractVector{T}, slowsum)
```

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+                break
+            else
+                throw(e)
+            end
+        end
+    end
+    # this will happen if eltype of some columns do not support zero
+    # or all columns have eltype Missing
+    isnothing(sumz) && return nothing
+    init = fill(sumz, length(cols[1]))
```
nalimilan, line 409:
```suggestion
    init = fill!(Tables.allocatecolumn(typeof(sumz), length(cols[1])), sumz)
```

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+    for col in cols
+        try
+            zi = zero(eltype(col))
+            if isnothing(sumz)
+                sumz = zi
+            elseif !ismissing(zi) # zi is missing if eltype is Missing
+                sumz += zi
+            end
+        catch e
+            if e isa MethodError && e.f === Base.zero
```
nalimilan, line 398:
```suggestion
            if e isa MethodError && e.f === zero
```

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+@noinline _table_transformation(df_sel::AbstractDataFrame, fun) = fun(Tables.columntable(df_sel))
+
+function _sum_fast(cols::Vector{<:AbstractVector})
+    isempty(cols) && throw(ArgumentError("No columns selected for reduction"))
+    return sum(cols)
+end
+
+function _sum_skipmissing_fast(cols::Vector{<:AbstractVector})
+    isempty(cols) && throw(ArgumentError("No columns selected for reduction"))
+    sumz = nothing
```
nalimilan, line 388: Technically, `zero(eltype(col))` could return `nothing` for weird custom types. I wonder whether `@isdefined` couldn't be used instead.

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+        if isnothing(fastsum)
+            slowsum = _table_transformation(df_sel, fun)
+            T = mapreduce(typeof, promote_type, slowsum)
+            return convert(Vector{T}, slowsum)
+        else
+            return fastsum
+        end
     else
-        return fun(tbl)
+        return _table_transformation(df_sel, fun)
```
nalimilan, line 432: Also narrow element type in this branch?

src/abstractdataframe/selection.jl
```diff
@@ -392,6 +440,9 @@ function _transformation_helper(df::AbstractDataFrame, col_idx::AbstractVector{I
         return _empty_selector_helper(fun.fun, nrow(df))
     else
         cdf = eachcol(df)
+        if typeof(fun) === typeof(Base.:+) || typeof(fun) === ByRow{typeof(Base.:+)}
```
nalimilan, line 443:
```suggestion
        if fun === + || fun === ByRow(+)
```

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+    df_sel = select(df, col_idx.cols, copycols=false)
+    if ncol(df_sel) == 0 && fun isa ByRow
         return _empty_astable_helper(fun.fun, nrow(df))
+    elseif typeof(fun) === typeof(Base.sum) || typeof(fun) === ByRow{typeof(Base.sum)}
+        return _sum_fast(map(identity, eachcol(df_sel)))
+    elseif typeof(fun) === ByRow{typeof(Base.:∘(Base.sum, Base.skipmissing))}
+        fastsum = _sum_skipmissing_fast(map(identity, eachcol(df_sel)))
+        if isnothing(fastsum)
+            slowsum = _table_transformation(df_sel, fun)
+            T = mapreduce(typeof, promote_type, slowsum)
```
nalimilan, line 426: Maybe this could be slightly faster?
```suggestion
            T = isconcretetype(nonmissingtype(T)) ? T : mapreduce(typeof, promote_type, slowsum)
```

src/abstractdataframe/selection.jl
```diff
@@ -376,12 +376,60 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+end
+
 function _transformation_helper(df::AbstractDataFrame, col_idx::AsTable, (fun,)::Ref{Any})
-    tbl = Tables.columntable(select(df, col_idx.cols, copycols=false))
-    if isempty(tbl) && fun isa ByRow
+    df_sel = select(df, col_idx.cols, copycols=false)
+    if ncol(df_sel) == 0 && fun isa ByRow
         return _empty_astable_helper(fun.fun, nrow(df))
+    elseif typeof(fun) === typeof(Base.sum) || typeof(fun) === ByRow{typeof(Base.sum)}
+        return _sum_fast(map(identity, eachcol(df_sel)))
```
nalimilan, line 421:
```suggestion
        return _sum_fast(collect(eachcol(df_sel)))
```
I guess specialization on column type is intended?

nalimilan:
Sounds good. I'd implement the same reductions as in grouping. Actually we could maybe reuse the same wrapper type, which could in theory allow users to implement their own reductions.

src/abstractdataframe/selection.jl
bkamins: I would prefer to use `map` here for the following reason:
```
julia> df = DataFrame(rand(2,3), :auto)
2×3 DataFrame
 Row │ x1        x2        x3
     │ Float64   Float64   Float64
─────┼──────────────────────────────
   1 │ 0.693707  0.583899  0.388987
   2 │ 0.976835  0.576875  0.893499

julia> map(identity, eachcol(df))
3-element Vector{Vector{Float64}}:
 [0.6937071807288211, 0.9768351037375032]
 [0.58389881666453, 0.5768753747948956]
 [0.3889871446622468, 0.8934986529968019]

julia> collect(eachcol(df))
3-element Vector{AbstractVector{T} where T}:
 [0.6937071807288211, 0.9768351037375032]
 [0.58389881666453, 0.5768753747948956]
 [0.3889871446622468, 0.8934986529968019]
```
Why do you think `collect` would be better?

src/abstractdataframe/selection.jl
bkamins: same as above

src/abstractdataframe/selection.jl
bkamins: I am not sure what you mean here. `T` is undefined before this line AFAICT.

src/abstractdataframe/selection.jl
bkamins: It is not needed. All elements of the result will already have the same CONCRETE type due to the fact that we update this vector:
```
init = fill!(Tables.allocatecolumn(typeof(sumz), length(cols[1])), sumz)
```
which has a concrete type.

src/abstractdataframe/selection.jl
bkamins: I prefer to use `map` because:
```
julia> df = DataFrame(rand(2,3), :auto)
2×3 DataFrame
 Row │ x1        x2        x3
     │ Float64   Float64   Float64
─────┼──────────────────────────────
   1 │ 0.693707  0.583899  0.388987
   2 │ 0.976835  0.576875  0.893499

julia> map(identity, eachcol(df))
3-element Vector{Vector{Float64}}:
 [0.6937071807288211, 0.9768351037375032]
 [0.58389881666453, 0.5768753747948956]
 [0.3889871446622468, 0.8934986529968019]

julia> collect(eachcol(df))
3-element Vector{AbstractVector{T} where T}:
 [0.6937071807288211, 0.9768351037375032]
 [0.58389881666453, 0.5768753747948956]
 [0.3889871446622468, 0.8934986529968019]
```

src/abstractdataframe/selection.jl
bkamins: same as above. do you agree?

src/abstractdataframe/selection.jl
bkamins: I was thinking about it and concluded it should not happen, as `zero` should return value of type that is passed, but I will fix it for safety.

src/abstractdataframe/selection.jl
bkamins: OK - done

src/abstractdataframe/selection.jl
bkamins: OK - I see what you mean, but it needs another syntax.

bkamins:
> Actually we could maybe reuse the same wrapper type, which could in theory allow users to implement their own reductions.

I guess we should use another wrapper type to achieve what you propose. However, before implementing this let us agree on the following.

`sum∘skipmissing` is optimized only for `AsTable(...) => ByRow(sum∘skipmissing)`

while `sum` is optimized for `... => +`, `... => ByRow(+)`, `AsTable(...) => sum`, and `AsTable(...) => ByRow(sum)`.

Do we want to put the optimizations for `sum` in all these places? Or only for `AsTable(...) => ByRow(sum)` and maybe `AsTable(...) => ByRow(sum)`?

The issue is that if we create a wrapper I need to know if I invoke it in both `AsTable(...) => ...` or also in `... => ...`. What do you think?

I would only call the wrapper in `AsTable(...) => ...`, then the question is if we want to keep the fast path for `... => +` and `... => ByRow(+)`.

bkamins:
In the end I think it is better not to use the wrapper type as you proposed, but just use multiple dispatch. I have implemented how I propose to handle this. Users will be allowed to add methods to `table_transformation` function.

pdeffebach:
Is adding a separate table_transformation step really less work than adding a `AsVector` wrapper? Since one of the major slowdowns seems to be constructing a large named tuple. Seems easier to at least construct the object a user requests and then dispatch the fast path based on the function type.

bkamins:
> Is adding a separate `table_transformation` step really less work than adding a `AsVector` wrapper?

The issue is two-fold:
1. performance - as you can see in the codes if we do e.g. `ByRow(sum∘skipmissing)` we actually do not process it by-row, but in a columnar fashion. This has the following consequence:

This PR:
```
julia> df = DataFrame(ones(10^7, 100), :auto);

julia> @time select(df, AsTable(:) => ByRow(sum∘skipmissing));
  2.946524 seconds (3.20 M allocations: 268.327 MiB, 14.51% gc time, 50.74% compilation time)

julia> @time select(df, AsTable(:) => ByRow(sum∘skipmissing));
  1.031689 seconds (255 allocations: 76.314 MiB, 0.59% gc time)

julia> @time select(df, AsTable(:) => ByRow(sum∘skipmissing));
  1.034903 seconds (255 allocations: 76.314 MiB, 0.34% gc time)
```
Current `main`:
```
julia> df = DataFrame(ones(10^7, 100), :auto);

julia> @time select(df, AsTable(:) => ByRow(sum∘skipmissing));
  5.808807 seconds (2.30 M allocations: 217.624 MiB, 6.73% gc time, 36.07% compilation time)

julia> @time select(df, AsTable(:) => ByRow(sum∘skipmissing));
  3.735678 seconds (365 allocations: 76.322 MiB, 0.07% gc time)

julia> @time select(df, AsTable(:) => ByRow(sum∘skipmissing));
  3.523625 seconds (365 allocations: 76.322 MiB, 0.05% gc time)
```
So as you can see it is over 3x slower because it processes data in a row-wise not whole column fashion.

So for performance we need custom code paths.

2. Correctness. `AsVector` would still require a custom code path for e.g. `ByRow(sum∘skipmissing)` to create an `init` value of proper type. And each reduction probably requires a different code for `init` value.

---

Having said that I for now opted-out from `AsVector` as we can add optimizations to `AsTable` path without complicating the minilanguage. Indeed the benefit of `AsVector` would be that users would be able to write "relatively fast" reductions that we did not think of (i.e. that are not hard coded). The question is how often do you think such a need arises? This is a discussion that would be good to have here before we finalize this PR.

bkamins:
E.g. note that apart from learning cost then you would have to handle it in DataFramesMeta.jl - so as you can see this adds technological debt.

OTOH - even if we go for `AsVector` these optimizations do not hurt anyway as they are "transparent" to the user.

bkamins:
Thinking of it more, what I think we could add is a `df::Bool` kwarg to `AsTable` constructor.
By default `df=false` and we have the current behavior.
However, if `df=true` then:
* normally a `echcol(df)` would be passed to the function (so it would be have the same way as normal `AsTable` but without compilation cost)
* with `ByRow` we would go through `eachrow(df)` so again - the same way as we normally do with `Tables.namedtupleiterator` but without compilation

This is orthogonal to this PR, but would allow to avoid `AsVector`. How would it play with DataFramesMeta.jl?

pdeffebach:
Thank you for the response. 

I don't disagree that we need fast paths internally that never construct the object of interest. But `AsTable(:)` is a bad thing to do with 10k columns in any context. I do worry that you might see people thinking that it's fast because `AsTable(:) => sum` is fast. It seems like there are two layers of magic here when we could only have one. 

I was not aware that `eachcol` supported `getpropery`! That does make it the natural object to use. 

As for DataFramesMeta, I am thinking something like 

```
@transform df :sum_columns = sum($AsTable(:))
```

and you are only allowed one `$AsTable(:)` call in the expression, since you can't have `[AsTable(:), :y] => ...`. 

This would get unwrapped to `AsTable(:) => f => :sum_column`. The same would work for `AsVector`.

bkamins:
What I was essentially proposing is to have:
* `AsTable(:)` => producing named tuples
* `AsTable(:, df=true)` => producing type unstable `eachcol`

The reason is that `AsTable(:, df=true)` still produces a table (i.e. you can index into it using column names not only column number). But I am OK to use another name for `AsTable(:, df=true)` if we prefer, but `AsVector` seems not the best as it is too closely linked to the promise of a vector (and drops names - in case they were useful). So maybe `AsColumns` would be better if you do not like `AsTable(:, df=true)`.

@nalimilan - what do you think?

pdeffebach:
I like `AsColumns`.

src/abstractdataframe/selection.jl
nalimilan: Sorry, I always forget that `collect(x)` isn't always identical to `map(identity, x)`.

nalimilan:
`AsColumns` doesn't sound super explicit. Things like `transform(df, [:x, :y] => f)` also pass columns to `f`. That would have to be called `AsEachCol`... I kind of like the idea of making `AsTable` more flexible, though I'd be tempted to pass a function (e.g. `eachol` vs `Tables.columntable`) or a type so that we can extend this later if needed.

That said, I'm not sure we need to allow users to do this, at least maybe not before having tested this PR for a while. Using optimizations under the hood sounds like a good idea, even if I admit that it could be a bit surprising. But custom reductions (that we wouldn't be able to optimize automatically) are quite rare AFAICT. In grouped reductions, we haven't got any request to add support for reductions beyond the standard ones.

bkamins:
> But custom reductions (that we wouldn't be able to optimize automatically) are quite rare

This was my original thinking in this PR. @pdeffebach - do you agree to the proposal by @nalimilan that we leave it out for now and implement custom reductions (giving an option to easily extend the mechanism)?

pdeffebach:
I think we can move forward with this PR, yeah, and leave it under the hood.

To summarize my line of thinking, this PR does two things

1. Avoids construction of a large named-tuple, which is costly
2. Uses a faster algorithm for calculating the `sum` by row. 

Option 2. will always have to be some sort of under the hood magic. But option 1 is something we can support via a new input type.

bkamins:
> Option 2. will always have to be some sort of under the hood magic. But option 1 is something we can support via a new input type.

Yes, but - as @nalimilan - we can leave option 1 in general for later and now focus on option 2. I will add other reductions to this PR soon (when I process all other GitHub notifications I am getting 😄).

src/abstractdataframe/selectionfast.jl
```diff
@@ -0,0 +1,152 @@
...
+        end
+    end
+    # this will happen if eltype of some columns do not support zero
+    # or all columns have eltype Missing
+    sumz_undefined && return nothing
+    sumv = fill!(Tables.allocatecolumn(typeof(sumz / 0), length(cols[1])), sumz)
+    lenv = zeros(Int, length(sumv))
+
+    for col in cols
+        sumv .= ifelse.(ismissing.(col), sumv, sumv .+ col)
```
bkamins, line 147: @nalimilan - do you think we need to do `Statistics._mean_promote` here? I think it should be OK the way we have it now, but these things are always tricky.

nalimilan: AFAICT it's not needed when the eltype is concrete, but it is when that's not the case. Note that `zero(Real) === 0` so that can happen even when `zero` is defined. Also, to match `mean`, accumulation has to be done using the return eltype, i.e. above do `zi = zero(eltype(col))/0`.

src/abstractdataframe/selectionfast.jl
```diff
@@ -0,0 +1,180 @@
...
+        lenv .+= .!ismissing.(col)
+    end
+    sumv ./= lenv
+    return sumv
+end
+
+@noinline table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(minimum))) =
```
nalimilan, line 154: Do you also plan to support `minimum∘skipmissing`?

src/abstractdataframe/selectionfast.jl
```diff
@@ -0,0 +1,180 @@
...
+    end
+end
+
+@noinline function table_transformation(df_sel::AbstractDataFrame, ::ByRow{typeof(length)})
+    @assert ncol(df) > 0
+    fill(ncol(df_sel), nrow(df_sel))
+end
+
+@noinline table_transformation(df_sel::AbstractDataFrame,
+                               ::ByRow{typeof(length∘skipmissing)}) =
```
nalimilan, line 88: By the way, would it make sense to support `first∘skipmissing`, `last∘skipmissing` and `coalesce`?

src/abstractdataframe/selectionfast.jl
```diff
@@ -0,0 +1,180 @@
...
+    local sumz
+    sumz_undefined = true
+    for col in cols
+        try
+            zi = zero(eltype(col))
+            if sumz_undefined
+                sumz_undefined = false
+                sumz = zi
+            elseif !ismissing(zi) # zi is missing if eltype is Missing
+                sumz += zi
```
nalimilan, line 61: To ensure small integers are promoted to `Int` (worth testing):
```suggestion
                sumz = Base.add_sum(sumz, zi)
```

Above we need something like `zi = Base.add_sum(zero(eltype(col)), zero(eltype(col)))` too.

src/abstractdataframe/selectionfast.jl
```diff
@@ -0,0 +1,180 @@
...
+        end
+    end
+    # this will happen if eltype of some columns do not support zero
+    # or all columns have eltype Missing
+    sumz_undefined && return nothing
+    init = fill!(Tables.allocatecolumn(typeof(sumz), length(cols[1])), sumz)
+
+    return foldl(cols, init=init) do l, r
+        l .= ifelse.(ismissing.(r), l, l .+ r)
+    end
```
nalimilan, line 79: Is `foldl` really better than a `for` loop here? Anyway, better use the same approach here and in `mean` below.

src/abstractdataframe/selectionfast.jl
```diff
@@ -0,0 +1,180 @@
...
+end
+
+@noinline function table_transformation(df_sel::AbstractDataFrame, ::ByRow{typeof(length)})
+    @assert ncol(df) > 0
+    fill(ncol(df_sel), nrow(df_sel))
```
nalimilan, line 84:
```suggestion
    return fill(ncol(df_sel), nrow(df_sel))
```

src/abstractdataframe/selection.jl
```diff
@@ -392,7 +392,15 @@ function _transformation_helper(df::AbstractDataFrame, col_idx::AbstractVector{I
...
-        return fun(map(c -> cdf[c], col_idx)...)
+        cols = map(c -> cdf[c], col_idx)
+        if (fun === +) || fun === ByRow(+)
+            return _sum_fast(cols)
+        elseif fun === ByRow(min)
+            return _minmax_row_fast(cols, min)
+        elseif fun === ByRow(max)
+            return _minmax_row_fast(cols, max)
+        end
+        return fun(cols...)
```
nalimilan, line 403: Shouldn't other reductions be checked here too?

Also:
```suggestion
        else
            return fun(cols...)
        end
```

src/abstractdataframe/selection.jl
```diff
@@ -376,27 +376,71 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+    isconcretetype(nonmissingtype(eltype(slowsum))) && return slowsum
+    T = mapreduce(typeof, promote_type, slowsum)
+    return convert(AbstractVector{T}, slowsum)
+end
+
 function _sum_skipmissing_fast(cols::Vector{<:AbstractVector})
     isempty(cols) && throw(ArgumentError("No columns selected for reduction"))
-    sumz = nothing
+    local sumz
+    sumz_undefined = true
```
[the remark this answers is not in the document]
bkamins: because if an error is thrown in `zero` I cannot "undefine" `sumz` in the `catch` part of the code.

src/abstractdataframe/selection.jl
```diff
@@ -376,27 +376,71 @@ _transformation_helper(df::AbstractDataFrame, col_idx::Int, (fun,)::Ref{Any}) =
...
+`AsTable(...)` selector.
+"""
+default_table_transformation(df_sel::AbstractDataFrame, fun) =
+    fun(Tables.columntable(df_sel))
+
+@noinline table_transformation(df_sel::AbstractDataFrame, ::typeof(sum)) =
+    _sum_fast(map(identity, eachcol(df_sel)))
+
+@noinline table_transformation(df_sel::AbstractDataFrame, ::ByRow{typeof(sum)}) =
+    table_transformation(df_sel, sum)
```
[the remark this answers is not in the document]
bkamins: Right - it was from earlier versions. I will add it to `default_table_transformation` just for safety that we do not trigger compilation there.

src/abstractdataframe/selectionfast.jl
bkamins: Yes - this PR is WIP. I will let you know when I am done with all the reductions I plan to add.

src/abstractdataframe/selectionfast.jl
bkamins: Yes - I planned `first` and `last` variants.

For `coalesce` - I have not planned it. As it takes positional arguments we are probably bound by the implementation of `coalesce` in Julia Base anyway. Do you think it would be useful?

src/abstractdataframe/selectionfast.jl
bkamins: I will change to loop

src/abstractdataframe/selection.jl
bkamins: What other reductions do you have in mind? Only `+`, `min` and `max` make sense for positional arguments. I will add `*` when I implement it.

bkamins: Actually given your comment for `Base.add_sum` above, maybe we should drop the fast path for positional arguments altogether? Given e.g.:
```
julia> sum([0x1, 0x1])
0x0000000000000002

julia> 0x1+0x1
0x02
```

nalimilan: Right, so there's only `*` left.

For `ByRow(+)`, we should indeed use `reduce(+, ...)` rather than `sum(...)`. So that's different from the fast path for `sum`.

src/abstractdataframe/selectionfast.jl
nalimilan: I imagine it can be used as `ByRow(coalesce)`. Isn't that the same kind of function as `min` or `max`?

bkamins:
@nalimilan - I went through current implementation and tried to make it correct.

src/abstractdataframe/selectionfast.jl
```diff
@@ -0,0 +1,243 @@
...
+
+function table_transformation(df_sel::AbstractDataFrame, ::typeof(minimum))
+    return reduce(min, map(identity, eachcol(df_sel)))
+end
+
+function table_transformation(df_sel::AbstractDataFrame, ::typeof(maximum))
+    return reduce(max, map(identity, eachcol(df_sel)))
+end
+
+# TODO:
```
bkamins, line 227: @nalimilan - I have made a list of additional standard reductions that might be asked for. We can add all of them (the list is long, or maybe we split this PR into two and concentrate only on some of them here and I would add other in the follow up - this maybe would be easier to review?)

bkamins:
@nalimilan - this should be good for a review now. I have implemented half of the planned reductions that are most common. I leave the second half for the follow up PR not to make it too big for the review.

bkamins:
CI fails on specific Julia versions because of Julia Base implementation differences - I will look into it (unfortunately this was expected 😞).

The documentation failure is unrelated.

bkamins:
The first type of problem is on x86 archictectures where `Int32` overflow happens and we using `Float64` promotion produced a correct result instead of an incorrect one done by standard `sum` due to overflow.

bkamins:
The second problem is incorrect calculation of `mean` on Julia 1.0 because of not doing `eltype` promotion during accumulation phase:
```
```

PR #2877, opened 2021-09-13
let argument cols in names() also works on contents rather than just names

### LINKED ISSUE #2747: Allow selecting columns based on predicate on column contents
Based on a discussion from slack:
> [Sebastian Rollén]: Potentially dumb question: How might I `select` only columns that don't have missing data in them? In R/dplyr I would use select_if, so I was looking for something similar
> 
> [Sebastian Rollén]: my best approach so far is to choose the column indices with `indices = map(x -> !any(ismissing, x), eachcol(df))` and then `df[:, indices]` , but I feel like there's probably a nicer solution?
>
> [Peter Deffebach]: you can use a type in `names`, something like `names(df, <:Missing)`. But this is all columns that allow missings, not the ones with missings.
>
> [Sebastian Rollén]: hm, that probably won't work in this case as all my columns allow missings as they are `unstack`ed from the same column
>
> [Sebastian Rollén]: I'm explicitly looking for the columns that don't have any missing values so I can filter out the other ones and then run `disallowmissing` so that I can get rid of the type union
>
> [Bogumił Kamiński]: What you use now is a currently recommended way. Currently in `names` if you pass predicate it looks at column name not at column contents. We have decided for this as it seemed more common. But agreed it is also useful to find columns based on a predicate of their contents. Maybe open an issue and we can discuss there how to design it.

It's currently possible to `select` a dataframe using a predicate in `names`, but this predicate applies to the types of the columns, not the contents of the columns themselves. The idea here is to have some function that allows `select`ing a column based on the contents of the columns instead, similar to `select_if` in dplyr

xluo127:
Trying to fix #2747. 

When a predicate function is applied, by setting another argument `content`  from `false` to `true`, it works for contents rather than names.

```
julia> df1 = DataFrame([missing 1 2 3; missing 2 missing 4; 1 4 2 5], :auto)
3×4 DataFrame
 Row │ x1       x2      x3       x4
     │ Int64?   Int64?  Int64?   Int64?
─────┼──────────────────────────────────
   1 │ missing       1        2       3
   2 │ missing       2  missing       4
   3 │       1       4        2       5

julia> names(df1, (function(x) x isa String end))
4-element Vector{String}:
 "x1"
 "x2"
 "x3"
 "x4"

julia> names(df1, (function(x) x isa String end), true)
String[]

julia> names(df1, (function(col) mean(skipmissing(col)) >= 4 end), true)
1-element Vector{String}:
 "x4"
```

This additional argument `content` seems not that good, but right now I do not have a better idea.

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -88,7 +88,8 @@ end
 
 Base.names(df::AbstractDataFrame, T::Type) =
     [String(n) for (n, c) in pairs(eachcol(df)) if eltype(c) <: T]
-Base.names(df::AbstractDataFrame, fun::Function) = filter!(fun, names(df))
+#Base.names(df::AbstractDataFrame, fun::Function) = filter!(fun, names(df))
+Base.names(df::AbstractDataFrame, fun::Function, predicate = 0) = predicate == 0 ? filter!(fun, names(df)) : names(df, fun.(eachcol(df)))
 
 # _names returns Vector{Symbol} without copying
 _names(df::AbstractDataFrame) = _names(index(df))
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -62,7 +62,7 @@ abstract type AbstractDataFrame end
 
 """
     names(df::AbstractDataFrame)
-    names(df::AbstractDataFrame, cols)
+    names(df::AbstractDataFrame, cols, content::Bool = false)
 
 Return a freshly allocated `Vector{String}` of names of columns contained in `df`.
 
@@ -74,6 +74,8 @@ selector (this is useful in particular with regular expressions, `Cols`, `Not`, 
   are returned
 * a `Function` predicate taking the column name as a string and returning `true`
   for columns that should be kept
+
+`content` is only specified to `true` when a predicate `Function` is applied on columns rather than names of them.
 
 See also [`propertynames`](@ref) which returns a `Vector{Symbol}`.
 """
@@ -89,7 +91,7 @@ end
 Base.names(df::AbstractDataFrame, T::Type) =
     [String(n) for (n, c) in pairs(eachcol(df)) if eltype(c) <: T]
 #Base.names(df::AbstractDataFrame, fun::Function) = filter!(fun, names(df))
-Base.names(df::AbstractDataFrame, fun::Function, predicate = 0) = predicate == 0 ? filter!(fun, names(df)) : names(df, fun.(eachcol(df)))
+Base.names(df::AbstractDataFrame, fun::Function, content::Bool = false) = content == false ? filter!(fun, names(df)) : names(df, fun.(eachcol(df)))
 
 # _names returns Vector{Symbol} without copying
 _names(df::AbstractDataFrame) = _names(index(df))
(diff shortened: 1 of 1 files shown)


bkamins:
Yes - this is a discussion we had some time ago and concluded that filtering by column name is more useful.
However, indeed - we can have both. We just need to discuss the syntax.
The original thinking was that
```
names(df, fun.(eachcol(df))
```
is already quite short so adding special syntax did not seem to be that useful, e.g.:
```
names(df, fun, content=true)
```
is longer to type and requires learning what it means (as opposed to `names(df, fun.(eachcol(df))` which is immediately clear for anyone who knows Julia Base).

CC @pdeffebach @nalimilan for opinions.

If we agree to add this change can you please add tests and NEWS.md entry later. Thank you!

nalimilan:
Yes, I'm still reserved about adding this given the syntax that @bkamins showed is already reasonable.

bkamins:
Maybe the solution (instead of adding a functionality) is to add a documentation that explains clearly what should be used.
In particular adding documentation for `names` for `GroupedDataFrame`, `DataFrameRows`, `GroupKey`, `DataFrameColumns`, and `DataFrameRow` should be made (I think it is a good "introductory level" PR to make).

Also there is an issue with e.g. `GroupedDataFrame` and `DataFrameRow`, if you would write `names(gdf, predicate, content=true)` it would be a bit confusing that you are using the whole columns of `parent(gdf)` and not e.g. groups of `gdf`. Simiar for `DataFrameRow`.

xluo127:
Thanks for your comments. Is there anything I can do to help, like adding a documentation?

bkamins:
Yes. If you are willing to help then I recommend you to:
* improve docstring for `names` withe a reference to `GroupedDataFrame`, `DataFrameRows`, `GroupKey`, `DataFrameColumns`, and `DataFrameRow` (you need to carefully check which types support which arguments for `names`)
* add an example with `names(df, fun.(eachcol(df))` to an examples section of docstring
* check if some more examples should not be added in the documentation; in /docs/src/man/getting_started.md probably is a best place

If any of these points are not clear to you please ask and I can guide you. Thank you!

xluo127:
I see, thanks! I am going to work on these points.

bkamins:
@xluo127 - on StackOverflow in https://stackoverflow.com/questions/69178300/find-a-subset-of-columns-of-a-data-frame-that-have-some-missing-values I have posted some examples how `eachcol` typically can be used to select columns based on their contents. Maybe you will find it useful when writing your PR.

xluo127:
OK, thank you.

xluo127:
@bkamins I think src/abstractdataframe/abstractdataframe.jl should be the place where to improve docstring for `names`, but yet I see no example section there. Do you think is a good idea to add some examples for some legal types, or do I just have to add an example for `names(df, fun.(eachcol(df))`? 

And for the third point that you gave, do you mean that putting some straightforward examples into /docs/src/man/getting_started.md rather than adding more examples in the documentation will be more suitable? 

By the way, are there more steps I have to follow before opening a new PR? Of course, I will check my work carefully. But if I miss something, opening a bad PR may increase your workflow.

bkamins:
1. In https://github.com/JuliaData/DataFrames.jl/blob/527fa8e9c5cebf33471263256bdb22e107d7a163/src/abstractdataframe/abstractdataframe.jl#L64 add additional accepted signatures
2. Then below add `# Examples` section like in other docstings. Add different examples that you think are good to show.
3. In https://raw.githubusercontent.com/JuliaData/DataFrames.jl/main/docs/src/man/getting_started.md you have `Column names can be obtained as strings using the `names` function:` part, where probably it would be also good to add some more examples (as now we only give `names(df)` example).
4. Opening the PR - just please open it (or push changes to this PR reverting the current commit) and I will guide you.

Thank you1

bkamins:
Closing in favor of #2882

PR #2878, opened 2021-09-13
Allow constructing Matrix from empty dataframe

jakobnissen:
Before:
```
julia> Matrix(DataFrame())
[ stacktrace omitted ]
```

After:
```
julia> Matrix(DataFrame())
0×0 Matrix{Union{}}
```

I chose an eltype of `Union{}` because the element type cannot be determined without any columns, and because it's the natural generalization of the type promotion between the columns.

### DIFF
--- a/test/conversions.jl
+++ b/test/conversions.jl
@@ -6,6 +6,7 @@ const ≅ = isequal
 
 @testset "Constructors to Base types" begin
     df = DataFrame()
+    @test isa(Matrix(df), Matrix)
     df[!, :A] = 1:5
     df[!, :B] = 1.0:5.0
     @test isa(Matrix(df), Matrix{Float64})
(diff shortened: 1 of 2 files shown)


test/conversions.jl
```diff
@@ -6,6 +6,7 @@ const ≅ = isequal
 
 @testset "Constructors to Base types" begin
     df = DataFrame()
+    @test isa(Matrix(df), Matrix)
```
bkamins, line 9:
```suggestion
    @test Matrix(df) isa Matrix{Union{}}
    @test size(Matrix(df)) == (0, 0)
```

bkamins:
Looks good. I will merge it once CI passes (I have updated the test a bit). Thank you!

bkamins (approved):

bkamins:
@nalimilan - OK to merge?

Issue #2875, opened 2021-09-12
Make Cols more flexible


bkamins:
Possible useful additions:
* `Cols(<:Type)`
* `Cols(<:Function)`

that would match how the `names` function works.

@nalimilan - the only limitation is that currently `AbstractIndex` is not aware of data frame columns, and we handle `Cols` on `AbstractIndex` level. I will have to think how to handle this cleanly (as the simplest solution is to add special methods on `AbstractDataFrame` level, but it is not nice as then some of the methods for handling columns would be defined at index level, and some at data frame level).

bkamins:
Adding `Cols(<:Type)` this would require a significant redesign of the package. We would have to pass around `AbstractDataFrame` or `DataFrameRow` in places were we currently pass around `AbstactIndex` (and there are many such places). Also with `DataFrameRow` is would be especially tricky to add.

I would constrain ourselves for now to `Cols(<:Function)` which should be easier.

PR #2881, opened 2021-09-13
Allow predicate in Cols

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2875

I was thinking that we might make it even more flexible and allow predicate to be mixed with other selectors in `Cols`, but for now I proposed a restrictive version. @nalimilan - do you think we should be more flexible?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -70,6 +70,8 @@
 * the `DataFrame` constructor when matrix is passed to it as a first
   argument now allows `copycols` keyword argument
   ([#2829](https://github.com/JuliaData/DataFrames.jl/pull/2859))
+* `Cols` now accepts a predicate accepting column names as strings.
+  ([#2880](https://github.com/JuliaData/DataFrames.jl/pull/2880))
 
 ## Bug fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -71,7 +71,7 @@
   argument now allows `copycols` keyword argument
   ([#2859](https://github.com/JuliaData/DataFrames.jl/pull/2859))
 * `Cols` now accepts a predicate accepting column names as strings.
-  ([#2880](https://github.com/JuliaData/DataFrames.jl/pull/2880))
+  ([#2881](https://github.com/JuliaData/DataFrames.jl/pull/2881))
 
 ## Bug fixes
 
--- a/docs/src/lib/indexing.md
+++ b/docs/src/lib/indexing.md
@@ -27,16 +27,16 @@ The rules for a valid type of index into a column are the following:
     * a regular expression, which gets expanded to a vector of matching column names;
     * a `Not` expression (see [InvertedIndices.jl](https://github.com/mbauman/InvertedIndices.jl));
       the `Not(idx)` selects all indices not in the passed `idx`;
-    * an `Cols` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
-      the `Cols(idxs...)` selects the union of the selections in `idxs`; in particular
+    * a `Cols` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
+      `Cols(idxs...)` selects the union of the selections in `idxs`; in particular
       `Cols()` selects no columns and `Cols(:)` selects all columns; a special rule is
-      `Cols(predicate)`, where `precicate` is a predicate function; in this case
-      the columns whose names passed to the `predicate` predicate as strings return `true`
+      `Cols(predicate)`, where `predicate` is a predicate function; in this case
+      the columns whose names passed to `predicate` as strings return `true`
       are selected.
-    * `Between` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
-       the `Between(first, last)` selects the columns between `first` and `last`;
-    * `All` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
-      the `All()` selects all columns, equivalent to `:`;
+    * a `Between` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
+       `Between(first, last)` selects the columns between `first` and `last`;
+    * an `All` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
+      `All()` selects all columns, equivalent to `:`;
     * a colon literal `:`.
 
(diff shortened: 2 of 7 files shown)


src/other/index.jl
```diff
@@ -221,6 +221,17 @@ end
...
+@inline Base.getindex(x::AbstractIndex, idx::Cols{<:Tuple{Function}}) =
+    findall(idx.cols[1], names(x))
+
+"""
+    Cols(f::Function)
+
+Select the columns whose names passed to the `f` predicate as strings return `true`.
+As a special case if `:` is passed (the `Colon` function) select all columns.
+"""
+Cols
```
nalimilan, line 234: Shouldn't this be documented in the `select`, etc. docstrings?

nalimilan:
Being restrictive sounds OK for now. Why does `:` need a special support? I would have thought it could be treated like any selector that can be included in `Cols`?

bkamins:
> Why does `:` need a special support?

Ah - you are right. `:` is not a function but a constructor so it is not a subtype of `Function` so special treatment is not needed. I will fix this.

src/other/index.jl
bkamins: I think we should add this docstring. Apart from this all special selectors now are only mentioned in https://github.com/JuliaData/DataFrames.jl/blob/main/docs/src/lib/indexing.md. We should decide on a good place to document them in general, especially when we will start allowing broadcasting them. Do you see any good place?

bkamins:
Ah - I was right originally `Colon` is a constructor but it is defined as:
```
struct Colon <: Function
end
```
so it is a `Function`. My original definition thus was required as otherwise `Colon(:)` treats `:` as a predicate. Things are tough.

src/other/index.jl
bkamins: @nalimilan I tried to improve the documentation. The PR should be good to re-review now.

docs/src/lib/indexing.md
```diff
@@ -26,7 +26,17 @@ The rules for a valid type of index into a column are the following:
...
+    * an `Cols` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
+      the `Cols(idxs...)` selects the union of the selections in `idxs`; in particular
+      `Cols()` selects no columns and `Cols(:)` selects all columns; a special rule is
+      `Cols(predicate)`, where `precicate` is a predicate function; in this case
+      the columns whose names passed to the `predicate` predicate as strings return `true`
+      are selected.
+    * `Between` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
+       the `Between(first, last)` selects the columns between `first` and `last`;
+    * `All` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
+      the `All()` selects all columns, equivalent to `:`;
```
nalimilan, line 39:
```suggestion
      the `Not(idx)` selects all indices not in the passed `idx`;
    * a `Cols` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
      `Cols(idxs...)` selects the union of the selections in `idxs`; in particular
      `Cols()` selects no columns and `Cols(:)` selects all columns; a special rule is
      `Cols(predicate)`, where `predicate` is a predicate function; in this case
      the columns whose names passed to `predicate` as strings return `true`
      are selected.
    * a `Between` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
       `Between(first, last)` selects the columns between `first` and `last`;
    * an `All` expression (see [DataAPI.jl](https://github.com/JuliaData/DataAPI.jl));
      `All()` selects all columns, equivalent to `:`;
```

docs/src/man/working_with_dataframes.md
```diff
@@ -255,18 +255,49 @@ julia> df[!, Not(:x1)]
...
+
+julia> df[:, Cols(x -> startswith(x, "x"))] # keep columns whose name starts with "x"
+1×2 DataFrame
+ Row │ x1     x2
+     │ Int64  Int64
+─────┼──────────────
+   1 │     2      3
+```
+
+The following examples show a more complex use of `Cols` selector and move all
```
nalimilan, line 297:
```suggestion
The following examples show a more complex use of the `Cols` selector, which moves all
```

docs/src/man/working_with_dataframes.md
```diff
@@ -255,18 +255,49 @@ julia> df[!, Not(:x1)]
...
+columns whose names match `r"x"` regular expression respectively to the front
+and to the end of a data frame:
```
nalimilan, line 299:
```suggestion
and to the end of the data frame:
```

src/other/index.jl
nalimilan: We really need page(s) about indexing and `select`/`transform`/... which would be more detailed than Getting Started, but less technical that the API reference.

It's a bit weird that we add a docstring to `Cols` that only makes sense for DataFrames. Maybe we should document this in DataAPI?

bkamins: I have opened https://github.com/JuliaData/DataAPI.jl/pull/42 (and then we can remove the docstring here)
and https://github.com/JuliaData/DataFrames.jl/issues/2887 (to handle this in the future)

docs/src/lib/internals.md
```diff
@@ -16,4 +16,5 @@ getmaxwidths
 ourshow
 ourstrwidth
 @spawn_for_chunks
+Cols
```
bkamins, line 19:
```suggestion
```

src/other/index.jl
bkamins, line 234:
```suggestion
```

NEWS.md
```diff
@@ -70,6 +70,8 @@
 * the `DataFrame` constructor when matrix is passed to it as a first
   argument now allows `copycols` keyword argument
   ([#2859](https://github.com/JuliaData/DataFrames.jl/pull/2859))
+* `Cols` now accepts a predicate accepting column names as strings.
+  ([#2880](https://github.com/JuliaData/DataFrames.jl/pull/2880))
```
bkamins, line 74:
```suggestion
  ([#2881](https://github.com/JuliaData/DataFrames.jl/pull/2881))
```

bkamins:
@nalimilan - this should be good to have a look at again.

nalimilan (approved):

bkamins:
Thank you!

PR #2892, opened 2021-09-30
Update show.jl

Moelf:
untangle double negation


Moelf:
I realized I can't just `cancel` them because they are just default, but this is pretty confusing to keep track of

PR #2894, opened 2021-10-05
Specify why leftjoin! needs at most one match

rikhuijzer:
I did read the documentation of the `leftjoin!`, but only spotted the caveat about the one match on `df2` after @bkamins pointed it out.

Therefore, I propose to change the wording a bit and explain to the reader *why* this assumption is necessary. Knowing the why should make it easier for people to remember the caveat. I hope that I guessed the reason right.

### DIFF
--- a/src/join/inplace.jl
+++ b/src/join/inplace.jl
@@ -6,9 +6,10 @@
 Perform a left join of two data frame objects by updating the `df1` with the
 joined columns from `df2`.
 
-A left join includes all rows from `df1`.
-Rows and columns from `df1` are left untouched.
-Each row in `df1` must have at most one match in `df2` based on `on` columns.
+A left join includes all rows from `df1` and leaves all rows and columns from `df1`
+untouched. Note that each row in `df1` must have at most one match in `df2`. Otherwise,
+this function would not be able to execute the join in-place since new rows need to be
+added to `df1`.
 
 # Arguments
 - `df1`, `df2`: the `AbstractDataFrames` to be joined


src/join/inplace.jl
```diff
@@ -6,9 +6,10 @@
 Perform a left join of two data frame objects by updating the `df1` with the
 joined columns from `df2`.
 
-A left join includes all rows from `df1`.
-Rows and columns from `df1` are left untouched.
-Each row in `df1` must have at most one match in `df2` based on `on` columns.
+A left join includes all rows from `df1` and leaves all rows and columns from `df1`
+untouched. Note that each row in `df1` must have at most one match in `df2`. Otherwise,
+this function would not be able to execute the join in-place since new rows need to be
+added to `df1`.
```
bkamins, line 12: yes - exactly. We either fill the row with the only matching row from `df2` or fill it with `missing` if we do not get a match.
The point is - as you have commented - that `df1` part must stay untouched.

bkamins (approved):

src/join/inplace.jl
rikhuijzer: > mentally it is easier for the user to have added new columns at the end of the data frame that is left untouched

Wow yes, agreed. For me, it is very hard to wrap my head around the idea that everything in the right table doesn't move in memory (non allocating) and that things magically appear at the left table (via pointers if I'm not mistaken)

Thank you for the extra explanation. It's nice to learn the thought behind the decisions

bkamins: This is exactly the point - with `leftjoin!` if you ask any user what should happen it is pretty obvious and an intuitive answer, even of the beginner, is the same as the correct answer.
For `rightjoin!` there are several options how it could work so it could lead to confusion.

Issue #2876, opened 2021-09-12
In src => fun => dst allow transformation function in dst


bkamins:
`dst` could take input column name and transform it.

Issue #2893, opened 2021-10-04
renamecols function for transform


jtrakk:
I would like to pass a `renamecols` function to `transform`. `innerjoin` takes a `renamecols` function and `transform` takes a `renamecols::Bool`. I want to replace the Bool with a function parameter that defaults to `column_function(c) = "$(c)_function"` and can be replaced with `identity` for users who currently pass `renamecols=false`.

bkamins:
This is a good proposal in line with some other proposals to allow `dst` to be a function in `src => fun => dst`, see https://github.com/JuliaData/DataFrames.jl/issues/2876.
We will just need to work out the best approach.

jtrakk:
Your proposal is better than mine. Closing.

PR #2897, opened 2021-10-07
allow transformation destination to be a function

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2876

The crucial API decision wis what function should take. I assume it is passed a string or a vector of strings.
The vector is passed if multicolumn selector is used (and `AsTable` is always considered to be multicolumn selector in particular)

### DIFF
--- a/docs/src/man/split_apply_combine.md
+++ b/docs/src/man/split_apply_combine.md
@@ -51,11 +51,11 @@ each subset of the `DataFrame`. This specification can be of the following forms
    `function` returns a single value or a vector; the generated name is created by
    concatenating source column name and `function` name by default (see examples below).
 3. a `cols => function => target_cols` form additionally explicitly specifying
-   the target column or columns, which must be single name (as a `Symbol` or a string),
+   the target column or columns, which must be a single name (as a `Symbol` or a string),
    a vector of names or `AsTable`. Additionally it can be a `Function` which
    takes a string or a vector of strings as an argument containing names of columns
-   selected by `cols`, in which case `target_cols` becomes a return value of this function
-   (all accepted types of values except `AsTable` are allowed).
+   selected by `cols`, and returns the target columns names (all accepted types
+   except `AsTable` are allowed).
 4. a `col => target_cols` pair, which renames the column `col` to `target_cols`, which
    must be single name (as a `Symbol` or a string), a vector of names or `AsTable`.
 5. a `nrow` or `nrow => target_cols` form which efficiently computes the number of rows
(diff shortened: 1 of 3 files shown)


bkamins:
> I just wonder whether we should also pass the function, as it could be useful to generate a column name e.g. in a loop over multiple functions and/or inputs.

Could you please give an example or explain it a bit more as I am not sure what you mean here.

> Though that would make the API less convenient. What kind of use case do we expect?

You mean use cases of the change I have implemented?

The most common will be `cols .=> fun .=> identity` to avoid changing column names I think.

nalimilan:
> Could you please give an example or explain it a bit more as I am not sure what you mean here.

I mean something like:
```julia
julia> combine(df, [:x, :y, :z] .=> [mean median] =.> string)
4×6 DataFrame
 Row │ meanx     meany    meanz     medianx   mediany   medianz 
     │ Float64   Float64  Float64   Float64   Float64   Float64  
─────┼───────────────────────────────────────────────────────────
   1 │ 0.483742   0.4106  0.287058   0.47259   0.32765  0.247748
```

Given that it's almost the same as the default names maybe that's not useful. It *could* be useful when there are many input variable names but you would like to use only one of them in the output rather than the "etc" default. That's not super likely but i thought I'd mention it just to be sure.

bkamins:
> That's not super likely but i thought I'd mention it just to be sure.

Ah - now I understand. I think it would complicate the API too much.

> So that's `renamecols=false`, right? I hope we have other use cases or it doesn't seem super appealing. :-)

Yes - `identity` or `string` is just `renamecols=false` but it is shorter to type. You asked me about a most common use case.

Recently @jtrakk asked for this in https://github.com/JuliaData/DataFrames.jl/issues/2893 so maybe we can get some more comment when it would be useful.

jtrakk:
My case was like this:

```jl
transform(df, [:X,:Y,:Z] .=> (v->v.-mean(v)) => (s->"$(s)demeaned"))
```

The case of "for each variable, for each function"

```jl
combine(df, [:x, :y, :z] .=> [mean median] .=> string)
```
is an interesting one and does suggest taking the function as well, but I agree it would complicate the API. Perhaps there could be a wrapper

```jl
combine(df, [:x, :y, :z] .=> [mean median] .=> Rename((s,f)->"$s$f"))
```

and passing only a function like `lowercase` could be a shortcut for `Rename((s,f)->lowercase(s))`, but that seems worth postponing until someone needs it.

bkamins:
Thank you!

Essentially as @jtrakk shows the use case is when the function name does not describe well the operation you perform and you prefer to give an explicit name for the transformation. This is especially useful when transformations are anonymous functions.

Issue #2901, opened 2021-10-08
Strange behaviour with non-ASCII column names


wentasah:
When a DataFrame contains non-ASCII column name like here:
```julia
d = DataFrame("power_µW" => [1,2,3])
```
access to the column leads to an "interesting" error:
```julia
d.power_µW
# ERROR: ArgumentError: column name :power_μW not found in the data frame; 
# existing most similar names are: :power_µW
```

This is weird, because typing `d.` and pressing `TAB` completes exactly the above command.

When using the following syntax, everything works as expected:
```julia
d."power_µW"
# 3-element Vector{Int64}:
#  1
#  2
#  3
```

Happens with DataFrames 1.2.2 as well as with [a93c6f00] DataFrames v1.2.2 `https://github.com/JuliaData/DataFrames.jl.git#main`.
```
julia> versioninfo()
Julia Version 1.6.2
Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
```

bkamins:
This is a limitation in Julia design that opts-in to canonicalize Unicode identifiers. Here is an explanation https://github.com/JuliaLang/julia/issues/5434. CC @stevengj as he was the OP so that he is aware of this consequence.

A more verbose example showing the problem and how to work-around it:
```
julia> a, b = "xμ", "yµ"
("xμ", "yµ")

julia> collect(a)
2-element Vector{Char}:
 'x': ASCII/Unicode U+0078 (category Ll: Letter, lowercase)
 'μ': Unicode U+03BC (category Ll: Letter, lowercase)

julia> collect(b)
2-element Vector{Char}:
 'y': ASCII/Unicode U+0079 (category Ll: Letter, lowercase)
 'µ': Unicode U+00B5 (category Ll: Letter, lowercase)

julia> d = DataFrame(a => 1, b => 2)
1×2 DataFrame
 Row │ xμ     yµ    
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2

julia> d.xμ # autocompleted
1-element Vector{Int64}:
 1

julia> d.yµ # autocompleted
ERROR: ArgumentError: column name :yμ not found in the data frame; existing most similar names are: :yµ and :xμ
Stacktrace:
 [1] lookupname
   @ C:\Users\bogum\.julia\packages\DataFrames\vuMM8\src\other\index.jl:291 [inlined]
 [2] getindex
[...]

PR #2904, opened 2021-10-08
Try to detect unicode normalization issues in column names

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2901

### DIFF
--- a/src/other/index.jl
+++ b/src/other/index.jl
@@ -313,12 +313,6 @@ function _norm_eq(a1::String, a2::String, idx::Symbol)
             refc1 = get(_julia_charmap, c1, c1)
             refc2 = get(_julia_charmap, c2, c2)
             @assert refc1 == refc2
-            if VERSION >= v"1.8-"
-                suffix = "use the Unicode.normalize function setting its " *
-                         "chartransform keyword argument to Unicode.julia_chartransform."
-            else
-                suffix = "use only $refc1 (codepoint: $(UInt32(refc1))) in column names."
-            end
             throw(ArgumentError("column name :$idx not found in the " *
                                 "data frame. However there is a similar " *
                                 "column name in the data frame where character $c2 " *
@@ -326,7 +320,9 @@ function _norm_eq(a1::String, a2::String, idx::Symbol)
                                 "(codepoint: $(UInt32(c1))). The " *
                                 "error is most likely caused by the Julia parser which " *
                                 "normalizes `Symbol` literals containing such " *
-                                "characters. In order to avoid such problems " * suffix))
+                                "characters. If the passed name is intended " *
+                                "to match the column name found in the data frame " *
+                                "replace the non-matching character with the correct one."))
         end
     end
     throw(AssertionError("Unreachable reached. Please report issue to " *
@@ -347,7 +343,7 @@ function normalized_match_test(l::Dict{Symbol, Int}, idx::Symbol)
                     throw(ArgumentError("column name :$idx not found in the " *
                                         "data frame. However, after applying Unicode " *
                                         "normalization to it, the passed column name " *
-                                        "matches a column name found in the data frame." *
+                                        "matches a column name found in the data frame. " *
                                         "You can use the `Unicode.normalize` " *
                                         "function to normalize the passed column name."))
                 else
@@ -372,7 +368,7 @@ function normalized_match_test(l::Dict{Symbol, Int}, idx::Symbol)
                 throw(ArgumentError("column name :$idx not found in the " *
                                     "data frame. However, the passed column name " *
                                     "matches a column name found in the data frame " *
-                                    "after applying Unicode normalization to the latter." *
+                                    "after applying Unicode normalization to the latter. " *
                                     "It is recommended to use " *
                                     "normalized column names to avoid ambiguity. " *
                                     "In order to normalize column names in " *
--- a/src/other/index.jl
+++ b/src/other/index.jl
@@ -288,7 +288,7 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
         return [s for (d, s) in dist if d <= maxd]
 end
 
-TODO: revisit after JuliaLang/julia#42561 has been merged
+# TODO: revisit after JuliaLang/julia#42561 has been merged
 const _julia_charmap = Dict{Char,Char}(
     Char(0x025B) => Char(0x03B5),
     Char(0x00B5) => Char(0x03BC),
@@ -310,9 +310,9 @@ function _norm_eq(a1::String, a2::String, idx::Symbol)
             refc1 == refc2 || return false
             throw(ArgumentError("column name :$idx not found in the " *
                                 "data frame. However there is a similar " *
-                                "column name in the data frame where character $c1 " *
-                                "(codepoint: $(UInt32(c1))) is used is instead of $c2 " *
-                                "(codepoint: $(UInt32(c2))). The " *
+                                "column name in the data frame where character $c2 " *
+                                "(codepoint: $(UInt32(c2))) is used is instead of $c1 " *
+                                "(codepoint: $(UInt32(c1))). The " *
                                 "error is most likely caused by the Julia parser which " *
                                 "normalizes `Symbol` literals containing such " *
                                 "characters. In order to avoid such problems use only " *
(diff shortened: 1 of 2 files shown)


bkamins:
The only problem is that this fix does not resolve the OP issue as still:
```
julia> d = DataFrame("power_µW" => [1,2,3])
3×1 DataFrame   
 Row │ power_µW 
     │ Int64    
─────┼──────────
   1 │        1 
   2 │        2 
   3 │        3 

julia> d.power_µW
ERROR: ArgumentError: column name :power_μW not found in the data frame; existing most similar names are: :power_µW
```
as this kind of change is not covered by `Unicode.normalize`.

nalimilan:
What's annoying is that we don't give any concrete solution when the parser replaced some chars with others during normalization. We should probably wait until @stevengj's PR is merged.

stevengj:
If there is no ambiguity, I'm not quite clear on why you are throwing an error rather than giving the equivalent (modulo normalization) column?

nalimilan:
Also performance will be worse when we need to check all column names, so better signal the problem to the user than silently being slower.

stevengj:
> So that would mean doing first a lookup of raw names, and if that fails a lookup on normalized names?

Yes, but furthermore in `getproperty` I think you should only look up normalized names, and should assume that the Symbol argument has already been normalized by Julia.

stevengj:
@bkamins, it's already been decided that `Symbol("...")` constructs a programmatic symbol that may not be type-able by the user, and is not normalized.

bkamins:
> it's already been decided that `Symbol("...")` constructs a programmatic symbol that may not be type-able by the user, and is not normalized.

Yes. I understand this. What I do not understand is why you think that:

```
julia> nt = (; Symbol("noe\u0308l") => ["noel"])
(noël = ["noel"],)

julia> julia> nt.noël # autocompleted
ERROR: type NamedTuple has no field noël
```

is OK (as this is what I understand your comment implies).

While you think that this behavior should be changed:

```
julia> df = DataFrame(nt)
1×1 DataFrame  
 Row │ noël   
     │ String  
─────┼──────── 
   1 │ noel    

julia> df.noël
ERROR: ArgumentError: column name :noël not found in the data frame; existing most similar names are: :noël
```

Or putting it in different words why:

> you could still allow such names in the non-normalized lookup, but you would omit names with normalization conflicts from the normalized lookup table—hence only a normalized lookup of an ambiguous normalized name (e.g. via `getproperty`) would be an error.

should apply to `DataFrame`, but should not apply to `NamedTuple`?

The reason I am pressing for clarifying this (and avoid just implementing it) is that there is an intended duality between these two types. `NamedTuple` of vectors of equal length and `DataFrame` should be useable in exactly the same way with the exception that `NamedTuple` is type stable and `DataFrame` is not type stable. This e.g. has a consequence that internally in DataFrames.jl we construct such `NamedTuples` to improve performance of some operations via a function barrier and having such discrepancy would be problematic.

src/other/index.jl
```diff
@@ -290,9 +290,80 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
         return [s for (d, s) in dist if d <= maxd]
 end
 
+# TODO: revisit after JuliaLang/julia#42561 has been merged
```
bkamins, line 293:
```suggestion
# see JuliaLang/julia#42561; we keep our own list to ensure things work consistently on earlier versions of Julia
```

src/other/index.jl
```diff
@@ -288,9 +288,66 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+    Char(0x00B7) => Char(0x22C5),
+    Char(0x0387) => Char(0x22C5),
+    Char(0x2212) => Char(0x002D)
+)
+
+function _norm_eq(a1::String, a2::String, idx::Symbol)
+    s1 = iterate(a1)
+    s2 = iterate(a2)
+
+    while !(s1 === nothing || s2 === nothing)
```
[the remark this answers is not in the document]
bkamins: I have re-implemented it to provide a simpler logic.

src/other/index.jl
```diff
@@ -288,9 +288,80 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+                                "data frame. However there is a match of " *
+                                "Unicode normalized passed column name with " *
+                                "a normalized column name found in the " *
+                                "data frame. In the passed data $case not " *
+                                "normalized. It is recommended to use " *
+                                "normalized column names and then refer to them " *
+                                "using normalized names to avoid ambiguity. " *
+                                "In order to normalize column names in " *
+                                "an existing data frame `df` do " *
+                                "`using Unicode; rename!(Unicode.normalize, df)`."))
```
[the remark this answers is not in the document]
bkamins: I have made three separate error messages (tests cover all options).

bkamins:
@nalimilan - I have updated the PR. The current approach is not to make any breaking changes, but just give more informative error messages.

In the future if we get reports from users that what we do now is problematic we could go back to discuss changes that @stevengj proposed (however, maybe adding special errors will be enough and this way we do not make breaking changes).

stevengj:
> Note that it is even super hard to type in "power_µW" in Julia 

Not on a Mac, where option-m is `µ` (micro).

bkamins:
> people kept confusing `µ` (micro) and `μ` (mu) in real code.

This is indeed a valid issue (it is interesting how such things influence experience; in my region Mac is quite rare).

However, I would still recommend that we take the path:
* first improve the error message in 1.3 release and wait for the feedback;
* based on the feedback decide if we want to make a breaking change.

src/other/index.jl
```diff
@@ -290,9 +290,96 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+            idx_ok = idx_raw == idxs
+            x_ok = x_raw == xs
+            if !idx_ok
+                if x_ok
+                    throw(ArgumentError("column name :$idx not found in the " *
+                                        "data frame. However there is a match of " *
+                                        "Unicode normalized passed column name with " *
+                                        "column name found in the data frame. " *
+                                        "You can use the `Unicode.normalize` " *
+                                        "function to normailize the passed column name."))
```
nalimilan, line 347: This would be more explicit IMO, as otherwise "normalized passed column name" could be understood as "the passed name is already normalized".
```suggestion
                                        "data frame. However, after applying Unicode " *
                                        "normalization to it, the passed column name " *
                                        "matches a column name found in the data frame." *
                                        "You can use the `Unicode.normalize` " *
                                        "function to normalize the passed column name."))
```

src/other/index.jl
```diff
@@ -290,9 +290,96 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+                else
+                    throw(ArgumentError("column name :$idx not found in the " *
+                                        "data frame. However there is a match of " *
+                                        "Unicode normalized passed column name with " *
+                                        "a normalized column name found in the " *
+                                        "data frame. In the call both passed column and " *
```
nalimilan, line 353:
```suggestion
                                        "data frame. However, there is a match " *
                                        "after applying Unicode normalization both " *
                                        "to the passed column names " *
                                        "and to the column names of the data frame. " *
                                        "In the call both passed column and " *
```

src/other/index.jl
```diff
@@ -290,9 +290,96 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+                                        "matching column found in the data frame were not " *
+                                        "normalized. It is recommended to use " *
+                                        "normalized column names and then refer to them " *
+                                        "using normalized names to avoid ambiguity. " *
+                                        "In order to normalize column names in " *
+                                        "an existing data frame `df` do " *
+                                        "`using Unicode; rename!(Unicode.normalize, df)`. " *
+                                        "Similarly, you can use the `Unicode.normalize` " *
+                                        "function to normailize the passed column name."))
```
nalimilan, line 362:
```suggestion
                                        "function to normalize the passed column name."))
```

src/other/index.jl
```diff
@@ -290,9 +290,96 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+                end
+            else
+                @assert !x_ok
+                throw(ArgumentError("column name :$idx not found in the " *
+                                    "data frame. However there is a match of " *
+                                    "passed column name with a normalized column name " *
+                                    "found in the data frame. It is recommended to use " *
```
nalimilan, line 369:
```suggestion
                                    "data frame. However, the passed column name " *
                                    "matches a column name found in the data frame " *
                                    "after applying Unicode normalization to the latter." *
                                    "It is recommended to use " *
```

src/other/index.jl
```diff
@@ -290,9 +290,96 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+    # but are equal only because we performed character mapping
+    for (c1, c2) in zip(a1, a2)
+        if c1 != c2
+            refc1 = get(_julia_charmap, c1, c1)
+            refc2 = get(_julia_charmap, c2, c2)
+            @assert refc1 == refc2
+            throw(ArgumentError("column name :$idx not found in the " *
+                                "data frame. However there is a similar " *
+                                "column name in the data frame where character $c2 " *
+                                "(codepoint: $(UInt32(c2))) is used is instead of $c1 " *
```
nalimilan, line 319:
```suggestion
                                "(codepoint: $(UInt32(c2))) is used instead of $c1 " *
```

src/other/index.jl
```diff
@@ -290,9 +290,96 @@ function fuzzymatch(l::Dict{Symbol, Int}, idx::Symbol)
...
+                                "(codepoint: $(UInt32(c1))). The " *
+                                "error is most likely caused by the Julia parser which " *
+                                "normalizes `Symbol` literals containing such " *
+                                "characters. In order to avoid such problems use only " *
+                                "$refc1 (codepoint: $(UInt32(refc1))) in column names."))
```
nalimilan, line 324: Now that https://github.com/JuliaLang/julia/pull/42561has been merged, could we print the command that fixes column names on Julia ≥ 1.8?

nalimilan:
> > people kept confusing `µ` (micro) and `μ` (mu) in real code.
> 
> This is indeed a valid issue (it is interesting how such things influence experience; in my region Mac is quite rare).

FWIW, on French PC keyboards, we also have micro available (with Shift+a key close to Return). Apparently some variants of US and Polish keyboards also support it via AltGr+M. That's indeed probably the main reason why people get confused.

src/other/index.jl
bkamins: OK - I have changed the error message under `v"1.8-"`.

src/other/index.jl
```diff
@@ -313,15 +313,20 @@ function _norm_eq(a1::String, a2::String, idx::Symbol)
             refc1 = get(_julia_charmap, c1, c1)
             refc2 = get(_julia_charmap, c2, c2)
             @assert refc1 == refc2
+            if VERSION >= v"1.8-"
+                suffix = "use the Unicode.normalize function setting its " *
+                         "chartransform keyword argument to Unicode.julia_chartransform."
```
nalimilan, line 318: How about a more direct suggestion like this?
```suggestion
                suffix = "normalize column names of the data frame by calling " *
                         "`using Unicode; rename!(n -> Unicode.normalize(n, chartransform=Unicode.julia_chartransform), df)`."
```

bkamins: This is a half-measure solution (just like the one below earlier). I will propose a better error (as we need to signal the user if the passed name or the name in data frame was incorrect)

bkamins: @nalimilan In the end I have reverted the simpler error message. The issue is that the number of potential cases we would have to handle is too large to warrant covering all of them given how rare the case is. I now just write which character should be replaced by which.

Just to give you a flavor of what happens:

`"\u0387"` after standard normalization via `Unicode.normalize` becomes `"\u00B7"` which after Julia-specific normalization (`chartransform=Unicode.julia_chartransform`) becomes "\u22C5".

I think that the problematic case is rare enough that we can just write what characters do not match and let the user decide how to handle this.

nalimilan: OK, I've lost track of all the possible cases to handle...

nalimilan (approved):

bkamins:
Thank you!

Issue #2866, opened 2021-09-08
Add a validator for src => transform => dst


bkamins:
In order to improve readability of error messages

bkamins:
@pdeffebach - can you give me please an example when we produce something that is not clear? I have checked and for incorrect selectors we produce:
```
julia> select(df, 'a')
ERROR: ArgumentError: Unrecognized column selector: a
```

nathanrboyer:
I think the best suggestion for this was a function to help construct pairs. 
```
make_pair(source = [:a, :b], fun = f, dest = AsTable)`
transform(df, make_pair(...))
```
There should also be functionality added to make `typeof(source)`, `typeof(f(source))`, and `typeof(dest)` work.

This might also help with #2870 if the function syntax would make it easier to programatically create a list of `args...` to pass to `transform` and the like.

bkamins:
Could you please explain what is the advantage of:
```
make_pair(source = [:a, :b], fun = f, dest = AsTable)
```
over
```
[:a, :b] => f => AsTable
```
?

For me it seems more verbose and less readable.

> There should also be functionality added to make `typeof(source)`, `typeof(f(source))`, and `typeof(dest)` work.

I am not fully clear what you mean here? Could you please provide an example of assumed input and expected output?

Thank you!

nathanrboyer:
It is only advantageous if the new function (or package) can provide additional error checking, type checking, and validation. Ideally, it could tell you what part of your pair is invalid or needs broadcasted.

bkamins:
>  Ideally, it could tell you what part of your pair is invalid or needs broadcasted.

OK - now I understand.

The problem is that it is extremely hard (if not impossible). The problem is that Julia is very flexible what you can define to work (e.g. consider https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects-1 which are objects, but are callable - you cannot detect such things by checking types unfortunately)

But I will think what can be done.

nathanrboyer:
I'll call on @pdeffebach and @CameronBieganek to add their thoughts too.

nathanrboyer:
I'm afraid I don't know enough to be helpful with the implementation details. Could you start by supporting more basic types and just print a warning "Validation not yet supported" for functors and such? I imagine advanced users would construct the pairs themselves anyway.

pdeffebach:
[...]

PR #2907, opened 2021-10-12
Updated Basic Usage of Transformation Functions

nathanrboyer:
Documentation update related to #2866. This is a work in progress and my first pull request. Contributions welcome.


docs/src/man/basics.md
```diff
@@ -1510,31 +1510,69 @@ julia> german[Not(5), r"S"]
...
+When multiple columns are selected,
+the `function` will receive the columns as multiple arguments in the
+order they are selected like `function(column1, column2, column3)`.
+Alternatively, the selected columns can be "slurped" into a
+single argument using `function(columns...)`.
+In more advanced usage, the `function` itself can use column selectors
+in its definition.
+
+`new_column_name` may be a `String` or a `Symbol`.
+(*Soon `new_column_name` will also accept a renaming function.*)
```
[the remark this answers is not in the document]
nathanrboyer: Can you only use the renaming function in the `source_column_selector => function => new_column_names` form? If I use a renaming function in `source_column_selector => new_column_names`, it thinks I am trying to do `source_column_selector => function`.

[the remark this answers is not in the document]
bkamins: Can you please share the code you are using as I am not clear what is not working for you?

If in `a => b` the `b` part is a function it will be interpreted as transformation specification. It is the only way it could be interpreted, as otherwise how could DataFrames.jl know if `b` should be a transformation or a column renaming function?

If you pass `a => b => c` then `c` can be one of:
* `AsTable`
* list of column names
* a function taking input column names from `a` and producing target column names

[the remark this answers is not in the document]
nathanrboyer: You confirmed my suspicion. I didn't know if there was a special way around the `a => b` renaming limitation.
```julia
julia> using DataFrames

julia> df = DataFrame(x=[1,7,3], y=[6,0,2])
3×2 DataFrame
 Row │ x      y     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      6
   2 │     7      0
   3 │     3      2

julia> newname(x::String) = x*"_new"
newname (generic function with 1 method)

julia> transform(df, :x => (x -> 2x) => newname)
3×3 DataFrame
 Row │ x      y      x_new 
     │ Int64  Int64  Int64
─────┼─────────────────────
   1 │     1      6      2
   2 │     7      0     14
   3 │     3      2      6

julia> transform(df, :x => newname)
ERROR: MethodError: no method matching newname(::Vector{Int64})
Closest candidates are:
  newname(::String) at REPL[3]:1
Stacktrace:
 [1] _transformation_helper(df::DataFrame, col_idx::Int64, ::Base.RefValue{Any})
   @ DataFrames C:\Users\nboyer.AIP\.julia\dev\DataFrames\src\abstractdataframe\selection.jl:397
 [2] select_transform!(::Base.RefValue{Any}, df::DataFrame, newdf::DataFrame, transformed_cols::Set{Symbol}, copycols::Bool, allow_resizing_newdf::Base.RefValue{Bool})
   @ DataFrames C:\Users\nboyer.AIP\.julia\dev\DataFrames\src\abstractdataframe\selection.jl:575
 [3] _manipulate(df::DataFrame, normalized_cs::Vector{Any}, copycols::Bool, keeprows::Bool)
   @ DataFrames C:\Users\nboyer.AIP\.julia\dev\DataFrames\src\abstractdataframe\selection.jl:1444
 [4] manipulate(::DataFrame, ::Any, ::Vararg{Any, N} where N; copycols::Bool, keeprows::Bool, renamecols::Bool)
   @ DataFrames C:\Users\nboyer.AIP\.julia\dev\DataFrames\src\abstractdataframe\selection.jl:1372
 [5] #select#466
   @ C:\Users\nboyer.AIP\.julia\dev\DataFrames\src\abstractdataframe\selection.jl:992 [inlined]
 [6] transform(df::DataFrame, args::Any; copycols::Bool, renamecols::Bool)
   @ DataFrames C:\Users\nboyer.AIP\.julia\dev\DataFrames\src\abstractdataframe\selection.jl:1070
 [7] transform(df::DataFrame, args::Any)
   @ DataFrames C:\Users\nboyer.AIP\.julia\dev\DataFrames\src\abstractdataframe\selection.jl:1059
 [8] top-level scope
   @ REPL[6]:1
```

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,95 @@ julia> german[Not(5), r"S"]
...
+  - *may only select one column unless broadcasting*
+  - *cannot be used with `subset` or `subset!`*
+- `source_column_selector`
+  - selects source column(s) without transforming them
+  - *often used with `select` or `select!` for isolating or moving columns*
+
+#### `source_column_selector`
+The most basic `source_column_selector` is a column name, but there are many more ways to
+select columns as explained in the first section of the 
+[Indexing API documentation](https://dataframes.juliadata.org/stable/lib/indexing/#Indexing).
```
nathanrboyer, line 1550: How do I make this an internal link?

docs/src/man/basics.md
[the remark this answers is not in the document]
bkamins: > the `a => b` renaming limitation.

If you pass one column (as this is what you do) just write `a => identity => b` or just `a => newname(string(a))`.

The function interface is intended for the cases where you rather perform multiple transformations in one shot.

docs/src/man/basics.md
bkamins: Please check the Documenter.jl documentation. I never remember this. Maybe just writing `#Indexing` is enough?

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,97 @@ julia> german[Not(5), r"S"]
...
+      - `transformation_function(source_dataframe, [transformation1, transformation2])`
+      - `transformation_function(source_dataframe, [transformation1 transformation2])`
+    - Any of the `transformation` pair syntax forms shown above can also use
+      broadcasting with `.=>` to apply multiple transformations at once.
+      See the next section for examples.
+    - More complex transformations are available through the syntax
+      `transformation_function(function, source_dataframe)`.
+      This usage is more advanced,
+      so it is discussed in a separate section
+      [here](**TODO: Add a link to relevant info.**).
```
nathanrboyer, line 1601: I couldn't find a good place to link to for the transformation_function(function, source_dataframe) form. Even the API documentation doesn't seem to cover it much. I do not want to cover it in this section either. Maybe some external blog post exists?

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,97 @@ julia> german[Not(5), r"S"]
 
 ## Basic Usage of Transformation Functions
```
nathanrboyer, line 1511: I am pretty much done updating this section. I will work on the Broadcasting with Transformation Functions section next when I find some time.

docs/src/man/basics.md
nathanrboyer: I think `[Indexing](@ref)`? https://juliadocs.github.io/Documenter.jl/dev/man/syntax/#@ref-link

nathanrboyer:
I think that is fine and easy enough to work around. I just wanted to make sure I documented correctly.

docs/src/man/basics.md
bkamins: `transformation_function(function, source_dataframe)` is the same as `transformation_function(source_dataframe, function)` - there is nothing special to it except that it allows `do` syntax (and that is why `subset` and `subset!` do not support it)

nathanrboyer: Oh, okay.

nathanrboyer:
I am pretty much done updating the Basic Usage of Transformation Functions section. I will work on the Broadcasting with Transformation Functions section next when I find some time.

bkamins:
@nathanrboyer - please let me know when you are done then I will review.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
-  and stored in `target_column_name` column.
-- `source_column => transformation`;
-  In this scenario we apply the transformation function to `source_column` and the
-  target column names is automatically generated.
-- `source_column => target_column_name` renames the `source_column` to `target_column_name`.
-- `source_column` just keep the source column as is in the result without any transformation;
-
-These rules are typically called transformation mini-language.
-
+In DataFrames.jl there are seven functions that can be used to transform data frame columns:
```
bkamins, line 1513: It seems to me that `subset` is not transforming columns of a data frame. Maybe use some more general name for the operation if you want to keep `subset` in the list? (keeping in mind that you do not list `filter` here)

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+
+| Function     | Memory Usage                     | Column Retention                             | Row Retention                                       |
+| ------------ | -------------------------------- | -------------------------------------------- | --------------------------------------------------- |
+| `transform`  | Creates a new data frame.        | Retains both source and transformed columns. | Retains same number of rows as source data frame.   |
```
bkamins, line 1517: maybe there is a better word than *transformed*? the reason is that we have `transform` function. I think it would be clearest if this "word" would be first introduced like:

> all the functions take a data frame as a first argument and then a list of operations that should be performed on the columns of this data frame

In this wording we could later use *operation* word.

(I am giving vague suggestions as I am not a native speaker and here every word matters)

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+| `select`     | Creates a new data frame.        | Retains only transformed columns.            | Retains same number of rows as source data frame.   |
+| `select!`    | Modifies an existing data frame. | Retains only transformed columns.            | Retains same number of rows as source data frame.   |
+| `subset`     | Creates a new data frame.        | Retains both source and transformed columns. | Number of rows is determined by the transformation. |
+| `subset!`    | Modifies an existing data frame. | Retains both source and transformed columns. | Number of rows is determined by the transformation. |
+| `combine`    | Creates a new data frame.        | Retains only transformed columns.            | Number of rows is determined by the transformation. |
+
+### Constructing Transformation Pairs
+All of the functions above use the same syntax which is commonly
+`transformation_function(source_dataframe, transformation)`.
+The `transformation` argument is a `Pair` which defines the
```
bkamins, line 1528: the naming here should be adjusted accordingly to what we agree above.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+transformation to be applied to the `source_dataframe`,
+and it can take any of the forms listed below:
+
+- `source_column_selector => function`
```
bkamins, line 1532: maybe show the cases here in the order you explain them later (and hint that you are going to add these explanations)

Also add that these are the most common patterns.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+    - *cannot be used with `subset` or `subset!`*
+- `source_column_selector => new_column_names`
+  - renames source column
+    - *may only select one column unless broadcasting*
+    - *may be used to split a column containing collections into multiple columns*
+    - *cannot be used with `subset` or `subset!`*
+- `source_column_selector`
+  - selects source column(s) without transforming them
+    - *often used with `select` or `select!` for isolating or moving columns*
+
```
bkamins, line 1547: add somewhere that for `select` operations must return boolean vector.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+#### `source_column_selector`
+The most basic `source_column_selector` is a column name,
+but there are many more ways to select columns as explained
```
bkamins, line 1550: I think it would be good to give some examples

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+in the first section of the [Indexing](@ref) API.
```
bkamins, line 1551: A special case of `source_column_selector` is `AsTable` that should be mentioned somewhere.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+Slurping and splatting with `...` can be used to define a function which can
+accept any number of columns returned by `source_column_selector`.
+For example, the function `f(columns...) = max.(columns...)` will return a new column
+containing the largest element from each row for any number of input columns.
+
+#### `new_column_names`
+The results of the transformation will be placed into new columns
+with names defined by `new_column_names`.
+`new_column_names` may be a string, a symbol, a vector of strings,
+or a vector of symbols.
```
bkamins, line 1574: or `AsTable`.

docs/src/man/basics.md
```diff
@@ -2068,3 +2132,212 @@ options of the transformation mini-language. More advanced examples, in particul
...
+
+## Broadcasting with Transformation Functions
+
+[Broadcasting](https://docs.julialang.org/en/v1/manual/arrays/#Broadcasting)
+is a convenient way to apply multiple transformations
+within one call to a transformation function.
+Broadcasting within the `Pair` of a `transformation` is no different than
+broadcasting in base Julia.
+The broadcasting `.=>` will be expanded into a vector of pairs,
+and this expansion will occur before the tranformation function is invoked.
```
bkamins, line 2144: maybe give some example here before moving to explanation of technical aspects what happens underneath?

docs/src/man/basics.md
```diff
@@ -2068,3 +2132,212 @@ options of the transformation mini-language. More advanced examples, in particul
...
+
+To illustrate these concepts, let us first examine the `Type` of a basic `Pair`.
+In DataFrames, a `Symbol`, `String`, or `Integer`
```
bkamins, line 2147:
```suggestion
In DataFrames.jl, a `Symbol`, `String`, or `Integer`
```

docs/src/man/basics.md
```diff
@@ -2068,3 +2132,212 @@ options of the transformation mini-language. More advanced examples, in particul
...
+   2 │     4     10
+   3 │     6     12
+```
+
+This broadcasting approach is even more useful if we use column selectors which
+generate a vector of column names for us rather than writing them all out manually.
+Consider a data frame with temperature data at three different locations taken over time.
+
+```julia
+julia> df = DataFrame(Time = 1:4, Temperature1 = [20, 23, 25, 28], Temperature2 = [33, 37, 41, 44], Temperature3 = [15, 10, 4, 0])
```
bkamins, line 2279: can you add newlines to avoid overlong lines?

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+| `transform`  | Creates a new data frame.        | Retains both source and transformed columns. | Retains same number of rows as source data frame.   |
+| `transform!` | Modifies an existing data frame. | Retains both source and transformed columns. | Retains same number of rows as source data frame.   |
+| `select`     | Creates a new data frame.        | Retains only transformed columns.            | Retains same number of rows as source data frame.   |
+| `select!`    | Modifies an existing data frame. | Retains only transformed columns.            | Retains same number of rows as source data frame.   |
+| `subset`     | Creates a new data frame.        | Retains both source and transformed columns. | Number of rows is determined by the transformation. |
+| `subset!`    | Modifies an existing data frame. | Retains both source and transformed columns. | Number of rows is determined by the transformation. |
+| `combine`    | Creates a new data frame.        | Retains only transformed columns.            | Number of rows is determined by the transformation. |
+
+### Constructing Transformation Pairs
+All of the functions above use the same syntax which is commonly
```
nalimilan, line 1526:
```suggestion

All of the functions above use the same syntax which is commonly
```

docs/src/man/basics.md
nalimilan: Also, better say `source_columns_selector` or just `source_columns` as multiple columns can be selected.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+transformation to be applied to the `source_dataframe`,
+and it can take any of the forms listed below:
+
+- `source_column_selector => function`
+  - passes source column(s) to function
+  - automatically names the resulting column(s)
+- `source_column_selector => function => new_column_names`
+  - passes source column(s) to function
+  - names the resulting column(s) `new_column_names`
+    - *cannot be used with `subset` or `subset!`*
```
nalimilan, line 1538: This should have the same indentation level as previous bullets, as it's not a sub-point of the previous one.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+- `source_column_selector => new_column_names`
+  - renames source column
+    - *may only select one column unless broadcasting*
```
nalimilan, line 1541: You mean `.=>` broadcasting? Then I wouldn't mention it as it's a different syntax from the one described in this point.

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+| `subset`     | Creates a new data frame.        | Retains both source and transformed columns. | Number of rows is determined by the transformation. |
+| `subset!`    | Modifies an existing data frame. | Retains both source and transformed columns. | Number of rows is determined by the transformation. |
+| `combine`    | Creates a new data frame.        | Retains only transformed columns.            | Number of rows is determined by the transformation. |
+
+### Constructing Transformation Pairs
+All of the functions above use the same syntax which is commonly
+`transformation_function(source_dataframe, transformation)`.
+The `transformation` argument is a `Pair` which defines the
+transformation to be applied to the `source_dataframe`,
+and it can take any of the forms listed below:
```
nalimilan, line 1530:
```suggestion
and it can take any of the following forms (see below for a description of each component):
```

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+column passed as a vector.
+If you instead want to apply a function to each element in the column,
+then you can wrap your element-wise function in `ByRow` like
+`ByRow(my_elementwise_function)`.
+This will apply `my_elementwise_function` to every element in the column
+and then collect the results back into a vector.
+
+When multiple columns are selected by `source_column_selector`,
+the `function` will receive the columns as multiple positional arguments in the
+order they are selected like `f(column1, column2, column3)`.
```
nalimilan, line 1564:
```suggestion
`function` will receive the columns as multiple positional arguments in the
order they are selected, e.g. `f(column1, column2, column3)`.
```

docs/src/man/basics.md
```diff
@@ -1510,32 +1510,96 @@ julia> german[Not(5), r"S"]
...
+In the `source_column_selector => function` transformation form,
+`new_column_names` will be generated automatically as the function name
+appended to the source column name with an underscore.
+However, if keyword argument `renamecols=false` is passed
+to the transformation function,
+then the new columns will retain their original source names
+instead of using automatically generated names.
+
+#### More Information
+This transformation pair syntax is sometimes referred to as a mini-language.
```
nalimilan, line 1588:
```suggestion
This transformation pair syntax is sometimes referred to as a the DataFrames
mini-language or domain-specific language (DSL).
```

docs/src/man/basics.md
```diff
@@ -2068,3 +2132,212 @@ options of the transformation mini-language. More advanced examples, in particul
...
+─────┼─────────────────────────────────────────────────
+   1 │     1            20            33            15
+   2 │     2            23            37            10
+   3 │     3            25            41             4
+   4 │     4            28            44             0
+```
+
+To convert all of the temperature data in one transformation,
+we just need to define a conversion function and broadcast
+all of the rows except for `Time` to it.
```
nalimilan, line 2292:
```suggestion
it to all of the columns except `Time`.
```

docs/src/man/basics.md
```diff
@@ -2068,3 +2132,212 @@ options of the transformation mini-language. More advanced examples, in particul
...
+ Row │ Time   Temperature1  Temperature2  Temperature3
+     │ Int64  Int64         Int64         Int64
+─────┼─────────────────────────────────────────────────
+   1 │     1           293           306           288
+   2 │     2           296           310           283
+   3 │     3           298           314           277
+   4 │     4           301           317           273
+```
+
+The function `celsius_to_kelvin` above was defined to operate on scalar values.
```
nalimilan, line 2309: The rest of this section deals with broadcasting over rows, while the part above was about columns. Can you try to make this difference clearer? The combination of both can be confusing.

docs/src/man/basics.md
```diff
@@ -2068,3 +2132,212 @@ options of the transformation mini-language. More advanced examples, in particul
...
+If a function is used as part of a transformation `Pair`,
+like in the `source_column_selector => function => new_column_names` form,
+then the function is repeated in each pair of the resultant vector.
+This is an easy way to apply a function to multiple columns at the same time.
+
+```julia
+julia> f(x) = 2 * x
+f (generic function with 1 method)
+
+julia> ["x", "y"] .=> f .=> ["a", "b"]
```
nalimilan, line 2238: While you're at it, it would make sense to show an example of a trick which works very well but is hard to discover by yourself: `["x", "y"] .=> [f f2]` to apply several functions to several columns.

nalimilan (commented):
Thanks! I feel like this deserves its own section of the manual now. We should stop adding things to First Steps. Maybe we could just keep a short presentation of transformation functions with a few examples, and refer to the dedicated section for the detailed reference.

We also have some material in Split-Apply-Combine which would ideally be harmonized, but that can be done later.

docs/src/man/basics.md
nathanrboyer: Yes, okay.

docs/src/man/basics.md
nathanrboyer: I went back and forth on the best formatting here. I wasn't sure if italics was obvious enough to distinguish the main points from the side-notes. Open to suggestions.

docs/src/man/basics.md
nathanrboyer: Last line should say "columns". The columns are broadcast to the function based on the direction of the `=>` arrow. The function is not broadcast to the columns right?

docs/src/man/basics.md
nathanrboyer: Sure

docs/src/man/basics.md
nalimilan: I think it's the function which is broadcast, as it's repeated to match the number of columns. Just like in `[1, 2] .+ 1`.

docs/src/man/basics.md
nathanrboyer: I didn't touch the Examples section between the two sections I wrote since it covered most things. It would probably be better to rewrite and split it up, but that might be a separate PR.

docs/src/man/basics.md
nathanrboyer: I ignored the existence of `AsTable` here.

1. I've never used it.
2. I don't totally understand it.
3. It seems to break most of the rules I've neatly outlined.
4. Probably not for beginners?

Might be better to write a separate section on it.

docs/src/man/basics.md
nathanrboyer: I didn't touch the Examples section between the two sections I wrote since it covered most things. It would probably be better to rewrite and split it up, but that might be a separate PR.

docs/src/man/basics.md
nathanrboyer: Yes, forgot that was possible 😅

docs/src/man/basics.md
nathanrboyer: I will keep `selector` as this was one of the main confusions I had. You do not use the source columns themselves (e.g. `df.x`). To me `column` just indicates the type of `selector`, and a `selector` could select one or more columns. `columns_selector` feels a little clunky to me personally.

docs/src/man/basics.md
nathanrboyer: `operation` is good.

Yes, I'm not totally happy with having `transform`, `transformation_function`, and `transformation` all mean different things.

Similarly, I somewhat dislike the name `function` in `source_column_selector => function` since you cannot write:
```julia
source_column_selector = "x"
function(x) = 2x #error
transform(df, source_column_selector => function) #error
```
However, I ultimately left the `function` name since I think it is the most descriptive.

docs/src/man/basics.md
nathanrboyer: How about "*may only select one column (unless `.=>` broadcasting)*" or "*may only select one column*". @bkamins previously mentioned that I should add this point.

docs/src/man/basics.md
nathanrboyer: Maybe `source_column_names` is better, but I don't know if `Not()` or `1` would count as names.

docs/src/man/basics.md
bkamins: So let us agree on the following terminology:
```
source_column_selector => operation_function => new_column_names
```
and the whole `Pair` syntax is `operation`. The interpretation of `operation` is defined in `operation mini-language`. The functions that interpret `operation` using the `operation mini-language` meaning are called `functions performing operations` or `functions supporting operation mini-language` (both forms are acceptable).

If we (@nathanrboyer, @nalimilan and @bkamins + @pdeffebach if available) agree on this terminology (i.e. if in this thread we all agree that this is what we like) then:
* this PR should be accordingly adjusted (both in "code" sections and in descriptions)
* whole documentation and manual should be re-written to use it consistently (in a separate PR; I can do this if no-one else wants after this PR is merged)

@nathanrboyer - before making these changes let us first agree that we like the wording. Let me summarize it in bullets again:

1. `source_column_selector => operation_function => new_column_names`
2. whole `Pair` syntax is `operation`.
3. interpretation of `operation` is defined in `operation mini-language`
4. functions (8 of them in total) that interpret `operation` using the `operation mini-language` meaning are called `functions performing operations` or `functions supporting operation mini-language` (both forms are acceptable).

docs/src/man/basics.md
bkamins: If we agree on what I have proposed above I think `source_column_selector` is OK, and I would give several examples of it to the user (like `1`, `:a`, `"a"`, `[:a, :b]`, `Not("a")` etc.) in the place where you would define it so that it is clear what we mean.

docs/src/man/basics.md
bkamins: ⚠️ is a nice emoticon for this I think + italics

docs/src/man/basics.md
bkamins: What I mean is to put mini-examples everywhere in the text, see e.g. my proposal in https://github.com/JuliaData/DataFrames.jl/pull/2907/files#r792878207 where I show what I mean by these examples

docs/src/man/basics.md
bkamins: > Might be better to write a separate section on it.

This is OK (and better to be done at the end)

> It seems to break most of the rules I've neatly outlined.

The rule is simple - instead of passing columns as consecutive positional argument you pass a `NamedTuple` as a single argument.

> Probably not for beginners?

Even beginners want to do row aggregation often like:
```
julia> df = DataFrame(ones(1, 100), :auto)
1×100 DataFrame
 Row │ x1       x2       x3       x4       x5       x6       x7       x8       x9       x10      x11      x12      x13      x14      x15      x16      x17    ⋯
     │ Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float6 ⋯
─────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │     1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1.0      1. ⋯
                                                                                                                                             84 columns omitted

julia> select(df, AsTable(:) => sum)
1×1 DataFrame
 Row │ x1_x2_etc_sum
     │ Float64
─────┼───────────────
   1 │         100.0
```
I think this example or similar with `mean` should be included somewhere as it is super common.

In general use `AsTable` if you have a function that *accepts a collection as its input* as opposed to list of arguments.

Note that these two operations are equivalent (also might be included to show the difference):

```
julia> df = DataFrame([1 2 3 4], :auto)
1×4 DataFrame
 Row │ x1     x2     x3     x4
     │ Int64  Int64  Int64  Int64
─────┼────────────────────────────
   1 │     1      2      3      4

julia> select(df, r"x" => ByRow(min))
1×1 DataFrame
 Row │ x1_x2_etc_min
     │ Int64
─────┼───────────────
   1 │             1

julia> select(df, AsTable(r"x") => ByRow(minimum))
1×1 DataFrame
 Row │ x1_x2_etc_minimum
     │ Int64
─────┼───────────────────
   1 │                 1
```
If someone understands the difference between `min` and `minimum` then the need for `AsTable` will be clear after seeing this example.

bkamins: Of course there are more advanced uses of `AsTable` (where you would refer to the actual column names as you can extract them out from a `NamedTuple`, but the ones I have listed above are most common).

docs/src/man/basics.md
bkamins: Same here - `AsTable` can be discussed separately, but it is a common operation for splitting a column into multiple columns with automatic column naming:
```
julia> df = DataFrame(in=[(a=1,b=2), (a=3, b=4)])
2×1 DataFrame
 Row │ in
     │ NamedTup…
─────┼────────────────
   1 │ (a = 1, b = 2)
   2 │ (a = 3, b = 4)

julia> select(df, :in => AsTable)
2×2 DataFrame
 Row │ a      b
     │ Int64  Int64
─────┼──────────────
   1 │     1      2
   2 │     3      4

julia> select(df, :in => [:out_a, :out_b]) # you can also give the names of the columns explicitly
2×2 DataFrame
 Row │ out_a  out_b
     │ Int64  Int64
─────┼──────────────
   1 │     1      2
   2 │     3      4
```

docs/src/man/basics.md
bkamins: But the question is (maybe you are right though) if people will be able to digest the rules without seeing the examples?

docs/src/man/basics.md
nathanrboyer: 3. I prefer `operation syntax`. The language is still Julia. 🤷‍♂️
4. I count 7 functions in my table. What am I missing?
    Could these be called `manipulations` or `manipulation functions`?
    I think it would have to clarified to be "functions performing DataFrame.jl `operations`" since all functions perform operations in general.

bkamins: 1. `operation syntax` is OK
2. yes `7` (I just multiplied 2x4, but forgotten that `combine` does not have `!` variant :))
3. `manipulation` instead of `operation` is OK for me (could you please just translate my four bullets under your proposal so that it is clear what you think would be clearest for you). Thank you!

nathanrboyer: In my post above I was suggesting the following:
1. `source_column_selector => operation_function => new_column_names`
2. whole Pair syntax is `operation`
3. interpretation of `operation` is defined in `operation syntax`
4. functions that interpret `operation` using the `operation syntax` are called `data frame manipulation functions`

bkamins: I am OK with this. @nalimilan + @pdeffebach : agreed?

nathanrboyer: I'll also suggest:

1. `source_column_selector => manipulation_function => new_column_names`
2. the whole `Pair` syntax is a `manipulation`
3. interpretation of `manipulations` is `manipulation syntax`
4. functions that interpret `manipulation syntax` are `functions supporting manipulation syntax` or `functions for manipulating dataframes`

bkamins:
@nathanrboyer - what is the status of this PR?

nathanrboyer:
Waiting for input from @pdeffebach and @nalimilan to reach a consensus on terminology. https://github.com/JuliaData/DataFrames.jl/pull/2907#discussion_r792872612

docs/src/man/basics.md
nalimilan: Sorry, I had missed this thread. I'm fine with the "operation" terminology from https://github.com/JuliaData/DataFrames.jl/pull/2907#discussion_r793008180, which I think we already use in a handful of places (though most of the time we indeed say "transformation"). Though note that this will require adapting docstrings to be fully consistent (should be easy).

I'm less fond of saying "manipulation", as to me this refers to functions such as `combine`, `select` and `transform`, rather than to the *operations* that are passed to them using the pair syntax.

nathanrboyer: > 4. Functions that interpret `operation`s using `operation syntax` are called `data frame manipulation functions`.

@nalimilan, do you want to propose anything different for the above? `functions performing operations` from @bkamins suggestion felt too general to be referring only to the these seven functions. Can you think of a way to make this more clear?

nathanrboyer: Okay, I just wanted to make sure you were okay with "manipulation" in that context.

pdeffebach:
I agree with @bkamins that `operation` is the word to use here. I have adapted the same convention in the DataFramesMeta.jl documentation.

bkamins:
@nathanrboyer - I am removing a milestone from this PR. The reason is that we are making a lot of changes before the 1.4 release so this PR probably will not be finished very soon.

Of course once you come back to it we will merge it.

bkamins:
@nathanrboyer - it is still a draft. Right?

nathanrboyer:
Yes, sorry. I just picked up working on it again. I was hoping it wouldn't alert you yet. Should I avoid `git push` to my fork until I'm finished?

bkamins:
No - pushing is OK. Then I will ignore changes in this PR until it is marked as ready for review. Thank you!

nathanrboyer:
`basics.md` contains the only content changes.
`notes.md` contains questions I had while writing. The answers to some of these questions should be added to the documentation or to pull requests.
The other changed files should be deleted. I will need help figuring out the correct Git process to get this branch compatible again with master.
Feedback and reorganization welcome, thanks!

notes.md
```diff
@@ -0,0 +1,215 @@
+# Why does `:` only sometimes work?
+I don't like `:` being listed as a `source_column_selector` when it often errors.
+
+`select(df, :)`
+: works
+
+`transform(df, Cols(:) => ByRow(max))`
+: works
+
+`transform(df, : => ByRow(max))` errors
```
bkamins, line 10: `(:)` is required. It is a limitation of Julia parser

notes.md
```diff
@@ -0,0 +1,215 @@
...
+: ERROR: syntax: whitespace not allowed after ":" used for quoting
+
+Similarly, regular expressions also sometimes must be wrapped in `Cols`.
```
bkamins, line 13: This is also a limitation of the fact that regular expressions are defined in Base Julia. We cannot adjust their behavior. Therefore we have `Cols` wrapper that can be put around anything.

bkamins: `Cols` around regex is only needed with broadcasting

notes.md
```diff
@@ -0,0 +1,215 @@
...
+This seems inconsistent.
+I don't understand the rules for when `Cols` is and isn't needed.
```
bkamins, line 15: The only two syntax problems are the ones you asked above. This can be added in your PR.

notes.md
```diff
@@ -0,0 +1,215 @@
...
+julia> transform(df, Cols(r"Temp") .=> (t->(t.-32)*5/9), renamecols=false)
+3×3 DataFrame
+ Row │ Time   TopTemp  BottomTemp
+     │ Int64  Float64  Float64
+─────┼────────────────────────────
+   1 │     3  21.1111     10.0
+   2 │     4  22.7778     12.7778
+   3 │     5  37.7778     26.6667
+
+julia> transform(df, r"Temp" .=> (t->(t.-32)*5/9), renamecols=false) # breaks without Cols
```
bkamins, line 37: `r"Temp"` lists columns AFTER broadcasting.

`Cols(r"Temp")` lists columns WITHIN broadcasting.

notes.md
```diff
@@ -0,0 +1,215 @@
...
+```
+```julia
+julia> df = DataFrame([1 2 3 4], :auto)
+1×4 DataFrame
+ Row │ x1     x2     x3     x4
+     │ Int64  Int64  Int64  Int64
+─────┼────────────────────────────
+   1 │     1      2      3      4
+
+julia> select(df, r"x" => ByRow(min)) # works without Cols
```
bkamins, line 48: yes, because multiple positional arguments are passed to `min`.

notes.md
```diff
@@ -0,0 +1,215 @@
...
+2×2 DataFrame
+ Row │ a      b
+     │ Int64  Int64
+─────┼──────────────
+   1 │     1      2
+   2 │     3      4
+```
+Then I should also be able to do the reverse to pack named tuples and get `df` back:
+```julia
+julia> df3 = select(df2, AsTable(:) => :in)
```
bkamins, line 77: Currently it is done with:
```
select(df2, AsTable(:) => ByRow(identity) => :in)
```

notes.md
```diff
@@ -0,0 +1,215 @@
...
+ERROR: ArgumentError: Unrecognized column selector: AsTable(Colon()) => :in
+```
+
+# Why are types truncated when there is still a lot of room?
+
+Also, why is `NamedTup...` truncated in the output above? There is room.
```
bkamins, line 83: because when typesetting the header we do not know the column width yet. However, maybe @ronisbr could improve this in the future.

notes.md
bkamins: a very similar example is given in the documentation of `AsTable`

notes.md
```diff
@@ -0,0 +1,215 @@
...
+julia> select(df, :)
+3×4 DataFrame
+ Row │ id     first_name  last_name  age
+     │ Int64  String      String     Int64
+─────┼─────────────────────────────────────
+   1 │     1  José        Garcia        61
+   2 │     2  Emma        Marino        24
+   3 │     3  Nathan      Boyer         33
+
+julia> select(df, AsTable(:))
```
bkamins, line 110: It cannot be used because it would be equivalent to:
```
julia> select(df2, AsTable(:) => identity)
ERROR: ArgumentError: Table returned but a single output column was expected
```
which is currently disallowed (for user safety, i.e. user has to explicitly decide how to handle multi-column output).

notes.md
```diff
@@ -0,0 +1,215 @@
...
+ERROR: ArgumentError: Unrecognized column selector: AsTable(Colon())
+```
+
+# Is summary table correct?
+Can `subset` ever create new columns?
```
bkamins, line 115: yes

notes.md
```diff
@@ -0,0 +1,215 @@
...
+Can `combine` ever create more than one row?
```
bkamins, line 116: yes

notes.md
```diff
@@ -0,0 +1,215 @@
...
+   4 │     4      8
+
+julia> add_new(s) = "new_" * s
+add_new (generic function with 1 method)
+
+julia> rename(df, :a => add_new)
+ERROR: MethodError: no method matching rename!(::DataFrame, ::Vector{Pair{Symbol, typeof(add_new)}})
+```
+
+I would like it to behave as below:
```
bkamins, line 139: this could be added. Can you please open a separate issue?

notes.md
```diff
@@ -0,0 +1,215 @@
...
+   1 │     1      5
+   2 │     2      6
+   3 │     3      7
+   4 │     4      8
+```
+
+# Cannot automatically group from NamedTuples?
+I am suprised that this does not work:
+```julia
+julia> df = DataFrame(data = [(a=1,b=2), (b=3,a=4)]) # vector of `NamedTuple`s
```
bkamins, line 168: this is intentional. Why would you expect this to work? This is clearly bad data. It should be a `Dict` not a `NamedTuple` if you want the order to be ignored:
```
julia> transform(df, :data => ByRow(Dict∘pairs)=> AsTable)
2×3 DataFrame
 Row │ data            a      b
     │ NamedTup…       Int64  Int64
─────┼──────────────────────────────
   1 │ (a = 1, b = 2)      1      2
   2 │ (b = 3, a = 4)      4      3
```

bkamins:
I answered all your questions in notes.md.

What I think will be simplest for you is to create a new branch from current `main` and make a new PR (later we will close this one and just work on a new one). This will be least work on your side I think. In that new PR please include only the things that should be merged (and we can use this PR of discussion on things you want to ask)

notes.md
ronisbr: Hi everyone!

When we were discussing how to use PrettyTables.jl to print DataFrames, we decided to use the previous mechanism to provide the type name (the function is called `batch_compacttype`). We can avoid those "unnecessary" contractions by using `crop_subheader = true` and specifying a minimum size for the columns. For example:

```
julia> df = DataFrame(a = (a = 1, b = 2), b = (c = 3, d = 4));

julia> show(df, header = (names(df), [eltype(c) for c in eachcol(df)]), crop_subheader = true)
1×2 DataFrame
 Row │ a               b
     │ NamedTuple{(:…  NamedTuple{(:…
─────┼────────────────────────────────
   1 │ (a = 1, b = 2)  (c = 3, d = 4)
```

Notice that the type sometimes is strange due to the cropping. That was the reason we decided to keep the previous behavior.

ronisbr: Yes! Sure! I will add to the list :)

notes.md
nathanrboyer: Okay, the current implementation makes sense to me now. `ByRow` should be required here.

notes.md
nathanrboyer: By the way, I thought it was fun to leave my name in the documentation, but I can change that if it is in bad taste.

notes.md
nathanrboyer: This is the only comment I'll push back on. I have the impression that order matters for tuples but not named tuples: keyword arguments can be supplied in any order, and named tuple destructuring works by name not by order. The difference between `NamedTuple` and `Dict` to me is mutability. As long as all of the named elements exist in the `data` column, I would still consider the data good. Even if not, `missing` values could be inserted as needed.
In any case, I will open a separate issue to discuss.

notes.md
nathanrboyer: I will.

notes.md
bkamins: `Tables.columntable(df)`

notes.md
bkamins: Mutability is also a difference.
But note that we have `Dict` and `OrderedDict`. `OrderedDict` is kind-of mutable equivalent of `NamedTuple` (with a bit different API).

notes.md
nathanrboyer: It looks like the same not-expanded-yet issue occurs with `(:)`, but then I'm confused why `2:3` works fine. Shouldn't the `2:3` behavior match the `(:)` behavior?
```julia
julia> df = DataFrame(a = 1:3, b = 4:6, c = 7:9)
3×3 DataFrame
 Row │ a      b      c     
     │ Int64  Int64  Int64
─────┼─────────────────────
   1 │     1      4      7
   2 │     2      5      8
   3 │     3      6      9

julia> transform(df, (:) .=> ByRow(x -> x + 1))
ERROR: MethodError: no method matching (::var"#23#24")(::Int64, ::Int64, ::Int64)

julia> transform(df, Cols(:) .=> ByRow(x -> x + 1))
3×6 DataFrame
 Row │ a      b      c      a_function  b_function  c_function 
     │ Int64  Int64  Int64  Int64       Int64       Int64
─────┼─────────────────────────────────────────────────────────
   1 │     1      4      7           2           5           8
   2 │     2      5      8           3           6           9
   3 │     3      6      9           4           7          10

julia> transform(df, 2:3 .=> ByRow(x -> x + 1))
3×5 DataFrame
 Row │ a      b      c      b_function  c_function 
     │ Int64  Int64  Int64  Int64       Int64
─────┼─────────────────────────────────────────────
   1 │     1      4      7           5           8
   2 │     2      5      8           6           9
   3 │     3      6      9           7          10
```

notes.md
nathanrboyer: I guess you are right.
```julia
julia> (; a=3, b=4) == (; a=3, b=4)
true

julia> (; a=3, b=4) == (; b=4, a=3)
false

julia> Dict(:a=>3, :b=>4) == Dict(:a=>3, :b=>4)
true

julia> Dict(:a=>3, :b=>4) == Dict(:b=>4, :a=>3)
true
```

notes.md
bkamins: `(:)` cannot be broadcasted for the same reason as regex - it is defined in Base.

There is no problem with `2:3` because it explicitly specifies which columns to pick (you do not need to know `df` to determine that). While `:` or regrex are dynamic - they can be resolved only in the context of `df` to know what columns are picked. That is why `Cols` wrapper is needed.

PR #2916, opened 2021-10-19
hardening haskey

bkamins:
Follows https://github.com/JuliaLang/julia/pull/42679, as otherwise CI fails


bkamins:
> Better wait a bit to see whether that's the case.

I will merge this as the error messages I introduced are desired anyway (we will throw informative `ArgumentError` message instead of `MethodError` that is probably not easy to understand).

PR #2920, opened 2021-10-27
Add general fast aggregation for wide tables with collect

bkamins:
This PR finishes the issue of wide aggregations I think.
The idea is to special case composition with `collect` so the benefit is that we do not introduce any new syntax to the minilanguage.

With this PR:
```
julia> using DataFrames

julia> using Statistics

julia> df = DataFrame(rand(100, 10_000), :auto);

julia> @time combine(df, AsTable(:) => std∘collect);
  1.174799 seconds (2.79 M allocations: 200.640 MiB, 6.38% gc time, 97.97% compilation time)

julia> @time combine(df, AsTable(:) => std∘collect);
  0.024292 seconds (49.67 k allocations: 35.170 MiB, 38.42% gc time)

julia> @time combine(df, AsTable(:) => ByRow(std∘collect));
  0.394846 seconds (888.92 k allocations: 51.886 MiB, 2.47% gc time, 96.65% compilation time)

julia> @time combine(df, AsTable(:) => ByRow(std∘collect));
  0.013849 seconds (19.17 k allocations: 1.288 MiB)
```

On DataFrames.jl 1.2.2:
```
julia> using DataFrames

julia> using Statistics

julia> df = DataFrame(rand(100, 10_000), :auto);

julia> @time combine(df, AsTable(:) => std∘collect);
  9.210553 seconds (3.98 M allocations: 236.613 MiB, 0.91% gc time, 99.63% compilation time)

julia> @time combine(df, AsTable(:) => std∘collect);
  0.034955 seconds (59.37 k allocations: 36.159 MiB, 32.80% gc time)

julia> @time combine(df, AsTable(:) => ByRow(std∘collect));
 10.378314 seconds (2.54 M allocations: 150.910 MiB, 0.31% gc time, 98.69% compilation time)

julia> @time combine(df, AsTable(:) => ByRow(std∘collect));
  0.099637 seconds (1.02 M allocations: 78.263 MiB, 5.47% gc time)
```

### DIFF
--- a/src/abstractdataframe/selectionfast.jl
+++ b/src/abstractdataframe/selectionfast.jl
@@ -116,8 +116,8 @@ function default_table_transformation(df_sel::AbstractDataFrame, fun)
         end
         v = Vector{T}(undef, ncol(df_sel))
         cols = collect(cT, eachcol(df_sel))
-        reduction = isreadonly(fun.fun.outer)
-        return _fast_row_aggregate_collect(fun.fun.outer, v, cols, reduction)
+        readonly = isreadonly(fun.fun.outer)
+        return _fast_row_aggregate_collect(fun.fun.outer, v, cols, readonly)
     elseif fun isa ComposedFunction{<:Any, typeof(collect)}
         # this will narrow down eltype of the resulting vector
         # but will not perform conversion
@@ -127,18 +127,18 @@ function default_table_transformation(df_sel::AbstractDataFrame, fun)
     end
 end
 
-function _populate_v!(v::Vector, cols::Vector, len::Int, i::Int, reduction::Bool)
+function _populate_v!(v::Vector, cols::Vector, len::Int, i::Int, readonly::Bool)
     for j in 1:len
         @inbounds v[j] = cols[j][i]
     end
-    return reduction ? v : copy(v)
+    return readonly ? v : copy(v)
 end
 
-function _fast_row_aggregate_collect(fun, v::Vector, cols::Vector, reduction::Bool)
+function _fast_row_aggregate_collect(fun, v::Vector, cols::Vector, readonly::Bool)
     len = length(v)
     n = length(cols[1])
     @assert all(x -> length(x) == n, cols)
-    return [fun(_populate_v!(v, cols, len, i, reduction)) for i in 1:n]
+    return [fun(_populate_v!(v, cols, len, i, readonly)) for i in 1:n]
 end
 
 # this is slower than _sum_fast below, but is required if we want
--- a/src/abstractdataframe/selectionfast.jl
+++ b/src/abstractdataframe/selectionfast.jl
@@ -17,7 +17,8 @@ It is guaranteed that `df_sel` has at least one column.
 The main use of special `table_transformation` methods is to provide more
 efficient than the default implementations of requested `fun` transformation.
 
-This function is part of the public API of DataFrames.jl.
+This function might become a part of the public API of DataFrames.jl in the
+future, currently it should be considered experimental.
 
 Fast paths are implemented within DataFrames.jl for the following functions `fun`:
 * `sum`, `ByRow(sum), `ByRow(sum∘skipmissing)`
@@ -56,13 +57,14 @@ table_transformation(df_sel::AbstractDataFrame, fun) =
 """
     isreduction(fun)
 
-Trait returning a `Bool` indicator if function `fun` is a reduction.
-Reduction function guarantees not to modify nor return in any form the passed
-argument. By default it returns `false`.
+Trait returning a `Bool` indicator if function `fun` is a reduction. Reduction
+function guarantees not to modify nor return in any form the passed argument. By
+default it returns `false`.
 
-This function is part of the public API of DataFrames.jl. Adding a method
-to `isreduction` for a specific function `fun` will improve performance
-of `AsTable(...) => ByRow(fun∘collect)` operation.
+This function might become a part of the public API of DataFrames.jl in the
+future, currently it should be considered experimental. Adding a method to
+`isreduction` for a specific function `fun` will improve performance of
+`AsTable(...) => ByRow(fun∘collect)` operation.
 """
 isreduction(::Any) = false
 isreduction(::typeof(sum)) = true
--- a/src/abstractdataframe/selectionfast.jl
+++ b/src/abstractdataframe/selectionfast.jl
@@ -23,6 +23,9 @@ Fast paths are implemented within DataFrames.jl for the following functions `fun
 * `sum`, `ByRow(sum), `ByRow(sum∘skipmissing)`
 * `length`, `ByRow(length)`, `ByRow(length∘skipmissing)`
 * `mean`, `ByRow(mean), `ByRow(mean∘skipmissing)`
+* `ByRow(var), `ByRow(var∘skipmissing)`
+* `ByRow(std), `ByRow(std∘skipmissing)`
+* `ByRow(median), `ByRow(median∘skipmissing)`
 * `minimum`, `ByRow(minimum)`, `ByRow(minimum∘skipmissing)`
 * `maximum`, `ByRow(maximum)`, `ByRow(maximum∘skipmissing)`
 * `fun∘collect` and `ByRow(fun∘collect)` where `fun` is any function
@@ -387,3 +390,18 @@ end
 function table_transformation(df_sel::AbstractDataFrame, ::typeof(maximum))
     return reduce(max, map(identity, eachcol(df_sel)))
 end
+
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(std))) =
+    table_transformation(df_sel, ByRow(std∘collect))
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(std∘skipmissing))) =
+    table_transformation(df_sel, ByRow(std∘skipmissing∘collect))
+
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(var))) =
+    table_transformation(df_sel, ByRow(var∘collect))
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(var∘skipmissing))) =
+    table_transformation(df_sel, ByRow(var∘skipmissing∘collect))
+
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(median))) =
+    table_transformation(df_sel, ByRow(median∘collect))
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(median∘skipmissing))) =
+    table_transformation(df_sel, ByRow(median∘skipmissing∘collect))
--- a/src/abstractdataframe/selectionfast.jl
+++ b/src/abstractdataframe/selectionfast.jl
@@ -55,38 +55,38 @@ table_transformation(df_sel::AbstractDataFrame, fun) =
     default_table_transformation(df_sel, fun)
 
 """
-    isreduction(fun)
+    isreadonly(fun)
 
-Trait returning a `Bool` indicator if function `fun` is a reduction. Reduction
-function guarantees not to modify nor return in any form the passed argument. By
-default it returns `false`.
+Trait returning a `Bool` indicator if function `fun` is only reading the passed
+argument. Such a function guarantees not to modify nor return in any form the
+passed argument. By default `false` is returned.
 
 This function might become a part of the public API of DataFrames.jl in the
 future, currently it should be considered experimental. Adding a method to
-`isreduction` for a specific function `fun` will improve performance of
+`isreadonly` for a specific function `fun` will improve performance of
 `AsTable(...) => ByRow(fun∘collect)` operation.
 """
-isreduction(::Any) = false
-isreduction(::typeof(sum)) = true
-isreduction(::typeof(sum∘skipmissing)) = true
-isreduction(::typeof(length)) = true
-isreduction(::typeof(mean)) = true
-isreduction(::typeof(mean∘skipmissing)) = true
-isreduction(::typeof(var)) = true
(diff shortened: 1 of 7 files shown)


src/abstractdataframe/selectionfast.jl
```diff
@@ -56,8 +60,46 @@ This is a default implementation called when `AsTable(...) => fun` is requested.
...
+        if length(vT) == 1 # homogenous type
+            T = eltype(vT[1])
+            cT = vT[1]
+        elseif length(vT) == 2 # small union
+            T = Union{eltype(vT[1]), eltype(vT[2])}
+            cT = Union{vT[1], vT[2]}
+        else # large union
+            T = Any
+            cT = Any
+        end
```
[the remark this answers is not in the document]
bkamins: OK - I will switch to `promote_typejoin` for the "large union case". The point is that if there are two types then we want `Union` not common type, as common supertype will be abstract, which will be slower than union:
```
julia> Base.promote_typejoin(Int, Float64)
Real

julia> x = Real[rand() for i in 1:10^7];

julia> y = Union{Int, Float64}[rand() for i in 1:10^7];

julia> @time sum(x);
  0.296293 seconds (10.07 M allocations: 157.023 MiB, 10.39% compilation time)

julia> @time sum(x);
  0.520884 seconds (10.00 M allocations: 152.588 MiB, 48.70% gc time)

julia> @time sum(y);
  0.073167 seconds (74.97 k allocations: 4.675 MiB, 67.77% compilation time)

julia> @time sum(y);
  0.024217 seconds (1 allocation: 16 bytes)
```

src/abstractdataframe/selectionfast.jl
```diff
@@ -349,21 +391,17 @@ function table_transformation(df_sel::AbstractDataFrame, ::typeof(maximum))
...
+
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(var))) =
+    table_transformation(df_sel, ByRow(var∘collect))
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(var∘skipmissing))) =
+    table_transformation(df_sel, ByRow(var∘skipmissing∘collect))
+
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(median))) =
+    table_transformation(df_sel, ByRow(median∘collect))
+table_transformation(df_sel::AbstractDataFrame, ::typeof(ByRow(median∘skipmissing))) =
+    table_transformation(df_sel, ByRow(median∘skipmissing∘collect))
```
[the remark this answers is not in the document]
bkamins: `median!` will not work with `skipmissing`.

src/abstractdataframe/selectionfast.jl
```diff
@@ -349,21 +391,17 @@ function table_transformation(df_sel::AbstractDataFrame, ::typeof(maximum))
     return reduce(max, map(identity, eachcol(df_sel)))
 end
 
-# TODO:
-# Add these transformations in the future
```
bkamins, line 353: @nalimilan: the question about removal of this was lost somewhere. I implemented the transformations that seemed useful. For instance I skipped `first` as it did not seem very common to use in case of row-wise aggregation (usually the user knows what the first selected column is). Also - since the PR is now fully general if someone really wants to do `first` one can add `collect` to composed function and have it. I did not want to add a lot of definitions that would not be used in practice.

src/abstractdataframe/selectionfast.jl
```diff
@@ -56,8 +57,46 @@ This is a default implementation called when `AsTable(...) => fun` is requested.
...
+    elseif fun isa ComposedFunction{<:Any, typeof(collect)}
+        # this will narrow down eltype of a resulting vector
+        # but will not perform conversion
+        return fun(map(identity, eachcol(df_sel)))
+    else
+        return fun(Tables.columntable(df_sel))
+    end
+end
+
+function _populate_v(v, cols, len, i)
```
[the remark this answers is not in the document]
bkamins: done

src/abstractdataframe/selectionfast.jl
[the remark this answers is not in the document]
bkamins: I have added `promote_typejoin` for large unions for target `eltype` as it makes sense for some functions (e.g. in StatsBase.jl) since they require `Real` eltype.

bkamins:
@nalimilan This should be good for another round of review. The nightly failure should be fixed in https://github.com/JuliaData/DataFrames.jl/pull/2922.

bkamins:
I have merged main so now all tests pass.

src/abstractdataframe/selectionfast.jl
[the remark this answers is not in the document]
nalimilan: My reasoning was that `collect` in Base calls `promote_typejoin`, so if we want to get exactly the same result we should use that. In theory using a `Union` instead could be problematic, e.g. if input types are `Int8` and `Int16` and the call function tries to assign `typemax(Int16)+1` to the vector it will fail with a `Union{Int8, Int16}` eltype but it will work with a `promote_typejoin(Int8, Int16) === Signed` eltype. So the behavior will be different from what you would expect for `collect`. For `Float16` and `Float32` you could get a lower precision. These are contrived examples, but are we sure we can ignore this kind of thing? That's a tradeoff...

src/abstractdataframe/selectionfast.jl
nalimilan: Yeah that was an outdated comment so I had removed it after posting it by mistake. :-)

src/abstractdataframe/selectionfast.jl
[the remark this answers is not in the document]
bkamins: Right. I was checking:
```
julia> collect(Any[1, 2.0])
2-element Vector{Any}:
 1
 2.0
```
and `promote_typejoin` is not called here, but it is called for `NamedTuple`:
```
julia> collect((a=1, b=2.0))
2-element Vector{Real}:
 1
 2.0
```

I will change the code the way you propose - it is slower but safer. The crucial case with `Missing` is handled correctly.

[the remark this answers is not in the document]
nalimilan: OK. How about adding a test that covers this very special pattern (for both 2 and 3 columns)?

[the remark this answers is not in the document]
bkamins: done

src/abstractdataframe/selectionfast.jl
```diff
@@ -49,15 +53,91 @@ julia> sum(Any[1.0, typemax(Int), typemax(Int)])
...
+    else
+        return fun(Tables.columntable(df_sel))
+    end
+end
+
+function _populate_v!(v::Vector, cols::Vector, len::Int, i::Int, reduction::Bool)
+    for j in 1:len
+        @inbounds v[j] = cols[j][i]
+    end
+    return reduction ? v : copy(v)
```
bkamins, line 132: @nalimilan following your last comment I have made also the following change. I make sure to reuse the working vector for rows only for functions that are reductions (i.e. are guaranteed not to reuse or resize the collected vectors). This makes sure that the following is correct:
```
julia> df = DataFrame(rand(10, 5), :auto)
10×5 DataFrame
 Row │ x1         x2         x3         x4         x5       
     │ Float64    Float64    Float64    Float64    Float64  
─────┼──────────────────────────────────────────────────────
   1 │ 0.569168   0.441972   0.134152   0.0920753  0.935849
   2 │ 0.338254   0.917705   0.447445   0.313503   0.41281
   3 │ 0.0164164  0.936971   0.791789   0.678257   0.551078
   4 │ 0.685475   0.731142   0.405247   0.715096   0.27323
   5 │ 0.658448   0.885822   0.541997   0.796662   0.331429
   6 │ 0.679871   0.806971   0.574006   0.700705   0.374933
   7 │ 0.964579   0.873352   0.485584   0.729399   0.963421
   8 │ 0.0093121  0.478524   0.0092773  0.217818   0.928038
   9 │ 0.124796   0.0379127  0.139757   0.0520611  0.49285
  10 │ 0.0715453  0.275452   0.197599   0.212332   0.930255

julia> combine(df, AsTable(All()) => ByRow(identity∘collect))
10×1 DataFrame
 Row │ x1_x2_etc_identity_collect        
     │ Vector{Float64}
─────┼───────────────────────────────────
   1 │ [0.569168, 0.441972, 0.134152, 0…
   2 │ [0.338254, 0.917705, 0.447445, 0…
   3 │ [0.0164164, 0.936971, 0.791789, …
   4 │ [0.685475, 0.731142, 0.405247, 0…
   5 │ [0.658448, 0.885822, 0.541997, 0…
   6 │ [0.679871, 0.806971, 0.574006, 0…
   7 │ [0.964579, 0.873352, 0.485584, 0…
   8 │ [0.0093121, 0.478524, 0.0092773,…
   9 │ [0.124796, 0.0379127, 0.139757, …
  10 │ [0.0715453, 0.275452, 0.197599, …
```
at the expense of additional copying when the operation is not a reduction.

nalimilan: Actually I had also just realized the same issue. :-) Indeed it sounds safer to restrict to some functions known as safe. 

Though `isreduction` is a bit too specific, right? This is actually more something like `isnonmutating` AFAICT. (This is again a case where being able to declare that a function is pure would be useful.)

src/abstractdataframe/selectionfast.jl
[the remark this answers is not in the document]
nalimilan: Thanks. Do we have tests checking that we get a `Union{Missing, T}` eltype?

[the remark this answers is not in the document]
bkamins: You have them here: https://github.com/JuliaData/DataFrames.jl/pull/2920/files#diff-ba97391c34f47e50271ac08c7d754fc3eed90c8601d0e1de262dfa65ec0bbd3eR2170-R2180

src/abstractdataframe/selectionfast.jl
bkamins: No - this is more than pure. We require some kind of reduction.
The point is that
not only we are not allowed to resize the passed argument
but also we are not allowed to return it in any form (pure function could do this).

For example `identity` is pure, but as a result (without my current fix) it would produce the same vector in all rows (it would contain the last row of the source table in my example above).

This of course extends not only to just returning a vector, but also wrapping it, making a view, etc. - all this disallowed.

nalimilan: Ah right. Though functions which are not reductions at all are still fine as they make copies, e.g. `copy`, `normalize`, `sort`... Not sure how to call this. `iscopying`? `isaliased`? `isinplace`?

bkamins: > Though functions which are not reductions at all are still fine as they make copies

Yes

---

Maybe `is_reading_only` as this is essentially what we require - that the passed vector is only used for reading its contents?

Anyway - I have made this function not part of a public API for now.

nalimilan: Yeah, or maybe just `isreadonly`?

bkamins: OK - changed.

src/abstractdataframe/selectionfast.jl
```diff
@@ -49,15 +54,92 @@ julia> sum(Any[1.0, typemax(Int), typemax(Int)])
...
+            # is required by some functions, eg. in StatsBase.jl
+            T = mapreduce(eltype, Base.promote_typejoin, vT)
+            # use Any for cT, as the common type of columns will be abstract
+            # anyway, so it is better to use Any as it should reduce compilation latency
+            cT = Any
+        end
+        v = Vector{T}(undef, ncol(df_sel))
+        cols = collect(cT, eachcol(df_sel))
+        reduction = isreadonly(fun.fun.outer)
+        return _fast_row_aggregate_collect(fun.fun.outer, v, cols, reduction)
```
nalimilan, line 120:
```suggestion
        readonly = isreadonly(fun.fun.outer)
        return _fast_row_aggregate_collect(fun.fun.outer, v, cols, readonly)
```

src/abstractdataframe/selectionfast.jl
```diff
@@ -49,15 +54,92 @@ julia> sum(Any[1.0, typemax(Int), typemax(Int)])
...
+    else
+        return fun(Tables.columntable(df_sel))
+    end
+end
+
+function _populate_v!(v::Vector, cols::Vector, len::Int, i::Int, reduction::Bool)
+    for j in 1:len
+        @inbounds v[j] = cols[j][i]
+    end
+    return reduction ? v : copy(v)
```
nalimilan, line 134:
```suggestion
function _populate_v!(v::Vector, cols::Vector, len::Int, i::Int, readonly::Bool)
    for j in 1:len
        @inbounds v[j] = cols[j][i]
    end
    return readonly ? v : copy(v)
```

nalimilan (approved):

src/abstractdataframe/selectionfast.jl
```diff
@@ -49,15 +54,92 @@ julia> sum(Any[1.0, typemax(Int), typemax(Int)])
...
+end
+
+function _fast_row_aggregate_collect(fun, v::Vector, cols::Vector, reduction::Bool)
+    len = length(v)
+    n = length(cols[1])
+    @assert all(x -> length(x) == n, cols)
+    return [fun(_populate_v!(v, cols, len, i, reduction)) for i in 1:n]
```
bkamins, line 141:
```suggestion
function _fast_row_aggregate_collect(fun, v::Vector, cols::Vector, readonly::Bool)
    len = length(v)
    n = length(cols[1])
    @assert all(x -> length(x) == n, cols)
    return [fun(_populate_v!(v, cols, len, i, readonly)) for i in 1:n]
```

bkamins:
Thank you!

PR #2923, opened 2021-11-02
Update ci.yml in preparation of Julia 1.6 LTS

bkamins:
Since Julia 1.6 will be a new LTS I propose to add it to our tests. With GitHub Actions this is painless. I would still keep Julia 1.0 in the tests as it does not hurt and it seems we will be able to support it for some time still.


nalimilan (approved):

bkamins:
I will merge this when 1.7 is out. Thank you!

Issue #2924, opened 2021-11-02
subset doesn't accept a vector of transformations


knuesel:
It seems inconsistent that `subset` won't accept a vector of transformations, like `select`, etc.

This came up while writing an answer to [this](https://discourse.julialang.org/t/dataframe-filtering-based-on-conditions-comparing-one-column-to-other-columns/70802) Discourse thread: I was surprised to find that the following doesn't work:

```julia
df = DataFrame(rand(8,5), :auto)
columns = names(df, Not(Between(:x3, ncol(df))))

# This triggers an ArgumentError:
subset(df, vcat.(columns, "x5") .=> .<)
```

so I need instead to write `subset(df, (vcat.(columns, "x5") .=> .<)...)`.

bkamins:
It was intentional - to avoid complicating the things too much and the intention was to require using `...`.
It can be easily added though.
(this is the usual @nalimilan brand - "if you are unsure make it an error so that later you can decide whatever you prefer")

@nalimilan - do you have any opinion here?

nalimilan:
Sure!

bkamins:
@knuesel - please see https://github.com/JuliaData/DataFrames.jl/pull/2926 if it fixes your issues.

knuesel:
@bkamins yes that fixes it, thanks!

PR #2926, opened 2021-11-03
Allow passing multiple columns to subset

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2924

### DIFF
--- a/docs/src/man/working_with_dataframes.md
+++ b/docs/src/man/working_with_dataframes.md
@@ -494,8 +494,8 @@ Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
 functions provided in Julia Base, which also allow subsetting a data frame.
 These methods are defined so that DataFrames.jl implements the Julia API
 for collections, but it is generally recommended to use the [`subset`](@ref)
-and [`subset!`](@ref) functions instead, as they are both faster and
-more consistent with other DataFrames.jl functions.
+and [`subset!`](@ref) functions instead, as they consistent with other
+DataFrames.jl functions (as opposed to [`filter`](@ref) and [`filter!`](@ref)).
 
 ### Selecting and transforming columns
 
--- a/docs/src/man/working_with_dataframes.md
+++ b/docs/src/man/working_with_dataframes.md
@@ -484,14 +484,14 @@ julia> subset(df, :x => x -> iseven.(x), skipmissing=true)
    2 │      4
 ```
 
-The [`subset`](@ref) was designed in a way that is
-consistent how column transformations are specified in functions like
+The [`subset`](@ref) function has been designed in a way that is
+consistent with how column transformations are specified in functions like
 [`combine`](@ref), [`select`](@ref), and [`transform`](@ref). Examples of column
 transformations accepted by these functions are provided in the following
 section.
 
 Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
-functions provided in Julia Base and they also allow to subset a data frame.
+functions provided in Julia Base, which also allow subsetting a data frame.
 The use of these two functions is discouraged as they have an API that is not
 consistent with other transformation functions in DataFrames.jl. They are kept
 in the API for legacy reason, but it is recommended to use the [`subset`](@ref)
--- a/docs/src/man/working_with_dataframes.md
+++ b/docs/src/man/working_with_dataframes.md
@@ -492,7 +492,7 @@ section.
 
 Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
 functions provided in Julia Base, which also allow subsetting a data frame.
-These methods are defined so that data frames implement the Julia API
+These methods are defined so that DataFrames.jl implements the Julia API
 for collections, but it is generally recommended to use the [`subset`](@ref)
 and [`subset!`](@ref) functions instead, as they are both faster and
 more consistent with other DataFrames.jl functions.
--- a/docs/src/man/working_with_dataframes.md
+++ b/docs/src/man/working_with_dataframes.md
@@ -492,10 +492,10 @@ section.
 
 Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
 functions provided in Julia Base, which also allow subsetting a data frame.
-The use of these two functions is discouraged as they have an API that is not
-consistent with other transformation functions in DataFrames.jl. They are kept
-in the API for legacy reason, but it is recommended to use the [`subset`](@ref)
-and [`subset!`](@ref) functions instead.
+These methods are defined so that data frames implement the Julia API
+for collections, but it is generally recommended to use the [`subset`](@ref)
+and [`subset!`](@ref) functions instead, as they are both faster and
+more consistent with other DataFrames.jl functions.
 
 ### Selecting and transforming columns
 
--- a/docs/src/man/working_with_dataframes.md
+++ b/docs/src/man/working_with_dataframes.md
@@ -494,8 +494,8 @@ Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
 functions provided in Julia Base and they also allow to subset a data frame.
 The use of these two functions is discouraged as they have an API that is not
 consistent with other transformation functions in DataFrames.jl. They are kept
-in the API for legacy reason, but it is recommended to use the `subset` and
-`subset!` functions instead.
+in the API for legacy reason, but it is recommended to use the [`subset`](@ref)
+and [`subset!`](@ref) functions instead.
 
 ### Selecting and transforming columns
 
--- a/docs/src/man/working_with_dataframes.md
+++ b/docs/src/man/working_with_dataframes.md
@@ -484,15 +484,18 @@ julia> subset(df, :x => x -> iseven.(x), skipmissing=true)
    2 │      4
 ```
 
-Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
-functions provided in Julia Base and they also allow to subset a data frame.
-Please refer to their documentation for details.
-
-It is worth to mention that the [`subset`](@ref) was designed in a way that is
+The [`subset`](@ref) was designed in a way that is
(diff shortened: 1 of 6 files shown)


nalimilan (approved):

bkamins:
@nalimilan - can you please re-review. I have forgotten to add NEWS.md entry + update documentation. I have also added a clear explanation that `filter` is not recommended and `subset` should be used instead in the documentation.

docs/src/man/working_with_dataframes.md
```diff
@@ -484,16 +484,19 @@ julia> subset(df, :x => x -> iseven.(x), skipmissing=true)
    2 │      4
 ```
 
-Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
-functions provided in Julia Base and they also allow to subset a data frame.
-Please refer to their documentation for details.
-
-It is worth to mention that the [`subset`](@ref) was designed in a way that is
+The [`subset`](@ref) was designed in a way that is
 consistent how column transformations are specified in functions like
```
nalimilan, line 488:
```suggestion
The [`subset`](@ref) function has been designed in a way that is
consistent with how column transformations are specified in functions like
```

docs/src/man/working_with_dataframes.md
```diff
@@ -484,16 +484,19 @@ julia> subset(df, :x => x -> iseven.(x), skipmissing=true)
...
 [`combine`](@ref), [`select`](@ref), and [`transform`](@ref). Examples of column
 transformations accepted by these functions are provided in the following
 section.
 
+Additionally DataFrames.jl extends the [`filter`](@ref) and [`filter!`](@ref)
+functions provided in Julia Base and they also allow to subset a data frame.
```
nalimilan, line 494:
```suggestion
functions provided in Julia Base, which also allow subsetting a data frame.
```

docs/src/man/working_with_dataframes.md
```diff
@@ -484,16 +484,19 @@ julia> subset(df, :x => x -> iseven.(x), skipmissing=true)
...
+The use of these two functions is discouraged as they have an API that is not
+consistent with other transformation functions in DataFrames.jl. They are kept
+in the API for legacy reason, but it is recommended to use the [`subset`](@ref)
+and [`subset!`](@ref) functions instead.
```
nalimilan, line 498: I find this a bit strong: if that was the case we would deprecate them. Maybe something like this?
```suggestion
These methods are defined so that data frames implement the Julia API
for collections, but it is generally recommended to use the [`subset`](@ref)
and [`subset!`](@ref) functions instead, as they are both faster and
more consistent with other DataFrames functions.
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -34,8 +34,6 @@ The following are normally implemented for AbstractDataFrames:
 * [`allowmissing`](@ref) : add support for missing values in columns
 * [`allowmissing!`](@ref) : add support for missing values in columns in-place
 * `similar` : a DataFrame with similar columns as `d`
-* `filter` : remove rows
-* `filter!` : remove rows in-place
```
nalimilan, line 38: Likewise, I'd keep this as these are implemented.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1008,8 +1006,13 @@ end
...
 
-Return a data frame containing only rows from `df` for which `fun`
-returns `true`.
+Return a data frame containing only rows from `df` for which `fun` returns
+`true`.
+
+The use of this function is discouraged as it has API that is inconsistent with
+other transformation functions like [`select`](@ref). Instead it is recommended
+to use the [`subset`](@ref) function. The [`filter`](@ref) function is supported
+to ensure backward compatibility.
```
nalimilan, line 1015: Same here (and below). Mentioning that `subset` is faster should be enough IMO.

src/abstractdataframe/subset.jl
```diff
@@ -95,17 +95,22 @@ end
...
+which all values produced by transformation(s) `args` for a given row are
+`true`. All transformations must produce vectors containing `true` or `false`
+(and optionally `missing` if `skipmissing=true`).
+
+Each argument passed in `args` can be any specifier following the rules
+described for [`select`](@ref) with the restriction that:
+* specifying target column name is not allowed as `subset` does not create new
+  columns;
+* every passed transformation must return a scalar or a vector (returning
+  multi-column values is not supported).
```
nalimilan, line 107: This only excludes `AsTable`, right? Mentioning it would be useful, as "scalar or vector" seems to include all types one may think of.

src/abstractdataframe/abstractdataframe.jl
bkamins: Actually - after thinking about it I would remove this list. It is hard to maintain (contains e.g. `length` now and does not list many functions that are implemented).

src/abstractdataframe/subset.jl
bkamins: This excludes: `AbstractDataFrame`, `NamedTuple`, `DataFrameRow`, `AbstractMatrix`. I will change the description to reflect this.

src/abstractdataframe/abstractdataframe.jl
bkamins: it also contains `convert`.

src/abstractdataframe/abstractdataframe.jl
bkamins: done

src/abstractdataframe/subset.jl
```diff
@@ -84,17 +95,24 @@ end
...
+* specifying target column name is not allowed as `subset` does not create new
+  columns;
+* every passed transformation must return a scalar or a vector (returning
+  `AbstractDataFrame`, `NamedTuple`, `DataFrameRow` or `AbstractMatrix`
+  is not supported).
+
+Note that as the `subset` function works in exactly the same way as other
+transformation functions defined in DataFrames.jl this is a preferred way
+to subset rows of a data frame or grouped data frame. In particular it uses a
+different set of rules for specifying transformations than [`filter!`](@ref)
```
bkamins, line 113:
```suggestion
different set of rules for specifying transformations than [`filter`](@ref)
```

bkamins:
I have updated the docs following your suggestions.

src/abstractdataframe/subset.jl
```diff
@@ -194,13 +212,20 @@ which all values produced by transformation(s) `args` for a given row is `true`.
...
+Each argument passed in `args` can be any specifier following the rules
+described for [`select`](@ref) with the restriction that:
+* specifying target column name is not allowed as `subset!` does not create new
+  columns;
+* every passed transformation must return a scalar or a vector (returning
+  `AbstractDataFrame`, `NamedTuple`, `DataFrameRow` or `AbstractMatrix`
+  is not supported).
+
+Note that as the `subset` function works in exactly the same way as other
+transformation functions defined in DataFrames.jl this is a preferred way to
```
nalimilan, line 224:
```suggestion
transformation functions defined in DataFrames.jl this is the preferred way to
```

src/abstractdataframe/subset.jl
```diff
@@ -84,17 +95,24 @@ end
...
+Each argument passed in `args` can be any specifier following the rules
+described for [`select`](@ref) with the restriction that:
+* specifying target column name is not allowed as `subset` does not create new
+  columns;
+* every passed transformation must return a scalar or a vector (returning
+  `AbstractDataFrame`, `NamedTuple`, `DataFrameRow` or `AbstractMatrix`
+  is not supported).
+
+Note that as the `subset` function works in exactly the same way as other
+transformation functions defined in DataFrames.jl this is a preferred way
```
nalimilan, line 111:
```suggestion
transformation functions defined in DataFrames.jl this is the preferred way
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2,41 +2,14 @@
     AbstractDataFrame
 
 An abstract type for which all concrete types expose an interface
-for working with tabular data.
+for working with tabular data. In DataFrames.jl the `DataFrame` and
+`SubDataFrame` types are subtypes of `AbstractDataFrame`.
 
 # Common methods
```
nalimilan, line 8:
```suggestion
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2,41 +2,14 @@
     AbstractDataFrame
 
 An abstract type for which all concrete types expose an interface
-for working with tabular data.
+for working with tabular data. In DataFrames.jl the `DataFrame` and
+`SubDataFrame` types are subtypes of `AbstractDataFrame`.
```
nalimilan, line 6: I'd put this after the sentence below which gives the overall definition.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -1123,6 +1102,11 @@ _filter_helper_astable(f, nti::Tables.NamedTupleIterator)::BitVector = (x -> f(x
 
 Remove rows from data frame `df` for which `fun` returns `false`.
 
+This method is defined so that DataFrames.jl implements the Julia API for
+collections, but it is generally recommended to use the [`subset!`](@ref)
+function instead, as it is both faster and more consistent with other
```
nalimilan, line 1107: Actually I forgot that you could pass `cols` here to ensure type stability, and `subset` isn't faster in that case, right? Maybe we should only mention consistency?

Also, I'd only mention this at the end of the docstring, possibly in a `!!! note`. Better not add too much noise for people who do want to use that function.

src/abstractdataframe/abstractdataframe.jl
bkamins: done

src/abstractdataframe/abstractdataframe.jl
bkamins: changed all docstrings for `filter` and `subset` and the manual entry.

nalimilan (approved):

bkamins:
Thank you!

PR #2934, opened 2021-11-11
Correctly handle functors when auto-generating column names

bkamins:
An interesting corner-case fix of:
```
julia> struct Identity
       end

julia> function (::Identity)(x)
           return x
       end

julia> i = Identity()
Identity()

julia> df = DataFrame(x=1:3)
3×1 DataFrame
 Row │ x
     │ Int64
─────┼───────
   1 │     1
   2 │     2
   3 │     3

julia> combine(df, :x => identity∘i)
ERROR: MethodError: no method matching nameof(::Identity)
Closest candidates are:
  nameof(::Module) at reflection.jl:16
  nameof(::DataType) at reflection.jl:214
  nameof(::UnionAll) at reflection.jl:215
  ...
Stacktrace:
 [1] funname(f::Identity)
```

### DIFF
--- a/src/other/utils.jl
+++ b/src/other/utils.jl
@@ -120,8 +120,13 @@ function gennames(n::Integer)
 end
 
 function funname(f)
-    n = nameof(f)
-    String(n)[1] == '#' ? :function : n
+    try
+        n = nameof(f)
+        return String(n)[1] == '#' ? :function : n
+    catch
+        # handle the case of functors that do not support nameof
+        return :function
+    end
 end
 
 funname(c::ComposedFunction) = Symbol(funname(c.outer), :_, funname(c.inner))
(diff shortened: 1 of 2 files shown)


bkamins:
Nightly fails because of https://github.com/JuliaLang/julia/pull/42991.

src/other/utils.jl
```diff
@@ -120,8 +120,13 @@ function gennames(n::Integer)
 end
 
 function funname(f)
-    n = nameof(f)
-    String(n)[1] == '#' ? :function : n
+    try
```
nalimilan, line 123: Maybe better use `applicable` rather than `try`?

bkamins: ok - changed

nalimilan (approved):

bkamins:
Thank you!

Issue #2804, opened 2021-06-24
Update the broadcasted getproperty when Julia 1.7 is out


bkamins:
Just to keep track of it. See https://github.com/JuliaData/DataFrames.jl/pull/2655.

nalimilan:
It's difficult to discuss intuitions as they vary a lot depending on previous experience, context, etc. IMO the main advantage of this change would be to have `df.x` be always equivalent to `df[!, :x]`. Given how complex the rules are, it seems important to avoid having three different syntaxes (`df.x`, `df[:, :x]` and `df[!, :x]`).

The downside is of course that we introduce an inconsistency between `df.x .= 1` and `v = df.x; v .= 1`.

Link to previous discussion (credits to @bkamins): https://github.com/JuliaData/DataFrames.jl/pull/2655.

bkamins:
> to have `df.x` be always equivalent to `df[!, :x]`

this is my major personal motivation

bkamins:
The plan is that we will change this behavior in DataFrames.jl 1.4. In 1.3 release an explicit warning message will be printed that the behavior will change in 1.4.

PR #2937, opened 2021-11-14
plan for a change in broadcasting rules in Julia 1.7

bkamins:
This is a transition PR clearly communicating to users what will happen in DataFrames.jl 1.4 release, see https://github.com/JuliaData/DataFrames.jl/issues/2804.

Since the change planned in DataFrames.jl 1.4 will be mildly breaking and will introduce an inconsistency between Julia versions an additional transition period is I think a good approach.

### DIFF
--- a/src/other/broadcasting.jl
+++ b/src/other/broadcasting.jl
@@ -139,7 +139,8 @@ if isdefined(Base, :dotgetproperty)
         else
             # TODO: remove the deprecation in DataFrames.jl 1.4 release
             Base.depwarn("In the 1.4 release of DataFrames.jl this operation will allocate a new column " *
-                         "instead of performing an in-place assignment.", :dotgetproperty)
+                         "instead of performing an in-place assignment. Use `v = df.col; col .= ...` instead.",
+                         :dotgetproperty)
             return getproperty(df, col)
         end
     end
--- a/src/other/broadcasting.jl
+++ b/src/other/broadcasting.jl
@@ -127,8 +127,6 @@ function Base.dotview(df::AbstractDataFrame, ::typeof(!), cols)
     return LazyNewColDataFrame(df, cols isa AbstractString ? Symbol(cols) : cols)
 end
 
-# TODO: remove the deprecations when Julia 1.7 functionality is commonly used
-#       by the community
 if isdefined(Base, :dotgetproperty)
     function Base.dotgetproperty(df::AbstractDataFrame, col::SymbolOrString)
         if columnindex(df, col) == 0
@@ -139,7 +137,8 @@ if isdefined(Base, :dotgetproperty)
             # TODO: double check that this is tested
             return LazyNewColDataFrame(df, Symbol(col))
         else
-            Base.depwarn("In the future this operation will allocate a new column " *
+            # TODO: remove the deprecation in DataFrames.jl 1.4 release
+            Base.depwarn("In the 1.4 release of DataFrames.jl this operation will allocate a new column " *
                          "instead of performing an in-place assignment.", :dotgetproperty)
             return getproperty(df, col)
         end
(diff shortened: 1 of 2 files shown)


bkamins:
For a reference. The current behavior is:
```
julia> df = DataFrame(x=1:3)
3×1 DataFrame
 Row │ x
     │ Int64
─────┼───────
   1 │     1
   2 │     2
   3 │     3

julia> df.x .= 'x'
3-element Vector{Int64}:
 120
 120
 120

julia> df # people typically expect that `:x` column contains 'x' Char values after the previous operation
3×1 DataFrame
 Row │ x
     │ Int64
─────┼───────
   1 │   120
   2 │   120
   3 │   120

julia> df.x .= 1.5 # people typically expect this operation to work
ERROR: InexactError: Int64(1.5)
```
and after the change both operations will produce the expected result described in the comment under Julia 1.7 and newer. Under Julia 1.6 and older they will remain as they are now as it is impossible to fix this behavior there.

src/other/broadcasting.jl
```diff
@@ -139,7 +137,8 @@ if isdefined(Base, :dotgetproperty)
             # TODO: double check that this is tested
             return LazyNewColDataFrame(df, Symbol(col))
         else
-            Base.depwarn("In the future this operation will allocate a new column " *
+            # TODO: remove the deprecation in DataFrames.jl 1.4 release
+            Base.depwarn("In the 1.4 release of DataFrames.jl this operation will allocate a new column " *
                          "instead of performing an in-place assignment.", :dotgetproperty)
```
nalimilan, line 142:
```suggestion
            Base.depwarn("In the 1.4 release of DataFrames.jl this operation will allocate a new column " *
                         "instead of performing an in-place assignment. Use `v = df.col; col .= ...` instead.",
                         :dotgetproperty)
```

nalimilan (approved):

src/other/broadcasting.jl
```diff
@@ -139,8 +137,10 @@ if isdefined(Base, :dotgetproperty)
             # TODO: double check that this is tested
             return LazyNewColDataFrame(df, Symbol(col))
         else
-            Base.depwarn("In the future this operation will allocate a new column " *
-                         "instead of performing an in-place assignment.", :dotgetproperty)
+            # TODO: remove the deprecation in DataFrames.jl 1.4 release
+            Base.depwarn("In the 1.4 release of DataFrames.jl this operation will allocate a new column " *
+                         "instead of performing an in-place assignment. Use `v = df.col; col .= ...` instead.",
```
bkamins, line 142:
```suggestion
                         "instead of performing an in-place assignment. "
                         "To perform an in-place assignment `df[:, col] .= ...` instead.",
```

src/other/broadcasting.jl
```diff
@@ -139,8 +137,11 @@ if isdefined(Base, :dotgetproperty)
             # TODO: double check that this is tested
             return LazyNewColDataFrame(df, Symbol(col))
         else
-            Base.depwarn("In the future this operation will allocate a new column " *
-                         "instead of performing an in-place assignment.", :dotgetproperty)
+            # TODO: remove the deprecation in DataFrames.jl 1.4 release
+            Base.depwarn("In the 1.4 release of DataFrames.jl this operation will allocate a new column " *
+                         "instead of performing an in-place assignment. "
+                         "To perform an in-place assignment `df[:, col] .= ...` instead.",
```
bkamins, line 143:
```suggestion
                         "To perform an in-place assignment use `df[:, col] .= ...` instead.",
```

src/other/broadcasting.jl
```diff
@@ -139,8 +137,11 @@ if isdefined(Base, :dotgetproperty)
             # TODO: double check that this is tested
             return LazyNewColDataFrame(df, Symbol(col))
         else
-            Base.depwarn("In the future this operation will allocate a new column " *
-                         "instead of performing an in-place assignment.", :dotgetproperty)
+            # TODO: remove the deprecation in DataFrames.jl 1.4 release
+            Base.depwarn("In the 1.4 release of DataFrames.jl this operation will allocate a new column " *
+                         "instead of performing an in-place assignment. "
+                         "To perform an in-place assignment use `df[:, col] .= ...` instead.",
```
bkamins, line 143: this is what I would recommend.

nalimilan (approved):

bkamins:
Thank you!

PR #2938, opened 2021-11-15
Change join tests to reduce memory consumption

bkamins:
CI on 32-bit machine very often fails due to out of memory error.
This PR aims to reduce memory footprint of the most problematic testset.


nalimilan (approved):

bkamins:
Thank you!

PR #2941, opened 2021-11-19
Add examples for `issorted` docstrings.

Chandu-4444:
I have added a few examples in `issorted` docstrings. And mentioned about lexicographic sorting at `issorted` and `sortperm`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -112,7 +112,7 @@
 
 * `delete!` is deprecated in favor of `deleteat!`
   ([#2854](https://github.com/JuliaData/DataFrames.jl/issues/2854))
-* In `sort`, `sort!`, `issorted` and `soerperm` it is now documented
+* In `sort`, `sort!`, `issorted` and `sortperm` it is now documented
   that the result of passing an empty column selector uses lexicographic
   ordering of all columns, but this behavior is deprecated.
   ([#2941](https://github.com/JuliaData/DataFrames.jl/issues/2941))
--- a/NEWS.md
+++ b/NEWS.md
@@ -113,7 +113,8 @@
 * `delete!` is deprecated in favor of `deleteat!`
   ([#2854](https://github.com/JuliaData/DataFrames.jl/issues/2854))
 * In `sort`, `sort!`, `issorted` and `soerperm` it is now documented
-  that passing an empty column selector leads to an undefined result
+  that the result of passing an empty column selector uses lexicographic
+  ordering of all columns, but this behavior is deprecated.
   ([#2941](https://github.com/JuliaData/DataFrames.jl/issues/2941))
 
 ## Planned changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -112,6 +112,9 @@
 
 * `delete!` is deprecated in favor of `deleteat!`
   ([#2854](https://github.com/JuliaData/DataFrames.jl/issues/2854))
+* In `sort`, `sort!`, `issorted` and `soerperm` it is now documented
+  that passing an empty column selector leads to an undefined result
+  ([#2941](https://github.com/JuliaData/DataFrames.jl/issues/2941))
 
 ## Planned changes
 
--- a/src/abstractdataframe/sort.jl
+++ b/src/abstractdataframe/sort.jl
@@ -351,7 +351,7 @@ See other methods for a description of other keyword arguments.
 
 # Examples
 ```jldoctest
-julia> df = DataFrame(a = [1,2,3,4], b = [4,3,2,1])
+julia> df = DataFrame(a = [1, 2, 3, 4], b = [4, 3, 2, 1])
 4×2 DataFrame
  Row │ a      b     
      │ Int64  Int64 
@@ -480,7 +480,9 @@ end
 
 Return a permutation vector of row indices of data frame `df` that puts them in
 sorted order according to column(s) `cols`.
-If `cols` were not specified, it returns row indices that sorts `df` lexicographically
+Order on multiple columns is done lexicographically.
+
+If `cols` select no columns, it returns row indices that sorts `df` lexicographically
 (this behaviour is deprecated and will change in future versions).
 
 `cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
--- a/src/abstractdataframe/sort.jl
+++ b/src/abstractdataframe/sort.jl
@@ -332,7 +332,7 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
 ########################
 
 """
-    issorted(df::AbstractDataFrame, cols;
+    issorted(df::AbstractDataFrame, cols=All();
              lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
 
 Test whether data frame `df` sorted by column(s) `cols`.
@@ -371,7 +371,7 @@ julia> issorted(df, :b, rev=true)
 true
 ```
 """
-function Base.issorted(df::AbstractDataFrame, cols=[];
+function Base.issorted(df::AbstractDataFrame, cols=All();
                        lt=isless, by=identity, rev=false, order=Forward)
     # exclude AbstractVector as in that case cols can contain order(...) clauses
     if cols isa MultiColumnIndex && !(cols isa AbstractVector)
@@ -387,7 +387,7 @@ function Base.issorted(df::AbstractDataFrame, cols=[];
 end
 
 """
-    sort(df::AbstractDataFrame, cols;
+    sort(df::AbstractDataFrame, cols=All();
          alg::Union{Algorithm, Nothing}=nothing, lt=isless, by=identity,
          rev::Bool=false, order::Ordering=Forward, view::Bool=false)
 
@@ -463,14 +463,14 @@ julia> sort(df, [:x, order(:y, rev=true)])
    4 │     3  b
 ```
 """
-@inline function Base.sort(df::AbstractDataFrame, cols=[]; alg=nothing, lt=isless,
+@inline function Base.sort(df::AbstractDataFrame, cols=All(); alg=nothing, lt=isless,
                            by=identity, rev=false, order=Forward, view::Bool=false)
     rowidxs = sortperm(df, cols, alg=alg, lt=lt, by=by, rev=rev, order=order)
     return view ? Base.view(df, rowidxs, :) : df[rowidxs, :]
 end
 
 """
-    sortperm(df::AbstractDataFrame, cols;
+    sortperm(df::AbstractDataFrame, cols=All();
              alg::Union{Algorithm, Nothing}=nothing, lt=isless, by=identity,
              rev::Bool=false, order::Ordering=Forward)
 
@@ -532,7 +532,7 @@ julia> sortperm(df, [:x, order(:y, rev=true)])
  1
 ```
 """
-function Base.sortperm(df::AbstractDataFrame, cols=[];
+function Base.sortperm(df::AbstractDataFrame, cols=All();
                   alg=nothing, lt=isless, by=identity, rev=false, order=Forward)
     if !(by isa Base.Callable || (by isa AbstractVector && eltype(by) <: Base.Callable))
         msg = "'by' must be a Function or a vector of Functions. " *
--- a/src/abstractdataframe/sort.jl
+++ b/src/abstractdataframe/sort.jl
@@ -483,13 +483,11 @@ end
 
 Return a permutation vector of row indices of data frame `df` that puts them in
 sorted order according to column(s) `cols`.
-Order on multiple columns is done lexicographically.
-
-If `cols` select no columns, it returns row indices that sorts `df` lexicographically
-(this behaviour is deprecated and will change in future versions).
+Order on multiple columns is computed lexicographically.
 
 `cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
-If `cols` selects no columns, return permutation vector based on sorting all columns.
+If `cols` selects no columns, return permutation vector based on sorting all columns
+(this behaviour is deprecated and will change in future versions).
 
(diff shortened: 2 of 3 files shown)


src/abstractdataframe/sort.jl
```diff
@@ -336,13 +336,40 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
              lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
 
 Test whether data frame `df` sorted by column(s) `cols`.
+If no column is specified, it checks if the `df` is sorted lexicographically.
```
bkamins, line 339: This is an important comment. Can you please add it also to `sort` and `sort!`. Also can you make it more precise that "no column is specified" means "column selector selects no columns".

@nalimilan - this rule is non-intuitive, but changing it would be breaking. Should we deprecate this behavior for 2.0 release (it would be a separate PR).

src/abstractdataframe/sort.jl
```diff
@@ -336,13 +336,40 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
...
+julia> issorted(df)
+true
+
+julia> issorted(df, :a)
+true
+
+julia> issorted(df, :b)
+false
+
+julia> issorted(df, :b, rev = true)
```
bkamins, line 370: the style is not to put spaces around `=` when passing kwargs
```suggestion
julia> issorted(df, :b, rev=true)
```

src/abstractdataframe/sort.jl
bkamins: @Chandu-4444 - if what I write is not clear please comment and I will give you examples.

bkamins: > Can you please elaborate?

`sort` and related functions were developed in very early days of DataFrames.jl when there was no precise rules for the whole ecosystem. Therefore some inconsistencies slipped through.

A proper definition of `sort` should be (I am skipping kwargs):
```
Base.sort(df::AbstractDataFrame, cols=All())
```
however it is
```
Base.sort(df::AbstractDataFrame, cols=[])
```
and as a consequence if you pass any column selector that selects no columns the data frame is sorted lexicographically.

While technically it should happen if you select `All()` columns. If you select no columns a proper behavior should be that the data frame is not reordered (it is left as is as you have not requested on what the data frame should be sorted).

We have two options:
* leave things as is forever (probably there is not much harm with that as it is unlikely that someone does sorting and wants the data frame to be left untouched)
* make a deprecation of how selecting no columns is handled and in 2.0 release change it the way I describe above.

Chandu-4444: I get it now, thanks for that excellent explanation. Perhaps this column selector can be made a required argument (Like the one Python has in Pandas)?

bkamins: > Perhaps this column selector can be made a required argument

This would be even more breaking. By proposing the change to:
```
Base.sort(df::AbstractDataFrame, cols=All())
```
we are not breaking calls like `sort(df)` which are quite likely to exist in current user code.

We try to limit breaking changes. Calls like `sort(df, [])` are quite unlikely, so we could deprecate them and change their behavior in 2.0.

Let us wait for an opinion from @nalimilan.

nalimilan: That's indeed unfortunate. I'm not sure we should make a breaking change though given that it's not super likely to confuse people (if you call `sort` with no columns, it's most likely a mistake anyway and you'll get unexpected results), so I'd opt for a deprecation.

src/abstractdataframe/sort.jl
```diff
@@ -336,13 +336,40 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
              lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
 
 Test whether data frame `df` sorted by column(s) `cols`.
+If column selector selects no columns, it checks if the `df` is sorted lexicographically.
```
nalimilan, line 339:
```suggestion
If `cols` selects no columns, check whether `df` is sorted on all columns.
```
I'd also put this below the next sentence to keep the first description as simple as possible.

src/abstractdataframe/sort.jl
```diff
@@ -368,6 +395,7 @@ Return a data frame containing the rows in `df` sorted by column(s) `cols`.
 Sorting on multiple columns is done lexicographically.
 
 `cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+If column selector selects no columns, then the sorting would be done lexicographically.
```
nalimilan, line 398:
```suggestion
If `cols` selects no columns, sort `df` on all columns.
```

src/abstractdataframe/sort.jl
```diff
@@ -448,8 +476,10 @@ end
 
 Return a permutation vector of row indices of data frame `df` that puts them in
 sorted order according to column(s) `cols`.
+If `cols` were not specified, it returns row indices that sorts `df` lexicographically.
 
 `cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+If column selector selects no columns, it returns row indices that sorts `df` lexicographically.
```
nalimilan, line 482:
```suggestion
If `cols` selects no columns, return permutation vector based on sorting all columns.
```

src/dataframe/sort.jl
```diff
@@ -8,6 +8,7 @@ Sort data frame `df` by column(s) `cols`.
 Sorting on multiple columns is done lexicographicallly.
 
 `cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+If column selector selects no columns, it sorts the `df` lexicographically.
```
nalimilan, line 11:
```suggestion
If `cols` selects no columns, sort `df` on all columns.
```

src/abstractdataframe/sort.jl
```diff
@@ -448,8 +476,10 @@ end
 
 Return a permutation vector of row indices of data frame `df` that puts them in
 sorted order according to column(s) `cols`.
+If `cols` were not specified, it returns row indices that sorts `df` lexicographically.
```
nalimilan, line 479: Rather than repeating this here (which would have to be put in all sorting functions), how about putting `cols=All()` in the signature above? We could also change the actual implementation to make it clearer.

src/abstractdataframe/sort.jl
bkamins: OK - let us then document that passing no columns is an undefined behavior.

src/abstractdataframe/sort.jl
nalimilan: Ah, if the idea was to mention that sorting is done lexicographically, this could be mentioned here, but without referring to the case were `cols` selects no columns, as it applies more generally. It would also make sense to explain what "lexicographically" means exactly as not everybody is familiar with this term.

src/abstractdataframe/sort.jl
bkamins: This is what I want to do. I will make these changes to this PR and then let @Chandu-4444 to sync the docs following your suggestions and writing that passing no columns leads to an undefined row order.

bkamins:
@Chandu-4444 - I have updated the method signatures and added NEWS.md entry following our discussion. Now can you please sync the documentation accordingly? Thank you. I am marking it as 1.3 release PR (we wait for Julia 1.7 anyway and this should be quick to merge)

Chandu-4444:
> If `cols` were not specified, it returns row indices that sorts `df` lexicographically.

Changing this to the line below would be fine I guess (In all the docstrings)?:

> Passing an empty columns selector leads to an undefined result.

nalimilan:
Rather than saying "undefined", wouldn't it be better to say what happens now and that this will change in the future? That's less scary and more useful.

Chandu-4444:
> Rather than saying "undefined", wouldn't it be better to say what happens now and that this will change in the future? That's less scary and more useful.

Yeah! Like this? :
> If `cols` were not specified, it returns row indices that sorts `df` lexicographically. 
> This behaviour will change from version 1.3.

nalimilan:
Well the exact wording will depend on the function (cf. the suggestions I made above). So I'd just add "(this behavior is deprecated and will change in future versions)" to each of them.

NEWS.md
```diff
@@ -112,6 +112,9 @@
 
 * `delete!` is deprecated in favor of `deleteat!`
   ([#2854](https://github.com/JuliaData/DataFrames.jl/issues/2854))
+* In `sort`, `sort!`, `issorted` and `soerperm` it is now documented
+  that passing an empty column selector leads to an undefined result
```
bkamins, line 116:
```suggestion
  that the result of passing an empty column selector uses lexicographic
  ordering of all columns, but this behavior is deprecated.
```

src/abstractdataframe/sort.jl
```diff
@@ -332,19 +332,47 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
 ########################
 
 """
-    issorted(df::AbstractDataFrame, cols;
+    issorted(df::AbstractDataFrame, cols=All();
              lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
 
 Test whether data frame `df` sorted by column(s) `cols`.
```
bkamins, line 338:
```suggestion
Test whether data frame `df` sorted by column(s) `cols`.
Checking against multiple columns is done lexicographically.

```

src/abstractdataframe/sort.jl
```diff
@@ -332,19 +332,49 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
...
 
 If `rev` is `true`, reverse sorting is performed. To enable reverse sorting
 only for some columns, pass `order(c, rev=true)` in `cols`, with `c` the
 corresponding column index (see example below).
+
 See other methods for a description of other keyword arguments.
+
+# Examples
+```jldoctest
+julia> df = DataFrame(a = [1,2,3,4], b = [4,3,2,1])
```
bkamins, line 354:
```suggestion
julia> df = DataFrame(a = [1, 2, 3, 4], b = [4, 3, 2, 1])
```

src/abstractdataframe/sort.jl
```diff
@@ -435,21 +467,24 @@ julia> sort(df, [:x, order(:y, rev=true)])
...
 end
 
 """
-    sortperm(df::AbstractDataFrame, cols;
+    sortperm(df::AbstractDataFrame, cols=All();
              alg::Union{Algorithm, Nothing}=nothing, lt=isless, by=identity,
              rev::Bool=false, order::Ordering=Forward)
 
 Return a permutation vector of row indices of data frame `df` that puts them in
 sorted order according to column(s) `cols`.
```
bkamins, line 482:
```suggestion
sorted order according to column(s) `cols`.
Order on multiple columns is done lexicographically.

```

src/abstractdataframe/sort.jl
```diff
@@ -435,21 +467,24 @@ julia> sort(df, [:x, order(:y, rev=true)])
...
+If `cols` were not specified, it returns row indices that sorts `df` lexicographically
```
bkamins, line 483:
```suggestion
If `cols` select no columns, it returns row indices that sorts `df` lexicographically
```

src/abstractdataframe/sort.jl
```diff
@@ -332,19 +332,49 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
...
 
 """
-    issorted(df::AbstractDataFrame, cols;
+    issorted(df::AbstractDataFrame, cols=All();
              lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
 
 Test whether data frame `df` sorted by column(s) `cols`.
+Checking against multiple columns is done lexicographically.
+
+If `cols` selects no columns, check whether `df` is sorted on all columns
```
bkamins, line 341:
```suggestion
`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
If `cols` selects no columns, check whether `df` is sorted on all columns
```

bkamins:
I have reviewed the PR pushing appropriate changes (in particular introducing deprecation warning) and making them consistent across functions. Can you please have a look if all is OK? Thank you!

src/abstractdataframe/sort.jl
```diff
@@ -435,21 +470,26 @@ julia> sort(df, [:x, order(:y, rev=true)])
...
 
 """
-    sortperm(df::AbstractDataFrame, cols;
+    sortperm(df::AbstractDataFrame, cols=All();
              alg::Union{Algorithm, Nothing}=nothing, lt=isless, by=identity,
              rev::Bool=false, order::Ordering=Forward)
 
 Return a permutation vector of row indices of data frame `df` that puts them in
 sorted order according to column(s) `cols`.
+Order on multiple columns is done lexicographically.
```
nalimilan, line 486:
```suggestion
Order on multiple columns is computed lexicographically.
```

src/abstractdataframe/sort.jl
```diff
@@ -435,21 +470,26 @@ julia> sort(df, [:x, order(:y, rev=true)])
...
+
+If `cols` select no columns, it returns row indices that sorts `df` lexicographically
+(this behaviour is deprecated and will change in future versions).
 
 `cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+If `cols` selects no columns, return permutation vector based on sorting all columns.
```
nalimilan, line 492:
```suggestion
`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
If `cols` selects no columns, return permutation vector based on sorting all columns
(this behaviour is deprecated and will change in future versions).
```

nalimilan (approved):

bkamins:
Thank you!

Issue #2438, opened 2020-09-18
Add reverse and reverse! functions similar to sort and sort!


ppalmes:
- there are cases where you just want to reverse the order of dataframe and currently there is no implementation of the reverse() function

Chandu-4444:
I'm trying to implement this. I wrote a function named `reverse` and put that in `src/abstractdataframe/reverse.jl`. But when I try to use my function, it didn't get recognized. But I placed the same function in sort.jl and it worked only that the docstrings I wrote weren't recognized (And they didn't show up during `?reverse`). Is there somekind of procedure to create new functions in a module?

Chandu-4444:
> I'm trying to implement this. I wrote a function named `reverse` and put that in `src/abstractdataframe/reverse.jl`. But when I try to use my function, it didn't get recognized. But I placed the same function in sort.jl and it worked only that the docstrings I wrote weren't recognized (And they didn't show up during `?reverse`). Is there somekind of procedure to create new functions in a module?

I figured out that I need to add the new files that hold usable functions in `ModuleName.jl` file. This works fine, and now my function is working. But I didn't understand why my docstrings aren't getting shown when I use `?reverse`.

bkamins:
added in 1.4 release

Issue #2936, opened 2021-11-14
add keepat! and insert! and pushfirst!


bkamins:
In the future consider adding `keepat!` support (introduced in Julia 1.7).

Chandu-4444:
Can I try adding `insert!`? Can you tell me how and where to start writing this function?

bkamins:
I think `insert!` is hard to start with (unless you are really experienced with Julia).

You would need to use:

* https://github.com/JuliaData/DataFrames.jl/blob/main/src/dataframe/dataframe.jl#L1416
* https://github.com/JuliaData/DataFrames.jl/blob/main/src/dataframe/dataframe.jl#L1666
* https://github.com/JuliaData/DataFrames.jl/blob/main/src/dataframerow/dataframerow.jl#L501

and define similar `insert!` three methods for `insert!`. I encourage to review this source code and decide, but you will see that it is not an easy code and `insert!` will be a bit more complex than `push!` (not much but a still).

`keepat!` should be much easier as you just need to call `deleteat!` with an inverse selector (at least in the easiest scenario, i.e. not pushing performance to the limits). The only thing to be careful about is that `keepat!` will either need to be imported from `Base` or defined internally and exported depending on Julia version the user is on.

bkamins:
If you are looking for easy PRs I would recommend:
* https://github.com/JuliaData/DataFrames.jl/issues/2845
* https://github.com/JuliaData/DataFrames.jl/issues/2443
* https://github.com/JuliaData/DataFrames.jl/issues/2438

PR #2944, opened 2021-11-23
Add  `reverse` prototype

Chandu-4444:
I tried to make a sample `reverse` function for [#2438](https://github.com/JuliaData/DataFrames.jl/issues/2438). Please look at this and suggest to me the improvements. I'll add those suggestions and will make `reverse!` too. Finally I'll add tests to these functions.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -417,7 +417,7 @@
   with `ByRow` is now accepted ([#2476](https://github.com/JuliaData/DataFrames.jl/pull/2476))
 * add `permutedims` method for `AbstractDataFrame` ([#2447](https://github.com/JuliaData/DataFrames.jl/pull/2447))
 * add support for `Cols` from DataAPI.jl ([#2495](https://github.com/JuliaData/DataFrames.jl/pull/2495))
-* add `reverse` function for `DataFrame` that reverses the rows
+* add `reverse` function for `AbstractDataFrame` that reverses the rows
   ([#2944](https://github.com/JuliaData/DataFrames.jl/pull/2944))
 
 ## Deprecated
--- a/NEWS.md
+++ b/NEWS.md
@@ -417,6 +417,8 @@
   with `ByRow` is now accepted ([#2476](https://github.com/JuliaData/DataFrames.jl/pull/2476))
 * add `permutedims` method for `AbstractDataFrame` ([#2447](https://github.com/JuliaData/DataFrames.jl/pull/2447))
 * add support for `Cols` from DataAPI.jl ([#2495](https://github.com/JuliaData/DataFrames.jl/pull/2495))
+* add `reverse` function for `DataFrame` that reverses the rows
+  ([#2944](https://github.com/JuliaData/DataFrames.jl/pull/2944))
 
 ## Deprecated
 
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -2250,30 +2250,27 @@ This function reverses the current order of the rows in `df`.
 ```jldoctest
 julia> df = DataFrame(a = 1:5, b = 6:10, c = 11:15)
 5×3 DataFrame
- Row │ a      b      c     
-     │ Int64  Int64  Int64 
+ Row │ a      b      c 
+     │ Int64  Int64  Int64
 ─────┼─────────────────────
    1 │     1      6     11
    2 │     2      7     12
    3 │     3      8     13
    4 │     4      9     14
    5 │     5     10     15
-   
+
 julia> reverse(df)
 5×3 DataFrame
-Row │ a      b      c     
-    │ Int64  Int64  Int64 
-────┼─────────────────────
-  1 │     5     10     15
-  2 │     4      9     14
-  3 │     3      8     13
-  4 │     2      7     12
-  5 │     1      6     11
+ Row │ a      b      c
+     │ Int64  Int64  Int64
+─────┼─────────────────────
+   1 │     5     10     15
+   2 │     4      9     14
+   3 │     3      8     13
+   4 │     2      7     12
+   5 │     1      6     11
 ```
 """
 function Base.reverse(df::AbstractDataFrame)
-    for column in 1:size(df, 2)
-        df[!, column] = Base.reverse(df[!, column])
-    end
-    return df
+    return df[nrow(df):-1:1, :]
 end
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -2248,7 +2248,7 @@ This function reverses the current order of the rows in `df`.
 # Examples
 
 ```jldoctest
-julia> df = DataFrame(a = 1:5, b = 6:10, c = 11:15)
+julia> df = DataFrame(a=1:5, b=6:10, c=11:15)
 5×3 DataFrame
  Row │ a      b      c 
      │ Int64  Int64  Int64
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -552,7 +552,7 @@ access missing values.
 
 # Examples
 ```jldoctest
-julia> df = DataFrame(i=1:10, x=0.1:0.1:1.0, y='a':'j');
+julia> df = DataFrame(i=1:10, x=0.1:0.1:1.0, y='a':'j')
 
 julia> describe(df)
 3×7 DataFrame
@@ -759,9 +759,9 @@ Use `findall(completecases(df))` to get the indices of the rows.
 # Examples
 
 ```jldoctest
-julia> df = DataFrame(i = 1:5,
-                            x = [missing, 4, missing, 2, 1],
-                            y = [missing, missing, "c", "d", "e"])
+julia> df = DataFrame(i=1:5,
+                        x=[missing, 4, missing, 2, 1],
+                        y=[missing, missing, "c", "d", "e"])
 5×3 DataFrame
  Row │ i      x        y
      │ Int64  Int64?   String?
@@ -850,9 +850,9 @@ See also: [`completecases`](@ref) and [`dropmissing!`](@ref).
 # Examples
 
 ```jldoctest
-julia> df = DataFrame(i = 1:5,
-                      x = [missing, 4, missing, 2, 1],
-                      y = [missing, missing, "c", "d", "e"])
+julia> df = DataFrame(i=1:5,
+                      x=[missing, 4, missing, 2, 1],
+                      y=[missing, missing, "c", "d", "e"])
 5×3 DataFrame
  Row │ i      x        y
      │ Int64  Int64?   String?
@@ -927,9 +927,9 @@ get converted using [`disallowmissing!`](@ref).
 See also: [`dropmissing`](@ref) and [`completecases`](@ref).
 
 ```jldoctest
-julia> df = DataFrame(i = 1:5,
-                      x = [missing, 4, missing, 2, 1],
-                      y = [missing, missing, "c", "d", "e"])
+julia> df = DataFrame(i=1:5,
+                      x=[missing, 4, missing, 2, 1],
+                      y=[missing, missing, "c", "d", "e"])
 5×3 DataFrame
  Row │ i      x        y
      │ Int64  Int64?   String?
@@ -1018,7 +1018,7 @@ See also: [`filter!`](@ref)
 
 # Examples
 ```jldoctest
-julia> df = DataFrame(x = [3, 1, 2, 1], y = ["b", "c", "a", "b"])
+julia> df = DataFrame(x=[3, 1, 2, 1], y=["b", "c", "a", "b"])
 4×2 DataFrame
  Row │ x      y
      │ Int64  String
@@ -1144,7 +1144,7 @@ See also: [`filter`](@ref)
 
 # Examples
 ```jldoctest
-julia> df = DataFrame(x = [3, 1, 2, 1], y = ["b", "c", "a", "b"])
+julia> df = DataFrame(x=[3, 1, 2, 1], y=["b", "c", "a", "b"])
 4×2 DataFrame
  Row │ x      y
      │ Int64  String
@@ -1177,7 +1177,7 @@ julia> filter!(:x => x -> x == 3, df)
 ─────┼───────────────
    1 │     3  b
 
-julia> df = DataFrame(x = [3, 1, 2, 1], y = ["b", "c", "a", "b"]);
+julia> df = DataFrame(x=[3, 1, 2, 1], y=["b", "c", "a", "b"]);
 
 julia> filter!([:x, :y] => (x, y) -> x == 1 || y == "b", df)
 3×2 DataFrame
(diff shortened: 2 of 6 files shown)


bkamins:
I think that this function is short enough to include it just in `abstractdataframe/abstractdataframe.jl`

Chandu-4444:
> I think that this function is short enough to include it just in `abstractdataframe/abstractdataframe.jl`

Sure! And I think I need to include `reverse!` in the same file right?

bkamins:
No, because `reverse`. The only trick is that `reverse!` should be something like:
```
function reverse!(df::AbstractDataFrame)
    foreach(col -> reverse!(col), eachcol(df))
    return df
end
```

The only issue is if we want to handle the case that some of the columns in `df` are read-only, but I think it is OK to just throw an error in the middle (it might render the `df` corrupted though). @nalimilan - what do you think?

When you write tests please make sure to test both `DataFrame` and `SubDataFrame`.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2239,3 +2239,41 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
 
 Base.setindex!(::AbstractDataFrame, ::Any, ::Union{Symbol, Integer, AbstractString}) =
     throw(ArgumentError("syntax df[column] is not supported use df[!, column] instead"))
+
+"""
+    reverse(df::AbstractDataFrame)
+
+This function reverses the current order of the rows in `df`.
```
bkamins, line 2246: please turn on trimming of trailing whitespace in your editor.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2239,3 +2239,41 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
...
+  1 │     5     10     15
+  2 │     4      9     14
+  3 │     3      8     13
+  4 │     2      7     12
+  5 │     1      6     11
+```
+"""
+function Base.reverse(df::AbstractDataFrame)
+    for column in 1:size(df, 2)
+        df[!, column] = Base.reverse(df[!, column])
```
bkamins, line 2276: as commented earlier this implementation is incorrect

src/abstractdataframe/abstractdataframe.jl
bkamins, line 2246: This function should create a new data frame with reversed rows. `reverse!` would reverse rows in the passed data frame in-place.

test/dataframe.jl
```diff
@@ -2032,4 +2032,15 @@ end
...
 end
 
+@testset "reverse --DataFrame" begin
+    df = DataFrame(a = 1:5, b = 5:-1:1)
+    @test reverse(df) == DataFrame(a = 5:-1:1, b = 1:5)
+    @test reverse(DataFrame(a = 1, b = 1)) == DataFrame(a = 1, b = 1)
+end
+
+@testset "reverse --SubDataFrame" begin
+    df = view(DataFrame(a = 1:5, b = 5:-1:1, c = 11:15), 1:3, :)
```
bkamins, line 2042: please also test if the result is a `DataFrame` and add a test when view is created with other column selector than `:`, as `:` is a special case in views.

bkamins:
Ah - and I have forgotten that we need NEWS.md entry after we are done adding new things.

Chandu-4444:
Sure, I'll do these!

Chandu-4444:
I've done adding `reverse`. I'll continue with `reverse!` next. But I'm sure about handling read-only columns. May be check for read-only columns prior to processing `df`?

bkamins:
Let us do `reverse!` in a separate PR and discuss it there with @nalimilan. Also have a look at https://github.com/JuliaData/DataFrames.jl/pull/2946/files to see how I think aliasing in `reverse!` should be handled (`sort!` is more general than reverse so the same issues apply)

NEWS.md
```diff
@@ -417,6 +417,8 @@
   with `ByRow` is now accepted ([#2476](https://github.com/JuliaData/DataFrames.jl/pull/2476))
 * add `permutedims` method for `AbstractDataFrame` ([#2447](https://github.com/JuliaData/DataFrames.jl/pull/2447))
 * add support for `Cols` from DataAPI.jl ([#2495](https://github.com/JuliaData/DataFrames.jl/pull/2495))
+* add `reverse` function for `DataFrame` that reverses the rows
```
bkamins, line 420:
```suggestion
* add `reverse` function for `AbstractDataFrame` that reverses the rows
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2239,3 +2239,38 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
...
+   1 │     5     10     15
+   2 │     4      9     14
+   3 │     3      8     13
+   4 │     2      7     12
+   5 │     1      6     11
+```
+"""
+function Base.reverse(df::AbstractDataFrame)
+    return df[nrow(df):-1:1, :]
+end
```
bkamins, line 2276:
```suggestion
Base.reverse(df::AbstractDataFrame) = df[nrow(df):-1:1, :]
```

test/dataframe.jl
```diff
@@ -2032,4 +2032,17 @@ end
     end
 end
 
+@testset "reverse --DataFrame" begin
```
bkamins, line 2035:
```suggestion
@testset "reverse DataFrame" begin
```

test/dataframe.jl
```diff
@@ -2032,4 +2032,17 @@ end
...
+    df = DataFrame(a = 1:5, b = 5:-1:1)
+    @test reverse(df) == DataFrame(a = 5:-1:1, b = 1:5)
+    @test reverse(DataFrame(a = 1, b = 1)) == DataFrame(a = 1, b = 1)
+end
+
+@testset "reverse --SubDataFrame" begin
```
bkamins, line 2041:
```suggestion
@testset "reverse SubDataFrame" begin
```

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2239,3 +2239,38 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
...
+
+"""
+    reverse(df::AbstractDataFrame)
+
+This function reverses the current order of the rows in `df`.
+
+# Examples
+
+```jldoctest
+julia> df = DataFrame(a = 1:5, b = 6:10, c = 11:15)
```
bkamins, line 2251: can you please remove spaces everywhere in this PR around `=` as kwarg? This is the style in Julia codes that you should not have spaces here. I have made one suggestion, but in other places the same applies.

bkamins, line 2251:
```suggestion
julia> df = DataFrame(a=1:5, b=6:10, c=11:15)
```

Chandu-4444: Sure... I was confused by looking at other places, so used spaces around `=`. Now I'll remove those.

Chandu-4444:
> Let us do `reverse!` in a separate PR and discuss it there with @nalimilan. Also have a look at https://github.com/JuliaData/DataFrames.jl/pull/2946/files to see how I think aliasing in `reverse!` should be handled (`sort!` is more general than reverse so the same issues apply)

Ok then

src/abstractdataframe/abstractdataframe.jl
bkamins: Can you x-link me a place in DataFrames.jl repo where we have spaces around `=`. We should remove them. Thank you!

test/dataframe.jl
```diff
@@ -2032,4 +2032,17 @@ end
     end
 end
 
+@testset "reverse DataFrame" begin
+    df = DataFrame(a = 1:5, b = 5:-1:1)
```
bkamins, line 2036: there are still spaces around `=`

src/abstractdataframe/abstractdataframe.jl
Chandu-4444: Umm.. Ok but, what's x-link btw?

bkamins: cross link - i.e. in which line of which file you have spotted the inconsistency. Thank you

Chandu-4444: There are quite a few places to mention:
https://github.com/JuliaData/DataFrames.jl/blob/73f11b6a0558cca2847b79223b84ab35526d5be6/src/abstractdataframe/abstractdataframe.jl#L158
https://github.com/JuliaData/DataFrames.jl/blob/73f11b6a0558cca2847b79223b84ab35526d5be6/test/dataframe.jl#L660-L682
https://github.com/JuliaData/DataFrames.jl/blob/73f11b6a0558cca2847b79223b84ab35526d5be6/src/abstractdataframe/abstractdataframe.jl#L279

These are few places I've observed. May be I'll remove those spaces around `=` for kwargs in this PR itself.

Chandu-4444:
I removed most of the spaces in the files I've worked on in this PR. This'll probably not confuse beginners like me :)
But I'm sure there were other places too where this has happened. I'll try to remove those as I work with these issues https://github.com/JuliaData/DataFrames.jl/issues/2936#issuecomment-975818873.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -759,9 +759,9 @@ Use `findall(completecases(df))` to get the indices of the rows.
 # Examples
 
 ```jldoctest
-julia> df = DataFrame(i = 1:5,
-                            x = [missing, 4, missing, 2, 1],
-                            y = [missing, missing, "c", "d", "e"])
+julia> df = DataFrame(i=1:5,
+                        x=[missing, 4, missing, 2, 1],
+                        y=[missing, missing, "c", "d", "e"])
```
bkamins, line 764:
```suggestion
                      x=[missing, 4, missing, 2, 1],
                      y=[missing, missing, "c", "d", "e"])
```

test/dataframe.jl
```diff
@@ -657,28 +657,28 @@ end
...
-    describe_output = DataFrame(variable = [:number, :number_missing, :string,
+    df = DataFrame(number=[1, 2, 3, 4],
+                   number_missing=[1, 2, 3, missing],
+                   string=["a", "b", "c", "d"],
+                   string_missing=["a", "b", "c", missing],
+                   dates=Date.([2000, 2001, 2003, 2004]),
+                   catarray=CategoricalArray([1, 2, 1, 2]))
+
+    describe_output = DataFrame(variable=[:number, :number_missing, :string,
                                             :string_missing, :dates, :catarray],
```
bkamins, line 668:
```suggestion
    describe_output = DataFrame(variable=[:number, :number_missing, :string,
                                          :string_missing, :dates, :catarray],
```

test/dataframe.jl
```diff
@@ -657,28 +657,28 @@ end
...
+                                q25=[1.75, 1.5, nothing, nothing, nothing, nothing],
+                                median=[2.5, 2.0, nothing, nothing, VERSION >= v"1.7.0-beta1.2" ? Date(2002) : nothing, nothing],
+                                q75=[3.25, 2.5, nothing, nothing, nothing, nothing],
+                                max=[4.0, 3.0, "d", "c", Date(2004), 2],
+                                nunique=[nothing, nothing, 4, 3, 4, 2],
+                                nmissing=[0, 1, 0, 1, 0, 0],
+                                first=[1, 1, "a", "a", Date(2000), 1],
+                                last=[4, missing, "d", missing, Date(2004), 2],
+                                eltype=[Int, Union{Missing, Int}, String,
                                           Union{Missing, String}, Date, CategoricalValue{Int, UInt32}])
```
bkamins, line 682:
```suggestion
                                eltype=[Int, Union{Missing, Int}, String,
                                        Union{Missing, String}, Date, CategoricalValue{Int, UInt32}])
```

test/dataframe.jl
```diff
@@ -698,7 +698,7 @@ end
         describe(df, :detailed)
 
     # Test that it works on a custom function
-    describe_output.test_std = describe_output.std
+    describe_output.test_std=describe_output.std
```
bkamins, line 701: we need space here - this is not a keyword argument of a function:
```suggestion
    describe_output.test_std = describe_output.std
```

bkamins:

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2239,3 +2239,36 @@ Base.getindex(::AbstractDataFrame, ::Union{Symbol, Integer, AbstractString}) =
 
 Base.setindex!(::AbstractDataFrame, ::Any, ::Union{Symbol, Integer, AbstractString}) =
     throw(ArgumentError("syntax df[column] is not supported use df[!, column] instead"))
+
+"""
+    reverse(df::AbstractDataFrame)
+
+This function reverses the current order of the rows in `df`.
```
nalimilan, line 2246: Better use a similar formulation as `sort`:
```suggestion
Return a data frame containing the rows in `df` in reversed order.
```

test/dataframe.jl
```diff
@@ -2032,4 +2032,17 @@ end
...
 
+@testset "reverse DataFrame" begin
+    df = DataFrame(a=1:5, b=5:-1:1)
+    @test reverse(df) == DataFrame(a=5:-1:1, b=1:5)
+    @test reverse(DataFrame(a=1, b=1)) == DataFrame(a=1, b=1)
+end
+
+@testset "reverse SubDataFrame" begin
+    df = DataFrame(a=1:10, b=10:-1:1, c=11:20)
+    @test reverse(view(df, 1:3, 1:2)) == DataFrame(a=[3, 2, 1], b=[8, 9, 10])
```
nalimilan, line 2043: Also check that the result is a `DataFrame` (and not a `SubDataFrame`) as @bkamins noted.

src/abstractdataframe/abstractdataframe.jl
nalimilan: Better do this separately I'd say, to keep commits clean.

bkamins: @Chandu-4444 - I agree with @nalimilan in general (and please use separate PRs for clean up tasks).

@nalimilan - however, since @Chandu-4444 already did the job, I am OK to accept it in this PR.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -552,7 +552,7 @@ access missing values.
 
 # Examples
 ```jldoctest
-julia> df = DataFrame(i=1:10, x=0.1:0.1:1.0, y='a':'j');
+julia> df = DataFrame(i=1:10, x=0.1:0.1:1.0, y='a':'j')
```
bkamins, line 555:
```suggestion
julia> df = DataFrame(i=1:10, x=0.1:0.1:1.0, y='a':'j');
```

bkamins: if you recommend to remove `;` here you need to add output - but it would be better done in another PR as @nalimilan commented.

bkamins (approved):
Looks good. Thank you! Let us wait for @nalimilan to approve

bkamins:
@Chandu-4444 - thank you. I hope you enjoyed it (although as you can see even simple PRs are hard, therefore I recommend starting with such rather than hard PRs which often turn out super hard 😄).

See in https://github.com/JuliaData/DataFrames.jl/pull/2946 what things were uncovered when I was thinking about `reverse!`.

Chandu-4444:
> @Chandu-4444 - thank you. I hope you enjoyed it (although as you can see even simple PRs are hard, therefore I recommend starting with such rather than hard PRs which often turn out super hard 😄).
> 
> See in https://github.com/JuliaData/DataFrames.jl/pull/2946 what things were uncovered when I was thinking about `reverse!`.

Yeah, now I get it. Thanks a lot!

PR #2946, opened 2021-11-23
Define sort! for AbstractDataFrame and fix issues of kwargs in sorting functions

bkamins:
@nalimilan - do you see any risks in making `sort!` more flexible and allow `SubDataFrame`?
(I will add tests if we agree on the design)

I have also proposed to be more careful and error if columns alias but are not identical, but we might decide not to add this extra check.


bkamins:
@nalimilan - I have added tests. It should be good for a review

In general what I propose is I think better. Already the previous method has quadratic complexity in the number of columns, but I think it is safer to error immediately when unsafe aliasing is used.

nalimilan:
Looks good!

> In general what I propose is I think better. Already the previous method has quadratic complexity in the number of columns, but I think it is safer to error immediately when unsafe aliasing is used.

I just wonder whether we should optimistically check and permute each column, and undo the changes if needed (as it should be super rare). But maybe that wouldn't make a difference for performance.

nalimilan (approved):

bkamins:
> But maybe that wouldn't make a difference for performance.

This is the point. It would only improve performance if we get an error which should be super rare.

nalimilan:
> This is the point. It would only improve performance if we get an error which should be super rare.

Wait, isn't that the contrary? It could be faster when no error happens, but it would be slower if it does because we would need to restore already permuted columns to their original state. But I agree the gain should be small anyway.

bkamins:
> undo the changes if needed

But how would you detect you need to undo this operation. The cost of doing this detection is quadratic in number of columns. So if we want to detect it we can do this detection immediately (and this is what I do now). Additionally we need to detect exact aliases and avoid permuting them (this was the behavior that we already had).

A different situation is e.g. in `push!` where indeed we can cheaply detect a problem post factum (as we can just check if column lengths are correct). However, doing `sort!` only shuffles values so we have no easy way to detect the problem.

nalimilan:
I just meant we could have a single loop over columns with the contents of the two loops you have now, and in case an error happens we would roll back any already applied changes. But forget that, it's probably not worth it.

bkamins:
I ended up having to standardize everything. Now all kwargs, following the rules we set in the 1.0 release have to be either scalars or vectors (tuples are not allowed - as it was announced we will not allow tuples). I have also improved docstrings, test coverage, and error checking.

src/abstractdataframe/sort.jl
```diff
@@ -14,6 +14,24 @@
 #                  which allows a user to specify column specific orderings
 #                  with "order(column, rev=true, ...)"
 
+function _check_sort_args(lt, by, rev, order)
```
nalimilan, line 17: Why not put this in function signatures instead? People should be used to the kind of `MethodError` that is printed. If we start checking the type of all arguments manually the codebase is going to get quite large. :-)

src/abstractdataframe/sort.jl
```diff
@@ -333,31 +351,45 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
 ## Actual sort functions
 ########################
 
+const SORT_ARGUMENTS =
 """
-    issorted(df::AbstractDataFrame, cols=All();
-             lt=isless, by=identity, rev::Bool=false, order::Ordering=Forward)
+`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR). If
+`cols` selects no columns, check whether `df` is sorted on all columns (this
+behaviour is deprecated and will change in future versions).
```
nalimilan, line 358: This text (that we just added) needs to be adapted a bit depending on the function.

src/abstractdataframe/sort.jl
```diff
@@ -333,31 +351,45 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
...
+The `by` keyword lets you provide a function that will be applied to each
+element before comparison; the `lt` keyword allows providing a custom "less
+than" function. If both `by` and `lt` are specified, the `lt` function is
+applied to the result of the `by` function.
 
-`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
+All the keyword arguments can be passed as a single value, which is applied to
+all columns or as a vector of length equal to the number of columns that the
+operation is performed on. In such a case the passed values are applied to
+consecutive columns.
```
nalimilan, line 372:
```suggestion
All the keyword arguments can be either a single value, which is applied to
all columns, or a vector of length equal to the number of columns that the
operation is performed on. In such a case each entry is used for the
column in the corresponding position in `cols`.
```

src/abstractdataframe/sort.jl
```diff
@@ -333,31 +351,45 @@ Sort.defalg(df::AbstractDataFrame, o::Ordering; alg=nothing, cols=[]) =
...
-Checking against multiple columns is done lexicographically.
+If `rev` is `true`, reverse sorting is performed. To enable reverse sorting only
+for some columns, pass `order(c, rev=true)` in `cols`, with `c` the
+corresponding column index (see example below).
 
-`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
-If `cols` selects no columns, check whether `df` is sorted on all columns
-(this behaviour is deprecated and will change in future versions).
+The `by` keyword lets you provide a function that will be applied to each
+element before comparison; the `lt` keyword allows providing a custom "less
```
nalimilan, line 365:
```suggestion
The `by` keyword allows providing a function that will be applied to each
cell before comparison; the `lt` keyword allows providing a custom "less
```

src/abstractdataframe/sort.jl
bkamins: Totally agreed. For some reason I started copying the old design. Now all kwargs have proper type restrictions.

src/abstractdataframe/sort.jl
bkamins: right - fixed

src/abstractdataframe/sort.jl
```diff
@@ -378,45 +392,54 @@ true
...
+         order::Union{Ordering, AbstractVector{<:Ordering}}=Forward,
+         view::Bool=false)
 
 Return a data frame containing the rows in `df` sorted by column(s) `cols`.
 Sorting on multiple columns is done lexicographically.
 
-`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
-If `cols` selects no columns, sort `df` on all columns
+`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR). If
+`cols` selects no columns, return permutation vector based on sorting all columns
```
nalimilan, line 431:
```suggestion
`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
 If `cols` selects no columns, sort `df` on all columns
```

src/abstractdataframe/sort.jl
```diff
@@ -470,35 +493,39 @@ julia> sort(df, [:x, order(:y, rev=true)])
...
 """
-@inline function Base.sort(df::AbstractDataFrame, cols=All(); alg=nothing, lt=isless,
-                           by=identity, rev=false, order=Forward, view::Bool=false)
+@inline function Base.sort(df::AbstractDataFrame, cols=All();
+                          alg::Union{Algorithm, Nothing}=nothing,
+                          lt::Union{Function, AbstractVector{<:Function}}=isless,
+                          by::Union{Function, AbstractVector{<:Function}}=identity,
+                          rev::Union{Bool, AbstractVector{Bool}}=false,
+                          order::Union{Ordering, AbstractVector{<:Ordering}}=Forward,
+                          view::Bool=false)
```
nalimilan, line 502:
```suggestion
                           alg::Union{Algorithm, Nothing}=nothing,
                           lt::Union{Function, AbstractVector{<:Function}}=isless,
                           by::Union{Function, AbstractVector{<:Function}}=identity,
                           rev::Union{Bool, AbstractVector{Bool}}=false,
                           order::Union{Ordering, AbstractVector{<:Ordering}}=Forward,
                           view::Bool=false)
```

src/abstractdataframe/sort.jl
```diff
@@ -470,35 +493,39 @@ julia> sort(df, [:x, order(:y, rev=true)])
...
 Return a permutation vector of row indices of data frame `df` that puts them in
 sorted order according to column(s) `cols`.
 Order on multiple columns is computed lexicographically.
 
-`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
-If `cols` selects no columns, return permutation vector based on sorting all columns
-(this behaviour is deprecated and will change in future versions).
+`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR). If
+`cols` selects no columns, check whether `df` is sorted on all columns (this
+behaviour is deprecated and will change in future versions).
```
nalimilan, line 521:
```suggestion
`cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
If `cols` selects no columns, return permutation vector based on sorting all columns
(this behaviour is deprecated and will change in future versions).
```

src/abstractdataframe/sort.jl
```diff
@@ -378,45 +392,54 @@ true
 ```
 """
 function Base.issorted(df::AbstractDataFrame, cols=All();
-                       lt=isless, by=identity, rev=false, order=Forward)
+                       lt::Union{Function, AbstractVector{<:Function}}=isless,
+                       by::Union{Function, AbstractVector{<:Function}}=identity,
+                       rev::Union{Bool, AbstractVector{Bool}}=false,
+                       order::Union{Ordering, AbstractVector{<:Ordering}}=Forward,
+)
```
nalimilan, line 399:
```suggestion
                       order::Union{Ordering, AbstractVector{<:Ordering}}=Forward)
```

bkamins:
@nalimilan - let me know when you think it is OK and I will do a final check and merge (there is so much copy-paste in this PR that I want to double check everything before merging).

src/abstractdataframe/sort.jl
```diff
@@ -378,45 +392,53 @@ true
...
+         rev::Union{Bool, AbstractVector{Bool}}=false,
+         order::Union{Ordering, AbstractVector{<:Ordering}}=Forward,
+         view::Bool=false)
 
 Return a data frame containing the rows in `df` sorted by column(s) `cols`.
 Sorting on multiple columns is done lexicographically.
 
 `cols` can be any column selector ($COLUMNINDEX_STR; $MULTICOLUMNINDEX_STR).
-If `cols` selects no columns, sort `df` on all columns
+ If `cols` selects no columns, sort `df` on all columns
```
nalimilan, line 430:
```suggestion
If `cols` selects no columns, sort `df` on all columns
```

nalimilan (approved):

bkamins:
Thank you!

PR #2948, opened 2021-11-24
Make transformation docstring more precise

bkamins:


src/abstractdataframe/selection.jl
```diff
@@ -57,7 +57,9 @@ const TRANSFORMATION_COMMON_RULES =
        it must be single name (as a `Symbol` or a string).
     6. vectors or matrices containing transformations specified by the `Pair` syntax
        described in points 2 to 5
-    8. a function which will be called with a `SubDataFrame` corresponding to each group;
+    7. a function which will be called with a `SubDataFrame` corresponding to each group
+       when `GroupedDataFrame` is processed or the passed source data frame if
+       `AbstractDataFrame` is processed;
```
nalimilan, line 62:
```suggestion
       if a `GroupedDataFrame` is processed, or with the data frame itself if
       an `AbstractDataFrame` is processed;
```

docs/src/man/split_apply_combine.md
```diff
@@ -63,7 +63,9 @@ each subset of the `DataFrame`. This specification can be of the following forms
    it must be single name (as a `Symbol` or a string).
 6. vectors or matrices containing transformations specified by the `Pair` syntax
    described in points 2 to 5
-8. a function which will be called with a `SubDataFrame` corresponding to each group;
+7. a function which will be called with a `SubDataFrame` corresponding to each group
+   when `GroupedDataFrame` is processed or the passed source data frame if
+   `AbstractDataFrame` is processed;
```
nalimilan, line 68: This section only covers split-apply-combine, so the input is supposed to be a `GroupedDataFrame` AFAICT. Though I guess we'd better keep these in sync for consistency?

bkamins: Yes - this is what I think. Also in general at some point this section should be rewritten at some point not to make such a strong distinction between split-apply-combine and other transformations as currently they are the same conceptually.

bkamins, line 68:
```suggestion
   if a `GroupedDataFrame` is processed, or with the data frame itself if
   an `AbstractDataFrame` is processed;
```

nalimilan (approved):

bkamins:
Thank you!

PR #2949, opened 2021-11-24
Catch OutOfMemoryError

bkamins:
We still get `OutOfMemoryError` on 32 bit machines. I change them to `@warning` and allow tests to pass in this PR

### DIFF
--- a/test/join.jl
+++ b/test/join.jl
@@ -1513,45 +1513,164 @@ end
     df2 = DataFrame(id=[1:10^6; 10^8+1:10^8+4])
     df2.right_row = axes(df2, 1)
 
-    @test innerjoin(df1, df2, on=:id) ≅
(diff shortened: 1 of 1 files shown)


nalimilan:
How about setting an environment variable on CI only to skip these tests? I don't like the idea that failures may go unnoticed because of a limitation that affects only CI on 32-bit.

bkamins:
> I don't like the idea that failures may go unnoticed because of a limitation that affects only CI on 32-bit.

Only `OutOfMemory` failures would go unnoticed with my proposal. However, I have fixed it and now only `OutOfMemory` failures on 32 bit machine would go unnoticed (i.e. if we have `OutOfMemory` on 64-bit machine we will get an error).

nalimilan (approved):

bkamins:
Thank you!

PR #2955, opened 2021-12-02
Generate precompile statements for Julia 1.7

### LINKED ISSUE #2921: Regenerate precompile statements for 1.3 release
Preferably this should be done after Julia 1.7 release.

bkamins:
Fixes #2921

@nalimilan - I have removed the extra precompile statements for `all=true` to reduce the size of the code parsed (it slightly reduces the load time) and since it was only used in H2O tests which are dead now (I have retained some minimal things that are done when `all=true` but they are minor).


bkamins:
Compilation time of 1.3 release (first timing is with precompilation):
```
$ julia -e "@time using DataFrames"
 16.620894 seconds (1.89 M allocations: 131.628 MiB, 0.18% gc time, 0.11% compilation time)

bogum@DESKTOP-4FNINJ9 C:\WORK
$ julia -e "@time using DataFrames"
  1.108809 seconds (1.84 M allocations: 129.151 MiB, 1.34% gc time, 0.67% compilation time)

bogum@DESKTOP-4FNINJ9 C:\WORK
$ julia -e "@time using DataFrames"
  1.113932 seconds (1.84 M allocations: 129.147 MiB, 1.33% gc time, 0.64% compilation time)

bogum@DESKTOP-4FNINJ9 C:\WORK
$ julia -e "@time using DataFrames"
  1.075805 seconds (1.84 M allocations: 129.151 MiB, 1.38% gc time, 0.72% compilation time)
```
vs 1.2.2 (first timing is with precompilation):
```
$ julia -e "@time using DataFrames"
 24.227153 seconds (1.92 M allocations: 140.367 MiB, 0.14% gc time, 0.08% compilation time)

$ julia -e "@time using DataFrames"
  1.164665 seconds (1.90 M allocations: 138.797 MiB, 2.87% gc time, 0.64% compilation time)

$ julia -e "@time using DataFrames"
  1.124894 seconds (1.90 M allocations: 138.797 MiB, 3.13% gc time, 0.68% compilation time)

$ julia -e "@time using DataFrames"
  1.173578 seconds (1.90 M allocations: 138.801 MiB, 2.98% gc time, 0.65% compilation time)
```
(and I think it is worth to reduce the `all` part to get those smaller load times)

nalimilan:
You mean that statements that are only used when `all=true` make `using DataFrames` slower, even after the initial precompilation? I find it surprising (and really bad).

bkamins:
I have run the test on DataFrames 1.2.2.

With `all=true` precompilation statements:
```
$ julia -e "@time using DataFrames"
 25.981184 seconds (1.94 M allocations: 141.029 MiB, 0.13% gc time, 0.08% compilation time)
```
and now manually in DataFrames 1.2.2 I removed these statements:
```
$ julia -e "@time using DataFrames"
 20.441751 seconds (1.93 M allocations: 139.962 MiB, 0.18% gc time, 0.11% compilation time)
```

bkamins:
After precompilation the relevant thing is number of allocations (as timings are naturally somewhat variable).

In 1.2.2 with removed `all=true` precompilation statements:
```
1.90 M allocations: 137.934 MiB
```
and when keeping these statements:
```
1.91 M allocations: 139.007 MiB
```
so it is a bit better to remove them (not much but bettet)

bkamins:
I have run more tests and the difference in compilation time after first precompilation is very small, so it can be ignored. Still I would also prefer to reduce precompilation time + not have so much code in the package that is not used.

nalimilan:
OK. I guess we'll see whether H2Oai or another benchmark would benefit from adding more precompile statements in the future.

nalimilan (approved):

bkamins:
Thank you!

PR #2966, opened 2021-12-15
better disallowmissing error message

bkamins:
Improve the error message in case column conversion fails, e.g.:
```
julia> df = DataFrame(x=1:3, y=[1,2,missing])
3×2 DataFrame
 Row │ x      y       
     │ Int64  Int64?  
─────┼────────────────
   1 │     1        1 
   2 │     2        2 
   3 │     3  missing 

julia> disallowmissing(df)
ERROR: ArgumentError: Missing value found in column number 2 

julia> disallowmissing!(df)
ERROR: ArgumentError: Missing value found in column number 2
```

### DIFF
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -2060,30 +2060,22 @@ function Missings.disallowmissing(df::AbstractDataFrame,
     for i in axes(df, 2)
         x = df[!, i]
         if i in idxcols
-            local y
+            y = x
             if Missing <: eltype(x)
-                if error
-                    try
-                        y = disallowmissing(x)
-                    catch e
-                        row = findfirst(ismissing, x)
-                        if row !== nothing
+                try
+                    y = disallowmissing(x)
+                catch e
+                    row = findfirst(ismissing, x)
+                    if row !== nothing
+                        if error
                             col = _names(df)[i]
                             throw(ArgumentError("Missing value found in column " *
                                                 ":$col in row $row"))
-                        else
-                            rethrow(e)
                         end
-                    end
-                else
-                    if any(ismissing, x)
-                        y = x
                     else
-                        y = disallowmissing(x)
+                        rethrow(e)
                     end
                 end
-            else
-                y = x
             end
             push!(newcols, y === x ? copy(y) : y)
         else
--- a/src/abstractdataframe/abstractdataframe.jl
+++ b/src/abstractdataframe/abstractdataframe.jl
@@ -2061,11 +2061,14 @@ function Missings.disallowmissing(df::AbstractDataFrame,
         x = df[!, i]
         if i in idxcols
             if Missing <: eltype(x)
-                # any(ismissing, x) is cheap so we accept paying it
+                # finding missing is cheap so we accept paying it
                 # to get a better error message
-                if any(ismissing, x)
+                row = findfirst(ismissing, x)
+                if row !== nothing
                     if error
-                        throw(ArgumentError("Missing value found in column number $i"))
+                        col = _names(df)[i]
+                        throw(ArgumentError("Missing value found in column :$col " *
+                                            "in row $row"))
                     else
                         y = x
                     end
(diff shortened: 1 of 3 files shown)


bkamins:
doctests failure is unrelated. I fix it in https://github.com/JuliaData/DataFrames.jl/pull/2967.

jtrakk:
Thanks. 

I don't use column numbers much so a name would be more familiar to me.

bkamins:
Fixed. Is it better now?

jtrakk:
Yep. Cheers.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2060,10 +2060,23 @@ function Missings.disallowmissing(df::AbstractDataFrame,
     for i in axes(df, 2)
         x = df[!, i]
         if i in idxcols
-            if !error && Missing <: eltype(x) && any(ismissing, x)
-                y = x
+            if Missing <: eltype(x)
+                # finding missing is cheap so we accept paying it
+                # to get a better error message
+                row = findfirst(ismissing, x)
```
nalimilan, line 2066: Wouldn't it be more logical to do `try; y = disallowmissing(x); catch... ` and only then check whether `missing` is present? Is this check so cheap when there are no missing values in the data?

bkamins: I have thought about it. It adds around 20% to computation time:
```
julia> x = Union{Int, Missing}[1 for i in 1:10^8];

julia> @time findfirst(ismissing, x)
  0.063825 seconds (15.35 k allocations: 869.997 KiB, 23.10% compilation time)

julia> @time findfirst(ismissing, x)
  0.049497 seconds

julia> @time disallowmissing(x);
  0.210153 seconds (112.43 k allocations: 769.128 MiB, 1.87% gc time, 18.54% compilation time)

julia> @time disallowmissing(x);
  0.190172 seconds (2 allocations: 762.939 MiB, 7.27% gc time)

julia> @time disallowmissing(x);
  0.253140 seconds (2 allocations: 762.939 MiB, 32.77% gc time)
```
which I thought is acceptable (I do not assume that `disallowmissing` will be used in any hot loop - it is most likely a one time operation).

But if you prefer `try`-`catch` then I can change it as it would indeed save some time.

bkamins: OK - I have changed this.

src/abstractdataframe/abstractdataframe.jl
```diff
@@ -2060,10 +2060,30 @@ function Missings.disallowmissing(df::AbstractDataFrame,
...
+                            rethrow(e)
+                        end
+                    end
+                else
+                    if any(ismissing, x)
+                        y = x
+                    else
+                        y = disallowmissing(x)
+                    end
+                end
```
nalimilan, line 2084: Instead of this, we could just do `error || continue` in the `if row !== nothing` branch above.

bkamins: We could not do `continue` because we still need to execute line 2088, so instead we would need to write inside `if row !== nothing`:

```
if error
    col = _names(df)[i]
    throw(ArgumentError("Missing value found in column :$col in row $row"))
else
    y = x
end
```

Do you think it would be simpler to understand by the potential readers of the code?
I used this explicit logic because when I read the old condition:
```
if !error && Missing <: eltype(x) && any(ismissing, x)
```
I had to spend some time to make sure it was OK (although I think I have written it some time ago).

Therefore I would vote for leaving things as they are as it is I think easier to understand if some user wants to inspect the code.

nalimilan: OK. But I'd use `try... catch` when `!error` too.

bkamins: I have changed `disallowmissing` and `disallowmissing!` the way that I think you wanted.

nalimilan: OK. But I'd use `try... catch` when `!error` too.

nalimilan (approved):

bkamins:
Thank you!

PR #2968, opened 2021-12-15
update docs for join on (left, right) tuple

visr:
I tried to use a tuple but in DataFrames 1.3 I got this error:

ERROR: ArgumentError: Using a `Tuple{Symbol, Symbol}` or a vector containing such tuples as a value of `on` keyword argument is not supported: use `Pair{Symbol, Symbol}` instead.

So I'm assuming this option is no longer supported.

### DIFF
--- a/docs/src/man/joins.md
+++ b/docs/src/man/joins.md
@@ -125,7 +125,7 @@ julia> crossjoin(people, jobs, makeunique = true)
 ```
 
 In order to join data frames on keys which have different names in the left and right tables,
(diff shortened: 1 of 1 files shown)


bkamins (approved):
Thank you!

bkamins:
I will wait with merging this till @nalimilan approves https://github.com/JuliaData/DataFrames.jl/pull/2967 and it gets merged to make sure docs build without error.

nalimilan (approved):

bkamins:
Thank you!

